PR #1658, opened 2024-05-17
chore(weave): Retry request failures

nickpenaranda:
https://wandb.atlassian.net/browse/WB-18851

- weave should make best effort to avoid crashing user script. to that end, we'll wrap network requests made by `RemoteHTTPTraceServer` with retry logic that will continue to retry for up to 36 hours, with EBO.

Tested manually:

- [x] Retries with EBO until success or deadline
- [x] Process waits for retries to succeed or fail
- [x] Failed, retryable requests succeed when service becomes available
- [x] Retries and final failure are logged to stderr


weave/trace_server/remote_http_trace_server.py
```diff
@@ -34,6 +36,29 @@ class ServerInfoRes(BaseModel):
     (32 - 1) * 1024 * 1024
 )  # 32 MiB (real limit) - 1 MiB (buffer)
 
+REMOTE_REQUEST_RETRY_DURATION = 60 * 60 * 36  # 36 hours
+REMOTE_REQUEST_RETRY_MAX_INTERVAL = 60 * 5  # 5 minutes
+
+
+def _is_retryable_exception(e: Exception) -> bool:
+    # TODO: Identify unretriable exceptions
```
tssweeney, line 44: I think the majority of exceptions are not retriable and are actually cases of:
* too much data being sent to clickhouse
* incorrect data (example - id missmatch)
* authentication error
...

The networking issues are a smaller subset. I believe this implementation will result in continuously retrying invalid requests, and ultimately hang the user thread.

nickpenaranda: i changed this fn so it will retry any 429 or non-4xx http errors, as well as any other errors.  as long as the backend correctly returns 4xx for those conditions, this should work as intended.  Otherwise, we should fix the backend return codes.

tssweeney (approved):

PR #1659, opened 2024-05-17
chore(weave): adding tooltip for overflowing text in modified dropdown

mbarrramsey:
Jira: https://wandb.atlassian.net/browse/WB-14570

Adding a tooltip for overflowing text in the modified dropdown. This adds a tooltip if the selected option or list of options is overflowing.

Before

https://github.com/wandb/weave/assets/126013019/d2957f7e-84e4-46e5-8d14-24e19371cfae

Screenshot 2024-05-17 at 11 40 48 AM

After

https://github.com/wandb/weave/assets/126013019/e6ab66f0-17b5-4428-bde0-093150719ad7

https://github.com/wandb/weave/assets/126013019/3409d2c0-6159-4c84-93f8-17bb23fca377


weave-js/src/common/components/elements/ModifiedDropdown.less
```diff
@@ -0,0 +1,14 @@
+.ui.dropdown {
+    .text {
+      line-height: 1.2;
+    }
+    .menu .item  {
+      line-height: 1.2 !important;
```
mbarrramsey, line 6: Need this to override the list of options line height for some reason

weave-js/src/common/components/elements/ModifiedDropdown.tsx
```diff
@@ -425,7 +422,17 @@ const ModifiedDropdown: FC<ModifiedDropdownProps> = React.memo(
       ) : (
         <>{children}</>
       );
-
+    const selectedOption = propsOptions.find(opt => opt.value === value);
+    const renderTrigger = () => {
+      if (searchQuery) {
+        return '';
+      }
```
connieelee, line 429: **question:** what does this do?

weave-js/src/common/components/elements/ModifiedDropdown.tsx
```diff
@@ -460,3 +468,46 @@ const ModifiedDropdown: FC<ModifiedDropdownProps> = React.memo(
 );
 
 export default ModifiedDropdown;
+
+const OptionWithTooltip = ({text}: any) => {
```
connieelee, line 472: **suggestion:** could we avoid the `any` type here?

```suggestion
const OptionWithTooltip = ({text}: {text: string}) => {
```

connieelee (approved):

weave-js/src/common/components/elements/ModifiedDropdown.tsx
mbarrramsey: This makes the option hidden when the user is typing a search query to avoid both text showing up at once:
Screenshot 2024-05-21 at 12 12 49 PM

weave-js/src/common/components/elements/ModifiedDropdown.tsx
mbarrramsey: Adding the string type but needed to cast `as string` due to the Option type having many possible types
Screenshot 2024-05-21 at 12 14 22 PM

mbarrramsey:
Responded to comments @connieelee, let me know if there are any more questions, thanks!

connieelee:
@mbarrramsey lgtm!

PR #1662, opened 2024-05-20
fix: when there is a network error, display error panel instead of no results

jamie-rasmussen:
When we have a network error, the Operations and Objects pages show our "No objects yet" UI instead of indicating that there was an error. Network errors are common in development (fix-gov issue), and it can be confusing what has happened given that you are possibly switching between a prod and local datastore.

The traces page also has this issue, but saving that for a different PR (need to think about export button in header for example).  Also, @tssweeney could we get rid of cgDataModelHooks.ts entirely?

Misleading current behavior:
image

After this PR:
image

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectVersionsPage.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectVersionsPage.tsx
@@ -7,6 +7,7 @@ import {
 import _ from 'lodash';
 import React, {useEffect, useMemo, useState} from 'react';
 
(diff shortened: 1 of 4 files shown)


tssweeney (approved):
Awesome!

1. Yes (: We can now completely blow away cg stuff ... was planning on doing that soon.
2. Where in this PR is error non null??

jamie-rasmussen:
> 2. Where in this PR is error non null??

Not in this PR, but error is being set here:
https://github.com/wandb/weave/blob/master/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts#L99

PR #1664, opened 2024-05-20
feat(weave): frontend to allow deletion of single call

gtarpenning:
Single call deletion path. Also works for evaluations.

https://wandb.atlassian.net/browse/WB-18627

depends on: 
- https://github.com/wandb/core/pull/21555
- https://github.com/wandb/weave/pull/1673

Screenshot 2024-05-20 at 11 54 54 AM

Screenshot 2024-05-22 at 4 36 16 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
@@ -664,27 +664,6 @@ export const usePeekLocation = () => {
   }, [peekPath]);
 };
 
-export const useClosePeekOrNavigateUp = () => {
-  /*
-  If we are peeking, just close the peek drawer
-  If we are not peeking, navigate up one level, unless we are at the root
-  */
-  const history = useHistory();
-  const peekLoc = usePeekLocation();
-  const closePeek = useClosePeek();
-
-  return () => {
-    if (peekLoc != null) {
-      closePeek();
-    } else {
-      // Not peeking (fullscreen) so navigate to the parent page
-      const path = history.location.pathname.split('/');
-      path.pop();
-      history.replace(path.join('/'));
-    }
-  };
-};
-
 export const WeaveHeaderExtrasContext = createContext<{
   extras: {[key: string]: HeaderExtra};
   addExtra: (key: string, value: HeaderExtra) => void;
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
@@ -664,6 +664,27 @@ export const usePeekLocation = () => {
   }, [peekPath]);
 };
 
+export const useClosePeekOrNavigateUp = () => {
+  /*
+  If we are peeking, just close the peek drawer
+  If we are not peeking, navigate up one level, unless we are at the root
+  */
+  const history = useHistory();
+  const peekLoc = usePeekLocation();
+  const closePeek = useClosePeek();
+
+  return () => {
+    if (peekLoc != null) {
+      closePeek();
+    } else {
+      // Not peeking (fullscreen) so navigate to the parent page
+      const path = history.location.pathname.split('/');
+      path.pop();
+      history.replace(path.join('/'));
+    }
+  };
+};
+
 export const WeaveHeaderExtrasContext = createContext<{
   extras: {[key: string]: HeaderExtra};
   addExtra: (key: string, value: HeaderExtra) => void;
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
@@ -51,6 +51,7 @@ const WFDataModelFromTraceServerCallsOnlyProvider: FC = ({children}) => {
     return {
       useCall: tsWFDataModelHooks.useCall,
       useCalls: tsWFDataModelHooks.useCalls,
+      useCallsDeleteFunc: tsWFDataModelHooks.useCallsDeleteFunc,
       useOpVersion: cgWFDataModelHooks.useOpVersion,
       useOpVersions: cgWFDataModelHooks.useOpVersions,
       useObjectVersion: cgWFDataModelHooks.useObjectVersion,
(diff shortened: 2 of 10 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
```diff
@@ -417,6 +417,7 @@ export const browse3ContextGen = (
         primaryDim
       )}&secondaryDim=${encodeURIComponent(secondaryDim)}`;
     },
+    refetchCalls: () => {},
```
tssweeney, line 420: This does not belong in the Url context. I'll keep reading and suggest an alternative place

weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
```diff
@@ -664,6 +665,27 @@ export const usePeekLocation = () => {
...
+  const history = useHistory();
+  const peekLoc = usePeekLocation();
+  const closePeek = useClosePeek();
+
+  return () => {
+    if (peekLoc != null) {
+      closePeek();
+    } else {
+      // Not peeking (fullscreen) so navigate to the parent page
+      const path = history.location.pathname.split('/');
```
tssweeney, line 682: i'm not quite sure what the intention is here

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
```diff
@@ -51,6 +51,7 @@ const WFDataModelFromTraceServerCallsOnlyProvider: FC = ({children}) => {
     return {
       useCall: tsWFDataModelHooks.useCall,
       useCalls: tsWFDataModelHooks.useCalls,
+      useCallsDelete: tsWFDataModelHooks.useCallsDelete,
```
tssweeney, line 54: nit/question: `useCallsDelete` follows a slightly different pattern (it returns a function). I wonder if there is a naming convention that could imply this pattern difference?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -202,16 +203,27 @@ const useCalls = (
...
   opts?: {skip?: boolean}
-): Loadable<CallSchema[]> => {
+): LoadableWithRefetch<CallSchema[]> => {
   const getTsClient = useGetTraceServerClientContext();
   const loadingRef = useRef(false);
   const [callRes, setCallRes] =
     useState<traceServerClient.TraceCallsQueryRes | null>(null);
   const deepFilter = useDeepMemo(filter);
+
+  const [lastFetch, setLastFetch] = useState<number | null>(null);
```
tssweeney, line 213: this whole block might be cleaner if instead of relying on states to trigger hooks, we instead:
1. Create a callback called `doFetch` that does the fetch and updates the state
2. Call `doFetch` inside the current useEffect
(by this point it is a pure refactor)
3. `refetch` is then literally just the `doFetch` method (:

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -191,6 +191,9 @@ export const CallsTable: FC<{
 
   // Fetch the calls
   const calls = useCallsForQuery(entity, project, effectiveFilter);
+  if (calls.refetch) {
+    baseContext.refetchCalls = calls.refetch;
```
tssweeney, line 195: This pattern would not be ideal in situations where there are 2 calls table on the screen at the same time (as there is when you open an eval). In general, it is not good practice to assign some state like this to a global symbol. Two comments:
1. If you wanted to do a pattern like this, you would want to have a react context provider somewhere above that allows children to "register" and "unregister" refetch triggers.
2. But even that is a little odd - and you might want to reach for a reducer pattern with listeners. 

Still, all these options are a bit odd - you really want someway to "broadcast" changes.

Got to run to dinner. Will respond later

tssweeney, line 195: This pattern would not be ideal in situations where there are 2 calls table on the screen at the same time (as there is when you open an eval). In general, it is not good practice to assign some state like this to a global symbol. Two comments:
1. If you wanted to do a pattern like this, you would want to have a react context provider somewhere above that allows children to "register" and "unregister" refetch triggers.
2. But even that is a little odd - and you might want to reach for a reducer pattern with listeners. 

Still, all these options are a bit odd - you really want someway to "broadcast" changes.

Let's see... in the best case scenario, upon deletion, two events occur:
1. All components on the page which include the deleted element adjust
2. If the primary view is of the deleted element, we navigate away.

Furthermore, when we have renaming, we want similar behavior: all cases of the name change are reflected immediately. 

I can think of 4 families of solutions:
1. Don't do anything and just let the next user-invoked navigation reflect the changes.
2. Hard refresh the whole page.
3. Broadcast change events to listening components to manually reflect the change.
    * Note: this is closest to your current implementation (where "reflect" is facilitated via a refresh) 
4. Automatically update any query hooks with the updated information.
    * The challenge here is not to be overly aggressive causing full-page reloads. Might be widespread refectors
5. are there others?...

What i don't like about #3 is that we put the responsibility on every rendering component to "listen" listen for changes. This is a lot a coupling in the design.

#4 is probably the best option across the board if done correctly. How does this relate to polling?

I wonder if the same mechanism which is hypothetically triggered in a poll event should be triggered after a mutation event? Where such a mechanism is: "refresh all active queries, without re-triggering loading/empty states". Yeah, doing this correctly might require a more widespread refactor.

Ok, so options...:
1. (low-cost version of #4) I wonder if a low-cost change to `OptionalTraceServerClientContextProvider` that allows mutation hooks to refresh the client would be jarring? Might be worth a try.
2. Do what you are doing now, but with a proper context
3. Somehow have the other call hooks react to the delete event

weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
gtarpenning: This is totally a hack, but when you are in fullscreen mode (which you can navigate to directly with a link) there is nothing to close once you have deleted the call you are looking at. So we have to navigate up the tree after deletion as the current page is no longer valid...

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
gtarpenning: brilliant comment, will work through these and have a soln by tue

tssweeney (approved):
SHIP IT!

PR #1665, opened 2024-05-20
chore(weave): Use new WANDB_POD_BASE_URL var

jwlee64:
Introduces a new env var called WANDB_POD_BASE_URL
uses that url for the majority of calls needing wandb_base_url except for urls called from the frontend

internal jira
https://wandb.atlassian.net/browse/WB-18847


weave-js/vite-plugin-block-cjs.ts
```diff
@@ -266,6 +266,7 @@ const ALLOWED_CJS_MODULES = [
   'lowlight/lib/core',
   'refractor',
   'refractor/core',
+  'yet-another-react-lightbox',
```
jwlee64, line 269: fixes build

weave/wandb_file_manager.py
```diff
@@ -52,7 +52,7 @@ def _local_path_and_download_url(
     if manifest_entry is None:
         return None
     md5_hex = hashutil.b64_to_hex_id(hashutil.B64MD5(manifest_entry["digest"]))
-    base_url = weave_env.wandb_base_url()
+    base_url = weave_env.wandb_frontend_base_url()
```
jwlee64, line 55: this returns urls to download artifacts so we need the frontend url base to download from frontend. This is the most questionable change

weave/environment.py
```diff
@@ -136,14 +136,19 @@ def analytics_disabled() -> bool:
 
 
 def weave_server_url() -> str:
-    base_url = wandb_base_url()
```
jwlee64, line 139: weave server url is accessed from the frontend

weave/environment.py
```diff
@@ -136,14 +136,19 @@ def analytics_disabled() -> bool:
...
+    base_url = wandb_frontend_base_url()
     default = "https://weave.wandb.ai"
     if base_url != "https://api.wandb.ai":
         default = base_url + "/weave"
     return os.getenv("WEAVE_SERVER_URL", default)
 
 
 def wandb_base_url() -> str:
+    pod_url = os.getenv("WANDB_POD_BASE_URL", "").rstrip("/")
```
nickpenaranda, line 147: nit: i'd prefer if this were named something like `WANDB_PRIVATE_BASE_URL` -- pod is ambiguous here, and a misnomer in some cases

weave/wandb_file_manager.py
nickpenaranda: This is right for the general case.  But i think your intuition is correct, there's probably a subset of call sites that can use the private url

nickpenaranda (commented):
LGTM thou

nickpenaranda (approved):
LGTM though had a comment on the env var naming and there's some follow-up work to look at what calls `_local_path_and_download_url` to identify which paths can use the private URL.

PR #1669, opened 2024-05-21
feat: Pluggable object serialization

shawnlewis:
This enables users to register Serializers with Weave, like so:

```python
import faiss


from weave.trace import serializer


def save_instance(obj: faiss.Index, artifact, name: str) -> None:
    with artifact.writeable_file_path(f"{name}.faissindex") as write_path:
        faiss.write_index(obj, write_path)


def load_instance(
    artifact,
    name: str,
) -> faiss.Index:
    return faiss.read_index(artifact.path(f"{name}.faissindex"))


serializer.register_serializer(faiss.Index, save_instance, load_instance)
```

After the `register_serializer` call, if the user tries to save an object that has a `faiss.Index` attribute, Weave will store the full FAISS index instead of just a __repr__ string.

This replaces the prior mechanism that relied on the older weave.Type. We were only relying on that for Op saving. The new implementation is fully compatible and generates objects with the same digests.

We will also save the load_instance method as an op, and add a reference to the load op from the saved object, so that the object can be correctly deserialized in a Python runtime that does not have the serializer registered.

So the effect is that objects with custom typed attributes are fully relocatable.

This PR also improves op serialization, notably it makes it so that ops that classes refered to by ops are captured correctly.

TODO:
- [x] Fix anonymous op issues, they are now annotated with @weave.op so I need to fix digests
- [x] Consider if we want `<module>.<name>` in the encoded weave type string, or just `<name>`
- [x] Register Serializers by ID string instead of class, so we don't need to import libraries to register serializers.
  - I chose not to do this. It's complicated to handle registering superclasses.

### DIFF
--- a/weave/trace/custom_objs.py
+++ b/weave/trace/custom_objs.py
@@ -108,7 +108,9 @@ def encode_custom_obj(obj: Any) -> Optional[dict]:
     serializer.save(obj, art, "obj")
 
     # Save the load_instance function as an op, and store a reference
-    # to that op in the saved value record.
+    # to that op in the saved value record. We don't do this if what
+    # we're saving is actually an op, since that would be self-referential
+    # (the op loading code is always present, we don't need to save/load it).
     load_op_uri = None
     if serializer.id() != "Op":
         # Ensure load_instance is an op
--- a/weave/trace/custom_objs.py
+++ b/weave/trace/custom_objs.py
@@ -146,19 +146,17 @@ def decode_custom_obj(
         if not isinstance(ref, ObjectRef):
             raise ValueError(f"Expected ObjectRef, got {load_instance_op_uri}")
         wc = require_graph_client()
-        # Can return None if op load fails. I think this might happen
-        # if a library is not installed?
-        # TODO: check this. We need to raise an error in this case
-        # if a library is not installed!
         load_instance_op = wc.get(ref)
+        if load_instance_op == None:  # == to check for None or BoxedNone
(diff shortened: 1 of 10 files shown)


weave/trace/serializer.py
```diff
@@ -0,0 +1,69 @@
...
+def get_serializer_by_id(id: str) -> Optional[Serializer]:
+    for serializer in SERIALIZERS:
+        if serializer.id() == id:
+            return serializer
+    return None
+
+
+def get_serializer_for_obj(obj: Any) -> Optional[Serializer]:
+    for serializer in SERIALIZERS:
+        if isinstance(obj, serializer.target_class):
```
tssweeney, line 67: nit: might want to find the closest in MRO

weave/trace/custom_objs.py
```diff
@@ -94,34 +99,68 @@ def writeable_file_path(self, path: str) -> Generator[str, None, None]:
...
         # will not be recoverable with client.get
         return None
     art = MemTraceFilesArtifact()
-    weave_type.save_instance(obj, art, "obj")
+    serializer.save(obj, art, "obj")
+
+    # Save the load_instance function as an op, and store a reference
+    # to that op in the saved value record.
+    load_op_uri = None
+    if serializer.id() != "Op":
```
tssweeney, line 113: would be nice to comment why this exclusion is appropriate

tssweeney (approved):
Dude, this is freaking legit!

PR #1655, opened 2024-05-16
feat(weave): backend to allow deletion of calls

gtarpenning:
moved to clean branch: https://github.com/wandb/weave/pull/1673

PR #1673, opened 2024-05-22
feat(weave): backend for deletion of calls

gtarpenning:
https://wandb.atlassian.net/browse/WB-18625
https://wandb.atlassian.net/browse/WB-18626
https://wandb.atlassian.net/browse/WB-18628

Adds: 
- `calls_delete` on the client
- `call.delete()` on the call object
- `calls_delete()` server method 

depends on https://github.com/wandb/core/pull/21555 for `wb_user_id` to appear on deletion rows.

(old pr: https://github.com/wandb/weave/pull/1655)


tssweeney (approved):
👍 Nice Job!

PR #1674, opened 2024-05-22
docs: Mistral docs v2

tcapelle:
New version with better formatting, using @shawnlewis doc as ref.


tcapelle:
Can we merge this?

PR #1679, opened 2024-05-23
chore(weave): serve dev frontend if running dev trace server

gtarpenning:
https://wandb.atlassian.net/browse/WB-18776

Screenshot 2024-05-23 at 4 10 14 PM


jwlee64 (approved):

weave/environment.py
```diff
@@ -144,11 +146,18 @@ def weave_server_url() -> str:
...
 
 def wandb_base_url() -> str:
-    pod_url = os.getenv("WANDB_PRIVATE_BASE_URL", "").rstrip("/")
-    return pod_url if pod_url != "" else wandb_frontend_base_url()
+    pod_url = os.getenv("WANDB_PRIVATE_BASE_URL")
+    if pod_url:
+        return pod_url.rstrip("/")
+
+    settings = Settings()
+    return os.environ.get("WANDB_BASE_URL", settings.base_url).rstrip("/")
```
tssweeney, line 154: this divergence from `wandb_frontend_base_url` implies something is no longer consistent

weave/environment.py
```diff
@@ -144,11 +146,18 @@ def weave_server_url() -> str:
...
 
 
 def wandb_frontend_base_url() -> str:
+    if wf_trace_server_url() == "http://127.0.01:6345":
```
tssweeney, line 158: This might be true right now, but may not always be true - for example, what if you are running a local trace server against QA? i think we should just get in the habit of setting `WANDB_BASE_URL` if we want the printout to be correct.

tssweeney: in fact, this raises yet a bigger point - what are all the URLs we need to have set? Can we list those somewhere centrally? Are there common "presets"? Seems like we at least have:
* where the frontend is served
* where the gorilla is located
* where the trace_server is located

weave/environment.py
gtarpenning: I think I understand what you are referring to, which is that `WANDB_BASE_URL` is being used to set both gorilla and the frontend link. We currently just **assume** that the frontend is being served at `app + ` the `WANDB_BASE_URL` path. Which is obviously not always the case, as in local dev or qa. Thus, there is a legitimate divergence. 

I think setting the base url from the frontend url (what we are doing currently) is a bit weird right? Shouldn't these things not necessarily rely on each other?

weave/environment.py
gtarpenning: Can we add an ENV var: `WANDB_FRONTEND_URL`? 
Then:
- frontend: `WANDB_FRONTEND_URL`
- gorilla:`WANDB_BASE_URL`
- trace: `WF_TRACE_SERVER_URL`

@tssweeney

PR #1688, opened 2024-05-29
chore(app): add hide text option to ModifiedDropdown

mbarrramsey:
Jira: https://wandb.atlassian.net/browse/WB-18777

Updates the `ModifiedDropdown` component to support hiding the input text prop so that we can display a JSX label as a dropdown option on its own. This will be used in the clone/ create report modal in an upcoming PR, https://github.com/wandb/core/pull/21754.

Before: 

Screenshot 2024-05-29 at 12 50 58 PM

After:

https://github.com/wandb/core/assets/126013019/e5c35eed-c45f-4e93-bbe1-05f18f15ede6

### DIFF
--- a/weave-js/src/common/components/elements/ModifiedDropdown.tsx
+++ b/weave-js/src/common/components/elements/ModifiedDropdown.tsx
@@ -215,12 +215,9 @@ const ModifiedDropdown: FC<ModifiedDropdownProps> = React.memo(
 
     const getOptionProps = (opt: Option, hideText: boolean) => {
       const {text, ...props} = opt;
-      props.text = hideText ? (
-        {}
-      ) : (
+      props.text = hideText ? null : (
         <OptionWithTooltip text={opt.text as string} />
       );
-
       return props;
     };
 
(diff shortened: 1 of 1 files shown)


weave-js/src/common/components/elements/ModifiedDropdown.tsx
```diff
@@ -477,7 +484,7 @@ type OptionWithTooltipProps = {
   text: string | null;
 };
 
-const OptionWithTooltip: React.FC<OptionWithTooltipProps> = ({text}) => {
+export const OptionWithTooltip: React.FC<OptionWithTooltipProps> = ({text}) => {
```
mbarrramsey, line 487: To be used in https://github.com/wandb/core/pull/21754

weave-js/src/common/components/elements/ModifiedDropdown.tsx
```diff
@@ -58,6 +58,14 @@ const simpleSearch = (options: DropdownItemProps[], query: string) => {
     .value();
 };
 
+const getOptionProps = (opt: Option, hideText: boolean) => {
+  const {text, ...props} = opt;
+  props.text = hideText ? null : (
+    <OptionWithTooltip text={opt.text as string} />
+  );
+  return props;
```
mbarrramsey, line 66: Adding the option to hide the text prop so that we can still search it's value, but only display the JSX label if we'd like

onx2 (approved):

PR #1694, opened 2024-05-30
chore(weave): Remove CG Data Provider

tssweeney:
Big tech debt cleanup: removal of `weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/cgDataModelHooks.ts`. This file was useful during a transition period to trace server, but is no longer useful and just causes extra code to conform to interfaces. Let's remove it!

While doing this, I realized that we had a bug when the trace server is not connected (ie. private deployments).

Before:
Screenshot 2024-05-29 at 19 20 54

After:
Screenshot 2024-05-29 at 19 19 02


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/EmptyContent.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import {TargetBlank} from '../../../../../../common/util/links';
+import {Link, TargetBlank} from '../../../../../../common/util/links';
```
jamie-rasmussen, line 3: Looks like this Link doesn't have the standard styling, maybe you want https://github.com/wandb/weave/blob/jamie/feedback-schema/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/Links.tsx#L37

jamie-rasmussen (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/EmptyContent.tsx
tssweeney: fixed! thanks

PR #1697, opened 2024-05-30
fix(app): Removing infinite loop on modified dropdown render

jtulk:
https://wandb.atlassian.net/browse/WB-19077

Effect updating state causing infinite render. Unsure as to the deeper root cause but this fixes the error.

### DIFF
--- a/weave-js/src/common/components/elements/ModifiedDropdown.tsx
+++ b/weave-js/src/common/components/elements/ModifiedDropdown.tsx
@@ -487,7 +487,7 @@ const OptionWithTooltip: React.FC<OptionWithTooltipProps> = ({text}) => {
     }
     const isOverflow =
       optionRef.current.scrollWidth > optionRef.current.clientWidth;
-    if (optionRef.current && isOverflow !== showTooltip) {
+    if (isOverflow !== showTooltip) {
       setShowTooltip(isOverflow);
     }
   }, [showTooltip, text]);


weave-js/src/common/components/elements/ModifiedDropdown.tsx
```diff
@@ -482,12 +482,15 @@ const OptionWithTooltip: React.FC<OptionWithTooltipProps> = ({text}) => {
...
-    if (optionRef.current) {
-      setShowTooltip(
-        optionRef.current.scrollWidth > optionRef.current.clientWidth
-      );
+    if (!optionRef.current) {
+      return;
+    }
+    const isOverflow =
+      optionRef.current.scrollWidth > optionRef.current.clientWidth;
+    if (optionRef.current && isOverflow !== showTooltip) {
```
onx2, line 490: is `optionRef.current && ` required here now that you are checking above?

onx2 (approved):

weave-js/src/common/components/elements/ModifiedDropdown.tsx
```diff
@@ -482,12 +482,15 @@ const OptionWithTooltip: React.FC<OptionWithTooltipProps> = ({text}) => {
...
-      setShowTooltip(
-        optionRef.current.scrollWidth > optionRef.current.clientWidth
-      );
+    if (!optionRef.current) {
+      return;
+    }
+    const isOverflow =
+      optionRef.current.scrollWidth > optionRef.current.clientWidth;
+    if (isOverflow !== showTooltip) {
+      setShowTooltip(isOverflow);
```
onx2, line 491: could this just be `setShowTooltip(isOverflow);`? React shouldn't cause an update if it ends up the same boolean value 🤔

onx2: oh wait duh 😂 , that was the previous code. I wonder if this is because of the ref

PR #1699, opened 2024-05-30
chore(weave): backend to allow call renaming

gtarpenning:
- https://wandb.atlassian.net/browse/WB-19044
- https://wandb.atlassian.net/browse/WB-19090
- https://wandb.atlassian.net/browse/WB-19092
- https://wandb.atlassian.net/browse/WB-19165

requires using `argMax`, which is not a `simpleAggregationFunction`. 

Depends on: https://github.com/wandb/core/pull/21850


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -313,9 +329,11 @@ def calls_query_stream(
...
             limit=req.limit,
             offset=req.offset,
-            order_by=None
-            if not req.sort_by
-            else [(s.field, s.direction) for s in req.sort_by],
+            order_by=(
+                None
+                if not req.sort_by
+                else [(s.field, s.direction) for s in req.sort_by]
+            ),
```
gtarpenning, line 336: lint?

weave/trace_server/trace_server_interface.py
```diff
@@ -196,6 +200,18 @@ class CallsQueryStatsRes(BaseModel):
     count: int
 
 
+class CallRenameReq(BaseModel):
+    project_id: str
+    call_id: str
+    # wb_user_id gets generated from auth params
+    wb_user_id: typing.Optional[str] = None
```
tssweeney, line 207: When is this None? If it is only expected to be None over the wire, then you should consider a specific Req class inside the server definition itself.

gtarpenning: ah okay, this is the same as Delete so I can do both that way...

weave/weave_client.py
```diff
@@ -561,6 +572,16 @@ def delete_call(self, call: Call) -> None:
             )
         )
 
+    @trace_sentry.global_trace_sentry.watch()
+    def rename_call(self, call: Call, display_name: str) -> None:
```
tssweeney, line 576: Seeing this pattern makes me wonder, "should all operations follow this pattern?:"

* LOWEST LEVEL: `trace_server_interface` has capability directly or indirectly via other capabilities
* `remote_http_trace_server` -> exposes http interface in language specific binding (could be generated someday)
* `weave_client`: contains some `[action]_[class](self, obj: Class, params)` to do the action operating on objects
* HIGHEST LEVEL `[Class].action(params)` - OO style of action

Maybe so? I like that clarity in design.

weave/trace_server/migrations/003_add_display_name.down.sql
```diff
@@ -0,0 +1,25 @@
+ALTER TABLE calls_merged_view MODIFY QUERY
```
tssweeney, line 1: please include a header and comments on what changes to make it easier for reviewers and readers

weave/trace_server/migrations/003_add_display_name.up.sql
```diff
@@ -0,0 +1,29 @@
+ALTER TABLE call_parts
+    ADD COLUMN display_name Array(Tuple(DateTime64(3), String));
+
+ALTER TABLE calls_merged
```
tssweeney, line 4: might be nice to explain this data model choice here

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1266,6 +1296,15 @@ def _raw_call_dict_to_ch_call(
     return SelectableCHCallSchema.model_validate(call)
 
 
+def _ch_call_display_name_tuple_to_str(
```
tssweeney, line 1299: preferably this happens inside the view so that we can sort/filter on it

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -922,7 +952,7 @@ def _select_calls_query_raw(
         for col in columns:
             if col in ["project_id", "id"]:
                 merged_cols.append(f"{col} AS {col}")
-            elif col in ["input_refs", "output_refs"]:
+            elif col in ["input_refs", "output_refs", "display_name"]:
```
tssweeney, line 955: perhaps it is time to extract this list to a constant maintained near the type def?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1282,6 +1312,7 @@ def _ch_call_to_call_schema(ch_call: SelectableCHCallSchema) -> tsi.CallSchema:
         exception=ch_call.exception,
         wb_run_id=ch_call.wb_run_id,
         wb_user_id=ch_call.wb_user_id,
+        display_name=ch_call.display_name[1] if ch_call.display_name else None,
```
tssweeney, line 1315: even further, the `[1]` indexing should be done in the db layer. we want to be able to select `display_name` directly, hiding the implementation details as to how that is achieved.

weave/trace_server/trace_server_interface.py
gtarpenning: I realize this is a bit lame but im not 100% sure what you mean, or where the best place for this type to live...

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -380,14 +403,24 @@ def calls_delete(self, req: tsi.CallsDeleteReq) -> tsi.CallsDeleteRes:
...
         ]
-        self._flush_immediately = False
-        for insertable in insertables:
-            self._insert_call(insertable)
-        self._flush_calls()
-        self._flush_immediately = True
+
+        with self.call_batch():
+            for insertable in insertables:
+                self._insert_call(insertable)
```
gtarpenning, line 409: how kosher is it to include small unrelated tweaks?

weave/trace_server/migrations/003_add_display_name.down.sql
```diff
@@ -0,0 +1,31 @@
+/*
```
tssweeney, line 1: I think you are going to need to bump the migration numbers!

gtarpenning: oops was slightly stale @tssweeney

weave/trace_server/trace_server_interface.py
tssweeney: ok let's chat... basically:

We should have `class CallRenameReq(BaseModel):` here require the `wb_user_id` and the trace server itself should have a class with such field omitted

tssweeney: honestly, it is a small nit, so maybe not worth it. In Jamie's `https://github.com/wandb/weave/pull/1718/files#diff-2628396670e408d29134ac19cc8f01d7d12e9e60821a514e02b0139a03e05d8c`, the id is required actually

gtarpenning: Right, but doesn't `call_rename` have to be identical in: 
- `remote_http_trace_server.py`
- `clickhouse_trace_server_batched.py`
- `TraceServerInterface` in `trace_server_interface.py`?

Over the wire we _don't_ need `wb_run_id`, but in the other places we do. @tssweeney

weave/trace_server/trace_server_interface.py
```diff
@@ -356,17 +360,17 @@ def call_read(self, req: CallReadReq) -> CallReadRes:
...
+    # def calls_delete(self, req: CallsDeleteReq) -> CallsDeleteRes:
+    #     ...
 
     @abc.abstractmethod
     def calls_query_stats(self, req: CallsQueryStatsReq) -> CallsQueryStatsRes:
         ...
 
-    @abc.abstractmethod
-    def call_rename(self, req: CallRenameReq) -> CallRenameRes:
-        ...
```
gtarpenning: this is what I was talking about

tssweeney (approved):

PR #1709, opened 2024-05-31
perf: Fix full table scans for objects, files, tables

shawnlewis:
We were scanning the whole clickhouse tables for object, file, and table reads, all the time.

Fix so that we only scan the subset of those tables that we need for the specific project referenced in the user query.

This should speed up W&B production reads from about 3s for object_versions, down to just 20ms in many cases (2 orders of magnitude faster). We should get similar improvements for files and table reads. as well. 

```
# There are four major kinds of things that we query:
# - calls,
# - object_versions,
# - tables
# - files
#
# calls are identified by ID.
#
# object_versions, tables, and files are identified by digest. For these kinds of
# things, we dedupe at merge time using Clickhouse's ReplacingMergeTree, but we also
# need to dedupe at query time.
#
# Previously, we did query time deduping in *_deduped VIEWs. But it turns out
# clickhouse doesn't push down the project_id predicate into those views, so we were
# always scanning whole tables.
#
# Now, we've just written the what were views before into this file directly as
# subqueries, and put the project_id predicate in the innermost subquery, which fixes
# the problem.
```

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -1,5 +1,27 @@
 # Clickhouse Trace Server
 
+# A note on query structure:
+# There are four major kinds of things that we query:
+# - calls,
+# - object_versions,
+# - tables
+# - files
+#
+# calls are identified by ID.
+#
+# object_versions, tables, and files are identified by digest. For these kinds of
+# things, we dedupe at merge time using Clickhouse's ReplacingMergeTree, but we also
+# need to dedupe at query time.
+#
+# Previously, we did query time deduping in *_deduped VIEWs. But it turns out
+# clickhouse doesn't push down the project_id predicate into those views, so we were
+# always scanning whole tables.
+#
+# Now, we've just written the what were views before into this file directly as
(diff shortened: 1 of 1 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -806,8 +853,21 @@ def file_create(self, req: tsi.FileCreateReq) -> tsi.FileCreateRes:
         return tsi.FileCreateRes(digest=digest)
 
     def file_content_read(self, req: tsi.FileContentReadReq) -> tsi.FileContentReadRes:
+        # Must dedupe files
```
tssweeney, line 856: still a todo?

shawnlewis: No, sorry those are comments that are intended to say that we're performing deduping in the query. I could clarify those.

tssweeney: oh, no it is just a note as to why the query is so wild

shawnlewis: The comments are updated to be more clear now.

tssweeney (approved):
Approved - functionally sound testing locally

PR #1711, opened 2024-06-01
chore: make it clearer that unabridged text is copied in use tabs

jamie-rasmussen:
Internal Jira: https://wandb.atlassian.net/browse/WB-19087

Multiple people have reported not realizing that the `<ref_uri>` placeholder  in code examples is replaced with the full ref string in the copied text.

This updates replaces that placeholder with an abridged version of the ref string and updates the tooltip content to indicate that the unabridged string is copied.

Before:
Screenshot 2024-05-31 at 10 11 13 PM

After:
Screenshot 2024-05-31 at 10 11 01 PM

### DIFF
--- a/weave-js/src/util/refs.ts
+++ b/weave-js/src/util/refs.ts
@@ -1,0 +1,28 @@
+const PROTOCOL = 'weave://';
+
+export const makeRefCall = (
+  entity: string,
+  project: string,
+  callId: string
+): string => {
+  return `${PROTOCOL}/${entity}/${project}/call/${callId}`;
+};
+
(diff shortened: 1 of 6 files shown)


weave-js/src/util/refs.ts
```diff
@@ -0,0 +1,28 @@
+const PROTOCOL = 'weave://';
+
+export const makeRefCall = (
```
jamie-rasmussen, line 3: These two small utility functions aren't used in this PR but will be in upcoming feedback UI code.

jlzhao27 (approved):

PR #1712, opened 2024-06-03
chore(weave): Implement Predicate Pushdown for Calls

tssweeney:
This PR implements predicate pushdown for the primary filter clause (not yet implemented for the dynamic mongo queries). It was raised that our filter conditions are being applied after grouping, which is certainly not as efficient as applying the filter pre-grouped (since applying after grouping requires operating on far more rows than needed). In this PR we do the following:
1. Refactor many symbols/params from `conditions` to `having_conditions` to specifically mean the conditions applied after the group by (in the `having` clause of the query)
2. Modify the query builder functions to accept `start_event_conditions` and `end_event_conditions` which can be used to apply filters _before_ the aggregation. If these conditions are supplied, an inner query is executed which generates an `id IN (...)` condition that is applied in the `WHERE` clause.
3. Modify the filter builder to produce these more specific conditions when appropriate

Testing:
* Current filter tests cover all sorts of cases here
* Also tested in the UI manually

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -1741,7 +1741,7 @@ def _process_calls_filter_to_conditions(
 ) -> FilterToConditions:
     """Converts a CallsFilter to a list of conditions for a clickhouse query."""
     param_builder = param_builder or ParamBuilder()
-    conditions = []
+    having_conditions = []
     raw_fields_used = set()
 
     if filter.op_names:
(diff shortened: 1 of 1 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -350,20 +362,27 @@ def calls_delete(self, req: tsi.CallsDeleteReq) -> tsi.CallsDeleteRes:
...
-        proj_cond = "project_id = {project_id: String}"
+        # Note: i think this project condition is redundant
+        proj_cond = "calls_merged.project_id = {project_id: String}"
         proj_params = {"project_id": req.project_id}
 
         # get all parents
         parents = self._select_calls_query(
             req.project_id,
-            conditions=[proj_cond, "id IN {ids: Array(String)}"],
+            start_event_conditions=[
```
tssweeney, line 372: These changes should make deletion a bit faster

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -992,10 +1017,15 @@ def _select_calls_query_raw(
...
 
-        if not conditions:
-            conditions = ["1 = 1"]
+        if not having_conditions:
+            having_conditions = ["1 = 1"]
 
-        conditions_part = _combine_conditions(conditions, "AND")
+        having_conditions_part = _combine_conditions(having_conditions, "AND")
+
+        where_conditions_part = _make_calls_where_condition_from_event_conditions(
```
tssweeney, line 1025: Here is one of the new clauses

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1082,10 +1115,15 @@ def _calls_query_stats_raw(
...
```
tssweeney, line 1123: Another new clause

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1706,13 +1745,22 @@ def _transform_external_field_to_internal_field(
...
+    start_event_conditions: list[str]
+    end_event_conditions: list[str]
+    fields_used: set[str]
+
+
 def _process_calls_filter_to_conditions(
     filter: tsi._CallsFilter,
     param_builder: typing.Optional[ParamBuilder] = None,
-) -> tuple[list[str], set[str]]:
+) -> FilterToConditions:
```
tssweeney, line 1758: This method has been refactored to return specific conditions where possible, using correct namespacing

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1898,3 +1952,31 @@ def process_operand(operand: tsi_query.Operand) -> str:
     conditions.append(filter_cond)
 
     return conditions, raw_fields_used
+
+
+def _make_calls_where_condition_from_event_conditions(
```
tssweeney, line 1957: create an aggregate filter

shawnlewis (approved):

PR #1718, opened 2024-06-04
feat: add feedback trace server APIs

jamie-rasmussen:
Internal Jira: https://wandb.atlassian.net/browse/WB-18697

Adds Trace Server APIs for create / query / purge. These APIs are based on a new "ORM" layer that helps us have more similar code between ClickHouse and SQLite.

Emoji utilities were previously reviewed in core, bringing them over to this repo to correspond to change in where validation logic lives.

@tssweeney - there is obviously a bunch of code duplication here with the mongo processing in Calls that we should figure out what to do about.

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -45,7 +45,11 @@ from . import environment as wf_env
 from . import clickhouse_trace_server_migrator as wf_migrator
 from .errors import InvalidRequest, RequestTooLarge
 from .emoji_util import detone_emojis
-from .feedback import TABLE_FEEDBACK, validate_feedback_create_req
+from .feedback import (
+    TABLE_FEEDBACK,
+    validate_feedback_create_req,
+    validate_feedback_purge_req,
+)
 from .orm import Table, Column, ParamBuilder, Row
 
 
@@ -969,6 +973,7 @@ class ClickHouseTraceServer(tsi.TraceServerInterfacePostAuth):
         #       should we select matching ids, and then DELETE FROM WHERE id IN (...)?
         #       This would allow us to return the number of rows deleted, and complain
         #       if too many things would be deleted.
+        validate_feedback_purge_req(req)
         query = TABLE_FEEDBACK.purge()
         query = query.project_id(req.project_id)
         query = query.where(req.query)
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -31,6 +31,7 @@ import typing
 import hashlib
 import dataclasses
 import logging
+from zoneinfo import ZoneInfo
 
 from clickhouse_connect.driver.client import Client as CHClient
 from clickhouse_connect.driver.query import QueryResult, StreamContext
@@ -42,6 +43,7 @@ from pydantic import BaseModel
 from . import environment as wf_env
 from . import clickhouse_trace_server_migrator as wf_migrator
 from .errors import RequestTooLarge
+from .orm import Table, Column, ParamBuilder, Row
 
 from .trace_server_interface_util import (
     extract_refs_from_values,
@@ -64,6 +66,21 @@ MAX_FLUSH_AGE = 15
 FILE_CHUNK_SIZE = 100000
 
 MAX_DELETE_CALLS_COUNT = 100
+
+
+TABLE_FEEDBACK = Table(
+    "feedback",
+    [
+        Column("id", "string"),
+        Column("project_id", "string"),
+        Column("weave_ref", "string"),
+        Column("wb_user_id", "string"),
+        Column("creator", "string", nullable=True),
+        Column("created_at", "datetime"),
+        Column("feedback_type", "string"),
+        Column("payload", "json", db_name="payload_dump"),
+    ],
+)
 
 
 class NotFoundError(Exception):
@@ -904,6 +921,52 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
             raise ValueError("Missing chunks")
         return tsi.FileContentReadRes(content=b"".join(chunks))
 
+    def feedback_create(
+        self, req: tsi.FeedbackCreateReqForInsert
+    ) -> tsi.FeedbackCreateRes:
+        feedback_id = generate_id()
+        created_at = datetime.datetime.now(ZoneInfo("UTC"))
+        # TODO: Any validation on weave_ref?
+        # TODO: What should max payload size be?
+        payload = _dict_value_to_dump(req.payload)
+        if len(payload) > 1024:
+            raise ValueError("Feedback payload too large")
+        row: Row = {
+            "id": feedback_id,
+            "project_id": req.project_id,
+            "weave_ref": req.weave_ref,
+            "wb_user_id": req.wb_user_id,
+            "creator": req.creator,
+            "feedback_type": req.feedback_type,
+            "payload": payload,
+            "created_at": created_at,
+        }
+        TABLE_FEEDBACK.insert(row).execute(self.ch_client)
+        return tsi.FeedbackCreateRes(
+            id=feedback_id, created_at=created_at, wb_user_id=req.wb_user_id
+        )
+
+    def feedback_query(self, req: tsi.FeedbackQueryReq) -> tsi.FeedbackQueryRes:
+        query = TABLE_FEEDBACK.select()
+        query = query.project_id(req.project_id)
+        query = query.fields(req.fields)
+        query = query.where(req.query)
+        query = query.order_by(req.sort_by)
+        query = query.limit(req.limit).offset(req.offset)
+        result = query.execute(self.ch_client)
+        return tsi.FeedbackQueryRes(result=result)
+
+    def feedback_purge(self, req: tsi.FeedbackPurgeReq) -> tsi.FeedbackPurgeRes:
+        # TODO: Instead of passing conditions to DELETE FROM,
+        #       should we select matching ids, and then DELETE FROM WHERE id IN (...)?
+        #       This would allow us to return the number of rows deleted, and complain
+        #       if too many things would be deleted.
+        query = TABLE_FEEDBACK.purge()
+        query = query.project_id(req.project_id)
+        query = query.where(req.query)
+        result = query.execute(self.ch_client)
+        return tsi.FeedbackPurgeRes(result=result)
+
     # Private Methods
     @property
     def ch_client(self) -> CHClient:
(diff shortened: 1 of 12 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -904,6 +921,52 @@ def file_content_read(self, req: tsi.FileContentReadReq) -> tsi.FileContentReadR
...
+        query = TABLE_FEEDBACK.select()
+        query = query.project_id(req.project_id)
+        query = query.fields(req.fields)
+        query = query.where(req.query)
+        query = query.order_by(req.sort_by)
+        query = query.limit(req.limit).offset(req.offset)
+        result = query.execute(self.ch_client)
+        return tsi.FeedbackQueryRes(result=result)
+
+    def feedback_purge(self, req: tsi.FeedbackPurgeReq) -> tsi.FeedbackPurgeRes:
```
tssweeney, line 959: This feels extremely risky (a user would be able to purge an entire project's feedback in 1 call?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -904,6 +921,52 @@ def file_content_read(self, req: tsi.FileContentReadReq) -> tsi.FileContentReadR
...
 
+    def feedback_create(
+        self, req: tsi.FeedbackCreateReqForInsert
+    ) -> tsi.FeedbackCreateRes:
+        feedback_id = generate_id()
+        created_at = datetime.datetime.now(ZoneInfo("UTC"))
+        # TODO: Any validation on weave_ref?
+        # TODO: What should max payload size be?
+        payload = _dict_value_to_dump(req.payload)
+        if len(payload) > 1024:
```
tssweeney, line 932: is `1024` pegged to something, or just a safeguard and more-or-less arbitrary?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -904,6 +921,52 @@ def file_content_read(self, req: tsi.FileContentReadReq) -> tsi.FileContentReadR
             raise ValueError("Missing chunks")
         return tsi.FileContentReadRes(content=b"".join(chunks))
 
+    def feedback_create(
+        self, req: tsi.FeedbackCreateReqForInsert
+    ) -> tsi.FeedbackCreateRes:
+        feedback_id = generate_id()
+        created_at = datetime.datetime.now(ZoneInfo("UTC"))
+        # TODO: Any validation on weave_ref?
+        # TODO: What should max payload size be?
```
tssweeney, line 930: 1024 seems like a pretty sufficient cap for now - couple paragraphs in note form

weave/trace_server/orm.py
```diff
@@ -0,0 +1,561 @@
+"""
```
tssweeney, line 1: <3 loving this!

weave/trace_server/clickhouse_trace_server_batched.py
jamie-rasmussen: That was a concern for me too. Batch deletion seems useful, but we could make it so we only allow purging a list of ids. We could also keep this as is but make the Python API more restrictive.

weave/trace_server/clickhouse_trace_server_batched.py
jamie-rasmussen: I picked a number rather arbitrarily - small because it is easier to loosen than tighten, but I'm open to alternatives.

weave/trace_server/trace_server_interface.py
```diff
@@ -318,6 +319,75 @@ class FileContentReadRes(BaseModel):
...
+
+class Feedback(FeedbackCreateReqForInsert):
+    id: str
+    created_at: datetime.datetime
+
+
+class FeedbackQueryReq(BaseModel):
+    project_id: str = Field(examples=["entity/project"])
+    fields: typing.Optional[list[str]] = Field(
+        default=None, examples=[["id", "feedback_type", "payload.note", "count(*)"]]
```
tssweeney, line 365: interesting idea to combine `count(*)` in the field selection. I'm curious to read more below how this is handled - both from a sql injection perspective but also if you have count & fields, then you need 2 queries

weave/trace_server/trace_server_interface.py
```diff
@@ -318,6 +319,75 @@ class FileContentReadRes(BaseModel):
...
+            }
+        ]
+    )
+
+
+class FeedbackCreateReqForInsert(FeedbackCreateReq):
+    wb_user_id: str
+
+
+# The response provides the additional fields needed to convert a request
```
tssweeney, line 348: what do you mean by this? Is there a back-and-forth here?

weave/trace_server/trace_server_interface.py
```diff
@@ -318,6 +319,75 @@ class FileContentReadRes(BaseModel):
...
+    # TODO: Might be nice to have shortcut for single field and implied ASC direction
+    # TODO: I think _SortBy shouldn't have leading underscore
+    sort_by: typing.Optional[typing.List[_SortBy]] = None
+    limit: typing.Optional[int] = Field(default=None, examples=[10])
+    offset: typing.Optional[int] = Field(default=None, examples=[0])
+
+
+class FeedbackQueryRes(BaseModel):
+    # Note: this is not a list of Feedback because user can request any fields.
+    result: list[dict[str, typing.Any]]
```
tssweeney, line 378: too bad we can't have Partials

weave/trace_server/trace_server_interface.py
```diff
@@ -318,6 +319,75 @@ class FileContentReadRes(BaseModel):
...
+    # TODO: I think I would prefer to call this order_by to match SQL, but this is what calls API uses
+    # TODO: Might be nice to have shortcut for single field and implied ASC direction
+    # TODO: I think _SortBy shouldn't have leading underscore
+    sort_by: typing.Optional[typing.List[_SortBy]] = None
+    limit: typing.Optional[int] = Field(default=None, examples=[10])
+    offset: typing.Optional[int] = Field(default=None, examples=[0])
+
+
+class FeedbackQueryRes(BaseModel):
+    # Note: this is not a list of Feedback because user can request any fields.
```
tssweeney, line 377: interesting that we support cases where the fields are not present. alternative idea for the "fields" in the request to only control payload fields, which would result in a list of value feedbacks.

weave/trace_server/trace_server_interface.py
jamie-rasmussen: I haven't added `GROUP BY` or `HAVING` to ORM yet. Kind of hacked in `count(*)` because it was useful for tests. Re: injection, I'm only allowing that specific string, we'd have to get fancier if we wanted to support functions more generally.

weave/trace_server/trace_server_interface.py
```diff
@@ -318,6 +319,75 @@ class FileContentReadRes(BaseModel):
...
+    result: list[dict[str, typing.Any]]
+
+
+class FeedbackPurgeReq(BaseModel):
+    project_id: str = Field(examples=["entity/project"])
+    # TODO: Should we make this required?
+    query: typing.Optional[Query] = None
```
tssweeney, line 384: I still don't like this ... feels very easy to mistakenly delete everything.

weave/trace_server/orm.py
```diff
@@ -0,0 +1,621 @@
...
+        self.name = name
+        self.db_name = db_name
+        self.type = type
+        self.nullable = nullable
+
+    def dbname(self) -> str:
+        return self.db_name or self.name
+
+    def create_sql(self) -> str:
+        sql_type = "TEXT"
```
tssweeney, line 125: should this be `self.type`?

weave/trace_server/trace_server_interface.py
jamie-rasmussen: Opinion on how I should mitigate that? Add a limit on how many can rows be deleted at once? Change the API to only allow deletion by ID? Keep it as a (required) Query parameter but raise if it is too generic (by some heuristic)?

weave/trace_server/trace_server_interface.py
jamie-rasmussen: The create request has a "partial" feedback object, the server does things like assign the id, determine the creation timestamp, possibly augment the payload (e.g. in the emoji case we augment it with the shortcode). So the server response includes all of the fields that are needed to make a complete "Feedback" object that will match what you would get if you had queried for all of the feedback. 

So e.g. in the Python SDK, the response from creation allows us to update the list of feedback without having to do a separate roundtrip query or complete refresh.

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: i think we might want to heavily restrict this to a list of ids for now

weave/trace_server/trace_server_interface.py
jamie-rasmussen: In the case of feedback there are probably not a lot of cases (although supporting COUNT(*) is already useful) where you wouldn't want the payload, but if think about API patterns for more complicated objects, like calls, I think there are more cases where you want the ability to only pull down the fields you need.

weave/trace_server/trace_server_interface.py
tssweeney: if it is just useful for tests, i might not advertise it as a primary example in case you want to change the way aggregations are specified when you are doing an agg operation

weave/trace_server/trace_server_interface.py
tssweeney: I see - makes sense

weave/trace_server/trace_server_interface.py
tssweeney: Definitely agree that it is useful to filter fields. However, in practice, there are a bunch of id/date/name like fields that are low cost. Then there are big ones (inputs, output, attributes, summary). Similarly here you have a few simple fields, then a possibly huge field (payload). My mental model for calls has been to always return the "simple fields". Then essentially have "partials" for the others. In such cases, the typing would be consistent because the big fields are still dicts, just empty.

I don't think this is a blocking comment btw, but maybe something to consider (both ways) as we start to harden the API

weave/trace_server/trace_server_interface.py
tssweeney: If you want to retain the query capability, I'd prefer to just allow a single query shape that matches what you need (which i think could just be an "and" over ids)

weave/trace_server/orm.py
jamie-rasmussen: It seemed like in SQLite we were mapping all of the field types ORM currently supports (string, json, datetime) to a TEXT column, but this would need to be updated with an actual mapping to support other column types.

tssweeney (approved):

PR #1723, opened 2024-06-06
feat: python API for feedback

jamie-rasmussen:
Adds feedback APIs to the Python SDK.

Note: So far I have only added the feedback object to Calls and added querying for feedback at the project level. We would want to add feedback to objects like Datasets in a subsequent PR.

Some examples of use:

Screenshot 2024-06-06 at 4 26 22 PM

Screenshot 2024-06-06 at 4 28 39 PM

Screenshot 2024-06-06 at 4 37 02 PM


weave/weave_client.py
```diff
@@ -625,6 +644,59 @@ def set_call_display_name(
     def remove_call_display_name(self, call: Call) -> None:
         self.set_call_display_name(call, None)
 
+    def feedback(
+        self,
+        query: Optional[Union[Query, str]] = None,
+        reaction: Optional[str] = None,
+        offset: int = 0,
+        limit: int = 100,
+    ) -> Feedbacks:
```
tssweeney, line 653: Can you please add documentation to this method? I see two things that might be confusing for me:
1. `query` is implemented as `query_or_feedback_id`. I think you made this decision in order to make the first positional argument dual purpose and therefore nicer syntax. Anyway, would be good to make that clear
2. `reaction` -> is this supposed to be the literal emoji or shorthand? Both? Good to document

weave/weave_client.py
```diff
@@ -664,12 +736,6 @@ def ref_value_input_to(self, ref: "ref_base.Ref") -> list[Call]:
     def ref_output_of(self, ref: ObjectRef) -> typing.Optional[Call]:
         raise NotImplementedError()
 
-    def add_feedback(self, run_id: str, feedback: typing.Any) -> None:
```
tssweeney, line 667: was `add_feedback` implemented elsewhere? Why remove if not?

weave/pydantic_util.py
tssweeney, line 1: nit: maybe call this `rich_pydantic_util`

weave/weave_client.py
tssweeney: nvm, i see it is implemented on `RefFeedback`

weave/feedback.py
```diff
@@ -0,0 +1,241 @@
...
+                        ],
+                    }
+                }
+            ),
+        )
+        self.client.server.feedback_purge(req)
+        if self.items:
+            self.items = [f for f in self.items if f.id != feedback_id]
+
+    def _repr_pretty_(self, p: Any, cycle: bool) -> None:
```
tssweeney, line 233: is this automatically called by ipython? maybe i am unfamiliar with is stub?

weave/feedback.py
```diff
@@ -0,0 +1,241 @@
...
+        # expose that in this Python API.
+        return list(tsi.Feedback(**r) for r in response.result)
+
+    def purge(self, feedback_id: str) -> None:
+        req = tsi.FeedbackPurgeReq(
+            project_id=self.project_id,
+            query=Query(
+                **{
+                    "$expr": {
+                        "$eq": [
```
tssweeney, line 221: might be nice if this also included the ref in the query expression. Else, you could inadvertently purge an id that does not belong to that ref

weave/weave_client.py
jamie-rasmussen: Yeah, this isn't necessary for the new APIs, just me trying to cleanup some unimplemented stuff that seemed like it could cause confusion.

weave/feedback.py
```diff
@@ -0,0 +1,241 @@
...
+
+from . import graph_client_context
+from . import util
+from weave import pydantic_util
+from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.interface.query import Query
+from weave.trace.refs import parse_uri
+
+
+class Feedbacks:
```
tssweeney, line 16: I am seeing that we have 4 data structures:
1. `tsi.Feedback` -> The common `Feedback` data pydantic schema
2. `Feedbacks` -> A container of `tsi.Feedback`s 
   * used primarily for display
   * returned from `RefFeedback.query`
   * Implements an iterator protocol over `tsi.Feedback`
3. `RefFeedback` -> A "pointer" to feedback for a refable thing.
   * Implements an iterator protocol over `tsi.Feedback`
   * Lazily fetches data
   * Supports mutation operations (because it is relates to a specific "ref")
4. Implied: `Iterator[tsi.Feedback]`

I'm not sure exactly why, but `Feedbacks` and `RefFeedback` seem to overlap in an important way. 

Ok, i think either:
1. `RefFeedback` wants to inherit from `Feedbacks`
2. `RefFeedback` should contain a `Feedbacks` instead of `items`

For some reason i also feel like they are the same thing. `RefFeedback` and `Feedbacks` both can be thought of as `FeedbackQuery`. I think the only difference is that `RefFeedback` has the added property that you can add/purge. The other difference is that `Feedbacks` is pre-materialized, but i think you could also make `Feedbacks` lazy.

Ok, talking this out, consider changing to two classes: `FeedbackQuery` and `RefFeedbackQuery(FeedbackQuery)`. 

`FeedbackQuery` basically is merging of both `RefFeedback` and `Feedbacks` (except for `add/purge`). It is lazy, iterator, nicely prints, etc... It has an internal `_query` property holding the current query. It also has a `query` method that returns a new `FeedbackQuery` with the query append to the existing query.

Then, `RefFeedbackQuery` is a subclass with an additional `weave_ref` property. This is where add/purge goes. 

The result of this is that the laziness/items cache/printing/iteration/query building/execution are all centralized and only the specific abilities of the RefFeedback are pulled out to the other class.

weave/feedback.py
jamie-rasmussen: Yes, this is called automatically by ipython. https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

Shawn called out this pretty repr stuff as the bit he's most likely to want to change for consistency across APIs, but he was OK with it going in while he's trying out different possibilities.

weave/feedback.py
jamie-rasmussen: Actually, I'm going to defer this - it's a good idea, but I'd have to first update the server side APIs, because also in the name of safety we put in restrictions on the purge endpoint that the purge query must be only a ref or list of refs.

weave/refs.py
```diff
@@ -0,0 +1,28 @@
...
+    def _add_table_columns(self, table: Table) -> None:
+        table.add_column("Ref", overflow="fold")
+
+    def _item_to_row(self, item: str) -> list:
+        return [item]
+
+    def parsed(self) -> list[AnyRef]:
+        return [parse_uri(ref) for ref in self.items]
+
+    def call_refs(self) -> "Refs":
```
jamie-rasmussen, line 27: I'd like to add a `calls` method to this to go directly to a `list[TraceObject]`. This would allow you to write something like:
```
need_review = client.feedback(reaction="⚠️")
need_review_calls = need_review.refs().calls()
# Save the calls to a dataset or ...
```
But I think this should wait until weave.init settles down.

weave/feedback.py
```diff
@@ -0,0 +1,230 @@
...
+        self.project = project
+
+        self.show_refs = show_refs
+        self._query = query
+        self.offset = offset
+        self.limit = limit
+
+        self.feedbacks = None
+
+    def __iter__(self) -> Iterator[tsi.Feedback]:
```
tssweeney, line 114: any particular reason you excluded the len/index operations?

weave/feedback.py
```diff
@@ -0,0 +1,230 @@
...
+        return Feedbacks(self.show_refs, (tsi.Feedback(**r) for r in response.result))
+
+    def execute(self) -> Feedbacks:
+        if self.feedbacks is not None:
+            return self.feedbacks
+        self.feedbacks = self.refresh()
+        return self.feedbacks
+
+    def refs(self) -> Refs:
+        return self.execute().refs()
```
tssweeney, line 143: i wonder if this should be doing some sort of unique operation over the refs?

jamie-rasmussen: It does.

weave/weave_client.py
```diff
@@ -625,6 +644,78 @@ def set_call_display_name(
     def remove_call_display_name(self, call: Call) -> None:
         self.set_call_display_name(call, None)
 
+    def feedback(
+        self,
+        query: Optional[Union[Query, str]] = None,
+        reaction: Optional[str] = None,
```
tssweeney, line 650: i would consider putting `reaction` as a `kw-only` arg since it is syntactic sugar over query

weave/feedback.py
tssweeney: oh nice, missed this: `uris = list(set(feedback.weave_ref for feedback in self.items))`

tssweeney: hmmm well now that i think about this, doesn't that mean that ordering is not guaranteed anymore? Which would be unexpected if the order changes for each attempt to get calls for a certain reaction

jamie-rasmussen: Good catch! For some reason I was under the mistaken belief that they made sets preserve insertion order in 3.7 when they guaranteed it for dicts. Will fix it to use dict.fromkeys.

jamie-rasmussen: As my kids are fond of saying, "this is sus". :-) 
https://github.com/wandb/weave/blob/jamie/feedback-python-api/weave/tests/test_client_trace.py#L378

weave/feedback.py
jamie-rasmussen: Nope, just missed during refactor.

weave/refs.py
```diff
@@ -0,0 +1,39 @@
...
+        return Refs(ref for ref in self.items if isinstance(parse_uri(ref), CallRef))
+
+    # TODO: Perhaps there should be a Calls that extends AbstractRichContainer
+    def calls(self) -> list[TraceObject]:
```
tssweeney, line 32: 2 notes:
1. This could be optimized by grouping all the ids together then executing a single request instead of N network requests
2. I feel like this wants to be on `FeedbackQuery` itself, so that my code is:
```
client.feedback(...).calls()
```

tssweeney (approved):
Approving, with comments. I think it would be good to get this in an keep iterating if need

weave/refs.py
jamie-rasmussen: Thanks, can definitely address these in a followup.

PR #1726, opened 2024-06-06
chore(weave): Fixes Stream response format in OpenAI

tssweeney:
Our OpenAI integration was not returning a `Stream` object from the sync streaming API. This was not caught in the past because most people just use it for it's iterator interface and everything works. However, Langchain actually uses the context block (enter/exit) dunder methods. In order to bring our integration back into compliance, I implement a similar class as the `WeaveStream` which is the sister class to `WeaveAsyncStream`. These now follow a common pattern and are true subclasses of the original openai class.

This should have no functional change for users and hits existing tests.

This came to my attention because the growth team is implementing Langchain and hit this underlying issue. 
```
 File "/opt/anaconda3/envs/weave/lib/python3.10/site-packages/langchain_openai/chat_models/base.py", line 480, in _stream
    with self.client.create(messages=message_dicts, **params) as response:
AttributeError: __enter__
```


weave/flow/chat_util.py
```diff
@@ -22,75 +22,83 @@ def __init__(self, chunk_iter: Iterator[ChatCompletionChunk]) -> None:
...
-                        }
-                    )
 
-                # choice fields
-                choice = self.output_choices[chunk_choice.index]
-                if chunk_choice.finish_reason:
-                    choice["finish_reason"] = chunk_choice.finish_reason
-                if chunk_choice.logprobs:
-                    choice["logprobs"] = chunk_choice.logprobs
+    def _process_chunk(self, chunk: ChatCompletionChunk) -> None:
```
tssweeney, line 35: This looks like a huge change, but if you hide whitespace, you see that it is just moving some logic to a helper function.

weave/monitoring/openai/openai.py
```diff
@@ -152,21 +187,14 @@ def _streaming_create(
...
-                    yield chunk
-                result = wrapped_stream.final_response()
-                result_with_usage = ChatCompletion(
-                    **result.model_dump(exclude_unset=True),
-                    usage=token_usage(messages, result.choices),
-                )
-                finish_run(result_with_usage.model_dump(exclude_unset=True))
-
-        return _stream_create_gen()  # type: ignore
+            base_stream = self._base_create(*args, **kwargs)
```
tssweeney, line 190: This logic is moved into the class

weave/monitoring/openai/openai.py
```diff
@@ -152,21 +187,14 @@ def _streaming_create(
...
+            stream = WeaveStream(
+                base_stream=base_stream,
+                messages=messages,
+                finish_run=finish_run,
+            )
+
+        return stream
```
tssweeney, line 197: Notice we get rid of type ignore!

weave/monitoring/openai/openai.py
```diff
@@ -84,6 +84,41 @@ async def __aiter__(self) -> AsyncIterator[ChatCompletionChunk]:
         self._finish_run(result_with_usage.model_dump(exclude_unset=True))
 
 
+class WeaveStream(Stream):
```
tssweeney, line 87: New class - pretty simple. It follows the same pattern as the async class above and uses the same code that was previously in a function below.

weave/tests/test_monitoring_openai.py
```diff
@@ -369,10 +369,52 @@ def teardown():
     unpatch()
 
 
+class MockSyncResponse:
```
tssweeney, line 372: Follow same pattern as async

shawnlewis (approved):

weave/flow/chat_util.py
```diff
@@ -22,75 +22,83 @@ def __init__(self, chunk_iter: Iterator[ChatCompletionChunk]) -> None:
...
+                choice["message"]["content"] += chunk_choice.delta.content
+            if chunk_choice.delta.role:
+                choice["message"]["role"] = chunk_choice.delta.role
 
-                # function call
-                if chunk_choice.delta.function_call:
-                    raise NotImplementedError("Function calls not supported")
+            # function call
+            if chunk_choice.delta.function_call:
+                raise NotImplementedError("Function calls not supported")
```
parambharat, line 69: @tssweeney:  I would like to understand why function calls are not supported here. I know this is deprecated in Openai SDK in favor of tool calls, but libraries like LangChain currently still use this feature, and this causes the following test case related to agents break in LangChain: 

https://github.com/wandb/weave/blob/6cd218ec90f20264ef45598123f3e02e37aa11a0/weave/integrations/langchain/langchain_test.py#L305

PR #1728, opened 2024-06-07
Simple HF model wrapping

tcapelle:
Adds a doc page on how to wrap a local HF model with Weave

### DIFF
--- a/docs/docs/guides/ecosystem/huggingface.md
+++ b/docs/docs/guides/ecosystem/huggingface.md
@@ -27,6 +27,9 @@ You have fine-tuned a model from the Hugging Face HUB a now you want to interact
         pipe = pipeline("text-generation", model=model_name, temperature=temperature)
         return pipe(messages)
     ```
+    
+    this is very inefficient as the model is instantiated every time the function is called.
+
 2. Explicitly using the `AutoModel` and `generate` methods:
     ```python
     import weave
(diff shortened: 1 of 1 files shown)


tcapelle:
Maybe @NiWaRe you can refactor and test this?

docs/docs/guides/ecosystem/huggingface.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Huggingface
+
+Recording traces of LLM applications in a centralized database is crucial during both development and production phases. These traces are invaluable for debugging and for creating a dataset of challenging examples to test and enhance your application.
+
+You can use models from the Hugging Face ecosystem with Weave to keep track of all your interactions and generate insights.
+
+## Using a models locally
+
+You have fine-tuned a model from the Hugging Face HUB a now you want to interact with it. You can start tracing your model by wrapping the model call with a `weave.op()`:
```
NiWaRe, line 15: @tcapelle I just had a look at the other ecosystem docs (OpenAI, Anthropic, Mistral) and for them the same main issue applies as for this HF pipeline - we init a new model object on every predict call. 

If the model inference happens in the cloud re-creating the model object comes down to re-creating a client object that is cheap to re-init (that's why for OpenAI, etc. it didn't make a big difference) but if the inference happens locally then the model object can't be re-init on every inference call (downloading, loading in memory, etc. would take time). 

Imo would make sense to create dedicated page in the docs about model init best practices (regarding the constructor) and related to that for weave objects and serialization. And then from the HF page (and also the other ecosystem pages) we could link to them - what do you think?

docs/docs/guides/ecosystem/huggingface.md
```diff
@@ -0,0 +1,116 @@
...
+
+    @weave.op
+    def generate(prompt: str, temperature: float = 0.7) -> str:
+        messages = [{"role": "user", "content": prompt}]
+        input_ids = tokenizer.apply_chat_template(messages, return_tensors="pt").to(model.device)
+        output = model.generate(input_ids, temperature=temperature, do_sample=True)
+        return outputs[0][input_ids.shape[-1]:] # we remove the input from the generation
+    ```
+
+3. Explicitly using the `AutoModel` and `generate` methods:
```
NiWaRe, line 49: @tcapelle we would need to find a slightly different name for this approach than for Nr. 2 (for now it's the same) - maybe also just skip the Nr. 2 and only show this one to push the best practice of having an explicit prompt formatting step?

tcapelle: push push, feel free to make any change

docs/docs/guides/ecosystem/huggingface.md
tcapelle: agree

PR #1729, opened 2024-06-07
fix(weave): Allow client to get ref without explicit `weave.init`

andrewtruong:
Fixes an issue where Tables did not auto-deref as expected when `weave.init` was not first explicitly called.

### DIFF
--- a/weave/trace/vals.py
+++ b/weave/trace/vals.py
@@ -116,11 +116,19 @@ def pydantic_getattribute(self: BaseModel, name: str) -> Any:
             return object.__getattribute__(self, "ref")
         except AttributeError:
             return None
-    res = attribute_access_result(self, attribute, name)
+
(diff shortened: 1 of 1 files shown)


weave/trace/vals.py
```diff
@@ -112,11 +116,19 @@ def pydantic_getattribute(self: BaseModel, name: str) -> Any:
...
     return res
 
 
-def attribute_access_result(self: object, val_attr_val: Any, attr_name: str) -> Any:
+def attribute_access_result(
+    self: object,
+    val_attr_val: Any,
+    attr_name: str,
+    *,
+    server: Optional[TraceServerInterface],
```
andrewtruong, line 130: Take a new `server` kwarg instead of using gc

shawnlewis (approved):

PR #1733, opened 2024-06-07
chore(weave): Tests for Threading Support

tssweeney:
We are going to be modifying how our context state / stack is managed across threads in the near-term. I wanted to make sure the desired behavior is properly tested. This new test has 3 flavors (1 of which is commented out because it current fails!).

When we fix threading support, we should be able to uncomment out and enable this test.


gtarpenning (approved):
is sort of independent enough + confusing + has considerable helper apparatus to consider another test file. just a thought.

tssweeney:
> is sort of independent enough + confusing + has considerable helper apparatus to consider another test file. just a thought.

totally agreed. I have been shoving everything in this test file so it is easier to keep track of what is "new weave" vs "old weave"... sigh...

PR #1736, opened 2024-06-07
fix(weave): Make weave.init work with threading/flask

andrewtruong:
Updates `get_graph_client` to work in cases where weave is init'd in the same process but on a different thread.

This allows a single weave.init() to serve as the entrypoint for:
1. Multithreading (e.g. ThreadPoolExecutor); and
2. Flask servers (where handlers are on different threads)


weave/state/graph_client.py
```diff
@@ -0,0 +1,42 @@
...
+    # These outer guards are to avoid expensive lock acquisition
+    if client is not None and _global_graph_client is None:
+        with lock:
+            if _global_graph_client is None:
+                _global_graph_client = client
+
+    elif client is None and _global_graph_client is not None:
+        with lock:
+            if _global_graph_client is not None:
+                _global_graph_client = client
```
andrewtruong, line 26: Most operations will set the client globally, which means it should "just work" in threaded cases

weave/state/graph_client.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def set_graph_client_context(client: Optional["WeaveClient"]) -> None:
+    context_state._graph_client.set(client)
```
andrewtruong, line 30: We also provide a context version in case you need that

weave/state/graph_client.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_graph_client() -> Optional["WeaveClient"]:
+    if (context_client := context_state._graph_client.get()) is not None:
+        return context_client
+    return _global_graph_client
```
andrewtruong, line 36: Getting a client prefers the context first and fall back to the global second.

This enables the user to just write `weave.init(...)` and have it "just work" for threaded cases like a Streamlit app or ThreadPoolExecutor

weave/api.py
```diff
@@ -331,10 +332,10 @@ def serve(
     def run():
         # This function doesn't return, because uvicorn.run does not
         # return.
-        with _graph_client_context.set_graph_client(client):
-            with _wandb_api.wandb_api_context(wandb_api_ctx):
-                with attributes(trace_attrs):
-                    uvicorn.run(app, host="0.0.0.0", port=port)
+        state.graph_client.set_graph_client_global(client)
```
shawnlewis, line 335: I don't think we should need to set the graph client here any more at all.

shawnlewis (approved):

PR #1739, opened 2024-06-08
chore(weave): Add attributes to Call object and allow explicit root

tssweeney:
This PR supports the Langchain integration work and does two things:
1. Makes sure the attributes payload is attached to the Calls (this is just good to do in general)
2. Allows the user to explicitly force a root trace (as opposed to inheriting from context stack
3. Moves parent to 3rd param and sets better default


weave/weave_client.py
```diff
@@ -442,6 +443,28 @@ def create_call(
...
+        run_context.push_call(call)
+        return call
+
+    @trace_sentry.global_trace_sentry.watch()
+    def create_call_outside_execution(
+        self,
+        op: Union[str, Op],
+        parent: Optional[Call],
+        inputs: dict,
+        attributes: dict = {},
```
jamie-rasmussen, line 458: Should avoid using a mutable object as a default value

jamie-rasmussen (approved):

weave/weave_client.py
```diff
@@ -457,9 +459,21 @@ def create_call(
...
     ) -> Call:
+        """Create, log, and push a call onto the runtime stack.
+
+        Args:
+            op (Union[str, Op]): The operation to log.
+            parent (Optional[Call]): The parent call. If parent is not provided, the current run is used as the parent.
+            inputs (dict): The inputs to the operation.
+            display_name (Optional[str], optional): The display name for the call. Defaults to None.
+            attributes (Optional[dict], optional): The attributes for the call. Defaults to None.
+            _use_stack (bool, optional): Whether to push the call onto the runtime stack. Defaults to True.
```
shawnlewis, line 475: Why prefix with "_"? I think its just as obscure as "parent". I'd just call it "use_stack".

I think the docstring could cover both cases of stack use: If True, read parent from call_context, and push call onto call_context.

Should parent and use_stack be mutually exclusive? I mean should we throw an error if both are passed?

Ideally parent would have a default value of None and be moved to after inputs. This would mean we need to update positional callers, but that's probably not too bad right now since they're all in our codebase.

weave/weave_client.py
```diff
@@ -457,9 +459,21 @@ def create_call(
         op: Union[str, Op],
         parent: Optional[Call],
         inputs: dict,
-        attributes: dict = {},
+        attributes: Optional[dict] = None,
         display_name: Optional[str] = None,
+        *,
+        _use_stack: bool = True,
     ) -> Call:
+        """Create, log, and push a call onto the runtime stack.
```
shawnlewis, line 467: Since you're adding the docstring, I'm going to review it.

Style: Please use the style in api.py for publish. My little doc generator script expects that style and I made some decisions in choosing it. In particular, don't put the type next to the parameter. The type information is already annotated in the function signature, no need to reproduce it.

weave/weave_client.py
```diff
@@ -457,9 +459,21 @@ def create_call(
...
+
+        Args:
+            op (Union[str, Op]): The operation to log.
```
shawnlewis, line 470: The Op that this is a call for. If a string is passed, a code-less op will be created using "op" as the id.

shawnlewis (changes requested):

weave/weave_client.py
tssweeney: fixed

weave/weave_client.py
tssweeney: fixed

weave/weave_client.py
tssweeney: removed underscore. I like to use that when the param is not something we want to be used externally and therefore have more liberties to change

tssweeney: we don't want mutual exclusivity. We want to be able to specify a parent and to not use the stack (this is the exact case in Langchain integration)

weave/weave_client.py
tssweeney: fixed

weave/weave_client.py
tssweeney: moved parent to 3rd

shawnlewis (approved):

PR #1748, opened 2024-06-11
fix(weave): format values for csv export in datagrid

gtarpenning:
https://wandb.atlassian.net/browse/WB-19203

This pr: 
- adds getter methods to datagrid for csv export. we use renderCell methods that prevent clean export. 
- dump objects and lists to string using `JSON.stringify`, otherwise lists mess up the columns of export, and objects just show up as `[object Object]`

Before:
Screenshot 2024-06-11 at 9 50 51 AM

After: 
Screenshot 2024-06-11 at 10 26 07 AM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
@@ -326,7 +326,11 @@ function buildCallsTableColumns(
       valueGetter: cellParams => {
         const val = (cellParams.row as any)[key];
         if (Array.isArray(val) || typeof val === 'object') {
-          return JSON.stringify(val);
+          try {
+            return JSON.stringify(val);
+          } catch {
+            return val;
+          }
         }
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
```diff
@@ -406,6 +427,9 @@ function buildCallsTableColumns(
     // Should probably have a custom filter here.
     filterable: false,
     sortable: false,
+    valueGetter: cellParams => {
```
jamie-rasmussen, line 430: For a case like this, should the valueGetter have the same logic re: UNSET as renderCell, and then I think maybe we don't need renderCell?

gtarpenning: Yeah probably, although there are cases where the UI must be different because of rendering. Ill fix this one @jamie-rasmussen .

jamie-rasmussen (approved):

PR #1749, opened 2024-06-11
fix(weave): Mutate and return the same object when mutating inside an op

andrewtruong:


weave/trace/vals.py
```diff
@@ -119,6 +119,10 @@ def pydantic_getattribute(self: BaseModel, name: str) -> Any:
 
     server = gc.server if (gc := get_graph_client()) else None
     res = attribute_access_result(self, attribute, name, server=server)
+
+    # We need this because we override getattribute.  If we don't do this, we can't guarantee
+    # that `obj.a is obj.a` because the lookup may fail and try to create a new object.
```
shawnlewis, line 124: "because the lookup may fail and try to create a new object"

I'd say this different.

The newly wrapped res may be something mutable like a list. Ensure we replace the original attribute on self, so that mutations correctly reflect on `self.<attribute>`

Ensure we store the new wrapped value on self.

shawnlewis (approved):

PR #1759, opened 2024-06-12
WIP: Refactor client

andrewtruong:


weave/andrew_client/client.py
```diff
@@ -0,0 +1,106 @@
...
+class Client:
+    """Now client is just a convenient way work with the Weave API."""
+
+    entity: str
+    project: str
+    api: WeaveApi = field(repr=False)
+
+    # TODO: add user-facing convenience methods here; for lower level, reach into api directly
+    def get(self, ref): ...
+    def save(self, val, name, branch): ...
```
andrewtruong, line 106: Users primarily interact with these convenience methods.  If they want to go lower level, they can reach into the embedded `api`.

weave/andrew_client/client.py
```diff
@@ -0,0 +1,106 @@
...
+        objects=ConcreteObjects(),
+        ...
+    )
+
+    api.get_call(...)  # this is delegated to api.calls.get_call(...)
+    """
+
+    calls: Calls
+    objects: Objects
+    ...
```
andrewtruong, line 72: This lower level API provides the standard REST-ish commands that are currently on the current `WeaveClient`

weave/andrew_client/client.py
```diff
@@ -0,0 +1,106 @@
...
+            calls=SqliteServerCalls(),
+            objects=SqliteServerObjects(),
+        )
+
+    @classmethod
+    def with_some_weird_mixture(cls):
+        return cls(
+            calls=RemoteHTTPTraceServerCalls(),
+            objects=SqliteServerObjects(),
+        )
```
andrewtruong, line 93: Now you can cleanly specify a backend and even patch in a mixture if you need it (e.g. during tests)

weave/andrew_client/client.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@runtime_checkable
+class Objects(Protocol):
+    def get_object(self): ...
+    def create_object(self): ...
+
+
+@runtime_checkable
+class WeaveApi(Calls, Objects, Protocol): ...
```
andrewtruong, line 54: This abstraction makes it easier for internal devs to work with just a section of the code at a time.

weave/andrew_init/global_state.py
```diff
@@ -0,0 +1,27 @@
+# forget contexts, let's just use globals for now and then revisit
```
andrewtruong, line 1: Not sure if we are still using contextvars.  Either way, we should make sure all shared state is in one place that's easy to see...

weave/andrew_init/lifecycle.py
```diff
@@ -0,0 +1,44 @@
...
+    # zero out global state
+    # utils.sentry_reset_scope()
+    trace_sentry.global_trace_sentry.end_session()
+
+    # TODO: move gc context to this new global to re-enable
+    # autopatch.reset_autopatch()
+    global_state.set_client(None)
+
+    # flush any messages stuck in the queue
+    ...
```
andrewtruong, line 35: `init` and `finish` now operate on global state.  Setup code is simpler and more readable because they're moved to where they're actually needed.  Most of the setup is for the server.

weave/andrew_init/utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # print init message
+    from .. import urls
+
+    username = get_username()
+    print(f"Logged in as Weights & Biases user: {username}.")
+    print(f"View Weave data at {urls.project_weave_root_url(entity, project)}")
+
+    return api_key
+    # remote_server = init_weave_get_server(api_key)
```
andrewtruong, line 78: I'm not sure how much of this is actually necessary, but it's here for now.

PR #1761, opened 2024-06-12
feat(weave): Cost + tokens in UI

jwlee64:
Main Changes
I may split out the three changes and push them in the order below

1. adds chips to trace tree with a tooltip
https://github.com/wandb/weave/pull/1767

2. adds a table in the call summary with the costs
https://github.com/wandb/weave/pull/1766/files

3. removes the summary.usage columns in calls table, and adds two columns for tokens and cost
https://github.com/wandb/weave/pull/1765

https://github.com/wandb/weave/assets/25037002/3a327787-022f-4c56-b0ff-779842a671d9

internal notion 
https://www.notion.so/wandbai/Design-Displaying-Tokens-and-Cost-in-Weave-cf547c76b3c54faeb47b34fb86f62145
internal jira
https://wandb.atlassian.net/browse/WB-19140

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallSummary.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallSummary.tsx
@@ -59,20 +59,24 @@ export const CallSummary: React.FC<{
           ...(Object.keys(summary).length > 0 ? {Summary: summary} : {}),
         }}
       />
-      <Divider />
       {span.summary.usage && (
-        <div>
-          {/* This styling is similar to what is is SimpleKeyValueTable */}
-          <p
-            style={{
-              fontWeight: 600,
-              marginRight: 10,
-              paddingRight: 10,
-            }}>
-            Usage
-          </p>
-          <CostTable usage={span.summary.usage as {[key: string]: UsageData}} />
-        </div>
+        <>
+          <Divider />
+          <div>
+            {/* This styling is similar to what is is SimpleKeyValueTable */}
+            <p
+              style={{
+                fontWeight: 600,
+                marginRight: 10,
+                paddingRight: 10,
+              }}>
+              Usage
+            </p>
+            <CostTable
+              usage={span.summary.usage as {[key: string]: UsageData}}
+            />
+          </div>
+        </>
       )}
     </div>
   );
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallSummary.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallSummary.tsx
@@ -10,6 +10,8 @@ import {CallSchema} from '../wfReactInterface/wfDataModelHooksInterface';
 import {CostTable} from './CostTable';
 import {UsageData} from './TraceUsageStats';
 
+const SUMMARY_FIELDS_EXCLUDED_FROM_GENERAL_RENDER = ['_latency_s', 'usage'];
+
 export const CallSummary: React.FC<{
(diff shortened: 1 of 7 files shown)


weave-js/src/util/llmTokenCosts.ts
```diff
@@ -0,0 +1,181 @@
...
+  'gpt-3.5-turbo-0125',
+  'gpt-3.5-turbo-instruct',
+  'davinci-002',
+  'babbage-002',
+  'claude-3-opus-20240229',
+  'claude-3-sonnet-20240229',
+  'claude-3-haiku-20240307',
+];
+
+export type MODEL = (typeof modelTuple)[number];
```
tssweeney, line 25: nit: instead of defining the modelTuple as a hard-coded list, you could get the keys from `LLM_TOKEN_COSTS`. and `export type MODEL` would be defined as a key of `LLM_TOKEN_COSTS`

tssweeney, line 25: nit: types are rarely all-caps, any reasoning for this in particular?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
```diff
@@ -448,7 +485,11 @@ const useAllDynamicColumnNames = (
       tableData.forEach(row => {
         Object.keys(row).forEach(key => {
           const keyAsPath = stringToPath(key);
-          if (isDynamicCallColumn(keyAsPath)) {
+          // Removes summary.usage columns, as they have their own dedicated column.
+          if (
+            isDynamicCallColumn(keyAsPath) &&
+            !key.startsWith('summary.usage')
```
tssweeney, line 491: nit: i think the `!key.startsWith('summary.usage')` might be better placed as a filter on top of the allDynamicColumns passed into `function buildCallsTableColumns(`... since the implementation of `buildCallsTableColumns` deals with adding derived columns for these fields, it would be nice to co-locate this logic

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallSummary.tsx
```diff
@@ -18,7 +21,9 @@ export const CallSummary: React.FC<{
   );
   const summary = _.fromPairs(
     Object.entries(span.summary ?? {}).filter(
-      ([k, a]) => !k.startsWith('_') && k !== 'latency_s' && a != null
+      ([k, a]) =>
+        // Display all summary fields, but remove usage stats because we have a separate table
+        !k.startsWith('_') && k !== 'latency_s' && a != null && k !== 'usage'
```
tssweeney, line 26: can we make a const called `SUMMARY_FIELDS_EXCLUDED_FROM_GENERAL_RENDER` and add `latency_s and `usage`?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CustomGridTreeDataGroupingCell.tsx
```diff
@@ -176,19 +182,32 @@ export const CustomGridTreeDataGroupingCell: FC<
...
+                sx={{
+                  overflow: 'hidden',
+                  textOverflow: 'ellipsis',
+                  whiteSpace: 'nowrap',
+                  flex: '1 1 auto',
+                }}>
+                {opNiceName(call.spanName)}
+              </Box>
             </Box>
+            {call?.rawSpan?.summary && (
```
tssweeney, line 205: what does the UI look like when this is not there? (I think this is true for an in-progress call?)

PR #1763, opened 2024-06-12
chore(weave): Simplify TraceServerInterface definition and fix inheritance 

tssweeney:
We introduced a violation of inheritance rules with `TraceServerInterfacePostAuth`. We added this because the interface definition for the HTTP server differs slightly from that of the internal trace storage layers (namely the inclusion of wb_user_id which is populated by the server - and therefore not required).

This was problematic for a few reasons:
1. Most importantly: it made it so that `RemoteHTTPTraceServer` was no longer compatible with the other servers. This is not a good place to be in.
2. It violates inheritance rules
3. It introduces additional complexity (not that big of a deal

To remedy this, I:
1. Removed the `TraceServerInterfacePostAuth`
2. Added runtime checks to assert the presence of the required field in the implementations which require it.

### DIFF
--- a/weave/tests/test_client_feedback.py
+++ b/weave/tests/test_client_feedback.py
@@ -10,7 +10,7 @@ def test_feedback_apis(client):
     project_id = client._project_id()
 
     # Emoji from Jamie
-    req = tsi.FeedbackCreateReqForInsert(
+    req = tsi.FeedbackCreateReq(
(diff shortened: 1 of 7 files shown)


gtarpenning (approved):
lgtm, @jamie-rasmussen might want to 👀  too though

PR #1764, opened 2024-06-12
chore(weave): Brings over ref conversion logic from WB to trace server

tssweeney:
This PR basically copies a few functions currently defined in the WB Trace Server into the weave trace_server directory. These are going to be used by a common class that can help streamline tests soon.


weave/trace_server/trace_server_converter.py
```diff
@@ -0,0 +1,116 @@
...
+from . import refs_internal as ri
+
+
+def universal_ext_to_int_ref_converter(
+    obj: typing.Any, convert_ext_to_int_project_id: typing.Callable[[str], str]
+) -> typing.Any:
+    """Takes any object and recursively replaces all external references with
+    internal references. The external references are expected to be in the
+    format of `weave:///entity/project/...` and the internal references are
+    expected to be in the format of `weave-internal:///project_id/...`.
```
jamie-rasmussen, line 14: Is it actually `weave-trace-internal:///...`

jamie-rasmussen (approved):

weave/trace_server/trace_server_converter.py
tssweeney: damn. actually used code-pilot to write this and still missed something.

PR #1765, opened 2024-06-12
chore(weave): Add cost and token column to calls table

jwlee64:
Screenshot 2024-06-12 at 2 37 40 PM
Screenshot 2024-06-12 at 2 37 35 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
@@ -96,6 +96,11 @@ export const useCallsTableColumns = (
     effectiveFilter
   );
 
+  // Filters summary.usage. because we add a derived column for tokens and cost
+  const filteredDynamicColumnNames = allDynamicColumnNames.filter(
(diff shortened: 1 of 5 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
```diff
@@ -303,6 +308,11 @@ function buildCallsTableColumns(
   }) as GridColumnGroupingModel;
 
   for (const key of allDynamicColumnNames) {
+    // We don't want to show usage stats in the table, because we have a separate derived column for that.
+    if (key.startsWith('summary.usage')) {
```
tssweeney, line 312: sorry to comment here again, but there is aline above `const tree = buildTree([...allDynamicColumnNames]);` that should filter this too.

I think you want something like:
```
const usableDyanmicColumnNames = allDynamicColumnNames.filter(...)
```
then use `usableDyanmicColumnNames` where `allDynamicColumnNames` is now

jwlee64: makes sense

tssweeney (approved):

PR #1769, opened 2024-06-12
docs: adds docstrings to objects

lukas:
I added some light docstrings that would have helped me a lot integrating weave into my project. Happy to add more if this is useful. Also happy to reformat if we have a docstring standard I don't know about.


scottire (approved):

scottire:
Thanks for your contribution, @lukas.

PR #1771, opened 2024-06-13
chore: Generalize Converter Logic 

tssweeney:
Benefits:
* (Critical!) Ref formats are now consistent across backends
* `wb_user_id` auto injected simulating server behavior
* Hashes are now consistent between backends
* Every test works for both backends (except VCR tests, but that is a different issue)
* The ID conversion logic is now an aspect of every test, reducing untested surface area in the server
   * As a result, internal server code: https://github.com/wandb/core/pull/22059 gets a huge reduction

TODO:
- [x] Land https://github.com/wandb/weave/pull/1772
- [x] Implement user_id conversions
    - [x] Remember to do this in filters!
- [x] Implement run_id conversions
- [x] Attempt to standardize some of the code structure in `weave/trace_server/external_to_internal_trace_server_adapter.py` - prone to errors if developers are not careful
- [x] (optional, but should do) Uptake these changes in (https://github.com/wandb/core/pull/22059)
  - [x] Run the test suite against the full server (requires working out some auth / running just locally?)
  - [x] Things to be careful with:
     - [x] obviously this touches auth, so just be careful
     - [x] for loop-based conversions, the naive implementation will be too slow (see Nick's changes as a guide)


weave/tests/test_client_trace.py
```diff
@@ -710,8 +710,12 @@ def basic_op(in_val: dict, delay) -> dict:
         assert inner_res.calls[2].inputs["in_val"]["prim"] == last
 
 
+def client_is_sql_lite(client):
```
jamie-rasmussen, line 713: I'd use `sqlite` instead of `sql_lite`

jamie-rasmussen (approved):

PR #1780, opened 2024-06-13
chore(weave): Split out old weave code

andrewtruong:
**This PR:**
1. Is a first-pass attempt to split out the "old" weave code;
2. Only splits out "old" from "new".  The old code is moved to `weave.legacy`, while the new code remains the top level.
3. Renames many of the relative imports to absolute imports (e.g. `from .. import xyz` becomes `from weave.legacy import xyz`)

**This PR does not:**
1. Try to remove "old" weave dependencies from the "new" code (but some of that may happen by splitting)
2. Update the components in core -- https://github.com/wandb/core/pull/22146

Following this PR, we'll also do a:
1. 1-time format of the repo using `ruff` -- https://github.com/wandb/weave/pull/1801


shawnlewis:
Force merging because of bi-directional dependency issue.

PR #1782, opened 2024-06-13
chore: fix double json encoding of feedback payload

jamie-rasmussen:
The ORM layer is supposed to handle the dump/load of json columns from a dict to a string column in the db. I had added a separate JSON dump to string when I added a size limit check on the payload, but I should have still been passing in the payload as a dict. Added a test for this.

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -996,7 +996,7 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
             "wb_user_id": req.wb_user_id,
             "creator": req.creator,
             "feedback_type": req.feedback_type,
-            "payload": payload,
+            "payload": req.payload,
             "created_at": created_at,
         }
         prepared = TABLE_FEEDBACK.insert(row).prepare(database_type="clickhouse")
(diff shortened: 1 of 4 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -996,7 +996,7 @@ def feedback_create(self, req: tsi.FeedbackCreateReq) -> tsi.FeedbackCreateRes:
             "wb_user_id": req.wb_user_id,
             "creator": req.creator,
             "feedback_type": req.feedback_type,
-            "payload": payload,
+            "payload": req.payload,
```
tssweeney, line 999: i see... the ORM does the encoding. Nice. 

idea: can the ORM enforce MAX_PAYLOAD to avoid double stringification?

tssweeney (approved):

weave/trace_server/clickhouse_trace_server_batched.py
jamie-rasmussen: Yes, in some databases I would expect the database itself to have the length constraint on the column, but I don't think that's the ClickHouse way, so doing it at our ORM level makes sense. I can add that as a follow on. Since we do have quite a strict size limit, the second encoding is expected to be fast, but it would be better to only do it once.

PR #1783, opened 2024-06-14
chore(weave): Mark test as flaky

andrewtruong:
This test seems to flake periodically in CI.  I haven't been able to reproduce, but based on the error message I think this might be a case where the object doesn't de-ref as expected.


tssweeney (approved):
Approving, but it would really be nice to have a comment here explaining the observed behavior with some notes as to any ideas. We shouldn't get in the habit of just hiding flakes

PR #1795, opened 2024-06-17
chore(weave): Quick fix for anthropic cost

jwlee64:
anthropic returns 
input_tokens
output_tokens

instead of
prompt_tokens
completion_tokens

This is not a sustainable fix, will have to come back and refactor, the way we set usage in finish_call or a better parsing method in the frontend as the models we want to support expand


weave-js/src/util/llmTokenCosts.ts
```diff
@@ -3,106 +3,109 @@ export const isValidLLMModel = (model: string): model is Model => {
   return model in Object.keys(LLM_TOKEN_COSTS);
 };
 
-export const LLM_TOKEN_COSTS: Record<string, {input: number; output: number}> =
```
jwlee64, line 6: getting rid of the this typing allows the above type, `keyof typeof LLM_TOKEN_COSTS` to correctly be set, instead of just being string

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/TraceUsageStats.tsx
```diff
@@ -91,6 +93,16 @@ export const getUsageFromCellParams = (params: {[key: string]: any}) => {
   return usage;
 };
 
+// This needs to updated eventually, to either include more possible keys or to be more dynamic
+// accounts for openai and anthropic usage objects (prompt_tokens, input_tokens)
+export const getInputTokens = (usage: UsageData) => {
+  return usage.prompt_tokens || usage.input_tokens || 0;
```
jamie-rasmussen, line 99: I think you would want to use `??` and check for input_tokens first. Same comment for output calculation.

weave-js/src/util/llmTokenCosts.ts
```diff
@@ -3,106 +3,109 @@ export const isValidLLMModel = (model: string): model is Model => {
...
+    input: 0.00025,
+    output: 0.00125,
+  },
+};
 
 export const getLLMTokenCost = (
   model: string,
   type: 'input' | 'output',
-  tokens: number
+  tokens: number = 0
```
jamie-rasmussen, line 103: What's the reason for adding the default 0 on these methods?

weave-js/src/util/llmTokenCosts.ts
jamie-rasmussen: I think something like this should also work, but I'm ok with it as is.
```
type KnownModel =
  | 'gpt-4o'
  | 'gpt-4o-2024-05-13'
  | 'gpt-4o-2024-05-13'
  | 'gpt-4-turbo-2024-04-09'
  ...

type Costs = {
  input: number;
  output: number;
};

export const LLM_TOKEN_COSTS: Record<KnownModel | 'default', Costs> = {
```

jamie-rasmussen (approved):

PR #1796, opened 2024-06-17
chore(app): Enhance getTimeSegmentsInTimeZone Function to Include Day Property

MBakirWB:
Jira Reference:
https://wandb.atlassian.net/browse/WB-19359

Support team will be reducing support hours for non paying users to Tuesday, Wednesday, Thursday

This PR enhances the getTimeSegmentsInTimeZone function by including the day of the week in its return value to use as part of conditional logic in core

Change:
Definition of New Types:
WeekTimeSegments: An extended type of TimeSegments that also includes the day property.

Function Update:
The getTimeSegmentsInTimeZone function is updated to return an object of type WeekTimeSegments, which includes the time segments (hour, minute, second) and now the day of the week.

### DIFF
--- a/weave-js/src/common/util/time.ts
+++ b/weave-js/src/common/util/time.ts
@@ -236,7 +236,7 @@ type TimeSegments = {
   second: number;
 };
 
-export function getTimeSegmentsInTimeZone(timeZone: string): TimeSegments {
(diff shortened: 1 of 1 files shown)


MBakirWB:
* * *
* * *

MBakirWB:
recheck

jcrawley (approved):

PR #1797, opened 2024-06-17
chore: Revise intro notebook to match demo flow

tssweeney:


docs/intro_notebook.ipynb
```diff
@@ -51,45 +51,50 @@
...
+   "id": "0921bbef-d089-4960-b624-505172f90d76",
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Log in to your W&B account\n",
-    "import weave"
+    "# Set your OpenAI API key\n",
+    "from set_env import set_env\n",
+    "\n",
+    "# Put your OPENAI_API_KEY in the secrets panel to the left 🗝️\n",
```
jamie-rasmussen, line 61: Do we want this notebook to be colab-specific?

docs/intro_notebook.ipynb
```diff
@@ -135,27 +138,26 @@
...
    "metadata": {},
    "outputs": [],
    "source": [
     "import weave\n",
-    "from openai import OpenAI\n",
     "\n",
     "weave.init(PROJECT)\n",
     "\n",
-    "@weave.op\n",
+    "@weave.op()\n",
```
jamie-rasmussen, line 160: Is there a reason to put the parens in? I think we've been moving towards stripping them out.

jamie-rasmussen (approved):

docs/intro_notebook.ipynb
tssweeney: right now this is actually broken with code capture :/

docs/intro_notebook.ipynb
tssweeney: set_env is general purpose exactly to address this. Do agree the comment is specific.

PR #1802, opened 2024-06-18
feat: add feedback tab to call details page

jamie-rasmussen:
Add a tab to the Call details page for displaying feedback.

Feedback:
Screenshot 2024-06-18 at 3 35 19 PM

No feedback case:
Screenshot 2024-06-18 at 3 35 06 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/wfDataModelHooksInterface.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/wfDataModelHooksInterface.ts
@@ -152,8 +152,6 @@ export type FeedbackKey = {
   entity: string;
   project: string;
   weaveRef: string;
-  // If true, do a prefix search on weaveRef instead of an exact match
-  includeExtra?: boolean;
 };
 
 export type WFDataModelHooksInterface = {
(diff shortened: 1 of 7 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/wfDataModelHooksInterface.ts
```diff
@@ -148,6 +148,14 @@ type AppendRefMutation = {
 
 export type RefMutation = SetRefMutation | AppendRefMutation;
 
+export type FeedbackKey = {
+  entity: string;
+  project: string;
+  weaveRef: string;
+  // If true, do a prefix search on weaveRef instead of an exact match
```
tssweeney, line 155: This field confuses me:
1. is seems like the name of the field implies the exact opposite of the comment
2. not sure when/why this is used
3. It doesn't seem to actually be used??

jamie-rasmussen: I'll remove this for now. In the prototype this code was extracted from you could attach feedback to a row in a table, and then when you are viewing the feedback tab for the table it would do a prefix search on the object ref instead of an exact match. This would allow it to display the feedback attached to the overall table but also surface the deeper feedback on rows.

I missed this detail when switching to the mongo-style query API. I'll probably put it back in but not until I get around to the PR displaying feedback on things other than calls, which will probably be a few PRs down the road.

tssweeney: i see! makes sense, thanks for the followup

tssweeney (approved):

PR #1743, opened 2024-06-10
feat(integration): add CrewAI integration

ayulockin:
This PR auto-patches the key methods of `Crew`, `Agent` and `Task` classes. These three classes together form an Agentic pipeline.

This PR utilises #1661 to trace underlying Langchain chains and LLM calls.

PR #1661, opened 2024-05-20
feat(integration): add langchain integration

parambharat:
This PR implements a Tracer using weave for langchain. 
Adds support for [WB-19624](https://wandb.atlassian.net/browse/WB-19624)

[WB-19624]: https://wandb.atlassian.net/browse/WB-19624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1804, opened 2024-06-18
feat(integration): update OpenAI integration to adhere to new design pattern

ayulockin:
This PR updates the existing OpenAI integration:
- to adhere to the new instrumentation design patter,
- to simplify usage token computation in the case of streaming using the newly introduced `stream_options={"include_usage": True}`.
- to get away with a lot of stream wrapping logic
- to unblock other integrations like #1743 and #1661

# Current usage patterns

## `Completion`

```
from openai import OpenAI
import weave

weave.init("test-openai")

client = OpenAI()

    model="gpt-4",
    messages=[
      {
        "role": "system",
        "content": "You are AGI. You will be provided with a message, and your task is to respond using emojis only."
      },
      {
        "role": "user",
        "content": "How are you?"
      }
    ],
    temperature=0.8,
    max_tokens=64,
    top_p=1,
)
print(response)
```

### `ChatCompletion` with `stream=True`

Add `stream=True` to the above `client.chat.completions.create`. This will not log usage tokens.

### `ChatCompletion` with `stream=True` and `stream_options={"include_usage": True}`

Add `stream=True` to the above `client.chat.completions.create` as well as `stream_options={"include_usage": True}`. It will log the usage token. 

❓ What should be the pattern here? Should we try to set `stream_options={"include_usage": True}` as a default if `stream=True`? Or we follow the OpenAI's pattern where they are expecting the user to set `stream_options={"include_usage": True}`.

## `AsyncCompletion`

```
import os
import asyncio
from openai import AsyncOpenAI

client = AsyncOpenAI()

async def main() -> None:
    chat_completion = await client.chat.completions.create(
        messages=[
            {
                "role": "user",
                "content": "Say this is a test",
            }
        ],
        model="gpt-3.5-turbo",
        stream=True,
        stream_options={"include_usage": True}
    )

    async for chunk in chat_completion:
        print(chunk)

asyncio.run(main())
```

This usage pattern also works with `stream=True` and to get usage tokens pass `stream_options={"include_usage": True}`.

## Function call and Tool call with the above configurations

Check out tests! Both `functions` argument (deprecated but frameworks like Langchain, LlamaIndex, etc, use it.) and `tool_calls` are supported.

# Questions and Todos

- [x] function calling and tool calling to be added/supported
- [x] The `op.name` is not setting the desired name, thus rendered name in the UI is `Completion.create` and `AsyncCompletion.create` instead of `openai.chat.completions.create`.
- [x] Add tests
- [x] Update docs
- [x] Fix OpenAI's [NOT_GIVEN](https://github.com/openai/openai-python/blob/f3e6e634a86d5789ab1274ae27f43adc842f4ba8/src/openai/_types.py#L134) issue.
- [x] Automatically set `stream_options={"include_usage": True}` when `stream=True` such that we track usage but adhere to user's usage of this flag.


weave/integrations/openai/cassettes/openai_test/test_openai_async_quickstart.yaml
```diff
@@ -0,0 +1,1951 @@
...
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate, br, zstd
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.31.0
+    method: GET
+    uri: https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json
```
tssweeney, line 14: In this, and all other cassettes, please remove the `litellm` block. This is will reduce these files immensely

weave/integrations/openai/openai_test.py
```diff
@@ -0,0 +1,705 @@
+import os
```
tssweeney, line 1: Top tier test suite .... chefs kiss

Please take a look at `weave/tests/test_monitoring_openai.py` ... these are our current tests for openai. It would be good to ensure that we cover everything tested in those

tssweeney (commented):
This looks like an incredible start. a few notes:

1. We should make sure that this plays nicely with the other integrations (i am sure it does)
2. Once the PR is fully complete, we should do a pixel-level comparison to current implementation to ensure that the UI is correct. In fact, what we should do is log to a project with current impl, then log the same with this. Then in the UI verify:
   1. There are no new Ops
   2. The outputs are exactly the same

weave/integrations/openai/openai_test.py
ayulockin: Checked! I think my tests are superset of `weave/tests/test_monitoring_openai.py`.

ayulockin:
Hey @tssweeney, I need help with the tests. There are two kind of tests that are failing:

1. 1st type has to do with the op name not set to `openai.chat.completion.create`.
2. 2nd type is related to JSON parsing on the `openai` side. Interestingly this is failing for all the tests where `stream` is not set.

I tried creating new cassettes but it didn't fix. I also tried to add `filter_body` but to no avail.

weave/integrations/litellm/litellm_test.py
```diff
@@ -146,7 +146,7 @@ def test_litellm_quickstart_stream(
         if chunk.choices[0].delta.content:
             all_content += chunk.choices[0].delta.content
 
-    exp = """Hello! I'm just a virtual assistant so I don't have feelings, but I'm here and ready to help you with anything you need. How can I assist you today?"""
```
tssweeney, line 149: why did the litellm tests change? seems like you can just revert all changes w.r.t lite llm, no?

weave/integrations/openai/openai_test.py
```diff
@@ -0,0 +1,773 @@
...
+    return r
+
+
+@pytest.mark.skip_clickhouse_client  # TODO:VCR recording does not seem to allow us to make requests to the clickhouse db in non-recording mode
+@pytest.mark.vcr(
+    filter_headers=["authorization"],
+    allowed_hosts=["api.wandb.ai", "localhost"],
+    before_record_request=filter_body,
+)
+def test_openai_quickstart(client: weave.weave_client.WeaveClient) -> None:
```
tssweeney, line 26: for all tests, can you please add assertions on: `op_name` and `finished_at` != None and `finished_at` < `started_at`

weave/integrations/openai/openai_sdk.py
```diff
@@ -0,0 +1,275 @@
...
+                        type=tool_call.type,
+                        function=Function(
+                            name=tool_call.function.name,
+                            arguments=tool_call.function.arguments,
+                        ),
+                    )
+                )
+        return _tool_calls
+
+    if isinstance(value, ChatCompletionChunk):
```
tssweeney, line 60: when is `value` NOT a ChatCompletionChunk? the input type should be a union in such cases

weave/integrations/openai/openai_sdk.py
```diff
@@ -0,0 +1,275 @@
...
+                }
+                for choice in value.choices
+            ],
+            created=value.created,
+            model=value.model,
+            object="chat.completion",
+            system_fingerprint=value.system_fingerprint,
+            usage=value.usage if hasattr(value, "usage") else None,
+        )
+        return final_value.model_dump(exclude_unset=True, exclude_none=True)
```
tssweeney, line 83: this dump is unfortunate. I look forward to the day that this is not needed

weave/trace/op_extensions/accumulator.py
```diff
@@ -15,6 +15,11 @@
 
 from weave.trace.op import FinishCallbackType, Op
 
+try:
```
tssweeney, line 18: I have an idea to make changes to this file much less invasive. For context, there are a lot of changes in this file just to manage the passing of around of the `skip_last` flag and managing what to do in such circumstances. This feels like we might be trying to do too much with this one function. Here is alternative proposal:

1. Change the `def add_accumulator` method's `accumulator` arg to `make_accumulator` which is a function that accepts the inputs and returns an accumulator function. Note: this will require changing a few other integrations to wrap their accumulator with a lambda... but it will be worth it i think.
2. From there, in your `openai_sdk` integration, you can make your accumulator raise a `StopIteration` exception when: a) you would normally skip last and b) you are processing a usage payload (which is always the last one).

Doing this will ensure that the functionality of this module is not overloaded.

weave/integrations/litellm/litellm_test.py
ayulockin: yeah I can change it back. They changed because I ran the test in record mode which changed the response slightly, hence the update. but I can revert it back.

weave/integrations/openai/openai_sdk.py
ayulockin: yeah ability to do it server side would be neater I guess.

weave/integrations/openai/openai_test.py
ayulockin: will do. didn't do it originally because the op name were not setting to the desired name. now that it's happening, adding these checks too.

weave/integrations/litellm/litellm_test.py
ayulockin: commented above.

weave/integrations/openai/openai_sdk.py
ayulockin: thanks for catching this.

weave/trace/op_extensions/accumulator.py
ayulockin: > Change the def add_accumulator method's accumulator arg to make_accumulator which is a function that accepts the inputs and returns an accumulator function.

This is a nice idea to remove the clutter of checking for one specific case (integration) in the `accumulator.py` file. I implemented it and it works fine.

> From there, in your openai_sdk integration, you can make your accumulator raise a StopIteration exception when: a) you would normally skip last and b) you are processing a usage payload (which is always the last one).

Say I raise `StopIteration` when `skip_last` is True and it the last item in the iterator with `usage` info. It will not return the said value (with usage info) to the user. I achieved this however, since we are not returning the final `acc` (that has usage info), it is not logged to weave. Need to handle this in an ugly way maybe.

ayulockin: Ignore, I was able to do it in an elegant way. TIL `StopIteration(value)` is a thing.

ayulockin: You rock <3

ayulockin:
Hey @tssweeney what would be your suggestion to `raise StopAsyncIteration`?

ayulockin:
Links to possible scenarios:

* Simple Quickstart: [weave](https://wandb.ai/wandbot/test-openai/weave/calls?filter=%7B%22opVersionRefs%22%3A%5B%22weave%3A%2F%2F%2Fwandbot%2Ftest-openai%2Fop%2Fopenai.chat.completions.create%3A*%22%5D%7D&peekPath=%2Fwandbot%2Ftest-openai%2Fcalls%2Fa331408f-3e16-4fcb-bcbe-c10dcc3d36ac) | `stream=True` but no `stream_options` set, we expect the usage to be captured and iterator stopping at the 2nd last item.
* Simple Quickstart Async: [weave](https://wandb.ai/wandbot/test-openai/weave/calls?filter=%7B%22opVersionRefs%22%3A%5B%22weave%3A%2F%2F%2Fwandbot%2Ftest-openai%2Fop%2Fopenai.chat.completions.create%3A*%22%5D%7D&peekPath=%2Fwandbot%2Ftest-openai%2Fcalls%2Fe5e85fef-f9cf-4e83-a30a-3da643452f47) | same as above.
* Simple Quickstart with `stream_options` set: [weave](https://wandb.ai/wandbot/test-openai/weave/calls?filter=%7B%22opVersionRefs%22%3A%5B%22weave%3A%2F%2F%2Fwandbot%2Ftest-openai%2Fop%2Fopenai.chat.completions.create%3A*%22%5D%7D&peekPath=%2Fwandbot%2Ftest-openai%2Fcalls%2Fb5342a7e-75f4-435e-b0a4-5b1960ce898c) | notice `stream_options` in the UI
* Simple Quickstart Async with `stream_options` set: [weave](https://wandb.ai/wandbot/test-openai/weave/calls?filter=%7B%22opVersionRefs%22%3A%5B%22weave%3A%2F%2F%2Fwandbot%2Ftest-openai%2Fop%2Fopenai.chat.completions.create%3A*%22%5D%7D&peekPath=%2Fwandbot%2Ftest-openai%2Fcalls%2F8611a68e-463f-43b2-9eee-0a073fe3ca99)
* Function call with `stream_options` set in sync: [weave](https://wandb.ai/wandbot/test-openai/weave/calls?filter=%7B%22opVersionRefs%22%3A%5B%22weave%3A%2F%2F%2Fwandbot%2Ftest-openai%2Fop%2Fopenai.chat.completions.create%3A*%22%5D%7D&peekPath=%2Fwandbot%2Ftest-openai%2Fcalls%2F4c7db363-a826-4dfd-a439-b60e02c07adf) | notice functions in the input; function_call in the output message body
* Function call in async mode without `stream_options`: [weave](https://wandb.ai/wandbot/test-openai/weave/calls?filter=%7B%22opVersionRefs%22%3A%5B%22weave%3A%2F%2F%2Fwandbot%2Ftest-openai%2Fop%2Fopenai.chat.completions.create%3A*%22%5D%7D&peekPath=%2Fwandbot%2Ftest-openai%2Fcalls%2F38c30153-a880-4b11-bf00-69b865e4f20a)
* Tool call sync ([weave](https://wandb.ai/wandbot/test-openai/weave/calls?filter=%7B%22opVersionRefs%22%3A%5B%22weave%3A%2F%2F%2Fwandbot%2Ftest-openai%2Fop%2Fopenai.chat.completions.create%3A*%22%5D%7D&peekPath=%2Fwandbot%2Ftest-openai%2Fcalls%2F5b2b09c9-0d19-4d35-a7a2-5ef7d763c133)) and async ([weave](https://wandb.ai/wandbot/test-openai/weave/calls?filter=%7B%22opVersionRefs%22%3A%5B%22weave%3A%2F%2F%2Fwandbot%2Ftest-openai%2Fop%2Fopenai.chat.completions.create%3A*%22%5D%7D&peekPath=%2Fwandbot%2Ftest-openai%2Fcalls%2F29974bdc-d23b-4357-a75c-949db89bd965)) | notice mention of tool in the UI

weave/integrations/litellm/cassettes/litellm_test/test_litellm_quickstart_stream_async.yaml
```diff
@@ -1,7 +1,7 @@
 interactions:
 - request:
     body: '{"messages": [{"content": "Hello, how are you?", "role": "user"}], "model":
-      "gpt-3.5-turbo", "stream": true}'
+      "gpt-3.5-turbo-0125", "stream": true, "stream_options": {"include_usage": true}}'
```
tssweeney, line 4: again, seems like these changes to cassettes can be removed

weave/trace/op_extensions/accumulator.py
```diff
@@ -134,9 +143,9 @@ def __init__(
 
         atexit.register(weakref.WeakMethod(self._call_on_close_once))
 
-    def _call_on_close_once(self) -> None:
+    def _call_on_close_once(self, e: Optional[S] = None) -> None:
```
tssweeney, line 146: i don't think you use `e: Optional[S] = None` anymore

weave/integrations/openai/openai_test.py
tssweeney: Can you please add these assertions (:

tssweeney (changes requested):
This is sooooo close, a few notes:
1. Please revert your LLM cassettes
2. Please remove the unused params
3. Please add assertions on op_name and timestamps

weave/integrations/openai/openai_test.py
```diff
@@ -0,0 +1,798 @@
...
+
+    res = client.server.calls_query(tsi.CallsQueryReq(project_id=client._project_id()))
+    assert len(res.calls) == 1
+
+    exp = '{"name":"Sachin Tendulkar","team":"India","highest_score":248}'
+    assert response.choices[0].message.tool_calls[0].function.arguments == exp
+
+    call = res.calls[0]
+    assert op_name_from_ref(call.op_name) == "openai.chat.completions.create"
+    assert call.started_at is not None
```
tssweeney, line 589: add `ended_at is not None` as well

tssweeney (approved):
Very amazing job (:

parambharat:
@ayulockin  and @tssweeney: I wanted to inform you that the patching in the integration affects streaming when we use the Openai client as a context manager.  

I merged this branch into the [Langchain Integration](https://github.com/wandb/weave/pull/1661) and noticed that streaming with the context manager breaks. 

Here are failing tests in the Langchain Integration
1. [test_simple_chain_stream](https://github.com/wandb/weave/blob/429bcf0058ea4c4dfc1246e8a198ecc0c4b521c7/weave/integrations/langchain/langchain_test.py#L80)
2. [test_agent_run_with_tools](https://github.com/wandb/weave/blob/429bcf0058ea4c4dfc1246e8a198ecc0c4b521c7/weave/integrations/langchain/langchain_test.py#L313)
3. [test_agent_run_with_function_call](https://github.com/wandb/weave/blob/429bcf0058ea4c4dfc1246e8a198ecc0c4b521c7/weave/integrations/langchain/langchain_test.py#L408)

Also, here's a minimal reproducible example test for the Openai client.

```
def test_simple_openai_streaming() -> None:
    from openai import OpenAI
    import weave

    client = OpenAI()

    # # This works
    # weave.init("test_openai_stream")
    # chat = client.chat.completions.create(
    #     model="gpt-3.5-turbo",
    #     messages=[{"role": "user", "content": "Hello, how are you?"}],
    #     stream=True,
    # )
    # for response in chat:
    #     print(response)

    # This raises an error when we patch the OpenAI client
    weave.init("test_openai_stream")
    with client.chat.completions.create(
        model="gpt-3.5-turbo",
        messages=[{"role": "user", "content": "Hello, how are you?"}],
        stream=True,
    ) as chat:
        for response in chat:
            print(response)


test_simple_openai_streaming()
```

ayulockin:
Thanks for reporting @parambharat. I am taking a look.

ayulockin:
This piece of code with or without `stream=True` doesn't work as the patched `client.chat.completions.create` is no more a context manager

```
weave.init("test_openai_stream")
with client.chat.completions.create(
    model="gpt-3.5-turbo",
    messages=[{"role": "user", "content": "Hello, how are you?"}],
    stream=True,
) as chat:
    for response in chat:
        print(response)
```

 as evident by the error:
 
 ```
  File "/Users/ayushthakur/integrations/weave-crewai/weave/trash.py", line 336, in <module>
    test_simple_openai_streaming()
  File "/Users/ayushthakur/integrations/weave-crewai/weave/trash.py", line 327, in test_simple_openai_streaming
    with client.chat.completions.create(
AttributeError: __enter__
```

File: weave/integrations/openai/openai_sdk.py [hunk 11]
```diff
@@ -0,0 +1,286 @@
...
+                acc, value, skip_last=not _openai_stream_options_is_set(inputs)
+            ),
+            should_accumulate=should_use_accumulator,
+            on_finish_post_processor=openai_on_finish_post_processor,
+        )
+
+    return wrapper
+
+
+symbol_pathers = [
```
parambharat, line 272: @ayulockin minor nit: typo here -> should be `symbol_patchers` instead of `symbol_pathers`

PR #1813, opened 2024-06-20
docs(best-practices) Some recipes that are useful 

tcapelle:
- Instrumenting with weave
- Weave and Pydantic: Model, Object, etc...
  - Using `model_validator` and `Field`, for delayed attributes
  - Example of using this for `client` and a `HF` model.


docs/docs/best-practices.md
```diff
@@ -0,0 +1,168 @@
...
+class LLMModel(weave.Model):
+    model_name: str
+    prompt_template: str
+    client: openai.AsyncClient = Field(init=False)
+    
+    # highlight-next-line
+    @model_validator(mode="before")
+    def create_client(cls, values: dict):
+        values['client'] = openai.AsyncClient()
+        return values
```
andrewtruong, line 100: I think in general we should discourage using something like `@model_validator` since it's more likely to break things.

This might be more simply expressed as:
```
_client: openai.AsyncClient = PrivateAttr(default_factory=openai.AsyncClient)
```

docs/docs/best-practices.md
```diff
@@ -0,0 +1,168 @@
...
+    def load_model_and_tokenizer(cls, v):
+        "Pydantic validator to load the model and the tokenizer"
+        model_id = v["model_id"]
+        if model_id is None:
+            raise ValueError("model_id is required")
+        model = AutoModel.from_pretrained(model_id)
+        tokenizer = LlamaTokenizerFast.from_pretrained(model_id)
+        v["model"] = model
+        v["tokenizer"] = tokenizer
+        return v
```
andrewtruong, line 142: You might want to use this instead:
https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_post_init

tcapelle: @NiWaRe look at this =)

docs/docs/best-practices.md
NiWaRe: @andrewtruong how would you give people the possibility of having a constructor? If people don't want to pass in the finished objects to the model the question is what pydantic option to use to allow users some pydantic-compatible constructor to create their specific model attributes (either just a client, or an actual loaded in model, RAG model, etc.)?

andrewtruong: I'm not sure I understand what you mean by this.  Can you give me an example?

docs/docs/best-practices.md
NiWaRe: @andrewtruong I played around with different options in order to have this constructor-like function to create the model objects: 
- @model_validator(mode='before')
- @model_validator(mode='after')
- model_post_init
- working with @property functions without pydantic
- calling the super constructor after adding a new model attribute in the __init__

What you suggest `model_post_init` gives the most intuitive naming for its purpose, works with fully init attributes (they can be called with self instead of by the values dict),  and feels more pydantic than the last two options I mentioned - do you see any other benefits in using model_post_init? Would you set this as a general best practice for a constructor init of attributes?

docs/docs/best-practices.md
NiWaRe: The code in the below thread for example - instead of passing in the ready model object you pass in the model name and init the model object dynamically as an attribute (either through any openai, anthropic, litellm client or as an actual local model you load in).

docs/docs/best-practices.md
tcapelle: This seems to work:
```python
class LlamaModel(weave.Model):
    """A model class for MetaAI-LLama models"""
    model_id: str
    temperature: float = 0.5
    max_new_tokens: int = 128
    model: LlamaForCausalLM = None
    tokenizer: LlamaTokenizerFast = None

    def model_post_init(self, __context):
        "Pydantic validator to load the model and the tokenizer"
        self.model = LlamaForCausalLM.from_pretrained(model_id)
        self.tokenizer = LlamaTokenizerFast.from_pretrained(model_id)

```

andrewtruong: Yes, just use `model_post_init`.

For this case, I would write:
```py
class LlamaModel(weave.Model):
    """A model class for MetaAI-LLama models"""
    model_id: str
    temperature: float = 0.5
    max_new_tokens: int = 128

    _model: LlamaForCausalLM = PrivateAttr()
    _tokenizer: LlamaTokenizerFast = PrivateAttr()

    def model_post_init(self, __context):
        self._model = LlamaForCausalLM.from_pretrained(model_id)
        self._tokenizer = LlamaTokenizerFast.from_pretrained(model_id)
```

docs/docs/best-practices.md
andrewtruong: Yep see the example below

docs/docs/best-practices.md
tcapelle: lemme try 😋

NiWaRe: This seems a bit verbose but it seems like a workflow many people could follow. The main question for me is whether we want to have separate private attributes that are meant to be left empty (in which case we just use @andrewtruong 's code above - maybe with an additional default to None) or whether users find it easier to have only model attribute that will either already be the model object or will be turned into the model object by the post init.

```python
from transformers import PreTrainedModel, PreTrainedTokenizerBase

class FineTunedModelHF(weave.Model): 
    model: Optional[Union[PreTrainedModel, str]]
    tokenizer: Optional[Union[PreTrainedTokenizerBase, str]]
    
    def model_post_init(self, __context: Any) -> None:
        if isinstance(self.model, str):
            if self.model.startswith("weave://"):
                model_ref = weave.ref(vector_store)
                self.model = model_ref.get()
            else:
                 self.model = AutoModel.from_pretrained(model_id)
 
     # same for tokenizer
     # ....
```

tcapelle: Model checkpoints are huge, so I don't feel to push people to store them on W&B

NiWaRe: @tcapelle if you don't add a serializer then the models will not be stored on Weave (will only show the __str__) - or what do you mean? Definitely wouldn't push people to store their foundation models in Weave.

andrewtruong: @NiWaRe 
1. Good idea -- the generic versions should work for all models
2. I think we should discourage the `Union[<complex_type>, str]` pattern because it mixes config (`str`) with the actual object (`<complex_type>`).  Also, I'm not sure if we want to be able to json-dump `weave.Model`, but if we did it wouldn't be easy for these complex types.

NiWaRe: Hi @tcapelle @andrewtruong , a couple of updates that came up during a recent feedback project I'm working on with Scott (see #1873): 

- I agree that not all complex_type objects should be serializable for a better display in the UI or for a saving of the file. It also might be confusing to promote the Union type as you said Andrew, so I tried to follow your suggestion for a new `WeaveVectorStore` class (see [here](https://github.com/wandb/weave/blob/fa753a5ab546ad7632ac8ab85cc2b3b28318d755/examples/serious_rag/weave_utils.py#L155)) and marked the actual faiss Index as `_index` and used the `model_post_init` function to set it during initialization. The problem now is that `_index` doesn't show up in my Weave UI and will also not be stored for me to retrieve it again. 
- Removing the underscore and just working with `model_post_init` will also not work because then Weave will save the `index` variable with a `None` value (the default value).
- For some reason using the `model_validator(mode="after")` will not change index at all. 

What would you say should our guideline be here? The simplest thing would probably be that custom_type objects can't be visualized nor stored as arguments in Weave but that custom weave.Objects would need to be created for a serialized view and object would need to be published as weave.Objects in order to save them (e.g. for the index above)? What will happen if they still do pass in custom_type objects without defining it as private?

J2-D2-3PO:
@tcapelle doing some drivebys of open docs PRs. I also notice this created a new page that doesn't yet have a location in `sidebar.ts`. Anything I can help with here? Figuring out menu set up, copy review once ready maybe?

PR #1814, opened 2024-06-20
chore: add more ruff rules (no current violations)

jamie-rasmussen:
These are docstring related rules. Turning on some that are currently not violated and adding comments that explain why some are explicitly not enabled.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,17 @@ filterwarnings = ["error"]
 preview = true
 select = [
   "D200", # https://docs.astral.sh/ruff/rules/fits-on-one-line/
+  "D201", # https://docs.astral.sh/ruff/rules/no-blank-line-before-function/
+  # "D203", # https://docs.astral.sh/ruff/rules/one-blank-line-before-class/ - skipped, we use Google style
+  "D204", # https://docs.astral.sh/ruff/rules/one-blank-line-after-class/
+  # "D206", # https://docs.astral.sh/ruff/rules/indent-with-spaces/ - redundant with formatter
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -67,6 +67,17 @@ filterwarnings = ["error"]
 preview = true
 select = [
   "D200", # https://docs.astral.sh/ruff/rules/fits-on-one-line/
+  "D201", # https://docs.astral.sh/ruff/rules/no-blank-line-before-function/
+  # "D203", # https://docs.astral.sh/ruff/rules/one-blank-line-before-class/ - skipped, we use Google style
+  "D204", # https://docs.astral.sh/ruff/rules/one-blank-line-after-class/
+  # "D206", # https://docs.astral.sh/ruff/rules/indent-with-spaces/ - redundant with formatter
+  # "D207", # https://docs.astral.sh/ruff/rules/under-indentation/ - redundant with formatter
+  # "D208", # https://docs.astral.sh/ruff/rules/over-indentation/ - redundant with formatter
```
andrewtruong, line 75: Should we remove these commented out ones, or move them to exclude?

tssweeney (approved):

pyproject.toml
jamie-rasmussen: My thinking was that keeping them as comments in sorted order with the enabled ones would make it easier to see remaining gaps in what rules we have investigated, but I could also move them into ignore if you think that would be less messy.

andrewtruong: yeah I would keep the comments but move them to exclude, that way it's more explicit what we're intentionally not using

jamie-rasmussen: ok I'll do that

PR #1831, opened 2024-06-25
feat(weave): Add `call` func to decorated op to get back a Call object

andrewtruong:
This PR adds a `call` method to ops that can be used to:
1. Literally "call" the func; and
2. Return a `Call` object back (useful for adding feedback)


weave/trace/op.py
```diff
@@ -140,6 +149,17 @@ async def _run_async() -> Coroutine[Any, Any, Any]:
...
             return on_output(res)
 
+    def call(self, *args: Any, **kwargs: Any) -> "Call":
+        client = client_context.weave_client.require_weave_client()
+        run = self._create_run(*args, **kwargs)
+        self._execute_run(run, *args, **kwargs)
+        return self._retry_fetch_call(client, run.id)
+
+    @staticmethod
+    @retry
```
jamie-rasmussen, line 159: I think this will retry forever with no backoff? We should probably limit it somehow.

weave/trace/op.py
```diff
@@ -140,6 +149,17 @@ async def _run_async() -> Coroutine[Any, Any, Any]:
         else:
             return on_output(res)
 
+    def call(self, *args: Any, **kwargs: Any) -> "Call":
+        client = client_context.weave_client.require_weave_client()
+        run = self._create_run(*args, **kwargs)
```
jamie-rasmussen, line 154: Would be nice to not perpetuate the "run" terminology

andrewtruong: let's do a second rename PR that changes it to call

jamie-rasmussen (approved):

PR #1832, opened 2024-06-25
fix(weave): Fix evals (and remove WeaveList dep)

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19041


shawnlewis (approved):
LGTM if you think there is good enough test coverage to show that behavior remains the same under this change.

PR #1833, opened 2024-06-25
fix(weave): Dont add extra decorator unnecessarily when calling deco factory without parens

andrewtruong:
This PR fixes codegen so that it doesn't add an extra decorator when using the `weave.op` (no parens) decorator factory.

Previously, we would do:
```py
@weave.op()
@weave.op
def func(...): ...
```

Now it will correctly be:
```py
@weave.op
def func(...): ...
```


weave/trace/op_type.py
```diff
@@ -19,6 +20,9 @@
 from . import serializer
 from .op import Op
 
+weave_op_pattern = re.compile(r"@weave\.op(\(\))?")
```
jamie-rasmussen, line 23: These are global constants so I'd use capital `WEAVE_OP_PATTERN`

jamie-rasmussen (approved):

jamie-rasmussen (commented):
We'll have to revisit this if we add any arguments to the decorator.

PR #1846, opened 2024-06-26
feat: Cohere chat models integration

tcapelle:
Cohere API integration with Weave


tcapelle:
@tssweeney I need help with 2 things:
- the output of the cohere SDK is a `pydantic.v1.BaseModel` that we don't parse correctly
- I think the wrapped function in cohere SDK patches after the weave patch, and that's why the name is ignored. For the streaming version the naming is correctly displayed, I suppose the accumulator fixes the naming.

tcapelle:
The only thing that may need changing is how the patching is applied.

As you know @tssweeney the cohere Client modifies the chat function on the init, so we need to patch before creating a client. Two options for this:
- We call `weave.init` before creating the client (this is weird in general and wouldn't be how most of of the users do, but... it is all over our docs: anthropic, mistral, etc...)
- We manually patch before creating a client. Something closer to where the imports happen.

.gitignore
```diff
@@ -27,4 +27,5 @@ weave/tests/zlib.txt
 weave/tests/execute.json
 weave-js/src/generated
 .integration-deps
+cape_dev/
```
tssweeney, line 30: what is this?

docs/docs/guides/integrations/cohere.md
```diff
@@ -0,0 +1,123 @@
+---
```
tssweeney, line 1: @scottire may want to review docs as well

weave/integrations/cohere/cohere_test.py
```diff
@@ -0,0 +1,232 @@
...
+    )
+
+    exp = response.text
+    assert exp.strip() != ""
+    res = client.server.calls_query(tsi.CallsQueryReq(project_id=client._project_id()))
+    assert len(res.calls) == 1
+    call = res.calls[0]
+    assert call.exception is None and call.ended_at is not None
+    output = _get_call_output(call)
+    assert output.text == exp
```
tssweeney, line 49: please also add assertions for: `started_at < ended_at`, `ended_at != None`, `exception == null`, and `op_name` are all correct. This comment applies to all tests

weave/integrations/file::memory:?cache=shared
tssweeney, line 1: what is this??

tssweeney (changes requested):
Looks really good!! (: Only a few comments to tend to

tssweeney (approved):

PR #1848, opened 2024-06-26
feat: make get_current_call a top level API

jamie-rasmussen:


weave/call_context.py
```diff
@@ -64,6 +64,13 @@ def pop_call(call_id: typing.Optional[str]) -> None:
 
 
 def get_current_call() -> typing.Optional["Call"]:
+    """Get a reference to the currently running call.
+
+    This can be used to access the call's id, feedback, etc.
```
shawnlewis, line 69: I think this line in the docstring is going to mislead people a bit. This is not the most useful API if what you want is to add feedback to call after calling it. We should point the user to op.call for that case. In other words, I think the op.call style is going to be more frequently used than this. But this is a top-level API, at a higher depth within the API reference docs. So we should probably use this to point people at the other one.

jamie-rasmussen: Docstring updated. We can revisit once I get around to adding cross-linking support to genpydocs.

andrewtruong (approved):

PR #1861, opened 2024-06-27
fix(weave): Move metadata into its own field to avoid collisions with other pkgs

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19388

This PR moves metadata that was previously top level (`name`, `description`) to a new top-level `metadata` field which prevents collision with other libraries that also have `name` as a top-level field (e.g. `openai`)

### DIFF
--- a/weave/tests/test_graph_client.py
+++ b/weave/tests/test_graph_client.py
@@ -37,7 +37,7 @@ def test_flask_server(flask_server):
     url = flask_server
     response = requests.get(url)
     assert response.status_code == 200
-    assert response.text == "T41nPvr6zpAr4ClFpZeH1p4xXE2x2bH2JeUXyOsQ4nY"
+    assert response.text == "PHOGkwSOn7DqLgIUNgUAq7d2vXpOmG8NGLltn6slzeU"
 
 
 def test_graph_client_global_accessible_in_thread(client):
--- a/weave/tests/test_graph_client.py
+++ b/weave/tests/test_graph_client.py
@@ -40,7 +40,7 @@ def test_flask_server(flask_server):
     assert response.text == "T41nPvr6zpAr4ClFpZeH1p4xXE2x2bH2JeUXyOsQ4nY"
 
 
-def test_weave_client_global_accessible_in_thread(client):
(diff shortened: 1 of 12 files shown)


tssweeney (approved):
Primary comment is that this should be under a different namespace - maybe `weave_meta` or something like that. Perhaps even underscored

weave/tests/test_graph_client.py
```diff
@@ -0,0 +1,59 @@
+import queue
```
tssweeney, line 1: accidentally included?

weave/tests/test_weave_client.py
```diff
@@ -288,7 +288,7 @@ async def score(target, model_output):
         return target == model_output
 
     evaluation = Evaluation(
-        name="my-eval",
+        metadata={"name": "my-eval"},
```
tssweeney, line 291: this feels very odd to me - does it feel odd to you?

andrewtruong: Why?

weave/tests/test_graph_client.py
andrewtruong: this branch is super outdated

tssweeney (changes requested):
Remove unused code before landing

weave/tests/test_weave_client.py
tssweeney: well, as a user, i am thinking to myself: i want to name this evaluation, why is it some nested dictionary? It feels like the underlying data model is leaking into user-land. If we didn't have this name collision issue, we would not choose this API. There might not be a good alt, just want to discuss / see if there are other ideas

andrewtruong: It feels right to me.  This dict format is just taking advantage of pydantic so you don't need to use the class.  You could alternatively write:
```python
metadata=Metadata(name="my-eval")
```

That said, I wonder if having a very general metadata dict is also useful -- basically simple k,v tags for objects

andrewtruong: For `Evaluation` specifically, we could also have `name` as the top-level then forward the value to `metadata`

Issue #1809, opened 2024-06-19
OSError: source code not available.


masterbatcoderman10:
I'm trying to run a class method with the decorator added to it, and while it works fine sometimes out of a sudden it is prone to throwing an `OSError: source code not available`.

This error only occurs with the latest version of `weave`, 0.50.5.

jamie-rasmussen:
Thanks for reporting this! If there is a full stack trace could you post it here? And could you tell us a little more information about your environment, e.g. Python version and whether you are running in a notebook or not?

tssweeney:
Adding internal ticket for bug tracker: https://wandb.atlassian.net/browse/WB-19479

andrewtruong:
Hi @masterbatcoderman10 , this should be addressed with the above PR!

PR #1865, opened 2024-06-30
fix(weave): Add better handling for functions missing source

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19479
- https://wandb.atlassian.net/browse/WB-18719
- https://github.com/wandb/weave/issues/1809

This PR provides better handling for functions that are missing source (e.g. if they were `exec`'d into existence).  Instead of crashing the user script, we pass the call and put a placeholder saying we weren't able to save their code.

Some improvements from here:
1. Add a reason why we couldn't save it
2. Promote "anonymous op" (not sure if we need this)


weave/tests/test_exec.py
```diff
@@ -0,0 +1,16 @@
...
+        import weave
+        
+        @weave.op()
+        def add(a: int, b: int) -> int:
+            return a + b
+
+        ref = weave.publish(add)
+        """
+    )
+    exec(code)
```
andrewtruong, line 16: is there a way to get the "saved op code" back?

weave/trace/op_type.py
```diff
@@ -271,7 +298,7 @@ def get_code_deps(
                 import_code += fn_import_code
 
                 code += fn_code
-                code.append(textwrap.dedent(inspect.getsource(var_value)))
+                code.append(get_source_notebook_safe(var_value))
```
shawnlewis, line 301: This line can still throw?

weave/tests/test_exec.py
shawnlewis: You can do this: https://github.com/wandb/weave-analysis/blob/7d5ea67862f610ea220c547430bc25baad2a08c3/robot.py#L120.

Not a very nice API yet...

weave/trace/ipython.py
```diff
@@ -0,0 +1,55 @@
...
+    return full_source
+
+
+def get_class_source(cls: Callable) -> str:
+    """Get the latest source definition of a class in the notebook."""
+    notebook_source = get_notebook_source()
+    tree = ast.parse(notebook_source)
+    class_name = cls.__name__
+
+    # We need to walk the entire tree and get the last one since that's the most version of the cls
```
shawnlewis, line 46: This could certainly be incorrect if there is more than one definition of the same class name. We should have some token in the code comment that we can use to indicate this like "WARNING:"

weave/trace/op_type.py
```diff
@@ -224,7 +251,7 @@ def get_code_deps(
         ]
         return {"import_code": [], "code": [], "warnings": warnings}
 
-    source = textwrap.dedent(inspect.getsource(fn))
+    source = get_source_or_fallback(fn)
```
shawnlewis, line 254: This is not just called on ops, but if it hits the fallback, it will make functions called "op_fn".

So this whole algorithm can generate a bunch of code like this

```
def op_fn(*args, **kwargs):  # type: ignore
    # Code-capture unavailable for this op
    pass

def op_fn(*args, **kwargs):  # type: ignore
    # Code-capture unavailable for this op
    pass
    
...
```

Where we repeat the same op_fn block over and over again. 

I think we need to name these "blank functions" what the function we're trying to save would have been called.

You should write more tests to show this behavior.

shawnlewis (changes requested):

weave/trace/ipython.py
andrewtruong: I think this is safe because this inspection is happening at publish time.  The flow is roughly:
```py
class A(weave.Object):
  a: str
class A(weave.Object):
  b: str
a = A()
weave.publish(a)  # This will publish A:v0 with b: str which is what I would expect
```

This is also expected:
```py
class A(weave.Object):
  a: str
a = A()
weave.publish(a)  # Publishes A:v0 with a: str

class A(weave.Object):
  b: str
a = A()
weave.publish(a)  # Publishes A:v1 with b: str
```

andrewtruong: I think this is mostly safe because this inspection is happening at publish time.  The flow is roughly:
```py
[In:1]
class A(weave.Object):
  a: str

[In:2]
class A(weave.Object):
  b: str
a = A()
weave.publish(a)  # This will publish A:v0 with b: str which is what I would expect
```

This is also expected:
```py
[In: 1]
class A(weave.Object):
  a: str
a = A()
weave.publish(a)  # Publishes A:v0 with a: str

[In: 2]
class A(weave.Object):
  b: str
a = A()
weave.publish(a)  # Publishes A:v1 with b: str
```

There may be this gotcha:
```py
[In: 1]
class A(weave.Object):
  a: str
a = A()

[In: 2]
weave.publish(a)  # Maybe Publishes A:v0 with b: str

class A(weave.Object):
  b: str
a = A()
weave.publish(a)  # No publish since duplicate?
```

weave/trace/op_type.py
```diff
@@ -200,6 +200,85 @@ def get_source_notebook_safe(fn: typing.Callable) -> str:
...
+def get_source_or_fallback(fn: typing.Callable) -> str:
+    if isinstance(fn, Op):
+        fn = fn.resolve_fn
+
+    func_name = fn.__name__
+    try:
+        sig_str = reconstruct_signature(fn)
+    except Exception as e:
+        print(f"Failed to reconstruct signature {e=}")
+        sig_str = "(*args, **kwargs)"
```
shawnlewis, line 268: why is this is a separate var instead of inline in the template string?

shawnlewis (approved):
Approved with comment!

weave/trace/op_type.py
andrewtruong: I think it's more readable.  Right now you can read as:

"Try to reconstruct the signature as a string.  If Exception, use args kwargs"

PR #1866, opened 2024-06-30
fix(weave): Handle exceptions raised by scoring funcs during evaluation

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-18523

This PR changes `Evaluation.evaluate` to not crash the user script if an exception is raised in a scoring function.  Instead, the error is captured and relayed to the user as part of the eval

### DIFF
--- a/weave/flow/eval.py
+++ b/weave/flow/eval.py
@@ -18,7 +18,6 @@ from weave.flow.scorer import (
     Scorer,
     auto_summarize,
     get_scorer_attributes,
-    new_auto_summarize,
     transpose,
 )
 from weave.trace.env import get_weave_parallelism
@@ -256,7 +255,7 @@ class Evaluation(Object):
                     scored = summarize_fn(score_table)
                     summary[scorer_name] = scored
             else:
-                model_output_summary = new_auto_summarize(vals)
+                model_output_summary = auto_summarize(vals)
                 if model_output_summary:
                     summary[name] = model_output_summary
 
--- a/weave/flow/eval.py
+++ b/weave/flow/eval.py
@@ -18,6 +18,7 @@ from weave.flow.scorer import (
     Scorer,
     auto_summarize,
     get_scorer_attributes,
+    new_auto_summarize,
     transpose,
 )
 from weave.trace.env import get_weave_parallelism
@@ -255,7 +256,7 @@ class Evaluation(Object):
                     scored = summarize_fn(score_table)
                     summary[scorer_name] = scored
             else:
-                model_output_summary = auto_summarize(vals)
+                model_output_summary = new_auto_summarize(vals)
                 if model_output_summary:
                     summary[name] = model_output_summary
 
--- a/weave/flow/eval.py
+++ b/weave/flow/eval.py
@@ -234,7 +234,16 @@ class Evaluation(Object):
                 raise OpCallError(message)
             # this is a catchall in case the user's scorer function throws
             except Exception as e:
-                result = {"error": str(e), "example": example}
+                exc_type = type(e).__name__
+                exc_val = str(e)
+                exc_traceback = traceback.format_exc()
+
+                result = {
+                    "exception_type": exc_type,
+                    "exception_value": exc_val,
+                    "exception_traceback": exc_traceback,
+                    "example": example,
+                }
             scores[scorer_name] = result
 
         return {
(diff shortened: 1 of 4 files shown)


weave/flow/eval.py
```diff
@@ -248,6 +248,9 @@ async def predict_and_score(
                     """
                 )
                 raise OpCallError(message)
+            # this is a catchall in case the user's scorer function throws
+            except Exception as e:
```
shawnlewis, line 252: This makes debugging really bad. Tracebacks will no longer be printed to the console.

weave/flow/eval.py
```diff
@@ -248,6 +248,9 @@ async def predict_and_score(
...
+                result = {"error": str(e), "example": example}
```
shawnlewis, line 253: Don't reproduce the whole example here. That's going to really mess up with the auto-columns we show in the UI.

Can you make a realistic example with an LLM judge that tries to json.loads() an LLM call response, and has like 50 examples in the dataset? Those fail sporadically. Then look in the UI to understand the impact this change has on the user experience.

weave/flow/eval.py
```diff
@@ -268,8 +271,14 @@ async def summarize(self, eval_table: list) -> dict:
                     scorer_name, _, summarize_fn = get_scorer_attributes(scorer)
                     scorer_stats = transpose(vals)
                     score_table = scorer_stats[scorer_name]
-                    scored = summarize_fn(score_table)
-                    summary[scorer_name] = scored
+
+                    if errs := [
+                        s for s in score_table if isinstance(s, dict) and "error" in s
+                    ]:
+                        summary[scorer_name] = {"errors": errs}
```
shawnlewis, line 278: This reproduces the list of errors that we've already logged in each predict_and_score call. Since we're using the trace itself as the data model for evaluations, I don't think we should do this.

shawnlewis (changes requested):

weave/flow/eval.py
andrewtruong: What would you do differently?

Some alternatives:
1. Add a debug flag to let the user decide if they want the eval to raise (will stop the eval)
2. Save the whole traceback instead of just `str(e)`

weave/flow/eval.py
andrewtruong: I think this has some nice properties actually:
1. In the outputs, errors link to the exact rows that broke making debugging easier
2. Errors show as N/A in the UI which seems clean to me

image
image

weave/flow/eval.py
andrewtruong: See comment above -- I think this has nice properties

weave/flow/eval.py
shawnlewis: We need to at least print the whole traceback to the console

weave/flow/scorer.py
```diff
@@ -46,20 +46,24 @@ def auto_summarize(data: list) -> dict[str, Any] | None:
...
         return {}
-    data = [x for x in data if x is not None]
+
     val = data[0]
 
     if isinstance(val, (bool, box.BoxedBool)):
         return {
             "true_count": (true_count := sum(1 for x in data if x)),
-            "true_fraction": true_count / len(data),
+            "true_proportion": true_count / len(data),
```
shawnlewis, line 55: You can't go changing the term here, that will break comparison to all the old evals people have done.

_fraction is also a better choice IMO, I've used that for a long time for this type of thing and people understand it. Originally recommended to me by a principle engineer at Google as the "right term" for this :)

weave/flow/eval.py
andrewtruong: It would be super noisy if you had an exception in a scorer function and had many examples though.

Imagine a 100 line trace * 100 examples -- very noisy and could lag if you print to console!

shawnlewis: You have to show developers tracebacks in the console if you want them to be able to fix them.

I tried this before with swallowing exceptions. It sucks. I need to know the exact traceback and line of code a problem came from, when I'm developing a scorer.

Try developing your own scorer function without it!

andrewtruong: But you will see the traceback in the UI...

Let's lay out a few things that we certainly want and don't want to narrow the solution space:
1. The user definitely needs to see a traceback somewhere to figure out how to fix their error.
2. We definitely can't print a large volume of tracebacks to the console because it will be extremely noisy and also slow down the user's code (try printing 10k, 100k lines to term)

Some solutions:
1. Current option -- user sees tracebacks in the UI only (not console).  Since we don't print to console, there is no possibility of program slowdown
2. Current option + capture exceptions and raise them in python at the end of evaluation

PR #1869, opened 2024-07-01
chore(weave): unbreak docs

jwlee64:
fix broken link


jamie-rasmussen (approved):
Linking to master feels a little brittle, but broken docs is worse.

PR #1870, opened 2024-07-01
docs(weave): Add llama docs to sidebar

jwlee64:
before

Screenshot 2024-07-01 at 12 25 03 PM

after

Screenshot 2024-07-01 at 12 24 55 PM

### DIFF
--- a/docs/docs/guides/integrations/index.md
+++ b/docs/docs/guides/integrations/index.md
@@ -2,7 +2,7 @@
 
 The Weave ecosystem contains automatic logging integrations for popular AI libraries, and Weave components for common AI workflows. Weave will automatically trace calls made via the following libraries:
 
-- [openai](/guides/integrations/openai): the `openai` library.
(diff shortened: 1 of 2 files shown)


docs/docs/guides/integrations/index.md
```diff
@@ -5,3 +5,4 @@ The Weave ecosystem contains automatic logging integrations for popular AI libra
 - [openai](/guides/integrations/openai): the `openai` library.
 - [anthropic](/guides/integrations/anthropic): the `anthropic-sdk-python` library.
 - [mistralai](/guides/integrations/mistral): the `mistralai` library.
+- [llamaIndex](/guides/integrations/llamaindex): the `llamaIndex` library.
```
jamie-rasmussen, line 8: I think we should be capitalizing these correctly, i.e. `LlamaIndex` if we are referring to the framework, `llama-index` if we are referring to the Python package.

jamie-rasmussen (approved):

PR #1872, opened 2024-07-02
docs(integration): Add `LiteLLM` example to `/guides/integrations/`

ash0ts:
Added an example identical to `anthropic` and `openai` for `LiteLLM`


weave/weave_client.py
```diff
@@ -489,7 +489,8 @@ def finish_call(
             summary = sum_dict_leaves([child.summary or {} for child in call._children])
         elif isinstance(output, dict) and "usage" in output and "model" in output:
             summary["usage"] = {}
-            summary["usage"][output["model"]] = {"requests": 1, **output["usage"]}
+            model_key = str(output["model"])  # Convert ObjectRef to string
```
tssweeney, line 492: Let's work on a different fix here: https://github.com/wandb/weave/pull/1891

tssweeney (changes requested):
Let's work on a different fix here: https://github.com/wandb/weave/pull/1891

weave/weave_client.py
ash0ts: Reverted my changes to allow for the changes in #1891

ash0ts:
@tssweeney Merged the fixes

tssweeney (approved):

PR #1873, opened 2024-07-02
chore(docs): feedback tutorial

scottire:


scottire:
This looks great. If we're only saying "Weave" - we should only say "Models" for consistency. Other than that, this is ready to go in my opinion
image

PR #1881, opened 2024-07-03
feat: new column management popup

jamie-rasmussen:
For the Calls grid, replace the built-in column management popup with our own. This addresses several issues:

* Action to open dialog was buried in the column overflow menu, hard to discover
* State is not kept in the URL, which is a pain for sharing with others or if you spent a lot of time customizing to get the view you want
* Popup is not movable and obscures the grid content underneath
* Does not use W&B styling so inconsistent with rest of app.

Before:
Screenshot 2024-07-03 at 4 50 58 PM

After:
Screenshot 2024-07-03 at 4 50 38 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
@@ -1,0 +1,192 @@
+/**
+ * Button/popup for grid column visibility state.
+ */
+
+import {Popover} from '@mui/material';
+import {GridColDef, GridColumnVisibilityModel} from '@mui/x-data-grid-pro';
+import {Switch} from '@wandb/weave/components';
+import classNames from 'classnames';
+import React, {useCallback, useRef, useState} from 'react';
+
+import {Button} from '../../../../../Button';
+import {DraggableGrow} from '../../../../../DraggablePopups';
+import {TextField} from '../../../../../Form/TextField';
+import {Tailwind} from '../../../../../Tailwind';
+import {ColumnInfo} from '../../types';
+import {TraceCallSchema} from '../wfReactInterface/traceServerClient';
+
+type ManageColumnsButtonProps = {
+  columnInfo: ColumnInfo;
+  columnVisibilityModel: GridColumnVisibilityModel;
+  setColumnVisibilityModel: (model: GridColumnVisibilityModel) => void;
+};
+
+export const ManageColumnsButton = ({
+  columnInfo,
+  columnVisibilityModel,
+  setColumnVisibilityModel,
+}: ManageColumnsButtonProps) => {
+  const [search, setSearch] = useState('');
+  const ref = useRef<HTMLDivElement>(null);
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
+  const onClick = (event: React.MouseEvent<HTMLElement>) => {
+    setAnchorEl(anchorEl ? null : ref.current);
+    setSearch('');
+  };
+
+  const open = Boolean(anchorEl);
+  const id = open ? 'simple-popper' : undefined;
+
+  const toggleColumnShow = useCallback(
+    (key: string) => {
+      const on = columnVisibilityModel[key] ?? true;
+      const newModel = {
+        ...columnVisibilityModel,
+        [key]: !on,
+      };
+      setColumnVisibilityModel(newModel);
+    },
+    [columnVisibilityModel, setColumnVisibilityModel]
+  );
+  const onHideAll = () => {
+    const newModel = columnInfo.cols.reduce((acc, col) => {
+      if (col.hideable ?? true) {
+        acc[col.field] = false;
+      }
+      return acc;
+    }, {} as GridColumnVisibilityModel);
+    setColumnVisibilityModel(newModel);
+  };
+  const onShowAll = () => {
+    const newModel = columnInfo.cols.reduce((acc, col) => {
+      // If a column is not hideable, we also don't need to explicitly show it.
+      if (col.hideable ?? true) {
+        acc[col.field] = true;
+      }
+      return acc;
+    }, {} as GridColumnVisibilityModel);
+    setColumnVisibilityModel(newModel);
+  };
+
+  return (
+    <>
+      <span ref={ref}>
+        <Button
+          variant="ghost"
+          icon="column"
+          tooltip="Manage columns"
+          onClick={onClick}
+        />
+      </span>
+      <Popover
+        id={id}
+        open={open}
+        anchorEl={anchorEl}
+        anchorOrigin={{
+          vertical: 'bottom',
+          horizontal: 'center',
+        }}
+        transformOrigin={{
+          vertical: 'top',
+          horizontal: 'center',
+        }}
+        slotProps={{
+          paper: {
+            sx: {
+              overflow: 'visible',
+            },
+          },
+        }}
+        onClose={() => setAnchorEl(null)}
+        TransitionComponent={DraggableGrow}>
+        <Tailwind>
+          <div className="min-w-[300px] p-12">
+            <div className="handle flex items-center pb-8">
+              <div className="flex-auto font-semibold">Manage columns</div>
+              <div>
+                <Button
+                  size="small"
+                  variant="ghost"
+                  icon="close"
+                  tooltip="Close column management"
+                  onClick={e => {
+                    e.stopPropagation();
+                    setAnchorEl(null);
+                  }}
+                />
+              </div>
+            </div>
+            <div className="mb-8">
+              <TextField
+                placeholder="Filter columns"
+                autoFocus
+                value={search}
+                onChange={setSearch}
+              />
+            </div>
+            <div className="max-h-[300px] overflow-auto">
+              {columnInfo.cols.map((col: GridColDef<TraceCallSchema>) => {
+                const value = col.field;
+                const idSwitch = `toggle-vis_${value}`;
+                const checked = columnVisibilityModel[col.field] ?? true;
+                const label = col.headerName ?? value;
+                const disabled = !(col.hideable ?? true);
+                if (
+                  search &&
+                  !label.toLowerCase().includes(search.toLowerCase())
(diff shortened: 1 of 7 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
```diff
@@ -0,0 +1,192 @@
...
+      const on = columnVisibilityModel[key] ?? true;
+      const newModel = {
+        ...columnVisibilityModel,
+        [key]: !on,
+      };
+      setColumnVisibilityModel(newModel);
+    },
+    [columnVisibilityModel, setColumnVisibilityModel]
+  );
+  const onHideAll = () => {
```
gtarpenning, line 51: when i click this i get a strange blank space in my data grid... 
Screenshot 2024-07-09 at 9 02 53 PM

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
```diff
@@ -0,0 +1,192 @@
...
+    <>
+      <span ref={ref}>
+        <Button
+          variant="ghost"
+          icon="column"
+          tooltip="Manage columns"
+          onClick={onClick}
+        />
+      </span>
+      <Popover
```
gtarpenning, line 81: this looks so good. One comment is that when dragging around the window my mouse turns into a horizontal scroll icon, i might expect a grabber. like: 

```
cursor: grabbing;
```

or something like that.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const onHideAll = () => {
+    const newModel = columnInfo.cols.reduce((acc, col) => {
+      if (col.hideable ?? true) {
+        acc[col.field] = false;
+      }
+      return acc;
+    }, {} as GridColumnVisibilityModel);
+    setColumnVisibilityModel(newModel);
+  };
+  const onShowAll = () => {
```
gtarpenning, line 60: is `showAll` the default? When I click `showAll` I get the constructed link, which is quite large, we could just use the vanilla link with no filter options right? Otherwise storing the info in the link is soo nice.

gtarpenning: for example, to me, these two links show an identical table: 

```http://localhost:9000/griffin_wb/prod-evals-july/weave/traces```

```http://localhost:9000/griffin_wb/prod-evals-july/weave/traces?cols=%7B%22feedback%22%3Atrue%2C%22derived.status_code%22%3Atrue%2C%22inputs.self%22%3Atrue%2C%22inputs.model%22%3Atrue%2C%22output.score_answer_format.correct.true_count%22%3Atrue%2C%22output.score_answer_format.correct.true_fraction%22%3Atrue%2C%22output.score_answer_expected.correct.true_count%22%3Atrue%2C%22output.score_answer_expected.correct.true_fraction%22%3Atrue%2C%22output.model_latency.mean%22%3Atrue%2C%22wb_user_id%22%3Atrue%2C%22started_at%22%3Atrue%2C%22derived.tokens%22%3Atrue%2C%22derived.cost%22%3Atrue%2C%22derived.latency%22%3Atrue%7D```

File: weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx [hunk 4]
```diff
@@ -0,0 +1,192 @@
...
+                          toggleColumnShow(col.field);
+                        }}
+                        disabled={disabled}>
+                        <Switch.Thumb size="small" checked={checked} />
+                      </Switch.Root>
+                      <label
+                        htmlFor={idSwitch}
+                        className={classNames(
+                          'ml-4',
+                          disabled ? '' : 'cursor-pointer'
```
gtarpenning, line 161: i wonder if we have a common 'disabled' mouse choice...

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
```diff
@@ -0,0 +1,192 @@
...
+                        onCheckedChange={isOn => {
+                          toggleColumnShow(col.field);
+                        }}
+                        disabled={disabled}>
+                        <Switch.Thumb size="small" checked={checked} />
+                      </Switch.Root>
+                      <label
+                        htmlFor={idSwitch}
+                        className={classNames(
+                          'ml-4',
```
gtarpenning, line 160: Mega nit: feels the tiniest bit tight. not sure if everyone agrees. 

Screenshot 2024-07-09 at 9 14 27 PM

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
```diff
@@ -0,0 +1,192 @@
...
+                  !label.toLowerCase().includes(search.toLowerCase())
+                ) {
+                  return null;
+                }
+                return (
+                  <div key={value}>
+                    <div
+                      className={classNames(
+                        'flex items-center py-2',
+                        disabled ? 'opacity-40' : ''
```
gtarpenning, line 145: I noticed in [figma](https://www.figma.com/design/01KWBdMZg5QM9SRS1pQq0z/Design-System----Robot-Styles?node-id=4755-27877&t=YYZd1MLxt8BEFvrh-0) we use `opacity-30` on some of the disabled components, like checkboxes.

gtarpenning: not sure if applicable, just noting

gtarpenning (commented):
Looks awesome, I left a few small comments on some visual stuff.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
gtarpenning: like its currently: 
image

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
gtarpenning: compared to the before image it does now appear like there is less of a gap: 
image

jamie-rasmussen: I'll bump it to `ml-6`, we can increase it more if it still feels tight.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
jamie-rasmussen: This matches what we do for other disabled controls like buttons.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
jamie-rasmussen: Good question! Right now, all columns are shown by default. However, not that long ago we had to disable a nice feature we used to have (was going to be complicated with server side pagination), where we would automatically hide columns that were mostly blank. I'm not sure if we will bring that back or not. I also think it is quite possible that we will introduce other columns that are hidden by default, for example, we might want to add separate columns for Input Tokens and Output Tokens - we probably don't need to show both of those by default in addition to Total tokens, but they could become available via this dialog.

If we do get to a situation where not all columns are shown by default, we'd probably want to add another action like "Reset". So we'll probably need to revisit no matter what.

While keeping the URL short would be nice, it's already unreadable as soon as you open a peek drawer, and it's about to get even worse once we start keeping filter state etc in there. (Which will be a massive UX improvement other than URL length.)

So I'm inclined to keep it as is, but if you think it would be better for Show all to reset to default I'm OK with that too.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
jamie-rasmussen: Nice suggestion. I don't know why react-draggable is using `ew-resize` for this. I'll override it.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
jamie-rasmussen: I will follow up with Yangyi.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
jamie-rasmussen: Yes, you can get this in prod too with the Material UI column selector, so unrelated to this PR. I think it is a grid bug in the interaction between hidden and pinned columns (the blank space corresponds to the pinned but hidden feedback column).

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/ManageColumnsButton.tsx
jamie-rasmussen: After this PR goes in I can update our other popups to have the same grabbing cursor behavior.

gtarpenning (approved):
Awesome

PR #1882, opened 2024-07-04
chore: don't load segment plugin if no apikey

jamie-rasmussen:
Internal Slack: https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1720104688407169


jwlee64 (approved):
thank you!

PR #1886, opened 2024-07-08
fix(weave): Update box impl to use python None and bool instead of BoxedNone and BoxedBool

andrewtruong:
resolves:
- https://wandb.atlassian.net/browse/WB-19587

### DIFF
--- a/weave/trace/vals.py
+++ b/weave/trace/vals.py
@@ -16,8 +16,8 @@ from pydantic import BaseModel
 from pydantic import v1 as pydantic_v1
 
 from weave.client_context.weave_client import get_weave_client
-from weave.legacy import box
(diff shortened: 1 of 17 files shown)


weave/trace/vals.py
```diff
@@ -481,6 +481,8 @@ def make_trace_obj(
     box_val = box.box(val)
     if isinstance(box_val, pydantic_v1.BaseModel):
         box_val.__dict__["ref"] = new_ref
+    elif box_val is None or isinstance(box_val, bool):
```
tssweeney, line 484: please add a larger comment here as to why we have this exception and what the implications are

tssweeney (approved):
Approving with change request

PR #1891, opened 2024-07-08
chore(weave): Fix LiteLLM + Anthropic Issue

tssweeney:
Fixes summary rollups so that we don't operate on refs!

### DIFF
--- a/weave/weave_client.py
+++ b/weave/weave_client.py
@@ -487,9 +487,16 @@ class WeaveClient:
         summary = {}
         if call._children:
             summary = sum_dict_leaves([child.summary or {} for child in call._children])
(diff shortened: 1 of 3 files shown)


weave/weave_client.py
```diff
@@ -487,9 +487,16 @@ def finish_call(
         summary = {}
         if call._children:
             summary = sum_dict_leaves([child.summary or {} for child in call._children])
-        elif isinstance(output, dict) and "usage" in output and "model" in output:
+        elif (
```
tssweeney, line 490: we want to use `original_output` here so that we don't accidentally start operating on refs

andrewtruong (approved):

PR #1895, opened 2024-07-08
chore(weave): Fix anonymous user mode

tssweeney:


jamie-rasmussen (changes requested):
I think we need a more comprehensive solution, the call details page will still crash for anonymous users, anonymous users should see a readonly view instead of nothing, and for public projects we need to handle the case of user being auth'ed but not a member of the project. I'll put up a PR to address.

PR #1896, opened 2024-07-09
fix(weave): Change box impl to use python bool instead of BoxedBool

andrewtruong:


shawnlewis (approved):

andrewtruong:
merged into: https://github.com/wandb/weave/pull/1886

PR #1898, opened 2024-07-09
fix: feedback was crashing for anonymous viewer

jamie-rasmussen:
Internal Slack: https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1720477278890919

I didn't consider public projects and anonymous users when implementing the feedback UI.

### DIFF
--- a/weave-js/src/common/hooks/useViewerUserInfo.ts
+++ b/weave-js/src/common/hooks/useViewerUserInfo.ts
@@ -34,7 +34,7 @@ export const useViewerUserInfo = (): UserInfoResponse => {
 };
 
 const VIEWER_QUERY = gql`
-  query Viewer {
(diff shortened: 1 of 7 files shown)


weave-js/src/common/hooks/useViewerUserInfo.ts
```diff
@@ -34,7 +34,7 @@ export const useViewerUserInfo = (): UserInfoResponse => {
 };
 
 const VIEWER_QUERY = gql`
-  query Viewer {
+  query Viewer2 {
```
jamie-rasmussen, line 37: Definition was conflicting after changes in useViewerInfo - need to combine these files.

tssweeney (approved):

PR #1901, opened 2024-07-09
fix(weave): Refactor @op deco to work with `inspect` lib (`isfunction`, `iscoroutinefunction`)

andrewtruong:
### Resolves
- https://wandb.atlassian.net/browse/WB-19428
- https://wandb.atlassian.net/browse/WB-19461
- https://wandb.atlassian.net/browse/WB-19462
- https://wandb.atlassian.net/browse/WB-19686

This PR makes the `@op` deco work with `inspect.isfunction` and `inspect.iscoroutinefunction` which are commonly used in other patching libs

There are a few tradeoffs, discussed here:
https://www.notion.so/wandbai/Weave-op-decorator-changes-a9fcd15d75ec4b1b917bbe657a3b8bd7?pvs=4


weave/flow/agent.py
```diff
@@ -1,4 +1,4 @@
-from typing import Any
```
shawnlewis, line 1: Why this change? Maybe just revert this file?

weave/legacy/ops_arrow/ops.py
```diff
@@ -1,11 +1,11 @@
 from weave.legacy.ops_arrow.arraylist_ops import *
 from weave.legacy.ops_arrow.boolean import *
-from weave.legacy.ops_arrow.convert_ops import *
-from weave.legacy.ops_arrow.date import *
+from weave.legacy.ops_arrow.convert_ops import *  # type: ignore
```
shawnlewis, line 3: unrelated?

weave/legacy/ops_primitives/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .. import context_state as _context_state
+from weave.legacy import context_state as _context_state
```
shawnlewis, line 1: unrelated?

weave/tests/test_op2.py
```diff
@@ -0,0 +1,178 @@
+import inspect
```
shawnlewis, line 1: Can this file have a better name?

weave/trace/vals.py
```diff
@@ -145,6 +146,12 @@ def attribute_access_result(
 ) -> Any:
     # Not ideal, what about properties?
     if callable(val_attr_val):
+        sig = inspect.signature(val_attr_val)
```
shawnlewis, line 149: I think this code should be a function in op.py called like bind_op()

weave/weave_client.py
```diff
@@ -409,7 +413,16 @@ def create_call(
                 self._anonymous_ops[op] = _build_anonymous_op(op)
             op = self._anonymous_ops[op]
         if isinstance(op, Op):
-            op_def_ref = self._save_op(op)
+            if isinstance(op, partial):
```
shawnlewis, line 416: can this block be written as a function in op.py?

weave/trace/op.py
```diff
@@ -256,3 +62,264 @@ def _apply_fn_defaults_to_inputs(
...
+    c = _create_call(op, *args, **kwargs)
+    return _execute_call(op, c, *args, **kwargs)
+
+
+def calls(op: Op) -> "CallsIter":
+    client = client_context.weave_client.require_weave_client()
+    return client._op_calls(op)
+
+
+# Legacy decos
```
shawnlewis, line 209: What is legacy decos? We don't need this. legacy.decorator_op has the legacy decorator, but we shouldn't need to redefine anything here.

weave/tests/test_client_trace.py
```diff
@@ -1316,8 +1316,10 @@ def op_with_custom_type(self, v):
     assert obj2.op_with_custom_type(1) == 2
 
     my_op_ref = weave_client.get_ref(CustomType.op_with_custom_type)
-    with pytest.raises(MissingSelfInstanceError):
-        my_op2 = my_op_ref.get()
+    assert my_op_ref is None, "Must call get on the instance!"
```
tssweeney, line 1319: Love seeing this! Huge (:

weave/tests/test_weaveflow.py
```diff
@@ -256,14 +256,15 @@ def get_item(row):
     assert res == {"in": 1, "out": 1}
 
 
-def test_op_method_name(client):
-    class MyModel(weave.Model):
-        a: int
+class MyModel(weave.Model):
```
tssweeney, line 259: is this just style or is there an issue with defining the class inside the func?

weave/tests/test_op2.py
```diff
@@ -0,0 +1,178 @@
...
+    @op
+    async def amethod(self, a: int) -> int:
+        return a + 1
+
+
+a = A()
+b = B()
+
+
+def test_sync_func(client):
```
tssweeney, line 52: It would be really nice if all these tests also had a version where the user publishes the object / op first, then "gets" the op and uses it.

weave/trace/op.py
```diff
@@ -256,3 +62,264 @@ def _apply_fn_defaults_to_inputs(
...
+    The interface for Op-ified functions and methods.
+
+    Op was previously a class, and has been converted to a Protocol to allow
+    functions to pass for Op.  This is needed because many popular packages are
+    using the `inspect` module for control flow, and Op instances don't always
+    pass those checks.  In particular, `inspect.iscoroutinefunction` always
+    fails for classes, even ones that implement async methods or protocols.
+
+    Some of the attributes are carry-overs from when Op was a class.  We should
+    consider removing the unnecessary ones where possible.
```
tssweeney, line 79: it would be nice to note explicitly which of them fall into this category

weave/trace/op.py
```diff
@@ -256,3 +62,264 @@ def _apply_fn_defaults_to_inputs(
...
+    name: str
+    signature: inspect.Signature
+    ref: Optional[ObjectRef]
+    resolve_fn: Callable
+
+    call: Callable[..., Any]
+    calls: Callable[..., "CallsIter"]
+
+    # not sure if this is the best place for this, but kept for compat
+    _set_on_output_handler: Callable[[OnOutputHandlerType], None]
```
tssweeney, line 91: Both of these are used in integrations - the contract needn't be maintained if you have alternative ways of achieving the same outcome

weave/trace/op.py
```diff
@@ -256,3 +62,264 @@ def _apply_fn_defaults_to_inputs(
...
+    __self__: Any
+
+
+def _set_on_output_handler(func: Op, on_output: OnOutputHandlerType) -> None:
+    if func._on_output_handler is not None:
+        raise ValueError("Cannot set on_output_handler multiple times")
+    func._on_output_handler = on_output
+
+
+def _is_method_alt(func: Callable) -> bool:
```
tssweeney, line 104: what does this function do?

weave/trace/op.py
tssweeney: would it be possible to put these in a different file and selectively import??

andrewtruong: we can try, but let's do that in a different PR.  This has enough dangerous changes as is!

weave/trace/vals.py
andrewtruong: clarified as `maybe_bind_method` in `op.py`

weave/weave_client.py
andrewtruong: clarified as `maybe_unbind_method` in `op.py`

weave/trace/op.py
andrewtruong: clarified as `_is_unbound_method` in `op.py`

tssweeney (approved):
🚢

PR #1902, opened 2024-07-09
chore: Fixes Tim/fix crash with different scorers

tssweeney:
Example: https://app.wandb.test/lavanyashukla/nims_rag_2/weave/compare-evaluations?evaluationCallIds=["7ca7836f-e055-42eb-94c5-1625b73af789"%2C"1024b6c8-6707-4e61-9474-463ad424a067"%2C"fad0e96b-e9e3-4a84-8ce0-ad61dce5987e"%2C"d6358605-5c1d-4561-819d-179bb433a7a7"] 

This fixes the page crash when not all evaluations have the same scorers.


weave/weave_client.py
```diff
@@ -817,3 +819,26 @@ def op_fn(*args, **kwargs):  # type: ignore
...
+                dict_res[k] = redact_sensitive_keys(v)
+        return dict_res
+
+    elif isinstance(obj, list):
+        list_res = []
+        for v in obj:
+            list_res.append(redact_sensitive_keys(v))
+        return list_res
+
+    return obj
```
nickpenaranda, line 844: do we also need a case for `isinstance(obj, tuple)`?

nickpenaranda (approved):
approving in case the answer to my question in comment is "no"

weave/weave_client.py
tssweeney: added in https://github.com/wandb/weave/pull/1906

PR #1905, opened 2024-07-09
chore: EvalCompare: Limit comparison to 5 and show help tooltip

tssweeney:
* Limits evaluation selection to 5 evaluations
* Shows help tooltip when disabled

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -12,6 +12,7 @@ import {
   Chip,
   FormControl,
   ListItem,
+  Tooltip,
 } from '@mui/material';
 import {Box, Typography} from '@mui/material';
 import {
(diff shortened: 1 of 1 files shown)


gtarpenning (commented):
can we do something more like: 

```
useEffect(() => {
    if (!isEvaluateTable) {
      return;
    }
    const disabledDueToMax = selectedCalls.length >= MAX_EVAL_COMPARISONS;
    const disabledDueToNonSuccess = selectedCalls.some(call => call.exception != null || call.ended_at == null);
    let tooltipText = '';
    if (disabledDueToNonSuccess) {
      tooltipText = 'Cannot compare non-successful evaluations';
    } else if (disabledDueToMax) {
      tooltipText = `Comparison limited to ${MAX_EVAL_COMPARISONS} evaluations`;
    }
    addExtra('compareEvaluations', {
      node: (
        <CompareEvaluationsTableButton
          onClick={() => {
            history.push(
              router.compareEvaluationsUri(entity, project, selectedCalls)
            );
          }}
          disabled={selectedCalls.length === 0 || disabledDueToMax || disabledDueToNonSuccess}
          tooltipText={tooltipText}
        />
      ),
      order: 3,
    });
```

This will then allow any selection, so that users can delete failed evaluations for example, but still prevent comparisons when necessary.

gtarpenning:
Like this maybe? @tssweeney  
eval-warning

gtarpenning (approved):

PR #1909, opened 2024-07-09
chore(docs):  Added gemini, together, open router and local models pages support docs via our OpenAI sdk compatibility

altryne:
Given our native support of OpenAI SDK and many companies supporting that format by default, as per slack discussion, added a few pages that outline how Weave can be utilized with those companies even before we have a native integration with their pytho clients. 

Also added a local models category, with links to the most popular local model runners with support for OpenAI SDK


altryne:
closed in favor of #1910

PR #1923, opened 2024-07-11
fix: Stop redacting inside of reffed objects

tssweeney:
We had a production issue caused by https://github.com/wandb/weave/commit/0a78fcd7132b3e3387b41fd3f855695a1c7fd6db made it so that the eval calls lost their refs in inputs, which broke some un-tested contracts relied on by the UI. This fixes that code in a general way and adds specific tests for these issues.

I also verified in the UI that this generates the correct data and renders well in the unpatched UI


weave/weave_client.py
```diff
@@ -818,6 +818,19 @@ def op_fn(*args, **kwargs):  # type: ignore
 
 
 def redact_sensitive_keys(obj: typing.Any) -> typing.Any:
+    # We should NEVER mutate reffed objects.
+    #
+    # 1. This code builds new objects that no longer have refs
+    # 2. Even if we did an in-place edit, that would invalidate the ref (since
+    # the ref is to the object's digest)
+
+    try:
```
jamie-rasmussen, line 827: Is there a reason we need this broad try/except?

weave/tests/test_evaluations.py
```diff
@@ -0,0 +1,120 @@
...
+
+    calls = client.server.calls_query(
+        tsi.CallsQueryReq(
+            project_id=client._project_id(),
+        )
+    )
+
+    # Flatten calls is a technique we use in the integration tests to assert the correct
+    # ordering of calls. This is used to assert that the calls are in the correct order
+    # as well as nested in the correct way. The returned list is the ordered list if calls
```
jamie-rasmussen, line 77: if => of

jamie-rasmussen (approved):

nickpenaranda (approved):

weave/tests/test_evaluations.py
tssweeney: fixed

weave/weave_client.py
tssweeney: I was under the impression this was risky, but you are correct, it should be safe. We actually recursively do this check immediately after this call so we should be good. Remove the try/except

weave/weave_client.py
```diff
@@ -818,6 +818,15 @@ def op_fn(*args, **kwargs):  # type: ignore
 
 
 def redact_sensitive_keys(obj: typing.Any) -> typing.Any:
+    # We should NEVER mutate reffed objects.
+    #
+    # 1. This code builds new objects that no longer have refs
+    # 2. Even if we did an in-place edit, that would invalidate the ref (since
+    # the ref is to the object's digest)
+    ref = get_ref(obj)
+    if ref:
```
andrewtruong, line 827: nit:
```suggestion
    if get_ref(obj):
```

weave/tests/test_evaluations.py
```diff
@@ -0,0 +1,120 @@
...
+    calls = client.server.calls_query(
+        tsi.CallsQueryReq(
+            project_id=client._project_id(),
+        )
+    )
+
+    # Flatten calls is a technique we use in the integration tests to assert the correct
+    # ordering of calls. This is used to assert that the calls are in the correct order
+    # as well as nested in the correct way. The returned list is the ordered list of calls
+    # with the depth of each call.
```
andrewtruong, line 78: this should be a docstring of the `flatten_calls` func

weave/tests/test_evaluations.py
```diff
@@ -0,0 +1,120 @@
...
+    res = []
+    for child in children:
+        res.append((child, depth))
+        res.extend(flatten_calls(calls, child.id, depth + 1))
+
+    return res
+
+
+def op_name_from_ref(ref: str) -> str:
+    return ref.split("/")[-1].split(":")[0]
```
andrewtruong, line 28: should be in `refs.py`

weave/tests/test_evaluations.py
```diff
@@ -0,0 +1,120 @@
...
+        ("match_score2", 2),
+        ("Evaluation.predict_and_score", 1),
+        ("MyModel.predict", 2),
+        ("match_score1", 2),
+        ("match_score2", 2),
+        ("Evaluation.summarize", 1),
+    ]
+    assert got == exp
+
+    def isObjectRefWithName(val: str, name: str):
```
andrewtruong, line 100:
```suggestion
def is_object_ref_with_name(val: str, name: str):
```

andrewtruong: aside: are these utils generally useful?  Should we make them available as helpers inside `refs.py`?

andrewtruong (approved):

weave/weave_client.py
tssweeney: fixed

weave/tests/test_evaluations.py
tssweeney: fixed

weave/tests/test_evaluations.py
tssweeney: fixed. Mostly a set of test utils that are probably common. can pull these out later

PR #1926, opened 2024-07-11
chore: add grab cursor to more popups

jamie-rasmussen:
For the user and string popups, it was not obvious that the header could be grabbed and dragged to reposition the popup. This adds the same cursor treatment that was added for the column management popup in https://github.com/wandb/weave/pull/1881

Recommend viewing diff with "Hide whitespace"

Before:

Screenshot 2024-07-10 at 11 20 26 PM

After:

Screenshot 2024-07-10 at 11 20 13 PM
Screenshot 2024-07-10 at 11 19 57 PM

### DIFF
--- a/weave-js/src/components/DraggablePopups.tsx
+++ b/weave-js/src/components/DraggablePopups.tsx
@@ -5,7 +5,6 @@
 import Grow from '@mui/material/Grow';
 import Tooltip, {tooltipClasses, TooltipProps} from '@mui/material/Tooltip';
 import * as Colors from '@wandb/weave/common/css/color.styles';
-import classNames from 'classnames';
 import React, {useCallback, useState} from 'react';
 import Draggable from 'react-draggable';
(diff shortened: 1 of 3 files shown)


weave-js/src/components/DraggablePopups.tsx
```diff
@@ -75,13 +74,14 @@ export const DraggableHandle = ({children}: DraggableHandleProps) => {
   const [isDragging, setIsDragging] = useState(false);
   const onMouseDown = useCallback(() => setIsDragging(true), [setIsDragging]);
   const onMouseUp = useCallback(() => setIsDragging(false), [setIsDragging]);
+  const style: React.CSSProperties = {
```
jamie-rasmussen, line 77: This change is just so we don't have to wrap everything in a `<Tailwind>`

gtarpenning (approved):
sweet

PR #1930, opened 2024-07-11
feat(weave): bulk deletion mode

gtarpenning:
https://wandb.atlassian.net/browse/WB-19700

- changes checkbox to weave style 
- adds bulk delete button, confirm/cancel buttons, and confirmation modal
- adds tooltip to compare button when disabled because in delete mode
- adds tooltip to trashcan when more than 10 items are selected
- adds bulk selection checkmark
- adds colored red rows when row is selected in bulk delete mode 
- adds analytics to deletion

delete button logic:
- if you have any selected rows, clicking the trashcan brings up the delete confirmation modal
- if you have nothing selected, clicking the trashcan starts bulk delete mode
   - can select any row, including crashed or ongoing traces
   - rows selected turn red
   - two new buttons appear, "Confirm" and "Exit delete mode"
      - "Confirm" brings up the confirmation delete modal
      - "Exit delete mode" de-selects all rows and stops bulk delete mode

bulkdelete-confirm-red


weave-js/src/components/PagePanelComponents/Home/Browse3/StyledDataGrid.tsx
```diff
@@ -77,5 +79,11 @@ export const StyledDataGrid = styled(
         backgroundColor: backgroundColorHoveredSelected,
       },
     },
+    '&.bg-red-100': {
```
tssweeney, line 82: nit: Using `bg-red-100` implies that you are using some `red-100`, but i am seeing a more complicated color here. Since we are codifying the rules for "deletion styling", i would recommend changing this to `selected-for-deletion` classname. Then making that a constant that is used here and in the CallsTable. This way the class describes the state, rather than the implementation of the state

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/OverflowMenu.tsx
```diff
@@ -101,11 +108,21 @@ DialogActions.displayName = 'S.DialogActions';
 
 const MAX_DELETED_CALLS_TO_SHOW = 10;
 
-const ConfirmDeleteModal: FC<{
-  calls: CallSchema[];
+export const ConfirmDeleteModal: FC<{
+  entity: string;
+  project: string;
+  calls: Array<{callId: string; name: string}>;
```
tssweeney, line 114: seems odd to create a new shape for call - why not just pass in the CallSchema and parse out the name inside? Feels odd for the caller to do this work

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/OverflowMenu.tsx
```diff
@@ -45,7 +47,12 @@ export const OverflowMenu: FC<{
     <>
       {confirmDelete && (
         <ConfirmDeleteModal
-          calls={selectedCalls}
+          entity={entity}
```
tssweeney, line 50: While in practice, all the calls for a page are the same entity/project, it seems weird to pass this in from the caller. Instead, i would extract the first entity/project from the incoming calls inside the delete modal. As a bonus, since the delete calls requires an entity, project, you could extract unique entity/project pairs, then delete those calls in groups.

weave-js/src/components/PagePanelComponents/Home/Browse3/StyledDataGrid.tsx
gtarpenning: ack yeah thats a mistake, good idea...

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/OverflowMenu.tsx
gtarpenning: This allows for a bit more flexibility, by only requiring exactly what we need, but I do also see your point. Especially with the comment below. This is of course how it works right now, so I will just not include this change in the pr...

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/OverflowMenu.tsx
gtarpenning: right, yeah ill add that group delete.

tssweeney (approved):
Approving with some code comments

jwlee64 (approved):
nice all the event stuff looks good

PR #1938, opened 2024-07-12
feat(weave): Add caching and slicing for CallsIter

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19833

This PR adds client-side caching for `CallsIter`.

In the images below, the first cell is "before", and the second cell is "after" (with this PR)

Sample of 10 calls at page size=10
image

Sample of 100 calls at page size=10
image


tssweeney (approved):
Really nice change!

PR #1949, opened 2024-07-16
fix: Weave Object's Ops should be saved as ops!

tssweeney:
1. Fixes issue where bound ops are not actually bound to their parent self
2. Fixes object publishing to correctly publish the op


weave/weave_client.py
```diff
@@ -739,8 +739,11 @@ def _save_op(self, op: Op, name: Optional[str] = None) -> Ref:
 
         if inspect.ismethod(op):
             # this "func" is both a method AND an Op, but we need to cast to get dot access to its attributes
-            op = cast(op, Op)  # type: ignore
-        op.ref = op_def_ref  # type: ignore
+            op = cast(Op, op)
```
tssweeney, line 742: this was flipped, which resulted in the ref not actually getting assigned to the object!

weave/trace/op.py
```diff
@@ -350,7 +356,7 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
             wrapper.calls = partial(calls, wrapper)  # type: ignore
 
             wrapper.__call__ = wrapper  # type: ignore
-            wrapper.__self__ = wrapper  # type: ignore
+            wrapper.__self__ = None  # type: ignore
```
tssweeney, line 359: `__self__` was completely unused before - coopting it to mean "the self argument for ops that are gotten"

weave/trace/op.py
```diff
@@ -372,6 +372,8 @@ def maybe_bind_method(func: Callable, self: Any = None) -> Union[Callable, Metho
     If self is None, return the function as is.
     """
     if (sig := inspect.signature(func)) and sig.parameters.get("self"):
+        if inspect.ismethod(func) and id(func.__self__) != id(self):
+            raise ValueError("Cannot re-bind a method to an new object")
```
andrewtruong, line 376: Should you be able to?

weave/weave_client.py
andrewtruong: Wow, great catch!

andrewtruong (approved):

weave/weave_client.py
andrewtruong, line 742: This is no longer required.  The cast was only done to assign to `ref`
```suggestion

```

weave/trace/op.py
tssweeney: I can't think of a case where hitting this path is intentional, but seems more like a programming error. Maybe if we have a real use case for it we can remove

PR #1950, opened 2024-07-16
fix: remove refs from objects in other projects

gtarpenning:
This pr: 
- fix sqlite db inconsistency with digest uniqueness (objects + files): https://wandb.atlassian.net/browse/WB-19863
- remove refs from objects when saving where `ref.project != client.project`: https://wandb.atlassian.net/browse/WB-19838

Using ops that have refs to previous weave.init calls should be checked to make sure that they have the correct project. if not, remove the previous ref and reconstruct it.


weave/weave_client.py
```diff
@@ -90,6 +90,11 @@ def get_ref(obj: Any) -> Optional[ObjectRef]:
     return getattr(obj, "ref", None)
 
 
+def remove_ref(obj: Any) -> None:
+    if get_ref(obj) is not None:
```
nickpenaranda, line 94: This is duplicative w/ L671.  Seems like we want all of L671-675 in its own function `remove_stale_ref(obj)` or just inline the `del obj.ref` instead of calling this function

gtarpenning: Okay I understand, but I think I disagree. Because we want to return early if there is no stale ref, we probably want to keep that logic in the `_save_nested_objects` fn rather than a helper for the 4 lines. We absolutely could do the `del` inline, but its sort of ugly. If there was any meaningful computational hit to doing that attr check (`get_ref`) then sure, but it should be darn near instant, and we get a safer helper function (`remove_ref`) that mirrors the `get_ref` impl. 

Does that make sense? @nickpenaranda

weave/weave_client.py
```diff
@@ -90,6 +90,11 @@ def get_ref(obj: Any) -> Optional[ObjectRef]:
...
+        del obj.ref
```
andrewtruong, line 95:
```suggestion
        obj.ref = None
```

weave/weave_client.py
andrewtruong: I think remove ref is just setting `obj.ref = None` -- should be ok to inline

weave/weave_client.py
```diff
@@ -662,8 +667,13 @@ def _save_object_basic(
         return ref
 
     def _save_nested_objects(self, obj: Any, name: Optional[str] = None) -> Any:
-        if get_ref(obj) is not None:
-            return
+        ref = get_ref(obj)
+        if ref is not None:
+            if ref.project == self.project:
```
andrewtruong, line 672:
```suggestion
        if (ref := get_ref(obj)) is not None and ref.project == self.project:
```

weave/weave_client.py
nickpenaranda: ah right, I misunderstood what we're returning from.  ok.

re: parity w/ `get_ref`, we may not necessarily want that, as one mutates the object and the other does not.  what happens to the old ref once removed? i guess this is already an issue anyway.

weave/weave_client.py
```diff
@@ -662,8 +667,13 @@ def _save_object_basic(
...
+                return
```
andrewtruong, line 673: Why do you return here?  Is it because the obj is already saved in the other project?

gtarpenning: Its essentially a cached ref, no need to make a new ref if the object already has one. This is currently the case for anytime the obj has a ref, we are just adding the check for when the projects don't match @andrewtruong

andrewtruong: I would add that as a comment

gtarpenning: yep

weave/weave_client.py
gtarpenning: Woah do we not support 3.7? Sick @andrewtruong

andrewtruong:
You might also want to add a test where you:
1. Get the ref back; and
2. Try to do the same thing in a 3rd project; and
3. Get that ref back

The op should work each time you get it back

weave/tests/test_weave_client.py
```diff
@@ -480,6 +480,26 @@ def add3(x, y, z):
     assert obj2["b"].name == "op-add3"
 
 
+def test_op_mismatch_project_ref(client):
```
tssweeney, line 483: please also add some more tests:
1. Publish an object to another project, get it, then re-publish it to a new project
2. Publish an object to another project, get it, then add it to a nested property of the a new object entirely, then publish, ensuring all refs are correct
3. Publish an op and object to a project, then use them with a new project - verify all correctness.

weave/tests/test_weave_client.py
```diff
@@ -753,8 +817,8 @@ def test_obj_dedupe(client):
     client._save_object({"a": 2}, "my-obj")
     res = client._objects()
     assert len(res) == 2
-    assert res[0].version_index == 0
-    assert res[1].version_index == 1
+    assert res[0].version_index in [0, 1]
```
tssweeney, line 820: why is this needed? seems like the order is important?

gtarpenning: 1. is it important? 
2. The sqlite primary key addition changed the ordering (i think)

we could add ordering to the query

@tssweeney

tssweeney (approved):

weave/tests/test_weave_client.py
tssweeney: might want to add the ordering to match CH

weave/weave_client.py
```diff
@@ -748,8 +756,14 @@ def _save_object_basic(
         return ref
 
     def _save_nested_objects(self, obj: Any, name: Optional[str] = None) -> Any:
-        if get_ref(obj) is not None:
-            return
+        if (ref := get_ref(obj)) is not None:
+            # Obj has an existing ref, check if ref is to the correct project
```
tssweeney, line 760: optional: consider adding a block before this:

```
if ALLOW_MIXED_PROJECT_REFS:
   return
```

That way we can have a flag `ALLOW_MIXED_PROJECT_REFS` which helps us to a) change this logic if needed; b) isolate the feature-specific code here.

gtarpenning: cool, yep adding

PR #1952, opened 2024-07-16
fix(weave): The `call` special func should never raise

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19805


andrewtruong:
Closing in favour of: https://github.com/wandb/weave/pull/1968

PR #1954, opened 2024-07-16
fix(weave): Log client version with call attributes

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19733

This PR adds the current weave client version as an attribute to all calls.  It should help with debugging!


jamie-rasmussen (commented):
I expect we might want to introduce other attributes (OS? Python version?). Would it be useful to group these as a "weave" object inside the attributes instead of just a prefix? 

Also, probably not an issue in practice, but maybe we want to prevent users from using the reserved prefix?

tssweeney:
> I expect we might want to introduce other attributes (OS? Python version?). Would it be useful to group these as a "weave" object inside the attributes instead of just a prefix?
> 
> Also, probably not an issue in practice, but maybe we want to prevent users from using the reserved prefix?

+100 -> yes, we definitely want to have a reserved key-space for such attributes

weave/tests/test_client_trace.py
```diff
@@ -75,7 +77,15 @@ def my_op(a: int) -> int:
         exception=None,
         output=6,
         summary={},
-        attributes={},
+        attributes={
+            "weave": {
+                "client_version": weave.version.VERSION,
+                "source": "python-sdk",
```
jamie-rasmussen, line 83: could we also add `sys.version`

jamie-rasmussen (approved):

tssweeney (approved):

PR #1955, opened 2024-07-16
feat(weave): Add threading helpers to simplify thread usage

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19692

This PR adds `concurrent.futures.ThreadPoolExecutor` and `threading.Thread` variants that behave like the original and also copy context to each of the threads.  This should resolve issues users have where the context is not copied by default.

### DIFF
--- a/weave/tests/test_client_trace.py
+++ b/weave/tests/test_client_trace.py
@@ -1746,7 +1746,6 @@ def test_mapped_execution(client, mapper):
     assert_valid_trace(roots[root_ndx], first_val)
     root_ndx += 1
     for outer in middle_vals_outer:
-        print(f"{middle_vals_outer=}")
         assert_valid_trace(roots[root_ndx], outer)
         root_ndx += 1
 
--- a/weave/tests/test_client_trace.py
+++ b/weave/tests/test_client_trace.py
@@ -1610,15 +1610,16 @@ def map_with_threads_no_executor(fn, vals):
 
     for v in vals:
         thread = Thread(target=task_wrapper, args=(v,))
+        thread.start()
         threads.append(thread)
 
-    # run threads in batches
-    for i in range(0, len(threads), max_workers):
-        batch = threads[i : i + max_workers]
-        for thread in batch:
-            thread.start()
-        for thread in batch:
-            thread.join()
+        if len(threads) >= max_workers:
+            for thread in threads:
+                thread.join()
+            threads = []
+
+    for thread in threads:
+        thread.join()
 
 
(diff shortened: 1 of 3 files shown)


weave/tests/test_client_trace.py
```diff
@@ -1729,6 +1746,7 @@ def assert_valid_batched_trace(root_call):
     assert_valid_trace(roots[root_ndx], first_val)
     root_ndx += 1
     for outer in middle_vals_outer:
+        print(f"{middle_vals_outer=}")
```
tssweeney, line 1749: dev tracks

weave/trace/util.py
```diff
@@ -0,0 +1,54 @@
+from concurrent.futures import ThreadPoolExecutor as _ThreadPoolExecutor
+from contextvars import Context, copy_context
+from functools import partial
+from threading import Thread as _Thread
+from typing import Any, Callable, Iterable, Iterator, Optional
+
+
+class ContextAwareThreadPoolExecutor(_ThreadPoolExecutor):
```
tssweeney, line 8: please add docstrings to this class explaining when/why it is needed.

tssweeney: same with the below

weave/trace/util.py
```diff
@@ -0,0 +1,76 @@
...
+            context = self.contexts.pop(0)
+        return context.run(fn, *args, **kwargs)
+
+
+class ContextAwareThread(_Thread):
+    """A Thread that runs functions with the context of the caller.
+
+    This is a drop-in replacement for Thread that ensures that calls behave as
+    expected inside the thread.  You can achieve the same effect without this
+    class by instead writing:
```
tssweeney, line 63: Missing code?

weave/trace/util.py
```diff
@@ -0,0 +1,76 @@
+from concurrent.futures import ThreadPoolExecutor as _ThreadPoolExecutor
+from contextvars import Context, copy_context
+from functools import partial
+from threading import Thread as _Thread
+from typing import Any, Callable, Iterable, Iterator, Optional
+
+
+class ContextAwareThreadPoolExecutor(_ThreadPoolExecutor):
+    """A ThreadPoolExecutor that runs functions with the context of the caller.
```
tssweeney, line 9: I see you have added some comments, but as a user, I still don't know when/why to use this. You have described _what_ this does from a developer's perspective, but if an external user reads this I don't think they will quite understand what problem it is solving or why to use it

tssweeney (approved):

PR #1956, opened 2024-07-16
feat: store calls grid sort state in URL

jamie-rasmussen:
Internal Jira: https://wandb.atlassian.net/browse/WB-19839

Store the calls grid column sort state in the URL rather than in local component state. This is another step towards having URLs that a user can bookmark or share and not have to spend a lot of time re-configuring.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
@@ -9,7 +9,7 @@ import {
   Toolbar,
   Typography,
 } from '@mui/material';
-import {GridColumnVisibilityModel} from '@mui/x-data-grid-pro';
+import {GridColumnVisibilityModel, GridSortModel} from '@mui/x-data-grid-pro';
 import {LicenseInfo} from '@mui/x-license-pro';
 import {useWindowSize} from '@wandb/weave/common/hooks/useWindowSize';
 import {Loading} from '@wandb/weave/components/Loading';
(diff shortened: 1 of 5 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -691,6 +707,8 @@ const CallsPageBinding = () => {
       onFilterUpdate={onFilterUpdate}
       columnVisibilityModel={columnVisibilityModel}
       setColumnVisibilityModel={setColumnVisibilityModel}
+      sortModel={sortModel}
+      setSortModel={setSortModel}
```
gtarpenning, line 711: we might end up wanting a context for this (?)

weave-js/src/components/PagePanelComponents/Home/Browse3/grid/sort.ts
```diff
@@ -0,0 +1,41 @@
...
+  GridSortModel,
+} from '@mui/x-data-grid-pro';
+import _ from 'lodash';
+
+const isValidSortDirection = (obj: any): obj is GridSortDirection => {
+  return obj === 'asc' || obj === 'desc' || obj === null || obj === undefined;
+};
+
+const isValidSortItem = (obj: any): obj is GridSortItem => {
+  if (!_.isPlainObject(obj)) {
```
gtarpenning, line 13: what does this do?

weave-js/src/components/PagePanelComponents/Home/Browse3/grid/sort.ts
```diff
@@ -0,0 +1,41 @@
...
+  jsonString: string,
+  def: T = null as T
+): T => {
+  try {
+    const parsed = JSON.parse(jsonString);
+    if (_.isArray(parsed) && parsed.every(isValidSortItem)) {
+      return parsed as T;
+    }
+  } catch (e) {
+    // Ignore
```
gtarpenning, line 38: what gets returned in this case? `def: T = null as T`  resolves to, null ?

gtarpenning (approved):
lgtm, a few clarifying questions, might be worth comments idk...

weave-js/src/components/PagePanelComponents/Home/Browse3/grid/sort.ts
jamie-rasmussen: Checks whether the passed in thing is a JavaScript object like `{"key": "value"}` - `_.isObject` would also return true for things like arrays or functions.

gtarpenning: ah, right, thx

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
jamie-rasmussen: Yes, I expect we'll eventually want either a container object or a context since the number of props is going to keep ballooning as we add more things to state (filters should come soon, but could also be column order, column pinning, row selection, etc.) I need to think about what would be nicest.

weave-js/src/components/PagePanelComponents/Home/Browse3/grid/sort.ts
gtarpenning: oh `def` is `default`? that makes way more sense, thanks!

PR #1964, opened 2024-07-17
fix(weave): Make `call` return both the result and the call object

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19853


andrewtruong:
Closing in favour of: https://github.com/wandb/weave/pull/1968

PR #1965, opened 2024-07-17
chore(docs): add Models tutorial and docs CONTRIBUTING

morganmcg1:
Breaks out introduction to Weave Models from the Evaluations tutorial. The reason being that Weave Models are a powerful tool for versioning and should be highlighted independently of Evaluations for clarity, instead of introducing the concept alongside Evaluations & Scorers.

New models tutorial covers:
- Define, Init and call Weave Model with example
- Shows how to export, modify and use a tracked Weave Model

This PR adds a basic DOCS_CONTRIBUTING.md file just to have these instructions documented somewhere, based off of [W&B docodile readme](https://github.com/wandb/docodile)


jamie-rasmussen:
Closed in favor of https://github.com/wandb/weave/pull/2064

PR #1968, opened 2024-07-17
fix(weave): Don't raise when calling `call` special func, and return both res and `Call` obj

andrewtruong:
## Resolves:
- https://wandb.atlassian.net/browse/WB-19805
- https://wandb.atlassian.net/browse/WB-19853

This PR changes the `call` special func to:
1. Return both the result and call
2. Never raise -- any exceptions are captured in the `call` and the result is `None`.

NB:
1. I'm slightly concerned about the "return async wrapper" pattern we have because it may introduce complexity when we try to cover other types of funcs like generators, async generators, etc.
2. For now, this is OK.

### DIFF
--- a/weave/integrations/openai/openai_sdk.py
+++ b/weave/integrations/openai/openai_sdk.py
@@ -235,7 +235,9 @@ def should_use_accumulator(inputs: typing.Dict) -> bool:
     return isinstance(inputs, dict) and bool(inputs.get("stream"))
 
 
-def create_wrapper(name: str) -> typing.Callable[[typing.Callable], typing.Callable]:
+def create_wrapper_sync(
+    name: str,
+) -> typing.Callable[[typing.Callable], typing.Callable]:
     def wrapper(fn: typing.Callable) -> typing.Callable:
         "We need to do this so we can check if `stream` is used"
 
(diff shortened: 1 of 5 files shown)


weave/integrations/openai/openai_sdk.py
```diff
@@ -269,17 +271,54 @@ def _openai_stream_options_is_set(inputs: typing.Dict) -> bool:
     return wrapper
 
 
+# Surprisingly, the async `client.chat.completions.create` does not pass
+# `inspect.iscoroutinefunction`, so we can't dispatch on it and must write
```
tssweeney, line 275: i wonder if `create` is sort of like our old Op class where it defines a `__acall__` for non-stream and `__aiter__` + `__anext__` for stream? Could we look for that in our op code?

andrewtruong: I think it's because of their use of these decorators:
https://github.com/openai/openai-python/blob/main/src/openai/resources/completions.py#L571-L572

https://github.com/openai/openai-python/blob/main/src/openai/resources/completions.py#L1030-L1031


If you try to use them, you'll notice that any function decorated with these do not pass `inspect.iscoroutinefunction`

Normally `@overload` would because the pattern is:
```
@overload
async def func(): ...
@overload
async def func(x:int): ...
async def func(x:int,y:int): ...  # this comes last and is not decorated so it will pass

But they don't have that.

tssweeney (approved):

PR #1971, opened 2024-07-18
feat(weave): Add limited support for mutations

andrewtruong:
## Resolves:
- https://wandb.atlassian.net/browse/WB-19588

This PR allows for de-ref'd objects to be mutated and saved.  Supported objects include:
- `WeaveObject`
- `WeaveList`
- `WeaveDict`

It does not include "incremental" mutations where only the mutated parts are updated.  Instead, this will just send the entire latest version of that object across the wire.

### DIFF
--- a/weave/api.py
+++ b/weave/api.py
@@ -10,6 +10,8 @@ import dataclasses
 from typing import Any
 import threading
 
+from weave.trace.vals import Traceable
+
 from . import urls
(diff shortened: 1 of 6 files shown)


andrewtruong:
Closed in favour of:
https://github.com/wandb/weave/pull/2022

PR #1972, opened 2024-07-18
fix: call deletion

jamie-rasmussen:
Internal Jira:
https://wandb.atlassian.net/browse/WB-19877
https://wandb.atlassian.net/browse/WB-19878

Fix two problems with call deletion. Bad project_id getting sent to API and dev console key warning.

Before:
Screenshot 2024-07-17 at 11 38 19 PM
Screenshot 2024-07-17 at 11 36 33 PM


gtarpenning (approved):
ouch

PR #1975, opened 2024-07-18
fix: simplify delete error and show in modal

gtarpenning:
https://wandb.atlassian.net/browse/WB-19880

This PR: 
- simplifies the error message
- only set the error when catching from `Promise.all`
- only close the modal on successful deletion, otherwise display error

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/OverflowMenu.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/OverflowMenu.tsx
@@ -154,6 +154,16 @@ export const ConfirmDeleteModal: FC<{
       return;
     }
     setDeleteLoading(true);
+    userEvents.deleteClicked({
+      callIds: calls.map(
+        call => `${call.entity}/${call.project}/${call.callId}`
+      ),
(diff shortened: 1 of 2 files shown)


tssweeney (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/OverflowMenu.tsx
```diff
@@ -159,33 +159,38 @@ export const ConfirmDeleteModal: FC<{
...
-        callIds: calls.map(
-          call => `${call.entity}/${call.project}/${call.callId}`
-        ),
-        numCalls: calls.length,
-        userId: userInfoLoaded?.id ?? '',
-        organizationName: orgName,
-        entityName: userInfoLoaded?.username ?? '',
+    Promise.all(deletePromises)
+      .then(() => {
+        userEvents.deleteClicked({
```
jamie-rasmussen, line 167: Not sure how we are using this event but it seems a little off
- it gets logged on success, not click
- why is field called "entityName" if it is capturing the username?

jamie-rasmussen (approved):

PR #1981, opened 2024-07-18
chore: inaccessible project IDs set to special private str

gtarpenning:
should fix: https://wandb.atlassian.net/browse/WB-19245
depends on: https://github.com/wandb/core/pull/22755

This pr: 
- Returns a string instead of a ref when the project ID is `None`

We display the string, as it becomes the `op_name`. We can either handle that case specially in the frontend, or choose this string discerningly. 

private-ref


weave/trace_server/refs_internal.py
```diff
@@ -10,6 +10,7 @@
 
 WEAVE_INTERNAL_SCHEME = "weave-trace-internal"
 WEAVE_SCHEME = "weave"
+PRIVATE_WEAVE_STR = "<private>"
```
tssweeney, line 13: I would use something more distinct, for example `<<PRIVATE_REF_DATA>>`

tssweeney: i wonder if we should do something like:
`weave-private://///{rest}` (yes, 5 slashes).

tssweeney: that would still meet our criteria of not being an official "ref", but also retain the "rest" component (name, version, etc...) if we did want to parse it in the future

gtarpenning: yeah I like it

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/util.ts
```diff
@@ -18,11 +19,19 @@ export const isPrimitive = (val: any) => {
...
-    typeof value === 'string' &&
-    (value.startsWith(WANDB_ARTIFACT_REF_PREFIX) ||
-      value.startsWith(WEAVE_REF_PREFIX))
-  );
+  if (typeof value !== 'string') {
+    return false;
+  }
+  if (value.startsWith(WEAVE_REF_PREFIX_EMPTY)) {
+    return false;
+  }
```
gtarpenning, line 27: not technically a ref, don't try to parse it

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -1323,21 +1327,38 @@ const isValidTraceCall = (callRes: traceServerClient.TraceCallSchema) => {
...
 
+export const emptyRefToSimpleName = (ref: string) => {
+  const trimmed = ref.replace(WEAVE_REF_PREFIX_EMPTY, '');
+  try {
+    return trimmed.split('/')[1].split(':')[0];
+  } catch (e) {
+    return trimmed;
+  }
+};
+
```
gtarpenning, line 1338: simple parse for our empty ref that goes from the fake URI to just the object_id

weave/trace_server/trace_server_converter.py
```diff
@@ -91,6 +91,8 @@ def replace_ref(ref_str: str) -> str:
                 project_id
             )
         external_project_id = int_to_ext_project_cache[project_id]
+        if not external_project_id:
+            return f"{ri.WEAVE_SCHEME}://///{tail}"
```
tssweeney, line 95: this is not correct - it should not be `WEAVE_SCHEME` because we have empty project/name

gtarpenning: oooh I misread your previous comment, now i see `weave-private:`...

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -226,7 +226,7 @@ def ops_query(self, req: tsi.OpQueryReq) -> tsi.OpQueryRes:
         original_project_id = req.project_id
         req.project_id = self._idc.ext_to_int_project_id(original_project_id)
         res = self._ref_apply(self._internal_trace_server.ops_query, req)
-        for op in res.ops:
+        for op in res.op_objs:
```
gtarpenning, line 229: I was getting type errors here, but i dont understand how this is working...

weave/trace_server/trace_server_converter.py
```diff
@@ -91,6 +91,8 @@ def replace_ref(ref_str: str) -> str:
                 project_id
             )
         external_project_id = int_to_ext_project_cache[project_id]
+        if not external_project_id:
+            return f"{ri.WEAVE_PRIVATE_SCHEME}://///{tail}"
```
gtarpenning, line 95: empty entity/project

weave/trace_server/external_to_internal_trace_server_adapter.py
tssweeney: nice

we don't implement these on server so no one every hit them

tssweeney (approved):

PR #1982, opened 2024-07-18
add costs to stream query

jwlee64:
current attempt at creating a cost table, and handling cost calculation at read time

adds join and group by to orm
adds cost query, which can chain a couple queries on the end of a call query to add the costs to the calls summarys
updates the frontend to uptake the costs from summary.costs

Screenshot 2024-07-24 at 12 39 31 PM
Screenshot 2024-07-24 at 12 39 07 PM


weave/trace_server/current_calls_query.sql
```diff
@@ -0,0 +1,192 @@
+-- This is the current call stream query
```
jwlee64, line 1: this is just a copy of the query executed on the trace page for more readability

weave/trace_server/trace_server_interface.py
```diff
@@ -51,6 +51,8 @@ class CallSchema(BaseModel):
 
     deleted_at: typing.Optional[datetime.datetime] = None
 
+class CallSchemaWithCosts(CallSchema):
+    costs: typing.Optional[typing.Dict[str, typing.Any]] = None
```
tssweeney, line 55: you might just want to put costs in the `summary` dictionary under a `[weave][costs]` key. This would avoid messing with the underlying data model and keep usage / costs co-located.

jwlee64: that does make more sense

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,34 @@
...
+    /*
+    `pricing_level`: The level at which the token pricing applies. It can be 'default' for global pricing,
+    'project' for project-specific pricing, or 'org' for organization-specific pricing.
+    */
+    pricing_level String,
+
+    /*
+    `filter_id`: The filter identifier for the token pricing. It can be 'default' or any other string. This allow with type to uniquely identify where the pricing applies.
+    */
+    filter_id String,
```
tssweeney, line 11: nit: `pricing_level_id`

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,34 @@
...
+
```
tssweeney, line 12: might want to consider adding a provider (think about the case where the provider has different pricing for the same model id)

weave/trace_server/calls_query_builder.py
```diff
@@ -427,11 +427,155 @@ def as_sql(self, pb: ParamBuilder, table_alias: str = "calls_merged") -> str:
...
+        # because we are selecting from a subquery, we need to prefix the fields, but we dont want to run all aggregations again
+        # inparticular argMaxMerge throws an error when run twice on display_name
+        select_fields_sql = ", ".join(
+            'all_calls.' + field.field if isinstance(field, CallsMergedAggField) and field.agg_fn.__contains__('Merge') else field.as_select_sql(pb, "all_calls") for field in self.select_fields
+        )
+    
         raw_sql = f"""
-        WITH filtered_calls AS ({filter_query._as_sql_base_format(pb, table_alias)})
-        {outer_query._as_sql_base_format(pb, table_alias, id_subquery_name="filtered_calls")}
+        -- This is the current call stream query
```
tssweeney, line 452: all this stuff should def be moved to helper functions

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -172,6 +167,11 @@ class SelectableCHCallSchema(BaseModel):
 
     deleted_at: typing.Optional[datetime.datetime] = None
 
+    attributes_dump: str
```
jwlee64, line 170: we parse the returned sql based on where its returned, since we are modifying summary_dump, i moved it to the end

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/TraceUsageStats.tsx
```diff
@@ -72,41 +111,46 @@ export const TraceUsageStats = ({
   );
 };
 
-export const getUsageFromCellParams = (params: {[key: string]: any}) => {
-  const usage: {[key: string]: UsageData} = {};
+export const getCostFromCellParams = (params: {[key: string]: any}) => {
```
tssweeney, line 114: We are starting to add some expected structure to `TraceCallSchema`. I think the typing contracts might be better if we modify `TraceCallSchema` in a way like:

```
# Move CostData def here

type WeaveComputedSummaryType = {
  "cost"?: CostData,
  // Other computed summary fields can be added here
}

type SummaryKeyedDictType = KeyedDictType & {
  "weave"?: WeaveComputedSummaryType;
}

export type TraceCallSchema = {
  # <snip>
  summary?: SummaryKeyedDictType;
  # <snip>
};
```

The more clearly defines the expectation from the server inside the the server interface. Then, if we decide to change the server output, we have a single type definition to change in an authoritative location. 

Furthermore, your functions like `getCostFromCellParams` that are specific to rendering should operate on `WeaveComputedSummaryType`, rather than a free-form dict.

tssweeney: Note: note required for this PR, but this pattern can also be used for other computed fields we will soon add as well as the `weave` payload in attributes

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CustomGridTreeDataGroupingCell.tsx
```diff
@@ -202,7 +202,7 @@ export const CustomGridTreeDataGroupingCell: FC<
             </Box>
             {call?.rawSpan?.summary && (
               <TraceUsageStats
-                usage={call.rawSpan.summary.usage}
+                costData={call.rawSpan.summary.costs}
```
tssweeney, line 205: i'd recommend converting this to `call.traceCall.summary` (see the typing comment below)

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/TraceUsageStats.tsx
tssweeney: i am pretty sure you can also declare the pyandtic type on the server like:

```
class KeyedDictType(BaseModel):
    class Config:
        extra = "allow"

class SummaryKeyedDictType(KeyedDictType):
    weave: ...
```

this would be really nice to formalize the representation so we can reason about it in the future

weave/trace_server/calls_query_builder.py
```diff
@@ -427,11 +430,36 @@ def as_sql(self, pb: ParamBuilder, table_alias: str = "calls_merged") -> str:
...
+        WITH filtered_calls AS ({filter_query._as_sql_base_format(pb, table_alias)}),
+
+        -- Then we get all the calls we want, with all the data we need, with heavy filtering
+        all_calls AS ({outer_query._as_sql_base_format(pb, table_alias, id_subquery_name="filtered_calls")}),
+
+        -- Add Cost to the summary dump, of each call
+        {cost_query(pb, "all_calls", self.project_id, final_select_fields)}
         """
 
+        # raw_sql = f"""
```
tssweeney, line 456: comment cleanup

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: This indicates that something is not quite correct. We should always be selecting an explicit set of columns, then directly reading those columns out.

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    /*
+    `pricing_level`: The level at which the token pricing applies. It can be 'default' for global pricing,
+    'project' for project-specific pricing, or 'org' for organization-specific pricing.
+    */
+    pricing_level String,
+
+    /*
+    `pricing_level_id`: The pricing level identifier for the token pricing. It can be 'default' or any other string(e.g. org id or project id). This allow with type to uniquely identify where the pricing applies.
+    */
+    pricing_level_id String,
```
tssweeney, line 11: what is the value here if `pricing_level=default`?

jwlee64: "default" as well

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,39 @@
...
+
+    /*
+    `provider`: The initial provider of the LLM, some LLMs are provided other than the default provider.
+    */
+    provider_id String,
```
tssweeney, line 16: thank you for adding this - although currently it is unused, correct? Since we don't get provider out from our calls yet?

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,39 @@
...
+
+    /*
+    `prompt_token_cost`: The cost of an prompt token in the specified LLM.
+    */
+    prompt_token_cost Float,
+
+    /*
+    `completion_token_cost`: The cost of an completion token in the specified LLM.
+    */
+    completion_token_cost Float
```
tssweeney, line 36: I am assuming the unit is USD? Similar to provider, i anticipate different units in the near future

weave/trace_server/migrations/005_add_cost.up.sql
jwlee64: yes, but we do have that info in the table courtesy of
https://github.com/AgentOps-AI/tokencost/blob/main/tokencost/model_prices.json

weave/trace_server/current_calls_query.sql
```diff
@@ -0,0 +1,203 @@
...
+
+-- We get the lightly filtered call ids
```
tssweeney, line 3: you will likely need to modify / add to `weave/trace_server/tests/test_calls_query_builder.py` which will allow you to more formally define your sql expectations

weave/trace_server/migrations/005_add_cost.up.sql
jwlee64: hmm, eg "credits" or others
like in https://wandb.atlassian.net/browse/WB-19503

weave/trace_server/orm.py
```diff
@@ -495,10 +549,19 @@ def _transform_external_field_to_internal_field(
             field = prefix + "_dump"
 
     # validate field
-    if field not in all_columns and field.lower() != "count(*)":
+    if (
+        field not in all_columns
+        and field.lower() != "count(*)"
+        and all(
+            substr not in unprefixed_field.lower()
+            # TODO Josiah: this is a hack to allow some select fields to pass through, but it's not a good solution
```
tssweeney, line 557: is this something that will be addressed in this PR?

weave/trace_server/calls_query_builder.py
```diff
@@ -427,11 +430,36 @@ def as_sql(self, pb: ParamBuilder, table_alias: str = "calls_merged") -> str:
             outer_query.limit = self.limit
             outer_query.offset = self.offset
 
+        # because we are selecting from a subquery, we need to prefix the fields, but we dont want to run all aggregations again
+        # inparticular argMaxMerge throws an error when run twice on display_name
+        # We also filter out summary_dump, because we add costs to summary dump in the select statement
+        final_select_fields = [
```
tssweeney, line 436: it seems like most of this should happen inside of `cost_query` ... since you already are passing down the `all_calls`. That way the exclusion of summary_dump is close to where it gets excluded

weave/trace_server/calls_query_builder.py
```diff
@@ -427,11 +430,36 @@ def as_sql(self, pb: ParamBuilder, table_alias: str = "calls_merged") -> str:
...
+            "all_calls." + field.field
+            if isinstance(field, CallsMergedAggField)
+            and field.agg_fn.__contains__("Merge")
```
tssweeney, line 439: i think what you want to do here is something different. essentially all the aggregation resolution happens in your `all_calls` block. I think the cleaner solution is to simply rip through the fields, converting each to a plain `CallsMergedField` which will strip the aggs

weave/trace_server/migrations/005_add_cost.up.sql
tssweeney: nice! very good

weave/trace_server/migrations/005_add_cost.up.sql
tssweeney: different currencies, credits, tokens, etc...

weave/trace_server/current_calls_query.sql
```diff
@@ -0,0 +1,203 @@
...
+            arrayJoin(
+                arrayMap(
+                    kv -> (kv.1, kv.2),
+                    JSONExtractKeysAndValuesRaw(usage_raw)
+                )
+            ) AS kv,
+            kv.1 AS llm_id,
+            JSONExtractInt(kv.2, 'requests') AS requests,
+            -- Some libraries return input_tokens and output_tokens, others prompt_tokens and completion_tokens
+            if(
```
tssweeney, line 70: This is really interesting - another great chance to standardize this with a better type system. For backwards compat, we should keep both, but i think we might want to formally define the expectations at the interface layer and also conform to that in our python client

tssweeney (changes requested):
Dude, nice work - that SQL is insane! A few followups (you may want to break this up into more PRs if easier):

Requested change summary: 
Server:
1. Please write some tests to assert expected functionality. You don't actually need to hit any LLM API endpoints, but can fake the token usage. This should be both SQL tests as well as functional tests
2. Please cleanup dev tracks
3. Please undo the property changes to `SelectableCHCallSchema` and fix the related issue by having the caller control the columns


Interfaces:
1. Please formally type your new schema inside of `traceServerClientTypes.ts` and `trace_server_interface.py`
   * In addition, now is the time to move everything under the `summary.weave` key
2. Define expected usage payload in `trace_server_interface.py` and implement conforming changes in python

PR #1987, opened 2024-07-19
chore: Fix evaluation serialization in notebooks

tssweeney:
For some reason op serialization was not happy with the custom exception. Should solve correctly, but for now this works


andrewtruong:
I think the problem might be the code discovery -- we probably don't include the custom exception code

andrewtruong (approved):

PR #2000, opened 2024-07-23
chore: Make summary accept a Table object

tssweeney:

PR #2009, opened 2024-07-24
chore: Add table update methods

tssweeney:
This PR introduces a new method / endpoint to our universal interface that allows users to update Table rows via the API: `table_update(self, req: TableUpdateReq) -> TableUpdateRes`.

I was originally motivated for performance / bug reasons (see discussion below), but this is actually an important API for some other features on the horizon, namely:
1. Python SDK Dataset Mutations
2. UI Client Dataset Mutations (think "Add to Dataset")

Ok, now onto the original motivating factor:

Evaluations with large datasets / model predictions fail when getting to the summary step. This is because the input to summary is some multiple of the dataset size, often blowing the limit we have set for our servers. As a practical workaround, we can represent the summary input as a table (https://github.com/wandb/weave/pull/2000). However, we are currently limited to X MB Tables (as dictated by the service network boundary). While we could just notch that up, it is not a good solution. After landing this PR, and the service layer PR: https://github.com/wandb/core/pull/22869, we can implement a routine on the client for very large tables that will chunk table building into a series of updates, allowing us to build massive tables. This will not only fix the Evaluation issue, but also more generally support much more massive datasets.

### DIFF
--- a/weave/trace_server/trace_server_interface.py
+++ b/weave/trace_server/trace_server_interface.py
@@ -292,14 +292,87 @@ class TableCreateReq(BaseModel):
     table: TableSchemaForInsert
 
 
+"""
+The `TableUpdateSpec` pattern is as follows, where `OPERATION` is globally unique. This
+follows a similar pattern as our `Query` definitions.
+
+```
+class Table[OPERATION]SpecPayload(BaseModel):
+    ... # Payload for the operation
+
+
+class Table[OPERATION]Spec(BaseModel):
+    [OPERATION]: Table[OPERATION]SpecInner
+```
+
+Fundamentally, this allows us to 
+easily distinguish different operation types over the wire, and is quite readable. 
+Consider the payload:
+
+```
+{
+    updates: [
+        {append: {row: ROW_DATA}},
+        {pop: {index: POP_INDEX}},
+        {insert: {index: INSERT_INDEX, row: ROW_DATA}},
+    ]
+}
+```
+
+Consider that if we did not have this nesting, we would have:
+{
+    updates: [
+        {row: ROW_DATA},
+        {index: POP_INDEX},
+        {index: INSERT_INDEX, row: ROW_DATA},
+    ]
+}
+
+Which would require parsing the keys to make a heuristic "guess" as to what operation each entry is. 
+This is unacceptably fragile. An alternative is to include a "update_type" literal. This would certainly work,
+but stylistically, I prefer the former as it requires fewer JSON characters and is nicer for Pydantic to parse.
+{
+    updates: [
+        {update_type: 'append', row: ROW_DATA},
+        {update_type: 'pop', index: POP_INDEX},
+        {update_type: 'insert', index: INSERT_INDEX, row: ROW_DATA},
+    ]
+}
+"""
+
+
+class TableAppendSpecPayload(BaseModel):
+    row: dict[str, typing.Any]
+
+
+class TableAppendSpec(BaseModel):
+    append: TableAppendSpecPayload
+
+
+class TablePopSpecPayload(BaseModel):
+    index: int
+
+
+class TablePopSpec(BaseModel):
+    pop: TablePopSpecPayload
+
+
+class TableInsertSpecPayload(BaseModel):
+    index: int
+    row: dict[str, typing.Any]
+
+
+class TableInsertSpec(BaseModel):
+    insert: TableInsertSpecPayload
+
+
+TableUpdateSpec = typing.Union[TableAppendSpec, TablePopSpec, TableInsertSpec]
+
+
 class TableUpdateReq(BaseModel):
     project_id: str
-    initial_digest: str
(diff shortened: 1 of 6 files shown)


weave/trace_server/trace_server_interface.py
```diff
@@ -292,6 +292,20 @@ class TableCreateReq(BaseModel):
     table: TableSchemaForInsert
 
 
+class TableUpdateReq(BaseModel):
+    project_id: str
+    initial_digest: str
+    # insertions performed before popping so that it is easier for
+    # caller to know the index of insertion
+    insert_rows: typing.Optional[list[tuple[int, dict[str, typing.Any]]]] = None
```
nickpenaranda, line 300: I think there's a subtle ambiguity here when we have multiple entries in `insert_rows`. Can we clearly specify and test the intended semantics on inserting multiple rows? I can see 3 interpretations:

1. Sequentially insert each row before row `n` of the table, updating indices as you go.
2. Simultaneously insert each row before row `n` of the original table, where multiple rows having the same insert index are inserted in order of the request.
3. Simultaneously insert each row before row `n` of the original table, where multiple rows having the same insert index are inserted in reverse order of the request.

Tables aside, consider the zero-indexed list `A B C` and `insert_rows` payload of `(1, D), (2, E), (2, F), (3, G)`.  Under each interpretation, the result would be:

1. Sequential: `A D F G E B C`
2. Simultaneous in-order `A D B E F C G`
3. Simultaneous reverse-order `A D B F E C G`

I modified the test in `test_weave_client` and verified it is currently consistent with the sequential interpretation, which does make OK sense if the intent is to mirror a series of calls to `list.insert(idx, ...)`

but I'd argue that either simultaneous interpretations could be equally valid.  The simultaneous interpretation makes it easier to make multiple edits without the knowledge of what other changes may be happening.

I don't have a strong preference but I think the tests should be explicit about this behavior

weave/trace_server/trace_server_interface.py
```diff
@@ -292,6 +292,20 @@ class TableCreateReq(BaseModel):
...
+    pop_digests: typing.Optional[list[str]] = None
+    append_rows: typing.Optional[list[dict[str, typing.Any]]] = None
```
nickpenaranda, line 302: Is this necessary? Feels like this is just a special case of `insert_rows`. is it here for convenience? we should be explicit that these are guaranteed to be in tail position, regardless of whatever inserting/deleting we are doing in this same request.

weave/trace_server/trace_server_interface.py
```diff
@@ -292,6 +292,20 @@ class TableCreateReq(BaseModel):
     table: TableSchemaForInsert
 
 
+class TableUpdateReq(BaseModel):
+    project_id: str
+    initial_digest: str
+    # insertions performed before popping so that it is easier for
+    # caller to know the index of insertion
+    insert_rows: typing.Optional[list[tuple[int, dict[str, typing.Any]]]] = None
+    pop_digests: typing.Optional[list[str]] = None
```
nickpenaranda, line 301: are table rows guaranteed to be unique? what if the user intends to dedupe rows?

weave/trace_server/trace_server_interface.py
```diff
@@ -292,6 +292,20 @@ class TableCreateReq(BaseModel):
     table: TableSchemaForInsert
 
 
+class TableUpdateReq(BaseModel):
```
nickpenaranda, line 295: I'm sure you put a lot of thought here and don't want to discount it, but did you consider an interface where instead of separate lists of inserts, deletes and appends, there's a single list and each element is an insert, delete or append?  this clears up a great deal of ambiguity, and more akin to what i've encountered in the past.  cf. bigtable

weave/trace_server/trace_server_interface.py
tssweeney: Hey @nickpenaranda - thanks for the thoughtful comments. Per our discussion in slack, i revised the interface to be a sequential series of operations. This addresses this comment, as well as your other ones below.

weave/trace_server/trace_server_interface.py
```diff
@@ -292,13 +292,103 @@ class TableCreateReq(BaseModel):
...
+
+
 class TableRowSchema(BaseModel):
     digest: str
     val: typing.Any
 
 
 class TableCreateRes(BaseModel):
     digest: str
+    # TODO: Add Row Digests
```
tssweeney, line 391: removing.... this is coming in the next PR

weave/trace_server/trace_server_interface.py
```diff
@@ -292,6 +292,95 @@ class TableCreateReq(BaseModel):
     table: TableSchemaForInsert
 
 
+"""
```
nickpenaranda, line 295: 👏

nickpenaranda (approved):
The API looks good but I'm having a hard time following the clickhouse implementation.  how does this work in the case where we are only deleting rows? how does it maintain row order when inserting?

PR #2019, opened 2024-07-24
fix: Allow > 32MB datasets

tssweeney:
In our server layer, we have a number of endpoints which we would expect to have large request or response payloads. In most cases we have solved that with streaming, batching, or pagination. However, with `table/create` it is still all-or-nothing. This means we have a practical limit (the network configured limit or the memory limit of the container) on the size of tables (which is the underlying data structure for Datasets). This PR augments the writer flow such that we create a binary partition of the dataset rows when they exceed some configurable limit, breaking the creation into a create + log_2(N) updates. This is possible thanks to the new Update api and the fact that a updates can be serialized!

### DIFF
--- a/weave/conftest.py
+++ b/weave/conftest.py
@@ -461,12 +461,26 @@ def network_proxy_client(client):
     """
     app = FastAPI()
 
+    records = []
+
     @app.post("/table/create")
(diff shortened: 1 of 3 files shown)


nickpenaranda (approved):
nice!

PR #2022, opened 2024-07-25
feat(weave): Add limited support for object mutations (except `WeaveTable`)

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19588
- https://wandb.atlassian.net/browse/WB-19786

This PR allows for de-ref'd objects to be mutated and saved.  Supported objects include:
- `WeaveObject`
- `WeaveList`
- `WeaveDict`

Notably, this does not include:
- `WeaveTable`

It also does not include "incremental" mutations where only the mutated parts are updated.  Instead, this will just send the entire latest version of that object across the wire.


weave/trace/box.py
```diff
@@ -10,22 +10,29 @@
...
 import numpy as np
 
+from weave.ref_base import Ref
+
 T = TypeVar("T")
 
 
 class BoxedInt(int):
     _id: int | None = None
+    ref: Ref | None = None
```
andrewtruong, line 20: I think these refs were implicit before ([they would have been set here](https://github.com/wandb/weave/pull/2022/files#diff-cfc7f0d1bd8608f55950aef0b9d455bb13ac01875de41808f90aef73cd0cd207R598)).  I'm making them explicit.

weave/trace/vals.py
```diff
@@ -350,29 +424,44 @@ def __init__(
...
+        new_ref = self.ref.with_key(key) if self.ref else None
+        v = super().get(key, default)
+        return make_trace_obj(v, new_ref, self.server, self.root)
 
     def __setitem__(self, key: str, value: Any) -> None:
+        full_path: tuple[str, ...]
         if not isinstance(self.ref, ObjectRef):
-            raise ValueError("Can only set items on object refs")
+            # We used to raise ValueError here to only set attributes on
+            # object refs but I don't understand why that's necessary.
```
andrewtruong, line 440: Anyone know why we used to raise here?

weave/weave_client.py
```diff
@@ -125,9 +125,11 @@ def _get_direct_ref(obj: Any) -> Optional[Ref]:
 
 
 def map_to_refs(obj: Any) -> Any:
-    ref = _get_direct_ref(obj)
-    if ref:
+    if ref := _get_direct_ref(obj):
         return ref
+    if isinstance(obj, ObjectRef):
+        obj = obj.get()
```
tssweeney, line 131: why would we not return `obj` here?

weave/weave_client.py
```diff
@@ -142,6 +144,8 @@ def map_to_refs(obj: Any) -> Any:
         return [map_to_refs(v) for v in obj]
     elif isinstance(obj, dict):
         return {k: map_to_refs(v) for k, v in obj.items()}
+    elif isinstance(obj, WeaveObject):
+        return map_to_refs(obj._val)
```
tssweeney, line 148: this branch happens when we save a dirty object that is then stripped of the ref, correct? I wonder if this implies that stripping the ref should return the inner `val._val`?

weave/trace/vals.py
```diff
@@ -493,7 +590,8 @@ def make_trace_obj(
 
         pass
     else:
-        setattr(box_val, "ref", new_ref)
+        if hasattr(box_val, "ref"):
```
tssweeney, line 593: why only if it already has the attr?

weave/tests/test_weave_client_mutations.py
```diff
@@ -0,0 +1,250 @@
...
+
+    thing = Thing(a="hello", b=1, c=4.2)
+    ref = weave.publish(thing)
+
+    thing2 = ref.get()
+    assert thing2.a == "hello"
+    assert thing2.b == 1
+    assert thing2.c == 4.2
+
+    thing2.a = "new"  # TODO: Should we ignore this?
```
tssweeney, line 20: i don't understand the TODO here - can you elaborate?

weave/tests/test_weave_client_mutations.py
```diff
@@ -0,0 +1,250 @@
+from pydantic import Field
```
tssweeney, line 1: Loving this test suite

weave/trace/vals.py
tssweeney: What are the other Ref types you are now handling?

weave/trace/vals.py
```diff
@@ -350,29 +421,43 @@ def __init__(
...
+        if isinstance(value, Traceable):
+            value.parent = self
+
         super().__setitem__(key, value)
-        self.root.add_mutation(self.ref.extra, "setitem", key, value)
 
-    def keys(self):  # type: ignore
-        return super().keys()
+    def keys(self) -> Generator[Any, Any, Any]:  # type: ignore
+        yield from super().keys()
```
tssweeney, line 454: why the yield?

tssweeney: just to match the others maybe?

weave/weave_client.py
andrewtruong: Do you mean in `_save_object`?  I've done it this way because some objects don't have `val._val`, e.g. `WeaveList`, or `WeaveDict`

weave/weave_client.py
andrewtruong: Thanks for catching, this was just for testing

weave/trace/vals.py
andrewtruong: The intent of this section is `if self.ref is None`, but I kept this comment and structure because I wasn't sure why we needed to raise.

weave/trace/vals.py
```diff
@@ -87,19 +87,30 @@ def make_mutation(
...
+        self.ref = None
+        if self.parent is not self and self.parent is not None:
+            self.parent._mark_dirty()
 
     def add_mutation(
         self, path: tuple[str, ...], operation: MutationOperation, *args: Any
     ) -> None:
+        """Adds a mutation to the object.
+
+        TODO: This function is currently unused"""
```
tssweeney, line 109: i'd elaborate to say: `we are tracking mutations for bookkeeping purposes but are not actually using them yet. In the future, we may choose to only send mutation updates over the wire` ... something that helps the reader understand

weave/trace/vals.py
```diff
@@ -202,15 +208,31 @@ def __getattribute__(self, __name: str) -> Any:
...
             return object.__setattr__(self, __name, __value)
         else:
+            full_path: tuple[str, ...]
             if not isinstance(self.ref, ObjectRef):
-                raise ValueError("Can only set attributes on object refs")
-            object.__getattribute__(self, "root").add_mutation(
-                self.ref.extra, "setattr", __name, __value
-            )
-            return object.__setattr__(self._val, __name, __value)
+                # We used to raise ValueError here to only set attributes on
```
tssweeney, line 224: maybe this was because until now we would have expected every single WeaveObject to have a ref?. not clear to me either

weave/trace/vals.py
```diff
@@ -202,15 +208,31 @@ def __getattribute__(self, __name: str) -> Any:
...
+                # object refs but I don't understand why that's necessary.
+                full_path = (__name,)
+            else:
+                full_path = self.ref.extra + (__name,)
+            base_root = object.__getattribute__(self, "root")
```
tssweeney, line 229: here, and in other places we are getting the base_root, but not using it?

weave/weave_client.py
```diff
@@ -142,6 +142,8 @@ def map_to_refs(obj: Any) -> Any:
         return [map_to_refs(v) for v in obj]
     elif isinstance(obj, dict):
         return {k: map_to_refs(v) for k, v in obj.items()}
+    elif isinstance(obj, WeaveObject):
```
tssweeney, line 145: might be nice to add a little comment and say that this is only expected under the condition that the WeaveObject was previously dirtied

weave/trace/vals.py
```diff
@@ -202,15 +208,31 @@ def __getattribute__(self, __name: str) -> Any:
...
+            "ref",
+            "server",
+            "root",
+            "mutations",
+            "_is_dirty",
+            "parent",
+        ]:
             return object.__setattr__(self, __name, __value)
         else:
+            full_path: tuple[str, ...]
```
tssweeney, line 222: full_path is never used?

weave/trace/vals.py
andrewtruong: I'll just remove commentary here since it's not part of this PR and will be part of the actual mutation-applying in a subsequent PR.  This function existed before this PR.

tssweeney (approved):

PR #2027, opened 2024-07-25
feat(ui): Add isInteractive prop for Tags and Pills

connieelee:
https://wandb.atlassian.net/browse/WB-19989

Currently our Tags all have hover state styles (darker on hover). Meanwhile Pills never have hover styles.

I got feedback from design that Tags and Pills should both have hover states, but only if the element is interactive (e.g. is removable or triggers a tooltip). I added an `isInteractive` prop to `Tag` and `Pill` components so that hover styles are only applied if the element is explicitly marked as interactive.

Testing: in storybook

- RemovableTag always has hover styles
- Tag has hover styles only if `isInteractive`
- Pill has hover styles only if `isInteractive`

https://github.com/user-attachments/assets/4735489b-b367-47f2-a5de-5eaba1e19d9f

### DIFF
--- a/weave-js/src/components/Tag/Tag.tsx
+++ b/weave-js/src/components/Tag/Tag.tsx
@@ -81,7 +81,7 @@ export const Tag: FC<TagProps> = ({
   return nakedTag;
 };
 
-export type RemovableTagProps = TagProps & {
+export type RemovableTagProps = Omit<TagProps, 'isInteractive'> & {
   removeAction: ReactElement<typeof RemoveAction>;
 };
 export const RemovableTag: FC<RemovableTagProps> = ({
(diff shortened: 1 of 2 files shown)


weave-js/src/components/Tag/Tag.tsx
```diff
@@ -86,7 +94,7 @@ export const RemovableTag: FC<RemovableTagProps> = ({
 }) => {
   const labelRef = useRef<HTMLParagraphElement>(null);
   const isTooltipEnabled = isTagLabelTruncated(labelRef);
-  const classes = useTagClasses({color});
+  const classes = useTagClasses({color, isInteractive: true});
```
jamie-rasmussen, line 97: Should RemovableTagProps Omit 'isInteractive'?

jo-fang (commented):
tested in dark mode - lgtm

thank you for doing this - this has been in backlog for ages now

jamie-rasmussen (approved):

weave-js/src/components/Tag/Tag.tsx
connieelee: good call, will do!

PR #2028, opened 2024-07-25
fix(weave): take in has trace backend in sidebar

jwlee64:


weave-js/src/components/FancyPage/useProjectSidebar.ts
```diff
@@ -9,12 +9,16 @@ export const useProjectSidebar = (
...
   hasWeaveData: boolean,
-  isLargeWorkspaceModeEnabled: boolean
+  isLargeWorkspaceModeEnabled: boolean,
+  hasTraceBackend: boolean = true
 ): FancyPageSidebarItem[] => {
-  const isNoData = !hasModelsData && !hasWeaveData;
-  const isModelsOnly = hasModelsData && !hasWeaveData;
-  const isWeaveOnly = !hasModelsData && hasWeaveData;
-  const isBothData = hasModelsData && hasWeaveData;
+  const showModelsData = hasModelsData || !hasTraceBackend;
```
jamie-rasmussen, line 15: I think the names are really confusing here, maybe we can think of something better. Like with showModelsData - probably "SidebarItems" would be better than "Data", and it's kind of weird to have it be true but hasModelsData be false.

weave-js/src/components/FancyPage/useProjectSidebar.ts
```diff
@@ -9,13 +9,20 @@ export const useProjectSidebar = (
...
+  hasTraceBackend: boolean = true
 ): FancyPageSidebarItem[] => {
-  const isNoData = !hasModelsData && !hasWeaveData;
-  const isModelsOnly = hasModelsData && !hasWeaveData;
-  const isWeaveOnly = !hasModelsData && hasWeaveData;
-  const isBothData = hasModelsData && hasWeaveData;
-  const isShowAll = isNoData || isBothData;
+  // Should show models sidebar items if we have models data or if we don't have a trace backend
+  const showModelsSidebarItems = hasModelsData || !hasTraceBackend;
+  // Should show weave sidebar items if we have weave data and we have a trace backend
```
jamie-rasmussen, line 17: Is is possible to have weave data and not have a trace backend?

jamie-rasmussen (approved):

weave-js/src/components/FancyPage/useProjectSidebar.ts
jwlee64: probably not, but i think we query gorilla for the stream run there, so in theory hasWeavedata could be true with no tracebackend

PR #2030, opened 2024-07-25
feat(ui): Accept tooltip position for Buttons

connieelee:
- Button currently uses a semantic-based Tooltip --> replaced with radix Tooltip
- Button accepts new `tooltipPosition` prop which allows you to configure align/side of tooltip contents

Tested in storybook

https://github.com/user-attachments/assets/1d88a96e-8cc3-48ba-a0f1-9f1800f92ffe


weave-js/src/components/Button/Button.tsx
```diff
@@ -6,6 +6,7 @@
 
 import classNames from 'classnames';
 import React, {ReactElement} from 'react';
+import {PopupProps} from 'semantic-ui-react';
```
jamie-rasmussen, line 9: Sad that we can't get rid of semantic UI.

jamie-rasmussen (approved):

weave-js/src/components/Button/Button.tsx
connieelee: if I had more time, i would've replaced the semantic popup with our radix tooltip 🙈

connieelee: You know what, you've inspired me. I might as well do it now while I'm touching this code. Please hold

weave-js/src/components/RadixTooltip/index.tsx
```diff
@@ -57,7 +57,7 @@ export const Portal = RadixTooltip.Portal;
  */
 export const Content = React.forwardRef(
   ({className, children, ...props}: RadixTooltip.TooltipContentProps, ref) => (
-    <Tailwind>
+    <Tailwind style={{display: 'contents'}}>
```
connieelee, line 60: side note: i'm starting to think we should just add `display: contents` on the Tailwind wrapper by default because we actually do this quite a lot throughout the application

weave-js/src/components/Button/Button.tsx
connieelee: Didn't end up taking long at all :tada: updated to use radix tooltip instead of semantic one

weave-js/src/components/Button/Button.tsx
```diff
@@ -142,10 +148,22 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       </Tailwind>
     );
 
+    const [isTooltipOpen, setIsTooltipOpen] = useState(false);
     if (tooltip) {
-      // span is needed so tooltip works on disabled buttons
-      return <Tooltip content={tooltip} trigger={<span>{button}</span>} />;
+      return (
+        <Tooltip.Provider>
```
jamie-rasmussen, line 154: It feels weird to me that provider is at this level. I think this is what determines things like popup delay? Not sure the radix defaults match the guidelines from design.

jamie-rasmussen (changes requested):
This seems to break the buttons in Weave, Slacked you a screenshot

weave-js/src/components/Button/Button.tsx
connieelee: Yeah, the provider should be configured at the root of the application. We'd need to do this for both weave and app. There was [an attempt](https://github.com/wandb/core/pull/20803) at this a while back, but looks like it ended up being abandoned due to CI struggles. If I have spare time in the next few sprints I could try taking another stab at it.

weave-js/src/components/Button/Button.tsx
```diff
@@ -82,70 +88,81 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
-              className
-            )
-          )}
-          {...htmlAttributes}>
-          {startIcon ? <Icon name={startIcon} /> : null}
-          {children}
-          {endIcon ? <Icon name={endIcon} /> : null}
-        </button>
-      </Tailwind>
+      <button
```
connieelee, line 91: hide whitespace

weave-js/src/components/RadixTooltip/index.tsx
```diff
@@ -57,12 +57,13 @@ export const Portal = RadixTooltip.Portal;
...
       <RadixTooltip.Content
         className={twMerge(
           'night-aware',
-          'rounded text-sm leading-[140%] shadow-md',
-          'bg-moon-900 p-12 text-moon-200 dark:bg-moon-50 dark:text-moon-800'
+          'z-[999999] rounded text-sm leading-[140%] shadow-md',
```
connieelee, line 64: had to add this ridiculous z-index value otherwise tooltips will be rendered below the app's top nav bar. this wasn't a problem when we were using the semantic tooltip because they set this _even more_ ridiculous z-index

Screenshot 2024-07-25 at 6 44 18 PM

weave-js/src/components/RadixTooltip/index.tsx
```diff
@@ -57,12 +57,13 @@ export const Portal = RadixTooltip.Portal;
...
+          'bg-moon-900 p-12 text-moon-200 dark:bg-moon-50 dark:text-moon-800',
+          className
```
connieelee, line 66: just added comma + `className` at the end of the list here so we can specify style overrides if needed. not utilized in this PR, but I think may be useful one day and we'll be glad we don't need to open another weave PR

weave-js/src/components/Button/Button.tsx
```diff
@@ -82,70 +88,85 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
 
+    const [isTooltipOpen, setIsTooltipOpen] = useState(false);
     if (tooltip) {
-      // span is needed so tooltip works on disabled buttons
-      return <Tooltip content={tooltip} trigger={<span>{button}</span>} />;
+      return (
+        <Tailwind style={wrapperStyles}>
+          <Tooltip.Provider>
+            <Tooltip.Root open={isTooltipOpen} onOpenChange={setIsTooltipOpen}>
+              <Tooltip.Trigger className="cursor-pointer" asChild>
```
connieelee, line 154: verified disabled state as well -- cursor will not be pointer if button is disabled

weave-js/src/components/Button/Button.tsx
```diff
@@ -32,6 +33,7 @@ export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
   children?: ReactElement | string;
   active?: boolean;
   tooltip?: string;
+  tooltipProps?: TooltipContentProps;
```
connieelee, line 36: previously picked out `align` and `side` props specifically, but realized we might want full flexibility including `alignOffset`, `sideOffset`, and even `className` so I made this prop broader

weave-js/src/components/Button/Button.tsx
```diff
@@ -82,70 +85,81 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
+      </button>
     );
 
+    const [isTooltipOpen, setIsTooltipOpen] = useState(false);
     if (tooltip) {
-      // span is needed so tooltip works on disabled buttons
-      return <Tooltip content={tooltip} trigger={<span>{button}</span>} />;
+      return (
+        <Tailwind style={wrapperStyles}>
+          <Tooltip.Provider>
```
jamie-rasmussen, line 149: From the radix docs, it looks like the default delayDuration is 700ms
https://www.radix-ui.com/primitives/docs/components/tooltip#provider

The semantic tooltip sets it to 500ms which was a value from Cécile

jamie-rasmussen (commented):
It appears this PR changes the behavior of disabled buttons with respect to tooltips - I think they should still show, as they do on master. (Can be useful for explaining why a button is disabled.)

weave-js/src/components/Button/Button.tsx
connieelee: worry not! i'm using our wrapped version of Tooltip.Provider which has the 500ms delay configured:

https://github.com/wandb/weave/blob/dc10ac1f0053de29ebbe8f075c7e965769b6a7d7/weave-js/src/components/RadixTooltip/index.tsx#L13

jamie-rasmussen: Oh, I missed that, 👍

connieelee:
@jamie-rasmussen thanks for your patience and thorough testing 🙇‍♀️ apologies again for the hot mess this has been

verified with/without tooltip and with/without disable state


https://github.com/user-attachments/assets/84b34902-d7c9-45e5-9336-8beffc5d3030

jamie-rasmussen (approved):

PR #2040, opened 2024-07-29
feat(weave): Add configurable user settings

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19657
- https://wandb.atlassian.net/browse/WB-20055
- https://wandb.atlassian.net/browse/WB-20115

This PR adds a simple framework for configurable user settings that can be overriden by env vars.  For example, users can disable Weave tracing with the env var `WEAVE_DISABLED=true`


weave/trace/settings.py
```diff
@@ -0,0 +1,72 @@
+import os
+from contextvars import ContextVar
+from typing import Any, Optional, Union
+
+from pydantic import BaseModel, ConfigDict, PrivateAttr
+
+disabled = ContextVar("disabled", default=False)
```
tssweeney, line 7: curious why you chose to use multiple ContextVars as opposed to 1 `settings` that has multiple properties

weave/trace/settings.py
```diff
@@ -0,0 +1,72 @@
...
+print_call_link = ContextVar("disabled", default=True)
+
+
+class UserSettings(BaseModel):
```
tssweeney, line 11: It seems like the order of precedent is:

ENV_VAR > init param > default

As such, i wonder if there is a more pluggable way to define settings.

For example, to add a setting right now, you need to:
1. Setup a context var
2. explicitly add a line to `def apply(self) -> None:`
3. add a method like `def should_disable_weave() -> bool:`, which need to also encode the env var.

I think an alternative approach would be something like:

```
class Setting():
   keyname: str
   default_value: any
   parse_env: Callable    # override to process env var


UserSettings.register(Setting("disabled", False))

# Then elsewhere

from weave.trace.settings import settings

settings.get("disabled")
```


this might make the system more maintainable

tssweeney: add one line to add a setting `UserSettings.register(Setting("disabled", False))`, and let the system handle the rest

andrewtruong: How would a user pass in their own settings in this case?

Also it looks like `settings` is a dict or class, but that might not be threadsafe.

In general, I think we should keep it simple.  I've added helpers where they make sense and docs to make it more clear what a user needs to do.

weave/trace/settings.py
```diff
@@ -30,33 +36,26 @@ class UserSettings(BaseModel):
...
-
-
-def str2bool_truthy(v: str) -> bool:
-    return v.lower() in ("yes", "true", "1", "on")
+        for name in self.model_fields:
+            context_var = _context_vars.get(name)
+            context_var.set(getattr(self, name))
 
 
 def should_disable_weave() -> bool:
```
andrewtruong, line 53: Explicit definition better for autocomplete

weave/trace/settings.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def _reset(self) -> None:
+        for name, field in self.model_fields.items():
+            setattr(self, name, field.default)
+
+    def apply(self) -> None:
+        if self._is_first_apply:
+            self._is_first_apply = False
+        else:
+            self._reset()
```
tssweeney, line 46: i don't understand when this branch would get called

weave/trace/settings.py
tssweeney: I'm still not convinced this is the best pattern to use It feels awkward to me that we have separate access methods for each setting which is divorced from the setting specification (here, represented as a pydantic model field). However, we can iterate - not worth to many back and forths here.

weave/trace/settings.py
```diff
@@ -0,0 +1,87 @@
...
+        user_settings = UserSettings()
+    if isinstance(settings, dict):
+        user_settings = UserSettings.model_validate(settings)
+    if isinstance(settings, UserSettings):
+        user_settings = settings
+
+    user_settings.apply()
+
+
+_context_vars = {
```
tssweeney, line 74: I am pretty sure this will bust in notebooks. Checkout `weave/legacy/context_state.py` which has some special logic for handling context

tssweeney (approved):

PR #2043, opened 2024-07-30
feat: store pinned column state in URL

jamie-rasmussen:
Internal Jira: https://wandb.atlassian.net/browse/WB-20138

Same kind of change as in https://github.com/wandb/weave/pull/1976 and https://github.com/wandb/weave/pull/1956

The Material UI grid allows you to pin columns to the left or right. This state was previously just kept in the component but is now stored in the URL. This is another step towards having URLs that a user can bookmark or share.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
@@ -440,7 +440,6 @@ function buildCallsTableColumns(
     align: 'center',
     sortable: false,
     resizable: false,
-    disableColumnMenu: true,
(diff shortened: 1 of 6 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
```diff
@@ -440,7 +440,6 @@ function buildCallsTableColumns(
     align: 'center',
     sortable: false,
     resizable: false,
-    disableColumnMenu: true,
```
jamie-rasmussen, line 443: Doesn't seem to be a good reason to not allow user to pin/unpin or hide the user column from the column menu.

jwlee64 (approved):
lgtm

PR #2045, opened 2024-07-31
chore(weave): Add sanitization and data validation for calls

andrewtruong:


weave/integrations/langchain/langchain.py
```diff
@@ -107,7 +101,7 @@ def _persist_run(self, run: Run) -> None:
         def _persist_run_single(self, run: Run) -> None:
             """Persist a run."""
             run_dict = _run_to_dict(run, as_input=True)
-            run_name = make_valid_run_name(run.name)
+            run_name = sanitize_call_name(run.name)
```
jamie-rasmussen, line 104: The `<` case in `make_valid_run_name` is redundant with the regexp, but the `>` case is different - I'm guessing in this context `>` is at the end and a trailing `_` looks bad.

weave/trace/util.py
```diff
@@ -95,3 +95,32 @@ def run(self) -> None:
 # rename for cleaner export
 ThreadPoolExecutor = ContextAwareThreadPoolExecutor
 Thread = ContextAwareThread
+
+
+def sanitize_call_name(call_name: str) -> str:
+    """Sanitizes calls following the standard set in our Langchain integration.
+
+    - Alphanumeric + period, underscore, hyphen, space, backslash.
```
jamie-rasmussen, line 103: The backslash here is escaping the dash which has a special meaning (range, like in `a-z`) inside the regexp. We don't want to allow backslashes in names.

weave/trace/util.py
```diff
@@ -95,3 +95,32 @@ def run(self) -> None:
...
+    from urllib.parse import quote
+
+    return quote(call_name)
+
+
+def sanitize_call_name2(call_name: str) -> str:
+    """Sanitizes calls following the W&B Artifacts standard.
+
+    - Alphanumeric + period, underscore, and hyphen.
+    - All other chars are converted to underscores."""
```
jamie-rasmussen, line 123: Looks like you are actually removing them

PR #2046, opened 2024-07-31
fix: object viewer autoexpand

jamie-rasmussen:
Internal Jira: https://wandb.atlassian.net/browse/WB-20207

When we have the peek drawer open, the ObjectViewer for the inputs/outputs may stay mounted but with a changing `data` prop. When the data being rendered changes we need to recompute the default expansion mode, clear out the state for manually expanded nodes, etc.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
@@ -1,7 +1,13 @@
 import Box from '@mui/material/Box';
 import {GridRowId, useGridApiRef} from '@mui/x-data-grid-pro';
 import _ from 'lodash';
-import React, {useCallback, useContext, useMemo, useState} from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import styled from 'styled-components';
 
 import {isWeaveObjectRef, parseRef} from '../../../../../../react';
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
```diff
@@ -153,6 +157,18 @@ const ObjectViewerSectionNonEmpty = ({
...
+  useEffect(() => {
+    const isSimple = isSimpleData(data);
+    const newMode = isSimple || isExpanded ? 'expanded' : 'collapsed';
+    if (newMode === 'expanded') {
+      onClickExpanded();
+    } else {
+      onClickCollapsed();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [data, isExpanded]);
```
gtarpenning, line 170: q: what are we ignoring here? could we make those functions callbacks?

gtarpenning (approved):
this makes sense, the peek drawer not unmounting has considerable (and not alwayr obvious) downstream affects.... Nice

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
jamie-rasmussen: It's ignoring onClickExpanded and onClickCollapsed - not sure about making them callbacks, they both use mode state and set it, I think you would just end up with two different exhaustive-deps disablings?

gtarpenning: right... 👍🏻

PR #2048, opened 2024-07-31
feat(weave): backend for new costs query

jwlee64:
ahead of this I merged
orm updates to add join and groupby
moving the clickhouse schema objs to new file
add the llm_token_prices table

creates some query funcs in token_costs.py to add costs under summary.weave.costs to the calls query
enables costs in on the req include_costs = True


weave/trace_server/trace_server_interface.py
```diff
@@ -206,6 +240,7 @@ class CallEndRes(BaseModel):
 class CallReadReq(BaseModel):
     project_id: str
     id: str
+    add_cost: typing.Optional[bool] = False
```
tssweeney, line 243: nit: can we call this `include_costs`

weave/trace_server/trace_server_interface.py
```diff
@@ -253,6 +288,7 @@ class CallsQueryReq(BaseModel):
     # Sort by multiple fields
     sort_by: typing.Optional[typing.List[_SortBy]] = None
     query: typing.Optional[Query] = None
+    add_cost: typing.Optional[bool] = False
```
tssweeney, line 291: can you add a unit test that exercises this path - going through the client?

weave/trace_server/trace_server_interface.py
```diff
@@ -19,11 +19,25 @@ class ExtraKeysTypedDict(TypedDict):
 # https://docs.pydantic.dev/2.8/concepts/strict_mode/#dataclasses-and-typeddict
 ExtraKeysTypedDict.__pydantic_config__ = ConfigDict(extra="allow")  # type: ignore
 
-
-class WeaveSummarySchema(ExtraKeysTypedDict):
-    status: typing.Optional[typing.Literal["success", "error", "running"]]
-    nice_trace_name: typing.Optional[str]
-    latency: typing.Optional[int]
+cost_string_fields = [
```
tssweeney, line 22: I would recommend moving these lists into your DB layer. I think they are more descriptive of the on-disk state than the over-the-wire type

weave/trace_server/sqlite_trace_server.py
```diff
@@ -1097,11 +1097,16 @@ def _make_derived_summary_map(
     status = make_call_status_from_exception_and_ended_at(exception, ended_at_dt)
     latency = None if not ended_at_dt else (ended_at_dt - started_at_dt).microseconds
     display_name = display_name or op_name_simple_from_ref_str(op_name)
+    summary = summary_dump or {}
```
tssweeney, line 1100: this is an ugly artifact of these typing shanagans - i think we should punt on that instead of adding all these nulls over the wire

weave/trace_server/calls_query_builder.py
```diff
@@ -216,6 +217,7 @@ class CallsQuery(BaseModel):
     order_fields: list[OrderField] = Field(default_factory=list)
     limit: typing.Optional[int] = None
     offset: typing.Optional[int] = None
+    should_add_cost: bool = False
```
tssweeney, line 220: nit: `include_costs`

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -352,13 +353,18 @@ def calls_query_stream(
...
+        cq.add_cost(req.add_cost or False)
 
         # TODO (Perf): By allowing a sub-selection of columns
         # we will gain increased performance by not having to
         # fetch all columns from the database. Currently all use
         # cases call for every column to be fetched, so we have not
         # implemented this yet.
-        columns = all_call_select_columns
+        # We put summary_dump last so that when we compute the costs and summary its in the right place
+        columns = [
```
tssweeney, line 364: might want to be 
```
columns = all_call_select_columns
if req.add_cost:
   # move to end

File: weave/trace_server/clickhouse_trace_server_batched.py [hunk 7]
```diff
@@ -1425,12 +1431,17 @@ def _make_derived_summary_map(
         None if not ended_at or not started_at else (ended_at - started_at).microseconds
     )
     display_name = display_name or op_name_simple_from_ref_str(op_name)
+    summary = summary_dump or {}
+    if "_weave" not in summary:
```
tssweeney, line 1435: i think this goes away if we drop the pydantic stuff

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,49 @@
+CREATE TABLE llm_token_prices (
```
tssweeney, line 1: how does this get populated? Should we have a mechanism of bootstrapping this / updating from some public location?

weave/trace_server/token_costs.py
```diff
@@ -0,0 +1,462 @@
...
+    Column(name="completion_token_cost", type="float"),
+    Column(name="prompt_token_cost_unit", type="string"),
+    Column(name="completion_token_cost_unit", type="string"),
+]
+LLM_TOKEN_PRICES_TABLE_NAME = "llm_token_prices"
+LLM_TOKEN_PRICES_TABLE = Table(
+    name=LLM_TOKEN_PRICES_TABLE_NAME, cols=LLM_TOKEN_PRICES_COLUMNS
+)
+
+CALLS_MERGED_COLUMNS = [
```
tssweeney, line 20: can we derive this from the other list to have a single source of truth?

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,49 @@
...
+
+    /*
+    `completion_token_cost`: The cost of a completion token in the specified LLM.
+    */
+    completion_token_cost Float,
+
+    /*
+    `completion_token_cost_unit`: The unit of completion token cost in the specified LLM.
+    */
+    completion_token_cost_unit String
```
jamie-rasmussen, line 46: Do we need some fields for auditing purposes - which user inserted this row, when it was inserted?

weave/trace_server/clickhouse_schema.py
```diff
@@ -0,0 +1,107 @@
+import datetime
```
jwlee64, line 1: moved from clickhouse_trace_server_batched. You can import there and token_costs, to keep ch definitions as 1 source of truth

weave/trace_server/trace_server_interface.py
jwlee64: when I go through the client it hits the sqlite server rather than clickhouse, Is there something that I need to enable to correctly test through client?

tssweeney: every test that uses the `client` fixture will be run twice in CI - once for SQLite and once for CH. To force CH, you can run pytest with `--weave_server=clickhouse`. Checkout the `client` fixture for details

weave/weave_client.py
```diff
@@ -310,7 +317,7 @@ def make_client_call(
         id=server_call.id,
         inputs=from_json(server_call.inputs, server_call.project_id, server),
         output=output,
-        summary=server_call.summary,
+        summary=dict(server_call.summary if server_call.summary else {}),
```
tssweeney, line 320: can you explain why this change is needed? Seems to effect some tests and I am curious what was the need for the here (perhaps left over from pydantic?)

weave/trace_server/trace_server_interface.py
```diff
@@ -11,13 +12,67 @@
...
+
+class SummaryInsertMap(ExtraKeysTypedDict, total=False):
+    usage: typing.Dict[str, LLMUsageSchema]
+
+
+class SummaryMap(SummaryInsertMap, total=False):
+    weave: typing.Optional[WeaveSummarySchema]
+
+
+class WeaveAttributeSchema(TypedDict):
```
tssweeney, line 60: unused?

weave/trace_server/trace_server_interface.py
```diff
@@ -43,14 +98,20 @@ class CallSchema(BaseModel):
     output: typing.Optional[typing.Any] = None
 
     ## Summary: a summary of the call
-    summary: typing.Optional[typing.Dict[str, typing.Any]] = None
+    summary: typing.Optional[SummaryMap] = None
```
tssweeney, line 101: It seems like you found the golden solution here. Worth checking with @gtarpenning and @andrewtruong to check to see if the degenerate effects they noticed are exhibited here. In Griffin's PR we decided to punt on the pydantic typing for now. IDK how much trouble vs benefit this is causing you

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,59 @@
...
+
+    /*
+    `inserted_by`: User ID of the user who inserted the record, or the system if the record was inserted by the system.
+    */
+    inserted_by String,
+
+    /*
+    `inserted_at`: When the record was inserted.
+    */
+    inserted_at DateTime64(3) DEFAULT now64(3),
```
tssweeney, line 56: we use `created_at` in all other tables - can we use the same name here?

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -0,0 +1,59 @@
...
+
+) ENGINE = ReplacingMergeTree()
```
tssweeney, line 58: why `ReplaceingMergeTree` over `MergeTree`?

weave/trace_server/update_cost_table.py
```diff
@@ -0,0 +1,60 @@
+# script to load data into llm_token_prices table in ClickHouse
```
tssweeney, line 1: I think this routine might need some modifications to be production ready - perhaps it should be excluded from this PR?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1496,7 +1405,7 @@ def _end_call_for_insert_to_ch_insertable_end_call(
         id=end_call.id,
         exception=end_call.exception,
         ended_at=end_call.ended_at,
-        summary_dump=_dict_value_to_dump(end_call.summary),
+        summary_dump=_dict_value_to_dump(dict(end_call.summary)),
```
tssweeney, line 1408: curious what happens with the nested typed dicts? is `dict` enough to fix this? Should this be implemented in `_dict_value_to_dump`?

weave/trace_server/clickhouse_schema.py
tssweeney: nice - good idea. There will likely be some merge conflicts soon here, just fyi

weave/tests/test_weave_client.py
```diff
@@ -305,8 +305,9 @@ def test_calls_query(client):
                 "os_version": platform.version(),
                 "os_release": platform.release(),
                 "sys_version": sys.version,
-            },
+            }
         },
+        summary={},
```
tssweeney, line 310: seems odd to me that we now have an empty dict where we didn't before - is this intentional?

weave/trace_server/clickhouse_trace_server_batched.py
jwlee64: im going to move this to _any_value_to_dump and I confirmed this works for at least 5 deep nested dicts

weave/weave_client.py
jwlee64: I can modify this ternary so we dont return {} and mess up tests, but this is necessary because we return SummaryMap, but i left call.summary as a dict, because the rollup of arbritrary keys got a bit complex using typeddicts

.github/workflows/test.yaml
```diff
@@ -133,7 +133,7 @@ jobs:
           DD_SERVICE: weave-python
           DD_ENV: ci
           WEAVE_SENTRY_ENV: ci
-        run: CI=1 WB_SERVER_HOST=http://wandbservice WF_CLICKHOUSE_HOST=weave_clickhouse WEAVE_SERVER_DISABLE_ECOSYSTEM=1 source /root/venv/bin/activate && pip install 'dspy-ai>=0.1.5' && pip install pydantic -U  && cd weave && pytest --job-num=${{ matrix.job_num }} --timeout=90  ./flow ./integrations ./tests ./legacy/ops_arrow ./legacy/ecosystem ./trace_server ./trace --ddtrace --durations=5
+        run: CI=1 WB_SERVER_HOST=http://wandbservice WF_CLICKHOUSE_HOST=weave_clickhouse WEAVE_SERVER_DISABLE_ECOSYSTEM=1 source /root/venv/bin/activate && pip install 'dspy-ai>=0.1.5' && pip install pydantic -U  && cd weave && pytest -m "not skip_sqlite_client" --job-num=${{ matrix.job_num }} --timeout=90  ./flow ./integrations ./tests ./legacy/ops_arrow ./legacy/ecosystem ./trace_server ./trace --ddtrace --durations=5
```
jwlee64, line 136: follows the pattern of `skip_clickhouse_client` adding costs only works in clickhouse as of now, because of all the json and array functions I do to calculated the costs. 

I probably could add the json stuff to the orm and make it work for both sqlite and clickhouse, but that is not an insignificant lift. For now I thought it better to write two tests, one for sqlite to make sure include_costs does nothing, and another for clickhouse that include_costs adds a cost object

weave/tests/test_weave_client.py
```diff
@@ -1165,3 +1166,96 @@ def test_table_partitioning(network_proxy_client):
...
+    callsNoCost = list(client.calls(include_costs=False))
+
+    assert len(callsWithCost) == len(callsNoCost)
+    assert len(callsWithCost) == 4
+
+    noCostCallSummary = callsNoCost[0].summary
+    withCostCallSummary = callsWithCost[0].summary
+
+    assert withCostCallSummary.get("weave", "bah") != "bah"
+    assert len(withCostCallSummary["weave"]["costs"]) > 0
```
tssweeney, line 1231: is this the only place where you are validating that your cost calculation is correct? Are there any other places where maybe you test:
* mixed models
* models that have 2 entries in the cost table over time,
* ...

The primary end-results?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1397,7 +1406,7 @@ def _end_call_for_insert_to_ch_insertable_end_call(
         id=end_call.id,
         exception=end_call.exception,
         ended_at=end_call.ended_at,
-        summary_dump=_dict_value_to_dump(end_call.summary),
+        summary_dump=_any_value_to_dump(end_call.summary),
```
tssweeney, line 1409: this is not quite right. We want to ensure that `summary_dump` is in fact a dictionary dump. so `_dict_value_to_dump` is correct. I am not sure what is the motivation for this change as it opens us up to potentially non-dict data

tssweeney: i think you want what was here before: `_dict_value_to_dump(dict(end_call.summary))`. 

I think the key here is that end_call.summary is now Expected to be a TypedDict after your typing changes.

weave/trace_server/migrations/006_seed_costs.up.sql
```diff
@@ -0,0 +1,506 @@
+INSERT INTO llm_token_prices 
+    (id, pricing_level, pricing_level_id, provider_id, llm_id, effective_date, prompt_token_cost, prompt_token_cost_unit, completion_token_cost, completion_token_cost_unit, created_by, created_at) 
+VALUES 
+    (generateUUIDv4(), 'default', 'default', 'openai', 'gpt-4', now(), 3e-05, 'USD', 6e-05, 'USD', 'system', now()),
```
tssweeney, line 4: please use UUIDv7 https://clickhouse.com/docs/en/sql-reference/functions/uuid-functions#generateUUIDv7

weave/trace_server/migrations/005_add_cost.up.sql
```diff
@@ -1,4 +1,8 @@
 CREATE TABLE llm_token_prices (
+     /*
+    `id`: The unique identifier for the call. This is typically a UUID.
+    */
+    id String,
```
tssweeney, line 5: just flagging - very bad to add changes to a prior migration - i know we just landed this, but better form is to alter the table in yet another migration. Else, any commits between the land of the first table and now are invalid to deploy

weave/tests/test_weave_client.py
```diff
@@ -1165,3 +1166,167 @@ def test_table_partitioning(network_proxy_client):
...
+    noCostCallSummary = callsNoCost[0].summary
+    withCostCallSummary = callsWithCost[0].summary
+
+    assert withCostCallSummary.get("weave", "bah") != "bah"
+    assert len(withCostCallSummary["weave"]["costs"]) == 2
+
+    gpt4cost = withCostCallSummary["weave"]["costs"]["gpt-4"]
+    gpt4ocost = withCostCallSummary["weave"]["costs"]["gpt-4o"]
+
+    # rounds to nearest integer avoids .999999992345234
```
tssweeney, line 1254: tip: use `pytest.approx` - it is used in various parts of the code base for reference.

weave/trace_server/migrations/006_seed_costs.up.sql
jwlee64: i believe our version of clickhouse is too old for v7

tssweeney: whaaa? according to what?

tssweeney (approved):

PR #2052, opened 2024-08-01
feat(weave): Add limited mutation support for `WeaveTable`

andrewtruong:
- **test**
- **test**
- **test**
- **test**
- **tidy**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **cleanup**
- **test**
- **undo mutations change (save for later)**
- **test**
- **test**


weave/table.py
```diff
@@ -28,3 +31,23 @@ def __getitem__(self, key: int) -> dict:
...
+
+    def add(self, row: dict) -> None:
+        """Add a row to the table."""
+        self.rows.append(row)
+
+    def remove(self, index: int) -> None:
+        """Remove a row at the given index from the table."""
+        self.rows.pop(index)
+
+    def to_pandas(self) -> "pd.DataFrame":
```
tssweeney, line 46: Can pandas support be decoupled from this PR?

weave/table.py
```diff
@@ -28,3 +31,23 @@ def __getitem__(self, key: int) -> dict:
 
     def __iter__(self) -> Iterator:
         return iter(self.rows)
+
+    def __eq__(self, other: Any) -> bool:
+        return self.rows == other
+
+    def add(self, row: dict) -> None:
```
tssweeney, line 38: i would use the same names as python (append, pop) - i suspect that Table might end up wanting to implement the entire list protocol

weave/flow/dataset.py
```diff
@@ -64,3 +68,18 @@ def convert_to_table(cls, rows: Any) -> weave.Table:
                     "Attempted to construct a Dataset row with an empty dict."
                 )
         return rows
+
+    @weave.op
```
tssweeney, line 72: why are these ops?

weave/trace/vals.py
```diff
@@ -318,12 +341,32 @@ def __getitem__(self, key: Union[int, slice, str]) -> Any:
...
-    def append(self, val: Any) -> None:
-        if not isinstance(self.ref, ObjectRef):
-            raise ValueError("Can only append to object refs")
+    def append(self, row: typing.Dict) -> None:
+        self._load_rows_if_not_exists()
+        assert self._loaded_rows is not None
+        self._loaded_rows.append(row)
+        self._mark_dirty()
+
+    def add(self, row: typing.Dict) -> None:
```
tssweeney, line 353: why have `add` and `append`?

weave/trace/vals.py
```diff
@@ -318,12 +341,32 @@ def __getitem__(self, key: Union[int, slice, str]) -> Any:
...
+        self.append(row)
+
+    def remove(self, index: int) -> None:
```
tssweeney, line 356: `pop`?

weave/tests/test_dataset.py
```diff
@@ -0,0 +1,50 @@
...
+    # TODO: sqlite server path is broken
+    ds = weave.Dataset(rows=[{"a": 1, "b": 2}, {"a": 3, "b": 4}])
+    ref = weave.publish(ds)
+
+    ds2 = ref.get()
+    ds2.remove(0)
+    assert ds2.rows == [{"a": 3, "b": 4}]
+
+
+def test_gotten_dataset_to_frame(client):
```
tssweeney, line 43: i think there are some key tests missing:

```
ds = weave.Dataset(rows=[{"a": 1, "b": 2}, {"a": 3, "b": 4}])
ref = weave.publish(ds)

ds2 = ref.get()
ds2.rows[1]['a'] = 42
ref2 = weave.publish(ds2)

ds3 = ref2.get()
assert ds3.rows[1]['a'] == 42
```

andrewtruong: I don't think the table interface supports that mutation -- only append, pop, insert.  Do we want to add "cell editing" mutations?

weave/table.py
tssweeney: Well this is the correct question. Specifically, a Table is assumed to be a List[Dict], so that is one distinction. But other than that, we probably always want to use Table. Again, table is purely an implementation detail, and should not be something the user has to think about.

weave/flow/dataset.py
tssweeney: Interesting. So the root problem you are facing is that when to `weave.get` a dataset, the resulting object is a WeaveObject, and therefore do not have these methods bound. 

This feels a bit like a hack, with a few undesirable properties:
1. A bunch of calls tracked that the user probably does not care about
2. Ops that are not pure (in this case there is a side-effect). Ops are intended to be pure.
3. I think the result of this behavior is that a new version dataset/table gets created after every single method. I think this is more than desirable.

Do you agree that these are issues? Did you consider alternatives?

weave/tests/test_dataset.py
tssweeney: The existence of server-side table update operation is an implementation detail. We should be thinking about the user patterns that we want to support. 

You might claim that we should explicitly error when setting the "cell", or perhaps we cleanly support it. But users will definitely write this kind of code.

If you want to support with existing apis, a cell set is just a pop followed by an insert.

Anyhow, i think the thing to keep in mind here is that our goal is to correctly track the data for the user (as opposed to directly mapping the underlying mutation logic). In their mind, `rows` is just a list of dicts. The underlying updates/mutations are current implementation details / faculties we have available, but not something the user cares about.

weave/trace/vals.py
tssweeney: i think that is a good idea ;)

weave/flow/dataset.py
andrewtruong: RE: Properties
1. You might care about tracking row-adds to a dataset, but I think atomic unit is the "squash commit append", not per-row appends
2. I didn't know that Ops needed to be pure.  Mutations are inherently not pure, so maybe we need to rethink this.
3. Agree -- per method is noisy, but per "squash commit append" feels like it would be useful.

One alternative is to operate on the lower level `rows` object.  This feels wrong because the user is "reaching into" the Dataset object to interact with something that ostensibly should be an implementation detail, but it does immediately satisfy our constraints and `Dataset.append` effectively delegates to `Dataset.rows.append` anyways.

Another option is making something like a "baby op" which doesn't do all the versioning on op call but rather collects the mutations and only does 1 call at publish time.  That feels more correct but is also more work.

I think I prefer the 2nd option, but would not implement until we have the full mutation system working.

tssweeney: I think there might be a mixing of objectives going on. 

Objective 1: Allow a "gotten" dataset to have these functions
Objective 2: Track additional parts of the mutation process.

IMO, we should just do #1. The versioning happens at the object level and we can always diff the datasets / row properties if we want to know what changed between commits

weave/weave_client.py
```diff
@@ -805,6 +813,9 @@ def _save_nested_objects(self, obj: Any, name: Optional[str] = None) -> Any:
                 self._save_nested_objects(v)
         elif isinstance(obj, Op):
             self._save_op(obj)
+        # TODO: Kinda hacky way to dispatching Dataset with rows: Table
+        elif isinstance(obj, WeaveObject) and hasattr(obj, "rows"):
```
tssweeney, line 817: this might be more robust if we iterated through the fields and looked for table properties?

tssweeney (approved):

shawnlewis (approved):

PR #2053, opened 2024-08-01
fix: Evaluations are robust to empty / null key entries

tssweeney:
This can happen when the model or scorer is non deterministic (ie. uses LLM to generate JSON) or when there are conditional keys in the output. A user recently had all "Nones" for an "error" key.

### DIFF
--- a/weave/tests/test_evaluations.py
+++ b/weave/tests/test_evaluations.py
@@ -586,3 +586,41 @@ async def test_eval_supports_non_op_funcs(client):
 
     # # 2: Assert that the model was correctly oped
     # assert shouldBeModelRef.startswith("weave:///")
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


vanpelt (approved):

weave/tests/test_evaluations.py
```diff
@@ -586,3 +586,40 @@ def function_model(sentence: str) -> dict:
 
     # # 2: Assert that the model was correctly oped
     # assert shouldBeModelRef.startswith("weave:///")
+
+
+@pytest.mark.asyncio
+async def test_eval_is_robust_to_missing_values(client):
+    # At least 1 None
+    # All dicts have "d": None
```
jamie-rasmussen, line 594: Did `d` become `always_none`?

jamie-rasmussen (approved):

PR #2054, opened 2024-08-01
chore(weave): Cleanup api.py

andrewtruong:
- **test**
- **test**

### DIFF
--- a/weave/api.py
+++ b/weave/api.py
@@ -17,7 +17,6 @@ from weave.legacy import graph_mapper as _graph_mapper
 
 from . import storage as _storage
 from . import ref_base as _ref_base
-from weave.legacy import artifact_wandb as _artifact_wandb
 from weave.legacy import wandb_api as _wandb_api
 
 from . import weave_internal as _weave_internal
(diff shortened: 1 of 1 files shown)


weave/api.py
andrewtruong, line 1: Do we need this file anymore?

weave/trace_api.py
andrewtruong, line 1: Now the Trace API is decoupled, and has nice formatting and cmd-click!

weave/trace_api.py
```diff
@@ -0,0 +1,273 @@
...
+__docspec__ = [
+    init,
+    publish,
+    ref,
+    get_current_call,
+    finish,
+]
+
+
+__all__ = [
```
tssweeney, line 254: we should trim this down significantly

tssweeney: not right now, but in general

andrewtruong: Agreed

tssweeney (approved):

jwlee64 (approved):
looks good from my side

PR #2058, opened 2024-08-01
feat: feedback deletion in UI

jamie-rasmussen:
Internal Jira: https://wandb.atlassian.net/browse/WB-20018

Adds a button to the feedback grid, allowing you to delete feedback items. A future improvement would be to let people delete their notes from the notes popup.

Screenshot 2024-08-01 at 2 35 49 PM
Screenshot 2024-08-01 at 2 36 25 PM

This PR also adds listeners for feedback create and purge events. This keeps the various places that show feedback in sync as you make changes.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/FeedbackGrid.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/FeedbackGrid.tsx
@@ -1,11 +1,13 @@
 import {Box} from '@mui/material';
 import _ from 'lodash';
-import React from 'react';
+import React, {useEffect} from 'react';
 
+import {useViewerInfo} from '../../../../../common/hooks/useViewerInfo';
 import {Alert} from '../../../../Alert';
 import {Loading} from '../../../../Loading';
 import {Tailwind} from '../../../../Tailwind';
 import {useWFHooks} from '../pages/wfReactInterface/context';
+import {useGetTraceServerClientContext} from '../pages/wfReactInterface/traceServerClientContext';
 import {FeedbackGridInner} from './FeedbackGridInner';
 
 type FeedbackGridProps = {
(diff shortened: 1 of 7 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/FeedbackGrid.tsx
```diff
@@ -21,14 +23,22 @@ export const FeedbackGrid = ({
...
+
   const {useFeedback} = useWFHooks();
   const query = useFeedback({
     entity,
     project,
     weaveRef,
   });
 
-  if (query.loading) {
+  const getTsClient = useGetTraceServerClientContext();
```
tssweeney, line 35: any reason in particular you are directly using the tsClient here instead of putting the listener mount inside the useFeedback hook? Similar questions for `getTsClient().feedbackPurge({` in the `FeedbackGridActions.tsx` component? Maybe you prefer to skip the hooks interface all together?

tssweeney: fwiw, i feel like refactoring / removing the wfHooks interface, so this comment is not an opposition

tssweeney (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/FeedbackGrid.tsx
jamie-rasmussen: If we do refactor I think it would be good if our API layer in general supported refetching and polling like we get with apollo and a standard error interface for every API.

tssweeney: +1 - this totally agree

PR #2062, opened 2024-08-02
chore(cookbook): Improving causal reasoning using DSPy

soumik12345:


scottire (changes requested):
- The prompt optimization step takes too long for users to run through, please make this shorter
- Please run it through something to check for grammar or spelling mistakes
- Please add warnings when users are going to spend a few dollars
- Please trim some details about the DSPy implementation and explain the Weave APIs where possible

scottire (approved):

PR #2070, opened 2024-08-05
Feedback Cookbook [WIP]

NiWaRe:
Following the new guide on cookbooks this is a short-form text-code tutorial about how to gather and use expert and user feedback with Weave.


docs/docs/guides/cookbooks/summarization/.gitignore
NiWaRe, line 1: This folder was empty so I moved it out of the project.

PR #2074, opened 2024-08-05
chore: Implement write-time db field validation

tssweeney:
Currently our primary validation happens at the type-level through pydantic. However, there are a number of fields that are user defined, which it would be good for us to enforce the content, beyond just the typing. Here, i introduce validation for:

* User-controllable fields:
  * `call_id` -> enforce correct UUID
  * `trace_id` -> enforce correct UUID
  * `parent_id` -> enforce correct UUID
  * `display_name` -> enforce a limit of 128 characters
  * `op_name` -> enforce either a valid op internal ref, or a limit of 128 characters
  * `object_id` -> enforce a limit of 128 characters
  * `refs_list` -> enforce all entries are valid internal refs.

* Server-controlled fields:
  * `project_id` -> make sure it is the expected base64 string
  * `wb_user_id` -> make sure it is the expected base64 string
  * `wb_run_id`  -> make sure it is the `project_id:run-id` string

Note: i would not expect this to create any issues for users other than:
1. Names that are exceeding 128 characters (which according to my research is not actively happening)
2. Scripts that are currently creating invalid refs (which is better to error now then let them get written to disk).

Note: this is a stepping stone towards better refs - I am not doing any charset validation here, just length, encoding, and ref pars-ability.


weave/conftest.py
```diff
@@ -345,8 +345,9 @@ def int_to_ext(self, key, default):
         return self._int_to_ext_map[key]
 
 
-def simple_hash(s: str) -> str:
-    return hashlib.sha256(s.encode()).hexdigest()
+def b64(s: str) -> str:
```
tssweeney, line 348: This change only changes the test harness id mapping. Now we have more formal requirements on the shape of our ids this needed to be updated

weave/tests/test_op_decorator_behaviour.py
```diff
@@ -132,7 +132,7 @@ def test_sync_method_call(client, weave_obj, py_obj):
             entity="shawn",
             project="test-project",
             name="A",
-            digest="nCFXG0JUooNi5QBQGsOPNhmNVRr8ueUKvEUDQ2TeSoc",
+            digest="tGCIGNe9xznnkoJvn2i75TOocSfV7ui1vldSrIP3ZZo",
```
tssweeney, line 135: this and other digest changes are due to the project id change from hexdigest to base64 - this does not indicate a change in the underlying digest logic.

weave/tests/trace_server_clickhouse_conftest.py
```diff
@@ -108,3 +118,107 @@ def server_healthy(num_retries=1):
 
     # wait for the server to start
     return server_healthy(num_retries=30)
+
+
+class TwoWayMapping:
```
tssweeney, line 123: This entire block was taken from to `weave/conftest.py` so that we could share the converter logic

pytest.ini
```diff
@@ -14,5 +14,6 @@ filterwarnings =
     ignore::DeprecationWarning:flask.*:
     ignore::DeprecationWarning:transformers.*:
     ignore::DeprecationWarning
-
 addopts = -v
+markers =
+    skip_clickhouse_client: marks tests as requiring clickhouse client to be skipped
```
tssweeney, line 19: small print cleanup for tests

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -97,6 +97,17 @@ class CallStartCHInsertable(BaseModel):
     wb_user_id: typing.Optional[str] = None
     wb_run_id: typing.Optional[str] = None
 
+    _project_id_v = field_validator("project_id")(validation.project_id_validator)
```
tssweeney, line 100: bulk of the changes are these field_validators

weave/trace_server/validation.py
```diff
@@ -0,0 +1,67 @@
+import typing
+
+from weave.trace_server import refs_internal
+
+from . import validation_util
+
+
+def project_id_validator(s: typing.Optional[str]) -> typing.Optional[str]:
```
gtarpenning, line 8: Why do we accept None here?

weave/trace_server/validation_util.py
```diff
@@ -0,0 +1,55 @@
...
+    if lower_s != s:
+        raise ValueError(f"UUIDs must be lowercase: {s}")
+
+    try:
+        s_prime = str(uuid.UUID(s))
+    except ValueError:
+        raise ValueError(f"Invalid UUID: {s}")
+
+    if s_prime != s:
+        raise ValueError(f"Invalid UUID: {s}")
```
gtarpenning, line 19: are these errors the same for some security reason? might be nice to keep them separate, in case there is a bug that somehow only affects the UUID constructor for example.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -108,6 +119,11 @@ class CallEndCHInsertable(BaseModel):
     input_refs: typing.List[str] = []  # sadly, this is required
     output_refs: typing.List[str]
 
+    _project_id_v = field_validator("project_id")(validation.project_id_validator)
```
gtarpenning, line 122: would be nice if we can somehow dry this out, maybe with a RequiredField base model or something?

gtarpenning (approved):

weave/trace_server/validation.py
tssweeney: nice catch - fixed

weave/trace_server/validation_util.py
tssweeney: Nice - i just updated each error to be unique

PR #2075, opened 2024-08-05
chore: Fix CI build by pinning typing versions

tssweeney:


andrewtruong (approved):
Do we need all the types though?  What types are we actually using?

PR #2077, opened 2024-08-06
chore: Add validation rules and url encoding to handle outlier characters in refs

tssweeney:
There are two user-defined fields that can be included in refs (objectName, and refExtraValues). This PR adds a few things to make our refs more robust:

1. Properly quote / unqote special characters for these fields
2. Validate other fields so that the above assumption is codified in the system and we will can protect against edge cases in the future.

When paired with https://github.com/wandb/weave/pull/2074, we actually have proper ref guarantees in all cases (i believe). Now, it is still possible for clients themselves to construct invalid refs, but at least we will catch it and error server-side now. And there is a built-in fix of properly encoding the violating fields.


weave/trace_server/refs_internal.py
```diff
@@ -17,12 +17,74 @@
...
 
+valid_edge_names = (
+    DICT_KEY_EDGE_NAME,
+    LIST_INDEX_EDGE_NAME,
+    OBJECT_ATTR_EDGE_NAME,
+    TABLE_ROW_ID_EDGE_NAME,
+)
+
+
+def quote_select(s: str, quote_chars: tuple[str, ...] = ("/", ":", "%")) -> str:
```
gtarpenning, line 28: kind of a weird name, don't we mean "escape" or "safe_encode"?

gtarpenning: I guess we are quoting hmm

weave/trace_server/refs_internal.py
```diff
@@ -17,16 +17,82 @@
...
+    OBJECT_ATTR_EDGE_NAME,
+    TABLE_ROW_ID_EDGE_NAME,
+)
+
 
 class InvalidInternalRef(ValueError):
     pass
 
 
+def quote_select(s: str, quote_chars: tuple[str, ...] = ("/", ":", "%")) -> str:
```
andrewtruong, line 32: Are we missing the whitespace char here?  It could be common in e.g. names

weave/trace_server/refs_internal.py
```diff
@@ -17,16 +17,82 @@
...
+        s = s.replace(c, urllib.parse.quote(c))
+
+    return s
+
+
+def validate_extra(extra: list[str]) -> None:
+    if len(extra) % 2 != 0:
+        raise InvalidInternalRef("Extra fields must be key-value pairs.")
+    for i, e in enumerate(extra):
+        if i % 2 == 0:
```
andrewtruong, line 55: What does this branching mean?  It's not immediately obvious on reading

weave/trace_server/refs_internal.py
```diff
@@ -17,16 +17,82 @@
...
+
+
 @dataclasses.dataclass
 class InternalTableRef:
     project_id: str
     digest: str
 
+    def __post_init__(self) -> None:
+        validate_no_slashes(self.project_id, "project_id")
+        validate_no_slashes(self.digest, "digest")
```
andrewtruong, line 94: While these will guarantee no slash at init time, they won't if the user changes these fields after.

I think in general refs should be immutable, so we can just set `@dataclass(frozen=True)`

Otherwise, if refs are mutable, then we need to use a property system like:

```py
@dataclass
class InternalTableRef:
    _project_id: field(init=False, repr=False)
    
    def __post_init__(self):
        self.project_id = self._project_id
        
    @property
    def project_id(self):
        return self._project_id
    
    @project_id.setter
    def project_id(self, value):
        validate_no_slashes(value, "project_id")
        self._project_id = value
```

weave/trace_server/refs_internal.py
```diff
@@ -38,19 +104,25 @@ class InternalObjectRef:
     version: str
     extra: list[str] = dataclasses.field(default_factory=list)
 
+    def __post_init__(self) -> None:
+        validate_no_slashes(self.project_id, "project_id")
+        validate_no_slashes(self.version, "version")
+        validate_no_colons(self.version, "version")
+        validate_extra(self.extra)
```
andrewtruong, line 111: See validation comment above

weave/trace_server/refs_internal.py
```diff
@@ -60,10 +132,17 @@ class InternalCallRef:
     id: str
     extra: list[str] = dataclasses.field(default_factory=list)
 
+    def __post_init__(self) -> None:
+        validate_no_slashes(self.project_id, "project_id")
+        validate_no_slashes(self.id, "id")
+        # Note: we don't actually support extra fields for calls, but when
+        # we do, we need to add edge names to the known list
+        validate_extra(self.extra)
```
andrewtruong, line 140: See validation comment above

weave/trace_server/refs_internal.py
```diff
@@ -90,20 +169,28 @@ def parse_internal_uri(uri: str) -> Union[InternalObjectRef, InternalTableRef]:
     if kind == "table":
         return InternalTableRef(project_id=project_id, digest=remaining[0])
     elif kind == "object":
-        name, version = remaining[0].split(":")
+        name_encoded, version = remaining[0].split(":")
+        name = urllib.parse.unquote_plus(name_encoded)
+        extra = remaining[1:]
+        if len(extra) == 1 and extra[0] == "":
+            extra = []
```
andrewtruong, line 176: While correct now, I feel these large blocks are prone to errors in future.  Can we pull the quote/unquote up one level?

weave/trace_server/refs_internal.py
tssweeney: the objective is not to make a valid URL for browsers, bur rather to ensure that we can parse the ref URI. in which case whitespace is perfectly valid

weave/trace_server/refs_internal.py
tssweeney: i'll make it more clear. basically extra is always a series of `edge/value` pairs, where edge is a set of known edge types, and value is user-defined

weave/trace_server/refs_internal.py
tssweeney: great call - we should make these frozen - will do that now.

weave/trace_server/refs_internal.py
tssweeney: ack

weave/trace_server/refs_internal.py
tssweeney: pulled into a common method

andrewtruong (approved):

PR #2081, opened 2024-08-07
chore: Enable TOC, Breadcrumbs, and Collapsing Sections

tssweeney:
Enables basic navigation utilities that were for some reason disabled:
* Table of contents on all pages
* Breadcrumbs in the header
* Collapsing Sidebar sections

Note to reviewers:
1. All the `sidebar_position` are no-ops because we define the sections in the sidebar.ts

Screenshot 2024-08-06 at 19 21 21


jwlee64 (approved):
Why did we hide table of contents in the first place?

tssweeney:
> Why did we hide table of contents in the first place?

We hide it for the API docs, and I think the pattern just got copied over and over again

PR #2083, opened 2024-08-07
feat: simple column selection in call stream query

gtarpenning:
https://wandb.atlassian.net/browse/WB-20285

will also solve: https://wandb.atlassian.net/browse/WB-20325

This pr:
- uses the new callQuery `req.columns` to determine which columns to read
- adds `required_call_columns` which are the columns that will be returned with any request, regardless of column selection: `["id", "trace_id", "project_id", "op_name", "started_at"]`

This pr does **not**:
- properly handle nested selection of json fields (like `summary.total`) on db read
    - if `summary.total` is requested, we pop off `total` and treat it like `summary`, filtering down to `summary.total` in the server layer

Future followup:
- db-side json extract for more efficient querying


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1327,8 +1325,8 @@ def _ch_call_to_call_schema(ch_call: SelectableCHCallSchema) -> tsi.CallSchema:
         op_name=ch_call.op_name,
         started_at=_ensure_datetimes_have_tz(ch_call.started_at),
         ended_at=_ensure_datetimes_have_tz(ch_call.ended_at),
-        attributes=_dict_dump_to_dict(ch_call.attributes_dump),
-        inputs=_dict_dump_to_dict(ch_call.inputs_dump),
+        attributes=_nullable_any_dump_to_any(ch_call.attributes_dump or "{}"),
+        inputs=_nullable_any_dump_to_any(ch_call.inputs_dump or "{}"),
```
tssweeney, line 1329: I would recommend modifying `SelectableCHCallSchema` to correctly type the response payload so that this would be caught by typer

weave/trace_server/sqlite_trace_server.py
```diff
@@ -431,29 +439,30 @@ def process_operand(operand: tsi_query.Operand) -> str:
...
-                    output_refs=None if row[12] is None else json.loads(row[12]),
-                    summary=json.loads(row[13]) if row[13] else None,
-                    wb_user_id=row[14],
-                    wb_run_id=row[15],
-                    display_name=row[17] if row[17] != "" else None,
-                )
-                for row in query_result
-            ]
-        )
+        print("QUERY RESULT", query_result)
```
tssweeney, line 442: leftover

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -252,17 +249,17 @@ def calls_query_stream(
...
             project_id=req.project_id, include_costs=req.include_costs or False
         )
-
-        # TODO (Perf): By allowing a sub-selection of columns
-        # we will gain increased performance by not having to
-        # fetch all columns from the database. Currently all use
-        # cases call for every column to be fetched, so we have not
-        # implemented this yet.
         columns = all_call_select_columns
+        if req.columns:
```
tssweeney, line 253: it would be nice to add a comment here saying that while `columns` is user-provided, the query builder itself enforces the correct column names and prevents injection. Without such context, this looks scary to me (:

tssweeney (approved):

PR #2084, opened 2024-08-07
chore: Add automation to support server doc gen and fix python doc gen v0

tssweeney:
This PR looks huge, but that is just because there are a lot of generated files and the yarn.lock changed a lot. Changes:

1. Big one: `make generate_reference_docs` now generates python and service API docs
2. Yarn update on docusaurus related packages
3. Remove width constraint on docs to allow for more natural sizing
4. Added a little dropdown for the reference docs (see screenshot below)
5. Fixed python doc gen which was broken

Screenshot 2024-08-07 at 20 35 56

**The following is hidden:**
Screenshot 2024-08-07 at 20 37 12

Note: currently the API docs are hidden until we make our pre-release checklist, but the system is now setup.


docs/src/css/custom.css
```diff
@@ -22,8 +22,8 @@
   --ifm-color-primary-lighter: #359962;
   --ifm-color-primary-lightest: #3cad6e;
   --ifm-code-font-size: 85%;
-  --ifm-container-width: 800px;
-  --ifm-container-width-xl: 800px;
+  /* --ifm-container-width: 800px;
```
jamie-rasmussen, line 25: Could we remove entirely or add a comment about why this is commented out

jamie-rasmussen (approved):

docs/src/css/custom.css
tssweeney: removed

PR #2085, opened 2024-08-08
feat(weave): Add improvements to Dataset

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-19384

This PR adds limited mutation support to `Dataset` and "ref-get-Datasets", including new `append`, `pop`, and comparison methods, allowing users to save and pick up progress from existing datasets.

For "ref-get-Datasets", special `append` and `pop` methods are patched on to the WeaveObject to facilitate similar behaviour without resorting to ops that would otherwise be noisy.


weave/flow/dataset.py
```diff
@@ -64,3 +64,21 @@ def convert_to_table(cls, rows: Any) -> weave.Table:
...
+
+    def __eq__(self, other: Any) -> bool:
+        return self.rows == other
+
+    def __iadd__(self, rows: list[dict]) -> None:
+        if not all(isinstance(row, dict) for row in rows):
+            raise ValueError("Can only append dicts to Dataset")
+        for row in rows:
+            self.rows.append(row)
+        return self
```
andrewtruong, line 76: `__iadd__` support for `ds += [...]`

This is experimental, please see the other `__iadd__` on `WeaveObject` to see if we want to support this pattern.

From a user perspective it does feel more ergonomic, but these classes can quickly get branchy with a bunch of different cases like we have in `make_trace_object`.  I find the branching harder to read, but maybe it's ok?

weave/tests/test_dataset.py
```diff
@@ -0,0 +1,125 @@
...
+
+    ds3 = ref2.get()
+    assert ds3 == [
+        {"a": 1, "b": 2},
+        {"a": 3, "b": 4},
+        {"a": 5, "b": 6},
+    ]
+
+
+@pytest.mark.xfail(reason="Table API does not properly handle objects yet")
```
andrewtruong, line 27: TODO in a follow up PR.  This is related to Thomas's `Point(x,y)` dataset issue:
https://wandb.atlassian.net/browse/WB-20236

weave/trace/vals.py
```diff
@@ -220,8 +220,24 @@ def __repr__(self) -> str:
...
         return self._val == other
 
+    def __iadd__(self, other: Any) -> Any:
+        if getattr(self._val, "_class_name", None) == "Dataset":
+            rows = other
+            if not all(isinstance(row, dict) for row in rows):
+                raise ValueError("Can only append dicts to Dataset")
+            for row in rows:
+                self._val.rows.append(row)
+            return self
```
andrewtruong, line 234: This branching is ok, but could get noisy!

weave/trace/vals.py
```diff
@@ -536,9 +552,22 @@ def make_trace_obj(
 
     if not isinstance(val, Traceable):
         if isinstance(val, ObjectRecord):
-            return WeaveObject(
-                val, ref=new_ref, server=server, root=root, parent=parent
-            )
+            res = WeaveObject(val, ref=new_ref, server=server, root=root, parent=parent)
```
shawnlewis, line 555: Why not initialize a Dataset here instead?

shawnlewis (changes requested):

weave/trace/vals.py
andrewtruong: good point -- we're special casing our known objects here so may as well just return the actual `Dataset`

PR #2087, opened 2024-08-08
feat: integration Mistral V1.0 support

tcapelle:
This adds support for the newly released Mistral v1.0 SDK: https://github.com/mistralai/client-python/blob/main/MIGRATION.md
- Moves mistralai<1.0 to v0 and adds support for +v1.0 in the folder v1.
- I also changed the logic on the accumulator to return a `CompletionChunk` instead of a `CompletionResponse` (similar to OpenAI)


sophiamyang (approved):

weave/integrations/mistral/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    mistral_version = "1.0"  # we need to return a patching function
+
+if version.parse(mistral_version) < version.parse("1.0.0"):
+    from .v0.mistral import mistral_patcher
+
+    print(
+        f"Using MistralAI version {mistral_version}. Please consider upgrading to version 1.0.0 or later."
+    )
+else:
+    from .v1.mistral import mistral_patcher as mistral_patcher
```
vanpelt, line 17: nit: no need for "as" here.

weave/trace/op.py
```diff
@@ -45,6 +45,11 @@
 except ImportError:
     ANTHROPIC_NOT_GIVEN = None
 
+try:
+    from mistralai.types.basemodel import UNSET as MISTRAL_NOT_GIVEN
+except ImportError:
+    ANTHROPIC_NOT_GIVEN = None
```
vanpelt, line 51: This should be `MISTRAL_NOT_GIVEN`

tcapelle: oups

weave/integrations/mistral/__init__.py
tcapelle: agree

tcapelle: ruff doesn't like this, it changes it back to `from .v1.mistral import mistral_patcher as mistral_patcher`

andrewtruong:
> I have failing tests outside of my scope:
> 
> image

I don't see this -- maybe just your env?

tssweeney (approved):

tcapelle:
image

PR #2096, opened 2024-08-08
chore(weave): Add join and group by to orm

jwlee64:
Add join and group by to orm

### DIFF
--- a/weave/trace_server/orm.py
+++ b/weave/trace_server/orm.py
@@ -89,6 +89,7 @@ ColumnType = typing.Literal[
     "string",
     "datetime",
     "json",  # Represented as string in ClickHouse
+    "float",
 ]
 
 
@@ -198,9 +199,21 @@ class PreparedSelect(BaseModel):
     fields: list[str]
 
 
+class Join:
+    join_type: str
+    table: Table
+    query: tsi.Query
+
+    def __init__(self, join_type: str, table: Table, query: tsi.Query):
+        self.join_type = join_type
+        self.table = table
+        self.query = query
+
+
 class Select:
     table: Table
     all_columns: list[str]
+    joins: list[Join]
 
     action: Action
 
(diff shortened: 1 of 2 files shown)


tssweeney (approved):

weave/trace_server/tests/test_orm.py
```diff
@@ -173,3 +174,117 @@ def test_select_fields():
...
+
+    assert (
+        prepared.sql
+        == """SELECT id, name
+FROM users
+JOIN roles ON (table1.id = table2.id)"""
+    )
+
+
+def test_special_join():
```
jamie-rasmussen, line 225: Maybe a more descriptive name, `test_join_with_join_type`

weave/trace_server/tests/test_orm.py
```diff
@@ -173,3 +174,117 @@ def test_select_fields():
...
+        ],
+    )
+
+    select = (
+        table1.select()
+        .fields(["id", "name"])
+        .join(
+            "",
+            table2,
+            tsi.Query(
```
jamie-rasmussen, line 202: I know we want to guard against SQL injection, but it feels wordy and weird to me to use this MongoDB syntax for the join condition.

jwlee64: I too think its wordy, but i feel like having to parse which conditions require the mongo syntax and which dont feels annoying as well, vs everything in the mongo syntax

weave/trace_server/orm.py
```diff
@@ -210,18 +223,27 @@ class Select:
...
         self._query = None
         self._order_by = None
         self._limit = None
         self._offset = None
+        self._group_by = None
+
+    def join(self, join_type: str, table: Table, query: tsi.Query) -> "Select":
+        self.joins.append(Join(join_type, table, query))
+        for col in table.cols:
+            self.all_columns.append(col.dbname())
```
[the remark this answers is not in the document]
jwlee64: yes you do I made a change in the orm file that allows for table prefixes

PR #2097, opened 2024-08-08
chore(weave): change retry to warn and error

jwlee64:
Changing retry logs from info to warn and error under suspicion of swallowing important info

### DIFF
--- a/weave/trace_server/remote_http_trace_server.py
+++ b/weave/trace_server/remote_http_trace_server.py
@@ -67,7 +67,7 @@ def _is_retryable_exception(e: Exception) -> bool:
 
 
 def _log_retry(retry_state: tenacity.RetryCallState) -> None:
-    logger.info(
+    logger.warn(
         "retry_attempt",
         extra={
             "fn": retry_state.fn,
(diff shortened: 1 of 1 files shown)


weave/trace_server/remote_http_trace_server.py
```diff
@@ -67,7 +67,7 @@ def _is_retryable_exception(e: Exception) -> bool:
 
 
 def _log_retry(retry_state: tenacity.RetryCallState) -> None:
-    logger.info(
+    logger.warn(
```
tssweeney, line 70: this by default prints out almost nothing useful. Can you please modify these such that the printed output will help us debug in the future?

weave/trace_server/remote_http_trace_server.py
```diff
@@ -67,23 +67,29 @@ def _is_retryable_exception(e: Exception) -> bool:
...
-    logger.info(
-        "retry_attempt",
+    logger.warning(
+        "Retry attempt %d for function '%s'",
+        retry_state.attempt_number,
+        retry_state.fn.__name__,
         extra={
-            "fn": retry_state.fn,
-            "attempt_number": retry_state.attempt_number,
             "exception": str(retry_state.outcome.exception()),
```
tssweeney, line 75: wouldn't the exception be the most useful?

jwlee64: i feel like showing the exception every retry could blow up console
And since we see exception on failure, i thought no need to show on every retry

jwlee64: after reading the code a bit more I am finding that we rery every 5 mins for up to 36 hours, it makes sense to show exception

PR #2102, opened 2024-08-09
chore: Python Docs: Link to source, better pydantic objects

tssweeney:
Adds:
* Link to source code
* Better support for objects inheriting from BaseModel

Screenshot 2024-08-09 at 13 55 29

### DIFF
--- a/docs/scripts/generate_python_sdk_docs.py
+++ b/docs/scripts/generate_python_sdk_docs.py
@@ -176,13 +176,10 @@ def fix_pydantic_model(text, obj, module_name):
     return text
 
 
-def generate_module_doc_string(
-    module,
-    src_root_path="/Users/timothysweeney/Workspace/github/wandb/core/services/weave-python/weave-public/",
-):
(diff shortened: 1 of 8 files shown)


docs/scripts/generate_python_sdk_docs.py
```diff
@@ -75,32 +114,91 @@ def replace_with_empty(match):
...
+
+    if text.endswith("---"):
+        text = text[:-3]
+
+    return text
+
+
+def generate_module_doc_string(
+    module,
+    src_root_path="/Users/timothysweeney/Workspace/github/wandb/core/services/weave-python/weave-public/",
```
jwlee64, line 181: do we need a different root path here

jwlee64 (approved):
other than the one comment lgtm

docs/scripts/generate_python_sdk_docs.py
tssweeney: 👍 nice - fixing.

PR #2103, opened 2024-08-09
chore: Remove private underscores from interface classes

tssweeney:
These underscores were leftover from an error of experimentation and should be public now

Effects:
```
_CallsFilter 
_SortBy 
_OpVersionFilter 
_ObjectVersionFilter
_TableRowFilter
```

### DIFF
--- a/weave/trace_server/trace_server_interface.py
+++ b/weave/trace_server/trace_server_interface.py
@@ -222,7 +222,7 @@ class CallsDeleteRes(BaseModel):
     pass
 
 
-class _CallsFilter(BaseModel):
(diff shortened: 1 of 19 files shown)


weave/trace_server/trace_server_interface.py
```diff
@@ -222,7 +222,7 @@ class CallsDeleteRes(BaseModel):
     pass
 
 
-class _CallsFilter(BaseModel):
+class CallsFilter(BaseModel):
```
tssweeney, line 225: Primary changes are here

jamie-rasmussen (approved):

PR #2109, opened 2024-08-12
fix: Langchain integration no longer creates prohibitively long names

tssweeney:


weave/integrations/integration_utilities.py
```diff
@@ -0,0 +1,67 @@
...
+    if len(name) <= MAX_RUN_NAME_LENGTH:
+        return name
+
+    trim_amount_needed = len(name) - MAX_RUN_NAME_LENGTH
+    parts = name.split(".")
+    last_part = parts[-1]
+
+    if len(last_part) <= trim_amount_needed:
+        # In this case, the last part is shorter than the amount we need to trim.
+        raise ValueError("Unable to create a valid run name from: " + name)
```
vanpelt, line 24: Hmm, couldn't we do something else here?  I suppose we only hit this case if we had a MEGA long function name which should be very rare so probably it's ok.

vanpelt (approved):

weave/integrations/integration_utilities.py
tssweeney: this should never happen in practice, since the first part is hard coded to `langchain.XYZ` where `XYZ` is a short classname.

vanpelt: I see, I misread this initially.  This would blow up if a user had a really long parts of the string at the beginning. 👍

weave/integrations/langchain/langchain.py
```diff
@@ -188,9 +184,12 @@ def _persist_run_single(self, run: Run) -> None:
                             # of an entire Parent object.
                             parent_run = self.gc.call(wv_current_run.parent_id)
 
+            fn_name = make_pythonic_function_name(run.name)
+            complete_op_name = f"langchain.{run.run_type.capitalize()}.{fn_name}"
+            complete_op_name = truncate_op_name(complete_op_name)
             call = self.gc.create_call(
```
gtarpenning, line 190: we do use this pattern in the llamaindex integration too, but they have hardcoded "op" names which are all reasonably sized. We are relying on them staying well behaved, but because they probably will be this is just a nit.

gtarpenning (approved):
lgtm

weave/integrations/langchain/langchain.py
tssweeney: 👍 agreed

PR #2118, opened 2024-08-13
chore: Setup automated semver bumps

tssweeney:


weave/version.py
andrewtruong, line 1: Why do we have this file?  Shouldn't version just be in `pyproject.toml`?

tssweeney: This is used to have the version available at runtime

andrewtruong: Can't you use `weave.__version__`?

tssweeney: weave.__version__ IS defined by `version.py`: 
Screenshot 2024-08-16 at 14 12 15

andrewtruong: hmm, I was expecting something more like this:
https://stackoverflow.com/a/67097076

tssweeney: This works, i would like to use it as is (:

andrewtruong (approved):

PR #2122, opened 2024-08-13
chore(weave): Inline pydantic and dataclass objects

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-20357


andrewtruong:
closing in favour of: https://github.com/wandb/weave/pull/2147

PR #2123, opened 2024-08-13
fix: limit massive trace queries in ui

gtarpenning:


gtarpenning:
breaks detail trace tree view

PR #2126, opened 2024-08-13
chore: Disallow cross-project reference lookups

tssweeney:
This PR disallows cross-project reference lookups. We implicitly enabled this feature here: https://github.com/wandb/weave/pull/1967, but for a number of reasons, we want to disable it.

Please see https://github.com/wandb/core/pull/23266 for additional commentary

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -128,14 +128,6 @@ required_obj_select_columns = list(set(all_obj_select_columns) - set([]))
 ObjRefListType = list[refs_internal.InternalObjectRef]
 
 
-def make_ref_cache_key(ref: refs_internal.InternalObjectRef) -> str:
-    return f"{ref.project_id}/{ref.name}/{ref.version}"
-
-
-def make_obj_cache_key(obj: SelectableCHObjSchema) -> str:
-    return f"{obj.project_id}/{obj.object_id}/{obj.digest}"
-
-
 class ClickHouseTraceServer(tsi.TraceServerInterface):
     def __init__(
         self,
@@ -727,6 +719,10 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
 
         # Lookup data for each project, scoped to each project
         final_result_cache: typing.Dict[str, typing.Any] = {}
+
+        def make_ref_cache_key(ref: refs_internal.InternalObjectRef) -> str:
+            return ref.uri()
+
         for project in refs_by_project_id:
             project_refs = refs_by_project_id[project]
             project_results = self._refs_read_batch_within_project(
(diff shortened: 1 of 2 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -702,60 +712,77 @@ def refs_read_batch(self, req: tsi.RefsReadBatchReq) -> tsi.RefsReadBatchRes:
...
-                if len(conds) > 0:
-                    conditions = [combine_conditions(conds, "OR")]
-                    objs = self._select_objs_query(
-                        project_id, conditions=conditions, parameters=parameters
-                    )
-                    for obj in objs:
-                        root_val_cache[make_obj_cache_key(obj)] = json.loads(
-                            obj.val_dump
-                        )
+            for ref_index, ref in enumerate(refs):
```
tssweeney, line 753: Mostly an indentation change - hide whitespace for easier review

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -897,7 +953,7 @@ def resolve_extra(extra: list[str], val: typing.Any) -> typing.Any:
                         extra_result.remaining_extra, extra_result.val
                     )
 
-        return tsi.RefsReadBatchRes(vals=[r.val for r in extra_results])
+        return [r.val for r in extra_results]
```
gtarpenning, line 956: why can we return just the list here?

gtarpenning: ah i see on line 736

gtarpenning (approved):
this change somehow makes this behemoth function (now two) easier to reason about. nice

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -702,16 +704,43 @@ def refs_read_batch(self, req: tsi.RefsReadBatchReq) -> tsi.RefsReadBatchRes:
...
+            )
+            for ref, result in zip(project_refs, project_results):
+                final_result_cache[make_ref_cache_key(ref)] = result
+
+        # Return the final data payload
+        vals = [final_result_cache[make_ref_cache_key(ref)] for ref in parsed_refs]
+        return tsi.RefsReadBatchRes(vals=vals)
+
+    def _refs_read_batch_within_project(
+        self, project_id_scope: str, parsed_refs: ObjRefListType
```
gtarpenning, line 739: not 100% sure what the value is in changing `project_id` to `project_id_scope`

tssweeney: mainly for code readability - basically `project_id` is part of a `ref`, so it was a little confusing. `project_id_scope` just is a more clear noun

weave/trace_server/trace_server_converter.py
```diff
@@ -45,8 +47,14 @@ def replace_ref(ref_str: str) -> str:
...
-        if isinstance(obj, str) and obj.startswith(weave_prefix):
-            return typing.cast(B, replace_ref(obj))
+        if isinstance(obj, str):
+            if obj.startswith(weave_prefix):
+                return typing.cast(B, replace_ref(obj))
+            elif obj.startswith(weave_internal_prefix):
+                # It is important to raise here as this would be the result of
+                # an external client attempting to write internal refs directly.
+                # We want to maintain full control over the internal refs.
+                raise ValueError("Encountered internal ref in external data.")
```
gtarpenning, line 57: Is there a reason we might want to obfuscate the reason to raise here? I can't actually think of a good reason, other than, if this only happens when someone is doing something they shouldn't, returning a handy error explaining why doesn't feel right.

tssweeney: Hmm - interesting. These will get transformed into 500s i believe, but really should be a 400.

tssweeney: ok, modified the tests to assert on status code and removed these messages.

PR #2137, opened 2024-08-14
feat: ref expansion in the server

gtarpenning:
This PR implements server-side ref expansion in the calls query. When `expand_columns` are passed (dot.notation strings corresponding to columns), all refs in those exact columns will be expanded. 

Example: 
```python
r = calls_query_stream(req=CallsQueryReq(project_id=client._project_id()))
# [{"id"...., "output": "weave:///object/...."}]
r = calls_query_stream(
     req=CallsQueryReq(
            expand_columns=["output.object"],
     )
)
# [{"id"...., "output": {"object": {"hello": 123}}}]
```

In the case of nested ref expansion, where a field is requested inside an object that is itself a ref, multiple expand column strings must be provided. For example, to fully de-ref:

```python
primative = 123
object2 = {"b": primative}
object3 = {"c": object2}

@weave.op
def fn():
   return object3

# to get 123: 
expand_columns = ["output", "output.c", "output.c.b"]
```

To do the above ref expansion performantly:
- calls are de-reffed in batches
    * batches start out as size `10`, so we can quickly return something to the stream, but batch size doubles with each successive batch. Max batch size is capped by the number of refs we request at any given time to an arbitrary limit of `1000`. 
- refs are sent off for resolution in batches as well
    * batches consist of refs at the same depth within the call objects. The first batch of refs will consist of all refs at depth 0 (ex. `["output", "inputs"]`) and increasing from there. The second batch might be columns: `["output.object", "inputs.param1", "inputs.param2"]`.

Known limitations: 
- This does not handle the case when the expand column refers to a list of refs. For example `inputs.self.scorers` when there are two scorers. There are a couple ways this could be supported. 
   1. inspect and detect when an expand column points to a list, then add special handling to iterate through the values and deref them. We might consider returning a dict with the keys as the ref indexes of the original list, values the expanded refs. Either way this will require refactor of the clickhouse batched handling. 
   2. Require specific references to indices in the list when expanding. Example: `inputs.self.scorers[0], inputs.self.scorers[1]`.  

Server PR:
- https://github.com/wandb/core/pull/23329


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +294,143 @@ def calls_query_stream(
...
+        return ".".join(parts) or None, cur
+
+    def _get_refs_to_resolve(
+        self, calls: list[dict[str, typing.Any]], expand_columns: typing.List[str]
+    ) -> typing.Dict[tuple[int, str, str], str]:
+        # First get refs from call batch, store them where we can lookup them later
+        # format: (call_index, column_name, column_prefix) -> ref
+        refs_to_resolve: typing.Dict[tuple[int, str, str], str] = {}
+        for i, call in enumerate(calls):
+            for col in expand_columns:
```
gtarpenning, line 374: order expansion columns by prefix

"output.model"

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +294,143 @@ def calls_query_stream(
...
+        # Find refs in the call batch, or skip when provided (if recursing)
+        refs_to_resolve = refs_to_resolve or self._get_refs_to_resolve(
+            calls, expand_columns
+        )
+        if not refs_to_resolve:
+            return calls
+
+        vals = self._refs_read_batch(list(refs_to_resolve.values()), ref_cache)
+        if len(vals) != len(refs_to_resolve):
+            raise ValueError(f"Expected {len(refs_to_resolve)} refs, got {len(vals)}")
```
gtarpenning, line 404: should always be t

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +294,143 @@ def calls_query_stream(
...
+
+        unresolved_refs = {}
+        for (i, col, col_prefix), val in zip(refs_to_resolve, vals):
+            if col_prefix != col:
+                next_part = col.replace(col_prefix, "")
+                if next_part.startswith("."):
+                    next_part = next_part[1:]
+                for part in next_part.split("."):
+                    if part not in val:
+                        raise ValueError(
```
gtarpenning, line 414: this can be improved with prefix management

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +294,143 @@ def calls_query_stream(
...
+        if not req.expand_columns:
+            for row in raw_res:
+                yield tsi.CallSchema.model_validate(
+                    _ch_call_dict_to_call_schema_dict(dict(zip(select_columns, row)))
+                )
+
+        else:
+            batch_size = 10
+            batch = []
+            ref_cache: typing.Dict[str, typing.Any] = {}
```
gtarpenning, line 307: uper bound that?

gtarpenning: lets use an LRU cache

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +293,108 @@ def calls_query_stream(
...
+                    for call in hydrated_batch:
+                        yield tsi.CallSchema.model_validate(call)
+
+                    batch = []
+                    # dynamically increase the batch size
+                    batch_size = min(1000, batch_size * 2)
+
+            hydrated_batch = self._hydrate_calls(batch, req.expand_columns, ref_cache)
+            for call in hydrated_batch:
+                yield tsi.CallSchema.model_validate(call)
```
gtarpenning, line 327: if we abandon this, then we can avoid re-nesting the call dicts

weave/trace_server/trace_server_common.py
```diff
@@ -0,0 +1,49 @@
+import copy
+from collections import OrderedDict
+from typing import Any, Dict
+
+
+def get_nested_key(d: Dict[str, Any], col: str) -> Any:
```
jamie-rasmussen, line 6: I'm kind of confused by this function, maybe some unit tests or more docstring would help. E.g. 
`get_nested_key({"a": {"b": {"c": "d"}}}, "foo")` seems like it would give you `{}`, why is that a different case than "a.b.c.e"?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +297,101 @@ def calls_query_stream(
...
+
+    def _hydrate_calls(
+        self,
+        calls: list[dict[str, typing.Any]],
+        expand_columns: typing.List[str],
+        ref_cache: typing.Dict[str, typing.Any],
```
jamie-rasmussen, line 345: Could this be typed `LRUCache`?

jamie-rasmussen (approved):

weave/trace_server/trace_server_common.py
```diff
@@ -0,0 +1,49 @@
...
+    curr[keys[-1]] = copy.deepcopy(val)
+
+
+class LRUCache(OrderedDict):
+    def __init__(self, max_size: int = 1000, *args: Any, **kwargs: Dict[str, Any]):
+        self.max_size = max_size
+        super().__init__(*args, **kwargs)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        if len(self) >= self.max_size:
```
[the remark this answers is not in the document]
gtarpenning: 100% correct, although in practice its a cache, so we would expect the consumer to get before setting generally. I guess if gets stale you might just set... a bug regardless, nice.

weave/trace_server/trace_server_common.py
```diff
@@ -0,0 +1,49 @@
...
+    get_nested_key({"a": {"b": {"c": "d"}}}, "a.b.c") -> "d"
+    get_nested_key({"a": {"b": {"c": "d"}}}, "a.b.c.e") -> None
+    """
+
+    def _get(dictionary: Dict[str, Any], key: str) -> Any:
+        if isinstance(dictionary, dict):
+            return dictionary.get(key, {})
+        return None
+
+    keys = col.split(".")
```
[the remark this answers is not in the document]
gtarpenning: yes :'( Do you think there is any hope for escaping dict keys? or are we doomed to a major rewrite?

weave/trace_server/trace_server_common.py
gtarpenning: Fixed, and added tests.

weave/trace_server/trace_server_common.py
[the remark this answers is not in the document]
jamie-rasmussen: I think it will be a pain for us at some point, but one we can incur once we've gotten a complaint about it.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +297,103 @@ def calls_query_stream(
...
+            columns = expand_column_by_depth[depth]
+            refs_to_resolve = self._get_refs_to_resolve(calls, columns)
+            if not refs_to_resolve:
+                continue
+
+            vals = self._refs_read_batch(list(refs_to_resolve.values()), ref_cache)
+            assert len(vals) == len(refs_to_resolve), "ref read batch inconsistency"
+
+            for (i, col), val in zip(refs_to_resolve, vals):
+                set_nested_key(calls[i], col, val)
```
gtarpenning, line 394: @shawnlewis  mentioned we might want to inject `{_ref: <ref_uri>}` to de-reffed objects here

gtarpenning: I think we should punt this to a followup pr.

gtarpenning: Took a stab in [1c52fa8](https://github.com/wandb/weave/pull/2137/commits/1c52fa87249b680caefbac712e4284d3a54fc129)

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTableButtons.tsx
```diff
@@ -90,6 +103,16 @@ export const ExportSelector = ({
       ? visibleColumns
       : undefined;
     const startTime = Date.now();
+
+    visibleColumns.sort((a, b) => b.length - a.length);
+    const minimalColumns: string[] = [];
+    for (const col of visibleColumns) {
+      if (minimalColumns.some(minimalCol => minimalCol.startsWith(col))) {
```
tssweeney, line 110: Why are you calculating "minimalColumns"? This seems incorrect. If my visible columns are: `a` and `a.b`, why do i only send over `a.b`?

weave/trace_server/refs_internal.py
```diff
@@ -217,3 +217,7 @@ def string_will_be_interpreted_as_ref(s: str) -> bool:
     return s.startswith(f"{WEAVE_INTERNAL_SCHEME}:///") or s.startswith(
         f"{WEAVE_SCHEME}:///"
     )
+
+
+def is_ref_str(val: Any) -> bool:
```
tssweeney, line 222: nit: `is_ref_str` might instead be: `will_be_interpreted_as_ref_str`. Just because we are only checking the schema, not the valid shape of the ref

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +297,104 @@ def calls_query_stream(
...
+            columns = expand_column_by_depth[depth]
+            refs_to_resolve = self._get_refs_to_resolve(calls, columns)
+            if not refs_to_resolve:
+                continue
+
+            refs = list(refs_to_resolve.values())
+            vals = self._refs_read_batch(refs, ref_cache)
+            for (i, col), val, ref in zip(refs_to_resolve, vals, refs):
+                if isinstance(val, dict):
+                    val["_ref"] = ref
```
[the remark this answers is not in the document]
tssweeney: ok, so in the main app, when manually expanding refs, i actually put everything expanded as follows:
```
{
   __ref__: ref,
   __value__: actual value
}
```

see 

```
export const makeRefExpandedPayload = <T = any>(
  originalRef: string,
  refData: T
): ExpandedRefWithValue<T> => {
  return {
    [EXPANDED_REF_REF_KEY]: originalRef,
    [EXPANDED_REF_VAL_KEY]: refData,
  };
};
```

The approach as implemented in this PR does not support primitives and technically could have a collision with a key of `_ref`.

Another approach is to have a map of {path: ref} expansions.

You are probably ok for now, but i suspect this will cause us a little mind bender in the future.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +297,104 @@ def calls_query_stream(
...
+            expand_column_by_depth[col.count(".")].append(col)
+
+        for depth in sorted(expand_column_by_depth.keys()):
+            columns = expand_column_by_depth[depth]
+            refs_to_resolve = self._get_refs_to_resolve(calls, columns)
+            if not refs_to_resolve:
+                continue
+
+            refs = list(refs_to_resolve.values())
+            vals = self._refs_read_batch(refs, ref_cache)
```
tssweeney, line 391: This is a security gap. Please use the _refs_read_batch_within_project and scope to the incoming project.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTableButtons.tsx
tssweeney: I see, so `columns` should be the "leaf columns"

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -292,10 +297,104 @@ def calls_query_stream(
...
+                batch.append(call_dict)
+
+                if len(batch) >= batch_size:
+                    hydrated_batch = self._hydrate_calls(
+                        batch, req.expand_columns, ref_cache
+                    )
+                    for call in hydrated_batch:
+                        yield tsi.CallSchema.model_validate(call)
+
+                    # *** Dynamic Batch Size ***
```
tssweeney, line 325: there are edges where this will fail the 1000 limit. I think you should probably make a way to skip that 1000 limit in these cases so that we don't accidentally fail an entire request do to an invalid estimation here.

gtarpenning: will have to anyway to do the above comment (using `_refs_read_batch_within_project`)

weave/trace_server/clickhouse_trace_server_batched.py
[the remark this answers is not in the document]
gtarpenning: I added a check to not overwrite any existing `_ref` key. happy to update this and make it more robust in the future, including adding a way of attaching refs to primitives.

tssweeney (approved):

PR #2139, opened 2024-08-15
fix: update weave client with formatted project name

gtarpenning:
https://wandb.atlassian.net/browse/WB-20396

### DIFF
--- a/weave/legacy/wandb_interface/project_creator.py
+++ b/weave/legacy/wandb_interface/project_creator.py
@@ -39,12 +39,17 @@ def wandb_logging_disabled() -> typing.Iterator[None]:
     wandb.termerror = original_termerror
 
 
-def ensure_project_exists(entity_name: str, project_name: str) -> None:
+def ensure_project_exists(entity_name: str, project_name: str) -> typing.Optional[str]:
     with wandb_logging_disabled():
         return _ensure_project_exists(entity_name, project_name)
 
 
-def _ensure_project_exists(entity_name: str, project_name: str) -> None:
+def _ensure_project_exists(entity_name: str, project_name: str) -> typing.Optional[str]:
+    """
+    Ensures that a W&B project exists by trying to access it, returns the project_name,
+    which is not guaranteed to be the same if the provided project_name contains invalid
+    characters.
+    """
     wandb_logging_disabled()
     api = InternalApi({"entity": entity_name, "project": project_name})
     # Since `UpsertProject` will fail if the user does not have permission to create a project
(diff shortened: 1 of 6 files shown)


weave/legacy/wandb_interface/project_creator.py
```diff
@@ -72,4 +77,6 @@ def _ensure_project_exists(entity_name: str, project_name: str) -> None:
                 raise UnableToCreateProject(
                     f"Failed to create project {entity_name}/{project_name}"
                 )
-    return
```
tssweeney, line 75: Is there ever a case that we need entity too?

tssweeney: in the future we might want to get the project id from here too for client-side digest calcs

tssweeney: I'd recommend making a structured return object that we can extend

gtarpenning: right, on it

gtarpenning: @tssweeney to get the entity back we are going to have to make a change in the wandb sdk

tssweeney: we don't need it right now ... just want to make sure the return type is a class that is structured

weave/legacy/wandb_interface/project_creator.py
```diff
@@ -72,4 +81,5 @@ def _ensure_project_exists(entity_name: str, project_name: str) -> None:
...
+    project_name_maybe: typing.Optional[str] = None if not project else project["name"]
```
tssweeney, line 84: nit: project should never be none due to check above

tssweeney (approved):

weave/trace_server/trace_server_interface.py
```diff
@@ -565,9 +565,15 @@ class FileContentReadRes(BaseModel):
     content: bytes
 
 
+class EnsureProjectExistsRes(BaseModel):
+    project_name: typing.Optional[str]
```
tssweeney, line 569: this should not be optional

weave/weave_client.py
```diff
@@ -380,7 +380,10 @@ def __init__(
         self.ensure_project_exists = ensure_project_exists
 
         if ensure_project_exists:
-            self.server.ensure_project_exists(entity, project)
+            resp = self.server.ensure_project_exists(entity, project)
+            if resp.project_name:
```
tssweeney, line 384: this is not needed when project_name is not optional

Issue #1380, opened 2024-03-27
`@weave.type()` does not work with pydantic `BaseModel`


willbakst:
Example buggy code below outputs `AttributeError: 'Book' object has no attribute '__pydantic_fields_set__'`

```python
import weave
from pydantic import BaseModel


@weave.type()
class Book(BaseModel):
    title: str
    author: str


book = Book(title="title", author="author")
```

scottire:
This should be fixed if you inherit instead from `weave.Object` rather than using `weave.type()`

```python
import weave
from pydantic import BaseModel


class Book(weave.Object, BaseModel):
    title: str
    author: str

@weave.op()
def use_book(book: Book):
    print(f"Title: {book.title}, Author: {book.author}")
    return book.title

weave.init('william')
book = Book(title="title", author="author")
print(book.__pydantic_fields_set__)
use_book(book)
print(book.__pydantic_fields_set__)
```

mikeedjones:
I'm not sure that this pretty drastic change to the codebase should be necessary to enable tracing of functions which use pydantic? Is there an issue to address this and improve developer experience? Can the args be recast inside weave.op instead?

patricknnamdi:
For pre-existing codebases having to adapt every model to include weave.Object is quite inconvenient. Additionally, it would be immensely helpful if this was in the docs somewhere. Unfortunately, getting weave to work with my application wasn't as simple as adding decorators everywhere due to this issue.

andrewtruong:
Hey @mikeedjones , @patricknnamdi 

You're right -- inheriting from `weave.Object` should not be necessary to enable function tracing, and fortunately it's not!

Hopefully the minimal example below is more clear at highlighting:
1. You don't need to change your objects to inherit from `weave.Object`;
2. You just need to apply the `@weave.op` decorator on top of the function you want to trace
3. We don't muck with your existing objects, but they are still tracked and available to view in the UI

```py
import weave
from pydantic import BaseModel


class Book(BaseModel):
    title: str
    author: str


@weave.op
def use_book(book: Book) -> str:
    return f"Hello, we are using {book.title} by {book.author}"


weave.init("testing78")

book = Book(title="Much Ado About Nothing", author="William Shakespeare")
use_book(book)
```
![image](https://github.com/user-attachments/assets/721f03d6-3391-4be1-a091-75d3d6e6e98c)
![image](https://github.com/user-attachments/assets/669bdd3c-eccd-4a8f-8786-5da310b83413)

[...]

PR #2140, opened 2024-08-15
chore(weave): Add client-side quoting for object names

andrewtruong:
## Summary
1. Adds client-side checks to ensure names are valid before hitting the Trace Server layer

__NOTE: Do not merge until companion UI fixes land__

## Resolves:
- https://wandb.atlassian.net/browse/WB-20402

## Related to:
- https://github.com/wandb/weave/issues/1380#issuecomment-2291658250 (invalid char caught at lower level)


andrewtruong:
Close in favour of https://github.com/wandb/weave/pull/2145

PR #2141, opened 2024-08-15
chore(weave): Use non-mutable defaults for clickhouse schema

andrewtruong:
## Summary
1. Don't use mutable defaults!


weave/trace_server/clickhouse_schema.py
```diff
@@ -16,7 +16,9 @@ class CallStartCHInsertable(BaseModel):
     attributes_dump: str
     inputs_dump: str
     input_refs: typing.List[str]
-    output_refs: typing.List[str] = []  # sadly, this is required
+    output_refs: typing.List[str] = Field(
+        default_factory=list
+    )  # sadly, this is required
```
andrewtruong, line 21: I'm not sure what this "is required" means.  What does it mean?

gtarpenning: Inserting into the DB fails without `[]`, as the insertion does a concat with the existing list, so if its a `nonetype` the db throws

gtarpenning: what does this object look like if you don't specify `output_refs` now?

andrewtruong: Ok I think that's fine behaviour then, not sure why the "sadly" part

This has the same behaviour as before, just not using a mutable default.

andrewtruong: Here's what it will look like:

image

gtarpenning (approved):

PR #2144, opened 2024-08-15
chore(weave): Tidy up Trace Server Interface

andrewtruong:
## Summary
1. Simplifies the `TraceServerInterface` class and makes it a Protocol
2. Simplifies `typing` imports


weave/trace_server/trace_server_interface.py
```diff
@@ -66,43 +65,41 @@ class CallSchema(BaseModel):
     # Name of the calling function (op)
     op_name: str
     # Optional display name of the call
-    display_name: typing.Optional[str] = None
+    display_name: Optional[str] = None
 
     ## Trace ID
```
gtarpenning, line 70: can we clean up the comments while we are here? These should all be single `#`

gtarpenning (approved):
Another micro nit/styling thing that can be cleaned

PR #2147, opened 2024-08-16
chore(weave): Inline pydantic and dataclass objects on weave.Object

andrewtruong:
## Summary
1. Inlines pydantic and dataclass objects, keeping the `ObjectRecord` shape but skipping `obj_create` (and inlined objects have no refs)
2. Saves _____ time for (benchmark obj) when saving deeply nested objects
3. Avoids cluttering the top-level Object space in the UI (and also from the object query)

## Resolves
1. https://wandb.atlassian.net/browse/WB-20357

### DIFF
--- a/weave/integrations/anthropic/anthropic_test.py
+++ b/weave/integrations/anthropic/anthropic_test.py
@@ -1,5 +1,4 @@
 import os
-from typing import Any
 
 import pytest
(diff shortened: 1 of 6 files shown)


andrewtruong:
closing in favour of: https://github.com/wandb/weave/pull/2271

PR #2150, opened 2024-08-16
chore(weave): Add `invoke` as a valid "infer method"

andrewtruong:
## Summary
1. Adds `invoke` as an option for inference method

## Resolves
- https://wandb.atlassian.net/browse/WB-20407

### DIFF
--- a/weave/flow/model.py
+++ b/weave/flow/model.py
@@ -50,5 +50,5 @@ def get_infer_method(model: Model) -> Callable:
         if (infer_method := getattr(model, name, None)) is not None:
             return infer_method
     raise MissingInferenceMethodError(
-        f"Missing a method with name like ({INFER_METHOD_NAMES})"
+        f"Missing a method with name in ({INFER_METHOD_NAMES})"
     )
--- a/weave/flow/model.py
+++ b/weave/flow/model.py
@@ -41,7 +41,7 @@ class Model(Object):
             if infer_method := getattr(self, name, None):
                 return infer_method
         raise MissingInferenceMethodError(
-            f"Missing a method with name like ({INFER_METHOD_NAMES})"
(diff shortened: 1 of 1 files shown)


weave/flow/model.py
```diff
@@ -32,20 +37,18 @@ def predict(self, input_data: str) -> dict:
...
-    for infer_method_names in ("predict", "infer", "forward"):
-        infer_method = getattr(model, infer_method_names, None)
-        if infer_method:
+    for name in INFER_METHOD_NAMES:
+        if (infer_method := getattr(model, name, None)) is not None:
             return infer_method
-    raise ValueError(
-        f"Model {model} does not have a predict, infer, or forward method."
+    raise MissingInferenceMethodError(
+        f"Missing a method with name like ({INFER_METHOD_NAMES})"
```
gtarpenning, line 53:
```suggestion
        f"Missing a method with name in ({INFER_METHOD_NAMES})"
```

I think "like" is a bit too broad, we need it to explicitly be included right?

weave/flow/model.py
```diff
@@ -32,20 +37,18 @@ def predict(self, input_data: str) -> dict:
...
-        for infer_method_names in ("predict", "infer", "forward"):
-            infer_method = getattr(self, infer_method_names, None)
-            if infer_method:
+        for name in INFER_METHOD_NAMES:
+            if infer_method := getattr(self, name, None):
                 return infer_method
-        raise ValueError(
-            f"Model {self} does not have a predict, infer, or forward method."
+        raise MissingInferenceMethodError(
+            f"Missing a method with name like ({INFER_METHOD_NAMES})"
```
gtarpenning, line 44: same

gtarpenning (approved):
copy nit

PR #2145, opened 2024-08-15
DNM: fix: Clients now correctly quote/unquote ref parts

tssweeney:
**Summary:** This PR hopefully 🤞 solves >90% of our remaining Ref/Naming issues - meaning users should be able to freely create entities/projects (as long as Gorilla accepts them) and name their objects/ops anything they want (baring a `%` symbol). Under the hood, I close a few remaining constraints / parsing issues on the server side and mostly modify the UI and Python clients to correctly read/write quoted refs.

---

High Level Improvements:
1. Python Client will always produce valid refs (ie. it correctly quotes when needed)
   * Benefit to user: no more crashed python scripts when bad chars are in the name
   * Note: the one case of `%` in the name is still illegal. We sanitize on the python client for this specific case.
2. Python Client is able to consume quoted refs (ie. it correctly unquotes when needed)
   * Benefit to user: able to correctly walk ref data edges with bad chars in the name
3. UI Client will always produce valid refs (ie. it correctly encodes when needed)
   * Benefit to user: able to link safely when dealing with bad characters
4. UI Client is able to consume encoded refs (ie. it correctly decodes when needed)
   * Benefit to user: able to correctly display links with bad characters

Smaller Details:
* Adds correct quoting of URLs to printing URLs
* Adds server-side entity/project de-quoting

Companion PR: https://github.com/wandb/core/pull/23333
* Correctly routes when parts are encoded
* Correctly de-quoting encoded refs on project lookup

---

This PR finishes up the Ref work to ensure we support any ref the user throws at us. 

**Context:**
In a recent PR: https://github.com/wandb/weave/pull/2077, I implemented a more robust parser and checker to ensure Refs and Names conform to expectations. This stopped the "bleeding" where we were storing invalid Refs.

However, our two clients Python and UI had not been updated to reflect the new rules. In fact, the Python client was still creating bad refs, and the UI was overly restrictive with what it thought a ref could be. So double edge sword. As a result, users still hit crashes (either python write time or ui read time).

Moreover, our Ref parsing logic is a mess (especially in the UI) as it is spread across many locations & riddled with legacy junk.

*A little terminology note*: in PY, URL escaping is called `quoting` and in TS it is `encoding`. When used in this PR, they follow the language-specific convention, but logically mean the same concept.

**Understanding Refs:**
See https://www.notion.so/wandbai/WIP-Robust-Refs-550bf06f4a244a1ab4fdd26133952e2c for more details.

The important part to understand is the ref format:

```
Calls:
[...]

PR #2156, opened 2024-08-16
fix: Fixes issues with special characters in object name

tssweeney:
Simplified version of https://github.com/wandb/weave/pull/2145 that takes a different approach to fixing the problem by implementing constraints on the names rather than url encoding.

Allowed characters: `\w._-`

This PR does 3 things:
1. Client:
   1. Sanitize object names when writing names
2. Backend:
   1. Enforces a strict character subset on object names
   2. Simplified url quoting/unquoting of ref extras
3. UI:
   1. Updates `Browse3` to use decoded params from the URL, which avoids creating invalid refs on UI side.
   2. Correctly handles URL encoded ref extra parts.

The effective result of this is:
1. (Client side) All updated clients will effectively never create an invalid object ever again
2. (Backend) Even old clients will be denied the ability to make objects that are not valid
3. (UI) Correctly reads/writes refs in the encoded format.

Deployment can be an any order, hence a common PR to centralize all these changes


weave/trace_server/refs_internal.py
```diff
@@ -29,18 +31,15 @@ class InvalidInternalRef(ValueError):
     pass
 
 
-def quote_select(s: str, quote_chars: tuple[str, ...] = ("/", ":", "%")) -> str:
+def extra_value_quoter(s: str, quote_chars: tuple[str, ...] = ("%", "/")) -> str:
```
tssweeney, line 34: I think this is correct - perhaps we just want to quote everything? `urllib.parse.quote(s, safe="")`? That might be the better approach

weave/weave_client.py
```diff
@@ -999,4 +1004,16 @@ def redact_sensitive_keys(obj: typing.Any) -> typing.Any:
     return obj
 
 
+def sanitize_object_name(name: str) -> str:
+    # Replaces any non-alphanumeric characters with a single dash and removes
+    # any leading or trailing dashes. This is more restrictive than the DB
+    # constraints and can be relaxed if needed.
+    res = re.sub(r"[._-]+", "-", re.sub(r"[^\w._]+", "-", name)).strip("-_")
```
tssweeney, line 1011: i am wondering if this will result in any changes to integration names?

weave/trace_server/validation.py
```diff
@@ -57,7 +58,20 @@ def wb_run_id_validator(s: typing.Optional[str]) -> typing.Optional[str]:
     return s
 
 
+def _validate_object_name_charset(name: str) -> None:
+    # Object names must be alphanumeric with dashes
+    invalid_chars = re.findall(r"[^\w._-]", name)
```
tssweeney, line 63: This is quite restrictive - could frustrate people?

weave-js/src/react.test.ts
```diff
@@ -34,7 +34,7 @@ describe('parseRef', () => {
     });
     it('parses a ref with spaces in entity', () => {
       const parsed = parseRef(
-        'weave:///Entity%20Name/project/object/artifact-name:artifactversion'
+        'weave:///Entity Name/project/object/artifact-name:artifactversion'
```
tssweeney, line 37: This was not a valid ref - in fact it was a bug in ref creation from Browse3

weave-js/src/react.test.ts
```diff
@@ -102,20 +102,6 @@ describe('parseRef', () => {
         weaveKind: 'object',
       });
     });
-    it('parses a ref with escaped spaces in name and projectName', () => {
```
tssweeney, line 105: Same here, not a valid ref

tssweeney:
> Attaching [this ticket ](https://wandb.atlassian.net/browse/WB-20526)from triage to this pr, curly braces should be supported.

This PR should sanitize these away

weave/trace_server/refs_internal.py
```diff
@@ -29,23 +31,8 @@ class InvalidInternalRef(ValueError):
...
-    # We have to be careful here since quoting creates new "%" characters.
-    # We don't want to double-quote characters.
-    if "%" in quote_chars and quote_chars[-1] != "%":
-        raise ValueError("Quoting '%' must be the last character in the list.")
-
-    for c in quote_chars:
-        s = s.replace(c, urllib.parse.quote(c))
-
-    return s
+def extra_value_quoter(s: str) -> str:
```
gtarpenning, line 34: Screenshot 2024-08-27 at 3 37 44 PM
consider including a note on which chars this does not quote

weave/trace_server/validation_util.py
```diff
@@ -58,6 +58,6 @@ def require_internal_ref_uri(
 
 
 def require_max_str_len(s: str, length: int) -> str:
-    if len(s) >= length:
+    if len(s) > length:
```
gtarpenning, line 61: we were truncating to 128 right? If so, this off by one fix will make the sentry error way less common presumably?

gtarpenning: we could add a test for this, that is **doesn't** throw when at 128

gtarpenning (approved):

weave/trace_server/validation_util.py
tssweeney: adding a test

PR #2165, opened 2024-08-19
fix: serialization bug when a Ref is put directly into a dict

shawnlewis:
Here's the test case:

```
def test_ref_in_dict(client):
    ref = client._save_object({"a": 5}, "d1")

    # Put a ref directly in a dict.
    ref2 = client._save_object({"b": ref}, "d2")

    obj = weave.ref(ref2.uri()).get()
    assert obj["b"] == {"a": 5}
```

Prior to the fix, we would actually save the Ref object itself as an ObjectRecord. After the fix we correctly save the ref as its uri string.

### DIFF
--- a/weave/tests/test_weave_client.py
+++ b/weave/tests/test_weave_client.py
@@ -1324,3 +1324,13 @@ def test_summary_tokens_cost_sqlite(client):
 
     assert noCostCallSummary is None
     assert withCostCallSummary is None
+
+
+def test_ref_in_dict(client):
+    ref = client._save_object({"a": 5}, "d1")
+
+    # Put a ref directly in a dict.
+    ref2 = client._save_object({"b": ref}, "d2")
+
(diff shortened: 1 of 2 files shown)


weave/tests/test_weave_client.py
```diff
@@ -1324,3 +1324,13 @@ def test_summary_tokens_cost_sqlite(client):
...
     assert withCostCallSummary is None
+
+
+def test_ref_in_dict(client):
+    ref = client._save_object({"a": 5}, "d1")
+
+    # Put a ref directly in a dict.
+    ref2 = client._save_object({"b": ref}, "d2")
+
+    obj = weave.ref(ref2.uri()).get()
```
andrewtruong, line 1335: Why do this instead of just `ref2.get()`?

andrewtruong (approved):

PR #2071, opened 2024-08-05
docs(cookbook): Governance Dashboard

NiWaRe:
Following the new guide on cookbooks this is a short-form text-code tutorial about how to use the Weave APIs to create a custom live prod dashboard (as requested by many companies). As a short term solution, as mid-term this is a best practice of how to integrate with Weave.

PR #2166, opened 2024-08-19
chore(docs): add feedback cookbook

scottire:


docs/docs/reference/gen_notebooks/feedback_prod.md
```diff
@@ -0,0 +1,179 @@
...
+
+:::
+
+
+
+<img src="http://wandb.me/logo-im-png" width="400" alt="Weights & Biases" />
+<!--- @wandbcode{feedback-colab} -->
+
+
+It is often hard to automatically evaluate a generated LLM response so, depending on your risk tolerance, you can gather direct user feedback to use to find areas to improve.
```
NiWaRe, line 20: "you can gather direct user feedback to use to find areas to improve." -> "you can gather direct user feedback to find areas to improve."

docs/docs/reference/gen_notebooks/feedback_prod.md
```diff
@@ -0,0 +1,179 @@
...
+import weave
+from openai import OpenAI
+import streamlit as st
+
+st.title("Add feedback")
+
+# highlight-next-line
+@weave.op
+def chat_response(prompt):
+    stream = client.chat.completions.create(
```
NiWaRe, line 55: If we don't put in an actual model from Weave we should defintely reference to another upcoming cookbook or docs page for how to deploy models from Weave (to close the loop from experimentation to production).

For #2071  I'm using something like this: 

```python
@weave.op()
def production_chat_response(model_url: str, prompt: str):
    """
    The main model function that will take the prompt and the model_url to generate a response.
    Optionally, this function call also be decorated with @weave.op(). 
    """
    
    if model_url.startswith("weave://"):
        # option 1 - get the model object from weave
        RagModel = get_weave_model(model_url)
        
        # NOTE: we use the .call object so that the call is returned and feedback can be added
        # TODO: check answer from Adam if i still need to a the specific model object in the call
        #data, call = asyncio.run(RagModel.predict.call(query=prompt, n_documents=2))  
        data = asyncio.run(RagModel.predict(prompt))   
        result_content = data['result']['content']
        call = weave.get_current_call()  

    elif model_url.startswith("http"):
        # option 2 - the model is already served to a certain endpoint
        response = requests.post(
            model_url,
            json={
                "query": prompt,
                "n_documents": 2,
            },
        )
        data = response.json()
        result_content = data['result']['result']['content']

        # TODO: this works when this is a weave.op - how to get the last call otherwise?
        call = weave.get_current_call() 
    else:
        print("Please pass in a URL to a weave object or http endpoint.")
 ```

File: docs/docs/reference/gen_notebooks/feedback_prod.md [hunk 3]
```diff
@@ -0,0 +1,179 @@
...
+You can run this with `streamlit run chatbot.py`.
+
+Now, you can interact with this application and click the feedback buttons after each response. 
+Visit the Weave UI to see the attached feedback.
+
+## Explanation
+
+If we consider our decorated prediction function as:
+
+We can use it as usal to deliver some model response to the user:
```
NiWaRe, line 117: This line has to be deleted I believe (it belongs to the next code snippet, see below).

docs/docs/reference/gen_notebooks/feedback_prod.md
```diff
@@ -0,0 +1,179 @@
...
+    # Your prediction logic here
+    some_result = 'hello world'
+    return some_result
+```
+
+We can use it as usal to deliver some model response to the user:
+
+
+```python
+with weave.attributes({'session': '123abc'}): # attach arbitrary attributes to the call
```
NiWaRe, line 137: Maybe quick explanation that this allows you to define other columns that "inputs" and "outputs" and allows you to effectively separate different production types (testing, etc.). Had his question come up in some demos.

docs/docs/reference/gen_notebooks/feedback_prod.md
```diff
@@ -0,0 +1,179 @@
...
+    result = predict(input_data="your data here") # user question through the App UI
+```
+
+To attach feedback, you need the `call` object, which is obtained by using the `.call()` method *instead of calling the function as normal*:
+
+
+```python
+result, call = predict.call(input_data="your data here")
```
NiWaRe, line 145: Should we also mention `call = weave.get_current_call() ` as alternative?

docs/docs/reference/gen_notebooks/feedback_prod.md
```diff
@@ -0,0 +1,179 @@
...
+calls = thumbs_down.refs().calls()
+for call in calls:
+    print(call.inputs)
+    print(call.feedback)
+```
+
+First, we gather all feedback of a certain type using `client.feedback(reaction="👎")`, then we can get the associalted calls with `.refs().calls()`. Finally, we can get the inputs and feedback of each call. 
+
+:::tip 
+After some post-processing, this can be used to build a `weave.Dataset` for use in `weave.Evaluation`.
```
NiWaRe, line 172: Here we have to add the reference to the cookbook on evaluating based on production data that's coming soon.

docs/notebooks/feedback_prod.ipynb
```diff
@@ -0,0 +1,307 @@
+{
```
NiWaRe, line 1: I checked out the branch and ran locally without any issues - only when clicking on the GColab I'm getting:

`There was an error loading this notebook. Ensure that the file is accessible and try again.
https://github.com/wandb/weave/blob/master/docs/notebooks/feedback_prod.ipynb`

That should be good as soon as everything is merged though right?

NiWaRe (commented):
hi @scottire , did the review of the prod logging feedback cookbook 👍

docs/docs/reference/gen_notebooks/feedback_prod.md
scottire: I think it's fine to leave that out here

docs/notebooks/feedback_prod.ipynb
```diff
@@ -0,0 +1,333 @@
...
+    "    # Your prediction logic here\n",
+    "    some_result = \"hello world\"\n",
+    "    return some_result"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can use it as usal to deliver some model response to the user:"
```
jamie-rasmussen, line 211: usal => usual

docs/notebooks/feedback_prod.ipynb
```diff
@@ -0,0 +1,333 @@
...
+   "outputs": [],
+   "source": [
+    "!pip install weave openai streamlit"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First, create a file called `secrets.toml` and add an OpenAI key so it works with `[st.secrets](https://docs.streamlit.io/develop/api-reference/connections/st.secrets)`. You can [sign up](https://platform.openai.com/signup) on the OpenAI platform to get your own API key. "
```
jamie-rasmussen, line 56: Why is st.secrets link escaped?

docs/notebooks/feedback_prod.ipynb
```diff
@@ -0,0 +1,333 @@
...
+    "vscode": {
+     "languageId": "plaintext"
+    }
+   },
+   "source": [
+    "\n",
+    "It is often hard to automatically evaluate a generated LLM response so, depending on your risk tolerance, you can gather direct user feedback to find areas to improve.\n",
+    "\n",
+    "In this tutorial, we'll use a custom RAG chatbot as an example app with which the users can interact and which allows us to collect user feedback.\n",
+    "We'll use Streamlit to build the interface and we'll capture the LLM interactions and feedback in Weave."
```
jamie-rasmussen, line 29: I'm kind of confused about what's happening at a meta level here - we're using a notebook but we don't actually want the notebook to be executed, it is just for generating the markdown file?

docs/notebooks/feedback_prod.ipynb
jamie-rasmussen, line 56: Storing secret keys in a plaintext file seems like encouraging bad practices - does getting OPENAI_API_KEY from the environment not work? If we are telling people to put it in a file, we should probably at least direct them to `~/.streamlit/` which is less likely to get checked in and leak the key.

docs/notebooks/feedback_prod.ipynb
```diff
@@ -0,0 +1,333 @@
...
+   "metadata": {},
+   "source": [
+    "## Retrieving Feedback "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can pull feedback and their associated calls using weaves export APIs. "
```
jamie-rasmussen, line 270: weave => Weave's. Also not sure I would call these `export` APIs vs. something like "query APIs", or maybe just saying "Weave's Python client".

docs/notebooks/feedback_prod.ipynb
scottire: this is how they're recommending it [here](https://docs.streamlit.io/deploy/streamlit-community-cloud/deploy-your-app/secrets-management), along with telling users to gitignore the file

docs/notebooks/feedback_prod.ipynb
scottire: there's no great way to allow users to run streamlit in notebooks. Open to suggestions of how to make this possible in future.

docs/notebooks/feedback_prod.ipynb
scottire: fixed

docs/notebooks/feedback_prod.ipynb
scottire: fixed

docs/notebooks/feedback_prod.ipynb
jamie-rasmussen: Maybe for the future we should extend docs generation to allow omitting the "This is a notebook / Open in Colab" header.

Instructions like `!pip install weave openai streamlit` definitely make it seem like this is intended to execute in a notebook.

jamie-rasmussen (approved):

PR #2168, opened 2024-08-19
fix(weave): Fix weave ISO date parsing

ericakdiaz:
To unblock and fix the failing tests https://github.com/wandb/core/pull/23351, this makes weave compatible with ISO strings that have a "Z". The regular datetime library only supports that as of python 3.11, so using dateutil instead fixes it.

Did some manual testing that it seems to work as expected with a "Z" instead of throwing an error

image


adrnswanberg (approved):

requirements.txt
```diff
@@ -42,6 +42,9 @@ numpy<2.0.0
 # Segment logging
 analytics-python>=1.2.9
 
+# Used for ISO date parsing (analytics-python uses 2.8.2, so match it to avoid multiple versions)
```
jamie-rasmussen, line 45: I think we generally want to avoid strictly pinning versions if we can.

ericakdiaz: oh ok, i had thought since `analytics-python` strictly pins it https://github.com/segmentio/analytics-python/blob/master/requirements.txt#L7, we should too? but if it's not an issue I can change it to >= :)

jamie-rasmussen: If they've strictly pinned the dependency the resolver should give us that version too. I'd still list our dependency as `>=`. I don't really know what the deal is with `analytics-python` but their README says

> This library is in maintenance mode. It will send data as intended, but receive no new feature support and only critical maintenance updates from Segment.

ericakdiaz: got it! just updated it!

jamie-rasmussen (approved):

PR #2176, opened 2024-08-20
chore: type constraints, class maps, and runtime validation of props

onx2:
This PR does the following:
- Replaces variables for each variant with a mapping
- Moves static values outside of function
- Uses JSX element instead of interpolation 
- Adds compile & runtime constraints for props:
  - Icon OR label button
- Sensible defaults for certain a11y attributes
- Prevents external overrides of explicitly added internal html attributes

### DIFF
--- a/weave-js/src/components/Button/Button.tsx
+++ b/weave-js/src/components/Button/Button.tsx
@@ -109,65 +109,66 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
     /**
      * Runtime validation of props' type constraints
      */
-    const ButtonInner = () => {
-      if (icon) {
-        if (children || startIcon || endIcon) {
-          console.warn(
-            'Button: `startIcon`, `endIcon`, and `children` are ignored when using `icon`.'
-          );
+    const ButtonInternal = () => {
+      const validatedChildren = () => {
+        if (icon) {
+          if (children || startIcon || endIcon) {
+            console.warn(
+              'Button: `startIcon`, `endIcon`, and `children` are ignored when using `icon`.'
+            );
+          }
+          return <Icon name={icon} role="presentation" />;
         }
-        return <Icon name={icon} role="presentation" />;
-      }
 
-      if (!children) {
-        console.error('Button: `children` is required when not using `icon`.');
-        return null;
-      }
+        if (!children) {
+          console.error(
+            'Button: `children` is required when not using `icon`.'
+          );
+          return null;
+        }
+
+        return (
+          <>
+            {startIcon ? <Icon name={startIcon} role="presentation" /> : null}
+            {children}
+            {endIcon ? <Icon name={endIcon} role="presentation" /> : null}
+          </>
+        );
+      };
 
       return (
-        <>
-          {startIcon ? <Icon name={startIcon} role="presentation" /> : null}
-          {children}
-          {endIcon ? <Icon name={endIcon} role="presentation" /> : null}
-        </>
+        <button
+          ref={ref}
+          {...htmlAttributes}
+          type="button"
+          aria-disabled={htmlAttributes.disabled}
+          aria-label={htmlAttributes['aria-label'] || tooltip}
+          className={twMerge(
+            classNames(
+              // Display
+              'inline-flex items-center justify-center',
+              // Font
+              "font-['Source_Sans_Pro'] font-semibold",
+              // Disabled
+              'disabled:pointer-events-none disabled:opacity-35',
+              // Focused
+              'focus-visible:outline focus-visible:outline-[2px] focus-visible:outline-teal-500',
+              // Other
+              'night-aware whitespace-nowrap rounded border-none',
+              sizeClasses[size],
+              variantClasses[variant],
+              // Conditional classes
+              {
+                [iconOnlySizeClasses[size]]: Boolean(icon),
+                [activeVariantClasses[variant]]: active,
+              },
+              className
+            )
+          )}>
+          {validatedChildren()}
+        </button>
       );
     };
-
-    const ButtonInternal = () => (
-      <button
-        ref={ref}
-        {...htmlAttributes}
-        type="button"
-        aria-disabled={htmlAttributes.disabled}
-        aria-label={htmlAttributes['aria-label'] || tooltip}
-        className={twMerge(
-          classNames(
-            'night-aware',
-            'inline-flex',
-            'items-center',
-            'justify-center',
-            'whitespace-nowrap',
-            'rounded',
-            'border-none',
-            "font-['Source_Sans_Pro']",
-            'font-semibold',
-            'disabled:pointer-events-none',
-            'disabled:opacity-35',
-            'focus-visible:outline',
-            'focus-visible:outline-[2px]',
-            'focus-visible:outline-teal-500',
-            sizeClasses[size],
-            variantClasses[variant],
-            {
-              [iconOnlySizeClasses[size]]: Boolean(icon),
-              [activeVariantClasses[variant]]: active,
-            },
-            className
-          )
-        )}>
-        <ButtonInner />
-      </button>
-    );
 
     const [isTooltipOpen, setIsTooltipOpen] = useState(false);
 
--- a/weave-js/src/components/Button/Button.tsx
+++ b/weave-js/src/components/Button/Button.tsx
@@ -7,8 +7,8 @@
 import {type TooltipContentProps} from '@radix-ui/react-tooltip';
 import classNames from 'classnames';
 import React, {
-  CSSProperties,
   forwardRef,
+  type PropsWithChildren,
   type ReactElement,
   useState,
 } from 'react';
@@ -41,7 +41,6 @@ export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
   active?: boolean;
   tooltip?: string;
   tooltipProps?: TooltipContentProps;
-  twWrapperStyles?: React.CSSProperties;
 } & (IconButtonProps | LabelButtonProps);
 
 const sizeClasses = {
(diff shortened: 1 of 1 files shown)


weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
-      startIcon = icon,
+      startIcon,
       endIcon,
       active,
       children,
       className = '',
       tooltip,
       tooltipProps,
-      twWrapperStyles = {},
+      twWrapperStyles,
```
onx2, line 79: `undefined` can be using in spread operators so no need for this default assignment

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
+};
+
+export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
   (
     {
       size = 'medium',
       variant = 'primary',
       icon,
-      startIcon = icon,
+      startIcon,
```
onx2, line 72: This may be convenient for consumers but makes for inconsistency between call sites leading to code that is harder to refactor and reason with.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
-    const isLarge = size === 'large';
-    const isPrimary = variant === 'primary';
-    const isSecondary = variant === 'secondary';
-    const isGhost = variant === 'ghost';
-    const isQuiet = variant === 'quiet';
-    const isDestructive = variant === 'destructive';
-
-    const hasBothIcons = startIcon && endIcon;
-    const hasOnlyOneIcon = hasIcon && !hasBothIcons;
-    const isIconOnly = hasOnlyOneIcon && !children;
```
onx2, line 74: replaced with mappings

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
       display: 'inline-flex',
       width: className.includes('w-full') ? '100%' : undefined,
       ...twWrapperStyles,
     };
 
-    const button = (
+    /**
+     * Runtime validation of props' type constraints
+     */
+    const ButtonInner = () => {
```
onx2, line 98: Enforces the type constraints above with helpful console output for invalid states.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
+          {children}
+          {endIcon ? <Icon name={endIcon} role="presentation" /> : null}
+        </>
+      );
+    };
+
+    const ButtonInternal = () => (
       <button
         ref={ref}
+        {...htmlAttributes}
```
onx2, line 125: We want to override the html attrs with our explicitly defined ones so this should come first.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
         type="button"
+        aria-disabled={htmlAttributes.disabled}
+        aria-label={htmlAttributes['aria-label'] || tooltip}
```
onx2, line 128: reasonable defaults for a11y attributes, if possible

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
     if (tooltip) {
       return (
         <Tailwind style={wrapperStyles}>
           <Tooltip.Provider>
             <Tooltip.Root open={isTooltipOpen} onOpenChange={setIsTooltipOpen}>
               <Tooltip.Trigger asChild>
                 {/* span is needed so tooltip works on disabled buttons */}
-                <span>{button}</span>
+                <span>
+                  <ButtonInternal />
```
onx2, line 152: Use JSX element instead of interpolation.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
  * https://www.figma.com/file/01KWBdMZg5QM9SRS1pQq0z/Design-System----Robot-Styles?type=design&node-id=5956-31813&mode=design&t=Gm4WWGWwgjdfUUTe-0
  */
 
-import {TooltipContentProps} from '@radix-ui/react-tooltip';
+import {type TooltipContentProps} from '@radix-ui/react-tooltip';
```
onx2, line 7: Importing with `type` explicitly makes it easier to reason with b/c components and constructor functions both use PascalCase in their naming conventions. so without this, the imports are indistinguishable.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
 import {Tailwind} from '../Tailwind';
-import {ButtonSize, ButtonVariant} from './types';
+import type {ButtonSize, ButtonVariant} from './types';
 
-export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
-  /**
-   * Shorthand alias for `startIcon` since most button icons are startIcons.
-   * If both props are provided, `startIcon` value will be used out of the two.
-   */
+type IconButtonProps = {
```
onx2, line 22: These changes add the following constraints at compile time:
- `tooltip` must be present if using `icon`
- `icon` mustn't be used with `children`, `startIcon`, or `endIcon`
- `children` is expected if not using `icon`

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
+
+export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
   size?: ButtonSize;
   variant?: ButtonVariant;
   children?: ReactElement | string;
   active?: boolean;
   tooltip?: string;
   tooltipProps?: TooltipContentProps;
   twWrapperStyles?: React.CSSProperties;
+} & (IconButtonProps | LabelButtonProps);
```
onx2, line 46: Must be either an "icon button" or "label button" - this is how those type constraints are applied along with the base props.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
+
+const sizeMap = {
```
onx2, line 48: The mappings used for classes below. Static values can be placed outside of the function so they aren't re-created on every render.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,153 @@
...
-    const isIconOnly = hasOnlyOneIcon && !children;
-
     /**
      * The Tailwind wrapper is a div, which is a block-element. However, a normal button
      * wouldn't be wrapped and would be inline-block by default. These styles are a
      * necessary workaround to ensure Button is styled correctly despite the wrapper.
      */
-    const wrapperStyles = {
+    const wrapperStyles: CSSProperties = {
       display: 'inline-flex',
```
onx2, line 90: This _may_ be able to be simple display: contents; + ...twWrapperStyles

weave-js/src/components/Button/Button.tsx
onx2: ⚠️ This is a breaking change.

It would be nice if weave was a package so we could optionally upgrade in a PR that updates the app with the necessary call site changes. I'm open to suggestions on how to make breaking changes without actually breaking the app.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,155 @@
...
+    /**
+     * Runtime validation of props' type constraints
+     */
+    const ButtonInner = () => {
+      if (icon) {
+        if (children || startIcon || endIcon) {
+          console.warn(
+            'Button: `startIcon`, `endIcon`, and `children` are ignored when using `icon`.'
+          );
+        }
```
onx2, line 104: These things are nice but it would also be nice to not bundle console logs or output them in production environments.

weave-js/src/components/Button/Button.tsx
connieelee: what! mind blown

weave-js/src/components/Button/Button.tsx
connieelee: you wouldn't need to block deploys, this PR would just be blocked until all the call sites are updated

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,155 @@
...
-   */
+type IconButtonProps = {
   icon?: IconName;
-  /**
-   * Icon displayed before the children
-   */
+  children?: never;
+  startIcon?: never;
+  endIcon?: never;
+  tooltip: string;
```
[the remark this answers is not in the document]
connieelee: the only thing is, if the aria label and the tooltip _are_ in fact the same, then we should probably use [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) to label the button by the tooltip contents

weave-js/src/components/Button/Button.tsx
onx2: Sure, either work fine IMO but if `{buttonInner}` feels better than `<ButtonInner />` I can switch it over

weave-js/src/components/Button/Button.tsx
onx2: This is somewhat of a subjective thing. I'm personally in favor of reducing options and adding them as the need arises rather than starting wide open.

weave-js/src/components/Button/Button.tsx
onx2: It probably varies as the `startIcon` and `icon` were originally interchangeable, but either way we'd need to update the call sites.

weave-js/src/components/Button/Button.tsx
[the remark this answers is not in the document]
onx2: Interesting point! I guess it depends on _how_ the tooltip is rendered -- if it is detached from the button then the aria-labelledby make sense on the button, but we could also just assign the tooltip's value to aria-label on the button.

As for design not wanting a tooltip in certain areas, I'd be curious when this is applicable. If someone uses a screen reader b/c of poor vision or something, an aria label is needed to understand what an icon button does in the same way that it would read the children of the button. The flip side of someone without those impairments is that they need the tooltip to understand what the icon button does, and the children of the button could just be read. 🤔 I'd defer to WCAG specs & design team on this topic but we can definitely remove the necessity for it right now.

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,153 +4,155 @@
...
-                isQuiet && active,
-              'hover:text-moon-800 dark:hover:text-moon-200': isQuiet,
-              'hover:bg-oblivion/[0.05] dark:hover:bg-moonbeam/[0.05]': isQuiet,
-
-              // destructive
-              'bg-red-500 text-white hover:bg-red-450': isDestructive,
-              'bg-red-450': isDestructive && active,
-            },
+            sizeMap[size] + (icon && !children ? ' h-24 w-24 p-0' : ''),
+            variantMap[variant] + (active ? ' bg-teal-450' : ''),
```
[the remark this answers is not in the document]
onx2: Updated, I think this is right now?

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,163 +4,172 @@
...
+          </Tooltip.Trigger>
+          <Tooltip.Portal>
+            <Tooltip.Content {...tooltipProps}>{tooltip}</Tooltip.Content>
+          </Tooltip.Portal>
+        </Tooltip.Root>
+      </Tooltip.Provider>
+    );
+  }
+
+  return <>{children}</>;
```
connieelee, line 174: **nitpick:** should not need to wrap in fragment?

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,163 +4,172 @@
...
+              </>
+            )}
+          </button>
+        </OptionalTooltip>
+      </Tailwind>
+    );
   }
 );
+
+function OptionalTooltip({
```
connieelee, line 151: **praise:** nice

weave-js/src/components/Button/Button.tsx
```diff
@@ -4,163 +4,172 @@
...
-
-    return <Tailwind style={wrapperStyles}>{button}</Tailwind>;
+    return (
+      <Tailwind
+        /**
+         * The Tailwind wrapper is a div, which is a block-element. However, a normal button
+         * wouldn't be wrapped and would be inline-block by default. These styles are a
+         * necessary workaround to ensure Button is styled correctly despite the wrapper.
+         */
+        style={{display: 'contents'}}>
```
connieelee, line 126: hmm it would be nice if this just worked, but i did some quick testing in our storybook and it doesn't behave the same as before

before:
Screenshot 2024-08-20 at 2 30 31 PM

after:
Screenshot 2024-08-20 at 2 30 11 PM

connieelee (commented):
over all i think the changes you're making are good ideas. it'll just be a matter of sequencing your PRs correctly and doing thorough testing. Button is used so heavily, we'll need to QA with extra diligence. might be good to even give other teams some heads up and have them help spot check as well.

last time i made a Button change (updated the tooltip from semantic to radix), Jamie helped me find a few in-app bugs that i wouldn't have found by testing in just storybook or parts of the app i normally work on

weave-js/src/components/Button/Button.tsx
onx2: oh yikes, thanks for checking 😅 

IMO the change here to make it align with the previous version should be implemented as a class rather than overriding things in the tailwind wrapper. I also think exposing the tailwind props is fraught with peril. I'll take a look at what I can do to fix this though 👍🏼

PR #2177, opened 2024-08-20
Add PII Cookbook

scottire:
Adding the PII cookbook using Presidio in addition to a couple of PII datasets


jamie-rasmussen:
Closing in favor of https://github.com/wandb/weave/pull/2305

PR #2180, opened 2024-08-20
chore: add $in to orm query builder 

gtarpenning:
This PR:
- adds `$in` function to the Mongo style query system, which accepts a get field on the left, and a list of literals on the right. 
- tests of usage

Usage:
```python
query = {
    "$expr": {
        "$in": [
            {"$getField": "id"}, 
            [{"$literal": "id123"}, {"$literal": "id1111"}]
        ]
    }
}

res = weave.client.server.calls_query_stream(
    tsi.CallsQueryReq.model_validate(
        dict(
            project_id=project_id,
            query=query,
        )
    )
)
```

https://wandb.atlassian.net/browse/WB-20048


weave/tests/trace/test_client_trace.py
```diff
@@ -2297,6 +2297,58 @@ def test_obj(val):
...
+
+    test_op(1, [1, 2, 3])
+    test_op(2, [1, 2, 3])
+    test_op(3, [5, 6, 7])
+    test_op(4, [8, 2, 3])
+
+    call_ids = [call.id for call in test_op.calls()]
+    assert len(call_ids) == 4
+
+    strs = [str(call_id) for call_id in call_ids]
```
jamie-rasmussen, line 2313: It looks like call.id can be None (though I'm not sure when that can occur) but if not None it is already a `str`, so not sure what this is here for.

jamie-rasmussen (approved):

weave/tests/trace/test_client_trace.py
gtarpenning: I was getting a weird boxed_str pydantic warning, but now that appears to be resolved so i will revert, thx.

PR #2185, opened 2024-08-21
feat(weave): Add customizable display names

andrewtruong:
## Summary
1. Supports customizable display names, with function support (fired at call time).  When using a function, it always takes the call object, so you can use `call.attributes` or do conditionals based on `call.inputs`, etc.

```py
    def custom_attribute_name(call):
        model = call.attributes["model"]
        revision = call.attributes["revision"]
        now = call.attributes["date"]

        return f"{model}__{revision}__{now}"

    @op(display_name=custom_attribute_name)
    def func():
        return 1

    with weave.attributes(
        {
            "model": "finetuned-llama-3.1-8b",
            "revision": "v0.1.2",
            "date": "2024-08-01",
        }
    ):
        func()

    with weave.attributes(
        {
            "model": "finetuned-gpt-4o",
            "revision": "v0.1.3",
            "date": "2024-08-02",
        }
    ):
        func()

```

image

## Resolves
- https://wandb.atlassian.net/browse/WB-20320


docs/docs/guides/tracking/tracing.md
```diff
@@ -97,3 +97,55 @@ sentence = "There are many fruits that were found on the recently discovered pla
...
+
+3. The `display_name` can also be a function that takes in a `Call` object and returns a string. The `Call` object will be passed automatically when the function is called, so you can use it to dynamically generate names based on call inputs, attributes, etc.
+
+   ```py
+   def custom_attribute_name(call):
+       model = call.attributes["model"]
+       revision = call.attributes["revision"]
+       now = call.attributes["date"]
+
+       return f"{model}__{revision}__{now}"
```
adrnswanberg, line 127: One common pattern I would think is just taking the fn name and adding a ts string or similar. Can we have the example show that? I wouldn't know from looking at this doc how to pull the "op name" out of the "call" object.

andrewtruong: I don't think this is easy to do atm, but you're right we should make it easy and show users how to do it.

Basically right now the op_name will be the fully qualified weave ref, which is probably not what the user wants.

adrnswanberg: Really? How does the op name eventually get attached to the call then?

andrewtruong: `op_name` is the ref atm:
image

I'm assuming then in the UI it's parsed

adrnswanberg: @tssweeney is it safe to parse out a friendly op name here?

weave/trace/op.py
```diff
@@ -107,6 +110,7 @@ class Op(Protocol):
     """
 
     name: str
+    display_name: str
```
tssweeney, line 113: I think we are missing something a bit critical here: serialization and deserialization.  (maybe this is something we want to acknowledge is a gap in the spirit of MVP?). But when you "get" an op, the custom display name function is not coming long for the ride.

docs/docs/guides/tracking/tracing.md
tssweeney: `op_name` is defined as either:
1. (most common case): A ref pointing to the source op
2. a simple string

The second case is what we loosely refer to as an "anonymous call" where we don't have a pointer to a producing Op.

Therefore "op_name" is probably not the best field name...

It is not safe to abandon that, but it is safe to parse it out. This is in fact what we do in the UI.

weave/trace/op.py
andrewtruong: Yep, ack this is a gap.  We'll have to think about how to handle this.

docs/docs/guides/tracking/tracing.md
andrewtruong: Added a property `func_name` on call that will do the parsing for users.

weave/trace/util.py
```diff
@@ -119,6 +119,20 @@ def is_notebook() -> bool:
     return True
 
 
+def _get_func_name_from_maybe_ref(s: str) -> str:
+    """This is a hack and should not be needed. We should be able to auto-resolve this for the user.
+    Keeping this here for now, but it should be removed in the future once we have a better solution.
+    """
+    # it's not a ref, just return same thing
```
tssweeney, line 126: i would recommend using the parser in `refs.py` rather than doing this prefix matching.

weave/trace/weave_client.py
```diff
@@ -143,13 +144,17 @@ class Call:
...
-    display_name: Optional[str] = None
+    display_name: Optional[Union[str, Callable[["Call"], str]]] = None
     attributes: Optional[dict] = None
     # These are the live children during logging
     _children: list["Call"] = dataclasses.field(default_factory=list)
 
     _feedback: Optional[RefFeedbackQuery] = None
 
+    @property
+    def func_name(self) -> str:
```
tssweeney, line 155: can you please add doc string about this

weave/trace/util.py
andrewtruong: much better, thanks!

weave/trace/op.py
tssweeney: can we add code note to this effect?

weave/trace/weave_client.py
```diff
@@ -143,13 +143,26 @@ class Call:
...
+        """
+        The decorated function's name that produced this call.
+
+        This is different from `op_name` which is usually the ref of the op.
+        """
+        if self.op_name.startswith("weave:///"):
+            ref = parse_uri(self.op_name)
```
tssweeney, line 161: Note: `parse_uri` does throw under certain conditions where the ref string is not parsable. (Shouldn't occur often, but technically is possible)

tssweeney (approved):

weave/trace/weave_client.py
andrewtruong: I can live with that here.  If we detect errors here I'll add more handling

shawnlewis (changes requested):
OK thanks for adding the docs, nice write-up.

Now after reading it I'm confused about the design though.

We need to separate two concepts that I think are mixed together:
- op naming
- calling naming

Right now the display_name parameter controls one or the other depending on how you call it? Let's rethink.

andrewtruong:
`display_name` name is just for the call, it does not affect the op name.  Maybe it can be clarified as `call_display_name`?

So we would have:
1. `call_display_name` -- a str or func that generates the display name of a call (at call-time)
2. `func_name` -- the name of the decorated function/op name
3. `op_name` -- most confusing here; this is usually the ref to the op itself.

shawnlewis (approved):

PR #2194, opened 2024-08-21
feat: guess renderer for string values

jamie-rasmussen:
Internal Jira: https://wandb.atlassian.net/browse/WB-20485

If we have high confidence that a string value is markdown via heuristics we can default the string viewer to the corresponding mode, saving the user a few clicks and making things look nicer by default.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ValueViewString.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ValueViewString.tsx
@@ -72,12 +72,35 @@ const PreserveWrapping = styled.div`
 `;
 PreserveWrapping.displayName = 'S.PreserveWrapping';
 
+// Regular expressions for common Markdown syntax
+// Note this is intentionally limited in scope to reduce false positives.
+const LIKELY_MARKDOWN_PATTERNS: RegExp[] = [
+  /```[\s\S]*```/, // Code block
+  /\[.+\]\(.+\)/, // Links [text](url)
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ValueViewString.tsx
```diff
@@ -72,12 +72,35 @@ const PreserveWrapping = styled.div`
...
+  /```[\s\S]*```/, // Code block
+  /\[.+\]\(.+\)/, // Links [text](url)
+  /!\[.*\]\(.+\)/, // Images ![alt](url)
+];
+
+const isLikelyMarkdown = (value: string): boolean => {
+  return LIKELY_MARKDOWN_PATTERNS.some(pattern => pattern.test(value));
+};
+
+const getDefaultFormat = (value: string): string => {
```
gtarpenning, line 87:
```suggestion
const getDefaultFormat = (value: string): 'Markdown' | 'Text' => {
```

gtarpenning: I think that plays nicer with the type suggestions, until we properly type it

gtarpenning (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ValueViewString.tsx
jamie-rasmussen: I tightened up the typing.

gtarpenning (approved):

PR #2191, opened 2024-08-21
chore: move header buttons in preparation for saved views design

jamie-rasmussen:
Internal Figma: https://www.figma.com/design/mufhcDVpsRg9HJrLM61PCN/Weave-(Latest)?node-id=1089-18178&t=U28KGlyGddcrndUf-0
Internal Jira: https://wandb.atlassian.net/browse/WB-20484

We're planning to use the header space for saved view controls. Moving export and delete calls grid controls to "filter bar" area.

Need to check with design on where the Compare button should be.

Before:
Screenshot 2024-08-21 at 2 31 51 PM

After:
Screenshot 2024-08-21 at 2 31 39 PM

PR #2199, opened 2024-08-22
chore: move evaluation compare button

jamie-rasmussen:
Follow up to https://github.com/wandb/weave/pull/2191
After consulting with @adamwdraper this is where the Compare button should go in preparation for a top Saved Views header.

Before:
Screenshot 2024-08-22 at 1 26 34 PM

After:
Screenshot 2024-08-22 at 1 26 04 PM

Note that after this change I think the WeaveHeaderExtrasContext is unused and we should consider removing it.


gtarpenning (approved):
Not sure I love this,

PR #2213, opened 2024-08-26
chore: add derived summary fields

gtarpenning:
https://wandb.atlassian.net/browse/WB-20561

Add derived fields to the summary under the `weave` key, immediately after db read. 

When a call is running, the derived fields will look like: 

```python
"summary": {
   "weave": {
      "status": "running",
      "trace_name": "name",
   }
}
```

When completed:
```python
"summary": {
   "weave": {
      "status": "success",
      "trace_name": "name",
      "latency_ms": 1222,
      "usage": {....}
   }
}
```


weave/trace/weave_client.py
```diff
@@ -308,7 +308,7 @@ def sum_dict_leaves(dicts: list[dict]) -> dict:
         for k, v in d.items():
             if isinstance(v, dict):
                 result[k] = sum_dict_leaves([result.get(k, {}), v])
-            else:
+            elif v is not None:
```
gtarpenning, line 311: None is now allowed at a value in the summary

weave/trace_server/trace_server_interface.py
```diff
@@ -45,7 +45,7 @@ class LLMCostSchema(LLMUsageSchema):
 
 class WeaveSummarySchema(ExtraKeysTypedDict, total=False):
     status: Optional[Literal["success", "error", "running"]]
-    nice_trace_name: Optional[str]
+    trace_name: Optional[str]
     latency: Optional[int]
```
tssweeney, line 49: can you comment what fidelity latency is? It is an int, so i am hoping it is at least ms

weave/tests/trace/test_client_trace.py
```diff
@@ -105,7 +116,7 @@ def test_trace_server_call_start_and_end(client):
     start = tsi.StartedCallSchemaForInsert(
         project_id=client._project_id(),
         id=call_id,
-        op_name="test_name",
+        op_name="weave:///test_entity/test_project/op/test_name:test",
```
tssweeney, line 119: question: it should very much be allowed for op_name to NOT be a ref - does that still work?

weave/trace_server/trace_server_common.py
```diff
@@ -1,6 +1,90 @@
...
+            it's "summary".
+
+    Returns:
+        The derived summary fields.
+    """
+    display_name = call_dict.get("display_name")
+    started_at_dt = _make_datetime_from_any(call_dict["started_at"])
+    ended_at_dt = _make_datetime_from_any(call_dict.get("ended_at"))
+
+    status = "success"
```
tssweeney, line 33: recommend that you use constants over magic strings

weave/trace_server/trace_server_common.py
```diff
@@ -1,6 +1,90 @@
...
+    summary = _load_json_maybe(call_dict.get(summary_key)) or {}
+
+    weave_summary = summary.get("weave", {})
+    weave_summary["trace_name"] = display_name
+    weave_summary["status"] = status
+    if latency is not None:
+        weave_summary["latency"] = latency
+
+    summary["weave"] = weave_summary
+    return cast(tsi.SummaryMap, summary)
```
tssweeney, line 59: instead of casting, can you directly construct this?

weave/trace_server/trace_server_common.py
```diff
@@ -1,6 +1,90 @@
...
+
+
+def _make_datetime_from_any(
+    dt: Optional[Union[str, datetime.datetime]],
+) -> Optional[datetime.datetime]:
+    """
```
tssweeney, line 65: why are we dealing with two different formats here? under what conditions are they different?

weave/trace_server/trace_server_common.py
```diff
@@ -1,6 +1,90 @@
...
+    if dt is None:
+        return None
+    if isinstance(dt, str):
+        return datetime.datetime.fromisoformat(dt)
+    elif isinstance(dt, datetime.datetime):
+        return dt
+    return None
+
+
+def _load_json_maybe(value: Any) -> Any:
```
tssweeney, line 78: same question here - why are we sometimes strings and sometimes dicts?

weave/trace_server/trace_server_interface.py
tssweeney: in the past, i have favored using a suffix like `latency_[UNIT]` to make this very clear

weave/trace_server/trace_server_common.py
gtarpenning: yo we have two different databases, in clickhouse these bad boys are json, in sqlite they need loading

gtarpenning: we could massage the data before

tssweeney: oh i see. i think i would pass a flag in top of the main function saying something like `data_is_stringified` or something like that. That the callers pass in. Then perform the proper paths from there. The reason i say this is i want to avoid cases where we accidentally have the wrong stored/in-mem format somewhere upstream, and this function happily consumes it.

A bit of a nit. you could also just comment this in the those functions.

weave/tests/trace/test_client_trace.py
gtarpenning: yep, removing this change to make that explicit.

weave/trace_server/trace_server_common.py
gtarpenning: I can't seem to construct this while keeping mypy happy, because while we control the `weave` key in the summary, there are also arbitrary user controlled keys. `tsi.SummaryMap` allows this, but mypy doesn't know that...

weave/trace_server/trace_server_common.py
gtarpenning: Ripped this out

weave/trace_server/trace_server_common.py
gtarpenning: removed

weave/trace_server/trace_server_common.py
```diff
@@ -1,27 +1,24 @@
...
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, cast
 
 from weave.trace_server import refs_internal as ri
-from weave.trace_server.trace_server_interface import (
-    FeedbackQueryReq,
-    FeedbackQueryRes,
-    Query,
-)
+from weave.trace_server import trace_server_interface as tsi
```
jwlee64, line 7: nit: maybe just add SummaryMap and TraceStatus

tssweeney (approved):

PR #2243, opened 2024-08-27
feat(weave): Add python api for costs

jwlee64:
client.add_costs
client.purge_costs
client.query_costs

client.server.cost_create
client.server.cost_purge
client.server.cost_query


weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -314,3 +314,22 @@ def feedback_query(self, req: tsi.FeedbackQueryReq) -> tsi.FeedbackQueryRes:
...
+    def cost_purge(self, req: tsi.CostPurgeReq) -> tsi.CostPurgeRes:
+        req.project_id = self._idc.ext_to_int_project_id(req.project_id)
+        return self._ref_apply(self._internal_trace_server.cost_purge, req)
+
+    def cost_query(self, req: tsi.CostQueryReq) -> tsi.CostQueryRes:
+        original_project_id = req.project_id
+        req.project_id = self._idc.ext_to_int_project_id(original_project_id)
+        res = self._ref_apply(self._internal_trace_server.cost_query, req)
+        for cost in res.results:
+            if "pricing_level_id" in cost:
```
gtarpenning, line 331: this will have to be changed if we allow different scope costs? When is this not the project_id? When do we expect this error to happen? what has to break upstream?

weave/trace_server/token_costs.py
```diff
@@ -438,3 +445,22 @@ def is_project_id_sql_injection_safe(project_id: str) -> None:
...
+def validate_cost_purge_req(req: tsi.CostPurgeReq) -> None:
+    """For safety, we currently only allow purging by cost id."""
+    expr = req.query.expr_.model_dump()
+    keys = list(expr.keys())
+    if len(keys) != 1:
+        raise InvalidRequest(MESSAGE_INVALID_PURGE)
+    if keys[0] == "eq_":
+        validate_purge_req_one(expr, MESSAGE_INVALID_PURGE)
+    elif keys[0] == "in_":
+        validate_purge_req_in(expr, MESSAGE_INVALID_PURGE)
```
gtarpenning, line 462:
```suggestion
    if keys[0] in ["eq_",  "in_"]:
        validate_purge_req_one(expr, MESSAGE_INVALID_PURGE)
```

weave/trace_server/validation.py
```diff
@@ -92,3 +93,58 @@ def object_id_validator(s: str) -> str:
 
 def refs_list_validator(s: typing.List[str]) -> typing.List[str]:
     return [validation_util.require_internal_ref_uri(ref) for ref in s]
+
+
+MESSAGE_INVALID_PURGE = "Can only purge by specifying one or more ids"
```
gtarpenning, line 98: I see why we are defining this again, but sort of weird to have two identical constants. I think we should keep this message here, don't allow passing it in, and change the `raise InvalidRequest(MESSAGE_INVALID_PURGE)` message in `token_costs.py` to something more general, like: `"Invalid format for purge, please specify one or more IDs using Mongo...."` etc.

weave/tests/trace/test_client_cost.py
```diff
@@ -0,0 +1,209 @@
...
+            project_id=project_id,
+            costs=costs,
+            wb_user_id="VXNlcjo0NTI1NDQ=",
+        )
+    )
+
+    cost_ids = res.ids
+    assert len(cost_ids) == 1
+
+    with pytest.raises(InvalidRequest):
```
gtarpenning, line 178: We could try to hit every `InvalidRequest` raise here, codecov

gtarpenning (commented):
questions!

weave/trace_server/external_to_internal_trace_server_adapter.py
jwlee64: This check will have to be extended when org costs are implemented, Ill add a comment.
Im not sure when this check would be triggered. We do have the check for both the feedback and calls query

weave/trace/weave_client.py
```diff
@@ -752,6 +757,103 @@ def feedback(
             query=query, reaction=reaction, offset=offset, limit=limit
         )
 
+    def add_costs(self, costs: Dict[str, CostCreateInput]) -> CostCreateRes:
+        """Add a cost to the current project.
+
+        Args:
+            cost: The cost to add.
```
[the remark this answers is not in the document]
jwlee64: I can extend the doc string.
when a new cost with the same llm_id is added, we add a new cost field with a different uuid and effective date

weave/tests/trace/test_client_cost.py
jwlee64: let me add a test for the validation utils

weave/trace_server/validation.py
jwlee64: The constants are slightly different.
In feedback we add "...feedback ids...", and in costs we add "...costs ids...", to assist in where the call site is, Ill update the constant names and make the messages clearer

weave/trace_server/trace_server_interface.py
```diff
@@ -590,6 +590,55 @@ class EnsureProjectExistsRes(BaseModel):
...
+    # TODO: From FeedbackQueryReq,
+    # TODO: I think I would prefer to call this order_by to match SQL, but this is what calls API uses
+    # TODO: Might be nice to have shortcut for single field and implied ASC direction
+    sort_by: Optional[List[SortBy]] = None
+    limit: Optional[int] = Field(default=None, examples=[10])
+    offset: Optional[int] = Field(default=None, examples=[0])
+
+
+class CostQueryRes(BaseModel):
+    # Note: this is not a list of costs because user can request any fields.
```
tssweeney, line 629: it would be nicer to have optional fields but make them explicit here

weave/trace_server/trace_server_interface.py
```diff
@@ -590,6 +590,55 @@ class EnsureProjectExistsRes(BaseModel):
     project_name: str
 
 
+class CostCreateInput(BaseModel):
+    prompt_token_cost: Optional[float] = 0.015
```
tssweeney, line 594: why have these defaults? Seems like we would want to require the user to provide the cost?

weave/trace_server/trace_server_interface.py
```diff
@@ -590,6 +590,55 @@ class EnsureProjectExistsRes(BaseModel):
...
+    completion_token_cost: Optional[float] = 0.015
+    prompt_token_cost_unit: Optional[str] = "USD"
+    completion_token_cost_unit: Optional[str] = "USD"
+    effective_date: Optional[datetime.datetime] = None
```
tssweeney, line 598: what does None mean for effective data and provider?

tssweeney (approved):

PR #2247, opened 2024-08-27
fix: object with dupe ref not loading in ui

gtarpenning:
https://wandb.atlassian.net/browse/WB-18553

This pr slightly changes the way we render refs in the object viewer. During replacement of refs with values, instead of relying on deleting refs once we see them, keep track of if we replaced. Currently we don't handle the case when we might want to replace multiple identical refs, as we have already deleted the ref from our lookup. 

Prod:
Screenshot 2024-08-27 at 12 41 58 PM

This branch:
Screenshot 2024-08-27 at 12 42 42 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
@@ -142,6 +142,8 @@ export const ObjectViewer = ({
       if (
         isRef(context.value) &&
         refValues[context.value] != null &&
+        // If this is a ref and the parent has been visited, we already resolved
+        // this ref. Example: `a._ref` where `a` is already in resolvedRefPaths
         !resolvedRefPaths.has(context.parent?.toString() ?? '')
       ) {
         dirty = true;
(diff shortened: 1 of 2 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
```diff
@@ -137,11 +137,19 @@ export const ObjectViewer = ({
     }
     let resolved = data;
     let dirty = true;
+    const resolvedRefPaths = new Set<string>();
     const mapper = (context: TraverseContext) => {
-      if (isRef(context.value) && refValues[context.value] != null) {
+      const parentPathStr = context.path
+        .ancestor(context.path.length() - 1)
```
jamie-rasmussen, line 143: Maybe add a `parent` method to `ObjectPath` that does this

jamie-rasmussen (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
gtarpenning: good idea, done

PR #2251, opened 2024-08-27
fix: use a more conservative `isWeaveRef` check before parsing strings

tssweeney:
Due to history of bad refs, there are a number of places in the UI which condition on `isRef` . `isRef` is just a prefix matcher, not a true validation of if the string is a fully compatible ref. As a result, we get a number of full-screen crashes. This replaces those calls with `isWeaveRef` which is a much stricter enforcement of the valid string.


weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
```diff
@@ -223,8 +226,25 @@ export const FIELD_DESCRIPTIONS: Record<string, string> = {
...
+export type WeaveRefString = string & {[WeaveRefStringSymbol]: never};
+
+/**
+ * `isWeaveRef` is a very conservative check that will ensure the passed
+ * in value is a valid ref string - capabable of being safely parsed into
+ * a Weave ref object. It ensures that the value is a string with the correct
+ * prefix, is parsible, and matches the latest "weave trace" style refs. It
+ * should be used as the appropriate type guard before parsing a ref.
+ */
+export const isWeaveRef = (value: any): value is WeaveRefString => {
```
gtarpenning, line 242: this looks great, do we still need `isRef`? Do we ever have to parse non weave refs?

gtarpenning (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
tssweeney: fair - un exported `isRef` and scoped it to a new name reflecting the prefix matching it is doing now

PR #2256, opened 2024-08-28
fix(ui): Fixes bad transitive import of visitParent

nhusher:
Updates unist-util-visit to the latest, which appears to resolve an issue with its transitive dependency unist-util-visit-parents. At some point they were incompletely migrated to es modules and some of the imports broke. This has the effect of breaking the custom charts panel in workspaces.

Updating to the latest version appears to have resolved the issue.


onx2 (approved):
seems legit to me 😆

jamie-rasmussen (approved):

nhusher:
recheck

PR #2258, opened 2024-08-28
chore: cosmetic refactor, improve import readability 

gtarpenning:
this pr: 
- `refs_internal` -> `ri`
- adds direct imports from `typing`, these classes are pretty unambiguous, no need to include `typing` in use

purely cosmetic refactor, all the extra import junk was driving me crazy


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -28,9 +28,20 @@
...
 import logging
 import threading
-import typing
 from collections import Counter, defaultdict
 from contextlib import contextmanager
+from typing import (
+    Any,
+    Dict,
+    Iterator,
+    List,
```
jamie-rasmussen, line 37: Since we require Python 3.9+ I believe we can just use e.g. `list` instead of `typing.List`

jamie-rasmussen (approved):

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: Fixing

PR #2260, opened 2024-08-28
chore(weave): Cleanup integrations tests and inline pydantic and dataclass objects

andrewtruong:
## Summary
1. Uses higher-level `client.calls()` instead of lower level `client.server.calls_query()`
 api for tests
    1. The `langchain` tests still use the lower level because the ordering is different there (seems to be BF?)
2. Moves scattered utilities into `integration_utilities.py`


weave/integrations/groq/groq_test.py
```diff
@@ -462,28 +415,23 @@ def run_conversation(user_prompt: str) -> Union[str, None]:
...
-    flatened_calls_list = [
-        (op_name_from_ref(c.op_name), d)
-        for (c, d) in flatten_calls(weave_server_respose.calls)
-    ]
-    assert flatened_calls_list == [
-        ("run_conversation", 0),
-        ("groq.chat.completions.create", 1),
-        ("get_game_score", 1),
-        ("groq.chat.completions.create", 1),
+    calls_list = [_get_op_name(c.op_name) for c in calls]
```
tssweeney, line 421: why did you remove the flatten calls? We are not only asserting order, but also depth. Those assertions are important, but lost here

weave/integrations/llamaindex/llamaindex_test.py
```diff
@@ -43,18 +20,17 @@ def assert_calls_correct_for_quickstart(calls: list[tsi.CallSchema]) -> None:
...
-        ("llama_index.query", 0),
-        ("llama_index.retrieve", 1),
-        ("llama_index.embedding", 2),
-        ("llama_index.synthesize", 1),
-        ("llama_index.chunking", 2),
-        ("llama_index.chunking", 2),
-        ("llama_index.templating", 2),
-        ("llama_index.llm", 2),
-        ("openai.chat.completions.create", 3),
+        "llama_index.query",
```
tssweeney, line 25: same here (and all other places this got dropped

weave/integrations/groq/groq_test.py
andrewtruong: Oh I see, I misunderstood what the index represented.

`client.calls()` is already flat and doesn't have that nesting atm.  Should there be an option for that?

tssweeney (approved):

PR #2269, opened 2024-08-29
chore(weave): Update client func names to be less confusing

andrewtruong:
## Summary
1. For clarity, renames:
    1. `call -> get_call`
    2. `calls -> get_calls`
    3. `feedback -> get_feedback`
2. This lines up with other methods like:
    1. `finish_call`
    2. `fail_call`.
3. Deprecates the functions `call`, `calls`, and `feedback`, adding warnings to update to the new name

image

## Resolves
- https://wandb.atlassian.net/browse/WB-20562


jamie-rasmussen (approved):
Code changes LGTM but recommend getting @shawnlewis review for general API naming pattern we want.

shawnlewis (approved):

PR #2271, opened 2024-08-29
chore(weave): Inline pydantic and dataclass objects

andrewtruong:
## Summary
1. Inlines pydantic and dataclass objects such that they don't have their own refs, unless explicitly published.
    ```py
    class Something(BaseModel):
        a: str
        b: int
    
    class WeaveObject(weave.Object):
        a: str
        s: Something    
    
    s = Something('a', 2)
    wo = WeaveObject('b', s)
    
    weave.publish(s)   # `s` object gets a ref
    weave.publish(wo)  # `s` object does not get a ref
    ```
  
    This should speed up the write path for deeply nested objects

## Resolves
- https://wandb.atlassian.net/browse/WB-20357

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -827,19 +827,23 @@ class WeaveClient:
             if ref.project == self.project:
                 return
             remove_ref(obj)
+        # Must defer import here to avoid circular import
+        from weave.flow.obj import Object
 
-        if isinstance(obj, (pydantic.BaseModel, pydantic.v1.BaseModel)):
+        if isinstance(obj, Object):
             obj_rec = pydantic_object_record(obj)
             for v in obj_rec.__dict__.values():
                 self._save_nested_objects(v)
             ref = self._save_object_basic(obj_rec, name or get_obj_name(obj_rec))
             obj.__dict__["ref"] = ref
+        elif isinstance(obj, (pydantic.BaseModel, pydantic.v1.BaseModel)):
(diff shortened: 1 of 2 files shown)


weave/trace/weave_client.py
```diff
@@ -799,19 +799,23 @@ def _save_nested_objects(self, obj: Any, name: Optional[str] = None) -> Any:
...
 
-        if isinstance(obj, (pydantic.BaseModel, pydantic.v1.BaseModel)):
+        if isinstance(obj, Object):
             obj_rec = pydantic_object_record(obj)
             for v in obj_rec.__dict__.values():
                 self._save_nested_objects(v)
             ref = self._save_object_basic(obj_rec, name or get_obj_name(obj_rec))
             obj.__dict__["ref"] = ref
+        elif isinstance(obj, (pydantic.BaseModel, pydantic.v1.BaseModel)):
+            obj_rec = pydantic_object_record(obj)
```
tssweeney, line 812: Let's sync on this - idk if `pydantic_object_record` is really the right dump here. We get those extra types/base classes/etc which seem to only make sense if the object is deserializable.

shawnlewis: That is the correct thing to do. It should be the same thing as what we actually save objects as.

tssweeney: hmmm interesting. Ok

shawnlewis (approved):

PR #2274, opened 2024-08-29
fix: Relax Ref Constraints and Add Helper Text to Error

tssweeney:
We are in the midst of introducing new constraints, deploying new versions, etc. This PR improves the constrain enforcement (and bumps the version so we can release in the morning - again).

1. When the Python client attempts to write an object with an un-sanitized name, we reject. However, the user message is something like the following error. This is helpful if they can change the name, but note useful if they cannot (due to pydantic saving). So the first part of this fix is to return a CTA to upgrade to `>0.50.0` (which will do the sanitization client-side)

```
HTTPError: 500 Server Error: Internal Server Error for url: /obj/create. {
  "reason": "1 validation error for ObjCHInsertable\nobject_id\n  Value error, Invalid object name:Name With Space. Contains invalid characters: [' ', ' '] [type=value_error, input_value='Name With Space', input_type=str]\n    For further information visit https://errors.pydantic.dev/2.7/v/value_error"
}
```

2. Since there are a number of "bad refs" in the system that correspond to unconstrained names, many users will be reading those existing refs. The way I implemented the constraint enforcement used post-init validation on the ref objects themselves. This is probably too aggressive. I modified this such that the write-time constraints are still enforced (stopping newly minted bad refs from entering the system), but lossened the read-time constraints so that these current users are not rejected when they read such refs.

3. I bumped the version so that we can deploy without waiting for another round of CI.

I feel like no matter what this is going to make people mad - since they can't have spaces or symbols in their object names after these changes. But we have to bite the bullet and draw the line sometime


weave/version.py
```diff
@@ -1 +1 @@
-VERSION = "0.51.0"
+VERSION = "0.51.1"
```
tssweeney, line 1: Version Bump for release

weave/trace_server/validation.py
```diff
@@ -62,8 +62,9 @@ def _validate_object_name_charset(name: str) -> None:
     # Object names must be alphanumeric with dashes
     invalid_chars = re.findall(r"[^\w._-]", name)
     if invalid_chars:
+        invalid_char_set = list(set(invalid_chars))
         raise ValueError(
-            f"Invalid object name: {name}. Contains invalid characters: {invalid_chars}"
+            f"Invalid object name: {name}. Contains invalid characters: {invalid_char_set}. Please upgrade your `weave` package to `>0.51.0` to prevent this error."
```
tssweeney, line 67: Better error message for user

weave/trace/refs.py
```diff
@@ -58,7 +58,8 @@ def __post_init__(self) -> None:
         refs_internal.validate_no_slashes(self.digest, "digest")
         refs_internal.validate_no_colons(self.digest, "digest")
         refs_internal.validate_extra(list(self.extra))
-        validation.object_id_validator(self.name)
+        refs_internal.validate_no_slashes(self.name, "name")
```
tssweeney, line 61: Change validation from the database validation to the correct parse validation

PR #2286, opened 2024-08-29
chore: limit expand columns to explicitly opened columns

gtarpenning:
This pr:
- Now only sets expand_columns to one the user has clicked on to expand
- Still allow export even if viewer query breaks


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTableButtons.tsx
```diff
@@ -207,14 +205,12 @@ export const ExportSelector = ({
                 />
               )}
             </DraggableHandle>
-            {!viewerLoading && !orgNameLoading && (
```
gtarpenning, line 210: this is to not block on failed query

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -522,6 +522,10 @@ export const CallsTable: FC<{
       : [];
   }, [allRowKeys, columnVisibilityModel, tableData]);
 
+  const refColumnsToExpand = useMemo(() => {
+    return Array.from(expandedRefCols).filter(col => columnsWithRefs.has(col));
```
tssweeney, line 526: under what conditions does `expandedRefColumns` contain cols not in `columnsWithRefs`?

tssweeney (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
gtarpenning: Many actually, although I did not bother to figure out why. Example: 

When I first load the page:
Screenshot 2024-08-29 at 4 46 08 PM

When I expand "response":
Screenshot 2024-08-29 at 4 46 31 PM

tssweeney: oh - that is because there is some silly thing we did that tries to default to including inputs.example

```
const [expandedRefCols, setExpandedRefCols] = useState<Set<string>>(
    new Set<string>().add('inputs.example')
  );
```

gtarpenning: Right, made it simpler to account for that

tssweeney (approved):

PR #2287, opened 2024-08-29
chore(app): make generated color of labeled tags deterministic

bcsherma:


bcsherma:

weave-js/src/core/model/graph/editing/hash.ts
```diff
@@ -3,7 +3,7 @@ import _ from 'lodash';
 import type {EditingNode, EditingOp} from './types';
 
 // TODO: is this unique enough for now?
-const cyrb53 = (str: string, seed = 0) => {
+export const cyrb53 = (str: string, seed = 0) => {
```
gtarpenning, line 6: lol this is crazy

weave-js/src/components/Tag/Tag.tsx
```diff
@@ -22,12 +23,18 @@ export const DEFAULT_TAG_ICON = 'tag';
...
   isInteractive,
+  label,
 }: {
   color?: TagColorName;
   isInteractive?: boolean;
+  label?: string;
 }) {
-  const defaultColorRef = React.useRef(color ?? getRandomTagColor());
-  const tagColor = color ?? defaultColorRef.current;
+  const randomColor = useRef(color || label ? undefined : getRandomTagColor());
```
gtarpenning, line 32:
```suggestion
  const randomColor = useRef((color ?? label) ? undefined : getRandomTagColor());
```

gtarpenning: Actually can we just make this block use conditionals? I think it might be worth it. Like: 

```javascript
function makeTagColor(color, label) {
   if (color) {
      return color
   }
   if (label) {
      return getTagColorByString(label)
   }
   return getRandomTagColor()
}
```

or something like that?

gtarpenning (commented):
This looks safe so if its pressing i feel comfortable stamping, but it is sort of getting messy

weave-js/src/components/Tag/Tag.tsx
gtarpenning: or even
```javascript
const tagColor = color ? color : label ? getTagColorByString(label) : getRandomTagColor()
```

bcsherma: The thing is we don't want to `useRef` for the color if it is assigned by the label. If you remove tags from the beginning of the list the colors stay the same by position, because of the `useRef`. We want the color to be a hash of the label and never vary by list order.

https://github.com/user-attachments/assets/74442a72-cdb9-44c1-a895-e8933f4248b5

weave-js/src/core/model/graph/editing/hash.ts
bcsherma: perfect stack overflow copypasta https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript

weave-js/src/components/Tag/Tag.tsx
gtarpenning: okay facts

Then maybe this? 

```javascript
const randomColor = useRef((color ?? label) ? undefined : getRandomTagColor());
const tagColor = color ? color : label ? getTagColorByString(label) : randomColor.current
````

gtarpenning: what you did was perfect i'm being a mega nit.

gtarpenning (approved):

connieelee (approved):

PR #2290, opened 2024-08-30
chore(weave): Change `get_current_call` to raise when None

andrewtruong:


andrewtruong:
closed in favour of https://github.com/wandb/weave/pull/2379

PR #2297, opened 2024-09-03
chore: show object name if set

gtarpenning:
https://wandb.atlassian.net/browse/WB-19884

Handle object `name` exactly the same as `description`, rather than always filtering out `name`. 

Prod: 
Screenshot 2024-09-03 at 1 40 05 PM

This branch:
Screenshot 2024-09-03 at 1 39 46 PM


tssweeney (approved):
Does this fix the table as well?

PR #2302, opened 2024-09-04
fix: Make Image serialization background process

tssweeney:
This PR backgrounds & parallelizes Image file uploads. Calling 50 ops that make accept and return images goes from ~21s to ~2 seconds. Moreover, the upload time is nearly entirely at program exit.

(After  |    Before)
Screenshot 2024-09-10 at 19 48 59


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
tssweeney, line 1: This is intentionally checked in and helps vsocde know where to find our tests. Unrelated to images

weave/trace/async_job_queue.py
```diff
@@ -0,0 +1,120 @@
+import atexit
+import concurrent.futures
+import threading
+from concurrent.futures import Future
+from typing import Any, Callable, TypeVar
+
+T = TypeVar("T")
+
+MAX_WORKER_DEFAULT = 2**6
```
gtarpenning, line 9:
```suggestion
MAX_WORKER_DEFAULT = 2**6  # 64
```

weave/trace/async_job_queue.py
```diff
@@ -0,0 +1,120 @@
...
+
+    Attributes:
+        executor (concurrent.futures.ThreadPoolExecutor): The executor used to run jobs.
+        _lock (threading.Lock): A lock to ensure thread-safe operations.
+        _is_shutdown (bool): A flag indicating whether the queue has been shut down.
+        _active_jobs (set): A set to keep track of active jobs and callbacks.
+        _max_workers (int): The maximum number of worker threads to use.
+
+    Args:
+        max_workers (int): The maximum number of worker threads to use. Defaults to 5.
```
gtarpenning, line 28:
```suggestion
        max_workers (int): The maximum number of worker threads to use. Defaults to 64.
```

weave/trace/async_job_queue.py
```diff
@@ -0,0 +1,120 @@
...
+        If the queue has been shut down, it will be restarted automatically.
+
+        Args:
+            func: The function to be executed.
+            *args: Positional arguments to pass to the function.
+            **kwargs: Keyword arguments to pass to the function.
+
+        Returns:
+            A Future representing the execution of the job.
+        """
```
gtarpenning, line 63: i feel like an example might be more useful here

weave/trace/weave_client.py
```diff
@@ -408,7 +412,9 @@ def __init__(
         self.project = project
         self.server = server
         self._anonymous_ops: dict[str, Op] = {}
+        self.async_job_queue = AsyncJobQueue()
         self.ensure_project_exists = ensure_project_exists
+        atexit.register(self._cleanup)
```
gtarpenning, line 417: should we put this on a dunder method? post_init? @andrewtruong would know where

weave/trace/async_job_queue.py
gtarpenning: too many?

gtarpenning (approved):

weave/trace/weave_client.py
tssweeney: Not sure what you are recommending here. We do call cleanup in __del__ as well.

weave/trace/async_job_queue.py
tssweeney: reduced to 8

weave/trace/async_job_queue.py
tssweeney: ty

weave/trace/async_job_queue.py
tssweeney: added

PR #2189, opened 2024-08-21
chore(docs): adding presidio pii cookbook from Hesham

NiWaRe:
In this tutorial, we'll demonstrate how to utilize Weave while preventing your Personally Identifiable Information (PII) data from being incorporated into Weave or the LLMs you employ.

To protect our PII data, we'll employ Microsoft's [Presidio](https://microsoft.github.io/presidio/), a Python-based data protection SDK. This tool provides redaction and replacement functionalities, both of which we will implement in this tutorial.

-- PII tutorial by Hesham

PR #2305, opened 2024-09-04
chore(docs): Re-do the PII cookbook based on #2189

ash0ts:
#2189 But with additional details


hesham-wandb (commented):
Hi Anish, 

Nico and I took a look at the new changes. Overall, looks good, but there are a couple of minor formatting issues; ex: "# Method 2: Weave Model Setup". The other thing is we think the optional section "Encrypting our data" is a great, however, we think it might be a little too much to have all that material in one tutorial. So maybe it can be it own thing? like can we add a few sentences about what the optional step is and a link to the colab or github repo or a new tutorial with this? 

CC: @niware

ash0ts:
Hey @hesham-wandb! Fixed the title of the line you pointed out.

> The other thing is we think the optional section "Encrypting our data" is a great, however, we think it might be a little too much to have all that material in one tutorial. So maybe it can be it own thing? like can we add a few sentences about what the optional step is and a link to the colab or github repo or a new tutorial with this?

I'm not sure this warrants its own cookbook. Should I try to make the whole thing collapsible? Not sure what your idea for this could be but would be happy to make the changes

tssweeney (approved):

PR #2314, opened 2024-09-05
fix: Fix Feedback for dedicated by using correct Apollo hook

tssweeney:
Internal Jira: https://wandb.atlassian.net/browse/WB-20703

Allows the gorillaGQL url to be passed into browse3.

### DIFF
--- a/weave-js/src/apollo.ts
+++ b/weave-js/src/apollo.ts
@@ -29,7 +29,7 @@ const authMiddleware = new ApolloLink((operation, forward) => {
 
 export const makeGorillaApolloClient = (
   gorillaApolloEndpoint: string = `${window.WEAVE_CONFIG.WANDB_BASE_URL}/graphql`
-): ApolloClient<object> => {
+) => {
   return new ApolloClient({
     link: ApolloLink.from([
       authMiddleware,
--- a/weave-js/src/apollo.ts
+++ b/weave-js/src/apollo.ts
@@ -26,7 +26,9 @@ const authMiddleware = new ApolloLink((operation, forward) => {
   });
 });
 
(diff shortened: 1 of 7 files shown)


weave-js/src/apollo.ts
```diff
@@ -26,7 +26,9 @@ const authMiddleware = new ApolloLink((operation, forward) => {
   });
 });
 
-export const apolloClient = new ApolloClient({
+// This is a hard-coded client that is used in the old weave.wandb.ai and should
+// be removed with that codebase. Using this is INVALID in any proper wandb app.
+export const deprecatedApolloClientDoNotUse = new ApolloClient({
```
onx2, line 31: I think if you mark this in JSDocs with `@deprecated` it'll strikethrough in editors like VSCode which could help avoid usage.

weave-js/src/components/UserLink.tsx
```diff
@@ -233,7 +232,7 @@ type UserLinkProps = {
   hasPopover?: boolean; // Can you click to open more a popup
 };
 
-const fetchUser = (userId: string) => {
+const fetchUser = (userId: string, apolloClient: ApolloClient<object>) => {
```
onx2, line 235: Is the concrete type known here so we can avoid `object`?

tssweeney: nope (: it is just the cache shape which we don't care about - `any` would also suffice

jamie-rasmussen (approved):

weave-js/src/components/UserLink.tsx
onx2: I think you might be able to omit the type parameter then? 🤔

PR #2317, opened 2024-09-05
maybe prevent another infinite loop?

onx2:


onx2:
@bcsherma do you think we need this additional check here?

bcsherma:
> @bcsherma do you think we need this additional check here?

@onx2 I don't think so. `match[0]` is the matching text, and if `match[0]` is falsey and we break if its empty

PR #2319, opened 2024-09-05
feat: object and op deletion backend + basic sdk + basic fe

gtarpenning:
Combo pr for testing / draft review

https://wandb.atlassian.net/browse/WB-20784

This pr: 
- adds a flexible `obj_delete` api that can delete specific versions of an object, or all versions of the object. 
- server api inserts an object with identical primary keys and a `deleted_at` field. **Before insert, we read from the db, and insert the full data payload, ensuring no data is lost.** `ReplacingMergeTree` deduplicates at merge, keeping the latest version of each row grouped by the primary (order by) key. 
    * "The most recently created part (the last insert) will be the last one in the selection. Thus, after deduplication, the very last row from the most recent insert will remain for each unique sorting key." [link](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree)
- updates the read path to filter out objects with `deleted_at` set. 
- the client in this pr catches the deleted object error when de-reffing nested objects, instead of throwing. 
- adds a trash button to the object versions page
- updates the objects query to refetch on delete

TODO:
- add restriction on deleting object versions to no more than 100 at a time -- we _do_ read the `val_dump` here so its important to add a sane limit. 
- investigate performance implications for massive data deletion requests. Example: try to delete an object with 100,000 versions. This should be fine without batching, as we do an objects query without grabbing `val_dump`, and deletion occurs as an insertion. 

Api usage: 
```python

api = weave.init()
obj = obj.get()

# delete a single version of an object
api.delete_object_version(obj)
# delete all versions of an object
api.delete_object_all_versions(obj.object_id)
```

Depends on: 
- [server pr](https://github.com/wandb/core/pull/23718)
- [types pr](https://github.com/wandb/weave/pull/2320)

Screenshot 2024-09-09 at 3 53 03 PM

delete-obj-1

Deleted object in the call detail view: 
Screenshot 2024-12-11 at 11 52 36 AM

Deleting in the calls table (refresh needs some work):
test-delete


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1400,21 +1442,25 @@ def _select_objs_query(
...
-                        kind,
-                        object_id
+                        PARTITION BY project_id, kind, object_id
                         ORDER BY created_at ASC
                     ) AS _version_index_plus_1,
                     _version_index_plus_1 - 1 AS version_index,
+                    row_number() OVER (
+                        PARTITION BY project_id, kind, object_id
+                        ORDER BY (deleted_at IS NOT NULL), created_at ASC
+                    ) AS alive_version_index,
```
gtarpenning, line 1460: the current index of this row, of those that are not deleted

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1400,21 +1442,25 @@ def _select_objs_query(
...
                     count(*) OVER (PARTITION BY project_id, kind, object_id) as version_count,
-                    if(_version_index_plus_1 = version_count, 1, 0) AS is_latest
+                    COUNT(*) OVER (PARTITION BY project_id, kind, object_id ORDER BY (deleted_at is not NULL)) AS deleted_count,
```
gtarpenning, line 1462: count of all deleted at rows

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectVersionPage.tsx
```diff
@@ -500,3 +533,96 @@ const OpVersionCallsLink: React.FC<{
...
+    $align === 'left' ? 'flex-start' : 'flex-end'} !important;
+  padding: 32px 32px 32px 32px !important;
+`;
+DialogActions.displayName = 'S.DialogActions';
+
+const DeleteModal: React.FC<{
+  object: ObjectVersionSchema;
+  open: boolean;
+  onClose: () => void;
+}> = ({object, open, onClose}) => {
```
gtarpenning, line 564: TODO: dry this with the call delete modal

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -648,6 +655,40 @@ def objs_query(self, req: tsi.ObjQueryReq) -> tsi.ObjQueryRes:
...
+            )
+        ).obj
+
+        deleted_at = datetime.datetime.now()
+        ch_obj = ObjDeleteCHInsertable(
+            project_id=req.project_id,
+            object_id=req.object_id,
+            digest=req.digest,
+            kind=get_kind(db_obj.val),
+            val_dump=json.dumps(db_obj.val),
```
tssweeney, line 679: interestingly, if we made this value something like `DELETED`, then we would never need to clean this up

gtarpenning: oh nice! we could add a flag here for "hard delete" which sets that

weave/trace/vals.py
```diff
@@ -122,6 +125,20 @@ def save(self) -> ObjectRef:
         raise NotImplementedError("Traceable.save not implemented")
         # return self.server.mutate(self.ref, mutations)
 
+    def delete(self) -> None:
+        if self.ref is None:
+            raise ValueError("Cannot delete object that is not saved")
+        if not isinstance(self.ref, (ObjectRef, OpRef)):
```
tssweeney, line 131: i wonder how to handle datasets, where really they want to delete the table

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
```diff
@@ -153,10 +155,17 @@ export const ObjectViewer = ({
...
     for (const [r, v] of _.zip(refs, resolvedRefData)) {
-      if (!r || !v) {
+      if (!r) {
         // Shouldn't be possible
         continue;
       }
+      if (!v) {
+        // Value for ref not found, probably deleted
+        refValues[r] = {
+          _weave_is_deleted_ref: objectRefDisplayName(parseRef(r)).label,
```
gtarpenning, line 165: this is terrible, we need a better way of storing/rendering deleted refs in the object viewer.

PR #2341, opened 2024-09-10
chore(weave): change raise to print for netrc existing with api key

jwlee64:
## Description
Changes the reaction to netrc and apikey existing to a print and not a raise when not in prod

## Testing

How was this PR tested?

### DIFF
--- a/weave/legacy/weave/environment.py
+++ b/weave/legacy/weave/environment.py
@@ -251,9 +251,7 @@ def weave_wandb_api_key() -> typing.Optional[str]:
     env_api_key = _wandb_api_key_via_env()
     netrc_api_key = _wandb_api_key_via_netrc()
     if env_api_key and netrc_api_key:
-        raise errors.WeaveConfigurationError(
-            "WANDB_API_KEY should not be set in both ~/.netrc and the environment."
-        )
+        print("WANDB_API_KEY should not be set in both ~/.netrc and the environment.")
     return env_api_key or netrc_api_key
 
 
--- a/weave/legacy/weave/environment.py
+++ b/weave/legacy/weave/environment.py
@@ -250,14 +250,15 @@ def _wandb_api_key_via_netrc_file(filepath: str) -> typing.Optional[str]:
 def weave_wandb_api_key() -> typing.Optional[str]:
     env_api_key = _wandb_api_key_via_env()
     netrc_api_key = _wandb_api_key_via_netrc()
-    if env_api_key and netrc_api_key and wandb_production():
-        raise errors.WeaveConfigurationError(
-            "WANDB_API_KEY should not be set in both ~/.netrc and the environment."
-        )
-    elif env_api_key and netrc_api_key and env_api_key != netrc_api_key:
-        print(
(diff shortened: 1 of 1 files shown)


weave/legacy/weave/environment.py
```diff
@@ -251,9 +251,7 @@ def weave_wandb_api_key() -> typing.Optional[str]:
     env_api_key = _wandb_api_key_via_env()
     netrc_api_key = _wandb_api_key_via_netrc()
     if env_api_key and netrc_api_key:
-        raise errors.WeaveConfigurationError(
-            "WANDB_API_KEY should not be set in both ~/.netrc and the environment."
-        )
+        print("WANDB_API_KEY should not be set in both ~/.netrc and the environment.")
```
jamie-rasmussen, line 254: * If the contents are identical I think we could be silent.
* I'd change the message to explain a bit better. Maybe something like:
"There are different credentials in the netrc file and the environment. Using the environment value." 
* Not worth fixing but just noting that NETRC file might not be located at ~/.netrc (Windows, I think we might support the NETRC env var for alternate path, etc.)

jwlee64: makes sense
_wandb_api_key_via_netrc() -> i believe checks the windows location as well

jamie-rasmussen: Right, just meant that on windows this message would have an incorrect path (`~/.netrc` vs. `~/_netrc`)

weave/legacy/weave/environment.py
```diff
@@ -250,9 +250,9 @@ def _wandb_api_key_via_netrc_file(filepath: str) -> typing.Optional[str]:
 def weave_wandb_api_key() -> typing.Optional[str]:
     env_api_key = _wandb_api_key_via_env()
     netrc_api_key = _wandb_api_key_via_netrc()
-    if env_api_key and netrc_api_key:
-        raise errors.WeaveConfigurationError(
-            "WANDB_API_KEY should not be set in both ~/.netrc and the environment."
+    if env_api_key and netrc_api_key and env_api_key != netrc_api_key:
```
tssweeney, line 253: i would still raise in production

weave/legacy/weave/environment.py
```diff
@@ -250,10 +250,14 @@ def _wandb_api_key_via_netrc_file(filepath: str) -> typing.Optional[str]:
 def weave_wandb_api_key() -> typing.Optional[str]:
     env_api_key = _wandb_api_key_via_env()
     netrc_api_key = _wandb_api_key_via_netrc()
-    if env_api_key and netrc_api_key:
+    if env_api_key and netrc_api_key and wandb_production():
```
tssweeney, line 253: nit:

```
    if env_api_key and netrc_api_key:
        if wandb_production():
          raise errors.WeaveConfigurationError(
              "WANDB_API_KEY should not be set in both ~/.netrc and the environment."
          )
        elif env_api_key != netrc_api_key:
          print(
              "There are different credentials in the netrc file and the environment. Using the environment value."
          )
```

PR #2345, opened 2024-09-11
chore(docs): Document low-level Python API for creating calls

adrnswanberg:
Adds an example to docs of how to use `weave.create_call` and `weave.finish_call` to trace function calls.

Relevant new section:

image

### DIFF
--- a/docs/docs/guides/tracking/tracing.mdx
+++ b/docs/docs/guides/tracking/tracing.mdx
@@ -143,7 +143,7 @@ You can also manually create Calls using the API directly.
         weave.finish_call(call, output="Hello, World!")
 
     # Call your function
-    print(my_function.call("World"))
+    print(my_function("World"))
     ```
     </TabItem>
     <TabItem value="service_api" label="HTTP API">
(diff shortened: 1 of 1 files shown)


docs/docs/guides/tracking/tracing.mdx
```diff
@@ -127,30 +124,53 @@ with weave.attributes({'env': 'production'}):
...
+        # Start a call
+        call = weave.create_call(op="my_function", inputs={"name": name})
+
+        # ... your function code ...
+
+        # End a call
+        weave.finish_call(call, output="Hello, World!")
+
+    # Call your function
+    print(my_function.call("World"))
```
jamie-rasmussen, line 146: Should this just be `my_function("World")`

adrnswanberg: good catch -- copied too much of the existing code 🤦 

while we're at it -- any reason to present the other examples on this page this way?

jamie-rasmussen: Sorry, not sure what you mean

adrnswanberg: image

^ here's the example we have for using @weave.op. Not sure why we want to highlight that you can do .call here. Should we just simplify?

jamie-rasmussen: Oh I see - yes I agree, no reason to use `.call` here

tssweeney (approved):

PR #2351, opened 2024-09-11
fix(weave): Object viewer properly expands refs on expand all

jwlee64:
## Description

-Fixes https://wandb.atlassian.net/browse/WB-20839
https://weightsandbiases.slack.com/archives/C01T8BLDHKP/p1726001368443599

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
@@ -76,8 +76,8 @@ const getRefFromObjectPath = (obj: any, path: string) => {
     );
 
     if (parsedPath) {
-      let property = parsedPath[1]; // e.g., "something"
-      let index = parsedPath[3]; // e.g., "0"
+      const property = parsedPath[1]; // e.g., "something"
+      const index = parsedPath[3]; // e.g., "0"
 
       // If acc is undefined, return undefined, or get the property if it exists
       if (acc && acc[property]) {
@@ -88,7 +88,7 @@ const getRefFromObjectPath = (obj: any, path: string) => {
         }
         if (index !== undefined) {
           // Get the array element if index is present
-          return acc[parseInt(index)];
+          return acc[parseInt(index, 10)];
         }
         return acc;
       }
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
```diff
@@ -90,6 +124,22 @@ export const ObjectViewer = ({
...
+  // We need this, because expandAll adds a bunch of ids to expandedIds, but not to expandedRefs.
+  for (const id of expandedIds) {
+    if (typeof id === 'number') {
+      continue;
+    }
+    if (!(id in expandedRefs)) {
+      // ref is a ref or undefined
+      const ref = getRefFromObjectPath(resolvedData, id);
+      if (isWeaveRef(ref)) {
+        addExpandedRef(id, ref);
```
tssweeney, line 138: This will create N (N=number of refs) rerenders and is subject to thrashing. let's refactor. See comment above

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewer.tsx
```diff
@@ -64,6 +64,40 @@ const getRefs = (data: Data): string[] => {
 
 type RefValues = Record<string, any>; // ref URI to value
 
+// Returns a ref or undefined from a path like "value.something[0].nested" or "value.nested" and an object
+const getRefFromObjectPath = (obj: any, path: string) => {
```
tssweeney, line 68: When i see path parsing like this, I hear warning bells. It seems like this is fragile and leveraging a contract that is not explicit. I think we can do better here. Specifically, the expandId is the stringified id from the traverse module. i'd recommend creating a single useEffect:

```typescript
  // This effect will ensure that all "expandedIds" whose value is a ref
  // have the ref added to the `expandedRefs` state.
  useEffect(() => {
    const expandRefsToAdd: {[ref: string]: string} = {};
    const mapper = (context: TraverseContext) => {
      const contextPath = context.path.toString();
      if (
        expandedIds.includes(contextPath) &&
        isWeaveRef(context.value) &&
        expandedRefs[context.value] == null
      ) {
        expandRefsToAdd[context.value] = contextPath;
      }
      return context.value
    };
    mapObject(data, mapper);
    setExpandedRefs(eRefs => ({...eRefs, ...expandRefsToAdd}));
  }, [data, expandedIds, expandedRefs, setExpandedRefs]);
```

I think something like this will solve the problem

tssweeney (approved):

PR #2353, opened 2024-09-11
fix: strip large fields after failed clickhouse insert (3.5MB row size, 1MB single field size)

gtarpenning:
## Description

- Fixes [WB-19623](https://wandb.atlassian.net/browse/WB-19623)

This pr: 
- adds constant representing the single row maximum data size the clickhouse client can log
- when inserts fail, either in the objects or calls table: strip out large fields and retry

## Testing
- adds two tests

[WB-19623]: https://wandb.atlassian.net/browse/WB-19623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -161,6 +161,41 @@ def _flush_calls(
...
+                bytes_size = len(item.model_dump_json().encode("utf-8"))
+                seen_bytes += bytes_size
+                if bytes_size > CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT:
+                    mb = bytes_size / (1024 * 1024)
+                    id_, mode = (
+                        (item.req.start.id, "start")
+                        if isinstance(item, StartBatchItem)
+                        else (item.req.end.id, "end")
+                    )
+                    print(
```
tssweeney, line 179: let's use `logging.error`

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -161,6 +161,41 @@ def _flush_calls(
...
+                        f"Dropping call {mode} ({id_}) with size {mb:.2f} MiB larger than "
+                        f"{CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT / (1024 * 1024):.2f} MiB."
+                        " If logging images, please encode them as `PIL.Image`"
```
tssweeney, line 182: It seems like we want to re-use this error when encountering a failed batch of that specific error as well.

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -161,6 +161,41 @@ def _flush_calls(
...
+                bytes_size = len(item.model_dump_json().encode("utf-8"))
+                seen_bytes += bytes_size
+                if bytes_size > CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT:
+                    mb = bytes_size / (1024**2)
+                    if isinstance(item.req, tsi.CallStartReq):
+                        id_, mode = (item.req.start.id, "start")
+                        item.req.start.inputs = {}
+                    else:
+                        id_, mode = (item.req.end.id, "end")
+                        item.req.end.output = {}
```
tssweeney, line 179: Output = None

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -161,6 +161,41 @@ def _flush_calls(
...
+            seen_bytes = 0
+            for i, item in enumerate(batch):
+                # below takes tens of miliseconds for the large payloads we skip
+                bytes_size = len(item.model_dump_json().encode("utf-8"))
+                seen_bytes += bytes_size
+                if bytes_size > CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT:
+                    mb = bytes_size / (1024**2)
+                    if isinstance(item.req, tsi.CallStartReq):
+                        id_, mode = (item.req.start.id, "start")
+                        item.req.start.inputs = {}
```
tssweeney, line 176: Additional consideration:

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -161,6 +161,41 @@ def _flush_calls(
             self._flush_calls(batch[split_idx:], _should_update_batch_size=False)
             return
 
+        final_batch = []
+        if encoded_bytes > CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT:
```
tssweeney, line 165: it would be nice to somehow mark the call or its inputs / outputs in these cases so we know that this happened and can surface that info to the user

weave/tests/trace/test_weave_client.py
```diff
@@ -1183,6 +1183,24 @@ def test_table_partitioning(network_proxy_client):
...
+
+    @weave.op
+    def save_large_dict(len: int):
+        return {f"{i}": "x" for i in range(len)}
+
+    save_large_string(10_000_000)
+    save_large_dict(10_000_000)
+
+    calls = client.get_calls()
+    assert len(calls) == 2
```
tssweeney, line 1201: probably want to add more assertions here to ensure that you logic works. As implemented, i don't think your redaction code is actually running

weave/trace_server_bindings/remote_http_trace_server.py
gtarpenning: Yeah, I think the right thing is to go with the JSONBlob serializer fix instead. This is more of a workaround...

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -161,6 +164,42 @@ def _flush_calls(
...
+            seen_bytes = 0
+            for i, item in enumerate(batch):
+                # below takes tens of miliseconds for the large payloads we skip
+                bytes_size = len(item.model_dump_json().encode("utf-8"))
+                seen_bytes += bytes_size
+                if bytes_size > CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT:
+                    mb = bytes_size / (1024**2)
+                    if isinstance(item.req, tsi.CallStartReq):
+                        id_, mode = (item.req.start.id, "start")
+                        item.req.start.inputs = ENTITY_TOO_LARGE_ERROR_PAYLOAD
```
tssweeney, line 179: For sentinel: i think we could start with just a string, similar to our redaction string: `<EXCEED_LIMITS>`. Furthermore, for the `inputs` one, you should keep all input keys, and replace the large values with sentinel

gtarpenning: "you should keep all input keys" this is non trivial. What if the inputs have 5 keys each with 1MB payloads? We could iterate through all the values and keep track, and only replace the values once a running tally exceeds our limits... But this would totally fail in the case of thousands of input keys (admittedly not likely) or very large input keys, as the keys themselves would add to the data cap and we would still blow through our limit. @tssweeney

tssweeney: I see. OK, well let's just do `{_weave: {error: ""}}` so we can put other keys inside there

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -123,6 +125,68 @@ def from_env(cls, should_batch: bool = False) -> "RemoteHTTPTraceServer":
...
+            # the large values with a placeholder.
+            if bytes_size > row_bytes_limit:
+                if isinstance(item.req, tsi.CallStartReq):
+                    id_, mode = (item.req.start.id, "start")
+                    for key in item.req.start.inputs:
+                        if _num_bytes(item.req.start.inputs[key]) > val_byte_limit:
+                            item.req.start.inputs[key] = ENTITY_TOO_LARGE_ERROR
+                else:
+                    id_, mode = (item.req.end.id, "end")
+                    for key in item.req.end.output:
```
tssweeney, line 171: almost (: ... output is not necessarily a dict

gtarpenning: ahhh

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -36,6 +36,8 @@ class ServerInfoRes(BaseModel):
 REMOTE_REQUEST_BYTES_LIMIT = (
     (32 - 1) * 1024 * 1024
 )  # 32 MiB (real limit) - 1 MiB (buffer)
+CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT = 3.5 * 1024 * 1024  # 3.5 MiB
+ENTITY_TOO_LARGE_ERROR = "<EXCEED_LIMITS>"
```
tssweeney, line 40: `exceeds`? maybe ...

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -161,6 +224,13 @@ def _flush_calls(
             self._flush_calls(batch[split_idx:], _should_update_batch_size=False)
             return
 
+        if encoded_bytes > CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT:
+            batch = self._strip_large_values_from_batch(batch, encoded_bytes)
+
+        if len(batch) == 0:
+            return
+
+        encoded_data = Batch(batch=batch).model_dump_json().encode("utf-8")
```
tssweeney, line 233: i think you are double encoding in the case that you don't need to strip

tssweeney: as in: encoding twice

gtarpenning: yeah nice. moved second encoding to inside the `if encoded_bytes > CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT:` condition

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -123,6 +125,68 @@ def from_env(cls, should_batch: bool = False) -> "RemoteHTTPTraceServer":
...
+        replace them with placeholder values.
+        """
+
+        def _num_bytes(data: t.Any) -> int:
+            """
+            Calculate the number of bytes in a string.
+
+            This can be computationally expensive, only call when necessary.
+            """
+            return len(str(data).encode("utf-8"))
```
tssweeney, line 144: what about data that does not have a string representation? seems like this could error

weave/tests/trace/test_weave_client.py
```diff
@@ -1200,6 +1206,100 @@ def test_table_partitioning(network_proxy_client):
     )
 
 
+def test_call_batch_size_handling(network_proxy_client, caplog):
```
tssweeney, line 1209: it is worth putting a similar test at the client level to ensure `remote_client._strip_large_values_from_batch(` gets called at the right spot

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -145,6 +145,10 @@ class NotFoundError(Exception):
 ObjRefListType = list[ri.InternalObjectRef]
 
 
+CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT = 3.5 * 1024 * 1024  # 3.5 MiB
+ENTITY_TOO_LARGE_PAYLOAD = '{"_weave": "<EXCEEDS_LIMITS>"}'
```
tssweeney, line 149: Again, i would nest the `"<EXCEEDS_LIMITS>"`. Else we are sort of stuck with a string value here

gtarpenning: oh sorry, I understand.

tssweeney (approved):

PR #2357, opened 2024-09-11
feat: log path to image as Path, auto save image

gtarpenning:
## Description

- Fixes WB-NNNNN

This PR:
- inspects `pathlib.Path` type log values, and if they are an image, opens them as `PIL.Image` and saves them. 

## Testing

- adds test

### DIFF
--- a/weave/type_serializers/Image/image_test.py
+++ b/weave/type_serializers/Image/image_test.py
@@ -1,3 +1,7 @@
+import os
+import tempfile
+from pathlib import Path
+
 from PIL import Image
 
 import weave
@@ -100,3 +104,20 @@ def test_image_as_call_io_refs(client: WeaveClient) -> None:
     assert image_as_solo_output_call.output.tobytes() == exp_bytes
     assert image_as_input_and_output_part_call.inputs["in_img"].tobytes() == exp_bytes
     assert image_as_input_and_output_part_call.output["out_img"].tobytes() == exp_bytes
+
+
+def test_image_from_path(client: WeaveClient) -> None:
+    with tempfile.TemporaryDirectory() as temp_dir:
+        path_str = os.path.join(temp_dir, "img.png")
+        img = Image.new("RGB", (512, 512), "purple")
+        img.save(path_str)
+        exp_bytes = img.tobytes()
+
+        @weave.op
+        def save_image(path: Path) -> None:
+            return {"out-img-path": path}
+
+        save_image(Path(path_str))
+
+        call = save_image.calls()[0]
+        assert call.output["out-img-path"].tobytes() == exp_bytes
--- a/weave/trace/serialize.py
+++ b/weave/trace/serialize.py
@@ -1,4 +1,5 @@
 import typing
+from pathlib import Path
 from typing import Any
 
 from weave.trace import custom_objs
@@ -30,6 +31,14 @@ def to_json(obj: Any, project_id: str, server: TraceServerInterface) -> Any:
 
     if isinstance(obj, (int, float, str, bool)) or obj is None:
         return obj
+
+    if isinstance(obj, Path):
+        # only support image files for now
+        if obj.suffix in [".png", ".jpg", ".jpeg", ".gif", ".tiff"]:
+            # Load the image
+            from PIL import Image
+
+            obj = Image.open(obj)
 
     encoded = custom_objs.encode_custom_obj(obj)
     if encoded is None:


weave/trace/weave_client.py
```diff
@@ -1191,4 +1202,39 @@ def sanitize_object_name(name: str) -> str:
...
+            convert_paths_to_images(v)
+    elif isinstance(obj, str):
+        if image_pattern.match(obj):
+            # Load the image, if PIL is available and file exists
+            try:
+                from PIL import Image
+
+                return PathImage(img=Image.open(obj), path=obj)
+            except (ImportError, FileNotFoundError) as e:
+                logger.warning(f"Failed to open image file: {obj}. {e}")
```
gtarpenning, line 1227: do we want to warn here? silently fail?

weave/type_serializers/Image/image_test.py
```diff
@@ -100,3 +105,47 @@ def test_image_as_call_io_refs(client: WeaveClient) -> None:
...
+        def save_stuff_with_image_embedded(path: Path) -> dict:
+            return {"out": f"Some random text + a path: {path}"}
+
+        save_stuff_with_image_embedded(Path(path_str))
+        call = save_stuff_with_image_embedded.calls()[0]
+        assert call.output["out"] == f"Some random text + a path: {path_str}"
+
+
+def test_image_from_remote_path(client: WeaveClient) -> None:
+    remote_path = "https://www.gstatic.com/webp/gallery/1.jpg"
```
gtarpenning, line 136: we could host this ourself but I think this image has been around since the dawn of the internet and probably isn't going anywhere anytime soon. I suspect that hosting an image we control might be more volatile.

weave/trace/settings.py
```diff
@@ -53,14 +53,29 @@ class UserSettings(BaseModel):
...
+
     If True, saves code for ops so they can be reloaded for later use.
     Can be overrided with the environment variable `WEAVE_CAPTURE_CODE`
-    
+
     WARNING: Switching between `save_code=True` and `save_code=False` mid-script
     may lead to unexpected behaviour.  Make sure this is only set once at the start!
     """
 
+    convert_paths_to_images: bool = True
```
gtarpenning, line 64: defaults to true!

weave/trace/settings.py
```diff
@@ -53,14 +53,29 @@ class UserSettings(BaseModel):
...
+    If True, image file paths will be converted to PathImage objects.
+    Can be overrided with the environment variable `WEAVE_CONVERT_PATHS_TO_IMAGES`
+
+    Example:
+        @weave.op
+        def generate_image(prompt: str):
+            text, url = generate_text_and_image(prompt)
+            return {"text": text, "image": url}
+
+        If WEAVE_CONVERT_PATHS_TO_IMAGES=true, `image` will be converted to PathImage
```
gtarpenning, line 76: change example to falsy

weave/trace/weave_client.py
```diff
@@ -693,6 +703,9 @@ def finish_call(
         *,
         postprocess_output: Optional[Callable[..., Any]] = None,
     ) -> None:
+        if should_convert_paths_to_images():
+            output = convert_paths_to_images(output)
+        self._save_nested_objects(output)
```
andrewtruong, line 708: are you doing this to trigger img saving?

weave/trace/weave_client.py
```diff
@@ -1464,4 +1477,57 @@ def sanitize_object_name(name: str) -> str:
...
+
+
+def convert_paths_to_images(obj: Any) -> Any:
+    """
+    Iterate through the object and find paths to local and remote images.
+
+    If the path is a local image, open it and return a PathImage object.
+    If the path is a remote image, download it and return a PathImage object.
+    """
+    if not pil_dependency:
```
andrewtruong, line 1501: nit: rename to `pil_package_available` -- more obvious when reading

weave/trace/weave_client.py
gtarpenning: yeah, if the setting is True, we rip through the entire output and replace any image paths with real images.

andrewtruong: I meant the `_save_nested_objects` part

gtarpenning: oh, yes, we do the conversion first so that the actual image (bytes) get saved instead of the url. This path uses grabs the `ImagePath` serializer to save it

weave/trace/weave_client.py
```diff
@@ -1464,4 +1466,78 @@ def sanitize_object_name(name: str) -> str:
...
+    elif local_image_pattern.match(obj):
+        try:
+            return PathImage(img=Image.open(obj), path=obj)
+        except Exception as e:
+            logger.warning(f"Failed to open image file: {obj}. {e}")
+
+    return obj
+
+
+def parse_obj(obj: Any, *parsers: Callable[[Any], Any]) -> Any:
```
gtarpenning, line 1508: rough stab, kinda like this?

gtarpenning, line 1508: took a rough stab, kinda like this? @andrewtruong

weave/trace/weave_client.py
```diff
@@ -1232,7 +1255,7 @@ def _save_nested_objects(self, obj: Any, name: Optional[str] = None) -> Any:
         # Case 1: Object:
         # Here we recurse into each of the properties of the object
         # and save them, and then save the object itself.
-        if isinstance(obj, Object):
+        if isinstance(obj, Object) or isinstance(obj, PathImage):
```
gtarpenning, line 1258: this is a hack, but still not working....

weave/trace/weave_client.py
```diff
@@ -641,11 +651,17 @@ def create_call(
         unbound_op = maybe_unbind_method(op)
         op_def_ref = self._save_op(unbound_op)
 
-        inputs_redacted = redact_sensitive_keys(inputs)
```
andrewtruong, line 644: did you remove redaction?

gtarpenning: uh, no, now its registered in `parse_obj_functions`

weave/type_serializers/Image/image.py
```diff
@@ -42,3 +54,64 @@ def load(artifact: MemTraceFilesArtifact, name: str) -> "Image.Image":
...
+
+def is_remote_image(path: str) -> bool:
+    return remote_image_pattern.match(path) is not None
+
+
+def is_base64_image(path: str) -> bool:
+    return base64_image_pattern.match(path) is not None
+
+
+class PathImage(BaseModel):
```
andrewtruong, line 79: I think you can simplify this class a lot by using classmethods, e.g.

```py
PathImage.from_data(...)
PathImage.from_path(...)
PathImage.from_url(...)
```

weave/type_serializers/Image/image.py
```diff
@@ -42,3 +54,64 @@ def load(artifact: MemTraceFilesArtifact, name: str) -> "Image.Image":
...
     if dependencies_met:
         serializer.register_serializer(Image.Image, save, load)
+
+
+# match local image file paths
+image_suffix = r".*\.(png|jpg|jpeg|gif|tiff)"
+local_image_pattern = re.compile(rf"^{image_suffix}$", re.IGNORECASE)
+remote_image_pattern = re.compile(rf"https://.*\.{image_suffix}", re.IGNORECASE)
+# match base64 encoded images
+base64_image_pattern = re.compile(r"^data:image/.*;base64,", re.IGNORECASE)
```
andrewtruong, line 64: move to top and caps as const

weave/trace/weave_client.py
```diff
@@ -1572,4 +1577,60 @@ def sanitize_object_name(name: str) -> str:
...
+    """
+    # Dont mutate reffed objects
+    if get_ref(obj):
+        return obj
+
+    # recurse through container types
+    if isinstance(obj, dict):
+        # Do any dict-type parsing
+        for parser in parsers:
+            param_types = typing.get_type_hints(parser)
```
andrewtruong, line 1614: This is probably ok, but worth double checking because the behaviour of `get_type_hints` can change between python versions: https://docs.python.org/3/library/typing.html#typing.get_type_hints

weave/trace/weave_client.py
```diff
@@ -1572,4 +1577,60 @@ def sanitize_object_name(name: str) -> str:
...
+            if "obj" in param_types and param_types["obj"] == dict:
+                obj = parser(obj)
```
andrewtruong, line 1616: do you need to pass the whole object?

weave/trace/weave_client.py
```diff
@@ -1572,4 +1577,60 @@ def sanitize_object_name(name: str) -> str:
...
+    """
+    if is_remote_image(obj) or is_local_image(obj):
+        return PathImage(path=obj, data=None)
+    elif is_base64_image(obj):
+        return PathImage(data=obj, path=None)
+
+    return obj
+
+
+def parse_obj(obj: Any, parsers: list[Callable[..., Any]]) -> Any:
```
andrewtruong, line 1594: How much time does this add to the user's process?

gtarpenning: In my testing, almost zero, except for the remote images, which incurs a network request. I was able to log about ~200 calls per second including multiple small image payloads.

weave/trace/weave_client.py
gtarpenning: for the redaction above yes

PR #2360, opened 2024-09-11
feat: Add a refresh button to the Calls Table

tssweeney:
In this PR, I add a little refresh button to the calls table. It simply updates the table and is way way better than doing a full page refresh. I was going to implement polling, but:
a) I want to avoid a lot of server pings
b) To really make polling effective, we probably want to add more smarts to:
   1. Quickly determine which ids need refreshing
   2. Refetch those ids in particular

This is a little more advanced than the timebox I gave myself, so stopping with basic refresh for now.

Screenshot 2024-09-12 at 14 38 12

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterPanel.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterPanel.tsx
@@ -21,7 +21,7 @@ type FilterPanelProps = {
 
 export const FilterPanel = (props: FilterPanelProps) => {
   return (
-    <div className="min-w-90 flex-auto self-stretch overflow-hidden">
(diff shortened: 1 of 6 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterPanel.tsx
```diff
@@ -21,7 +21,7 @@ type FilterPanelProps = {
 
 export const FilterPanel = (props: FilterPanelProps) => {
   return (
-    <div className="min-w-90 flex-auto self-stretch overflow-hidden">
+    <div className="min-w-90 flex-auto self-stretch">
```
tssweeney, line 24: This minor change allows the Filter button to not disappear when the main view is very small

jwlee64 (approved):

PR #2363, opened 2024-09-12
fix(weave): Remove erroneous self-binding of weave ops on classes

andrewtruong:
## Description

https://wandb.atlassian.net/browse/WB-20637

This removes auto-self-binding that could sometimes bind to the wrong instance, leading to unexpected results.

## Testing

Unit tests

### DIFF
--- a/docs/docs/guides/tracking/tracing.mdx
+++ b/docs/docs/guides/tracking/tracing.mdx
@@ -79,6 +79,29 @@ def my_function(name: str):
 print(my_function.call("World"))
 ```
 
+:::note
+If your op is a method on a class, you need to pass the instance as the first argument to the op (see example below).
+:::
+
+
+```python showLineNumbers
+import weave
+
+# Initialize Weave Tracing
+weave.init("intro-example")
+
+class MyClass:
+    # Decorate your method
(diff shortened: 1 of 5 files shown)


docs/docs/guides/tracking/tracing.mdx
```diff
@@ -79,6 +79,29 @@ def my_function(name: str):
...
+class MyClass:
+    # Decorate your method
+    @weave.op
+    def my_method(self, name: str):
+        return f"Hello, {name}!"
+
+instance = MyClass()
+
+# Call your method -- Weave will automatically track inputs and outputs
+instance.my_method.call(instance, "World")
```
tssweeney, line 102: What is the error message that occurs when this is not the case - can we give a good instructional error?

andrewtruong: It's the default `ValueError: missing arg ____`, but we can make it extra explicit since this is a special func

andrewtruong: You get an `OpCallError` with clear instructions to pass self:
image

tssweeney (approved):

PR #2367, opened 2024-09-12
feat: Make Image serialization background process

tssweeney:
This PR is a patch to the previously reverted: https://github.com/wandb/weave/pull/2302 ... Essentially i made the exit procedure safer and tested in:
* local ipython
* local python
* colab

This is a good view to see changes on top of the previous pr: https://github.com/wandb/weave/pull/2367/files/1224f719c58b7289ecccc1cfaba8603b56893665..2316cb120567a657eb79ed1e0bc67e9b6d7db1bf

The original implementation had a bug that when running in ipython, explicitly exiting would throw:

```
Exception ignored in: <function WeaveClient.__del__ at 0x141ae9080>
  File "/Users/jamie/source/wandb/core/services/weave-python/weave-public/weave/trace/weave_client.py", line 1116, in __del__
  File "/Users/jamie/source/wandb/core/services/weave-python/weave-public/weave/trace/weave_client.py", line 1120, in _cleanup
  File "/Users/jamie/source/wandb/core/services/weave-python/weave-public/weave/trace/weave_client.py", line 1111, in flush
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

This was because the `__del__` method was running after the other symbol was deleted. This cleans up such cases

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -1115,7 +1115,7 @@ class WeaveClient:
         self.async_job_queue.flush()
         if self._server_is_flushable:
             # We don't want to do an instance check here because it could
-            # be suseptibale to shutdown race conditions. So we save a boolean
+            # be susceptible to shutdown race conditions. So we save a boolean
             # _server_is_flushable and only call this if we know the server is
             # flushable. The # type: ignore is safe because we check the type
             # first.
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -1114,7 +1114,12 @@ class WeaveClient:
         # Used to wait until all currently enqueued jobs are processed
         self.async_job_queue.flush()
         if self._server_is_flushable:
-            self.server.call_processor.wait_until_all_processed()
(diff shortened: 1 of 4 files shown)


weave/trace/weave_client.py
```diff
@@ -1106,18 +1110,27 @@ def _op_runs(self, op_def: Op) -> Sequence[Call]:
...
-    def flush(self) -> None:
+    def _flush(self) -> None:
+        # Used to wait until all currently enqueued jobs are processed
         self.async_job_queue.flush()
-        if isinstance(self.server, RemoteHTTPTraceServer):
-            if self.server.should_batch:
-                self.server.call_processor.wait_until_all_processed()
+        if self._server_is_flushable:
+            # We don't want to do an instance check here because it could
+            # be suseptibale to shutdown race conditions. So we save a boolean
```
jamie-rasmussen, line 1118: suseptibale => susceptible

jamie-rasmussen (approved):

PR #2368, opened 2024-09-12
feat: Tim/disable weave completely

tssweeney:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?


andrewtruong (approved):
nit: We should have a more direct error msg like `WeaveDisabledError` when the user tries to use the client (and fails) when weave is disabled.  Right now you'll see something like this `HTTPError` which is less clear about why the call failed
image

image

PR #2370, opened 2024-09-12
fix: dont crash the async batch handler when a batch raises

gtarpenning:
When a batch in the `AsyncBatchProcessor` fails, for any reason, log to `logger.error` instead of crashing. 

Improve the error message for when a table insertion fails due to a row being too large. This can happen in any of the tables that use insert. 

In the `http_remote_trace_server` pass through the raised error when its our special `413` error.


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1534,7 +1534,11 @@ def _insert(
                 # │    return 1 << (21 - int(log(row_size, 2)))
                 # │ValueError: negative shift count
                 # when we try to insert something that's too large.
-                raise RequestTooLarge("Could not insert record")
+                raise InsertTooLarge(
```
gtarpenning, line 1537: Thoughts on copy here, the user should see this.

weave/trace_server/async_batch_processor.py
```diff
@@ -1,10 +1,12 @@
 import atexit
+import logging
 import time
 from queue import Queue
 from threading import Event, Lock, Thread
 from typing import Callable, Generic, List, TypeVar
 
 T = TypeVar("T")
+logger = logging.getLogger(__name__)
```
gtarpenning, line 9: will this work correctly to get the weave logger?

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: I would maybe put a CTA - if you need help, please email / post here.... etc.

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -166,6 +166,10 @@ def _flush_calls(
             data=encoded_data,
             auth=self._auth,
         )
+        if r.status_code == 413:
+            # handle 413 explicitly to provide actionable error message
+            reason = json.loads(r.text)["reason"]
+            raise requests.HTTPError(f"413 Client Error: {reason}", response=r)
```
tssweeney, line 172: does the 413 error code come along for the ride here? We want to make sure that it is the same code

weave/trace_server/async_batch_processor.py
```diff
@@ -54,7 +56,10 @@ def _process_batches(self) -> None:
                 current_batch.append(self.queue.get())
 
             if current_batch:
-                self.processor_fn(current_batch)
+                try:
```
tssweeney, line 59: This blind catch-all feels like a hammer. Do we always want to catch in every exception case? Maybe. Worth investigating

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: Also, if you can make it shorter. Ask gpt

tssweeney (approved):

weave/trace_server_bindings/remote_http_trace_server.py
gtarpenning: Yeah, the response has the status code still

PR #2373, opened 2024-09-12
feat(weave): Add post-processing options for call inputs and outputs

andrewtruong:
## Description

https://wandb.atlassian.net/browse/WB-20474

This PR adds `postprocess_inputs_func` and `postprocess_outputs_func` options onto the op decorator to allow users to customize what data gets logged to weave.  These transforms do not affect the function call itself, just what data gets logged.

## Testing

Unit tests


jamie-rasmussen:
The names like "postprocess_inputs_func" seem kind of long to me. I think we could at least drop `_func`. Maybe something like "amend_inputs"?

I think output should be singular to match other places in system.

weave/trace/weave_client.py
```diff
@@ -560,9 +560,13 @@ def create_call(
             op_str = op
 
         inputs_redacted = redact_sensitive_keys(inputs)
+        if op.postprocess_inputs:
+            inputs_postprocessed = op.postprocess_inputs(inputs_redacted)
```
jamie-rasmussen, line 564: I'm wondering whether it would be useful to have `postprocess_inputs` work on the `inputs` rather than `inputs_redacted` - that way, if we were redacting something the user didn't want redacted they would have a way to opt out. But our current redaction is pretty limited and I doubt there are many false positives, so this is probably for the best. Just raising it for consideration.

weave/trace/weave_client.py
```diff
@@ -627,11 +631,22 @@ def create_call(
...
+        exception: Optional[BaseException] = None,
+        *,
+        postprocess_output: Optional[Callable[..., Any]] = None,
     ) -> None:
-        self._save_nested_objects(output)
         original_output = output
-        output = map_to_refs(original_output)
+
+        if postprocess_output:
+            postprocessed_outputs = postprocess_output(original_output)
```
jamie-rasmussen, line 644: outputs => output for consistency

docs/docs/guides/tracking/ops.md
```diff
@@ -22,3 +22,32 @@ Functions decorated with `@weave.op()` will behave normally (without code versio
 :::
 
 Ops can be [served](/guides/tools/serve) or [deployed](/guides/tools/deploy) using the Weave toolbelt.
+
+## Customize logged inputs and outputs
+
+If you want to change the data that is logged to weave without modifying the original function (e.g. to hide sensitive data), you can pass `postprocess_inputs` and `postprocess_outputs` to the op decorator.
```
jamie-rasmussen, line 28: outputs => output

docs/docs/guides/tracking/ops.md
```diff
@@ -22,3 +22,32 @@ Functions decorated with `@weave.op()` will behave normally (without code versio
...
+
+`postprocess_inputs` takes in a dict where the keys are the argument names and the values are the argument values, and returns a dict with the transformed inputs.
+
+`postprocess_outputs` takes in any value which would normally be returned by the function and returns the transformed output.
```
jamie-rasmussen, line 32: outputs => output

weave/tests/trace/test_op_decorator_behaviour.py
```diff
@@ -263,3 +259,38 @@ async def test_gotten_object_method_is_callable_with_call_func(client, weave_obj
...
+def test_postprocessing_funcs(client):
+    def postprocess_inputs(inputs: dict[str, Any]) -> dict[str, Any]:
+        d = {}
+        for k, v in inputs.items():
+            if k in {"hide_me", "and_me"}:
+                continue
+            d[k] = v
+        return d
+
+    def postprocess_output(outputs: dict[str, Any]) -> dict[str, Any]:
```
jamie-rasmussen, line 273: outputs => output

jamie-rasmussen (approved):

weave/trace/weave_client.py
andrewtruong: Maybe redacting could be part of postprocess.  We could let the user pass a list of funcs?

jamie-rasmussen: Thinking about this more, I think the cleanest would be to keep it how you have it, and then if we later discover that people want to turn off redaction we could  introduce a separate `redact` boolean keyword arg on the opt that defaults to True.

PR #2377, opened 2024-09-13
fix(weave): Show link when calling async ops

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-20883

### DIFF
--- a/weave/trace/op.py
+++ b/weave/trace/op.py
@@ -234,7 +234,6 @@ def _execute_call(
     if inspect.iscoroutinefunction(func):
 
         async def _call_async() -> Coroutine[Any, Any, Any]:
-            call_context.push_call(call)
             try:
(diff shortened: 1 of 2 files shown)


weave/trace/op.py
```diff
@@ -234,7 +234,6 @@ def handle_exception(e: Exception) -> Any:
     if inspect.iscoroutinefunction(func):
 
         async def _call_async() -> Coroutine[Any, Any, Any]:
-            call_context.push_call(call)
```
jamie-rasmussen, line 237: Is it correct to pop_call both here and in `finish_call`?

andrewtruong: oops yeah, deleted the extraneous pop

jamie-rasmussen (approved):

PR #2379, opened 2024-09-13
chore(weave): Add new `require_current_call` which raises if the current call is `None`

andrewtruong:
## Description

https://wandb.atlassian.net/browse/WB-20672

1. Adds a new `require_current_call` which raises if there is no current call

## Testing

Unit tests


weave/trace/call_context.py
```diff
@@ -54,7 +58,7 @@ def pop_call(call_id: typing.Optional[str]) -> None:
     _call_stack.set(new_stack)
 
 
-def get_current_call() -> typing.Optional["Call"]:
+def require_current_call() -> "Call":
```
jamie-rasmussen, line 61: I think I prefer `get_current_call`

andrewtruong: Would you prefer the optional to be `maybe_get_current_call` too?

jamie-rasmussen: Worth getting opinions from others, but I'd propose just have a single method - if people are using the throwing version and finding they are writing a lot of `try/catch` around it they can always make their own four line `maybe_get_current_call` helper, and I think there's value in keeping our API simpler.

andrewtruong: The implementation we have today is `maybe_get_current_call`.

I think @shawnlewis raised that he really wanted a version that would raise if the current call is None.

With an explicit raise version, user code might look like:
```py
try:
    call = client.require_current_call()
except NoCurrentCallError:
    handle(...)
do_something(...)
```

Without, you can still get away with this:
```py
if (call := client.get_current_call()) is None:
    handle(...)
do_something(...)
```

shawnlewis (approved):

PR #2380, opened 2024-09-13
feat: first class prompt object

jamie-rasmussen:
## Description

Internal Jira: 
https://wandb.atlassian.net/browse/WB-20795
https://wandb.atlassian.net/browse/WB-20796
https://wandb.atlassian.net/browse/WB-20818

Internal Notion:
https://www.notion.so/wandbai/First-Class-Prompts-de9ed92bcb9f4bfca0ff98ada9b12ad8?pvs=4

Implement prompt management and a chat view of calls.

## Testing

How was this PR tested?


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallChat.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }
+  return true;
+};
+
+// Does this call look like a chat formatted object?
+export const isCallChat = (call: CallSchema): boolean => {
+  if (!('messages' in call.rawSpan.inputs)) {
+    return false;
+  }
+  const {messages} = call.rawSpan.inputs;
```
tssweeney, line 37: we should be avoiding `rawSpan` as much as possible and instead using the `traceCall` (TraceCallSchema)

tssweeney:
```
  // See note above `RawSpanFromStreamTableEra` regarding the `rawSpan` field.
  // We should wean off of this and move the needed properties to `CallSchema`.
  rawSpan: RawSpanFromStreamTableEra;
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallChat.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const {inputs, output} = call.rawSpan;
+
+  const {messages} = inputs;
+  const inputRefs = messages
+    .map((m: {content: any}) => (isWeaveRef(m.content) ? m.content : null))
+    .filter(Boolean);
+  const {useRefsData} = useWFHooks();
+  const inputData = useRefsData(inputRefs);
+
+  const outputData = useRefData(output?._result);
```
tssweeney, line 64: This blindly assumes `output` is a ref. I think this and all other cases need to be adjusted. With Andrew's inlining stuff, it is not guaranteed to be a ref

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallChat.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const [refOutputChoice, setRefOutputChoice] = useState('');
+  const [refOutputMessage, setRefOutputMessage] = useState('');
+
+  // When the output data changes, get the choice ref
+  useEffect(() => {
+    const result = outputData.result;
+    if (!result || result._type !== 'ChatCompletion') {
```
tssweeney, line 71: why this conditional?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallChat.tsx
```diff
@@ -0,0 +1,148 @@
+/**
+ * Code to determine if a call is a chat call, and if so,
+ * load the message data and render it.
+ */
+
+import _ from 'lodash';
```
tssweeney, line 6: It seems like this component could be split up such that it accepts a pure TS type definition of an OpenAI messages format and decoupled from the concept of a call. (say `LLMChatView`)

Then we have two higher level paths:
1. Given a call, determine if it looks like a chat (optionally resolving refs while making this determination), and if so, pass the resolved pure representation to `LLMChatView`
2. Given a `Prompt`, resolve it to the pure rep, then pass to `LLMChatView`

As i read it, this couples Call schema to the chat view processing itself, which makes it harder to reuse and dependent on that structure & ref locations.

docs/docs/guides/core-types/prompts.md
```diff
@@ -0,0 +1,300 @@
+# Prompts
```
tssweeney, line 1: Absolutely 🔥 docs

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PromptPage/PromptTab.tsx
```diff
@@ -0,0 +1,58 @@
...
+export const PromptTab = ({entity, project, data}: PromptTabProps) => {
+  const {loading, prompt} = usePrompt(entity, project, data);
+  if (loading) {
+    return <LoadingDots />;
+  }
+  if (!prompt) {
+    return null;
+  }
+
+  const hasPlaceholders = prompt.placeholders.length > 0;
```
tssweeney, line 31: I am not seeing placeholders in the ui, for example here: https://beta.wandb.ai/jamie-rasmussen/prompt-demo/weave/objects/travel_agent/versions/hawXYs4DrtroeANocVIc55PYRziAJXoRjFfmXpn71fc

jamie-rasmussen: Yeah, I think I broke this when I decided it would be better for the placeholders to be recomputed than stored as part of the object. Might be for the best, the accordion material UI component I put in looked really ugly.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PromptPage/types.ts
```diff
@@ -0,0 +1,31 @@
+// TODO: lock this down
```
tssweeney, line 1: Oh i missed this as it relates to my comment on `CallChat`. This is more what i was looking for. It seems like everything in this directory (except `PromptTab`) could be in it's own dir: `LLMMessageList`

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PromptPage/types.ts
```diff
@@ -0,0 +1,31 @@
...
+// TODO: Can we tighten this up?
+export type ToolCall = Record<string, any>;
+
+export type Message = {
+  role: string;
+  content?: string | MessagePart[];
+  tool_calls?: ToolCall[];
+};
+
+export type Messages = Message[];
```
tssweeney, line 31: `Messages` should be exactly OpenAI's messages. Is that correct?

jamie-rasmussen: Not exactly the same because I wanted the component to be able to handle our placeholders. I do want to refine these types to get closer to what OpenAI actually has.

weave/flow/prompt/message.py
```diff
@@ -0,0 +1,183 @@
...
+        return bound
+
+    def add_text(self, text: str) -> "Message":
+        self.messages.append({"type": "text", "text": text})
+        return self
+
+    def add_image_url(self, url: str, detail: Optional[Detail] = None) -> "Message":
+        image_url = dict(url=url)
+        if detail:
+            image_url["detail"] = detail
```
tssweeney, line 160: how is `detail` used?

weave/flow/prompt/message.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+# TODO: Restrict more
+
+Detail = Literal["low", "high", "auto"]
+
+
+class Message:
+    content: MessageParts
+    messages: list[dict]
```
tssweeney, line 120: it is not clear to me why there is `messages` here

weave/flow/prompt/message.py
jamie-rasmussen: https://platform.openai.com/docs/guides/vision/low-or-high-fidelity-image-understanding

weave/flow/prompt/common.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Literal
+
+Templating = Literal["weave", "none"]
```
tssweeney, line 3: What are these templating formats? how is `weave` different than `fstrings?`

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PromptPage/types.ts
tssweeney: Can you add a comment on the variation points? due to the union, this appears to be a safe superset. I think the important part is that we can essentially feed openai-compatible outputs into here - which it does imply?

weave/flow/prompt/common.py
jamie-rasmussen: This is something I'm still thinking about. "none" turns off the template parsing, which I think we probably want to do for assistant responses by default.

My fuzzy idea for our parameters is that we would generate a form UI from them in the prompt playground. So compared to f-strings where the controls are for things like numeric precision it would be more like `type:slider min:0 max:100`.

weave/flow/prompt/message.py
```diff
@@ -0,0 +1,183 @@
...
+        if self.content:
+            bound["content"] = self.content.bind(params)
+        if self.messages:
+            # TODO: Support messages with placeholders
+            bound["content"] = self.messages
+        return bound
+
+    def add_text(self, text: str) -> "Message":
+        self.messages.append({"type": "text", "text": text})
+        return self
```
vanpelt, line 155: We should add `add_image` which converts it to a data uri and calls `add_image_url`.

vanpelt:
Oooo, was just messing around with f-strings.  There's no clever way I've seen capture the original `f"..."` template, however if we made our own little `f` equivalent we could capture everything we would need to 🔥 .

```python
import inspect
class f:
    def __init__(self, format_str, **kwargs):
        self.format_str = format_str
        frame = inspect.currentframe().f_back
        caller_locals = frame.f_locals
        fmt_dict = kwargs.get("defaults", {})
        fmt_dict.update(caller_locals)
        self.evaluated_str = format_str.format(**fmt_dict)

    def __str__(self):
        return self.evaluated_str

    def original(self):
        return self.format_str

name = "Alice"
tracked_string = weave.f("Hello, {name}!", name="my_prompt", defaults={"name": "Bob"})
=> 'Hello, Alice!'
```

jamie-rasmussen:
> Oooo, was just messing around with f-strings. There's no clever way I've seen capture the original `f"..."` template, however if we made our own little `f` equivalent we could capture everything we would need to 🔥 .

Someone also mentioned https://pypi.org/project/fvalues/

vanpelt:
> > Oooo, was just messing around with f-strings. There's no clever way I've seen capture the original `f"..."` template, however if we made our own little `f` equivalent we could capture everything we would need to 🔥 .
> 
> Someone also mentioned https://pypi.org/project/fvalues/

Oooo 👀 , this is very cool.  Maybe too magical:

> Calls to the constructor F() are magically detected using the executing library, and f-strings within are parsed to extract their components.

But I want to play with it now, I imagine we could just borrow some of their ideas.

PR #2382, opened 2024-09-13
chore(weave): add_costs -> more clear add_cost

jwlee64:
## Description

Make add cost take in explicit fields and not cost create input

## Testing

How was this PR tested?

### DIFF
--- a/weave/tests/trace/test_client_cost.py
+++ b/weave/tests/trace/test_client_cost.py
@@ -207,3 +207,18 @@ def test_purge_only_ids(client):
             ),
         )
     )
+
+
+def test_cost_create_invalid_costs(client):
+    try:
+        client.add_costs({"my_model": {"prompt_token_cost": "not_a_number"}})
+        assert False, "Expected ValueError"
+    except ValueError as e:
+        assert "prompt_token_cost" in str(e)
+
+    try:
+        client.add_costs({"my_model": {}})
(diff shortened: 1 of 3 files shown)


weave/tests/trace/test_client_cost.py
```diff
@@ -207,3 +207,18 @@ def test_purge_only_ids(client):
             ),
         )
     )
+
+
+def test_cost_create_invalid_costs(client):
+    try:
```
tssweeney, line 213: use `with pytest.raises(ValueError):`

weave/trace/weave_client.py
```diff
@@ -797,11 +799,18 @@ def add_costs(self, costs: Dict[str, CostCreateInput]) -> CostCreateRes:
...
 
         Args:
-            costs: Dictionary of costs to add to the project. In the form of {llm_id: cost}.
+            costs: Dictionary of costs to add to the project. In the form of `{llm_id: {prompt_token_cost, completion_token_cost}}`.
 
         """
+        parsed_costs = {}
+        for llm_id, cost_dict in costs.items():
+            try:
+                parsed_costs[llm_id] = CostCreateInput(**cost_dict)
```
tssweeney, line 808: what if the user passes a `CostCreateInput` here. this will be wrong

tssweeney (approved):

PR #2387, opened 2024-09-16
fix: Simplify the exit procedures to reduce flush errors

tssweeney:
Currently, our system has a few async mechanics to handle writing data to Weave in a non-blocking way. As currently implemented, I used both `atexit` handlers AND `__del__` overrides. This system was admittedly over-engineered and overkill. As a result, in some scenarios, we were running into error at exit time for some machines. 

This PR does the following:
1. Remove all the `__del__` handling. Upon reflection, this is not needed and makes everything more complicated. Remove a bunch of code relating to this
2. Refactor the code to just use `atexit`

The net effect of this PR is that:
1. The only classes that register exit functions are those that explicitly construct the threads (much much cleaner). This is in contrast to my prior approach which was trying to also execute cleanup at `__del__` moments. However, this is overkill.
2. As a result of the above, the complete flush events will not occur until program exit (meaning if all references to a client are deleted - invoking a GC), we will retain exit handlers to their underlying threads.

### DIFF
--- a/weave/trace/async_job_queue.py
+++ b/weave/trace/async_job_queue.py
@@ -111,21 +111,14 @@ class AsyncJobQueue:
                   If False, outstanding jobs are cancelled and the executor is shut down immediately.
         """
         with self._lock:
+            # Lock and flag avoids race conditions with multiple shutdowns
             if self._is_shutdown:
                 return
             self._is_shutdown = True
             if atexit is not None:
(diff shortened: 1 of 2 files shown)


gtarpenning (approved):
This makes sense, maybe we can add a backlog ticket to test this? w/ different python versions?

PR #2393, opened 2024-09-16
fix: Running evals should use in-memory data instead of re downloading everything

tssweeney:
This PR implements 2 fixes:
1. We ensure that we pass typechecks for various object types - there is an existing issue where:
   1. We don't do the full saving pass on Table rows
   2. We replace the in-memory data with the fetched data (both slow and insufficient for passing type checks & prop access. For example, dataclasses in this case are pure strings!

2. We assert that no queries are made. This is critical because all the data is already in memory and so any query is wasting user time.

Note: we are still blocked on the entire dataset write before we start the eval. Fixing this will come in a followup PR


weave/conftest.py
```diff
@@ -176,15 +176,38 @@ def __getattribute__(self, name):
         if callable(attr) and not name.startswith("_") and name != "flush":
 
             def wrapper(*args, **kwargs):
-                res = attr(*args, **kwargs)
                 self_super._flush()
+                res = attr(*args, **kwargs)
```
tssweeney, line 180: This was just a minor change to the test class - not specifically related to this PR, but i noticed the issue while debugging. Should flush first

weave/trace/vals.py
```diff
@@ -514,6 +514,12 @@ def make_trace_obj(
             root=root,
             parent=parent,
         )
+        # Use in memory rows! This is the case where we are making
+        # a trace object from an existing Table! If we don't do this
+        # then the WeaveTable will try to fetch all the rows from the
+        # server, throwing away the in memory rows. This is really expensive
+        # when we are doing evaluations!
+        val._rows = rows
```
tssweeney, line 522: This is the actual fix! Just a two-liner (:

weave/trace/vals.py
```diff
@@ -270,8 +300,47 @@ def __eq__(self, other: Any) -> bool:
...
+        if (
+            self.ref is None
+            or self.table_ref is None
+            or self.table_ref.row_digests is None
+            or self._prefetched_rows is None
+        ):
+            logger.error(
+                "Expected all row digests and prefetched rows to be set, falling back to remote iteration"
+            )
+            yield from self._remote_iter()
```
gtarpenning, line 323: i think this should throw instead of falling back, when could this ever occur outside of a programming error? Feels a bit too magical to call "local_iter" and get server results. 

Could also have a parent method `iter` that farms out the iter to either local or remote...

weave/trace/vals.py
```diff
@@ -505,7 +590,7 @@ def make_trace_obj(
                     "Expected Table.ref or Table.table_ref to be TableRef"
                 )
             val_ref = val_table_ref
-
+        rows = val.rows
```
gtarpenning, line 593: sort of weird

weave/tests/trace/test_evaluations.py
```diff
@@ -633,3 +637,79 @@ def function_score(scorer_res, model_output) -> dict:
...
+
+    # Before this test (and fix) we were making extra requests
+    # to reconstruct the table and objects in the evaluation.\
+    # These assertions ensure that we aren't making those extra requests.
+    # There is no reason to query the table, objects, or files
+    # as everything is in memory
+    access_log = client.server.attribute_access_log
+    assert "table_query" not in access_log
+    assert "obj_read" not in access_log
+    assert "file_content_read" not in access_log
```
gtarpenning, line 715: it could make sense so also write a test where there _are_ in the `attribute_access_log` to cover our bases.

gtarpenning (approved):
Stamped with a couple small comments

weave/trace/vals.py
gtarpenning: ah right

tssweeney: 👍

weave/trace/vals.py
tssweeney: I agree. I changed the method to `_local_iter_with_remote_fallback`. The reason i am being careful here is that we are dealing with variations in the backend and i would rather print an error than crash user code in the near term.

PR #2394, opened 2024-09-17
feat: Table Creation and Update endpoints return row digests

tssweeney:
This is a particularly interesting an nuanced change.

**Context**
As part of my other work relating to fixing an Evaluation bug and improving overall Eval write-time performance, there is a real need to return the row digests as part of the creation payload - else things will be slow.

**Solution**
Add `row_digests` and `updated_row_digests` to the server's response payload. (pretty easy)

**Complexity**
We don't have a great mechanism for client/server version checking and feature toggling at this point. So i added these fields with a (temporary) default constructor. The net effect is that:
1. New Client + Old Server: digest is empty list on client
2. Old Client + New Server: digest is happily thrown away on the client
3. New Client + New Server: digests flow through.

In the subsequent PRs, i will use these digests for performance improvements.


gtarpenning (approved):
This feels pretty straightforward, in the client we can easily condition on whether we get a digest to speed up (short circuit) further query.

tssweeney:
> This feels pretty straightforward, in the client we can easily condition on whether we get a digest to speed up (short circuit) further query.

Yes, this is exactly what i am working on in the next round of PRs. But wanted to get the server stuff in for release cut

PR #2078, opened 2024-08-06
feat(integration): Add integration with Google GenerativeAI SDK

soumik12345:
This PR adds the autopatch integration with [Google GenerativeAI SDK](https://ai.google.dev/api?lang=python).

## Tasks:

- [x] Non async, non streaming
- [x] Non async, streaming
- [x] Async, non streaming
- [x] Async, streaming
- [x] Add tests
- [ ] Add docs

### Non async, non streaming

```python
import os
import google.generativeai as genai
import weave

genai.configure(api_key=os.environ["GOOGLE_API_KEY"])
weave.init(project_name="google_ai_studio-test")
model = genai.GenerativeModel("gemini-1.5-flash")
response = model.generate_content("Write a story about an AI and magic")
```

### Non async, streaming

```python
import os
import google.generativeai as genai
import weave

genai.configure(api_key=os.environ["GOOGLE_API_KEY"])
weave.init(project_name="google_ai_studio-test")
model = genai.GenerativeModel("gemini-1.5-flash")
response = model.generate_content("Write a story about an AI and magic", stream=True)
chunks = [chunk.text for chunk in response]
```

### Async, non streaming

```python
import os
import asyncio
import google.generativeai as genai
import weave

genai.configure(api_key=os.environ["GOOGLE_API_KEY"])
weave.init(project_name="google_ai_studio-test")
model = genai.GenerativeModel("gemini-1.5-flash")

async def async_generate():
    response = await model.generate_content_async(
        "Write a story about an AI and magic"
    )
    return response

response = asyncio.run(async_generate())
```

### Async Streaming

```python
import os
import asyncio
import google.generativeai as genai
import weave

genai.configure(api_key=os.environ["GOOGLE_API_KEY"])
weave.init(project_name="google_ai_studio-test")
model = genai.GenerativeModel("gemini-1.5-flash")

async def get_response():
    async for chunk in await model.generate_content_async(
        "Write a story about an AI and magic", stream=True
    ):
        if chunk.text:
            print(chunk.text)
        print("_" * 80)


response = asyncio.run(get_response())
```

PR #2399, opened 2024-09-17
feat(weave): Add integration for Google Genai SDK

soumik12345:
Continuing from https://github.com/wandb/weave/pull/2078

This PR adds the autopatch integration with [Google GenerativeAI SDK](https://ai.google.dev/api?lang=python).

## Tasks:

- [x] Non async, non streaming
- [x] Non async, streaming
- [x] Async, non streaming
- [x] Async, streaming
- [x] Add tests
- [x] Add docs

### Non async, non streaming

```python
import os
import google.generativeai as genai
import weave

genai.configure(api_key=os.environ["GOOGLE_API_KEY"])
weave.init(project_name="google_ai_studio-test")
model = genai.GenerativeModel("gemini-1.5-flash")
response = model.generate_content("Write a story about an AI and magic")
```

### Non async, streaming

```python
import os
import google.generativeai as genai
import weave

genai.configure(api_key=os.environ["GOOGLE_API_KEY"])
weave.init(project_name="google_ai_studio-test")
model = genai.GenerativeModel("gemini-1.5-flash")
response = model.generate_content("Write a story about an AI and magic", stream=True)
chunks = [chunk.text for chunk in response]
```

### Async, non streaming

```python
import os
import asyncio
import google.generativeai as genai
import weave

genai.configure(api_key=os.environ["GOOGLE_API_KEY"])
weave.init(project_name="google_ai_studio-test")
model = genai.GenerativeModel("gemini-1.5-flash")

async def async_generate():
    response = await model.generate_content_async(
        "Write a story about an AI and magic"
    )
    return response

response = asyncio.run(async_generate())
```

### Async Streaming

```python
import os
import asyncio
import google.generativeai as genai
import weave

genai.configure(api_key=os.environ["GOOGLE_API_KEY"])
weave.init(project_name="google_ai_studio-test")
model = genai.GenerativeModel("gemini-1.5-flash")

async def get_response():
    async for chunk in await model.generate_content_async(
        "Write a story about an AI and magic", stream=True
    ):
        if chunk.text:
            print(chunk.text)
        print("_" * 80)


response = asyncio.run(get_response())
```


requirements.test.txt
```diff
@@ -24,4 +24,4 @@ semver==2.13.0             # Used for switching logic based on package versions
 flask
 uvicorn>=0.27.0
 
-pillow
\ No newline at end of file
+pillow
```
andrewtruong, line 27: is this just for your test?  If so, put it in the tox file

tox.ini
```diff
@@ -47,6 +48,7 @@ commands =
     cerebras: pytest {posargs} integrations/cerebras/
     cohere: pytest {posargs} integrations/cohere/
     dspy: pytest {posargs} integrations/dspy/
+    google-genai: pytest integrations/integrations/
```
andrewtruong, line 51: this should be `integrations/google_ai_studio`

.github/workflows/test.yaml
```diff
@@ -169,6 +169,7 @@ jobs:
             'cohere',
             'dspy',
             'groq',
+            'google-genai',
```
andrewtruong, line 172: nit: these should also be `google_ai_studio`

tox.ini
andrewtruong, line 1: you need to add the `google_ai_studio` shard here: https://github.com/wandb/weave/blob/91955c39f6f3a48ad3bbf4908b009662a6c5ba5f/tox.ini#L17

tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -0,0 +1,107 @@
...
+    def children_of_parent_id(id: Optional[str]) -> list[tsi.CallSchema]:
+        return [call for call in calls if call.parent_id == id]
+
+    children = children_of_parent_id(parent_id)
+    res = []
+    for child in children:
+        res.append((child, depth))
+        res.extend(flatten_calls(calls, child.id, depth + 1))
+
+    return res
```
andrewtruong, line 23: You shouldn't need these lower-level interfaces for tests.  See other integration tests for examples:
https://github.com/wandb/weave/blob/master/tests/integrations/anthropic/anthropic_test.py

weave/trace/op_extensions/accumulator.py
```diff
@@ -82,9 +83,12 @@ def __iter__(self) -> "_IteratorWrapper":
...
         if not hasattr(self._iterator_or_ctx_manager, "__next__"):
-            raise TypeError(
-                f"Cannot call next on an iterator of type {type(self._iterator_or_ctx_manager)}"
-            )
+            try:
+                self._iterator_or_ctx_manager = iter(self._iterator_or_ctx_manager)  # type: ignore
+            except TypeError:
+                raise TypeError(
+                    f"Cannot call next on an iterator of type {type(self._iterator_or_ctx_manager)}"
+                )
```
andrewtruong, line 91: Why do you need this?

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
```diff
@@ -0,0 +1,88 @@
...
+    acc: "GenerateContentResponse", value: "GenerateContentResponse"
+) -> "GenerateContentResponse":
+    if acc is None:
+        acc = value
+    if not acc._done:
+        return value
+    for candidate_idx in range(len(value.candidates)):
+        value_candidate = value.candidates[candidate_idx]
+        for part_idx in range(len(value_candidate.content.parts)):
+            value_part = value_candidate.content.parts[part_idx]
```
andrewtruong, line 23: these look like they should be `enumerate`

tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -0,0 +1,107 @@
...
+def test_content_generation_async(client: WeaveClient) -> None:
+    import google.generativeai as genai
+
+    genai.configure(api_key=os.environ.get("GOOGLE_API_KEY", "DUMMY_API_KEY"))
+    model = genai.GenerativeModel("gemini-1.5-flash")
+
+    async def async_generate() -> Any:
+        return await model.generate_content_async("Write a story about an AI and magic")
+
+    asyncio.run(async_generate())
```
andrewtruong, line 92: This looks like it should be in an async test

```
@pytest.mark.asyncio
async def test_content_generation_async(client): ...
```

tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+from weave.integrations.integration_utilities import op_name_from_ref
+from weave.trace.weave_client import WeaveClient
+
+
+@pytest.mark.skip_clickhouse_client
+def test_content_generation(client: WeaveClient) -> None:
+    import google.generativeai as genai
+
+    genai.configure(api_key=os.environ.get("GOOGLE_GENAI_KEY"))
```
andrewtruong, line 13: Do you need to explicitly pass this?

tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+    output = call.output
+    assert output is not None
+
+
+@pytest.mark.skip_clickhouse_client
+def test_content_generation_stream(client: WeaveClient) -> None:
+    import google.generativeai as genai
+
+    genai.configure(api_key=os.environ.get("GOOGLE_GENAI_KEY", "DUMMY_API_KEY"))
```
andrewtruong, line 34: same here

tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -0,0 +1,75 @@
...
+        op_name_from_ref(call.op_name)
+        == "google.generativeai.GenerativeModel.generate_content"
+    )
+    output = call.output
+    assert output is not None
+
+
+@pytest.mark.asyncio
+@pytest.mark.skip_clickhouse_client
+async def test_content_generation_async(client: WeaveClient) -> None:
```
andrewtruong, line 57: nit: You don't need types on the test funcs.  I don't think they're helpful

tests/integrations/google_ai_studio/google_ai_studio_test.py
soumik12345: Yes, that's what the doc says.

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def gemini_accumulator(
+    acc: Union["GenerateContentResponse", None], value: "GenerateContentResponse"
+) -> "GenerateContentResponse":
+    if acc is None:
+        acc = value
+    if not acc._done:
+        return value
+    if acc != value:
```
andrewtruong, line 20: Is this `if` correct?  What if the acc is done and it's the same as the value (e.g. you return just 1 token)

weave/trace/op_extensions/accumulator.py
```diff
@@ -82,11 +83,10 @@ def __iter__(self) -> "_IteratorWrapper":
 
     def __next__(self) -> Generator[None, None, V]:
         if not hasattr(self._iterator_or_ctx_manager, "__next__"):
-            raise TypeError(
-                f"Cannot call next on an iterator of type {type(self._iterator_or_ctx_manager)}"
-            )
+            self._iterator_or_ctx_manager = iter(self._iterator_or_ctx_manager)  # type: ignore
         try:
             value = next(self._iterator_or_ctx_manager)  # type: ignore
+            # print(f"{value.candidates[0].content.parts[0].text=}")
```
andrewtruong, line 89: remove

weave/trace/op_extensions/accumulator.py
```diff
@@ -82,11 +83,10 @@ def __iter__(self) -> "_IteratorWrapper":
 
     def __next__(self) -> Generator[None, None, V]:
         if not hasattr(self._iterator_or_ctx_manager, "__next__"):
-            raise TypeError(
-                f"Cannot call next on an iterator of type {type(self._iterator_or_ctx_manager)}"
-            )
+            self._iterator_or_ctx_manager = iter(self._iterator_or_ctx_manager)  # type: ignore
```
andrewtruong, line 86: Why are you putting a type ignore here?

Is this safe?  I think it assumes that whatever is yielded by the contextmanager is iter-able

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
soumik12345: If i don't add this `if`, the first token is being accumulated twice, both in the weave trace as well as in the output of the function. I agree that this is not the best way to tackle this issue and I'm open to suggestions to improve the logic here.

weave/trace/op_extensions/accumulator.py
soumik12345: If I don't put type ignore here, its failing the mypy checks.

andrewtruong: The point of mypy is to check if it's type-safe, and the ignore here suggests that it's not.  Can you please remove the ignore and fix?

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
```diff
@@ -0,0 +1,87 @@
+import importlib
+from functools import wraps
+from typing import TYPE_CHECKING, Any, Callable, Union
+
+if TYPE_CHECKING:
+    from google.generativeai.types.generation_types import GenerateContentResponse
```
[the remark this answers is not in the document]
soumik12345: Okay, I'm putting this back; otherwise, CI is failing.

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
andrewtruong: Are you saying that adding the accumulator modifies the output of the function?  It shouldn't do that -- we should never modify the user's output

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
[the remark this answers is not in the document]
andrewtruong: why does it fail?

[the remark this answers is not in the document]
andrewtruong: oh I see, it's a flaw in how we're autopatching -- ok this is fine for now

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
soumik12345: Yeah, I placed the check to prevent modification of output.

andrewtruong: sorry, I don't understand.

Here's what I think I'm reading:
1. If the acc is fresh, use the first value of the iterator
2. While the acc is not finished, use the next values
3. If acc != value, then aggregate the candidate completions

So open questions are:
1. How does the above modify the user's output?  It shouldn't.  We should only be "listening" to the stream
2. Why do we need the acc != value check?  I guess there's an interaction with `_done`, but I'm not sure how

soumik12345: Done.

weave/trace/op_extensions/accumulator.py
tcapelle: We added this back then when the class was extended to support context manager for Anthropic, I pushed a hacky fix

andrewtruong (approved):

PR #2403, opened 2024-09-17
chore(weave): Split out Trace and Query tests; move trace tests into dedicated tox envs across py39-312

andrewtruong:
## Description

https://wandb.atlassian.net/browse/WB-20964

This PR splits out tests into dedicated tox envs for {trace, trace_server, and first party integrations} across the python versions we support.

1. Py312 tests are legitimately failing, but are commented out to be addressed in a follow-up PR because those changes would otherwise be noisy here.  The issues seem to stem from a behaviour change in `runtime_checkable` which breaks `isinstance(obj, Op)` checks:
https://docs.python.org/3/library/typing.html#typing.runtime_checkable

2. This PR requires the companion:
https://github.com/wandb/core/pull/23983

## Testing

this is testing :)


.github/workflows/test.yaml
```diff
@@ -73,6 +78,88 @@ jobs:
           yarn prettier
           yarn run tsc
 
+  test2:
```
tssweeney, line 81: test2?

.github/workflows/test.yaml
```diff
@@ -73,6 +78,88 @@ jobs:
...
+          WEAVE_SENTRY_ENV: ci
+          CI: 1
+          WB_SERVER_HOST: http://wandbservice
+          WF_CLICKHOUSE_HOST: weave_clickhouse
+          WEAVE_SERVER_DISABLE_ECOSYSTEM: 1
+        run: |
+          tox -e ${{ matrix.tox-shard }}-py${{ matrix.python-version-major }}${{ matrix.python-version-minor }} -- \
+            -m "weave_client and not skip_clickhouse_client" \
+            --weave-server=clickhouse
+      - name: Run tox (Clickhouse Only)
```
tssweeney, line 153: "clickhouse only" is the one above

.github/workflows/test.yaml
```diff
@@ -133,7 +220,7 @@ jobs:
           --timeout=90 \
           --ddtrace \
           --durations=5 \
-          ./integrations ./legacy ./trace_server ./trace ./tests
+          ./legacy
```
tssweeney, line 223: This should be removed. Anything depending on the `client` fixture in legacy is misplaced. I just verified this by looking at the 3 cases

.github/workflows/test.yaml
```diff
@@ -153,81 +240,4 @@ jobs:
...
-          SPLIT_INDEX: ${{ strategy.job-index }}
-          # - uses: actions/upload-artifact@v3
-      - name: Upload logs
-        uses: actions/upload-artifact@v3
-        if: failure()
-        with:
-          name: weave-server-logs
-          path: /tmp/weave/log
-          retention-days: 3
+          ./legacy
```
tssweeney, line 243: Rename to `Query Service Tests`

requirements.legacy.txt
```diff
@@ -0,0 +1,61 @@
+# These are the base Weave requirements, enough for weave tracking and evaluation
```
tssweeney, line 1: This can be severely trimmed down - much of this is trace specific. fine to be a followup

requirements.txt
```diff
@@ -1,5 +1,10 @@
 # These are the base Weave requirements, enough for weave tracking and evaluation
 # to work.
+# -r requirements.legacy.txt
+# COPIED FROM LEGACY:
```
tssweeney, line 4: why not put this comment next to where you added these reqs?

.github/workflows/test.yaml
```diff
@@ -73,6 +78,88 @@ jobs:
...
+            'litellm',
+            'llamaindex',
+            'mistral0',
+            'mistral1',
+            'openai',
+          ]
+      fail-fast: false
+    services:
+      wandbservice:
+        image: us-central1-docker.pkg.dev/wandb-production/images/local-testcontainer:master
```
tssweeney, line 113: why this base?

weave/docker/Dockerfile.python-client.test
```diff
@@ -10,6 +10,6 @@ RUN python3 -m venv venv
 RUN --mount=type=cache,target=/root/.cache /bin/bash -c \
     "source venv/bin/activate && \
     pip install --upgrade pip && \
-    pip install -r requirements.test.txt"
+    pip install -r requirements.legacy.test.txt"
```
tssweeney, line 13: This doesn't seem right - the legacy tests should be using `weave/legacy/Dockerfile.test`. This Dockerfile would be used for your tox tests (or deleted if we want to use wandbs)

weave/legacy/tests/test_wb_data_types.py
```diff
@@ -535,7 +535,7 @@ def test_annotated_images_in_tables(fake_wandb):
 
 def test_annotated_legacy_images_in_tables(fake_wandb):
     # Mocking this property makes the payload look like the legacy version.
-    from wandb.data_types import _ImageFileType
+    from wandb.sdk.data_types.image import _ImageFileType
```
tssweeney, line 538: this is fixed in master

.github/workflows/test.yaml
andrewtruong: That's the container we use to test things that need gorilla.

Do you think there's a better container to use?

.github/workflows/test.yaml
andrewtruong: ok, I found 4 cases and commented them out for now.

requirements.legacy.txt
andrewtruong: Is the minimal set described here?  If yes we can replace:
https://github.com/wandb/core/blob/master/services/weave-python/requirements.txt

requirements.txt
andrewtruong: This is a temporary "inlining" of `-r requirements.legacy.txt` because the `-r ...` syntax is not valid for pip installs using pyproject.toml.  I think this form is more obvious that it's a copy-paste vs. annotating individual reqs

In future, we'll remove this altogether and just use pyproject.toml

weave/docker/Dockerfile.python-client.test
andrewtruong: I'm not very familiar here, but I'm following this part of the tests:
https://github.com/wandb/weave/blame/andrew/tox3/.github/workflows/test.yaml#L37-L40

Which seems to be:
1. A unit test image which is used for the legacy tests below
2. An integration test image which is used for the weave tests in core

.github/workflows/test.yaml
andrewtruong: I'll bring `test_weave_finish_unsets_client` over.

I don't think the Monitoring tests are valuable anymore.  We don't have that path, and the OpenAI and basic logging stuff has its own set of tests already.

.github/workflows/test.yaml
```diff
@@ -12,7 +12,30 @@ on:
   push:
 
 jobs:
-  build-container:
+  # ==== Query Service Jobs ====
+  weavejs-lint-compile:
```
tssweeney, line 16: technically `weave-js` is a UI package, not Query Service

.github/workflows/test.yaml
```diff
@@ -12,7 +12,30 @@ on:
...
+          cd weave-js
+          yarn install --frozen-lockfile
+          yarn generate
+          yarn eslint --max-warnings=0
+          yarn tslint
+          yarn prettier
+          yarn run tsc
+
+  build-container-query-service:
     name: Build test container
```
tssweeney, line 39: i'd rename this too

.github/workflows/test.yaml
```diff
@@ -34,57 +57,20 @@ jobs:
...
-          set -e
-          cd weave-js
-          yarn install --frozen-lockfile
-          yarn generate
-          yarn eslint --max-warnings=0
-          yarn tslint
-          yarn prettier
-          yarn run tsc
+      - name: Build legacy (query sevice) unit test image
+        run: python3 weave/docker/docker_build.py build_deps weave-test-python-client builder . weave/legacy/Dockerfile.client.test
```
tssweeney, line 61: again, this is not really about testing the `client` as much as it testing the query service. name seems misleading

weave/legacy/tests/test_weave_api.py
```diff
@@ -34,23 +34,23 @@ def test_create_list_rename_delete():
...
-
-    foo()
-    assert len(list(weave_client.get_calls())) == 1
-
-    weave.finish()
-
-    foo()
-    assert len(list(weave_client.get_calls())) == 1
-    assert weave.trace.weave_init._current_inited_client is None
+# def test_weave_finish_unsets_client(client):
```
tssweeney, line 37: this is a proper `trace` test

tssweeney: oh, looks like you did! maybe delete then?

andrewtruong: done

tssweeney (approved):

PR #2409, opened 2024-09-17
chore: add invalid field 403 error

gtarpenning:
## Description

- Fixes [WB-20985](https://wandb.atlassian.net/browse/WB-20985)

Raises a more specific error that can be handled as 403

Depends on: https://github.com/wandb/core/pull/23966

## Testing

How was this PR tested?

[WB-20985]: https://wandb.atlassian.net/browse/WB-20985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/calls_query_builder.py
+++ b/weave/trace_server/calls_query_builder.py
@@ -34,6 +34,7 @@ import sqlparse
 from pydantic import BaseModel, Field
 
 from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.errors import InvalidFieldError
 from weave.trace_server.interface import query as tsi_query
 from weave.trace_server.orm import (
     ParamBuilder,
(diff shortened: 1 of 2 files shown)


tssweeney (approved):

gtarpenning:
@tssweeney  We might need to revert this FYI. Not urgent, but I can't get the core-side PR to handle this exception properly. AFAICT the fastAPI middlewear can't catch errors that occur in a generated/stream response... Because we have already set headers? I get this gigantic and totally opaque TaskGroup error....

Example...
[error.txt](https://github.com/user-attachments/files/17062715/error.txt)

gtarpenning:
We could do something like this: https://github.com/fastapi/fastapi/discussions/10138#discussioncomment-8216436

PR #2410, opened 2024-09-17
feat: Improve serialization & deserialization of custom objects using in mem cache

tssweeney:
Currently our serialization procedures (to/from json) also handle custom objects. This means: possible file I/O AND network transmission.

Horribly, we don't cache any of this, so every time the same Image is passed through the call stack it gets re-serialzied every time! This means that our evaluations with Images likely serialize and de-serialize the same image multiple times. With this change, we create an in-memory cache of such serialization and deserialization results. Everything uses WeakRefs so we don't have an ever-expanding memory footprint.

Attention Reviewers:
1. The entry-point for the change is refactoring the serialize.py to/from json methods to have sub-methods for custom weave objects. These sub methods are literally the same code as before, with a little code at the top and bottom for cache lookups / storage.
2. Most of the logic resides in the 2 new classes `CustomWeaveTypeSerializationCache` and `WeakKeyDictionarySupportingNonHashableKeys` which have ample comments and unit tests.
3. The entire end to end system is also tested to ensure we have minimal network calls.

### DIFF
--- a/weave/trace/serialize.py
+++ b/weave/trace/serialize.py
@@ -1,5 +1,7 @@
+import json
 import typing
-from typing import Any
+import weakref
+from typing import Any, Optional
 
 from weave.trace import custom_objs
 from weave.trace.object_record import ObjectRecord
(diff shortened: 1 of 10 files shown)


weave/trace/serialize.py
```diff
@@ -110,3 +99,65 @@ def from_json(obj: Any, project_id: str, server: TraceServerInterface) -> Any:
         return parse_uri(obj)
 
     return obj
+
+
+# _[to/from]_json_custom_weave_type are used to serialize and deserialize
```
tssweeney, line 104: Reviewers start here! I pulled out the to/from json stuff into common functions and short circuit using the cache. All the code changes are in the new data structures

tssweeney:
This is not going to work well due to mutations.

PR #2412, opened 2024-09-17
perf: push down conditions in objects query

gtarpenning:
## Description

Push down conditions into the lowest subquery for performance. Also clean up unused query vars. 

This will take a bite out of https://wandb.atlassian.net/browse/WB-21011

## Testing

In clickhouse testing, this performance improvement leads to ~ 2 orders of magnitude improvement. 

This should affect both `objs/query` and `obj/read` by changing the underling object select query, but most of the improvement will be to the `obj/read` path

### This will not change the performance of `objs/query` when no `object_id` is specified, meaning that it will actually not be that much of an improvement for the first loading of `ops`/`objects`... More work to do there. 

Example: 

Existing query:
Screenshot 2024-09-18 at 12 42 23 PM

New query:
Screenshot 2024-09-18 at 12 39 40 PM


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -537,12 +537,11 @@ def ops_query(self, req: tsi.OpQueryReq) -> tsi.OpQueryRes:
                 conds.append("object_id IN {op_names: Array(String)}")
                 parameters["op_names"] = req.filter.op_names
 
-            if req.filter.latest_only:
-                conds.append("is_latest = 1")
-
         ch_objs = self._select_objs_query(
             req.project_id,
             conditions=conds,
+            parameters=parameters,
```
gtarpenning, line 543: why was this never passed in??

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1430,11 +1433,12 @@ def _select_objs_query(
                         ) AS rn
                     FROM object_versions
                     WHERE project_id = {{project_id: String}}
+                       AND {conditions_part}
                 )
                 WHERE rn = 1
             )
-            WHERE project_id = {{project_id: String}} AND
```
gtarpenning, line 1436: why do we need this?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1371,11 +1371,13 @@ def _select_objs_query(
         conditions: Optional[list[str]] = None,
         limit: Optional[int] = None,
         parameters: Optional[Dict[str, Any]] = None,
+        is_latest: bool = False,
```
gtarpenning, line 1374: this is the only thing that can't actually go in the heavy filter, because it is really the reason for the subquery

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: ops_query is not used :( so it likely has a few bugs

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: abundance of caution - probably left over. can never let things get through that are not the right project for security reasons

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1430,11 +1437,11 @@ def _select_objs_query(
                         ) AS rn
                     FROM object_versions
                     WHERE project_id = {{project_id: String}}
+                       AND {conditions_part}
```
tssweeney, line 1440: this should go next to `WHERE rn = 1` i think. else your versions will be off if you include digest in this condition set

tssweeney (approved):

PR #2418, opened 2024-09-18
perf: use objs_query param `metadata_only` to get objects without their values

gtarpenning:
## Description

https://wandb.atlassian.net/browse/WB-21042

Pass in `metadata_only` to **not** load the val_dump. has a pretty mega impact on the `objs/query` that we use in the `ops` and `objects` page. That query currently barely completes in the worst projects, example below:

Prod:
minutes (and crashes) on large objects

This change:
Screenshot 2024-09-18 at 1 30 02 PM

## Testing

How was this PR tested?


weave/trace_server/trace_server_interface.py
```diff
@@ -377,6 +377,9 @@ class ObjectVersionFilter(BaseModel):
 class ObjQueryReq(BaseModel):
     project_id: str
     filter: Optional[ObjectVersionFilter] = None
+    # Defaults to true, pass in False for a more performant
+    # query that does not return the object value
+    include_val_dump: bool = True
```
tssweeney, line 382: "dump" is an internal detail - users think of this as `val`.

I think the `include_*` pattern is used to include something that does not normally come for free. I think instead we might want to use the `column` pattern from calls to allow a sub selection of fields.

tssweeney: Also, tests (:

gtarpenning: yes yes tests, first making sure you think the approach is sound. ill play with the request

weave/trace_server/sqlite_trace_server.py
```diff
@@ -1079,14 +1080,29 @@ def _select_objs_query(
...
         offset: Optional[int] = None,
         sort_by: Optional[list[tsi.SortBy]] = None,
     ) -> list[tsi.ObjSchema]:
         conn, cursor = get_conn_cursor(self.db_path)
         pred = " AND ".join(conditions or ["1 = 1"])
-        query = f"""SELECT * FROM objects
-                    WHERE deleted_at IS NULL AND project_id = ? AND {pred}"""
+        val_dump_part = "'{}' as val_dump" if metadata_only else "val_dump"
+        query = f"""
+            SELECT
```
tssweeney, line 1092: thank you! * is bad

tssweeney (approved):

PR #2419, opened 2024-09-18
chore: Refactor & Comment Saving Methods

tssweeney:
The dataflow for saving objects has been notoriously complex and undocumented. This PR changes that by:
1. Co-locating common functions
2. Commenting profusely
3. Re-arranging some switch statements to be grouped more logically.

There should be no logical changes as a result of this PR

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -26,7 +26,6 @@ from weave.trace.op import Op, maybe_unbind_method
 from weave.trace.op import op as op_deco
 from weave.trace.refs import CallRef, ObjectRef, OpRef, Ref, TableRef, parse_op_uri
 from weave.trace.serialize import from_json, isinstance_namedtuple, to_json
-from weave.trace.serializer import get_serializer_for_obj
 from weave.trace.table import Table
 from weave.trace.util import deprecated
 from weave.trace.vals import WeaveObject, WeaveTable, make_trace_obj
@@ -440,6 +439,28 @@ class WeaveClient:
 
     @trace_sentry.global_trace_sentry.watch()
     def save(self, val: Any, name: str, branch: str = "latest") -> Any:
+        """Do not call directly, use weave.publish() instead.
+
+        Args:
+            val: The object to save.
+            name: The name to save the object under.
+            branch: The branch to save the object under. Defaults to "latest".
+
+        Returns:
+            A deserialized version of the saved object.
+        """
+        # Adding a second comment line for developers that is not a docstring:
+        # Save an object to the weave server and return a deserialized version of it.
+
+        # Note: This is sort of a weird method becuase:
+        # 1. It returns a deserialized version of the object (which will often not pass type-checks)
+        # 2. It is slow (because it re-downloads the object from the weave server)
+        # 3. It explicitly filters out non ObjectRefs, which seems like a useless constraint.
+        #
+        # Because of these reasons, `weave.publish()` directly calls `_save_object()`
+        # and then returns the raw ObjectRef. I (Tim) think we should consider refactoring
+        # `save()`. I am not sure when as an end user you would ever want to use this method.
+
         ref = self._save_object(val, name, branch)
         if not isinstance(ref, ObjectRef):
             raise ValueError(f"Expected ObjectRef, got {ref}")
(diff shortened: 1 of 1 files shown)


weave/trace/weave_client.py
```diff
@@ -961,32 +982,136 @@ def query_costs(
...
+
+        Args:
+            val: The object to save.
+            name: The name to save the object under.
+            branch: The branch to save the object under. Defaults to "latest".
+
+        Returns:
+            An ObjectRef to the saved object.
+        """
+        if isinstance(val, WeaveTable):
```
tssweeney, line 1001: I added these no-op conditions for future reference, but did not want to introduce logical changes in this PR

weave/trace/weave_client.py
```diff
@@ -1022,61 +1150,31 @@ def _save_object_basic(
             ref = OpRef(self.entity, self.project, name, response.digest)
         else:
             ref = ObjectRef(self.entity, self.project, name, response.digest)
-        # TODO: Try to put a ref onto val? Or should user code use a style like
-        # save instead?
+
+        # Attach the ref to the object
```
tssweeney, line 1154: This already happened in all cases where we called this function. Adding here is the correct thing to do and simplified other code

weave/trace/weave_client.py
```diff
@@ -1121,27 +1227,6 @@ def _objects(self, filter: Optional[ObjectVersionFilter] = None) -> list[ObjSche
...
-    def _save_op(self, op: Op, name: Optional[str] = None) -> Ref:
-        if op.ref is not None:
-            return op.ref
-
-        if name is None:
-            name = op.name
-
-        return self._save_and_attach_ref(op, name)
-
-    def _save_and_attach_ref(self, op: Any, name: Optional[str] = None) -> Ref:
```
tssweeney, line 1133: Removed completely!

weave/trace/weave_client.py
```diff
@@ -99,6 +98,19 @@ def remove_ref(obj: Any) -> None:
             obj.ref = None
 
 
+def set_ref(obj: Any, ref: Optional[Ref]) -> None:
```
gtarpenning, line 101:
```suggestion
def set_ref(obj: Any, ref: Optional[Ref]) -> None:
   """Attempts to set the ref for all possible object types."""
```

weave/trace/weave_client.py
```diff
@@ -440,6 +452,28 @@ def __init__(
 
     @trace_sentry.global_trace_sentry.watch()
     def save(self, val: Any, name: str, branch: str = "latest") -> Any:
+        """Do not call directly, use weave.publish() instead.
```
gtarpenning, line 455: can we mark this deprecated? or `def save` -> `def _save`? Docstring is good but still will probably be used directly.

weave/trace/weave_client.py
```diff
@@ -961,41 +995,159 @@ def query_costs(
...
+            - If the ref is to the current project, do nothing.
+            - If the ref is to a different project, remove it. (to avoid cross-project references)
+        2. If the `obj` value can be "reffed" (according to one of the above cases), invoke
+            the appropriate "save" function for that type of object, and attach the ref result to `obj`.
+            - `_save_object_basic` (for `weave.flow.obj.Object` instances)
+            - `_save_op` (for `weave.trace.op.Op` instances)
+            - `_save_table` (for `weave.trace.Table` and `weave.trace.vals.WeaveTable` instances)
+        3. Otherwise, traverse all values within `obj` recursively, applying the above logic to each value.
+        Important notes to developers: This method does not return anything - it _mutates_ the
+        values that it traverses (specifically, it attaches `ref` values to them)
```
gtarpenning, line 1059: this attaches refs, and also importantly fires off create queries to the server? Might want to include something about how this function actually incurs server latency

weave/trace/weave_client.py
gtarpenning: nice

gtarpenning (approved):
clean

weave/trace/weave_client.py
tssweeney: Ack - added in next commit

weave/trace/weave_client.py
tssweeney: It is not "deprecated" per-se. But since it is in the open, we can't really change it at this point. I think the right thing is to fix the issues in the list (:

weave/trace/weave_client.py
tssweeney: added in next commit

PR #2428, opened 2024-09-19
feat(weave): Add setting to control code capture

andrewtruong:
## Description

https://wandb.atlassian.net/browse/WB-20322

Adds a setting to control code capture, either in code or by env var with `WEAVE_CAPTURE_CODE`

## Testing

How was this PR tested?

### DIFF
--- a/weave/tests/trace/test_trace_settings.py
+++ b/weave/tests/trace/test_trace_settings.py
@@ -106,4 +106,13 @@ def test_should_capture_code_setting(client):
     assert "Code-capture disabled" not in inspect.getsource(func2)
 
 
-def test_should_capture_code_env(client): ...
+def test_should_capture_code_env(client):
+    os.environ["WEAVE_CAPTURE_CODE"] = "false"
+    ref = weave.publish(func)
+    func2 = ref.get()
+    assert "Code-capture disabled" in inspect.getsource(func2)
+
+    os.environ["WEAVE_CAPTURE_CODE"] = "true"
+    ref = weave.publish(func)
+    func2 = ref.get()
+    assert "Code-capture disabled" not in inspect.getsource(func2)
--- a/weave/tests/trace/test_trace_settings.py
+++ b/weave/tests/trace/test_trace_settings.py
@@ -105,7 +105,7 @@ def test_should_capture_code_setting(client):
     ref = weave.publish(test_func)
     test_func2 = ref.get()
     code2 = test_func2.art.path_contents["obj.py"].decode()
-    assert "Code-capture disabled" in code2
+    assert "Code-capture was disabled" in code2
 
     parse_and_apply_settings(UserSettings(capture_code=True))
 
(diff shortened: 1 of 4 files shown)


weave/tests/trace/test_trace_settings.py
```diff
@@ -90,3 +93,53 @@ def test_print_call_link_env(client):
...
+
+    ref = weave.publish(test_func)
+    test_func2 = ref.get()
+    code2 = test_func2.art.path_contents["obj.py"].decode()
+    assert "Code-capture disabled" in code2
+
+    parse_and_apply_settings(UserSettings(capture_code=True))
+
+    # TODO: Not safe to change capture_code setting mid-script because the op's ref
+    # does not know about the setting change.
```
andrewtruong, line 113: is there a simple way around this?

Currently you have this problem:
```py
def func(): ...

os.environ["WEAVE_CODE_CAPTURE"] = False
ref = weave.publish(func)

os.environ["WEAVE_CODE_CAPTURE"] = True
ref2 = weave.publish(func)

assert ref == ref2  # True
```

Which happens because the code function has not changed

andrewtruong: I think this is OK behaviour as long as we document it.  I think it would be rare to flip between capturing and not capturing code mid-script

weave/trace/weave_client.py
```diff
@@ -496,6 +496,7 @@ def get(self, ref: ObjectRef) -> Any:
                     digest=ref.digest,
                 )
             )
+            print(f"{read_res=}")
```
gtarpenning, line 499: dev trax

weave/trace/op_type.py
```diff
@@ -276,7 +288,6 @@ def get_source_or_fallback(fn: typing.Callable, *, warnings: list[str]) -> str:
         sig_str = reconstruct_signature(fn)
     except Exception as e:
         warnings.append(f"Failed to reconstruct signature: {e}")
-        sig_str = "(*args, **kwargs)"
```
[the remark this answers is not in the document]
andrewtruong: oops, we do

gtarpenning (approved):
lgtm but I only have minimal context

weave/trace/op_type.py
```diff
@@ -267,6 +269,16 @@ def get_source_or_fallback(fn: typing.Callable, *, warnings: list[str]) -> str:
     if isinstance(fn, Op):
         fn = fn.resolve_fn
 
+    if not settings.should_capture_code():
+        # This digest is kept for op versioning purposes
+        digest = str_digest(inspect.getsource(fn))
+        return textwrap.dedent(
+            f"""
+            def func(*args, **kwargs):
+                ...  # Code-capture disabled for this op (digest: {digest})
```
tssweeney, line 278: nit: "Code-capture was disabled while saving this op" (code capture is not a setting on the op itself)

tssweeney (approved):

PR #2429, opened 2024-09-19
chore(weave): auto detect right-to-left text in tables

bcsherma:
## JIRA Issue(s)

- https://wandb.atlassian.net/browse/WB-20787

## Description

This PR updates the plain text display for table cells to format arabic/hebrew text right-to-left (RTL). A regex specifying unicode ranges for arabic, hebrew, and related scripts is tested against the text. If matches are found, RTL formatting is applied. Specifically, we set `direction: rtl` and `text-align: right` on the div containing the text.

Before:

Screenshot 2024-09-19 at 4 51 17 PM

After:

Screenshot 2024-09-19 at 4 50 25 PM

## Questions

This PR adds a regex check for these character ranges for all table cells rendering plain text. The code already parses the text to detect if its a url and applies special formatting in this case, so this pr does not worsen the asymptotic complexity of the related code.

### DIFF
--- a/weave-js/src/components/Panel2/PanelString.tsx
+++ b/weave-js/src/components/Panel2/PanelString.tsx
@@ -14,6 +14,9 @@ import * as S from './PanelString.styles';
 import {TooltipTrigger} from './Tooltip';
 import {WeaveFormatContext} from './WeaveFormatContext';
 
+const rtlChars =
+  /[\u0590-\u05FF\u0600-\u06FF\u0750-\u077F\u08A0-\u08FF\uFB50-\uFDFF\uFE70-\uFEFF]/;
+
 const inputType = {
   type: 'union' as const,
   members: [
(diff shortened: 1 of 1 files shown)


weave-js/src/components/Panel2/PanelString.tsx
```diff
@@ -53,6 +53,27 @@ function isURL(text: string): boolean {
...
+ * - U+08A0 to U+08FF
+ * - U+FB50 to U+FDFF
+ * - U+FE70 to U+FEFF
+ * Matching arabic, hebrew, and other related scripts.
+ *
+ * @param text - The text to be checked for RTL characters.
+ * @returns `true` if the text contains RTL characters, otherwise `false`.
+ */
+const isRTL = (text: string): boolean => {
+  const rtlChars =
```
gtarpenning, line 72: you could technically compile this first. If we are running this against all text in every panel we might want to take that micro perf gain.

weave-js/src/components/Panel2/PanelString.tsx
```diff
@@ -224,16 +245,19 @@ export const PanelString: React.FC<PanelStringProps> = props => {
         console.error(e);
       }
     }
+    const textStyle: React.CSSProperties = isRTL(fullStr)
```
gtarpenning, line 248: we could also limit the regex check to just a portion of the str if its gigantic. Like im just imaging tables of novels or something, you can probably tell if its hebrew and should be RTL from just the first... 1000 chars? Idk

gtarpenning (approved):
I wonder how this will affect existing panels, and if anyone will want to op out of this?

gtarpenning:
In the ticket it _does_ mention an explicit option... like a button "Plain text (right to left)" Idk that would probably get super annoying to change every panel always though...

weave-js/src/components/Panel2/PanelString.tsx
bcsherma: Yeah I'm limiting to first 100

PR #2434, opened 2024-09-19
feat: Minimize blocking calls while Tracing 

tssweeney:
Successor to https://github.com/wandb/weave/pull/2384

# Non-Blocking Tracing Path for Improved Performance

This PR aims to implement a non-blocking tracing path that enhances resilience to errors and improves overall performance. The primary goal is to minimize, ideally eliminate, blocking time in the main tracing path by offloading I/O operations (network, disk, etc.) to non-main threads.

## Motivation

1. Reduce latency in LLM Applications: Users expect Weave to add minimal overhead to their applications.
2. Improve evaluation performance: Current slow performance due to dataset uploads and other blocking operations discourages some users from utilizing evaluations.
3. Optimize resource usage: By backgrounding calls, we can amortize evaluation time across LLM inference time, significantly reducing the impact on user experience.

## User-Facing Changes

- Faster execution of `weave.op` tracked methods
- Improved performance for evaluations
- Potential increase in exit time due to background uploads (observable at cell-level in notebooks and program-level in scripts)

## Implementation Details

1. Lazy Ref Attributes:
   - `digest`, `extra`, and related components are now calculated on-demand.
   - Implemented using `_digest`, `_row_digests`, and `_extra` attributes with `@property` decorators.
   - Maintains backward compatibility: `ref.digest` and `ref.extra` still work but block until calculation is complete.
   - Derivative values (e.g., `uri`) inherit this lazy behavior.

2. FutureExecutor:
   - Replaces the `AsyncJobQueue` with `FutureExecutor` for a more versatile and robust solution.
   - Introduces `defer` and `then` functions for Promise-like operations with Futures.
   - Global `FutureExecutor` for non-priority tasks.
   - WeaveClient-specific `FutureExecutor` for prioritized tasks (e.g., file uploads).

3. Non-Blocking Call Lifecycle:
   - `create_call` and `finish_call` methods are now entirely non-blocking.
   - Asynchronous operations include: object saving, op saving, file saving, JSON conversion, and API requests.
4. This change also makes us resilient to all sorts of errors emitted from the server - allowing us to pass the resilience tests previously skipped

## Developer Considerations

1. Error Handling: Errors may surface later in the execution flow (e.g., object URI creation might reveal earlier write errors).
2. Thread Safety: Standard multithreading precautions apply.
3. Execution Flow: Main process manages logical connections between Futures and their usage. Early termination may result in incomplete server synchronization.
4. Future Improvements: This architecture facilitates potential implementation of a sidecar process for I/O task management.


weave/conftest.py
```diff
@@ -185,15 +186,19 @@ class TestOnlyFlushingWeaveClient(weave_client.WeaveClient):
     read operation(s) are performed.
     """
 
+    def set_autoflush(self, value: bool):
```
andrewtruong, line 189: Does this need to be a setter?  Maybe simpler as just `flush: bool`

weave/trace/weave_client.py
```diff
@@ -163,7 +169,7 @@ def map_to_refs(obj: Any) -> Any:
 class Call:
     """A Call represents a single operation that was executed as part of a trace."""
 
-    op_name: str
+    _op_name: Union[str, Future[str]]
```
andrewtruong, line 172: this doesn't feel right on a public api

weave/trace/concurrent/futures.py
```diff
@@ -0,0 +1,244 @@
...
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+class FutureExecutor:
+    """A utility for thread-local threadpool execution and promise-like chaining.
+
+    This class provides a thread-safe way to submit and execute jobs asynchronously
+    using thread-local ThreadPoolExecutors. It ensures proper shutdown of the executors when the
+    object is deleted or when the program exits.
```
andrewtruong, line 55: Why did you choose threadpool instead of using asyncio?

weave/trace/refs.py
```diff
@@ -43,22 +85,56 @@ def with_attr(self, attr: str) -> "RefWithExtra":
...
+        return self._digest
 
     def __post_init__(self) -> None:
-        refs_internal.validate_no_slashes(self.digest, "digest")
-        refs_internal.validate_no_colons(self.digest, "digest")
-        refs_internal.validate_extra(list(self.extra))
+        if isinstance(self._digest, str):
+            refs_internal.validate_no_slashes(self._digest, "digest")
+            refs_internal.validate_no_colons(self._digest, "digest")
+        # TODO: FIx me!
```
andrewtruong, line 137: fix what?

weave/trace/serialize.py
```diff
@@ -9,37 +9,59 @@
...
     TraceServerInterface,
 )
+from weave.trace_server.trace_server_interface_util import bytes_digest
 
+if typing.TYPE_CHECKING:
+    from weave.trace.weave_client import WeaveClient
 
-def to_json(obj: Any, project_id: str, server: TraceServerInterface) -> Any:
+
+def to_json(obj: Any, project_id: str, client: "WeaveClient") -> Any:
```
andrewtruong, line 18: Do you need the whole client, or can you get away with just the server?

weave/trace/weave_client.py
tssweeney: This is a good comment - i agree. Which is why it is underscored. The public API is just `.op_name`. Obviously this pattern is used throughout this PR. Do you have an alternative?

weave/trace/concurrent/futures.py
tssweeney: A few important reasons:
1. Thread jobs start immediately after submission, rather than waiting for an `await` call
2. You can use threads anywhere (including within asyncio coroutines), whereas asyncio coroutines can only be "awaited" within an `async` context
3. Personally, threads are easier to reason about in terms of execution flow.

Threads are the OGs, asyncio have additional constraints (namely #2) that make it challenging. I would hate to have to rework all our code (and our customer's code) just to get a defered value.

weave/trace/serialize.py
tssweeney: You need the client in particular, so that we can schedule the file creation.

weave/tests/trace/test_evaluation_performance.py
```diff
@@ -0,0 +1,165 @@
...
+
+    assert log == ["ensure_project_exists"]
+
+    with paused_client(client) as client:
+        res = await evaluation.evaluate(predict)
+        assert res["score"]["true_count"] == 1
+        log = [l for l in client.server.attribute_access_log if not l.startswith("_")]
+        assert log == ["ensure_project_exists"]
+
+    log = [l for l in client.server.attribute_access_log if not l.startswith("_")]
```
andrewtruong, line 125: Can you help me grok this?

It seems like:
1. You have a test checking for the internal sever calls, where the first is the `ensure_project_exists` check
2. There is the same thing(?) with a `paused_client` context?  What is this for?

weave/tests/trace/test_evaluation_performance.py
```diff
@@ -0,0 +1,165 @@
...
+
+    counts = {}
+    for l in log:
+        if l not in counts:
+            counts[l] = 0
+        counts[l] += 1
```
andrewtruong, line 131: nit: you can use `Counter` here

weave/trace/concurrent/futures.py
andrewtruong: Ok, that's fair.  Do you expect end-users to ever have to use this executor, or will it be mainly internal (including but not exceeding the integrations surface)?

weave/trace/weave_client.py
```diff
@@ -641,8 +654,10 @@ def create_call(
         attributes._set_weave_item("os_release", platform.release())
         attributes._set_weave_item("sys_version", sys.version)
 
+        op_name_future = self.future_executor.defer(lambda: op_def_ref.uri())
```
andrewtruong, line 657: Can you help me understand how this works?

It seems like we:
1. Save an op, which generates a ref
2. That ref's URI is not necessarily available yet, so we make a future of it and pass it as the value of `call._op_name`
3. The `call_start` is also deferred (but we don't wait for the uri to resolve first?)
4. The `end_call` is also deferred (also not clear if there are guarantees this always comes after call_start -- what if you start and stop immediately?)

weave/trace/serialize.py
andrewtruong: Does it make sense to move this method to the client then?  I think the client should have a way to easily get the `project_id`, so you'd basically instead have:
```py
res = client.jsonify(obj)
```

weave/trace/weave_client.py
andrewtruong: You could use an alternate constructor, but I'm not sure it's much prettier here.

The alt would be something like:
```py
@classmethod
def from_future(cls, fut): ...
```

weave/trace/concurrent/futures.py
andrewtruong: Also, have you seen AnyIO?
https://anyio.readthedocs.io/en/stable/

This is what I instinctually jumped to when I saw this problem.

andrewtruong (approved):

PR #2441, opened 2024-09-20
fix(weave): Make op methods work in py312

andrewtruong:


weave/tests/trace/test_client_trace.py
```diff
@@ -1708,7 +1708,7 @@ def _wrapped_fn(*args):
     [
         map_simple,
         map_with_threads_no_executor,
-        map_with_thread_executor,
+        # map_with_thread_executor,  # <-- Flakes in CI
```
tssweeney, line 1711: is this new or old?

tssweeney (approved):

weave/tests/trace/test_client_trace.py
andrewtruong: old, just commenting out again because it seems to flake a lot

PR #2446, opened 2024-09-20
fix: Fix rendering of singular ref outputs

tssweeney:
This PR does a few things:
1. Auto-expand the first layer of inputs and outputs
2. Make sure that single-ref outputs are viewed in the object viewer
3. Fix the object viewer to support expanding direct child refs
Good example: https://app.wandb.test/adrian2/support-model-eval/weave/calls/01919101-4953-79e3-b797-68e9e506abf9


gtarpenning (approved):
Yikes yes this is better.  Also I tested on the dupe ref workspace and it works 🍏

PR #2447, opened 2024-09-20
fix: LiteLLM Cost Logging

tssweeney:
This fixes LiteLLM Cost logging for non-streaming use cases. Streaming use cases are not yet solved. Maybe someone else can take a look at how to build an accumulator inside the APIResponse object in a durable way. For now this at least solves the non-stream cases


weave/integrations/litellm/litellm_test.py
```diff
@@ -149,6 +144,7 @@ def test_litellm_quickstart_stream(
     for chunk in chat_response:
         if chunk.choices[0].delta.content:
             all_content += chunk.choices[0].delta.content
+    del chat_response
```
andrewtruong, line 147: why do you need this?

weave/integrations/litellm/litellm_test.py
```diff
@@ -11,12 +11,12 @@
...
-# parts that are affected by this change to unblock CI
-USES_RAW_OPENAI_RESPONSE_IN_ASYNC = (
-    semver.compare(litellm._version.version, "1.42.11") > 0
-)
+# Changed the output format for OpenAI to use APIResponse.
+# We can handle this in non-streaming mode, but in streaming mode, we
+# have no way of correctly capturing the output and not messing up the
+# users' code (that i can see). In these cases, model cost is not captured.
+
+USES_RAW_OPENAI_RESPONSE = semver.compare(litellm._version.version, "1.42.11") > 0
```
andrewtruong, line 19: better to use `importlib.metadata` here I think

weave/integrations/openai/openai_sdk.py
```diff
@@ -10,6 +10,38 @@
     from openai.types.chat import ChatCompletionChunk
 
 
+def maybe_unwrap_api_response(value: typing.Any) -> typing.Any:
```
andrewtruong, line 13: So we need this because LiteLLM is using the wrapped version for some reason?

Any idea why they do that?

jwlee64 (approved):

weave/integrations/openai/openai_sdk.py
tssweeney: That is exactly why. They are requesting the raw responses and then consuming the headers

weave/integrations/litellm/litellm_test.py
tssweeney: oops, dev tracks

weave/integrations/litellm/litellm_test.py
tssweeney: ok, changed

andrewtruong (approved):

PR #2448, opened 2024-09-21
fix: Ensure patching fails gracefully

tssweeney:
This PR improves our resilience to patching errors. It correctly catches all patching errors and logs the issue only once to the user.

This also introduces a test suite to:
1. Assert that we correctly catch these issues (and sets up testing for next pr)
2. Makes sure that all gobbled errors are thrown in test suites


weave/trace_server/async_batch_processor.py
```diff
@@ -64,6 +65,8 @@ def _process_batches(self) -> None:
                     if e.response.status_code == 413:
                         # 413: payload too large, don't raise just log
                         logger.error(f"Error processing batch: {e}")
+                        if get_raise_on_captured_errors():
```
gtarpenning, line 68: nice, was going to suggest that

weave/trace/op_extensions/error_once.py
```diff
@@ -0,0 +1,12 @@
+from typing import Callable
+
+LOG_ONCE_MESSAGE_SUFFIX = " (subsequent messages of this type will be suppressed)"
+logged_messages = []
+
+
+def log_once(log_method: Callable[[str], None], message: str) -> None:
```
gtarpenning, line 7: a bit confusing wrt the file name `error_once`. Also a docstring here would be nice (with the cannonical example)

gtarpenning: this is to be used mainly in generators? for loops?

gtarpenning (approved):
We could really use codecov! I know @jamie-rasmussen  mentioned it but it would be especially useful for these kinds of changes (broad but shallow).

weave/trace/op_extensions/log_once.py
```diff
@@ -0,0 +1,28 @@
+from typing import Callable
+
+LOG_ONCE_MESSAGE_SUFFIX = " (subsequent messages of this type will be suppressed)"
+logged_messages = []
+
+
+def log_once(log_method: Callable[[str], None], message: str) -> None:
```
andrewtruong, line 7: Is the intent to use this with many different loggers?

If just python logging, in future you can use the builtin `addFilter`

andrewtruong (approved):

PR #2470, opened 2024-09-24
feat: Expose TableQueryStats and Paginate Large Tables in UI

tssweeney:
When viewing large datasets (or requesting them via API), we can crash once the dataset is large enough. This PR implements a few things:
1. A stats endpoint for row calculation server-side
2. Uptakes pagination support in the UI and adds a few cosmetic niceties
3. Critical! Refactors our Table query layer. Currently, it is not optimized and there are horrible performance characteristics - namely: querying ANY dataset in a project costs as much as loading the entire project's rows! This is VERY BAD!

Pairs with: https://github.com/wandb/core/pull/24158/files

TODO:
- [ ] Split Backend and UI (will do after approval)
   * https://github.com/wandb/weave/pull/2535
   * https://github.com/wandb/weave/pull/2534

Future:
* Make a `stream` endpoint that supports row-level streaming like calls (this is what will get us even better performance


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -766,118 +774,104 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
...
-            """
-        if parameters is None:
-            parameters = {}
-        if limit:
-            query += " LIMIT {limit: UInt64}"
-            parameters["limit"] = limit
-        if offset:
-            query += " OFFSET {offset: UInt64}"
-            parameters["offset"] = offset
+        if sort_fields is None or len(sort_fields) == 0:
```
andrewtruong, line 818: nit: `if not sort_fields:`

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -766,118 +774,104 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
...
+                conds.append(
+                    f"tr.digest IN {{{pb.add_param(req.filter.row_digests)}: Array(String)}}"
+                )
 
-        sort_clause = ""
-        pb = ParamBuilder()
+        sort_fields = []
         if req.sort_by:
-            sort_fields = []
             for i, sort in enumerate(req.sort_by):
```
andrewtruong, line 786: why do you need to enumerate?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -766,118 +774,104 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
...
         return tsi.TableQueryRes(rows=rows)
 
     def _table_query(
         self,
         project_id: str,
         digest: str,
-        conditions: Optional[list[str]] = None,
+        pb: ParamBuilder,
+        *,
+        sql_safe_conditions: Optional[list[str]] = None,
```
andrewtruong, line 813: Why did `conditions` change to `sql_safe_conditions`?  What does that mean?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -766,118 +774,104 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
...
-            sort_clause=sort_clause,
         )
         return tsi.TableQueryRes(rows=rows)
 
     def _table_query(
         self,
         project_id: str,
         digest: str,
-        conditions: Optional[list[str]] = None,
+        pb: ParamBuilder,
```
andrewtruong, line 811: not blocking: The way `pb` is passed in here feels weird.  Maybe I'm just not familiar with this pattern yet

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -766,118 +774,104 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
...
-            parameters={
-                "project_id": project_id,
-                "digest": digest,
-                **parameters,
-            },
-        )
+        if (
+            len(sort_fields) == 1
+            and sort_fields[0].field.field == ROW_ORDER_COLUMN_NAME
+            and (sql_safe_conditions is None or len(sql_safe_conditions) == 0)
```
andrewtruong, line 829: nit: `not sql_safe_conditions`

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -766,118 +774,104 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
...
+            query = make_natural_sort_table_query(
+                project_id,
+                digest,
+                pb,
+                limit=limit,
+                offset=offset,
+                natural_direction=sort_fields[0].direction,
+            )
+        else:
+            sql_safe_sort_clause = f"ORDER BY {', '.join([sort_field.as_sql(pb, TABLE_ROWS_ALIAS) for sort_field in sort_fields])}"
```
andrewtruong, line 840: this looks complex -- I would break into 2 lines

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -766,118 +774,104 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
...
+                limit=limit,
+                offset=offset,
+            )
+
+        query_result = self.ch_client.query(query, parameters=pb.get_params())
 
         return [
             tsi.TableRowSchema(digest=r[0], val=json.loads(r[1]))
             for r in query_result.result_rows
         ]
```
andrewtruong, line 856: nit: more clear as:
```py
return [
    tsi.TableRowSchema(digest=digest, val=json.loads(value))
    for digest, value in query_result.result_rows
]
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -766,118 +774,104 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
...
 
+    def table_query_stats(self, req: tsi.TableQueryStatsReq) -> tsi.TableQueryStatsRes:
+        parameters: Dict[str, Any] = {
```
andrewtruong, line 859: -> `req.model_dump()`

weave/trace_server/trace_server_interface.py
```diff
@@ -815,6 +829,7 @@ def objs_query(self, req: ObjQueryReq) -> ObjQueryRes: ...
     def table_create(self, req: TableCreateReq) -> TableCreateRes: ...
     def table_update(self, req: TableUpdateReq) -> TableUpdateRes: ...
     def table_query(self, req: TableQueryReq) -> TableQueryRes: ...
+    def table_query_stats(self, req: TableQueryStatsReq) -> TableQueryStatsRes: ...
```
andrewtruong, line 832: Per the implementations, it looks like this endpoint returns the number of rows for a query.  This is surprising because `stats` suggests to me something like a mean/median/mode.  This endpoint may be more clearly named as `table_query_num_rows`

weave/trace_server/table_query_builder.py
```diff
@@ -0,0 +1,103 @@
...
+
+from weave.trace_server.orm import ParamBuilder
+
+# Constants for table and column names
+TABLE_ROWS_ALIAS = "tr"
+VAL_DUMP_COLUMN_NAME = "val_dump"
+ROW_ORDER_COLUMN_NAME = "row_order"
+
+
+def make_natural_sort_table_query(
```
andrewtruong, line 11: I understand the utility of this optimization, but shouldn't the query engine find the fastest way to get data for you given your params?  There don't seem to be any obvious optimization fences in the other query.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/traceServerClientTypes.ts
```diff
@@ -188,6 +188,17 @@ export type TraceTableQueryReq = {
     row_digests?: string[];
   };
   limit?: number;
+  offset?: number;
+  sort_by?: SortBy[];
+};
+
+export type TraceTableQueryStatsReq = {
```
andrewtruong, line 195: can't wait to generate these instead!

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: fixed

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: fixed

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: Added comment:
```
# using the `sql_safe_*` prefix is a way to signal to the caller
# that these strings should have been santized by the caller.
```

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: `pb` is used in a lot of places. It is a helper class that can be used to accumulate placeholders and values used for parameter building

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: Fixed

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: Done

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: yup, changed

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: i do not want to blindly dump for params - i like the explicit nature of this as i want full clear control over the params sent to clickhous

weave/trace_server/trace_server_interface.py
tssweeney: we also have `calls_query_stats` which is similar. These are intended to support additional column stats and other summary info for a query in the future

weave/trace_server/table_query_builder.py
tssweeney: i'm not sure i understand this comment. The performance improvement is quite dramatic by applying the limit/offset in the inner query as the free-form val_dumps field is skipped

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: You can do `model_dump(include={'project_id', 'digest'})`

But you are right -- in this case where the dict is small, just making your own dict is also more performant

weave/trace_server/table_query_builder.py
andrewtruong: ok, I don't know enough about clickhouse to comment deeply here.

I was just surprised that the query engine doesn't figure out the most performant way for you.

weave/trace_server/trace_server_interface.py
tssweeney: it will expand, yes. right now it is just `count`, but you can imagine it could contain stats per column, for example

Why is that bad?

andrewtruong (approved):

tssweeney:
Closing as this was split into 2 PRs

PR #2484, opened 2024-09-25
chore: add support for version number obj query in sqlite

gtarpenning:
Add support for reading objects with digests like: `v1` in sqlite. Don't skip sqlite in test.


jwlee64 (approved):
nice

tssweeney (approved):

PR #2486, opened 2024-09-25
fix(weave): align tables column header cell text based on panel type

domphan-wandb:
## Description

**Internal JIRA:** [WB-17238](https://wandb.atlassian.net/browse/WB-17238)

The JIRA reports a difficulty in difficulty in reading Tables because column header cells are left justified while numerical value cells are right justified.

This PR fixes that by **using the same text alignment for the column header cell as the value cell**. This solution differs from the proposed solution on the JIRA (center all header cells and cells) because typically numerical columns are right justified for visual clarity. It's easier to interpret the magnitude of cells when numerical values have the same level of precision. Non-numerical column header cells are defaulted to `center` text alignment.

Previous PR that introduced the issue: https://github.com/wandb/weave/pull/875

### Implementation Details
I leveraged the new `WeaveFormatContext` from the PR mentioned above to add a `columnFormat` field to specifies the column headers' `textAlign` value.

### Screenshots

Before:
image

After:
image

**One thing to note,** on hover-over, the `EllipsisIcon` to trigger the Popup will overlap the column name text due to its negative margin.
image

I thought this behavior is acceptable because it currently happens when the column widths are resized smaller than the column header text.

edit(10/02): See update below

## Testing

How was this PR tested?
Visually with my [pytorch-intro project](https://wandb.ai/dominic-phan-weights-and-biases/pytorch-intro?nw=nwuserdominicphan). 

[WB-17238]: https://wandb.atlassian.net/browse/WB-17238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Update to handle overlapping indicators

I took a display approach that [material-ui data grid uses](https://mui.com/x/react-data-grid/) to place the **indicators on the opposite side in reverse** for numerical columns. This mitigates the ellipses stacking on the right justified column name.

Here's the column header on hoverover:
image

When pinned on hoverover:
image

Pinned, sorted on hoverover:
image

Without hoverover:
image

### DIFF
--- a/weave-js/src/components/Panel2/PanelTable.styles.ts
+++ b/weave-js/src/components/Panel2/PanelTable.styles.ts
@@ -32,7 +32,6 @@ export const ColumnName = styled.div`
     color: ${globals.primary};
   }
   overflow: hidden;
-  text-align: left;
 `;
 ColumnName.displayName = 'S.ColumnName';
 
--- a/weave-js/src/components/Panel2/PanelTable.styles.ts
+++ b/weave-js/src/components/Panel2/PanelTable.styles.ts
@@ -98,18 +98,23 @@ FilterIcon.displayName = 'S.FilterIcon';
 
 export const ColumnAction = styled.div<{isHovered?: boolean}>`
   cursor: pointer;
+  height: 100%;
+  background-color: transparent;
+`;
+ColumnAction.displayName = 'S.ColumnAction';
+
+export const ColumnActionContainer = styled.div<{isHovered?: boolean}>`
+  display: flex;
   padding: 5px 0px 0px 0px;
+  font-size: 20px;
   flex: 0 0 auto;
   height: 100%;
-  background: white;
-  font-size: 20px;
 
   :hover {
     background-color: ${hexToRGB(OBLIVION, 0.04)};
   }
-  background-color: transparent;
 `;
-ColumnAction.displayName = 'S.ColumnAction';
+ColumnActionContainer.displayName = 'S.ColumnActionContainer';
 
 export const TableAction = styled.div<{
   highlight?: boolean;
--- a/weave-js/src/components/Panel2/PanelTable/util.ts
+++ b/weave-js/src/components/Panel2/PanelTable/util.ts
@@ -29,7 +29,7 @@ import {WeaveFormatContextType} from '../WeaveFormatContext';
 import * as Table from './tableState';
 import {useTableStateWithRefinedExpressions} from './tableStateReact';
 
-// Formatting for PanelNumbers and PanelStrings inside Tables
+// Formatting for PanelNumber and PanelString headers and values inside Tables
 export const getColumnCellFormats = (colType: Type): WeaveFormatContextType => {
   const t = nullableTaggableStrip(colType);
   const numberFormat =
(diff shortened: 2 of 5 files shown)


domphan-wandb:

jamie-rasmussen (approved):
Even though this PR doesn't introduce the problem with the ellipsis it does seem like it would be encountered a lot more frequently with this change. This is part of a larger problem with indicators in the header e.g pinned columns. 

I don't know how hard it would be to fix the underlying issue, but you might want to take a crack at it before merging this one in.

weave-js/src/components/Panel2/PanelTable/util.ts
```diff
@@ -42,9 +42,14 @@ export const getColumnCellFormats = (colType: Type): WeaveFormatContextType => {
         }
       : {};
   const stringFormat = {spacing: t === 'string'};
+  const columnFormat = {
+    textAlign: t === 'number' ? ('right' as const) : ('center' as const),
```
onx2, line 46: is `as const` necessary? This a type exported from React for CSS styles if you want to use that. non-blocker though

weave-js/src/components/Panel2/PanelTable.styles.ts
```diff
@@ -99,18 +98,23 @@ FilterIcon.displayName = 'S.FilterIcon';
...
   cursor: pointer;
+  height: 100%;
+  background-color: transparent;
+`;
+ColumnAction.displayName = 'S.ColumnAction';
+
+export const ColumnActionContainer = styled.div<{isHovered?: boolean}>`
+  display: flex;
   padding: 5px 0px 0px 0px;
+  font-size: 20px;
```
onx2, line 109: Do we have a common font-size variable we can access here instead of a hard-coded px? non-blocking as this was here before

weave-js/src/components/Panel2/PanelTable.styles.ts
```diff
@@ -99,18 +98,23 @@ FilterIcon.displayName = 'S.FilterIcon';
...
   flex: 0 0 auto;
   height: 100%;
-  background: white;
```
onx2, line 105: This background was probably here because of sticky headers, have you tested that use-case?

weave-js/src/components/Panel2/PanelTable.styles.ts
domphan-wandb: As far as I know, I don't think so. This `font-size` property is used only to size the indicator and ellipses icons

weave-js/src/components/Panel2/PanelTable.styles.ts
domphan-wandb: I cannot scroll on the table because of the pagination, but I wanted to note that before I changed the divs up to `ColumnActionContainer` (for containing `ColumnAction`s) and `ColumnAction` (for each icon) the CSS for ColumnAction had both `background: white` and `background-color: transparent` in that order.

weave-js/src/components/Panel2/PanelTable/util.ts
domphan-wandb: I tried using it, but I got type errors from it, so I used `as const` because it was used, (I assume to resolve the issue), in `numberFormat`

here's the type error:
image

onx2 (approved):

PR #2495, opened 2024-09-26
chore: Improve test coverage - specifically asserting no error logs

tssweeney:
This PR improves some of our test mechanics such that tests are closer to real user-environment & we ensure that no `ERROR` logs are created.

Many lines changed, but conceptually, quite simple:
1. Replace `always_raise_on_captured_errors` with `logging_error_check` fixture in `weave/conftest.py`.
  * This is important because `always_raise_on_captured_errors` was intended to "turn on" raising in a number of places where we currently try/except. However, the OpenAI bug that I introduced was the result of NOT raising when we needed to, and this test fixture "hid" that issue. Instead, `logging_error_check` ensures that we do not emit any `logger.error` messages during runtime, which: a) makes pytests run in the same way that user environments do; b) is a better way of catching these sorts of issues in the future.
2. Move `logger.error` AFTER `get_raise_on_captured_errors` in all cases (`weave/trace/async_job_queue.py`, `weave/trace/op.py`, `weave/trace/op_extensions/accumulator.py`, `weave/trace/patcher.py`, `weave/trace/vals.py`, `weave/trace_server/async_batch_processor.py`.
    * This does not effect users, but makes sure our tests in `weave/tests/trace/test_tracing_resilience.py` are more correct. Refactor `weave/tests/trace/test_tracing_resilience.py` to conform

Nice to-haves
1. `pytest.ini` now contains a list of our markers so reduce warnings in tests
2. Cleanup some unused markers
3. `weave/tests/trace/test_tracing_resilience.py`: Change `TestException` to `DummyTestException` which makes pytest happier

---


pytest.ini
```diff
@@ -9,3 +9,6 @@ filterwarnings =
 addopts = -v
 markers =
     skip_clickhouse_client: marks tests as requiring clickhouse client to be skipped
+    weave_client: alias for `client` fixture used for CI dispatching. Do not use directly in code.
```
andrewtruong, line 12: Sorry, where/how is this used?  It's not obvious to me

pytest.ini
```diff
@@ -9,3 +9,6 @@ filterwarnings =
...
+    flaky: marks tests as flaky (CI will retry them if they fail)
```
andrewtruong, line 13: I think you get this for free with:
https://pypi.org/project/pytest-rerunfailures/

We should add it back to `requirements.test.txt`

pytest.ini
tssweeney: in CI: `-m "weave_client and not skip_clickhouse_client" \`, and in code:

```
def pytest_collection_modifyitems(config, items):
    # Add the weave_client marker to all tests that have a client fixture
    for item in items:
        if "client" in item.fixturenames:
            item.add_marker(pytest.mark.weave_client)
```

weave/tests/trace/test_tracing_resilience.py
```diff
@@ -31,104 +31,104 @@ def test_resilience_to_user_code_errors(client):
...
+    with raise_on_captured_errors(True):
+        with pytest.raises(DummyTestException):
+            do_test()
 
     # The user's exception should be raised - even if we're not capturing errors
-    with raise_on_captured_errors(False):
-        with pytest.raises(TestException):
-            do_test()
+    with pytest.raises(DummyTestException):
+        do_test()
```
andrewtruong, line 45: This is a test about testing, right?  It seems like it should be in its own file noting so

tssweeney: No, it is a test about trace resilience

pytest.ini
tssweeney: hmmm, we have `pytest.mark.flaky` in a few places... does you comment imply these are not actually working right now?

weave/trace/async_job_queue.py
```diff
@@ -85,9 +85,9 @@ def callback(f: Future[T]) -> None:
                 self._active_jobs.remove(f)
             exception = f.exception()
             if exception:
-                logger.error(f"Job failed with exception: {exception}")
                 if get_raise_on_captured_errors():
                     raise
+                logger.error(f"Job failed with exception: {exception}")
```
andrewtruong, line 90: Why does order here matter?

pytest.ini
andrewtruong: Yeah if the plugin is not present then the marks are ignored (tests are still run, but no retries for flake)

You can confirm by doing:
```
tox -e trace-py312 -- --markers
```

The flaky marker does not show up which means it's being ignored atm

tssweeney: ok, i can add this to this PR and remove this line

weave/trace/async_job_queue.py
tssweeney: if `get_raise_on_captured_errors` is set (something only done for testing), then we don't want the logger to log (this is true for all cases of this pattern) We effectively get the error bubbled up as a proper exception

pytest.ini
tssweeney: done

andrewtruong (approved):

weave/conftest.py
```diff
@@ -84,10 +84,62 @@ def pytest_collection_modifyitems(config, items):
     items[:] = selected_items
 
 
+PYTEST_CURRENT_TEST_ENV_VAR = "PYTEST_CURRENT_TEST"
```
andrewtruong, line 87: where is this env var set?

tssweeney: Automatically by pytest

tssweeney: https://docs.pytest.org/en/latest/example/simple.html#pytest-current-test-environment-variable

weave/conftest.py
```diff
@@ -84,10 +84,62 @@ def pytest_collection_modifyitems(config, items):
...
 @pytest.fixture(autouse=True)
-def always_raise_on_captured_errors():
-    with raise_on_captured_errors():
-        yield
+def logging_error_check(request, log_collector):
+    yield
+    if "disable_logging_error_check" in request.keywords:
+        return
+    error_logs = log_collector.get_error_logs()
+    if len(error_logs) > 0:
```
andrewtruong, line 139: nit: `if error_logs:`

tssweeney: yup fixed

PR #2496, opened 2024-09-26
perf: split single object read into two queries

gtarpenning:
## Description

When querying for a single object by digest, break up the query into two parts. This should improve performance on objects with many versions (or lots of data)
1. list all objects and calculate the `version_count` (unused afaict), `version_index` which must see all objects. 
2. make a second query for the val of the object, using the found `version_index` if the digest is a version.

### Performance: 

Existing queries that crash due to massive memory requirements (`50GB+`) now can be resolved in around 1 second, with a memory requirement less than `100MB` total.

Table of performance of the two queries, as tested in CH Cloud on an object with `462,000` versions. 

| Type | Time (ms)   | Size (MB) | Rows Read |
| -------- | -------- | ------- | ------- |
| version count  | 887    | 69.5 | 618,088 |
| val_dump | 152     | 12.62 | 72,810 |

## Testing

- existing tests
- manual clickhouse testing, [query here](https://console.clickhouse.cloud/services/aef2798b-2436-46a6-b8e5-1d067fb98ace/console/query/9ccb8f5e-2614-450c-a070-005f39947e99?param_project_id=UHJvamVjdEludGVybmFsSWQ6Mzk4MDIzNzg%3D&param_object_id=ModelInput&param_version_digest=wgVNJEuUrmXcXSZLUIWliY8dpjjGxh2HWSN6ziYX8Yc)


weave/trace_server/clickhouse_schema.py
```diff
@@ -150,4 +150,5 @@ class SelectableCHObjSchema(BaseModel):
     base_object_class: typing.Optional[str]
     digest: str
     version_index: int
+    version_count: int
```
gtarpenning, line 153: we already calculate this, and it would be great from a performance standpoint. Is it too confusing for users though? on the object schema?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1413,6 +1405,144 @@ def _insert_call_batch(self, batch: list) -> None:
...
+            SELECT
+                project_id,
+                object_id,
+                created_at,
+                kind,
+                base_object_class,
+                refs,
+                val_dump,
+                if (kind = 'op', 1, 0) AS is_op
+            FROM (
```
tssweeney, line 1483: If you extract everything except `val_dump` from the first query, then this query can avoid the 2-layer query by ignoring row_number stuff

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -582,29 +582,21 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
-                parameters["version_digest"] = req.digest
-        objs = self._select_objs_query(
-            req.project_id,
-            conditions=conds,
-            object_id_conditions=object_id_conditions,
-            parameters=parameters,
-        )
-        if len(objs) == 0:
-            raise NotFoundError(f"Obj {req.object_id}:{req.digest} not found")
+                condition_part = f"digest = '{req.digest}'"
```
tssweeney, line 592: parameterize

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: Like we allow for re-inserting objects, so there could be more than one with the same primary key (briefly). Re that issue that the data team mentioned wrt to changing `created_at` times.

gtarpenning: although in this case val_dump would be the same... Okay i see

gtarpenning: The `created_at` could be different, but we probably don't care

tssweeney: get created at from the first one

gtarpenning: yessir, check it now

tssweeney (approved):

gtarpenning:
Going with [this](https://github.com/wandb/weave/pull/2512) pr

PR #2500, opened 2024-09-26
chore(weave): add width to autocomplete

mbarrramsey:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

Adds minWidth param and ref to the autocomplete component so we can measure the full width of the text input for use in https://github.com/wandb/core/pull/24247.

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/Form/AutoComplete.tsx
+++ b/weave-js/src/components/Form/AutoComplete.tsx
@@ -59,7 +59,7 @@ const getStyles = (props: AdditionalProps) => {
                 opacity: 1,
               },
               '& .MuiInputBase-input': {
-                padding: '0px', // Adjust padding as needed
+                padding: '0px',
                 minHeight: `${HEIGHTS[size]} !important`,
               },
             },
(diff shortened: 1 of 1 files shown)


connieelee (approved):

weave-js/src/components/Form/AutoComplete.tsx
```diff
@@ -145,15 +145,17 @@ type AdditionalProps = {
   size?: SelectSize;
   isDarkMode?: boolean;
   maxWidth?: number;
+  minWidth?: number;
   hasInputValue?: boolean;
+  inputRef?: React.MutableRefObject<HTMLDivElement | null>;
```
connieelee, line 150: **nitpick:** `inputRef` naming seems misleading if it's actually a div element under there?

mbarrramsey: Maybe `inputDivRef`? or `autocompleteDivRef`?

connieelee: we should probably leave out "input" entirely. if you inspect the elements, there's an actual `input` element nested inside, but this ref is definitely not being attached to that.

have you tried `React.forwardRef()`?

mbarrramsey: I did try the forwardRef, but I don't think it works well with generics 🤔 see https://github.com/wandb/core/pull/24247/files#r1777844929. We could try using the wrapper approach but it seems like the custom name is similar?

connieelee: ahh i see. maybe something like `containerRef` or `autocompleteRef` seems fine then!

PR #2507, opened 2024-09-27
fix: select columns for child call table preview

jamie-rasmussen:
## Description

https://wandb.atlassian.net/browse/WB-21218

Also sorts attributes after inputs and output in the column management modal.

Before:
Screenshot 2024-09-27 at 9 57 37 AM

After:
Screenshot 2024-09-27 at 9 57 17 AM

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -143,6 +143,9 @@ export const CallsTable: FC<{
 
   paginationModel?: GridPaginationModel;
   setPaginationModel?: (newModel: GridPaginationModel) => void;
+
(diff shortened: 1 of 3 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -140,6 +140,9 @@ export const CallsTable: FC<{
 
   paginationModel?: GridPaginationModel;
   setPaginationModel?: (newModel: GridPaginationModel) => void;
+
+  // Can include glob for prefix match, e.g. "inputs.*"
+  shownCols?: string[];
```
tssweeney, line 145: Nit: I think this is operating more like a filter - if a column does not match this filter, is it not even selectable in the column picker. I might call this: `allowedColumnPatterns` or something a bit more specific.

tssweeney (approved):

PR #2510, opened 2024-09-27
fix: some unfilterable cols showing up in filters

jamie-rasmussen:
## Description

Some field IDs changed in https://github.com/wandb/weave/pull/2339 but the filters UI was still referring to the previous values, causing them to not get filtered out as they should be.

## Testing

How was this PR tested?


gtarpenning (approved):
my bad...

PR #2512, opened 2024-09-27
perf: split metadata/val in objects query 

gtarpenning:
https://github.com/wandb/weave/pull/2496 but better

## Description

This pr: 
- removes the `val_dump` read from the initial select objects query
- adds a second query for val dump conditioned on digest, and combines the two query results then returns

This should improve performance on objects with many versions (or lots of data)
1. list all objects and calculate the `version_count` (unused afaict), `version_index` which must see all objects. 
2. make a second query for the val of the object, using the found digests if not provided.

## Performance: 

Existing queries that crash due to massive memory requirements (`50GB+`) now can be resolved in around 1 second, with a memory requirement less than `100MB` total.

## Testing

- existing tests
- manual clickhouse testing, [query here](https://console.clickhouse.cloud/services/aef2798b-2436-46a6-b8e5-1d067fb98ace/console/query/9ccb8f5e-2614-450c-a070-005f39947e99?param_project_id=UHJvamVjdEludGVybmFsSWQ6Mzk4MDIzNzg%3D&param_object_id=ModelInput&param_version_digest=wgVNJEuUrmXcXSZLUIWliY8dpjjGxh2HWSN6ziYX8Yc)


weave/tests/trace/test_weave_client.py
```diff
@@ -1446,3 +1447,49 @@ def test_object_version_read(client):
     )
     assert obj_res.obj.val == {"a": 5}
     assert obj_res.obj.version_index == 5
+
```
tssweeney, line 1450: I would add a clause to this test that loops through all the known digests and validates that they return the correct payload

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1553,19 +1545,50 @@ def _select_objs_query(
...
-                                "val_dump",
                                 "digest",
                                 "is_op",
                                 "version_index",
                                 "version_count",
                                 "is_latest",
+                                "val_dump",
                             ],
-                            row,
+                            list(row) + ["{}"],  # type: ignore
```
tssweeney, line 1555: would be nice to correctly type this

gtarpenning: As an optional?

gtarpenning: easy, on it...

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1553,19 +1545,50 @@ def _select_objs_query(
...
+            GROUP BY digest
+        """
+        parameters = {
+            "project_id": project_id,
+            "object_ids": object_ids,
+            "digests": digests,
+        }
+        query_result = self._query_stream(query, parameters)
+        # Map digest to val_dump
+        object_values: Dict[str, Any] = {}
```
tssweeney, line 1583: I think you want the key to be `object_id:digest` inside the `object_values` map. while it is true that the same digest for multiple object_ids would be the same data, it could be that they were deleted in one case.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1844,7 +1865,7 @@ def _ch_obj_to_obj_schema(ch_obj: SelectableCHObjSchema) -> tsi.ObjSchema:
         digest=ch_obj.digest,
         kind=ch_obj.kind,
         base_object_class=ch_obj.base_object_class,
-        val=json.loads(ch_obj.val_dump),
+        val=json.loads(ch_obj.val_dump) if ch_obj.val_dump else None,
```
gtarpenning, line 1868: this makes me nervous that there are other places we blindly assume `val` is jsonable @tssweeney

tssweeney: This would be a bug - `val` should never be None. Let's figure out when this is the case

tssweeney: correction, `ch_obj.val_dump`

gtarpenning: This pr would change that to allow None though right? Isn't that what you meant by correctly typing this? SelectableObject would allow an optional val_dump

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1578,7 +1600,7 @@ def _query_stream(
         query: str,
         parameters: Dict[str, Any],
         column_formats: Optional[Dict[str, Any]] = None,
-    ) -> Iterator[QueryResult]:
+    ) -> Iterator[tuple]:
```
gtarpenning, line 1603: This does not return a `QueryResult`! its a tuple...

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1550,19 +1542,49 @@ def _select_objs_query(
...
-                                "val_dump",
                                 "digest",
                                 "is_op",
                                 "version_index",
                                 "version_count",
                                 "is_latest",
+                                "val_dump",
                             ],
-                            row,
+                            list(row) + ["{}"],
```
tssweeney, line 1552: nit: might be nice to comment what this `+['{}']` does

tssweeney (approved):

PR #2518, opened 2024-09-29
chore(weave): Conditionally run weave_query CI only if query code is changed

andrewtruong:
This PR changes CI to only run query service tests if the its dirs are changed, which is currently defined here:
https://github.com/wandb/weave/pull/2518/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R19-R20

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,15 +11,38 @@ name: test
 on:
   push:
 
+env:
+  QUERY_SERVICE_PATHS: 'weave-js/ weave_query/'
+
 jobs:
+  check_paths:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -108,6 +149,8 @@ jobs:
   weavejs-lint-compile:
     name: WeaveJS Lint and Compile
     runs-on: ubuntu-latest
+    needs: check_paths
+    if: ${{ needs.check_paths.outputs.should_run_query_tests == 'true' }}
```
tssweeney, line 153: Hmmm, why run this when python files change?

PR #2519, opened 2024-09-29
chore(weave): Move to nox (speed up unit tests by ~17%)

andrewtruong:
This PR:
1. Moves from `tox` to `nox` (https://nox.thea.codes/en/stable/) -- speeds up unit tests ~17%
2. Moves `requirements*.txt` into `pyproject.toml` (as `dependencies` and `optional-dependencies`)
3. Some linting

To see new testing env names, run `nox -l`.  Here's what they look like now:

```
* lint
* tests-3.9(shard='trace')
* tests-3.9(shard='trace_server')
* tests-3.9(shard='anthropic')
...
```

So you can run `nox -e "tests-3.9(shard='trace')"` to do the python 3.9 trace tests


pyproject.toml
```diff
@@ -28,10 +28,72 @@ classifiers = [
   "Topic :: Software Development :: Widget Sets",
 ]
 requires-python = ">=3.9"
-dynamic = ["version", "dependencies", "optional-dependencies"]
+dynamic = ["version"]
+dependencies = [
+  "pydantic>=2.0.0",
```
tssweeney, line 33: it would be nice to keep the dependency comments as they are currently

tssweeney (approved):
Approving with comments

PR #2521, opened 2024-09-30
chore(weave): Cleanup trace and flow imports

andrewtruong:
The PR looks like a lot, but basically:
1. Moves individual `flow` imports into `flow/__init__.py`
2. Moves old trace imports at `api.py` into `trace/__init__.py` (and update paths that previously used api)
3. The package `weave/__init__.py` imports * from both of these


weave/__init__.py
```diff
@@ -2,16 +2,12 @@
 
 from weave import version
 
-from weave.trace.api import *
+from weave.trace import *
```
tssweeney, line 5: nit: I think this re-exporting stuff is nice because it allows us to allow users to access symbols differently than their source location. However, import * seems not ideal for this case as we might accidentally re-export more symbols than is desirable here.

weave/flow/__init__.py
```diff
@@ -0,0 +1,15 @@
+from weave.flow.agent import Agent, AgentState
+from weave.flow.dataset import Dataset
+from weave.flow.eval import Evaluation, Scorer
+from weave.flow.model import Model
+from weave.flow.obj import Object
+
+__docspec__ = [
```
tssweeney, line 7: docspec is only needed if we want this particular file to be doced. Specifically, we don't want to doc-spec something twice. This can be removed

weave/flow/eval.py
```diff
@@ -49,7 +49,7 @@ def async_call(
     return asyncio.to_thread(func, *args, **kwargs)
 
 
-class EvaluationResults(weave.Object):
+class EvaluationResults(Object):
```
tssweeney, line 52: i think it is cleaner to use `weave.Object` in most cases

weave/trace/__init__.py
```diff
@@ -0,0 +1,274 @@
+"""The top-level functions for Weave Trace API."""
```
tssweeney, line 1: I'm not sure I understand the benefit of having this in `__init__.py` in general i try to stay away from populating `init.py` for a few reasons:
1. `__init__.py` are automatically executed when doing things like `from weave.trace.X import Y`
2. It makes it less clear what the purpose is of the file.

Can you share why you think `api.py` should be replaces with this?

weave/flow/__init__.py
tssweeney: Just to be clear: `__docspec__` is only used by our doc generator. We should expect each symbol to appear in only one `__docspec__` list. Since these are already in the top-level init, no need to have them here

weave/flow/eval.py
andrewtruong: It causes an import cycle in this case which is why it's changed

weave/trace/__init__.py
andrewtruong: I added `api.py` as a stop-gap to help split out the `query` and `trace` apis.  Now that this is just trace, we don't need it.

The pattern I want is `from weave.trace import *` gives you all the symbols you would need in 90% of use cases.

I agree that having definitions in `__init__.py` is bad form.  In general `__init__` should just be imports, not defining anything.  I considered pulling out the definitions in this PR but it seemed noisier (we can pull symbols out in another PR)

If you prefer, I can pull the symbols out in this PR

weave/__init__.py
tssweeney: I don't think that the author of trace.init should be able to control what gets exported here. Yes, __all__ would do that, but personally i think that moves the control to the wrong spot

weave/flow/eval.py
tssweeney: oh, it is because of the `__init__` problems i called out below

andrewtruong: They are changed at the `__init__` level -- you can try adding it back, you'll get a cycle!

weave/trace/__init__.py
tssweeney: Hmmm, i disagree with `from weave.trace import *`. I would never do that in my own code. IMO, `import *` is not a clean pattern we should be promoting.

Even so, i would prefer `__init__.py` to only be used for explicitly re-exporting symbols or doing module-level initialization you want to run on each import of sub modules. I am not seeing the advantage here.

tssweeney: In addition, we should not be working towards `weave.trace` for users anyway. They should not even know about trace server, so all of weave is "trace"

tssweeney: I'd like to see some example of why `import *` is better. I only see less characters typed. Other than that, it feels like there are many footguns and you much with the global namespace.

tssweeney: I think we are prematurely optimizing here. `flow` was an artifact of `weaveflow` and not something users should ever think about. I don't see how this adds user value, but in exchange we have more circular imports and implicit symbols being imported and exported

PR #2541, opened 2024-09-30
fix: `include_feedback` breaks when > 10 results AND no expand fields listed

tssweeney:
Simplify dynamic batch size calculation in calls stream to be `10` then `100` then a max of `500`

### DIFF
--- a/weave/tests/trace/test_client_trace.py
+++ b/weave/tests/trace/test_client_trace.py
@@ -2685,20 +2685,6 @@ def test_calls_stream_column_expansion_dynamic_batch_size(client, batch_size):
     for i in range(batch_size):
         assert calls[i].output == i
 
-    res = client.server.calls_query_stream(
-        tsi.CallsQueryReq(
-            project_id=client._project_id(),
-            columns=["output"],
-            expand_columns=[
-                "output.a.b.c.d.e.f.g.h.i.j.k.l.m.n.o.p.q.r.s.t.u.v.w.x.y.z"
-            ],
-        )
-    )
-    calls = list(res)
-    assert len(calls) == batch_size
-    for i in range(batch_size):
-        assert calls[i].output == i
-
 
 class Custom(weave.Object):
     val: dict
--- a/weave/tests/trace/test_client_trace.py
+++ b/weave/tests/trace/test_client_trace.py
@@ -277,9 +277,7 @@ def simple_line_call_bootstrap(init_wandb: bool = False) -> OpCallSpec:
     @weave.op()
     def multiplier(
         a: Number, b
-    ) -> (
-        int
-    ):  # intentionally deviant in returning plain int - so that we have a different type
+    ) -> int:  # intentionally deviant in returning plain int - so that we have a different type
         return a.value * b
 
     @weave.op()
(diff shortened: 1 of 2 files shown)


weave/tests/trace/test_client_trace.py
```diff
@@ -2664,6 +2664,42 @@ def return_nested_object(nested_obj: NestedObject):
...
+    assert len(calls) == batch_size
+    for i in range(batch_size):
+        assert calls[i].output == i
+
+    res = client.server.calls_query_stream(
+        tsi.CallsQueryReq(
+            project_id=client._project_id(),
+            columns=["output"],
+            expand_columns=[
+                "output.a.b.c.d.e.f.g.h.i.j.k.l.m.n.o.p.q.r.s.t.u.v.w.x.y.z"
```
andrewtruong, line 2693: I don't understand what this is for.  How does it relate to this test?

weave/tests/trace/test_client_trace.py
```diff
@@ -2664,6 +2664,42 @@ def return_nested_object(nested_obj: NestedObject):
     assert call_result.output == nested_ref.uri()
 
 
+@pytest.mark.parametrize("batch_size", [1, 11, 101, 120])
```
andrewtruong, line 2667: why these numbers?

weave/tests/trace/test_client_trace.py
gtarpenning: Hmm, probably not necessary now, the previous calculation accounted for the depth of the nested ref, so in this case it would be depth 27, which would have affected the batching. now it doesn't, but if this test were written in the first place it would have triggered the bug. Ill delete it.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -348,15 +348,8 @@ def calls_query_stream(self, req: tsi.CallsQueryReq) -> Iterator[tsi.CallSchema]
...
-                    # count the number of columns at each depth
-                    depths = Counter(col.count(".") for col in expand_columns)
-                    # take the max number of columns at any depth
-                    max_count_at_ref_depth = max(depths.values())
-                    # divide max refs that we can resolve 1000 refs at any depth
-                    max_size = 1000 // max_count_at_ref_depth
-                    # double batch size up to what refs_read_batch can handle
-                    batch_size = min(max_size, batch_size * 2)
+                    # *** Dynamic increase from 10 to 500 ***
+                    batch_size = min(500, batch_size * 10)
```
tssweeney, line 352: Let's move this to a top level constant

weave/tests/trace/test_client_trace.py
gtarpenning: this tests batch of 10, 100, and 500. 120 because what if theres an off by one error and we are just getting lucky 🧠

andrewtruong: You should add a comment -- these look like magic numbers otherwise.  I would make the 10, 100, 500 consts and re-use them here with that comment

gtarpenning: Because of how the tests run synchronously generating 500 calls actually gets pretty expensive, like 60+ seconds expensive. 120 completed in 15 seconds ish. I will add a comment

andrewtruong: yeah you dont have to do 500, just comments RE: what these numbers mean

andrewtruong (approved):

PR #2547, opened 2024-10-01
chore(weave): Add conditional ci

andrewtruong:
1. Conditionally runs tests based on which dirs have changed.
2. Updates the notify job payload to also include a signal for whether or not core integration tests should be run.  Core is always notified and the commit is always added, but tests are conditionally run based on the payload

This PR cannot be fully tested e2e until both sides are merged into master because `repository_dispatch` only applies on the master branch (of core), but I have tested the individual pieces and it seems to work

Companion to: https://github.com/wandb/core/pull/24362

### DIFF
--- a/.github/workflows/check-which-tests-to-run.yaml
+++ b/.github/workflows/check-which-tests-to-run.yaml
@@ -2,6 +2,11 @@ name: Check which tests to run
 
 on:
   workflow_call:
+    outputs:
+      weave_query_tests:
+        value: ${{ jobs.check.outputs.weave_query_tests}}
+      core_integration_tests:
+        value: ${{ jobs.check.outputs.core_integration_tests }}
 
 env:
   WEAVE_QUERY_PATHS: 'weave_query/'
--- a/.github/workflows/check-which-tests-to-run.yaml
+++ b/.github/workflows/check-which-tests-to-run.yaml
@@ -1,5 +1,8 @@
 name: Check which tests to run
 
+on:
+  workflow_call:
+
 env:
   WEAVE_QUERY_PATHS: 'weave_query/'
   CORE_INTEGRATION_PATHS: 'weave-js/ weave_query/'
--- a/.github/workflows/check-which-tests-to-run.yaml
+++ b/.github/workflows/check-which-tests-to-run.yaml
@@ -1,0 +1,46 @@
+name: Check which tests to run
+
+env:
+  WEAVE_QUERY_PATHS: 'weave_query/'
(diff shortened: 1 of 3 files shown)


.github/workflows/check-which-tests-to-run.yaml
```diff
@@ -0,0 +1,71 @@
...
+      weave_js_tests:
+        value: ${{ jobs.check.outputs.weave_js_tests }}
+      core_integration_tests:
+        value: ${{ jobs.check.outputs.core_integration_tests }}
+
+env:
+  WEAVE_QUERY_PATHS: 'weave_query/'
+  WEAVE_JS_PATHS: 'weave-js/'
+  CORE_INTEGRATION_PATHS: 'weave-js/ weave_query/'
+  # Everything else is implicitly trace server
```
andrewtruong, line 17: do we need 1 more for trace server?

tssweeney: Once we fully isolate trace sdk and trace server, we could further split this

.github/workflows/test.yaml
```diff
@@ -20,6 +23,8 @@ jobs:
     # runs-on: ubuntu-latest
     env:
       REGISTRY: us-east4-docker.pkg.dev/weave-support-367421/weave-images
+    needs: check-which-tests-to-run
+    if: needs.check-which-tests-to-run.outputs.core_integration_tests
```
tssweeney, line 27: I think this one and the next one should be `if: needs.check-which-tests-to-run.outputs.weave_query_tests`

davidwallacejackson (approved):

.github/workflows/test.yaml
```diff
@@ -12,6 +12,9 @@ on:
   push:
 
 jobs:
```
tssweeney, line 14: Can we make all the tests run on master every time for a while until we are confident in this?

.github/workflows/check-which-tests-to-run.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
```
tssweeney, line 21: Dude this is LEGIT!

.github/workflows/test.yaml
andrewtruong: yep

.github/workflows/test.yaml
andrewtruong: done

tssweeney (approved):

PR #2557, opened 2024-10-01
feat: seed cost migrations

jwlee64:
## Description
https://wandb.atlassian.net/browse/WB-20683

adds new migration
updates trace server migrator to accept an alias and different cost directory
adds a script that pulls from litellm costs, and compares to the current costs.json, and creates a migration from the diff

2700 lines are the costs.json
600 lines are tests
550 is the migrations

about ~ 300 lines of actual code

## Testing

running make migrate in core


weave/trace_server/clickhouse_trace_server_migrator.py
```diff
@@ -10,18 +10,30 @@
 
 class ClickHouseTraceServerMigrator:
     ch_client: CHClient
+    migration_dir: str
```
jwlee64, line 13: main change to this file is adding an optional migration_dir field, and passing ins a test_db_migration_alias to apply this allows us to do something like

ClickHouseTraceServerMigrator(ch_client, "costs/migrations")
ClickHouseTraceServerMigrator.apply_migrations(target_db, target_db_migration_alias="costs")

this will then create a different entry in
migrations under cost

weave/trace_server/costs/create_new_migration.py
```diff
@@ -0,0 +1,255 @@
+# This script creates a new migration for the llm_token_prices table
```
jwlee64, line 1: this is a script we would run periodically to generate a new migration for the cost table, or we can create manual migrations

weave/trace_server/costs/costs.json
tssweeney, line 1: can this file now be removed?

jwlee64: No, im using it to be the source of truth for the what the last migration updated the costs to.
so that i can run create_cost_migration and see which costs have changed without querying the db, I could just query the db though, and delete the file

tssweeney: Oh i see - maybe we can call it that? like `last_cost_generation_checkpoint.json`

weave/trace_server/clickhouse_trace_server_migrator.py
tssweeney: It seems like a lot of this complexity could be avoided if we simply add the cost migrations to the existing migration directory. Then we don't need to modify any of the migration related scripts and there is a single true version for an entire DB

jwlee64: yah that makes sense, I just didnt want to bloat our migration directory with constant cost updates, since i figured we'd want to make frequent cost migrations

PR #2558, opened 2024-10-01
perf: Improve Table Query Performance

tssweeney:
10k row dataset, 10kb per row.

- (offset 200, limit 100)
  - Prod: Elapsed: 1.439s Read: 33,970 rows (225.50 MB)
  - _Master: Elapsed: 5.691s Read: 4,481,567 rows (18.02 GB)_
  - **PR: Elapsed: 0.098s Read: 34,531 rows (225.53 MB)**
- Filtered by Digest
  - Prod: Elapsed: 0.163s Read: 31,321 rows (225.39 MB)
  - _Master: Elapsed: 7.913s Read: 4,481,939 rows (18.02 GB)_
  - **PR: Elapsed: 0.052s Read: 24,222 rows (20.48 MB)**

### DIFF
--- a/weave/trace_server/table_query_builder.py
+++ b/weave/trace_server/table_query_builder.py
@@ -34,7 +34,6 @@ def make_natural_sort_table_query(
             row_digests_selection = f"arraySlice({row_digests_selection}, 1 + {{{pb.add_param(offset)}: Int64}})"
         else:
             row_digests_selection = f"arraySlice({row_digests_selection}, 1 + {{{pb.add_param(offset)}: Int64}}, {{{pb.add_param(limit)}: Int64}})"
-    
 
     query = f"""
     SELECT DISTINCT tr.digest, tr.val_dump, t.row_order
--- a/weave/trace_server/table_query_builder.py
+++ b/weave/trace_server/table_query_builder.py
@@ -23,30 +23,34 @@ def make_natural_sort_table_query(
     """
     project_id_name = pb.add_param(project_id)
     digest_name = pb.add_param(digest)
-    sql_safe_dir = "ASC" if natural_direction == "ASC" else "DESC"
 
-    sql_safe_limit = (
-        f"LIMIT {{{pb.add_param(limit)}: Int64}}" if limit is not None else ""
-    )
-    sql_safe_offset = (
-        f"OFFSET {{{pb.add_param(offset)}: Int64}}" if offset is not None else ""
-    )
+    row_digests_selection = "row_digests"
+    if natural_direction.lower() == "desc":
+        row_digests_selection = f"reverse({row_digests_selection})"
+    if limit is not None and offset is None:
+        offset = 0
+    if offset is not None:
+        if limit is None:
+            row_digests_selection = f"arraySlice({row_digests_selection}, 1 + {{{pb.add_param(offset)}: Int64}})"
+        else:
+            row_digests_selection = f"arraySlice({row_digests_selection}, 1 + {{{pb.add_param(offset)}: Int64}}, {{{pb.add_param(limit)}: Int64}})"
+    
 
     query = f"""
     SELECT DISTINCT tr.digest, tr.val_dump, t.row_order
     FROM table_rows tr
-    RIGHT JOIN (
+    INNER JOIN (
         SELECT row_digest, row_number() OVER () AS row_order
-        FROM tables
+        FROM (
+            SELECT {row_digests_selection} as row_digests
+            FROM tables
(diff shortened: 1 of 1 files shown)


weave/trace_server/table_query_builder.py
```diff
@@ -23,30 +23,33 @@ def make_natural_sort_table_query(
...
+        if limit is None:
+            row_digests_selection = f"arraySlice({row_digests_selection}, 1 + {{{pb.add_param(offset)}: Int64}})"
+        else:
+            row_digests_selection = f"arraySlice({row_digests_selection}, 1 + {{{pb.add_param(offset)}: Int64}}, {{{pb.add_param(limit)}: Int64}})"
 
     query = f"""
     SELECT DISTINCT tr.digest, tr.val_dump, t.row_order
     FROM table_rows tr
-    RIGHT JOIN (
+    INNER JOIN (
```
tssweeney, line 41: This is one key to the solution (INNER vs RIGHT)

weave/trace_server/table_query_builder.py
```diff
@@ -23,30 +23,33 @@ def make_natural_sort_table_query(
...
         SELECT row_digest, row_number() OVER () AS row_order
-        FROM tables
+        FROM (
+            SELECT {row_digests_selection} as row_digests
```
tssweeney, line 44: Second insight - we can do the offset/limit inside the sub selection avoiding uneeded joins

weave/trace_server/table_query_builder.py
```diff
@@ -23,30 +23,33 @@ def make_natural_sort_table_query(
...
+            FROM tables
+            WHERE project_id = {{{project_id_name}: String}}
```
tssweeney, line 46: 3rd insight - do the project/digest filter before the ARRAY JOIN

weave/trace_server/table_query_builder.py
```diff
@@ -88,12 +91,15 @@ def make_standard_table_query(
     (
         SELECT DISTINCT tr.digest, tr.val_dump, t.row_order
         FROM table_rows tr
-        RIGHT JOIN (
+        INNER JOIN (
             SELECT row_digest, row_number() OVER () AS row_order
-            FROM tables
+            FROM (
```
tssweeney, line 96: Similar changes here

weave/trace_server/table_query_builder.py
```diff
@@ -23,30 +23,33 @@ def make_natural_sort_table_query(
...
         ARRAY JOIN row_digests AS row_digest
-        WHERE project_id = {{{project_id_name}: String}}
-        AND digest = {{{digest_name}: String}}
-        ORDER BY row_order {sql_safe_dir}
-        {sql_safe_limit}
-        {sql_safe_offset}
     ) AS t ON tr.digest = t.row_digest
     WHERE tr.project_id = {{{project_id_name}: String}}
-    ORDER BY row_order {sql_safe_dir}
+    ORDER BY row_order ASC
```
tssweeney, line 52: Always ASC now that the order is handled by potential reversal in digest selection

weave/trace_server/table_query_builder.py
gtarpenning: `INNER` is intersection, `RIGHT` is all of the second table + intersection. We can get away with using just the intersection? Intuitively this will now be fewer rows.

weave/trace_server/table_query_builder.py
gtarpenning: does this affect `row_number() OVER () AS row_order`? Do we get the correct row number if we specify an offset?

gtarpenning (approved):
I think this makes sense, asked a couple clarifying questions just to make sure.

weave/trace_server/table_query_builder.py
gtarpenning: why do we even use `row_order`? can we remove?

gtarpenning: oh its being used to dedupe

weave/trace_server/table_query_builder.py
tssweeney: Yes, the edge case here is if we accidentally deleted a row from the table_rows that was referenced in the table itself. This would be a bad error. However, most of the perf in this PR comes from this `INNER` clause...

weave/trace_server/table_query_builder.py
tssweeney: Ahhh - row number is not actually the row number of the table, but row number of the query (in this case)

PR #2560, opened 2024-10-01
chore: Remove jupyter pre-commit

tssweeney:
This is legacy


andrewtruong (commented):
We still have notebooks in trace that could be formatted.  Are you sure you want to remove?

PR #2563, opened 2024-10-01
fix: Better filtering by ref values

tssweeney:
Filtering by values within a reffed object is technically quite hard. However, Scott made the great recommendation today to have the following rules:
1. Don't allow the user to provide free-text input to columns which are expanded columns
  * As a result, this means that users will not be able to accidentally filter by something that isn't supported
2. If a user option-clicks a field, use the parent ref as the filter, rather than trying to filter by the value
  * As a result, users will still feel like they can effectively filter the results. Note: if two distinct objects have the same value for a given property, the filter will only produce rows where the ref corresponds to the clicked cell. This is likely still going to be a friction point.

This Pr does those two things (:


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -268,14 +271,78 @@ export const CallsTable: FC<{
...
+  // This is a specific helper that is used when the user attempts to option-click
+  // a cell that is a child cell of an expanded ref. In this case, we want to
+  // add a filter on the parent ref itself, not the child cell. Once we can properly
+  // filter by reffed values on the backend, this can be removed.
+  const getFieldAndValueForRefExpandedFilter = useCallback(
+    (field: string, rowId: string) => {
+      if (columnIsRefExpanded(field)) {
+        // In this case, we actually just want to filter by the parent ref itself.
+        // This means we need to:
+        // 1. Determine the column of the highest level anscestor column with a ref
```
jamie-rasmussen, line 288: anscestor => ancestor

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -268,14 +271,78 @@ export const CallsTable: FC<{
...
+        // 2. Get the value of that corresponding cell (ref column @ row)
+        // 3. Add a filter for that ref on that column.
+        // The aknoweldged drawback of this approach is that we are not filtering by that
```
jamie-rasmussen, line 291: aknoweldged => acknowledged

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -268,14 +271,78 @@ export const CallsTable: FC<{
...
+        // cell's value, but rather the entire object itself. This still might confuse users,
+        // but is better than returning nothing.
+        const fieldParts = field.split('.');
+        let ancestorField: string | null = null;
+        let targetRef: string | null = null;
+        for (let i = 0; i <= fieldParts.length; i++) {
+          const ancestorFieldCandidate = fieldParts.slice(0, i).join('.');
```
jamie-rasmussen, line 298: `slice(0,0)` will be an empty list, I think we don't need to check that case?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -306,6 +373,18 @@ export const CallsTable: FC<{
     onAddFilter
   );
 
+  // This contains columns which are suitable for selection and raw data
+  // entry. Noteably, not children of expanded refs.
```
jamie-rasmussen, line 377: Noteably => Notably

jamie-rasmussen (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
tssweeney: 👍

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
tssweeney: 👍

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
tssweeney: 👍

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
tssweeney: 👍

PR #2581, opened 2024-10-02
Griffin/objs delete with table delete

gtarpenning:
https://github.com/wandb/weave/pull/2319 + table deletion


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -642,6 +690,53 @@ def objs_query(self, req: tsi.ObjQueryReq) -> tsi.ObjQueryRes:
...
+            req.object_id,
+            req.digest,
+            # If purging, don't read the value from the db
+            metadata_only=bool(req.permanently_delete),
+        )
+
+        # Check if any of the object refs are tables, and delete them if
+        # permenantly_delete is set
+        if db_obj.refs:
+            for raw_ref in db_obj.refs:
```
gtarpenning, line 712: super rogue, but would something like this work? @tssweeney

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -872,6 +967,49 @@ def table_query_stats(self, req: tsi.TableQueryStatsReq) -> tsi.TableQueryStatsR
...
+        WHERE project_id = {project_id:String} AND digest IN {row_digests:Array(String)}
+        """
+        parameters = {
+            "project_id": project_id,
+            "row_digests": row_digests,
+        }
+        self.ch_client.query(delete_query, parameters=parameters)
+
+        # 3. delete the table by digest
+        delete_table_query = """
```
gtarpenning, line 1003: could also leave this, as a memory of the deleted table. tables with zero row_digests are considered deleted?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -872,6 +967,49 @@ def table_query_stats(self, req: tsi.TableQueryStatsReq) -> tsi.TableQueryStatsR
 
         return tsi.TableQueryStatsRes(count=count)
 
+    def _table_delete(self, project_id: str, digest: str) -> None:
```
tssweeney, line 970: This is going to need to be smarter in that it only deletes rows once they no longer are used by any other table

gtarpenning: Ugh right....

PR #2599, opened 2024-10-03
chore(weave): Remove old decorator overloads

andrewtruong:
Removes old decorator overloads and makes a much nicer op decorator!

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -22,7 +22,7 @@ from weave.trace.object_record import (
     dataclass_object_record,
     pydantic_object_record,
 )
-from weave.trace.op import Op, is_op, maybe_unbind_method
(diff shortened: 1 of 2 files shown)


weave/trace/weave_client.py
```diff
@@ -1395,6 +1395,7 @@ def op_fn(*args, **kwargs):  # type: ignore
 
     op_fn.__name__ = name
     op = op_deco(op_fn)
+    op = as_op(op)
```
gtarpenning, line 1398: what does this do?

gtarpenning (approved):
lets let this soak for a bit before deploying the sdk, hot path and all.

weave/trace/weave_client.py
andrewtruong: it's casting to the `Op` type

PR #2609, opened 2024-10-04
chore(query): Update .gitignore

andrewtruong:


gtarpenning (approved):

.gitignore
```diff
@@ -3,31 +3,20 @@
...
 gha-creds-*.json
-.vscode/launch.json
+.vscode
 .mypy_cache
 .hypothesis
 .env
-yarn-error.log*
-config/.admin.env
-config/.wb_api.env
-weave/tests/zlib.txt
```
gtarpenning, line 26: so much nicer

gtarpenning: but just to be clear, this is gone?

gtarpenning: what if i already have an admin.env file in that folder, will this now appear in git?

andrewtruong: We don't have a `config` dir anymore -- do you need it?

PR #2076, opened 2024-08-06
feat: Models, Datasets, and Custom Objects now show properties as columns when viewing all versions

tssweeney:
Before:
Screenshot 2024-08-05 at 19 23 59

After:
Screenshot 2024-08-05 at 19 23 29

Remaining Todos (followups):
* [ ] Add the ability to expand refs
* [ ] Paginate & stream responses similar to calls
* [ ] Add the ability to sort / filter on values
* [ ] Add the ability to sort / filter on expanded values (blocked by general support for expansion operations)
* [ ] Add sort / filter state to URL

PR #2612, opened 2024-10-05
fix: broken obj grid links

jamie-rasmussen:
## Description

Fixes https://wandb.atlassian.net/browse/WB-21208

https://github.com/wandb/weave/pull/2076 added display of object properties as columns in the grid when viewing all versions. Unfortunately, it is possible that a property name conflicts with the field IDs that are used for the non-dynamic columns, and this was happening for the first column, "object" in this case.

There's certainly a more robust fix possible here, did a quick one in the interests of time.

In the before image, note how the first column is getting overwritten by the "object" column definition from the sixth column due to the field ID conflict. This also generates a bunch of developer console errors.

Before:
Screenshot 2024-10-04 at 10 12 29 PM

After:
Screenshot 2024-10-04 at 10 12 47 PM

## Testing

How was this PR tested?


gtarpenning (approved):
Aha! Sorry for the red herring, nice fix.

PR #2619, opened 2024-10-08
fix: Improve TableQuery performance by adding a streaming interface

tssweeney:
In order to support larger datasets, we need to intrude a streaming interface similar to calls that allow us to stream results from the server as opposed to sending the entire payload over the wire at once. This follows the exact same pattern as the Calls Query [Stream] methods.

The existing table_query path uses the new `table_query_stream` path which ensure that all existing tests for table_query are also exercising the new path.


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -770,6 +770,12 @@ def add_new_row_needed_to_insert(row_data: Any) -> str:
         return tsi.TableUpdateRes(digest=digest, updated_row_digests=updated_digests)
 
     def table_query(self, req: tsi.TableQueryReq) -> tsi.TableQueryRes:
+        rows = list(self.table_query_stream(req))
```
tssweeney, line 773: By using the stream method here, we effectively share all the same code and existing tests

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -439,6 +441,14 @@ def table_query(
             "/table/query", req, tsi.TableQueryReq, tsi.TableQueryRes
         )
 
+    def table_query_stream(
+        self, req: tsi.TableQueryReq
+    ) -> Iterator[tsi.TableRowSchema]:
+        # Need to manually iterate over this until the stram endpoint is built and shipped.
```
jwlee64, line 447: typo: stram -> stream

jwlee64 (approved):
lgtm

PR #2622, opened 2024-10-08
chore(weave): Move shared trace server components into `trace_server_common` and exclude `trace_server` from dist

andrewtruong:
1. Move files under `trace_server` that are used by the client into `trace_server_common`
2. `trace_server` is excluded from the build

Companion: https://github.com/wandb/core/pull/24677


weave/trace_server_common/requests.py
tssweeney, line 1: i don't think this is used anywhere in trace_server itself and should just be moved near `remote_http_trace_server`

weave/trace/weave_init.py
```diff
@@ -185,6 +184,8 @@ def init_weave_get_server(
 
 
 def init_local() -> InitializedClient:
+    from weave.trace_server import sqlite_trace_server
```
tssweeney, line 187: red flag: did we introduce an import loop?

weave/trace_server_common/async_batch_processor.py
tssweeney, line 1: this is only used in `remote_http_trace_server` - suggest removing from common and putting there

tssweeney (commented):
This is also going to require a core change to update the `trace_server_code` make command to include this common dir

weave/trace/weave_init.py
andrewtruong: not afaik

andrewtruong:
Added core companion

weave/trace/weave_init.py
tssweeney: then why move here?

tssweeney (approved):

weave/trace/weave_init.py
andrewtruong: This file should not depend on `trace_server` at all since the contents of that dir will move into `core` in future.  This is a step in that direction.

Also, afaik the `init_local` and `init_local_client` are not actually used by end-users atm.  It's just there for internal use against sqlite

PR #2625, opened 2024-10-08
chore(weave): Seed cost table

jwlee64:
## Description
Seeds cost table on migration

adds a script for devs to run to update cost_checkpoint.json
adds script that runs on migration to add any updated costs in cost_checkpoint.json to the db

## Testing

unit testing and manually running make migrate_local


weave/trace_server/clickhouse_trace_server_migrator.py
```diff
@@ -35,12 +37,16 @@ def apply_migrations(
         )
         if len(migrations_to_apply) == 0:
             logger.info(f"No migrations to apply to `{target_db}`")
+            if should_insert_costs(target_version):
```
tssweeney, line 40: we should probably not do such migrations if we are migrating down

weave/trace_server/costs/update_costs.py
```diff
@@ -0,0 +1,131 @@
+# This script populates costs.json with the latest costs from litellm
+# We store up to 3 historical costs for each model
+import json
```
tssweeney, line 3: should the execution of this be added to a makefile command?

weave/trace_server/costs/insert_costs.py
```diff
@@ -0,0 +1,187 @@
...
+import os
+import uuid
+from datetime import datetime, timezone
+from typing import Dict, List, Optional, Tuple, TypedDict
+
+from clickhouse_connect.driver.client import Client
+
+COST_FILE = "cost_checkpoint.json"
+
+logger = logging.getLogger(__name__)
```
tssweeney, line 17: might want to make the default log level info for this logger

weave/trace_server/costs/update_costs.py
```diff
@@ -0,0 +1,131 @@
...
+        new_costs = fetch_new_costs()
+    except Exception as e:
+        print("Failed to fetch new costs:", e)
+        return
+
+    for k, v in new_costs.items():
+        if k not in costs:
+            costs[k] = [v]
+        elif costs[k] and (
+            costs[k][-1]["input"] != v["input"] or costs[k][-1]["output"] != v["output"]
```
tssweeney, line 110: might want to make a small note that this avoids updating a previous date if it already has the same entry

weave/trace_server/costs/cost_checkpoint.json
```diff
@@ -0,0 +1,4386 @@
+{
```
tssweeney, line 1: Curious how your query logic works. lets say i have a call from time T, and a cost created at time T+1. If there are no other costs registered for that model/provider pair, do i use the "future" cost?

If the answer is "no", we might want to backfill this a bit.

weave/trace_server/clickhouse_trace_server_migrator.py
jwlee64: I added a check to make sure we are not migrating down

tssweeney (approved):

PR #2628, opened 2024-10-08
chore(ui): Adjust weave0 timestamp resolver and documentation for parity

domphan-wandb:
## Description
[Internal Jira [WB-19303]](https://wandb.atlassian.net/browse/WB-19303)

This PR adds a docstring stating the expected time unit for the to_timestamp op.

[WB-19303]: https://wandb.atlassian.net/browse/WB-19303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/core/ops/primitives/number.ts
+++ b/weave-js/src/core/ops/primitives/number.ts
@@ -655,37 +655,33 @@ export const opNumberCos = OpKinds.makeStandardOp({
   resolver: ({n}) => Math.cos(n),
 });
 
-const timestampSecondUpperBound = 60 * 60 * 24 * 365 * 1000; // first 1000 years
-const timestampMilliSecondUpperBound = timestampSecondUpperBound * 1000;
-const timestampMicroSecondUpperBound = timestampMilliSecondUpperBound * 1000;
-const timestampNanoSecondUpperBound = timestampMicroSecondUpperBound * 1000;
+// Use min and max timestamps from python builtin datetime library
+// for consistency between weave0 and weave_query resolver imeplemntation
+const PY_DATETIME_MAX_MS = 50000000000000;
+const PY_DATETIME_MIN_MS = -50000000000000;
 
 // we will start by making this a simple millisecond converter, but
 // in the future we can make the unit customizable.
 export const opNumberToTimestamp = OpKinds.makeStandardOp({
   name: 'number-toTimestamp',
   argTypes: {val: 'number'},
-  description: `Converts a ${docType('number')} to a ${docType(
-    'timestamp'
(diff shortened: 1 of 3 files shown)


weave-js/src/core/ops/primitives/number.ts
```diff
@@ -665,10 +665,12 @@ const timestampNanoSecondUpperBound = timestampMicroSecondUpperBound * 1000;
...
-  description: `Converts a ${docType('number')} to a ${docType(
-    'timestamp'
-  )}. Values less than ${timestampSecondUpperBound} will be converted to seconds, values less than ${timestampMilliSecondUpperBound} will be converted to milliseconds, values less than ${timestampMicroSecondUpperBound} will be converted to microseconds, and values less than ${timestampNanoSecondUpperBound} will be converted to nanoseconds.`,
-  argDescriptions: {val: 'Number to convert to a timestamp'},
+  description: `Converts a ${docType(
+    'number'
+  )} representing unix time in milliseconds to a ${docType('timestamp')}.`,
+  argDescriptions: {
+    val: 'Number (unix time in milliseconds) to convert to a timestamp',
+  },
```
dannygoldstein, line 673: based on the resolver, 

```typescript
  resolver: ({val}) => {
    if (val < timestampSecondUpperBound) {
      return Math.floor(val * 1000);
    } else if (val < timestampMilliSecondUpperBound) {
      return Math.floor(val);
    } else if (val < timestampMicroSecondUpperBound) {
      return Math.floor(val / 1000);
    } else if (val < timestampNanoSecondUpperBound) {
      return Math.floor(val / 1000 / 1000);
    } else {
      return null;
    }
```

it looks like the original doc string is correct

domphan-wandb: Gotcha. According to the weave_query timestamp function, it looks like its implementation deviates intentionally from weave0. https://github.com/wandb/weave/blob/26a8a6560cf1e0e15ce08e357489658551d3a126/weave_query/weave_query/timestamp.py#L69-L75

Specifically, it does not handle the conversion from seconds to timestamp.

Should I remove that seconds conversion from the js resolver?

I'll tweak the docstring back to the original, but leave out this part `Values less than ${timestampSecondUpperBound} will be converted to seconds`

dannygoldstein: I think it is safe to update the JS resolver to match the while loop in the weave1 resolver, and then update the JS docstring. The blast radius for that change is low, because weave0 resolvers are seldom used anymore.

dannygoldstein (commented):
commenting to move this out of my graphite inbox

dannygoldstein (approved):

jwlee64 (approved):

PR #2633, opened 2024-10-08
feat: add finish handler for ops

jamie-rasmussen:
## Description

Adds the ability to set a "finish" handler on an Op. The intended use is as an extension point for integrations. For example, I intend to use this functionality in the Gemini integration to copy information about token usage from the Gemini call response object (whose structure only the Gemini integration should have to know anything about) into our standard token usage format in the call's summary dictionary.

## Testing

How was this PR tested?

### DIFF
--- a/weave/trace/op.py
+++ b/weave/trace/op.py
@@ -82,6 +82,8 @@ def print_call_link(call: "Call") -> None:
 
 FinishCallbackType = Callable[[Any, Optional[BaseException]], None]
 OnOutputHandlerType = Callable[[Any, FinishCallbackType, Dict], Any]
+# Call, original function output, exception if occurred
+OnFinishHandlerType = Callable[["Call", Any, Optional[BaseException]], None]
 
 
 def value_is_sentinel(param: Any) -> bool:
@@ -149,6 +151,9 @@ class Op(Protocol):
     _set_on_output_handler: Callable[[OnOutputHandlerType], None]
     _on_output_handler: Optional[OnOutputHandlerType]
 
+    _set_on_finish_handler: Callable[[OnFinishHandlerType], None]
+    _on_finish_handler: Optional[OnFinishHandlerType]
+
     __call__: Callable[..., Any]
     __self__: Any
 
@@ -166,6 +171,12 @@ def _set_on_output_handler(func: Op, on_output: OnOutputHandlerType) -> None:
     if func._on_output_handler is not None:
         raise ValueError("Cannot set on_output_handler multiple times")
     func._on_output_handler = on_output
+
+
+def _set_on_finish_handler(func: Op, on_finish: OnFinishHandlerType) -> None:
+    if func._on_finish_handler is not None:
+        raise ValueError("Cannot set on_finish_handler multiple times")
+    func._on_finish_handler = on_finish
 
 
 def _is_unbound_method(func: Callable) -> bool:
@@ -232,6 +243,7 @@ def _execute_call(
             output,
             exception,
             postprocess_output=__op.postprocess_output,
+            op=__op,
         )
         if not call_context.get_current_call():
             print_call_link(call)
@@ -496,6 +508,9 @@ def op(
 
             wrapper._set_on_output_handler = partial(_set_on_output_handler, wrapper)  # type: ignore
             wrapper._on_output_handler = None  # type: ignore
+
+            wrapper._set_on_finish_handler = partial(_set_on_finish_handler, wrapper)  # type: ignore
+            wrapper._on_finish_handler = None  # type: ignore
 
             wrapper._tracing_enabled = True  # type: ignore
 
--- a/weave/trace/op.py
+++ b/weave/trace/op.py
@@ -242,7 +242,6 @@ def _execute_call(
             call,
             output,
             exception,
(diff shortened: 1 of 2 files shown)


weave/trace/op.py
```diff
@@ -232,6 +243,7 @@ def finish(output: Any = None, exception: Optional[BaseException] = None) -> Non
             output,
             exception,
             postprocess_output=__op.postprocess_output,
+            op=__op,
```
jamie-rasmussen, line 246: Can probably get rid of the `postprocess_output` if we're passing the entire `__op`

weave/trace/weave_client.py
```diff
@@ -722,8 +723,7 @@ def finish_call(
             postprocessed_output = original_output
         self._save_nested_objects(postprocessed_output)
 
-        output = map_to_refs(postprocessed_output)
-        call.output = output
+        call.output = map_to_refs(postprocessed_output)
```
jamie-rasmussen, line 726: I found it a little confusing to be redefining the `output` parameter to `finish_call`. Most code uses the `original_output` alias, small update for this change below.

weave/trace/weave_client.py
```diff
@@ -749,6 +749,7 @@ def finish_call(
             if isinstance(usage, dict) and isinstance(model, str):
                 summary["usage"] = {}
                 summary["usage"][model] = {"requests": 1, **usage}
+        call.summary = summary
```
jamie-rasmussen, line 752: Moving summary assignment from below to keep that logic together.

weave/trace/weave_client.py
andrewtruong: makes sense

weave/trace/weave_client.py
andrewtruong: makes sense

weave/trace/op.py
andrewtruong: why do you need to pass the whole `op`?

weave/trace/weave_client.py
```diff
@@ -776,13 +782,6 @@ def send_end_call() -> None:
 
         self.future_executor.defer(send_end_call)
 
-        # Descendent error tracking disabled til we fix UI
-        # Add this call's summary after logging the call, so that only
-        # descendents are included in what we log
-        # summary.setdefault("descendants", {}).setdefault(
-        #     call.op_name, {"successes": 0, "errors": 0}
-        # )["successes"] += 1
```
andrewtruong, line 784: is this fixed?

andrewtruong:
This is a step towards lifecycle callbacks for the op!

weave/trace/weave_client.py
jamie-rasmussen: No, this code has been commented out since it entered the code base six months ago in https://github.com/wandb/weave/pull/1414 

I could leave it in if you'd prefer.

weave/trace/op.py
jamie-rasmussen: Just seemed a little cleaner to pass in the Op object rather than to have separate arguments for the handlers.

weave/trace/weave_client.py
andrewtruong: @tssweeney do we need this?

weave/trace/op.py
andrewtruong: ok I would pick one -- if you want to just pass the `op` that's fine, but then we should take out `postprocess_output` for consistency

weave/trace/weave_client.py
jamie-rasmussen: There's a screenshot in the PR I linked above that illustrates why this never made it in.

There's a redesign for the Summary tab in the queue, so if we want to revive this it's probably a good time.
https://www.notion.so/wandbai/Trace-drawer-items-43c3106f7d2848758d02f89cee2b3540?pvs=4#10ae2f5c7ef3808e8e5de7424dfa0584

We have a long standing issue where we try to calculate the existence of descendent errors in the UI to show a different status pill, but we don't do it in a consistent way. If we were going to log this kind of stat in the summary we could use it for that. I think it should be scoped under the "weave" key if we do that.

weave/trace/op.py
```diff
@@ -82,6 +82,8 @@ def print_call_link(call: "Call") -> None:
 
 FinishCallbackType = Callable[[Any, Optional[BaseException]], None]
 OnOutputHandlerType = Callable[[Any, FinishCallbackType, Dict], Any]
+# Call, original function output, exception if occurred
+OnFinishHandlerType = Callable[["Call", Any, Optional[BaseException]], None]
```
tssweeney, line 86: Curious why `_on_output_handler` was insufficient here?

jamie-rasmussen: It only gets the output value, the finish handler needs the whole call - in the Gemini case the handler is used to modify the summary object, and what it needs to set in the summary depends on parsing information (the model name) out of the inputs.

weave/trace/weave_client.py
andrewtruong: Ok I would keep this comment and also include reference to your doc future readers' clarity

jamie-rasmussen: I can do that.

PR #2636, opened 2024-10-08
chore(weave): Add materialized views that will make it easier to do billing calculations

adrnswanberg:
## Description

Add materialized views that provide usage stats for our major tables.

## Testing

How was this PR tested?
It hasn't, yet :)


weave/trace_server/migrations/007_stats_views.down.sql
```diff
@@ -0,0 +1,11 @@
+DROP TABLE IF EXISTS weave_trace_db.calls_merged_stats;
```
tssweeney, line 1: calls merged is one of the few that we don't need actually

adrnswanberg: still helpful to make counting up usage dramatically faster

tssweeney: fair enough - any reason you did not do the other tables?

weave/trace_server/migrations/007_stats_views.down.sql
```diff
@@ -0,0 +1,11 @@
...
+DROP VIEW IF EXISTS weave_trace_db.calls_merged_stats_view;
+
+DROP TABLE IF EXISTS weave_trace_db.object_versions_stats;
+DROP VIEW IF EXISTS weave_trace_db.object_versions_stats_view;
+
+DROP TABLE IF EXISTS weave_trace_db.table_rows_stats;
+DROP VIEW IF EXISTS weave_trace_db.table_rows_stats_view;
+
+DROP TABLE IF EXISTS weave_trace_db.files_stats;
+DROP VIEW IF EXISTS weave_trace_db.files_stats_view;
```
tssweeney, line 11: Note: you don't actually need the `weave_trace_db` prefix here, or anywhere as the DB is a param of the entire migration process

weave/trace_server/migrations/007_stats_views.up.sql
```diff
@@ -0,0 +1,138 @@
...
+(
+    project_id String,
+    id String,
+    trace_id SimpleAggregateFunction(any, Nullable(String)),
+    parent_id SimpleAggregateFunction(any, Nullable(String)),
+    op_name SimpleAggregateFunction(any, Nullable(String)),
+    started_at SimpleAggregateFunction(any, Nullable(DateTime64(3))),
+    attributes_size_bytes SimpleAggregateFunction(any, Nullable(UInt64)),
+    inputs_size_bytes SimpleAggregateFunction(any, Nullable(UInt64)),
+    input_refs SimpleAggregateFunction(groupArrayArray, Array(String)),
```
tssweeney, line 99: not sure if you wanted to trim down the fields, but you could probably drop many of these if we just want the size stats.

weave/trace_server/migrations/007_stats_views.up.sql
```diff
@@ -0,0 +1,138 @@
...
+CREATE MATERIALIZED VIEW object_versions_stats_view
+TO weave_trace_db.object_versions_stats
+AS
+SELECT
+    project_id,
+    kind,
+    object_id,
+    digest,
+    anySimpleState(base_object_class) AS base_object_class,
+    anySimpleState(refs) AS refs,
```
tssweeney, line 81: similar here, do we really need these other fields?

weave/trace_server/migrations/007_stats_views.down.sql
adrnswanberg: you mean like tables? we only need this for tables we bill on. and besides feedback, no other tables have user-defined fields of arbitrary length

tssweeney: feedback does, but i guess we don't bill that right now

adrnswanberg: added one for feedback

weave/trace_server/migrations/007_stats_views.down.sql
adrnswanberg: fixed

weave/trace_server/migrations/007_stats_views.up.sql
adrnswanberg: felt easier to include them..they might come in handy for quick analysis? we can always drop them later. easier to drop than include later as backfills are hard.

adrnswanberg:
@tssweeney Finally got around to testing this -- all working ok.

tssweeney (approved):

tssweeney (changes requested):
Please bump the migration number (requesting changes to make sure we don't forget)

PR #2637, opened 2024-10-08
chore(ui): make eval section header draggable

gtarpenning:
## Description

- Fixes [WB-21422](https://wandb.atlassian.net/browse/WB-21422)

Drag around the Eval buttons, to set the order of how the evaluations appear in the evaluation compare page. 

draggable-not-stupid

[WB-21422]: https://wandb.atlassian.net/browse/WB-21422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/compareEvaluationsContext.tsx
```diff
@@ -83,30 +77,32 @@ export const CompareEvaluationsProvider: React.FC<{
...
       state: initialState.result,
-      setBaselineEvaluationCallId,
+      setSelectedCallIdsOrdered,
       setComparisonDimensions,
       setSelectedInputDigest,
       setSelectedMetrics,
       addEvaluationCall: (newCallId: string) => {
         const newEvaluationCallIds = [...evaluationCallIds, newCallId];
         setEvaluationCallIds(newEvaluationCallIds);
+        setSelectedCallIdsOrdered(newEvaluationCallIds);
```
gtarpenning, line 87: keep these separate, so that when you reorder them we don't re-query the db. We can also extend this to not re-query when an evaluation is removed, but this is complicated and out of scope for v0.

tssweeney: seems a little weird to have both. i would instead create a helper hook that can take a list of ids to fetch which only reruns when the set changes. For example, you could re-order the list passed by the caller alphabetically, query, then return back in the original order. This way the component does not need to worry about maintaining two lists

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/ecpState.ts
```diff
@@ -117,17 +115,21 @@ export const useEvaluationComparisonState = (
  * evaluation call is first.
  */
 export const getOrderedCallIds = (state: EvaluationComparisonState) => {
-  const initial = Object.keys(state.data.evaluationCalls);
-  moveItemToFront(initial, state.baselineEvaluationCallId);
-  return initial;
+  return state.selectedCallIdsOrdered.filter(
+    callId => state.data.evaluationCalls[callId] != null
```
tssweeney, line 119: this smells - hacky way to enforce deletion on your ordered list

tssweeney (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/shoppingCart/ShoppingCart.tsx
```diff
@@ -7,9 +7,9 @@ import React from 'react';
 import {useHistory} from 'react-router-dom';
 import {SortableContainer} from 'react-sortable-hoc';
 
-import {queryToggleString, searchParamsSetArray} from '../urlQueryUtil';
+import {ShoppingCartItemDefs} from '../../../compare/types';
```
jamie-rasmussen, line 10: Seems like this shouldn't be reaching back into compare.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ComparisonDefinitionSection/ComparisonDefinitionSection.tsx
```diff
@@ -222,12 +261,26 @@ const AddEvaluationButton: React.FC<{
...
                     className="pb-8 pt-8 font-['Source_Sans_Pro'] text-base font-normal text-moon-800"
-                    onClick={() => {
-                      addEvaluationCall(call.callId);
-                    }}>
+                    onClick={() => addEvaluationCall(call.callId)}>
                     <>
-                      <span>{call.displayName ?? call.spanName}</span>
-                      <Id id={call.callId} type="Call" className="ml-0 mr-4" />
+                      <span
+                        style={{
```
jamie-rasmussen, line 267: I think it would be nice if we start using Tailwind for stuff like this but no need to change now.

jamie-rasmussen (commented):
I'm noticing that the shopping cart items have the "Click to select this item" tooltip / cursor / hover effect but it appears you can't actually click to do the subselection as you can in the generic comparison view.

I also think we still want the colored circle and the id value to show in the item?

jamie-rasmussen (approved):

PR #2638, opened 2024-10-09
fix(weave_query): Handle paths with colons when escaping artifact path

nicholaspun-wandb:
## Description

JIRA: [WB-21299](https://wandb.atlassian.net/browse/WB-21299)

When the artifact URI does not have a version, but the path has a colon (`:`), our current implementation to find the path will mistakenly treat part of it as the version string. Instead of splitting on the version delimiter (`:`) and then trying to find the path, we can reverse the order and split on the path delimiter (`/`) first, allowing paths to include `:`.

The PR also updates the tests to include the following cases:
* Standard artifact path: `<name>:<version>/<path>`
* No version and path does not have `:`
* With version and path has `:`
* No version and path has `:` (this is the case we were previously erroring on)

## Testing

* Added unit tests 
* Verified that `test_zlib_playback` now passes

[WB-21299]: https://wandb.atlassian.net/browse/WB-21299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave_query/tests/test_wb.py
+++ b/weave_query/tests/test_wb.py
@@ -1776,16 +1776,66 @@ def test_is_valid_version_string():
     for v in ["v01", "v0009"]:
         assert not artifact_wandb.is_valid_version_index(v)
 
+class TestEscapeArtifactPath:
+    wandb_prefix = "wandb-client-artifact://"
+    artifact_name = "12347187287418787843872388177814"
+    file_path = "table #3.table.json"
+    file_path_with_colon = "(batch: 1) table #3.table.json"
+    version = "latest"
+    # version_with_colon = "latest:v0"
 
-def test_artifact_path_character_escaping():
-    name = 12347187287418787843872388177814
-    path = "table #3.table.json"
-    result = wb_util.escape_artifact_path(
-        f"wandb-client-artifact://{name}:latest/{path}"
-    )
-    uri = artifact_wandb.WeaveWBLoggedArtifactURI.parse(result)
+    def test_standard_artifact_path(self):
+        """
+        Artifact path of the form: "<name>:<version>/<path>" should be escaped properly.
+        """
+        result = wb_util.escape_artifact_path(
+            f"{self.wandb_prefix}{self.artifact_name}:{self.version}/{self.file_path}"
+        )
+        uri = artifact_wandb.WeaveWBLoggedArtifactURI.parse(result)
+        assert uri.path == self.file_path
 
-    assert uri.path == path
+    def test_artifact_path_with_no_version(self):
+        """
+        Artifact path of the form: "<name>/<path>" should be escaped properly.
+        """
+        result = wb_util.escape_artifact_path(
+            f"{self.wandb_prefix}{self.artifact_name}/{self.file_path}"
+        )
+        uri = artifact_wandb.WeaveWBLoggedArtifactURI.parse(result)
+        assert uri.path == self.file_path
+
+    def test_artifact_path_with_version_and_file_with_colon(self):
+        """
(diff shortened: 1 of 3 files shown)


weave_query/tests/test_wb.py
```diff
@@ -1776,16 +1776,66 @@ def test_is_valid_version_string():
...
+    # def test_artifact_path_with_version_with_colon_and_file_with_colon(self):
+    #     """
+    #     Artifact path of the form: "<name>:<version>/<path>" should be escaped properly,
+    #     where <version> contains a colon and <path> contains a colon.
+    #     """
+    #     result = wb_util.escape_artifact_path(
+    #         f"{self.wandb_prefix}{self.artifact_name}:{self.version_with_colon}/{self.file_path_with_colon}"
+    #     )
+    #     uri = artifact_wandb.WeaveWBLoggedArtifactURI.parse(result)
+    #     assert uri.path == self.file_path_with_colon
```
nicholaspun-wandb, line 1838: The resulting URI `wandb-client-artifact://12347187287418787843872388177814:latest:v0/%28batch%3A%201%29%20table%20%233.table.json` is not actually valid (We run into a `errors.WeaveInvalidURIError`)

dannygoldstein (approved):

weave_query/tests/test_wb.py
dannygoldstein: what is the error this raises? can you share it here?

nicholaspun-wandb: Yup, here it is:
```
weave_query/artifact_wandb.py:1087: in parse
    return cls.from_parsed_uri(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'weave_query.artifact_wandb.WeaveWBLoggedArtifactURI'>
uri = 'wandb-client-artifact://12347187287418787843872388177814:latest:v0/%28batch%3A%201%29%20table%20%233.table.json', schema = 'wandb-client-artifact'
netloc = '12347187287418787843872388177814:latest:v0', path = '(batch: 1) table #3.table.json', params = '', query = {}, fragment = ''

    @classmethod
    def from_parsed_uri(
        cls,
        uri: str,
        schema: str,
        netloc: str,
        path: str,
        params: str,
        query: dict[str, list[str]],
        fragment: str,
    ):
        path = parse.unquote(path.strip("/"))
        spl_netloc = netloc.split(":")
        if len(spl_netloc) == 1:
            name = spl_netloc[0]
            version = None
        elif len(spl_netloc) == 2:
            name, version = spl_netloc
        else:
>           raise errors.WeaveInvalidURIError(f"Invalid WB Client Artifact URI: {uri}")
E           weave_query.errors.WeaveInvalidURIError: Invalid WB Client Artifact URI: wandb-client-artifact://12347187287418787843872388177814:latest:v0/%28batch%3A%201%29%20table%20%233.table.json

weave_query/artifact_wandb.py:1041: WeaveInvalidURIError
```

dannygoldstein: I think this should work if we make `spl_netloc = netloc.split(':', 1)`

jwlee64 (approved):

PR #2646, opened 2024-10-09
chore(weave): Enforce conventional commits

andrewtruong:


.github/workflows/pr.yaml
```diff
@@ -10,26 +11,25 @@ jobs:
...
-          # Allowed types: https://www.notion.so/wandbai/Pull-Requests-17966b997841407b95b7c36414ae0634
+          # Allowed types: See CONTRIBUTING.md
           types: |
             feat
             fix
+            refactor
             docs
             style
-            refactor
-            perf
```
gtarpenning, line 24: 😢  I liked `perf`

gtarpenning: `perf` and `test` are the only ones of those i have personally used 🤷🏻

PR #2647, opened 2024-10-09
chore(weave): Add note on providing reproducible examples

andrewtruong:


jamie-rasmussen (approved):
Should we suggest giving us additional information like links to projects, Weave version, Python version, OS?

PR #2648, opened 2024-10-09
chore(weave): Move weave python under sdk

andrewtruong:
1. Moves the weave python code and tests under `sdks/python` in prep for other langs  The `README.md` and `LICENSE` are symlinked from `sdks/python` to the repo root for now
2. Keeps `weave/clear_cache.py` for compat


sdks/python/weave/clear_cache.py
tssweeney, line 1: This move is going to break local builds again

andrewtruong: I think it breaks helm, yeah good point.

Gonna keep `clear_cache.py` where it is and leave a note.

tssweeney: I think you might want to modify the docker file to actually load from weave_query dir, rather than this file

andrewtruong: We already do:
https://github.com/wandb/core/blame/master/onprem/local/Dockerfile#L46

But I don't think that would work for helm:
https://github.com/wandb/helm-charts/blob/main/charts/operator-wandb/charts/weave/templates/deployment.yaml#L93

tssweeney: hmm... ok. no more changes needed here... but FWIW, my understanding is that:
1. https://github.com/wandb/core/blame/master/onprem/local/Dockerfile#L46 -> this docker file puts the file at the correct `weave/clear_cache.py` dir
2. https://github.com/wandb/helm-charts/blob/main/charts/operator-wandb/charts/weave/templates/deployment.yaml#L93 picks that up.

Implied by your comment is that https://github.com/wandb/helm-charts/blob/main/charts/operator-wandb/charts/weave/templates/deployment.yaml#L93 relies on `clear_cache.py` being in this repository at this location, but i am not sure that is true.

PR #2680, opened 2024-10-11
feat(weave): dictify unknown class instances

jamie-rasmussen:
## Description

Instead of giving up and resorting to `repr` we try to recursively convert objects into a dict.
See internal doc https://www.notion.so/wandbai/Better-Rendering-of-Gemini-Calls-114e2f5c7ef38037953df7b7a9e5a52b?pvs=4 for more details. This will allow us to build a chat view for Gemini and solves other problems with not logging important data in the trace.

Before:
Screenshot 2024-10-15 at 12 19 41 PM

After:
Screenshot 2024-10-09 at 2 08 00 PM

## Testing

How was this PR tested?

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -1477,25 +1477,23 @@ def test_named_reuse(client):
 
 
 def test_unknown_input_and_output_types(client):
-    class MyUnserializableClassA:
+    class MyUnknownClassA:
         a_val: float
 
         def __init__(self, a_val) -> None:
             self.a_val = a_val
 
-    class MyUnserializableClassB:
+    class MyUnknownClassB:
         b_val: float
 
         def __init__(self, b_val) -> None:
             self.b_val = b_val
 
     @weave.op()
-    def op_with_unknown_types(
-        a: MyUnserializableClassA, b: float
-    ) -> MyUnserializableClassB:
-        return MyUnserializableClassB(a.a_val + b)
+    def op_with_unknown_types(a: MyUnknownClassA, b: float) -> MyUnknownClassB:
+        return MyUnknownClassB(a.a_val + b)
 
-    a = MyUnserializableClassA(3)
+    a = MyUnknownClassA(3)
     res = op_with_unknown_types(a, 0.14)
 
     assert res.b_val == 3.14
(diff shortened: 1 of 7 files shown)


tests/trace/test_client_trace.py
```diff
@@ -1508,25 +1506,39 @@ def op_with_unknown_types(
 
     assert len(inner_res.calls) == 1
     assert inner_res.calls[0].inputs == {
-        "a": repr(a),
+        "a": {
+            "__class__": {
+                "module": "test_client_trace",
+                "qualname": "test_unknown_input_and_output_types.<locals>.MyUnknownClassA",
```
andrewtruong, line 1512: Why keep `<locals>`?

jamie-rasmussen: I'm just storing the value exactly as it comes back from `obj.__class__.__qualname__` - but I'm not using the qualname for anything and I'm questioning whether it is worth saving or not in general.

andrewtruong:
What happens when you try to load this back?  I assume it comes back as  `WeaveDict`, but maybe we would prefer something else?

jamie-rasmussen:
> What happens when you try to load this back? I assume it comes back as `WeaveDict`, but maybe we would prefer something else?

Yes, WeaveDict. Do you have something else in mind?

tests/trace/test_weave_client.py
```diff
@@ -675,7 +675,14 @@ def __init__(self, a):
     ref = client._save_object(obj, "my-np-array")
     obj2 = client.get(ref)
     # Expect None for now
```
andrewtruong, line 677: is this comment still relevant?

weave/trace/serialize.py
```diff
@@ -73,10 +74,11 @@ def _build_result_from_encoded(
     return result
 
 
-REP_LIMIT = 1000
+STR_LENGTH_LIMIT = 1000
```
andrewtruong, line 77: nit: STR_LEN_LIMIT (because you'd do `len(s)`)

weave/trace/serialize.py
```diff
@@ -86,11 +88,103 @@ def fallback_encode(obj: Any) -> Any:
...
+            return stringify(obj)
+        else:
+            seen.add(obj_id)
+
+    if maxdepth > 0 and depth > maxdepth:
+        # TODO: If obj at this point is a simple type,
+        #       maybe we should just return it rather than stringify
+        return stringify(obj)
+
+    if isinstance(obj, (int, float, str, bool, type(None))):
```
andrewtruong, line 121: nit: `if is_primitive`

weave/trace/serialize.py
```diff
@@ -86,11 +88,103 @@ def fallback_encode(obj: Any) -> Any:
...
+        except Exception:
+            raise ValueError("to_dict failed")
+
+    result = {}
+    result["__class__"] = {
+        "module": obj.__class__.__module__,
+        "qualname": obj.__class__.__qualname__,
+        "name": obj.__class__.__name__,
+    }
+    if hasattr(obj, "__len__") and hasattr(obj, "__getitem__"):
```
andrewtruong, line 148: This might be more flexible as `isinstance(obj, Sequence)`

weave/trace/serialize.py
```diff
@@ -86,11 +88,103 @@ def fallback_encode(obj: Any) -> Any:
...
+        # Custom list-like object
+        try:
+            n = len(obj)
+            for i in range(n):
+                result[i] = dictify(obj[i], maxdepth, depth + 1, seen)
```
andrewtruong, line 153: and this can be
```py
for i, item in enumerate(obj):
    result[i] = dictify(item, maxdepth, depth+1, seen)
```

weave/trace/serialize.py
```diff
@@ -86,11 +88,103 @@ def fallback_encode(obj: Any) -> Any:
...
+ALWAYS_STRINGIFY = (logging.Logger,)
+
+
+def fallback_encode(obj: Any) -> Any:
+    # TODO: Should we try to compute an object size and skip if too big?
+    if isinstance(obj, ALWAYS_STRINGIFY):
+        return stringify(obj)
+    try:
+        # Note: Max depth not picked scientifically, just trying to keep things under control.
+        return dictify(obj, maxdepth=10)
```
andrewtruong, line 183: Maybe this can be pulled into a const that is later configurable

andrewtruong (approved):
lgtm

tests/trace/test_weave_client.py
jamie-rasmussen: Nope

tssweeney (approved):
In the spirit of moving fast, i can approve this, but i think we will want to add a `_type: "PythonObject"` or something to as a sibling the "__class__". This would fit the pattern of: when recursing weave objects, when you encounter a dict with "_type", that tells you to stop recursing blindly and route to a type-specific interpretation of the payload.

PR #2681, opened 2024-10-11
chore(weave): Add evaluation scores as feedback

tssweeney:
This PR is strictly anticipatory. For performance and overall correct DB design, we are likely going to move to storing scores as feedback. This PR simply adds scores as feedback for now so that we can explore new query patterns and UI patterns moving forward.

Now, each time a score is generated for a prediction, a feedback record will be created according to :
```
class ScoreTypePayload(TypedDict):
    name: str
    op_ref: str
    call_ref: str
    results: dict
```

This should have 0 user effect in the short term. We will want to experiment with a few different data models before locking it in. In the mid term, this will lead to better performance and capabilities. Specifically: retroactive scoring, cached evals, and HITL scorers


tests/trace/test_evaluations.py
```diff
@@ -420,7 +431,14 @@ async def test_evaluation_data_topology(client):
...
     assert predict_call.output == model_output
-    assert predict_call.summary == predict_usage
+    summary = {
+        **predict_call.summary,
+        "weave": {
+            **predict_call.summary["weave"],
+            "feedback": [],
+        },
+    }
+    assert summary == with_empty_feedback(predict_usage)
```
andrewtruong, line 441: why are you doing this?

tests/trace/test_evaluations.py
```diff
@@ -747,3 +780,36 @@ def function_score(image, dc, model, obj, text, model_output) -> bool:
...
+        )
+    )
+    assert len(calls.calls) == 1
+    assert calls.calls[0].summary["weave"]["feedback"]
+    feedbacks = calls.calls[0].summary["weave"]["feedback"]
+    assert len(feedbacks) == 1
+    feedback = feedbacks[0]
+    assert feedback["feedback_type"] == SCORE_TYPE_NAME
+    assert feedback["payload"]["name"] == "score"
+    assert feedback["payload"]["op_ref"] == get_ref(score).uri()
```
andrewtruong, line 814: nit: we have some inconsistencies in our code.  Some places use `get_ref(obj)`, whereas others use `obj.ref`

tests/trace/test_scores.py
```diff
@@ -0,0 +1,42 @@
...
+        return {"in_range": input_x < model_output}
+
+    call_res, call = my_op.call(1)
+    assert call_res == 2
+    score_res, score_call = my_score.call(1, call_res)
+    assert score_res == {"in_range": True}
+    res_fut = client._send_score_call(call, score_call)
+    assert isinstance(res_fut, Future)
+    res = res_fut.result()
+    assert isinstance(res, str)
```
andrewtruong, line 25: May be worth commenting -- this wasn't obvious on first read.

It seems like what's happening is:
1. Invoke an op
2. Invoke a scoring op
3. "attach" a score_call to the call?

weave/flow/eval.py
```diff
@@ -164,7 +178,23 @@ async def predict_and_score(
...
+            model_call = None
+            if is_op(model_predict):
+                # I would expect this path to always be hit, but keeping the other
+                # path for backwards compatibility / safety
+                model_predict = as_op(model_predict)
+                if model_self is not None:
+                    model_predict_args = {
+                        **model_predict_args,
+                        "self": model_self,
+                    }
```
andrewtruong, line 190: nit: do you need to try all of this, or just the part below?

weave/flow/eval.py
```diff
@@ -221,7 +254,24 @@ async def predict_and_score(
...
-                result = await async_call(score_fn, **score_args)
+                if is_op(score_fn) and model_call:
+                    # I would expect this path to always be hit, but keeping the other
+                    # path for backwards compatibility / safety
+                    score_fn = as_op(score_fn)
+                    if scorer_self is not None:
+                        score_args = {
+                            **score_args,
+                            "self": scorer_self,
+                        }
```
andrewtruong, line 265: nit: same q here

weave/trace/weave_client.py
```diff
@@ -235,6 +245,13 @@ def ui_url(self) -> str:
             raise ValueError("Can't get URL for call without ID")
         return urls.redirect_call(entity, project, self.id)
 
+    @property
+    def ref(self) -> CallRef:
+        entity, project = self.project_id.split("/")
```
andrewtruong, line 250: seems like a code smell -- this obj has `project_id` but not `entity, project`, but the `CallRef` has `entity, project` and no `project_id`?

weave/trace/weave_client.py
```diff
@@ -1042,6 +1059,95 @@ def query_costs(
...
+    @trace_sentry.global_trace_sentry.watch()
+    def _add_score(
+        self,
+        *,
+        call_ref_uri: str,
+        score_name: str,
+        score_results: Any,
+        scorer_call_ref_uri: str,
+        scorer_op_ref_uri: str,
+    ) -> str:  # , supervision: dict):
```
andrewtruong, line 1103: move into sig?

weave/trace/weave_client.py
```diff
@@ -1042,6 +1059,95 @@ def query_costs(
...
+            "op_ref": scorer_op_ref_uri,
+            "call_ref": scorer_call_ref_uri,
+            "results": results_json,
+        }
+
+        freq = FeedbackCreateReq(
+            project_id=self._project_id(),
+            weave_ref=call_ref_uri,
+            feedback_type=SCORE_TYPE_NAME,  # should this be score_name?
+            payload=payload,
```
andrewtruong, line 1145: is this payload arbitrary and you are just defining a specific type of payload of shape `ScoreTypePayload`?

tests/trace/test_evaluations.py
tssweeney: i think `get_ref` and `set_ref` should be used everywhere, since `ref` is often tacked on in different ways based on the object and it seems more robust to just have helpers here.

andrewtruong: Are there other ways than `.ref`?  I'm not against using `get_ref`, but I do think we should be consistent

tests/trace/test_evaluations.py
tssweeney: I am now requesting feedback which is used below, but don't want to make assertions about it here. Stripping the feedback makes that nice. I cleaned up the code to make it cleaner.

tests/trace/test_evaluations.py
tssweeney: we should probably standardize:
```

def get_ref(obj: Any) -> Optional[ObjectRef]:
    return getattr(obj, "ref", None)


def remove_ref(obj: Any) -> None:
    if get_ref(obj) is not None:
        if "ref" in obj.__dict__:  # for methods
            obj.__dict__["ref"] = None
        else:
            obj.ref = None


def set_ref(obj: Any, ref: Optional[Ref]) -> None:
    """Try to set the ref on "any" object.

    We use increasingly complex methods to try to set the ref
    to support different kinds of objects. This will still
    fail for python primitives, but those can't be traced anyway.
    """
    try:
        obj.ref = ref
    except:
        try:
            setattr(obj, "ref", ref)
        except:
            try:
                obj.__dict__["ref"] = ref
            except:
                raise ValueError(f"Failed to set ref on object of type {type(obj)}")
```

tests/trace/test_scores.py
tssweeney: ok

weave/flow/eval.py
tssweeney: just `async_call_op` or `async_call`

weave/trace/weave_client.py
tssweeney: yeah, it would be nice for that to be unified across out objects

weave/trace/weave_client.py
tssweeney: done

weave/trace/weave_client.py
tssweeney: yup!

weave/flow/eval.py
andrewtruong: I would make the try block just over that then

andrewtruong (approved):
lgtm

PR #2682, opened 2024-10-11
chore(weave): Add docs to CI and build process

andrewtruong:


Makefile
```diff
@@ -0,0 +1,17 @@
...
+setup-docs-ci:
+	pip install -e .[docs]
+	playwright install
+
+	cd docs && \
+	npm install --global yarn && \
+	npm install
+
+docs: 
+	cd docs && make generate_all
```
tssweeney, line 12: one note: generate_all hits the live server AND the frontend app. This means every single commit will trigger a few pings. Probably fine, but a bit fragile

tssweeney (approved):

Makefile
andrewtruong: What's the alternative?

PR #2688, opened 2024-10-13
chore(weave): Add Object field deprecation decorator

andrewtruong:


weave/flow/obj.py
```diff
@@ -23,6 +45,7 @@ class Object(BaseModel):
         arbitrary_types_allowed=True,
         protected_namespaces=(),
         extra="forbid",
+        populate_by_name=True,
```
gtarpenning, line 48: not necessary, but I had to look up what this meant, might be nice to add a comment about what this does and why we turned it on.

gtarpenning (approved):
Are we actually using this anywhere?

weave/flow/obj.py
gtarpenning: (this will affect all aliases right?)

andrewtruong: Yeah, affects all aliases, though I don't think we use any atm.  In future it will be used to support deprecating fields.

weave/flow/obj.py
```diff
@@ -12,6 +13,27 @@
 from weave.trace.vals import WeaveObject, pydantic_getattribute
 from weave.trace.weave_client import get_ref
 
+logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
+
+
+def deprecated_field(new_field_name: str) -> Callable[[Callable[[Any], T]], property]:
```
tssweeney, line 21: what is this for? not seeing it used anywhere

andrewtruong: This is being added to facilitate some work @tcapelle is doing where:
1. He wants to change the name of a field; but
2. We want to make it backwards-compatible

https://weightsandbiases.slack.com/archives/C0785BR1RNZ/p1728746308456249?thread_ts=1728633360.048499&cid=C0785BR1RNZ

PR #2696, opened 2024-10-15
fix(weave): Use sdk-local deserializer instead of saved deserializer for known types like Images

andrewtruong:
## Description

https://wandb.atlassian.net/browse/WB-21517

- Prefer using the SDK's local registered deserializer instead of the saved deserializer for known types like Images

## Testing

Unit tests and manual testing

### DIFF
--- a/weave/trace/custom_objs.py
+++ b/weave/trace/custom_objs.py
@@ -15,7 +15,11 @@ class DecodeCustomObjectError(Exception):
 
 # in future, could generalize as
 # {target_cls.__module__}.{target_cls.__qualname__}
-KNOWN_TYPES = ["PIL.Image.Image", "wave.Wave_read"]
+KNOWN_TYPES = {
+    "Op",
+    "PIL.Image.Image",
+    "wave.Wave_read",
+}
 
 
 def encode_custom_obj(obj: Any) -> Optional[dict]:
@@ -71,7 +75,7 @@ def _decode_custom_obj(
 def decode_custom_obj(
     weave_type: dict,
     encoded_path_contents: Mapping[str, Union[str, bytes]],
-    load_instance_op_uri: Optional[str],
+    load_instance_op_uri: Optional[str] = None,
 ) -> Any:
     _type = weave_type["type"]
     found_serializer = False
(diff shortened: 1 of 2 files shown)


weave/trace/custom_objs.py
```diff
@@ -51,8 +57,12 @@ def decode_custom_obj(
     encoded_path_contents: Mapping[str, Union[str, bytes]],
     load_instance_op_uri: Optional[str],
 ) -> Any:
-    load_instance_op = None
-    if load_instance_op_uri is not None:
+    if weave_type["type"] in KNOWN_TYPES:
+        serializer = get_serializer_by_id(weave_type["type"])
```
tssweeney, line 61: what happens if the serialization changes? It seems like we should first "try" the local version, then fallback to the remote version, then error

andrewtruong: ok yeah that's safer

weave/trace/custom_objs.py
```diff
@@ -52,31 +56,35 @@ def decode_custom_obj(
     encoded_path_contents: Mapping[str, Union[str, bytes]],
     load_instance_op_uri: Optional[str],
 ) -> Any:
-    load_instance_op = None
-    if load_instance_op_uri is not None:
+    _type = weave_type["type"]
+    found_serializer = False
+
+    # First, try to load the object using a known serializer
+    if _type in KNOWN_TYPES:
```
tssweeney, line 63: this is still not quite right - we need to try with the local one first, else fallback to the the remote one

andrewtruong: sorry, asked for review too soon before pushing commit -- can you check again

weave/trace/custom_objs.py
```diff
@@ -9,6 +9,19 @@
...
 
 
+class DecodeCustomObjectError(Exception):
+    """An error that occurs while decoding a custom object."""
+
+
+# in future, could generalize as
+# {target_cls.__module__}.{target_cls.__qualname__}
+KNOWN_TYPES = {
+    "Op",
```
tssweeney, line 19: does this work for Ops? I thought ops were a special hard coded one? I thought at one time we had special Op handling in `def from_json(obj: Any, project_id: str, server: TraceServerInterface) -> Any:`

tssweeney: I stand corrected. nvm

tssweeney (approved):

PR #2710, opened 2024-10-15
feat(weave): add cost logging to Gemini integration

jamie-rasmussen:
## Description

Updates the Gemini integration to make use of the new "finish handler" support. We register a callback that extracts the token usage information from the Gemini response object and converts it into the format our costs code expects.

Before:
Screenshot 2024-10-15 at 2 18 04 PM

After:
Screenshot 2024-10-15 at 2 31 36 PM

## Testing

How was this PR tested?


tests/integrations/google_ai_studio/google_ai_studio_test.py
andrewtruong, line 1: these tests will fail until `jamie/dictify2` is in `master`

andrewtruong (approved):

PR #2712, opened 2024-10-15
chore(weave): fix seed costs

jwlee64:
## Description
The target db of the client passed in was incorrect, specify the target db for insert costs

## Testing

running make migrate local

### DIFF
--- a/weave/trace_server/costs/insert_costs.py
+++ b/weave/trace_server/costs/insert_costs.py
@@ -148,7 +148,8 @@ def sum_costs(data: Dict[str, List[CostDetails]]) -> float:
     return total_costs
 
 
-def insert_costs(client: Client) -> None:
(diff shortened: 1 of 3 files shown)


weave/trace_server/costs/insert_costs.py
```diff
@@ -148,7 +148,8 @@ def sum_costs(data: Dict[str, List[CostDetails]]) -> float:
     return total_costs
 
 
-def insert_costs(client: Client) -> None:
+def insert_costs(client: Client, target_db: str) -> None:
+    client.database = target_db
```
tssweeney, line 152: this matches existing pattern, but would be a bit odd if the same client was used in different target_dbs. No action, just calling out

tssweeney (approved):

PR #2714, opened 2024-10-15
feat(weave): save audio as wave.Wave_read objects to weave

gtarpenning:
## Description

SDK support for audio. This pr enables the saving of audio in a similar paradigm as images. Simply pass a `wave.Wave_read` object to the inputs or outputs of an op and weave will serialize the audio for future playback.

Example: 

```python
import weave
import wave
from openai import OpenAI

weave.init()
client = OpenAi()

@weave.op
def make_audio_file_streaming(text: str):
    with client.audio.speech.with_streaming_response.create(
        model="tts-1",
        voice="alloy",
        input=text,
        response_format="wav",
    ) as res:
        res.stream_to_file("output-stream.wav")
        return wave.open("output-stream.wav")

wave_file = make_audio_file_streaming("Hello world!")
```

Screenshot 2024-10-15 at 3 06 48 PM

https://wandb.atlassian.net/browse/WB-21557


weave/type_serializers/Audio/audio.py
```diff
@@ -0,0 +1,51 @@
...
+try:
+    from openai._legacy_response import HttpxBinaryResponseContent
+
+    dependencies_met = True
+except ImportError:
+    pass
+
+
+def save_httpx(
+    obj: "HttpxBinaryResponseContent", artifact: MemTraceFilesArtifact, name: str
```
tssweeney, line 18: `HttpxBinaryResponseContent` is always a .wav file??? seems unlikely.

weave/type_serializers/Audio/audio.py
```diff
@@ -0,0 +1,51 @@
...
+) -> None:
+    with artifact.new_file("audio.wav", binary=True) as f:
+        for data in obj.iter_bytes():
+            f.write(data)  # type: ignore
+
+
+def load_httpx(artifact: MemTraceFilesArtifact, name: str) -> BufferedReader:
```
tssweeney, line 25: load should always be the inverse of save. `BufferedReader != HttpxBinaryResponseContent`

weave/type_serializers/Audio/audio.py
```diff
@@ -0,0 +1,51 @@
+import wave
+from io import BufferedReader
+
+from weave.trace import serializer
+from weave.trace.custom_objs import MemTraceFilesArtifact
+
+dependencies_met = False
+
+try:
+    from openai._legacy_response import HttpxBinaryResponseContent
```
tssweeney, line 10: I would just remove this stuff and add a docs entry on how to make it work

weave/type_serializers/Audio/audio.py
```diff
@@ -0,0 +1,51 @@
...
+except ImportError:
+    pass
+
+
+def save_httpx(
+    obj: "HttpxBinaryResponseContent", artifact: MemTraceFilesArtifact, name: str
+) -> None:
+    with artifact.new_file("audio.wav", binary=True) as f:
+        for data in obj.iter_bytes():
+            f.write(data)  # type: ignore
```
tssweeney, line 22: won't this exhaust the iterator?

weave/type_serializers/Audio/audio.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def load_httpx(artifact: MemTraceFilesArtifact, name: str) -> BufferedReader:
+    path = artifact.path("audio.wav")
+    return open(path, "rb")
```
andrewtruong, line 27: if you open, make sure to close!

tssweeney (approved):

weave/type_serializers/Audio/audio.py
```diff
@@ -0,0 +1,26 @@
...
+        wav_file = wave.open(fp, "wb")
+        wav_file.setparams(params)
+        wav_file.writeframes(frames)
+        wav_file.close()
+    obj.rewind()
+
+
+def load(artifact: MemTraceFilesArtifact, name: str) -> wave.Wave_read:
+    path = artifact.path(AUDIO_FILE_NAME)
+    return wave.open(path, "rb")
```
andrewtruong, line 22: you need to close otherwise it can cause resource leak

andrewtruong (changes requested):

weave/type_serializers/Audio/audio.py
gtarpenning: Do you have a second to pair on this? I understand the concern and am not quite sure the correct way to remedy it.... @andrewtruong

andrewtruong (approved):

PR #2715, opened 2024-10-15
chore(ui): audio player for logged wav objects

gtarpenning:
## Description

https://wandb.atlassian.net/browse/WB-21558

Very simple UI for playing audio. There are three different states the new `AudioPlayer` component can be in: 
- controls: has a play/pause button and the timestamps
- slider: show play/pause, the time, a download link, and a scrubbable waveform visualization
- mini: show play/pause

The width of the component determines what state the viewer will be in. Because columns are restricted to 150px minimum in the calls table there doesn't appear to be any place where we use the mini viewer. 

Screenshot 2024-10-15 at 2 58 05 PM
Screenshot 2024-10-15 at 2 58 13 PM
Screenshot 2024-10-15 at 3 03 53 PM

Screenshot 2024-10-16 at 3 01 36 PM

TODO:
 - better support for expanding the viewer. When its small there should be some indication that it can get larger.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/CustomWeaveTypeDispatcher.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/CustomWeaveTypeDispatcher.tsx
@@ -29,9 +29,6 @@ const customWeaveTypeRegistry: {
   'PIL.Image.Image': {
     component: PILImageImage,
   },
-  'openai._legacy_response.HttpxBinaryResponseContent': {
-    component: AudioPlayer,
-  },
   'wave.Wave_read': {
     component: AudioPlayer,
   },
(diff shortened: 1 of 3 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/CustomWeaveTypeDispatcher.tsx
```diff
@@ -28,6 +29,12 @@ const customWeaveTypeRegistry: {
   'PIL.Image.Image': {
     component: PILImageImage,
   },
+  'openai._legacy_response.HttpxBinaryResponseContent': {
```
tssweeney, line 32: similar to other PR, i really don't think `HttpxBinaryResponseContent` is just for audio

tssweeney (approved):

PR #2718, opened 2024-10-16
fix(ui): Fix "Ghost" columns in tables

tssweeney:
Have you been plagued by ghost columns? Columns that shouldn't be in the normal view but seem to be left over from the last one? Yah, me to. It was because of me.

# Before:

Anyway, the fix is to correctly _not_ update the state if by the time the network requests complete, the new dependency has changed. Classic react bug.

# After:

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
@@ -29,115 +29,36 @@ export const useClientSideCallRefExpansion = (
   const [isExpanding, setIsExpanding] = useState(false);
 
   useEffect(() => {
+    let mounted = true;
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
```diff
@@ -29,115 +29,36 @@ export const useClientSideCallRefExpansion = (
...
-            refsNeeded.add(value);
-          }
-        });
-      });
-
-      const refsNeededArray = Array.from(refsNeeded);
-
-      if (refsNeededArray.length === 0) {
-        setExpandedCalls(traceCalls);
+    if (calls.result.length === 0) {
```
tssweeney, line 39: All this PR does is extract `const doExpansionIteration = async (` into a top-level function and call it from the hook here. The only logical change is `if (calls.result === callResultStart && mounted) {` conditional on update

PR #2724, opened 2024-10-16
feat(weave): Update DSPy integration compatibility with `dspy.LM`

soumik12345:
Update DSPy integration to make it compatible with `dspy.LM` in accordance with the [migration guide](https://github.com/stanfordnlp/dspy/blob/main/examples/migration.ipynb).


andrewtruong:
Seems like all dspy shards are failing.  Can you reproduce if you do `nox -e {env}` locally?

PR #2731, opened 2024-10-17
fix(weave): include scores from scorers with illegal characters in name

gtarpenning:
## Description

- Fixes [WB-21559](https://wandb.atlassian.net/browse/WB-21559)

This pr includes: 
- Hack frontend fix for scorers with illegal chars in the name. The scorer becomes a ref, which sanitizes the name, then when we look them up in the score summary the names don't match. Use a regex with a wildcard to match up the sanitized names to the raw summary value ones. 
- a python fix that sanitizes the name so that the name in the summary will match the artifact name. 

### Note: Scorers with names like `v1` and `v1!` will be treated as identical...

Master: 
Screenshot 2024-10-17 at 10 02 55 AM

This branch:
Screenshot 2024-10-17 at 10 02 02 AM

This is the UI for the above two runs, while the unsanitized scorers look different, we can still compare them in the UI. 
Screenshot 2024-10-17 at 10 38 48 AM

### Testing
Added test

[WB-21559]: https://wandb.atlassian.net/browse/WB-21559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/flow/scorer.py
```diff
@@ -102,6 +103,11 @@ def get_scorer_attributes(
         summarize_fn = auto_summarize  # type: ignore
     else:
         raise ValueError(f"Unknown scorer type: {scorer}")
+
+    if not scorer_name:
+        raise ValueError("Scorer name is required")
```
gtarpenning, line 108: required by lint for some reason, should be impossible

tssweeney (approved):

tests/trace/test_evaluate.py
```diff
@@ -264,3 +264,25 @@ def model(col_a, col_b):
         "scores": {scorer_results[s]: True for s in scorers},
     }
     assert all(o == expected_output for o in outputs)
+
+
+@pytest.mark.parametrize(
+    "scorer_name",
+    ["my scorer", "my-scorer()*&^%$@#/", "my-scorer", "       my scorer     "],
```
gtarpenning, line 271: these names should all end up being the same @andrewtruong

weave/scorers/base_scorer.py
```diff
@@ -107,6 +108,10 @@ def get_scorer_attributes(
         summarize_fn = auto_summarize  # type: ignore
     else:
         raise ValueError(f"Unknown scorer type: {scorer}")
+
+    if scorer_name:
+        scorer_name = sanitize_object_name(scorer_name)
+
     return (scorer_name, score_fn, summarize_fn)  # type: ignore
```
andrewtruong, line 115: oh I see because we have this return of `scorer_name` here...

andrewtruong (approved):
good catch!

PR #2732, opened 2024-10-17
revert(weave): dictify unknown class instances

jamie-rasmussen:
Reverts wandb/weave#2680

See internal Slack: https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1729185249582949


jamie-rasmussen:
Because this didn't revert cleanly we decided to fix forward in https://github.com/wandb/weave/pull/2733

PR #2742, opened 2024-10-21
docs(weave): Hallucination_Cookbook

MG-Microsoft:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

 Hallucination_Cookbook
## Testing

How was this PR tested?
google colab


docs/docs/reference/gen_notebooks/hallucination.md
```diff
@@ -0,0 +1,450 @@
...
+
+```
+
+Loading Hallucination detection model:
+
+```python
+%%capture
+
+# Load the model with custom code
+model = AutoModelForSequenceClassification.from_pretrained(
```
vanpelt, line 107: Would be good to add a little context here about why we chose this hallucination model and what some other options could be for hallucination detection.

docs/docs/reference/gen_notebooks/hallucination.md
```diff
@@ -0,0 +1,450 @@
...
+
+# Set up embedding model
+embedding_model = OpenAIEmbedding(model="text-embedding-ada-002")
+
+
+```
+
+## 📥 Download and Load Documents
+
+We'll download a PDF document from a URL and create an index using LlamaIndex. Please note taht this can be your own vector database with your data indexed for your RAG Chatbot.
```
vanpelt, line 132: `taht` -> `that`

docs/docs/reference/gen_notebooks/hallucination.md
```diff
@@ -0,0 +1,450 @@
...
+
+
+@weave.op()
+def check_statement_hallucination(context: str, statement: str) -> Dict[str, Any]:
+    '''
+    Detects hallucination in the answer using the provided context and model answer.
+    '''
+
+    #statements = sent_tokenize(statement)
+    #pairs = [(context, statement) for statement in statements]
```
vanpelt, line 257: Why these comments?

docs/docs/reference/gen_notebooks/hallucination.md
```diff
@@ -0,0 +1,450 @@
...
+    )
+    answer = response.choices[0].message.content.strip()
+    return answer
+
+
+```
+
+### 3. Break Down Answer into Statements
+
+This function breaks down the answer into simpler statements without pronouns.
```
vanpelt, line 213: Why are we doing this?

docs/docs/reference/gen_notebooks/hallucination.md
```diff
@@ -0,0 +1,450 @@
...
+    evaluation = Evaluation(
+        dataset=dataset,  # the dataset we have defined earlier
+        scorers=[hallucination_scorer],  # the scorer function
+    )
+
+    # Run the evaluation
+    summary = asyncio.run(evaluation.evaluate(evaluator_model))
+
+    print(f"Completed evaluation with model: {model_name}\n")
+
```
vanpelt, line 403: Would be good to show some results here, like we see the crappier model hallucinates more or something.

docs/docs/reference/gen_notebooks/hallucination.md
MG-Microsoft: when you get the asnwer from LLM ( lets say in RAG) there are multiple statemtns provided even in one single asnwer so we need to run hallucination evalaution to check whic ones are fact. For instance, you ask about What is weave? Chatbot says: Weave is a product releaed at THIS Date by WANDB. So there are. two statements here and we need to check for both in case of hallucination

docs/docs/reference/gen_notebooks/hallucination.md
MG-Microsoft: will get removed

docs/docs/reference/gen_notebooks/hallucination.md
MG-Microsoft: will get fixed

docs/docs/reference/gen_notebooks/hallucination.md
MG-Microsoft: Will add it

docs/docs/reference/gen_notebooks/hallucination.md
vanpelt: That's great, we should tell the person reading this that 😉

docs/docs/reference/gen_notebooks/hallucination.md
vanpelt: Let's add a caption to the image explaining what this is telling us.  Something like in our example 3.5 turbo had a factual score of 1 while gpt-4o got 0.8 which means 4o is hallucinating more which doesn't make any sense 🤣

MG-Microsoft: Yes it sounds surprising, but here is the reason observed( I ran all the code with more changes to make it more observable in weave and still get the response for GPT3.5 better ). I added this to notebooks too:

_During the evaluation for this particular usecase, it was observed that GPT-4o tends to produce more hallucinations than GPT-3.5, which may seem unexpected at first. However, this difference can be explained by how hallucinations manifest in Retrieval-Augmented Generation (RAG).


A common type of hallucination in RAG is factual but ungrounded. For example, given the premise "The capital of France is Berlin," a generated response stating "The capital of France is Paris" would be considered a hallucination—even though it is factually correct—because the generated statement is not derived from the provided retrieval context.

This issue arises when the model generates answers that are based on its pre-trained knowledge rather than the specific textual data retrieved as part of the RAG process.

GPT-4o, with its tendency to produce more detailed and precise responses, sometimes is more prone to this type of hallucination. As its answers are often longer and more comprehensive, they contain remote facts or statements that may not align with or be verifiable from the retrieved context. As a result, the evaluation score for GPT-4 may appear lower, even though some of the hallucinated statements might still be factually accurate._

docs/docs/reference/gen_notebooks/hallucination.md
MG-Microsoft: Added to the notebook :) 

_In this step, the response generated by the LLM is divided into multiple individual statements. The reason for breaking down the answer is to evaluate each statement independently for potential hallucinations. Since a single answer can contain a mix of factual and hallucinated information, checking each statement separately ensures a more fine-grained analysis.

This approach helps identify whether every part of the response is grounded in the retrieved context, as some statements might appear factually correct but are not supported by the provided retrieval. This breakdown allows us to better detect factual but ungrounded (hallucinated) content, improving the reliability of the evaluation process._

vanpelt (approved):

PR #2745, opened 2024-10-21
feat(weave): adding charts to the traces page

mbarrramsey:
## Description

See example page, https://beta.wandb.ai/vanpelt/openui-hosted/weave/calls?betaVersion=9b2425666fb0efa1f67f98cc8b55e84b00ea06d0

Adds charts for Latency, Errors, and Requests to the top of the Weave traces page. Based on figma: https://www.figma.com/design/mSF3IydmSfbUAVLCi2YlNZ/Weave-charts(hack)?node-id=65-10462&node-type=frame&t=VN4fvFeQ9QDXa3mV-0

https://github.com/user-attachments/assets/261d071c-9428-406c-b4cc-6576a1f43dfe

I would recommend moving to a different plotting library if we _must_ support more custom tooltip styling or x axis date formatting since Plotly has limited capabilities for this.

## Testing

How was this PR tested?


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsCharts.tsx
```diff
@@ -0,0 +1,134 @@
...
+    () => [{field: 'started_at', sort: 'desc'}],
+    []
+  );
+  const calls = useCallsForQueryCharts(
+    entity,
+    project,
+    filter,
+    filterModelProp,
+    0,
+    1000,
```
mbarrramsey, line 46: Can update to 10,000 or any arbitrary number

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsCharts.tsx
```diff
@@ -0,0 +1,133 @@
...
+      const started_at = call.traceCall?.started_at;
+      if (!started_at) {
+        return;
+      }
+      const ended_at = call.traceCall?.ended_at;
+
+      const isError =
+        call.traceCall?.exception !== null &&
+        call.traceCall?.exception !== undefined &&
+        call.traceCall?.exception !== '';
```
mbarrramsey, line 82: Filtering out non-errors (so chart will start at the time the first error occurred)

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/Charts.tsx
```diff
@@ -0,0 +1,263 @@
...
+  binSizeMinutes?: number;
+}> = ({height, chartData, binSizeMinutes = 60}) => {
+  const divRef = useRef<HTMLDivElement>(null);
+
+  const plotlyData: Plotly.Data[] = useMemo(() => {
+    const groupedData = _(chartData)
+      .groupBy(d => {
+        const date = moment(d.started_at);
+        const roundedMinutes =
+          Math.floor(date.minutes() / binSizeMinutes) * binSizeMinutes;
```
mbarrramsey, line 88: Grouping by the hour, we can change this if there's a preference

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/SimplePageLayout.tsx
```diff
@@ -58,7 +58,7 @@ export const SimplePageLayout: FC<{
         flexDirection: 'column',
         flexGrow: 1,
         height: '100%',
-        overflow: 'hidden',
+        overflow: 'hidden', // here
```
mbarrramsey, line 61: Leaving this comment in for now -- these are the two places that cause the issues with padding in CallsCharts. Looks like other existing margins are also not getting applied evenly
Screenshot 2024-10-28 at 6 38 26 PM

tssweeney: I'm not sure I understand the `//here` comment. Can we just remove and we will revisit? Or maybe describe this more?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
```diff
@@ -131,6 +131,59 @@ export const useCallsForQuery = (
   }, [callResults, calls.loading, total, costs.loading, costResults, refetch]);
 };
 
+export const useCallsForQueryCharts = (
```
tssweeney, line 134: this does not have anything "Charts" specific, it is more of a refactor of the first section of `useCallsForQuery`. I'd recommend renaming this and adding a comment that it should be merged with `useCallsForQuery` (assuming you don't want to do that (: )

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/Charts.tsx
tssweeney: I think the binning logic really wants to be:
1. Select all the data given the filters and count
2. Calculate the delta between the first and last result ("totalTimeRange")
3. Divide "TotalTimeRange" by a configurable, fixed number (say 30), this will give you "BinSize" - this should be the same bin size across all charts.
4. Allocate each call into the corresponding bin.

Fixing the bin size is not a viable since it doesn't adapt to the traffic volume.

tssweeney (approved):
Approving with comments

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/Charts.tsx
mbarrramsey: I think I see what you're saying -- grouping by the hour won't always work since our data may only be over one hour or less for example (or too many hours). We can calculate the bin size based on a fixed number

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterMetrics.tsx
```diff
@@ -0,0 +1,24 @@
+import {Switch} from '@wandb/weave/components';
```
tssweeney, line 1: nit: given that this is specifically used in the calls table and not really a general component, i would co-locate it in the page/CallsPage dir

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
```diff
@@ -32,10 +32,12 @@ export const useCallsForQuery = (
...
   gridFilter: GridFilterModel,
-  gridSort: GridSortModel,
-  gridPage: GridPaginationModel,
-  expandedColumns: Set<string>,
-  columns?: string[]
+  gridSort?: GridSortModel,
+  gridPage?: GridPaginationModel,
+  expandedColumns?: Set<string>,
+  columns?: string[],
+  offset?: number,
```
tssweeney, line 39: nit: i don't prefer having parameters that override other parameters. I think you can just pass in a GridPaginationModel confirming payload. Where pageSize = limit and page = offset // limit

```
export interface GridPaginationModel {
    /**
     * Set the number of rows in one page.
     * If some of the rows have children (for instance in the tree data), this number represents the amount of top level rows wanted on each page.
     * @default 100
     */
    pageSize: number;
    /**
     * The zero-based index of the current page.
     * @default 0
     */
    page: number;
}
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -168,7 +170,8 @@ export const CallsTable: FC<{
   allowedColumnPatterns,
 }) => {
   const {loading: loadingUserInfo, userInfo} = useViewerInfo();
-
+  const [isMetricsChecked, setMetricsChecked] = useState(false);
+  console.log(isMetricsChecked);
```
tssweeney, line 174: dev tracks

tssweeney (approved):

PR #2746, opened 2024-10-21
feat(weave): Patch `send_message` in Google GenAI SDK

soumik12345:
Patch `ChatSession.send_message` in Google GenAI SDK.


tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -119,3 +119,34 @@ async def test_content_generation_async(client):
...
+    chat = model.start_chat()
+    chat.send_message(
+        (
+            "What is the sum of the first 50 prime numbers? "
+            "Generate and run code for the calculation, and make sure you get all 50."
+        )
+    )
+
+    calls = list(client.calls())
+    assert len(calls) == 2
```
andrewtruong, line 140: This is surprising.  Where is the 2nd call coming from?

tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -119,3 +119,34 @@ async def test_content_generation_async(client):
...
+
+    call = calls[0]
+    assert call.started_at < call.ended_at
+    trace_name = op_name_from_ref(call.op_name)
+    assert trace_name == "google.generativeai.ChatSession.send_message"
+    assert call.output is not None
```
andrewtruong, line 146: can you make a more relevant assert using a mock? This is just checking if anything populated at all.

tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -119,3 +119,34 @@ async def test_content_generation_async(client):
...
+
+    call = calls[1]
+    assert call.started_at < call.ended_at
+    trace_name = op_name_from_ref(call.op_name)
+    assert trace_name == "google.generativeai.GenerativeModel.generate_content"
+    assert call.output is not None
```
andrewtruong, line 152: same here

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
```diff
@@ -32,7 +32,10 @@ def gemini_accumulator(
 def gemini_on_finish(
     call: Call, output: Any, exception: Optional[BaseException]
 ) -> None:
-    original_model_name = call.inputs["self"].model_name
+    if hasattr(call.inputs["self"], "model_name"):
+        original_model_name = call.inputs["self"].model_name
+    elif hasattr(call.inputs["self"], "model"):
+        original_model_name = call.inputs["self"].model.model_name
```
andrewtruong, line 38: what about the case where it has neither?  Is that valid (then you need a default), or invalid (then you need an error msg)

tests/integrations/google_ai_studio/google_ai_studio_test.py
```diff
@@ -119,3 +119,34 @@ async def test_content_generation_async(client):
...
+
+@pytest.mark.retry(max_attempts=5)
+@pytest.mark.skip_clickhouse_client
+def test_send_message(client):
+    import google.generativeai as genai
+
+    genai.configure(api_key=os.getenv("GOOGLE_GENAI_KEY"))
+    model = genai.GenerativeModel(model_name="gemini-1.5-pro", tools="code_execution")
+    chat = model.start_chat()
+    chat.send_message(
```
andrewtruong, line 132: are you gonna support the async and streaming cases too?

tests/integrations/google_ai_studio/google_ai_studio_test.py
soumik12345: This is because `send_message` is using `GenerativeModel.generate_content` under the hood which we're already patching.

tests/integrations/google_ai_studio/google_ai_studio_test.py
soumik12345: Added support for the async case, however, if `stream=True`, then a bug is being encountered where the run doesn't complete; here's a [sample trace](https://wandb.ai/geekyrakshit/test-google-ai-studio/weave/calls/0192d51c-987e-7950-b690-d36382f1bb97).

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
soumik12345: Added error message.

tests/integrations/google_ai_studio/google_ai_studio_test.py
andrewtruong: Thanks, I would add that comment

tests/integrations/google_ai_studio/google_ai_studio_test.py
andrewtruong: did you look into why that's happening?

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
```diff
@@ -19,7 +19,35 @@ def gemini_accumulator(
...
+                    acc.candidates[i].content.parts[
+                        j
+                    ].code_execution_result.output += (
+                        value_part.code_execution_result.output
+                    )
+                    acc.candidates[i].content.parts[
+                        j
+                    ].code_execution_result.status = (
+                        value_part.code_execution_result.status
+                    )
```
andrewtruong, line 50: what does this mean?

andrewtruong:
I think these are still live tests, and the Google live tests have been failing in CI.  Can you mock this?

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
soumik12345: This workaround had to be made due to the unique way the type `Part` is defined in the google generativeai library. Parts consist of unique `oneoff` fields for different type of outputs such as text, code, code output, etc. By `oneoff` it means that an instance of `Part` can only have only any one of these fields set, i.e, it can either be text or code or code output, etc. For streaming scenarios like code execution where the `value` might consist of candidates with only a single type of part, cannot be accumulated to the same part in `acc` which would immediately overwrite all the previous values from a different part type. Hence, all the diffferent types of part must be checked and accumulated in diffferent parts in the same candidate.

andrewtruong: Can you document and/or make it more obvious in the code?  When I look at this I'm not sure what to make of it.  A few annotations will go a long way

andrewtruong:
Your pytest markers should be called flaky -- see example: https://github.com/wandb/weave/blame/master/tests/integrations/google_ai_studio/google_ai_studio_test.py#L50

Please also fix linting issues!

andrewtruong (approved):
Approving with the understanding that you have thoroughly checked the integration since these tests are skipped

PR #2750, opened 2024-10-22
docs(weave): Audio with weave 

wylerz:
## Description
Adds a cookbook containing a basic and advanced demo.

Basic demo uses OAI Conversational API with audio model to go from user input text to model response audio, and logs this in weave.

Advanced demo uses OAI Realtime API to implement a basic e2e audio (audio in, audio out) assistant where all interactions (User audio transcriptions, model response audio, and model response transcription) is logged in weave.

Demo: https://www.youtube.com/watch?v=lnnd73xDElw
## Testing

Primary cookbook section tested on MacOS, Colab. 
Advanced Realtime API section Tested on MacOS, needs testing on Windows for advanced usage.


vanpelt (commented):
This is awesome!  Only question for me would be do we really need to define all of those types / pydantic objects?  I would imagine they are in the latest official SDK no?

wylerz:
> This is awesome! Only question for me would be do we really need to define all of those types / pydantic objects? I would imagine they are in the latest official SDK no?
@vanpelt 

OAI official library doesn't support Realtime API yet and doesn't provide any types associated with that API.

I made a choice here to setup pydantic schema for the entire Realtime API as I figured:
(1) Readability: The rest of the code readability improves drastically not having to inline nested JSON objects
(2) Future Modification: The API schema enforces some typing and such, so if users wish to modify this code for their own use cases they can easily just work the the Pydantic schema rather than having to look at OAI docs (which are quite sparse as of right now)
(3) Appearance: The code block renders as folded in the markdown on our cookbooks page, so it doesn't take up a huge amount of space


Happy to hardcode JSON if we don't want to work with the pydantic schema.

wylerz:
> That's reasonable, I would imagine the official SDK will have types soon so let's call that out in a comment.

@vanpelt Done. Are we good to merge? The failing CI test is due to broken links that (from my testing) are in fact not broken. Will certainly double check after merge.

jamie-rasmussen (approved):
Has been reviewed by Scott as per https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1729880805188469

PR #2752, opened 2024-10-22
chore(weave): consistent font color for manage columns popup

harukatab:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

After
Screenshot 2024-10-22 at 12 15 50 AM

## Testing

How was this PR tested?


harukatab:
I think i was color blinded or hallucinated that it was using pitch black for fonts. not true so no fix needed.

PR #2755, opened 2024-10-22
fix(weave): global df in op fails during op saving

gtarpenning:
## Description

This pr: 
- Add pandas shard for testing pandas interactions with weave
- add test for dataframe manipulation as side effect of op
- add try/catch when generating code dependencies to prevent crash

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -240,7 +240,7 @@ jobs:
             'mistral1',
             'notdiamond',
             'openai',
-            'pandas',
+            'pandas-test',
           ]
       fail-fast: false
     services:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -240,6 +240,7 @@ jobs:
             'mistral1',
             'notdiamond',
             'openai',
+            'pandas',
           ]
       fail-fast: false
     services:
--- a/noxfile.py
+++ b/noxfile.py
@@ -40,15 +40,12 @@ def lint(session):
         "mistral1",
         "notdiamond",
         "openai",
+        "pandas",
     ],
 )
 def tests(session, shard):
     if session.python.startswith("3.13") and shard in PY313_INCOMPATIBLE_SHARDS:
         session.skip(f"Skipping {shard=} as it is not compatible with Python 3.13")
-
-    # shard specific installs
-    if shard == "trace":
-        session.install("pandas")
 
     session.install("-e", f".[{shard},test]")
     session.chdir("tests")
--- a/noxfile.py
+++ b/noxfile.py
@@ -40,7 +40,7 @@ def lint(session):
         "mistral1",
         "notdiamond",
         "openai",
-        "pandas",
+        "pandas-test",
     ],
 )
 def tests(session, shard):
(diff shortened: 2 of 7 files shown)


tests/trace/test_weave_client.py
```diff
@@ -1528,3 +1529,22 @@ def my_op(a: int) -> int:
...
+    df = pd.DataFrame({"a": ["a", "b", "c"]})
+
+    @weave.op()
+    def my_op(a: int) -> int:
+        # modify the global df
+        prev_val = df.loc[df.index[0], "a"]
+        df.loc[df.index[0], "a"] = a
+        return prev_val
+
+    res = my_op("d")
```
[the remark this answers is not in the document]
gtarpenning: oops

noxfile.py
```diff
@@ -46,6 +46,10 @@ def tests(session, shard):
     if session.python.startswith("3.13") and shard in PY313_INCOMPATIBLE_SHARDS:
         session.skip(f"Skipping {shard=} as it is not compatible with Python 3.13")
 
+    # shard specific installs
+    if shard == "trace":
+        session.install("pandas")
```
gtarpenning, line 51: whats a less terrible way of adding dependencies for specific shards? Pandas isn't small... @andrewtruong

andrewtruong: You can make a separate shard just for these tests which have the trace deps + pandas

.github/workflows/test.yaml
```diff
@@ -240,6 +240,7 @@ jobs:
             'mistral1',
             'notdiamond',
             'openai',
+            'pandas',
```
andrewtruong, line 243: I would call this `pandas-test` to show that it's just for testing, not a user-facing extra

tests/integrations/pandas/test_pandas.py
```diff
@@ -0,0 +1,22 @@
...
+
+import weave
+
+
+def test_op_save_with_global_df(client):
+    df = pd.DataFrame({"a": ["a", "b", "c"]})
+
+    @weave.op()
+    def my_op(a: str) -> str:
+        # modify the global df
```
andrewtruong, line 11: technically it's just outside the op scope

weave/trace/op_type.py
```diff
@@ -320,9 +320,18 @@ def get_code_deps(
...
             warnings: list[str], any warnings that occurred during the process.
     """
-    return _get_code_deps(fn, artifact, {}, depth)
+    try:
+        return _get_code_deps(fn, artifact, {}, depth)
+    except Exception as e:
+        print(f"Error getting code deps for {fn}: {e}")
+        return {
+            "import_code": [],
+            "code": ["<error>"],
```
andrewtruong, line 332: nit: Should pull this out into a const

weave/trace/op_type.py
```diff
@@ -355,7 +364,13 @@ def _get_code_deps(
         var_value = None
         if isinstance(fn, py_types.FunctionType):
             var_value = resolve_var(fn, var_name)
-        if var_value == fn:
+
+        try:
+            # Some objects throw on equality comparison (like dataframes)
+            equivalent = var_value == fn
+        except ValueError:
```
andrewtruong, line 371: Does pandas specifically throw `ValueError`?

andrewtruong (approved):

weave/trace/op_type.py
gtarpenning: yeah, should we make this broader?

andrewtruong: Nah, it's probably ok -- we have the outer try/except that will catch other errors if any

PR #2770, opened 2024-10-24
feat(weave): link to artifacts from wandb models

levinandrew:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?


levinandrew:

weave-js/src/react.tsx
```diff
@@ -553,15 +553,20 @@ const RE_WEAVE_CALL_REF_PATHNAME = new RegExp(
...
+  let maxSplitsToMake: number;
+  let allowableNumSeparations: number[];
 
   const isWandbArtifact = url.protocol.startsWith('wandb-artifact');
   const isLocalArtifact = url.protocol.startsWith('local-artifact');
   const isWeaveRef = url.protocol.startsWith('weave');
   if (isWandbArtifact) {
-    splitLimit = 4;
+    maxSplitsToMake = 4;
+    // artifact URI expected as <entity>/<project>/<artifact name>:<artifact version - ie, v0>/<optional path>
```
levinandrew, line 564: expects in this form where version is v0, v1, or an artifact alias because that is how the SDK currently supports downloading an artifact (see https://docs.wandb.ai/guides/artifacts/download-and-use-an-artifact/). if we want to be able to reference the specific artifact version by digest on both the models and weave side, I think we'd need to add SDK support for retrieving an artifact by digest

weave-js/src/components/PagePanelComponents/Home/Browse2/SmallRef.tsx
```diff
@@ -76,11 +78,92 @@ export const objectRefDisplayName = (
...
+    <Link
+      $variant="secondary"
+      style={{
+        width: '100%',
+        minHeight: '38px',
+        display: 'flex',
+        alignItems: 'center',
+      }}
+      as="a"
+      href={`${window.location.origin}/${objRef.entityName}/${objRef.projectName}/artifacts/model/${objRef.artifactName}/${objRef.artifactVersion}`}
```
levinandrew, line 152: the hardcoded `artifacts/model` takes you to the artifacts registry in the models folder, at the screenshot below for example

if you ever wanted to reference another type of artifact (ie, `prompt` from below), you'd either have to 
- make a call to gorilla in this component to lookup the artifact by name, and get more metadata to construct the link

_OR_

- pass the artifact type as part of the artifact URI when creating the weave object (in your python script that's calling the weave SDK code)

 
image

weave-js/src/components/PagePanelComponents/Home/Browse2/SmallRef.tsx
```diff
@@ -76,8 +78,93 @@ export const objectRefDisplayName = (
   throw new Error('Unknown ref type');
 };
 
-export const SmallRef: FC<{
```
levinandrew, line 79: the biggest number of line changes of this PR are in this file. here I just wanted to separate the `SmallRef` for weave-objects and for artifacts into separate components. 

to do that and avoid code duplication, I moved the "item" into this `SmallRefBox` component

PR #2774, opened 2024-10-24
chore(weave): Add back hook to check for weave data

ericakdiaz:
## Description

This basically reverts the cleanup from https://github.com/wandb/weave/pull/2132. I believe that the limit issue should've been fixed in https://github.com/wandb/weave/pull/2420, so hopefully we should be able to check if a project has weave data more quickly.

## Testing

How was this PR tested?

This should have no effect since it's unused, will follow up in core with a change similar to this https://github.com/wandb/core/pull/20661/commits/3ca824472beff5023cd0c4926e1f95d9cbc63f1c and see how the performance is now.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
@@ -10,9 +10,9 @@
 
 import React, {createContext, FC, useContext} from 'react';
 
+import {useHasTraceServerClientContext} from './traceServerClientContext';
 import {tsWFDataModelHooks} from './tsDataModelHooks';
 import {WFDataModelHooksInterface} from './wfDataModelHooksInterface';
-import {useHasTraceServerClientContext} from './traceServerClientContext';
 
 const WFDataModelHooksContext = createContext<WFDataModelHooksInterface | null>(
   null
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
```diff
@@ -35,3 +36,44 @@ export const WFDataModelAutoProvider: FC<{
...
+  entity: string,
+  project: string
+) => {
+  const hasTraceServer = useHasTraceServerClientContext();
+  const objs = tsWFDataModelHooks.useRootObjectVersions(
+    entity,
+    project,
+    {},
+    1,
+    true,
```
ericakdiaz, line 54: added metadata_only param

jamie-rasmussen (approved):
Needs lint

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
```diff
@@ -35,3 +36,44 @@ export const WFDataModelAutoProvider: FC<{
...
+    {
+      skip: !hasTraceServer,
+    }
+  );
+
+  const calls = tsWFDataModelHooks.useCalls(
```
tssweeney, line 60: you could optimize this by selecting only the `id` column for example as well

ericakdiaz: done!

tssweeney (approved):

PR #2775, opened 2024-10-24
chore(ui): Creating a wave loader to start migrating app loaders

jtulk:
## Description
Preliminary work for WB-21477

Extends the tracked loader to use a new loading animation. Extends the tailwind config to support the animation.

### DIFF
--- a/weave-js/src/common/components/WandbLoader.tsx
+++ b/weave-js/src/common/components/WandbLoader.tsx
@@ -2,6 +2,7 @@
  * with an upgrade of react-spring, so we've switched back to the semantic loader.
  * The react-spring version also used 100% cpu, we should use an animated gif
  * instead if we want a custom loader */
+import {WaveLoader} from '@wandb/weave/components/Loaders/WaveLoader';
 import React from 'react';
 import {Loader, StrictLoaderProps} from 'semantic-ui-react';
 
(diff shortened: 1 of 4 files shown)


weave-js/src/common/components/WandbLoader.tsx
```diff
@@ -86,3 +88,48 @@ export const TrackedWandbLoader = ({
...
+        const additionalData = profilingCb ? profilingCb() : {};
+        const trackedData = {
+          componentId: data.id,
+          duration: data.duration,
+          ...additionalData,
+        };
+        if (onComplete) {
+          onComplete(name, trackedData);
+        }
+        const randomNum = Number(Math.random().toString().slice(-2)); // take the last two digits off a random number
```
connieelee, line 118: **nitpick:** looks like this is just copied from above but not a fan of the number->string->number acrobatics here. something like `Math.floor(Math.random() * 100)` feels more elegant and is likely faster (though i'm sure the cost is extremely negligible 😛)

weave-js/src/common/components/WandbLoader.tsx
```diff
@@ -86,3 +88,48 @@ export const TrackedWandbLoader = ({
...
+  profilingCb,
+  samplingRate = 0.1,
+  track,
+  onComplete,
+  onStart,
+  size,
+}: TrackedWandbLoaderProps & {
+  size: 'small' | 'huge';
+}) => {
+  useLifecycleProfiling(
```
connieelee, line 104: **question:** seems like contents of this call are copy-pasted from the existing tracked loader above. are there subtle differences i'm missing? could we pull this logic out and reuse for both comps (and any other future loaders)?

connieelee (approved):

weave-js/src/common/components/WandbLoader.tsx
connieelee: story of my life 😆

tssweeney (approved):

PR #2779, opened 2024-10-25
perf(weave): split calls query when using a heavy condition

gtarpenning:
## Description

[WB-21659](https://wandb.atlassian.net/browse/WB-21659)

This PR adds functionality to the CallsQuery builder to optionally generate two queries instead of one when conditions are right. If the query has a heavy filter or order by, we can't push down into the subquery, which causes a heinous join and bad performance. To remedy this we split the query into two parts, resolving the set of all possible call IDs, then adding that as a param to the param builder, then making a second query. This dance is managed by a helper function: `_construct_query_str_maybe_do_pre_query` which always returns a valid sql query string. If we need to make a pre-query to get the IDs it happens in the helper, and returns the formatted next query. 

## Testing

added test

[WB-21659]: https://wandb.atlassian.net/browse/WB-21659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/trace_server/test_calls_query_builder.py
```diff
@@ -6,12 +6,27 @@
 from weave.trace_server.orm import ParamBuilder
 
 
+def assert_sql(cq: CallsQuery, exp_queries, exp_params):
+    pb = ParamBuilder("pb")
+    queries = cq.as_sql(pb)
+    params = pb.get_params()
+
+    for qr, qe in zip(queries, exp_queries):
```
andrewtruong, line 14: what are `qr, qe`?

gtarpenning: queryRequest, expectedQuery

tssweeney (approved):

PR #2780, opened 2024-10-25
feat(ui): Add simple language detection for code browser

andrewtruong:
## Description

Extremely simple language detection :) 

## Testing

How was this PR tested?


weave-js/src/components/PagePanelComponents/Home/Browse2/Browse2OpDefCode.tsx
```diff
@@ -37,10 +37,29 @@ export const Browse2OpDefCode: FC<{uri: string; maxRowsInView?: number}> = ({
     );
   }
 
+  const detectLanguage = (code: string) => {
```
jamie-rasmussen, line 40: For the purposes of unit testing and not recreating the function on each render I think it would be a little cleaner to pull this out out the class and just have it as a top-level function in the file that also takes uri as an argument.

jamie-rasmussen (approved):

PR #2786, opened 2024-10-25
feat(ui): Add JS SDK options for Use tab

andrewtruong:
- **test**
- **test**
- **test**


weave-js/src/components/PagePanelComponents/Home/Browse2/Browse2OpDefCode.tsx
```diff
@@ -37,10 +37,29 @@ export const Browse2OpDefCode: FC<{uri: string; maxRowsInView?: number}> = ({
     );
   }
 
+  const detectLanguage = (code: string) => {
```
jamie-rasmussen, line 40: Seems like this was pulled out in https://github.com/wandb/weave/pull/2780?

jamie-rasmussen (approved):

PR #2791, opened 2024-10-28
first draft of router chain implementation with weave

umakrishnaswamy:
## Description

adds router chain cookbook example to cookbooks
- includes how to track results of prompt router chain
- includes how to evaluate final outputs using weave 

## Testing

How was this PR tested?

yarn start


scottire:
thanks for the contribution! 
Can you refactor to not subclass `weave.Object`. It's not something we really suggest users to do because it should just be a function or a `weave.Model`

scottire:
This looks great, can you change `model_output` to be `output` (a recent API change), other than that, this is good to go

umakrishnaswamy:

umakrishnaswamy:
recheck

PR #2796, opened 2024-10-28
chore(ui): ejecting markdown from tailwind defaulting

jtulk:
## Description

Part of fix for WB-21646


connieelee (approved):

connieelee (commented):
are there any config files that need to be updated here in weave too?

jamie-rasmussen (approved):

PR #2799, opened 2024-10-28
chore: add feedback replace endpoint that purges -> creates in one step

gtarpenning:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

Adds `feedback_replace` endpoint that purges then deletes in one request. This is necessary to allow quick edits in the new human feedback flow, as we want to auto-save once the user stops typing, but we don't want many different versions of the feedback. Without this endpoint the extra round-trip makes this very clunky to use. 

## Testing

test

### DIFF
--- a/tests/trace/test_feedback.py
+++ b/tests/trace/test_feedback.py
@@ -1,5 +1,11 @@
 import pytest
 
+from weave.trace_server.trace_server_interface import (
+    FeedbackCreateReq,
+    FeedbackQueryReq,
+    FeedbackReplaceReq,
+)
+
 
 def test_client_feedback(client) -> None:
     feedbacks = client.get_feedback()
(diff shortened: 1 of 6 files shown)


tests/trace/test_feedback.py
```diff
@@ -62,3 +68,68 @@ def test_custom_feedback(client) -> None:
...
+        project_id="test/project",
+        weave_ref="weave:///test/project/obj/123:abc",
+        feedback_type="note",
+        payload={"note": "Updated feedback"},
+        feedback_id=initial_feedback.id,
+    )
+    replaced_feedback = client.server.feedback_replace(replace_req)
+
+    # Verify the replacement
+    assert replaced_feedback.id == initial_feedback.id
```
tssweeney, line 104: this strikes me as odd - that it is the same underlying id. Can you share why you did not generate a new id?

tssweeney: i would expect this to be the opposite - the new feedback id should explicitly NOT be the original

tests/trace/test_feedback.py
```diff
@@ -45,6 +50,40 @@ def test_client_feedback(client) -> None:
...
+
+    feedback_note_res = client.server.feedback_create(
+        tsi.FeedbackCreateReq(
+            project_id=client._project_id(),
+            weave_ref=ref,
+            feedback_type="wandb.note",
+            payload=note,
+        )
+    )
+    assert feedback_note_res.payload == note
```
gtarpenning, line 84: this test fails in prod, without the small change to `feedback_create`

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1349,6 +1342,8 @@ def feedback_create(self, req: tsi.FeedbackCreateReq) -> tsi.FeedbackCreateRes:
             req.payload["detoned"] = detoned
             req.payload["detoned_alias"] = emoji.demojize(detoned)
             res_payload = req.payload
+        else:
+            res_payload = req.payload
```
gtarpenning, line 1346: I think this is a tiny bug, unless this was an intentional omission? cc @jamie-rasmussen

weave/trace_server/sqlite_trace_server.py
```diff
@@ -991,6 +989,8 @@ def feedback_create(self, req: tsi.FeedbackCreateReq) -> tsi.FeedbackCreateRes:
             req.payload["detoned"] = detoned
             req.payload["detoned_alias"] = emoji.demojize(detoned)
             res_payload = req.payload
+        else:
```
tssweeney, line 992: nit: put res_payload = req.payload outside of the conditional if both branches do it

tssweeney (approved):

weave/trace_server/clickhouse_trace_server_batched.py
jamie-rasmussen: I think it was intentional - in the case of emojis the server is doing this extra work computing the alias, detoned version, etc. In other cases there's not really a point in sending the payload back to the client, which already has a complete version which it included in the request, it would just waste bandwidth.

gtarpenning: just seems weird you can get a complete payload for emojis but nothing for all others. I get it though, changing this slightly

PR #2801, opened 2024-10-28
chore(weave): basic sdk for human feedback spec types

gtarpenning:
## Description

[WB-21670](https://wandb.atlassian.net/browse/WB-21670)

Basic SDK types for human feedback. Use the objects api to create a feedback spec, which can be used in the UI to generate human scorer columns 

## Testing

adds test. eventually we will want an integration test going from the type to loading the UI, but thats blocked on CI not sucking and the frontend landing

[WB-21670]: https://wandb.atlassian.net/browse/WB-21670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/interface/base_object_classes/human_annotation_column.py
```diff
@@ -0,0 +1,16 @@
+from typing import Optional
+
+from weave.trace_server.interface.base_object_classes import base_object_def
+
+
+class HumanAnnotationColumn(base_object_def.BaseObject):
```
tssweeney, line 6: note: "HumanAnnotationColumn" is a permanent name. It is sort of like a table name, so we have to be sure we like it (: I think there is a subtle world where there are AnnotationColumns that are not filled out by Humans... which might mean we drop the human part?

weave/trace_server/interface/base_object_classes/human_annotation_column.py
```diff
@@ -0,0 +1,16 @@
...
+    json_schema: dict
+
+    # If true, all unique creators will have their
+    # own value for this feedback type. Otherwise,
+    # by default, the value is shared and can be edited.
+    unique_among_creators: bool = False
+
+    # If provided, this feedback type will only be shown
+    # when a call is generated from the given op
+    op_scope: Optional[list[str]] = None
```
tssweeney, line 16: is this expected to be a ref string or op name?

weave/trace_server/interface/base_object_classes/human_annotation_column.py
```diff
@@ -0,0 +1,16 @@
+from typing import Optional
+
+from weave.trace_server.interface.base_object_classes import base_object_def
+
+
+class HumanAnnotationColumn(base_object_def.BaseObject):
+    json_schema: dict
```
tssweeney, line 7: this is probably fine for now. you could add a field validator that uses https://python-jsonschema.readthedocs.io/en/latest/validate/ to check for a schema error

weave/trace_server/interface/base_object_classes/human_annotation_column.py
gtarpenning: I think we want ref strings? like the frontend format `"weave:///griffin_wb/prod-evals-aug/op/Evaluation.summarize:*"`

weave/trace_server/interface/base_object_classes/human_annotation_column.py
gtarpenning: yep, that sounds right, do we then want a field within the class to denote that it came from a human?

weave/trace_server/interface/base_object_classes/human_annotation_column.py
tssweeney: i see in your tests that this is just the name itself. you might consider using a ref here. then if you wanted to match any version you can use `weave:///e/p/name:*` (just like we support with `op_name` in the call filter

tssweeney: oh, you commented before i finished reviewing. Looks like we are on the same page

gtarpenning: haha yeah okay fixing

weave/trace_server/interface/base_object_classes/human_annotation_column.py
tssweeney: There currently is `creator` and `wb_user_id` on the feedback table. I think using those are appropriate (which should be automatic right now)

weave/trace_server/interface/base_object_classes/annotation_column.py
```diff
@@ -0,0 +1,58 @@
...
+    op_scope: Optional[list[str]] = Field(
+        default=None,
+        examples=[
+            ["weave:///entity/project/op/name:digest"],
+            ["weave:///entity/project/op/name:*"],
+        ],
+    )
+
+    @field_validator("json_schema")
+    def validate_json_schema(cls, v: dict) -> dict:
```
tssweeney, line 56: interesting. there are sort of two things here:
1. if you want to explicitly type a subset of allowed json schemas (like you did here), just create a few BaseModels and make a union - no need for jsonschema.
2. If you want to allow any json schema, do:
```
try:
   jsonschema.validate(None,v)
except jsonschema.exceptions.SchemaError:
   raise InvalidJson
except jsonschema.exceptions.ValidationError:
   pass # we don't care that `None` does not conform
return v
```

weave/trace_server/interface/base_object_classes/annotation_column.py
```diff
@@ -0,0 +1,58 @@
+from typing import Optional
+
+import jsonschema
```
tssweeney, line 3: did you need to add this package? I wonder if the trace server already has this transitively.

gtarpenning: I dont understand what this means... I see that the trace server has it in the requirements, how do I access it "transitively"?

tssweeney: yeah, this is what I mean. in our pyproject.toml (inside core), there is no `jsonschema`. However, since `litellm` depends on `jsonschema`, you get it for free in trace server (when deployed).

tssweeney: but for these CI tests to pass you need to add jsonschema to the dev reqs

tssweeney: inside of 
```
test = [
  "nox",
  "pytest>=8.2.0",
  "pytest-asyncio>=0.23.6",
  "pytest-cov>=5.0.0",
  "pytest-xdist>=3.1.0",
  "pytest-rerunfailures>=12.0",
  "pytest-rerunfailures>=14.0",
  "clickhouse_connect==0.7.0",
  "fastapi>=0.110.0",
  "sqlparse==0.5.0",

  # Integration Tests
  "pytest-recording>=0.13.2",
  "vcrpy>=6.0.1",

  # serving tests
  "flask",
  "uvicorn>=0.27.0",
  "pillow",
  "filelock",
  "httpx",
]
```

gtarpenning: Yeah got it thanks

weave/trace_server/interface/base_object_classes/annotation_column.py
```diff
@@ -0,0 +1,48 @@
...
+from pydantic import Field, field_validator
+
+from weave.trace_server.interface.base_object_classes import base_object_def
+
+
+class AnnotationColumn(base_object_def.BaseObject):
```
gtarpenning, line 9: I keep going back and forth but I think this should actually be `AnnotationSpec`. Column doesn't really make that much sense.

tssweeney (approved):

PR #2806, opened 2024-10-29
fix(weave): Allow back-compat for "old-style" scorers and evaluations using `model_output` kwarg

andrewtruong:
## Description

Allows for back-compat with old-style scorers that used the `model_output` kwarg.

1. If any "old-style" scorers are used, the evaluation will return `model_output` and a warning is displayed to the user.
2. If ALL "new-style" scorers are used, the evaluation will return `output`

"old-style" scorers are defined as:
1. scoring functions that have a `model_output` argument; or
2. scoring classes that have a `score` method with a `model_output` argument.

"new-style" scorers have renamed `model_output -> output`

## Testing

Unit tests with both old-style and new-style scorers, and combinations of new-style, old-style, and classes.


tests/trace/test_evaluate_oldstyle.py
andrewtruong, line 1: This file re-implements the old-style tests, specifically:
1. outputs containing `model_output`; and 
2. scorers that take a `model_output` param

tests/trace/test_evaluate.py
```diff
@@ -153,29 +152,3 @@ def score(self, target, output):
             "mean": pytest.approx(0, abs=1),
         },
     }
-
-
-def test_multiclass_f1_score(client):
```
andrewtruong, line 158: This test is moved to `test_evaluate_oldstyle`

weave/scorers/base_scorer.py
```diff
@@ -16,7 +16,7 @@ class Scorer(Object):
         description="A mapping from column names in the dataset to the names expected by the scorer",
     )
 
-    def score(self, input: Any, target: Any, output: Any) -> Any:
+    def score(self, *, output: Any, **kwargs: Any) -> Any:
```
andrewtruong, line 19: This is more representative of the actual score func signature.  Some scorers don't take input, or may take other args

weave/flow/eval.py
```diff
@@ -339,7 +344,12 @@ async def predict_and_score(
                     raise ValueError(
                         f"{score_fn} expects arguments: {score_arg_names}, provide a preprocess_model_input function that returns a dict with those keys."
                     )
-            score_args["output"] = model_output
+            if "model_output" in score_arg_names:
+                util.warn_once(
+                    logger, "model_output is deprecated, please use output instead"
+                )
+                self._output_key = "model_output"
```
tssweeney, line 351: what happens if some scores use the new output and some use the old? I don't think this should be stored on the eval. just return `output` from predict and score

andrewtruong: are you saying for all scorers to return `output` instead of `model_output`?

tssweeney: i'm saying that `self._output_key` does not make sense. The output key for the arg is per-scorer, not global to the eval

andrewtruong: I think it does make sense to keep the output key consistent per eval though.  Otherwise if you mix and match scorers, should you get both `output` and `model_output` keys?  That doesn't seem right to me.

tssweeney: i think it is worth being clear here - all we are talking about is the key for the output of predict and score (i think?). Which if you miz am match is going to flip around given current implementation. the call stack is 

* predict_and_score
  * predict
  * score 1
  * score 2
  * score 3

the output of `predict_and_score` should not be keyed based on the last scorer param

weave/scorers/base_scorer.py
tcapelle: agree, and `input` is not necessary an available col.

weave/flow/eval.py
```diff
@@ -127,6 +133,20 @@ def model_post_init(self, __context: Any) -> None:
             else:
                 raise ValueError(f"Invalid scorer: {scorer}")
             scorers.append(scorer)
+
+        # Determine output key based on scorer types
+        if has_oldstyle_scorers(scorers):
+            self._output_key = "model_output"
```
tssweeney, line 139: Can we please fix this so we can merge it? it should not be assigning to self

weave/flow/eval.py
```diff
@@ -397,7 +420,7 @@ async def predict_and_score(
             scores[scorer_name] = result
 
         return {
-            "output": model_output,
+            self._output_key: model_output,
```
tssweeney, line 423: this should always be "output" even if you have some scorers that use "model_output"

weave/flow/eval.py
```diff
@@ -441,7 +463,7 @@ async def eval_example(example: dict) -> dict:
             except Exception as e:
                 print("Predict and score failed")
                 traceback.print_exc()
-                return {"output": None, "scores": {}}
+                return {self._output_key: None, "scores": {}}
```
tssweeney, line 466: this should always be "output" even if you have some scorers that use "model_output"

weave/flow/eval.py
```diff
@@ -458,7 +480,7 @@ async def eval_example(example: dict) -> dict:
             #     f"Evaluating... {duration:.2f}s [{n_complete} / {len(self.dataset.rows)} complete]"  # type:ignore
             # )
             if eval_row is None:
-                eval_row = {"output": None, "scores": {}}
+                eval_row = {self._output_key: None, "scores": {}}
```
tssweeney, line 483: this should always be "output" even if you have some scorers that use "model_output"

weave/flow/eval.py
```diff
@@ -370,7 +393,7 @@ async def predict_and_score(
                     for param in score_signature.parameters.values()
                     if param.default == inspect.Parameter.empty
                 ]
-                required_arg_names.remove("output")
+                required_arg_names.remove(self._output_key)
```
tssweeney, line 396: this should dynamic based on the scorer

weave/flow/eval.py
```diff
@@ -127,6 +133,20 @@ def model_post_init(self, __context: Any) -> None:
...
+        # Determine output key based on scorer types
+        if has_oldstyle_scorers(scorers):
+            self._output_key = "model_output"
+            util.warn_once(
+                logger,
+                "Using 'model_output' key for compatibility with older scorers. Please update scorers to use 'output' parameter.",
+            )
+        else:
+            self._output_key = "output"
+            util.warn_once(
```
tssweeney, line 146: why are we warning here?

tests/trace/test_evaluate.py
```diff
@@ -155,27 +154,77 @@ def score(self, target, output):
     }
 
 
-def test_multiclass_f1_score(client):
+@pytest.mark.asyncio
+async def test_basic_evaluation_with_mixed_scorer_styles(client):
```
tssweeney, line 158: @andrewtruong i added this to show the expected behavior.

andrewtruong: I see what you're trying to show here and I'm ok with this behaviour, but this will break for anyone who currently relies on the "output" key being `model_output`.

The alternative is safer -- if we detect any "old-style" scorers, then we use `model_output` (does not break older code).  If only new-style, then we use `output`.  If there's a mix, we use `output` and give a warning.

tssweeney: ok, i don't love this, but i see your perspective.

tssweeney (approved):

weave/flow/eval.py
tcapelle: I think this should be `required_arg_names.remove(score_output_name)` instead or we could just remove this line.

PR #2810, opened 2024-10-29
chore(weave): Add `weave_ts`, `dev`, and `app` scopes

andrewtruong:


jamie-rasmussen (approved):
Feels like we should be consistent about dash vs. underscore. We might also consider allowing `dev`  and `app` (for common components, matching core repo)

andrewtruong:
Aren't those UI changes?

jamie-rasmussen:
> Aren't those UI changes?

Yes. What I would like to do is reduce friction for contributors coming from the models side of the house. They will by habit be used to specifying "app" for frontend changes. For us, the distinction between common components and the Weave UI is probably not that useful of one to make, though maybe it serves as a bit of a hint to reflect on whether your change might affect the core app.

PR #2811, opened 2024-10-29
feat(weave): prompts as first class objects

jamie-rasmussen:
## Description

First release of first-class Prompt object including SDK, UI, and docs.

https://wandb.atlassian.net/browse/WB-20795
https://wandb.atlassian.net/browse/WB-20796

## Testing

How was this PR tested?

### DIFF
--- a/docs/docs/guides/core-types/prompts.md
+++ b/docs/docs/guides/core-types/prompts.md
@@ -1,0 +1,373 @@
+# Prompts
+
+Creating, evaluating, and refining prompts is a core activity for AI engineers.
+Small changes to a prompt can have big impacts on your application's behavior.
+Weave lets you create prompts, save and retrieve them, and evolve them over time.
+Some of the benefits of Weave's prompt management system are:
+
+- Unopinionated core, with a batteries-included option for rapid development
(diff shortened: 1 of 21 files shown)


docs/docs/guides/core-types/prompts.md
```diff
@@ -0,0 +1,373 @@
+# Prompts
```
tssweeney, line 1: Very nice writeup

weave/trace/refs.py
```diff
@@ -144,6 +144,19 @@ def uri(self) -> str:
             u += "/" + "/".join(refs_internal.extra_value_quoter(e) for e in self.extra)
         return u
 
+    def objectify(self, obj: Any) -> Any:
```
tssweeney, line 147: This is really nice. ccing @andrewtruong for viz. Not needed for this PR, but i would imagine that this sort of dispatching logic would be really nice for other object types as well.

Note: here we are assigning the `ref` to obj.ref. However, in the event that the user mutates the prompt, this ref should be invalidated. We have this issue with other objects as well.

tssweeney (approved):
🚢 !! First class prompts!

PR #2812, opened 2024-10-29
chore(ui): check Op code for literal credentials

jamie-rasmussen:
## Description

https://wandb.atlassian.net/browse/WB-21683

After:

Screenshot 2024-10-29 at 1 06 41 PM

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse2/Browse2OpDefCode.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse2/Browse2OpDefCode.tsx
@@ -3,6 +3,7 @@ import Box from '@mui/material/Box';
 import {Loading} from '@wandb/weave/components/Loading';
 import React, {FC} from 'react';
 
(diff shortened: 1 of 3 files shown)


gtarpenning (approved):
I wonder if the message doesn't convey enough information for the user to act on it? Like if I see `'<Redacted: string contains Authorization pattern>'` I'm not sure I would instantly know that this is generated from `{"authorization": ...}`. Although i guess we should always be sanitizing and this is just for existing key stuff?

jamie-rasmussen:
> I wonder if the message doesn't convey enough information for the user to act on it? Like if I see `'<Redacted: string contains Authorization pattern>'` I'm not sure I would instantly know that this is generated from `{"authorization": ...}`. Although i guess we should always be sanitizing and this is just for existing key stuff?

Yes, this is a last line of defense in case of sanitization failures. I agree that it isn't a very actionable message but not clear what we would want user to do about it either.

PR #2818, opened 2024-10-30
feat(weave): enable dictify of inputs and outputs

jamie-rasmussen:
## Description

Internal Jira: https://wandb.atlassian.net/browse/WB-21699

Adds sanitization to the dictify method, then turns dictify back on for call inputs and outputs only. (An explicit flag needs to be passed for `to_json` to dictify, so code saving won't use it.)

For later we should also take a closer look at why code saving is using to_json - it doesn't seem right to me as you're not going to get runnable code from that and also I think the memory address that is part of the default repr is going to cause your op versions to increase unexpectedly.

---

Update: @tssweeney pointed out that dictify'ing the output of dspy made it unusable when previously it was OK (due to the custom repr). Given the purpose of a repr, it seemed to me that if a custom one was made we should by default use that instead of dictify'ing. However, in the Gemini integration, we explicitly want the inputs and output dictify-ed so that we don't truncate the strings and can build the chat view from structured data. 

## Testing

How was this PR tested?

### DIFF
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -666,7 +666,6 @@ def test_saveload_customtype(client):
     assert obj2.b == "x"
 
 
-@pytest.mark.skip(reason="Re-enable after dictify is fixed")
 def test_save_unknown_type(client):
     class SomeUnknownThing:
         def __init__(self, a):
@@ -675,14 +674,7 @@ def test_save_unknown_type(client):
     obj = SomeUnknownThing(3)
     ref = client._save_object(obj, "my-np-array")
     obj2 = client.get(ref)
-    assert obj2 == {
-        "__class__": {
-            "module": "test_weave_client",
-            "qualname": "test_save_unknown_type.<locals>.SomeUnknownThing",
-            "name": "SomeUnknownThing",
-        },
-        "a": 3,
-    }
+    assert obj2 == repr(obj)
 
 
 def test_save_model(client):
--- a/weave/integrations/google_ai_studio/google_ai_studio_sdk.py
+++ b/weave/integrations/google_ai_studio/google_ai_studio_sdk.py
@@ -5,10 +5,17 @@ from typing import TYPE_CHECKING, Any, Callable, Optional
 import weave
 from weave.trace.op_extensions.accumulator import add_accumulator
 from weave.trace.patcher import MultiPatcher, SymbolPatcher
+from weave.trace.serialize import dictify
 from weave.trace.weave_client import Call
 
 if TYPE_CHECKING:
     from google.generativeai.types.generation_types import GenerateContentResponse
+
+
+def gemini_postprocess_inputs(inputs: dict[str, Any]) -> dict[str, Any]:
+    if "self" in inputs:
+        inputs["self"] = dictify(inputs["self"])
+    return inputs
 
 
 def gemini_accumulator(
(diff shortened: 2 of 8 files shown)


tests/integrations/dspy/dspy_test.py
```diff
@@ -106,21 +106,37 @@ def test_dspy_inline_signatures(client: WeaveClient) -> None:
     call_1, _ = flattened_calls[0]
     assert call_1.exception is None and call_1.ended_at is not None
     output_1 = call_1.output
-    assert (
-        output_1
-        == """Prediction(
-    sentiment='Positive'
-)"""
-    )
+    assert output_1 == {
```
tssweeney, line 109: This does not seem desirable. Specifically: 
1. the output no longer contains `Prediction's "positive" `sentiment` value.
2. It is much more verbose
3. It is capturing information about the class object (eg. `completions`) which it appears is a member of the class.

tests/trace/test_weave_client.py
```diff
@@ -675,14 +674,7 @@ def __init__(self, a):
...
     obj2 = client.get(ref)
-    assert obj2 == {
-        "__class__": {
-            "module": "test_weave_client",
-            "qualname": "test_save_unknown_type.<locals>.SomeUnknownThing",
-            "name": "SomeUnknownThing",
-        },
-        "a": 3,
-    }
+    assert obj2 == repr(obj)
```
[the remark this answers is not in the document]
jamie-rasmussen: Yes, this test (which I had changed but then skipped) was using the old logic which was more aggressive about what would be dict'ified.

weave/integrations/google_ai_studio/google_ai_studio_sdk.py
```diff
@@ -5,12 +5,19 @@
...
 from weave.trace.op_extensions.accumulator import add_accumulator
 from weave.trace.patcher import MultiPatcher, SymbolPatcher
+from weave.trace.serialize import dictify
 from weave.trace.weave_client import Call
 
 if TYPE_CHECKING:
     from google.generativeai.types.generation_types import GenerateContentResponse
 
 
+def gemini_postprocess_inputs(inputs: dict[str, Any]) -> dict[str, Any]:
```
tssweeney, line 15: why is this needed? Shouldn't the to_json in call start handle this now?

jamie-rasmussen: It's because of this:
> Given the purpose of a repr, it seemed to me that if a custom one was made we should by default use that instead of dictify'ing. However, in the Gemini integration, we explicitly want the inputs and output dictify-ed so that we don't truncate the strings and can build the chat view from structured data.

In the latest iteration we no longer attempt to dict'ify if there is a custom repr, which the Gemini objects have, so we have to explicitly do it as part of the integration.

tssweeney: I see. Interestingly, this makes me realize that this PR effectively has 2 distinct user-facing changes:
1. Post-processing for Gemini
2. A more general purpose input/output routine.

Given that #2 runs the risk of bigger payloads / aggressive dumps, would it make sense to split this to land Gemini first, which we know is less likely to need revision?

jamie-rasmussen: OK, I changed the general inputs/output case to `use_dictify=False` and updated the PR.

weave/trace/serialize.py
```diff
@@ -168,12 +197,12 @@ def dictify(
                 else:
```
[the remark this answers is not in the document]
jamie-rasmussen: Do you have an example? The ones I'm thinking of would be handled by the `callable` check.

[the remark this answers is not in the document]
tssweeney: let's say we had something like:
```
class APIResponse():
   response_class: Type

res = APIResponse(response_class=MyBaseModel)
```

My understanding is that this would walk into MyBaseModel and traverse all of it's properties as well (which are types by that point). Consider https://github.com/openai/openai-python/blob/6e42e7879a8e22db9268d1ef2f3b80428534e752/src/openai/_response.py#L48 which is the base of all openai responses

tssweeney (approved):

weave/trace/serialize.py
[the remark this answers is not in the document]
jamie-rasmussen: I think it wouldn't, as `BaseAPIResponse` is `callable`:
```
from openai._response import BaseAPIResponse
assert callable(BaseAPIResponse)
```

PR #2819, opened 2024-10-30
chore(ui): prevent formatter from auto-removing React import

andrewtruong:


weave-js/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": false,
     "isolatedModules": true,
-    "jsx": "preserve",
+    "jsx": "react",
```
andrewtruong, line 12: This setting is already used in the frontend
https://github.com/wandb/core/blame/master/frontends/app/tsconfig.json#L15

jamie-rasmussen (approved):

PR #2828, opened 2024-10-31
chore(ui): fix row selection in object versions page

harukatab:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

- When a user clicks a row in Objects page (ObjectVersionsPage.tsx), it didn't keep the correct row selection from the URL. This PR fixes that by parsing the URL into RefURI and using the consistent peek ID to match with row IDs

Before (left) vs. After (right)

https://github.com/user-attachments/assets/fef69fe7-bd9a-44cf-9603-1854e345f792

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectVersionsPage.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectVersionsPage.tsx
@@ -51,7 +51,10 @@ import {
   isTableRef,
   makeRefExpandedPayload,
 } from './wfReactInterface/tsDataModelHooksCallRefExpansion';
-import {objectVersionKeyToRefUri} from './wfReactInterface/utilities';
+import {
+  objectVersionKeyToRefUri,
+  parseUrlPathToRefUri,
+} from './wfReactInterface/utilities';
 import {
   KnownBaseObjectClassType,
   ObjectVersionSchema,
@@ -332,7 +335,10 @@ const ObjectVersionsTable: React.FC<{
   // Highlight table row if it matches peek drawer.
   const query = useURLSearchParamsDict();
   const {peekPath} = query;
-  const peekId = peekPath ? peekPath.split('/').pop() : null;
+
+  const {baseRouter} = useWeaveflowRouteContext();
+  const peekId = parseUrlPathToRefUri(peekPath, baseRouter);
+
   const rowIds = useMemo(() => {
     return rows.map(row => row.id);
   }, [rows]);
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
@@ -286,6 +286,43 @@ export const objectVersionNiceString = (ov: ObjectVersionSchema) => {
   return result;
 };
 
+export function parseUrlPathToRefUri(
+  urlPath: string | undefined
+): string | null {
+  if (!urlPath) {
+    return null;
+  }
+
+  // Remove leading and trailing slashes
+  const cleanPath = urlPath.replace(/^\/|\/$/g, '');
+
+  // Split the path into segments
+  const segments = cleanPath.split('/');
+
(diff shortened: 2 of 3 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
```diff
@@ -286,6 +286,39 @@ export const objectVersionNiceString = (ov: ObjectVersionSchema) => {
...
+
+  // Construct RefUri for object version
+  return objectVersionKeyToRefUri({
+    scheme: 'weave',
+    entity,
+    project,
+    objectId,
+    versionHash,
+    path: '',
+    refExtra: '',
```
harukatab, line 317: path isn't in the peekPath passed into this function. e.g. `peekPath = "/haruka/test-weave/objects/grammar/versions/ARz6FRfbWRsPMipJTShviHZicJVQyJTtLsDXI9gSXnM"`. Is it valid to hard code `path: 'obj'`? 

Also, i think `refExtra` isn't passed in a lot of `objectVersionKeyToRefUri` call sites so I'm comfortable leaving the refExtra empty but if anyone has a suggestion,  let me know!

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectVersionsPage.tsx
```diff
@@ -332,7 +335,9 @@ const ObjectVersionsTable: React.FC<{
   // Highlight table row if it matches peek drawer.
   const query = useURLSearchParamsDict();
   const {peekPath} = query;
-  const peekId = peekPath ? peekPath.split('/').pop() : null;
+
+  const peekId = parseObjectVersionUrlPathToRefUri(peekPath);
```
gtarpenning, line 339: we might want to name this to reflect that it is the URI, or a comment saying that peekId == rowId == objectURI

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
tssweeney: path and ref extra should both be null/empty. these are old things

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
```diff
@@ -286,6 +286,39 @@ export const objectVersionNiceString = (ov: ObjectVersionSchema) => {
...
+  const segments = cleanPath.split('/');
+
+  // We need at least 6 segments to be a valid object version path
+  if (segments.length < 6) {
+    return null;
+  }
+  const entity = segments[0];
+  const project = segments[1];
+  const objectId = segments[3];
+  const versionHash = segments[5];
```
gtarpenning, line 307: it feels a bit suspicious that this is the first place we are doing a Url to object key conversion. I understand why, but it feels like the object page router would be doing this as well. That being said I can't find a useful existing function to do this....

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
```diff
@@ -286,6 +286,39 @@ export const objectVersionNiceString = (ov: ObjectVersionSchema) => {
   return result;
 };
 
+export function parseObjectVersionUrlPathToRefUri(
```
tssweeney, line 289: i'd consider using `usePeekLocation` which does some of this handling for you, and you can just map that to the URI. This will reduce the number of places we are doing manual URI parsing

gtarpenning (approved):
I think this is good, but @jamie-rasmussen might want to once-over?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
gtarpenning: yeah like Tim said: `usePeekLocation`

PR #2832, opened 2024-10-31
chore(weave_ts): Cleanup netrc

andrewtruong:
## Description
Adds better .netrc parsing

## Testing

Unit tests


jamie-rasmussen (approved):
Looking at https://everything.curl.dev/usingcurl/netrc.html it appears that there may be netrc files in the wild this code (before and after) would have trouble parsing, e.g. single line definitions. But if it hasn't been a problem so far probably won't in future. No action required, just noting it.

PR #2862, opened 2024-11-04
Haize Labs <> wandb weave integration 

lithafnium:
## Description

Hey folks! This is Steve from Haize Labs, cc'ing @scottire for visibility. Excited to get this integration going! 

For those just getting involved, Haize Labs is in the business of stress-testing AI systems. One part of our evaluation suite is the act of "haizing," or finding and surfacing new vulnerabilities through automated, dynamic redteaming. More info on our work can be found on our [blog](https://blog.haizelabs.com/). 

Given the observability implications of our attack suite, the teams at Haize and weave thought a collaboration was only natural. 

With that said, the current developer flow looks like the following: 
1. Generate an api key on our platform: https://platform.haizelabs.com/
2. Instantiate a `HaizeEvaluation` and a `HaizeDataset,` along with a function representing calling your AI model or application. The function expects a single string input and should only return a single string. 
3. Run `evaluate.evaluation(func, dependencies=[])`, where dependencies contains a list of additional packages needed for your function to execute. We essentially run the developer's function in an isolated sandbox such that our attack suite can query and call to guide our optimization process. 

Example terminal output: 
image

Example results on wandb: 
image

Some questions to consider: 
1. Where's the best place to put this code? 
2. What additional testing should be done before launching this integration? 
3. Any fields that a user might want to get logged? 

Any feedback would be appreciated! Super excited to get this out there!

### DIFF
--- a/weave/flow/haize_eval.py
+++ b/weave/flow/haize_eval.py
@@ -1,0 +1,249 @@
+from typing import Callable, Union, Optional, List, Dict, Any, Set
+from rich import print
+from rich.table import Table
+from rich.live import Live
+from rich.console import Console, Group
+from rich.spinner import Spinner
+from weave.trace.op import Op
+from pydantic import BaseModel
+from collections import defaultdict
+from pydantic import field_validator
+
+import cloudpickle
+import requests
+import weave
+import uuid
+import asyncio
+import base64
+import aiohttp
+import signal
+import platform
+
+console = Console()
+
+statuses = ["SAFE", "WARNING", "UNSAFE"]
+
+
+class Message(BaseModel):
+    role: str
+    content: str
+
+
+class Result(BaseModel):
+    id: str
+    conversation: List[Message]
+    categories: List[str]
+    score: float
+    status: int
+
+
+class ResultOutput(BaseModel):
+    intent: str
+    results: List[Result]
+
+
+def print_table(new_results):
+    table = Table()
+    headers = ["Intent", "# Safe", "# Warning", "# Unsafe"]
+    for header in headers:
+        table.add_column(header)
+
+    rows = [[intent, *[str(r) for r in new_results[intent]]] for intent in new_results]
+    rows.sort(key=lambda x: x[0])
+
+    for row in rows:
+        table.add_row(*row, end_section=True)
+
+    return Group(
+        table,
+        Spinner("dots", text="[bold green]Running Haize..."),
+    )
+
+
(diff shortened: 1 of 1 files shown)


scottire:
Thanks for the contribution! We'll review it soon

andrewtruong:
Hey @lithafnium , thanks for the PR!

A few questions:
1. IIUC "haizing" properly, you're basically taking a model output, passing it through your proprietary scoring system and returning SAFE/UNSAFE -- is that correct?
2. I see you're also cloudpickling the function -- why do you need to do that?  Are the inputs/outputs not sufficient?
3. What are "behaviors", and why do they need to be part of the dataset row?

lithafnium:
> Hey @lithafnium , thanks for the PR!
> 
> A few questions:
> 
> 1. IIUC "haizing" properly, you're basically taking a model output, passing it through your proprietary scoring system and returning SAFE/UNSAFE -- is that correct?
> 2. I see you're also cloudpickling the function -- why do you need to do that?  Are the inputs/outputs not sufficient?
> 3. What are "behaviors", and why do they need to be part of the dataset row?

Hey @andrewtruong ! Thanks for taking a look. 

As you may know, we at Haizelabs are in the business of stress testing models. One such approach that we take is  automated, diverse redteaming, which we call "haizing." As part of our product offering we have a full suite of proprietary redteaming algorithms that surface vulnerabilities in the form of attack prompts in order to induce a given harmful behavior. 

To answer your questions specifically, 
1. We don't simply pass in model outputs through our judging system. We run our algorithms on our backend to produce attack prompts, pass the prompts to the model/AI application, and use the responses as feedback  in order to come up and develop new attack prompts. This leads to point no. 2:
2. The reason why we need to pickle the function is to upload the model/function to an isolated sandbox such that our attack algorithms can query them for a response. This approach can modified, if need be. We essentially just need a method to get inference from the provided models that our redteaming algos. can query. 
3. Behaviors are a list of "intents" or "instructions" that you don't wish your model to carry out. Some examples include "how to make a bomb" or "write me a hacking script to break into a server" 

Let me know if you have any questions! 

cc'ing @scottire

andrewtruong:
Thanks, so now I'm imagining the workflow like:
1. Upload model code and a set of "behaviours" to Haize
3. You create a series of inputs and evaluate the model's outputs to see if they are "behaving"
3. You return a list of rows where each row contains the input, output, and "behaviours".

Conceptually, I'm mapping Haize's "behaviour" to Weave's "Scorer".

Does that sound right to you?

PR #2863, opened 2024-11-05
chore(weave): Add util to strip EXIF Metadata (and save 30-50% filesize per img)

andrewtruong:
EXIF: https://en.wikipedia.org/wiki/Exif

Strips identifying metadata and makes for smaller images that load faster!


tssweeney (approved):

strip_exif.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
```
tssweeney, line 1: nit - belongs in scripts

PR #2864, opened 2024-11-05
feat(weave_query): Add string op to parse number with thousands and decimal separators

nicholaspun-wandb:
## Description

- Handles [WB-16442](https://wandb.atlassian.net/browse/WB-16442)
- Creates a new string op that will parse numbers with a thousands separator.

## Testing

Added unit tests

[WB-16442]: https://wandb.atlassian.net/browse/WB-16442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave_query/tests/test_basic_ops.py
+++ b/weave_query/tests/test_basic_ops.py
@@ -94,12 +94,12 @@ def test_string_ops():
     # assert weave.use(foo in foobar) == True # Broken
     # assert weave.use(foobar in foo) == False # Broken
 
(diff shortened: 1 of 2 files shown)


domphan-wandb:
I think this solution is pretty sweet especially since you removed the locale calls. Do you think it would feasible to add a second optional argument to represent the decimal symbol if needed? For example, I think `1.234.567,14` is valid in many European countries.

jwlee64 (approved):

PR #2865, opened 2024-11-05
chore(weave): Implement enhaced feedback structure and mvp filter/query layer

tssweeney:
This PR lays the groundwork for the next leg of feedback types in our system. Specifically, we have two "classes" of feedback: `runnables` and `annotations`. "Runnables" are feedbacks that are generated by running a program (think: Op, Configured Action, Scorer), while "Annotations" are feedbacks created by humans with specific types (aka human in the loop, aka custom columns, etc...).

There were three problems to solve with this emerging data model:
1. How do we store additional metadata about these feedbacks linking them to other objects in our system.
2. How do we group / collect feedbacks that belong to the same logical "column" or concept?
3. Given the above, how can we filter/sort without reading large JSON dumped columns?

After much iteration and discussion, the solution that seemed most suitable is as follows:
1. `feedback_type` how has 2 special prefixes: `wandb.runnable` and `wandb.annotation`, where the total type should be `wandb.runnable.RUNNABLE_NAME` or `wandb.annotation.ANNOTATION_NAME`. here, `RUNNABLE_NAME` or `ANNOTATION_NAME` are the `name` (aka `object_id`) components of the backing Object or Op. This is the most common group key and indexed in Clickhouse already.
2. I have added 4 new columns to the feedback table. Note, I originally had these as fields in the payload itself, but this would result in more complex, heavy lookups & more ridged structure over the payload itself. This approach allows us to put our foreign keys in columns that can be indexed in the future if needed.:
    - `annotation_ref`: The ref pointing to the annotation definition for this feedback.
    - `runnable_ref`: The ref pointing to the runnable definition for this feedback.
    - `call_ref`: The ref pointing to the resulting call associated with generating this feedback.
    - `trigger_ref`: The ref pointing to the trigger definition which resulted in this feedback.
3. When these types of feedback are entered into the DB, our server now will enforce that these ref values are filled out when required and matched the correct format. Moreover, the payloads themselves conform to a very simple structure: 
```
class AnnotationPayloadSchema(BaseModel):
    value: Any


class RunnablePayloadSchema(BaseModel):
    output: Any
```
4. Finally, I implement basic (not optimized yet) support for filter and sort on the calls query using the notation `feedback[feedback_type].payload.json.selector`. This allows us to specify the feedback type (while supporting dots) and match our other field access patterns.

----

With all of this together, we can have code like:
```python
@weave.op
def my_scorer(x: int, output: int) -> int:
    expected = ["a", "b", "c", "d"][x]
    return {
        "model_output": output,
        "match": output == expected,
    }

@weave.op
def my_model(x: int) -> str:
    return [
        "a",
        "x",  # intentional "mistake"
        "c",
        "y",  # intentional "mistake"
    ][x]

ids = []
for x in range(4):
    _, c = my_model.call(x)
    ids.append(c.id)
    # Note: `_apply_scorer` is not user-facing (yet!) but will be made public during the eval api project.
    c._apply_scorer(my_scorer)
```

... then query ...

```python
calls = client.server.calls_query_stream(
    tsi.CallsQueryReq(
        filter=tsi.CallsFilter(op_names=[get_ref(my_model).uri()]),
        # Filter down to just correct matches
        query={
            "$expr": {
                "$eq": [
                    {
                        "$getField": "feedback.[wandb.runnable.my_scorer].payload.output.match"
                    },
                    {"$literal": "true"},
                ]
            }
        },
        # Sort by the model output desc
        sort_by=[
            {
                "field": "feedback.[wandb.runnable.my_scorer].payload.output.model_output",
                "direction": "desc",
            }
        ],
    )
)
```

This can easily be extended to support different aggregation logic and specific version selectors.


tests/trace/test_evaluations.py
```diff
@@ -961,7 +962,14 @@ def score(text, model_output) -> bool:
     feedbacks = calls.calls[0].summary["weave"]["feedback"]
     assert len(feedbacks) == 1
     feedback = feedbacks[0]
-    assert feedback["feedback_type"] == SCORE_TYPE_NAME
-    assert feedback["payload"]["name"] == "score"
-    assert feedback["payload"]["op_ref"] == get_ref(score).uri()
-    assert feedback["payload"]["results"] == True
+    assert feedback["feedback_type"] == "wandb.runnable.score"
```
tssweeney, line 965: This is ok to change as the UI/query layer does not consume it yet.

tests/trace/test_scores.py
```diff
@@ -39,9 +38,8 @@ def my_score(input_x: int, model_output: int) -> int:
 
     assert len(calls) == 2
     feedback = calls[0].summary["weave"]["feedback"][0]
-    assert feedback["feedback_type"] == SCORE_TYPE_NAME
+    assert feedback["feedback_type"] == "wandb.runnable.my_score"
```
tssweeney, line 41: again, safe to change now that we have a good format

weave/trace/feedback_types/score.py
```diff
@@ -1,13 +0,0 @@
-# This type is still "Beta" and the underlying payload might change as well.
-# We're using "beta.1" to indicate that this is a pre-release version.
-from typing import TypedDict
-
-SCORE_TYPE_NAME = "wandb.score.beta.1"
```
tssweeney, line 5: we learned from this - no longer needed

weave/trace_server/calls_query_builder.py
```diff
@@ -209,7 +258,8 @@ def as_sql(self, pb: ParamBuilder, table_alias: str) -> str:
         if self._consumed_fields is None:
             self._consumed_fields = []
             for field in conditions.fields_used:
-                self._consumed_fields.append(get_field_by_name(field))
+                # TODO: Verify that this is "ok" since before we were just looking at field name
```
gtarpenning, line 261: didn't we change `get_field_by_name` in this pr to make this work?

weave/trace_server/calls_query_builder.py
```diff
@@ -556,23 +614,36 @@ def _as_sql_base_format(
...
 
+        feedback_join_sql = ""
+        feedback_where_sql = ""
+        if needs_feedback:
+            feedback_where_sql = (
+                f" AND calls_merged.project_id = {_param_slot(project_param, 'String')}"
+            )
+            feedback_join_sql = f"""
+            LEFT JOIN feedback
+            ON (feedback.weave_ref = concat('weave-trace-internal:///', {_param_slot(project_param, 'String')}, '/call/', calls_merged.id))
```
gtarpenning, line 635: any reason to do this concat in the query vs outside and pass it in?

weave/trace_server/feedback.py
```diff
@@ -28,6 +35,65 @@
...
+
+
+@overload
+def _ensure_ref_is_valid(
+    ref: str,
+    expected_type: Tuple[Type[T], ...],
+) -> T: ...
+
+
+def _ensure_ref_is_valid(
```
gtarpenning, line 72: this feels like just a util fn?

weave/trace_server/feedback.py
```diff
@@ -39,6 +105,72 @@ def validate_feedback_create_req(req: tsi.FeedbackCreateReq) -> None:
                 f"Invalid payload for feedback_type {req.feedback_type}: {e}"
             )
 
+    # Validate the required fields for the feedback type.
+    is_annotation = req.feedback_type.startswith(ANNOTATION_FEEDBACK_TYPE_PREFIX)
+    is_runnable = req.feedback_type.startswith(RUNNABLE_FEEDBACK_TYPE_PREFIX)
```
gtarpenning, line 110: these might be nice as helper functions

```
def is_annotation(feedback_type: str) -> bool
```

weave/trace_server/orm.py
```diff
@@ -48,10 +48,20 @@ def __init__(
...
+            if param_value in self._param_to_name:
+                return self._param_to_name[param_value]
+        except TypeError:
+            pass
         param_name = self._prefix + str(len(self._params))
         self._params[param_name] = param_value
+        try:
+            self._param_to_name[param_value] = param_name
+        except TypeError:
+            pass
```
gtarpenning, line 64: what happens when this silently fails? This returns early if it already exists, then tries to set if it doesn't?

weave/trace_server/trace_server_interface.py
```diff
@@ -686,6 +686,18 @@ class FeedbackCreateReq(BaseModel):
             }
         ]
     )
+    annotation_ref: Optional[str] = Field(
```
gtarpenning, line 689: it would be nice if we could type this to a kind of ref, like `objectRef`, with a pydantic validator and then check its construction in the client.

gtarpenning (approved):
This new feedback query is going to be spicy in big projects, but looks good.  The calls query builder is also feeling... clunky. Generally this makes sense, I wonder how much of the implementation we can abstract away from the user when adding, but still create an intuitive way for them to get the data out. It's possible that we might want to have some way of auto-constructing queries client-side, i'm imagining users not finding the following easy to use.... 
```"$getField": "feedback.[wandb.runnable.my_scorer].payload.output.match"```

weave/trace_server/calls_query_builder.py
tssweeney: 👍

weave/trace_server/calls_query_builder.py
tssweeney: i don't think so? but not sure. We have to do a concat either way since the last part is dynamic

weave/trace_server/feedback.py
tssweeney: moved

weave/trace_server/feedback.py
tssweeney: extracted

weave/trace_server/orm.py
tssweeney: i made this code more readable and commented

weave/trace_server/trace_server_interface.py
tssweeney: Agreed

PR #2781, opened 2024-10-25
Team Friday Ship: Online Evals Integration Branch: Weave

tssweeney:

PR #2882, opened 2024-11-05
chore(weave): Online Evals Part 1: Backend support for on-demand LLM Judges

tssweeney:
Important: Based off of https://github.com/wandb/weave/pull/2865.

This PR extracts the backend-only changes to implement on-demand LLM Judges from the larger online evals PR: https://github.com/wandb/weave/pull/2781.

This is the first in a series of "extractions"
1. (this PR) Extract Backend Support for on-demand judgements. Key Features:
    - Method to execute small batches on-demand
    - Ability to use customer API keys (builds off the secret manager work)
    - Ability to store results in new feedback data model (the above https://github.com/wandb/weave/pull/2865)
    - Builds on the object validation PR (and naming for actions)
2. (https://github.com/wandb/core/pull/25442) Expose this method via a new endpoint (extract from https://github.com/wandb/core/pull/25146)
    - This will setup the necessary HTTP endpoints server-side
3. (https://github.com/wandb/weave/pull/2902) Extract the frontend changes from  https://github.com/wandb/weave/pull/2781 to support on-demand LLM Judges
     - This will allow users to setup and execute these judges from the UI
     - Setups up Scorer management page
     - Exposes metrics in the UI
4. Implementation of Online Evals. This will require core changes, frontend changes, and more backend changes


weave/actions_worker/tasks.py
```diff
@@ -0,0 +1,102 @@
+import json
+from functools import partial
+from typing import Any, Tuple
+
+from openai import OpenAI
+
+from weave.trace_server.feedback import RunnablePayloadSchema
```
tssweeney, line 7: remove this dep (move feedback to the correct location)

tests/integrations/litellm/test_actions_lifecycle_llm_judge.py
```diff
@@ -0,0 +1,211 @@
...
+)
+from weave.trace_server.trace_server_interface import (
+    ActionsExecuteBatchReq,
+)
+
+
+class DummySecretFetcher:
+    def fetch(self, secret_name: str) -> dict:
+        return {
+            "secrets": {secret_name: os.environ.get(secret_name, "DUMMY_SECRET_VALUE")}
```
andrewtruong, line 29: nit: `os.getenv`

tests/trace/test_actions_lifecycle.py
```diff
@@ -0,0 +1,115 @@
...
+from weave.trace_server.interface.base_object_classes.actions import (
+    ActionDefinition,
+)
+from weave.trace_server.trace_server_interface import (
+    ActionsExecuteBatchReq,
+    FeedbackCreateReq,
+)
+
+
+def test_action_lifecycle_word_count(client: WeaveClient):
```
andrewtruong, line 15: What are the significant differences between this test and the one above?

I can see (but dont understand the significance of):
1. Difference in `ActionDefinition` spec
2. An added manual feedback (?)

weave/trace_server/actions_worker/actions/contains_words.py
```diff
@@ -0,0 +1,24 @@
...
+    TraceServerInterface,
+)
+
+
+def do_contains_words_action(
+    project_id: str,
+    config: ContainsWordsActionSpec,
+    call: CallSchema,
+    trace_server: TraceServerInterface,
+) -> Any:
```
andrewtruong, line 18: A few questions here:
1. Will actions always be functions with this shape?
2. Will users be able to define their own actions?
3. Would it ever be preferable to write actions as classes?

tests/integrations/litellm/test_actions_lifecycle_llm_judge.py
tssweeney: ack

tests/trace/test_actions_lifecycle.py
tssweeney: the word count action spec is stuper simple whereas the llm one requires the full litellm code path.

The manual part is just a demonstration of being able to manually create this feedback type - really only need to validate this once. Could be pulled out to it's own pr

weave/trace_server/actions_worker/dispatcher.py
```diff
@@ -0,0 +1,135 @@
...
+    weave_ref = InternalCallRef(project_id, call_id).uri()
+    parsed_action_ref = parse_internal_uri(action_ref)
+    if not isinstance(parsed_action_ref, (InternalObjectRef, InternalOpRef)):
+        raise ValueError(f"Invalid action ref: {action_ref}")
+    action_name = parsed_action_ref.name
+    return trace_server.feedback_create(
+        FeedbackCreateReq(
+            project_id=project_id,
+            weave_ref=weave_ref,
+            feedback_type="wandb.runnable." + action_name,
```
andrewtruong, line 130: magic str, extract as const

weave/trace_server/actions_worker/actions/contains_words.py
tssweeney: 1. yes, we might add to this in the future - it is purely internal implementation detail, so allowed to change
2. No. Instead they would define their own op and define an action that hits that op.
3. Maybe. again, this is all backend implementation for now. I was just keeping it simple for now until we have more examples to pattern match

weave/trace_server/actions_worker/dispatcher.py
tssweeney: fixed

weave/trace_server/interface/base_object_classes/actions.py
andrewtruong, line 1: feels weird that these are here.  Why can't they be with their respective actions?

weave/trace_server/interface/base_object_classes/base_object_registry.py
```diff
@@ -1,3 +1,4 @@
+from weave.trace_server.interface.base_object_classes.actions import ActionDefinition
 from weave.trace_server.interface.base_object_classes.base_object_def import BaseObject
```
andrewtruong, line 2: Is this concept new?  Why do we have and register base objects?

weave/trace_server/interface/base_object_classes/actions.py
tssweeney: This is actually really important - these are inside the `base_object_classes` directory which defines the object classes that are validated. These are like "virtual tables". See https://github.com/wandb/weave/pull/2826 for a discussion of this

weave/trace_server/interface/base_object_classes/base_object_registry.py
tssweeney: yes, just landed: https://github.com/wandb/weave/pull/2826

tests/trace/test_actions_lifecycle.py
andrewtruong: ok maybe this comment in the code could be helpful for others to understand the difference as well

weave/trace_server/actions_worker/actions/contains_words.py
andrewtruong: RE:2 -- How do you define "an action that hits (an) op"?
RE:3 -- Ok.  I was initially concerned because you have a strict type def below, but we have total control over this part of the code so we can just change it later

tssweeney: #2 -> this is not supported yet, but is in the plan. It will be a new action type ("op_execute") maybe with it's own parameter definition

#3 -> 👍

tests/trace/test_actions_lifecycle.py
tssweeney: ack - adding now

weave/trace_server/interface/base_object_classes/actions.py
andrewtruong: My intuition here would have been to something like:
```py
# base.py
R = TypeVar("R")

class BaseActionSpec(BaseModel, Generic[R]):
    action_type: str

    def execute(self, project_id, config, call, trace_server): ...


# contains_words.py
class ContainsWordsAction(BaseActionSpec[bool]):
    action_type: Literal["contains_words"] = "contains_words"
    target_words: ...

    def execute(...) -> bool: ...


# registry.py
ActionSpecType = ContainsWordsActionSpec | ...

dispatch_map = {...}
```


Now everything about each action is in 1 file, which roll into the registry.  

But it seems like `base_object_classes` just defines schemas and no impl?  I guess it makes sense depending on your perspective...

weave/trace_server/actions_worker/actions/contains_words.py
andrewtruong: oh I see an action that literally calls the op -- makes sense

weave/trace_server/interface/base_object_classes/actions.py
tssweeney: Oh i see what you are saying. Yeah, this is an explicit separation of concerns. Here we are strictly defining the interface (which can be useed in the client, in the server & is used to generate frontend specs). THink of this as an extension of trace_server_interface and importantly should not contain implementation

andrewtruong: ok, thanks for clarifying

andrewtruong (approved):

PR #2883, opened 2024-11-05
style(weave): style chat view a bit

jwlee64:
## Description

system prompt is clear with border
user is cacuts and right justified
assistant messages have callout with robot nad no background
function responses have a border on top to separate from actual function

Video shows before then after

https://github.com/user-attachments/assets/2a11e14e-9aa8-46ee-b72b-5f4726395da6

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanel.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanel.tsx
@@ -19,6 +19,8 @@ export const MessagePanel = ({
   const isUser = message.role === 'user';
   const isTool = message.role === 'tool';
   const isSystemPrompt = message.role === 'system';
+  const hasToolCalls = !!message.tool_calls && message.tool_calls.length > 0;
+  const hasContent = !!message.content && message.content.length > 0;
 
   return (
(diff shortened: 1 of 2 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanel.tsx
```diff
@@ -16,25 +17,62 @@ export const MessagePanel = ({
...
+          // Max width for non-system prompts
+          'max-w-3xl': !isSystemPrompt,
+          'w-3/4': isTool || isStructuredOutput || hasToolCalls,
+          // Justify the message to the right if it's a user message, add cactus background
+          'ml-auto bg-cactus-300/[0.24]': isUser,
+          // Justify the message to the left if it's not a user message
+          'mr-auto': !isUser,
+        })}>
+        {isSystemPrompt && (
+          // We only show the role for system prompts
```
jamie-rasmussen, line 51: I want to use these classes for Prompt objects too, where the role is important information to show

jamie-rasmussen (commented):
Could you share links to the calls you were including in the video? I think I like some of these changes and not others, wondering if maybe we should discuss briefly tomorrow in the sync with the design team.

PR #2887, opened 2024-11-05
chore(weave): ruff: enable UP007

andrewtruong:
Implements: https://docs.astral.sh/ruff/rules/non-pep604-annotation/

### DIFF
--- a/weave/scorers/llm_utils.py
+++ b/weave/scorers/llm_utils.py
@@ -83,7 +83,7 @@ def create(
 
 
 def embed(
-    client: _LLM_CLIENTS, model_id: str, texts: Union[str, list[str]], **kwargs: Any
+    client: _LLM_CLIENTS, model_id: str, texts: str | list[str], **kwargs: Any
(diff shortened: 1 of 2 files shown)


weave/scorers/llm_utils.py
```diff
@@ -83,7 +83,7 @@ def create(
 
 
 def embed(
-    client: _LLM_CLIENTS, model_id: str, texts: Union[str, list[str]], **kwargs: Any
+    client: _LLM_CLIENTS, model_id: str, texts: str | list[str], **kwargs: Any
```
tssweeney, line 86: seems like this replacement could be everywhere - why just this one spot?

tssweeney (approved):

weave/scorers/llm_utils.py
andrewtruong: It's because these annotations are opt-in until py310 (we still support py39)

It only kicks in for places where we explicitly do `from __future__ import annotations`

PR #2900, opened 2024-11-06
fix(weave): Add deepcopy methods for boxed types

andrewtruong:
## Description

Follow-up to:
https://github.com/wandb/weave/pull/2868

Adds `__deepcopy__` methods to boxed types which resolves issues when using `Model.predict` with `litellm`

## Testing

Unit tests and manual testing of the e2e LiteLLM evaluation example

### DIFF
--- a/weave/trace/box.py
+++ b/weave/trace/box.py
@@ -87,7 +87,8 @@ class BoxedNDArray(np.ndarray):
         if obj is None:
             return
 
-    def __deepcopy__(self, memo: dict) -> "BoxedNDArray":
+    def __deepcopy__(self, memo: dict[int, Any] | None = None) -> "BoxedNDArray":
+        memo = memo or {}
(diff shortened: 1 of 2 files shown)


weave/trace/box.py
```diff
@@ -19,16 +19,31 @@ class BoxedInt(int):
     _id: int | None = None
     ref: Ref | None = None
 
+    def __deepcopy__(self, memo: dict) -> "BoxedInt":
+        res = BoxedInt(self)
+        memo[id(self)] = res
```
tssweeney, line 24: where is `memo` defined? what is that?

andrewtruong: It's part of the python magic method definition:
https://docs.python.org/3/library/copy.html#object.__deepcopy__

adrnswanberg (approved):

PR #2903, opened 2024-11-06
chore(weave): ruff: enable TRY004

andrewtruong:
Implements: https://docs.astral.sh/ruff/rules/type-check-without-type-error/


gtarpenning (approved):
interesting, i like it

PR #2906, opened 2024-11-06
chore(weave): ruff: enable C409

andrewtruong:
Implements: https://docs.astral.sh/ruff/rules/unnecessary-literal-within-tuple-call/

### DIFF
--- a/tests/trace_server/test_validation.py
+++ b/tests/trace_server/test_validation.py
@@ -42,7 +42,7 @@ def test_validate_purge_req_one():
     assert str(e.value) == "Expected value of type <class 'tuple'>, got <class 'dict'>"
 
     with pytest.raises(InvalidRequest) as e:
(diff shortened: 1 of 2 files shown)


tests/trace_server/test_validation.py
```diff
@@ -42,7 +42,7 @@ def test_validate_purge_req_one():
     assert str(e.value) == "Expected value of type <class 'tuple'>, got <class 'dict'>"
 
     with pytest.raises(InvalidRequest) as e:
-        validation.validate_purge_req_one({"eq_": tuple([{"get_field_": "id"}])})
+        validation.validate_purge_req_one({"eq_": ({"get_field_": "id"},)})
```
jamie-rasmussen, line 45: This is a case where I feel like the other syntax is more readable, but not worth putting in an exception for it.

jamie-rasmussen (approved):

PR #2913, opened 2024-11-06
chore(weave): ruff: enable C405

andrewtruong:
Implements: https://docs.astral.sh/ruff/rules/unnecessary-literal-set/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -167,6 +167,7 @@ select = [
   "F401",   # https://docs.astral.sh/ruff/rules/unused-import/
   "TID252", # https://docs.astral.sh/ruff/rules/relative-imports/#relative-imports-tid252
   "UP",     # https://docs.astral.sh/ruff/rules/#pyupgrade-up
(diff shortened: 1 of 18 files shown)


pyproject.toml
```diff
@@ -174,6 +174,7 @@ select = [
   "C403",   # https://docs.astral.sh/ruff/rules/unnecessary-list-comprehension-set/
   "C416",   # https://docs.astral.sh/ruff/rules/unnecessary-comprehension/
   "C417",   # https://docs.astral.sh/ruff/rules/unnecessary-map/
+  "C405",   # https://docs.astral.sh/ruff/rules/unnecessary-literal-set/
```
jamie-rasmussen, line 177: Would be nice to keep the rules sorted at least within category - makes it more obvious where gaps are.

jamie-rasmussen (approved):

pyproject.toml
andrewtruong: Agreed.  I was a bit sloppy here because the intent is to fast-follow replace all of this group with a blanket `C` check

https://github.com/wandb/weave/pull/2914/files

PR #2915, opened 2024-11-06
fix(ui): refetch when changing filter state from frozen -> not frozen

gtarpenning:
## Description

[WB-21591](https://wandb.atlassian.net/browse/WB-21591)

This PR refetches the calls query when filters have changed. This normally happens automatically, except in the case when the `effectiveFilter` changes without any actual change to the backend query. This frequently occurs during interactions in the UI when navigating back to an already loaded table; the filter changes `frozen` status but none of the actual filter variables. 

## Testing

👀 

[WB-21591]: https://wandb.atlassian.net/browse/WB-21591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -279,6 +279,11 @@ export const CallsTable: FC<{
       setCallsResult([]);
       setCallsTotal(0);
       callsEffectiveFilter.current = effectiveFilter;
+      // Refetch the calls IFF the filter has changed, this is a
+      // noop if the calls query is already loading, but if the filter
+      // has no effective impact (frozen vs. not frozen) we need to
+      // manually refetch
+      calls.refetch();
     } else if (!calls.loading) {
       setCallsResult(calls.result);
       setCallsTotal(calls.total);


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -279,6 +279,11 @@ export const CallsTable: FC<{
       setCallsResult([]);
       setCallsTotal(0);
       callsEffectiveFilter.current = effectiveFilter;
+      // Refetch the calls IFF the filter has changed, this is a
+      // noop if the calls query is already loading, but if the filter
+      // has no effective impact (frozen vs. not frozen) we need to
+      // manually refetch
+      calls.refetch();
```
andrewtruong, line 286: When this runs, won't this cause `calls` to change and fire the `useEffect` again?

gtarpenning: I don't think so, because this useEffect should fire the next render after useCalls is called, which would make this `calls.refetch` a noop, as the calls query will **always** be loading at that point. However, even if this does cause the useEffect to fire again, the `callsEffectiveFilter` ref should now be changed, so either calls is `loading` in which case we do nothing, or it has returned, in which case we set the calls and load the table.

adrnswanberg (approved):

PR #2922, opened 2024-11-07
chore(weave): Class-based scorer runnable ref fix

tssweeney:
Correctly reports runnable refs when the scorer is an object.

Now that I am getting into the nitty-gritty of getting scorers to work, i realized we are not reporting the Evaluation scorer feedback ref correctly when the scorer is an object. Today, we are reporting the Scorer.score fn, not the parent Scorer object. This fixes and tests for that

### DIFF
--- a/tests/trace/test_evaluations.py
+++ b/tests/trace/test_evaluations.py
@@ -973,3 +973,37 @@ async def test_feedback_is_correctly_linked(client):
             id=list(score.calls())[0].id,
         ).uri()
     )
+
+
+@pytest.mark.asyncio
+async def test_feedback_is_correctly_linked_with_scorer_subclass(client):
+    @weave.op
+    def predict(text: str) -> str:
+        return text
+
+    class MyScorer(Scorer):
+        @weave.op()
+        def score(self, text, model_output) -> bool:
+            return text == model_output
+
+    scorer = MyScorer()
+    eval = weave.Evaluation(
+        dataset=[{"text": "hello"}],
+        scorers=[scorer],
+    )
+    res = await eval.evaluate(predict)
+    calls = client.server.calls_query(
+        tsi.CallsQueryReq(
+            project_id=client._project_id(),
+            include_feedback=True,
+            filter=tsi.CallsFilter(op_names=[get_ref(predict).uri()]),
+        )
+    )
+    assert len(calls.calls) == 1
+    assert calls.calls[0].summary["weave"]["feedback"]
+    feedbacks = calls.calls[0].summary["weave"]["feedback"]
+    assert len(feedbacks) == 1
+    feedback = feedbacks[0]
+    assert feedback["feedback_type"] == "wandb.runnable.MyScorer"
+    assert feedback["payload"] == {"output": True}
+    assert feedback["runnable_ref"] == get_ref(scorer).uri()
--- a/tests/trace/test_evaluations.py
+++ b/tests/trace/test_evaluations.py
@@ -982,9 +982,9 @@ async def test_feedback_is_correctly_linked_with_scorer_subclass(client):
         return text
 
     class MyScorer(Scorer):
-        @weave.op()
-        def score(self, text, model_output) -> bool:
-            return text == model_output
+        @weave.op
+        def score(self, text, output) -> bool:
+            return text == output
 
     scorer = MyScorer()
     eval = weave.Evaluation(
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -42,6 +42,7 @@ from weave.trace.table import Table
 from weave.trace.util import deprecated
 from weave.trace.vals import WeaveObject, WeaveTable, make_trace_obj
 from weave.trace_server.ids import generate_id
+from weave.trace_server.interface.feedback_types import RUNNABLE_FEEDBACK_TYPE_PREFIX
 from weave.trace_server.trace_server_interface import (
     CallEndReq,
     CallSchema,
@@ -1178,7 +1179,7 @@ class WeaveClient:
         freq = FeedbackCreateReq(
             project_id=self._project_id(),
             weave_ref=weave_ref_uri,
-            feedback_type="wandb.runnable." + runnable_ref.name,
+            feedback_type=RUNNABLE_FEEDBACK_TYPE_PREFIX + "." + runnable_ref.name,
             payload=payload,
             runnable_ref=runnable_ref_uri,
             call_ref=call_ref_uri,
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -332,11 +332,11 @@ class Call:
         score_call_ref = get_ref(score_call)
         if score_call_ref is None:
             raise ValueError("Score call has no ref")
-        client._add_score(
-            call_ref_uri=self_ref.uri(),
+        client._add_runnable_feedback(
+            weave_ref_uri=self_ref.uri(),
             runnable_ref_name=score_name,
-            score_results=score_results,
-            scorer_call_ref_uri=score_call_ref.uri(),
+            output=score_results,
+            call_ref_uri=score_call_ref.uri(),
             runnable_ref_uri=scorer_op_ref.uri(),
         )
 
@@ -1106,7 +1106,10 @@ class WeaveClient:
 
     @trace_sentry.global_trace_sentry.watch()
     def _send_score_call(
-        self, predict_call: Call, score_call: Call, scorer_ref_uri: Optional[str] = None
+        self,
+        predict_call: Call,
+        score_call: Call,
+        scorer_object_ref_uri: Optional[str] = None,
     ) -> Future[str]:
         """(Private) Adds a score to a call. This is particularly useful
         for adding evaluation metrics to a call.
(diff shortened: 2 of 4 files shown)


tests/trace/test_evaluations.py
```diff
@@ -973,3 +973,37 @@ def score(text, model_output) -> bool:
...
+
+@pytest.mark.asyncio
+async def test_feedback_is_correctly_linked_with_scorer_subclass(client):
+    @weave.op
+    def predict(text: str) -> str:
+        return text
+
+    class MyScorer(Scorer):
+        @weave.op()
+        def score(self, text, model_output) -> bool:
```
andrewtruong, line 986: use new `output` notation

tests/trace/test_evaluations.py
```diff
@@ -973,3 +973,37 @@ def score(text, model_output) -> bool:
...
+
+
+@pytest.mark.asyncio
+async def test_feedback_is_correctly_linked_with_scorer_subclass(client):
+    @weave.op
+    def predict(text: str) -> str:
+        return text
+
+    class MyScorer(Scorer):
+        @weave.op()
```
andrewtruong, line 985: nit: no parens

tests/trace/test_evaluations.py
```diff
@@ -973,3 +973,37 @@ def score(text, model_output) -> bool:
...
+        def score(self, text, model_output) -> bool:
+            return text == model_output
+
+    scorer = MyScorer()
+    eval = weave.Evaluation(
+        dataset=[{"text": "hello"}],
+        scorers=[scorer],
+    )
+    res = await eval.evaluate(predict)
+    calls = client.server.calls_query(
```
andrewtruong, line 995: Do you need to go down to this level?  The test might be more appropriate with `client.calls()`

weave/trace/weave_client.py
```diff
@@ -1181,11 +1177,11 @@ def _add_score(
 
         freq = FeedbackCreateReq(
             project_id=self._project_id(),
-            weave_ref=call_ref_uri,
-            feedback_type="wandb.runnable." + score_name,
+            weave_ref=weave_ref_uri,
+            feedback_type="wandb.runnable." + runnable_ref.name,
```
andrewtruong, line 1181: may want to pull out this `wandb.runnable.` magic val as a const

weave/trace/weave_client.py
```diff
@@ -1114,37 +1118,35 @@ def send_score_call() -> str:
...
+                output=score_results,
+                call_ref_uri=scorer_call_ref_uri,
+                runnable_ref_uri=runnable_ref_uri,
             )
 
         return self.future_executor.defer(send_score_call)
 
     @trace_sentry.global_trace_sentry.watch()
-    def _add_score(
+    def _add_runnable_feedback(
```
andrewtruong, line 1143: what is a `runnable`?  Is that defined somewhere?

weave/trace/weave_client.py
```diff
@@ -1114,37 +1118,35 @@ def send_score_call() -> str:
...
         self,
         *,
+        weave_ref_uri: str,
```
andrewtruong, line 1146: we are starting to different URIs in a lot of places.  Maybe worth type narrowing to avoid errors

weave/trace/weave_client.py
```diff
@@ -1114,37 +1118,35 @@ def send_score_call() -> str:
...
-            scorer_op_ref_uri = score_call.op_name
-            scorer_op_ref = parse_uri(scorer_op_ref_uri)
-            if not isinstance(scorer_op_ref, OpRef):
-                raise TypeError(f"Invalid scorer op ref: {scorer_op_ref_uri}")
-            score_name = scorer_op_ref.name
+
+            # If scorer_object_ref_uri is provided, it is used as the runnable_ref_uri
+            # Otherwise, we use the op_name from the score_call. This should happen
+            # when there is a Scorer subclass that is the source of the score call.
+            runnable_ref_uri = scorer_object_ref_uri or score_call.op_name
```
andrewtruong, line 1130: our use of `op_name` here is very confusing.  IIRC this is actually a uri, but we don't call it that for some reason?

tests/trace/test_evaluations.py
```diff
@@ -973,3 +973,37 @@ def score(text, model_output) -> bool:
...
+        )
+    )
+    assert len(calls.calls) == 1
+    assert calls.calls[0].summary["weave"]["feedback"]
+    feedbacks = calls.calls[0].summary["weave"]["feedback"]
+    assert len(feedbacks) == 1
+    feedback = feedbacks[0]
+    assert feedback["feedback_type"] == "wandb.runnable.MyScorer"
+    assert feedback["payload"] == {"output": True}
+    assert feedback["runnable_ref"] == get_ref(scorer).uri()
```
andrewtruong, line 1009: shouldn't this be `runnable_uri`?

weave/trace/weave_client.py
```diff
@@ -1181,11 +1177,11 @@ def _add_score(
...
             project_id=self._project_id(),
-            weave_ref=call_ref_uri,
-            feedback_type="wandb.runnable." + score_name,
+            weave_ref=weave_ref_uri,
+            feedback_type="wandb.runnable." + runnable_ref.name,
             payload=payload,
-            runnable_ref=scorer_op_ref_uri,
-            call_ref=scorer_call_ref_uri,
+            runnable_ref=runnable_ref_uri,
+            call_ref=call_ref_uri,
```
andrewtruong, line 1184: these fields represent URIs so we should call them uris

tests/trace/test_evaluations.py
tssweeney: I really don't like the calls API - especially how it gets feedback. This is much better for asserting the data I am interested in asserting

tests/trace/test_evaluations.py
tssweeney: ack

tests/trace/test_evaluations.py
tssweeney: ack

tests/trace/test_evaluations.py
tssweeney: the feedback spec is `runnable_ref`

weave/trace/weave_client.py
tssweeney: that is exactly correct - op_name is actually the source op

weave/trace/weave_client.py
tssweeney: runnable is a new class of feedback. see https://github.com/wandb/weave/pull/2865

weave/trace/weave_client.py
tssweeney: what do you mean?

weave/trace/weave_client.py
tssweeney: Done

weave/trace/weave_client.py
tssweeney: changing this is outside the scope of this PR - it is the interface layer

tssweeney: are you referring to the variable or the req field?

tests/trace/test_evaluations.py
andrewtruong: My preference is to fix the calls API.  It feels wrong to reach into the server here.  What's wrong with how it gets feedback?

tssweeney: I do this in countless tests right now. all i want is to assert the raw data, not worry aobut the higher level calls iterator or opaque feedback iterator stuff

tssweeney: Not friendly: 

```
    @property
    def feedback(self) -> RefFeedbackQuery:
        if not self.id:
            raise ValueError("Can't get feedback for call without ID")
        if self._feedback is None:
            project_parts = self.project_id.split("/")
            if len(project_parts) != 2:
                raise ValueError(f"Invalid project_id: {self.project_id}")
            entity, project = project_parts
            weave_ref = CallRef(entity, project, self.id)
            self._feedback = RefFeedbackQuery(weave_ref.uri())
        return self._feedback

```

weave/trace/weave_client.py
andrewtruong: Is there a difference in shape between `weave_ref_uri` and `call_ref_uri`?  If we have many different types of URI floating around, can we help ourselves use the correct type of URI?

weave/trace/weave_client.py
andrewtruong: Can we rename it to be less confusing?

tests/trace/test_evaluations.py
andrewtruong: but that's confusing because this is a URI

tests/trace/test_evaluations.py
tssweeney: fwiw, i try to avoid using the WeaveClient directly unless i am explicitly trying to test functionality of the WeaveClient. I find it much more cumbersome

weave/trace/weave_client.py
andrewtruong: The field -- it's called `ref` but it's being assigned a `uri`.  I think it actually represents a URI?

weave/trace/weave_client.py
tssweeney: Hmmm, let's make sure we are on the same page:
  * The Feedback table now has a few key column (they all have the `_ref` suffix and are strings formatted like refs:
    * `weave_ref`
    * `call_ref`
    * `trigger_ref`
    * `runnable_ref`
    * `annotation_ref`
  * At runtime, I like to distinguish between runtime Refs (instances of Refs) and the string uri by using `_ref_uri` to make it super clear this is a string, not the runtime object.

Given that context, i am having a hard time groking your recommendation or callout. Can you elaborate?

weave/trace/weave_client.py
tssweeney: yes, but that follows the existing pattern of the feedback table and the interface. Changing this would impact quite a few layers and break pattern

tests/trace/test_evaluations.py
andrewtruong: I won't block on this, but I think reaching into the server is not ideal.  All of this code is at the "client level", so the checks should also be at the "client level".  If the client sucks we should fix it

weave/trace/weave_client.py
andrewtruong: Maybe I'm saying something silly or overkill

Here's what I observe:
1. We are using URIs in a lot of places in the SDK
2. They can have different shapes like [(1)](https://github.com/wandb/weave/blame/tim/basic_batch_actions/weave/trace/refs.py#L75), [(2)](https://github.com/wandb/weave/blame/tim/basic_batch_actions/weave/trace/refs.py#L148), [(3)](https://github.com/wandb/weave/blame/tim/basic_batch_actions/weave/trace/refs.py#L210), and a specific shape of URI maps to a specific type of Ref in our SDK.
3. Should we make a subtype of str for the URIs to avoid passing the wrong URI?

For example:
```py
T = TypeVar('T', bound='WeaveURI')

class WeaveCallURI(str):
    def __new__(cls: Type[T], content: str) -> T:
        # validate it conforms to f"weave:///{self.entity}/{self.project}/call/{self.id}"
       return cast(T, super().__new__(cls, content))
```

weave/trace/weave_client.py
andrewtruong: ok, unfortunate but I wont block here.

There are a few places where the nomenclature is kinda weird and I always get a little confused when I revisit them

andrewtruong (approved):
stamping as there are no blocking comments

PR #2924, opened 2024-11-07
chore(weave): Move deepcopy methods to Ref

andrewtruong:


weave/trace/refs.py
```diff
@@ -15,16 +15,33 @@ class WeaveDigestError(ValueError):
...
 
+    def __deepcopy__(self, memo: dict) -> "Ref":
+        d = {}
+        for f in fields(self):
+            v = getattr(self, f.name)
 
-@dataclasses.dataclass(frozen=True)
+            # Must resolve the future to remove the lock which causes issues when deepcopying
+            if isinstance(v, Future):
+                v = v.result()
```
tssweeney, line 33: you could just copy the future, no? Else this internal deep copy with all of a sudden block the user process

andrewtruong: yeah good point -- I thought you had to resolve, but shallow copy is simpler and better

tssweeney (commented):
🔥

tssweeney (approved):

PR #2929, opened 2024-11-07
chore(weave): Fix typo from revert

tssweeney:
Fixes typo from earlier revert


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1364,7 +1364,7 @@ def feedback_create(self, req: tsi.FeedbackCreateReq) -> tsi.FeedbackCreateRes:
             "wb_user_id": req.wb_user_id,
             "creator": req.creator,
             "feedback_type": req.feedback_type,
-            "payload": req.payload,
+            "payload": payload,
```
tssweeney, line 1367: this should be the dumped payload, not dictionary

PR #2934, opened 2024-11-07
chore(weave): adding onMouseEnter/ onMouseLeave handlers to button tooltip

mbarrramsey:
## Description

See https://github.com/wandb/core/pull/25285/files#r1833424789

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/Button/Button.tsx
+++ b/weave-js/src/components/Button/Button.tsx
@@ -156,14 +156,26 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
     );
 
     const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+    const handleMouseEnter = () => {
+      setIsTooltipOpen(true);
+    };
+
+    const handleMouseLeave = () => {
+      setIsTooltipOpen(false);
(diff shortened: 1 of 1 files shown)


weave-js/src/components/Button/Button.tsx
```diff
@@ -156,14 +156,26 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
       return (
         <Tailwind style={wrapperStyles}>
           <Tooltip.Provider>
             <Tooltip.Root open={isTooltipOpen} onOpenChange={setIsTooltipOpen}>
               <Tooltip.Trigger asChild>
-                {/* span is needed so tooltip works on disabled buttons */}
-                <span className="[display:inherit]">{button}</span>
+                <span
+                  onMouseEnter={handleMouseEnter}
+                  onMouseLeave={handleMouseLeave}>
```
connieelee, line 175: hmm this change doesn't seem right. there's probably something wrong with the call site where it's not working.

mbarrramsey: Sounds good -- investigating more

PR #2938, opened 2024-11-08
chore(weave): Rename ActionDefintion to ActionSpec to match naming patterns

tssweeney:
This is our only chance to change this (before it ships). Let's get it right

### DIFF
--- a/weave/trace_server/interface/base_object_classes/actions.py
+++ b/weave/trace_server/interface/base_object_classes/actions.py
@@ -5,7 +5,7 @@ from pydantic import BaseModel, Field, field_validator
 from weave.trace_server.interface.base_object_classes import base_object_def
 
 
-class LlmJudgeActionSpec(BaseModel):
+class LlmJudgeActionConfig(BaseModel):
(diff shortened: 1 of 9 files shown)


weave/trace_server/interface/base_object_classes/actions.py
```diff
@@ -15,19 +15,19 @@ class LlmJudgeActionSpec(BaseModel):
...
+ActionConfigType = Union[
+    LlmJudgeActionConfig,
+    ContainsWordsActionConfig,
 ]
 
 
-class ActionDefinition(base_object_def.BaseObject):
-    spec: ActionSpecType = Field(..., discriminator="action_type")
+class ActionSpec(base_object_def.BaseObject):
+    config: ActionConfigType = Field(..., discriminator="action_type")
```
tssweeney, line 30: Changed the nested property to `config` so that we don't have a nested "spec". This also matches the parameters in the dispatcher

gtarpenning (approved):
I think you have to re-generate too

PR #2942, opened 2024-11-08
feat(ui): object comparison

jamie-rasmussen:
## Description

Adds initial implementation of object/call comparison view.

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePage.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePage.tsx
@@ -1,0 +1,70 @@
+/**
+ * Dispatch to call or object specific comparison based on query parameters.
+ */
+
+import React from 'react';
+import {useHistory} from 'react-router-dom';
+
+import {Empty} from '../pages/common/Empty';
+import {
+  getParamArray,
+  queryGetBoolean,
+  queryGetDict,
+  queryGetString,
+} from '../urlQueryUtil';
+import {ComparePageCalls} from './ComparePageCalls';
+import {ComparePageObjects} from './ComparePageObjects';
+
+type ComparePageProps = {
+  entity: string;
+  project: string;
+};
+
+export const ComparePage = ({entity, project}: ComparePageProps) => {
+  const history = useHistory();
+  const d = queryGetDict(history);
+  const mode =
+    queryGetString(history, 'mode') === 'unified' ? 'unified' : 'parallel';
+  const baselineEnabled = queryGetBoolean(history, 'baseline', false);
+  const onlyChanged = queryGetBoolean(history, 'changed', false);
+  const callIds = getParamArray(d, 'call');
+
+  if (callIds.length) {
+    return (
+      <ComparePageCalls
+        entity={entity}
+        project={project}
+        callIds={callIds}
+        mode={mode}
+        baselineEnabled={baselineEnabled}
+        onlyChanged={onlyChanged}
+      />
+    );
+  }
+
+  const objs = getParamArray(d, 'obj');
+  if (objs.length) {
+    return (
+      <ComparePageObjects
+        entity={entity}
+        project={project}
+        objectIds={objs}
+        mode={mode}
+        baselineEnabled={baselineEnabled}
+        onlyChanged={onlyChanged}
+      />
+    );
+  }
+
+  // TODO: Link to docs if we make them
+  // Currently nothing links to this state but query parameters
+  // are sometimes dropped during dev reloading.
+  return (
+    <Empty
+      icon="baseline-alt"
+      heading="Compare objects or calls"
+      description="This page allows you to compare objects or calls."
+      moreInformation=""
+    />
+  );
+};
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePageObjectsLoaded.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePageObjectsLoaded.tsx
@@ -1,0 +1,399 @@
+/**
+ * This is the main layout of the compare page.
+ * It also handles the logic around ref expansion.
+ */
+import {GridRowId} from '@mui/x-data-grid-pro';
+import {Switch} from '@wandb/weave/components';
+import * as DropdownMenu from '@wandb/weave/components/DropdownMenu';
+import _ from 'lodash';
+import React, {useCallback, useEffect, useMemo, useState} from 'react';
+import {useHistory} from 'react-router-dom';
+
+import {Button} from '../../../../Button';
+import {Icon} from '../../../../Icon';
+import {TailwindContents} from '../../../../Tailwind';
+import {isWeaveRef} from '../filters/common';
+import {mapObject, TraverseContext} from '../pages/CallPage/traverse';
+import {SimplePageLayout} from '../pages/common/SimplePageLayout';
+import {useWFHooks} from '../pages/wfReactInterface/context';
+import {
+  queryGetString,
+  querySetArray,
+  querySetString,
+  queryToggleBoolean,
+} from '../urlQueryUtil';
+import {computeDiff, mergeObjects} from './compare';
+import {CompareGrid} from './CompareGrid';
+import {isSequentialVersions, parseSpecifier} from './hooks';
+import {getExpandableRefs, RefValues, RESOLVED_REF_KEY} from './refUtil';
+import {ShoppingCart} from './ShoppingCart';
+import {ComparableObject, Mode} from './types';
+
+type ComparePageObjectsLoadedProps = {
+  objectType: 'object' | 'call';
+  objectIds: string[];
+  mode: Mode;
+  baselineEnabled: boolean;
+  onlyChanged: boolean;
+  objects: ComparableObject[];
+  lastVersionIndices?: Record<string, number>;
+};
+
+// TODO: These are always going to be different so not useful to flag as such.
+//       But maybe seeing value for things like versionHash is still useful?
+const UNINTERESTING_PATHS_CALL = ['id'];
+const UNINTERESTING_PATHS_OBJ = [
+  'baseObjectClass',
+  'entity',
+  'path',
+  'objectId',
+  'project',
+  'scheme',
+  'versionHash',
+  'versionIndex',
+  'weaveKind',
+];
+
+const IconPlaceholder = () => <div className="h-18 w-18" />;
+
+export const ComparePageObjectsLoaded = ({
+  objectType,
+  objectIds,
+  mode,
+  baselineEnabled,
+  onlyChanged,
+  objects,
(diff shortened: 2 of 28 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
```diff
@@ -449,6 +463,26 @@ export const browse3ContextGen = (
...
+      entityName: string,
+      projectName: string,
+      callIds: string[]
+    ) => {
+      const params = callIds
+        .map(id => 'call=' + encodeURIComponent(id))
+        .join('&');
+      return `${projectRoot(entityName, projectName)}/compare?${params}`;
+    },
+    compareObjectsUri: (
```
tssweeney, line 476: Are Ops supported as objects?

weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
```diff
@@ -449,6 +463,26 @@ export const browse3ContextGen = (
         leaderboardName ? `/${leaderboardName}` : ''
       }${edit ? '?edit=true' : ''}`;
     },
+    compareCallsUri: (
+      entityName: string,
+      projectName: string,
+      callIds: string[]
+    ) => {
+      const params = callIds
+        .map(id => 'call=' + encodeURIComponent(id))
```
tssweeney, line 472: Curious why you went with IDs over refs?

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/refUtil.ts
```diff
@@ -0,0 +1,28 @@
+import {isWeaveRef} from '../filters/common';
+import {traverse, TraverseContext} from '../pages/CallPage/traverse';
+import {isExpandableRef} from '../pages/wfReactInterface/tsDataModelHooksCallRefExpansion';
+import {ComparableObjects} from './types';
+
+// When we replace a ref with its resolved data we insert this key
+// with the ref URI to keep track of the original value.
+export const RESOLVED_REF_KEY = '_ref';
```
tssweeney, line 8: I think we do something similar for call expansion - maybe an opportunity for some merging.

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePage.tsx
```diff
@@ -0,0 +1,70 @@
+/**
```
tssweeney, line 1: any reason you put this (i suppose the entire compare dir) outside of the pages dir?

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/hooks.ts
```diff
@@ -0,0 +1,121 @@
...
+  objectVersions: ObjectVersionSchema[];
+  lastVersionIndices: Record<string, number>; // Name to number
+};
+
+export const useObjectVersions = (
+  entity: string,
+  project: string,
+  objectVersionSpecifiers: string[]
+): ObjectVersionsResult => {
+  // TODO: Need to introduce a backend query (/objs/read?) to bulk get specific versions of objects
```
tssweeney, line 87: UGGG lets add that! You are completely right!

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ShoppingCart.tsx
```diff
@@ -0,0 +1,150 @@
...
+    oldIndex,
+    newIndex,
+  }: {
+    oldIndex: number;
+    newIndex: number;
+  }) => {
+    if (oldIndex === newIndex) {
+      return;
+    }
+    const {search} = history.location;
```
tssweeney, line 84: I noticed this elsewhere, but figured I would add one comment: I generally stay way from embedding URL management deep in components like this. Ideally the URL management would happen in a very high level top component (Browse3) and everything below is driven by params and callbacks. The reason for this is that if we ever want to export to reports or embed in other contexts where the state is no longer URL managed, then we incur a pretty hefty refactor.

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePageObjectsLoaded.tsx
```diff
@@ -0,0 +1,399 @@
...
+                      Unified
+                    </Button>
+                  </div>
+                )}
+                <div className="flex-grow" />
+                <DropdownMenu.Root open={isOpen} onOpenChange={setIsOpen}>
+                  <DropdownMenu.Trigger>
+                    <div className="cursor-pointer">
+                      <div className="flex items-center gap-10">
+                        {baselineEnabled
```
tssweeney, line 346: This toggle does not seem to do anything when there are only 2 things to compare

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePageObjectsLoaded.tsx
```diff
@@ -0,0 +1,399 @@
...
+          content: (
+            <TailwindContents>
+              <div className="flex items-center gap-4 px-16 pt-12">
+                {isSequential && (
+                  <Button
+                    size="small"
+                    disabled={first.version === 0}
+                    icon="back"
+                    variant="ghost"
+                    onClick={onPrev}
```
tssweeney, line 281: as a user, I am not sure what this does (especially when there are only 2 items?) unclear why these arrows are here, but not in the calls one (without reading the code)

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePage.tsx
```diff
@@ -0,0 +1,70 @@
...
+export const ComparePage = ({entity, project}: ComparePageProps) => {
+  const history = useHistory();
+  const d = queryGetDict(history);
+  const mode =
+    queryGetString(history, 'mode') === 'unified' ? 'unified' : 'parallel';
+  const baselineEnabled = queryGetBoolean(history, 'baseline', false);
+  const onlyChanged = queryGetBoolean(history, 'changed', false);
+  const callIds = getParamArray(d, 'call');
+
+  if (callIds.length) {
```
tssweeney, line 32: I feel like we might want to limit the number to a reasonable set: 
Screenshot 2024-11-19 at 15 23 18

tssweeney: this pretty much halts to page

weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
tssweeney: I see by playing with it that they are not - seem like we could support that in the future

weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
tssweeney: seems like you could do `ref=...` and the ref itself encodes that it is an object or a call. which would unify the model a bit.

It seems a little odd to me that the search params are mutually exclusive

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePage.tsx
jamie-rasmussen: At some point I'd like to do a more major reorganization of our UI directory structure. E.g. `PagePanelComponents` seems unnecessarily long when everything is already under `components`, it seems weird to have `Home` in the path to these, "Browse3" isn't a good name, etc.

"pages" seems redundant to me given all of this is under "PagePanelComponents" and I generally would rather have things organized/named by feature. But I can move it if you have a strong preference.

tssweeney: I agree with the need to refactor. In my mind `weave-js/src/components/PagePanelComponents/Home/Browse3` is the root of our weave application and the initial refactor should be to just hoist that to a top level dir. Everything outside of that `Browse3` that is not common W&B should be either removed or extracted. Given that framing, the `pages` dir serves as the "topmost" dir for our pages.

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePageObjectsLoaded.tsx
jamie-rasmussen: I will hide it when there are only two things.

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/refUtil.ts
jamie-rasmussen: Yeah, I think there's a lot of opportunity for merging functionality, though with all of the problems we have with things like import cycle I'm a little wary of introducing dependencies across different parts of the code and willing to live with more duplication than I'd prefer.

weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
jamie-rasmussen: Using refs would make the URLs substantially longer and uglier 

e.g. comparing three calls here:
`http://localhost:9000/wandb-designers/signal-maven/weave/compare?call=491f11f4-cccf-4f0b-b447-897254fa1534&call=d1e494c4-737f-49a8-be89-7703fec23b7b&call=2b58c817-607b-48ed-b4ce-4ec4fb816e95`
v.s. with refs it would be:
`http://localhost:9000/wandb-designers/signal-maven/weave/compare?ref=weave%3A%2F%2F%2Fwandb-designers%2Fsignal-maven%2Fcall%2F491f11f4-cccf-4f0b-b447-897254fa1534&ref=weave%3A%2F%2F%2Fwandb-designers%2Fsignal-maven%2Fcall%2Fd1e494c4-737f-49a8-be89-7703fec23b7b&ref=weave%3A%2F%2F%2Fwandb-designers%2Fsignal-maven%2Fcall%2F2b58c817-607b-48ed-b4ce-4ec4fb816e95`

weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx
jamie-rasmussen: Yes, that would be a good future enhancement to make things more consistent. Comparing calls was already scope creep I couldn't resist. :-)

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ComparePage.tsx
jamie-rasmussen: I'm going to limit it to 6 object columns in the grid and show a message. I think that will work for most cases while keeping the entire set in the shopping cart so you can select the items for a pairwise comparison.

tssweeney (approved):

PR #2955, opened 2024-11-12
chore(weave): update tooltip depending on if we can expand all

jwlee64:
## Description

when small
Screenshot 2024-11-12 at 12 29 38 PM

when expand is too big
Screenshot 2024-11-12 at 12 29 28 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
@@ -153,12 +153,15 @@ const ObjectViewerSectionNonEmpty = ({
     setMode('collapsed');
     setExpandedIds([]);
   };
+
+  const isExpandAllLarge = getGroupIds().length > EXPANDED_IDS_LENGTH;
+
   const onClickExpanded = () => {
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
```diff
@@ -153,17 +153,22 @@ const ObjectViewerSectionNonEmpty = ({
     setMode('collapsed');
     setExpandedIds([]);
   };
+
+  const isExpandAllSmall =
+    !!apiRef?.current?.getAllRowIds &&
+    getGroupIds().length < EXPANDED_IDS_LENGTH;
```
jamie-rasmussen, line 159: Would it make sense for this to be something like `getGroupIds().length - expandedIds.length < EXPANDED_IDS_LENGTH`? i.e. if you've expanded 990 / 1000 already the message would be "Expand all" instead of "Expand next 200 rows"

jwlee64: yah that makes sense to me

jamie-rasmussen (approved):

PR #2956, opened 2024-11-12
chore(weave): using chevron icon for runSelector dropdowns

mbarrramsey:
## Description

https://github.com/user-attachments/assets/51828e9c-5988-4d4f-b855-a05836f53445

Adding chevron up and down icon for new runs-selector-modified-dropdown class added in follow-up frontend PR: https://github.com/wandb/core/pull/25533


weave-js/src/common/components/elements/ModifiedDropdown.tsx
```diff
@@ -470,6 +471,15 @@ const ModifiedDropdown: FC<ModifiedDropdownProps> = React.memo(
           }
         }}
         trigger={renderTrigger()}
+        icon={
+          passProps.className === 'runs-selector-modified-dropdown' ? (
```
jamie-rasmussen, line 475: `runs-selector` seems like logic that shouldn't be in this layer. Maybe add a `useIcon` prop?

mbarrramsey: Sounds good -- added a `useIcon` prop

jamie-rasmussen (approved):

PR #2983, opened 2024-11-14
chore(weave): Make Call.children a property

andrewtruong:
This makes it so you can get the children of a call without the parens.  This is safe because the result is already a lazy iterator, and it's more idiomatic because `children` are a property of the call.

So instead of:
```py
for c in call.children(): ...
```

You can now do:
```py
for c in call.children: ...
```


andrewtruong:
Updated comments

jamie-rasmussen:
Quite possibly no customers are using `children()` in their own code but if they are, upgrading would break them, right? Not a reason not to do this if we think it is better but maybe worth noting in ChangeLog or something?

PR #2990, opened 2024-11-14
Update Quickstart to use gpt-4o-mini

johndmulhausen:
When you create a new API key with OpenAI it does not work with GPT 3.5 anymore. The only way I could get the quickstart to run with a new API key was to replace the model with `gpt-4o-mini` in every instance, as done in this PR.

## Testing

Try this by changing the model names by hand at https://colab.research.google.com/github/wandb/weave/blob/master/docs/intro_notebook.ipynb


johndmulhausen:

shawnlewis (approved):

PR #2995, opened 2024-11-15
feat(ui): create ToggleButtonGroup component

molliean:
## Description

Creates `ToggleButtonGroup` common component.

https://www.figma.com/design/01KWBdMZg5QM9SRS1pQq0z/Design-System----Robot-Styles?m=auto&node-id=12323-59672&t=GB4Y5hz8mGnYWOfC-1

Allows users to select a single option from a set of toggle buttons. It supports three sizes (`small`, `medium`, and `large`), optional icons, and disabled states. It must have at least two options.

`ToggleButtonGroup` is built on Radix UI's [Toggle Group](https://www.radix-ui.com/primitives/docs/components/toggle-group).

## Testing

Tested in Storybook (will be in separate `core` PR):

https://github.com/user-attachments/assets/e0604d6b-f108-49f2-9a17-60a80f5a1aee

How was this PR tested?


molliean:

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+import React from 'react';
+import {twMerge} from 'tailwind-merge';
+
+import {Button} from '../Button';
+import {IconName} from '../Icon';
+import {Tailwind} from '../Tailwind';
+import {ToggleButtonGroupSize} from './types';
+
+export type ToggleButtonGroupProps = {
+  options: string[];
```
ericakdiaz, line 11: could you merge this and icons into a single type?
`type Option = {title: string, icon?: IconName}`
`options: Option[];`

then you don't have to deal with weirdness of matching up the indices

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+  value: string;
+  size: ToggleButtonGroupSize;
+  isDisabled?: boolean;
+  icons?: Array<IconName | undefined>;
```
adityachoudhari26, line 15: maybe we should just create a new type `ToggleOption`
which includes the option label and the option icon. that way we don't have to do these two separate arrays and need to use the index to match them. instead if we create a type then the label + option become more tightly coupled

weave-js/src/components/ToggleButtonGroup/types.tsx
```diff
@@ -0,0 +1,7 @@
+export const ToggleButtonGroupSizes = {
+  Small: 'small',
+  Medium: 'medium',
+  Large: 'large',
+} as const;
+export type ToggleButtonGroupSize =
+  (typeof ToggleButtonGroupSizes)[keyof typeof ToggleButtonGroupSizes];
```
ericakdiaz, line 7: nit - would an enum work better here?

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
adityachoudhari26:
```
type ToggleOption = {
  label: string;
  icon?: IconName
}

export type ToggleButtonGroupProps = {
  ...
  options: ToggleOption[] // label and icon are tightly coupled
}

File: weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx [hunk 4]
```diff
@@ -0,0 +1,65 @@
...
+  onValueChange: (value: string) => void;
+};
+
+export const ToggleButtonGroup = React.forwardRef<
+  HTMLDivElement,
+  ToggleButtonGroupProps
+>(({options, value, size, isDisabled = false, icons, onValueChange}, ref) => {
+  if (options.length < 2) {
```
adityachoudhari26, line 23: should we even render the component if its in this invalid state? maybe we should just return null, and then add a jsdocs comment that this component should only be rendered if `options.length >= 2`

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+  return (
+    <Tailwind>
+      <ToggleGroup.Root
+        type="single" // supports single selection only
+        value={value}
+        onValueChange={handleValueChange}
+        className="flex gap-px"
+        ref={ref}>
+        {options.map((option, index) => (
```
adityachoudhari26, line 40:
```suggestion
        {options.map(({ label, icon}) => (
```

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
adityachoudhari26: i'm not sure about the specific ask for this component though so maybe we do nothing, but generally we should avoid console.error as it would show up on production saas (if some customer inspected the logs they would see this error)

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <ToggleGroup.Item key={option} value={option} disabled={isDisabled}>
+            <Button
+              icon={icons?.[index]}
```
adityachoudhari26, line 43: suggestion - maybe we should also outright disable the option for the one that is already selected. keep the check in your handler (newValue !== value) but add another layer of safety by not allowing the user to even click the option that is already toggled)

adityachoudhari26 (approved):

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
molliean: good idea, i'll update!

weave-js/src/components/ToggleButtonGroup/types.tsx
```diff
@@ -0,0 +1,6 @@
+export enum ToggleButtonGroupSizes {
+  small = 'small',
+  medium = 'medium',
+  large = 'large',
+}
+export type ToggleButtonGroupSize = keyof typeof ToggleButtonGroupSizes;
```
ericakdiaz, line 6: do you need this type? can you just do `size: ToggleButtonGroupSizes` in `ToggleButtonGroup.tsx`

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,75 @@
...
+            key={label}
+            value={label}
+            disabled={isDisabled || value === label}>
+            <Button
+              icon={icon}
+              size={size}
+              className={twMerge(
+                size === 'small' && 'gap-2 px-6 py-3 text-sm',
+                size === 'medium' && 'gap-3 px-10 py-4 text-base',
+                size === 'large' && 'gap-2 px-12 py-8 text-base',
```
ericakdiaz, line 58: you can compare directly to the enum like `ToggleButtonGroupSizes.small`

weave-js/src/components/ToggleButtonGroup/types.tsx
```diff
@@ -0,0 +1,5 @@
+export enum ToggleButtonGroupSizes {
+  small = 'small',
+  medium = 'medium',
+  large = 'large',
```
ericakdiaz, line 4: super nit - i think this is the convention
```
export enum ToggleButtonGroupSizes {
  Small,
  Medium,
  Large,
}
```
I don't think you have a need for them to be backed by strings here

ericakdiaz (approved):
looks good, just one small nit left :)

weave-js/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * ToggleButtonGroup component should only be rendered if options.length >= 2.
+ */
+export const ToggleButtonGroup = React.forwardRef<
+  HTMLDivElement,
+  ToggleButtonGroupProps
+>(({options, value, size, isDisabled = false, onValueChange}, ref) => {
+  if (options.length < 2) {
+    return null; // Do not render if there are fewer than two options
+  }
+
```
jamie-rasmussen, line 33: Could consider logging a warning if the provided value is not one of the options.

jamie-rasmussen (approved):

PR #2997, opened 2024-11-15
chore(weave): Move to future annotations for weave_client

andrewtruong:
Move to modern python syntax:
https://peps.python.org/pep-0604/

This excludes files that define classes that derive from `BaseModel` because it can cause strange behaviour in py39.  Once we move to py310+, we can update that code


gtarpenning (approved):
I like the `|` a lot... what is `# noqa: UP007`?

PR #2999, opened 2024-11-15
chore(weave): Retry HTTP500 errors

andrewtruong:
HTTP500 usually means ephemeral errors and are retry-able

Separately, we should work to improve the errors that the server sends back.


tests/trace_server/test_remote_http_trace_server.py
```diff
@@ -42,56 +42,49 @@ def test_ok(self, mock_post):
...
 
         resp4 = requests.Response()
-        resp4.status_code = 429
+        resp4.status_code = 504
 
         resp5 = requests.Response()
-        resp5.json = lambda: dict(
+        resp5.status_code = 429
+
+        resp6 = requests.Response()
```
andrewtruong, line 81: these look complicated but are just renames...

PR #3011, opened 2024-11-19
chore(weave): store microsecond precision timestamps for call start and end, use for sorting

bcsherma:
## Description

Adds a `started_at_micros` and `ended_at_micros` to the `call_parts` table and `calls_merged` materialized view so call execution duration can be tracked in microseconds.

The `call_parts` table gets two new columns while for the materialized view we need to adjust the query being used to select `*_micros` and aggregate them into the view.

The trace server will also save the `started_at` and `ended_at` timestamps it receives from the client (already in microsecodn resolution) to these new columns and use `started_at_microseconds` as the default ordering for when serving calls.

## Questions

How computationally expensive is it for clickhouse to modify the query for the materialized view to `calls_merged`? Will this cause interruptions for large tables like prod?

Are a bunch of trace server queries going to start returning both `started_at` and `started_at_micros`? Same for `ended_at`.


bcsherma:
* **#3012** <a href="https://app.graphite.dev/github/pr/wandb/weave/3012?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3011** <a href="https://app.graphite.dev/github/pr/wandb/weave/3011?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/3011?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

tests/trace/test_client_trace.py
```diff
@@ -186,18 +191,16 @@ def test_trace_server_call_start_and_end(client):
         )
     )
 
-    exp_ended_at = datetime.datetime.fromisoformat(
-        end.ended_at.isoformat(timespec="milliseconds")
-    )
-
```
bcsherma, line 192: this was just unused

weave/trace_server/trace_server_interface.py
```diff
@@ -94,6 +94,8 @@ class CallSchema(BaseModel):
 
     # Start time is required
     started_at: datetime.datetime
+    started_at_micros: Optional[datetime.datetime] = None
```
tssweeney, line 97: i think we can just use started_at and ended_at directly here with higher precision if avail

weave/trace_server/sqlite_trace_server.py
```diff
@@ -91,7 +91,9 @@ def setup_tables(self) -> None:
                 parent_id TEXT,
                 op_name TEXT,
                 started_at TEXT,
+                started_at_micros TEXT,
                 ended_at TEXT,
+                ended_at_micros TEXT,
```
tssweeney, line 96: since the data is exactly the same here, i don't think we need any new fields for sqlite?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1982,7 +1991,9 @@ def _ch_call_dict_to_call_schema_dict(ch_call_dict: dict) -> dict:
         "parent_id": ch_call_dict.get("parent_id"),
         "op_name": ch_call_dict.get("op_name"),
         "started_at": started_at,
+        "started_at_micros": started_at_micros,
         "ended_at": ended_at,
```
tssweeney, line 1995: here i would just use `started_at_micros or started_at`

PR #3015, opened 2024-11-19
chore(weave): Cleanup save object basic

andrewtruong:
Tidying this block

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -1335,13 +1335,13 @@ class WeaveClient:
         """Directly saves an object to the weave server and attach
         the ref to the object. This is the lowest level object saving logic.
         """
+        orig_val = val
+
         # The WeaveTable case is special because object saving happens inside
         # _save_object_nested and it has a special table_ref -- skip it here.
         if getattr(val, "_is_dirty", False) and not isinstance(val, WeaveTable):
             val.ref = None
 
-        is_opdef = is_op(val)
-        orig_val = val
         val = map_to_refs(val)
         if isinstance(val, ObjectRef):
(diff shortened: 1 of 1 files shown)


weave/trace/weave_client.py
```diff
@@ -1335,13 +1335,13 @@ def _save_object_basic(
         """Directly saves an object to the weave server and attach
         the ref to the object. This is the lowest level object saving logic.
         """
+        orig_val = val
```
andrewtruong, line 1338: moved to top for clarity

weave/trace/weave_client.py
```diff
@@ -1366,6 +1363,8 @@ def _save_object_basic(
         name = sanitize_object_name(name)
 
         def send_obj_create() -> ObjCreateRes:
+            # `to_json` is mostly fast, except for CustomWeaveTypes
+            # which incur network costs to serialize the payload
```
andrewtruong, line 1367: moved comment closer to code for clarity

weave/trace/weave_client.py
```diff
@@ -1377,21 +1376,20 @@ def send_obj_create() -> ObjCreateRes:
...
+        if is_op(orig_val):
             ref = OpRef(self.entity, self.project, name, digest_future)
         else:
             ref = ObjectRef(self.entity, self.project, name, digest_future)
 
         # Attach the ref to the object
         try:
             set_ref(orig_val, ref)
-        except:
+        except Exception:
```
andrewtruong, line 1392: blank except is actually `BaseException` which is not safe
https://stackoverflow.com/questions/27995057/why-is-it-recommended-to-derive-from-exception-instead-of-baseexception-class-in

tssweeney (approved):

PR #3020, opened 2024-11-19
chore(weave): human annotation labeling review comments

gtarpenning:
## Description

Crush a bunch of human feedback rough edges. 

This pr: 
+ Sidebar:
  - Store boolean types as boolean, not string
  - add better loading state, don't show CTA when still loading
  - add analytics event
  - increase "save" button padding

+ Trace table:
  - fix paging to keep feedback expanded when changing calls
  - swap icon position (feedback/trace expand)
  - `hide/show feedback sidebar` -> `hide/show feedback`
  - correctly display bool types in trace table 
  
+ Create form:
  - Add integer type (disallows float)
  - "enum" -> "options"
  - fix annotation form default state
  - description optional 
  - zod number `min/max` -> `minimum/maximum`
  - after scorer creation, set form data to null

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/StructuredFeedback/HumanAnnotation.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/StructuredFeedback/HumanAnnotation.tsx
@@ -167,7 +167,7 @@ const FeedbackComponentSelector: React.FC<{
   jsonSchema: Record<string, any>;
   focused: boolean;
   onAddFeedback: (value: any) => Promise<boolean>;
-  foundValue: string | number | null;
+  foundValue: string | number | boolean | null;
   feedbackSpecRef: string;
   setUnsavedFeedbackChanges: React.Dispatch<
     React.SetStateAction<Record<string, () => Promise<boolean>>>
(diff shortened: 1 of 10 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/StructuredFeedback/HumanAnnotation.tsx
```diff
@@ -194,12 +194,23 @@ const FeedbackComponentSelector: React.FC<{
     );
 
     switch (type) {
+      case 'integer':
+        return (
+          <NumericalFeedbackColumn
+            min={jsonSchema.minimum}
```
tssweeney, line 200: make sure your are respecting https://json-schema.org/understanding-json-schema/reference/numeric#range correctly (inclusive vs non inclusive)

tssweeney (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/StructuredFeedback/HumanAnnotation.tsx
gtarpenning: Yeah it should be inclusive by default, which is what the component expects. Rejecting iff: `(min != null && parsedVal < min) || (max != null && parsedVal > max)` so accepting when `val == min`

PR #3023, opened 2024-11-19
chore(weave): feedback payload 1024 -> 1024*1024

gtarpenning:
## Description

Per @shawnlewis 's request, bump limit on payload to basically maxed.

### DIFF
--- a/tests/trace/test_client_feedback.py
+++ b/tests/trace/test_client_feedback.py
@@ -228,7 +228,7 @@ def test_feedback_payload(client):
 def test_feedback_create_too_large(client):
     project_id = client._project_id()
 
-    value = "a" * 10000
+    value = "a" * 1024 * 1024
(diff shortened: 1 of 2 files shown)


tests/trace/test_client_feedback.py
```diff
@@ -228,7 +228,7 @@ def test_feedback_payload(client):
 def test_feedback_create_too_large(client):
     project_id = client._project_id()
 
-    value = "a" * 10000
+    value = "a" * 1024 * 1024
```
jamie-rasmussen, line 231: I am surprised this passes since the size check is for a value > than this (not >=) but InvalidRequest still gets raised?

jamie-rasmussen (commented):
Could you take a look at https://github.com/wandb/weave/pull/2926 and my comment on that - we should see what values this size do to the UI and we should update the docs.

PR #3029, opened 2024-11-20
chore(weave): Fix deps

tssweeney:
Adds jsonschema to deps


jamie-rasmussen:
Retroactively approved

PR #3036, opened 2024-11-20
docs(weave): Add new env vars catalog page under Tools

J2-D2-3PO:
Addresses https://wandb.atlassian.net/browse/DOCS-1053

Adds a new page summarizing and cataloging env vars under Product Walkthrough


docs/docs/guides/tools/env-vars.md
```diff
@@ -0,0 +1,38 @@
...
+|--------------------------|------------------------------|-----------------------------------------------------------------|
+| WEAVE_CAPTURE_CODE      | `False`                       | Disable code capture for `weave.op`.                                   |
+| WEAVE_DEBUG_HTTP        | 1                             | If set to `1`, turns on HTTP request and response logging for debugging. |
+| WEAVE_DISABLED          | `True`                        | If set to `True`, `weave.op` will behave like regular functions.      |
+| WEAVE_PARALLELISM       | `20`                          | In evaluations, the number of examples to evaluate in parallel. `1` runs examples sequentially, while `0` will effectively execute all background operations immediately in the main thread.    |
+| WEAVE_PRINT_CALL_LINK   | `False`                       | Suppress the printing of call URLs.                            |
+| WEAVE_TRACE_LANGCHAIN   | `False`                       |                                                                 |
+| WF_CLICKHOUSE_HOST      | `localhost`                  | The host of the clickhouse server.                                                                |
```
J2-D2-3PO, line 31: @jamie-rasmussen not sure if I was understanding the CLICKHOUSE vars correctly, added default values based on https://github.com/wandb/weave/blob/master/weave/trace_server/environment.py

jamie-rasmussen: I think `WF_CLICKHOUSE*` and `WF_TRACE_SERVER_URL` should not be in public documentation.

docs/sidebars.ts
```diff
@@ -76,7 +76,7 @@ const sidebars: SidebarsConfig = {
           collapsed: true,
           label: "Tools & Utilities",
           link: { type: "doc", id: "guides/tools/index" },
-          items: ["guides/tools/serve", "guides/tools/deploy"],
+          items: ["guides/tools/serve", "guides/tools/deploy", "guides/tools/env-vars"],
```
jamie-rasmussen, line 79: This categorization doesn't seem right to me, these aren't a "tool" or "utility". I would put directly under Product Walkthrough or we could consider adding a category like "Reference"?

docs/docs/guides/tools/env-vars.md
```diff
@@ -0,0 +1,38 @@
...
+# Example of setting environment variables in Python
+import os
+
+os.environ["WEAVE_PARALLELISM"] = "10"
+os.environ["WEAVE_PRINT_CALL_LINK"] = "false"
+```
+
+## Environment variables reference 
+
+Here’s a complete list of available environment variables for Weave, along with their usage:
```
jamie-rasmussen, line 21: I'd probably not call it a complete list, there will be other internal ones.

docs/docs/guides/tools/env-vars.md
```diff
@@ -0,0 +1,38 @@
...
+
+| Variable Name            | Default Values                | Description                                                     |
```
jamie-rasmussen, line 23: I think it is confusing to call this column Default Values since it seems to suggest what would happen if the environment variable is not set. E.g. the default behavior when WEAVE_DEBUG_HTTP is unset is to not do the additional output logging. We should maybe just omit the column and explain the behavior for each in the description.

docs/docs/guides/tools/env-vars.md
```diff
@@ -0,0 +1,38 @@
...
+|--------------------------|------------------------------|-----------------------------------------------------------------|
+| WEAVE_CAPTURE_CODE      | `False`                       | Disable code capture for `weave.op`.                                   |
+| WEAVE_DEBUG_HTTP        | 1                             | If set to `1`, turns on HTTP request and response logging for debugging. |
+| WEAVE_DISABLED          | `True`                        | If set to `True`, `weave.op` will behave like regular functions.      |
+| WEAVE_PARALLELISM       | `20`                          | In evaluations, the number of examples to evaluate in parallel. `1` runs examples sequentially, while `0` will effectively execute all background operations immediately in the main thread.    |
```
jamie-rasmussen, line 28: Where did this information about `0` come from? I don't think that's correct.

jamie-rasmussen (changes requested):

docs/docs/guides/tools/env-vars.md
J2-D2-3PO: ✅ Removed in https://github.com/wandb/weave/pull/3036/commits/427e2d3e7290e5bf558a5193e9e8d0a22b80e1a9

docs/sidebars.ts
J2-D2-3PO: I agree. In the current structure, it makes most sense to me to just put under Product Walkthrough. Updated in https://github.com/wandb/weave/pull/3036/commits/427e2d3e7290e5bf558a5193e9e8d0a22b80e1a9.

Overall, it would probably make more sense in a Reference section, but that gets into the docs structure refactor, which is a can of 🪱 s that we've not yet touched, so gonna punt that down the road

docs/docs/guides/tools/env-vars.md
J2-D2-3PO: ✅ Removed the word "complete" in https://github.com/wandb/weave/pull/3036/commits/427e2d3e7290e5bf558a5193e9e8d0a22b80e1a9

docs/docs/guides/tools/env-vars.md
J2-D2-3PO: ✅ Makes sense. Removed the "default" column in https://github.com/wandb/weave/pull/3036/commits/427e2d3e7290e5bf558a5193e9e8d0a22b80e1a9

docs/docs/guides/tools/env-vars.md
J2-D2-3PO: ✅ I was looking through the code and saw something about this, but I think I misunderstood. Removed in https://github.com/wandb/weave/pull/3036/commits/427e2d3e7290e5bf558a5193e9e8d0a22b80e1a9

docs/docs/guides/core-types/env-vars.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Environment variables reference 
+
+| Variable Name            | Description                                                     |
+|--------------------------|-----------------------------------------------------------------|
+| WEAVE_CAPTURE_CODE      | Disable code capture for `weave.op` if set to `false`.                                    |
+| WEAVE_DEBUG_HTTP        | If set to `1`, turns on HTTP request and response logging for debugging.  |
+| WEAVE_DISABLED          | If set to `true`, all tracing to Weave is disabled.      |
+| WEAVE_PARALLELISM       | In evaluations, the number of examples to evaluate in parallel. `1` runs examples sequentially. Default value is `20`.    |
+| WEAVE_PRINT_CALL_LINK   | Suppress the printing of call URLs. Default value is `false`.                            |
```
jamie-rasmussen, line 27: I'd say "if set to `false`." to match the first row.

jamie-rasmussen (approved):

PR #3044, opened 2024-11-21
chore(ui): Flatten nested keys on calls table

m-rgba:
## Description

| Before    | After |
| -------- | ------- |
| before | after |

- Flattens the headers in `buildDynamicColumns` so there will be max-3 dynamic headers. 
- The middle keys will be flattened into a string with dot notation (this is to avoid really large headers for customers passing nested data).


m-rgba:
Fixed FeatureBee link: https://beta.wandb.ai/?betaVersion=4227d5494b8a3e8795469c888d8495b9726b74da

m-rgba:
Design feedback:
- For objects with multiple levels, only the parent will be showing in the header. 
- Solutions: 
   - Try overflow-truncating from different areas.
   - Experiment possibly allowing one more level (so it'd be the level under inputs, outputs, attributes)

PR #3050, opened 2024-11-22
chore(weave): Wrap up loose ends with Annotations

tssweeney:
1. Rename `json_schema` to `field_schema`
  - [ ] Todo: verify UI still works
2. Adds backend validation when creating feedback records
  - [ ] Todo: add to sqlite & write better code
3. Allows user to pass a Pydantic BaseModel or Field to the field_schema and not worry about json.

### DIFF
--- a/weave/trace_server/interface/base_object_classes/annotation_spec.py
+++ b/weave/trace_server/interface/base_object_classes/annotation_spec.py
@@ -1,14 +1,14 @@
-from typing import Optional, Union, Type, Dict, Any
+from typing import Any, Optional
 
 import jsonschema
-from pydantic import BaseModel, field_validator, Field
+from pydantic import BaseModel, Field, field_validator, model_validator
 from pydantic.fields import FieldInfo
 
 from weave.trace_server.interface.base_object_classes import base_object_def
 
 
 class AnnotationSpec(base_object_def.BaseObject):
-    field_schema: Dict[str, Any] = Field(
+    field_schema: dict[str, Any] = Field(
         default={},
         description="Expected to be valid JSON Schema. Can be provided as a dict, a Pydantic model class, or a Pydantic Field",
         examples=[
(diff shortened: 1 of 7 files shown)


weave/trace_server/interface/base_object_classes/annotation_spec.py
```diff
@@ -41,12 +42,49 @@ class AnnotationSpec(base_object_def.BaseObject):
...
+        field_schema = data["field_schema"]
+
+        # Handle Pydantic Field
+        if isinstance(field_schema, FieldInfo):
+            data["field_schema"] = field_schema.json_schema()
+            return data
+
+        # Handle Pydantic model
+        if isinstance(field_schema, type) and issubclass(field_schema, BaseModel):
+            data["field_schema"] = field_schema.model_json_schema()  # type: ignore
```
gtarpenning, line 60: We currently don't support nested objects, we might want to throw with an actionable error mentioning they will have to create multiple specs. I will add a ticket to move as much of the human feedback UI to the Zod form, which will allow for nested.

weave/trace_server/interface/base_object_classes/annotation_spec.py
```diff
@@ -41,12 +42,49 @@ class AnnotationSpec(base_object_def.BaseObject):
...
+    def validate(self, payload: Any) -> bool:
+        """
+        Validates a payload against this annotation spec's schema.
+
+        Args:
+            payload: The data to validate against the schema
+
+        Returns:
+            bool: True if validation succeeds, False otherwise
+        """
```
gtarpenning, line 85: I think including the work "feedback" or in some way tying it to the feedback schema might make sense here, perhaps an example

gtarpenning (approved):
Sweet

shawnlewis (approved):

weave/trace_server/interface/base_object_classes/annotation_spec.py
tssweeney: I think we should allow the creation, but maybe in the UI there is a fallback that allows free-form json input or something

PR #3058, opened 2024-11-22
fix(weave_query): add refinement op to get types for artifactVersion-files op

domphan-wandb:
## Description
Fixes [WB-21719](https://wandb.atlassian.net/browse/WB-21719)

A user attempted to render a list of images from an artifact version using **`project.artifactVersion("<art_name>", "<version>").files.content`** which crashes the panel because the **`.content()`** op is meant to display the contents of a text file, not a png.

Since the intention of the user is to render a list of images and the expression **`project.artifactVersion().file("<image.(png|jpg|jpeg)>")`** renders one image file with a `PanelImage`, the **`.files()`** op should also render a list of images.

I added a refinement op for the **`artifactVersion-files`** op.

Accompanying weave-js changes: https://github.com/wandb/weave/pull/3083

## Testing

Before:
image
image

After:
image

My dev setup doesn't properly show the image assets due to an ssl issue, but in the next screenshot, you can see a table of images being rendered because the file type has been refined:
image

How was this PR tested?
- Manually

[WB-21719]: https://wandb.atlassian.net/browse/WB-21719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/core/ops/domain/artifactVersion.ts
+++ b/weave-js/src/core/ops/domain/artifactVersion.ts
@@ -179,6 +179,58 @@ export const opArtifactVersionFileCount = makeArtifactVersionOp({
   resolver: ({artifactVersion}) => artifactVersion.fileCount,
 });
 
+export const opArtifactVersionFilesType = makeArtifactVersionOp({
+  name: 'artifactVersion-_files_refine_output_type',
+  argTypes: artifactVersionArgTypes,
+  description: `Returns the type of a ${docType('list')} of ${docType('file', {
+    plural: true,
+  })} of the ${docType('artifactVersion')}`,
+  argDescriptions: {
+    artifactVersion: artifactVersionArgDescription,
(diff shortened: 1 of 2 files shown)


nickpenaranda (approved):
LGTM but you might also want a review from the weave team proper.  Also I'd be careful making this change in one PR since I think this will break if the app is updated before `weave-python`

domphan-wandb:
> LGTM but you might also want a review from the weave team proper. Also I'd be careful making this change in one PR since I think this will break if the app is updated before `weave-python`

thank you! you're saying the risk is someone deploying `fe prod` because I can get the `weave-python` changes out? I'll go ahead and break this apart so `weave-python` can be deployed first

PR #3060, opened 2024-11-22
fix(weave_query): prevent table state loss when encountering a new logged column

domphan-wandb:
## Description
- Fixes [WB-17336](https://wandb.atlassian.net/browse/WB-17336)

There was a user experience issue causing a table's state (derived columns, sorting, etc) to be wiped when the user logged an additional column with their table in a subsequent run.

Initially I couldn't reproduce it, but realized the issue only occurs after the first new column is added-- subsequent added columns did not reset the table state.

**Root cause**: the **`typeShapesMatch`** util function (only used to determine if tableState should be migrated) does not handle the case of a `type` being a `union<typedDict, typedDict>` and the `toType` being a `typedDict` where the union members match the property types of the `toType`'s `typedDict`.

This fix adds a check when the `toType` is a `typedDict` and the `type` is a `union<typedDict>`. It uses the existing typedDict checking logic to confirm all the members of the union are compatible. I also refactored some of the conditional checks to make it a bit more readable.

I also need to note that the existing behavior did not correctly check the case where two unions are passed in. It will by default return true. In order to reduce the chances of shipping a regression, I've left that untouched and left a comment in the code.

https://github.com/user-attachments/assets/53384985-f791-42f1-859a-108197a9a20b

## Testing
* Unit tests
* Local dev QA

[WB-17336]: https://wandb.atlassian.net/browse/WB-17336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/Panel2/PanelTable/util.ts
+++ b/weave-js/src/components/Panel2/PanelTable/util.ts
@@ -116,6 +116,13 @@ export const typeShapesMatch = (type: Type, toType: Type): boolean => {
   // This util's original behavior never properly handled comparing two union types together;
   // it just returned true. To minimize the risk of shipping a regression, I am leaving this
   // legacy behavior alone for now.
+
+  // This returns true when comparing:
+  //   Two compatible TypedDicts
+  //   One TypedDict to a compatible Union<TypedDict[]>
+  //   Two Unions of any type
+  //
+  //   Two Lists that contain any of the types mentioned above
   return true;
 };
 
--- a/weave-js/src/components/Panel2/PanelTable/util.ts
+++ b/weave-js/src/components/Panel2/PanelTable/util.ts
@@ -5,6 +5,7 @@ import {
   isListLike,
   isTaggedValue,
   isTypedDict,
+  isUnion,
   listObjectType,
   ListType,
   MemoizedHasher,
@@ -18,6 +19,7 @@ import {
   OutputNode,
   taggedValueValueType,
   Type,
+  TypedDictType,
   WeaveInterface,
 } from '@wandb/weave/core';
 import _ from 'lodash';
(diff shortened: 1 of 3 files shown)


weave-js/src/components/Panel2/PanelTable/util.ts
```diff
@@ -60,29 +62,60 @@ export const stripTag = (type: Type): Type => {
...
-        if (keyType === undefined || !typeShapesMatch(keyType, toKeyType)) {
-          return false;
-        }
-      }
     }
+
+    return propertyTypesAreCompatible(type, toType);
   }
+
+  // 2024/11/20, Note from Dom:
```
gtarpenning, line 115: this comment could be made more actionable by describing what case we return true here

weave-js/src/components/Panel2/PanelTable/util.ts
```diff
@@ -60,29 +62,60 @@ export const stripTag = (type: Type): Type => {
...
+    }
+    return true;
+  }
+
   type = stripTag(type);
   toType = stripTag(toType);
+
+  // List type handling
   if (isList(type) || isList(toType)) {
-    if (!isList(type) || !isList(toType)) {
```
gtarpenning, line 68: weird...

weave-js/src/components/Panel2/PanelTable/util.ts
```diff
@@ -60,29 +62,60 @@ export const stripTag = (type: Type): Type => {
...
+    // This is meant for special cases where a user adds another column to their table
+    // in a subsequent run.
+    if (isTypedDict(toType) && isUnion(type)) {
+      return type.members.every(
+        member =>
+          isTypedDict(member) &&
+          isTypedDict(toType) &&
+          propertyTypesAreCompatible(member, toType)
+      );
     }
```
gtarpenning, line 106: everything else is just refactor (?), this lgtm

gtarpenning (approved):

weave-js/src/components/Panel2/PanelTable/util.ts
domphan-wandb: Yup, besides the union checking code, I refactored to make it a bit more readable. Hope it didn't make the review hard to parse. Thanks for reviewing my PRs, @gtarpenning , I appreciate it! 🙏

dannygoldstein (approved):
nice pr! love the thorough tests and extensive write up. this pr was a pleasure to review

my one comment is that we should rename this function from `typeShapesMatch` to something else, like `isConcattable` or something (i leave it to you to think up a better name). as it stands, this function is somewhere in between `isAssignable()` and `typeShapesMatch()`: it does not check strict type shape equality (the toType can have  less columns than the fromType), but it is also not checking assignability (it would return true for string <- number).

domphan-wandb:
> nice pr! love the thorough tests and extensive write up. this pr was a pleasure to review
> 
> my one comment is that we should rename this function from `typeShapesMatch` to something else, like `isConcattable` or something (i leave it to you to think up a better name). as it stands, this function is somewhere in between `isAssignable()` and `typeShapesMatch()`: it does not check strict type shape equality (the toType can have less columns than the fromType), but it is also not checking assignability (it would return true for string <- number).

Thanks Danny! I'm bad at naming things, so I renamed it to `typesAreConcattable` 👍

PR #3069, opened 2024-11-25
fix(weave): netrc entry handling in Weave JS SDK login method

thanos-wandb:
## Description

- Fixes [WB-22046](https://wandb.atlassian.net/browse/WB-22046)

This PR resolves an issue in the Weave JS SDK's login method where the `.netrc` file was being modified or created incorrectly if the `apiKey` arg was empty.

Changes:
- Added a validation check to ensure the `apiKey` is provided and non-empty
- Throws an error if the `apiKey` is missing
- Prevents overwriting existing `.netrc` entries unless both `host` and `apiKey` are valid and non-empty

Key fixes:
- Ensures that `.netrc` entries remain intact and prevent removal of the password line for existing entries
- Avoids introducing extra blank lines between machine blocks in the `.netrc` file

## Testing

To reproduce the issue:
- Start with an existing `.netrc` file containing valid machine entries
- Execute the following JS script where `login` method is missing the `apiKey` arg
```
import * as weave from 'weave';
async function main() {
    await weave.login()
    await weave.init('examples');
    console.log('Hello there, Weave JS');
}
main().catch(console.error);
```
- The .netrc file gets modified incorrectly (password line of api.wandb.ai is removed and blank lines are introduced between machine blocks)

Fix testing:
- Verified that `.netrc` won't be created, if doesn't already exist, and no `apiKey` is provided
- Verified that previous valid entries in `.netrc` remain intact when `weave.login()` is invoked

[WB-22046]: https://wandb.atlassian.net/browse/WB-22046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdks/node/src/clientApi.ts
+++ b/sdks/node/src/clientApi.ts
@@ -24,6 +24,9 @@ export async function login(apiKey: string, host?: string) {
     console.warn('No host provided, using default host:', defaultHost);
     host = defaultHost;
   }
+  if (!apiKey) {
+    throw new Error('API key is required for login. Please provide a valid API key.');
+  }
   const {traceBaseUrl} = getUrls(host);
 
   // Test the connection to the traceServerApi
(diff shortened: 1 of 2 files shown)


sdks/node/src/clientApi.ts
```diff
@@ -24,6 +24,9 @@ export async function login(apiKey: string, host?: string) {
     console.warn('No host provided, using default host:', defaultHost);
     host = defaultHost;
   }
+  if (!apiKey) {
```
andrewtruong, line 27: This is specifically for `apiKey == ''`?

thanos-wandb: I think there was no validation in place to enforce this requirement before, but should raise an error now by adding that line. I tested with this script before:
```
import * as weave from 'weave';
async function main() {
    await weave.login()
    await weave.init('examples');
    console.log('Hello there, Weave JS');
}
main().catch(console.error);
```

andrewtruong (approved):

PR #3072, opened 2024-11-25
WIP: Callbacks

andrewtruong:
### Motivation
1. Provide first-class support for logging iterators, including generators.
2. Make the op execution machinery easier to understand

### Callbacks
The big change in this PR is introducing `Callbacks` and a `LifecycleHandler` which manages the callbacks during op execution.

This allows users to hook into and add custom code specific stages of op execution, like adding debugging statements during execution start, or accumulating a stream into the call's output as values are yielded.

The `Callback` API is designed to be simple and flexible:
```py
class Callback(Protocol):
    """A callback that can be registered with an op to handle lifecycle events.
    You can inherit from this class, or implement part of the protocol to add lifecycle
    hooks to an op.
    """

    def before_call_start(
        self, inputs: dict, parent: Call | None, attributes: dict | None
    ) -> None: ...
    def before_yield(self, call: Call, value: Any) -> None: ...
    def before_call_finish(self, call: Call) -> None: ...
    def after_error(self, call: Call, error: Exception) -> None: ...
```

### Reducers
Reducers are callables that take a value and accumulator and fold them into a new accumulator.  This is useful for logging the output an iterator, like a stream of text or a generator of values.

Here's a reducer for a stream of text:
```py
def concat(v: str, acc: str) -> str:
    return acc + v
```

The reducer implementation converts the reducer function into a ReducerCallback and attaches the accumulated result into the call's output.

### Integrations
Eventually we'll need to move all of the integrations to this pattern.  I've implemented Anthropic to give reviews a sense of how this change will affect other integrations.

### Op saving
If a reducer is specified and used in the op, it's included as part of the op saved code.


weave/integrations/anthropic/anthropic_sdk.py
```diff
@@ -1,89 +1,83 @@
...
+    def before_yield(self, call: Call, value: Any) -> None:
+        from anthropic.types import (
+            ContentBlockDeltaEvent,
+            Message,
+            MessageDeltaEvent,
+            TextBlock,
+            Usage,
+        )
+
+        if call.output is None:
```
andrewtruong, line 26: Accumulation happens directly on the `call.output` itself.  Since each `Call` is unique to the execution, this is safe.

One nice side effect of accumulating directly on `call` is that it's a more faithful representation of what the user sees.  If you stop the stream midway, the the output will look like a partial output

weave/integrations/anthropic/anthropic_sdk.py
```diff
@@ -1,89 +1,83 @@
...
+                type=value.message.type,
                 usage=Usage(input_tokens=0, output_tokens=0),
             )
-        else:
-            raise ValueError("Initial event must contain a message")
 
-    # Merge in the usage info if available
-    if hasattr(value, "message") and value.message.usage is not None:
-        acc.usage.input_tokens += value.message.usage.input_tokens
+        # Merge in the usage info if available
```
andrewtruong, line 40: These are non-changes; basically just renames `acc -> call.output`

weave/trace/op.py
```diff
@@ -734,4 +785,294 @@ def as_op(fn: Callable) -> Op:
...
+        try:
+            val = await self._aiterator.__anext__()
+        except StopAsyncIteration:
+            raise
+        self.op.lifecycle_handler.before_yield(self.call, val)
+        return val
+
+
+@contextmanager
+def _call_context(op: Op, call: Call, should_raise: bool):
```
andrewtruong, line 899: All op execution happens in this context.  This block basically says:
1. Do your thing
2. If you fail, run the error callback
3. Always try to finish the call out

weave/trace/op.py
```diff
@@ -734,4 +785,294 @@ def as_op(fn: Callable) -> Op:
...
+    async def _wrapped_async_generator() -> Any:
+        with _call_context(op, call, should_raise):
+            async for val in await func(*args, **kwargs):
+                op.lifecycle_handler.before_yield(call, val)
+                yield val
+
+    return _wrapped_async_generator()
+
+
+async def _exc_op_async(
```
andrewtruong, line 958: This will be renamed to `_execute_op_async`.  It's named this way for now because it makes the diff easier to read.

tests/trace/test_reducers.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # The output is accumulated
+    assert call.output == "the quick brown fox"
+
+    # The reducer is also captured in the op code
+    ref = weave.publish(generator)
+    op = ref.get()
+
+    reducer_src = textwrap.dedent(inspect.getsource(concat))
+    op_code = op.art.path_contents["obj.py"].decode()
```
andrewtruong, line 35: this is kinda sketchy.  We should probably make a property for this...

PR #3077, opened 2024-11-25
perf(ui): second cost query ignore all expensive columns

gtarpenning:
## Description

This pr removes all expensive columns from the second calls query intended to inject costs into the calls table. There are currently many queries that fail in the UI due to the sheer amount of data required to load the trace page, either causing a memory exception or a timeout on formatting. Most of these errors occur in the full fat cost query, lets see if this reduces the impact of these.

## Testing

👀 

### Openui
Prod: 
Screenshot 2024-11-25 at 2 23 35 PM

This branch:
Screenshot 2024-11-25 at 2 23 55 PM

### Heavy proj
Prod:
Screenshot 2024-11-25 at 2 29 24 PM

This branch:
Screenshot 2024-11-25 at 2 28 01 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/cost/costUtils.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/cost/costUtils.ts
@@ -102,7 +102,23 @@ export const addCostsToCallResults = (
 
   return callResults.map(call => {
     if (call.callId && costDict[call.callId]) {
-      return {...call, ...costDict[call.callId]};
+      if (!call.traceCall) {
+        return call;
+      }
+      return {
+        ...call,
+        traceCall: {
+          ...call.traceCall,
+          summary: {
+            ...call.traceCall?.summary,
+            usage: costDict[call.callId].traceCall?.summary?.usage,
+            weave: {
+              ...call.traceCall?.summary?.weave,
+              costs: costDict[call.callId].traceCall?.summary?.weave?.costs,
+            },
+          },
+        },
+      };
     }
     return call;
   });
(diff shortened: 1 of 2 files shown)


jwlee64 (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/cost/costUtils.ts
```diff
@@ -102,7 +102,22 @@ export const addCostsToCallResults = (
...
+      }
+      return {
+        ...call,
+        traceCall: {
+          ...call.traceCall,
+          summary: {
+            ...call.traceCall?.summary,
+            weave: {
+              ...call.traceCall?.summary?.weave,
+              costs: costDict[call.callId].traceCall?.summary?.weave?.costs,
```
gtarpenning, line 116: like here

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
```diff
@@ -104,12 +105,11 @@ export const useCallsForQuery = (
     undefined,
     sortBy,
```
jwlee64, line 106: i think its this sort by here, we try to sort by outputdump, but dont have the requested column

jwlee64: i think we can remove the sortby since we stitch the results together later anyway

gtarpenning: ah good call, that explains why its a very infrequent error.

PR #3083, opened 2024-11-25
fix(weave_query): Add artifactVersion-file type refinement to weave-js

domphan-wandb:
## Description

Fixes [WB-21719](https://wandb.atlassian.net/browse/WB-21719)

See https://github.com/wandb/weave/pull/3058 for full context. This PR was created after the fact to separate backend and frontend changes to prevent issues from deploy ordering.

## Testing

How was this PR tested?
* Manually

**Here's some weave0 screenshots**

Before:
image

After (yes, I should have used different images that weren't a green rectangle 😅):
image

[WB-21719]: https://wandb.atlassian.net/browse/WB-21719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/core/ops/domain/artifactVersion.ts
+++ b/weave-js/src/core/ops/domain/artifactVersion.ts
@@ -179,6 +179,58 @@ export const opArtifactVersionFileCount = makeArtifactVersionOp({
   resolver: ({artifactVersion}) => artifactVersion.fileCount,
 });
 
+export const opArtifactVersionFilesType = makeArtifactVersionOp({
+  name: 'artifactVersion-_files_refine_output_type',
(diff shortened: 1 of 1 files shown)


gtarpenning (approved):
I'm curious if we have analytics on op usage...

domphan-wandb:
> I'm curious if we have analytics on op usage...

Yeah, I added analytics to track the expressions that users submit under the action "Expression.SET_EXP", but tracking things at the Op level would probably make things way easier to analyze. I'll try to add it when I have some spare cycles.

PR #3088, opened 2024-11-26
fix(ui): improved comparison performance and object handling

jamie-rasmussen:
## Description

Internal Jira: https://wandb.atlassian.net/browse/WB-22043

This PR handles a few problems with the comparison page:

* We were autoexpanding all groups. With large objects this causes the page to repeatedly freeze for several seconds. Two mitigations:
  * Limit the depth we'll consider auto expanding for
  * Don't autoexpand rows with no changes (a better UX for emphasizing differences anyway)
* The code editor component was capturing scroll events, which could make the page appear frozen as you tried to scroll. 
* For non-ref objects and arrays, we now display a key/item count and a pill if it has changed. This helps user understand the change (especially if not expanded) but also mentally prepare for possible slowness expanding large objects.
Screenshot 2024-11-26 at 10 33 41 AM
Screenshot 2024-11-26 at 10 36 13 AM

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CodeDiff.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CodeDiff.tsx
@@ -17,6 +17,8 @@ type CodeDiffProps = {
   maxLines?: number;
 };
 
+const DEFAULT_MAX_LINES = 20;
+
 export const CodeDiff = ({
   oldValueRef,
   newValueRef,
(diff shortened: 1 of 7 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CodeDiff.tsx
```diff
@@ -79,14 +71,17 @@ export const CodeDiff = ({oldValueRef, newValueRef}: CodeDiffProps) => {
...
+          },
           padding: {top: 10, bottom: 10},
         }}
       />
     );
 
-  const maxRowsInView = 20;
   const totalLines = sanitizedNew.split('\n').length ?? 0;
-  const showLines = Math.min(totalLines, maxRowsInView);
+  const showLines = Math.min(totalLines, maxLines ?? 20);
```
gtarpenning, line 84: could be worth a constant here

gtarpenning: and in `CodeView`

gtarpenning (approved):

PR #3095, opened 2024-11-26
chore(weave): Allow removing the system message

jwlee64:
## Description
some users want to remove the last message, should not force users to have a message in the playground

## Testing

How was this PR tested?


jamie-rasmussen (approved):
Could you update the PR description to provide the context for the change

PR #3096, opened 2024-11-26
chore(weave): load model default properly

jwlee64:
## Description

before the model was not loaded into the settings, when opening a trace in playground
adds a smart default, if the model is not one of the models we support

389687742-2ca90dba-ddf8-401e-bf4a-998e43995b23

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/llmMaxTokens.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/llmMaxTokens.ts
@@ -121,3 +121,68 @@ export const LLM_MAX_TOKENS = {
 };
 
 export type LLMMaxTokensKey = keyof typeof LLM_MAX_TOKENS;
+
+export const LLM_MAX_TOKENS_KEYS: LLMMaxTokensKey[] = Object.keys(
+  LLM_MAX_TOKENS
+) as LLMMaxTokensKey[];
+
+// Helper function to calculate string similarity using Levenshtein distance
+const getLevenshteinDistance = (str1: string, str2: string): number => {
+  const track = Array(str2.length + 1)
+    .fill(null)
+    .map(() => Array(str1.length + 1).fill(null));
+
+  for (let i = 0; i <= str1.length; i++) {
+    track[0][i] = i;
+  }
+  for (let j = 0; j <= str2.length; j++) {
+    track[j][0] = j;
+  }
+
+  for (let j = 1; j <= str2.length; j++) {
+    for (let i = 1; i <= str1.length; i++) {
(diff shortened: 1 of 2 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/llmMaxTokens.ts
```diff
@@ -121,3 +121,68 @@ export const LLM_MAX_TOKENS = {
 };
 
 export type LLMMaxTokensKey = keyof typeof LLM_MAX_TOKENS;
+
+export const LLM_MAX_TOKENS_KEYS: LLMMaxTokensKey[] = Object.keys(
```
gtarpenning, line 125:
```suggestion
export const LLM_MAX_TOKENS_KEYS = Object.keys(
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/llmMaxTokens.ts
```diff
@@ -121,3 +121,68 @@ export const LLM_MAX_TOKENS = {
...
+  LLM_MAX_TOKENS
+) as LLMMaxTokensKey[];
+
+// Helper function to calculate string similarity using Levenshtein distance
+const getLevenshteinDistance = (str1: string, str2: string): number => {
```
gtarpenning, line 130: I think we are using `js-levenshtein` in the app, can you use that?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/usePlaygroundState.ts
```diff
@@ -106,6 +111,20 @@ export const usePlaygroundState = () => {
...
+        if (inputs.model) {
+          if (LLM_MAX_TOKENS_KEYS.includes(inputs.model as LLMMaxTokensKey)) {
+            newState.model = inputs.model as LLMMaxTokensKey;
+          } else {
+            const closestModel = findMostSimilarLLMName(
+              inputs.model,
+              LLM_MAX_TOKENS_KEYS
+            );
+            toast(
+              `We currently don't support ${inputs.model}, in the playground. We will default to ${closestModel}`
```
gtarpenning, line 123: do we always want to default to the most similar model? can it ever be a typo? I would expect this case to most likely happen with custom models, which they might not want to have auto-selected.

gtarpenning (commented):
I think the smart default might be a bit too magic. Deferring to the team

jwlee64:
@gtarpenning down to remove the smart default to just load the model in if it exist and if not go with the default model

PR #3101, opened 2024-11-27
fix(ui): Prevent Switch component from shrinking in flex containers

molliean:
## Description

https://wandb.atlassian.net/browse/WB-21938

Adds the `flex-shrink-0` class to the `Switch.Root` component to prevent shrinking in flex containers across viewport sizes. Affects the following files:

- `ArtifactDag.tsx`
- `SearchAutoExpandToggle.tsx`
- `WBTableActionsFilter.tsx`
- `OrgPrivacySettings.tsx`

## Testing

Org Privacy Settings:

Before:

Screenshot 2024-11-25 at 5 15 48 PM

After:

https://github.com/user-attachments/assets/b1c78765-145c-4772-8739-f5ce07dea0cc

How was this PR tested?


ssisk (approved):
approved since this fixes the issue, but if you have time you might try min-width instead

PR #3093, opened 2024-11-26
docs(weave): Update new feedback docs

J2-D2-3PO:
## Description

https://wandb.atlassian.net/browse/DOCS-1067

Updates existing Feedback feature docs https://weave-docs.wandb.ai/guides/tracking/feedback/

## Testing

- [x] Tested instructions as documented @J2-D2-3PO 
- [x] SME review

PR #3104, opened 2024-11-27
chore(ui): change human annotation docs link

gtarpenning:
## Description

Change docs link to match: https://github.com/wandb/weave/pull/3093


jamie-rasmussen (approved):
I remember some discussion of using wandb.me links for docs so we would get better analytics?

gtarpenning:
> I remember some discussion of using wandb.me links for docs so we would get better analytics? @jamie-rasmussen 

Oh... I see in the EMTPY state file there are actually 3 totally different kinds of links!

J2-D2-3PO:
@gtarpenning I changed the branch to merge into to DOCS-1067 (the branch for https://github.com/wandb/weave/pull/3093 that changes the links) because it seems like it could easier to roll everything up at once, and that way no links will be broken in deployment. LMK if you want me to change back though.

PR #3106, opened 2024-11-27
fix(ui): better defaults for trace/eval trace tree

gtarpenning:
## Description

While adding support for maintaining show-annotation state, I broke the trace tree defaults. Specifically we want to be able to show or hide the trace tree when the user sets it, but default to showing when a trace is normal, or hiding when the trace is an evaluation. 

## Testing

test-url-param-tree


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
```diff
@@ -237,6 +239,7 @@ function buildCallsTableColumns(
             fullWidth={true}
             preservePath={preservePath}
             color={TEAL_600}
+            isEval={isEval}
```
gtarpenning, line 242: necessary change because the opName passed into callLink can be displayName

jamie-rasmussen (approved):

Issue #1879, opened 2024-07-03
Add support for Gemini API


masterbatcoderman10:
I'm creating this issue to request support on the Gemini API, currently, tracked calls for the Gemini API do not show any information on the token/cost breakdown, therefore, please enable support as you have for Mistral, and Anthropic.

jwlee64:
Hi, 
I am one of the developer for weave, I can add cost support for gemini apis.
Could you send the link to the project your are working in so I can ensure that my fix works for your project?

masterbatcoderman10:
Dear Josiah

I am grateful for your commitment and your support, here is the link to my
project.

https://wandb.ai/lighthouseai/QA-monitoring/weave/traces

Thanks,
Mohammed Ali

morganmcg1:
@masterbatcoderman10 just to let you know that @soumik12345 is working on a PR at the moment to add support for Google AI Studio

https://github.com/wandb/weave/pull/2078

tcapelle:
The integration is merged now; I think token usage should be there: https://github.com/wandb/weave/pull/2399

masterbatcoderman10:
Thanks for the notification.

ramon-prieto:
@tcapelle this is still not working if using the `vertexai` library, which is the preferred method if accessing Gemini via GCP. Any chance it can be implemented?

soumik12345:
@ramon-prieto Thanks for raising the issue with the `vertexai` integration. I have raised a PR https://github.com/wandb/weave/pull/3110 to fix it.

PR #3110, opened 2024-11-29
fix(weave): dictify for vertexai integration

soumik12345:
- [x] Fix dictification for the vertexai integration as pointed out in https://github.com/wandb/weave/issues/1879#issuecomment-2506894735
- [x] Add patching for `chat.send_message`


weave/integrations/vertexai/vertexai_sdk.py
```diff
@@ -5,12 +5,25 @@
...
     from vertexai.generative_models import GenerationResponse
 
 
+def vertexai_postprocess_inputs(inputs: dict[str, Any]) -> dict[str, Any]:
+    if "self" in inputs:
+        model_name = (
+            inputs["self"]._model_name
+            if hasattr(inputs["self"], "_model_name")
+            else inputs["self"]._model._model_name
+        )
```
andrewtruong, line 21: nit: just use regular if/else, ternary here is more confusing

andrewtruong (approved):
Approving with the understanding that you have thoroughly checked the integration since these tests are skipped

PR #3117, opened 2024-12-02
docs(weave): Update Playground docs with Bedrock info

J2-D2-3PO:
Addresses https://wandb.atlassian.net/browse/DOCS-1089

Adds Bedrock models and note about credentials to existing Playground docs edited in https://wandb.atlassian.net/browse/DOCS-1065


jwlee64 (approved):
thank you!

PR #3129, opened 2024-12-03
fix(weave): Add explicit object preparation hook for types that need it (e.g. `PIL.Image.Image`)

andrewtruong:
Some libs, like PIL, lazily load their data.  This can have a strange interaction in combination with our async/threading model where the object is returned in an "unready" state.

This PR adds support for a `prepare_obj` step inside `make_trace_obj` which is can be used to fully load objects before they are returned to the user.  In the PIL case, this is simply calling `image.load()` to ensure all the bytes are in mem before returning.

Resolves: https://wandb.atlassian.net/browse/WB-22130


weave/trace/api.py
```diff
@@ -9,9 +9,9 @@
 from collections.abc import Iterator
 from typing import Any
 
-# TODO: type_serializers is imported here to trigger registration of the image serializer.
+# TODO: type_handlers is imported here to trigger registration of the image serializer.
```
andrewtruong, line 12: renamed to reflect that it does more than just serializing now

weave/trace/vals.py
```diff
@@ -643,7 +644,7 @@ def make_trace_obj(
             )
         )
         val = from_json(read_res.obj.val, val.entity + "/" + val.project, server)
-
+        initialize_object(val)
```
andrewtruong, line 647: The expected initialization path -- make sure the image is fully loaded before returning it

weave/trace/vals.py
```diff
@@ -753,6 +754,7 @@ def make_trace_obj(
     else:
         if hasattr(box_val, "ref"):
             setattr(box_val, "ref", new_ref)
+        initialize_object(box_val)
```
andrewtruong, line 757: Backup initialization path -- I wouldn't expect anything here, but included for safety

weave/type_handlers/__init__.py
```diff
@@ -0,0 +1,5 @@
+from weave.type_handlers.Audio import audio
+from weave.type_handlers.Image import image
+
+image.register()
+audio.register()
```
adrnswanberg, line 5: why not keep this pattern for registering preparers?

andrewtruong: yeah I guess we could put it here -- will change

andrewtruong: changed

weave/type_serializers/__init__.py
andrewtruong, line 1: this was not deleted, just renamed

gtarpenning (approved):
A couple nits, one meta comment: iirc this pr forceably loads pil images in vals.py on L647 so that we always have the whole image by the time we are serializing. We can't just call `image.load()` in the `save()` serializer method because out of the serializer scope we close the image read buffer/stream?

weave/trace/object_preparers.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ObjectPreparer(Protocol):
+    """An initializer to ensure saved Weave objects are safe to load back to their original types.
+
+    In many cases, this will be some form of deepcopy to ensure all the data is loaded
+    into memory before attempting to return the object.
+    """
+
+    def should_prepare(self, obj: Any) -> bool: ...
```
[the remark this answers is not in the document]
andrewtruong: I'm not certain this will always strictly be a type check.  Maybe it will end up being that way, but keeping it flexible for now

weave/trace/object_preparers.py
```diff
@@ -0,0 +1,26 @@
...
+
+def register(preparer: ObjectPreparer) -> None:
+    _object_preparers.append(preparer)
+
+
+def prepare_obj(obj: Any) -> None:
+    for initializer in _object_preparers:
+        if initializer.should_prepare(obj):
+            initializer.prepare(obj)
+            break
```
gtarpenning, line 26:
```suggestion
def get_preparer_for_obj(obj: Any) -> None:
      for initializer in _object_preparers:
        if initializer.should_prepare(obj):
           return initializer

def prepare_obj(obj: Any) -> None:
    preparer = get_preparer_for_obj(obj)
    if preparer:
        initializer.prepare(obj)
```

weave/trace/vals.py
```diff
@@ -753,6 +754,7 @@ def make_trace_obj(
     else:
         if hasattr(box_val, "ref"):
             setattr(box_val, "ref", new_ref)
+        prepare_obj(box_val)
```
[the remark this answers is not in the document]
andrewtruong: That's fair.  I put this mostly for safety but you're right we could probably remove or if we keep it we should do the preparation before boxing

weave/trace/object_preparers.py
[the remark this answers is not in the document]
andrewtruong: I don't understand this

andrewtruong: How is this different / what's the benefit?

gtarpenning: Just stylistically reinforces one preparer per type, rn it looks like we take the first one 🤷🏻  mega nit

andrewtruong: ok I see that, we can do this :)

andrewtruong:
@gtarpenning thanks for comments -- RE: your first question, yep that's the first case.  The second case is where you're loading the image back (e.g. you get a ref to a dataset that has images).

PR #3142, opened 2024-12-04
feat(ui): support base64 encoded images in markdown rendering

trane293:
## Description

- Fixes [WB-22155](https://wandb.atlassian.net/browse/WB-22155)

Resolves issues with base64-encoded PNG images not rendering correctly in dataset descriptions when using markdown format. The [SanitizationSchema](https://github.com/wandb/weave/blob/master/weave-js/src/common/util/markdown.ts#L22) restricted `src` protocols that were not `http` or `https`. Included `data` as a valid protocol since that is used to render base64 encoded images. 

## Testing

How was this PR tested?
- Local `invoker_fe_prod.ini` 
- Tested on Dataset cards, artifact descriptions and Reports

### Dataset Card
Screenshot 2024-12-03 at 6 51 10 PM

### Report
Screenshot 2024-12-03 at 6 48 05 PM

### Dataset Card
Screenshot 2024-12-03 at 6 50 19 PM

### Report
Screenshot 2024-12-03 at 6 48 46 PM

### DIFF
--- a/weave-js/src/common/util/markdown.ts
+++ b/weave-js/src/common/util/markdown.ts
@@ -57,7 +57,6 @@ export function buildSanitizationSchema(rules?: SanitizationRules) {
 }
 
 export function generateHTML(markdown: string, rules?: SanitizationRules) {
-  console.log('generateHTML', markdown);
   const sanitizationSchema = buildSanitizationSchema(rules);
   // IMPORTANT: We must sanitize as the final step of the pipeline to prevent XSS
   const vfile = (
@@ -80,13 +79,9 @@ export function generateHTML(markdown: string, rules?: SanitizationRules) {
     .processSync(markdown);
 
   if (typeof vfile.value === 'string') {
-    console.log('generateHTML before blankifyLinks', vfile.value);
     vfile.value = blankifyLinks(vfile.value);
-    console.log('generateHTML after blankifyLinks', vfile.value);
     vfile.value = shiftHeadings(vfile.value);
-    console.log('generateHTML after shiftHeadings', vfile.value);
   }
-  console.log('generateHTML', vfile.value);
   return vfile;
 }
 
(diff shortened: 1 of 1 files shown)


trane293:

weave-js/src/common/util/markdown.ts
```diff
@@ -20,7 +20,12 @@ import {blankifyLinks, shiftHeadings} from './html';
 
 // exported only for tests
 export const DEFAULT_SANITIZATION_SCHEMA = _.merge(gh, {
-  attributes: {'*': ['className', 'style']},
+  attributes: {
+    '*': ['className', 'style'],
+  },
+  protocols: {
+    src: ['http', 'https', 'data'],
```
ssisk, line 27: From a pure code perspective, this seems fine. However this definitely needs a security review - I’m not sure how secure it is to allow markdown to use arbitrary data (this is unfortunately not restricted to images). Could be fine, but definitely in the realm of “get a security expert to tell you” :slightly_smiling_face:.

weave-js/src/common/util/markdown.ts
```diff
@@ -20,7 +20,12 @@ import {blankifyLinks, shiftHeadings} from './html';
...
+  },
```
ssisk, line 28: This also has a pretty wide range of places it applies to - it’s used in the Markdown component that is used in 23 different places. One way we reduce the splash damage of this type of change is by requiring a caller to specify that we want to allow a specific change in security checks.

ibindlish: ➕

trane293: I had similar concerns about the XSS potential of this change. I tried testing using usual XSS scripts both in [HTML](https://owasp.org/www-community/attacks/xss/) and [Markdown](https://book.hacktricks.xyz/pentesting-web/xss-cross-site-scripting/xss-in-markdown), but couldn't definitively say that I'm confident this is safe. Given the fact this is used across the platform, the tests aren't there either. 

Let me see who can I talk to in order to get a better sense of the implications of this change.

PR #3153, opened 2024-12-05
chore(weave): capture request id for openai completions

bcsherma:
## Description

Openai integration captures the request id from completion responses and adds them to `request_id` on the call output.

### DIFF
--- a/tests/integrations/instructor/instructor_test.py
+++ b/tests/integrations/instructor/instructor_test.py
@@ -64,7 +64,7 @@ def test_instructor_openai(
     assert op_name_from_ref(call.op_name) == "openai.chat.completions.create"
     output = call.output
     output_arguments = json.loads(
-        output.choices[0].message.tool_calls[0].function.arguments
+        output["choices"][0]["message"]["tool_calls"][0]["function"]["arguments"]
     )
     assert "person_name" in output_arguments
     assert "age" in output_arguments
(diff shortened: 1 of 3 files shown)


bcsherma:
* **#3153** <a href="https://app.graphite.dev/github/pr/wandb/weave/3153?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/3153?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

tests/integrations/instructor/instructor_test.py
```diff
@@ -64,7 +64,7 @@ def test_instructor_openai(
     assert op_name_from_ref(call.op_name) == "openai.chat.completions.create"
     output = call.output
     output_arguments = json.loads(
-        output.choices[0].message.tool_calls[0].function.arguments
+        output["choices"][0]["message"]["tool_calls"][0]["function"]["arguments"]
```
andrewtruong, line 67: why do you have to change from dot to bracket notation?

bcsherma: The output here comes from the post processor I'm modifying. In practice `value` can have many types depending on integrations, sdk versions, and I think even the type of completion being requested. In some cases the post processor returns a pydantic object, in some cases it returns a dict. For the standard case where we receive a `ChatCompletion` we will now return a dict with this extra key. The alternative would be adding the attribute to the pydantic class which feels hackier, and ultimately I think the output just gets dumped later in our pipe anyway.

andrewtruong (approved):

PR #3158, opened 2024-12-05
fix(app): render bounding boxes on media panel without points

brianlund-wandb:
## Description
- Fixes [WB-16623](https://wandb.atlassian.net/browse/WB-16623)

When logging a point cloud via `wandb.Object3D` with an empty list of points we still end up including a single, empty point in the file we use to generate the panel.  
```
{
    "boxes":[
        {
            "color":[255, 0, 10],
            "corners":[ ... omitted for brevity ... ],
            "label":"Box"
        },
        {
            "color":[0, 255, 10],
            "corners":[... omitted for brevity...],
            "label":"Box-2"
        }
    ],
    "points":[
        []
    ],
    "type":"lidar/beta",
    "vectors":[]
}
```

The library we use to generate the panel, Babylon, fails to interpret that point, and generates a confusing panel.

image

The fix
* coerces that empty point to the position `[0, 0, 0]`
* ensures the camera focuses on the first bounding box (if it exists)

image

## Testing

Reproduction steps are included in the ticket.  The fix was tested locally, and includes a unit test for the coercion logic.

[WB-16623]: https://wandb.atlassian.net/browse/WB-16623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brianlund-wandb:

nhusher (approved):
this seems reasonable to me

PR #3170, opened 2024-12-06
docs(weave): Upload and document Service API notebook

J2-D2-3PO:
## Description

https://wandb.atlassian.net/browse/DOCS-1050

- Uploads [Service API notebook](https://colab.research.google.com/drive/18eDnFFiMQPnw5YGF40wa5G2tSwjE5itj?usp=sharing) created by @m-rgba 
- Documents the notebook for MD generation
- Adds generated MD as new guide
- Adjusts generated MD a bit for cosmetics

## Testing

- [ ] @J2-D2-3PO ran through the notebook to verify that code works / outputs for documentation
- [ ] @m-rgba verifies that the documented version is good to go

### DIFF
--- a/docs/docs/reference/gen_notebooks/weave_via_service_api.md
+++ b/docs/docs/reference/gen_notebooks/weave_via_service_api.md
@@ -1,0 +1,488 @@
+---
+title: Use the Service API to Log and Query Traces
+---
+
+
+:::note[This is a notebook]
+
+<a href="https://colab.research.google.com/github/wandb/weave/blob/master/docs/./notebooks/weave_via_service_api.ipynb" target="_blank" rel="noopener noreferrer" class="navbar__item navbar__link button button--secondary button--med margin-right--sm notebook-cta-button"><div><img src="https://upload.wikimedia.org/wikipedia/commons/archive/d/d0/20221103151430%21Google_Colaboratory_SVG_Logo.svg" alt="Open In Colab" height="20px" /><div>Open in Colab</div></div></a>
+
+<a href="https://github.com/wandb/weave/blob/master/docs/./notebooks/weave_via_service_api.ipynb" target="_blank" rel="noopener noreferrer" class="navbar__item navbar__link button button--secondary button--med margin-right--sm notebook-cta-button"><div><img src="https://upload.wikimedia.org/wikipedia/commons/9/91/Octicons-mark-github.svg" alt="View in Github" height="15px" /><div>View in Github</div></div></a>
+
+:::
+
+<!--- @wandbcode{service-api-colab} -->
+
+In the following notebook, you will learn how to use the Weave Service API to log traces. Specifically, you will use the Service API to:
+
+1. [Create a mock of a simple LLM call and response, and log it to Weave.](#simple-trace)
+2. [Create a mock of a more complex LLM call and response, and log it to Weave.](#complex-trace)
+3. [Run a sample lookup query on the logged traces.](#run-a-lookup-query)
+
+:::tip[View logged traces]
+You can view all of the Weave traces created when you run the code in this guide by going to the **Traces** tab in your Weave project (specified by `team_id\project_id`), and selecting the name of the trace.
+:::
+
+Before beginning, complete the [prerequisites](#prerequisites-set-variables-and-endpoints)
+
+## Prerequisites: Set variables and endpoints
+
+The following code sets the URL endpoints that will be used to access the Service API:
+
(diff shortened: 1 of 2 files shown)


J2-D2-3PO:
@wandb/weave-team help with a GH check requested please:

- _Issue_: [test/Python lint (push)](https://github.com/wandb/weave/actions/runs/12205030291/job/34051416625?pr=3170) failing on `ruff-format`
- _Fix attempt_: Ran `ruff check  notebooks/weave_via_service_api.ipynb --fix` locally and recommitted in [fe6fe58](https://github.com/wandb/weave/pull/3170/commits/fe6fe589bbae1d3e31826b602e8172eee0a26fb0)

m-rgba (approved):

J2-D2-3PO:
@wandb/weave-team, Node.js Tests is also failing as of my latest merge from master https://github.com/wandb/weave/pull/3170/commits/f5a61d9a48fed508b3bcf08f4ff9f623a454bda6. I don't think this has anything to do with the changes made in this PR

log:
https://github.com/wandb/weave/actions/runs/12242871380/job/34151142009?pr=3170

docs/docs/reference/gen_notebooks/weave_via_service_api.md
```diff
@@ -0,0 +1,488 @@
...
+
+<a href="https://colab.research.google.com/github/wandb/weave/blob/master/docs/./notebooks/weave_via_service_api.ipynb" target="_blank" rel="noopener noreferrer" class="navbar__item navbar__link button button--secondary button--med margin-right--sm notebook-cta-button"><div><img src="https://upload.wikimedia.org/wikipedia/commons/archive/d/d0/20221103151430%21Google_Colaboratory_SVG_Logo.svg" alt="Open In Colab" height="20px" /><div>Open in Colab</div></div></a>
+
+<a href="https://github.com/wandb/weave/blob/master/docs/./notebooks/weave_via_service_api.ipynb" target="_blank" rel="noopener noreferrer" class="navbar__item navbar__link button button--secondary button--med margin-right--sm notebook-cta-button"><div><img src="https://upload.wikimedia.org/wikipedia/commons/9/91/Octicons-mark-github.svg" alt="View in Github" height="15px" /><div>View in Github</div></div></a>
+
+:::
+
+<!--- @wandbcode{service-api-colab} -->
+
+In the following notebook, you will learn how to use the Weave Service API to log traces. Specifically, you will use the Service API to:
```
andrewtruong, line 16: We have a first-party `trace_server_interface` for this -- curious why you're using the raw HTTP endpoint instead of that?

J2-D2-3PO: The initial code was created by @m-rgba - maybe they can provide that context? I just documented the code as it was. @andrewtruong do you think it would be promoting an anti pattern by using the raw endpoint? If so, I can go back in and update the notebook to use trace_server_interface?

docs/docs/reference/gen_notebooks/weave_via_service_api.md
```diff
@@ -0,0 +1,488 @@
...
+# W&B variables
+team_id=""
+project_id=""
+wandb_token=""
+```
+
+## Simple trace
+The following sections walk you through creating a simple trace.
+
+1. [Start a complex trace](#start-a-complex-trace)
```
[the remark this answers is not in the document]
J2-D2-3PO: I did, fixed

docs/docs/reference/gen_notebooks/weave_via_service_api.md
m-rgba: Hey @andrewtruong - so when I initially created it:
- I made them while I was making a OpenWebUI plugin where I didn't want the user to have to install any packages.
- A customer then had asked for some examples using the REST API since they were having trouble with the client.

My argument for why I feel like this could be valuable:
- It's a good escape-hatch for anyone running where they can't use our client, or are having trouble with our client, but they want to send us traces.
- The service API is the most language neutral method for sending us traces.

If either of these are true, our current OpenAPI docs get the user like 95% there - but are pretty easy to flesh out some parts like - what's the shape of a completion we're expecting, or how should they format token counts.

100% open to this being an anti-pattern though, if we're going to be updating the API a lot or if there are other reasons from the product-adoption side. 
It also may be a both kind of thing where we can have examples for manually creating traces with the client as well.

andrewtruong: For python we should recommend the first-party stuff, and for other languages we can recommend this HTTP approach.

That said, I'm not against it.  We can still show python using the service API

andrewtruong (approved):

PR #3171, opened 2024-12-06
feat(weave): Add AzureOpenAI support for Scorers

tssweeney:


gtarpenning (approved):

andrewtruong:
is this sufficient?  Do we need a test in case the Azure version is different from regular openai?

tssweeney:
> is this sufficient? Do we need a test in case the Azure version is different from regular openai?

The Azure flavours are subclasses of OpenAI client, so i think it is literally just a drop in. I was hoping to get API keys to test

PR #3177, opened 2024-12-09
chore(weave): Add generic iterator for trace server API objects

andrewtruong:
Purely a refactor to make the `CallsIter` object more generic.  This will be used in a subsequent PR to implement something like "ObjectsIter"

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -10,7 +10,7 @@ import sys
 from collections.abc import Iterator, Sequence
 from concurrent.futures import Future
 from functools import lru_cache
-from typing import Any, Callable, Generic, Protocol, TypeVar, cast
+from typing import Any, Callable, Generic, Protocol, TypeVar, cast, overload
 
 import pydantic
 from requests import HTTPError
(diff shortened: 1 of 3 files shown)


weave/trace/weave_client.py
```diff
@@ -90,6 +90,118 @@
...
+
+
+class FetchFunc(Protocol[T]):
+    def __call__(self, offset: int, limit: int) -> list[T]: ...
+
+
+TransformFunc = Callable[[T], R]
+
+
+class PaginatedIterator(Generic[T, R]):
```
andrewtruong, line 104: This is basically a parameterized version of `CallsIter`

weave/trace/weave_client.py
```diff
@@ -90,6 +90,128 @@
...
+                project_id=project_id,
+                filter=filter,
+                offset=offset,
+                limit=limit,
+                include_costs=include_costs,
+            )
+        )
+        return response.calls
+
+    # TODO: Should be Call, not WeaveObject
```
andrewtruong, line 208: This may need to be done later.  Today, we wrap the "Call" in `WeaveObject` to give it nice "getting" properties, but the ideal state is to just have all of the functionality on `Call` itself.  That change is outside the scope of this PR.

tssweeney (approved):
Lovely refactor

PR #3189, opened 2024-12-10
fix(ui): Add start and stop timestamps to loader tracking

nhusher:
Useful for forwarding vitals to Datadog or other analytics providers.

### DIFF
--- a/weave-js/src/common/components/WandbLoader.tsx
+++ b/weave-js/src/common/components/WandbLoader.tsx
@@ -125,6 +125,8 @@ export const TrackedWaveLoader = ({
         const trackedData = {
           componentId: data.id,
           duration: data.duration,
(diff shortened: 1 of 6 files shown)


weave-js/src/common/components/WandbLoader.tsx
```diff
@@ -85,6 +85,8 @@ export const TrackedWandbLoader = ({
         const trackedData = {
           componentId: data.id,
           duration: data.duration,
+          start: data.start,
+          stop: data.stop,
```
nhusher, line 89: The concrete use case here is that I want to use Datadog RUM's vital metrics, which can either be given a start+stop or a start+duration. I could do this inside the wandb UI, but this seemed more correct

PR #3195, opened 2024-12-10
feat(weave): Add 'tracing_sample_rate' param to weave.op

adrnswanberg:
## Description

Introduces a new parameter, "tracing_sample_rate", to weave.op. This allows users to specify how frequently the op should actually be traced. 

The sample rate applies to the traced op *as well as all of its descendants*. If a root op is 'sampled out', then none of its descendant calls in that 'tree' will be traced either, regardless of the sampling rate set in those descendant ops.

## Testing

Tests were added.


weave/trace/op.py
```diff
@@ -407,16 +414,27 @@ def _do_call(
...
-        res = func(*pargs.args, **pargs.kwargs)
-    elif not op._tracing_enabled:
+    skip_tracing = (
+        settings.should_disable_weave()
+        or weave_client_context.get_weave_client() is None
+        or not op._tracing_enabled
+        or not get_tracing_enabled()
+    )
+
+    if skip_tracing:
```
andrewtruong, line 424: I try to avoid this.  The former is easier to debug which branch you're in if you're unexpectedly skipping tracing

andrewtruong:
We should update the docs for this feature too

tests/trace/test_client_trace.py
```diff
@@ -263,27 +263,27 @@ def simple_line_call_bootstrap(init_wandb: bool = False) -> OpCallSpec:
...
 
-    @weave.op()
+    @weave.op
     def adder(a: Number) -> Number:
         return Number(value=a.value + a.value)
 
     adder_v0 = adder
 
-    @weave.op()
+    @weave.op  # type: ignore
```
tssweeney, line 272: why is this the only one with type ignore?

tests/trace/test_client_trace.py
```diff
@@ -2997,3 +2998,192 @@ def foo():
...
+    def sometimes_traced(x: int) -> int:
+        nonlocal sometimes_traced_calls
+        sometimes_traced_calls += 1
+        return x + 1
+
+    # Never traced should execute but not be traced
+    for i in range(10):
+        never_traced(i)
+    assert never_traced_calls == 10  # Function was called
+    # NOTE: We can't assert here that never_traced.calls() is empty because that call requires
```
tssweeney, line 3030: you could just publish the op

weave/trace/op.py
```diff
@@ -407,37 +414,52 @@ def _do_call(
...
+    if not op._tracing_enabled:
+        res = func(*pargs.args, **pargs.kwargs)
+        return res, call
+    if not get_tracing_enabled():
         res = func(*pargs.args, **pargs.kwargs)
-    elif not op._tracing_enabled:
+        return res, call
+
+    current_call = call_context.get_current_call()
+    if current_call is None:
```
tssweeney, line 432: might be nice to say: if your parent is traced, then you are (as a comment)

tssweeney: Note: this specific case is not tested by your unit tests (setting a child sample rate to 0, and a parent to 1)

tests/trace/test_client_trace.py
tssweeney: well it would be nice to have a consistent assertion for # of calls. so anywhere you did not to .calls could have explicit publishing

tssweeney (approved):
Seems good. I would add the test to show how that once a parent call is traced, all children are traced.

weave/trace/op.py
adrnswanberg: Good point

adrnswanberg:
> Seems good. I would add the test to show how that once a parent call is traced, all children are traced.

Done

PR #3206, opened 2024-12-11
perf(docs): Disable GH checks on `docs` that don't apply to `docs`

J2-D2-3PO:
## Description

https://wandb.atlassian.net/browse/DOCS-1108

Proposes disabling GH action checks that don't apply to things that live in the `docs` folder. 

Reason:
1. Save 💰 and resources 
2. Speed up docs deployment / PR process. As of now, every docs PR has to pass 100+ checks, most of which don't apply to the docs themselves
3. Free up engineering from having to help docs team figure out some of the issues with tests randomly failing that don't apply anyways

## Testing

- [ ] Push changes to `docs` to test what happens on this branch
- [ ] Need to confirm with the team that all changes are ok first. Assuming that we only need to **keep the following tests for docs** (but if we can disable more then cheers )
   - `weave/.github/workflows/check-which-tests-to-run.yaml`
   -  `weave/.github/workflows/cla.yaml`
   - `weave/.github/workflows/deploy.yaml`
   - `weave/.github/workflows/pr.yaml`
   - `weave/.github/workflows/release.yaml`
   - `trace-docs` in `weave/.github/workflows/test.yaml`
       I added this to `trace-docs` to override the global docs disable on this test: `if: ${{ github.event_name == 'push' && (github.event.before == 'docs/**' || github.event.commits | contains('docs')) }}`. Alternatively, could break out into separate file, but need to check with team if this would cause other issues / more work

Help requested from @wandb/weave-team before proceeding further 
https://github.com/wandb/weave/pull/3206#issuecomment-2536698752

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -229,7 +229,7 @@ module = "weave_query.*"
 ignore_errors = true
 
 [tool.bumpversion]
-current_version = "0.51.2335"
+current_version = "0.51.2336-dev0"
 parse = """(?x)
     (?P<major>0|[1-9]\\d*)\\.
     (?P<minor>0|[1-9]\\d*)\\.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -229,7 +229,7 @@ module = "weave_query.*"
 ignore_errors = true
 
 [tool.bumpversion]
-current_version = "0.51.2313"
+current_version = "0.51.2314-dev0"
 parse = """(?x)
     (?P<major>0|[1-9]\\d*)\\.
     (?P<minor>0|[1-9]\\d*)\\.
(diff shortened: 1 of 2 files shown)


J2-D2-3PO:
@wandb/weave-team could I get someone to look at this before I proceed any further? 

I wanted to test the idea of disabling GH checks that don't apply to `docs` so that we could speed up the deployment process / not have to deal with changes and tests not related to docs work. 

For one, I'm not sure what's going on here.

image

J2-D2-3PO:
@tssweeney I'm not sure who to check with on the engineering team about this idea. Would this be ok to do? Would it cause other issues? If not, I think that this would be a useful process update

PR #3214, opened 2024-12-11
docs(weave): Remove api_key hardcodes in code examples

J2-D2-3PO:
Addresses https://wandb.atlassian.net/browse/DOCS-1120

Removes `api_key="..."` from examples like https://weave-docs.wandb.ai/tutorial-tracing_2 so that we aren't promoting bad practices. Leaves instances where the api_key is loaded from the environment e.g. `os.environ`

### DIFF
--- a/docs/docs/guides/integrations/notdiamond.md
+++ b/docs/docs/guides/integrations/notdiamond.md
@@ -68,7 +68,6 @@ preference_id = train_router(
     response_column="actual",
     language="en",
     maximize=True,
-    api_key=api_key,
 )
 ```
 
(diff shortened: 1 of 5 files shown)


docs/docs/guides/integrations/notdiamond.md
```diff
@@ -68,7 +68,6 @@ preference_id = train_router(
     response_column="actual",
     language="en",
     maximize=True,
-    api_key=api_key,
```
jamie-rasmussen, line 71: This isn't the OpenAI constructor like the others, but it does appear to be an optional parameter.

jamie-rasmussen (approved):

docs/docs/guides/integrations/notdiamond.md
J2-D2-3PO: Agreed, that's why I removed it - seemed like it would be promoting the antipattern

PR #3197, opened 2024-12-11
feat(weave): Support op configuration for autopatched functions (starting with OpenAI)

andrewtruong:
This PR adds a new kwarg to `weave.init`, called `autopatch_settings`.

Normally, we provide sane defaults to autopatched functions.  However, some users may want to customize how that patching happens, e.g.:
1. Configuring a custom `postprocess_inputs` to strip out PII
2. Configuring the names of autopatched functions
3. Disabling patching altogether

Now, users can configure this easily at init time before patching:
```py
client = weave.init(
    ...,
    autopatch_settings={
        "openai": {
            "op_settings": {
                "postprocess_inputs": ...,
                "postprocess_output": ...,
            }
        },
        "anthropic": {
            "enabled": False,
        }
    },
)

```

---

This PR provides the core functionality and implements OpenAI to show what integration changes would look like.  A second fast-follow PR implements the remaining integrations: https://github.com/wandb/weave/pull/3154

Resolves:
- https://wandb.atlassian.net/browse/WB-21231
- https://wandb.atlassian.net/browse/WB-21752

PR #3216, opened 2024-12-12
feat(weave): Support op configuration for autopatched functions for remaining integrations

andrewtruong:
This is a fast-follow PR to: https://github.com/wandb/weave/pull/3197

It implements op kwarg passthrough for the remaining integrations, except Langchain and Llamaindex because they don't use the op decorator.

### DIFF
--- a/weave/integrations/anthropic/anthropic_sdk.py
+++ b/weave/integrations/anthropic/anthropic_sdk.py
@@ -1,27 +1,26 @@
+from __future__ import annotations
+
 import importlib
 from collections.abc import AsyncIterator, Iterator
 from functools import wraps
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Optional,
-    Union,
-)
+from typing import TYPE_CHECKING, Any, Callable
 
 import weave
+from weave.trace.autopatch import IntegrationSettings, OpSettings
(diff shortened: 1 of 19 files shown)


weave/integrations/anthropic/anthropic_sdk.py
```diff
@@ -187,28 +180,58 @@ def wrapper(fn: Callable) -> Callable:
...
-            lambda: importlib.import_module("anthropic.resources.messages"),
-            "AsyncMessages.stream",
-            create_stream_wrapper(name="anthropic.AsyncMessages.stream"),
-        ),
-    ]
-)
+def get_anthropic_patcher(
+    settings: IntegrationSettings | None = None,
+) -> MultiPatcher | NoOpPatcher:
+    if settings is None:
```
tssweeney, line 186: It seems like all these patchers have the same header pattern of:
```
    if settings is None:
        settings = IntegrationSettings()

    if not settings.enabled:
        return NoOpPatcher()

    global _anthropic_patcher
    if _anthropic_patcher is not None:
        return _anthropic_patcher
```

I wonder if this can be packaged into a common util so that integration authors do not need to remember to copy/paste this

tssweeney (approved):
Approving, but i do think you could make a few abstractions that would make the code cleaner and more maintainable.

andrewtruong:
merging this to get the Nvidia stuff in, then will revisit

PR #3217, opened 2024-12-12
chore(ui): Playground: make editor component resizeable

m-rgba:
## Issue: 
Editor is currently setting a max-height when auto-expanding. 
If we unlock max-height the editing view could be too long, pushing away the Cancel and Save controls.

## Solution:
- Added a resizeable condition to the editor input component (disable if autogrow is enabled).
- Gave it a larger starting height.
- Disabled `setIsShowingMore` for max-height on edit so the user can expand the input.

### Note:
Could we keep passing that generated height of the message content into the editor component?
- I think the perfect solution would be to un-cap the editor and have it auto-grow - we'd need some form of sticky action bar for the editor though to solve the issue addressed above.
- This PR will help fix the immediate usability issue though. 

#### Before
CleanShot 2024-12-12 at 11 52 45@2x

#### After
CleanShot 2024-12-12 at 11 51 59@2x

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanel.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanel.tsx
@@ -47,6 +47,13 @@ export const MessagePanel = ({
     }
   }, [message.content, contentRef?.current?.scrollHeight]);
 
+  // Set isShowingMore to true when editor is opened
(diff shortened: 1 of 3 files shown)


m-rgba:
Updated FeatureBee: https://beta.wandb.ai/?betaVersion=4bcca949107e3e1d31ef4cef5e68eebf075ee049

jwlee64 (approved):

PR #3218, opened 2024-12-12
chore(ui): Playground: add copy button, tooltip on toggles

m-rgba:
## Description
- Added a tooltip for model toggles in Playground settings (to help with readability for longer model names).
- Added copy button functionality to the messages in the stack.
~~Updated icons and costing info to use `TraceCostStats`.~~

| Tooltip  |  Copy Button | ~~Trace Costing + Icons~~ |
| --- | --- | --- |
| CleanShot 2024-12-12 at 13 39 51@2x | CleanShot 2024-12-12 at 13 40 08@2x | CleanShot 2024-12-12 at 13 41 04@2x |

Trace costs updated in:
https://github.com/wandb/weave/pull/3227


m-rgba:
Updated FeatureBee: https://beta.wandb.ai/?betaVersion=fb49a2b45e9c9bc92f9b02e6d74b08560e36809c

jwlee64 (commented):
can we split out the cost fetch I think we fetch the call higher in the tree and can pass that down rather than fetching it here, then we might not have to do the traceCallSchema-> CallSchema, to fetching the costs

jwlee64 (approved):
approving, with the caveat of moving the cost fetching

m-rgba:
Removed the cost fetching elements from the PR

PR #3231, opened 2024-12-13
fix(weave): optimize finding next tick up or down

brianlund-wandb:
## Description

- Fixes [WB-22089](https://wandb.atlassian.net/browse/WB-22089)
- Possibly fixes [WB-9411](https://wandb.atlassian.net/browse/WB-9411), need adequate test case

Fixes `getNearestTick()` in `SliderInput.tsx` by including the existing value of the slider to to determine whether to look above or below the previous value.
Optimized function for better performance over large lists.
Included a new test file for basic unit tests of the function.
Refactored logic from event handler (`update()`) function into `getNearestTick()` to consolidate logic.

## Testing

Added new unit tests to verify basic functionality.
Created a project with varying steps to ensure slider increments appropriately.

[WB-22089]: https://wandb.atlassian.net/browse/WB-22089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-9411]: https://wandb.atlassian.net/browse/WB-9411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave-js/src/common/components/elements/SliderInput.tsx
```diff
@@ -172,17 +172,38 @@ const SliderInput: React.FC<SliderInputProps> = React.memo(
...
     const diff = Math.abs(tick - val);
-    if (diff >= minDiff) {
-      break;
+
+    // Only update closest if the tick is in the right direction
+    if (
+      diff < minDiff &&
+      ((increasing && tick >= val) || (!increasing && tick <= val))
+    ) {
+      closest = tick;
```
peytonsarmiento, line 198: could we have something to break out early if tick is equal to val?

weave-js/src/common/components/elements/SliderInput.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    const previous = 500000;
+    const val = 500001;
+    const start = Date.now();
+    const actual = getClosestTick(ticks, val, previous);
+    const end = Date.now();
+    const duration = end - start;
+    expect(actual).toBe(500002);
+    expect(duration).toBeLessThanOrEqual(1);
```
peytonsarmiento, line 66: these tests are awesome! could you add one for testing the bounds as well? like one where the val is less than the smallest tick and also one where the val is larger than the largest tick?

weave-js/src/common/components/elements/SliderInput.tsx
brianlund-wandb: Yeah, lets not go searching through half the list if the slider increments to a valid step, heh.  Good call.

weave-js/src/common/components/elements/SliderInput.test.tsx
brianlund-wandb: Oh I like it, sure.  That logic lives in the `update()` function that calls `getClosestTick()` when it has a list of ticks. I might suck that logic into this function instead, and leave `update()` as a, "get the right value and synchronize the inputs fields."

brianlund-wandb:
@peytonsarmiento should be ready for another look, but I want to highlight a couple of changes:
* **removed unused `allowGreaterThanMax` prop** - couldn't find any usage, implementing ticket, [WB-5666](https://wandb.atlassian.net/browse/WB-5666), appears outdated. I don't want to leave dead code lying around, but I would greatly appreciate any ideas you have to verify that property is indeed unused.
* **potential performance cost to `ticks.includes()`** - Hard to tell if the explicit value check (i.e. new value is in the list of ticks) buys us anything unless we can utilize the `stepMeta.stepSet` instead of `stepMeta.steps`.  In this context, a set seems an appropriate data structure (ordered, unique), but I wondered if there are considerations with changing the input types that I'm overlooking.  At this point I left it as a list, but am happy to make the change.

brianlund-wandb:
Per my last comment about the performance of `ticks.includes()`, the test failure is pertinent.

I updated the test to need more iterations through the search hoping to get my local machine to exceed that 1 millisecond threshold, and it seems that check costs ~0.5 milliseconds for the 5 million ticks.
Swapping from a list to a set probably isn't worth the effort, since we should be able to get through this list in less than 0.1 milliseconds.

peytonsarmiento (approved):
This looks good! I would just double check that the Smoothing Input isn't changed from this - https://github.com/wandb/core/pull/7093/files

brianlund-wandb:
@peytonsarmiento thanks for calling out that smoothing feature.

Dug into the details with @KyleGoyette and realized that slider is one of our `SettingSlider` components (`LinePlotSmoothing` -> `SmoothingSetting` -> `SettingSlider`), not the `SliderInput` I'm changing here. Whether it _should_ use this component feels like a different question. At this point, I'm going to assume that it's not worth filing a ticket to refactor the smoothing component to use this one, but I see potential for refactoring the smoothing component to reuse the `SliderInput`.

PR #3253, opened 2024-12-16
docs(weave): Don't use scary warnings if we don't need to

J2-D2-3PO:
https://wandb.atlassian.net/browse/DOCS-1107

Replaces `:::warning` callouts with `:::important` where we don't need to scare the reader 👻

### DIFF
--- a/docs/docs/guides/evaluation/scorers.md
+++ b/docs/docs/guides/evaluation/scorers.md
@@ -226,7 +226,7 @@ In Weave, Scorers are used to evaluate AI outputs and return evaluation metrics.
     ### Mapping Column Names with `columnMapping`
     :::important
 
-    In TypeScript, this feature is currently on the `Evaluation` object, not individual scorers!
+    In TypeScript, this feature is currently on the `Evaluation` object, not individual scorers.
 
     :::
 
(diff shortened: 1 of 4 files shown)


docs/docs/guides/evaluation/scorers.md
```diff
@@ -224,7 +224,7 @@ In Weave, Scorers are used to evaluate AI outputs and return evaluation metrics.
     ```
 
     ### Mapping Column Names with `columnMapping`
-    :::warning
+    :::important
 
     In TypeScript, this feature is currently on the `Evaluation` object, not individual scorers!
```
jamie-rasmussen, line 229: Do you want to remove this exclamation point too?

jamie-rasmussen (approved):

docs/docs/guides/evaluation/scorers.md
J2-D2-3PO, line 229:
```suggestion
    In TypeScript, this feature is currently on the `Evaluation` object, not individual scorers.
```

J2-D2-3PO: Good call, removed in https://github.com/wandb/weave/pull/3253/commits/ddb8af965d4b410474856e6d75c2be8c9abe257f

PR #3263, opened 2024-12-16
style(weave): Hide function drawer textarea draggable

jwlee64:
## Description

After
Screenshot 2024-12-16 at 2 54 45 PM

Before
Screenshot 2024-12-16 at 2 54 19 PM

## Testing

How was this PR tested?


gtarpenning (approved):
Why dont we want resize?

jwlee64:
you can end up in some weird states, minimizes potential user error

jwlee64:
you can end up in some weird states, minimizes potential user error

PR #3268, opened 2024-12-17
feat(weave): delete objects and ops from the sdk

gtarpenning:
## Description

[WB-22374](https://wandb.atlassian.net/browse/WB-22374)

Allow deletion of ops and objects by their refs in the sdk. 

Example: 

```python

ref = weave.save({"name": "new-obj"}, "my first object")

# now delete
ref.delete()

```

## Testing

Tests!

[WB-22374]: https://wandb.atlassian.net/browse/WB-22374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/trace/test_weave_client.py
```diff
@@ -1544,3 +1544,98 @@ def func():
...
+
+    # create 3 versions of the object
+    obj["a"] = 6
+    weave_obj2 = client.save(obj, "my-obj")
+    obj["a"] = 7
+    weave_obj3 = client.save(obj, "my-obj")
+    obj["a"] = 8
+    weave_obj4 = client.save(obj, "my-obj")
+
+    # delete weave_obj3 with class method
```
andrewtruong, line 1567: technically just a method

tests/trace/test_weave_client.py
```diff
@@ -1544,3 +1544,98 @@ def func():
...
+    weave_obj = client.save(obj, "my-obj")
+    assert client.get(weave_obj.ref) == obj
+
+    client.delete_object_version(weave_obj.ref)
+    with pytest.raises(weave.trace_server.errors.ObjectDeletedError):
+        client.get(weave_obj.ref)
+
+    # create 3 versions of the object
+    obj["a"] = 6
+    weave_obj2 = client.save(obj, "my-obj")
```
andrewtruong, line 1561: Given this API, you'd think the symmetric API here would be `obj.save()`

tests/trace/test_weave_client.py
```diff
@@ -1544,3 +1544,98 @@ def func():
...
+            project_id=client._project_id(),
+            names=["my-obj"],
+            sort_by=[tsi.SortBy(field="created_at", direction="desc")],
+        )
+    )
+    assert len(versions.objs) == 2
+
+    # iterate over the versions, confirm the indexes are correct
+    assert versions.objs[0].version_index == 3
+    assert versions.objs[1].version_index == 1
```
andrewtruong, line 1590: will this order always be guaranteed?  If not, it may be more lenient to test that the set of version indices includes (1,3)

tests/trace/test_weave_client.py
```diff
@@ -1544,3 +1544,98 @@ def func():
...
+    obj3 = {"c": obj2_ref}
+    obj3_ref = client.save(obj3, "obj3").ref
+
+    # Delete obj1
+    client.get(obj1_ref).delete()
+
+    # Make sure we can't get obj1
+    with pytest.raises(weave.trace_server.errors.ObjectDeletedError):
+        out = client.get(obj1_ref)
+        print(out)
```
andrewtruong, line 1621: remove

tests/trace/test_weave_client.py
```diff
@@ -1544,3 +1544,98 @@ def func():
...
+
+    # Make sure we can get obj2, but the ref to object 1 should return None
+    assert client.get(obj2_ref) == {"b": None}
```
andrewtruong, line 1624: Is this what you expect?  I feel at minimum this should return a warning that the ref was deleted.  It might be safer to set this as a sentinel "DELETED"

weave/trace/vals.py
```diff
@@ -636,15 +637,21 @@ def make_trace_obj(
...
+                )
             )
-        )
-        val = from_json(read_res.obj.val, val.entity + "/" + val.project, server)
-        prepare_obj(val)
+            val = from_json(read_res.obj.val, val.entity + "/" + val.project, server)
+            prepare_obj(val)
+        except ObjectDeletedError:
+            # Catch error case where an object has been deleted. Val here is likely
+            # part of a nested object, return None to indicate a deleted object.
```
andrewtruong, line 652: I think it might be safer to have a `DELETED_OBJECT` sentinel.  That makes it very clear to the user that an object used to exist here, but now it doesn't.

The case where `None` is bad is where you have a signature like:
```
class Thing:
    maybe_object: Object | None
```

If you then delete maybe_object, it's ambiguous if `None` means "unset" or "deleted"

tests/trace/test_weave_client.py
gtarpenning: Yeah this order has to be the same, always want to get the versions out in order.

andrewtruong (approved):

PR #3269, opened 2024-12-17
feat(weave_query): create api client for trace server

domphan-wandb:
## Description

[WB-21928](https://wandb.atlassian.net/browse/WB-21928)

This PR creates a client api, `wandb_trace_server_api.py`, to make http requests to the weave trace server.

## Testing

How was this PR tested?

Manually through local development

[WB-21928]: https://wandb.atlassian.net/browse/WB-21928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave_query/weave_query/io_service.py
+++ b/weave_query/weave_query/io_service.py
@@ -35,6 +35,7 @@ from weave_query import (
     uris,
     wandb_api,
     wandb_file_manager,
+    wandb_trace_server_api
 )
 
 tracer = engine_trace.tracer()  # type: ignore
@@ -217,6 +218,7 @@ class Server:
         self.register_handler_fn("ensure_file", self.handle_ensure_file)
         self.register_handler_fn("direct_url", self.handle_direct_url)
         self.register_handler_fn("sleep", self.handle_sleep)
+        self.register_handler_fn("query_traces", self.handle_query_traces)
 
         if process:
             self.request_handler = aioprocessing.AioProcess(
@@ -343,6 +345,7 @@ class Server:
             self.wandb_file_manager = wandb_file_manager.WandbFileManagerAsync(
                 fs, net, await wandb_api.get_wandb_api()
             )
+            self.wandb_trace_server_api = wandb_trace_server_api.WandbTraceApiAsync(net)
 
             self._request_handler_ready_event.set()
             while True:
@@ -417,6 +420,24 @@ class Server:
         ):
             raise errors.WeaveInternalError("invalid scheme ", uri)
         return await self.wandb_file_manager.ensure_file(uri)
+    
+    async def handle_query_traces(
+        self,
+        project_id: str,
+        filter: typing.Optional[dict] = None,
+        limit: typing.Optional[int] = None,
+        offset: typing.Optional[int] = None,
+        sort_by: typing.Optional[list] = None,
+        query: typing.Optional[dict] = None,
+    ) -> list[dict]:
+        return await self.wandb_trace_server_api.query_calls_stream(
+            project_id,
+            filter=filter,
+            limit=limit,
+            offset=offset,
+            sort_by=sort_by,
+            query=query,
+        )
 
     async def handle_ensure_file_downloaded(
         self, download_url: str
@@ -481,12 +502,24 @@ class AsyncConnection:
 
     async def close(self) -> None:
         self.connected = False
+        sentinel = ServerResponse(http_error_code=200, error=False, client_id=self.client_id, id=-1, value=None)
+        await self.response_queue.async_put(sentinel)
+        await self.response_task
 
     async def handle_responses(self) -> None:
-        while self.connected:
-            resp = await self.response_queue.async_get()
-            self.response_queue.task_done()
-            self.requests[resp.id].set_result(resp)
+        try:
+            while self.connected:
+                resp = await self.response_queue.async_get()
+
+                if resp.id == -1:
+                    break
+
+                self.response_queue.task_done()
+                self.requests[resp.id].set_result(resp)
+        finally:
+            for future in self.requests.values():
+                if not future.done():
+                    future.cancel()
 
     async def request(self, name: str, *args: typing.Any) -> typing.Any:
         # Caller must check ServerResponse.error!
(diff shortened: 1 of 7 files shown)


weave_query/weave_query/io_service.py
```diff
@@ -481,12 +502,26 @@ def response_task_ended(self, task: asyncio.Task) -> None:
...
+        await self.response_queue.async_put(sentinel_response)
+        await self.response_task
 
     async def handle_responses(self) -> None:
-        while self.connected:
-            resp = await self.response_queue.async_get()
-            self.response_queue.task_done()
-            self.requests[resp.id].set_result(resp)
+        try:
+            while self.connected:
```
tssweeney, line 511: this is a nice change, but i would break into 2 PRs

weave_query/weave_query/weave_http.py
```diff
@@ -128,6 +129,29 @@ async def download_file(
...
+        cookies: typing.Optional[dict[str, str]] = None,
+        auth: typing.Optional[aiohttp.BasicAuth] = None,
+    ) -> list[dict]:
+        with tracer.trace("query_traces"):
+            results = []
+            async with self.session.post(url, json=payload, headers=headers, cookies=cookies, auth=auth) as response:
+                if response.status == 200:
+                    async for line in response.content:
+                        if line:
+                            decoded_line = line.decode('utf-8').strip()
```
tssweeney, line 146: subtly here: not all trace server endpoints return jsonl. therefore, the pass in of URL feels a little odd here. It is possible for a future dev to pass in another url for a different endpoint that would result in breakage.

tssweeney (commented):
How are you thinking about testing this? It would be nice to have a basic integration test to validate the query

domphan-wandb:
> How are you thinking about testing this? It would be nice to have a basic integration test to validate the query

Thanks for reviewing this @tssweeney ! I'll get the connection closing logic split out another PR. Do you know if there's an example of an existing integration test validating a query that I could refer to?

tssweeney:
> > How are you thinking about testing this? It would be nice to have a basic integration test to validate the query
> 
> Thanks for reviewing this @tssweeney ! I'll get the connection closing logic split out another PR. Do you know if there's an example of an existing integration test validating a query that I could refer to?

This is the first integration between the services of this nature, so there is not an existing example. You would want to have your test harness start a trace server and clickhouse instance to ensure the connection works. You could also mock the service and assert on the expected shape.

weave_query/weave_query/io_service.py
```diff
@@ -481,12 +502,26 @@ def response_task_ended(self, task: asyncio.Task) -> None:
 
     async def close(self) -> None:
         self.connected = False
+        sentinel_response = ServerResponse(http_error_code=200, error=False, client_id=self.client_id, id=SENTINEL_ID, value=None)
+        await self.response_queue.async_put(sentinel_response)
+        await self.response_task
```
dannygoldstein, line 507: AFAIK this is not used anywhere right now, all of our clients are sync. Is there a motivation to switch to AsyncConnections?

weave_query/weave_query/ops_domain/project_ops.py
```diff
@@ -259,3 +262,32 @@ def artifacts(
...
+        task = loop.create_task(conn.query_traces(
+            project_id,
+            filter=filter,
+            limit=limit,
+            offset=offset,
+            sort_by=sort_by,
+            query=query))
+        tasks.add(task)
+        await asyncio.wait(tasks)
+        return task.result()
```
dannygoldstein, line 293: is this weakreffing needed when dealing with only 1 task, and when the task is being awaited (instead of fired and forgotten)?

weave_query/weave_query/io_service.py
```diff
@@ -217,6 +218,7 @@ def __init__(
         self.register_handler_fn("ensure_file", self.handle_ensure_file)
         self.register_handler_fn("direct_url", self.handle_direct_url)
         self.register_handler_fn("sleep", self.handle_sleep)
+        self.register_handler_fn("query_traces", self.handle_query_traces)
```
dannygoldstein, line 221: while it was the intention for all IO to go through this service, in practice not all of the application actually abides by that, for example the graphql client executes queries directly from request handlers in the server. right now the IO service is used mainly for handling files that are present either the local fs or a remote fs. HTTP requests for REST apis generally are made from the server directly, and dont go through the IO service.

was there a specific motivation for putting this in the io service? as a first step, it may be simpler / easier to test to call out to the trace server directly with `requests` in the handler code

weave_query/weave_query/io_service.py
domphan-wandb: I saw this being used here: https://github.com/wandb/weave/blob/6a093a4f2d8088e0d5e02c6ca2a5688cc017ac0d/weave_query/weave_query/ops_domain/table.py#L773

I didn't see an example of op code making http requests, so I tried to implement something similar based on that use of io_service.

weave_query/weave_query/ops_domain/project_ops.py
domphan-wandb: I set up the weakreffing because `.wait` expected an iterable

weave_query/weave_query/io_service.py
domphan-wandb: thank you for the background, Danny, super helpful. This is my first time working with python async code (I didn't know python had async / await until a few weeks ago), so I thought I would try to follow what's being done for concurrency reasons. 

let me tweak these set of changes so the request doesn't go through the IO service and follow your suggestion 👍

weave_query/weave_query/wandb_trace_server_api.py
```diff
@@ -0,0 +1,74 @@
...
+from requests.auth import HTTPBasicAuth
+
+from weave_query import environment as weave_env
+from weave_query import engine_trace, server_error_handling
+
+from weave_query.context_state import WandbApiContext, _wandb_api_context
+
+tracer = engine_trace.tracer()  # type: ignore
+
+def get_wandb_api_context() -> typing.Optional[WandbApiContext]:
```
tssweeney, line 15: i'd underscore this

tssweeney (approved):
Hot

PR #3270, opened 2024-12-17
feat(weave_query): Create ops to fetch traces and traces types

domphan-wandb:
## Description

- [WB-21929 - Create project-traces op](https://wandb.atlassian.net/browse/WB-21929)
- [WB-21930 - Create project-tracesType op](https://wandb.atlassian.net/browse/WB-21930)

This PR adds two ops to fetch weave traces at the project level from a query panel.

In this first pass, the ops are meant to be used by copy and pasting the curl export of the traces calls into the op. The refinement currently makes the same call to the trace server to get types. Our team is okay with this perf issue in the mean time since this is a POC, but we will follow up with an improvement.

Also wanted to call out that feedback and costs are not currently implemented, but we'll figure out how to display it in a follow up.

https://github.com/user-attachments/assets/8db5e120-c578-4101-8336-e1e89910dea0

Traces in Weave:
image
Traces in Query Panel:
image

## Testing

How was this PR tested?
Manually through development environment.

### DIFF
--- a/weave-js/src/core/ops/domain/util.ts
+++ b/weave-js/src/core/ops/domain/util.ts
@@ -1,3 +1,5 @@
+import {dict, list, typedDict, union} from '../../model';
+
 /**
  * `connectionToNodes` is a helper function that converts a `connection` type
  * returned from gql to its list of nodes. In many, many location in our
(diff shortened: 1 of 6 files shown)


weave-js/src/core/ops/domain/util.ts
```diff
@@ -41,3 +43,27 @@ export const connectionToNodes = <T>(connection: MaybeConnection<T>): T[] =>
   (connection?.edges ?? [])
     .map(edge => edge?.node)
     .filter(node => node != null) as T[];
+
+const traceFilterPropertyTypes = {
```
tssweeney, line 47: i would move this to a new file. This file is more about gql typescript typing

weave_query/weave_query/ops_domain/project_ops.py
```diff
@@ -291,3 +291,68 @@ async def _get_project_traces(project, payload):
...
+    res = asyncio.run(_get_project_traces(project, payload))
+    if res:
+        return types.TypeRegistry.type_of(res)
+    else:
+        return types.TypeRegistry.type_of([])
+
+@op(
+    name="project-traces",
+    input_type=traces_input_types,
+    output_type=ops_arrow.ArrowWeaveListType(types.TypedDict({})),
```
tssweeney, line 342: optional: i think `output_type` could be even further narrowed since many of the trace fields are known up from

weave_query/weave_query/ops_domain/project_ops.py
```diff
@@ -291,3 +291,68 @@ async def _get_project_traces(project, payload):
...
+            entity {
+                id
+                name
+            }
+        """
+    ),
+    hidden=True
+)
+def traces_type(project, payload):
+    res = asyncio.run(_get_project_traces(project, payload))
```
tssweeney, line 333: future: would be nice for this to be cached somehow

tssweeney (approved):
Nice job!

PR #3279, opened 2024-12-18
feat(weave): Add mods page and menu item only for wandb admins

vanpelt:
## Description

This adds a new Mods page that we only expose in the overflow menu for wandb admins.  It depends on a number of changes in core.  I'll be breaking those changes out into seperate PR's with graphite and will link any dependent PR's here.

## Testing

How was this PR tested?


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ModsPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+type ModCategoryType = 'Labeling' | 'Analysis' | 'Demos';
+
+type ModCategories = {
+  [key in ModCategoryType]: Mod[];
+};
+
+const mods: ModCategories = {
+  Labeling: [
+    {
+      id: 'labeling/html',
```
tssweeney, line 36: seems odd to have a slash in the id

tssweeney (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ModsPage.tsx
vanpelt: Good eye, it is indeed odd but I did this on purpose although we may want to just not allow it.  The thinking was all of these mods will be in a git repo somewhere and organizing them in folders would be ideal.  I think I can get around this by having the id to directory mapping stored somewhere else...

PR #3282, opened 2024-12-18
chore(weave): Better error handling in sender thread

tssweeney:
Currently we re-raise errors in the call batch processor thread. This is pretty bad since it kills the background thread and results in no additional calls being logged.

### DIFF
--- a/weave/trace_server/async_batch_processor.py
+++ b/weave/trace_server/async_batch_processor.py
@@ -6,7 +6,6 @@ from threading import Event, Lock, Thread
 from typing import Callable, Generic, TypeVar
 
 from weave.trace.context.tests_context import get_raise_on_captured_errors
-from weave.trace_server import requests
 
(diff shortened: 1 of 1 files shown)


weave/trace_server/async_batch_processor.py
```diff
@@ -61,14 +60,10 @@ def _process_batches(self) -> None:
             if current_batch:
                 try:
                     self.processor_fn(current_batch)
-                except requests.HTTPError as e:
-                    if e.response.status_code == 413:
-                        # 413: payload too large, don't raise just log
-                        if get_raise_on_captured_errors():
-                            raise
-                        logger.exception(f"Error processing batch: {e}")
-                    else:
```
tssweeney, line 70: Currently, this will crash the processor thread, which stops future sends

gtarpenning (approved):

PR #3284, opened 2024-12-18
fix(app): improve rendering code preview on large files

onx2:
## Description

- Fixes WB-22387

This PR memoizes the code preview component, isolates the scroll to only the preview, and removes highlighting when the file gets large enough. This isn't a perfect solution but at least before the code freeze the page becomes usable.

## Testing

Tested on open ai large files >8MB

### DIFF
--- a/weave-js/src/common/components/FileBrowser.tsx
+++ b/weave-js/src/common/components/FileBrowser.tsx
@@ -681,22 +681,25 @@ const CodePreview: FC<CodePreviewProps> = memo(
           padding: 16,
           height: 'calc(100vh - 150px)',
           overflowY: 'auto',
+          maxWidth: '100%',
         }}>
-        <pre
-          style={{
-            maxWidth: '100%',
-          }}>
-          <code
-            style={{whiteSpace: 'pre-wrap', wordBreak: 'break-all'}}
-            ref={ref}
-            className={
-              language && isSmallFile != null
-                ? `language-${language}`
-                : undefined
-            }>
+        {isSmallFile ? (
+          <pre
+            style={{
+              maxWidth: '100%',
+            }}>
+            <code
+              style={{whiteSpace: 'pre-wrap', wordBreak: 'break-all'}}
+              ref={ref}
+              className={language != null ? `language-${language}` : undefined}>
+              {data}
+            </code>
+          </pre>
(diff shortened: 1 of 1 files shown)


weave-js/src/common/components/FileBrowser.tsx
```diff
@@ -608,103 +608,94 @@ interface CodePreviewProps {
...
-  useEffect(() => {
-    if (ref.current != null) {
-      Prism.highlightElement(ref.current);
+    // We don't pass a fallback to allow dev mode zero byte files to render
+    const loading = useLoadFile(file, {
+      onSuccess: setData,
+      onFailure: setError,
+    });
+    useEffect(() => {
+      if (ref.current != null && file.sizeBytes / 1024 < 1024) {
```
onx2, line 661: ignore highlighting for files larger than 1MB. This can probably be bumped up if we dig into the highlighting solution more

weave-js/src/common/components/FileBrowser.tsx
```diff
@@ -608,103 +608,94 @@ interface CodePreviewProps {
...
+    }
+
+    return (
+      <div
         style={{
-          maxWidth: '100%',
+          background: 'white',
+          border: '1px solid #eee',
+          padding: 16,
+          height: 'calc(100vh - 150px)',
```
onx2, line 678: the page craps out when we use the normal scroll bar in large files so having this preview scroll on its own prevents that

weave-js/src/common/components/FileBrowser.tsx
```diff
@@ -608,103 +608,94 @@ interface CodePreviewProps {
...
+            style={{whiteSpace: 'pre-wrap', wordBreak: 'break-all'}}
+            ref={ref}
+            className={language != null ? `language-${language}` : undefined}>
+            {data}
+          </code>
+        </pre>
+      </div>
+    );
+  },
+  (prevProps, nextProps) => {
```
onx2, line 695: simple memo by file url

weave-js/src/common/components/FileBrowser.tsx
jamie-rasmussen: This might be better as an in-code comment

PR #3286, opened 2024-12-18
feat(weave): run it n times for o1 models

jwlee64:
## Description

For o1 models
run it n times
stitch choices and usage together
Should be temporary until openai adds choices to o1 models

Screenshot 2024-12-18 at 10 23 02 AM

## Testing

How was this PR tested?

### DIFF
--- a/weave/trace_server/llm_completion.py
+++ b/weave/trace_server/llm_completion.py
@@ -32,19 +32,53 @@ def lite_llm_completion(
 
     # This allows us to drop params that are not supported by the LLM provider
     litellm.drop_params = True
+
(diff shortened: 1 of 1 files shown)


weave/trace_server/llm_completion.py
```diff
@@ -32,20 +32,54 @@ def lite_llm_completion(
 
     # This allows us to drop params that are not supported by the LLM provider
     litellm.drop_params = True
+
+    if "o1" not in inputs.model or inputs.n == 1:
```
tssweeney, line 36: is "o1" in sufficient? I would pull this into it's own helper

tssweeney (approved):

PR #3287, opened 2024-12-18
fix(weave): Unbreak editing and deleting choices

jwlee64:
## Description

on any state change we would reload the intial trace

deleting choices was broken
if you deleted all the choices it would break as well

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundPage.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundPage.tsx
@@ -108,11 +108,9 @@ export const PlaygroundPageInner = (props: PlaygroundPageProps) => {
         setPlaygroundStateFromTraceCall(callWithCosts.result.traceCall);
       }
     }
-  }, [
-    callWithCosts.loading,
-    setPlaygroundStateFromTraceCall,
-    callWithCosts.result,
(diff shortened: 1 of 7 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundPage.tsx
```diff
@@ -108,11 +108,9 @@ export const PlaygroundPageInner = (props: PlaygroundPageProps) => {
         setPlaygroundStateFromTraceCall(callWithCosts.result.traceCall);
       }
     }
-  }, [
-    callWithCosts.loading,
-    setPlaygroundStateFromTraceCall,
-    callWithCosts.result,
-  ]);
+    // Only set the call the first time the page loads, and we get the call
```
jwlee64, line 111: this was rerunnning every render because of callwithCosts.result

shawnlewis (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundPage.tsx
gtarpenning: hmmm, but isn't `callwithCosts.result` expensive? shouldn't that not be changing?

PR #3289, opened 2024-12-19
docs(weave): fix small error in EmbeddingSimilarityScorer docs

umakrishnaswamy:
currently it states that EmbeddingSimilarityScorer uses target_column as an argument to the scorer, which is not valid (and raises an error). i made a quick pr to update the docs to include column_mapping

## Testing

How was this PR tested?
yarn start


johndmulhausen:
These changes pertain to code that is no longer here; closing

PR #3293, opened 2024-12-19
WIP: add new op to retrieve table with steps

nicholaspun-wandb:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?


nicholaspun-wandb:
See https://github.com/wandb/weave/pull/3293. We can drive this completely from the new panel, so the new op won't be necessary

PR #3299, opened 2024-12-20
chore(ui): Token and cost formatting - higher precision

m-rgba:
## Description

- Updates cost formatting based on customer feedback (4 points of precision).
- Updated token formatting to make it more precise (don't collapse thousands).
- Right aligns both.

## Screenshot

| Before | After |
| -------- | ------- |
| before | after |


m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=e9c2832da95736056f7981dad579e03150398d8b

tssweeney (approved):

PR #3304, opened 2024-12-21
feat(weave): Add support for jpegs and pngs

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-22316
- https://wandb.atlassian.net/browse/WB-22453

This PR adds support for logging and viewing JPEGs.  Previously, all images would be saved as PNG, even if the source image was a JPEG.

If you somehow force-logged a JPEG, it still wouldn't work because the UI didn't support JPEGs.
(before)
image
(after)
image

Testing is mostly manual, with some client tests added for the different formats.  We should add playwright tests when they are available.  I have verified that the image that comes back is a jpeg (and you can confirm by inspecting/saving the img)


weave/trace/util.py
```diff
@@ -168,6 +168,59 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
     return deco
 
 
+KT = TypeVar("KT")
+VT = TypeVar("VT")
+
+
+class InvertableDict(MutableMapping[KT, VT]):
```
tssweeney, line 175: this is cool... but 2 things:
1. Seems a bit overkill: `def invert_dict(d): return {v: k for k, v in d.items()}`?
2. Would put this in it's own file since it has nothing specific to do with trace

weave/type_handlers/Image/image.py
```diff
@@ -49,15 +66,21 @@ def save(obj: Image.Image, artifact: MemTraceFilesArtifact, name: str) -> None:
...
+    with artifact.new_file(fname, binary=True) as f:
+        obj.save(f, format=ext_to_pil_format[ext])  # type: ignore
 
 
 def load(artifact: MemTraceFilesArtifact, name: str) -> Image.Image:
-    # Note: I am purposely ignoring the `name` here and hard-coding the filename. See comment
-    # on save.
-    path = artifact.path("image.png")
-    return Image.open(path)
+    for ext in pil_format_to_ext.values():
```
tssweeney, line 75: I really don't like this blind try/catch to find the right file name. Is there a way to know the target file up front?

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/PIL.Image.Image/PILImageImage.tsx
```diff
@@ -6,9 +6,11 @@ import {CustomWeaveTypePayload} from '../customWeaveType.types';
 
 type PILImageImageTypePayload = CustomWeaveTypePayload<
   'PIL.Image.Image',
-  {'image.png': string}
+  {'image.jpg'?: string; 'image.png'?: string}
```
tssweeney, line 9: i think this would be better as `{image.jpg: string} | {image.png: string}`

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/PIL.Image.Image/PILImageImage.tsx
```diff
@@ -21,10 +23,17 @@ export const PILImageImage: React.FC<{
   data: PILImageImageTypePayload;
 }> = props => {
   const {useFileContent} = useWFHooks();
+
+  const imageFileName = Object.keys(props.data.files)[0] as
```
tssweeney, line 27: This feels overly optimistic. I would instead get the keys and look for exact matches, then have a failure case

weave/trace/util.py
andrewtruong: 1. That's not guaranteed to be a bijection though because `invert_dict(invert_dict(x))` may not equal `x` (consider 2 keys map to 1)
2. Ok, we can move

weave/type_handlers/Image/image.py
andrewtruong: fixed

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/PIL.Image.Image/PILImageImage.tsx
```diff
@@ -21,11 +21,25 @@ export const PILImageImage: React.FC<{
...
+  const imageTypes = {
+    'image.jpg': 'jpg',
+    'image.png': 'png',
+  } as const;
+
+  const imageKey = Object.keys(props.data.files).find(
+    key => key in imageTypes
+  ) as keyof PILImageImageTypePayload['files'];
+  if (!imageKey) {
+    throw new Error('No image file found');
```
tssweeney, line 34: def don't throw. just render a empty / some info state. Throwing will crash the page

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/PIL.Image.Image/PILImageImage.tsx
```diff
@@ -21,11 +21,25 @@ export const PILImageImage: React.FC<{
...
+  const imageTypes = {
+    'image.jpg': 'jpg',
+    'image.png': 'png',
+  } as const;
+
+  const imageKey = Object.keys(props.data.files).find(
+    key => key in imageTypes
+  ) as keyof PILImageImageTypePayload['files'];
+  if (!imageKey) {
+    return <span>Img not found!</span>;
```
tssweeney, line 34: I would use `weave-js/src/components/PagePanelComponents/Home/Browse2/NotApplicable.tsx` this component

tssweeney, line 34: early returns must happen after all use hooks

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/PIL.Image.Image/PILImageImage.tsx
```diff
@@ -21,20 +22,36 @@ export const PILImageImage: React.FC<{
...
   const {useFileContent} = useWFHooks();
+
+  const imageTypes = {
+    'image.jpg': 'jpg',
+    'image.png': 'png',
+  } as const;
+
+  const imageKey = Object.keys(props.data.files).find(
+    key => key in imageTypes
+  ) as keyof PILImageImageTypePayload['files'];
```
tssweeney, line 33: This is still not correct since you are not guaranteed to find the key. This cast is invalid and causes line 37 to be incorrect.

tssweeney (approved):

PR #3305, opened 2024-12-23
chore(ui): Call table footer - pagination sizing, call-out for Alt/Option command

m-rgba:
## Description
- Updated sizing of pagination footer to be smaller / tighter.
- Added call-out for Alt/Option+click command for the datatable.
- De-duped the playground `isMac` command to use utils (same as here).

## Screenshot
Before:
before-pagination
After:
after-pagination


m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=f8473a6026e74f0dfc73616ebbbdcefa41fbb8cf

m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=7ad33d3337f41e198abe1c3dfb901a4fa2972494

m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=92ae47ee56e0e54e79d5579ce804febd489481d5

m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=01653c264202c91118dfb4740526cc89a2de3ea9

m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=61d68e5a6d942971c7e12795b7a94b0e04302a7f

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/DataTableView.tsx
```diff
@@ -392,7 +392,7 @@ export const DataTableView: FC<{
   const displayRows = 10;
   const hideFooter = USE_TABLE_FOR_ARRAYS && gridRows.length <= displayRows;
   const headerHeight = 40;
-  const footerHeight = 52;
+  const footerHeight = 40;
```
gtarpenning, line 395: what does this do?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTableButtons.tsx
```diff
@@ -666,34 +667,75 @@ export const PaginationButtons = () => {
...
+            border: `1px solid ${MOON_400}`,
+            fontWeight: '600',
+            padding: '0px 4px',
+            marginLeft: '4px',
+            marginRight: '4px',
+            borderRadius: '4px',
+          }}>
+          Click
+        </Box>
+        on a cell to filter by the value
```
gtarpenning, line 737: nit: this could be its own decomposed component. name suggestions: "CellSortCallout", "PaginationControlsSortCallout" etc

gtarpenning (changes requested):
Two small changes requested:

for small table sizes we probably want this text to be hidden (or something, maybe tooltip?) instead of wrapping.
Screenshot 2024-12-30 at 9 54 05 AM

When the table is being shown in a preview, like in the details drawer, we don't need to show this callout. It just adds complexity and does nothing (cells aren't clickable here). 
Screenshot 2024-12-30 at 9 55 07 AM

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/DataTableView.tsx
m-rgba: This is the height of the footer bar in px after the smaller button change. It gets passed to help calculate the height of the datatable above it.

m-rgba:
> When the table is being shown in a preview, like in the details drawer, we don't need to show this callout. It just adds complexity and does nothing (cells aren't clickable here).

Yes, will update - should hide it away when the screen is too wide and on the child views. Will also make it its own component 👍

jamie-rasmussen:
We intentionally haven't advertised Option+Click because it doesn't work for all columns.

jamie-rasmussen:
> The two that I saw that had issues were JSON columns (doing [object Object] as a bug) and the token / latency columns.

There are other that don't work - cost columns, status plus some more complicated column cases where user might or might not expect something to happen like trace name / ID, feedback, human annotation.

PR #3308, opened 2024-12-23
chore(ui): Simplified object viewer

m-rgba:
## Description
- Simplified object viewer buttons to use one button for collapse / expand.
- Show / hide functionality for clicking the header to hide/show the object viewer.
- Adjusted the sizing of the header for the object viewer.
- Adjusted padding for the object viewer in the calls panel.

## Screenshot [ToDo]

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
@@ -44,7 +44,7 @@ TitleRow.displayName = 'S.TitleRow';
 const Title = styled.div`
   flex: 1 1 auto;
   font-family: Source Sans Pro;
-  font-size: 16px;
+  font-size: 14px;
   font-weight: 600;
   line-height: 32px;
   letter-spacing: 0px;
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
@@ -22,6 +22,7 @@ import {WeaveCHTable, WeaveCHTableSourceRefContext} from './DataTableView';
 import {ObjectViewer} from './ObjectViewer';
 import {getValueType, traverse} from './traverse';
 import {ValueView} from './ValueView';
+import {Icon} from '../../../../../Icon';
 
 const EXPANDED_IDS_LENGTH = 200;
 
@@ -49,6 +50,13 @@ const Title = styled.div`
   line-height: 32px;
   letter-spacing: 0px;
   text-align: left;
+  display: flex;
+  align-items: center;
+  cursor: pointer;
+  
+  &:hover {
+    opacity: 0.8;
+  }
 `;
 Title.displayName = 'S.Title';
 
@@ -172,6 +180,14 @@ const ObjectViewerSectionNonEmpty = ({
     }
   };
 
+  const onToggleExpansion = () => {
+    if (mode === 'expanded') {
+      onClickCollapsed();
+    } else {
+      onClickExpanded();
+    }
+  };
+
   // On first render and when data changes, recompute expansion state
   useEffect(() => {
     const isSimple = isSimpleData(data);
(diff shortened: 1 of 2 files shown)


m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=2d72424eebbc1a0442a57445f93b361949321550

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/ObjectViewerSection.tsx
```diff
@@ -187,21 +203,24 @@ const ObjectViewerSectionNonEmpty = ({
...
-        <Title>{title}</Title>
-        <Button
-          variant="quiet"
-          icon="row-height-small"
-          active={mode === 'collapsed'}
-          onClick={onClickCollapsed}
-          tooltip="View collapsed"
-        />
+        <Title
+          onClick={() => setMode(mode === 'hidden' ? 'collapsed' : 'hidden')}>
```
gtarpenning, line 207: total nit but i've been thinking that 1 liner functions are more aesthetic than I previously thought, we might consider adding a local `isModeHidden` fn like:

```javascript
function isModeHidden(mode) {
   return mode === 'hidden'
}
function isModeExpanded...
```

gtarpenning: and then obviously use it everywhere we test the mode state

gtarpenning (approved):
I think this makes sense, I like the icon change. One micro tioli style suggestion.

Note: This will change our users experience of the object flow, they might go looking for this button we have removed. 

Also, not your problem, but the object viewer not auto resizing after collapse is annoying
object-viewer-empty

m-rgba:
> I think this makes sense, I like the icon change. One micro tioli style suggestion.
> 
> Note: This will change our users experience of the object flow, they might go looking for this button we have removed.
> 
> Also, not your problem, but the object viewer not auto resizing after collapse is annoying !

For sure - I thought it was an acceptable level of moving the cheese since we use the title collapse elsewhere for the entire viewer, and the collapse/expand buttons being separate buttons when they're really a toggle was kind of an anti-pattern and was cluttering it up.

Yeah I saw the height re-size as well but I thought it may be better for a separate PR cause I know there was some stuff around setting the explicit height and virtualization to take into account.

jamie-rasmussen:
> collapse/expand buttons being separate buttons when they're really a toggle was kind of an anti-pattern and was cluttering it up

They aren't a toggle because the user can manually expand or collapse items in the tree. If I have a huge deep tree in the collapsed state and I drill down 5 levels, in this design I can't collapse with a single click, I'd have to expand all before I could return to collapsed.

m-rgba:
> They aren't a toggle because the user can manually expand or collapse items in the tree. If I have a huge deep tree in the collapsed state and I drill down 5 levels, in this design I can't collapse with a single click, I'd have to expand all before I could return to collapsed.

@jamie-rasmussen I see where you're coming from on it - in that use case I could also make the argument that the user would loose scroll position if they need to expand / contract.

While I'd like to clean it up, I don't think I have enough data to move the cheese on it. Going to look at how I can wire up the telemetry to track sessions and usage for the object viewer. 

Will just update the icon for now to use our general expand icon, add the telemetry tags, and keep the title + chevron for the object viewer hide / collapse.

m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=1c8489d89638d3069b3517266889887ffa6d1867

m-rgba:
Updated preview after merge conflict / button change:
https://beta.wandb.ai/?betaVersion=9d77fad973c4aefcaa1c83edea7cfbd69cea0607

m-rgba:
Hey @jamie-rasmussen @gtarpenning - just wanted to check you're good if I merge this in?
Made the updates to the button and function logic recommended above 👍

jamie-rasmussen (commented):
I think there's still a bug, if I do - Expand tree, then hide with title, unhide with title, the "collapsed" button is selected but the tree still has expanded state

m-rgba:
Definitely my bad - updated to fix, splitting out the state of the viewer from the mode it's in.
Preview after fix: https://beta.wandb.ai/?betaVersion=362d40ab9f649f679b083d7a5d01157e3b367ab6

PR #3316, opened 2025-01-02
chore(weave): Implement public version of "apply scorer" (aka MVP Guardrails & Monitoring)

tssweeney:
Exposes a new `.apply_scorer` method to the `Call` interface which lays the foundation for guardrails and monitoring. Unifies logic with Evals such that the eval framework uses this code path


tests/trace/test_evaluation_performance.py
```diff
@@ -117,6 +117,7 @@ async def test_evaluation_performance(client: WeaveClient):
         res = await evaluation.evaluate(predict)
         assert res["score"]["true_count"] == 1
         log = [l for l in client.server.attribute_access_log if not l.startswith("_")]
+        print(log)
```
andrewtruong, line 120: intentional?

weave/trace/weave_client.py
```diff
@@ -864,8 +895,8 @@ def finish_call(
         else:
             postprocessed_output = original_output
         self._save_nested_objects(postprocessed_output)
-
-        call.output = map_to_refs(postprocessed_output)
+        output_as_refs = map_to_refs(postprocessed_output)
+        call.output = postprocessed_output
```
andrewtruong, line 899: The call output is not mapped to refs anymore?

weave/trace/op.py
```diff
@@ -417,15 +417,19 @@ def _do_call(
     # Handle all of the possible cases where we would skip tracing.
     if settings.should_disable_weave():
         res = func(*pargs.args, **pargs.kwargs)
+        call.output = res
```
andrewtruong, line 420: Why do we have this?  Is weave is disabled, would the call object get tracked anyways?

tests/trace/test_call_apply_scorer.py
```diff
@@ -0,0 +1,213 @@
...
+from weave.trace.weave_client import Call, Op, WeaveClient
+
+
+def do_assertions_for_scorer_op(
+    apply_score_res: ApplyScorerResult,
+    call: Call,
+    score_fn: Op | weave.Scorer,
+    client: WeaveClient,
+):
+    # assert apply_score_res.feedback_id is not None
```
andrewtruong, line 18: intentional?

tests/trace/test_call_apply_scorer.py
```diff
@@ -0,0 +1,213 @@
...
+    assert apply_score_res.score_call.id is not None
+    assert apply_score_res.result == 0
+
+    feedbacks = list(call.feedback)
+    assert len(feedbacks) == 1
+    target_feedback = feedbacks[0]
+    # assert target_feedback.id == apply_score_res.feedback_id
```
andrewtruong, line 25: intentional?

weave/trace/weave_client.py
tssweeney: Great eye! This is a pretty important change. Let me explain:
1.`map_to_refs` turns values into refs.
2. This is what we want to store on the server
3. However, as currently implemented, we also attach the ref to call.output. Technically this matches what you would get if you "get" the call.
4. Unfortunately, this means getting the output value for a call you just made results in a network request. This implementation keeps the ref (b/c it is attached to the ref prop, but maintain the value space

weave/trace/op.py
tssweeney: this allows the non-tracked `call` to have the correct output for downstream users

tests/trace/test_call_apply_scorer.py
tssweeney: cleaned

tests/trace/test_call_apply_scorer.py
tssweeney: cleaned

tests/trace/test_evaluation_performance.py
tssweeney: cleaned

weave/trace/op.py
andrewtruong: When would a downstream user need the call if weave is disabled?  Can you give me an example of how that would be used?

tssweeney: Well let's say they are using an API like:

```
res, call = my_op.call(...)
if call.output > 0.5:
    do_a()
else
    do_b()
```

This code path is valid when tracing is on, but brakes when you disable it.

andrewtruong: hmm, ok I see the point here.

This raises a separate concern about whether users can rely on the shape of `call` even when weave is disabled, but I won't block here

andrewtruong (approved):

PR #3318, opened 2025-01-03
fix(weave): ChatNVIDIA shouldn't depend on openai being installed

abraham-leal:
## Description

Previous impl relied on openai being installed for typing. That has been removed.
This also adds original inputs to the input / output reported

## Testing

All tests updated.

image

### DIFF
--- a/weave/integrations/langchain_nvidia_ai_endpoints/langchain_nv_ai_endpoints.py
+++ b/weave/integrations/langchain_nvidia_ai_endpoints/langchain_nv_ai_endpoints.py
@@ -63,7 +63,6 @@ def postprocess_output_to_openai_format(output: Any) -> dict:
                 }
             ],
             "model": message.get("model_name", ""),
-            "object": "ChatNVIDIA._generate",
             "tool_calls": message.get("tool_calls", []),
             "usage": enhanced_usage,
         }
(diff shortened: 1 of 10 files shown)


weave/integrations/langchain_nvidia_ai_endpoints/langchain_nv_ai_endpoints.py
```diff
@@ -66,15 +62,15 @@ def postprocess_output_to_openai_format(output: Any) -> dict:
...
             ],
-            created=int(time.time()),
-            model=message.get("model_name", ""),
-            object="chat.completion",
-            tool_calls=message.get("tool_calls", []),
-            system_fingerprint=None,
-            usage=enhanced_usage,
-        )
+            "model": message.get("model_name", ""),
+            "object": "ChatNVIDIA._generate",
```
tssweeney, line 66: what is this "object" field. Seems unlikely that our internal name is the correct thing to put here

abraham-leal: the object field is part of the ChatCompletion spec, it doesn't have to be here if we don't want it to be given we removed the dependency. In openai, this is "chat.completion", here i put not our internal name, but rather the internal name of the ChatNVIDIA library.

abraham-leal: Removed given its not needed now that this integration doesn't depend on openai, remove check for it in tests as well.

tssweeney (approved):

PR #3323, opened 2025-01-06
chore(weave): allow call.feedback.add() for annotations

gtarpenning:
## Description

[WB-22606](https://wandb.atlassian.net/browse/WB-22606)

Add annotation feedback with an annotation spec ref like: 

```python
spec_name = "wandb.annotation.number-spec"
annotation_ref = ref.uri()
payload = {"value": 3}

call.feedback.add(spec_name, payload, annotation_ref=annotation_ref)
```

## Testing

Add test

[WB-22606]: https://wandb.atlassian.net/browse/WB-22606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/test_annotation_feedback.py
+++ b/tests/trace/test_annotation_feedback.py
@@ -510,6 +510,6 @@ def test_annotation_feedback_sdk(client):
     # no wandb.annotation prefix
     with pytest.raises(
         ValueError,
-        match="Feedback type must conform to the format: 'wandb.annotation.<name>'.",
+        match="To add annotation feedback, feedback_type must conform to the format: 'wandb.annotation.<name>'.",
     ):
         calls[0].feedback.add("number_rating", {"value": 3}, annotation_ref=ref.uri())
--- a/tests/trace/test_annotation_feedback.py
+++ b/tests/trace/test_annotation_feedback.py
@@ -483,7 +483,7 @@ def test_annotation_feedback_sdk(client):
 
     # Add annotation feedback
     calls[0].feedback.add(
-        "wandb.annotation.number_rating",
+        "wandb.annotation.number-spec",
         {"value": 3},
         annotation_ref=ref.uri(),
     )
--- a/tests/trace/test_annotation_feedback.py
+++ b/tests/trace/test_annotation_feedback.py
@@ -1,5 +1,6 @@
 import pytest
 from pydantic import BaseModel, Field
+from requests import HTTPError
 
 import weave
 from weave.flow.annotation_spec import AnnotationSpec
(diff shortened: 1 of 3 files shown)


tests/trace/test_annotation_feedback.py
```diff
@@ -456,3 +456,57 @@ class FeedbackModel(BaseModel):
...
+
+    do_call()
+    do_call()
+
+    calls = do_call.calls()
+    assert len(list(calls)) == 2
+
+    # Add annotation feedback
+    calls[0].feedback.add(
+        "wandb.annotation.number-spec",
```
jamie-rasmussen, line 485: API wise I don't like that it isn't obvious how to construct this string

tests/trace/test_annotation_feedback.py
```diff
@@ -456,3 +456,57 @@ class FeedbackModel(BaseModel):
...
+        calls[0].feedback.add(
+            "wandb.annotation.number_rating", {"value": 3}, annotation_ref=""
+        )
+
+    # invalid annotation_ref
+    with pytest.raises(ValueError):
+        calls[0].feedback.add("number_rating", {"value": 3}, annotation_ref="ssss")
+
+    # no wandb.annotation prefix
+    with pytest.raises(Exception):
```
jamie-rasmussen, line 511: We would ideally raise something more informative in this case. Maybe add a comment to that effect?

jamie-rasmussen (approved):

tests/trace/test_annotation_feedback.py
gtarpenning: Yes completely...  We could add a property to the `AnnotationSpec` itself like "feedback_type" that constructs this? @jamie-rasmussen

jamie-rasmussen: Yes. I guess that also raises the question for me of whether we would want a convenience method on the `AnnotationSpec` that was something like: `number_spec.annotate_call(calls[0], {"value": 3})` - since you have to construct or get the spec anyway?

gtarpenning: oh interesting... that's a fun idea. Almost like the class is an annotator itself... 

The payload value could be abstracted too...

```python

rating_1_to_10_annotator = AnnotationSpec(...)

rating_1_to_10_annotator.annotate_call(call_id, 3)
```

gtarpenning: perfect

gtarpenning: ticketed both things separately

PR #3324, opened 2025-01-06
chore(weave): add query and parameters to failed calls query requests

gtarpenning:
## Description

[WB-22605](https://wandb.atlassian.net/browse/WB-22605)

Catch and re-raise every streaming error in calls_query_stream, with additional information logged. 

Why should we do this? 
- this block errors occasionally for a variety of reasons and we only capture the trace, not the query or parameters. Debugging becomes very challenging... 

## Testing

Manual testing locally, with a very easily tripped mem limit, example: 

Screenshot 2025-01-06 at 11 32 57 AM

[WB-22605]: https://wandb.atlassian.net/browse/WB-22605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1603,24 +1603,31 @@ def _query_stream(
...
-            if isinstance(stream.source, QueryResult):
-                summary = stream.source.summary
-            logger.info(
-                "clickhouse_stream_query",
-                extra={
-                    "query": query,
-                    "parameters": parameters,
-                    "summary": summary,
-                },
+        try:
```
tssweeney, line 1606: how does this manifest in the trace server? Should we do something on the server side such that this actually shows a 500?

tssweeney (approved):
Approving, but i am not sure where this logger will end up

PR #3327, opened 2025-01-06
chore(weave): Fixes dedicates instance connection issues with traces API

wylerz:
## Description

The URL generation logic correctly handles default `https://api.wandb.ai` url by changing trace url to `https://trace.wandb.ai`, but for non default hosts it fails to define traces url correctly (just uses `https://${resolvedHost}`).

Fix is to append `/traces` to the url for non default hosts.


andrewtruong (approved):
thanks!

PR #3325, opened 2025-01-06
chore(weave): Remove dead code paths from `Evaluation.evaluate` by implementing correct typing.

tssweeney:
This is a pure refactor PR and should not change user experience. I am working on a larger refactor of this part of the code base but wanted to extract this portion as an incremental change. Specifically:
1. More strictly defines the dataset list to be a list of dicts (already enforced in application logic) allowing 2 branches to be removed.
2. Expose `_post_init_dataset` and `_post_init_scorers` properties in order to have more tightly defined symbols during execution.
3. Remove unreachable code due to the above.

_Note: Recommend hiding whitespace when reviewing due to indentation changes_

Now includes changes from https://github.com/wandb/weave/pull/3328

PR #3328, opened 2025-01-06
chore(weave): Extract common functions in `eval.py` into re-usable common modules.

tssweeney:
This is another refactor PR to decouple Scoring and model prediction mechanics from the eval api itself. This builds off of https://github.com/wandb/weave/pull/3325. There should be no functional changes to the user. Specifically:
* `weave/trace/op_caller.py` now contains the helper functions for `async_call` and `async_call_op` which unifies op calling for async and sync calls under the same API.
* Extracts `apply_scorer_async` into `weave/scorers/base_scorer.py` prepping for re-use elsewhere
* Extracts `apply_model_async` into `weave/flow/model.py` for similar reasons.

### DIFF
--- a/weave/flow/model.py
+++ b/weave/flow/model.py
@@ -1,7 +1,18 @@
-from typing import Callable
+import inspect
+import textwrap
+import time
+import traceback
+from dataclasses import dataclass
+from typing import Any, Callable, Optional, Union
+
+from rich import print
 
 from weave.flow.obj import Object
-from weave.trace.op import Op, is_op
+from weave.trace.errors import OpCallError
+from weave.trace.isinstance import weave_isinstance
+from weave.trace.op import Op, as_op, is_op
+from weave.trace.op_caller import async_call_op
+from weave.trace.weave_client import Call
 
 INFER_METHOD_NAMES = {"predict", "infer", "forward", "invoke"}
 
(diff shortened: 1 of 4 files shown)


weave/flow/model.py
```diff
@@ -57,3 +68,110 @@ def get_infer_method(model: Model) -> Op:
     raise MissingInferenceMethodError(
         f"Missing a method with name in ({INFER_METHOD_NAMES})"
     )
+
+
+# Using `dataclass` because pydantic does not like `Call` as a property
```
andrewtruong, line 73: That's odd -- why?

weave/scorers/base_scorer.py
```diff
@@ -157,3 +160,194 @@ def _has_oldstyle_scorers(scorers: list[Union[Op, Scorer]]) -> bool:
...
+
+
+# Using `dataclass` because pydantic does not like `Call` as a property
+@dataclass
+class ApplyScorerSuccess:
+    result: Any
+    score_call: Call
+
+
+ApplyScorerResult = ApplyScorerSuccess
```
andrewtruong, line 172: There's no error case?

andrewtruong (approved):
lgtm

weave/flow/model.py
tssweeney: it doesn't like the future

weave/scorers/base_scorer.py
tssweeney: not yet (:

PR #3336, opened 2025-01-07
feat(weave): Add global post-processing options

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-22599

This allows users to configure custom post-processing functions to be applied globally and is intended to be give users fine-grain control over what data they actually want to log to weave.

For example, you can define a custom `redact_keys` function to hide anything with "key" in the name:
```py
import weave


def redact_keys(d: dict) -> dict:
    for k in d:
        if "key" in k.lower():
            d[k] = "REDACTED"
    return d


client = weave.init("testing306", global_postprocess_inputs=redact_keys)


@weave.op
def func(api_key: str, key_secret: str, name: str) -> int:
    return f"Hi {name}!"


func("123", "456", "John")
```

image

### DIFF
--- a/tests/trace/test_postprocessing.py
+++ b/tests/trace/test_postprocessing.py
@@ -1,0 +1,44 @@
+import weave
+from weave.trace import api
+
+
+def redact_keys(d: dict) -> dict:
+    print(f"Start redacting keys {d=}")
+    for k in d:
+        print(f"Looking at key {k=}")
+        if "key" in k.lower():
+            print(f"Redacting key {k=}")
+            d[k] = "REDACTED"
+    print(f"Finished redacting keys {d=}")
+    return d
+
+
+def redact_output(s: str) -> str:
+    if "key" in s.lower():
+        return "API KEY DETECTED IN STRING; REDACTED."
+    return s
+
+
+api._global_postprocess_inputs = redact_keys
+api._global_postprocess_output = redact_output
+
+
+def test_global_postprocessing(client) -> None:
+    @weave.op
+    def func(api_key: str, secret_key: str, name: str, age: int) -> str:
+        return (
+            f"Hello, {name}! You are {age} years old.  Also your api_key is {api_key}."
+        )
+
+    res = func(api_key="123", secret_key="456", name="John", age=30)
+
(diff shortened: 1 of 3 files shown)


weave/trace/api.py
```diff
@@ -28,12 +28,17 @@
...
+_global_postprocess_inputs: Callable[[Any], Any] | None = None
+_global_postprocess_output: Callable[[Any], Any] | None = None
+
 
 def init(
     project_name: str,
     *,
     settings: UserSettings | dict[str, Any] | None = None,
     autopatch_settings: AutopatchSettings | None = None,
+    global_postprocess_inputs: Callable[[Any], Any] | None = None,
```
jamie-rasmussen, line 40: I feel like the `global_` prefix may be unnecessary. 

Can the `Any`s be tightened?

At a global level, I wonder if it would be useful to also pass in an Op reference to this function, so you could do something like "delete the bulky images in the inputs, but only if the Op is (some autopatched thing)"

andrewtruong: I added `global_` here to suggest to the user that this will be applied to EVERY op.  Do you think the doc is sufficient?

Tightened up the Anys

There's also an autopatched op config on init (that's autopatch_settings).

tests/trace/test_postprocessing.py
```diff
@@ -0,0 +1,50 @@
...
+
+def redact_keys(d: dict) -> dict:
+    for k in d:
+        if "key" in k.lower():
+            d[k] = "REDACTED"
+    return d
+
+
+def redact_output(s: str) -> str:
+    if "key" in str(s).lower():
```
jamie-rasmussen, line 15: According to the input type, s is already a `str`

weave/trace/api.py
jamie-rasmussen: I feel like the `global_` is kind of implied from this being on the weave.init. I'd probably get a second opinion but I won't block on it.

I feel like the Op is useful information, but I'm OK waiting until a customer asks for it. If we want to keep the API simple another future possibility might be a "get_current_op" kind of method.

jamie-rasmussen (approved):

PR #3340, opened 2025-01-07
feat(weave): Make redaction applied to captured code

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-22604

Before
image

After
image


andrewtruong:
yeah that case is tougher and I'll reserve for another PR!

PR #3345, opened 2025-01-08
chore(app): Adding an optional name to label the number input

jtulk:
## Description
I want an easier way to select these in E2E tests, also helps with accessibility.


weave-js/src/common/components/elements/NumberInput.tsx
```diff
@@ -79,6 +80,7 @@ const NumberInput: React.FC<NumberInputProps> = props => {
   return (
     <div className="number-input__container" style={props.containerStyle}>
       <Input
+        aria-label={props.name}
```
nhusher, line 83: Can this prop be something other than `name`? Inputs can (and often should) have a `name` field which has important implications for form submissions. `label` or `ariaLabel` might be more appropriate?

nhusher (approved):

PR #3352, opened 2025-01-09
chore(dev): Use uuid v4 instead of v7 in error boundary

jamie-rasmussen:
## Description

I'm not sure why, but in development an older version of the uuid library is sometimes referenced instead of the ^11.0.3 version our package.json specifies. The older version doesn't support v7. Since we don't really need v7 for the React Error Boundary use case I'm switching it to use v4.

Before:
image

After:
image

## Testing

How was this PR tested?


andrewtruong (approved):
seems weird, but ok :)

PR #3355, opened 2025-01-09
chore(weave): Make REDACT_KEYS a set instead of tuple

andrewtruong:
This is easier to modify if the user needs to add extra keys


jamie-rasmussen (commented):
I feel like we should offer a documented public API if we want users to have this feature.

PR #3356, opened 2025-01-09
chore(weave): feature parity between callsStreamQuery and get_calls

gtarpenning:
## Description

[WB-22766](https://wandb.atlassian.net/browse/WB-22766)

The user should not be using raw server methods, but this is what we suggest in the python export window due to incomplete parity between the client convenience method and the raw server method. This pr closes that gap. 

## Testing

Adds test

[WB-22766]: https://wandb.atlassian.net/browse/WB-22766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/trace/test_weave_client.py
```diff
@@ -355,6 +355,114 @@ def test_calls_query(client):
...
+                "$contains": {
+                    "input": {"$getField": "inputs.s"},
+                    "substr": {"$literal": "str"},
+                }
+            }
+        }
+    )
+
+    # use all the parameters to get_calls
+    result = list(
```
andrewtruong, line 384: nit: `client_result`

gtarpenning:
Sort of concerning that these tests are failing yet the sqlite tests were all green when run locally....

gtarpenning:
> Sort of concerning that these tests are failing yet the sqlite tests were all green when run locally....

some sort of weird interaction in the client when using `get_calls`  with `include_costs` but only in the first part of the test...

tests/trace/test_weave_client.py
```diff
@@ -355,6 +355,112 @@ def test_calls_query(client):
...
+    )
+
+    # use all the parameters to get_calls
+    client_result = list(
+        client.get_calls(
+            filter=weave_client.CallsFilter(op_names=[call1.op_name]),
+            limit=1,
+            offset=0,
+            query=query,
+            sort_by=[weave_client.SortBy(field="started_at", direction="desc")],
```
andrewtruong, line 390: aren't these really `tsi.SortBy`?

weave/trace/weave_client.py
```diff
@@ -123,16 +124,28 @@ def __init__(
...
             raise ValueError("page_size must be greater than 0")
 
     @lru_cache
     def _fetch_page(self, index: int) -> list[T]:
+        if self.limit is not None or self.offset is not None:
+            # If limit/offset are provided, make a single request with those parameters
+            offset = self.offset or 0
+            if self.limit is not None:
+                return self.fetch_func(offset, self.limit)
+            return self.fetch_func(offset, self.page_size)
```
andrewtruong, line 146: Are page size and limit technically the same thing here?

weave/trace/weave_client.py
```diff
@@ -143,6 +156,12 @@ def _get_one(self, index: int) -> T | R:
         if index < 0:
             raise IndexError("Negative indexing not supported")
 
+        if self.limit is not None and index >= self.limit:
```
andrewtruong, line 159: Should this be `index >= limit + offset`?

andrewtruong: e.g. you may ask for item 2000, with limit 1000.  This is ok if you have offset 1500, giving you items in range (1500, 2500)

weave/trace/weave_client.py
```diff
@@ -143,6 +156,12 @@ def _get_one(self, index: int) -> T | R:
...
+            raise IndexError(f"Index {index} out of range")
+
+        if self.offset is not None:
+            index += self.offset
```
andrewtruong, line 163: Oh I see, you may need to move this up

weave/trace/weave_client.py
andrewtruong: Also, should limited queries also take advantage of pagination?

andrewtruong:
I think my confusion stems from:
1. I think pagination should apply even for limited queries
2. Maybe this is a bit confusing, but what "index space" do we think of the iterator being in.  For example, if you say `limit=1000`, does that mean that `iterator[1001]` raises IndexError?  What if `offset=500, limit=1000`.  Are the valid indices [0,999] or [500,1499]?

weave/trace/weave_client.py
gtarpenning: After playing around with this a bit, I think it makes the most sense for the iterator to behave most like a python list after the initial call. For example: 

```python
calls = get_calls(offset=5, limit=10)

# this should always work
call1 = calls[0]

# this should break
call8 = calls[7]
``` 

It just feels so wrong to ask for 5 calls back from the api and not be able to index into the first one...

weave/trace/weave_client.py
```diff
@@ -745,16 +790,46 @@ def get(self, ref: ObjectRef) -> Any:
     def get_calls(
         self,
         filter: CallsFilter | None = None,
```
andrewtruong, line 792: nit: Would put a `*` after self to force users to enter these as kwargs

andrewtruong (approved):

weave/trace/weave_client.py
gtarpenning: ah right, perfect.

PR #3362, opened 2025-01-10
feat(weave): Make ref-getting more pleasant

andrewtruong:
Resolves:
- https://wandb.atlassian.net/browse/WB-22632
- https://wandb.atlassian.net/browse/WB-22688
- https://wandb.atlassian.net/browse/WB-22699

This PR:
1. Adds a mechanism, `@register_object`, to register objects that can be converted from `WeaveObject` into their base python classes.  This can be used both internally and by end-users.
2. Adds a new classmethod `from_uri` to registered objects
4. Improves `ref(...).get()` to return the actual python class instead of `WeaveObject` for registered objects
5. Adds optional refs on all `weave.Object` which are initially unset, but can be set when the object is re-created from a ref

Now you can do:
```py
ds = Dataset(rows=[...])
ref = weave.publish(ds)

ds2 = ref.get()  # returns a `Dataset` instead of `WeaveObject`
ds3 = Dataset.from_uri(ref.uri())  # returns a `Dataset`
```


weave/flow/obj.py
```diff
@@ -51,6 +54,17 @@ class Object(BaseModel):
...
     __str__ = BaseModel.__repr__
 
+    @classmethod
+    def from_uri(cls, uri: str, *, objectify: bool = True) -> Self:
+        if not isinstance(cls, Objectifyable):
+            raise NotImplementedError(
+                f"`{cls.__name__}` must implement `from_obj` to support deserialization from a URI."
+            )
+
+        import weave
```
jamie-rasmussen, line 64: Could we just import the ref method?

jamie-rasmussen (approved):

PR #3363, opened 2025-01-10
feat(weave): Make it easier to get weave objects from URI

andrewtruong:
TODO: Ideally this is just defined on `Object`, but I don't think we can guarantee all objects will work just yet


andrewtruong:
merged into https://github.com/wandb/weave/pull/3362

PR #3366, opened 2025-01-10
feat(ui): Saved Views

jamie-rasmussen:
## Description

Internal Jira: https://wandb.atlassian.net/browse/WB-21701

Adds "Saved Views" to the UI. Views are saved as Weave objects. Table settings on the Traces and Evaluations pages like filters, sorting, page size, column visibility, and column pinning are stored in the view object.

New views can be saved, views can be updated, renamed, and deleted.

Some  screenshots:

Screenshot 2025-02-03 at 4 12 51 PM
Screenshot 2025-02-03 at 4 14 07 PM
Screenshot 2025-02-03 at 4 12 56 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
@@ -661,13 +661,17 @@ const CallPageBinding = () => {
   );
 };
 
-type CallsPageLastViewProps = {
+type RedirectToLastViewProps = {
   entity: string;
   project: string;
   tab: string;
 };
 
-const CallsPageLastView = ({entity, project, tab}: CallsPageLastViewProps) => {
+const RedirectToLastView = ({
+  entity,
+  project,
+  tab,
+}: RedirectToLastViewProps) => {
   const location = useLocation();
   const {loading: loadingProjectInfo, projectInfo} = useProjectInfo(
     entity,
@@ -1023,7 +1027,7 @@ const CallsPageBinding = () => {
   const query = useURLSearchParamsDict();
   const loadLastView = Object.keys(query).length === 0;
   if (loadLastView) {
-    return <CallsPageLastView entity={entity} project={project} tab={tab} />;
+    return <RedirectToLastView entity={entity} project={project} tab={tab} />;
   }
   return (
     <CallsPageLoadView
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
@@ -29,9 +29,22 @@ import {
   useParams,
 } from 'react-router-dom';
 
+import {
+  ProjectInfo,
+  useProjectInfo,
+} from '../../../common/hooks/useProjectInfo';
+import {
+  MaybeUserInfo,
+  UserInfo,
+  useViewerInfo,
+} from '../../../common/hooks/useViewerInfo';
+import {safeLocalStorage} from '../../../core/util/localStorage';
+import {useDeepMemo} from '../../../hookUtils';
 import {URL_BROWSE3} from '../../../urls';
+import {Alert} from '../../Alert';
 import {Button} from '../../Button';
 import {ErrorBoundary} from '../../ErrorBoundary';
+import {Loading} from '../../Loading';
 import {ComparePage} from './Browse3/compare/ComparePage';
 import {
   baseContext,
@@ -47,10 +60,7 @@ import {
 } from './Browse3/context';
 import {FullPageButton} from './Browse3/FullPageButton';
 import {getValidFilterModel} from './Browse3/grid/filters';
-import {
-  DEFAULT_PAGE_SIZE,
-  getValidPaginationModel,
-} from './Browse3/grid/pagination';
+import {getValidPaginationModel} from './Browse3/grid/pagination';
 import {getValidPinModel, removeAlwaysLeft} from './Browse3/grid/pin';
 import {getValidSortModel} from './Browse3/grid/sort';
 import {BoardPage} from './Browse3/pages/BoardPage';
@@ -79,6 +89,10 @@ import {OpsPage} from './Browse3/pages/OpsPage/OpsPage';
 import {OpVersionPage} from './Browse3/pages/OpsPage/OpVersionPage';
 import {OpVersionsPage} from './Browse3/pages/OpsPage/OpVersionsPage';
 import {PlaygroundPage} from './Browse3/pages/PlaygroundPage/PlaygroundPage';
+import {
+  getDefaultView,
+  getDefaultViewId,
+} from './Browse3/pages/SavedViews/savedViewUtil';
 import {ScorersPage} from './Browse3/pages/ScorersPage/ScorersPage';
 import {TablePage} from './Browse3/pages/TablePage';
 import {TablesPage} from './Browse3/pages/TablesPage';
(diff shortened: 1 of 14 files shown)


tssweeney:
Starting to dive into review here. Noticing a bug: save a view with a sort applied, change views, change back to the view with the saved sort - it is applied briefly, then resets and drops the sort

tssweeney:
Also, not blocking, but for me the interactions are very slow.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/SavedViews/savedViewUtil.ts
```diff
@@ -0,0 +1,117 @@
...
+  'cols',
+  'filter',
+  'filters',
+  'sort',
+  'pin',
+  'pageSize',
+];
+
+// Value of the object
+type ViewDefinition = Record<string, any>;
```
tssweeney, line 34: View definition is untyped?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsPage.tsx
```diff
@@ -47,68 +75,238 @@ export const CallsPage: FC<{
...
+      if (objectId === getDefaultViewId(table)) {
+        objectId = getNewViewId(table);
+      }
+      if (label === null) {
+        // If caller doesn't provide a new label, use the existing one.
+        label = baseView.val.label;
+      }
+      const className = 'SavedView';
+      tsClient
+        .objCreate({
```
tssweeney, line 145: As implemented, because `SavedView` is not defined on the server, this has no validation. This makes me a bit nervous that this critical data structure is enforced within this component. If you defined a builtin object, you could use `useCreateBuiltinObjectInstance` which would allow you to simply pass the relevant data:
```
    project_id: string;
    object_id: string;
    val: T;
    builtin_object_class?: string;
```

And have val be validated in typescript and enforced on the backend. This feels like a safer approach

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -642,123 +661,347 @@ const CallPageBinding = () => {
...
-      return {};
-    }
-  }, [query.filter, entity, project, tab]);
+type CallsPageLastViewProps = {
+  entity: string;
+  project: string;
+  tab: string;
+};
+
+const CallsPageLastView = ({entity, project, tab}: CallsPageLastViewProps) => {
```
tssweeney, line 670: I would call this `RedirectToLastView` to emphasize that it is a redirect

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -642,123 +661,347 @@ const CallPageBinding = () => {
...
+
+  // Traces table might be under 'calls'
+  const table = tab === 'evaluations' ? 'evaluations' : 'traces';
+  const [views, setViews] = useState<TraceObjSchema[] | null>(null);
+
+  const getTsClient = useGetTraceServerClientContext();
+  const tsClient = getTsClient();
+  const fetchViews = useCallback(() => {
+    tsClient
+      .objsQuery({
```
tssweeney, line 765: Similarly, this could be done with `useBaseObjectInstances`

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -642,123 +661,347 @@ const CallPageBinding = () => {
...
+  const location = useLocation();
+  const {loading: loadingProjectInfo, projectInfo} = useProjectInfo(
+    entity,
+    project
+  );
+  if (loadingProjectInfo) {
+    return <Loading />;
+  }
+  if (!projectInfo) {
+    return <Alert severity="error">Invalid project: {project}</Alert>;
```
tssweeney, line 680: Here and elsewhere, it seems like we could have a nicer landing 404 / landing page when something is not found

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsPage.tsx
```diff
@@ -47,68 +75,238 @@ export const CallsPage: FC<{
...
-      }
+
+    // Update with params from the view definition
+    for (const [key, value] of Object.entries(viewToLoad.val.definition)) {
+      newQuery.set(key, JSON.stringify(value));
+    }
+
+    newQuery.set('view', viewToLoad.object_id);
+    history.push({search: newQuery.toString()});
+    props.onRecordLastView(viewToLoad.object_id);
```
tssweeney, line 117: will this line run in an unmounted? I am not sure if the execution order ensures this runs before the URL changes

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsPage.tsx
```diff
@@ -47,68 +75,238 @@ export const CallsPage: FC<{
...
+              _bases: ['SavedView', 'Object', 'BaseModel'],
+              label,
+              definition: currentViewDefinition,
+            },
+          },
+        })
+        .then(res => {
+          const newQuery = new URLSearchParams(history.location.search);
+          newQuery.set('view', objectId);
+          history.push({search: newQuery.toString()});
```
tssweeney, line 163: same here regarding history

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsPage.tsx
```diff
@@ -47,68 +75,238 @@ export const CallsPage: FC<{
...
+
+  const tsClient = getTsClient();
+  const view = baseView.object_id;
 
-  const title = useMemo(() => {
-    if (isEvaluationTable) {
-      return 'Evaluations';
+  const currentViewDefinition = useCurrentViewDefinition(baseView);
+
+  const history = useHistory();
```
tssweeney, line 99: it is always a bit of a smell to me when a component uses history below the main routing component (in our case Browse3). It means that CallsPage itself is now strictly dependent on the URL and cannot be embedded elsewhere. This particular case seems challenging to separate, so maybe we can punt on this comment. However, we should acknowledge the impact of this change

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -642,123 +661,347 @@ const CallPageBinding = () => {
...
+    }, [query.filter, entity, project, table, viewDef])
+  );
+
   const onFilterUpdate = useCallback(
     filter => {
-      history.push(routerContext.callsUIUrl(entity, project, filter));
+      const {search} = history.location;
+      const newQuery = new URLSearchParams(search);
+      newQuery.set('filter', JSON.stringify(filter));
+      history.push({search: newQuery.toString()});
```
tssweeney, line 884: I noticed in various places in this PR (and the compare anything project), that we are often doing `history.push...` with some locally constructed information. The impact of this is that the URL contract is no longer tightly defined, but rather implicitly defined throughout the various component implementations. The idea of `weave-js/src/components/PagePanelComponents/Home/Browse3/context.tsx` is to centralize all path construction such that we have a single interface definition for construction of URLs. This unification also allows the same URL to work in peek drawers as well as main views. Specifically, something like `history.push(routerContext.callsUIUrl(entity, project, filter));` is "smart" enough to perform the appropriate URL update regardless of where it is rendered. It would be nice to keep all URL construction co-located for easier maintenance

jamie-rasmussen:
Closed in favor of https://github.com/wandb/weave/pull/4240

PR #3368, opened 2025-01-10
chore(weave_ts): Add object name sanitization

andrewtruong:
Add object sanitization comparable to what we have in python


sdks/node/src/weaveClient.ts
```diff
@@ -829,3 +829,23 @@ async function maybeFormatCode(code: string) {
   //     return code;
   //   }
 }
+
+function sanitizeObjectName(name: string): string {
+  // Replaces any non-alphanumeric characters with a single dash and removes
+  // any leading or trailing dashes. This is more restrictive than the DB
+  // constraints and can be relaxed if needed.
+  let res = name.replace(/[^\w._]+/g, '-'); // non-words
```
jamie-rasmussen, line 837: I think `_` here is redundant with `\w` but maybe it improves readability.

jamie-rasmussen (approved):

PR #3382, opened 2025-01-13
feat(weave): Allow Dataset to be constructed from Calls

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-22700

### DIFF
--- a/weave/flow/dataset.py
+++ b/weave/flow/dataset.py
@@ -1,5 +1,5 @@
 from collections.abc import Iterable, Iterator
-from typing import Any, Self
+from typing import Any
 
 from pydantic import field_validator
 from typing_extensions import Self
--- a/weave/flow/dataset.py
+++ b/weave/flow/dataset.py
@@ -1,5 +1,5 @@
-from collections.abc import Iterator
-from typing import Any
+from collections.abc import Iterable, Iterator
+from typing import Any, Self
 
 from pydantic import field_validator
 from typing_extensions import Self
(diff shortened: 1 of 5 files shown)


weave/trace/weave_client.py
```diff
@@ -534,6 +534,13 @@ class ApplyScorerSuccess:
             wc._send_score_call(self, score_call, scorer_ref_uri)
         return apply_scorer_result
 
+    def to_dict(self) -> dict:
```
tssweeney, line 537: I would more concretely type this so that the dict coming out has strongly typed fields instead of relying on the implicit fields from the dataclass

weave/trace/weave_client.py
```diff
@@ -534,6 +552,17 @@ class ApplyScorerSuccess:
...
+    def to_dict(self) -> CallDict:
+        d = {}
+        for field in dataclasses.fields(self):
+            if field.name.startswith("_"):
+                continue
+            d[field.name] = getattr(self, field.name)
+        d["op_name"] = self.op_name
+        d["display_name"] = self.display_name
+
+        return cast(CallDict, d)
```
tssweeney, line 564: this cast is not guaranteed. I would do something to at least validate the keys

andrewtruong: What case do you see missing here?

weave/flow/dataset.py
```diff
@@ -54,6 +55,11 @@ def from_obj(cls, obj: WeaveObject) -> Self:
             rows=obj.rows,
         )
 
+    @classmethod
+    def from_calls(cls, calls: Iterable[Call]) -> Self:
+        rows = [call.to_dict() for call in calls]
```
tssweeney, line 60: so much network traffic here. Hopefully in the future we can do something serverside

tssweeney (approved):

PR #3383, opened 2025-01-13
chore(weave): Make save/delete symmetric and available on weave builtins

andrewtruong:


weave/flow/obj.py
```diff
@@ -63,6 +64,23 @@ def from_uri(cls, uri: str, *, objectify: bool = True) -> Self:
             )
         return api.ref(uri).get(objectify=objectify)
 
+    def publish(self, name: Union[str, None] = None) -> ObjectRef:
+        # I'm using the `publish` term here for now, but ideally we call this `save`.
+        cls_name = self.__class__.__name__
+        if not is_registered(cls_name):
```
tssweeney, line 70: this seems overly restrictive, why enforce this?

weave/flow/obj.py
```diff
@@ -63,6 +64,23 @@ def from_uri(cls, uri: str, *, objectify: bool = True) -> Self:
...
+        return weave_publish(self, name)
+
+    def delete(self) -> None:
+        if self.ref is None:
+            raise ValueError(
+                "Can't delete an object without a ref -- has this object been saved?"
+            )
+
+        self.ref.delete()
+        self.name = f"DELETED OBJECT: {self.name}"
```
tssweeney, line 82: why change the name? this would break something like
```
obj.publish()
obj.delete()
obj.publish() # new published name is deleted
```

weave/trace/api.py
```diff
@@ -120,6 +120,23 @@ def publish(obj: Any, name: str | None = None) -> weave_client.ObjectRef:
     Returns:
         A weave Ref to the saved object.
     """
+    import weave
+
+    if not isinstance(obj, weave.Object):
+        return _publish(obj, name)
+
+    old_ref, obj.ref = obj.ref, None
```
tssweeney, line 128: can you elaborate on why this is here?

weave/flow/obj.py
andrewtruong: good point; fixed

andrewtruong: Good point.  The intent here is to let the user know that an object they have a handle to has been deleted.  Consider this case:
```py
obj = MyObject(...)
obj.delete()

obj  # looks like it is fine!  How do we warn the user that it's been deleted?
```

weave/flow/obj.py
andrewtruong: you're right, we don't need it

weave/flow/obj.py
tssweeney: you could have a deleted flag or something int he repr if you really wanted to. I am not sure why `obj` i bad here. maybe `delete` needs to be more clear that it is `delete_from_remote` or something?

weave/trace/api.py
andrewtruong: Added a comment to this part of the code

weave/flow/obj.py
andrewtruong: Without this, interacting with a deleted obj just feels like a normal obj.  One alternative might be to emit warnings when you interact with the obj, but it could get noisy

tssweeney: hmmm, why is `interacting with a deleted obj just feels like a normal obj.` a bad thing? We basically "unpublished" it from the remote, but it is still sitting there in mem

andrewtruong: hmm, maybe it's not!

PR #3384, opened 2025-01-13
feat(weave): Provide helper methods to bridge pandas and weave datasets

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-22701

Adds helpers to go back and forth between pandas dataframes.  Alternatively, consider [narwhals](https://github.com/narwhals-dev/narwhals) for multi-frame support


weave/flow/dataset.py
```diff
@@ -49,6 +52,19 @@ def from_calls(cls, calls: Iterable[Call]) -> Self:
         rows = [call.to_dict() for call in calls]
         return cls(rows=rows)
 
+    @classmethod
+    def from_pandas(cls, df: "pd.DataFrame") -> Self:
```
tssweeney, line 56: isn't the standard term `from_df` and `to_df`? maybe i am wrong. Can you point to any existing apis that are similar?

andrewtruong: No, `to_pandas` is less ambiguous:
- https://docs.xarray.dev/en/latest/generated/xarray.DataArray.to_pandas.html
- https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_pandas.html
- https://docs.rapids.ai/api/cudf/latest/user_guide/api_docs/api/cudf.dataframe.to_pandas/
- https://networkx.org/documentation/networkx-1.10/reference/generated/networkx.convert_matrix.to_pandas_dataframe.html

tssweeney (approved):

PR #3385, opened 2025-01-13
feat(weave): Support creating a dataset from ops

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-22706


tssweeney (commented):
Why do this? Is this in response to a customer/user? Seems like unnecessary sugar - makes a 1 liner from a 2 liner.

andrewtruong:
It's just sugar for expressing the user intent "How do I make a dataset of my function's calls" (as per: https://www.notion.so/Eval-and-Dataset-Workflow-API-Improvements-15fe2f5c7ef380c3a89ce031c36cb767?pvs=4#160e2f5c7ef380e7adf5d02da9390c7f)

I'm not against skipping this one

tssweeney:
> It's just sugar for expressing the user intent "How do I make a dataset of my function's calls" (as per: https://www.notion.so/Eval-and-Dataset-Workflow-API-Improvements-15fe2f5c7ef380c3a89ce031c36cb767?pvs=4#160e2f5c7ef380e7adf5d02da9390c7f)
> 
> I'm not against skipping this one

I think it is going to be very rare that you just want to create a dataset from a single op. you almost always are going to filter by something / look for a tag / look for an annotation / filter by a time window / etc... i might opt to skip this for now

PR #3387, opened 2025-01-13
feat(guardrails): Implement all the Guarddrils

soumik12345:
## Description

Implement all the guardrails from the [`safeguards`](https://github.com/soumik12345/safeguards) library.

### DIFF
--- a/weave/guardrails/prompt_injection_guardrail.py
+++ b/weave/guardrails/prompt_injection_guardrail.py
@@ -1,0 +1,66 @@
+import weave
+from pydantic import BaseModel
+from typing import Optional
+from weave.scorers.llm_scorer import InstructorLLMScorer
+from weave.scorers.llm_utils import OPENAI_DEFAULT_MODEL, create
+from weave.scorers.utils import stringify
+from weave.guardrails.prompts import (
+    PROMPT_INJECTION_GUARDRAIL_SYSTEM_PROMPT,
+    PROMPT_INJECTION_SURVEY_PAPER_SUMMARY,
+)
+
+
+class LLMGuardrailResponse(BaseModel):
+    injection_prompt: bool
+    is_direct_attack: bool
+    attack_type: Optional[str]
+    explanation: Optional[str]
+
+
+class PromptInjectionLLMGuardrail(InstructorLLMScorer):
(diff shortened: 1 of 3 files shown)


weave/guardrails/prompt_injection_guardrail.py
```diff
@@ -0,0 +1,69 @@
...
+            ],
+            model=self.model_id,
+            response_model=LLMGuardrailResponse,
+            temperature=self.temperature,
+            max_tokens=self.max_tokens,
+        )
+        return response.model_dump()
+
+    @weave.op
+    def guard(self, prompt: str) -> GuardrailResponse:
```
tssweeney, line 56: this should be called `score` since it seems like it is the entry point

ayulockin:
hey @tssweeney, thank you for the initial review. Can you find sometime to give it one more round of review. We would like to see this being merged eventually.

weave/guardrails/__init__.py
```diff
@@ -0,0 +1,3 @@
+from weave.guardrails.prompt_injection_guardrail import PromptInjectionLLMGuardrail
```
tssweeney, line 1: I would nest `guardrails` under `weave/scorers` since they are just a special case of scorer

weave/guardrails/prompt_injection_guardrail.py
```diff
@@ -0,0 +1,59 @@
...
+    attack_type: Optional[str]
+    explanation: Optional[str]
+
+
+class LLMGuardrailResponse(BaseModel):
+    safe: bool
+    reasoning: LLMGuardrailReasoning
+
+
+class PromptInjectionLLMGuardrail(InstructorLLMScorer):
```
tssweeney, line 27: this looks good - i just want to warn you that the `InstructorLLMScorer` accepts a `client` which is not a portable object, meaning anything written in this style will not be executable by the backend (ie. executing scores in the UI or in monitoring contexts)

soumik12345:
Hi @tssweeney, I addressed your feedback. Can you please take another pass through the PR?

tssweeney (approved):
Nice! Would be good to get this in the docs!

PR #3389, opened 2025-01-13
style(app): Fix runs table scroll in mobile

connieelee:
## Description

https://wandb.atlassian.net/browse/WB-19489
https://wandb.atlassian.net/browse/WB-19999

There are some scrolling issues with the runs table on mobile. The main fix is in this [core PR](https://github.com/wandb/core/pull/26776), but these minor CSS changes are also required to fully address the issues.

## Testing

See core PR

### DIFF
--- a/weave-js/src/common/css/globals.less
+++ b/weave-js/src/common/css/globals.less
@@ -58,25 +58,25 @@
 @primaryLight: #6ba6fa;
 @primaryText: @primary;
 
-@success: #00B5A0;
+@success: #00b5a0;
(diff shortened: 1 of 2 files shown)


weave-js/src/common/css/globals.less
connieelee, line 1: ooph, my auto-formatter decided this file wasn't up to par. the only real change is L232 to update `@fancyPageSidebarWidth`, which is wider than what the variable suggests

nhusher (approved):

PR #3390, opened 2025-01-13
feat(weave): adding predefined keyboard binding controls

brianlund-wandb:
## Description

- Implements https://wandb.atlassian.net/browse/WB-22252

Adds a property allowing for binding canned functions to keys.

I added an `enum` to define the specific actions we allow clients to bind to keys (in this case just increment and decrement).

## Testing

How was this PR tested?


weave-js/src/common/components/elements/SliderInput.tsx
```diff
@@ -75,6 +80,40 @@ const SliderInput: React.FC<SliderInputProps> = React.memo(
...
+        return handler(event);
+      }
+    };
+
+    useEffect(() => {
+      if (!isEmpty(keyboardBindings)) {
+        document.addEventListener('keydown', stepKeyboardListener, true);
+      }
+      return () => {
+        document.removeEventListener('keydown', stepKeyboardListener, true);
```
ericakdiaz, line 113: i'm curious if `document.addEventListener` is better vs using `onKeyDown` on the div?

brianlund-wandb: My only requirement is to make sure that the div doesn't need to have the focus to capture the event.  I can try it out and see if it still works the way I want, but that was the motivation for using the `document` instead of a more localized element.

PR #3397, opened 2025-01-14
feat(weave): Add simple way to export evaluations from SDK

andrewtruong:


tests/trace/test_client_trace.py
```diff
@@ -3236,3 +3236,87 @@ def test():
 
     assert len(test.calls()) == 2
     assert len(client.get_calls()) == 2
+
+
+@pytest.mark.parametrize(
+    "descendents_code",
+    [
+        "list(call.get_descendents())",
```
tssweeney, line 3244: this is pretty nasty. If you really want to parameterize this, pass in functions, not a raw eval

weave/trace/weave_client.py
```diff
@@ -828,6 +853,61 @@ def get(self, ref: ObjectRef, *, objectify: bool = True) -> Any:
...
+        if nested:
+            warn_once(
+                logger,
+                "Calling `get_call_descendents` with `nested=True` pulls all calls into "
+                "memory.  This may take a while!",
+            )
+            return self._get_call_descendents_nested(calls)
+        return self._get_call_descendents_flat(calls)
+
+    def _get_call_descendents_flat(self, calls: Iterable[Call]) -> CallsIter:
```
tssweeney, line 881: this is not correct if the call is not the root of the trace

PR #3406, opened 2025-01-14
feat(weave): PII Redaction for weave using Microsoft Presidio

niall-wb:
## Description

Jira Ticket: [Sensitive Data Redaction in Weave](https://wandb.atlassian.net/browse/WB-22744?atlOrigin=eyJpIjoiMmYwNGNlY2NiN2M3NDZiMTgyMGJkMDljNjY1MzVmNDQiLCJwIjoiaiJ9)

This PR integrates Microsoft Presidio to optionally enable users to redact Personally Identifiable Information (PII) from Weave traces before the data is transmitted to the Weave trace servers.

**Note**: Enabling the new PII redaction feature will result in increased processing times due to the analysis performed by Presidio.

## Key Changes:

**New Dependencies**: Introduces presidio-analyzer and presidio-anonymizer.

**PII Redaction Setting**: Adds a `redact_pii` UserSetting, which can be configured during the weave.init call to enable PII redaction.

**Customizable PII Fields**: Provides a default list of PII fields for redaction, which can be customized using the `redact_pii_fields` UserSetting.

**Telemetry with Sentry**: Implements Sentry to capture telemetry data related to the usage of this feature. Sentry was chosen due to the absence of existing telemetry libraries in Weave, and its event system provides the necessary data insights.

This is my first dive into the Weave SDK and I am open to any feedback related to how these changes could be better implemented. 

## Testing
How was this PR tested?

- New unit tests

## Summary by CodeRabbit

- **New Features**
  - Introduced automated redaction to mask sensitive personal information, ensuring enhanced data privacy.
  - Added configurable options that let users enable redaction and customize which data fields are processed.
  - Integrated privacy controls into client operations to conditionally apply redaction based on user settings.
  - Implemented event tracking to monitor when privacy features are activated.

- **Chores**
  - Updated project configuration with new contributor metadata and an optional dependency group to support these privacy enhancements.


weave/trace/settings.py
```diff
@@ -61,6 +61,26 @@ class UserSettings(BaseModel):
...
+    """Toggles PII redaction using Microsoft Presidio.
+
+    If True, redacts PII from trace data before sending to the server.
+    Can be overrided with the environment variable `WEAVE_REDACT_PII`
+
+    WARNING: PII redaction is an experimental feature, and may not always work.
+    It will also introduce significant performance overhead when redacting large payloads.
+    """
+
+    redact_pii_fields: list[str] = []
```
andrewtruong, line 74: is there an env var for this?  Did you test it?

weave/trace/weave_client.py
```diff
@@ -965,11 +967,17 @@ def create_call(
         unbound_op = maybe_unbind_method(op)
         op_def_ref = self._save_op(unbound_op)
 
-        inputs_redacted = redact_sensitive_keys(inputs)
+        inputs_sensitive_keys_redacted = redact_sensitive_keys(inputs)
+
+        if should_redact_pii():
+            prepared_inputs = redact_pii(inputs_sensitive_keys_redacted)
```
andrewtruong, line 973: why did you change the name?  I think redacted is more descriptive here

andrewtruong (commented):
I think you're also missing redaction in:
1. The serialize path
2. The code capture path

niall-wb:

J2-D2-3PO:
Draft docs are available here https://github.com/wandb/weave/pull/3618, please tag me when this PR is ready to ship and we can get the docs out (once they are SME-approved, of course)

weave/trace/weave_init.py
```diff
@@ -125,6 +126,18 @@ def init_weave(
     autopatch.autopatch(autopatch_settings)
 
     username = get_username()
+
+    # This is a temporary event to track the number of users who have enabled PII redacting.
+    if should_redact_pii():
```
tssweeney, line 131: This sort of event logging should probably be centralized and/or closer to the implementation of the feature

weave/trace/settings.py
```diff
@@ -61,6 +61,26 @@ class UserSettings(BaseModel):
...
+    """List of fields to redact.
+
+    If redact_pii is True, this list of fields will be redacted.
```
tssweeney, line 77: please add this to env var docs

weave/trace/weave_client.py
```diff
@@ -965,11 +967,17 @@ def create_call(
         unbound_op = maybe_unbind_method(op)
         op_def_ref = self._save_op(unbound_op)
 
-        inputs_redacted = redact_sensitive_keys(inputs)
+        inputs_sensitive_keys_redacted = redact_sensitive_keys(inputs)
```
tssweeney, line 970: `

weave/trace/weave_client.py
tssweeney, line 973: if this is heavy, it should be backgrounded by putting it into `send_start_call`

weave/trace/settings.py
niall-wb: Yes and yes, it works.

weave/trace/weave_init.py
niall-wb: I don't want to log a unique event every time a call to `redact_pii` is made. I just want to log an event whenever someone calls `weave.init` with this feature turned on. Is there a better place to put this?

niall-wb:
> I think you're also missing redaction in:
> 
> 1. The serialize path
> 2. The code capture path

For the serialization path, I am redacting the inputs and outputs from calls after they are serialized. 

For the code capture path, I'm not sure I understand why that would be necessary. How could potential user PII ever get included in the code?

weave/trace/settings.py
niall-wb: I'm having trouble finding the specific env var docs. That said, we have added documentation for this feature in [this pr](https://github.com/wandb/weave/pull/3618/files).

weave/trace/weave_init.py
andrewtruong: You can just log once when you call `redact_pii`.  On the first call log, else no-op

weave/trace/weave_client.py
```diff
@@ -1080,7 +1082,15 @@ def create_call(
         project_id = self._project_id()
 
         def send_start_call() -> None:
-            inputs_json = to_json(inputs_with_refs, project_id, self, use_dictify=False)
+            maybe_redacted_inputs_with_refs = inputs_with_refs
+            if should_redact_pii():
+                from weave.trace.pii_redaction import redact_pii
```
andrewtruong, line 1087: does this import need to be deferred?

weave/trace/weave_client.py
```diff
@@ -1080,7 +1082,15 @@ def create_call(
...
+
```
andrewtruong, line 1088: e.g. you could implement the log-once check here

weave/trace/weave_client.py
niall-wb: Yes, because redact_pii uses Presidio, which is an optional dependency that most users will only have when should_redact_pii is enabled

niall-wb:
> > For the code capture path, I'm not sure I understand why that would be necessary. How could potential user PII ever get included in the code?
> 
> For example, if the user's code was:
> 
> ```python
> openai_key = "..."
> 
> client = openai.OpenAI(api_key=openai_key)
> 
> @weave.op
> def do_something():
>     client.something(...)
> ```
> 
> Then the key could be captured in the code

@andrewtruong This feature is specifically for redacting PII data, such as names, emails, and SSNs. It wouldn’t redact API keys anyway. Handling API keys or other secrets in captured code is out of scope, and PII wouldn’t typically be hardcoded in our customers’ code.

pyproject.toml
```diff
@@ -36,15 +37,15 @@ dynamic = ["version"]
...
-  "diskcache==5.6.3",        # Used for data caching
+  "packaging>=21.0",              # For version parsing in integrations
+  "tenacity>=8.3.0,!=8.4.0",      # Excluding 8.4.0 because it had a bug on import of AsyncRetrying
+  "emoji>=2.12.1",                # For emoji shortcode support in Feedback
+  "uuid-utils>=0.9.0",            # Used for ID generation - remove once python's built-in uuid supports UUIDv7
+  "numpy>1.21.0",                 # Used in box.py and scorer.py (should be made optional)
+  "rich",                         # Used for special formatting of tables (should be made optional)
+  "gql[aiohttp,requests]",        # Used exclusively in wandb_api.py
+  "jsonschema>=4.23.0",           # Used by scorers for field validation
+  "diskcache==5.6.3",             # Used for data caching
```
andrewtruong, line 48: revert

weave/trace/settings.py
andrewtruong: Please specify the env var in the doc similar to here: https://github.com/wandb/weave/pull/3406/files#diff-eecf2dc4cef8a4d9f04ade3573762e9ab65230eb152d1b0669febf31ace4421cL58

weave/trace/trace_sentry.py
```diff
@@ -210,6 +210,27 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+            "message": event_name,
+            "level": "info",
+            "tags": tags or {},
+            "user": {
+                "username": username,
+            },
+        }
+
+        self.hub.capture_event(event_data)
+
```
adrnswanberg, line 233: if this is for analytics shouldn't we use segment? I don't know of other examples where we use sentry for analytics events.

andrewtruong (approved):

weave/trace/trace_sentry.py
niall-wb: Good question! I was told that there isn't any existing telemetry or analytics within the weave sdk, aside from sentry, but I've heard that something is "coming soon." I was also asked not to introduce any new deps. I'm using sentry as as a temporary fix until the new telemetry is implemented, though I don't have context on its timeline or specifics. Our goal is to track the number of users and projects using this new feature, and this allows us to gather that info, while anticipating a more robust solution soon. It's not ideal, but it meets our current needs.

adrnswanberg: is this usable downstream though? AFAIK there is no way to connect these sentry events to mixpanel or anything else we typically use.

niall-wb: No, my plan was just to query them from sentry's discover tab and export them as we need them until we I'm able to use something better. I'd be much happier with segment.

PR #3410, opened 2025-01-14
feat(weave): track user on object create

gtarpenning:
## Description

https://wandb.atlassian.net/browse/WB-22663

Track user on object create. Also read user id from the db. 

## Testing

Updates tests

### DIFF
--- a/weave/trace_server/migrations/010_add_user_to_objects.up.sql
+++ b/weave/trace_server/migrations/010_add_user_to_objects.up.sql
@@ -1,0 +1,48 @@
+ALTER TABLE object_versions
+    ADD COLUMN wb_user_id Nullable(String) DEFAULT NULL;
+
+CREATE OR REPLACE VIEW object_versions_deduped as
+    SELECT project_id,
+        object_id,
+        created_at,
+        deleted_at,
+        kind,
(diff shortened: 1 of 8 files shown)


jwlee64 (approved):
lgtm

weave/trace_server/migrations/010_add_user_to_objects.up.sql
```diff
@@ -0,0 +1,48 @@
+ALTER TABLE object_versions
+    ADD COLUMN wb_user_id Nullable(String) DEFAULT NULL;
+
+CREATE OR REPLACE VIEW object_versions_deduped as
```
tssweeney, line 4: object_versions_deduped is no longer used. In fact it should be dropped

tssweeney (commented):
I think you can remove the object_versions_deduped. Also keep in mind that this requires a core change as well.

gtarpenning:
> I think you can remove the object_versions_deduped. Also keep in mind that this requires a core change as well.

You just mean this 1-liner right? https://github.com/wandb/core/pull/26825 @tssweeney

tssweeney (approved):

PR #3429, opened 2025-01-16
chore(weave): Make PIL images threadsafe

tssweeney:
See comments in code:

```
This file exposes a function that makes the PIL ImageFile thread-safe: `make_image_file_thread_safe`.
There is a discussion here: https://github.com/python-pillow/Pillow/issues/4848#issuecomment-671339193 in which
the author claims that the Pillow library is thread-safe. However, my reasoning leads me to a different conclusion.
Specifically, the `ImageFile.load` method is not thread-safe. This is because `load` will both close and delete
an open file handler as well as modify properties of the ImageFile object (namely the `im` property which contains
the underlying image data). Inside of Weave we use threads to parallelize work which may involve Images. This bug
has presented itself not only in our own persistence layer, but also in user code where they are consuming loaded
images across threads.
We call `make_image_file_thread_safe` in the `__init__.py` file to ensure that the ImageFile class is thread-safe.
```

### DIFF
--- a/tests/trace/test_image_thread_safety.py
+++ b/tests/trace/test_image_thread_safety.py
@@ -1,0 +1,56 @@
+import os
+import random
+from concurrent.futures import ThreadPoolExecutor
+
+import PIL
+import pytest
+
+
+@pytest.fixture()
+def deferred_cleanup():
+    files_to_cleanup = []
+
+    def add_file(name: str):
+        files_to_cleanup.append(name)
+
+    def cleanup():
+        for name in files_to_cleanup:
+            os.remove(name)
+
+    try:
+        yield add_file
+    finally:
+        cleanup()
+
+
+# Repeat 5 times since we are trying to test for thread safety
+@pytest.mark.parametrize("i", range(5))
+def test_image_thread_safety(deferred_cleanup, i):
+    file_name = f"temp_test_{i}.png"
+
+    # 1. Create a random in-memory image
+    image = PIL.Image.new("RGB", (1024, 1024))
+    image.putdata(
+        [
+            (
+                int(255 * random.random()),
+                int(255 * random.random()),
+                int(255 * random.random()),
+            )
+            for _ in range(1024 * 1024)
+        ]
+    )
+
+    # 2. Save the image to a file
+    deferred_cleanup(file_name)  # Cleanup the file on exit
+    image.save(file_name)
+
+    # 3. Open the image in multiple threads and load it
+    # This simulates what can happen when users are working
+    # with images across threads.
+    image = PIL.Image.open(file_name)
+    with ThreadPoolExecutor(max_workers=10) as executor:
+        jobs = [executor.submit(image.load) for i in range(10)]
+
+    # Wait for all the threads to finish
+    res = [job.result() for job in jobs]
--- a/tests/trace/test_image_thread_safety.py
+++ b/tests/trace/test_image_thread_safety.py
@@ -1,56 +1,38 @@
-import os
 import random
+import tempfile
 from concurrent.futures import ThreadPoolExecutor
 
 import PIL
 import pytest
 
 
-@pytest.fixture()
-def deferred_cleanup():
-    files_to_cleanup = []
-
-    def add_file(name: str):
-        files_to_cleanup.append(name)
-
-    def cleanup():
-        for name in files_to_cleanup:
-            os.remove(name)
-
-    try:
-        yield add_file
-    finally:
-        cleanup()
-
-
 # Repeat 5 times since we are trying to test for thread safety
 @pytest.mark.parametrize("i", range(5))
-def test_image_thread_safety(deferred_cleanup, i):
-    file_name = f"temp_test_{i}.png"
+def test_image_thread_safety(i):
+    # Create a temporary file that will be automatically cleaned up
+    with tempfile.NamedTemporaryFile(suffix=".png") as temp_file:
+        # 1. Create a random in-memory image
+        image = PIL.Image.new("RGB", (1024, 1024))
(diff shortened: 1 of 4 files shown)


weave/initialization/pil_image_thread_safety.py
```diff
@@ -0,0 +1,52 @@
...
+import threading
+from functools import wraps
+
+_patch_lock = threading.Lock()
+_patched = False
+
+
+def make_image_file_thread_safe() -> None:
+    global _patched
+    with _patch_lock:
```
tssweeney, line 25: this might be overkill.

tests/trace/test_image_thread_safety.py
```diff
@@ -0,0 +1,56 @@
...
+        files_to_cleanup.append(name)
+
+    def cleanup():
+        for name in files_to_cleanup:
+            os.remove(name)
+
+    try:
+        yield add_file
+    finally:
+        cleanup()
```
andrewtruong, line 23: You can use TemporaryFile for this:
https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

weave/initialization/pil_image_thread_safety.py
andrewtruong, line 25: This with statement won't do anything since you return immediately -- remove

weave/initialization/pil_image_thread_safety.py
```diff
@@ -0,0 +1,52 @@
...
+        if _patched:
+            return
+        _patched = True
```
andrewtruong, line 28: you should move this after the try except block, probably as:
```py
try:
    _make_image_file_thread_safe()
except Exception as e:
    print(f"Failed to patch PIL.ImageFile.ImageFile: {e}.")
else:
    _patched = True
```

andrewtruong: Unless `_patched` here is meant to be "attempted patch"

weave/initialization/pil_image_thread_safety.py
```diff
@@ -0,0 +1,52 @@
...
+
+def _make_image_file_thread_safe() -> None:
+    from PIL.ImageFile import ImageFile
+
+    old_load = ImageFile.load
+    old_init = ImageFile.__init__
+
+    @wraps(old_init)
+    def new_init(self, *args, **kwargs):  # type: ignore
+        self._load_lock = threading.Lock()
```
andrewtruong, line 43: since we are patching this, it might be safer to make this namespaced as `self._weave_load_lock`

weave/initialization/pil_image_thread_safety.py
tssweeney: yup - fixing

weave/__init__.py
```diff
@@ -14,9 +14,14 @@
 from weave.flow.model import Model
 from weave.flow.obj import Object
 from weave.flow.prompt.prompt import EasyPrompt, MessagesPrompt, Prompt, StringPrompt
+from weave.initialization.pil_image_thread_safety import (
+    make_image_file_thread_safe as make_image_file_thread_safe,
+)
 from weave.trace.util import Thread as Thread
 from weave.trace.util import ThreadPoolExecutor as ThreadPoolExecutor
 
+make_image_file_thread_safe()
```
andrewtruong, line 23: I would put this in `image.py` instead of here

weave/initialization/pil_image_thread_safety.py
tssweeney: changed

andrewtruong (approved):
Just a few nits, otherwise lgtm

weave/__init__.py
tssweeney: If my goal is to ensure this is always ran, your proposal feels a bit implicit and exposed to unintended breakage:
`__init__.py -> from weave.trace.api import *`
`weave/trace/api.py -> from weave import type_handlers`
`weave.type_handlers -> from weave.type_handlers.Image import image`
`weave/type_handlers/Image/image.py -> runs script`

andrewtruong:
Since we're doing patching here, you may also want to include an "unpatch" func just in case.

tests/trace/test_image_thread_safety.py
tssweeney: ack

weave/__init__.py
andrewtruong: Sorry, can you clarify?  I don't understand the concern here

If folks want to use Images (which depend on PIL), they will have to go through `image.py` anyways

tssweeney: 1. I am operating under the assumption that we want this code to always run no matter what.

  - It should not require the user to explicitly patch
  - It should run even if we did nothing special to handle images (this bug can occur just in our eval loop)

2. Putting it in `image.py` means that the import chain described above must be maintained else it will no longer be ran at import

andrewtruong: It's not the end of the world to put it in `__init__.py`, but consider:
1. When would a user want the combination of wanting to use PIL images with weave, but not our PIL type handler?
2. The top level namespace is now noisier with `make_image_file_thread_safe` in it
3. Things related to images are not spatially close in the package

andrewtruong: Alternatively, consider putting the setup code in the `initialization` module and explicitly don't export `make_image_file_thread_safe`... so roughly:


```py
# weave/__init__.py
from weave.initialization import *

# weave/initialization/__init__.py
from pil_image_thread_safety import make_image_file_thread_safe

make_image_file_thread_safe()

__all__ = []
```

tssweeney: ok, taking the last suggestion here

PR #3438, opened 2025-01-17
chore(weave): Re-enable to_dict support in serialization layer

tssweeney:
With dictify=false on inputs and outputs, we lost the ability to respect to_dict. This patches in a little hack to put that back

### DIFF
--- a/weave/trace/serialization/dictifiable.py
+++ b/weave/trace/serialization/dictifiable.py
@@ -1,4 +1,4 @@
-from typing import Any, Protocol, TypeVar, runtime_checkable
+from typing import Any, Protocol, runtime_checkable
 
 
 @runtime_checkable
@@ -9,14 +9,14 @@ class Dictifiable(Protocol):
         """Convert the object to a dictionary representation."""
         ...
 
-T = TypeVar('T')
-def try_to_dict(obj: T) -> dict[str, Any] | None:
+
+def try_to_dict(obj: Any) -> dict[str, Any] | None:
     """
     Attempt to convert an object to a dictionary using the Dictifiable protocol.
-    
+
     Args:
         obj: Object to attempt to convert to dictionary
-        
+
     Returns:
         Dictionary representation if object implements Dictifiable protocol, None otherwise
     """
--- a/weave/trace/serialization/dictifiable.py
+++ b/weave/trace/serialization/dictifiable.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import Any, Protocol, runtime_checkable
 
 
--- a/weave/trace/serialization/dictifiable.py
+++ b/weave/trace/serialization/dictifiable.py
@@ -1,0 +1,30 @@
+from typing import Any, Protocol, TypeVar, runtime_checkable
+
+
+@runtime_checkable
+class Dictifiable(Protocol):
+    """Protocol for objects that can be converted to a dictionary representation."""
+
+    def to_dict(self) -> dict[str, Any]:
+        """Convert the object to a dictionary representation."""
+        ...
+
+T = TypeVar('T')
+def try_to_dict(obj: T) -> dict[str, Any] | None:
+    """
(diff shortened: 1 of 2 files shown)


weave/trace/serialize.py
```diff
@@ -57,6 +58,14 @@ def to_json(
             and not has_custom_repr(obj)
         ):
             return dictify(obj)
+
+        # TODO: I would prefer to only have this once in dictify? Maybe dictify and to_json need to be merged?
```
jamie-rasmussen, line 62: I think it is mostly confusing we have the function `dictify` but conceptually `try_to_dict` is also a "dictify" of sorts. Maybe we could change `use_dictify` to something non-boolean, like "dictify_level" or "dictify_strategy".

weave/trace/serialization/dictifiable.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any, Protocol, runtime_checkable
+
+
+@runtime_checkable
+class Dictifiable(Protocol):
+    """Protocol for objects that can be converted to a dictionary representation."""
+
+    def to_dict(self) -> dict[str, Any]:
```
jamie-rasmussen, line 8: I believe the return type here is misleading and the isinstance check below would pass even if the to_dict doesn't return a dict or returns one with non-string keys.

jamie-rasmussen (approved):

weave/trace/serialization/dictifiable.py
andrewtruong: The runtime will allow it but typecheckers will complain

weave/trace/serialize.py
andrewtruong: Conceptually I understand this as:
1. Functions may implement their own `to_dict`, which makes them `Dictifiable`
2. The current `dictify` is more of a "to_dict_of_last_resort" for objects that are not `Dictifiable`
3. To compose, rename the current `dictify->_to_dict_fallback`, then have a new func:
```py
def dictify(obj: Any) -> dict[str, Any]:
    if isinstance(obj, Dictifiable):
        return obj.to_dict()
    return _to_dict_fallback(obj)
```

weave/trace/serialization/dictifiable.py
jamie-rasmussen: In our serialization use case I don't think we need to require the keys are strings anyway.

weave/trace/serialize.py
jamie-rasmussen: 1. Yes
2. Yes, though it also handles the Dictifiable case.
3. The issue is the meaning of the `use_dictify` parameter. Tim is proposing we do `to_dict` even when use_dictify is false, a condition in which we wouldn't want to call the `_to_dict_fallback`

weave/trace/serialization/dictifiable.py
tssweeney: I think there is another subtlety here that might be useful to be stricter about:
* `dict` when used in `to_dict` or `from_dict` type methods refers specifically to a python dictionary where the keys are strings (or strings | int) and the values are Any. By this definition, it is only a top-level dict - leaves can be anything. This is most akin to pydantic's `model_dump`: https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump
* `json` could be similar, but specifically the output is a valid json string, and of course all leave nodes must be valid json primitives. Notebly, this needn't be a top-level dict. most similar to pydantic's `model_dump_json`: https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump_json

There is sort of an intermediate concept that we don't have a name for, call it: `portable_dict` which is the same as `dict` except that:
1. the top level can be a primitive or list
2. all leaves are primitives (it is ready to be jsoned)
3. any dict in the tree could have a `_type` key that is used to dispatch to a deserializer.

The current `dictify` is sort of like a recursive `dict` where leaves are expected to be primitives as well.

I feel like nailing this down (or a version of this) would be helpful in designing a clearer system here.

PR #3448, opened 2025-01-21
Add new FAQ to Calls page

tssweeney:
Answers:
* How do I disable tracing?
* How do I capture information about a call?

### DIFF
--- a/docs/docs/guides/tracking/tracing.mdx
+++ b/docs/docs/guides/tracking/tracing.mdx
@@ -5,6 +5,7 @@ import TracingCallsMacroImage from '@site/static/img/screenshots/calls_macro.png
 import TracingCallsFilterImage from '@site/static/img/screenshots/calls_filter.png';
 import BasicCallImage from '@site/static/img/screenshots/basic_call.png';
 
+
 # Calls
 
 <DesktopWindow 
@@ -699,17 +700,7 @@ The easiest way to get started is to construct a view in the UI, then learn more
 :::info[Comming Soon]
 ::: */}
 
-## Call FAQ
-
-{/* TODO:
-Common Questions / Variations:
-* Images
-* Ops
-* Cost?
-* General data model
-*/}
-
-#### Call Schema
+### Call Schema
 
 Please see the [schema](../../reference/python-sdk/weave/trace_server/weave.trace_server.trace_server_interface#class-callschema) for a complete list of fields.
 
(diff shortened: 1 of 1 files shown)


docs/docs/guides/tracking/tracing.mdx
```diff
@@ -748,4 +739,64 @@ Calculated Fields:
...
+
+### How do I disable tracing?
+
+**Environment Variable:**
+
+In situations where you want to unconditionally disable tracing for the entire program, you can set the environment variable `WEAVE_TRACING_ENABLED` to `false`.
+
+**Client Initialization:**
+
+Sometimes, you may want to conditionally enable tracing for a specific initialization based on some condition. In this case, you can initialize the client with the `tracing_enabled` flag.
```
jamie-rasmussen, line 752: I don't think this is correct. Isn't it `weave.init(... settings={"disabled": True})`

docs/docs/guides/tracking/tracing.mdx
```diff
@@ -748,4 +739,64 @@ Calculated Fields:
     * Duration
     * Status
 
+## FAQs
+
+### How do I disable tracing?
+
+**Environment Variable:**
+
+In situations where you want to unconditionally disable tracing for the entire program, you can set the environment variable `WEAVE_TRACING_ENABLED` to `false`.
```
jamie-rasmussen, line 748: I don't think this is correct, should be `WEAVE_DISABLED=true`

docs/docs/guides/tracking/tracing.mdx
```diff
@@ -748,4 +739,64 @@ Calculated Fields:
...
+```python
+import weave
+
+# Initialize the client
+client = weave.init(..., tracing_enabled=False)
+```
+
+**Context Manager:**
+
+Finally, you may want to conditionally disable tracing for a single function based on some application logic. In this case, you can use the context manager `with set_tracing_enabled(False)` which can be import from `weave.trace.context.call_context`.
```
jamie-rasmussen, line 763: import => imported

docs/docs/guides/tracking/tracing.mdx
tssweeney: 🤦 thank you

docs/docs/guides/tracking/tracing.mdx
tssweeney: 🤦 thank you

jamie-rasmussen (approved):

PR #3453, opened 2025-01-21
docs(weave): Playground docs copy consistency

J2-D2-3PO:
- Fixes https://wandb.atlassian.net/browse/DOCS-1210

Updates copy on Playground page so as to not confuse readers about available providers


J2-D2-3PO:
@johndmulhausen I thought I was able to merge without additional approval, but I guess not, could you approve?

PR #3454, opened 2025-01-22
chore(weave): improve error handling to raise with extra info

gtarpenning:
## Description

[WB-22605](https://wandb.atlassian.net/browse/WB-22605)

This pr:
- adds a special database exception that gets handled in the server (pr here)
- use the helper functions whenever possible

[WB-22605]: https://wandb.atlassian.net/browse/WB-22605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/errors.py
```diff
@@ -51,3 +52,22 @@ class ObjectDeletedError(Error):
     def __init__(self, message: str, deleted_at: datetime.datetime):
         self.deleted_at = deleted_at
         super().__init__(message)
+
+
+class ClickhouseQueryError(Error):
+    """Raised when a query to Clickhouse fails."""
+
+    def __init__(
```
tssweeney, line 60: you might want to include an optional user-safe message field that can be passed along.
you might want to include a unique query id

PR #3457, opened 2025-01-22
chore(weave): Improve Weave patch to support pre-constructed images

tssweeney:
Modifies the patching logic such that order of construction is irrelevant. Currently this will fail:

```python
import PIL
image = PIL.Image.new("RGB", (10, 10))
with NamedTemporaryFile(suffix=".png") as f:
    image.save(f.name)
    image = PIL.Image.open(f.name)
image.crop((0, 0, 10, 10)) # <---- Attribute error on `_weave_load_lock`
```

This modified implementation uses a `TheadSafeLockLookup` to ensure that the `_weave_load_lock` is always available for any object requesting it and only minted once per object.

### DIFF
--- a/weave/initialization/pil_image_thread_safety.py
+++ b/weave/initialization/pil_image_thread_safety.py
@@ -20,10 +20,14 @@ import threading
 from functools import wraps
 from typing import Any, Callable, Optional
 
+# Global state
+# `_patched` is a boolean that indicates whether the thread-safety patch has been applied
+# `_original_methods` is a dictionary that stores the original methods of the ImageFile class
+# `_new_lock_lock` is a lock that is used to create a new lock for each ImageFile instance
+# `_fallback_load_lock` is a global lock that is used to ensure thread-safe image loading when per-instance locking fails
 _patched = False
 _original_methods: dict[str, Optional[Callable]] = {"load": None}
 _new_lock_lock = threading.Lock()
-# Global fallback lock for thread-safe image loading when per-instance locking fails
 _fallback_load_lock = threading.Lock()
 
 
--- a/weave/initialization/pil_image_thread_safety.py
+++ b/weave/initialization/pil_image_thread_safety.py
@@ -18,28 +18,9 @@ We call `apply_threadsafe_patch_to_pil_image` in the `__init__.py` file to ensur
 import threading
 from functools import wraps
 from typing import Any, Callable, Optional
-from weakref import WeakKeyDictionary
 
 _patched = False
 _original_methods: dict[str, Optional[Callable]] = {"load": None}
-
-
-class TheadSafeLockLookup:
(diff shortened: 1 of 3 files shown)


weave/initialization/pil_image_thread_safety.py
```diff
@@ -4,27 +4,39 @@
...
 """
 
 import threading
 from functools import wraps
 from typing import Any, Callable, Optional
 
+# Global state
+# `_patched` is a boolean that indicates whether the thread-safety patch has been applied
+# `_original_methods` is a dictionary that stores the original methods of the ImageFile class
+# `_new_lock_lock` is a lock that is used to create a new lock for each ImageFile instance
```
andrewtruong, line 26: is this supposed to be `_new_load_lock`?

weave/initialization/pil_image_thread_safety.py
```diff
@@ -4,27 +4,39 @@
...
+# `_fallback_load_lock` is a global lock that is used to ensure thread-safe image loading when per-instance locking fails
 _patched = False
-_original_methods: dict[str, Optional[Callable]] = {"init": None, "load": None}
+_original_methods: dict[str, Optional[Callable]] = {"load": None}
+_new_lock_lock = threading.Lock()
```
andrewtruong, line 30: same

weave/initialization/pil_image_thread_safety.py
```diff
@@ -41,35 +53,61 @@ def apply_threadsafe_patch_to_pil_image() -> None:
...
+            # Note: this `_new_lock_lock` is global as opposed to per-instance, else
+            # it would be possible for the same ImageFile to be loaded by multiple threads
+            # thereby creating a race where different threads would be each minting their
+            # own lock for the same ImageFile!
+            if not hasattr(self, "_weave_load_lock"):
+                with _new_lock_lock:
+                    # Double-check pattern: verify the attribute still doesn't exist
+                    # after acquiring the lock to prevent race conditions
+                    if not hasattr(self, "_weave_load_lock"):
+                        setattr(self, "_weave_load_lock", threading.Lock())
```
andrewtruong, line 89: Can you not just do `self._weave_load_lock = threading.Lock()` here?

andrewtruong (approved):

weave/initialization/pil_image_thread_safety.py
tssweeney: no - this is the "lock" on creating a new lock

weave/initialization/pil_image_thread_safety.py
tssweeney: I liked using setattr here to have consistency with the other built `*attr` usages in this section. Particularly because `_weave_load_lock` is not a typed attribute of self

PR #3461, opened 2025-01-22
chore(weave): Adds test for recent dictify change

tssweeney:
I had stashed this test and forgot to actually check it in. It passes (:

### DIFF
--- a/tests/trace/test_dictifiable.py
+++ b/tests/trace/test_dictifiable.py
@@ -36,6 +36,10 @@ def test_dictifiable(client):
     call = func.calls()[0]
 
     assert call.inputs["d"] == {"attr": val}
-    assert call.inputs["nd"].startswith('<test_dictifiable.test_dictifiable.<locals>.NonDictifiable object at')
+    assert call.inputs["nd"].startswith(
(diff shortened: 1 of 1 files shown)


tests/trace/test_dictifiable.py
```diff
@@ -0,0 +1,45 @@
...
+        attr: int
+
+        def __init__(self, attr: int):
+            self.attr = attr
+
+        def to_dict(self):
+            return {"attr": self.attr}
+
+    @weave.op
+    def func(d: Dictifiable, nd: NonDictifiable) -> Dictifiable:
```
jamie-rasmussen, line 21: This return type seems wrong, it's a dict, not a Dictifiable?

tssweeney: fixed - i originally had it as dictifiable, but wanted to make sure both worked.

jamie-rasmussen (approved):

tests/trace/test_dictifiable.py
jamie-rasmussen: I see we exclude the `tests` dir from `mypy`

PR #3464, opened 2025-01-22
chore(ui): show object version timestamp in header metadata

jamie-rasmussen:
## Description

Adds the created timestamp to the metadata displayed for an object version.

Before:
Screenshot 2025-01-22 at 1 32 19 PM

After:
Screenshot 2025-01-22 at 1 32 09 PM


gtarpenning (approved):
Could to the same for ops too

PR #3467, opened 2025-01-22
docs(weave): Add troubleshooting page

J2-D2-3PO:
## Description

- https://wandb.atlassian.net/browse/DOCS-1186

Adds Troubleshooting page to the docs, with first tip from [this slack convo](https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1736846808980349).

We don't currently have a great place to put this page without a larger menu reorg, so slotting into Product Walkthrough for now

## Testing

- `yarn start` on local

### DIFF
--- a/docs/docs/guides/troubleshooting.md
+++ b/docs/docs/guides/troubleshooting.md
@@ -1,0 +1,11 @@
+# Troubleshooting 
+
+This page provides solutions and guidance for common issues you may encounter. As we continue to expand this guide, more troubleshooting topics will be added to address a broader range of scenarios.
+
+:::tip
+Do you have Weave troubleshooting advice to share with the community?  Click **Edit this page** at the bottom of this guide to contribute directly by submitting a pull request.
+:::
+
+## Reduce the number of trace table rows
+
+To reduce the number of trace table rows retrieved by a query, set the `pageSize` query parameter to a value less than the default value of `100`. 
--- a/docs/docs/guides/troubleshooting.md
+++ b/docs/docs/guides/troubleshooting.md
@@ -8,12 +8,12 @@ Do you have Weave troubleshooting advice to share with the community? Click **Ed
 
 ## Trace pages load slowly
 
-If trace pages are loading slowly, reduce the number of rows displayed to improve load time. You can either reduce the number of rows via the UI, or using query parameters.
+If trace pages are loading slowly, reduce the number of rows displayed to improve load time. The default value is `50`. You can either reduce the number of rows via the UI, or using query parameters.
 
 ### Adjust via the UI (recommended)
 
-Use the **Per page** control at the bottom-right of the Traces page to adjust the number of rows displayed. The default value is `50`, but you can also set to `10`, `25`, or `100`.
+Use the **Per page** control at the bottom-right of the Traces page to adjust the number of rows displayed. In addition to the default of `50`, you can also set to `10`, `25`, or `100`.
 
 ### Use query parameters 
 
-If you prefer a manual approach, you can modify the `pageSize` query parameter in your query URL to a value less than the default of `50`.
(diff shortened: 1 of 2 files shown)


docs/docs/guides/troubleshooting.md
```diff
@@ -0,0 +1,11 @@
...
+
+This page provides solutions and guidance for common issues you may encounter. As we continue to expand this guide, more troubleshooting topics will be added to address a broader range of scenarios.
+
+:::tip
+Do you have Weave troubleshooting advice to share with the community?  Click **Edit this page** at the bottom of this guide to contribute directly by submitting a pull request.
+:::
+
+## Reduce the number of trace table rows
+
+To reduce the number of trace table rows retrieved by a query, set the `pageSize` query parameter to a value less than the default value of `100`.
```
jamie-rasmussen, line 11: I just changed the default per page to 50 and added a UI control for this at the bottom right, which we should probably encourage people to use instead of query parameter hacking.

I think this is also missing the context that this is to deal with trace pages that are slow loading.

J2-D2-3PO: Ah that makes alot more sense, I didn't have that context. Updated the page so that:

- Section header reflects context
- Section describes UI option as recommended solution 
- Section still documents query param approach in case someone wants to use it. Maybe if they need more granular controls than the UI picklist?

J2-D2-3PO: @jamie-rasmussen here's the updated look https://github.com/wandb/weave/pull/3467/files

docs/docs/guides/troubleshooting.md
```diff
@@ -0,0 +1,19 @@
...
+
+:::tip
+Do you have Weave troubleshooting advice to share with the community? Click **Edit this page** at the bottom of this guide to contribute directly by submitting a pull request.
+:::
+
+## Trace pages load slowly
+
+If trace pages are loading slowly, reduce the number of rows displayed to improve load time. You can either reduce the number of rows via the UI, or using query parameters.
+
+### Adjust via the UI (recommended)
```
jamie-rasmussen, line 13: I find it a little odd that our h2 and h3 render the same - do you know if that's intentional?

docs/docs/guides/troubleshooting.md
```diff
@@ -0,0 +1,19 @@
...
+
+Use the **Per page** control at the bottom-right of the Traces page to adjust the number of rows displayed. The default value is `50`, but you can also set to `10`, `25`, or `100`.
+
+### Use query parameters 
+
+If you prefer a manual approach, you can modify the `pageSize` query parameter in your query URL to a value less than the default of `50`.
```
jamie-rasmussen, line 19: Can actually be anything less than the max of 100

docs/docs/guides/troubleshooting.md
```diff
@@ -0,0 +1,19 @@
+# Troubleshooting
```
jamie-rasmussen, line 1: Need to link this in sidebar somewhere (PR description no longer accurate)

docs/docs/guides/troubleshooting.md
J2-D2-3PO: Yeah me too, I suspect that is a Docusaurus thing or custom style. Not sure we can do much about it in this PR

docs/docs/guides/troubleshooting.md
J2-D2-3PO: Got it, updated in https://github.com/wandb/weave/pull/3467/commits/3402a7e8103b5be888b6d8e18d541420099af116

docs/docs/guides/troubleshooting.md
J2-D2-3PO: 💀 Ah my miss, added https://github.com/wandb/weave/pull/3467/commits/114fd919d6335ca2385bcde454568648ab3d284a

jamie-rasmussen (approved):

PR #3469, opened 2025-01-22
chore(ui): dont do the cost query if no calls found

gtarpenning:
## Description

Micro perf improvement by not querying for costs when we have no calls to display.

in prod we actually query for costs twice even when there are no calls. 

## Testing

Prod:
Screenshot 2025-01-22 at 1 57 10 PM

Branch:
Screenshot 2025-01-22 at 1 57 42 PM


jamie-rasmussen (approved):

jwlee64 (approved):
niceee

PR #3470, opened 2025-01-22
chore(ui): debounce server requests while typing in filter bar

gtarpenning:
## Description

[WB-22888](https://wandb.atlassian.net/browse/WB-22888)

Debounce typing in the trace table filter bar. 

## Testing

Prod:
debounce-prod

Branch:
debounce-branch

[WB-22888]: https://wandb.atlassian.net/browse/WB-22888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
@@ -29,6 +29,8 @@ import {FilterTagItem} from './FilterTagItem';
 import {GroupedOption, SelectFieldOption} from './SelectField';
 import {VariableChildrenDisplay} from './VariableChildrenDisplayer';
 
+const DEBOUNCE_MS = 500;
+
 type FilterBarProps = {
   filterModel: GridFilterModel;
   setFilterModel: (newModel: GridFilterModel) => void;
@@ -60,7 +62,7 @@ export const FilterBar = ({
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
 
   // local filter model is used to avoid triggering a re-render of the trace
-  // table on every keystroke. debounced 300 ms
+  // table on every keystroke. debounced DEBOUNCE_MS ms
   const [localFilterModel, setLocalFilterModel] = useState(filterModel);
   useEffect(() => {
     setLocalFilterModel(filterModel);
@@ -169,7 +171,10 @@ export const FilterBar = ({
 
   const debouncedSetFilterModel = useMemo(
     () =>
-      _.debounce((newModel: GridFilterModel) => setFilterModel(newModel), 300),
+      _.debounce(
+        (newModel: GridFilterModel) => setFilterModel(newModel),
+        DEBOUNCE_MS
+      ),
     [setFilterModel]
   );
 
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
```diff
@@ -143,29 +152,36 @@ export const FilterBar = ({
...
-      setFilterModel(newModel);
+      setLocalFilterModel(newModel);
     },
-    [filterModel, setFilterModel]
+    [localFilterModel]
+  );
+
+  const debouncedSetFilterModel = useMemo(
+    () =>
+      _.debounce((newModel: GridFilterModel) => setFilterModel(newModel), 300),
```
gtarpenning, line 172: make constant?

tssweeney: yeah, and I would increase the rate to more like 1 sec

gtarpenning: that adds 1 second of latency to every user though, and 99% probably aren't running into this perf stuff. what about 500ms? @tssweeney

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
```diff
@@ -143,29 +152,36 @@ export const FilterBar = ({
...
     (item: GridFilterItem) => {
-      const oldItems = filterModel.items;
+      const oldItems = localFilterModel.items;
       const index = oldItems.findIndex(f => f.id === item.id);
 
       if (index === -1) {
-        const newModel2 = {...filterModel, items: [item]};
-        setFilterModel(newModel2);
+        const newModel2 = {...localFilterModel, items: [item]};
+        setLocalFilterModel(newModel2);
```
tssweeney, line 183: i would instead create a callback: `updateFilterModel` which internally sets the "local" working copy AND calls the debounced one. Seems weird to have the dual call here

tssweeney (approved):
Approved with comments

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
gtarpenning: I agree this feels wrong, but it would only ever be used in this one place, the other times we `setLocalFilterModel` we want to either _not_ update the original filter or directly set it instead of using the debounced. 

I will try to appease the style gods while not being verbose

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
tssweeney: i think 500 is fine, sure. I general this live-filter effect is pretty rare in most applications. Instead there is a moment of "applying" the filter b/c this is so expensive.

gtarpenning: yeah true, i'll get the featurebee and you can play with 500

PR #3471, opened 2025-01-22
fix(ui): remove expandColumns with get_calls

gtarpenning:
## Description

Fix for frontend export code gen. Fixes 2 cases: 
- no need for expandColumns (sdk resolves refs)
- when selected calls, extra `}`

## Testing

Screenshot 2025-01-22 at 3 53 19 PM


jwlee64 (approved):
Lgtm

PR #3475, opened 2025-01-23
chore(weave): Cleanup a mem leak in our ThreadExecutor

tssweeney:
Context:

After some deep diving into our data flow, I identified at least one mem leak in the `ContextAwareThreadExecutor`. I am not sure this is a huge one, but does have some impact.

Before:

Screenshot 2025-01-23 at 01 24 29

After:

Screenshot 2025-01-23 at 01 24 36

---

Summary:

This pull request includes changes to the `weave/trace/util.py` file to improve the handling of contexts in the `submit` and `map` methods. The most important changes include modifying how contexts are managed and passed to the `_run_with_context` method, ensuring better context handling and reducing complexity.

Context handling improvements:

* [`weave/trace/util.py`](diffhunk://#diff-bb390e76b5555c67339ae3a82e757d5c601d508aa7ba7b9db4aebfc04332a537L64-L92): Removed the `contexts` attribute from the class and updated the `submit` method to pass the context directly to the `super().submit` call.
* [`weave/trace/util.py`](diffhunk://#diff-bb390e76b5555c67339ae3a82e757d5c601d508aa7ba7b9db4aebfc04332a537L64-L92): Updated the `map` method to create contexts for each item in the first iterable and pass them along with the iterables to the `super().map` call.
* [`weave/trace/util.py`](diffhunk://#diff-bb390e76b5555c67339ae3a82e757d5c601d508aa7ba7b9db4aebfc04332a537L64-L92): Simplified the `_run_with_context` method by removing the conditional context assignment and directly using the passed context.

### DIFF
--- a/weave/trace/util.py
+++ b/weave/trace/util.py
@@ -61,35 +61,38 @@ class ContextAwareThreadPoolExecutor(_ThreadPoolExecutor):
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
-        self.contexts: list[Context] = []
 
     # ignoring type here for convenience because otherwise you have to write a bunch of overloads
     # for py310+ and py39-
     def submit(self, fn: Callable, *args: Any, **kwargs: Any) -> Any:  # type: ignore
-        context = copy_context()
-        self.contexts.append(context)
-
-        wrapped_fn = partial(self._run_with_context, context, fn)
-        return super().submit(wrapped_fn, *args, **kwargs)
+        wrapped_fn = partial(self._run_with_context, fn)
+        return super().submit(wrapped_fn, copy_context(), *args, **kwargs)
 
     def map(
         self,
         fn: Callable,
-        *iterables: Iterable[Iterable],
+        *iterables: Iterable[Any],
         timeout: float | None = None,
         chunksize: int = 1,
     ) -> Iterator:
-        contexts = [copy_context() for _ in range(len(list(iterables[0])))]
-        self.contexts.extend(contexts)
+        first_iterable = list(iterables[0])
+        map_len = len(first_iterable)
 
-        wrapped_fn = partial(self._run_with_context, None, fn)
-        return super().map(wrapped_fn, *iterables, timeout=timeout, chunksize=chunksize)
+        # Create a context for each item in the first iterable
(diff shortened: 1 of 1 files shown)


weave/trace/util.py
```diff
@@ -61,35 +61,38 @@ def _wrapped_fn(*args):
...
-        return super().map(wrapped_fn, *iterables, timeout=timeout, chunksize=chunksize)
+        return super().map(
+            wrapped_fn, *map_iterables, timeout=timeout, chunksize=chunksize
+        )
 
     def _run_with_context(
-        self, context: Context, fn: Callable, *args: Any, **kwargs: Any
+        self, fn: Callable, context: Context, *args: Any, **kwargs: Any
     ) -> Any:
-        if context is None:
```
jwlee64, line 91: did we ever use this null check?

jwlee64 (approved):

weave/trace/util.py
gtarpenning: i think that was just for this callsite `wrapped_fn = partial(self._run_with_context, None, fn)`

gtarpenning: or maybe unused

tssweeney: Correct, it was passed in from the map as none. There were two problems here:
1. this was not typed as none
2. we should not be pushing global state onto a stack and popping it in a child thread. at the very least ordering feels unstable here. Anyway, now there is no global state to worry about.

weave/trace/util.py
```diff
@@ -61,35 +61,38 @@ def _wrapped_fn(*args):
...
         self,
         fn: Callable,
-        *iterables: Iterable[Iterable],
+        *iterables: Iterable[Any],
         timeout: float | None = None,
         chunksize: int = 1,
     ) -> Iterator:
-        contexts = [copy_context() for _ in range(len(list(iterables[0])))]
-        self.contexts.extend(contexts)
+        first_iterable = list(iterables[0])
```
andrewtruong, line 78: this can crash if `iterables == []`

andrewtruong (approved):
I remember there was a reason why I had to do this weird list of contexts thing, but I can't remember why.  Maybe it's gone now?

Approving since the tests pass and this does seem to fix a mem leak issue

PR #3476, opened 2025-01-23
fix: filter out LiteLLM GenericAPILogger enterprise feature debug message

morganmcg1:
The LiteLLM library outputs a debug message about being unable to import GenericAPILogger when it's imported. This is an expected message for non-enterprise users and doesn't indicate any issues, but it can be confusing for users who see it in their logs.

This change adds a custom filter to the LiteLLM logger that suppresses only this specific debug message about the enterprise feature while allowing all other debug messages to be shown.

Test Plan:
1. Set logging level to DEBUG
2. Call weave.init()
3. Verify that the LiteLLM debug message about GenericAPILogger enterprise feature is suppressed
4. Verify that other debug messages are still shown

### DIFF
--- a/weave/trace/weave_init.py
+++ b/weave/trace/weave_init.py
@@ -7,8 +7,20 @@ from weave.trace.context import weave_client_context as weave_client_context
 from weave.trace_server import sqlite_trace_server
 from weave.trace_server_bindings import remote_http_trace_server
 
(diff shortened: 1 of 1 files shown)


morganmcg1:

morganmcg1:

morganmcg1:

PR #3480, opened 2025-01-23
Update "Open App" link to actually open Weave

johndmulhausen:
## Description

Maybe there's a better dashboard link that gets you right into the Weave UI, but the generic dashboard link doesn't feel specific enough. Is there a better idea?


jamie-rasmussen (approved):
This seems fine to me - we can't really drop you into the main Weave UI without knowing a project to use.

PR #2920, opened 2024-11-06
feat(weave): link artifact references in weave trace

estellazx:
## Description
https://wandb.atlassian.net/browse/WB-21724
Building on @levinandrew's [PR](url)

New loom with feedback:
https://www.loom.com/share/e25073c52f5c4f73bb523e043b1672ec?sid=6b3d6655-da3f-4645-9543-a04d54d0b224

~~old loom:~~
~~https://www.loom.com/share/056ae7450d064784b6c8b3a79dfcbc97?sid=1ed27f9c-d9fe-42a8-b2bc-0a93cb676a41~~ 

## Testing
locally. Note, you need to have a 
1. a team entity 
2. registries set up (all this requires is you to go to `http://localhost:9000/orgs/gr-org/registry/` for example and it will auto set up Registries of type model)
3. a personal entity to have model registries set up (this is not a must you can just comment out the part where we test model registry code)

Test code example:
```
from typing import List
import weave

import wandb
import os

os.environ["WANDB_BASE_URL"] = "http://localhost:9001"
os.environ["WF_TRACE_SERVER_URL"] = "http://127.0.01:6345"

# Configuration Constants
# PROJECT = "YOUR-PROJECT"
# TEAM_ENTITY = "YOUR-TEAM"  # MAKE SURE THIS IS NOT A PERSONAL ENTITY
# PERSONAL_ENTITY = "YOUR-PERSONAL-ENTITY" # Most likely your username
# ORG_ENTITY = "YOUR-ORG_ENTITY"  # Fetch this from the UI banner {BASE_URL}/orgs/wandb/registry/model

artifact_name = "test"

# log a personal entity artifact
run = wandb.init(project=PROJECT, entity=PERSONAL_ENTITY)
model_artifact = wandb.Artifact(name=artifact_name, type="model")
run.log_artifact(model_artifact).wait()
ARTIFACT = f"wandb-artifact:///{model_artifact.qualified_name}"

# link to model registry artifact
run.link_artifact(model_artifact, f"model-registry/{artifact_name}")
MODEL_REG_URL = f"wandb-artifact:///{PERSONAL_ENTITY}/model-registry/{artifact_name}:latest"
run.finish()

# Log a team entity artifact
run = wandb.init(project=PROJECT, entity=TEAM_ENTITY)
model_artifact = wandb.Artifact(name=artifact_name, type="model")
run.log_artifact(model_artifact).wait()
ARTIFACT = f"wandb-artifact:///{model_artifact.qualified_name}"


[...]

PR #3484, opened 2025-01-23
chore(weave): artifact ref trace server changes

jamie-rasmussen:
## Description

Extracting just the trace server changes from https://github.com/wandb/weave/pull/2920 so they can be reviewed and deployed separately. These changes were authored by the artifact team; I reviewed and have made only one minor tweak.

### DIFF
--- a/weave/trace_server/refs_internal.py
+++ b/weave/trace_server/refs_internal.py
@@ -11,6 +11,7 @@ from typing import Any, Union
 WEAVE_INTERNAL_SCHEME = "weave-trace-internal"
 WEAVE_SCHEME = "weave"
 WEAVE_PRIVATE_SCHEME = "weave-private"
+ARTIFACT_REF_SCHEME = "wandb-artifact"
 
 DICT_KEY_EDGE_NAME = "key"
 LIST_INDEX_EDGE_NAME = "index"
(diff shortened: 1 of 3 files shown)


weave/trace_server/refs_internal.py
```diff
@@ -143,9 +144,23 @@ def uri(self) -> str:
...
+
+    def uri(self) -> str:
+        u = f"{ARTIFACT_REF_SCHEME}:///{self.project_id}/{self.id}"
+        return u
+
+
 def parse_internal_uri(
     uri: str,
-) -> Union[InternalObjectRef, InternalTableRef, InternalCallRef]:
+) -> Union[InternalObjectRef, InternalTableRef, InternalCallRef, InternalArtifactRef]:
```
andrewtruong, line 163: Feels big enough that we can pull out into a type.  It will be easier to reuse in the validation util below

andrewtruong (commented):
What is this used for?  Do we need tests here?

andrewtruong (approved):

weave/trace_server/refs_internal.py
jamie-rasmussen: will do

jamie-rasmussen:
> What is this used for? Do we need tests here?

It's to allow artifact references to be used in objects and traces, allowing those things to link back from the Weave UI to the Artifacts UI - there's a loom in the PR linked above that shows the ultimate feature.

Agree that there could be better tests in this part of the code.

PR #3486, opened 2025-01-23
chore(weave): remove unneeded expand_columns param

gtarpenning:
Remove unused param. Expand columns shouldn't do anything in the sdk because we resolve refs recursively anyway. 

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the call data query process by removing support for expanded column details. This update simplifies data retrieval and provides a more straightforward querying experience for end-users.


andrewtruong:
What does expand columns do?

gtarpenning:
> What does expand columns do? @andrewtruong 

This pushes ref expansion to the server. We don't need this in the SDK because of `make_trace_obj`

andrewtruong (approved):

PR #3487, opened 2025-01-24
chore(weave): Streaming Evals Pt1: lazy `async_foreach`

tssweeney:
Context:

Currently the `async_foreach` method used in eval loop eagerly exhausts the passed in iterable which makes streaming datasets not possible. 

This is part 1 of a few PRs that build up to streaming support in evals & support for huggingface datasets.

---

Summary:

This pull request includes significant improvements to the `async_foreach` function in `weave/flow/util.py` and adds comprehensive tests for this functionality in `tests/trace/test_flow_util.py`. The changes enhance the documentation, error handling, and concurrency control of the `async_foreach` function.

Enhancements to `async_foreach`:

* [`weave/flow/util.py`](diffhunk://#diff-4ea0910967848b60c0a39eef3ba931abba8016263740bba2b083a94bf26ebf87R19-R96): Added detailed docstring to `async_foreach` explaining its functionality, arguments, and usage examples. Improved concurrency control by adding a semaphore and task management to ensure tasks are processed in order and cleaned up on errors or cancellation.

Comprehensive tests added:

* [`tests/trace/test_flow_util.py`](diffhunk://#diff-75763b33c284c3dd83fa28f8aaabbe7b5877444afa17531b9451a517a6dfbbb6R1-R155): Added multiple tests for `async_foreach` to cover basic functionality, concurrency limits, lazy loading, error handling, empty input, task cancellation, and execution order. These tests ensure that the function behaves as expected under various conditions.

### DIFF
--- a/tests/trace/test_flow_util.py
+++ b/tests/trace/test_flow_util.py
@@ -1,0 +1,155 @@
+import asyncio
+
+import pytest
+
+from weave.flow.util import async_foreach
+
+
+@pytest.mark.asyncio
+async def test_async_foreach_basic():
+    """Test basic functionality of async_foreach."""
+    input_data = range(5)
+    results = []
+
+    async def process(x: int) -> int:
+        await asyncio.sleep(0.1)  # Simulate async work
+        return x * 2
+
+    async for item, result in async_foreach(
+        input_data, process, max_concurrent_tasks=2
+    ):
+        results.append((item, result))
+
+    assert len(results) == 5
+    assert all(result == item * 2 for item, result in results)
+    assert [item for item, _ in results] == list(range(5))
+
+
+@pytest.mark.asyncio
+async def test_async_foreach_concurrency():
+    """Test that max_concurrent_tasks is respected."""
+    currently_running = 0
+    max_running = 0
+    input_data = range(10)
+
+    async def process(x: int) -> int:
+        nonlocal currently_running, max_running
+        currently_running += 1
+        max_running = max(max_running, currently_running)
+        await asyncio.sleep(0.1)  # Simulate async work
+        currently_running -= 1
+        return x
+
+    max_concurrent = 3
+    async for _, _ in async_foreach(
+        input_data, process, max_concurrent_tasks=max_concurrent
+    ):
+        pass
+
+    assert max_running == max_concurrent
+
+
+@pytest.mark.asyncio
+async def test_async_foreach_lazy_loading():
+    """Test that items are loaded lazily from the iterator."""
+    items_loaded = 0
+
+    def lazy_range(n: int):
+        nonlocal items_loaded
+        for i in range(n):
+            items_loaded += 1
+            yield i
+
+    async def process(x: int) -> int:
+        await asyncio.sleep(0.1)
+        return x
+
+    # Process first 3 items then break
+    async for _, _ in async_foreach(lazy_range(100), process, max_concurrent_tasks=2):
+        if items_loaded >= 3:
+            break
+
+    # Should have loaded at most 4 items (3 + 1 for concurrency)
+    assert items_loaded <= 4
+
+
+@pytest.mark.asyncio
+async def test_async_foreach_error_handling():
+    """Test error handling in async_foreach."""
+    input_data = range(5)
+
+    async def process(x: int) -> int:
+        if x == 3:
+            raise ValueError("Test error")
+        return x
+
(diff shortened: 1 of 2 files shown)


tests/trace/test_flow_util.py
```diff
@@ -0,0 +1,155 @@
...
+    # Give a moment for any lingering tasks to complete if cleanup failed
+    await asyncio.sleep(0.1)
+
+    # Check that we got the expected number of results before cancellation
+    assert len(results) == 2
+
+
+@pytest.mark.asyncio
+async def test_async_foreach_execution_order():
+    """Test that results are yielded in input sequence order regardless of completion order."""
```
andrewtruong, line 136: Why does eval order matter?

weave/flow/util.py
```diff
@@ -16,18 +16,83 @@ async def async_foreach(
...
+        while active_tasks:
+            # Always wait for the first task in the list to complete
+            # This ensures we yield results in order
+            task = active_tasks.pop(0)  # Remove completed task from front of list
+            try:
+                item, result = await task
+                yield item, result
+
+                # Add a new task if there are more items
+                maybe_queue_next_task()
```
andrewtruong, line 82: Queuing up an item is blocked on awaiting the task, so I think you can only get `max_concurrent_tasks` here, therefore the semaphore above is not required

weave/flow/util.py
```diff
@@ -16,18 +16,83 @@ async def async_foreach(
...
+        for _ in range(max_concurrent_tasks):
+            maybe_queue_next_task()
+
+        while active_tasks:
+            # Always wait for the first task in the list to complete
+            # This ensures we yield results in order
+            task = active_tasks.pop(0)  # Remove completed task from front of list
+            try:
+                item, result = await task
+                yield item, result
```
andrewtruong, line 79: subjective: I think this would have been easier for me to read as:
```py
try:
    yield active_tasks.pop(0)
```

tests/trace/test_flow_util.py
tssweeney: This ensures we loop through the dataset in the correct order

weave/flow/util.py
```diff
@@ -16,18 +16,83 @@ async def async_foreach(
...
+        - Memory usage is bounded by max_concurrent_tasks
+        - All pending tasks are properly cleaned up on error or cancellation
+        - Results are yielded in the same order as the input sequence
+    """
     semaphore = asyncio.Semaphore(max_concurrent_tasks)
+    active_tasks: list[asyncio.Task] = []
 
     async def process_item(item: T) -> tuple[T, U]:
+        """Process a single item using the provided function with semaphore control."""
         async with semaphore:
```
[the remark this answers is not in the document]
tssweeney: The semaphore is needed because while we control task creation with the list, we still need to limit how many func(item) calls run simultaneously - create_task() schedules tasks immediately, but the semaphore ensures only max_concurrent_tasks are actually executing at once.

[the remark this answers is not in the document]
tssweeney: the semaphore ensures this property without relying on the implicit  enforcement from the implementation below

tests/trace/test_flow_util.py
andrewtruong: But why dos that matter?  Dataset rows should be independent, i.e. order does not matter

weave/flow/util.py
tssweeney: seems better to await here rather than returning the task itself?

tests/trace/test_flow_util.py
tssweeney: I suppose technically this is true, but it is nice to have predict and score calls in the correct order.

tssweeney: this also allows the summary method to receive the results in a stable order (which was the original reason i added this - since tests were failing)

weave/flow/util.py
andrewtruong: sorry, typo.  I meant `yield await active_tasks.pop(0)`

weave/flow/util.py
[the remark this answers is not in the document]
andrewtruong: I think the guarantee you're looking for in the semaphore could be also be given by a test with the added benefit of slight perf improvement (less overhead)

andrewtruong (approved):

PR #3501, opened 2025-01-28
perf(weave): push heavy conditions into WHERE for calls stream query 

gtarpenning:
## Description

Push heavy conditions into WHERE clause before aggregating in calls query, and add secondary string check using `LIKE` before any json extraction. 

This will lead to functional changes in results from the calls query, as conditioning on rows before aggregating can lead to incomplete results when `calls_merged` contains unmerged parts. Because of the light filter that handles `deleted_at` and other light filters, the incomplete case only occurs when filtering by heavy fields. For example, if I am filtering my trace table by `inputs` in a project that is actively receiving traces, its possible that the inputs part is correctly included in the result set but outputs won't be included _yet_ (or vice versa). 

The addition of `LIKE` in the query dramatically reduces S3 read, reducing peak RAM by 2-20+x depending on the string filter, as well as speeding up the query. This is possibly due to better index usage, although it does not appear in the query planner so the exact reason this improves performance is tbd. 

Example difference between master and branch query structure:
**Master**
```sql
WITH filtered_calls AS (...)
SELECT ...
FROM calls_merged
WHERE calls_merged.project_id = '<>'
HAVING position(JSON_VALUE(any(calls_merged.output_dump), '$."prompt"'), 'ripples') > 0
```

**Branch**
```sql
WITH filtered_calls AS (...)
SELECT ...
FROM calls_merged
WHERE calls_merged.project_id = '<>'
  AND calls_merged.output_dump LIKE '%prompt%ripples%'
  AND position(JSON_VALUE(calls_merged.output_dump, '$."prompt"'), 'ripples') > 0
```

## Testing

Example issue: 
Prod:
Screenshot 2025-02-20 at 4 14 01 PM
Branch:
Screenshot 2025-02-20 at 4 14 12 PM
After 30 seconds (with `min_age_to_force_merge_seconds` set):
Screenshot 2025-02-20 at 4 22 42 PM

QA testing w/ queries generated in-app:
Master:
Screenshot 2025-02-20 at 5 21 56 PM
It [OOMed](https://us5.datadoghq.com/apm/trace/67b7d5b400000000e560a1529e2906a4?spanID=10431225205892846943&timeHint=1740101044485.919) when I tested again...
qa-slow-filter-master-2

Branch:
Screenshot 2025-02-20 at 5 18 29 PM
qa-slow-filter-branch

## Summary by CodeRabbit

- **Tests**
  - Added new test cases to validate behavior of call stream filtering and query handling under heavy conditions.
  - Enhanced SQL query validation tests to ensure queries honor updated filtering, ordering, and data conditions.

- **Enhancements**
  - Refined query building capabilities to provide more flexible aggregation handling.
  - Improved condition processing logic to better distinguish between standard and heavy filtering scenarios for more accurate query results.
  - Updated handling of input data to ensure accurate reflection of state when processing attributes and inputs.

- **Database Migrations**
  - Updated settings for the `calls_merged` table to optimize merging operations.
  - Introduced new Bloom filter indexes on `inputs_dump` and `output_dump` columns to enhance query performance.


tests/trace_server/test_calls_query_builder.py
```diff
@@ -300,12 +298,11 @@ def assert_sql(cq: CallsQuery, exp_query, exp_params):
     query = cq.as_sql(pb)
     params = pb.get_params()
 
-    assert exp_params == params
-
     exp_formatted = sqlparse.format(exp_query, reindent=True)
     found_formatted = sqlparse.format(query, reindent=True)
 
     assert exp_formatted == found_formatted
+    assert exp_params == params
```
gtarpenning, line 305: easier to debug in this order

tssweeney (commented):
Before diving into the code, i have a concern:

I believe this will fail to return rows which have not been merged by the AMT. Referencing the example in the description: what happens if the start and end events are not merged? The resulting data will return the unmerged rows without their start events!

gtarpenning:
> Before diving into the code, i have a concern:
> 
> I believe this will fail to return rows which have not been merged by the AMT. Referencing the example in the description: what happens if the start and end events are not merged? The resulting data will return the unmerged rows without their start events!

Hmm, looking at the query plan I do think this is correct, although i'm not sure how often this will happen in practice. I am immediately confronted with dumb ways around this, like, always including all the start events if conditioning on an end event and vice versa @tssweeney

gtarpenning:
@tssweeney  We could also force merges by using `FINAL`... In large projects this will very likely be more performant than the groupby... In local testing, using a query that filters down to 200 rows, `FINAL` used 6x less memory than `GROUP BY`. 

I'm still not exactly sure the best way to repro the conditions that would lead to the issue, merges are hard to predict... And the aggregation functions _appear_ in my testing to actually be working as expected (ie, the query planner reports unmerged parts of the table, but doing filtering on the inputs still always returns the outputs as well). i'll use QA tomorrow.

tests/trace/test_client_trace.py
```diff
@@ -3236,3 +3236,59 @@ def test():
...
+        output={"d": 5},
+    )
+    client.server.call_end(tsi.CallEndReq(end=end))
+
+    res = _make_query("inputs.param.value1", "hello")
+    assert len(res) == 1
+    assert res[0].inputs["param"]["value1"] == "hello"
+
+    # Does the query return the output?
+    assert res[0].output["d"] == 5
```
gtarpenning, line 3294: This test highlights the error case that the query creates.

PR #3503, opened 2025-01-28
feat(ui): lightbox for images

jamie-rasmussen:
## Description

Partial fix for https://wandb.atlassian.net/browse/WB-21609

Brings back a lightbox viewer for logged images, also with a preview on hover, which we didn't have before.

Tested with new `wf` example, Image-Eval which creates a Weave dataset and evaluation from a W&B artifact that contains 100 image generation prompts and the dall-e output for them.

Before:
Screenshot 2025-01-27 at 10 12 44 PM

After:
Screenshot 2025-01-27 at 10 13 39 PM
Screenshot 2025-01-27 at 10 13 45 PM
image


bcsherma (approved):
🔥

PR #3507, opened 2025-01-28
chore(ui): new Radix based tooltip implementation

jamie-rasmussen:
## Description

https://wandb.atlassian.net/browse/WB-22063

Our current Tooltip component is used in many places and is just a styled version of Semantic UI's [Popup](https://semantic-ui.com/modules/popup.html). We would like to replace it with a component based on [Radix Tooltip](https://www.radix-ui.com/primitives/docs/components/tooltip). Unfortunately many places make direct use of Popup props. 

This is the final step of a multi-step plan to address this, which helps with the long term goal of not using Semantic and also will get rid of many occurrences of a warning that is cluttering the dev console.

Step 1 (done): https://github.com/wandb/weave/pull/3176 - clone the Semantic UI based Tooltip as TooltipDeprecated
Step 2 (done):  https://github.com/wandb/core/pull/27083 - Update core Tooltip uses that need the Semantic UI props to use TooltipDeprecated
Step 3 (this PR): Land the revised Radix-based Tooltip in the Weave repo.

## Testing

How was this PR tested?


weave-js/src/components/Tooltip.tsx
```diff
@@ -1,33 +1,91 @@
...
+  children,
+}: TooltipProps) => {
+  const defaultPosition = parsePosition(position);
+  const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+  return (
+    <Provider>
+      <Root open={isTooltipOpen} onOpenChange={setIsTooltipOpen}>
+        <Trigger asChild>
+          {/* span is needed so tooltip works on disabled buttons */}
+          <span className="[display:inherit]">{trigger}</span>
```
jwlee64, line 75: This is a nice touch, adding <span> around the triggers always annoyed me

weave-js/src/components/Tooltip.tsx
```diff
@@ -1,33 +1,91 @@
...
+        </Trigger>
+        <Portal>
+          <Content
+            style={{
+              // it's hard to state how silly this is, but the zIndex on semantic's modal is 2147483605 - so, that + 1
+              zIndex: 2147483606,
```
jwlee64, line 81: lol

jwlee64 (approved):

PR #3523, opened 2025-01-29
feat(weave): allow disabling capture of context

jamie-rasmussen:
## Description

Internal Slack: https://weightsandbiases.slack.com/archives/C072S236A75/p1738168167702489

Adds the ability to disable capture of contextual information such as Weave client version and operating system.

### DIFF
--- a/weave/trace/settings.py
+++ b/weave/trace/settings.py
@@ -61,6 +61,12 @@ class UserSettings(BaseModel):
     may lead to unexpected behaviour.  Make sure this is only set once at the start!
     """
 
+    capture_client_info: bool = True
+    """Toggles capture of client information (Python version, SDK version) for ops."""
+
+    capture_system_info: bool = True
+    """Toggles capture of system information (OS name and version) for ops."""
+
     client_parallelism: Optional[int] = None
     """
     Sets the number of workers to use for background operations.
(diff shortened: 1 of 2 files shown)


andrewtruong (approved):

weave/trace/weave_client.py
```diff
@@ -987,12 +987,13 @@ def create_call(
...
         attributes = AttributesDict(**attributes)
-        attributes._set_weave_item("client_version", version.VERSION)
-        attributes._set_weave_item("source", "python-sdk")
-        attributes._set_weave_item("os_name", platform.system())
-        attributes._set_weave_item("os_version", platform.version())
-        attributes._set_weave_item("os_release", platform.release())
-        attributes._set_weave_item("sys_version", sys.version)
+        if should_capture_context():
+            attributes._set_weave_item("client_version", version.VERSION)
+            attributes._set_weave_item("source", "python-sdk")
```
andrewtruong, line 992: I think we can always capture these.  Is there ever a case where this would be bad to capture?

weave/trace/settings.py
```diff
@@ -61,6 +61,13 @@ class UserSettings(BaseModel):
     may lead to unexpected behaviour.  Make sure this is only set once at the start!
     """
 
+    capture_context: bool = True
```
andrewtruong, line 64: nit: would rename to `capture_system_info`

weave/trace/weave_client.py
jamie-rasmussen: I split this into two separate flags. I agree that we would almost always want it for debugging, but I'm giving user the option under the theory that we should allow consistent configuration and give users control (esp. since they pay by usage)

PR #3528, opened 2025-01-29
chore(app): Handle dark mode in ToggleButtonGroup

ericakdiaz:
## Description

See report here https://weightsandbiases.slack.com/archives/C01KQ5KTDC3/p1738147450239079?thread_ts=1738108415.428959&cid=C01KQ5KTDC3. The component didn't have styling for dark mode, and also the rounded corners weren't working correctly because `Button` is the only child of its parent.

## Testing

How was this PR tested?

See testing in https://github.com/wandb/core/pull/27117


weave-js/src/components/ToggleButtonGroup.tsx
```diff
@@ -82,10 +86,13 @@ export const ToggleButtonGroup = React.forwardRef<
...
                   (isDisabled || optionIsDisabled) && 'cursor-auto opacity-35',
                   value === optionValue
-                    ? 'bg-teal-300/[0.48] text-teal-600 hover:bg-teal-300/[0.48]'
-                    : 'hover:bg-oblivion/7 bg-oblivion/5 text-moon-600 hover:text-moon-800',
-                  'first:rounded-l-sm', // First button rounded left
-                  'last:rounded-r-sm' // Last button rounded right
+                    ? 'bg-teal-300/[0.48] text-teal-600 hover:bg-teal-300/[0.48] dark:bg-teal-700/[0.48] dark:text-teal-400'
+                    : 'hover:bg-oblivion/7 bg-oblivion/5 text-moon-600 hover:text-moon-800 dark:bg-moonbeam/[0.05] hover:dark:bg-teal-700/[0.48] hover:dark:text-teal-400',
+                  i === 0
+                    ? 'rounded-l-sm rounded-r-none' // First button rounded left
```
brianlund-wandb, line 92: trying to wrap my head around tailwind, so i'm curious - the issue is the nesting right?  That `first:*`/`last:*` would end up referring to the first/last children of the `ToggleGroup.Item` not the `ToggleGroup.Root`?

Would it work to be explicit about the group?
```
<div class="group">
  <div class="group/item">
    <div class="bg-gray-500 text-white group-first/item:bg-black">lorem</div>
  </div>
  <div class="group/item">
    <div class="bg-gray-500 text-white group-first/item:bg-black">lorem</div>
  </div>
  <div class="group/item">
    <div class="bg-gray-500 text-white group-first/item:bg-black">lorem</div>
  </div>
</div>
```
In which case the `ToggleGroup.Root` would get the `group` class, the `ToggleGroup.Item` would get the `group/item` class, and the `Button` would get the `group-first\item:rounded-l-sm group-last\item:rounded-r-sm`

Not sure if that would cause other problems, more just trying to see if i understand tailwind's expectations.

brianlund-wandb (approved):

weave-js/src/components/ToggleButtonGroup.tsx
```diff
@@ -61,35 +61,38 @@ export const ToggleButtonGroup = React.forwardRef<
...
-                  value === optionValue
-                    ? 'bg-teal-300/[0.48] text-teal-600 hover:bg-teal-300/[0.48]'
-                    : 'hover:bg-oblivion/7 bg-oblivion/5 text-moon-600 hover:text-moon-800',
-                  'first:rounded-l-sm', // First button rounded left
-                  'last:rounded-r-sm' // Last button rounded right
-                )}>
-                {!iconOnly ? optionValue : <></>}
-              </Button>
-            </ToggleGroup.Item>
+            <div className="group" key={optionValue}>
```
peytonsarmiento, line 64: do we want the key in the div as well?

peytonsarmiento (approved):

weave-js/src/components/ToggleButtonGroup.tsx
ericakdiaz: i just didn't bother removing it from the child, but yes it should be on the div as the parent component that's being mapped over to prevent the React error "Warning: Each child in a list should have a unique “key” prop."
https://react.dev/learn/rendering-lists. having it on the ToggleGroup.Item doesn't really have an effect so i didn't remove it.

PR #3531, opened 2025-01-29
docs(weave): what gets logged

jamie-rasmussen:
## Description

Update tracing page with an FAQ section.


tssweeney (approved):
Lovely Doc

PR #3532, opened 2025-01-29
refactor(ui): split SmallRef into smaller files

jamie-rasmussen:
## Description

Breaks SmallRef file up by component to make it a bit more manageable. (May also help with circular dependency issues?)


gtarpenning (approved):
Just curious, why is the `smallRef` folder not `SmallRef`? (I realize I stamped that earlier change)

jamie-rasmussen:
> Just curious, why is the `smallRef` folder not `SmallRef`? (I realize I stamped that earlier change)

There's not a good reason - we don't have a coding standard and we're inconsistent. Starting with lower case matches the sibling folders.

gtarpenning:
👍🏻

PR #3542, opened 2025-01-30
chore(weave): update model providers, add azure

jwlee64:
## Description

Adds Azure to our list of providers in the playground
adds some logic to get specific secrets for Azure

### DIFF
--- a/weave/trace_server/llm_completion.py
+++ b/weave/trace_server/llm_completion.py
@@ -55,7 +55,7 @@ def lite_llm_completion(
 
 def get_bedrock_credentials(
     model_name: str,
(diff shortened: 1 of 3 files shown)


weave/trace_server/llm_completion.py
```diff
@@ -89,3 +94,33 @@ def get_bedrock_credentials(
         )
 
     return aws_access_key_id, aws_secret_access_key, aws_region_name
+
+
+def get_azure_credentials(model_name: str) -> tuple[Optional[str], Optional[str]]:
```
jamie-rasmussen, line 99: It seems like these can't be Optional - if either value is None you raise?

jamie-rasmussen (approved):

PR #3549, opened 2025-01-30
chore(ui): add missing field to TraceObjSchema

jamie-rasmussen:
## Description

Adding `deleted_at` field to TypeScript side to match Python declaration.
https://github.com/wandb/weave/blob/master/weave/trace_server/trace_server_interface.py#L181

I think the field has to always exist because of the default value, so `| null` seemed more correct than the `?` we have for `base_object_class` and `wb_user_id`.


gtarpenning (approved):
Makes sense, thx

PR #3556, opened 2025-01-31
feat(ui): Very basic MVP for viewing a Scorer's logs

tssweeney:
A note to reviewers. I will break this up into 3 PRs after I get approval for the purposes of rollout, but i figured it might be easier to review in bulk because everything fits together:

1. Server
2. UI
3. Python & Docs

---

In order to help complete the user flow for scoring, we need a way to actually get the scored calls. The user as a few downstream options: build a dataset, python analytics, view in UI, etc... However, as of right now our data fetch side is limited to just: view the scores on a call... we don't have an easy way to view all cals scored by some metric. This PR puts in place:

a. the backend query layer to fetch calls scored by a particular version of the scorer
b. adds a super lightweight (emphasis on super) to navigate to the calls table for a given scorer
c. adds a very basic query param to the client that allows fetching these calls

---

This docs page summarizes the work:

Screenshot 2025-01-31 at 14 56 34


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectsPage/ObjectVersionPage.tsx
```diff
@@ -283,6 +285,33 @@ const ObjectVersionPageInner: React.FC<{
...
               </div>
             )}
+            {isScorer && (
+              <div className="block">
+                <p className="text-moon-500">Score Log</p>
+                <CallsLink
+                  entity={entityName}
+                  project={projectName}
+                  neverPeek
+                  gridFilters={{
```
m-rgba, line 295: Bit of a bigger ask - been looking at cleaning up the filters area.
Would it be possible if we're pushing to the traces view to display an object, to just display the object name with an identifier of what it is?

As an example:
CleanShot 2025-01-31 at 12 10 52@2x

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectsPage/ObjectVersionPage.tsx
```diff
@@ -236,7 +238,7 @@ const ObjectVersionPageInner: React.FC<{
       }
       headerContent={
         <Tailwind>
-          <div className="grid w-full grid-flow-col grid-cols-[auto_auto_auto_1fr] gap-[16px] text-[14px]">
+          <div className="grid w-full grid-flow-col grid-cols-[auto_auto_auto_auto_1fr] gap-[16px] text-[14px]">
```
m-rgba, line 241: Now that we have more than a couple columns on all of the objects views we can set `grid-cols-auto` here. 
- Previously we were prescriptive about the column count so that it didn't feel weird having really large spaces between each column.
- Now that we have a delete icon on the farthest right column it will anchor all of the views.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectsPage/ObjectVersionPage.tsx
tssweeney: This will need to be a followup. Incremental change. Don't want to mess with filters here as the intention of this PR is to setup the edge

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectsPage/ObjectVersionPage.tsx
tssweeney: Uggg i _despise_ Tailwind. Yeah i will fix

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectsPage/ObjectVersionPage.tsx
```diff
@@ -283,6 +285,33 @@ const ObjectVersionPageInner: React.FC<{
                 <UserLink userId={objectVersion.userId} includeName />
               </div>
             )}
+            {isScorer && (
```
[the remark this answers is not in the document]
tssweeney: made the name changes. I suspect this will change very soon anyway maybe with it's own tab or something

J2-D2-3PO:
@tssweeney my proposed refactor of this section (never merged) contains a builtin scorers page - can we reconcile the two https://github.com/wandb/weave/blob/f0fb79782a1e4f6f1b413c616d1f6b22a793beda/docs/docs/guides/scorers/built-in-scorers.md?

tests/trace/test_scored_calls_query.py
```diff
@@ -0,0 +1,225 @@
...
+def assert_number_of_calls(client: WeaveClient, count: int):
+    stats = client.server.calls_query_stats(
+        CallsQueryStatsReq(
+            project_id=client._project_id(),
+        )
+    )
+    assert stats.count == count
+
+
+def output_field_for_name(name: str) -> str:
```
jamie-rasmussen, line 22: These seem like essentially the same methods you added in `feedback_types.py`?

tests/trace/test_scored_calls_query.py
```diff
@@ -0,0 +1,225 @@
...
+):
+    """
+    This is a unified test for both scorer and op tests. It ensures that we can query for calls
+    that have been scored by any type of scorer and ensures that we correctly differentiate between
+    versions of the same scorer.
+
+    We also ensure that the low-level and high-level queries return the same results. This is because
+    we will likely implement a better query at the service layer and we want to ensure that the high
+    level api continues to work correctly. We don't really want users to manually write the mongo
+    query, so that part of these tests are just here for correctness.
```
jamie-rasmussen, line 53: are just here => is just here

tests/trace/test_scored_calls_query.py
```diff
@@ -0,0 +1,225 @@
...
+
+    # This is required for the test to work
+    assert s0_v0_name == s0_v1_name
+    assert s0_v0_name != s1_v0_name
+
+    s0_name = s0_v0_name
+    s1_name = s1_v0_name
+
+    """
+    The above ordering is quire important. We now have the following scores:
```
jamie-rasmussen, line 93: quire => quite

weave/trace/weave_client.py
```diff
@@ -859,13 +870,44 @@ def get_calls(
...
+                full ref URI which will fetch scores for a specific version of the scorer.
 
         Returns:
             An iterator of calls.
         """
         if filter is None:
             filter = CallsFilter()
 
+        # This logic might be pushed down to the server soon, but for now it lives here:
+        if scored_by and len(scored_by) > 0:
```
jamie-rasmussen, line 884: len clause seems redundant since empty list would be falsey

weave/trace/weave_client.py
```diff
@@ -859,13 +870,44 @@ def get_calls(
             columns: A list of columns to include in the response. If None,
                all columns are included. Specifying fewer columns may be more performant.
                Some columns are always included: id, project_id, trace_id, op_name, started_at
+            scored_by: A list of Scorers to filter by. Multiple scorers are ANDed together. You can
```
jamie-rasmussen, line 873: It would be nice if you could also pass a single string here instead of a list with one item.
It would also be nice if instead of a full ref URL you could specify just `:v1` or the digest

weave/trace/weave_client.py
```diff
@@ -859,13 +870,44 @@ def get_calls(
...
         Returns:
             An iterator of calls.
         """
         if filter is None:
             filter = CallsFilter()
 
+        # This logic might be pushed down to the server soon, but for now it lives here:
+        if scored_by and len(scored_by) > 0:
+            exprs = []
+            if query is not None:
```
jamie-rasmussen, line 886: This doesn't seem like it belongs inside the check for scored_by - you should be able to specify the query and have it apply even if there is no scored_by.

jamie-rasmussen (approved):

docs/docs/guides/evaluation/scorers.md
```diff
@@ -303,493 +312,133 @@ In Weave, Scorers are used to evaluate AI outputs and return evaluation metrics.
...
 
-    model = XMLModel()
-    xml_scorer = ValidXMLScorer()
+#### Single Call API
 
-    dataset = [
-        {"input": "Generate a valid XML with a root element"},
-        {"input": "Create an invalid XML"}
-    ]
+To retrieve the calls for a single call, you can use the `get_calls` method.
```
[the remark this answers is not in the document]
tssweeney: fixed

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/Links.tsx
```diff
@@ -419,21 +420,31 @@ export const CustomLink: React.FC<{
...
+  callCount?: number;
   countIsLimited?: boolean;
   filter?: WFHighLevelCallFilter;
+  gridFilters?: GridFilterModel;
   neverPeek?: boolean;
   variant?: LinkVariant;
 }> = props => {
   const {peekingRouter, baseRouter} = useWeaveflowRouteContext();
   const router = props.neverPeek ? baseRouter : peekingRouter;
+  let label = 'View Traces';
```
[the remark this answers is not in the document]
tssweeney: yup

tests/trace/test_scored_calls_query.py
tssweeney: Yes, but i intentionally did not not want rely on that implementation since it is part of what I was testing

tests/trace/test_scored_calls_query.py
tssweeney: ok

tests/trace/test_scored_calls_query.py
tssweeney: fixed

weave/trace/weave_client.py
tssweeney: fair

weave/trace/weave_client.py
tssweeney: I am not following this logic. If we don't enter the scored by block then the query just gets passed along

jamie-rasmussen: I see now, my mistake

PR #3558, opened 2025-01-31
chore(ui): Add useProjectInfo hook

jamie-rasmussen:
## Description

A project like `jamie-rasmussen/2025-01-30_wf` has both an "id" and an "internalId". If you decode them they look like:
id: `Project:v1:2025-01-30_wf:jamie-rasmussen`
internalId: `ProjectInternalId:40869716`
So somewhat unexpectedly, the "id" value changes when you rename the project. The "internalId" does not change.

In the upcoming Saved Views work, we use local storage to remember the last view you used, on a per-project (and per-page, traces or evaluations) basis. This hook allows us to query for the "internalId" of the project, so that we can use it as part of the key.

### DIFF
--- a/wb_schema.gql
+++ b/wb_schema.gql
@@ -202,6 +202,7 @@ type Project implements Node {
   createdAt: DateTime!
   updatedAt: DateTime
   id: ID!
+  internalId: ID!
   name: String!
   runs(filters: JSONString, order: String, first: Int): RunConnection
   entity: Entity!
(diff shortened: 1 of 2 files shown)


tssweeney (approved):
We should be careful with internalid... as named, i believe it is not intended to be used externally. I added this edge i believe purely for weave's id swapping. Marie used this in workspaces/report views I believe in a similar way that you are using now to prevent naming issues. 

All that to be said, this is good, but i just want to prevent people from using this internal id in ways it shouldn't (since it shouldn't really leak)

PR #3560, opened 2025-01-31
fix(ui): more specific icons shown in SmallRef

jamie-rasmussen:
## Description

Fixes https://wandb.atlassian.net/browse/WB-21887

This improves a number of things about SmallRef.
* Currently, the component renders once with a version hash, then with the version number, and again but in a bolder font. I change it to show a loading dots indicator, and then the final form.
* We were showing a generic icon instead of a more specific one for prompts, models, datasets, etc.
* Deleted refs now have a tooltip with more information.

Before:
Screenshot 2025-01-31 at 10 05 46 AM

After:
Screenshot 2025-01-31 at 10 06 31 AM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/utilities.ts
@@ -298,6 +298,15 @@ export const isObjDeleteError = (error: Error | null): boolean => {
   return false;
 };
 
+export const getErrorReason = (error: Error): string | null => {
+  try {
+    const parsed = JSON.parse(error.message);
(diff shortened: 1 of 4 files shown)


gtarpenning (approved):
I'm a bit confused why we are able to get rid if the op hook entirely, but this works nicely in my poking around with it so 🚢 !

jamie-rasmussen:
> I'm a bit confused why we are able to get rid if the op hook entirely, but this works nicely in my poking around with it so 🚢 !

This is a little puzzling to me too, but Ops are also objects with the field "kind" having value "op" - the code for `useOpVersion` is almost identical to `useObjectVersion` other than your deletion related error handling and some minor structural difference in output shape, so was able to simplify the code.

PR #3570, opened 2025-02-01
chore(ui): add analytic event to metrics chart views

gtarpenning:
## Description

Add analytics for the metrics plot. Includes a latency metric which could be interesting to see if users are running into perf issues.


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsCharts.tsx
```diff
@@ -114,6 +115,29 @@ export const CallsCharts = ({
     columns
   );
 
+  const [callsQueryStartTime, setCallsQueryStartTime] = useState<number | null>(
```
tssweeney, line 118: Might want to pull this into its own hook to reduce component code

bcsherma (approved):

PR #3584, opened 2025-02-04
chore(weave): Separate Scorers module from the rest of weave

tcapelle:
This PR puts the Scorers as a standalone module. 
- This enables imports of dependencies at the top of the scorers (`litellm`, `transformers`, etc...)
- Moved the base `Scorer` class back to `weave.flow` to enable calling `weave.Scorer`
- Use `weave.Scorer` consistently.

This should not make any difference to the end user


tcapelle:
I think this is ready senpai @andrewtruong

tssweeney (approved):

andrewtruong (approved):

PR #3587, opened 2025-02-04
fix(ui): Add <0.0001 precision for costs

jwlee64:
## Description

we removed <$0.01 for more precision here
https://github.com/wandb/weave/pull/3299/files
adds <$0.001

before
screenshot_2025-02-04_at_9 53 27___am_480

after
Screenshot 2025-02-04 at 10 00 46 AM


m-rgba:
My bad, I thought at 4 precision we wouldn't need it anymore.
I guess the cost of 1 output token in 3.5 is 0.0000015 though.

m-rgba (approved):

PR #3588, opened 2025-02-04
chore(ui): Add to dataset design edits

m-rgba:
## Description

General:
- Tightened up padding across views.

Drawer / header / footer:
- Swapped in the resizeable drawer component.
- Tweaked close / back controls for page 1/2 of the drawer.
- Added fullscreen control for editing the dataset rows.
- Tweaked back button to secondary / tweaked Add text (to Add to dataset).

ResizeableDrawer component enhancements:
- Increased min-width to match peek drawer component.
- Added a proper height to account for the header bar for the fixed component.

Mapping:
- Feature: Added latest dataset selector to dropdown (revert [this commit](https://github.com/wandb/weave/pull/3588/commits/d6d1e5433d5befbbae2e566ef6e31b558d73bb29) to remove).
- Made the dropdown columns liquid vs. fixed.
- Added a max height / wrap / overflow scroll for the `from:` field mapping since many rows can be added.
- Tweaked the styles of the containers around the object / field mapping area.
- Swapped the loader with our dancing dots loader (other one is going to be deprecated).
- Added an error tag w/ tooltip to see any error messages.

Editing:
- Made the dataset editor full-width w/ no padding to fit the area.

| Before | After |
| -------- | ------- |
| before1 | after1 |
| before2 | after2 |

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
@@ -239,7 +239,7 @@ export const AddToDatasetDrawer: React.FC<AddToDatasetDrawerProps> = ({
           sx={{
             flex: 1,
             overflow: 'auto',
-            px: 2,
+            px: currentStep === 1 ? 2 : 0,
             display: 'flex',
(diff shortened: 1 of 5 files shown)


bcsherma (approved):
Looks great!

PR #3594, opened 2025-02-04
chore(weave): Reduced tests that run for doc-only changes

tssweeney:
Ensures that the heavy tests are skipped when changes are doc-only


J2-D2-3PO:
@tssweeney wanted to follow up on this, but I see the PR is closed? Was that intentional?

PR #3600, opened 2025-02-05
fix(ui): remove parent row indicator icon from trace tree

jamie-rasmussen:
## Description

Internal Slack: https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1738776616167189
Change to the tooltip component made this old icon peek out.

Before:
Screenshot 2025-02-05 at 12 20 34 PM

After:
Screenshot 2025-02-05 at 12 20 26 PM

PR #3602, opened 2025-02-05
Jamie/show up arrow

jamie-rasmussen:
## Description

Alternative to https://github.com/wandb/weave/pull/3600 - fix the parent row arrow indicator instead of removing it.

Screenshot 2025-02-05 at 1 46 40 PM


jwlee64 (approved):
lgtm

PR #3605, opened 2025-02-05
chore(weave): Allows scorer link to link into non-roots

tssweeney:
We probably need to re-expose trace roots toggle again (i think we can make it clearer). But in any case, scores are almost never on roots, so this PR allows the "view calls" link in the scorer view to show non-trace roots as well.


gtarpenning (approved):
Lets sync with @m-rgba with his new filter view, I think we can easily add this kind of toggle within the filter bar context.

PR #3617, opened 2025-02-06
security(weave_query): Upgrade Pyarrow to 17.0.0

domphan-wandb:
## Description

[WB-22924](https://wandb.atlassian.net/jira/software/c/projects/WB/boards/138?selectedIssue=WB-22924)

This PR upgrades and pins pyarrow to 17.0.0.

### 16.1.0 -> 17.0.0 Upgrade Issues

This [change in pyarrow](https://github.com/apache/arrow/pull/13894) introduced stricter checks when using `pa.ListArray.from_arrays().` Specifically, we cannot use both a null mask with an offset that has nulls.

```cpp
  if (null_bitmap != nullptr && offsets.null_count() > 0) {
    return Status::Invalid(
        "Ambiguous to specify both validity map and offsets with nulls");
  }
```

This new strictness caused many of our unit tests to fail:
image

**A strange observation**
I stumbled upon this accidentally, but just checking the null count in the offsets like `new_offsets.null_count` makes all the unit tests pass. 

My theory:
It looks like when we're getting new offsets during concatenation, we use compute functions (which are lazy), so when the offset is checked for nulls on the cpp side, it's possible that it's in an intermediate state with nulls. When evaluating the null_count in python, this forces the compute functions to evaluate, resulting in offsets without nulls.

### Solution
I created a new function called `safe_list_array_from_arrays` that handles the strictness. It checks the `null_count` of the offsets:
*  if there are nulls, it calls `pa.ListArray.from_arrays()` without a mask, then uses `pc.if_else` to set nulls.
* if there are no nulls, it calls `pa.ListArray.from_arrays()` with the mask.

**The outcome:**
`offset.null_count` is called which triggers evaluation of the offset array
a. If there are no nulls in the offset array, the mask is used, and the behavior is the same as before the upgrade
b. If there are nulls in the offset array, then a mask isn't used, but `pc.if_else` handles nulling values in the resulting array.

## Testing

I used the unit test failures to test my changes.

[WB-22924]: https://wandb.atlassian.net/browse/WB-22924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Chores**
  - Updated the key dependency to enforce version 17.0.0 for improved compatibility with current tests.
- **New Features**
  - Introduced an enhanced mechanism for constructing list arrays, ensuring more robust handling of null values and better data consistency during array operations.

### DIFF
--- a/weave-js/src/core/ops/primitives/list.ts
+++ b/weave-js/src/core/ops/primitives/list.ts
@@ -678,7 +678,7 @@ export const opGroupby = makeListOp({
     engine
   ) => {
     const tableRows: any[] = inputs.arr;
-    // console.log('OP GROUP BY RESOLVER', tableRows.length);
+    // console.log('OP GROUP BY RESOLVER', tableRows.length, "");
     const groupByFn = inputs.groupByFn;
     const groupSafeKeyFn =
       groupByFn.nodeType === 'void' ? groupByFn : toSafeCall(groupByFn);
--- a/weave-js/src/core/ops/primitives/list.ts
+++ b/weave-js/src/core/ops/primitives/list.ts
@@ -678,7 +678,7 @@ export const opGroupby = makeListOp({
     engine
   ) => {
     const tableRows: any[] = inputs.arr;
-    // console.log('OP GROUP BY RESOLVER', tableRows.length, "");
+    // console.log('OP GROUP BY RESOLVER', tableRows.length);
     const groupByFn = inputs.groupByFn;
     const groupSafeKeyFn =
       groupByFn.nodeType === 'void' ? groupByFn : toSafeCall(groupByFn);
--- a/weave_query/requirements.legacy.txt
+++ b/weave_query/requirements.legacy.txt
@@ -6,8 +6,7 @@ typing_extensions>=4.0.0
 
 # Definitely need arrow
 # TODO: Colab has 9.0.0, can we support?
-# TODO: 17.0.0 breaks a bunch of tests - can we move this requirement to just the engine?
-pyarrow>=14.0.1,<17.0.0
+pyarrow==17.0.0
 
 # pydantic integration, and required by openai anyway
 openai>=1.0.0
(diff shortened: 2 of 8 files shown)


nicholaspun-wandb:
* The nox tests should be resolved with a rebase to current master, looks like the clickhouse issues from a couple days ago
* The [build](https://app.circleci.com/pipelines/github/wandb/core/183316/workflows/fdc07527-8784-4903-b7db-528d6bfb2fd3/jobs/5085499) issue looks like a mismatch between `weave` and `weave_query`?

domphan-wandb:
> * The nox tests should be resolved with a rebase to current master, looks like the clickhouse issues from a couple days ago
> * The [build](https://app.circleci.com/pipelines/github/wandb/core/183316/workflows/fdc07527-8784-4903-b7db-528d6bfb2fd3/jobs/5085499) issue looks like a mismatch between `weave` and `weave_query`?

yeah, build ci issue won't pass until i can merge this in: https://github.com/wandb/core/commit/6024445f236bdcd91f082a55138bcfc17b3c97d1

ty, i'll rebase

nicholaspun-wandb (approved):

wandb-kc (approved):

wandb-kc:
Do we need to bump pyarrow here as well? https://github.com/wandb/core/blob/1fcfde879bdf5e8a5d39c69b189d9b2cc2795d67/services/weave-python/requirements.txt#L11

domphan-wandb:
> Do we need to bump pyarrow here as well? https://github.com/wandb/core/blob/1fcfde879bdf5e8a5d39c69b189d9b2cc2795d67/services/weave-python/requirements.txt#L11

Yup, here's the companion PR for it: https://github.com/wandb/core/pull/27312

When this gets stamped, I'll merge the core PR in first, then this PR

tssweeney (approved):

PR #3633, opened 2025-02-07
chore(weave): Switch to RLock for PIL Image thread safety

tssweeney:
Replaces threading.Lock with threading.RLock in our PIL thread safety implementation to prevent potential deadlocks if PIL's load() method needs to call itself recursively. This is a defensive change that maintains the same thread-safety guarantees while being more robust against recursive image loading scenarios, with negligible performance impact.

---

* Replaced `_new_lock_lock` and `_fallback_load_lock` with `threading.RLock()` to ensure reentrant locking. (`weave/initialization/pil_image_thread_safety.py`)
* Updated the `new_load` method to use `threading.RLock()` for the `_weave_load_lock` attribute to prevent race conditions with reentrant locking. (`weave/initialization/pil_image_thread_safety.py`)

### DIFF
--- a/weave/initialization/pil_image_thread_safety.py
+++ b/weave/initialization/pil_image_thread_safety.py
@@ -27,8 +27,8 @@ from typing import Any, Callable, Optional
 # `_fallback_load_lock` is a global lock that is used to ensure thread-safe image loading when per-instance locking fails
 _patched = False
 _original_methods: dict[str, Optional[Callable]] = {"load": None}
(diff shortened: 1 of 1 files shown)


gtarpenning (approved):
Would a test be appropriate? Double load PIL?

tssweeney:
> Would a test be appropriate? Double load PIL?

I'll add to the existing tests here: tests/trace/test_image_thread_safety.py

PR #3637, opened 2025-02-08
chore(weave): add query to calls_query_stats in `get_calls`

gtarpenning:
## Description

[WB-23221](https://wandb.atlassian.net/browse/WB-23221)

The query parameter can limit results, it must be passed to the calls stats function to get the length as well as the main streaming endpoint. 

Also move the query construction code to where the other helper is, for clarity on main client endpoint. 

## Testing

add test

[WB-23221]: https://wandb.atlassian.net/browse/WB-23221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/test_scored_calls_query.py
+++ b/tests/trace/test_scored_calls_query.py
@@ -126,6 +126,7 @@ async def perform_scorer_tests(
     calls_high_level = client.get_calls(scored_by=[s0_name])
     assert call_ids(calls_low_level) == expected_ids
     assert call_ids(calls_high_level) == expected_ids
+    assert len(calls_high_level) == len(calls_low_level)
 
     # 2. Query for calls that have been scored by s0:v0
     expected_ids = call_ids([call_scored_by_s0_v0, call_scored_by_s0_v0_and_s1_v0])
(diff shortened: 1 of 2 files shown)


tests/trace/test_scored_calls_query.py
```diff
@@ -126,6 +126,7 @@ def predict(x):
     calls_high_level = client.get_calls(scored_by=[s0_name])
     assert call_ids(calls_low_level) == expected_ids
     assert call_ids(calls_high_level) == expected_ids
+    assert len(calls_high_level) == len(calls_low_level)
```
tssweeney, line 129: nit: we are relying on an implementation detail: that get_calls returns a CallsIter that internally uses the stats endpoint to determine length. Let's take a world were we adding an optimization where after iterating through the list we update an internal size param so we can skip the stats call. We could do that and this test would not know any better. But as a result, it is no longer testing the functionality you desire. I'd recommend just explicitly calling the stats endpoint and checking that

tssweeney (approved):
Approving with comment

tests/trace/test_scored_calls_query.py
gtarpenning: yup adding

PR #3640, opened 2025-02-09
chore(weave): Add codegen for trace server interface

andrewtruong:
- **basic tools**
- **update pyproject**
- **add verbose client util**
- **starting stainless trace server**
- **server info**
- **call start**
- **more calls**
- **some objs**
- **more calls**
- **ops**
- **objs**
- **test**
- **update tools**
- **update init**
- **casting**
- **test**
- **update stainless config**
- **update stainless**
- **update remaining endpoints**
- **comment out tables query stream endpoint**
- **test**
- **update pin**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **test**
- **implement reference server**
- **test**
- **test**
- **test**
- **test**


tools/codegen/generate_config.yaml.template
andrewtruong, line 1: Template file for users to configure codegen.  Their `generate_config.yaml` wont be checked in

tools/codegen/stainless.js
andrewtruong, line 1: The stainless CLI -- here as a bundle for now, but hopefully will be standalone and installable in future

weave/utils/verbose_httpx_client.py
andrewtruong, line 1: Example verbose httpx client for easier debugging

weave/trace_server_bindings/stainless_http_trace_server.py
andrewtruong, line 1: This is basically `remote_http_trace_server` converted to use the Stainless client.  Much of the code is the same, except:
1. The generic requests are removed; and
2. Replaced with specific calls to the Stainless client

trace_server_reference/reference_server.py
andrewtruong, line 1: This is a stub implementation of the server.  We'll need additional checks core-side to ensure that it conforms to this interface.

The most reliable way to check this is to ensure both servers return the same `openapi.json`.

trace_server_reference/reference_server.py
```diff
@@ -0,0 +1,241 @@
...
+    req: tsi.CallsQueryStatsReq, auth_params: Auth
+) -> tsi.CallsQueryStatsRes: ...
+
+
+@router.post(
+    "/calls/stream_query",
+    tags=[CALLS_TAG_NAME],
+    # This section is required to get FastAPI to generate the correct OpenAPI schema.
+    response_class=StreamingResponse,
+    responses={
```
andrewtruong, line 123: This is unfortunate, but seems to be an issue with FastAPI

trace_server_reference/reference_server.py
andrewtruong, line 1: TODO:
1. Make an interface definition and generate this automatically with a script?

tools/codegen/generate.py
```diff
@@ -0,0 +1,543 @@
...
+            warning("Force killing server...")
+            server.kill()
+            server.wait()
+
+
+@cli.command()  # type: ignore
+@click.option("--python-path", help="Path to the Python code generation output")
+@click.option("--node-path", help="Path to the Node.js code generation output")
+@click.option("--typescript-path", help="Path to the TypeScript code generation output")
+def generate_code(
```
andrewtruong, line 118: This should branch and never be on main

weave/trace/weave_init.py
```diff
@@ -178,21 +177,23 @@ def init_weave_disabled() -> InitializedClient:
...
-        init_weave_get_server("DISABLED", should_batch=False),
+        init_weave_get_server("DISABLED", "DISABLED", should_batch=False),
         ensure_project_exists=False,
     )
 
     return InitializedClient(client)
 
 
 def init_weave_get_server(
+    username: str | None = None,
```
andrewtruong, line 188: this is an implementation detail leak -- the server does not recognize `username`

tssweeney (changes requested):
I think it would be good to break this into 3 PRs:
1. reference_server. As we discussed, I would not use a flat file, but rather a factory function that accepts a protocol/interface. This way it can be consumed from the trace server service directly.
   - 1.b: I would then create a core PR that uptakes this reference server effectively unifying the service definition layer. We should make sure that docgen works, and deploy this new version.
2. Stainless generation. Have a reliable mechanism for generating the stainless sdk (you already have this in the code-gen stuff). As part of this, please modify your implementation such that you don't install codegen directly on the user machine - this is not correct. Please make it an internal tool using Make.
    2.b - do we need to do anything in actions to ensure that master is always tracking?
3. Finally, a Pr that uptakes the stainless sdk inside the remote trace server. Rather than making a whole new stainless trace server, just make the few edits directly to the remote trace server

PR #3642, opened 2025-02-09
fix(weave): Add better backcompat for materializing older versions of objects like Evaluation

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-23123

This allows older objects that are missing optional keys available on newer objects to be loaded

## Summary by CodeRabbit

- **Tests**
  - Added automated tests to verify that evaluation objects load and perform correctly even when legacy attributes are missing.
  
- **Refactor**
  - Improved object creation processes for data evaluations and datasets, adopting a dynamic approach that adapts to changes and enhances error handling for scorer validations.


tests/trace/test_evaluate.py
```diff
@@ -324,3 +324,38 @@ def score(output):
         "model_latency": {"mean": pytest.approx(1, abs=1)},
     }
     assert time.time() - now < 5
+
+
+def test_evaluation_from_weaveobject_missing_evaluation_name(client):
+    dataset_rows = [{"input": "1 + 2", "target": 3}, {"input": "2**4", "target": 15}]
```
jamie-rasmussen, line 330: dataset_rows is also defined on line 9, any reason to duplicate?

jamie-rasmussen (approved):

tests/trace/test_evaluate.py
andrewtruong: Yes, for some reason it didn't like using the one outside.  I think it's better practice to scope these per test anyways.  We can convert to fixtures in another PR :)

PR #3660, opened 2025-02-12
fix(weave): wrap up datasets using separate "fastlane" future executor

gtarpenning:
## Description

Use a separate futureExecutor for jobs that are guaranteed to not be nested, ensuring the ability to directly execute. Complex potentially nested operations like creating tables/objects/calls stay on the original futureExecutor, but move `file_create` to the new `future_executor_fastlane`. In practice this resolves a longstanding issue rolling up Evaluations, where upload the dataset leads to deadlock and subsequent serial execution of file uploads (row by row baby). 

Adds new `get_parallelism_settings` function that manually counts the number of threads, to allocate 50% to the main executor and 50% to the background executor.

Example usage on a 16 core machine: 
```python
# default experience
client = weave.init("cool-project")
assert client.future_executor._max_workers == 10  # (16 + 4) / 2
assert client.future_executor_fastlane._max_workers == 10

# override total threads
client = weave.init("cool-project", settings={"client_parallelism": 64})
assert client.future_executor._max_workers == 32
assert client.future_executor_fastlane._max_workers == 32
```

## Testing

Examples run on [this](https://wandb.ai/griffin_wb/evaluators/weave/traces) project on a 10 core machine. 
```python
# PROD
client = weave.init("evaluators")
asyncio.run(_run_evaluation())
# > main workers: 14, total seconds ~50s

# default
client = weave.init("evaluators")
asyncio.run(_run_evaluation())
client.flush()
# > main workers: 7, fastlane workers: 7, total seconds ~20s

# override total threads
client = weave.init("evaluators", settings={"client_parallelism": 64})
asyncio.run(_run_evaluation())
client.flush()
# > main workers: 32, fastlane workers: 32, total seconds ~8s
```

- manual testing
- add unit tests

## Summary by CodeRabbit

- **New Features**
	- Enhanced concurrent task execution for operations, including optimizations to file handling, resulting in improved performance and responsiveness.
	- Introduced new settings for parallelism management to better handle task execution.
- **Tests**
	- Added a new test to validate the behavior of parallelism settings under various conditions.
	- Updated existing tests to ensure accurate validation of parallelism configurations.


tssweeney (commented):
I would simplify this a bit. Instead of adding additional logic to the future executor, this seems to be just something that is part of the weave client itself. We simply instantiate a second future executor with the same number of workers as Max parallelism. Then we use that future executor only in thefile call. This means that the future executor does not throw or enforce any business logic on the collar. It's up to the collar to decide what is best. And in this case, the collar knows that the send file features are direct.

tssweeney (approved):

PR #3661, opened 2025-02-12
fix(weave): flush in eval 

gtarpenning:
## Description

Expose `client.flush()` with comment. 

Currently, user scripts that complete before all data has synced causes data to be loaded serially. The background futureExecutor threadpool can not accept jobs once the main thread has completed. As a temporary solution, allow users to call `flush` after evaluation completion, keeping the main thread alive until all work has completed. 

## Summary by CodeRabbit

- **New Features**
  - Added a public operation that ensures all pending background tasks complete before proceeding, enhancing reliability and performance.

- **Documentation**
  - Updated the troubleshooting guide with a new section on managing long evaluation cleanup times. The guide now offers advice on optimizing performance through task flushing, adjusting parallel processing, and clarifies caching behavior.

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -1864,7 +1864,8 @@ class WeaveClient:
         """
         An optional flushing method for the client.
         Forces all background tasks to be processed, which ensures parallel processing
-        during main thread execution.
+        during main thread execution. Can improve performance when user code
+        completes before data has been uploaded to the server.
         """
         self._flush()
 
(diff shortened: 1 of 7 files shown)


weave/trace/weave_client.py
```diff
@@ -1860,6 +1860,15 @@ def _op_runs(self, op_def: Op) -> Sequence[Call]:
     def _ref_uri(self, name: str, version: str, path: str) -> str:
         return ObjectRef(self.entity, self.project, name, version).uri()
 
+    def flush(self) -> None:
+        """
+        An optional flushing method for the client.
+        Forces all background tasks to be processed, which ensures parallel processing
```
tssweeney, line 1866: I think you should add an entry in faq section or troubleshooting docs for evals

tssweeney (approved):

tssweeney (approved):

PR #3688, opened 2025-02-18
feat(weave): Add basic pairwise evaluation support

andrewtruong:
- **test**
- **test**
- **test**
- **test**


weave/flow/eval.py
```diff
@@ -222,7 +223,27 @@ async def predict_and_score(self, model: Union[Op, Model], example: dict) -> dic
         scorers = self._post_init_scorers
 
         for scorer in scorers:
-            apply_scorer_result = await model_call.apply_scorer(scorer, example)
+            if isinstance(scorer, PairwiseScorer):
```
tssweeney, line 226: This seems wrong to couple the logic here. I would have expected the implementation of `PairwiseScorer` to handle this internally

weave/flow/scorer.py
```diff
@@ -365,3 +366,53 @@ async def apply_scorer_async(
...
+    def model_post_init(self, __context: Any) -> None:
+        super().model_post_init(__context)
+
+        from weave.flow.eval import is_valid_model
+
+        if not is_valid_model(self.other_model):
+            raise ValueError("other_model must be a valid Model or Op instance")
+
+    @weave.op
+    async def score(self, output: Any, other_output: Any, **kwargs: Any) -> dict:
```
tssweeney, line 407: I think this should be something like
```
def score(self, output, *args, **kwargs):
    other_score = self.other_model.predct(*args, **kwargs)
    return self.pairwise_score(output, other_score, ....)
```

This move the logical control outside of eval

tests/trace/test_evaluations.py
```diff
@@ -1052,3 +1054,191 @@ def model(input: str) -> str:
...
+    """Test error handling when an invalid model is provided."""
+
+    class InvalidModel:
+        def predict(self, input_text: str):
+            return "invalid"
+
+    class PreferenceScorer(PairwiseScorer):
+        @weave.op
+        async def score(self, output: dict, input_text: str) -> dict:
+            other_output = await self._get_other_model_output(
```
tssweeney, line 1198: this feels awkward that the implementer of the scorer has to explicitly call this boilerplate. Is there anyway that the user can simply fillout a compare function?

andrewtruong: It's a tradeoff of this approach.

One alternative I considered is to let the user fill out a different compare function and then use that inside `score`, but it's also awkward because now the all `Scorers` require you to implement `score` but `PairwiseScorer` requires you to NOT implement `score` and instead implement some other methods

tssweeney: hmmm, i thought you were going to bringup column mapping as the main problem. i feel like if we documented that implementing a `PairwiseScorer` requires implementing the `pairwise_score` method - parallel to `Scorer` implementing the `score` method is actually quite nice - assuming the implementation code has less boilerplate. That being said, i think predict if you decide to go down that path, you'll hit issues with the column mapping.

As implemented today, I feel like this could just be documented as a technical guide - not sure why we need a first-class class if the user still needs to have boilerplate.

J2-D2-3PO: I will document this today

PR #3690, opened 2025-02-18
feat(weave): Add WeaveTrustScorer

tcapelle:
Add WeaveTrustScorer

## Summary by CodeRabbit

- **New Features**
  - Introduced the `WeaveTrustScorerV1` class for evaluating the trustworthiness of model outputs using multiple specialized scorers.
  - Added support for sentence piece tokenization through a new dependency.

- **Tests**
  - Expanded test coverage for `WeaveTrustScorerV1`, including validation of scoring functionality, text preprocessing, and input filtering.
  - Added new tests for scoring logic and parallel execution capabilities.
  - Implemented new fixtures to enhance testing of the `WeaveTrustScorerV1`.


tests/scorers/test_trust_scorer.py
```diff
@@ -0,0 +1,105 @@
+import pytest
+
+from tests.scorers.test_utils import TINY_MODEL_PATHS
+from weave.scorers.trust_scorer import WeaveTrustScorerV1
+from weave.scorers.utils import WeaveScorerResult
+
+
+# Dummy scorer to test _filter_inputs_for_scorer functionality.
+class DummyScorer:
```
andrewtruong, line 9: shouldn't this inherit from Scorer?

tests/scorers/test_trust_scorer.py
```diff
@@ -0,0 +1,105 @@
...
+    def score(self, output: str, query: str):
```
andrewtruong, line 10: shouldn't this have an op deco?

tests/scorers/test_trust_scorer.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test_score_all_with_dummy_scorers(trust_scorer):
+    # Replace the loaded scorers with dummy ones to simulate controlled outcomes.
+    class AlwaysPassScorer:
+        def score(self, **kwargs):
+            return WeaveScorerResult(passed=True, metadata={"score": 0.1})
+
+    class AlwaysFailScorer:
+        def score(self, **kwargs):
+            return WeaveScorerResult(passed=False, metadata={"score": 0.9})
```
andrewtruong, line 61: I recognize these are dummies (as are the ones below), but shouldn't they also inherit from Scorer and have an op deco?

Technically this does work, but it doesn't show the user what the happy path looks like

tests/scorers/test_trust_scorer.py
andrewtruong, line 1: I would expect you to call `scorer.score(...)` somewhere in this file to show the user what to expect when they call the scorer.  Currently this just tests the implementation

weave/scorers/trust_scorer.py
```diff
@@ -0,0 +1,376 @@
...
+        description="Path or name of the coherence model",
+        validate_default=True,
+    )
+    run_in_parallel: bool = Field(
+        default=True,
+        description="Whether to run scorers in parallel or sequentially, useful for debugging.",
+    )
+
+    # Define scorer categories
+    _critical_scorers: set[type[weave.Scorer]] = {
```
andrewtruong, line 151: I'm surprised this works.  Doesn't it have to be a private attr?

weave/scorers/trust_scorer.py
```diff
@@ -0,0 +1,376 @@
...
+        base_params: dict[str, Any] = {
+            "column_map": self.column_map,
+            "device": self.device,
+        }
+
+        # Load all scorers (both critical and advisory)
+        all_scorers = self._critical_scorers | self._advisory_scorers
+
+        for scorer_cls in all_scorers:
+            scorer_params = base_params.copy()
```
andrewtruong, line 193: Do you need this?  You are instantiating base_params fresh every time, so this is just an extra copy

weave/scorers/trust_scorer.py
```diff
@@ -0,0 +1,376 @@
...
+        inputs: dict[str, Any] = {"output": processed_output}
+        if processed_context is not None:
+            inputs["context"] = processed_context
+        if processed_query is not None:
+            inputs["query"] = processed_query
+
+        results = {}
+
+        if self.run_in_parallel:
+            with ThreadPoolExecutor() as executor:
```
andrewtruong, line 270: You will want to add a test in the Evaluation framework for this flag.  In there, we also have thread parallelism, and it's not clear how these two will interact

andrewtruong: Basically just add the trust scorer as a scorer and run the eval framework across ~100 rows

weave/scorers/trust_scorer.py
```diff
@@ -0,0 +1,376 @@
...
+                        )
+        else:
+            # Run scorers sequentially
+            for scorer_name, scorer in self._loaded_scorers.items():
+                try:
+                    results[scorer_name] = scorer.score(
+                        **self._filter_inputs_for_scorer(scorer, inputs)
+                    )
+                except Exception:
+                    pass
```
andrewtruong, line 295: Is this correct?

weave/scorers/trust_scorer.py
```diff
@@ -0,0 +1,376 @@
...
+                scores[name] = result.extras  # Toxicity returns category scores
+            elif hasattr(result, "extras") and "score" in result.extras:
+                scores[name] = result.extras["score"]
+
+        return {
+            "passed": passed,
+            "trust_level": trust_level,
+            "critical_issues": critical_issues,
+            "advisory_issues": advisory_issues,
+            "extras": {"raw_outputs": raw_results, "scores": scores},
```
andrewtruong, line 347: extras -> metadata?

weave/scorers/trust_scorer.py
```diff
@@ -0,0 +1,376 @@
...
+        }
```
andrewtruong, line 348: Why not just return the metadata dict that you use below?  Seems like you have to reconstruct it needlessly

tests/scorers/test_trust_scorer.py
```diff
@@ -0,0 +1,107 @@
...
+    assert results["AlwaysPass"].passed
+    assert not results["AlwaysFail"].passed
+
+
+def test_score_with_logic(trust_scorer):
+    # Create dummy critical and advisory scorers to verify nuanced trust logic.
+    class DummyCriticalScorer:
+        def score(self, **kwargs):
+            return WeaveScorerResult(
+                passed=False, metadata={"score": 0.8}, extras={"score": 0.8}
```
andrewtruong, line 82: not sure what these extras are

andrewtruong: The fact that they work suggests there's something wrong with WeaveScorerResult.  It shouldn't accept arbitrary extra keys

tests/scorers/test_trust_scorer.py
tcapelle: agree

tests/scorers/test_trust_scorer.py
tcapelle: yes they should

weave/scorers/trust_scorer.py
tcapelle: yeap, me to

weave/scorers/trust_scorer.py
tcapelle: This scorer is slow, do we want to really do this?

andrewtruong: Well you need to test the parallelism somehow.  It's not obvious to me that this is safe

weave/scorers/trust_scorer.py
tcapelle: yeap, remove this

tests/scorers/test_trust_scorer.py
tcapelle: this is wrong

weave/scorers/trust_scorer.py
```diff
@@ -0,0 +1,376 @@
...
+
+    @validate_call
+    @weave.op
+    def score(
+        self,
+        query: str,
+        context: Union[str, list[str]],
+        output: str,  # Pass the output of a LLM to this parameter for example
```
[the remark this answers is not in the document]
tcapelle: nothing is optional, everything is mandatory here =)

weave/scorers/trust_scorer.py
tcapelle: removed

weave/scorers/trust_scorer.py
tcapelle: we can turn off parallelism by default for the time being

andrewtruong: no, if you include parallelism then you need to test it.  Otherwise you have to remove it

tests/scorers/test_trust_scorer.py
```diff
@@ -0,0 +1,138 @@
...
+            "extra": "should_be_filtered",
+        }
+    ] * 100
+
+    @weave.op
+    def model(output):
+        return output
+
+    evaluation = weave.Evaluation(dataset=ds, scorers=[model])
+    asyncio.run(evaluation.evaluate(model))
```
andrewtruong, line 138: You should get the data back and assert it's correct

tcapelle: I am running a dummy model, there is real output here

andrewtruong: yeah but you need asserts to make sure the scores are correct -- see the eval tests for examples :)

tcapelle: Ok, made it work =)

weave/scorers/trust_scorer.py
tcapelle: added a test

tests/scorers/test_trust_scorer.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @weave.op
+    def model(output):
+        return output
+
+    evaluation = weave.Evaluation(dataset=ds, scorers=[trust_scorer])
+    eval_result = asyncio.run(evaluation.evaluate(model))
+    assert "passed" in eval_result["WeaveTrustScorerV1"]
+    assert "metadata" in eval_result["WeaveTrustScorerV1"]
+    assert len(eval_result["WeaveTrustScorerV1"]["metadata"]["raw_outputs"]) == 5
```
andrewtruong, line 141: You also know the sub-scorer scoring results because you mocked them right?  So you can add the scores here too I think

tcapelle: Ok!

tcapelle:
@SocketSecurity ignore pypi/sentencepiece@0.2.0

PR #3702, opened 2025-02-19
feat(weave): Add datetime serializer

andrewtruong:
- **test**
- **test**

### DIFF
--- a/weave/type_handlers/DateTime/datetime.py
+++ b/weave/type_handlers/DateTime/datetime.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import datetime
 import json
-from typing import Any
+from typing import TextIO, cast
 
 from weave.trace import serializer
 from weave.trace.custom_objs import MemTraceFilesArtifact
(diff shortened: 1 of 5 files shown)


weave/type_handlers/DateTime/datetime.py
```diff
@@ -0,0 +1,59 @@
...
+
+import datetime
+import json
+from typing import TextIO, cast
+
+from weave.trace import serializer
+from weave.trace.custom_objs import MemTraceFilesArtifact
+
+
+class DatetimeWithArtifact(datetime.datetime):
```
tssweeney, line 13: can you help me understand why we need this? Seems like a fault of some internal problem?

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/datetime.datetime/DateTimeView.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  // Handle null/undefined cases early
+  if (data == null) {
+    return <NotApplicable />;
+  }
+
+  // Handle direct ISO string input
+  if (typeof data === 'string') {
+    const stringTimestamp = new Date(data).getTime() / 1000;
+    return <Timestamp value={stringTimestamp} format="relative" />;
```
jamie-rasmussen, line 41: I think we shouldn't use relative time for this by default - can be disconcerting for user to see their call output appear to change its value.

A few ideas for follow up:
* We could make the component a little fancier like we do in `CompareGridCellValueTimestampProps` - user could click on the value and have it cycle through the absolute string representation, relative time, unix epoch int etc.
* We should probably do something to make it more visually obvious that what was logged was a datetime object and not a string representation - maybe a little icon prefix like this:
Screenshot 2025-02-20 at 1 27 50 PM

jamie-rasmussen (approved):

PR #3713, opened 2025-02-20
fix(weave_query): Handle escaped dots in run-history

domphan-wandb:
## Description

Fixes [WB-23402](https://wandb.atlassian.net/browse/WB-23402)

A few weeks ago there was a bug with `run.history` not returning dictionary values correctly because nested keys were flattened. We shipped [a fix](https://github.com/wandb/weave/pull/3536) that unescaped dots in the keys from pick operations to resolve the bug.

This PR fixes a regression with pick keys that have dots. They were being unescaped and since we overload the dot to provide key path and for use as a character, this led to X/Y selectors in panel plots like "row["mAP-val-0\.5:0\.95"]" not working correctly.

Now we unescape dots during compilation specifically for `run-history` ops.

## Testing
Tested through local dev

Pic of plot working again with selected X/Y dim:
image

Pic of history working with nested dict:
image

Pic of history pick working with nested dict:
image

[WB-23402]: https://wandb.atlassian.net/browse/WB-23402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected data formatting issues to ensure outputs present fully accurate column labels and values, enhancing reliability and consistency in reporting.
  - Ensured that downstream data displays now faithfully reflect the intended formatting.

- **Refactor**
  - Streamlined internal data processing routines to reduce complexity and boost performance during data transformation and retrieval.
  - Refined processes improve overall data handling efficiency for a smoother user experience.
  - Modified handling of column names to ensure proper processing of special characters.

### DIFF
--- a/weave_query/weave_query/compile.py
+++ b/weave_query/weave_query/compile.py
@@ -30,6 +30,7 @@ from weave_query import (
 )
 from weave_query.language_features.tagging import tagged_value_type_helpers
 from weave_query.op_def import OpDef
+from weave_query._dict_utils import unescape_dots
 
 # These call_* functions must match the actual op implementations.
 # But we don't want to import the op definitions themselves here, since
@@ -464,7 +465,7 @@ def compile_apply_column_pushdown(
                 )
             if "run-history" in node.from_op.name:
                 history_cols = list(run_cols.keys())
-
+                history_cols = [unescape_dots(hc) for hc in history_cols]
                 if len(history_cols) > 0:
                     return graph.OutputNode(
                         node.type,
--- a/weave_query/weave_query/graph.py
+++ b/weave_query/weave_query/graph.py
@@ -4,7 +4,7 @@ import typing
 
 from weave_query import weave_types
 from weave_query import uris, errors, storage
-from weave_query._dict_utils import unescape_dots
+
 if typing.TYPE_CHECKING:
     from weave_query import weave_inspector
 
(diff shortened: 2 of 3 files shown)


nicholaspun-wandb (commented):
Approving to resolve the immediate issue. Can we file a ticket to track the larger issue? This fix doesn't break [WB-22947](https://wandb.atlassian.net/browse/WB-22947), but a general solution to handle escaped dots in picks still need to be addressed (e.g. `runs.summary["level_one\.level_two"]`)

nicholaspun-wandb (approved):

domphan-wandb:
> Approving to resolve the immediate issue. Can we file a ticket to track the larger issue? This fix doesn't break [WB-22947](https://wandb.atlassian.net/browse/WB-22947), but a general solution to handle escaped dots in picks still need to be addressed (e.g. `runs.summary["level_one\.level_two"]`)

Okay, @nicholaspun-wandb and I discussed and nested dicts in run.summary function correctly by not escaping the dots at all.  This is different behavior than run.history. Not going to spend more time on this, but we created a ticket to keep track of this: https://wandb.atlassian.net/browse/WB-23421

tssweeney (approved):

PR #3715, opened 2025-02-20
chore(weave): Unblock scorers CI

tcapelle:
Fix scorers CI, missing dep.

Note: This appeared because I changed some of the underlying model checkpoints to use legacy tokenizer.


tcapelle:
@SocketSecurity ignore pypi/sentencepiece@0.2.0

andrewtruong:
Already fixed in HF (needs to be done in one place to pass CI)

PR #3717, opened 2025-02-20
chore(weave): dump apply_scorer response to json

gtarpenning:
## Description

Sdk for dump scorers --> we want the dict repr instead of class

## Testing

- add test

## Summary by CodeRabbit

- **New Features**
  - Improved handling of feedback submissions for `WeaveScorerResult`, allowing for enhanced output processing and serialization.
  - Added a new asynchronous test to validate scorer behavior with `WeaveScorerResult` outputs.

- **Refactor**
  - Enhanced the feedback submission process by standardizing output handling to ensure more consistent and reliable data formatting before it’s sent to the server.


weave/trace/weave_client.py
```diff
@@ -1568,11 +1568,15 @@ def _add_runnable_feedback(
...
+
         # Prepare the result payload - we purposely do not map to refs here
         # because we prefer to have the raw data.
-        results_json = to_json(output, self._project_id(), self)
-
-        # # Prepare the supervision payload
+        if isinstance(output, WeaveScorerResult):
+            results_json = output.model_dump_json()
+        else:
+            results_json = to_json(output, self._project_id(), self, True)
```
jamie-rasmussen, line 1578: Should we take out the `use_dictify`

gtarpenning: oh, yes

weave/trace/weave_client.py
```diff
@@ -1568,12 +1568,16 @@ def _add_runnable_feedback(
...
             raise TypeError(f"Invalid scorer op ref: {runnable_ref_uri}")
 
+        from weave.scorers.utils import WeaveScorerResult
+
         # Prepare the result payload - we purposely do not map to refs here
         # because we prefer to have the raw data.
-        results_json = to_json(output, self._project_id(), self)
-
-        # # Prepare the supervision payload
+        if isinstance(output, WeaveScorerResult):
```
tssweeney, line 1575: Please add a note that this is a super hack and should be moved somehow into to_json or something else. We should not need to look for `WeaveScorerResult` as customer defined scores could return pydantics

tssweeney: I think that adds to the hackiness and risk. The "correct" solution would be to fix to_json (and more generally our serialization layer) to be clear, centralized, and well defined. I am hesitant increase the surface of this particular shim

tests/scorers/test_call_apply_scorer.py
```diff
@@ -209,3 +210,64 @@ async def score(self, x, output):
         scorer, additional_scorer_kwargs={"correct_answer": 2}
     )
     do_assertions_for_scorer_op(apply_score_res, call, scorer, client)
+
+
+# @pytest.mark.asyncio
```
tssweeney, line 215: leftover?

tssweeney (approved):
Approving, but we really should fix this one correctly in the near-term. I worry that this is going to awkwardly stick around.

weave/trace/weave_client.py
andrewtruong: Let's do this:
https://github.com/wandb/weave/commit/e16aaea3e5ef6288d90e285e78a7e680ecce9ab0

andrewtruong: @tssweeney wdyt ^?

PR #3724, opened 2025-02-20
fix(weave): wait for async inserts to clear buffer

gtarpenning:
## Description

Candidate fix for ~ some ~ dropped inserts. According to [this ch doc](https://clickhouse.com/docs/en/optimize/asynchronous-inserts#enabling-asynchronous-inserts): "Our strong recommendation is to use async_insert=1,wait_for_async_insert=1 if using asynchronous inserts. Using wait_for_async_insert=0 is very risky because your INSERT client may not be aware if there are errors, and also can cause potential overload if your client continues to write quickly in a situation where the ClickHouse server needs to slow down the writes and create some backpressure in order to ensure reliability of the service." 

Swallowing insert errors sounds pretty bad, even if we have to take a small performance hit. Let's follow their "strong recommendation"? In practice there is no noticeable performance hit. 

## Testing

qa

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted asynchronous data submission settings to ensure operations complete before proceeding, which enhances error detection during background processing.


andrewtruong (approved):
Worth landing separately so we have time to see if it has any particular effects?

PR #3734, opened 2025-02-21
chore(weave): fix to async langchain batch integration 

gtarpenning:
## Description

[WB-23410](https://wandb.atlassian.net/browse/WB-23410)

Hack to unblock langchain integration. 

Pros: 
- this solves the error case 100% of the time

Cons: 
- the parent is not usable in the stack

## Testing

[prod](https://wandb.ai/griffin_wb/my_openai_app/weave/calls?cols=%7B%22attributes.weave.client_version%22%3Afalse%2C%22attributes.weave.os_name%22%3Afalse%2C%22attributes.weave.os_release%22%3Afalse%2C%22attributes.weave.os_version%22%3Afalse%2C%22attributes.weave.source%22%3Afalse%2C%22attributes.weave.sys_version%22%3Afalse%7D&peekPath=%2Fgriffin_wb%2Fmy_openai_app%2Fcalls%2F019529da-7a55-7c03-ba13-ca9a87474fc5%3Ftracetree%3D1%26feedbackExpand%3D0)

[branch](https://wandb.ai/griffin_wb/my_openai_app/weave/calls?cols=%7B%22attributes.weave.client_version%22%3Afalse%2C%22attributes.weave.os_name%22%3Afalse%2C%22attributes.weave.os_release%22%3Afalse%2C%22attributes.weave.os_version%22%3Afalse%2C%22attributes.weave.source%22%3Afalse%2C%22attributes.weave.sys_version%22%3Afalse%7D&peekPath=%2Fgriffin_wb%2Fmy_openai_app%2Fcalls%2F01952a17-6899-7c63-9694-2793aeca187f%3Ftracetree%3D1%26feedbackExpand%3D0)

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the handling of parent runs by shifting from remote lookups to efficient in-memory processing.
	- Removed redundant dependencies for a cleaner, optimized system implementation.
- **Tests**
	- Enhanced testing for batch processing with new and modified test cases to verify output and error handling.
	- Introduced a new test case specifically for handling parent runs in batch operations. 
	- Captured output during tests to assert correct logging behavior. 
- **Chores**
	- Added a new YAML file to facilitate integration testing of OpenAI API interactions.

[WB-23410]: https://wandb.atlassian.net/browse/WB-23410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/integrations/langchain/langchain.py
```diff
@@ -180,11 +180,17 @@ def _persist_run_single(self, run: Run) -> None:
...
-                            # Note: this is implemented as a network call - it would be much nice
-                            # to refactor `create_call` such that it could accept a parent_id instead
-                            # of an entire Parent object.
-                            parent_run = cast(
-                                Call, self.gc.get_call(wv_current_run.parent_id)
+                            # Hack in memory parent call to satisfy `create_call`
+                            # Impact is the parent won't actually represent the parent call
+                            # so we do NOT want to save the parent to the stack
+                            use_stack = False
+                            parent_run = Call(
```
andrewtruong, line 187: this is basically a placeholder for the parent and trace_ids right?

andrewtruong: you assume they will exist eventually, they just dont at this very instance

bcsherma (approved):

tests/integrations/langchain/langchain_test.py
```diff
@@ -198,6 +199,12 @@ def test_simple_chain_batch(
     calls = list(client.calls(filter=tsi.CallsFilter(trace_roots_only=True)))
     assert_correct_calls_for_chain_batch(calls)
 
+    log_lines = capsys.readouterr().out
+
+    # one parent call link
+    assert log_lines.count("/shawn/test-project/r/call") == 1
```
tssweeney, line 205: This style of assertion is very fragile. We should avoid asserting on IO unless we are explicitly trying to test I/O

gtarpenning: fair, although this kind of test would have caught the issue instead of a user reporting it. But I can remove it if you think that it's too fragile

weave/integrations/langchain/langchain.py
```diff
@@ -180,11 +180,17 @@ def _persist_run_single(self, run: Run) -> None:
...
                         else:
-                            # Note: this is implemented as a network call - it would be much nice
-                            # to refactor `create_call` such that it could accept a parent_id instead
-                            # of an entire Parent object.
-                            parent_run = cast(
-                                Call, self.gc.get_call(wv_current_run.parent_id)
+                            # Hack in memory parent call to satisfy `create_call`
+                            # Impact is the parent won't actually represent the parent call
+                            # so we do NOT want to save the parent to the stack
+                            use_stack = False
```
tssweeney, line 186: use_stack controls if the new call is added to the stack. This seems incorrect to me

gtarpenning: ah, controls both adding to parent children and to call stack. The intention was to not even try to construct the correct in memory repr as we will not be constructing the parent correctly in this parallel context. But we def could still add the calls to the `call_context`...

tests/integrations/langchain/langchain_test.py
tssweeney: It is just a very roundabout way of testing something. i like the notion of making sure errors are not emitted in stdout. but that should be done globally (checkout `logging_error_check` where i do exactly this)

The "thing" you are trying to fix is trace topology

PR #3735, opened 2025-02-21
chore(weave_query): add file ops for artifact memberships

ibindlish:
## Description

[WB-23340](https://wandb.atlassian.net/browse/WB-23340)

Context:
- Developing a new feature to create a role for W&B Registry that allows reading artifact metadata but not its files' contents
- Role-based auth needs to be resolved using collection information attached while querying the artifact. For this reason, the artifact membership is a better api to access since the artifact version uses the ID to hit the artifact directly, and we lose the collection context

- Upstream PR that exposes files() endpoint on the ArtifactCollectionMembership type https://github.com/wandb/core/pull/27563/
- ~~Upstream PR that fixes mds query to fetch an artifact using its commitHash, even if we're fetching the artifact via its portfolio membership https://github.com/wandb/core/pull/27670~~ This is no longer needed since i allowed the resolved uri to use the version index instead of ONLY the commit hash

- This PR adds the files op to artifactMembership (similar to the ops on the artifactVersion)

## Testing

- I modified the UI code in `PanelArtifactVersionFiles` to take an artifactMembershipNode as input instead of an artifactVersionNode, using which I fetched its files via the artifactMembership-file() Op. (temp branch with changes using which i tested the ops' results: https://github.com/wandb/core/pull/27717)
Was able to load the artifact files panels: 

image

## Summary by CodeRabbit

- **New Features**
  - Enhanced artifact membership operations by adding file retrieval capabilities with improved error handling and type resolution.
  - Centralized and standardized query definitions for artifact file operations, ensuring more consistent data handling.
  - Expanded GraphQL field generation to support the new artifact membership file retrieval, improving end-to-end system integration.
  - Introduced new operations for refining output types and handling file-related artifacts. 

- **Bug Fixes**
  - Improved version validation logic in the artifact URI resolution process for better accuracy.

[WB-23340]: https://wandb.atlassian.net/browse/WB-23340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave_query/weave_query/artifact_wandb.py
```diff
@@ -994,7 +994,7 @@ def uri_path(self) -> str:
 
     @property
     def resolved_artifact_uri(self) -> "WeaveWBArtifactURI":
-        if self.version and likely_commit_hash(self.version):
+        if self.version and (likely_commit_hash(self.version) or is_valid_version_index(self.version)):
```
ibindlish, line 997: As per this comment, the `resolved_artifact_uri` just needs to be a specific, immutable version like v4 instead of an alias for idempotency of cache keys. The existing check enforced the version in the uri to be a commit hash which caused issues while fetching the artifact in a portfolio collection since we don't store the commit hashes for those memberships. 

I initially went about updating the `FetchArtifactByCommitHash` to allow fetching an artifact via its portfolio membership and sequence commit hash but this fix to use the version index instead makes more sense. (Pr here https://github.com/wandb/core/pull/27670)

tssweeney (approved):

PR #3738, opened 2025-02-21
fix(weave): Only print root donut link

andrewtruong:
## Summary by CodeRabbit

- **Refactor**
  - Improved call link display so that only a single, accurate link is printed when nested operations complete successfully.
  
- **Tests**
  - Introduced a test that validates nested operations produce the expected number of calls and correctly formatted output.

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -1120,6 +1120,7 @@ class WeaveClient:
         project_id = self._project_id()
 
         _should_print_call_link = should_print_call_link()
+        _current_call = call_context.get_current_call()
 
         def send_start_call() -> bool:
             maybe_redacted_inputs_with_refs = inputs_with_refs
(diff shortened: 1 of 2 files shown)


weave/trace/weave_client.py
```diff
@@ -1151,9 +1152,8 @@ def send_start_call() -> bool:
 
         def on_complete(f: Future) -> None:
             try:
-                if f.result() and not call_context.get_current_call():
-                    if _should_print_call_link:
-                        print_call_link(call)
+                if _should_print_call_link and f.result() and not _current_call:
```
gtarpenning, line 1155: for readability can you assign this to a var or something? I don't think its immediately obvious why we wouldn't want to print the call if "not current call".

gtarpenning (approved):

weave/trace/weave_client.py
andrewtruong: yep

PR #3747, opened 2025-02-24
in progress

jamie-rasmussen:
## Description

Internal Slack: https://weightsandbiases.slack.com/archives/C0880TX4U5V/p1739987386042209

User was looking over the children of an evaluation and for each `predict_and_score` (~500 of them) was trying to get a value from the "example", which is a ref to a dataset row. Unfortunately, this causes the entire dataset to be retrieved. The user was doing this twice per child, so the end result was downloading the whole dataset about a 1000 times.

There are two changes here:
1. Constructing a `TableRowFilter` when the `TableRef` has extra  pointing to a particular row.
2. Passing the `TableRowFilter` through to the Trace Server API. I unfortunately don't know why this was commented out, but it seems to work.

## Testing

Created a dataset and ran code similar to what user was doing. First run is with this PR, second is master.
```
Logged in as Weights & Biases user: jamie-rasmussen.
View Weave data at https://wandb.ai/jamie-rasmussen/2025-02-20_slow-dataset/weave
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 483/483 [01:52<00:00,  4.28it/s]
inputs Time taken: 113.03 seconds
(.venv) ➜  suprot_2025-02-17 python slow_dataset.py
Logged in as Weights & Biases user: jamie-rasmussen.
View Weave data at https://wandb.ai/jamie-rasmussen/2025-02-20_slow-dataset/weave
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 483/483 [05:43<00:00,  1.40it/s]
inputs Time taken: 344.06 seconds
```


weave/trace/vals.py
```diff
@@ -703,11 +703,20 @@ def make_trace_obj(
 
             # need to deref if we encounter these
             if isinstance(val, TableRef):
+                table_row_filter = TableRowFilter()
```
tssweeney, line 706: Can you write a unit test that hits this path and demonstrates a reduced query?

jamie-rasmussen: Added

PR #3765, opened 2025-02-25
feat(weave): Add method to list objects

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-22237

## Summary by CodeRabbit

- **New Features**
  - Enhanced object version retrieval now groups historical object changes into intuitive collections. Users benefit from flexible filtering options and efficient lazy loading for faster, smoother access to object histories.


weave/trace/weave_client.py
```diff
@@ -1935,6 +1935,35 @@ def _objects(self, filter: ObjectVersionFilter | None = None) -> list[ObjSchema]
...
+        self,
+        *,
+        filter: ObjectVersionFilter | None = None,
+    ) -> list[ObjectVersionCollection]:
+        """
+        Get objects with versions grouped by object ID.
+
+        Args:
+            filter: Filter to apply to the objects query
+            limit: Maximum number of object collections to return
```
jamie-rasmussen, line 1949: limit does not appear to be an arg

weave/trace/weave_client.py
```diff
@@ -1935,6 +1935,35 @@ def _objects(self, filter: ObjectVersionFilter | None = None) -> list[ObjSchema]
...
+
+        Returns:
+            A list of ObjectVersionCollection objects, each containing versions of a single object
+
+        """
+        all_objects = self._objects(filter)
+        from collections import defaultdict
```
jamie-rasmussen, line 1956: Why not import at top?

weave/trace/weave_client.py
andrewtruong: fixed

weave/trace/weave_client.py
andrewtruong: done

weave/trace/weave_client.py
```diff
@@ -2111,4 +2196,71 @@ def elide_display_name(name: str) -> str:
     return name
 
 
-__docspec__ = [WeaveClient, Call, CallsIter]
+class ObjectVersionCollection:
+    """A collection of object versions for a single object.
+
+    This class provides access to all versions of a specific object.
+    By default, versions are sorted by version_index in ascending order
```
jamie-rasmussen, line 2203: Doesn't the constructor always sort?

weave/trace/weave_client.py
```diff
@@ -2111,4 +2196,71 @@ def elide_display_name(name: str) -> str:
...
+            )
+            self._weave_objects[index] = self._client.get(ref)
+
+        return self._weave_objects[index]
+
+    @property
+    def base_object_class(self) -> str | None:
+        """Get the base object class of this object."""
+        if self._raw_versions:
+            return self._raw_versions[0].base_object_class
```
jamie-rasmussen, line 2262: One unfortunate thing is we allow different types of objects to be published with the same id. Worth considering how that impacts things like this.

weave/trace/weave_client.py
```diff
@@ -2111,4 +2196,71 @@ def elide_display_name(name: str) -> str:
     return name
 
 
-__docspec__ = [WeaveClient, Call, CallsIter]
+class ObjectVersionCollection:
+    """A collection of object versions for a single object.
```
jamie-rasmussen, line 2200: The lazy loading seems to be a key motivation for this class, I'd mention it in here.

weave/trace/weave_client.py
```diff
@@ -2111,4 +2196,71 @@ def elide_display_name(name: str) -> str:
...
+            version = self._raw_versions[index]
+            entity, project = version.project_id.split("/", 1)
+
+            ref = ObjectRef(
+                entity=entity,
+                project=project,
+                name=version.object_id,
+                _digest=version.digest,
+            )
+            self._weave_objects[index] = self._client.get(ref)
```
jamie-rasmussen, line 2254: I think we generally do want the "objectified" version, e.g. a Dataset if that's what it is. However, there is some weirdness because some of the useful fields, like `wb_user_id`, `created_at`, and `version_index` are at the `ObjSchema` level. Maybe the best way to solve this would be to expose those as fields on the Dataset? But otherwise I think we would want `ObjectVersionCollection` to give you a way to get the `ObjSchema` out.

weave/trace/weave_client.py
andrewtruong: I believe the objects always come back in order.  This comment is just noting to users what the order is

weave/trace/weave_client.py
andrewtruong: are you saying the unique key (and therefore the grouping) is `(base_object_class, object_id)`?

weave/trace/weave_client.py
andrewtruong: done

weave/trace/weave_client.py
andrewtruong: I had these included on the ObjectVersionCollection in one iteration, but removed it.  Do you think it's worth adding back to the collection level?

weave/trace/weave_client.py
jamie-rasmussen: I found the comment confusing, maybe because it talks about "when created" but it sounds like you are talking about API response, and also this is something the user of the higher level APIs doesn't have control over.

weave/trace/weave_client.py
jamie-rasmussen: Yes, I think there's a better longer timeframe solution, but maybe useful to have something like `get_schema(index)` on the `ObjectVersionCollection`?

weave/trace/weave_client.py
jamie-rasmussen: I think you'd still want to group them by object_id alone, it is just that the base_object_class for the first version is not necessarily going to be the base_object_class for all versions.

andrewtruong: That seems bad -- when/why would the object class change across versions?

jamie-rasmussen: You can reproduce this with just:
```
weave.publish(['felix', 'jimbo', 'billie'], 'same-name')
weave.publish({"foo": "bar"}, 'same-name')
weave.publish("hello", 'same-name')
```
As to why we allow it, possibly just an oversight in the original design, or possibly it was intentional to allow you to change the type of objects without losing the semantic connected-ness across versions of the same object, e.g. maybe you change a plain string used as a prompt to a StringPrompt object or something like that.

PR #3766, opened 2025-02-25
chore(ui): Add to dataset design updates 

m-rgba:
## Description

| Before | After |
| -------- | ------- |
| before | after |

Updates:
- Spacing updates.
- Swap checkbox to left-hand side for show latest version.
- Uses checkboxes instead of custom radio for field selection.
- Uses table-like structure - updates select all / none / some functionality.
- Allows field-selection to be show before dataset name is entered.

## Summary by CodeRabbit

- **New Features**
  - Enhanced dataset selection to clearly distinguish between creating a new dataset and choosing an existing one.

- **UI Enhancements**
  - Upgraded the field selection interface with a standard checkbox that supports partial selections.
  - Improved layout and spacing across dataset schema mapping and selection steps for a cleaner, more intuitive experience.


m-rgba:
Updated Featurebee:
https://beta.wandb.ai/?betaVersion=a4342ebc6acbc1bfdd769392a1980a0e1d94f87a

bcsherma (approved):
Goodbye, custom radios 🥲

m-rgba:
> Goodbye, custom radios 🥲

I actually really liked the style of them, just thought the checkbox might be better so we can do that select all / some / none logic like a table.

PR #3771, opened 2025-02-25
docs(weave): Move pairwise workaround to FAQ

J2-D2-3PO:
## Description

Fixes https://wandb.atlassian.net/browse/DOCS-1290

- Moves pairwise workaround  doc to FAQ
- Adds note that this is a workaround and may change in the future

## Testing

`yarn start` on local


J2-D2-3PO:
This PR is not relevant anymore, we have https://github.com/wandb/weave/pull/3772

PR #3779, opened 2025-02-26
docs(weave): document saved views

J2-D2-3PO:
## Description

- Fixes https://wandb.atlassian.net/browse/DOCS-1105

_Summary:_ Documents new UI feature Saved Views:
_Applies to:_ Traces & Evals tabs
_What is documented:_ Table settings on the Traces and Evaluations pages like filters, sorting, page size, column visibility, and column pinning can be saved as a View.
Views can be accessed / updated & deleted from the accordion button in upper left corner of Traces & Evals pages. Views can be renamed by clicking the view name and editing the field.

_Feature PR:_ https://github.com/wandb/weave/pull/3366

## Notes for reviewer

https://github.com/wandb/weave/pull/3779#issuecomment-2689357131

## Testing

- [x] `yarn start` on local
- [ ] SME review

## Summary by CodeRabbit

- **Documentation**
	- Introduced new guides detailing the "Saved Views" feature, enabling users to save and access customized table configurations using both the UI and Python SDK.
	- Updated navigational hyperlinks to improve documentation structure and ensure accurate reference links.
	- Enhanced user guidance with a new sidebar entry linking to the saved views documentation.


J2-D2-3PO:
@jamie-rasmussen a few notes for your review:

1. Should I call it "view" or "Views"? I went with the latter but now it seems like I should just call it a "view" to match UI / it isn't a code construct like `Model`
2. I more or less duplicated the instructions across Traces and Evals subsections. I feel like both locations I chose are not great, but I'm trying to avoid adding more pages to the sidebar since we are about to clean that out anyways. Any thoughts on location?
3. I noticed that there is this handy UI text 
image
If we push this docs PR as is, there would be two separate sections, one for Traces at `/guides/tracking/tracing#create-a-trace-view-in-the-ui` and the other for Evals at `/guides/core-types/evaluations#create-an-evals-view-in-the-ui` so there would need to be different links depending on which tab the user is in

docs/docs/guides/tracking/tracing.md
```diff
@@ -739,6 +739,46 @@ Calculated Fields:
...
+### Create a View
+
+1. Navigate to your **Traces** tab.
+2. Adjust any of the following variables in your table configuration:
+   - Filters
+   - Sort order
+   - Page size
+   - Column visibility
+   - Column pinning
+3. In the upper right hand corner, click **Save view**. Weave automatically saves your View.
```
jamie-rasmussen, line 755: I think it is a little odd to say "automatically saves" when it is an explicit user action

docs/docs/guides/tracking/tracing.md
```diff
@@ -739,6 +739,46 @@ Calculated Fields:
...
+
+### Access a saved View
+
+1. Navigate to your **Traces** tab.
+2. Click the hamburger menu to the right of the **Trace** tab. A dropdown menu showing all saved Views displays.
```
jamie-rasmussen, line 760: I would say to the "left of the Page title" - the navigation sidebar is something that we've reordered many times and might again.

docs/docs/guides/core-types/datasets.md
```diff
@@ -8,7 +8,7 @@ Weave `Dataset`s help you to organize, collect, track, and version examples for
 This page describes:
 
 - Basic `Dataset` operations in Python and TypeScript and how to get started  
-- How to create a `Dataset` in Python and TypeScript from objects such as Weave [calls](../tracking/tracing.mdx)
+- How to create a `Dataset` in Python and TypeScript from objects such as Weave [calls](../tracking/tracing)
```
jamie-rasmussen, line 11: This change seems unrelated to saved views - Ideally we have one PR per logical change, but no need to fix this PR. (Maybe not quite as important for docs as code but helps if we later need to roll something back for example.)

jamie-rasmussen:
1. Can you point me to an example of where you are wondering about singular vs. plural? There will be a SavedView python object shortly.
2. I think ultimately we'll want to give Saved Views their own page and just have the traces and evals pages have a short section linking there. Another section on saved views that will be coming soon is programmatic use in Python. A dedicated page would also cut down on the duplication, which will remove the complication of keeping them in sync, having to have separate links in the UI, and duplicates in search results. But I don't have all the context on the upcoming sidebar changes so I defer to your judgement on the best place to put it and we can revisit as necessary.
3. Good point! Whichever way we go on 2 I can fast follow the code changes to point to the right place. Small FE changes like link location are easy to make and deploy - we should make sure the links are correct before we do a publicity push but I'm not too worried about them being perfect at the time of initial code merge.

J2-D2-3PO:
> 1. Can you point me to an example of where you are wondering about singular vs. plural? There will be a SavedView python object shortly.
> 2. I think ultimately we'll want to give Saved Views their own page and just have the traces and evals pages have a short section linking there. Another section on saved views that will be coming soon is programmatic use in Python. A dedicated page would also cut down on the duplication, which will remove the complication of keeping them in sync, having to have separate links in the UI, and duplicates in search results. But I don't have all the context on the upcoming sidebar changes so I defer to your judgement on the best place to put it and we can revisit as necessary.
> 3. Good point! Whichever way we go on 2 I can fast follow the code changes to point to the right place. Small FE changes like link location are easy to make and deploy - we should make sure the links are correct before we do a publicity push but I'm not too worried about them being perfect at the time of initial code merge.

1. Sorry I meant using "Saved View" vs "saved view" when referring to the general concept. I'm going with "saved view" similar to these docs https://learn.microsoft.com/en-us/dynamics365/fin-ops-core/dev-itpro/get-started/saved-views. For the code object, I used `SavedView`
2. Makes sense. I don't think the sidebar reorg is set, so I'm just going to add one more page to side bar for now :) I modified the traces and evals page to just include a short info blurb and a link out to the feature page 
3. Sounds good

docs/docs/guides/core-types/datasets.md
J2-D2-3PO: Removed

docs/docs/guides/tracking/tracing.md
J2-D2-3PO: Changed in https://github.com/wandb/weave/pull/3779/commits/fc85e1a11d99408790e2c1d2ce4aa84d70397cf5

docs/docs/guides/tracking/tracing.md
J2-D2-3PO: Fair, removed in https://github.com/wandb/weave/pull/3779/commits/fc85e1a11d99408790e2c1d2ce4aa84d70397cf5

docs/docs/guides/tools/saved-views.md
```diff
@@ -0,0 +1,187 @@
...
+```python
+import weave
+client = weave.init(<my_project>)
+
+view = weave.SavedView()
+```
+
+### Visualize the `SavedView` as a grid
+
+Use `.to_grid()` to represent the saved view as a grdi. Specify the maximum number of rows to display with `limit`.
```
J2-D2-3PO, line 29:
```suggestion
Use `.to_grid()` to represent the saved view as a grid. Specify the maximum number of rows to display with `limit`.
```

docs/docs/guides/tools/saved-views.md
```diff
@@ -0,0 +1,187 @@
...
+
+1. Navigate to your **Traces** or **Evals** tab.
+2. Adjust any of the following variables in your table configuration:
+   - Filters
+   - Sort order
+   - Page size
+   - Column visibility
+   - Column pinning
+3. Save the view using one of two options:
+   - In the upper right hand corner, click **Save view**.
```
J2-D2-3PO, line 149:
```suggestion
   - In the upper right-hand corner, click **Save view**. 
```

docs/docs/guides/tools/saved-views.md
```diff
@@ -0,0 +1,187 @@
...
+1. Follow the steps described in [Load a saved view](#load-a-saved-view).
+2. In the upper lefthand corner of the **Traces** or **Evals** tab, click the view name.
+3. Enter a new name for the view.
+4. To save the new view name, press **Enter**.
+
+### Edit a saved view
+
+1. Follow the steps described in [Load a saved view](#load-a-saved-view).
+2. Adjust your table configuration.
+3. In the upper right hand corner, click **Save view**.
```
J2-D2-3PO, line 169:
```suggestion
3. In the upper right-hand corner, click **Save view**. 
```

J2-D2-3PO (commented):
self nitpicks

docs/docs/guides/evaluation/guardrails_and_monitors.md
```diff
@@ -67,7 +67,7 @@ result, call = generate_text.call(input)  # Now you can use the call object with
 :::tip Why Use `.call()`?
 The Call object is essential for associating the score with the call in the database. While you can directly call the scoring function, this would not be associated with the call, and therefore not searchable, filterable, or exportable for later analysis.
 
-For more details about Call objects, see our [Calls guide section on Call objects](../tracking/tracing.mdx#getting-a-handle-to-the-call-object-during-execution).
```
J2-D2-3PO, line 70: Not sure why this change is showing up. Had originally modified it but changed it back. Not related to this PR

docs/docs/guides/evaluation/scorers.md
```diff
@@ -315,7 +315,7 @@ While this guide shows you how to create custom scorers, Weave comes with a vari
 
 To apply scorers to your Weave ops, you'll need to use the `.call()` method which provides access to both the operation's result and its tracking information. This allows you to associate scorer results with specific calls in Weave's database.
 
-For more information on how to use the `.call()` method, see the [Calling Ops](../tracking/tracing.mdx#calling-ops#getting-a-handle-to-the-call-object-during-execution) guide.
```
J2-D2-3PO, line 318: Not sure why this change is showing up. Had originally modified it but changed it back. Not related to this PR

J2-D2-3PO:
> Could you also document creating a new view and deleting them with this UI please: Screenshot 2025-02-28 at 8 14 42 AM

Yep added in as an option to save view. Also documented the delete functionality

J2-D2-3PO:
@jamie-rasmussen sounds like not much will change in terms of user facing functionality? LMK if I need to update this

docs/docs/guides/tools/saved-views.md
```diff
@@ -0,0 +1,187 @@
...
+```python
+view.set_columns("id", "op_name")
+```
+
+### Add columns
+
+Use `.add_column()` to add one or more new columns to the view. Specify one or more columns to be added.
+
+```python
+view.add_column("started_at", "Created")
```
jamie-rasmussen, line 54: `add_column` actually only allows adding a single column. The first argument is the field specifier, the second is an optional label to display as the column header - the default is to use the field specifier for that as well.

There is a separate `add_columns` method if you want to bulk add multiple columns, it doesn't allow overriding the label.

docs/docs/guides/tools/saved-views.md
```diff
@@ -0,0 +1,187 @@
...
+### Filter by operation name
+
+In Weave, every trace or eval is associated with an operation name.
+Use `.filter_op()` to filter the `SavedView` to only include calls where that specific operation was executed. 
+
+```python
+view.filter_op("Evaluation.predict_and_score")
+```
+
+:::tip
```
jamie-rasmussen, line 74: I would remove this tip - the `add_filter` method won't work except for a single version of the op that you specify with a digest, while the `filter_op` version above handles all versions of the Op. Also, we don't tell people what the correct URI format is.

docs/docs/guides/tools/saved-views.md
```diff
@@ -0,0 +1,187 @@
...
+```python
+view.add_filter("output.model_latency", ">=", 5)
+```
+
+#### Filter operators
+
+| Operator | Description | Example |
+|----------|-------------|---------|
+| `"contains"` | Checks if a string contains a substring. | `view.add_filter("output.status", "contains", "error")` |
+| `"equals"` | Checks if a string is exactly equal to a given value. | `view.add_filter("user.name", "equals", "Alice")` |
```
jamie-rasmussen, line 91: `user.name` is not a valid field. I would suggest using `input.category` for strings or `output.score` for numeric things.

docs/docs/guides/tools/saved-views.md
```diff
@@ -0,0 +1,187 @@
...
+| `"="` | Checks if a number is equal to a value. | `view.add_filter("output.score", "=", 80)` |
+| `"≠", "!="` | Checks if a number is not equal to a value. | `view.add_filter("metrics.loss", "!=", 0.5)` |
+| `"<"` | Checks if a number is less than a value. | `view.add_filter("age", "<", 30)` |
+| `"≤", "<="` | Checks if a number is less than or equal to a value. | `view.add_filter("metric.value", "<=", 100)` |
+| `">"` | Checks if a number is greater than a value. | `view.add_filter("output.score", ">", 90)` |
+| `"≥", ">="` | Checks if a number is greater than or equal to a value. | `view.add_filter("output.model_latency", ">=", 5)` |
+| `"is"` | Checks if a boolean field is `True` or `False`. | `view.add_filter("is_active", "is", True)` |
+| `"after"` | Checks if a date is after a given timestamp. | `view.add_filter("started_at", "after", "2024-01-01")` |
+| `"before"` | Checks if a date is before a given timestamp. | `view.add_filter("ended_at", "before", "2024-12-31")` |
+| `"is empty"` | Checks if a field is empty (`None` or `""`). | `view.add_filter("comments", "is empty", None)` |
```
jamie-rasmussen, line 102: I believe your doc is correct that the None value is required in this case but I should probably make it optional.

docs/docs/guides/evaluation/guardrails_and_monitors.md
```diff
@@ -66,7 +66,7 @@ result, call = generate_text.call(input)  # Now you can use the call object with
...
+For more details about Call objects, see our [Calls guide section on Call objects](../tracking/tracing#getting-a-handle-to-the-call-object-during-execution).
```
J2-D2-3PO, line 69:
```suggestion
For more details about Call objects, see the [Calls guide section on Call objects](../tracking/tracing.mdx#getting-a-handle-to-the-call-object-during-execution).
```

docs/docs/guides/tools/saved-views.md
J2-D2-3PO, line 54:
```suggestion
# Add a column with the field specifier and label "Created"
view.add_column("Created")
# Optionally, you can add a second argument to specify a different label name for the new column. By default, the field specifier is use for the label.
```

J2-D2-3PO: Got it, will update this example so that it's just showing the one arg and will add a code comment about changing the label using the optional arg

docs/docs/guides/tools/saved-views.md
```diff
@@ -0,0 +1,187 @@
...
+In Weave, every trace or eval is associated with an operation name.
+Use `.filter_op()` to filter the `SavedView` to only include calls where that specific operation was executed. 
+
+```python
+view.filter_op("Evaluation.predict_and_score")
+```
+
+:::tip
+`.filter_op()` is a shortcut for filtering by operation name. It ensures the correct URI format is used automatically. Alternatively, you can achieve the same result with `.add_filter("op_name", "equals", <operation_name>)`.
+:::
```
J2-D2-3PO, line 76:
```suggestion
```

docs/docs/guides/tools/saved-views.md
J2-D2-3PO, line 91:
```suggestion
| `"equals"` | Checks if a string is exactly equal to a given value. | `view.add_filter("input.category", "equals", "Alice")` |
```

J2-D2-3PO: Got it, updated the example to `input.category` since its string equality

jamie-rasmussen (approved):

PR #3781, opened 2025-02-26
perf(weave): add pre-groupby conditions to heavy query filter in calls query 

gtarpenning:
## Description

Add additional filters to calls query when querying by heavy fields. Because we must account for unmerged calls, all conditions must take into account individual call parts. Example: instead of adding: `inputs_dump.a = 'bob'` we must write `inputs_dump.a = 'bob' OR inputs_dump is NULL` to handle the case that we are only looking at an end call part. 

Prod:
```sql
WITH filtered_calls AS (...)
SELECT
    calls_merged.id AS id,
    any(calls_merged.inputs_dump) AS inputs_dump
FROM calls_merged
WHERE calls_merged.project_id = 'project'
HAVING (
    JSON_VALUE(any(calls_merged.inputs_dump), 'a') = 'bob'
)
```

Branch:
```sql
WITH filtered_calls AS (...)
SELECT
    calls_merged.id AS id,
    any(calls_merged.inputs_dump) AS inputs_dump
FROM calls_merged
WHERE calls_merged.project_id = 'project'
AND ((JSON_VALUE(calls_merged.inputs_dump, 'a') = 'bob')
    OR calls_merged.inputs_dump IS NULL)
HAVING (
    JSON_VALUE(any(calls_merged.inputs_dump), 'a') = 'bob'
)
```

## Testing

- unit tests
- end to end tests

## Summary by CodeRabbit

- **New Features**
  - Enhanced the system’s filtering and data aggregation capabilities to boost query performance and accuracy during high-demand operations.

- **Tests**
  - Expanded automated test coverage with new tests for heavy condition aggregation and batch processing, ensuring consistent and reliable system behavior.

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -3353,26 +3353,6 @@ def test_calls_stream_heavy_condition_aggregation_parts(client):
     res = _make_query("inputs.param.value1", "hello")
     assert len(res) == 1
     assert res[0].inputs["param"]["value1"] == "hello"
-
-    if client_is_sqlite(client):
-        # Does the query return the output?
-        with pytest.raises(TypeError):
-            # There will be no output because clickhouse hasn't merged the inputs and
-            # output yet
-            assert res[0].output["d"] == 5
-
-    # insert some more calls to encourage clickhouse to merge
-
-    @weave.op
-    def test():
-        return 1
-
-    test()
-    test()
-    test()
-
-    res = _make_query("inputs.param.value1", "hello")
-    assert len(res) == 1
     assert res[0].output["d"] == 5
 
 
@@ -3427,22 +3407,9 @@ def test_call_stream_query_heavy_query_batch(client):
     res = client.server.calls_query_stream(
         tsi.CallsQueryReq.model_validate(output_query)
     )
-    if not client_is_sqlite(client):
-        # in clickhouse we don't know how many calls are merged,
-        # and the query filters out started_at is NULL, so this will
-        # likely fail to return all 10 calls.
-        try:
-            assert len(list(res)) == 10
-            for call in res:
-                assert call.attributes["a"] == 5
-        except AssertionError:
-            # This can happen if the call_parts are not merged in the query,
-            # which is likely when we are inserting so few rows
-            pass
-    else:
-        assert len(list(res)) == 10
-        for call in res:
-            assert call.attributes["a"] == 5
+    assert len(list(res)) == 10
+    for call in res:
+        assert call.attributes["a"] == 5
 
     # now query for inputs by string. This should be okay,
     # because we don't filter out started_at is NULL
@@ -3473,16 +3440,5 @@ def test_call_stream_query_heavy_query_batch(client):
     )
     assert len(list(res)) == 10
     for call in res:
-        assert call.inputs["param"]["value1"] == "hello"
-        assert call.output["d"] == 5
-
-    # By making these queries, clickhouse normally merges the call_parts
-    # into calls_merged, and we should be able to query the outputs
-    # and get the correct results.
-    res1 = client.server.calls_query_stream(
-        tsi.CallsQueryReq.model_validate(output_query)
-    )
-    assert len(list(res1)) == 10
-    for call in res1:
         assert call.inputs["param"]["value1"] == "hello"
         assert call.output["d"] == 5
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -3388,7 +3388,7 @@ def test_call_stream_query_heavy_query_batch(client):
             id=call_ids[i],
             ended_at=datetime.datetime.now(tz=datetime.timezone.utc),
             summary={"c": 5},
-            output={"d": 5, "e": "f"},
+            output={"d": 5, "e": "f", "result": {"message": "completed"}},
         )
         client.server.call_end(tsi.CallEndReq(end=end))
 
@@ -3417,9 +3417,20 @@ def test_call_stream_query_heavy_query_batch(client):
         "project_id": project_id,
         "query": {
             "$expr": {
-                "$eq": [
-                    {"$getField": "inputs.param.value1"},
-                    {"$literal": "hello"},
+                "$and": [
+                    {
+                        "$eq": [
+                            {"$getField": "inputs.param.value1"},
+                            {"$literal": "hello"},
+                        ]
+                    },
+                    {
+                        "$contains": {
+                            "input": {"$getField": "output.result.message"},
+                            "substr": {"$literal": "COMPleted"},
+                            "case_insensitive": True,
+                        }
+                    },
                 ]
             }
         },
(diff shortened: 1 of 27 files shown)


tests/trace/test_client_trace.py
```diff
@@ -3300,3 +3300,157 @@ def test():
...
+    }
+    res = client.server.calls_query_stream(
+        tsi.CallsQueryReq.model_validate(output_query)
+    )
+    assert len(list(res)) == 10
+    for call in res:
+        assert call.attributes["a"] == 5
+
+    # now query for inputs by string. This should be okay,
+    # because we don't filter out started_at is NULL
```
gtarpenning, line 3415: remove

weave/trace_server/calls_query_builder.py
```diff
@@ -73,6 +76,7 @@ def is_heavy(self) -> bool:
 
 class CallsMergedAggField(CallsMergedField):
     agg_fn: str
+    use_agg_fn: bool = True
```
gtarpenning, line 79: here @bcsherma

weave/trace_server/calls_query_builder.py
```diff
@@ -65,6 +65,9 @@ def as_sql(
     def as_select_sql(self, pb: ParamBuilder, table_alias: str) -> str:
         return f"{self.as_sql(pb, table_alias)} AS {self.field}"
 
+    def is_heavy(self) -> bool:
```
tssweeney, line 68: seems like this is a property of `CallsMergedField`. not sure why it needed to be hoisted up to QueryBuilderField

weave/trace_server/calls_query_builder.py
```diff
@@ -781,7 +809,12 @@ def process_operand(operand: "tsi_query.Operand") -> str:
             )
         elif isinstance(operand, tsi_query.GetFieldOperator):
             structured_field = get_field_by_name(operand.get_field_)
-            field = structured_field.as_sql(param_builder, table_alias)
+            if not use_agg_fn and isinstance(structured_field, CallsMergedAggField):
+                structured_field.use_agg_fn = False
+                field = structured_field.as_sql(param_builder, table_alias)
```
tssweeney, line 814: i would discourage this pattern. Prefer to have `use_agg_fn` a parameter of as_sql, rather than flipping a property

tssweeney: Hey! Coderrabbit had a good comment below! Exactly correct

weave/trace_server/calls_query_builder.py
```diff
@@ -893,6 +926,118 @@ def process_calls_filter_to_conditions(
...
+    - attributes_dump
+    """
+    if not conditions:
+        return False
+
+    # Check if we have any OR operations between start and end fields
+    for condition in conditions:
+        if not hasattr(condition, "operand"):
+            continue
+        if not isinstance(condition.operand, tsi_query.OrOperation):
```
tssweeney, line 955: I actually think you can support OR using the design (you would just OR the parts instead of ANDing) them. Might want to add a comment about this since it is more of a feature addition rather than a logical impossibility.

weave/trace_server/calls_query_builder.py
```diff
@@ -781,7 +805,12 @@ def process_operand(operand: "tsi_query.Operand") -> str:
             )
         elif isinstance(operand, tsi_query.GetFieldOperator):
             structured_field = get_field_by_name(operand.get_field_)
-            field = structured_field.as_sql(param_builder, table_alias)
+            if not use_agg_fn and isinstance(structured_field, CallsMergedAggField):
+                structured_field.use_agg_fn = False
+                field = structured_field.as_sql(param_builder, table_alias)
+                structured_field.use_agg_fn = True  # reset to default
+            else:
+                field = structured_field.as_sql(param_builder, table_alias)
```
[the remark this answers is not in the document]
gtarpenning: nice

weave/trace_server/calls_query_builder.py
```diff
@@ -592,17 +605,27 @@ def _as_sql_base_format(
...
         if self.hardcoded_filter is not None:
             having_conditions_sql.append(self.hardcoded_filter.as_sql(pb, table_alias))
 
+        call_parts_predicate_filters_sql = ""
         if len(having_conditions_sql) > 0:
             having_filter_sql = "HAVING " + combine_conditions(
                 having_conditions_sql, "AND"
             )
 
+        if should_add_predicate_filters(self.query_conditions):
```
tssweeney, line 617: I would slightly adjust this to be more pluggable.

```
call_parts_predicate_filters_sql = build_call_parts_predicate_filter(self.query_conditions)
``` 

I think that the implementation for "should_add_predicate_filers" is coupled with the implementation of "make_call_parts_predicate_filters_sql" and therefore should be abstracted behind a common method. Looking at your algorithm, i can imagine some alternatives which would make this easier to modify

weave/trace_server/calls_query_builder.py
tssweeney: you resolved this conversation without any comment or code change?

weave/trace_server/calls_query_builder.py
```diff
@@ -592,17 +606,26 @@ def _as_sql_base_format(
...
         if self.hardcoded_filter is not None:
             having_conditions_sql.append(self.hardcoded_filter.as_sql(pb, table_alias))
 
+        call_parts_predicate_filters_sql = ""
         if len(having_conditions_sql) > 0:
             having_filter_sql = "HAVING " + combine_conditions(
                 having_conditions_sql, "AND"
             )
 
+        heavy_non_feedback_conditions = [
```
tssweeney, line 620: why not just have this inside the `make_call_parts_predicate_filters_sql`?

weave/trace_server/calls_query_builder.py
```diff
@@ -65,9 +65,6 @@ def as_sql(
     def as_select_sql(self, pb: ParamBuilder, table_alias: str) -> str:
         return f"{self.as_sql(pb, table_alias)} AS {self.field}"
 
-    def is_heavy(self) -> bool:
-        return False
-
```
gtarpenning: lol @tssweeney

weave/trace_server/calls_query_builder.py
```diff
@@ -893,6 +926,80 @@ def process_calls_filter_to_conditions(
     return conditions
 
 
+def make_call_parts_predicate_filters_sql(
+    pb: ParamBuilder, table_alias: str, heavy_conditions: list[Condition]
+) -> str:
```
tssweeney, line 923: I would add the comment here that the condition returned MUST account for unmatched parts and NEVER filter out more than the heavy filter would

weave/trace_server/calls_query_builder.py
```diff
@@ -893,6 +926,80 @@ def process_calls_filter_to_conditions(
...
+    This function creates SQL conditions that can be applied before the GROUP BY
+    to filter out rows that definitely won't match the heavy conditions.
+
+    For fields that may only exist in start or end parts, we add special handling
+    to avoid filtering out rows where the field is NULL (as they might be part of
+    a valid call when combined with other parts).
+
+    Returns an empty string if:
+    1. There are no heavy conditions
+    2. There are OR operations between start-only and end-only fields, which would
```
tssweeney, line 935: I actually think #2 is safe, just not using this algorithm

PR #3790, opened 2025-02-26
fix(weave): use of deprecated /calls/query endpoint

jamie-rasmussen:
## Description

1. Change server.calls_query to be implemented in terms of `calls_query_stream` so that it won't use the deprecated endpoint.
2. Change the higher level `client.get_call` and `client.get_calls` methods to use `calls_query_stream`

## Testing

How was this PR tested?

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the retrieval of call data using a direct streaming approach, reducing overhead and simplifying processing for enhanced performance.
  
- **New Features**
  - Updated the server query interface to support more flexible input types.
  - Deprecated the legacy query method in favor of the improved streaming mechanism.

These updates deliver a smoother, more efficient data fetching experience for end users.

### DIFF
--- a/weave/trace_server_bindings/remote_http_trace_server.py
+++ b/weave/trace_server_bindings/remote_http_trace_server.py
@@ -290,11 +290,12 @@ class RemoteHTTPTraceServer(tsi.TraceServerInterface):
     def calls_query(
         self, req: Union[tsi.CallsQueryReq, dict[str, Any]]
     ) -> tsi.CallsQueryRes:
-        return self._generic_request(
-            "/calls/query", req, tsi.CallsQueryReq, tsi.CallsQueryRes
-        )
+        # This previously called the deprecated /calls/query endpoint.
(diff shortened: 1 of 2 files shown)


weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -290,11 +290,12 @@ def call_read(self, req: Union[tsi.CallReadReq, dict[str, Any]]) -> tsi.CallRead
     def calls_query(
         self, req: Union[tsi.CallsQueryReq, dict[str, Any]]
     ) -> tsi.CallsQueryRes:
-        return self._generic_request(
-            "/calls/query", req, tsi.CallsQueryReq, tsi.CallsQueryRes
-        )
+        # This previously called the deprecated /calls/query endpoint.
```
jamie-rasmussen, line 293: I tried using the `@deprecated` decorator on this but it seems the default configuration of `warnings.warn` will not show it anyway if you're not calling it from main.

tssweeney (approved):

PR #3804, opened 2025-02-27
chore(weave): Make object viewer load much faster

tssweeney:
There is an effect which is intended to reset the expansion state when new data is loaded into the object viewer. This can result in a horrible render loop (for some examples, it can take 30s to resolve). I did not get to the bottom of exactly why, but i think it is because the MUI data grid is loading in fresh data while simultaneously we are expanding/collapsing nodes, and it seems to have more-than-linear loading time depending on the size of the old and new state. Removing this moves the load down to sub second in nearly all cases i found

## Summary by CodeRabbit

- **Refactor**
	- Improved component responsiveness and efficiency through optimized state management and event handling.
	- Enhanced performance by reducing unnecessary recalculations and streamlining the control flow. 
	- Removed unnecessary complexity related to data expansion logic, leading to a cleaner implementation.


jamie-rasmussen (approved):
I think we have to do this but I would definitely anticipate getting complaints about having to manually expand.

PR #3808, opened 2025-02-27
chore(ui): adds small and range datetime filter bar items

gtarpenning:
## Description

[WB-23782](https://wandb.atlassian.net/browse/WB-23782)

Adds a small version of the timestamp pil to look more like datadog.  Also adds a _range_ component to render in the filter bar when there is a before and an after. This just looks much nicer when we have two. We only use the small version when we have one after filter (which will become the default and therefore the vast majority of the time). 

this pr: 
- adds a small version of the `(date): after` operator that looks like `[cal] past <x>d` when displayed in the filter bar display section. 
- adds a range filter widget that is rendered when there is both an `after` and a `before` date operation on called and they are valid. This combines the two dates into one visual filter. Filter is deletable, which just deletes the before timestamp. Both timestamps have unique labels on hover. Can be edited from the filter edit window, as seperate rows. If two dates are invalid as a range, just renders them separately. 

## Testing
Prod:
Screenshot 2025-02-27 at 12 46 20 PM

Screenshot 2025-02-27 at 12 46 42 PM

Screenshot 2025-03-19 at 9 25 20 AM

Branch:
Screenshot 2025-02-27 at 12 45 45 PM
Screenshot 2025-02-27 at 12 45 34 PM

Range with specific time
Screenshot 2025-03-19 at 9 25 07 AM

Range with default time
Screenshot 2025-03-19 at 9 24 56 AM

datetime-combo

Error handling for invalid dates: 
datetime-error

## Summary by CodeRabbit

- **New Features**
	- Enhanced filter tag behavior for date values, including refined interaction controls.
	- Introduced concise micro-level timestamp formatting (e.g., "1h", "1d", "1w") with tooltip support for improved time displays.
	- Improved identification of date-specific filter criteria to ensure a smoother user experience.
	- Added functionality to combine date range filters for better filter management.
	- Implemented a new method for generating unique filter IDs to prevent conflicts.

[WB-23782]: https://wandb.atlassian.net/browse/WB-23782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave-js/src/components/Timestamp.tsx
```diff
@@ -22,6 +23,49 @@ type TimestampProps = {
...
   live?: boolean;
+
+  // If true, will omit the time when it's midnight (00:00)
+  dropTimeWhenDefault?: boolean;
+};
+
+// Format a time difference to a micro string (1h, 1d, 1w, etc.)
+const formatMicroTime = (then: moment.Moment): string => {
+  const now = moment();
+  const diffMs = now.diff(then);
```
bcsherma, line 34: I got confused here by the use of micro and "ms" - I think you are using micro to describe the small string format (eg 16d) but I was also thinking microsecond resolution time string, which is what our started and ended timestamps for calls use. Maybe we should call the short form strings something else

weave-js/src/components/Timestamp.tsx
```diff
@@ -22,6 +23,49 @@ type TimestampProps = {
...
+  }
+
+  // Otherwise use days for more precision
+  if (days >= 1) {
+    return days === 1 ? '1d' : `${days}d`;
+  } else if (hours >= 1) {
+    return hours === 1 ? '1h' : `${hours}h`;
+  } else {
+    return '<1h';
+  }
```
bcsherma, line 68: Lgtm

bcsherma (approved):
Nice improvement. We really gotta restyle the datetime thing in general though it looks pretty raw.

PR #3811, opened 2025-02-27
fix(app): various runs history step slider bug fixes

nicholaspun-wandb:
## Description

 - Account for matching keys across history. The incoming type will be a typedDict, instead of a union
 - Check for nullish step, instead of falsey, since steps can start at 0

## Testing

Local FE

Table keys are obtained properly and the table at step 0 is fetched:

https://github.com/user-attachments/assets/085fd064-fa49-461c-9a3a-61df96f9aedd

Previous behaviour maintained:

https://github.com/user-attachments/assets/61e6bfb2-dd5c-4818-9c49-3e10cbc97306

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved how tables are displayed in the run history view by ensuring consistent retrieval of table information.
  - Adjusted the stepper’s starting index to apply default selections only when valid, enhancing the overall navigation experience.


domphan-wandb (approved):
LGTM. Just curious about the demo videos. Is the slider jittery because there are less steps in the first recording than there are in the second?

nicholaspun-wandb:
I think the slider's trying to snap to closest valid value

PR #3812, opened 2025-02-27
chore(weave): add a default datetime filter on trace table

gtarpenning:
## Description

This pr implements a datetime filter by default on the traces table. Key features: 
- default date range adjusts to the most recent traces in the project, either 7 days, 30 days, or 180 days.
- this date range is then saved to the browser localStorage, with a 1 day TTL
- adds a browser storage util 
- does not apply to the evaluations page 
- continues to apply for all operations/filters/sorting, unless you explicitly delete the filter in which case it should not return until you refresh the page. 
- adds link to expand date range when no results for current one, which is dynamically set to buckets corresponding to the most recently created op version
     - this should be a hard path to hit
- refactors empty page into own component

## Testing
project-datetime-stash-4

https://www.loom.com/share/e43d44c91de543bcb31ace11cc94d8a7?sid=ceea6006-2ecd-4e2e-8734-3dfad7713c2a

## Summary by CodeRabbit

- **New Features**
  - Enhanced call records filtering with a default date filter that applies a 30-day range.
  - Introduced a new component to display an overlay when no calls are found, offering actionable suggestions based on filter settings.
  - Improved filter state tracking to better reflect when filters are applied or removed.
- **Bug Fixes**
  - Fixed issues with filter visibility and state management for date filters.
- **Style**
  - Refined notification messaging to clearly indicate when no calls are found within the specified date range.


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -1020,7 +1031,7 @@ export const CallsTable: FC<{
                   alignItems: 'center',
                 }}>
                 <Typography color="textSecondary">
-                  No calls found.{' '}
+                  No calls found for the specified date range.{' '}
```
gtarpenning, line 1034: condition 'date range' on if there is a date filter

tssweeney (approved):
LGTM - no state change, completely undoable (:

jamie-rasmussen (approved):
It feels a little weird to have "Past 30 days" as a filter but in the relative time picker dropdown "Past 1 mo", might be something we could align.

PR #3828, opened 2025-03-03
chore(weave): add rich progress bar to client flush

gtarpenning:
## Description

Adds a tqdm progress bar to convey the flushing progress. It accounts for all jobs that have been enqueued to either of the two futureExecutors in the client (main, and file-upload/fastlane). Also adds a `num_outstanding_jobs` attr that returns the number of futures left in queue. 

Because jobs can enqueue more jobs, the progress bar isn't linear, but still provides a useful visual aid. Additionally, because trace prints wait on call upserts, if many calls haven't been synced by the time we call `flush` the progress bar will be broken up by the donut links. Most calls will have been completed though, with the majority of flush time uploading files (for most heavy dataset evaluation).

## Testing
Screenshot 2025-03-03 at 2 06 16 PM

test-progress-bar-5

## Summary by CodeRabbit

- **New Features**
  - Enhanced background task processing with a visible progress indicator during flush operations.
  - Introduced real-time status updates via a callback mechanism to keep users informed.
  - Improved tracking of pending jobs to ensure smoother, more reliable task management.
  - Added new test functions to validate progress bar and callback functionalities.
  - Added properties to monitor outstanding jobs in both the WeaveClient and AsyncBatchProcessor.

- **Bug Fixes**
  - Ensured thread safety when accessing the number of outstanding jobs in processing classes.

### DIFF
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -3,11 +3,11 @@ import dataclasses
 import json
 import platform
 import sys
+import time
 
 import pydantic
 import pytest
 import requests
-from PIL import Image
 
 import weave
 import weave.trace_server.trace_server_interface as tsi
@@ -1890,25 +1890,11 @@ def test_global_attributes_with_call_attributes(client_creator):
 
 
 def test_flush_progress_bar(client):
-    # expensive op
     @weave.op
-    def make_image():
-        img = Image.new("RGB", (300, 300))
-        pixels = img.load()
-        for x in range(300):
-            for y in range(300):
-                pixels[x, y] = (x, y, 0)
-        return img
+    def op_1():
+        time.sleep(1)
 
-    ds2 = [make_image() for i in range(10)]
-
-    # really expensive op
-    @weave.op
-    def make_dataset(ds2):
-        new_ds = [make_image() for i in range(10)]
-        return ds2 + new_ds
-
-    make_dataset(ds2)
+    op_1()
 
     # flush with progress bar
     client.flush(use_progress_bar=True)
@@ -1918,29 +1904,27 @@ def test_flush_progress_bar(client):
     assert client._has_pending_jobs() == False
 
 
-def test_flush_status_callback(client, capsys):
+def test_flush_callback(client):
     @weave.op
-    def make_image():
-        img = Image.new("RGB", (300, 300))
-        pixels = img.load()
-        for x in range(300):
-            for y in range(300):
-                pixels[x, y] = (x, y, 0)
-        return img
+    def op_1():
+        time.sleep(1)
 
-    ds = [make_image() for i in range(10)]
+    op_1()
 
-    @weave.op
-    def use_dataset(ds):
-        return ds
+    def fake_logger(status):
+        assert "main_jobs" in status
 
-    use_dataset(ds)
+    # flush with callback
+    client.flush(callback=fake_logger)
 
-    def logger(status):
-        assert status["main_jobs"] == 0
+    # make sure there are no pending jobs
+    assert client._get_pending_jobs() == (0, 0, 0, 0)
+    assert client._has_pending_jobs() == False
 
-    # use custom logger callback
-    client.flush(callback=logger)
+    op_1()
+
+    # this should also work, the callback will override the progress bar
+    client.flush(callback=fake_logger, use_progress_bar=True)
 
     # make sure there are no pending jobs
     assert client._get_pending_jobs() == (0, 0, 0, 0)
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -1900,7 +1900,7 @@ def test_flush_progress_bar(client):
     client.flush(use_progress_bar=True)
 
     # make sure there are no pending jobs
-    assert client._get_pending_jobs() == (0, 0, 0, 0)
+    assert client._get_pending_jobs()["total_jobs"] == 0
     assert client._has_pending_jobs() == False
 
 
@@ -1912,13 +1912,13 @@ def test_flush_callback(client):
     op_1()
 
     def fake_logger(status):
-        assert "main_jobs" in status
+        assert "job_counts" in status
 
     # flush with callback
     client.flush(callback=fake_logger)
 
     # make sure there are no pending jobs
-    assert client._get_pending_jobs() == (0, 0, 0, 0)
+    assert client._get_pending_jobs()["total_jobs"] == 0
     assert client._has_pending_jobs() == False
 
     op_1()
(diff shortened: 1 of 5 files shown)


weave/trace/weave_client.py
```diff
@@ -1972,15 +1973,27 @@ def _ref_uri(self, name: str, version: str, path: str) -> str:
...
+        improve performance when user code completes before data has been
+        uploaded to the server.
         """
+        # Use the progress bar utility
+        flush_with_progress_bar(
+            main_executor=self.future_executor,
+            fastlane_executor=self.future_executor_fastlane,
+        )
+
+        # Make sure all jobs are processed, including `call_processor` jobs which
```
tssweeney, line 1990: seems like we should certainly have this included?

weave/trace/weave_client.py
```diff
@@ -1972,15 +1973,27 @@ def _ref_uri(self, name: str, version: str, path: str) -> str:
...
+        Flushes all background tasks to ensure they are processed.
+
+        This method blocks until all currently enqueued jobs are processed,
+        displaying a progress bar to show the status of the pending tasks.
+        It ensures parallel processing during main thread execution and can
+        improve performance when user code completes before data has been
+        uploaded to the server.
         """
+        # Use the progress bar utility
+        flush_with_progress_bar(
```
tssweeney, line 1985: Good first pass, but i would recommend refactoring this slightly. Your client_progress_bar is tightly coupled with the weave client implementation, which in itself is coupled with the lower level data structures. I would recommend a slightly different approach:
1. Define a typed dict / structured object `WeaveFlushStatus`
2. Make `_flush` accept an optional callback which in itself accepts a `WeaveFlushStatus`. Probably also specify a refresh interval
3. Move accounting logic from your `client_progress_bar` into the weave_client class itself, which is responsible for constructing the data payload
4. thin out your `client_progress_bar` such that it is purely responsible for interpreting the payload and updating the print. 

This will help isolate the abstractions. Then finally, `flush` can accept a parameter that specifies which logger to use. In theory, this progress bar won't render well in write-only logs like CI, so i would imagine a simple print of the progress state would be another logger-type.

weave/trace/weave_client.py
```diff
@@ -1970,34 +1975,190 @@ def _op_runs(self, op_def: Op) -> Sequence[Call]:
...
-            # We don't want to do an instance check here because it could
-            # be susceptible to shutdown race conditions. So we save a boolean
-            # _server_is_flushable and only call this if we know the server is
-            # flushable. The # type: ignore is safe because we check the type
-            # first.
-            self.server.call_processor.wait_until_all_processed()  # type: ignore
+            self.server.call_processor.stop_accepting_new_work_and_flush_queue()  # type: ignore
 
-    def _send_file_create(self, req: FileCreateReq) -> Future[FileCreateRes]:
+    def _get_pending_jobs(self) -> tuple[int, int, int, int]:
```
andrewtruong, line 2133: nit: make this a typeddict

tests/trace/test_weave_client.py
```diff
@@ -1886,3 +1887,45 @@ def my_op(a: int) -> int:
 
         # Local attributes override global ones
         assert call.attributes["env"] == "override"
+
+
+def test_flush_progress_bar(client):
```
andrewtruong, line 1892: this is not testing the pbar.

If you really want to test the pbar, you can capture redirect IO into a buf and confirm it looks the way you want

weave/trace/client_progress_bar.py
```diff
@@ -0,0 +1,122 @@
...
+    has_pending_jobs: bool
+
+
+def create_progress_bar_callback() -> Callable[[WeaveFlushStatus], None]:
+    """Create a callback function that displays a progress bar for flush status.
+
+    Returns:
+        A callback function that can be passed to WeaveClient._flush.
+    """
+    console = Console()
```
andrewtruong, line 46: Did you check that it doesn't overwrite user logs?  I'm not sure why but rich sometimes paints over my terminal which can be annoying

andrewtruong (approved):
lgtm, but please double check my testing comments

weave/trace/client_progress_bar.py
```diff
@@ -0,0 +1,122 @@
...
+    BarColumn,
+    Progress,
+    SpinnerColumn,
+    TaskProgressColumn,
+    TextColumn,
+    TimeElapsedColumn,
+)
+
+
+class WeaveFlushStatus(TypedDict):
```
tssweeney, line 20: nit: i would move this type def to weave_client

weave/trace/weave_client.py
```diff
@@ -1970,34 +1975,190 @@ def _op_runs(self, op_def: Op) -> Sequence[Call]:
...
         if self.future_executor_fastlane:
-            # If we have a separate upload worker pool, use it
-            return self.future_executor_fastlane.defer(self.server.file_create, req)
-        return self.future_executor.defer(self.server.file_create, req)
+            fastlane_jobs = self.future_executor_fastlane.num_outstanding_futures
+        call_processor_jobs = 0
+        if self._server_is_flushable:
+            call_processor_jobs = self.server.call_processor.num_outstanding_jobs  # type: ignore
+
+        total_jobs = main_jobs + fastlane_jobs + call_processor_jobs
```
tssweeney, line 2151: nit: "total jobs" seems a little necessary

weave/trace/weave_client.py
```diff
@@ -1970,34 +1975,190 @@ def _op_runs(self, op_def: Op) -> Sequence[Call]:
...
+        if use_progress_bar and callback is None:
+            callback = create_progress_bar_callback()
+        if callback is None:
+            self._flush()
+        else:
+            self._flush_with_callback(callback=callback)
+
+    def _flush_with_callback(
+        self,
+        callback: Callable[[WeaveFlushStatus], None] | None = None,
```
tssweeney, line 2034: callback should never be non here - would recommend removing that and simplifying the code a bit

gtarpenning: yes true

weave/trace/client_progress_bar.py
gtarpenning: My inclination is to keep the client as clean as possible, but I agree it does sort of want to be there.

weave/trace/client_progress_bar.py
gtarpenning: I think its fine? How exactly can I test that? 
Screenshot 2025-03-03 at 7 03 43 PM

gtarpenning: The trace donut link gets printed while the bar is going, I think its fine.

weave/trace/weave_client.py
gtarpenning: i'm going to keep it in the typedDict, removing this local var though

tests/trace/test_weave_client.py
gtarpenning: hmmm yeah, the issue is actually more generating a test that actually requires flushing. not sure there is a trivial way of doing it... scratching my head over here

andrewtruong: There are a few ways, but maybe the easiest is to use a mock server that is slow to respond / only accepts 1 connection.  Then you can queue up some jobs and see them go through 1-by-1

PR #3832, opened 2025-03-04
chore(weave): Improve DF Serialization

tssweeney:
Do not merge - the UI is not yet satisfactory for raw tables

Screenshot 2025-03-03 at 16 49 55


andrewtruong (commented):
Not clear to me it's always a good idea to dump to a weave table vs. just serialize to parquet -- have you considered the latter?

PR #3835, opened 2025-03-04
chore(weave): Add reference trace server and generator

andrewtruong:
- **test**
- **test**


andrewtruong:
@tssweeney I started down this path but don't love it.  Each server impl will need to customize different parts of the server anyways and it doesn't seem like we're getting much out of this

weave/trace_server/reference/generate.py
```diff
@@ -0,0 +1,374 @@
...
+
+        def _get_server(
+            auth_params: AuthParams = Depends(self.auth_dependency),
+        ) -> tsi.TraceServerInterface:
+            return self.server_factory(auth_params)
+
+        return _get_server
+
+
+def generate_routes(
```
andrewtruong, line 79: The routes are basically all delegating to the server, which is dependency injected per request.

Defining all the routes here ensures that the relevant parts of the OpenAPI schema we generate for any implementation will be the same.  Parts that may differ include properties of the app itself, security, etc.

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -188,6 +188,9 @@ def calls_delete(self, req: tsi.CallsDeleteReq) -> tsi.CallsDeleteRes:
             req.wb_user_id = self._idc.ext_to_int_user_id(req.wb_user_id)
         return self._ref_apply(self._internal_trace_server.calls_delete, req)
 
+    def call_start_batch(self, req: tsi.CallCreateBatchReq) -> tsi.CallCreateBatchRes:
+        return self._ref_apply(self._internal_trace_server.call_start_batch, req)
```
andrewtruong, line 192: is this correct?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -250,6 +250,10 @@ def call_start(self, req: tsi.CallStartReq) -> tsi.CallStartRes:
             trace_id=ch_call.trace_id,
         )
 
+    # TODO: This is not actually implemented at the lower level yet.
+    def call_start_batch(self, req: tsi.CallCreateBatchReq) -> tsi.CallCreateBatchRes:
+        return tsi.CallCreateBatchRes(res=[])
```
andrewtruong, line 255: is this safe?

weave/trace_server/reference/generate.py
```diff
@@ -0,0 +1,382 @@
...
+
+
+def noop_trace_server_factory(auth: AuthParams) -> tsi.TraceServerInterface:
+    # This type-ignore is safe, it's just used to instantiate a stub implementation
+    # without having to redefine all of the methods (which would be pointless because
+    # this is a stub that does nothing).
+    return NoopTraceServer()  # type: ignore
+
+
+class ServerDependency:
```
andrewtruong, line 48: This pattern is used to work with FastAPI's dependency injection system.

weave/trace_server/reference/generate.py
```diff
@@ -0,0 +1,382 @@
...
+        return hash(
+            (
+                tuple(self.headers.items()) if self.headers else None,
+                tuple(self.cookies.items()) if self.cookies else None,
+                self.auth,
+            )
+        )
+
+
+class NoopTraceServer(tsi.TraceServerInterface): ...
```
andrewtruong, line 38: This definition is required because you can't instantiate a protocol

weave/trace_server/trace_server_interface.py
```diff
@@ -915,6 +933,7 @@ def calls_query_stream(self, req: CallsQueryReq) -> Iterator[CallSchema]: ...
     def calls_delete(self, req: CallsDeleteReq) -> CallsDeleteRes: ...
     def calls_query_stats(self, req: CallsQueryStatsReq) -> CallsQueryStatsRes: ...
     def call_update(self, req: CallUpdateReq) -> CallUpdateRes: ...
+    def call_start_batch(self, req: CallCreateBatchReq) -> CallCreateBatchRes: ...
```
tssweeney, line 936: The various "no-op" `call_start_batches` gives the impression that this is not correct. Why are they no-ops? Seems like an unnecessary coupling of fastapi logic and lower level interface logic. I think either:
a) there should not be a `call_start_batch` here; or
b) they implementations should be real.

Also, probably should call this `call_batch` since it is not limited to start events

weave/trace_server/sqlite_trace_server.py
```diff
@@ -206,6 +206,10 @@ def call_start(self, req: tsi.CallStartReq) -> tsi.CallStartRes:
             trace_id=req.start.trace_id,
         )
 
+    # TODO: The Sqlite Trace Server does not support batching.
+    def call_start_batch(self, req: tsi.CallCreateBatchReq) -> tsi.CallCreateBatchRes:
+        return tsi.CallCreateBatchRes(res=[])
```
tssweeney, line 211: it is easy: just loop through

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -256,6 +256,11 @@ def call_start(
             "/call/start", req, tsi.CallStartReq, tsi.CallStartRes
         )
 
+    # TODO: Technically this is a no-op because of how the trace server implements
+    # the batching.
+    def call_start_batch(self, req: tsi.CallCreateBatchReq) -> tsi.CallCreateBatchRes:
+        return tsi.CallCreateBatchRes(res=[])
```
tssweeney, line 262: why not hit the endpoint here?

weave/trace_server/reference/server.py
```diff
@@ -0,0 +1,17 @@
...
+from weave.trace_server.reference.generate import (
+    ServerDependency,
+    generate_routes,
+    noop_trace_server_factory,
+)
+
+server_dependency = ServerDependency(server_factory=noop_trace_server_factory)
+trace_service_router = generate_routes(APIRouter(), server_dependency)
+app = FastAPI()
+app.include_router(trace_service_router)
```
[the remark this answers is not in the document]
tssweeney: I don't think so. it is totally ok for the fastapi server to be a superset of tsi. tsi.

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: why not actually implement this?

weave/trace_server/trace_server_interface.py
tssweeney: Alternatively: I think you could benefit from defining a `TraceServerAPI(tsi.TraceServerInterface)` that has additional functions (for example `call_start_batch`) if you did not want to couple the tsi interface with the service api - allowing the service api to be a true superset

tssweeney (commented):
Before approving this, a few things:
1. I think we should modify the pattern to either implement the call_start_batch methods OR define a subclass interface that allows the server definition to define `call_start_batch` but not force the engines to implement it.
2. I would like to see if this could be used inside the trace server itself - how does this change the implementation in core? We should have a PR that picks this up in core to ensure correctness

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: We could!  But this slice of the stack was already getting big and I didn't want to implement batching for all of the servers in this same PR.

weave/trace_server/sqlite_trace_server.py
andrewtruong: If ok with you, I prefer to implement in a follow up PR for all of the relevant servers at once.

weave/trace_server/reference/server.py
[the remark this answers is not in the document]
andrewtruong: That's true, but for the purposes of codegen with our specific repo structure, I think we'll need to do what I described above.  If not, we'll run into the circular reference problem discussed earlier.

To summarize:
1. Codegen must be done from `weave` and can't reference `core`
2. This implies we need a stub server implemented in `weave`, which is defined in this PR.
3. However, we also want the codegen generate all routes for the server defined in `core`.
4. Therefore, the stub and `core` servers must have the same interface!

weave/trace_server/trace_server_interface.py
andrewtruong: We can implement them for all the servers, I just decided it was a bit too much for one PR.

The name reflects what it's currently called in `core`, though we can change both.

Why would you want a `TraceServerAPI`?  It seems like going straight to `TraceServerInterface` is preferred

weave/trace_server_bindings/remote_http_trace_server.py
andrewtruong: see comments above

weave/trace_server/reference/server.py
```diff
@@ -0,0 +1,17 @@
...
+
+server_dependency = ServiceDependency(service_factory=noop_trace_server_factory)
+trace_service_router = generate_routes(APIRouter(), server_dependency)
+app = FastAPI()
+app.include_router(trace_service_router)
+
+if __name__ == "__main__":
+    import uvicorn
+
+    uvicorn.run(app, host="0.0.0.0", port=8000)
```
tssweeney, line 17: not blocking, just curious: do you actually need to start the server here?

weave/trace_server/trace_server_interface.py
```diff
@@ -951,3 +978,14 @@ def actions_execute_batch(
 
     # Execute LLM API
     def completions_create(self, req: CompletionsCreateReq) -> CompletionsCreateRes: ...
+
+
+class ServerInfoRes(BaseModel):
+    min_required_weave_python_version: str
+
+
+class TraceService(Protocol):
```
tssweeney, line 987: I would add a comment explaining what this is. Also, I could see the justification of putting this into it's own file to properly seperate the layers.

tssweeney (approved):

weave/trace_server/trace_server_interface.py
andrewtruong: done

weave/trace_server/reference/server.py
andrewtruong: Nope, but it doesn't hurt and is added for convenience in case you want to run the `server.py` file directly

PR #3840, opened 2025-03-04
fix(weave): fix repeated flushes and add `client.finish`

gtarpenning:
## Description

[WB-23701](https://wandb.atlassian.net/browse/WB-23701)

`flush` currently deactivates the async processor that handles uploading calls, we want to make sure this is really only called at the end of the user process. The only benefit to calling it is to prevent exiting, it should not be called repeatedly or before final ops are called, so rename that `finish`. 

Also fix the `client.flush` method so that it is robust to repeated calls. 

this pr: 
- fix `flush`
- adds `finish`
- changes tests to use `finish`
- add test for repeated flushes
- better typing

depends on https://github.com/wandb/weave/pull/3828

## Testing:

adds test 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method to check if the system is currently accepting new tasks.
  
- **Refactor**
  - Updated the process for completing operations to enhance reliability.
  - Adjusted test procedures to utilize the new completion method, improving consistency and simulation of real-world scenarios.

[WB-23701]: https://wandb.atlassian.net/browse/WB-23701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace/weave_client.py
```diff
@@ -2117,7 +2131,13 @@ def _flush(self) -> None:
         if self.future_executor_fastlane:
             self.future_executor_fastlane.flush()
         if self._server_is_flushable:
-            self.server.call_processor.stop_accepting_new_work_and_flush_queue()  # type: ignore
+            # We don't want to do an instance check here because it could
+            # be susceptible to shutdown race conditions. So we save a boolean
```
andrewtruong, line 2135: When would that happen?

gtarpenning: thats a tim comment that I removed in the previous pr. not exactly sure.

andrewtruong (commented):
Why not have both?  You might want to flush and keep going, and you always want to flush before you finish

weave/trace/weave_client.py
tssweeney: literally the symbol itself could be gone by the time it is needed here. it was erroring on me before

weave/trace/weave_client.py
```diff
@@ -2018,6 +2019,14 @@ def flush(
             callback: Optional callback function that receives status updates.
                       Overrides use_progress_bar.
         """
+        if self._server_is_flushable:
+            server = cast(RemoteHTTPTraceServer, self.server)
+            if not server.call_processor.is_accepting_new_work():
+                raise ValueError(
+                    "client.finish() called while async processor already stopping. client.finish() "
```
tssweeney: `async processor` is an implementation detail. I would instead logger.warn("`finish` has already been called") and return early

gtarpenning: i removed this, as we will now always start up another thread after the client forces a flush

gtarpenning: so its okay if we call finish multiple times.

gtarpenning: but also you are right, bad error message

andrewtruong (approved):

PR #3841, opened 2025-03-04
chore(weave): custom provider model built in object backend

jwlee64:
## Description
Adds 3 built in objects
provider
provider_model
LLM_model

runs the code gen to generate the frontend types 

adds handling in clickhouse_trace_server_batched and llm_completion for custom models

Diagram of how the new objects interact
Screenshot 2025-03-04 at 12 19 28 PM

video of it working (for this i spun up a server on my local that just passes stuff through to openai) caveat for this to work with litellm, the apis need to be the same as openai

https://github.com/user-attachments/assets/7d1cf44d-7426-4c6b-8331-46daf6e474f2

core branch with frontend - core/custom-beta
## Testing

How was this PR tested?

## Summary by CodeRabbit

- **New Features**
  - Introduced new schemas for language model configurations, enhancing the flexibility and structure of model parameters.
  - Added support for multiple model modes, including options for 'chat' and 'completion'.
  - Enhanced capabilities for managing provider information and model parameters, improving integration with custom providers.

- **Bug Fixes**
  - Improved error handling for API interactions and custom provider logic, ensuring a more reliable user experience.

- **Refactor**
  - Streamlined the logic for processing model requests and handling API keys, enhancing clarity and consistency.


weave/trace_server/interface/builtin_object_classes/provider.py
```diff
@@ -0,0 +1,14 @@
...
+
+class ReturnType(str, Enum):
+    OPENAI = "openai"
+
+
+class Provider(base_object_def.BaseObject):
+    base_url: str
+    api_key_name: str
+    extra_headers: dict[str, str] = {}
+    return_type: ReturnType = ReturnType.OPENAI
```
tssweeney, line 14: does this change between completions, response, and chat? What does "OpenAI" return type mean?

jwlee64: I was thinking change between opeanai, anthropic, and other types of provider apis

weave/trace_server/interface/builtin_object_classes/llm_model.py
```diff
@@ -0,0 +1,32 @@
...
+class ProviderModel(base_object_def.BaseObject):
+    provider: base_object_def.RefStr
+    max_tokens: int
+    mode: ModelMode = ModelMode.CHAT
+
+
+class LLMModel(base_object_def.BaseObject):
+    provider_model: base_object_def.RefStr
+    prompt: Optional[base_object_def.RefStr] = None
+    default_params: ModelParams = ModelParams()
```
tssweeney, line 32: use `Field` for default factory

weave/trace_server/interface/builtin_object_classes/llm_model.py
```diff
@@ -0,0 +1,32 @@
...
+class ModelParams(BaseModel):
+    temperature: Optional[float] = None
+    top_p: Optional[float] = None
+    max_tokens: Optional[int] = None
+    presence_penalty: Optional[float] = None
+    frequency_penalty: Optional[float] = None
+    stop: Optional[Union[str, list[str]]] = None
+
+
+class ProviderModel(base_object_def.BaseObject):
```
tssweeney, line 23: can you help me understand the purpose of a "providerModel"? Why not have these properties inside of `LLMModel`? Not saying to change, just understanding

weave/trace_server/interface/builtin_object_classes/llm_model.py
```diff
@@ -0,0 +1,32 @@
+from enum import Enum
+from typing import Optional, Union
+
+from pydantic import BaseModel
+
+from weave.trace_server.interface.builtin_object_classes import base_object_def
+
+
+class ModelMode(str, Enum):
```
tssweeney, line 9: For here and everywhere else, I would recommend being a bit more verbose in naming: `LLMProviderModelMode` - this will avoid name conflicts in the future

weave/trace_server/interface/builtin_object_classes/llm_model.py
```diff
@@ -0,0 +1,32 @@
...
+    top_p: Optional[float] = None
+    max_tokens: Optional[int] = None
+    presence_penalty: Optional[float] = None
+    frequency_penalty: Optional[float] = None
+    stop: Optional[Union[str, list[str]]] = None
+
+
+class ProviderModel(base_object_def.BaseObject):
+    provider: base_object_def.RefStr
+    max_tokens: int
```
tssweeney, line 25: was max tokens supposed to be here and in ModelParams?

jwlee64: max_tokens here means, max tokens allowed by the model (ie max max_tokens value in params)
in model params its the max tokens to use

weave/trace_server/interface/builtin_object_classes/llm_model.py
jwlee64: Theres no specific reason as to why not have these properties inside of LLMModel, maybe jsut separation of concerns
- ProviderModel: Represents the base configuration of an LLM on top of a provider
- LLMModel: Represents a specific usage configuration of a provider model

This separation provides several advantages:
1. Reusability: The same ProviderModel can be referenced by multiple LLMModel instances with different prompts and parameters
2. Separation of concerns:
   ProviderModel handles "which model am I using?"
   LLMModel handles "how am I using this model?"
3. Flexibility: You could swap out provider models while keeping the same prompt and parameters
4. Configuration management: Provider-specific details are isolated from usage-specific details

the current implementation kind of prioritizes flexibility between the two objects

tssweeney (commented):
Logically, this looks pretty solid. My main concern is that I am not seeing any tests! (:

weave/trace_server/interface/builtin_object_classes/provider.py
tssweeney: I guess I was wondering how it would change your code - but it looks like this is not used in the current implementation.

weave/trace_server/interface/builtin_object_classes/llm_model.py
jwlee64: Ah that does make more sense, thats why for the initial implementation we only use providerModel since its necessary for the configuration of the provider and not LLMModel since thats a separate milestone with prompts and the like 👍

weave/trace_server/llm_completion.py
```diff
@@ -124,3 +160,108 @@ def get_azure_credentials(model_name: str) -> tuple[str, str]:
...
+        )
+        provider_obj_res = obj_read_func(provider_obj_req)
+        provider_obj = provider_obj_res.obj
+
+        if provider_obj.base_object_class != "Provider":
+            raise InvalidRequest(
+                f"Object {provider_id} is not a Provider, it is a {provider_obj.base_object_class}"
+            )
+
+        # Extract provider information
```
tssweeney, line 223: you could convert the payload into the `Provider` class (and same for the model below) which would reduce the use of raw dictionary access

weave/trace_server/interface/builtin_object_classes/provider.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Provider(base_object_def.BaseObject):
+    base_url: str
+    api_key_name: str
+    extra_headers: dict[str, str] = Field(default_factory=dict)
+    return_type: ProviderReturnType = Field(default=ProviderReturnType.OPENAI)
+
+
+class ProviderModelMode(str, Enum):
```
tssweeney, line 19: this distinction does not seem to be respected anywhere in code - what is the use?

jwlee64: None currently, but if we ever want to add completion models it will be useful, I can remove and add when useful

weave/trace_server/interface/builtin_object_classes/llm_model.py
```diff
@@ -0,0 +1,20 @@
...
+class LLMModelDefaultParams(BaseModel):
+    temperature: Optional[float] = None
+    top_p: Optional[float] = None
+    max_tokens: Optional[int] = None
+    presence_penalty: Optional[float] = None
+    frequency_penalty: Optional[float] = None
+    stop: Optional[Union[str, list[str]]] = None
+
+
+class LLMModel(base_object_def.BaseObject):
```
tssweeney, line 17: I really agree with what you are doing here. We def need `LLModel`. However, I don't think it is strictly needed for this PR (model provider setup). Unless you need it here, i think it might be good to pull this one out of this PR. There is nearly an identical need in the on demand evals AND overlap with scorers. I think we might want to take a bit of time to make sure there is alignment there

tssweeney: Specifically, I think we will want to implement this as a `weave.Object` actually (and have server-side support for that - a little tech debt). That way it comes with it's own pre-packaged "predict" function and can run inside of evaluations.

jwlee64: that makes sense to me, no reason to keep this in until its actually implemented

tssweeney (approved):

PR #3843, opened 2025-03-05
chore(weave): Make queue size configurable

andrewtruong:
## Summary by CodeRabbit

- **New Features**
	- Introduced a configurable option to set the maximum size of the calls queue (default: 100,000) with the flexibility to override via an environment variable.
	- Enhanced the server’s call processing by dynamically applying this setting to better manage queue size during operation.

### DIFF
--- a/weave/trace/settings.py
+++ b/weave/trace/settings.py
@@ -206,9 +206,9 @@ def scorers_dir() -> str:
 
 def max_calls_queue_size() -> int:
     max_queue_size = _optional_int("max_calls_queue_size")
-    if max_queue_size == 0:
-        return 0
-    return max_queue_size or 100_000
+    if max_queue_size is None:
+        return 100_000
+    return max_queue_size
 
 
 def parse_and_apply_settings(
--- a/weave/trace/settings.py
+++ b/weave/trace/settings.py
@@ -205,7 +205,10 @@ def scorers_dir() -> str:
 
 
 def max_calls_queue_size() -> int:
-    return _optional_int("max_calls_queue_size") or 100_000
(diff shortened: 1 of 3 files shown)


weave/trace/settings.py
```diff
@@ -128,6 +128,14 @@ class UserSettings(BaseModel):
...
 
     Can be overridden with the environment variable `WEAVE_SCORERS_DIR`
+
+    """
+    max_calls_queue_size: int = 100_000
+    """
+    Sets the maximum size of the calls queue.  Defaults to 100_000.
+    Setting a value of 0 means the queue can grow unbounded.
+
+    Can be overridden with the environment variable `WEAVE_MAX_CALLS_QUEUE_SIZE`
```
jamie-rasmussen, line 138: Could you update `docs/docs/guides/core-types/env-vars.md`

weave/trace/settings.py
```diff
@@ -196,6 +204,10 @@ def scorers_dir() -> str:
     return _optional_str("scorers_dir")  # type: ignore
 
 
+def max_calls_queue_size() -> int:
+    return _optional_int("max_calls_queue_size") or 100_000
```
jamie-rasmussen, line 208: If the value is 0, wouldn't this get the default of 100_000

weave/trace/settings.py
andrewtruong: yep, done

jamie-rasmussen (approved):
Maybe confusing to leave the default `max_queue_size: int = 10_000` in `AsyncBatchProcessor`?

PR #3854, opened 2025-03-06
feat(weave): Python Saved View API

jamie-rasmussen:
## Description

Implements Saved Views in the SDK. Users can define a view in terms of columns, filters, sorting, etc. and persist them as built-in objects to a project. As part of this I implemented preliminary versions of some generally useful utilities, including a "Grid" class that can be used for tabular data, and "ObjectPath" which mirrors a TypeScript utility we have.

## Testing

Unit tests added.

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced tabular data component featuring interactive pagination, rich display, and export capabilities.
  - Added robust saved view functionality for filtering, sorting, and customizable column management including pinning options.
  - Improved navigation for complex, nested data structures with refined path utilities.
  - Implemented a streamlined client initialization process for more reliable session management.

- **Tests**
  - Expanded automated test coverage to ensure the reliability of grid operations, saved view interactions, and data traversal.
  - Added comprehensive test suites for the `Grid`, `SavedView`, and `ObjectPath` functionalities, focusing on validation and error handling.

### DIFF
--- a/tests/trace/test_saved_view.py
+++ b/tests/trace/test_saved_view.py
@@ -121,7 +121,7 @@ def test_saved_view_load(client):
     uri = saved_view.ref.uri()
     loaded_view = weave.SavedView.load(uri)
     assert loaded_view.name == saved_view.name
-    assert loaded_view.table == saved_view.table
+    assert loaded_view.view_type == saved_view.view_type
     assert loaded_view.base.definition.cols["attributes.weave.client_version"] is True
 
 
--- a/tests/trace/test_saved_view.py
+++ b/tests/trace/test_saved_view.py
@@ -1,5 +1,7 @@
 import datetime
 
+import pytest
+
 import weave
 from weave.flow.saved_view import filters_to_query, to_seconds
 from weave.trace.api import ObjectRef
@@ -22,10 +24,105 @@ def test_filters_to_query():
     assert filters_to_query(Filters()) is None
 
 
+def test_filter_op_without_client():
+    """If we haven't done an init, we don't know what entity/project a non-qualified op name is in."""
+    with pytest.raises(
+        ValueError, match="Must specify Op URI if entity/project is not known"
+    ):
+        weave.SavedView("traces", "My saved view").filter_op(
+            "Evaluation.predict_and_score"
+        )
+
+
+def test_filter_op_with_client(client):
+    view = weave.SavedView("traces", "My saved view").filter_op(
+        "Evaluation.predict_and_score"
+    )
+    assert view.base.definition.filter.op_names == [
+        "weave:///shawn/test-project/op/Evaluation.predict_and_score:*"
+    ]
+    view.filter_op(None)
+    assert view.base.definition.filter is None
+
+
+def test_filter_manipulation():
+    view = weave.SavedView("traces", "My saved view")
+
+    view.add_filter("inputs.model", "equals", "gpt-3.5-turbo")
+    assert view.base.definition.filters is not None
+    assert view.base.definition.filters.items[0].field == "inputs.model"
+    assert view.base.definition.filters.items[0].operator == "(string): equals"
+    assert view.base.definition.filters.items[0].value == "gpt-3.5-turbo"
+    view.remove_filters()
+    assert view.base.definition.filters is None
+
+    view.add_filter("inputs.model", "equals", "gpt-3.5-turbo")
+    view.remove_filter("inputs.model")
+    assert view.base.definition.filters is None
+
+    view.add_filter("inputs.model", "equals", "gpt-3.5-turbo")
+    view.remove_filter("inputs.model")
+    assert view.base.definition.filters is None
+
+
+def test_column_manipulation():
+    view = weave.SavedView("traces", "My saved view")
+
+    # Removing a column that doesn't exist doesn't error
+    view.remove_column("inputs.foo")
+
+    # Removing all columns is same as having none specified
+    view.add_column("inputs.foo")
+    view.remove_column("inputs.foo")
+    assert view.base.definition.columns is None
+
+    view.set_columns("id", "inputs.model")
+    assert len(view.base.definition.columns) == 2
+    assert view.base.definition.columns[0].path == ["id"]
+    assert view.base.definition.columns[0].label is None
+    assert view.base.definition.columns[1].path == ["inputs", "model"]
+    assert view.base.definition.columns[1].label is None
+
+    assert view.column_index(1) == 1
+    assert view.column_index("inputs.model") == 1
+    with pytest.raises(ValueError, match='Column "foo" not found'):
+        view.column_index("foo")
+
+    view.remove_column(0)
+    assert len(view.base.definition.columns) == 1
+    assert view.base.definition.columns[0].path == ["inputs", "model"]
+
+    view.rename_column(0, "new_inputs_model_name")
+    assert view.base.definition.columns[0].label == "new_inputs_model_name"
+
+    view.remove_columns()
+    assert view.base.definition.columns is None
+
+    view.add_column("inputs.foo")
+    view.insert_column(0, "inputs.bar")
+    assert len(view.base.definition.columns) == 2
+    assert view.base.definition.columns[0].path == ["inputs", "bar"]
+    assert view.base.definition.columns[1].path == ["inputs", "foo"]
+    view.remove_columns("inputs.bar")
+    assert len(view.base.definition.columns) == 1
+    assert view.base.definition.columns[0].path == ["inputs", "foo"]
+
+
 def test_saved_view_create(client):
     view = weave.SavedView("traces", "My saved view").hide_column("feedback").save()
     assert view.name == "My saved view"
     assert isinstance(view.ref, ObjectRef)
+
+
+def test_saved_view_load(client):
+    saved_view = weave.SavedView("traces", "My saved view")
+    saved_view.show_column("attributes.weave.client_version")
+    saved_view.save()
+    uri = saved_view.ref.uri()
+    loaded_view = weave.SavedView.load(uri)
+    assert loaded_view.name == saved_view.name
+    assert loaded_view.table == saved_view.table
+    assert loaded_view.base.definition.cols["attributes.weave.client_version"] is True
 
 
 def test_saved_view_column_pinning():
--- a/tests/trace/test_saved_view.py
+++ b/tests/trace/test_saved_view.py
@@ -1,0 +1,172 @@
+import datetime
+
+import weave
+from weave.flow.saved_view import filters_to_query, to_seconds
+from weave.trace.api import ObjectRef
+from weave.trace_server.interface.builtin_object_classes.saved_view import Filters
+
+
+def test_to_seconds():
+    assert to_seconds(None) is None
+    assert to_seconds("") is None
+    assert to_seconds(1) == 1
+    assert to_seconds(5.3) == 5.3
+    assert to_seconds("5.3") == 5.3
+    assert to_seconds("2025-03-02T00:00:00Z") == 1740873600.0
+    dt = datetime.datetime(2025, 3, 2, 0, 0, 0, tzinfo=datetime.timezone.utc)
+    assert to_seconds(dt) == 1740873600.0
+
+
+def test_filters_to_query():
+    assert filters_to_query(None) is None
+    assert filters_to_query(Filters()) is None
+
+
+def test_saved_view_create(client):
+    view = weave.SavedView("traces", "My saved view").hide_column("feedback").save()
+    assert view.name == "My saved view"
+    assert isinstance(view.ref, ObjectRef)
+
+
+def test_saved_view_column_pinning():
+    view = weave.SavedView("traces", "My saved view")
+
+    # Pin two columns to the left and two to the right.
+    # Confirm order including default pin is as expected.
+    view.pin_column_left("output.col_a")
+    view.pin_column_left("output.col_b")
+    view.pin_column_right("output.col_c")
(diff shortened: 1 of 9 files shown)


weave/__init__.py
```diff
@@ -44,4 +45,5 @@
     Evaluation,
     Scorer,
     AnnotationSpec,
+    SavedView,
```
tssweeney, line 48: We have called this `SavedViews` internally as the project name. Curious if you think that is the best end-user name. How might we support savings non-call-table views in the future? Do you see a conflict here? Not sure how much `Saved` adds to this? Perhaps `CallView`? Probably not the biggest deal, but it will be something we need to maintain moving forward

tests/trace/test_saved_view.py
```diff
@@ -0,0 +1,172 @@
+import datetime
```
tssweeney, line 1: Can you add some tests around the publishing / consuming of saved views - essentially testing the actual create/edit flow that you will need to hook up to the UI

weave/trace/context/weave_client_context.py
```diff
@@ -43,3 +45,36 @@ def require_weave_client() -> WeaveClient:
...
+    if entity is None and project is None:
+        if required and current_client is None:
+            raise WeaveInitError("You must call `weave.init(<project_name>)` first")
+        yield current_client
+    else:
+        from weave.trace.weave_init import init_weave
+
+        scoped_name = f"{entity}/{project}"
+        try:
+            initialized_client = init_weave(scoped_name, ensure_project_exists=False)
```
tssweeney, line 77: what happens if you have a current client AND you define an entity/project? Will that leave behind a global client different from the one already running?

weave/trace_server/interface/builtin_object_classes/saved_view.py
```diff
@@ -0,0 +1,89 @@
...
+
+class Column(BaseModel):
+    # Optional in case we want something like computed columns in the future.
+    path: Optional[list[PathElement]] = Field(default=None)
+    label: Optional[str] = Field(default=None)
+
+
+class SavedViewDefinition(BaseModel):
+    filter: Optional[CallsFilter] = Field(default=None)
+    filters: Optional[Filters] = Field(default=None)
```
tssweeney, line 63: I think we call this `query` in the other interface

weave/trace_server/interface/builtin_object_classes/saved_view.py
```diff
@@ -0,0 +1,89 @@
+from typing import Any, Literal, Optional
+
+from pydantic import BaseModel, Field
+
+from weave.trace.traverse import PathElement
+from weave.trace_server.interface.builtin_object_classes import base_object_def
+
+# Compare with definitions in weave/trace_server/trace_server_interface.py
```
tssweeney, line 8: This looks reasonable - are their any differences between this and the query layer worth calling out? Any reason we did not simply import some of those symbols?

weave/trace_server/interface/builtin_object_classes/saved_view.py
```diff
@@ -0,0 +1,89 @@
...
+    pin: Optional[Pin] = Field(default=None)
+    sort: Optional[list[SortBy]] = Field(default=None)
+    page_size: Optional[int] = Field(default=None)
+
+
+class SavedView(base_object_def.BaseObject):
+    # TODO: Should we have a spec_version literal?
+
+    # "traces" or "evaluations", type is str for extensibility
+    table: str
```
tssweeney, line 81: "table" here is sort of interesting. it sort of denormalizes part of the filter out of the definition. Can you explain this decision?

tssweeney (commented):
Overall looks really good. First read of code looks good. I want to use it a bit, but overall no obvious blockers. Nice work! (:

weave/__init__.py
jamie-rasmussen: I think it is reasonable to expect we'd want to extend this capability for other tables (e.g. datasets, prompts) at some point, so I'd avoid using CallView.

I'm inclined not to just call it "View". That will make it harder to e.g. find this specific code amongst things like the term "viewer" or database-level views.

I'm open to other possibilities (TableView?, SharedView?) but none have come to mind that I think are strictly better, so I'm inclined to keep the name SavedView.

weave/trace_server/interface/builtin_object_classes/saved_view.py
jamie-rasmussen: Just logically I want us to have the flexibility to make changes to the API layer independently, there may be backwards compatibility issues with the persisted views.

weave/trace_server/interface/builtin_object_classes/saved_view.py
jamie-rasmussen: It's kind of mirroring the split we have between the path and the query parameters in the URL. Given the idea that we might extend saved views to objects or other tables I don't really think of it as part of a filtering action. I don't feel super strongly about this if you think it would make more sense to be part of the SavedViewDefinition object.

tests/trace/test_saved_view.py
jamie-rasmussen: Added more tests including of loading back a published view.

weave/trace/context/weave_client_context.py
jamie-rasmussen: I don't believe so, the finally clause should restore the global client back to what it was.

weave/trace_server/interface/builtin_object_classes/saved_view.py
tssweeney: I see - I had interpreted SavedView as only targeting calls. But you are saying it could map to any saved view in our frontend. Given that, maybe `table` feels a bit misleading to me because there are other things called tables and it is more of a filter on the traces. Maybe just calling this `view_type` would help here.

weave/trace_server/interface/builtin_object_classes/saved_view.py
tssweeney: I see what you are saying. Given that the URL format and the API format are different, we have 2 shapes: URLFilter and APIFilter. It follows that we need to have converters in both directions. This PR (Having SavedViews contain a URLFilter) puts the responsibility to convert from URLFilter to APIFilter in both Python (`./flow/saved_view.py`) and Typescript (inside CallsPage). Which requires keeping this in sync.

Alternatively, we could have SavedViews contain an APIFilter. In that case the two conversions needed would be from APIFilter -> URLFilter and URLFilter -> APIFiler, but both happening on the UI side. This would decouple our datamodel from the specification of the URLFilter.

I think either approach can work, I had just imagined going with the latter for those reasons (hence the comment above about using the tsi types). If we are going to go with this approach, I just want us to be explicit with that decision and make sure we consider the alternative.

jamie-rasmussen: It's a fair point. I feel like the view functionality (UI and Python) is higher level than the API, e.g. there are notions of things like pinned columns in views that don't exist at the API level. This might be my bias leaking through but I'm not a fan of our Mongo-ish query API with things like `$literal` and less thans being implemented as not conditions on lists of greater than clauses etc. (And to be fair, I don't love things like `"(number): <"` from our URLFilter approach either.) It's going to be possible for users to see the internals of view definitions, diff them, etc. once the UI lands and overall I'd prefer it if they weren't exposed to what I see as the lower-abstraction-level Mongo syntax.

weave/trace_server/interface/builtin_object_classes/saved_view.py
jamie-rasmussen: I'll rename it to view_type.

weave/flow/saved_view.py
```diff
@@ -0,0 +1,727 @@
...
+            # This method is also used internally, we allow forcing all columns.
+            table_columns = None if all_columns else self.get_table_columns()
+            if table_columns:
+                columns = [c["path"].to_str() for c in table_columns]
+            else:
+                columns = None
+            query = filters_to_query(self.base.definition.filters)
+            filter = py_to_api_filter(self.base.definition.filter)
+            sort_by = py_to_api_sort_by(self.base.definition.sort)
+            return client.get_calls(
```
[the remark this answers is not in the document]
jamie-rasmussen: Yes, good catch, I will fix.

weave/trace_server/interface/builtin_object_classes/saved_view.py
```diff
@@ -0,0 +1,89 @@
...
+    # Optional in case we want something like computed columns in the future.
+    path: Optional[list[PathElement]] = Field(default=None)
+    label: Optional[str] = Field(default=None)
+
+
+class SavedViewDefinition(BaseModel):
+    filter: Optional[CallsFilter] = Field(default=None)
+    filters: Optional[Filters] = Field(default=None)
+
+    # cols is legacy column visibility config
```
tssweeney, line 65: why do we need to include something legacy here?

jamie-rasmussen: I will fix the comment, it is misleading - in the UI we are currently using a dict from field to boolean to determine column visibility. This is not ideal because it doesn't give you control over the order columns are displayed, e.g. if you wanted to display the User column before the Status you are out of luck.

I would like to move to a design where the columns shown are explicitly specified in order. The Python Saved Views implementation supports this, it will need to be added to the UI. Eventually, I want to drop the dict version.

weave/trace_server/interface/builtin_object_classes/saved_view.py
tssweeney: I think you make some good points, and I would split that feedback into a few categories:

1. Attributes of a SavedView that are not part of a query layer. Yes, absolutely things like pinned columns and page size need to be represented here in addition to query components. I am not sure the existence of additional fields justifies a deviation in the filter state, however.
2. Query spec itself (aka Mongo subset). I agree there are improvements to be supported (for example, allowing a literal value to be interpreted as `{$literal: value}`. This would be a great addition to the query layer itself. I would maintain that despite the verbosity of this syntax, it is functionally complete and well defined/tested. Extensions here will improve all things that use queries (for example, online monitoring will most certainly use a query that matches the query layer as opposed to a UI-oriented query). 
3. Showing users this query - I agree that this is less than ideal for displaying filter logic.

I think my main concern here is that once this lands, the filter spec defined here will need to be maintained forever moving forward - as well as the mapping from this representation to the api query definition in both languages. This means committing operators like `"(number): <"` deep in the DB. I am not sure the cosmetic improvement over the Mongo-like syntax justifies the additional specification to maintain and support.

For example, what happens if we add new query support at the API layer (for example a dedicated set of date operators), will we need to update this and the UI to allow those to be saved & queried? What about adding logical OR support to the UI? now that also requires updating this file as well.

Perhaps you see this as less constraining or perhaps more flexible? Maybe maintaining the mapping is not that bad? I definitely don't want to drag this out if it is not worth it, but I also feel obligated to raise these concerns to at least ensure we consider the consequences & tradeoffs.

jamie-rasmussen: I think the high-level to low-level converter is fairly straightforward. If we are going with Option 2, the user experience (e.g. `add_filter` API or the filters UI) is basically the same as the high-level representation, so we have to essentially maintain a high-to-low kind of conversion in multiple languages regardless.

If what we've stored is the low-level representation, it seems likely we also have to maintain a low-to-high conversion in multiple languages. If we didn't, when we load a Saved View definition back, we can't display something friendly to the user. And even if we decided we were OK with that, we'd still have to rewrite the high-level operations like add_filter in a way that they operated on the low-level representation instead of a high-level one, because we've not kept the information about what the high-level was. Operations like "remove the second filter" become more complicated. Even the basic stuff like add_filter becomes complicated unless we restrict the kinds of low-level queries we let the user construct. And if we limit the kinds of queries the user can construct, they aren't actually getting the tradeoff benefit of the harder to use if more powerful query language.

tssweeney: > I think the high-level to low-level converter is fairly straightforward. If we are going with Option 2, the user experience (e.g. add_filter API or the filters UI) is basically the same as the high-level representation, so we have to essentially maintain a high-to-low kind of conversion in multiple languages regardless.

Agreed - in both cases you have to implement high-to-low. But in Option 2 you only persist the low level, which leaves the high level free to change without worry.

> If what we've stored is the low-level representation, it seems likely we also have to maintain a low-to-high conversion in multiple languages. If we didn't, when we load a Saved View definition back, we can't display something friendly to the user. 

Agreed - this is the main tradeoff mentioned above (re: `think the ultimate tradeoff here is either...`). Either we maintain a low-to-high, or persist the high. My instinct is that persisting the high will have greater long-term cost as it is primarily a presentation syntax and by-definition a subset of the features.

---

Given the desire to have users construct this in python, did you consider writing a utility that allows for easier low-level query mutation/construction/display? It seems doable to also analyze the query to determine if it is suitable for higher level operations / display, else fallback to general

-- 

Fundamentally, my belief is that this higher level representation is just a convenience abstraction for the UI form/console logging. Using this state as a source of truth is both limiting in terms of expression and increases application complexity due to the need to convert from high to low in order to query. My reading is that you feel differently: that this form is actually suitable & desirable for storing query state - with the added benefit of increased readability and more straight forward modification.

Maybe we need to meet to discuss live-time. I'll set something up tomorrow.

jamie-rasmussen: OK, let's discuss it tomorrow.

jamie-rasmussen:
Closing in favor of https://github.com/wandb/weave/pull/4240

PR #3863, opened 2025-03-07
feat(weave): inline serializers, with two examples

jamie-rasmussen:
## Description

Resolves https://wandb.atlassian.net/browse/WB-23407

Takes a different approach than https://github.com/wandb/weave/pull/3702 - it adds the ability for "inline" serialization of custom objects.

PR includes two examples of this functionality, datetimes and markdown objects. A custom UI is added for each.

### Datetime
Screenshot 2025-03-18 at 4 31 17 PM
Screenshot 2025-03-18 at 4 31 23 PM

### Markdown
Screenshot 2025-03-18 at 4 30 46 PM
Screenshot 2025-03-18 at 4 30 05 PM
Screenshot 2025-03-18 at 4 30 31 PM

## Testing

Markdown and datetime in the `wf` values example.

## Summary by CodeRabbit

- **New Features**
	- Introduced an interactive Markdown viewer that allows toggling between formatted Markdown, code, and plain text views, complete with a copy-to-clipboard feature.
	- Added a dynamic timestamp display that presents dates in multiple formats with easy copy functionality.
	- Expanded custom content handling with additional views for displaying date/time and rich text, enhancing the user experience.
	- Added a new format selection menu for choosing between Markdown, Code, and Text formats.
	- Introduced a new component for rendering Markdown content with customizable display options.

### DIFF
--- a/tests/trace/test_custom_objs.py
+++ b/tests/trace/test_custom_objs.py
@@ -1,6 +1,25 @@
+from datetime import datetime, timezone
+
+import rich.markdown
 from PIL import Image
 
-from weave.trace.serialization.custom_objs import decode_custom_obj, encode_custom_obj
+import weave
+from weave.trace.serialization.custom_objs import (
+    KNOWN_TYPES,
+    decode_custom_obj,
+    decode_inline_obj,
+    encode_custom_obj,
+)
+
+
+def test_encode_custom_obj_unknown_type(client):
+    """No encoding should be done for unregistered types."""
+
+    class UnknownType:
+        pass
+
+    unknown = UnknownType()
+    assert encode_custom_obj(unknown) is None
 
 
 def test_decode_custom_obj_known_type(client):
@@ -14,3 +33,43 @@ def test_decode_custom_obj_known_type(client):
 
     assert isinstance(decoded, Image.Image)
     assert decoded.tobytes() == img.tobytes()
+
+
+def test_inline_custom_obj(client):
+    dt = datetime(2025, 3, 7, 0, 0, 0)
+    encoded = encode_custom_obj(dt)
+    assert encoded["_type"] == "CustomWeaveType"
+    assert encoded["weave_type"]["type"] == "datetime.datetime"
+    assert "files" not in encoded
+    assert "load_op" in encoded
+    assert encoded["val"] == "2025-03-07T00:00:00+00:00"
+
+    decoded = decode_inline_obj(encoded)
+    assert isinstance(decoded, datetime)
+    dt_with_tz = dt.replace(tzinfo=timezone.utc)
+    assert decoded == dt_with_tz
+
+
+def test_inline_custom_obj_needs_load_op(client):
+    """Test the condition that the current version of the SDK doesn't know how to load the object.
+
+    In that case we fallback to the saved load op."""
+    md = rich.markdown.Markdown("# Hello")
+
+    @weave.op
+    def return_markdown(md):
+        return md
+
+    _, call = return_markdown.call(md)
+    client.flush()
+
+    # Temporarily modify KNOWN_TYPES to remove markdown
+    global KNOWN_TYPES
+    original_known_types = KNOWN_TYPES.copy()
+    KNOWN_TYPES.remove("rich.markdown.Markdown")
+    try:
+        loaded = client.get_call(call.id)
+        loaded_markdown = loaded.inputs["md"]
+        assert isinstance(loaded_markdown, rich.markdown.Markdown)
+    finally:
+        KNOWN_TYPES = original_known_types
--- a/tests/trace/test_custom_objs.py
+++ b/tests/trace/test_custom_objs.py
@@ -6,8 +6,8 @@ from PIL import Image
 import weave
 from weave.trace.serialization.custom_objs import (
     KNOWN_TYPES,
-    decode_custom_obj,
-    decode_inline_obj,
+    decode_custom_files_obj,
+    decode_custom_inline_obj,
     encode_custom_obj,
 )
 
@@ -22,12 +22,12 @@ def test_encode_custom_obj_unknown_type(client):
     assert encode_custom_obj(unknown) is None
 
 
-def test_decode_custom_obj_known_type(client):
+def test_decode_custom_files_obj_known_type(client):
     img = Image.new("RGB", (100, 100))
     encoded = encode_custom_obj(img)
 
     # Even though something is wrong with the deserializer op, we can still decode
-    decoded = decode_custom_obj(
+    decoded = decode_custom_files_obj(
         encoded["weave_type"], encoded["files"], "weave:///totally/invalid/uri"
     )
 
@@ -44,7 +44,7 @@ def test_inline_custom_obj(client):
     assert "load_op" in encoded
     assert encoded["val"] == "2025-03-07T00:00:00+00:00"
 
-    decoded = decode_inline_obj(encoded)
+    decoded = decode_custom_inline_obj(encoded)
     assert isinstance(decoded, datetime)
     dt_with_tz = dt.replace(tzinfo=timezone.utc)
     assert decoded == dt_with_tz
--- a/weave-js/src/components/PagePanelComponents/Home/Browse2/CellValueMarkdown.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse2/CellValueMarkdown.tsx
@@ -1,0 +1,220 @@
+import {Popover} from '@mui/material';
+import {Tailwind} from '@wandb/weave/components/Tailwind';
+import copyToClipboard from 'copy-to-clipboard';
+import React, {ReactNode, useCallback, useRef, useState} from 'react';
+import styled from 'styled-components';
+
+import {toast} from '../../../../common/components/elements/Toast';
+import Markdown from '../../../../common/components/Markdown';
+import * as Colors from '../../../../common/css/color.styles';
+import {Button} from '../../../Button';
+import {CodeEditor} from '../../../CodeEditor';
+import {
+  DraggableGrow,
+  DraggableHandle,
+  PoppedBody,
+  StyledTooltip,
+  TooltipHint,
+} from '../../../DraggablePopups';
+import {Icon} from '../../../Icon';
+
+type CellValueMarkdownProps = {
+  value: string;
+};
+
+const Collapsed = styled.div`
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  cursor: pointer;
+`;
+Collapsed.displayName = 'S.Collapsed';
+
+const TooltipContent = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+TooltipContent.displayName = 'S.TooltipContent';
+
+const TooltipText = styled.div`
+  max-height: 35vh;
+  overflow: auto;
+  white-space: break-spaces;
+`;
+TooltipText.displayName = 'S.TooltipText';
+
+const Popped = styled.div`
+  border-radius: 4px;
+  padding: 4px 8px;
+  font-size: 0.8rem;
+  word-wrap: break-word;
+  background-color: #fff;
+  color: ${Colors.MOON_700};
+  border: 1px solid ${Colors.MOON_300};
+`;
+Popped.displayName = 'S.Popped';
+
+const Toolbar = styled.div`
+  display: flex;
+  align-items: center;
+  padding: 4px 0;
+`;
+Toolbar.displayName = 'S.Toolbar';
+
+const Spacer = styled.div`
+  flex: 1 1 auto;
+`;
+Spacer.displayName = 'S.Spacer';
+
+const truncateText = (text: string, maxLength: number) => {
+  if (text.length <= maxLength) {
+    return text;
+  }
+  return text.slice(0, maxLength) + '...';
+};
+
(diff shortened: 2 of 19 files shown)


tests/trace/test_custom_objs.py
```diff
@@ -1,6 +1,25 @@
...
+
+import rich.markdown
 from PIL import Image
 
-from weave.trace.serialization.custom_objs import decode_custom_obj, encode_custom_obj
+import weave
+from weave.trace.serialization.custom_objs import (
+    KNOWN_TYPES,
+    decode_custom_obj,
+    decode_inline_obj,
```
andrewtruong, line 10: Please update the readme to reflect this new path

tests/trace/test_custom_objs.py
```diff
@@ -14,3 +33,43 @@ def test_decode_custom_obj_known_type(client):
 
     assert isinstance(decoded, Image.Image)
     assert decoded.tobytes() == img.tobytes()
+
+
+def test_inline_custom_obj(client):
+    dt = datetime(2025, 3, 7, 0, 0, 0)
+    encoded = encode_custom_obj(dt)
```
andrewtruong, line 40: Why is this `encode_custom_obj` but below it's `decode_inline_obj`?  Is there no encode inline path?

weave/type_handlers/Markdown/markdown.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any
+
+from rich.markdown import Markdown
+
+from weave.trace.serialization import serializer
+
+
+def save(obj: Markdown) -> dict[str, Any]:
```
andrewtruong, line 8: nit: You probably want a `TypedDict` here

weave/trace/serialization/custom_objs.py
```diff
@@ -47,25 +48,61 @@ def encode_custom_obj(obj: Any) -> dict | None:
...
         # Calculating this URL is blocking, but we only have to pay it once per custom type
         load_instance_op_ref = wc._save_op(serializer.load, "load_" + serializer.id())  # type: ignore
         load_op_uri = load_instance_op_ref.uri()
 
-    encoded_path_contents = {
-        k: (v.encode("utf-8") if isinstance(v, str) else v)  # type: ignore
-        for k, v in art.path_contents.items()
-    }
-    return {
+    encoded = {
```
andrewtruong, line 59: The lack of typing here is unfortunate and makes it harder to understand what's going on.

Do I understand `encoded["val"]` to mean the inline-serialized object?

weave/trace/serialization/custom_objs.py
```diff
@@ -47,25 +48,61 @@ def encode_custom_obj(obj: Any) -> dict | None:
...
         "_type": "CustomWeaveType",
         "weave_type": {"type": serializer.id()},
-        "files": encoded_path_contents,
         "load_op": load_op_uri,
     }
 
+    # If the save method just takes one argument, it is an inline serializer
+    save_signature = inspect.signature(serializer.save)
+    param_count = len(save_signature.parameters)
```
andrewtruong, line 67: hmm I don't love this but maybe ok for now.

Longer term I think it would be better to type this literally as an inline vs. artifact serializer

weave/trace/serialization/serialize.py
```diff
@@ -93,8 +93,9 @@ def to_json(
                 for k, v in as_dict.items()
             }
         return fallback_encode(obj)
+    if "val" in encoded:
```
andrewtruong, line 96: would add a comment here noting that it's inline

andrewtruong: or even a func `is_inline`

weave-js/src/components/PagePanelComponents/Home/Browse2/CellValueMarkdown.tsx
andrewtruong, line 1: What's the difference between Browse2 and Browse3?

jamie-rasmussen: Browse2 was a previous iteration of the UI dating back to the "weaveflow" era. We had it and the new Weave UI going in parallel for a while and there were some components shared between the two that lived in Browse2. I have been slowly refactoring things to get rid of Browse2, e.g. https://github.com/wandb/weave/pull/3516 

The CellValue renderers are one of the last pieces remaining in Browse2. I'm putting these new files alongside the existing ones but intend to move them all into Browse3 shortly after this lands.

weave/trace/serialization/custom_objs.py
jamie-rasmussen: Yes. I agree that stronger typing would be better but would rather leave that for follow up - I think there may need to be several changes because we do things like pop the `_type` key off the dict.

weave/trace/serialization/custom_objs.py
jamie-rasmussen: I'll make the typing a little stricter and add some typeguard methods for readability. Doing it properly would require more of a change than I'd like to do right now because of circular imports.

weave/trace/serialization/serialize.py
jamie-rasmussen: Adding func

tests/trace/test_custom_objs.py
jamie-rasmussen: encode_custom_obj handles both inline and file-based. I agree the asymmetry is kind of ick. I think `from_json` should not be digging into the internals of custom obj decoding as much as it is. I was a little wary of making too extreme a change to this delicate and important part of the code, but I can take a crack at it if you would like.

weave/trace/serialization/custom_objs.py
andrewtruong: Ok I wont block here, but I think it might be easier than you're describing (at least for MVP)

For poppable keys, you can write something like:
```py
class EncodedDict(TypedDict):
    _type: NotRequired[str]
    ...
```

tests/trace/test_custom_objs.py
andrewtruong: If it's not too hard I would try to do that now, or as a fast follow.  At minimum we should document this in the serialization doc

weave/trace/serialization/custom_objs.py
jamie-rasmussen: `NotRequired` is actually giving me issues in 3.9, I think there's some bad interaction with our code serialization.

andrewtruong: oh?  It was introduced in py311 so you might have to do import from `typing_extensions`

https://peps.python.org/pep-0655/

andrewtruong: if it still breaks then it's fine, I won't block here

tests/trace/test_custom_objs.py
```diff
@@ -14,3 +33,43 @@ def test_decode_custom_obj_known_type(client):
...
+    client.flush()
+
+    # Temporarily modify KNOWN_TYPES to remove markdown
+    global KNOWN_TYPES
+    original_known_types = KNOWN_TYPES.copy()
+    KNOWN_TYPES.remove("rich.markdown.Markdown")
+    try:
+        loaded = client.get_call(call.id)
+        loaded_markdown = loaded.inputs["md"]
+        assert isinstance(loaded_markdown, rich.markdown.Markdown)
```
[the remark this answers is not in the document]
andrewtruong: hmm, well if you want to test that it will use the saved load op, I think you can just MagicMock this and check that the `load_op` func was called?

weave/trace/serialization/custom_objs.py
andrewtruong: ok, unfortunate but I wont block.  Thanks for checking Jamie!

tests/trace/test_custom_objs.py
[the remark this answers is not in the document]
jamie-rasmussen: But that wouldn't check that the deserialization actually worked, right?

tests/trace/test_custom_objs.py
jamie-rasmussen: The longer I looked the harder it seemed. I updated some method names for clarity and I'm adding notes to the README.

weave/trace/serialization/op_type.py
```diff
@@ -412,14 +412,18 @@ def _get_code_deps(
...
+                var_value_name
                 and hasattr(var_value, "__module__")
                 and var_value.__module__ != fn.__module__
             ):
-                import_line = f"from {var_value.__module__} import {var_value.__name__}"
-                if var_value.__name__ != var_name:
-                    import_line += f"as {var_name}"
+                import_line = f"from {var_value.__module__} import {var_value_name}"
+                if var_value_name != var_name:
+                    import_line += f" as {var_name}"
```
jamie-rasmussen, line 426: @andrewtruong - unrelated bug fix, there should be a space before the `as`

weave/trace/serialization/op_type.py
```diff
@@ -412,14 +412,18 @@ def _get_code_deps(
                     import_code.append(import_line)
 
         else:
+            # _name will work for typing_extensions.NotRequired in 3.9
```
jamie-rasmussen, line 415: This PR uses `TypedDict` and `NotRequired` as part of the markdown serializer. As part of our whole approach to custom objects, the "deserializer" Op is serialized as a backup. Serialization of the deserializer Op was failing on Python 3.9 because `NotRequired` on 3.9 requires using typing_extensions and the implementation doesn't have a `__name__` attribute.

andrewtruong: Nice catch!  I would put this comment in the code

weave/trace/serialization/serialize.py
```diff
@@ -290,10 +298,13 @@ def from_json(obj: Any, project_id: str, server: TraceServerInterface) -> Any:
                 {k: from_json(v, project_id, server) for k, v in obj.items()}
             )
         elif val_type == "CustomWeaveType":
-            files = _load_custom_obj_files(project_id, server, obj["files"])
-            return custom_objs.decode_custom_obj(
-                obj["weave_type"], files, obj.get("load_op")
-            )
+            if _is_inline_custom_obj(obj):
+                return custom_objs.decode_custom_inline_obj(obj)
+            else:
```
andrewtruong, line 303: nit: redundant `else`

weave/trace/serialization/serializer.py
```diff
@@ -36,14 +36,54 @@ def load_instance(
...
+from typing import Any, Callable, Union
+
+from typing_extensions import TypeGuard
+
+# Not locking down the return type for inline encoding but
+# it would be expected to be something like a str or dict.
+InlineSave = Callable[[Any], Any]
+# Second parameter should really be MemTraceFilesArtifact
+# but we are avoiding a circular import.
+FileSave = Callable[[Any, Any, str], None]
```
andrewtruong, line 50: You can do an `if TYPE_CHECKING: import ...` for this

weave/trace/serialization/serializer.py
```diff
@@ -36,14 +36,54 @@ def load_instance(
...
+Save = Union[InlineSave, FileSave]
```
andrewtruong, line 51: And with future you can do `InlineSave | FileSave` heree

jamie-rasmussen: I tried that but even with with future it fails on 3.9:
```
TypeError: unsupported operand type(s) for |: '_CallableGenericAlias' and '_CallableGenericAlias'
```

weave/trace/serialization/README.md
```diff
@@ -42,15 +42,22 @@ NOTE: Not all serialization is reversible, which can be surprising and frustrati
...
 
-1. `encode_custom_obj()`, which encodes custom objects using registered serializers
-2. `decode_custom_obj()`, which decodes custom objects using the appropriate serializer
+1. `encode_custom_obj()`, which encodes custom objects using registered serializers (both inline and file-based)
+2. `decode_custom_inline_obj()`, which decodes custom inline objects using the appropriate serializer
+3. `decode_custom_files_obj()`, which decodes custom file-based objects using the appropriate serializer
 
 #### Custom Object Serialization Flow
 
+This diagram shows the flow for file-based custom objects. Inline objects are similar but simpler.
```
andrewtruong, line 59: Can you show the inline diagram too?

weave/trace/serialization/serializer.py
jamie-rasmussen: Like this? (With an explanatory comment)
```
if TYPE_CHECKING:
    from weave.trace.serialization.mem_artifact import MemTraceFilesArtifact

    FileSave = Callable[[Any, MemTraceFilesArtifact, str], None]
else:
    FileSave = Callable[[Any, Any, str], None]
 ```

Comment (andrewtruong) [reply weave/trace/serialization/serializer.py:50]: I believe you can do this:

```py
if TYPE_CHECKING:
    from weave.trace.serialization.mem_artifact import MemTraceFilesArtifact

FileSave = Callable[[Any, "MemTraceFilesArtifact", str], None]
```

weave/trace/serialization/serializer.py
andrewtruong: ok no worries

weave/trace/serialization/README.md
jamie-rasmussen: I'll add one, but I think they are a little confusing (e.g. it looks like the decode is calling the from_json when it is the other way around)

PR #3865, opened 2025-03-07
fix(weave): @weave.op decorator destroys type information

zbirenbaum:
## Description

- WB-22816

Changes weave.op to preserve type info of the function it decorates. Return type modified to be either a decorator of the original function or the type of the original function instead of `Op`

## Testing

Imported weave, decorated a function, type signature was correct

## Summary by CodeRabbit

- **Refactor**
	- Enhanced type safety and consistency for decorated functions, ensuring that function interfaces remain unchanged for more reliable behavior.

### DIFF
--- a/weave/trace/op.py
+++ b/weave/trace/op.py
@@ -16,16 +16,14 @@ from typing import (
     Any,
     Callable,
     Optional,
+    ParamSpec,
     Protocol,
     TypedDict,
+    TypeGuard,
+    TypeVar,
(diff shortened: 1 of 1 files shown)


zbirenbaum:
This seems to break things like ` _set_on_finish_handler: Callable[[OnFinishHandlerType], None]` since the return type is no longer an `Op` which has that property. Needs some more thought on how to address

PR #3870, opened 2025-03-10
chore(weave): Swap to codegen server bindings

andrewtruong:
This PR swaps in the Stainless generated bindings into RemoteHTTPTraceServer


weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -65,26 +76,36 @@ def __init__(
                 self._flush_calls,
                 max_queue_size=max_calls_queue_size(),
             )
-        self._auth: Optional[tuple[str, str]] = None
         self.remote_request_bytes_limit = remote_request_bytes_limit
 
+        self.stainless_client = WeaveTrace(
+            username=entity_name if entity_name else "",
```
tssweeney, line 82: i don't understand why you are passing entity name around. It should always be blank

tests/trace_server/test_remote_http_trace_server.py
```diff
@@ -1,127 +0,0 @@
-import datetime
```
tssweeney, line 1: do we not longer want to test error code handling?

andrewtruong: not sure why this got deleted; undeleted

weave/trace_server_bindings/remote_http_trace_server.py
tssweeney: but why is it needed? it is not used in auth

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -166,205 +186,139 @@ def _flush_calls(
...
-        return self._generic_request(
-            "/calls/query_stats", req, tsi.CallsQueryStatsReq, tsi.CallsQueryStatsRes
+        stream = self.calls_query_stream(req)
+        return tsi.CallsQueryRes(calls=list(stream))
+
+    @validate_call
+    def calls_query_stream(self, req: tsi.CallsQueryReq) -> Iterator[tsi.CallSchema]:
+        # TODO: Technically this does not exist on the interface yet.
+        # return self.stainless_client.calls.stream_query(query=req.query)
+        return iter([])
```
tssweeney, line 241: This is a big bug - this is the call query!

PR #3872, opened 2025-03-12
fix(weave): Reduce duplicate file uploads

tssweeney:
Uses LRU cache to reduce duplicate file uploads of the same content digest

## Summary by CodeRabbit

- **New Features**
  - Introduced file upload caching to improve performance by reducing redundant network calls.
  - Implemented an in-memory caching mechanism with an LRU eviction policy for efficient file management.
  - Enhanced image dataset creation with specific colors for each image associated with questions.

- **Tests**
  - Added new tests for image-based evaluation.
  - Expanded test suites to validate cache behavior, performance under concurrent access, and stress conditions.
  - Improved resilience tests to ensure proper handling of server errors.


tests/trace/test_wave_client_file_cache_performance.py
```diff
@@ -0,0 +1,165 @@
...
+        f"Completed {num_threads * operations_per_thread} operations in {duration:.2f} seconds"
+    )
+    print(f"Performance: {ops_per_second:.2f} operations/second")
+
+    # Ensure the cache size hasn't exceeded max_size
+    assert cache.size() <= cache.max_size
+
+
+# Only run this test when explicitly requested, as it's resource-intensive
+@pytest.mark.skip(reason="Long-running stress test, run manually")
```
andrewtruong, line 119: Tests like these might be better run in a nightly shard.  If we skip we probably wont be running it frequently enough

weave/trace/weave_client_send_file_cache.py
```diff
@@ -0,0 +1,171 @@
...
+from typing import Generic, Optional, TypeVar
+
+from weave.trace_server.trace_server_interface import FileCreateReq, FileCreateRes
+
+# Define generic type variables
+K = TypeVar("K")  # Key type
+V = TypeVar("V")  # Value type
+
+
+class ThreadSafeLRUCache(Generic[K, V]):
```
andrewtruong, line 19: Isn't `functools.lru_cache` already thread safe?  Couldn't you just decorate:
```py
@lru_cache(...)
def _send_file_create(...)
```

andrewtruong: https://docs.python.org/3/library/functools.html#functools.lru_cache

tests/trace/test_image_evaluation.py
```diff
@@ -0,0 +1,51 @@
...
+
+    publish_all(
+        [
+            weave.Evaluation.evaluate,
+            weave.Evaluation.predict_and_score,
+            weave.Evaluation.summarize,
+            basic_scorer,
+            predict,
+        ]
+    )
```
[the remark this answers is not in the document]
tssweeney: yes exactly. else the tests will upload these as well and it is not as clear which files are uploaded

weave/trace/weave_client_send_file_cache.py
andrewtruong: In general I prefer the builtins because they are battle-tested and should be faster. 
 Did you consider wrapping the builtin `lru_cache`?

Otherwise, I'll add comments that assume you go with this impl

weave/trace/weave_client_send_file_cache.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def __init__(self, max_size: int = 1000) -> None:
+        """Initialize an LRU cache with a maximum size.
+
+        Args:
+            max_size: Maximum number of items to store in the cache.
+                     If set to 0 or negative, cache size is unlimited.
+        """
+        self._max_size = max(0, max_size)  # Ensure max_size is non-negative
+        self._cache: OrderedDict[K, V] = OrderedDict()
```
andrewtruong, line 30: This is unnecessary.  Dicts are defined to be ordered in the python language spec as of py37

weave/trace/weave_client_send_file_cache.py
```diff
@@ -0,0 +1,171 @@
...
+        If the key already exists, it will be moved to the end (most recently used position).
+        If max_size is reached, the least recently used item will be evicted.
+        """
+        with self._lock:
+            # If key exists, remove it so we can re-insert at the end (most recently used)
+            if key in self._cache:
+                self._cache.pop(key)
+
+            # If we've reached max size and max_size is not 0 (unlimited), remove oldest item
+            if self._max_size > 0 and len(self._cache) >= self._max_size:
```
andrewtruong, line 45: nit: `if 0 < self._max_size < len(self._cache)`

weave/trace/weave_client_send_file_cache.py
andrewtruong, line 19: This is more specifically a `Mapping[K, V]`

weave/trace/weave_client_send_file_cache.py
```diff
@@ -0,0 +1,171 @@
...
+    def put(self, key: K, value: V) -> None:
+        """Thread-safe insertion/update with LRU tracking.
+
+        If the key already exists, it will be moved to the end (most recently used position).
+        If max_size is reached, the least recently used item will be evicted.
+        """
+        with self._lock:
+            # If key exists, remove it so we can re-insert at the end (most recently used)
+            if key in self._cache:
+                self._cache.pop(key)
```
andrewtruong, line 42: With `OrderedDict`, there is a more efficient method `move_to_end` that achieves this effect.
https://docs.python.org/3/library/collections.html#collections.OrderedDict.move_to_end

weave/trace/weave_client_send_file_cache.py
```diff
@@ -0,0 +1,171 @@
...
+        """Thread-safe retrieval with LRU tracking.
+
+        If the key exists, it will be moved to the end (most recently used position).
+        """
+        with self._lock:
+            if key not in self._cache:
+                return None
+
+            # Move the key to the end by removing and reinserting it
+            value = self._cache.pop(key)
```
andrewtruong, line 63: see `move_to_end` above

weave/trace/weave_client_send_file_cache.py
tssweeney: lru_cache is insufficient here since in a multi-threaded context, the same file can be uploaded multiple times

weave/trace/weave_client_send_file_cache.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(self, max_size: int = 1000) -> None:
+        """Initialize an LRU cache with a maximum size.
+
+        Args:
+            max_size: Maximum number of items to store in the cache.
+                     If set to 0 or negative, cache size is unlimited.
+        """
+        self._max_size = max(0, max_size)  # Ensure max_size is non-negative
+        self._cache: OrderedDict[K, V] = OrderedDict()
+        self._lock = threading.RLock()
```
andrewtruong, line 31: A single lock here is ok, but it might be better to have per-key locks/conditions.

I think the current impl means that if I'm writing `key1` I can't look up `key2` until the write is complete?

weave/trace/weave_client_send_file_cache.py
andrewtruong: But can you wrap `lru_cache` with locks/conditions?

weave/trace/weave_client_send_file_cache.py
```diff
@@ -0,0 +1,171 @@
...
+
+            # If the new max_size is 0, it means unlimited size, so no eviction needed
+            if self._max_size == 0:
+                return
+
+            # If we have more items than the new max_size, evict the oldest ones
+            # Need to make a copy to keep LRU order intact during eviction
+            if len(self._cache) > self._max_size:
+                # Keep only the max_size most recently used items
+                keys_to_keep = list(self._cache.keys())[-self._max_size :]
```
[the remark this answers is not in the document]
tssweeney: ok, changed

weave/trace/weave_client_send_file_cache.py
```diff
@@ -0,0 +1,171 @@
...
+                new_cache: OrderedDict[K, V] = OrderedDict()
+                for key in keys_to_keep:
+                    new_cache[key] = self._cache[key]
+                self._cache = new_cache
```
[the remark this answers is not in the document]
tssweeney: fixed

weave/trace/weave_client_send_file_cache.py
tssweeney: yup

weave/trace/weave_client_send_file_cache.py
tssweeney: 👍

weave/trace/weave_client_send_file_cache.py
tssweeney: 👍

weave/trace/weave_client_send_file_cache.py
tssweeney: You are correct, but these are fairly quick operations. I think the complexity of per-key locks is not worth the refactor

andrewtruong (approved):

PR #3873, opened 2025-03-12
chore(weave): Add better exceptions (e.g. for KeyboardInterrupt)

andrewtruong:
Previously, a keyboard interrupt would result in a "forever running" state.  Now it correctly shows as an error.

Before
image

After
image

## Summary by CodeRabbit

- **Refactor**
  - Enhanced error handling to manage interruptions more gracefully, including unexpected exit signals and user-initiated terminations. This improvement aims to provide a smoother and more robust experience during occasional interruptions.

### DIFF
--- a/weave/trace/op.py
+++ b/weave/trace/op.py
@@ -339,7 +339,7 @@ def _execute_op(
 
     try:
         res = func(*args, **kwargs)
-    except Exception as e:
+    except BaseException as e:
         handle_exception(e)
     else:
         return process(res)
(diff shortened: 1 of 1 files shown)


weave/trace/op.py
```diff
@@ -339,7 +339,7 @@ async def _call_async() -> tuple[Any, Call]:
 
     try:
         res = func(*args, **kwargs)
-    except Exception as e:
+    except (Exception, SystemExit, KeyboardInterrupt) as e:
```
neutralino1, line 342: TIL `KeyboardInterrupt` does not inherit from `Exception` 🤝

neutralino1 (approved):
Thanks @andrewtruong

gtarpenning (approved):

PR #3882, opened 2025-03-17
feat(weave): OpenAI Agents SDK integration

andrewtruong:
https://wandb.atlassian.net/browse/WB-23857

Adds basic support for the OpenAI Agents SDK

TODO:
1. Responses API should be wrapped in OAI patcher and then consumed here
2. Consider emulating errors to get the UI to show failed or problematic steps
3. Consider moving more data out into metadata
image

## Summary by CodeRabbit

- **New Features**
  - Introduced an integration that captures and logs detailed agent interactions, including various execution stages.
  - Enhanced configuration settings now automatically manage the deployment and cleanup of the integration for better trace management and diagnostic visibility.
  - Added support for additional configuration related to OpenAI agents in the autopatching system.
  - Expanded testing capabilities to include OpenAI agents.
  - Added a new optional dependency for OpenAI agents to enhance project functionalities.
  - Introduced new test cases to validate the functionality of OpenAI agents, including homework-related queries.


tests/integrations/openai_agents/openai_agents_test.py
```diff
@@ -0,0 +1,260 @@
...
+    assert val.role == "assistant"
+    assert val.type == "message"
+    assert val.status == "completed"
+    assert (
+        val.content[0].text
+        == "Code calls to itself,  \nInfinite loops in silence,  \nPatterns emerge clear."
+    )
+
+
+# @pytest.mark.skip(reason="Not clear why VCR fails for this test")
```
zbirenbaum, line 56: Is this comment still accurate?

weave/integrations/openai_agents/openai_agents.py
zbirenbaum, line 1: Left a couple comments but this is a bit hard to parse. Would it be possible to separate out the handlers for Tracing and Spans into their own files since there is already an openai_agents dir so it's a bit clearer to review?

weave/integrations/openai_agents/openai_agents.py
```diff
@@ -0,0 +1,507 @@
...
+            "metrics": {},
+            "metadata": {},
+        }
+
+        # Create a call for this trace
+        wc = require_weave_client()
+        trace_call = wc.create_call(
+            op="openai_agent_trace",
+            inputs={"name": trace.name},
+            parent=call_context.get_current_call(),
```
zbirenbaum, line 87: How is the call structure being resolved for a trace? They have group_id but only spans have a parent_id

weave/integrations/openai_agents/openai_agents.py
```diff
@@ -0,0 +1,507 @@
...
+            attributes={"type": "task", "agent_trace_id": trace.trace_id},
+            display_name=trace.name,
+        )
+        self._trace_calls[trace.trace_id] = trace_call
+
+    def on_trace_end(self, trace: tracing.Trace) -> None:
```
zbirenbaum, line 93: not a requirement, but was thinking this could be organized a bit by having a separate processor implemented for spans and traces. Then just call `self.span_processor.on_trace_end(trace: Trace)`

andrewtruong: How would you get the "trace" call from the `SpanProcessor`?

weave/integrations/openai_agents/openai_agents.py
andrewtruong: In this context, `trace_call` is basically a dummy call that acts like the "trace" container.

All call structure is otherwise modelled with child calls

weave/integrations/openai_agents/openai_agents.py
andrewtruong: I would prefer this too, but it's not obvious how to get the "trace" call into the "span" processor

tests/integrations/openai_agents/openai_agents_test.py
andrewtruong: nope, thanks for catching

noxfile.py
```diff
@@ -92,6 +93,9 @@ def tests(session, shard):
         env["MISTRAL_API_KEY"] = session.env.get("MISTRAL_API_KEY")
         env["OPENAI_API_KEY"] = session.env.get("OPENAI_API_KEY")
 
+    if shard == "openai_agents":
+        env["OPENAI_API_KEY"] = session.env.get("OPENAI_API_KEY", "DUMMY_API_KEY")
```
zbirenbaum, line 97: Maybe I'm doing something wrong but this line replaces the API key to DUMMY_API_KEY even when passed explicitly like so:
`OPENAI_API_KEY=<api-key> nox -e "tests-3.13(shard='openai_agents')"`

Since this makes it impossible to re-record and I can't find a way to pass environment vars to the command directly (--env sets the virtual environment not the vars) it might be an issue.

zbirenbaum, line 97: This never reads from the env variable even when explicitly passed which breaks re-recording

tests/integrations/openai_agents/test_autopatch.py
zbirenbaum, line 1: remove empty file

zbirenbaum (approved):

PR #3883, opened 2025-03-17
fix(weave): correct sorting by trace name

gtarpenning:
## Description

[WB-23790](https://wandb.atlassian.net/browse/WB-23790)

Sort on display name in the trace table, for evals and traces. 

## Testing

todo

[WB-23790]: https://wandb.atlassian.net/browse/WB-23790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
	- The calls page now displays a more user-friendly label when an alternative name is provided. If a custom name is available, it is shown instead of the default identifier, offering clearer and more descriptive call records.


gtarpenning:
This does not actually work, we need to do server side sorting

PR #3886, opened 2025-03-17
feat(weave): Add basic support for OpenAI Responses

andrewtruong:
https://wandb.atlassian.net/browse/WB-23858

The response output is quite verbose

In future, we should make a decision RE: where to put the full response output data (some of it would fit under metadata/attributes, others output, etc.)

image

## Summary by CodeRabbit

- **New Features**
	- Enhanced API response handling with improved support for streaming, synchronous, asynchronous, and tool integration scenarios.
- **Tests**
	- Expanded test coverage across various interactions, ensuring reliable performance for both streaming and non-streaming responses.
	- New test cases added for validating synchronous and asynchronous response creation, streaming capabilities, and tool calling functionalities.

### DIFF
--- a/weave/integrations/openai/openai_sdk.py
+++ b/weave/integrations/openai/openai_sdk.py
@@ -382,8 +382,9 @@ def create_wrapper_async(settings: OpSettings) -> Callable[[Callable], Callable]
 
 
 def _pad_output(acc: Response, value: ResponseStreamEvent) -> Response:
-    while len(acc.output) <= value.output_index:
-        acc.output.append("")
+    if len(acc.output) <= value.output_index:
+        missing_len = value.output_index - len(acc.output) + 1
+        acc.output.extend([""] * missing_len)
     return acc
 
 


weave/integrations/openai/openai_sdk.py
```diff
@@ -380,6 +381,228 @@ def _openai_stream_options_is_set(inputs: dict) -> bool:
...
+            ResponseAudioDeltaEvent,
+            ResponseAudioTranscriptDeltaEvent,
+        ),
+    ):
+        # Not obvious how to handle these since there is no output_index
+        if not acc.output:
+            acc.output = [""]
+        acc.output[0] += value.delta
+
+    elif isinstance(value, ResponseTextAnnotationDeltaEvent):
```
zbirenbaum, line 486: It seems like the definition is the 'full' output would be the contents of the file + annotated change at the computed index. This comment shows how to compute the full output in conjunction with the annotation:
https://github.com/openai/openai-python/issues/1792#issuecomment-2596981504

Maybe just modify the output index for the thing it annotates to include the annotation? This is a pretty poorly documented case so might be fine to skip it for now

weave/integrations/openai/openai_sdk.py
```diff
@@ -380,6 +381,228 @@ def _openai_stream_options_is_set(inputs: dict) -> bool:
     return wrapper
 
 
+def _pad_output(acc: Response, value: ResponseStreamEvent) -> Response:
+    while len(acc.output) <= value.output_index:
+        acc.output.append("")
```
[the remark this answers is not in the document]
andrewtruong: ok, I think your way is more efficient

We're appending empty string for padding because later we set `acc.output[value.output_index] = ...`

weave/integrations/openai/openai_sdk.py
andrewtruong: I think that comment is for the `Assistants` API -- is it similar for `Responses`?

In any case, if this comes up we can add better support.  In practice, I'm finding that I always hit `ResponseCompletedEvent` which has everything assembled already

zbirenbaum: I think that's a good call since it's inconsequential in practice. Resolving this for now

zbirenbaum (approved):
LGTM

PR #3889, opened 2025-03-17
chore(weave): add highlight state when editing filters

gtarpenning:
## Description

[WB-23846](https://wandb.atlassian.net/browse/WB-23846)

Add highlight state when editing filter. 

## Testing
Screenshot 2025-03-17 at 4 56 49 PM
Screenshot 2025-03-17 at 4 56 58 PM

[WB-23846]: https://wandb.atlassian.net/browse/WB-23846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Enhanced filter editing experience with clear visual feedback when a filter is actively being edited.
  - The editing state now automatically resets when a filter is removed or editing is exited, ensuring seamless interactions.

- **Style**
  - Updated filter tag presentation with distinct color cues to differentiate between editing and default states.


neutralino1 (commented):
I wonder if we can leverage the `:focus` pseudo-class instead of a React state to do this. It would be more minimal, lower level, and more centralized (meaning we wouldn't have to do this for each individual field).

Alternatively, what about the `focus` and `blur` events?

gtarpenning:
> I wonder if we can leverage the `:focus` pseudo-class instead of a React state to do this. It would be more minimal, lower level, and more centralized (meaning we wouldn't have to do this for each individual field).
> 
> Alternatively, what about the `focus` and `blur` events?
@neutralino1 

Played around with this, afaict using `focus` is going to be hard, because we want to update a totally different component. we want the children to light up when selecting the filter in the edit window, only while actively editing. Maybe i'm misunderstanding though.

neutralino1 (approved):

PR #3891, opened 2025-03-18
feat(ui): Allow displaying W&B Run information in calls grid

jamie-rasmussen:
## Description

Add a (default hidden) column that shows run information for a call. Does a GQL lookup of the run display name. Only Option+Click filtering is supported, and that's kind of hacky.

Screenshot 2025-03-17 at 9 51 03 PM

Screenshot 2025-03-17 at 9 51 16 PM


jamie-rasmussen:
> lgtm, when you say default hidden, users will have to toggle it with the column management popup?

Correct - most of our users don't use Weave with W&B runs so hiding it instead of showing most people an empty column seemed to be the way to go.

PR #3892, opened 2025-03-18
chore(ui): Custom providers drawer design update

m-rgba:
## Description

Design update for custom AI provider drawer.

| Before | After |
| -------- | ------- |
| CleanShot 2025-03-18 at 13 24 21 | CleanShot 2025-03-24 at 08 32 45 |

- **ToDo**: Need to update with docs link in the callout when we're ready to publish.
- Updated header area for height and font weight.
- Updated font to use Source Sans Pro / made typography feel right.
- Updated spacing between form elements.
- Made some tweaks to padding and sizing to prevent page overflow when the screen is short.
- Updated padding around the form.
- Adjusted button sizes on the clear buttons to match the inline inputs.
- Added a sample endpoint description to the base URL field.
- Made buttons full width - added submit / edit logic to the text.

Functional changes:
- Used the ResizeablDrawer cause we should probably center around using those for most of these non-peek drawers.
- Added default header and model rows to save on an action for adding a new row.
- Stripped empty headers (similar to the stripping model logic).
- Stripped the trailing `/` for the baseURL onSubmit (cause I do that all the time other places when I'm setting up custom models and it's annoying to debug).


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/OverviewPage/AddProviderDrawer.tsx
```diff
@@ -155,82 +164,130 @@ export const AddProviderDrawer: React.FC<AddProviderDrawerProps> = ({
...
             <Box>
-              <Typography variant="subtitle2" sx={{mb: 1, fontWeight: 'bold'}}>
-                API key name
+              <Typography
+                sx={{...sharedTypographyStyle, mb: 1, fontWeight: '600'}}>
+                API key / token
               </Typography>
               <TextField
-                placeholder="Enter API key name"
+                placeholder="Enter API key / token..."
```
jwlee64, line 270: im worried this might be confusing,
i used name because putting the token itself here doesnt work
you have to put the name and then add the name and token to the team secrets

jwlee64 (approved):

neutralino1:
Maybe your target branch needs to be rebased to master as well. It seems the PR diff is showing many unrelated changes, which makes it hard to review atomically.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/OverviewPage/AddProviderDrawer.tsx
m-rgba: Hey @jwlee64 - updated with a callout about the OpenAI compatible endpoint and more clarity around the API secret stuff. Have a look and let me know what you think.

m-rgba:
Updated preview:
https://beta.wandb.ai/?betaVersion=9247c5b093dc72b1f083386763316195dc177fda

PR #3899, opened 2025-03-18
fix(weave_query): Handle null values with vectorized string split patterns

nicholaspun-wandb:
## Description

- Split from https://github.com/wandb/weave/pull/3900, additional fixes to the AWL-split op that I found while adding tests
- Raise an exception when the arrow data and pattern data have mismatching lengths (TODO: I need to add a test for this)
- Null values in the arrow data should evaluate to `None`

## Testing

Unit tests


nicholaspun-wandb:
Included in https://github.com/wandb/weave/pull/3942

PR #3902, opened 2025-03-18
chore(ui): Create TruncateByCharsWithTooltip HOC

connieelee:
## Description

https://wandb.atlassian.net/browse/WB-23679

Introduces `TruncateByCharsWithTooltip` higher order component, which

- Truncates text by max chars
- Supports start, middle, or end truncation
- Passes truncated text to children
- Automatically adds tooltip with full text if truncated

**Broader context:**

The end goal is to improve truncation behavior of our Tag/Pill/Alias common components. Currently, those comps only support end truncation after ~24 chars, and the tooltip behavior is inconsistent. For an upcoming feature, we need to display Alias tags with middle truncation after 16 chars. I wanted a unified/consistent way to update Tag/Pill/Alias and all their sub-variants to accept `maxChars` and `truncatedPart` props -- that's what `TruncateByCharsWithTooltip` gives us.

This PR just implements `TruncateByCharsWithTooltip` without using it anywhere yet. For ease of review, I split out the changes to Tag/Pill/Alias comps here: https://github.com/wandb/weave/pull/3904.

While Tag/Pill/Alias is our only use case at the moment, this is a generic component that could be used by any component that needs max char truncation + tooltip.

**Additional notes:**

- This is specifically useful if you need to truncate text **by char count**. In `core`, we have a [TruncateTextTooltip](https://github.com/wandb/core/blob/81858c42686166e59c5e080277c04d0a76c4ced1/frontends/app/src/components/common/TruncateText/TruncateTextTooltip.tsx#L50) component that supports start/middle/end truncation **by container width** (e.g. in runs table, truncate run names to fit resizable/dynamic column width). The implementation for that is much more complicated than just truncating by char, especially for middle truncation. If you know the exact char count you need to truncate to, then this component is likely a better fit.
- CSS has some built-in truncation capabilities, but only for start & end truncation. We also need middle truncation. It's also width based instead of by char count.

## Testing

- See unit tests
- See https://github.com/wandb/weave/pull/3904

## Summary by CodeRabbit

- **New Features**
	- Introduced a text truncation component that shortens long texts to a specified character limit and displays the full content via a tooltip on hover.
	- Offers configurable truncation options (start, middle, or end) along with flexible styling capabilities.
	- Enhanced the Tooltip component with a new optional property for improved flexibility.

- **Tests**
	- Added comprehensive tests to ensure reliable behavior for text shortening and tooltip display across various scenarios, including edge cases for truncation logic.


weave-js/src/components/TruncateByCharsWithTooltip/TruncateByCharsWithTooltip.tsx
```diff
@@ -0,0 +1,63 @@
...
+        noTriggerWrap
+        trigger={
+          <div role="button" className="w-fit cursor-auto">
+            {children({truncatedText})}
+          </div>
+        }>
+        {text}
+      </Tooltip>
+    </Wrapper>
+  );
```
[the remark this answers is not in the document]
connieelee: the `button` role is on purpose - it ensures the tooltip trigger is accessible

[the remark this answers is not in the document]
connieelee: our Tooltip component is built on radix, which has various accessibility features built in. currently `Tooltip` expects trigger to be a button already, so it sets `asChild={true}` on the trigger element. however, if the trigger element isn't already a button then we want to make sure to set `asChild={false}` and utilize radix's built-in accessibility features

weave-js/src/components/Tooltip.tsx
```diff
@@ -24,6 +24,9 @@ type TooltipProps = {
   /** Element to be rendered in-place where the popup is defined. */
   trigger: React.ReactNode;
 
+  /** Value for Tooltip.Trigger `asChild` prop */
+  isTriggerAsChild?: boolean;
```
connieelee, line 28: for accessibility, tooltip trigger should have button role and tabindex so it can receive focus. by default, radix will wrap your trigger element for you to include these accessibility features. however, if your tooltip trigger is already a button, then we would want to use the `asChild` prop.

it looks like this `Tooltip` component was set up to always set `asChild` to true. however, there are cases where we just want to use radix's built-in accessibility features, so it would be more appropriate to set `asChild` to false.

weave-js/src/components/TruncateByCharsWithTooltip/TruncateByCharsWithTooltip.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <Tooltip
+        trigger={
+          <div className="w-fit cursor-auto">{children({truncatedText})}</div>
+        }
+        isTriggerAsChild={false}
+        noTriggerWrap>
+        {text}
+      </Tooltip>
+    </Wrapper>
+  );
```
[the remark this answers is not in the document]
connieelee: radix handles this already

weave-js/src/components/TruncateByCharsWithTooltip/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Text may be truncated at the start, middle, or end (e.g.
+ * "…lo world", "hell…orld", or "hello wo…")
+ */
+export const truncateTextByChars = (
+  text: string,
+  maxChars: number,
+  truncatedPart: TruncatedPart
+): string => {
+  if (maxChars <= 0) {
+    return text;
```
connieelee, line 15: been going back and forth between returning full text or just "…" in this scenario. open to suggestion

weave-js/src/components/TruncateByCharsWithTooltip/TruncateByCharsWithTooltip.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  const truncatedText = truncateTextByChars(text, maxChars, truncatedPart);
+  return (
+    <Wrapper>
+      <Tooltip
+        trigger={
+          <div className="w-fit cursor-auto">{children({truncatedText})}</div>
+        }
+        isTriggerAsChild={false}
```
jo-fang, line 56: **suggestion (non-blocking):** Can we add comment on why this is false? I can see someone modifying comp to take in a `isTriggerAsChild` prop in the future without much thinking

weave-js/src/components/TruncateByCharsWithTooltip/utils.ts
jo-fang: I like that we're returning full text here, so callee (if i'm using this word correctly) can control how it should be displayed

jo-fang (approved):

weave-js/src/components/TruncateByCharsWithTooltip/utils.ts
connieelee: i kinda like the ellipsis to indicate that the text is fully truncated. in theory the caller/consumer has the full text anyway (that's where the input came from in the first place)

in reality this is really an edge case that probably doesn't matter all that much soooo i'll probably just leave it as is. it would be a simple change anyway if we do change our mind

PR #3903, opened 2025-03-18
chore(weave): weave client batch delete calls

gtarpenning:
## Description

[WB-23872](https://wandb.atlassian.net/browse/WB-23872)

Easily delete large batches of calls from the API with `delete_calls` by passing in a list of strings corresponding to call ids.

maximum delete size is 1000. 

deleting calls also deletes all children of that call. 

Example: 
```python
client = weave.init("my-project")

all_calls = client.get_calls()
first_1000_calls = all_calls[:1000]
first_1000_calls_ids = [c.id for c in first_1000_calls]

client.delete_calls(call_ids=first_1000_calls_ids)
# 💰
```

## Testing

Adds test

[WB-23872]: https://wandb.atlassian.net/browse/WB-23872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
	- Enhanced call record management now lets you delete multiple entries at once.
	- The update ensures that related records are properly handled and provides clear feedback for repeated or invalid deletion attempts.
- **Tests**
	- Added a new test function to verify the behavior of the deletion functionality under various scenarios.

### DIFF
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -606,7 +606,7 @@ def test_delete_calls(client):
     call1 = my_op(2)
     call2 = my_op(3)
 
-    calls = list(client.get_calls())
+    calls = client.get_calls()
     assert len(calls) == 3
 
     call_0_id = calls[0].id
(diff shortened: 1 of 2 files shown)


tests/trace/test_weave_client.py
```diff
@@ -597,6 +597,41 @@ async def score(target, model_output):
...
+def test_delete_calls(client):
+    @weave.op()
+    def my_op(a: int) -> int:
+        return a + 1
+
+    call0 = my_op(1)
+    call1 = my_op(2)
+    call2 = my_op(3)
+
+    calls = list(client.get_calls())
```
andrewtruong, line 609: you don't need to call list btw :)

tests/trace/test_weave_client.py
```diff
@@ -597,6 +597,41 @@ async def score(target, model_output):
...
+    assert len(calls) == 3
+
+    call_0_id = calls[0].id
+    call_1_id = calls[1].id
+    call_2_id = calls[2].id
+
+    client.delete_calls([call_0_id, call_1_id])
+    assert len(list(client.get_calls())) == 1
+    assert list(client.get_calls())[0].id == call_2_id
```
andrewtruong, line 618: nit: would have just called `get_calls()` once

tests/trace/test_weave_client.py
```diff
@@ -597,6 +597,41 @@ async def score(target, model_output):
...
+
+    client.delete_calls([call_0_id, call_1_id])
+    assert len(list(client.get_calls())) == 1
+    assert list(client.get_calls())[0].id == call_2_id
```
andrewtruong, line 622: is this testing idempotency?  I would just make a short comment above clarifying that

tests/trace/test_weave_client.py
```diff
@@ -597,6 +597,41 @@ async def score(target, model_output):
...
+
+    client.delete_calls([])
+    assert len(list(client.get_calls())) == 1
+    assert list(client.get_calls())[0].id == call_2_id
```
andrewtruong, line 626: same here

tests/trace/test_weave_client.py
gtarpenning: praise the lord

PR #3904, opened 2025-03-18
feat(ui): Improve Tag/Pill/Alias truncation

connieelee:
## Description

https://wandb.atlassian.net/browse/WB-23679

Stacked on https://github.com/wandb/weave/pull/3902

Currently Tag/Pill/Alias common comps only support end truncation after ~24 chars. For an upcoming feature, we need to display tags with middle truncation after 16 chars.

The current truncation strategy uses CSS, which works well for start/end truncation but not middle truncation. It also doesn't truncate to exact char counts (go to storybook and count the chars to see what I mean). For these reasons, we're switching to use the new `TruncateByCharsWithTooltip` component for unified & more robust truncation functionality.

- All tag/pill/alias comps with text now accept `maxChars` and `truncationPart` props. Defaults to 24 char end truncation to match existing behavior.
- Ensures we display tooltips if truncated (currently this only works for `RemovableTag` -- going forward we'll have truncation tooltips for for all tag/pill/alias variants)
- Currently, tag comps accept `Wrapper` prop, which is not defined for pill/alias comps. `TruncateByCharsWithTooltip` provides unified handling for this as well, so we can remove all the `Wrapper` logic from tag comps

## Testing

storybook

https://github.com/user-attachments/assets/0866a721-41fd-4968-b635-9103ba4292f7


weave-js/src/components/Tag/Alias.tsx
connieelee, line 1: recommend hide whitespace

jo-fang (approved):

PR #3910, opened 2025-03-19
merge

gtarpenning:
## Description

This pr: 
- creates a new clickhouse database function that queries for descendents using a recursive CTE query
- exposes endpoint on the client so that users can ask for descendents 

## Testing

adds test

### DIFF
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -2389,3 +2389,183 @@ def test_calls_query_sort_by_display_name_prioritized(client):
 
     # Verify they all have the same op_name
     assert call_list[0].op_name == call_list[1].op_name == call_list[2].op_name
+
+
+def test_calls_children_basic(client):
+    """Test basic functionality of calls_children including depth limits and multiple children."""
+
+    @weave.op()
+    def parent_op(x: int) -> int:
+        child1 = child_op1(x)
+        child2 = child_op2(x)
+        return child1 + child2
+
+    @weave.op()
+    def child_op1(x: int) -> int:
+        return grandchild_op(x)
+
+    @weave.op()
+    def child_op2(x: int) -> int:
+        return x * 3
+
+    @weave.op()
+    def grandchild_op(x: int) -> int:
+        return x * 2
+
+    # Create a call that will have multiple children and grandchildren
+    result = parent_op(5)
+    parent_call = client.get_calls(
+        query=tsi.CallsQueryReq(
+            project_id=client.project_id, filter=tsi.CallsFilter(op_names=["parent_op"])
+        )
+    ).next()
+
+    # Test getting immediate children only (depth=1)
+    immediate_children = list(
+        client.calls_children(
+            tsi.CallsChildrenReq(
+                project_id=client.project_id, call_ids=[parent_call.id], depth=1
+            )
+        )
+    )
+    assert len(immediate_children) == 2
+    assert {child.op_name for child in immediate_children} == {"child_op1", "child_op2"}
+
+    # Test getting all descendants (depth=2)
+    all_descendants = list(
+        client.calls_children(
+            tsi.CallsChildrenReq(
+                project_id=client.project_id, call_ids=[parent_call.id], depth=2
+            )
+        )
+    )
+    assert len(all_descendants) == 3
+    assert {child.op_name for child in all_descendants} == {
+        "child_op1",
+        "child_op2",
+        "grandchild_op",
+    }
+
+    # Test getting children for multiple parents
+    result2 = parent_op(10)
+    parent_calls = list(
+        client.get_calls(
+            query=tsi.CallsQueryReq(
+                project_id=client.project_id,
+                filter=tsi.CallsFilter(op_names=["parent_op"]),
+            )
+        )
+    )
+    multi_parent_children = list(
+        client.calls_children(
+            tsi.CallsChildrenReq(
+                project_id=client.project_id,
+                call_ids=[call.id for call in parent_calls],
+                depth=1,
+            )
+        )
+    )
+    assert len(multi_parent_children) == 4  # 2 children per parent
+
+
+def test_calls_children_edge_cases(client):
+    """Test edge cases for calls_children including deleted calls and leaf nodes."""
+
+    @weave.op()
+    def parent_op(x: int) -> int:
+        return child_op(x)
+
+    @weave.op()
+    def child_op(x: int) -> int:
+        return x * 2
+
+    @weave.op()
+    def leaf_op(x: int) -> int:
+        return x * 3
+
+    # Test leaf node (no children)
+    leaf_result = leaf_op(5)
+    leaf_call = client.get_calls(
+        query=tsi.CallsQueryReq(
+            project_id=client.project_id, filter=tsi.CallsFilter(op_names=["leaf_op"])
+        )
+    ).next()
+    leaf_children = list(
+        client.calls_children(
+            tsi.CallsChildrenReq(project_id=client.project_id, call_ids=[leaf_call.id])
+        )
+    )
+    assert len(leaf_children) == 0
+
+    # Test deleted parent
+    parent_result = parent_op(5)
+    parent_call = client.get_calls(
+        query=tsi.CallsQueryReq(
+            project_id=client.project_id, filter=tsi.CallsFilter(op_names=["parent_op"])
+        )
+    ).next()
+
+    # Delete parent and verify no children are returned
+    client.calls_delete(
+        tsi.CallsDeleteReq(project_id=client.project_id, call_ids=[parent_call.id])
+    )
+    deleted_parent_children = list(
+        client.calls_children(
+            tsi.CallsChildrenReq(
+                project_id=client.project_id, call_ids=[parent_call.id]
+            )
+        )
+    )
+    assert len(deleted_parent_children) == 0
+
+    # Test deleted child
+    parent_result2 = parent_op(10)
+    parent_call2 = client.get_calls(
+        query=tsi.CallsQueryReq(
+            project_id=client.project_id, filter=tsi.CallsFilter(op_names=["parent_op"])
+        )
+    ).next()
+    all_children = list(
+        client.calls_children(
+            tsi.CallsChildrenReq(
+                project_id=client.project_id, call_ids=[parent_call2.id]
+            )
+        )
+    )
+    assert len(all_children) == 1
+
+    # Delete child and verify it's no longer returned
+    client.calls_delete(
+        tsi.CallsDeleteReq(project_id=client.project_id, call_ids=[all_children[0].id])
+    )
+    remaining_children = list(
+        client.calls_children(
+            tsi.CallsChildrenReq(
+                project_id=client.project_id, call_ids=[parent_call2.id]
+            )
+        )
+    )
+    assert len(remaining_children) == 0
+
+    with pytest.raises(ValueError, match="Depth must be a positive integer"):
+        list(
+            client.calls_children(
+                tsi.CallsChildrenReq(
+                    project_id=client.project_id, call_ids=[parent_call2.id], depth=-1
+                )
+            )
+        )
+
+    with pytest.raises(
+        RequestTooLarge,
+        match="Cannot get more than 100000 children at once (requested: 100001).",
+    ):
+        list(
+            client.calls_children(
+                tsi.CallsChildrenReq(
+                    project_id=client.project_id,
+                    call_ids=[parent_call2.id],
+                    depth=100001,
+                )
+            )
+        )
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -129,14 +129,15 @@ from weave.trace_server.trace_server_interface_util import (
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
-MAX_FLUSH_COUNT = 10000
+MAX_FLUSH_COUNT = 10_000
 MAX_FLUSH_AGE = 15
 
-FILE_CHUNK_SIZE = 100000
+FILE_CHUNK_SIZE = 100_000
 
-MAX_DELETE_CALLS_COUNT = 1000
+MAX_DELETE_CALLS_COUNT = 1_000
 INITIAL_CALLS_STREAM_BATCH_SIZE = 50
 MAX_CALLS_STREAM_BATCH_SIZE = 500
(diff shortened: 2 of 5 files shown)


tests/trace/test_weave_client.py
```diff
@@ -2389,3 +2390,135 @@ def test_calls_query_sort_by_display_name_prioritized(client):
...
+
+    # Create a call that will have multiple children and grandchildren
+    result = parent_op(5)
+    parent_call = client.get_calls(filter=tsi.CallsFilter(op_names=["%parent_op:*"]))[0]
+
+    # Test getting immediate children only (depth=1)
+    immediate_children = client.call_descendants(parent_call, depth=1)
+    assert len(immediate_children) == 2
+    for child in immediate_children:
+        assert "child_op" in child.op_name
```
andrewtruong, line 2424: We have a util for this:
```py
from weave.integrations.integration_utilities import op_name_from_ref
```

tests/trace/test_weave_client.py
```diff
@@ -2389,3 +2390,135 @@ def test_calls_query_sort_by_display_name_prioritized(client):
...
+    def child_op2(x: int) -> int:
+        return x * 3
+
+    @weave.op()
+    def grandchild_op(x: int) -> int:
+        return x * 2
+
+    # Create a call that will have multiple children and grandchildren
+    result = parent_op(5)
+    parent_call = client.get_calls(filter=tsi.CallsFilter(op_names=["%parent_op:*"]))[0]
```
andrewtruong, line 2418: What does the `%` mean here?

weave/trace/weave_client.py
```diff
@@ -1325,6 +1391,36 @@ def fail_call(self, call: Call, exception: BaseException) -> None:
         """Fail a call with an exception. This is a convenience method for finish_call."""
         return self.finish_call(call, exception=exception)
 
+    @trace_sentry.global_trace_sentry.watch()
+    def call_descendants(
```
andrewtruong, line 1395: nit: `get_call_descendants`

weave/trace/weave_client.py
```diff
@@ -1325,6 +1391,36 @@ def fail_call(self, call: Call, exception: BaseException) -> None:
...
+        """Get all descendant calls of a given list of call IDs."""
+        res = self.server.calls_descendants(
+            CallsDescendantsReq(
+                project_id=self._project_id(),
+                call_ids=call_ids,
+                depth=depth,
+                limit=limit,
+            )
+        )
+        return list(res)
```
andrewtruong, line 1422: I would collapse into the previous function.  Instead, you can have it take flexible inputs and do the handling inside the func
```py
CallId = str
def call_descendants(self, call_ids: Call | CallId | Iterable[Call | CallId], ...): ...
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -456,6 +457,105 @@ def _expand_call_refs(
...
+            SELECT c.id, c.parent_id, d.depth + 1
+            FROM call_parts c
+            INNER JOIN descendants d ON c.parent_id = d.id
+            WHERE c.project_id = {{project_id:String}}
+                AND c.deleted_at IS NULL
+                {depth_condition}
+        )
+        SELECT DISTINCT id FROM descendants
+        LIMIT {{limit:Int32}}
+        SETTINGS allow_experimental_analyzer=1
```
andrewtruong, line 505: what does this do?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -456,6 +457,105 @@ def _expand_call_refs(
...
+        """
+        result = self._query(
+            query,
+            {
+                "project_id": project_id,
+                "call_ids": call_ids,
+                "depth": depth,
+                "limit": limit or MAX_CALLS_CHILDREN_LIMIT,
```
andrewtruong, line 513: I would put this `or` at the top to make it more obvious what the values will be

weave/trace/weave_client.py
```diff
@@ -1325,6 +1391,36 @@ def fail_call(self, call: Call, exception: BaseException) -> None:
...
+        """Get all descendant calls of a given call."""
+        res = self.server.calls_descendants(
+            CallsDescendantsReq(
+                project_id=self._project_id(),
+                call_ids=[call.id],
+                depth=depth,
+                limit=limit,
+            )
+        )
+        return list(res)
```
andrewtruong, line 1407: Why not be lazy?  I think it's better to let the user have control -- the iterator we return already has list-like access

weave/trace_server/trace_server_interface.py
```diff
@@ -251,6 +251,20 @@ class CallsDeleteRes(BaseModel):
     pass
 
 
+class CallsDescendantsReq(BaseModel):
+    project_id: str
+    call_ids: list[str]
+    limit: Optional[int] = None
+    # list of columns to select, defaults to all columns
```
andrewtruong, line 258: Instead of writing this as a comment, make it a field description.  This additionally gives us nice documentation when it's exported in the OpenAPI spec

weave/trace_server/trace_server_interface.py
```diff
@@ -251,6 +251,20 @@ class CallsDeleteRes(BaseModel):
...
+    columns: Optional[list[str]] = None
+
+    # If provided, the depth of the descendants to return.
+    # Default is all descendants. limit 1 would stop at grandchildren
```
andrewtruong, line 262: same here

tests/trace/test_weave_client.py
gtarpenning: this might actually be sqlinjection, lemme test something

neutralino1:
Thanks for this @gtarpenning.

Recurrent SQL queries seem complex and potentially inefficient. A few questions:

- Is this functionality actually necessary? Is this a customer request or just a nice-to-have?
- If this is indeed required, should we make adjustments to the data model to make these queries O(1)? There are ways to represent trees that make it easy to query all descendants of arbitrary nodes. See 👇 
Screenshot 2025-03-20 at 2 09 16 PM

gtarpenning:
> * Is this functionality actually necessary? Is this a customer request or just a nice-to-have?
> * If this is indeed required, should we make adjustments to the data model to make these queries O(1)? There are ways to represent trees that make it easy to query all descendants of arbitrary nodes. See 👇

@neutralino1 

1. It this functionality actually necessary? 
   - There have been users asking for this kind of thing in the past, but not exactly, nor right now. However, the immediate benefit is that it will enable much faster deletes. We still have my patch pr that is limiting deletion to checking 10,000 children. When deleting calls with tens of thousands of _siblings_, its possible that checking 10,000 isn't enough and we don't end up deleting all possible descendants. By using this recursive CTE algorithm, we get an order of magnitude better performance than what we currently have, which would allow us to remove the limit (or increase it 10x). 
   - It is actually pretty nice to find descendants from calls when using the python SDK. Often users want to iterate through a previously completed eval, for example, and conditionally grab the descendants based on criteria (like if the predict call didn't do what they expected). We currently expose a `call.children` function, but that leads to long recursive calls (only one level at a time) and the performance is shocking. 
   - We may use a function like this (your point is valid this is not exactly an end-game performant query) to improve the UI rendering as well. Tim made a lot of progress on the "trace tree" recently, but I think one of the reasons he mentioned this kind of function in standup was that we can wire it up to the trace tree rendering and make it even more performant. Currently we query the db for all calls with a common `trace_id`, which can include siblings, and then construct the trace tree in memory in the browser. At least that was how it worked until Tim's recent change.

2. I think this is a valid question, but perhaps more of a "yes _and_". My preference would be to merge a change like this and then consider improving the data model if there are performance concerns. It's very possible this endpoint gets very little use, in which case there would be no need to improve the query.

weave/trace/weave_client.py
```diff
@@ -1325,6 +1391,55 @@ def fail_call(self, call: Call, exception: BaseException) -> None:
         """Fail a call with an exception. This is a convenience method for finish_call."""
         return self.finish_call(call, exception=exception)
 
+    @trace_sentry.global_trace_sentry.watch()
+    def get_call_descendants(
```
gtarpenning, line 1395: should this be `get_calls_descendants`?

weave/trace/weave_client.py
```diff
@@ -1325,6 +1391,55 @@ def fail_call(self, call: Call, exception: BaseException) -> None:
...
+        self,
+        calls: Call | WeaveObject | CallId | Iterable[Call | WeaveObject | CallId],
```
andrewtruong, line 1397: The WeaveObject here is unfortunate :(

weave/trace/weave_client.py
```diff
@@ -1325,6 +1391,55 @@ def fail_call(self, call: Call, exception: BaseException) -> None:
...
+            limit: Maximum number of descendants to fetch
+
+        Returns:
+            An iterable of descendant calls
+        """
+        # Convert input to a list of items we can process
+        if isinstance(calls, (Call, WeaveObject, CallId)):
+            items_to_process = [calls]
+        elif isinstance(calls, Iterable):
+            items_to_process = list(calls)
```
andrewtruong, line 1415: I don't think you need to call `list` here

weave/trace/weave_client.py
```diff
@@ -1325,6 +1391,55 @@ def fail_call(self, call: Call, exception: BaseException) -> None:
...
+        else:
+            raise TypeError(
+                f"Invalid call, CallId, or Iterable: {calls}, expected Call, CallId (str), or Iterable[Call | CallId]"
+            )
+
+        call_ids: list[str] = []
+        for item in items_to_process:
+            if isinstance(item, (Call, WeaveObject)) and item.id is not None:
```
andrewtruong, line 1423: nit: Might be more informative to `ValueError` on `item.id is not None`

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -456,6 +457,108 @@ def _expand_call_refs(
                     val["_ref"] = ref.uri()
                 set_nested_key(calls[i], col, val)
 
+    def _get_call_descendent_ids(
```
andrewtruong, line 460: spelling: `descendant` not "ent"

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: Are there implications for non-clickhouse cloud users?  I don't think we have any now, but will we in future?

weave/trace_server/trace_server_interface.py
```diff
@@ -106,7 +106,7 @@ class CallSchema(BaseModel):
     # Exception is present if the call failed
     exception: Optional[str] = None
 
-    # Outputs
+    # Output
```
andrewtruong, line 109: What is the point of this comment?

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: clickhouse cloud is generally _stricter_ than other deployments, for example our dev setup does not require this extra setting

weave/trace_server/trace_server_interface.py
```diff
@@ -309,7 +337,7 @@ class CallsFilter(BaseModel):
 
 class SortBy(BaseModel):
     # Field should be a key of `CallSchema`. For dictionary fields
-    # (`attributes`, `inputs`, `outputs`, `summary`), the field can be
+    # (`attributes`, `inputs`, `output`, `summary`), the field can be
     # dot-separated.
     field: str  # Consider changing this to _FieldSelect
```
andrewtruong, line 342: I wont block here, but in general all comments for BaseModel should be Field descriptions :)

weave/trace_server/trace_server_interface.py
gtarpenning: oh, cursor likes `outputs` because it mirrors `inputs` but actually we only have one `output`. I just cleaned that up

andrewtruong (approved):
overall lgtm, please see comments and fix spelling issue!

weave/trace_server/trace_server_interface.py
gtarpenning: yeah, we should take another pass. i'll clean up my interface

weave/trace/weave_client.py
gtarpenning: `weave/trace/weave_client.py:1416: error: Incompatible types in assignment (expression has type "WeaveObject | Iterable[Call | WeaveObject | str]", variable has type "list[Call | WeaveObject | str]")  [assignment]`

gtarpenning: ill fix by defining items_to_process explicitly as an iterable

gtarpenning: nothing I do seems to work....

andrewtruong: seems like a type-checker mistake -- you already do an isinstance check for `WeaveObject` above

I won't block on this, it's probably fine

weave/trace/weave_client.py
```diff
@@ -874,6 +874,7 @@ def __repr__(self) -> str:
 
 
 BACKGROUND_PARALLELISM_MIX = 0.5
+CallId = str
```
andrewtruong, line 877: We might be able to make this tighter.  @zbirenbaum I think you are defining CallId in a more specific way now?

zbirenbaum: I think its best not to make it too tight. Call ID can be a UUID or a string of  a specific byte length, but the format might change as we review certain requirements such as performance. Keeping it a bit loose will make it easier to update later.

tests/trace/test_weave_client.py
```diff
@@ -2506,25 +2506,25 @@ def test_calls_descendants_edge_cases(client):
...
 
     with pytest.raises(ValueError, match="Limit must be a positive integer"):
-        list(client.get_call_descendants(parent_call2.id, limit=-1))
+        list(client.get_calls_descendants(parent_call2.id, limit=-1))
 
     with pytest.raises(
         RequestTooLarge, match="Cannot get more than 100000 children at once"
     ):
-        list(client.get_call_descendants(parent_call2.id, limit=100_001))
+        list(client.get_calls_descendants(parent_call2.id, limit=100_001))
```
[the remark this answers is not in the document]
gtarpenning: yeah, but then I will have to implement that validation in both the client and in the server....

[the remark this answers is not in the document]
gtarpenning: i'll make a todo

gtarpenning:
closing in favor of: https://github.com/wandb/weave/pull/5163

PR #3912, opened 2025-03-19
chore(weave): Remove graph view from trace view

tssweeney:
Graph view deps caused crash in prod for artifacts. need to align on versions


estellazx (approved):
this should fix artifact lineage tab from crashing

jwlee64 (approved):

PR #3916, opened 2025-03-20
fix(weave): Fixing filtering calls table by boolean

neutralino1:
## Description

- Fixes WB-23754

We use a `LIKE` optimization for "heavy" conditions. The `LIKE` condition template wraps the literal value in double quotes, which is good for strings, but boolean literals are not wrapped in quotes in JSON payloads.

One corner case of this fix is that when filtering by strings `"true"` or `"false"` (not the boolean), the optimization may also match any call with strings containing those substrings (e.g. `"trued"`). This is not a concern as the final filtering would remove those (tested). It just makes the optimization slightly less optimal in those rare corner cases.

## Testing

How was this PR tested?

- tested manually with boolean values and string values `"true"`.
- modified the `test_query_heavy_column_simple_filter_with_order_and_limit_and_mixed_query_conditions` unit test to cover the boolean case.


neutralino1:

gtarpenning (approved):
Ah great catch. Nice fix. I wonder if there are any other types I missed...

PR #3919, opened 2025-03-20
Set the WEAVE_TRACE_LANGCHAIN to control tracing in LangChain

chance-wnb:
## Changes
- Added environment variable control for LangChain tracing through `WEAVE_TRACE_LANGCHAIN`
- The environment variable is automatically set when `set_tracing_enabled()` is called
- The original value is preserved and restored when the context manager exits
- Added proper cleanup in case the environment variable didn't exist before

## Why
This change addresses a critical issue with LangChain tracing control. Currently, even when using `with set_tracing_enabled(False):`, LangChain traces will still be generated unless the `WEAVE_TRACE_LANGCHAIN` environment variable is explicitly set to "false". This creates a confusing situation where the tracing state isn't fully controlled by our context manager.

I considered two alternative approaches:

1. Adding a `weave_tracing_disabled()` function in the LangChain integration that would directly manipulate the `weave_tracing_callback_var`. However, this would require users to make additional changes on their side to make it work, which isn't ideal for user experience.

2. Directly operating on the `weave_tracing_callback_var` from within `set_tracing_enabled()`. While this would work, it would introduce circular dependency issues. While these could be solved, the resulting code would be more complex than the current approach.

The chosen solution of using an environment variable:
- Maintains clean separation of concerns
- Avoids circular dependencies
- Provides a simple, standard way to control LangChain tracing
- Requires no additional user-side changes

- Fixes WB-22810

## Testing

TBD ( I tested manually by applying the change to the file in my local pyenv, but I will do more thorough testing shortly)


weave/trace/context/call_context.py
```diff
@@ -147,11 +148,21 @@ def get_tracing_enabled() -> bool:
 @contextlib.contextmanager
 def set_tracing_enabled(enabled: bool) -> Iterator[None]:
     token = _tracing_enabled.set(enabled)
+
+    original_env_var = os.environ.get("WEAVE_TRACE_LANGCHAIN")
```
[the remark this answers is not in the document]
chance-wnb: yes, your concerns makes sense. Without doing a surgery in such a central piece, we can do the alternative 1 mentioned above. But that adds some extra work on the user side to use that explicitly and separately. So it is not completely ideal.

chance-wnb:

weave/trace/serialization/op_type.py
```diff
@@ -403,6 +403,9 @@ def _get_code_deps(
                         f"Cross-module op dependencies are not yet serializable {var_value}"
                     )
                 else:
+                    # This is to work around mypy not being able to infer that var_value is a FunctionType or type
+                    var_value = cast(Union[py_types.FunctionType, type], var_value)
```
tssweeney, line 407: unrelated?

weave/trace/op.py
```diff
@@ -296,12 +298,12 @@ def _create_call(
     )
 
 
-def _should_skip_tracing(op: Op) -> bool:
+def should_skip_tracing(op: Op | str) -> bool:
```
tssweeney, line 301: it seems odd to me to have this function accept a str, but not actually use the string-like value. This smells like the function's params are being modified to accommodate the caller in weave_client

weave/trace/op.py
```diff
@@ -331,6 +333,15 @@ def _placeholder_call() -> Call:
     )
 
 
+def is_placeholder_call(call: Call) -> bool:
```
tssweeney, line 336: if you are going to formalize this, I think it would be better to use a subclass as opposed to multiple value checks

tssweeney (commented):
Another round of adjustments (:

weave/trace/weave_client.py
```diff
@@ -1223,6 +1235,9 @@ def finish_call(
         *,
         op: Op | None = None,
     ) -> None:
+        if (op is not None and should_skip_tracing(op)) or is_placeholder_call(call):
```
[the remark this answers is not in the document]
chance-wnb: Agreed. Breaking up might make things cleaner.

weave/trace/serialization/op_type.py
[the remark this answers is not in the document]
chance-wnb: It thinks that var_value would be FunctionType, type or **Op**. so it looks the the is_op() TypeGuard doesn't function as expected (only here). 

Anyway, the change is now withdrawn.

weave/trace/op.py
chance-wnb: Yeah, I did try to accommodate the call signatures to cut corners. Now it's refactored out.

weave/trace/weave_client.py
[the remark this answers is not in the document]
chance-wnb: I tried to break up the function little bit, but now I feel it might a bit tedious. Please let me know if this matches what you envisioned.

tests/trace/test_weave_client.py
```diff
@@ -2389,3 +2390,26 @@ def test_calls_query_sort_by_display_name_prioritized(client):
...
+    @weave.op()
+    def test_op():
+        return "test"
+
+    # Test create_call with tracing enabled
+    call = await client.create_call(test_op, {})
+    assert isinstance(call, Call)
+    assert call._op_name == "test_op"  # Use string literal instead of __name__
+
+    # Test create_call with tracing disabled
```
tssweeney, line 2408: i would also assert the result here. something like `len(list(client.get_calls())) == 1` in the first case, and then also is 1 after disabled

chance-wnb: Thats a great feedback. Thanks! It is incorporated now.

tssweeney (approved):

PR #3922, opened 2025-03-20
fix(weave): handle noncompliant chunk response from Azure

jamie-rasmussen:
## Description

Internal Slack: https://weightsandbiases.slack.com/archives/C053FHBH73P/p1740558111134829?thread_ts=1740124248.544159&cid=C053FHBH73P

Per the OpenAI SDK, the `delta` field of a `Choice` is not `Optional`.
https://github.com/openai/openai-python/blob/3e69750d47df4f0759d4a28ddc68e4b38756d9ca/src/openai/types/chat/chat_completion_chunk.py#L88

However, the Azure implementation can return a `None` value for the delta. This issue notes one triggering condition which I was able to reproduce, which is when you have enabled custom content filtering settings with the Asynchronous_filter streaming setting.
https://github.com/openai/openai-python/issues/1677

A similar previous PR: https://github.com/wandb/weave/pull/3736

### DIFF
--- a/weave/integrations/openai/openai_sdk.py
+++ b/weave/integrations/openai/openai_sdk.py
@@ -169,6 +169,13 @@ def openai_accumulator(
             if chunk_choice.logprobs:
                 choice["logprobs"] = chunk_choice.logprobs
 
+            # See https://github.com/openai/openai-python/issues/1677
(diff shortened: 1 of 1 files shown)


neutralino1 (approved):
Thank you.

PR #3927, opened 2025-03-21
chore(ui): Header update (visuals only)

m-rgba:
## Description

| Before | After |
| -------- | ------- |
| CleanShot 2025-03-22 at 19 09 21 | CleanShot 2025-03-22 at 19 11 03 |

- Visual changes for trace / eval page header update.
- Emphasizes actions which can be taken after row selection.
  - Adds a bar which calls out how many traces are selected
- Updated default button styles to reduce line-work.
- Updated Ops selector to use the proper Source Sans Pro font.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -1105,9 +1105,24 @@ const OpSelector = ({
         <FormControl fullWidth sx={{borderColor: MOON_200}}>
           <Autocomplete
             PaperComponent={paperProps => <StyledPaper {...paperProps} />}
+            ListboxProps={{
+              sx: {
+                fontSize: '14px',
+                fontFamily: 'Source Sans Pro',
+                '& .MuiAutocomplete-option': {
+                  fontSize: '14px',
+                  fontFamily: 'Source Sans Pro',
+                },
+                '& .MuiAutocomplete-groupLabel': {
+                  fontSize: '14px',
+                  fontFamily: 'Source Sans Pro',
+                },
+              },
+            }}
             sx={{
               '& .MuiOutlinedInput-root': {
                 height: '32px',
+                fontFamily: 'Source Sans Pro',
                 '& fieldset': {
                   borderColor: MOON_200,
                 },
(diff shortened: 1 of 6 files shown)


m-rgba:
Updated:
https://beta.wandb.ai/?betaVersion=5a3572f9f51899415724cf84dc543ff6c179e665

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -754,168 +754,178 @@ export const CallsTable: FC<{
...
-                    .filter(row => selectedCalls.includes(row.id))
-                    .map(traceCallToUICallSchema)}
-                  confirmDelete={deleteConfirmModalOpen}
-                  setConfirmDelete={setDeleteConfirmModalOpen}
-                  onDeleteCallback={() => {
-                    setSelectedCalls([]);
-                  }}
+              )}
+              {filterModel && setFilterModel && (
+                <FilterPanel
```
gtarpenning, line 782: technically you can select calls, and then open the filter window to and click "filter selected calls" so this might be losing functionality. not sure who is actually doing that though.

gtarpenning: it is basically the only way to filter to a specific call

gtarpenning (approved):
We should make sure we don't remove functionality from users, so there should be *some* way to filter by selected calls. otherwise lgtm

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -754,168 +754,178 @@ export const CallsTable: FC<{
...
+                  selectedCalls={selectedCalls}
+                  clearSelectedCalls={clearSelectedCalls}
                 />
+              )}
+            </>
+          ) : (
+            <div className="flex items-center gap-8">
+              <Button
+                variant="ghost"
+                size="small"
```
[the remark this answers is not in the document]
m-rgba: Actually went small on the close button since we're using small text for the trace / eval count.
Also helps with the padding since it's on the furthest-left.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
m-rgba: So I think we can add something like a "Filter selected traces/evals" button the trace bar (maybe in a more dropdown).
The way it's currently in the filter-area isn't being used -- taking a look at the analytics of non-Gmail / non-W&B users, only one user in one session used the button.

PR #3930, opened 2025-03-21
fix(weave_query): ensure fallback for color void node is reached

domphan-wandb:
## Description

Fixes [WB-23789](https://wandb.atlassian.net/browse/WB-23789)

This PR fixes a regression from the feature (https://github.com/wandb/weave/pull/3700) to add custom run names.

The fallback to use the labelSelectFn as the node for the color dimension was never reached due to an early return.

## Testing
Before:

image

After:

image

Validation for custom run names:
image

How was this PR tested?
visually through invoker fe prod

### DIFF
--- a/weave-js/src/components/Panel2/PanelPlot/util.ts
+++ b/weave-js/src/components/Panel2/PanelPlot/util.ts
@@ -117,22 +117,28 @@ export const useVegaReadyTables = (series: SeriesConfig[], frame: Frame) => {
       const dims = allDims[i];
       const labelSelectFn = table.columnSelectFunctions[dims.label];
       const colorSelectFn = table.columnSelectFunctions[dims.color];
-      if (labelSelectFn.nodeType !== 'void') {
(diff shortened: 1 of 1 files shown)


weave-js/src/components/Panel2/PanelPlot/util.ts
```diff
@@ -117,22 +117,28 @@ export const useVegaReadyTables = (series: SeriesConfig[], frame: Frame) => {
...
-          if (isAssignableTo(labelType, maybe('run'))) {
-            runNode = labelSelectFn;
-          } else if (
-            labelSelectFn.nodeType === 'output' &&
-            labelSelectFn.fromOp.name === 'run-name'
-          ) {
-            runNode = labelSelectFn.fromOp.inputs.run;
-          } else {
-            return table;
-          }
```
nicholaspun-wandb, line 133: 🤦 Ah, good catch

nicholaspun-wandb (approved):

PR #3934, opened 2025-03-21
fix(weave): Anthropic message lists containing tool results do not render as a chat

neutralino1:
## Description

- Fixes WB-23912

When a `tool_result` message is present in a payload, it is not recognized as a chat payload. This PR fixes that and makes sure tool calls and results are properly displayed in the chat interface.

## Testing

- Tested with JSON and plain Python tool results.

**Display for plain Python objects.**
Note that the tool's response is not displayed in the same block as the tool call as expected. This is because the Anthropic tool result API schema is quite different from OpenAI's. Making both compliant would require significant refactoring that is beyond the scope of this PR. The solution below is suboptimal, but an acceptable tradeoff.
Screenshot 2025-03-25 at 9 35 04 PM

**Display for JSON objects.** (ignore old function display)
Screenshot 2025-03-25 at 8 59 15 AM


jwlee64:
we might want to make the tool calls use our ToolCalls.tsx component for standardized ui

Screenshot 2025-03-25 at 1 50 00 PM

neutralino1:
@jwlee64 thanks for this, did not know this existed. This looks much better, although I'll have to see how it looks within the chat bubble.

jwlee64 (approved):
awesome

PR #3939, opened 2025-03-24
perf(weave): download dataset with images in rows in parallel

gtarpenning:
## Description

[WB-23955](https://wandb.atlassian.net/browse/WB-23955)

Use the future executor to parallelize row requests. The only issue I can see is when the queue is full, but I think in that case this executes exactly as fast as it currently does. 

This pr: 
- utilizes the async_batch_processor to handle row download requests, massively speeding up row downloads when rows have objects in them (like images). 

## Testing

| Before | After |
| -------- | ------- |
| `29.7s` | `2.24s` | 
| Screenshot 2025-03-24 at 10 13 51 AM | Screenshot 2025-03-24 at 10 09 59 AM | 

Image dataset test (run eval against downloaded dataset):
| size | prod eval time (s) | branch  eval time (s) | 
| -------- | ------- | ------- |
| 10 | `1.92s` | `0.66s` |
| 100 | `14.66s` | `0.96s` |
| 1000 | `175.16s` | `20.16s` |

[WB-23955]: https://wandb.atlassian.net/browse/WB-23955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace/vals.py
```diff
@@ -439,6 +439,12 @@ def _remote_iter(self) -> Generator[dict, None, None]:
         if self.table_ref is None:
             return
 
+        wc = get_weave_client()
+        if not wc:
+            raise ValueError(
```
gtarpenning, line 444: can this ever happen?

weave/trace/vals.py
```diff
@@ -487,9 +495,25 @@ def _remote_iter(self) -> Generator[dict, None, None]:
...
+                    )
+
+                future = wc.future_executor.defer(
+                    lambda v=val, r=new_ref: process_row(v, r)
+                )
+                futures.append(future)
+
+            # Yield results as they complete
+            for future in futures:
+                yield future.result()
```
gtarpenning, line 516: very possible this messes up the ordering

gtarpenning: apparently not

weave/trace/vals.py
andrewtruong, line 1: can you share benchmarks for 1^(0..6) rows so we can get a sense of the magnitude of impact?

weave/trace/vals.py
```diff
@@ -487,9 +494,25 @@ def _remote_iter(self) -> Generator[dict, None, None]:
...
```
andrewtruong, line 515: Just want to make sure this is expected:

As written, if you raise on line 500, no further futures will be returned.

gtarpenning: I think in practice table_ref should be defined here. i'll move that check outside the process row fn.

gtarpenning: oh, there already is that check at the top. ill make this less aggro but just returning none?

gtarpenning: hmm. I think that is what we want right? During an eval, if anything fails I believe we fail the entire thing. Additionally, all i'm doing here is a thin wrapper over the existing functionality that downloaded in the main thread, which definitely would have errored. So I think this is safe. If we in the future decided to allow partial-evals, or "resuming" evals, or fixing errored traces in evals, we might have to rethink this. Unless you can see an obvious alternative?

andrewtruong: ok, just want to make sure you thought about it.  That makes sense to me :)

andrewtruong (approved):

PR #3905, opened 2025-03-19
chore(weave_query): Handle DictionaryArray inputs for various awl string ops

nicholaspun-wandb:
JIRA: https://wandb.atlassian.net/browse/WB-23892

## Description

Follow-up to https://github.com/wandb/weave/pull/3900: Most of the other string ops don't re-encode `DictionaryArray`s. This PR abstracts the process in a helper and modifies the following ArrowWeaveList string ops:
* Accepts a unary input type (i.e. the input the string itself with argument name `self`)
    * arrowweavelist_len (op: "ArrowWeaveListString-len")
    * isalpha (op: "ArrowWeaveListString-isAlpha")
    * isnumeric (op: "ArrowWeaveListString-isNumeric")
    * isalnum (op: "ArrowWeaveListString-isAlnum")
    * lower (op: "ArrowWeaveListString-lower")
    * upper (op: "ArrowWeaveListString-upper")
    * strip (op: "ArrowWeaveListString-strip")
    * lstrip (op: "ArrowWeaveListString-lStrip")
    * rstrip (op: "ArrowWeaveListString-rStrip")
* Additional one-liners
    * slice (op: "ArrowWeaveListString-slice")
    * replace (op: "ArrowWeaveListString-replace")

The following ops still need to be modified and I'll follow up in a separate PR:
* \_\_eq\_\_
* \_\_ne\_\_
* \_\_contains\_\_
* in\_\_
* string_add
* string\_right\_add
* append
* prepend
* split (only partially corrected from https://github.com/wandb/weave/pull/3900)
* partition
* startswith
* endswith
* join\_to\_str
* to_number 

## Testing

Unit tests

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive tests for string validations, including checks for letter-only, numeric, alphanumeric, and case conversion behaviors.

- **Refactor**
  - Improved the underlying string processing approach for better consistency and maintainability, enhancing how specific data structures are handled.

PR #3942, opened 2025-03-24
chore(weave_query): Handle DictionaryArray inputs for another subset of AWL string ops

nicholaspun-wandb:
## Description
JIRA: https://wandb.atlassian.net/browse/WB-23892

Follow-up to https://github.com/wandb/weave/pull/3905. This PR fixes the following ops:
* split
* partition
* startswith
* endswith
* join\_to\_str
* to_number 

A couple things to note:
* In several instances, we weren't handling null value cases (either for `arrow_data` or for the input string argument to the op). These are fixed and included in the tests.
* The `toNumber` op does not completely work correctly as floating point numbers will not be converted (e.g. 123.45) due to `utf8_is_numeric` ([docs](https://arrow.apache.org/docs/python/generated/pyarrow.compute.utf8_is_numeric.html)) considering that string to not be numeric. I tried fixing this for a bit, but it looks like most workarounds are pretty hacky. Since the focus of this PR is updating the ops to handle `DictionaryArray`s, I've opted to keep this behaviour in for this PR.

The following ops still need to be modified:
* \_\_eq\_\_
* \_\_ne\_\_
* \_\_contains\_\_
* in\_\_
* string_add
* string\_right\_add
* append
* prepend

## Testing

Unit Tests

### DIFF
--- a/weave_query/weave_query/ops_arrow/string.py
+++ b/weave_query/weave_query/ops_arrow/string.py
@@ -206,30 +206,19 @@ def prepend(self, other):
     output_type=ARROW_WEAVE_LIST_LIST_OF_STR_TYPE,
 )
 def split(self, pattern):
-    if isinstance(pattern, str):
-        if isinstance(self._arrow_data, pa.DictionaryArray):
-            # If we have a dictionary array, we need to split the dictionary and
-            # then re-encode it as a dictionary array.
-            return ArrowWeaveList(
-                pa.DictionaryArray.from_arrays(
-                    self._arrow_data.indices,
-                    pc.split_pattern(self._arrow_data.dictionary, pattern),
-                ),
-                types.optional(types.List(types.String())),
-                self._artifact,
+    def _split(arrow_data):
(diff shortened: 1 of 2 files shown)


weave_query/weave_query/ops_arrow/string.py
```diff
@@ -290,18 +296,18 @@ def startswith(self, prefix):
...
-    )
+    def _ends_with(arrow_data):
+        if isinstance(suffix, str):
+            return pc.ends_with(arrow_data, suffix)
+        else:
+            return pa.array(
+                s.as_py().endswith(p.as_py())
+                if s.as_py() is not None and p.as_py() is not None
+                else None
+                for s, p in zip(self._arrow_data, suffix._arrow_data)
```
domphan-wandb, line 307: nit: _ends_with has the parameter of `arrow_data` but the logic uses `self._arrow_data`. I think you can just pass in `arrow_data`

domphan-wandb (approved):

PR #3944, opened 2025-03-24
style(ui): Update Select component styles

connieelee:
## Description

Pre-req for https://wandb.atlassian.net/browse/WB-23683

We're building a new UI that displays a multi-select field next to a plain text field: https://www.figma.com/design/8v53a9ldK9DcVUkILt4CIn/Scratchpad?node-id=3102-175949&m=dev

But there are some styling discrepancies between our `TextField` and `Select` comps that are especially prominent when the two are displayed right next to each other.

Screenshot 2025-03-24 at 1 47 31 PM

- height: multi-select is taller than TextField (40px vs 32px)
- placeholder text: multi-select is smaller than TextField (14px vs 16px)
- border-radius: multi-select is rounder than TextField (4px vs 2px)
  - ...these are actually both wrong, it should be 5px [per design specs](https://www.figma.com/design/01KWBdMZg5QM9SRS1pQq0z/Design-System----Robot-Styles?node-id=16374-60224&t=ARJuNYDjkamP6I9W-0)
- "border": multi-select uses `box-shadow` to simulate border while TextField uses `outline`. this results in the borders to behave _ever so slightly_ differently:

https://github.com/user-attachments/assets/b0c6f9cf-a887-49d2-aa47-e7a86b343ae5

There are also some styling issues with multi-select that are totally unrelated to TextField:

Screenshot 2025-03-24 at 1 52 19 PM

- dropdown indicator (up/down chevron) should be hidden (see [designs](https://www.figma.com/design/8v53a9ldK9DcVUkILt4CIn/Scratchpad?node-id=3222-159265&m=dev))
- clear indicator should use our "close" icon
- selected items should should have 2px gaps

This PR addresses all of the issues listed above. For the border problem, I've converted Select to use outline like TextField does. Unfortunately, neither Select nor TextField actually fully match the [design specs in figma](https://www.figma.com/design/01KWBdMZg5QM9SRS1pQq0z/Design-System----Robot-Styles?node-id=16864-85003&t=UNBecuidxuPpDLgT-0). I think there must've been some changes to the designs since we originally implemented these components. That's probably something to be addressed separately (prob as part of the common comps epic)

## Testing

Tested in core app; things generally look right/good (will link associated core PR once that's available)

https://github.com/user-attachments/assets/98962305-966e-437d-a140-cba1eff23996

### DIFF
--- a/weave-js/src/components/Form/Select.tsx
+++ b/weave-js/src/components/Form/Select.tsx
@@ -243,7 +243,7 @@ const getStyles = <
         fontSize,
         cursor: props.cursor ?? 'default',
         border: 0,
-        borderRadius: '3px',
+        borderRadius: '5px',
         outline:
           state.menuIsOpen || state.isFocused
             ? `2px solid ${colorBorderOpen}`
--- a/weave-js/src/components/Form/Select.tsx
+++ b/weave-js/src/components/Form/Select.tsx
@@ -23,6 +23,7 @@ import {
 import {Icon} from '@wandb/weave/components/Icon';
 import React from 'react';
 import ReactSelect, {
+  ClearIndicatorProps,
   components,
   DropdownIndicatorProps,
   GroupBase,
@@ -54,7 +55,7 @@ const MIN_HEIGHTS: Record<SelectSize, number | undefined> = {
   small: undefined,
   medium: undefined,
   large: undefined,
-  variable: 40,
+  variable: 32,
 } as const;
 
 const LINE_HEIGHTS: Record<SelectSize, string | undefined> = {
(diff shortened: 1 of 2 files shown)


weave-js/src/components/Form/Select.tsx
```diff
@@ -107,6 +128,9 @@ const DropdownIndicator = <
 >(
   indicatorProps: DropdownIndicatorProps<Option, IsMulti, Group>
 ) => {
+  if (indicatorProps.isMulti) {
+    return null;
+  }
```
connieelee, line 133: hides dropdown indicator (chevron icon) for multi-select

weave-js/src/components/Form/Select.tsx
```diff
@@ -182,7 +206,7 @@ const getStyles = <
     },
     valueContainer: baseStyles => {
       const padding = PADDING[size];
-      return {...baseStyles, padding};
+      return {...baseStyles, padding, gap: '2px'};
```
connieelee, line 209: adds 2px gap between selected items

Screenshot 2025-03-24 at 2 39 16 PM

PR #3945, opened 2025-03-24
fix(weave): add a null check to prevent playground crash

jwlee64:
## Description
https://wandb.atlassian.net/browse/WB-23842

the link in this ticket now loads instead of a full page crash
important for failed calls that might not have output

## Testing

How was this PR tested?


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/usePlaygroundState.ts
```diff
@@ -219,7 +219,10 @@ export const parseTraceCall = (traceCall: OptionalTraceCallSchema) => {
 
   // Handles anthropic outputs
   // Anthropic has content and stop_reason as top-level fields
-  if (isAnthropicCompletionFormat(parsedTraceCall.output)) {
+  if (
+    isAnthropicCompletionFormat(parsedTraceCall.output) &&
```
jamie-rasmussen, line 223: I think reversing the order of these clauses would be clearer.

jamie-rasmussen (approved):

PR #3946, opened 2025-03-24
chore(weave): allow otel style ids for calls

zbirenbaum:
## Description

- Fixes WB-23975

Modify regex matching on frontend to match any ID with or without dashes to support OTEL style ID strings. Change validation of call IDs and trace IDs to include OTEL style ID strings.

### DIFF
--- a/weave-js/src/react.tsx
+++ b/weave-js/src/react.tsx
@@ -544,7 +544,7 @@ const RE_WEAVE_CALL_REF_PATHNAME = new RegExp(
     '/',
     PATTERN_PROJECT,
     '/call/',
-    '([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})', // Call UUID
+    '([0-9a-fA-F-]+)', // Call ID
     '/?', // Ref extra portion is optional
     PATTERN_REF_EXTRA, // Optional ref extra
     '$', // End of the string
--- a/weave-js/src/react.tsx
+++ b/weave-js/src/react.tsx
@@ -544,7 +544,7 @@ const RE_WEAVE_CALL_REF_PATHNAME = new RegExp(
     '/',
     PATTERN_PROJECT,
     '/call/',
-    '([0-9a-fA-F-]+)', // Call ID
+    '([0-9a-fA-F-]+)', // Match any ID with or without dashes to support OTEL style ID strings
     '/?', // Ref extra portion is optional
     PATTERN_REF_EXTRA, // Optional ref extra
     '$', // End of the string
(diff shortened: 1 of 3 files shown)


weave/trace_server/validation.py
```diff
@@ -11,11 +11,17 @@ def project_id_validator(s: str) -> str:
 
 
 def call_id_validator(s: str) -> str:
-    return validation_util.require_uuid(s)
+    try:
+        return validation_util.require_otel_span_id(s)
+    except:
```
andrewtruong, line 16: Need more explicit `except` here.  What are you catching?

weave-js/src/react.tsx
```diff
@@ -544,7 +544,7 @@ const RE_WEAVE_CALL_REF_PATHNAME = new RegExp(
     '/',
     PATTERN_PROJECT,
     '/call/',
-    '([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})', // Call UUID
+    '([0-9a-fA-F-]+)', // Call ID
```
andrewtruong, line 547: I would add a comment here with your comment from the PR about this being for OTEL support

weave/trace_server/validation_util.py
```diff
@@ -9,6 +9,28 @@ class CHValidationError(Exception):
...
 
 
+def require_otel_trace_id(s: str) -> str:
+    lower_s = s.lower()
+    if lower_s != s:
+        raise CHValidationError(f"Invalid Trace ID: {s}. Trace ID must be lowercase")
+
+    # 16 Bytes is a hex string of len 32
+    if len(s) != 32:
+        raise CHValidationError(f"Invalid Trace ID: {s}. Trace ID must be 8 bytes")
```
andrewtruong, line 19: 8 or 16 bytes?

weave-js/src/react.tsx
zbirenbaum: Done

weave/trace_server/validation.py
zbirenbaum: Logically this is similar to `require_otel_span_id(s) or require_uuid(s)`. The exceptions raised in the `require` methods cause execution to halt so this just checks if it is an OTEL UUID before allowing the `require_uuid` to throw.

I agree its a bit messy, I guess I could modify the original `require_uuid` to also check for otel format IDs if that's better, or return either a string or error from the requires and throw if it's an instance of error

andrewtruong: If you want to be super general then you can do `except Exception` which will catch most things.  Bare except is dangerous because it will also catch stuff like KeyboardInterrupt which you probably dont want

andrewtruong (approved):

PR #3971, opened 2025-03-25
chore(weave): 4 chats in playground

jwlee64:
## Description

adds a min width and x-scroll
allows up to 4 chats

Screenshot 2025-03-25 at 1 10 30 PM

## Testing

How was this PR tested?


gtarpenning (commented):
thunderdome????

m-rgba:
Just a couple small tweaks:
- Could we add margin-left 8px to the clear chat button (gives some space between the button and the model selector)?
- Could we make the min-width 620px (makes it not collapse the padding on the right)?

Either than that, looks slick!

gtarpenning:
The settings padding is a bit scuffed with 2 chats @jwlee64 
Screenshot 2025-04-07 at 10 23 02 AM

gtarpenning:
it does sort of just get worse with more chats, more header stuff gets obscured
Screenshot 2025-04-07 at 10 23 42 AM

gtarpenning:
right margin too maybe
Screenshot 2025-04-07 at 10 24 59 AM

gtarpenning:
Not sure we need so much margin on the right of the id too 
Screenshot 2025-04-07 at 10 25 22 AM

jwlee64:
I updated the beta link
@gtarpenning I believe what your observing is the intended behavior, if there are too many chats we crowd, otherwise you will just need to expand the view

PR #3973, opened 2025-03-25
chore(weave): warn user when weave is being used without `weave.init`

zbirenbaum:
## Description

- Fixes WB-23983

Provide a warning message to the user informing them that their traces will not be logged when weave.init has not been called.

### DIFF
--- a/weave/trace/op.py
+++ b/weave/trace/op.py
@@ -89,6 +89,7 @@ logger = logging.getLogger(__name__)
 CALL_CREATE_MSG = "Error creating call:\n{}"
 ASYNC_CALL_CREATE_MSG = "Error creating async call:\n{}"
 ON_OUTPUT_MSG = "Error capturing call output:\n{}"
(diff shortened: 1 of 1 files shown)


weave/trace/op.py
```diff
@@ -89,6 +89,7 @@ async def anext(obj: AsyncIterator[V], default: Optional[V] = None) -> V:  # noq
 CALL_CREATE_MSG = "Error creating call:\n{}"
 ASYNC_CALL_CREATE_MSG = "Error creating async call:\n{}"
 ON_OUTPUT_MSG = "Error capturing call output:\n{}"
+UNINITIALIZED_MSG = "Warning: Traces will not be logged. Call weave.init to log your traces to a project.\n"
```
andrewtruong, line 92: nit: remove `Warning: `

it will be included by default when you do `logger.warn`

andrewtruong (approved):

PR #3988, opened 2025-03-27
chore(ui): Fix nested buttons in removable tag/alias

connieelee:
## Description

`RemovableTag` and `RemovableAlias` with truncated texts cause this console error:

https://github.com/user-attachments/assets/19d82512-9b7c-4278-a260-379839fb43aa

This is because we wrap the tag/alias in a tooltip when truncated. Tooltip triggers get `button` role for accessibility. The remove icon button is included as part of the tooltip trigger, which is the problem. This change moves the remove button outside of the truncation wrapper.

## Testing

Console error no longer appears

https://github.com/user-attachments/assets/a5211b7f-950f-47db-a9d6-f13ce23f593a

### DIFF
--- a/weave-js/src/components/Tag/Alias.tsx
+++ b/weave-js/src/components/Tag/Alias.tsx
@@ -2,6 +2,7 @@ import React, {FC, ReactElement} from 'react';
 import {twMerge} from 'tailwind-merge';
 
 import {Icon} from '../Icon';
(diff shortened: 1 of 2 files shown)


weave-js/src/components/Tag/Alias.tsx
connieelee, line 1: hide whitespace

PR #3996, opened 2025-03-27
fix(ui): expand refs in call objects when adding to dataset

bcsherma:
## Description

Calls have refs in their inputs and outputs recursively resolved before being added to a dataset.

The calls table now passes selected calls by ID to the drawer. The drawer then queries for the calls by ID and then uses the `useClientSideCallRefExpansion` hook to expand the refs. This PR adds a `recursiveUnwrap` option to the hook which causes the provided paths for expansion to be recursively expanded, ie if a ref resolves to a value that contains a ref, the inner ref is also resolved.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
@@ -1,5 +1,5 @@
 import {Box, Typography} from '@mui/material';
-import React, {useCallback, useEffect} from 'react';
+import React, {useCallback, useEffect, useMemo} from 'react';
 import {toast} from 'react-toastify';
 
 import {maybePluralize} from '../../../../../core/util/string';
@@ -8,6 +8,7 @@ import {WaveLoader} from '../../../../Loaders/WaveLoader';
 import {useWeaveflowRouteContext} from '../context';
 import {ResizableDrawer} from '../pages/common/ResizableDrawer';
 import {useWFHooks} from '../pages/wfReactInterface/context';
+import {useClientSideCallRefExpansion} from '../pages/wfReactInterface/tsDataModelHooksCallRefExpansion';
 import {
   ACTION_TYPES,
   DatasetDrawerProvider,
@@ -26,7 +27,7 @@ interface AddToDatasetDrawerProps {
   project: string;
   open: boolean;
   onClose: () => void;
-  selectedCalls: CallData[];
+  selectedCallIds: string[];
 }
 
 const typographyStyle = {fontFamily: 'Source Sans Pro'};
@@ -34,7 +35,7 @@ const typographyStyle = {fontFamily: 'Source Sans Pro'};
 export const AddToDatasetDrawer: React.FC<AddToDatasetDrawerProps> = props => {
   return (
     <DatasetDrawerProvider
-      selectedCalls={props.selectedCalls}
+      selectedCallIds={props.selectedCallIds}
       onClose={props.onClose}
       entity={props.entity}
       project={props.project}>
@@ -48,7 +49,7 @@ export const AddToDatasetDrawerInner: React.FC<AddToDatasetDrawerProps> = ({
   onClose,
   entity,
   project,
-  selectedCalls,
+  selectedCallIds,
 }) => {
   const {
     state,
@@ -61,6 +62,7 @@ export const AddToDatasetDrawerInner: React.FC<AddToDatasetDrawerProps> = ({
     resetDrawerState,
     isNextDisabled,
     editorContext,
+    processRowsForEditor,
   } = useDatasetDrawer();
 
   const {
@@ -79,11 +81,83 @@ export const AddToDatasetDrawerInner: React.FC<AddToDatasetDrawerProps> = ({
   } = state;
 
   const {peekingRouter} = useWeaveflowRouteContext();
-  const {useRootObjectVersions, useTableUpdate, useObjCreate, useTableCreate} =
-    useWFHooks();
+  const {
+    useRootObjectVersions,
+    useTableUpdate,
+    useObjCreate,
+    useTableCreate,
+    useCalls,
+  } = useWFHooks();
   const tableUpdate = useTableUpdate();
   const objCreate = useObjCreate();
   const tableCreate = useTableCreate();
+
+  // Fetch call data using the selectedCallIds
+  const callsData = useCalls(
+    entity,
+    project,
+    {callIds: selectedCallIds},
+    selectedCallIds.length // limit to fetch all selected calls
+  );
+
+  // Recursively expand all refs in inputs and output
+  const expandRefColumns = useMemo(() => new Set(['inputs', 'output']), []);
+
+  // Use the enhanced useClientSideCallRefExpansion hook with recursiveUnwrap option,
+  // but only for inputs.* fields
+  const {expandedCalls, isExpanding} = useClientSideCallRefExpansion(
+    callsData,
+    expandRefColumns,
+    {expansionDepth: 1, expandAttr: true}
+  );
+
+  const isDataLoading = useMemo(() => {
+    const hasCallData =
+      !callsData.loading && callsData.result && callsData.result.length > 0;
+
+    if (hasCallData && expandedCalls.length === 0 && !isExpanding) {
+      return false;
+    }
+
+    return callsData.loading || isExpanding;
+  }, [callsData.loading, callsData.result, expandedCalls.length, isExpanding]);
+
+  // Transform fetched call data into the format expected by the drawer
+  const selectedCalls: CallData[] = useMemo(() => {
+    if (
+      !callsData.loading &&
+      callsData.result &&
+      callsData.result.length > 0 &&
+      expandedCalls.length === 0 &&
+      !isExpanding
+    ) {
+      return callsData.result
+        .filter(call => call?.traceCall != null)
+        .map(call => ({
+          digest: call.traceCall!.id,
+          val: call.traceCall!,
+        }));
+    }
+
+    // Normal case - use expanded calls when they're available
+    if (isDataLoading || expandedCalls.length === 0) {
+      return [];
+    }
+
+    // All references are already expanded by the useClientSideCallRefExpansion hook
+    return expandedCalls
+      .filter(call => call != null)
+      .map(call => ({
+        digest: call.id,
+        val: call,
+      }));
+  }, [
+    isDataLoading,
+    expandedCalls,
+    callsData.loading,
+    callsData.result,
+    isExpanding,
+  ]);
 
   // Access edit context methods through the drawer context's editorContext property
   const {getRowsNoMeta, convertEditsToTableUpdateSpec, resetEditState} =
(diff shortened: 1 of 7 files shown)


bcsherma:
* **#3996** <a href="https://app.graphite.dev/github/pr/wandb/weave/3996?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/3996?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
```diff
@@ -79,12 +81,86 @@ export const AddToDatasetDrawerInner: React.FC<AddToDatasetDrawerProps> = ({
...
   const tableUpdate = useTableUpdate();
   const objCreate = useObjCreate();
   const tableCreate = useTableCreate();
 
+  // Fetch call data using the selectedCallIds
+  const callsData = useCalls(
+    entity,
+    project,
+    {callIds: selectedCallIds},
+    selectedCallIds.length // limit to fetch all selected calls
```
gtarpenning, line 100: why?

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
```diff
@@ -79,12 +81,86 @@ export const AddToDatasetDrawerInner: React.FC<AddToDatasetDrawerProps> = ({
...
+    if (hasCallData && expandedCalls.length === 0 && !isExpanding) {
+      return false;
+    }
+
+    return callsData.loading || isExpanding;
+  }, [callsData.loading, callsData.result, expandedCalls.length, isExpanding]);
+
+  // Transform fetched call data into the format expected by the drawer
+  const selectedCalls: CallData[] = useMemo(() => {
+    // If call data is loaded but expansion either failed or is taking too long,
```
gtarpenning, line 127: how long is too long?

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/schemaUtils.ts
```diff
@@ -141,7 +175,7 @@ export const extractSourceSchema = (calls: CallData[]): SchemaField[] => {
   });
 
   return allFields
-    .filter(field => field.name !== 'inputs.self')
+    .filter(field => !field.name.startsWith('inputs.self'))
```
gtarpenning, line 178: is this because we now resolve the ref and there can be sub-fields?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
```diff
@@ -71,98 +82,350 @@ export const useClientSideCallRefExpansion = (
...
   traceCalls: traceServerClientTypes.TraceCallSchema[],
   expandedRefColumns: Set<string>,
-  client: traceServerClient.TraceServerClient
+  client: traceServerClient.TraceServerClient,
+  recursiveUnwrap: boolean = false,
+  iterationCount: number = 0,
+  maxIterations: number = 10
 ): Promise<traceServerClientTypes.TraceCallSchema[]> => {
+  // Safety check to prevent infinite recursion
+  if (iterationCount >= maxIterations) {
```
gtarpenning, line 91: good safety check, do we also want one one the raw # of traces? one ref could have hundreds of ref children, sort of worried about blowup here

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
```diff
@@ -71,98 +82,350 @@ export const useClientSideCallRefExpansion = (
...
-      const colParts = col.split('.');
-      let value: any = call;
-      for (const part of colParts) {
+
+  if (recursiveUnwrap) {
+    // Recursively find all references in specified columns/paths
+    const expandedRefColumnsList = Array.from(expandedRefColumns ?? []);
+
+    traceCalls.forEach(call => {
+      expandedRefColumnsList.forEach(column => {
```
gtarpenning, line 102: make helper? could be cleaner for top level logic reading

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
```diff
@@ -71,98 +82,350 @@ export const useClientSideCallRefExpansion = (
...
+              parent = target;
+            }
+
+            target = target[part as keyof typeof target];
+          }
+
+          // If we found the path, recursively replace refs in it
+          if (target !== undefined && parent !== null) {
+            // Replace refs in the target value
+            replaceAllRefsRecursively(target, refsDataMap);
```
gtarpenning, line 205: this resolved all refs for top level provided keys, so `inputs.model` could resolve lots and lots of sub refs right?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
```diff
@@ -71,98 +82,350 @@ export const useClientSideCallRefExpansion = (
...
+  // Base case: null or primitive value
+  if (!obj || typeof obj !== 'object') {
+    if (isExpandableRef(obj)) {
+      refsNeeded.add(obj);
+    }
+    return;
+  }
+
+  // Prevent circular references
+  if (visited.has(obj)) {
```
gtarpenning, line 295: i wonder if this blows up for large objects, we could hash it for better mem perf at the cost of cpu time

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
```diff
@@ -71,98 +82,350 @@ export const useClientSideCallRefExpansion = (
...
+    return;
+  }
+  visited.add(obj);
+
+  // Handle arrays
+  if (Array.isArray(obj)) {
```
gtarpenning, line 301: i wonder how inclusive this is, like wb sets? is that even a thing in js?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksCallRefExpansion.ts
```diff
@@ -71,98 +82,350 @@ export const useClientSideCallRefExpansion = (
...
+
+/**
+ * Recursively replaces all references in an object with their expanded values
+ * @param obj The object to process
+ * @param refsDataMap Map of reference URIs to their resolved values
+ * @param visited Set of objects already visited to prevent circular reference issues
+ * @param depth Current recursion depth
+ * @param maxDepth Maximum recursion depth to prevent stack overflow
+ */
+const replaceAllRefsRecursively = (
```
gtarpenning, line 354: maybe possible to dry with above?

gtarpenning (commented):
mostly questions, sort of concerned about the perf.  I think the backend will reject trying to read table refs, but we should make sure thats true otherwise one table ref with rows that have images would tank this

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/schemaUtils.ts
bcsherma: Precisely!

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
tssweeney: the call id lookup is already optimized

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
```diff
@@ -79,12 +81,86 @@ export const AddToDatasetDrawerInner: React.FC<AddToDatasetDrawerProps> = ({
...
+
+  // Recursively expand all refs in inputs and output
+  const expandRefColumns = useMemo(() => new Set(['inputs', 'output']), []);
+
+  // Use the enhanced useClientSideCallRefExpansion hook with recursiveUnwrap option,
+  // but only for inputs.* fields
+  const {expandedCalls, isExpanding} = useClientSideCallRefExpansion(
+    callsData,
+    expandRefColumns,
+    {recursiveUnwrap: true}
```
tssweeney, line 111: hmmm. I actually don't think you need / want recursive expansion. This could lead to enormous datasets. I think the main thing is to expand the refs at the top-level columns. consider a case as follows:

```
class MyModel(weave.Model):
   @weave.op
   def predict(self, col_a, col_b): ...
```

In this case, when doing the prediction on a dataset row, all three will be refs! Yes, self, col_a, and col_b are refs. (col_a and col_b) are dataset row value refs. Creating a dataset from these predictions should ideally make the exact same dataset as was used in the start.

However, if col_a in itself contains a ref, then it is safe to assume that the ref was also present in the original dataset. Recursively unwrapping here could lead to data explosion and would actually result in a different dataset.

bcsherma: I've replaced the yolo recursive expand with an `expansionDepth` option which will do recursive expansion of the provided columns up to the specified depth. And the add to dataset drawer will use the hook with depth 1 to get top level expansions of the inputs and outputs keys.

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
bcsherma: that was a junk comment

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
bcsherma: I've also added an option to the existing client side ref expansion hook to expand all `/attr/` references in the call after expanding the requested columns to the requested depth.

See these traces (and try adding to a dataset) https://app.wandb.test/wandb/people-bot/weave/traces?filter=%7B%22opVersionRefs%22:%5B%22weave:///wandb/people-bot/op/Agent-get_completion:*%22%5D%7D for an example. If we have a field like `inputs.completion_params.model` which is a ref to a dataset entry, we first expand `completion_params` ref to get the `/attr/` refs for the dataset row value and then we resolve the remaining `/attr/` refs.

tssweeney (approved):

PR #3997, opened 2025-03-28
feat(weave): Add otel tracing support

zbirenbaum:
## Description

- Towards WB-23859

Adds `otel_export` endpoint and supporting types + conversion to internal schema functionality to the trace server

## Testing

It's been tested on some examples but unit/integration tests are still needed


weave/trace_server/trace_server_interface.py
```diff
@@ -212,6 +218,20 @@ class TableSchemaForInsert(BaseModel):
...
 
 
+class OtelExportReq(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    project_id: str
+    traces: ExportTraceServiceRequest
+    wb_user_id: Optional[str] = Field(None, description=WB_USER_ID_DESCRIPTION)
+
+
+class OtelExportRes(JSONResponse):
```
tssweeney, line 228: Since your request abstraction is pydantic over proto, i'd recommend your response be pydantic over proto. The service layer should do any post processing to make it jsonable.

tssweeney: Critical: `trace_server_interface` does not "know" that it is exposed as a fastapi service

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -99,6 +99,14 @@ def ensure_project_exists(
     ) -> tsi.EnsureProjectExistsRes:
         return self._internal_trace_server.ensure_project_exists(entity, project)
 
+    def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
+        req.project_id = self._idc.ext_to_int_project_id(req.project_id)
+        # if req.wb_run_id is not None:
```
tssweeney, line 104: leftover

weave/trace_server/opentelemetry/python_spans.pyi
tssweeney, line 1: why did you need this interface definition file? It looks like your implementation is fully defined.

weave/trace_server/opentelemetry/python_spans.py
```diff
@@ -0,0 +1,342 @@
...
+        )
+        # TODO: (Followup) Get Exception from Output
+
+        summary_insert_map = tsi.SummaryInsertMap(usage={})
+
+        end_call = tsi.EndedCallSchemaForInsert(
+            project_id=project_id,
+            id=self.span_id,
+            ended_at=self.end_time,
+            exception=None,
```
tssweeney, line 275: should `exception` depend on status?

weave/trace_server/opentelemetry/python_spans.pyi
zbirenbaum: We can probably remove this now. this was generated by claude and I noticed that without it some autocomplete information wasn't working properly during development and it helped me keep track of where to implement functionality. Now it's likely unnecessary so I'll remove it.

weave/trace_server/opentelemetry/python_spans.pyi
```diff
@@ -0,0 +1,146 @@
...
+    def end_time(self) -> datetime.datetime: ...
+    @property
+    def duration_ns(self) -> int: ...
+    @property
+    def duration_ms(self) -> float: ...
+    @classmethod
+    def from_proto(cls, proto_span: PbSpan) -> Span: ...
+    def to_call(
+        self, project_id: str
+    ) -> tuple[tsi.StartedCallSchemaForInsert, tsi.EndedCallSchemaForInsert]: ...
```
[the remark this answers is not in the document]
tssweeney: I might take this back - looking at the implementation, it might be more pain than it is worth to handle the summary and other insert-specific mechanics

tests/trace_server/test_opentelemetry.py
```diff
@@ -0,0 +1,426 @@
...
+
+    return tsi.OtelExportReq(
+        project_id=project_id,
+        traces=request,
+        wb_user_id=None
+    )
+
+
+@pytest.fixture
+def mock_clickhouse_trace_server():
```
tssweeney, line 127: I think you instead want to use the `client` fixture. These mocks seem to hide some of the critical things you are testing.

weave/trace_server/opentelemetry/python_spans.py
zbirenbaum: I updated it to set the status and tested it out. The message doesn't seem to display properly in the dash unfortunately but it does when you click in.

image
image

weave/trace_server/trace_server_interface.py
zbirenbaum: I don't think we can make this a pydantic, according to the spec the response has to be a proto:

From https://opentelemetry.io/docs/specs/otlp/

> OTLP/gRPC Response
> The response MUST be the appropriate message (see below for the specific message to use in the [Full Success](https://opentelemetry.io/docs/specs/otlp/#full-success), [Partial Success](https://opentelemetry.io/docs/specs/otlp/#partial-success) and [Failure](https://opentelemetry.io/docs/specs/otlp/#failures) cases).
> 
> Full Success
> The success response indicates telemetry data is successfully accepted by the server.
> 
> If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with success.
> 
> On success, the server response MUST be a [Export\<signal\>ServiceResponse](https://github.com/open-telemetry/opentelemetry-proto/tree/v1.5.0/opentelemetry/proto/collector) message (ExportTraceServiceResponse for traces, ExportMetricsServiceResponse for metrics, ExportLogsServiceResponse for logs and ExportProfilesServiceResponse for profiles).
> 
> The server MUST leave the partial_success field unset in case of a successful response.

tssweeney: that is fine. i'm saying that your `trace_server_interface` shall not contain service primitives like `JSONRespose`

weave/trace_server/opentelemetry/python_spans.py
tssweeney: what is wrong with the dash?

zbirenbaum: I thought that the lowercase status field would display the error message rather than just `error` but maybe that's not the case

tssweeney: what is that lower case status? that looks like a custom field, not a standard field

zbirenbaum: I fixed it, misread a schema, we can resolve this now

weave/trace_server/trace_server_interface.py
```diff
@@ -212,6 +216,18 @@ class TableSchemaForInsert(BaseModel):
...
+class OtelExportReq(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    project_id: str
+    traces: ExportTraceServiceRequest
+    wb_user_id: Optional[str] = Field(None, description=WB_USER_ID_DESCRIPTION)
+
+
+# Spec requires that the response be of type Export<signal>ServiceResponse
+# https://opentelemetry.io/docs/specs/otlp/
+OtelExportRes: TypeAlias = "ExportTraceServiceResponse"
```
tssweeney, line 228: sorry. i should have been specific. Here, please use:

```
class OtelExportRes(BaseModel):
   response: ExportTraceServiceResponse
```

Then you can correctly handle the contract in the server layer. Your `OtelExportReq` already deviates from a pure otel sink

tssweeney: interesting. i wonder why it is ok with the request

tssweeney: you could also do a TypedDict

zbirenbaum: Nevermind, I'm getting a bad request error pretty much no matter what I do after swapping to this. I think it's going to take more work than just an ignore flag...

tssweeney: that comment implies something is not clear. The service itself needn't us this exact payload.

tests/trace_server/test_opentelemetry.py
zbirenbaum: Should I use test_custom_provider as inspiration? Or is there another example you recommend

tssweeney: there are very many such examples see `test_trace_client`. when you use the `client` fixture, you get `client.server` which is basically your clickhouse batched server

pyproject.toml
```diff
@@ -71,12 +71,14 @@ trace_server = [
...
   "litellm>=1.36.1",
+   # OTEL tracing support
+  "opentelemetry-proto>=1.12.0",
 ]
 trace_server_tests = [
   # BYOB - S3
   "moto[s3]>=5.0.0",
 ]
-docs = ["playwright", "lazydocs", "nbformat", "nbconvert"]
+docs = ["playwright", "lazydocs", "nbformat", "nbconvert", "opentelemetry-proto>=1.12.0"]
```
tssweeney, line 81: needed for docs?!

weave/trace_server/trace_server_interface.py
```diff
@@ -212,6 +221,26 @@ class TableSchemaForInsert(BaseModel):
...
+class OtelExportReq(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    project_id: str
+    traces: ExportTraceServiceRequest
+    wb_user_id: Optional[str] = Field(None, description=WB_USER_ID_DESCRIPTION)
+
+
+# Spec requires that the response be of type Export<signal>ServiceResponse
+# https://opentelemetry.io/docs/specs/otlp/
+class OtelExportRes(BaseModel):
```
tssweeney, line 233: I don't quite understand why you still went with BaseModel here instead of just a simple class given the workaround needed. What do we get from BaseModel here?

tests/trace_server/test_opentelemetry.py
```diff
@@ -0,0 +1,338 @@
...
+    assert isinstance(response, tsi.OtelExportRes)
+
+    # Query the calls
+    res = client.server.calls_query(
+        tsi.CallsQueryReq(
+            project_id=project_id,
+        )
+    )
+    # Verify that the start and end calls were merged into a single call
+    assert len(res.calls) == 1
```
tssweeney, line 142: i would add assertions similar to the other tests that verify the call's properties are as you expect.

weave/trace_server/trace_server_interface.py
zbirenbaum: Anything which expects a Basemodel such as the generic endpoint caller in the http server can take this as a parameter. We only really care about the nested field and validation is skipped for that so it adds a bit of flexibility at no cost.

weave/trace_server/trace_server_interface.py
zbirenbaum: think this is resolved so closing

pyproject.toml
zbirenbaum: Unfortunately yes. I asked Andrew about this and seems its a bug. When the dep is only in trace server, the docs job fails to run in CI since trace_server_interface includes the proto import. 

Docgen for trace server should require all the trace_server deps but it doesn't. This was better than making the proto package a top level requirement which was the other fix that worked

tests/trace_server/test_opentelemetry.py
zbirenbaum: I added some more checks, let me know your thoughts

weave/trace_server/trace_server_interface.py
```diff
@@ -212,6 +221,26 @@ class TableSchemaForInsert(BaseModel):
...
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    _response: ExportTraceServiceResponse
+
+    def __init__(self, **data: Any) -> None:
+        super().__init__(**data)
+        # Set the response type to ExportTraceServiceResponse
+        # TODO: Actually populate the error fields
```
zbirenbaum, line 240: The error field population might actually not be super important here. I think it's to communicate partial successes so that the client knows which span exports it should retry. If an error is raised the full thing should retry. Since we do this batched all in one go, if an error is raised and it wants to retry it should do everything anyways I think.

Since in the end we communicate back with JSONResponse instead of protobuf message I'm not sure the client even knows how to correctly parse this.

Other implementations just always return a newly initialized trace service response at the end of execution regardless so this is probably fine as is, but wanted to leave a comment in case it comes up later.

tests/trace_server/test_opentelemetry.py
```diff
@@ -0,0 +1,365 @@
...
+
+    call = res.calls[0]
+    export_span = export_req.traces.resource_spans[0].scope_spans[0].spans[0]
+    decoded_trace = hexlify(export_span.trace_id).decode("ascii")
+    decoded_span = hexlify(export_span.span_id).decode("ascii")
+
+    # TODO: Once postprocessing is added we should have tests for input, output and summary propogation
```
tssweeney, line 150: we should add these tests now for current func

pyproject.toml
tssweeney: oh, you could just make docs depend on server.... but i got it

tssweeney (approved):
Approving. I think there are still a number of little cleanups, but let's get this thing out

PR #3998, opened 2025-03-28
feat(weave): adds api capability to include storage size

chance-wnb:
## Overview
This PR adds the ability to query storage size information for calls in the trace server. It introduces two new features:
1. Individual call storage size
2. Rolled-up storage size for entire traces

## Changes
- Added new fields to `CallSchema`:
  - `storage_size_bytes`: Size of metadata storage for individual calls
  - `total_size_bytes`: Total size of metadata storage for entire traces
- Added new query parameters to `CallsQueryReq`:
  - `include_storage_size`: Returns storage size for each call
  - `include_rolled_up_storage_size`: Returns total size for root calls only
- Implemented storage size calculation in both SQLite and ClickHouse trace servers
- Added unit tests for both features

## Implementation Details
- For SQLite: Uses `length()` function to calculate storage size from JSON fields
- For ClickHouse: Uses `calls_merged_stats` table to get pre-calculated sizes
- Storage size is calculated as sum of:
  - attributes size
  - inputs size
  - output size
  - summary size

CallsQueryReq is added with two new fields `include_storage_size` , `include_rolled_up_storage_size` .

- `include_rolled_up_storage_size` is primarily used for the calls list view UI.
- `include_storage_size` should be included in the call details view
- `include_rolled_up_storage_size` if the FE can detect that the call about to be opened is the root call, include this as well when requesting the call details data.

### Client Tests (SQlite only for now)
1. `test_calls_query_with_storage_size`: Tests individual call storage size
2. `test_calls_query_with_rolled_up_storage_size`: Tests trace-level storage size
3. `test_calls_query_with_both_storage_sizes`: Tests both features together

### Query Builder Tests
1. `test_storage_size_fields`: Tests SQL generation for individual storage size queries
2. `test_rolled_up_storage_size`: Tests SQL generation for rolled-up storage size queries
3. `test_aggregated_data_size_field`: Tests the new `AggregatedDataSizeField` class

### Manual tests:
Despite there is no UI, I did manually set the `include_storage_size` and `include_rolled_up_storage_size` and did see the response contains the expected numbers.

## Notes
- This is marked as beta and subject to change
- This is only the server side change, I will also be responsible for frontend change later.
- Storage size calculation is optimized for ClickHouse backends
- Rolled-up storage size is only available for root calls (calls without parents)
- A mini design-doc can be found [here (internal)](https://www.notion.so/wandbai/Data-Size-display-1bde2f5c7ef38087bc03cb8accb95bfb)

- Addresses task WB-23748

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -181,6 +181,8 @@ def test_trace_server_call_start_and_end(client):
         "wb_run_id": None,
         "deleted_at": None,
         "display_name": None,
+        "storage_size_bytes": None,
+        "total_storage_size_bytes": None,
     }
 
     end = tsi.EndedCallSchemaForInsert(
@@ -227,6 +229,8 @@ def test_trace_server_call_start_and_end(client):
         "wb_run_id": None,
         "deleted_at": None,
         "display_name": None,
+        "storage_size_bytes": None,
+        "total_storage_size_bytes": None,
     }
 
 
--- a/weave/trace_server/calls_query_builder.py
+++ b/weave/trace_server/calls_query_builder.py
@@ -87,6 +87,21 @@ class CallsMergedAggField(CallsMergedField):
         return clickhouse_cast(f"{self.agg_fn}({inner})")
 
 
+class AggFieldWithTableOverrides(CallsMergedAggField):
+    # This is useful if you know the field must come from a predetemined table alias.
+
+    table_name: str
+
+    def as_sql(
+        self,
+        pb: ParamBuilder,
+        table_alias: str,
+        cast: Optional[tsi_query.CastTo] = None,
+        use_agg_fn: bool = True,
+    ) -> str:
+        return super().as_sql(pb, self.table_name, cast, use_agg_fn)
+
+
 class CallsMergedDynamicField(CallsMergedAggField):
     extra_path: Optional[list[str]] = None
 
@@ -201,6 +216,36 @@ class CallsMergedFeedbackPayloadField(CallsMergedField):
         raise NotImplementedError(
             "Feedback fields cannot be selected directly, yet - implement me!"
         )
+
+
+class AggregatedDataSizeField(CallsMergedField):
+    join_table_name: str
+
+    def is_heavy(self) -> bool:
+        return True
+
+    def as_sql(
+        self,
+        pb: ParamBuilder,
+        table_alias: str,
+        cast: Optional[tsi_query.CastTo] = None,
+    ) -> str:
+        # This field is not supposed to be called yet. For now, we just take the parent class's
+        # implementation. Consider re-implementation for future use.
+        return super().as_sql(pb, table_alias, cast)
+
+    def as_select_sql(self, pb: ParamBuilder, table_alias: str) -> str:
+        # It doesn't make sense for a non-root call to have a total storage size,
+        # even if a value could be computed.
+        conditional_field = f"""
+        CASE
+            WHEN any({table_alias}.parent_id) IS NULL
+            THEN any({self.join_table_name}.total_storage_size_bytes)
+            ELSE NULL
+        END
+        """
+
+        return f"{conditional_field} AS {self.field}"
 
 
 class QueryBuilderDynamicField(QueryBuilderField):
@@ -371,6 +416,8 @@ class CallsQuery(BaseModel):
     limit: Optional[int] = None
     offset: Optional[int] = None
     include_costs: bool = False
+    include_storage_size: bool = False
+    include_total_storage_size: bool = False
 
     def add_field(self, field: str) -> "CallsQuery":
         name = get_field_by_name(field)
@@ -569,7 +616,11 @@ class CallsQuery(BaseModel):
 
         # If so, build the two queries
         filter_query = CallsQuery(project_id=self.project_id)
-        outer_query = CallsQuery(project_id=self.project_id)
+        outer_query = CallsQuery(
+            project_id=self.project_id,
+            include_storage_size=self.include_storage_size,
+            include_total_storage_size=self.include_total_storage_size,
+        )
 
         # Select Fields:
         filter_query.add_field("id")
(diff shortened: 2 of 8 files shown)


tests/trace/test_client_trace.py
```diff
@@ -3471,3 +3475,116 @@ def test_call_stream_query_heavy_query_batch(client):
     assert len(list(res)) == 10
     for call in res:
         assert call.attributes["empty"] == ""
+
+
+def test_calls_query_with_storage_size(client):
+    """Test querying calls with storage size information"""
+    if not client_is_sqlite(client):
+        # due to limitations on how calls_merged_stats is calculated,
```
chance-wnb, line 3483: To successfully run the 3 tests for ClickHouse client, it requires that the materialized view of `calls_merged_stats` finish aggregation at the time the API call is made. This cannot be realized during the duration of a test case. I noticed that the ClickHouse engine almost decides when to do the aggregation at its own discretion. 

I cannot find a sensible way to control this behavior, so I gave up testing them for ClickHouse. I am open for suggestions.

weave/trace_server/calls_query_builder.py
```diff
@@ -699,10 +751,34 @@ def _as_sql_base_format(
...
 
+        storage_size_sql = ""
+        if self.include_storage_size:
+            storage_size_sql = """
+            LEFT JOIN (SELECT
+                id,
+                sum(COALESCE(attributes_size_bytes,0) + COALESCE(inputs_size_bytes,0) + COALESCE(output_size_bytes,0) + COALESCE(summary_size_bytes,0)) as storage_size_bytes
+            FROM calls_merged_stats
+            GROUP BY id) as storage_size_tbl
+            on calls_merged.id = storage_size_tbl.id
```
tssweeney, line 762: always also include project_id equality (here and in the other snippet)

weave/trace_server/trace_server_interface.py
```diff
@@ -118,6 +118,12 @@ class CallSchema(BaseModel):
 
     deleted_at: Optional[datetime.datetime] = None
 
+    # Size of metadata storage for this call
+    storage_size_bytes: Optional[int] = None
+
+    # Total size of metadata storage for the entire trace
+    total_size_bytes: Optional[int] = None
```
tssweeney, line 125: Please standardize terms. `total_size_bytes` <> `include_rolled_up_storage_size`.

I'd recommend:
* `storage_size_bytes` and
* `trace_storage_size_bytes`, where `trace_*` contains the rolled up storage for `trace root calls` only

tests/trace/test_client_trace.py
```diff
@@ -3471,3 +3475,118 @@ def test_call_stream_query_heavy_query_batch(client):
...
+            tsi.CallsQueryReq(
+                project_id=get_client_project_id(client), include_storage_size=True
+            )
+        )
+    )
+    assert len(calls) == 1
+    call = calls[0]
+
+    # Verify storage size is present and reasonable
+    assert call.storage_size_bytes > 0
```
tssweeney, line 3508: why not make the assertion explicit and specific?

tests/trace/test_client_trace.py
```diff
@@ -3471,3 +3475,118 @@ def test_call_stream_query_heavy_query_batch(client):
...
+    assert len(calls) == 2  # Parent and child calls
+
+    # Find parent and child calls
+    parent_call = next(c for c in calls if c.parent_id is None)
+    child_call = next(c for c in calls if c.parent_id is not None)
+
+    # Verify storage size fields
+    assert parent_call.storage_size_bytes > 0
+    assert (
+        parent_call.total_size_bytes > parent_call.storage_size_bytes
```
tssweeney, line 3589: here and other places, i would assert the specific values

weave/trace_server/calls_query_builder.py
```diff
@@ -203,6 +218,37 @@ def as_select_sql(self, pb: ParamBuilder, table_alias: str) -> str:
...
+        # implementation. Consider re-implementation for future use.
+        return super().as_sql(pb, table_alias, cast)
+
+    def as_select_sql(self, pb: ParamBuilder, table_alias: str) -> str:
+        # It doesn't make sense for a non-root call to have a rolled up storage size,
+        # even if a value could be computed.
+        conditional_field = f"""
+        CASE
+            WHEN any({table_alias}.parent_id) IS NULL
+            THEN any(rolled_up_cms.total_size_bytes)
```
tssweeney, line 244: why is table_alias parametrized but `rolled_up_cms` is hard coded?

chance-wnb: When `as_select_sql` is called, the caller doesn't send a secondary table alias, and it is unable to do so. `rolled_up_cms` is the joined table alias, which is currently indeed hard coded.

weave/trace_server/calls_query_builder.py
```diff
@@ -203,6 +218,37 @@ def as_select_sql(self, pb: ParamBuilder, table_alias: str) -> str:
         )
 
 
+class AggregatedDataSizeField(CallsMergedField):
+    def __init__(self, field: str = "total_size_bytes"):
```
tssweeney, line 222: i'd recommend not having defaults here forcing the developer to define the field in the spec below

weave/trace_server/calls_query_builder.py
tssweeney: I think the problem is that this implicitly binds the relationship between this part of the code and completely other part of the code that specifies "rolled_up_cms"

chance-wnb: That is indeed the case. Due to the current code layout, I haven't been able to find a delicate way to link them together. The existing `feedback` join table is also hard coded. 

I agree this does make the code hard to trace and follow. I am open to suggestions that can improve this situation.

tssweeney: can the constructor accept a join table alias?

tests/trace/test_client_trace.py
chance-wnb: This is great. TIL. Thanks!

However, this doesn't really work reliably, sometimes my test get the right number, sometimes not. orz.

When i manually inspect the DB using a ClickHouse console eg. it is always right.

My theory is that when multiple clients connect to an instance, they have different perspectives and are not necessarily in sync. I have one client(a driver client in test) to do `OPTIMIZE TABLE` and the other client (which is the trace_server) uses doesn't necessarily get the most up-to-date state.

Anyhow, I gave up the idea to carrying a real instance in tests, and write a thin layer of testing with mocks instead. It should help reduce new test flakiness.

weave/trace_server/trace_server_interface.py
chance-wnb: Thanks. They should be in sync now.

tests/trace/test_client_trace.py
chance-wnb: I have rewritten the tests completely using mocks. The previous tests logic is overly complicated and are relying on the real DB engines in runtime, which are not reliable at all. I no longer feel that it is worth the hassle to get them working while introducing great risk of flakiness.

weave/trace_server/calls_query_builder.py
chance-wnb: Along with using the constructor parameters, I also laid down some constants, hopefully it reduces some degree of hardcode-ness.

tests/trace/test_client_trace.py
tssweeney: Hey Chance. I appreciate your determination (: I put a meeting on the cal tomorrow to discuss. I would rather sync verbally so you can land this asap. IIUC your changes, the test refactor makes it so that you are no longer actually testing the authored functionality. Getting SQLlite to work in this case is far less important (and skippable) compared to ensuring the clickhouse implementation is correct.

tssweeney (approved):

PR #4003, opened 2025-03-28
chore(weave): read CH database settings in test

chance-wnb:
## Description

By default when running test against clickhouse-backed trace server, the test run will erase the data in the `default` database, which jeopardizes the local dev environment. One way to work around it is to use WF_CLICKHOUSE_DATABASE env var.

However, when I use this command to run a test against the clickhouse server:
```
WF_CLICKHOUSE_DATABASE=test nox -e "tests-3.12(shard='trace')" -- --weave-server=clickhouse
```

Because the conftest is hard coded to use the `default` database, the execution will fail. This PR fixes that. 

However I recommend using an alternative port, which is also supported in [this internal PR](https://github.com/wandb/core/pull/28414).

## Testing

Locally verified by running test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -570,7 +570,9 @@ def create_client(
     elif weave_server_flag == "clickhouse":
         ch_server = clickhouse_trace_server_batched.ClickHouseTraceServer.from_env()
         ch_server.ch_client.command("DROP DATABASE IF EXISTS db_management")
-        ch_server.ch_client.command("DROP DATABASE IF EXISTS default")
+        ch_server.ch_client.command(
+            f"DROP DATABASE IF EXISTS {ts_env.wf_clickhouse_database()}"
+        )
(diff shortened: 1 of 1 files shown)


tssweeney (approved):
This is good, but not a completion solution. `db_management` is going to be dropped which will drop migration info for other local dbs. for a more complete solution, we should only be dropping the row from db_management that corresponds to the DB in the next line.

chance-wnb:
> This is good, but not a completion solution. `db_management` is going to be dropped which will drop migration info for other local dbs. for a more complete solution, we should only be dropping the row from db_management that corresponds to the DB in the next line.

Yes, exactly. And that's part of the reason why I think using an alternative port is a better method. It is not a complete solution, and maybe we don't have to pursue a complete solution given a better alternative exists. A user who really wants to use `WF_CLICKHOUSE_DATABASE` should know its consequences.

PR #4005, opened 2025-03-29
feat(weave): Framework-agnostic Eval API

andrewtruong:
Adds a beta imperative API for weave Evaluation logging and an updated Evaluations page UI to view those logged traces.

User API:
```py
dataset = [{'text': ...,}, {'text': ...,}, ...]

ev = BetaEvaluationLogger()

for row in dataset:
    user_output = user_model_func(row["img"], row["number"])  # Bring your own function
    pred = ev.log_prediction(inputs=row, output=user_output)
    ev.log_summary({'coolness': 9001})  # Log any summary values you want!
```

Example UI:

image


tests/trace/test_evaluation_imperative.py
```diff
@@ -0,0 +1,28 @@
...
+        return a + b
+
+    return func
+
+
+def test_basic_evaluation(client, user_dataset, user_model):
+    ev = BetaEvaluationLogger()
+
+    for row in user_dataset:
+        model_output = user_model(row["a"], row["b"])
```
tssweeney, line 23: maybe i will learn about this later, but i would expect the hello world case to also include scores on predictions

weave/flow/eval2.py
```diff
@@ -0,0 +1,200 @@
...
+
+ID = str
+
+NOT_DEFINED = "Not defined for custom scoring"
+
+# Context variable to store the current output safely between threads.  This also
+# ensures that only 1 version of the predict method is saved because the code
+# contents are always the same.
+current_output: ContextVar[Any] = ContextVar("current_output", default=None)
+current_score: ContextVar[float | None] = ContextVar("current_score", default=None)
```
tssweeney, line 25: scores should support more than float. namely bool & maps

weave/flow/eval2.py
```diff
@@ -0,0 +1,200 @@
+from collections.abc import Iterator
```
tssweeney, line 1: general nit: "Beta" seems unnecessary? Why include it? Is there a better world that could be more descriptive?

weave/flow/eval2.py
```diff
@@ -0,0 +1,200 @@
...
+    hack_reference_to_predict_and_score: Call
+    hack_reference_to_evaluate: Call
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    async def log_score(
+        self, scorer_name: str, score: float, metadata: dict | None = None
+    ) -> BetaScoreLogger:
+        # Define the scorer class dynamically
+        cls_name = scorer_name.replace(".", "_")
```
tssweeney, line 76: i would more rigorously validate scorer name to be a valid classname (alphabetic strings, underscores, no leading numbers)

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/SummaryPlotsSection/SummaryPlotsSection.tsx
```diff
@@ -32,6 +31,8 @@ export const SummaryPlots: React.FC<{
   const {radarData, allMetricNames} = usePlotDataFromMetrics(state);
   const {selectedMetrics} = state;
 
+  console.log('state', state);
```
tssweeney, line 34: dev tracks

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/compositeMetricsUtil.ts
```diff
@@ -80,6 +82,8 @@ export const buildCompositeMetricsMap = (
 ): CompositeScoreMetrics => {
   const composite: CompositeScoreMetrics = {};
 
+  console.log('summaryData', summaryData);
```
tssweeney, line 85: dev tracks

weave/flow/eval2.py
```diff
@@ -0,0 +1,200 @@
...
+        with call_context.set_call_stack(
+            [self.hack_reference_to_evaluate, self.hack_reference_to_predict_and_score]
+        ):
+            await self.hack_reference_to_predict_and_score.apply_scorer(scorer_instance)
+
+        return BetaScoreLogger(prediction_id="123", score=score, metadata=metadata)
+
+
+class BetaEvaluationLogger(BaseModel):
+    model_id: ID = ""
```
tssweeney, line 98: I am not seeing `model_id` actually consumed anywhere. I think that `model_id` needs to be somewhat a property of the created model itself. I also think a dictionary would be good here as it can uniquely identify the model based on properties.

weave/flow/eval2.py
```diff
@@ -0,0 +1,200 @@
...
+    model_id: ID
+    inputs: dict
+    output: Any
+    hack_reference_to_predict_and_score: Call
+    hack_reference_to_evaluate: Call
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    async def log_score(
+        self, scorer_name: str, score: float, metadata: dict | None = None
```
tssweeney, line 73: similarly, i think scorer_name should support dictionaries

weave/flow/eval2.py
```diff
@@ -0,0 +1,200 @@
...
+    try:
+        yield
+    finally:
+        current_summary.reset(token)
+
+
+class BetaScoreLogger(BaseModel):
+    prediction_id: ID
+    score: float
+    metadata: dict | None
```
tssweeney, line 60: metadata does not seem to be consumed anywhere

weave/flow/eval2.py
```diff
@@ -0,0 +1,200 @@
...
+        assert self._evaluate_call is not None
+        return BetaPredictionLogger(
+            model_id="123",
+            inputs=inputs,
+            output=model_output,
+            hack_reference_to_predict_and_score=predict_and_score_call,
+            hack_reference_to_evaluate=self._evaluate_call,
+        )
+
+    def log_summary(self, summary: dict) -> None:
```
tssweeney, line 188: it would be nice if the summary could support a function that accepted prior rows

andrewtruong: I think we can leave this up to the user for now.  The user has full control and can implement their own roll-up func for summary

weave/flow/eval2.py
andrewtruong: I don't understand the dictionary comment, can you expand?

weave/flow/eval2.py
andrewtruong: I don't understand this one either, can you expand?

weave/integrations/integration_utilities.py
```diff
@@ -92,6 +92,11 @@ def op_name_from_ref(ref: str) -> str:
     return ref.split("/")[-1].split(":")[0]
 
 
+def op_name_from_call(call: Call) -> str:
+    op_name = cast(str, call._op_name)
```
tssweeney, line 96: Unsafe cast. could be a future. Either use `op_name` which will wait on the future or do a type check first

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,246 @@
...
+            # --- Setup the evaluation object ---
+            @weave.op(name="Evaluation.evaluate")
+            def evaluate(self: Evaluation, model: Model) -> None: ...
+
+            @weave.op
+            def predict_and_score(self: Evaluation, model: Model, inputs: dict) -> dict:
+                model_output = model.get_infer_method()(inputs)
+                return {
+                    "model_output": model_output,
+                    "scores": NOT_DEFINED,
```
tssweeney, line 184: The `NOT_DEFINED` value: a) looks funky in the UI; b: deviates from the standard type. I would prefer either None, or empty dict/0 respectively.

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,246 @@
...
+            # The following section is a "hacky" way to create Model and Evaluation
+            # objects that "look right" to our object saving system.
+
+            # --- Setup the model object ---
+            @weave.op
+            def predict(self: Model, inputs: dict) -> Any:
+                # Get the output from the context variable
+                return current_output.get()
+
+            self._pseudo_model.__dict__["predict"] = MethodType(
```
tssweeney, line 171: If I am not mistaken, there will be no difference between the models in these evaluations. I really think we need to have the constructor accept either a string (name/id) or a dictionary which is used as part of the model to distinguish it in the object system. This `id_type` should be the same as scorer_name in logging score as well.

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,246 @@
...
+    _logged_summary: bool = PrivateAttr(False)
+    _evaluate_call: Call | None = PrivateAttr(None)
+    _pseudo_model: Model = PrivateAttr(default_factory=lambda: Model())
+    _pseudo_evaluation: Evaluation = PrivateAttr(
+        default_factory=lambda: Evaluation(
+            dataset=weave.Dataset(rows=weave.Table([{"": ""}])),
+            scorers=[],
+        )
+    )
+    _starting_stack: list[Call] = PrivateAttr(default_factory=list)
```
tssweeney, line 156: this property is never accessed

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,246 @@
...
+    inputs: dict
+    output: Any
+    predict_and_score_call: Call
+    evaluate_call: Call
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    async def log_score(
+        self,
+        scorer_name: str,
```
tssweeney, line 100: see comment below - i think `scorer_name` should be a string or a dict

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,246 @@
...
+        score: ScoreType,
+        metadata: dict | None = None,
+    ) -> None:
+        # Define the scorer class dynamically
+        validate_scorer_name(scorer_name)
+        cls_name = scorer_name.replace(".", "_")
```
tssweeney, line 106: your validation asserts that scorer_name does not contain dots.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/ecpTypes.ts
```diff
@@ -150,3 +150,9 @@ export type MetricResult = {
   value: MetricValueType;
   sourceCallId: string;
 };
+
+// TraceCall shapes
+export type TraceCallData = {
+  callId: string;
+  traceCall: any;
```
tssweeney, line 157: I think this wants to be `TraceCallSchema`

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/CompareEvaluationsPage.tsx
```diff
@@ -85,6 +104,19 @@ export const CompareEvaluationsPageContent: React.FC<
...
 
+  const {useCalls} = useWFHooks();
+  const childCalls = useCalls(props.entity, props.project, {
+    parentIds: props.evaluationCallIds,
+  });
+
+  const predictAndScoreCalls = getCallsSubset(childCalls.result, [
+    'predict_and_score',
+  ]);
+  const summarizeCalls = getCallsSubset(childCalls.result, [
```
tssweeney, line 115: both of these should be memoed, else they will re-compute each time

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/CompareEvaluationsPage.tsx
```diff
@@ -185,6 +225,23 @@ const CompareEvaluationsPageInner: React.FC<{
     Object.keys(state.loadableComparisonResults.result?.resultRows ?? {})
       .length > 0;
   const resultsLoading = state.loadableComparisonResults.loading;
+
+  // Check if we should show the traceCalls UI
+  const isTraceCallsPath = props.traceCalls && props.traceCalls.length > 0;
+
+  if (isTraceCallsPath) {
```
tssweeney, line 232: Can you help explain what is going on here? when would traceCalls and summarizeCalls ever be empty?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/TraceCallsCompareEvaluationsPage.tsx
```diff
@@ -0,0 +1,518 @@
+import {Box} from '@material-ui/core';
```
tssweeney, line 1: I would move this file and the `TraceCallsSection` into a subdir inside of `CompareEvaluationsPage` since it is basically a completely alternative implementation

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/CompareEvaluationsPage.tsx
```diff
@@ -185,6 +225,23 @@ const CompareEvaluationsPageInner: React.FC<{
...
+    // Use the new TraceCallsCompareEvaluationsPage component
+    return (
```
tssweeney, line 234: Verified that current evals meet this condition, meaning all evals use this new view (which is breaks for current evals and is a visual regression). For expediency, let's land the python code separate from this

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,246 @@
...
+            return cast(float, current_score.get())
+
+        scorer_instance.__dict__["score"] = MethodType(score_method, scorer_instance)
+
+        # attach the score feedback to the predict call
+        with call_context.set_call_stack(
+            [self.evaluate_call, self.predict_and_score_call]
+        ):
+            with set_current_score(score):
+                await self.predict_and_score_call.apply_scorer(scorer_instance)
```
tssweeney, line 123: This is not correct - `apply_scorer` MUST be called on the model predict call

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,325 @@
...
+        )
+    )
+
+    @model_validator(mode="after")
+    def _validate_model(self) -> ImperativeEvaluationLogger:
+        if isinstance(self.model, str):
+            # Convert string to a model instance
+            # Create a dynamic model class with the string as the name
+            model_name = self.model
+            DynamicModel = type(model_name, (Model,), {})
```
tssweeney, line 233: don't we need to handle the dict case here?

andrewtruong: thanks for catching -- this is extraneous and now removed.  All casting is done in the annotation above.

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,314 @@
...
+    def log_score(self, scorer: Scorer | str, score: ScoreType) -> None:
+        """Log a score synchronously by calling the async method.
+
+        This is a convenience method for when you don't want to use async/await.
+        If called from within an existing event loop, use _alog_score instead.
+        """
+        try:
+            asyncio.run(self._alog_score(scorer, score))
+        except RuntimeError as e:
+            if "This event loop is already running" in str(e):
```
tssweeney, line 146: is it possible to run this immediately using the current event loop?

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,314 @@
...
+            return cast(ScoreType, current_score.get())
+
+        scorer.__dict__["score"] = MethodType(score_method, scorer)
+
+        # attach the score feedback to the predict call
+        with call_context.set_call_stack(
+            [self.evaluate_call, self.predict_and_score_call]
+        ):
+            with _set_current_score(score):
+                await self.predict_and_score_call.apply_scorer(scorer)
```
tssweeney, line 182: We don't apply scorer to the predict and score - we apply it to the predict call

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,314 @@
...
+                # Get the output from the context variable
+                return current_output.get()
+
+            self.model.__dict__["predict"] = MethodType(predict, self.model)
+
+            # --- Setup the evaluation object ---
+            @weave.op(name="Evaluation.evaluate")
+            def evaluate(self: Evaluation, model: Model) -> None: ...
+
+            @weave.op
```
tssweeney, line 248: can we call this (`name=Evaluation.predict_and_score`) to match the current datamodel

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,314 @@
...
+            def predict_and_score(self: Evaluation, model: Model, inputs: dict) -> dict:
+                model_output = model.get_infer_method()(inputs)
+                return {
+                    "model_output": model_output,
+                    "scores": {},
+                    "model_latency": None,
+                }
+
+            @weave.op
```
tssweeney, line 257: same here: `Evaluation.summarize`

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,314 @@
...
+
+            self.model.__dict__["predict"] = MethodType(predict, self.model)
+
+            # --- Setup the evaluation object ---
+            @weave.op(name="Evaluation.evaluate")
+            def evaluate(self: Evaluation, model: Model) -> None: ...
+
+            @weave.op
+            def predict_and_score(self: Evaluation, model: Model, inputs: dict) -> dict:
+                model_output = model.get_infer_method()(inputs)
```
tssweeney, line 250: the `inputs` param should be `example` to match current data model

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,314 @@
...
+                return {
+                    "model_output": model_output,
+                    "scores": {},
```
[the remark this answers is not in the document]
tssweeney: if we don't collect it here, the UI will need to do more work to query for this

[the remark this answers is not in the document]
andrewtruong: we can add it, but I think it will involve more hacks to the op deco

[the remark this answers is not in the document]
andrewtruong: basically we can have a "no finish mode" and then have explicitly write a finish in `eval_imperative`

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,314 @@
...
+        # dynamically create the model class
+
+        if not self._eval_started:
+            self._eval_started = True
+
+            # The following section is a "hacky" way to create Model and Evaluation
+            # objects that "look right" to our object saving system.
+
+            # --- Setup the model object ---
+            @weave.op
```
tssweeney: `

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,355 @@
...
+            raise ValueError("predict_call should not be None")
+
+        assert self._evaluate_call is not None
+        return ImperativeScoreLogger(
+            predict_and_score_call=predict_and_score_call,
+            evaluate_call=self._evaluate_call,
+            predict_call=predict_call,
+        )
+
+    def log_summary(self, summary: dict) -> None:
```
tssweeney, line 342: we should accumulate the unfinished predict_and_score calls and explicitly finish them here

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,355 @@
...
+            def evaluate(self: Evaluation, model: Model) -> None: ...
+
+            @weave.op(name="Evaluation.predict_and_score")
+            def predict_and_score(
+                self: Evaluation, model: Model, example: dict
+            ) -> dict:
+                predict_method = cast(Op, model.get_infer_method())
+                model_output, predict_call = predict_method.call(model, example)
+                current_predict_call.set(predict_call)
+                return {
```
tssweeney, line 291: might be good to not that this is thrown away

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,355 @@
...
+    # model_id: ID
+    predict_and_score_call: Call
+    evaluate_call: Call
+    predict_call: Call
+
+    _captured_scores: dict[str, ScoreType] = PrivateAttr(default_factory=dict)
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    def finish(self, *, scores: dict[str, ScoreType] | None = None) -> None:
```
tssweeney, line 151: what is the use of adding more scores here? This escapes the pattern of log_score

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,358 @@
...
```
tssweeney, line 151: Similar to "summarize" should we disallow adding scores after finish?

weave/flow/eval_imperative.py
andrewtruong: it was an escape hatch for testing

weave/flow/eval_imperative.py
andrewtruong: yes

weave/trace/op.py
```diff
@@ -514,6 +522,7 @@ def call(
     *args: Any,
     __weave: WeaveKwargs | None = None,
     __should_raise: bool = False,
+    __require_explicit_finish: bool = False,
```
tssweeney, line 525: can you comment on this intention of this param and warn about usage

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,371 @@
...
+        if self._logged_summary:
+            logger.warn("(NO-OP): Already called summary, returning.")
+            return
+        self._logged_summary = True
+        # Call the summarize method with the proper context
+        assert self._evaluate_call is not None
+        with _set_current_summary(summary):
+            self._pseudo_evaluation.summarize()
+
+        for pred in self._accumulated_predictions:
```
tssweeney, line 364: any reason you put this after the summary?

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,371 @@
...
+        assert self._evaluate_call is not None
+        pred = ImperativeScoreLogger(
+            predict_and_score_call=predict_and_score_call,
+            evaluate_call=self._evaluate_call,
+            predict_call=predict_call,
+        )
+        self._accumulated_predictions.append(pred)
+        return pred
+
+    def log_summary(self, summary: dict) -> None:
```
tssweeney, line 354: FUTURE: since we are in the business of accumulating, we could accumulate the scores and use our existing auto-summary to save the user from the hassle of summaries

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,371 @@
...
+    evaluate_call: Call
+    predict_call: Call
+
+    _captured_scores: dict[str, ScoreType] = PrivateAttr(default_factory=dict)
+    _has_finished: bool = PrivateAttr(False)
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    def finish(self) -> None:
+        scores = self._captured_scores
```
tssweeney, line 153: if has finished, warn/error?

weave/flow/eval_imperative.py
andrewtruong: re-ordered now, thanks

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,371 @@
...
+        wc.finish_call(
+            self.predict_and_score_call,
+            output={
+                "model_output": self.predict_call.output,
+                "scores": scores,
+                "model_latency": None,
+            },
+        )
+
+        self._has_finished = True
```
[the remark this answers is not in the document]
andrewtruong: I don't understand this comment

weave/flow/eval_imperative.py
andrewtruong: hmm yes, I guess we've gone this far we can accumulate for the user.

Agree to defer to a future PR

weave/flow/eval_imperative.py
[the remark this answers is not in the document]
tssweeney: you are accumulating `ImperativeScoreLogger` in memory which could just build and build. We should pop them off the list of _accumulated_predictions when finished so that we free that memory

weave/flow/eval_imperative.py
```diff
@@ -0,0 +1,408 @@
...
+            return value
+
+        raise TypeError("Unsupported type for casting")
+
+    return _convert_to_cls_inner
+
+
+def _cast_to_imperative_dataset(value: Dataset | list[dict] | str) -> Dataset:
+    if isinstance(value, str):
+        return Dataset(rows=weave.Table([{"dataset_id": value}]))
```
tssweeney, line 117: name here too if value is string?

PR #4008, opened 2025-03-31
fix(weave): Fix CrewAI tests

ayulockin:
## Description

fix the failing crewai tests. 

CrewAI updated their pydantic version from 2.10 -> 2.11. Handling the exclusion of the cross ref objs gracefully now.

note: You will see a removed line from the test. Earlier I was extracting one attribute from the cross referenced object which if we don't show it not a big deal. 

## Testing

How was this PR tested?


.github/workflows/test.yaml
```diff
@@ -238,7 +238,7 @@ jobs:
             "bedrock",
             "cerebras",
             "cohere",
-            # "crewai",
```
andrewtruong, line 241: Remove

andrewtruong (approved):

PR #4016, opened 2025-04-01
chore(app): Improve number input arrow hit box

ericakdiaz:
## Description

https://wandb.atlassian.net/browse/WB-24067

Tiny hit box and no visual feedback means clicking the arrows to step is difficult. Increased the size of the hitbox and added a background hover color. Also migrated the icon from semantic to our internal chevron icons. 

## Testing

How was this PR tested?

Before:

https://github.com/user-attachments/assets/c01febfe-b43a-43a4-9d36-8009e2e37439

After:

https://github.com/user-attachments/assets/0fb3ed5e-61bc-4a52-9252-1e83da7c94e3


nhusher (approved):

onx2:
> @onx2 are there accessibility things with the +/- widget that we should be thinking of? I looked through the list of ARIA roles on MDN and didn't see anything that fit well.

Yes, here is what the common component will have eventually:

- Customizable formatting – Support for internationalized number formatting and parsing including decimals, percentages, currency values, and units. Multiple currency formats are supported, including symbol, code, and name in standard or accounting notation.
- International – Support for the Latin, Arabic, Devanagari, Bengali, and Han positional decimal numbering systems in [over 30 locales](https://react-spectrum.adobe.com/react-aria/internationalization.html#supported-locales). The numbering system is automatically detected from user input, and input method editors such as Pinyin are supported.
- Validation – Keyboard input is validated as the user types so that only numeric input is accepted, according to the locale and numbering system. Values are automatically rounded and clamped according to configurable decimal, minimum, maximum, and step values. Custom client and server-side validation is also supported.
- Mobile friendly – An appropriate software keyboard is automatically selected based on the allowed values for ease of use.
- Accessible – Follows the [spinbutton](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/) ARIA pattern, with support for arrow keys, scroll wheel, and stepper buttons to increment and decrement the value. Extensively tested across many devices and [assistive technologies](https://react-spectrum.adobe.com/react-aria/accessibility.html#testing) to ensure announcements and behaviors are consistent.

[React Aria blog post about it](https://react-spectrum.adobe.com/blog/how-we-internationalized-our-numberfield.html)

PR #4021, opened 2025-04-01
chore(weave): refactor and improve calls query string optimization 

gtarpenning:
## Description

Refactor the current scheme that handling applying string optimizations in the calls query builder. This should unlock more easily adding further optimizations in the future, like ID datetime filtering. 

This pr:
- uses recursive parsing to handle deeply nested string comparison queries. the existing implementation does not handle this case correctly, and thus is technically broken for nested string filters. 
- adds a comprehensive test
- fixes sqlite implementation to allow more than 2 statements in an or condition

## Testing

- existing end to end tests + new nested filter test
- updated unit tests with a nested example

And in qa:

[proj w/ filter](https://qa.wandb.ai/weave-team/wtypes/weave/traces?filters=%7B%22items%22%3A%5B%7B%22id%22%3A0%2C%22field%22%3A%22started_at%22%2C%22operator%22%3A%22%28date%29%3A+after%22%2C%22value%22%3A%222025-03-27T22%3A38%3A02.413Z%22%7D%2C%7B%22id%22%3A1%2C%22field%22%3A%22output.float%22%2C%22operator%22%3A%22%28number%29%3A+%3C%22%2C%22value%22%3A55.402773825554895%7D%2C%7B%22id%22%3A2%2C%22field%22%3A%22output.message%22%2C%22operator%22%3A%22%28string%29%3A+contains%22%2C%22value%22%3A%22hell%22%7D%2C%7B%22id%22%3A3%2C%22field%22%3A%22output.no%22%2C%22operator%22%3A%22%28bool%29%3A+is%22%2C%22value%22%3Afalse%7D%5D%2C%22logicOperator%22%3A%22and%22%7D)

Screenshot 2025-04-03 at 3 38 46 PM


tests/trace_server/test_calls_query_builder.py
```diff
@@ -1012,8 +1005,9 @@ def test_calls_query_with_or_between_start_and_end_fields() -> None:
             any(calls_merged.inputs_dump) AS inputs_dump,
             any(calls_merged.output_dump) AS output_dump
         FROM calls_merged
-        WHERE
-            calls_merged.project_id = {pb_4:String}
+        WHERE calls_merged.project_id = {pb_6:String}
+            AND (((calls_merged.inputs_dump LIKE {pb_4:String} OR calls_merged.inputs_dump IS NULL)
```
tssweeney, line 1009: so this was completely incorrect before?

gtarpenning: We basically flattened nested conditions, no we retain the nested structure, which requires checking null at every reference @tssweeney

tests/trace_server/test_calls_query_builder.py
```diff
@@ -1369,44 +1362,6 @@ def test_calls_query_with_combined_like_optimizations_and_op_filter() -> None:
     )
 
 
-def test_calls_query_filter_by_empty_str() -> None:
```
tssweeney, line 1372: why remove?

tests/trace_server/test_calls_query_builder.py
```diff
@@ -1100,21 +1100,18 @@ def test_calls_query_with_complex_heavy_filters() -> None:
...
         WHERE
             calls_merged.project_id = {pb_1:String}
-        AND
-            (calls_merged.id IN filtered_calls)
-        AND (((calls_merged.inputs_dump LIKE {pb_10:String} OR calls_merged.inputs_dump IS NULL))
-        AND (((JSON_VALUE(calls_merged.inputs_dump, {pb_4:String}) > {pb_5:UInt64}) OR calls_merged.inputs_dump IS NULL))
-        AND ((calls_merged.output_dump LIKE {pb_11:String} OR calls_merged.output_dump IS NULL))
-        AND ((lower(calls_merged.output_dump) LIKE {pb_12:String} OR calls_merged.output_dump IS NULL)))
+          AND (calls_merged.id IN filtered_calls)
+          AND (
```
tssweeney, line 1104: reduction of conditions here is intentional?

tests/trace_server/test_calls_query_builder.py
gtarpenning: oh, sorry, in this case we just ejected when we saw an OR operation, so we _dont_ do the optimization currently. Now we can do the optimization!

weave/trace_server/calls_query_builder.py
```diff
@@ -784,7 +786,7 @@ def _as_sql_base_format(
         {feedback_where_sql}
         {id_mask_sql}
         {id_subquery_sql}
-        {call_parts_predicate_filters_sql}
+        {str_filter_opt_sql}
```
tssweeney, line 789: @chance-wnb just modified this code path. Please update with master before merging

tests/trace_server/test_calls_query_builder.py
gtarpenning: The reduction is technically a regression, we are no longer doing pre-having optimizations for integer comparisons. I'll make a ticket to add it back. The current impl focuses on strings, which dominates the current bad-query numbers.

weave/trace_server/calls_query_builder.py
gtarpenning: I think those changes should be in this version, just merged again to be safe (and got that print statement)

weave/trace_server/calls_query_builder.py
```diff
@@ -1353,7 +1255,150 @@ def _create_like_optimized_in_condition(
...
+    if _field_requires_null_check(field):
+        return f"({or_sql} OR {table_alias}.{field} IS NULL)"
+    return or_sql
+
+
+class OptimizationConditions(BaseModel):
+    str_filter_opt_sql: Optional[str] = None
+
+
+def process_query_to_optimization_sql(
```
tssweeney, line 1268: reasoning in the abstract here, it seems there are some conditions which are not processable (return None). let's call those `unprocessable_operand`. In that case, if you have a condition: `unprocessable_operand OR processable_operand`, then the predicate filter will filter only to values of `processable_operand`. But if the `unprocessable_operand` is TRUE for some filtered subset, then this would be overly restrictive. Does that make sense?

weave/trace_server/calls_query_builder.py
tssweeney: 👍

tssweeney (commented):
comments

weave/trace_server/calls_query_builder.py
gtarpenning: That does make sense, hmmm. We could return True?

tssweeney: I think we need a test like this. Basically the nested resolution allows ORs to be split up in such a way that value results get filtered out if one part of the OR is not processable. I think you have to say: If any part of an OR is not processable, the entire OR is not processable

gtarpenning: I'll add that test! See some red in CI, then fix it

gtarpenning: wait @tssweeney  I already did that on line 1307:

```
for op in operation.or_:
    result = process_operand(op)
    if result is None:
        return None  # If any operand can't be optimized, the whole OR can't be optimized <--------
    or_conditions.append(result)
```

gtarpenning: Ideally we would fallback to the normal processing fn if we ever can't optimize

tssweeney: oh nice!, yup that is perfect

tssweeney (approved):

PR #4029, opened 2025-04-02
chore(weave): Fix type generation to replace circular references with Anys

tssweeney:
When generating zod types, circular references cause issues. This PR identifies circular references and fixes them before generation occurs. More advanced replacement could be designed in the future if desired.

This unblocks saved views using Query in the type definition.

I explored more creative ways for unrolling nested structures, but they got overly complicated. Can always make this better in the future.

### DIFF
--- a/weave-js/scripts/generate-schemas.sh
+++ b/weave-js/scripts/generate-schemas.sh
@@ -3,13 +3,16 @@
 # Exit on error
 set -e
 
-SCHEMA_INPUT_PATH="../weave/trace_server/interface/builtin_object_classes/generated/generated_builtin_object_class_schemas.json"
-SCHEMA_OUTPUT_PATH="./src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/generatedBuiltinObjectClasses.zod.ts"
+SCHEMA_PATH="../weave/trace_server/interface/builtin_object_classes/generated/generated_builtin_object_class_schemas.json"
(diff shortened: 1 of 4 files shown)


weave-js/scripts/generate-schemas.sh
```diff
@@ -3,13 +3,16 @@
...
 
-SCHEMA_INPUT_PATH="../weave/trace_server/interface/builtin_object_classes/generated/generated_builtin_object_class_schemas.json"
-SCHEMA_OUTPUT_PATH="./src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/generatedBuiltinObjectClasses.zod.ts"
+SCHEMA_PATH="../weave/trace_server/interface/builtin_object_classes/generated/generated_builtin_object_class_schemas.json"
+OUTPUT_PATH="./src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/generatedBuiltinObjectClasses.zod.ts"
 
 echo "Generating schemas..."
 
+# First resolve circular references in the schema
+python3 ./resolve_schema_circular_refs.py "$SCHEMA_PATH"
```
jamie-rasmussen, line 12: This doesn't work when run from `make synchronize-base-object-schemas`  because the working directory will be `weave-public/weave-js`
```suggestion
python3 "$(dirname "$0")/resolve_schema_circular_refs.py" "$SCHEMA_PATH"
```

jamie-rasmussen (approved):

weave-js/scripts/generate-schemas.sh
tssweeney: Fixed! I thanks - I changed that right before pushing.

PR #4035, opened 2025-04-02
perf(weave): use uuidv7 timestamps to filter datetime conditions

gtarpenning:
## Description

[WB-24305](https://wandb.atlassian.net/browse/WB-24305)

This pr implements a query optimization using uuid7s, which are already indexed, as a pre-groupby datetime filter. This applies to all calls query requests (`calls_query_stats`, `calls_query`, `calls_query_stream`) that include a datetime filter. We now convert the timestamps provided into their equivalent uuid7s, which are indexed and sorted, massively decreasing the amount of granules that must be read into memory. This can occur before the groupby, because both call starts and ends have an id (the same id). 

What about ids that aren't uuidv7? We handle this case by adding a filter mask of `'ffffffffffffffff'` which is the max 8 byte id. 

Example stats query in prod: 
```sql
SELECT count() as c
FROM
(
    SELECT calls_merged.id AS id
    FROM calls_merged
    WHERE calls_merged.project_id = 'project_id'
    HAVING (any(calls_merged.started_at) > _CAST(1742472000, 'UInt64')) AND (NOT (any(calls_merged.started_at) > _CAST(1742475600, 'UInt64'))) AND (any(calls_merged.deleted_at) IS NULL) AND (NOT (any(calls_merged.started_at) IS NULL))
)
```

Example stats query with optimization:
```sql
SELECT count() as c
FROM
(
    SELECT calls_merged.id AS id
    FROM calls_merged
    WHERE calls_merged.project_id = 'project_id'
        AND (calls_merged.id <= 'ffffffffffffffff' OR
        (AND calls_merged.id > '0195b36c-b200-7000-8000-000000000000' AND calls_merged.id < '0195b3a3-a080-7000-8000-000000000000'))
    HAVING (any(calls_merged.started_at) > _CAST(1742472000, 'UInt64')) AND (NOT (any(calls_merged.started_at) > _CAST(1742475600, 'UInt64'))) AND (any(calls_merged.deleted_at) IS NULL) AND (NOT (any(calls_merged.started_at) IS NULL))
)
```

## Testing

add unit tests

qa-azure: 

prod stats query in big project with filter: [10 seconds](https://us5.datadoghq.com/apm/traces?query=env%3Amanaged-install%20operation_name%3Afastapi.request%20service%3Aweave-trace%20%40weave_trace_server.auth_scope.email%3Agriffin.tarpenning%40wandb.com&agg_m=count&agg_m_source=base&agg_t=count&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code%2C%40_span.count%2C%40_duration.by_service%2C%40weave_trace_server.auth_scope.email&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&refresh_mode=sliding&shouldShowLegend=true&sort=time&spanID=9852347362044463913&spanType=all&spanViewType=logs&storage=hot&timeHint=1743113551411&trace=67e5cd44000000002fed8ea33b1a5c919852347362044463913&traceID=67e5cd44000000002fed8ea33b1a5c91&view=spans&start=1743112662057&end=1743113562057&paused=false)

branch stats query in big project with filter: [650 ms](https://us5.datadoghq.com/apm/traces?query=env%3Amanaged-install%20operation_name%3Afastapi.request%20service%3Aweave-trace%20%40weave_trace_server.auth_scope.email%3Agriffin.tarpenning%40wandb.com%20-resource_name%3A%22POST%20%2Ftraces%2Ffeedback%2Fquery%22&agg_m=count&agg_m_source=base&agg_t=count&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code%2C%40_span.count%2C%40_duration.by_service%2C%40weave_trace_server.auth_scope.email&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&refresh_mode=sliding&shouldShowLegend=true&sort=time&spanID=14711125143075388032&spanType=all&spanViewType=metadata&storage=hot&timeHint=1743113262593&trace=67e5cc2d000000005d5baad8d59273f414711125143075388032&traceID=67e5cc2d000000005d5baad8d59273f4&view=spans&start=1743112740300&end=1743113640300&paused=false) 

Can be even better. In a specific high load dedicated server environment, string filter counts: 

| stat      | prod      | branch   |
|-----------|-----------|----------|
| mem       | 17GB      | 91MB     |
| rows read | 34,673,371| 76,744   |
| time      | 12.5 s    | 0.026 s  |

[WB-24305]: https://wandb.atlassian.net/browse/WB-24305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/calls_query_builder.py
```diff
@@ -1378,29 +1445,132 @@ def process_operand(operand: tsi_query.Operand) -> Optional[str]:
...
+
+    if not processed.str_filter_opt_sql and not processed.id_datetime_filters_sql:
         return OptimizationConditions()
 
-    processed.str_filter_opt_sql = "AND " + processed.str_filter_opt_sql
+    if processed.str_filter_opt_sql:
+        processed.str_filter_opt_sql = "AND " + processed.str_filter_opt_sql
+    if processed.id_datetime_filters_sql:
+        # Create non-uuidv7 condition
+        non_uuidv7_condition = f"{table_alias}.id <= 'ffffffffffffffff'"
```
tssweeney, line 1466: I think this `non-uuid7` condition should be appended inside of the `_create_datetime_optimization_sql` procedure so this caller does not need to know about those details

gtarpenning: I agree its ugly, I can play around with it. The reason its there is that if you have multiple id conditions, like "x > 111 and y < 1111" we only ever want one `id < ffffffff`. Thats fine for this example, but what if the datetime condition is deep in a nested and + or + in stack we still just only want the one condition at the top (if we ever touch an id pre-where, always always let these kinds of ids pass). I will figure out how to make it prettier, the recursion is annoying hmm.

weave/trace_server/calls_query_builder.py
```diff
@@ -1281,56 +1297,90 @@ def process_query_to_optimization_sql(
...
-    if not filterable_conditions:
+    if not conditions:
         return OptimizationConditions()
 
     def process_operation(
         operation: tsi_query.Operation,
     ) -> Optional[OptimizationConditions]:
         if isinstance(operation, tsi_query.OrOperation):
-            or_conditions = []
+            str_conditions = []
```
tssweeney, line 1310: I think this is fine, but given that you have a second use case now, i cold see a slightly more elegant function here for `process_operation`. Instead of hardcoding the str and datetime conditions, you could instead have a mapper/visitor pattern that allows the caller to specify the conditions to accumulate

weave/trace_server/calls_query_builder.py
```diff
@@ -1281,56 +1297,90 @@ def process_query_to_optimization_sql(
...
             for op in operation.or_:
                 result = process_operand(op)
+                # If any operand can't be optimized, the whole OR can't be optimized
                 if result is None:
-                    return None  # If any operand can't be optimized, the whole OR can't be optimized
-                or_conditions.append(result)
-            if not or_conditions:
-                return None
+                    return None
+                if not result.str_filter_opt_sql and not result.id_datetime_filters_sql:
```
tssweeney, line 1318: i think this is subtly buggy. You could hit a case where strs are not processable, but ids are, in which case you continue down the tree. Then it flips back and forth. The total condition would not be processess-able, but this routine would allow it. The previous short circuiting worked because you only had one applicable condition to check.

tssweeney (commented):
I think there is a subtle bug here. I could be wrong - let's analyze and go from there

weave/trace_server/calls_query_builder/optimization_builder.py
```diff
@@ -0,0 +1,448 @@
...
+    elif isinstance(operation, tsi_query.GtOperation):
+        return processor.process_gt(operation)
+    elif isinstance(operation, tsi_query.GteOperation):
+        return processor.process_gte(operation)
+    return None
+
+
+class OptimizationConditions(BaseModel):
+    str_filter_opt_sql: Optional[str] = None
+    id_datetime_filters_sql: Optional[str] = None
```
gtarpenning, line 148: could correlate these with the classes. Like one handler that registers the class and calls finalize

weave/trace_server/calls_query_builder/optimization_builder.py
```diff
@@ -0,0 +1,448 @@
...
+    else:
+        return None
+
+    # Return if literal isn't a string
+    if not isinstance(literal_operand, tsi_query.LiteralOperation) or not isinstance(
+        literal_operand.literal_, str
+    ):
+        return None
+
+    from weave.trace_server.calls_query_builder.calls_query_builder import (
```
gtarpenning, line 235: unfortunate but a really annoying refactor otherwise

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -49,7 +49,7 @@
 from weave.trace_server import refs_internal as ri
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.actions_worker.dispatcher import execute_batch
-from weave.trace_server.calls_query_builder import (
+from weave.trace_server.calls_query_builder.calls_query_builder import (
```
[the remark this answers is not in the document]
gtarpenning: It was not. Just organizing.

tssweeney (approved):

gtarpenning:
This does not work, and is a hack. Closing to prefer using an indexed datetime field

PR #4037, opened 2025-04-03
fix(weave): Loosen the regex check to accept potentially CJK characters

chance-wnb:
## Description

- Fixes WB-23821
- Fixes WB-21343

There's no point in restricting artifact names to only ASCII characters.

If we need these checks, they should be handled at the server or SDK level (during creation).

Since users often use CJK characters for dataset names and other objects without issues, I'm temporarily relaxing the frontend regex restrictions.

## Testing

I tested the scenario of 

1. Getting the dataset details which uses CJK characters

image

2. Compare evaluations using a dataset with CJK characters
image

These scenarios were previously broken without this change.

### DIFF
--- a/weave-js/src/react.tsx
+++ b/weave-js/src/react.tsx
@@ -516,7 +516,7 @@ const RE_WEAVE_OBJECT_REF_PATHNAME = new RegExp(
     '/',
     '(object|op)', // Weave kind
     '/',
-    '([a-zA-Z0-9-_/. ]{1,128})', // Artifact name
+    '([^\\#?%:]{1,128})', // Artifact name
     ':',
     '([*]|[a-zA-Z0-9]+)', // Artifact version, allowing '*' for any version
     '/?', // Ref extra portion is optional


weave-js/src/react.tsx
```diff
@@ -516,7 +516,7 @@ const RE_WEAVE_OBJECT_REF_PATHNAME = new RegExp(
     '/',
     '(object|op)', // Weave kind
     '/',
-    '([a-zA-Z0-9-_/. ]{1,128})', // Artifact name
+    '(.{1,128})', // Artifact name
```
neutralino1, line 519: Wasn't the original regex also preventing punctuation and special characters? e.g. `$#%&{}[]()` etc.
Also whitespaces too?

chance-wnb: Good catch! I will fix

chance-wnb: Quite interesting, while I was trying to replicate a case an invalid character is used. I can't. 

The reason being that in the [SDK code](https://github.com/wandb/weave/blob/8e04058a7fe32d6388aec57537bce2782857e702/weave/trace/weave_client.py#L2355), we already sanitize the object names. (In its implementation, unicode characters can be perserved)

So the chance that we could get an object name stored with invalid characters is low. 

Therefore, for this regex, I basically copied what's been done for the `const PATTERN_PROJECT = '([^\\#?%:]{1,128})'; // Project name`, I added an extra `/` exclusion. That ought to be sufficient.

neutralino1 (approved):
SDK/back-end parity. I like that! 🤠 

Thank you!

PR #4044, opened 2025-04-03
chore(weave): convert 12m to 1y

gtarpenning:
## Description

[WB-24241](https://wandb.atlassian.net/browse/WB-24241)

Convert anything thats a nice even (ish) multiple of a year into `1y` in the timestamp filter bar.

## Testing

prod:
Screenshot 2025-04-03 at 1 59 07 PM

branch:
Screenshot 2025-04-03 at 1 58 52 PM

[WB-24241]: https://wandb.atlassian.net/browse/WB-24241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/Timestamp.tsx
+++ b/weave-js/src/components/Timestamp.tsx
@@ -43,17 +43,16 @@ const formatSmallTime = (then: moment.Moment): string | null => {
     return `${years}y`;
   }
 
-  // Calculate months using moment's built-in month diff
   const monthDiff = now.diff(then, 'months');
-
-  // Get remaining days by moving forward the months and checking what's left
-  const afterMonths = then.clone().add(monthDiff, 'months');
-  const remainingDays = now.diff(afterMonths, 'days');
 
   // Always show months if 3 or more
   if (monthDiff >= 3) {
     return `${monthDiff}mo`;
   }
+
+  // Get remaining days by moving forward the months and checking what's left
+  const afterMonths = then.clone().add(monthDiff, 'months');
+  const remainingDays = now.diff(afterMonths, 'days');
 
   // Show months if exact multiple
   if (monthDiff > 0 && remainingDays === 0) {
@@ -71,7 +70,7 @@ const formatSmallTime = (then: moment.Moment): string | null => {
     return `${weeks}w`;
   }
 
-  // Otherwise use days for more precision
+  // Otherwise use days or more precise units
   if (days >= 1) {
     return days === 1 ? '1d' : `${days}d`;
   } else if (hours >= 1) {
--- a/weave-js/src/components/Timestamp.tsx
+++ b/weave-js/src/components/Timestamp.tsx
@@ -55,8 +55,8 @@ const formatSmallTime = (then: moment.Moment): string | null => {
     return `${monthDiff}mo`;
   }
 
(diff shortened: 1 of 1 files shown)


weave-js/src/components/Timestamp.tsx
```diff
@@ -38,6 +38,18 @@ const formatSmallTime = (then: moment.Moment): string | null => {
...
   const days = Math.floor(diffMs / (1000 * 60 * 60 * 24));
 
+  const years = now.diff(then, 'years');
+  if (years > 0) {
+    const remainingDays = days - years * 365;
+    if (remainingDays < 7) {
+      if (years === 1) {
+        return '1y';
+      } else if (years > 1) {
+        return `${years}y`;
```
jamie-rasmussen, line 48: Doesn't this return work for the years === 1 case too?

weave-js/src/components/Timestamp.tsx
```diff
@@ -38,6 +38,18 @@ const formatSmallTime = (then: moment.Moment): string | null => {
   const hours = Math.floor(diffMs / (1000 * 60 * 60));
   const days = Math.floor(diffMs / (1000 * 60 * 60 * 24));
 
+  const years = now.diff(then, 'years');
+  if (years > 0) {
+    const remainingDays = days - years * 365;
+    if (remainingDays < 7) {
```
jamie-rasmussen, line 44: I don't think I'm following the logic here - if remainingDays was >= 7 wouldn't we still want to show a year diff?

gtarpenning: we would show 53 weeks instead.

gtarpenning: idk, I sort of expect Martin to take a pass at this when he gets back

weave-js/src/components/Timestamp.tsx
gtarpenning: ... yes

weave-js/src/components/Timestamp.tsx
gtarpenning: I guess we dont need that level of fidelity, but what about 13 months vs 12 months?

jamie-rasmussen: I think some more comments or unit tests would be good - this function doesn't match my intuition, which is that the older something is the less I care about a precise number. What we get now is:
formatSmallTime(moment("2025-01-01")) => 3mo
formatSmallTime(moment("2024-01-01")) => 459d
formatSmallTime(moment("2021-01-01")) => 222w
And that feels wrong.

gtarpenning: hahahah great breakdown, i'm convinced. fixing

gtarpenning: Screenshot 2025-04-04 at 12 53 21 PM
Screenshot 2025-04-04 at 12 53 32 PM
Screenshot 2025-04-04 at 12 53 43 PM

Okay, its simpler, and more intuitive now

jamie-rasmussen (approved):

PR #4048, opened 2025-04-03
feat(app): Create a generic stepper panel for any incoming expressions that return a list type

nicholaspun-wandb:
## Description

- Work branched from [WB-24167](https://wandb.atlassian.net/browse/WB-24167)
- Creates a new stepper panel type that generalizes the work done with the runs history stepper. 
    - This new panel will work on any expression, including `runs.history`, which was the original ask. The `runs` behaviour is unaffected
    - The user can select the property key that they wish to step through, as well as the panel type they wish to render the output with.
- The existing runs history table stepper and plot stepper are unaffected by this PR. They can probably be removed, but we may want to keep the spec and ID and just pass it through to this new stepper component (so that user workspaces are unaffected)

## Screenshots

Existing behaviour with tables stepping is matched

https://github.com/user-attachments/assets/2ad3724b-173e-4779-bc83-6ac3d80425e7

Existing behaviour with plots stepping is matched

https://github.com/user-attachments/assets/a1bcde6a-ffd0-4b6f-a501-f92b0ef1ff46

User can use any expression

https://github.com/user-attachments/assets/770856ab-c650-4dcb-b9b7-d57923b94ba8

User can change the panel type

https://github.com/user-attachments/assets/297bc195-e40f-4c73-bd03-10924a17cbec

User can step through the result type as-is

https://github.com/user-attachments/assets/5d8fd5a1-5e1b-47ab-98d8-4484c0bc6f12

## Testing

Local FE Invoker

[WB-24167]: https://wandb.atlassian.net/browse/WB-24167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nicholaspun-wandb:
Beta link: https://beta.wandb.ai/home?betaVersion=846a65c7579c239c9e56ec50496d2a66039881b9

nicholaspun-wandb:
Beta link: https://beta.wandb.ai/home?betaVersion=a694867a6c9d67edeb29cc0de7f607940425a671

weave-js/src/components/Panel2/PanelStepper/component.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  useEffect(() => {
+    if (stepsNodeLoading || stepsNodeResult == null) {
+      return;
+    }
+
+    const newSteps: number[] = [...new Set<number>(stepsNodeResult)].sort(
+      (a, b) => a - b
+    );
+
```
domphan-wandb, line 36: I got a panel crash when going from a Combined Table (`runs.summary["table"]`) > Stepper. Is it happening on your end too?

```
UseNodeValueServerExecutionError: Node execution failed (useNodeValue): Cannot dispatch op "mapped_typedDict-pick"; ops [<OpDef: list-__getitem__ None>] matched first arg type, but no matching ops found for remaining arg types: [Const(String(), _step)]
```

nicholaspun-wandb: Ok, I've added an empty state (we'll need it later anyway in the case that no custom step metrics are found)

nicholaspun-wandb:
Updated beta link (will need a couple minutes to build): http://beta.wandb.ai/home?betaVersion=2f5c9643ba6a0a00afbcd8095e7b9ab3f9a8ae54

domphan-wandb (approved):

PR #4051, opened 2025-04-04
feat(weave): Display trace storage size in Trace view

chance-wnb:
## Description

- Fixes WB-23748

Add storage size column to CallsTable

Changes:
- Added storage size column to display total storage usage for each call
- Implemented loading state handling with LoadingDots component
- Added proper null checking for storage size values

The storage size column shows human-readable byte sizes (e.g., "1.5 MB") and handles loading states appropriately.

## Screenshot

image

## Design doc

Please also refer this [mini design doc(private)](https://www.notion.so/wandbai/Data-Size-display-1bde2f5c7ef38087bc03cb8accb95bfb?pvs=4#1cfe2f5c7ef3805e9f73f9ddc4575460) for greater context.

## Performance Study

At the beginning of implementing this PR, I debated myself how to load the data for this additional column.

The `/calls/stream_query` API  is capable of either:
1: accepting a group (about 50) of call_ids, and do table joins to calculated an aggregated result for each of the row_ids.
2: or it can accept a single row_id, and do the same expensive calculation of an aggregated result for that single row_id.

Given that the table join and the aggregation is expensive, I need to decide whether I make a single call with a group of ids, or invoke each id and have parallel requests on the frontend. 

If I do parallel requests, the code structure would be simpler, because each frontend component (a ColumnCellRenderer) can encapsulate the data loading logic within itself.

Doing a single request with a group ids might be harder to implement, because I have to centralize the data loading at one place and dispatch the results separately.

To make a sensible decision, during the first several hours, instead of writing the feature code, I did some study and measurement, here is the result:

```
Query Performance Analysis:
callsTableQuery.ts:180 

Batch Size: 1
callsTableQuery.ts:181 Average Total Time: 67.10ms
callsTableQuery.ts:182 Average DB Time: 17.60ms
callsTableQuery.ts:183 Average Network Time: 49.50ms
callsTableQuery.ts:184 Average Memory Usage: 0.04MB
callsTableQuery.ts:189 Error Rate: 0.00%
callsTableQuery.ts:180 

Batch Size: 10
callsTableQuery.ts:181 Average Total Time: 56.16ms
callsTableQuery.ts:182 Average DB Time: 16.20ms
callsTableQuery.ts:183 Average Network Time: 39.96ms
callsTableQuery.ts:184 Average Memory Usage: 0.03MB
callsTableQuery.ts:189 Error Rate: 0.00%
callsTableQuery.ts:180 

Batch Size: 50
callsTableQuery.ts:181 Average Total Time: 56.20ms
callsTableQuery.ts:182 Average DB Time: 15.80ms
callsTableQuery.ts:183 Average Network Time: 40.40ms
callsTableQuery.ts:184 Average Memory Usage: 0.07MB
callsTableQuery.ts:189 Error Rate: 0.00%
callsTableQuery.ts:180 

Batch Size: 100
callsTableQuery.ts:181 Average Total Time: 58.64ms
callsTableQuery.ts:182 Average DB Time: 15.60ms
callsTableQuery.ts:183 Average Network Time: 43.04ms
callsTableQuery.ts:184 Average Memory Usage: 0.13MB
callsTableQuery.ts:189 Error Rate: 0.00%
```
From these measurement, it can be seen that the DB query time is not significantly impacted by the size of IDs. At least within 100 number of ids, the query time is roughly un-impacted. 

On the other hand, network transportation takes the major portion of the time. 

Therefore, parallelizing request makes little sense in terms of UX. And centralizing the processing with a group of ID is the wiser move. 

## Testing

The PR is currently manually tested.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -265,6 +265,8 @@ export const CallsTable: FC<{
     [expandedRefCols]
   );
 
+  const shouldIncludeTotalStorageSize = effectiveFilter.traceRootsOnly;
+
   // Fetch the calls
   const calls = useCallsForQuery(
     entity,
@@ -273,7 +275,13 @@ export const CallsTable: FC<{
     filterModelResolved,
     paginationModelResolved,
     sortModelResolved,
-    expandedRefCols
+    expandedRefCols,
+    undefined,
+    {
+      // The total storage size only makes sense for traces,
+      // and not for calls.
+      includeTotalStorageSize: shouldIncludeTotalStorageSize,
+    }
   );
 
   // Here, we only update our local state once the calls have loaded.
@@ -445,7 +453,10 @@ export const CallsTable: FC<{
     columnIsRefExpanded,
     allowedColumnPatterns,
     onAddFilter,
-    calls.costsLoading
+    calls.costsLoading,
+    shouldIncludeTotalStorageSize,
+    shouldIncludeTotalStorageSize ? calls.storageSizeResults : null,
+    shouldIncludeTotalStorageSize && calls.storageSizeLoading
   );
 
   // This contains columns which are suitable for selection and raw data
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
@@ -51,13 +51,18 @@ export const useCallsForQuery = (
   gridPage: GridPaginationModel,
   gridSort?: GridSortModel,
   expandedColumns?: Set<string>,
-  columns?: string[]
+  columns?: string[],
+  options?: {
+    includeTotalStorageSize?: boolean;
+  }
 ): {
   costsLoading: boolean;
   result: CallSchema[];
   loading: boolean;
   total: number;
   refetch: () => void;
+  storageSizeLoading: boolean;
+  storageSizeResults: Map<string, number> | null;
 } => {
   const {useCalls, useCallsStats} = useWFHooks();
   const effectiveOffset = gridPage?.page * gridPage?.pageSize;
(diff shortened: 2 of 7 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
```diff
@@ -147,11 +183,15 @@ export const useCallsForQuery = (
...
         refetch,
+        storageSizeLoading: storageSize.loading,
+        storageSizeResults: null,
       };
     }
 
     return {
       costsLoading: costs.loading,
+      storageSizeLoading: storageSize.loading,
+      storageSizeResults,
```
chance-wnb, line 194: I am emitting this result as a map separately. I don't want it to participate the merging with calls and costCalls. I think the cell renderer can very easily access this info, which renders a 3-way merge with `callResults` and `costResults` unnessary.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
```diff
@@ -130,6 +135,37 @@ export const useCallsForQuery = (
     }
   );
 
+  const storageSizeCols = useMemo(() => ['id', 'total_storage_size_bytes'], []);
+  const storageSizeFilter = costFilter;
+
+  const storageSize = useCalls(
```
tssweeney, line 141: I would recommend trying to merging this storage second query with the cost second query. This would allow you to inly have a single second query

tssweeney: I see the comment below indicates you did not want to do this (perhaps it was slower?)

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -392,6 +401,7 @@ export const CallsTable: FC<{
         }
       : undefined;
 
+  const shouldIncludeTotalStorageSize = pageType === PageType.Traces;
```
tssweeney, line 404: I think this `PageType` is not exactly necessary or correct actually. I think what you want to do is look at the effective filter. `shouldIncludeTotalStorageSize = effectiveFilter.traceRootsOnly && !isEvaluation`

tssweeney (commented):
I think you can remove the PageType calculation and determine if you show the column purely based on the filter. Other than that, happy to approve.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
chance-wnb: That's entirely true. Thanks for the suggestion! I adopted it and feel that it s much more robust.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -265,6 +269,35 @@ export const CallsTable: FC<{
...
+  const shouldIncludeTotalStorageSize = useMemo(() => {
+    // if it's an evaluation table, then trace size should not be shown
+    if (isEvaluationTable) {
+      return false;
+    }
+    if (effectiveFilter.traceRootsOnly) {
+      return true;
+    }
+    // if there is a single op version ref, then trace size should not be shown
+    if (effectiveFilter.opVersionRefs?.length === 1) {
```
tssweeney, line 287: isn't your total size only valid for trace roots? `include_total_storage_size` is only accurate for cases where `traceRootsOnly=true`

chance-wnb: Thanks for catching that! 

Yes, I tested that even under the evaluation scenario, `effectiveFilter.traceRootsOnly` isn't set. So it seems safe to solely rely on `effectiveFilter.traceRootsOnly`. I have made it that way.

tssweeney (approved):
For completeness, I think you are now going to see sizes for evals (which might be ok)

PR #4060, opened 2025-04-07
chore(app): Add more instrumentation for query panel analytics

domphan-wandb:
## Description

Internal JIRA: [WB-24278](https://wandb.atlassian.net/browse/WB-24278)

This PR adds instrumentation to answer the following:
1. How often is the filter being opened and then used or not used
    a. Add instrumentation for OPEN_FILTER / CLOSE_FILTER events
    b. Add instrumentation for FILTER_DISCARD_CHANGES and REMOVE_FILTER events
2. How long does it take to get to the first query
    a. Add instrumentation for editor focus
    b. This will be used In combination with event from flushing the expression
3. How many invalid expression submissions are attempted
    a. Add instrumentation to invalid WeaveExpression enter key events 
5. How much time via clicks are users are spending on query panels
    a. Add dd action for qp click instrumentation

This PR also overhauls DD RUM by naming actions via the prop `data-dd-action-name.` When set on a clickable element, it'll use that to label unnamed click events happening in child elements.
* I think from here, we can further add more specific action names for different things like click next page, or resizing columns, etc.

[WB-24278]: https://wandb.atlassian.net/browse/WB-24278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gtarpenning (approved):
might be a bit noisy but lgtm. I'd test in qa if you haven't. Also you might want to port that old analytics stuff that Nick Pen made to the new form. Example: 
```
import {Analytics} from '../services/analytics';
Analytics.track('Add Organization Member Clicked', {
      entityNames: teams,
      organizationName: organization?.name,
      location: analyticsLocation,
    });
```

domphan-wandb:
> might be a bit noisy but lgtm. I'd test in qa if you haven't. Also you might want to port that old analytics stuff that Nick Pen made to the new form. Example:
> 
> ```
> import {Analytics} from '../services/analytics';
> Analytics.track('Add Organization Member Clicked', {
>       entityNames: teams,
>       organizationName: organization?.name,
>       location: analyticsLocation,
>     });
> ```

Thanks @gtarpenning ! Which events do you think would be noisy? I'd like to see if I can tweak them or find another event that'd achieve the same thing without the noise.

Also captured a [jira](https://wandb.atlassian.net/browse/WB-24313) for porting over analytics

PR #4066, opened 2025-04-07
fix(app): wider run description field

ArtsiomWB:
## Description
This PR Makes it so the description field is not as narrow as it currently is. At the moment, the current behavior when trying to edit the run description is to make the EditField narrow, every time you are in the edit state:

https://github.com/user-attachments/assets/80a0b346-5c81-4221-8ce3-41c6f6dbe341

With these changes, the EditField will enter the edit state remembering the current width of the current description string and will not enter the edit state as a narrow field again

https://github.com/user-attachments/assets/71bd432d-54be-45ed-b99a-f214ffee763c

- [Fixes WB-16773](https://wandb.atlassian.net/browse/WB-16773)

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/common/components/EditableField.tsx
+++ b/weave-js/src/common/components/EditableField.tsx
@@ -43,6 +43,7 @@ interface EditableFieldState {
   editing: boolean;
   origValue: string;
   currentValue: string;
+  fieldWidth?: number;
 }
 
 /**
@@ -78,8 +79,10 @@ export default class EditableField extends React.Component<
     editing: false,
     origValue: this.props.value,
     currentValue: this.props.value,
+    fieldWidth: undefined,
   };
   inputRef = React.createRef<Input>();
+  fieldRef = React.createRef<HTMLDivElement>();
 
   save = _.debounce((val: string) => {
     if (this.props.save) {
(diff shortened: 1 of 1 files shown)


onx2 (commented):
@ArtsiomWB have you tried using css for this? maybe setting the width `100%` rather than an explicit pixel value is a more robust approach. For example, this comment block that I'm typing in now does that and relies on the parent container to determine relative width.

ArtsiomWB:
Changed it to where it is set to 100% with css as well as can be edited in the horizontal direction. 

https://github.com/user-attachments/assets/1e0c62b4-21ad-4597-84b0-77500fcd928e

weave-js/src/common/components/EditableField.tsx
```diff
@@ -256,7 +257,7 @@ export default class EditableField extends React.Component<
 
     return (
       <div
-        className={className}
+        className={`${className}`}
```
onx2, line 260: Isn't this the same thing as before just more verbose?

onx2 (approved):
Looking at the behavior it seems weird that we can resize width, wdyt about a static width for both read and edit?

Approved but something to consider

ArtsiomWB:
Per Tristan's request made the field wider using css. 

https://github.com/user-attachments/assets/152cd3de-5238-41e6-9aa5-156a5423079a

onx2 (approved):

PR #4067, opened 2025-04-07
feat(ui): Conditionally add trace metadata to ObjectView

zbirenbaum:
## Description

- Fixes WB-24306

Adds a conditionally available section to the ObjectView which displays full otel trace metadata when the field is available.

image


bcsherma (approved):
🔥

PR #4069, opened 2025-04-07
chore(weave): Temporary patch for auth url referrer

tssweeney:
This is a temporary patch for `wandb` in order to implement basic referrer tracking in
auth URLs. This commit: https://github.com/wandb/wandb/commit/5db4d763516ba2a3e8d94a85ba67bdf6fc124423 adds a utility for referrer tracking, but is not yet released. Once released, this patch can be removed and instead https://github.com/wandb/weave/pull/4004/files can be merged. Since this will take more than a month, we need a short-term patch to ensure that we can start to track referrers.


gtarpenning (approved):
lgtm, kinda wild that we check any string, i wonder what cases might accidentally trigger this.

gtarpenning (approved):
👍🏻

PR #4070, opened 2025-04-07
fix(weave): require opentelemetry proto

bcsherma:
`import weave` fails with `ModuleNotFoundError: No module named 'opentelemetry'`. This PR moves `opentelemetry-proto` from an optional to required dependency. It's eagerly imported by `trace_server_interface.py` which runs on `import weave`. Alternate approach would be to conditionally import.


zbirenbaum (approved):
Don't change indent level tho

PR #4078, opened 2025-04-08
chore(ui): Trace tree - fix padding on un-nested node levels

m-rgba:
## Description

| Before | After |
| -------- | ------- |
| Screenshot 2025-04-08 at 3 28 19 PM | Screenshot 2025-04-08 at 3 27 23 PM |

- Fixes padding on un-nested node levels to match nested text (since they're on the same level).


zbirenbaum (approved):
Thanks! Looks much better in the pic

PR #3797, opened 2025-02-27
chore(weave): call stream supports sorting and filtering by latency, status, trace_name

bcsherma:
## Description

Adds support for `summary.weave.{latency_ms,status,trace_name}` as sort and filter fields for the calls query.

Summary fields are not (in general) persisted and must be computed on the fly for each query. This PR adds a new `CallsMergedSummaryField` model to the query builder which generate sql to compute the real values of the summary field in the db query so we can sort/filter the results by the field.

Also adds support for latency and status summary fields to the sqlite trace server implementation.

## Testing

Adds unit tests for the SQL generation and an end to end test that runs the queries against the clickhouse + sqlite trace servers and verifies the sort + filter results. Also tested in our GCP QA environment against a real clickhouse cloud instance.

## Summary by CodeRabbit

- **New Features**
  - Improved call record handling with advanced sorting and filtering options by status and latency, ensuring consistent ordering and more precise performance metrics.
  - Enhanced processing of operational data to deliver clearer insights into call performance (e.g., updated status and response time information).

- **Tests**
  - Expanded quality checks with new test functions for sorting and filtering calls by status and latency, ensuring the reliability and accuracy of these enhancements.

PR #4083, opened 2025-04-09
feat(ui): Add status filtering

jamie-rasmussen:
## Description

Wires up the UI to allow filtering by status, either through the filter popup form or Option+Click. Backend support for this was added in https://github.com/wandb/weave/pull/3797

Also makes Option+Click more useful. Currently doing Option+Click on the same value adds multiple duplicate filters. With this PR we change it so a second Option+Click will either toggle the filter off or replace it with an updated one (e.g. in the case where you have an `OR` of status values and want to filter to one.)

Screenshot 2025-04-09 at 11 32 51 AM

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterRow.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterRow.tsx
@@ -63,7 +63,8 @@ export const FilterRow = ({
   };
 
   const isOperatorDisabled =
-    isWeaveRef(item.value) || ['id', 'user'].includes(getFieldType(item.field));
+    isWeaveRef(item.value) ||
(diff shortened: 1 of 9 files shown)


bcsherma (approved):
Nice! Not worth blocking but I do think that the select highlight for the status chips looks a little funny and could use a design review. I think greying/blurring out states that aren't included in the filter + showing the included states as normal rather than with a blue highlight might be more intuitive. A tooltip for include/don't include when hovering on the state may also help users understand how to modify the filter.

Screenshot 2025-04-09 at 12 34 17 PM

jamie-rasmussen:
> Nice! Not worth blocking but I do think that the select highlight for the status chips looks a little funny and could use a design review.

Totally agree - a consequence of reusing our existing Button component and it not having a variant that works ideally here. I will follow up with design but a custom built control is probably the way to go.

PR #4085, opened 2025-04-09
feat(app): Allow selection of custom step metric in stepper panel

nicholaspun-wandb:
## Description

- Implements [WB-24236](https://wandb.atlassian.net/browse/WB-24236)
- Adds a selection to the stepper panel config to allow the user to select the step key.
- Valid step keys are those where the values are all integers, but can contain null values (e.g. in the case where a custom step key is logged for one run but not for another).
    - We ignore monotonicity since the data could span across runs (e.g. `runs.history`), in which case the entire sequence may not be monotonic 

## Testing

Local FE Invoker

https://github.com/user-attachments/assets/47de0543-7426-4b03-a619-fc9f1dbb718c

[WB-24236]: https://wandb.atlassian.net/browse/WB-24236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/Panel2/PanelStepper/index.tsx
+++ b/weave-js/src/components/Panel2/PanelStepper/index.tsx
@@ -1,20 +1,28 @@
+import {useWeaveContext} from '@wandb/weave/context';
 import {
+  constBoolean,
   constFunction,
   constNumber,
   constString,
+  isAssignableTo,
   Node,
   NodeOrVoidNode,
+  opBooleanAll,
   opFilter,
   opIndex,
   opIsNone,
+  opMap,
   opNot,
   opNumberEqual,
+  opNumberIsInteger,
+  opOr,
   opPick,
+  union,
   voidNode,
 } from '@wandb/weave/core';
-import React, {useMemo} from 'react';
+import React, {useEffect, useMemo, useState} from 'react';
 
-import {useNodeWithServerType} from '../../../react';
+import {useNodeValueExecutor, useNodeWithServerType} from '../../../react';
 import {usePanelStacksForType} from '../availablePanels';
 import * as Panel2 from '../panel';
 import {PanelStepper} from './component';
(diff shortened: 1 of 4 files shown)


weave-js/src/components/Panel2/PanelStepper/index.tsx
```diff
@@ -135,6 +205,56 @@ const PanelStepperEntryComponent: React.FC<PanelStepperEntryProps> = props => {
     });
   }
 
+  const numericalKeys = Object.keys(propertyKeysAndTypes).filter(
+    key =>
+      isAssignableTo(propertyKeysAndTypes[key], union(['none', 'number'])) &&
+      !['loss', 'accuracy', 'precision', 'recall'].some(metric =>
```
nicholaspun-wandb, line 211: I can remove this, just figured that sometimes these might slip through if the fit the `isInteger` criteria

nicholaspun-wandb:
Beta link: https://beta.wandb.ai/home?betaVersion=cd8d5a66b8686e349193258c6e00ed490f5ad180

weave-js/src/components/Panel2/PanelStepper/index.tsx
```diff
@@ -32,6 +38,71 @@ import {
   NONE_KEY_AND_TYPE,
 } from './util';
 
+const useIntegralChecks = (
```
nicholaspun-wandb, line 41: We need to do this roundabout method with `useNodeValueExecutor` and promises since we can't the `useNodeValue` hook in a callback (i.e. We wouldn't be able to write something `possibleKeys.filter((key) => useNodeValue(opIsInteger(key)) === true)`)

weave-js/src/components/Panel2/PanelStepper/index.tsx
```diff
@@ -32,6 +38,71 @@ import {
...
+        }
+      });
+
+      const checkResults = await Promise.all(promises);
+
+      checkResults.forEach(({key, result}) => {
+        newResults[key] = result;
+      });
+
+      hasRunRef.current = true;
```
nicholaspun-wandb, line 96: Couldn't find the right set of conditions to not have this run multiple times. It should run once when the panel is selected (the component is mounted) and doesn't change even when the user modifies the config.

weave-js/src/components/Panel2/PanelStepper/index.tsx
domphan-wandb: I like it. Can you add a comment to the code for this?

domphan-wandb (approved):

PR #4092, opened 2025-04-09
chore(weave): Add drawer to providers tab

jwlee64:
## Description
Use the configure drawer instead of linking to the settings page

Screenshot 2025-04-09 at 2 37 18 PM


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/OverviewPage/ProvidersTab.tsx
```diff
@@ -47,33 +47,21 @@ const columns: GridColDef[] = [
...
-          <Link
-            to={`/${params.row.entityName}/settings`}
-            $variant="secondary"
-            target="_blank"
-            rel="noopener noreferrer"
-            className="flex items-center gap-2">
-            Add team secret
-            <Icon name="forward-next" />
-          </Link>
+        {!params.row.status && (
```
chance-wnb, line 56: Why don't we not require `isAdmin` now?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/OverviewPage/ProvidersTab.tsx
```diff
@@ -142,6 +139,10 @@ export const ProvidersTabInner: React.FC<{
       missingSecrets,
       isAdmin,
       entityName,
+      configure: () => {
```
chance-wnb, line 142: It seems hacky to add a function to work around accessing the state setter in this component. 

Typically, we should not include functions. The DataGrid often serializes row data (e.g., for filtering, sorting, or exporting). Functions are not serializable and will be lost.

I think there is a MUI recommended way of achieving that.
https://mui.com/x/react-data-grid/api-object/?srsltid=AfmBOoqdeWGlM6LXDglC2UeP7yt8EG0RJ56mcI2TcJy7ARMRZY9yAduj

Or you can consider simply adding a React context. And have a "configureProvider(provider: string)" function in the context as a service.

jwlee64: sounds good I add a context

chance-wnb: I think the prop of `configure` on 153 is no longer needed.

chance-wnb: I think the prop of `configure` on 153 is no longer needed.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/OverviewPage/ProvidersTab.tsx
```diff
@@ -156,6 +168,11 @@ export const ProvidersTabInner: React.FC<{
     }
   );
 
+  const configureProvider = (provider: string) => {
```
chance-wnb, line 167: Lets use `useMemo()` here, so we don't constantly reset the provider value.

```typescript
const value = useMemo(() => {
  const configureProvider = (provider: string) => {
    setIsConfiguringProviderDrawerOpen(true);
    setConfiguringProvider(provider);
  };
  return {configureProvider};
}, []);
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/OverviewPage/ProvidersTab.tsx
```diff
@@ -172,11 +189,14 @@ export const ProvidersTabInner: React.FC<{
             </div>
 
             <div className="flex h-full min-h-[200px] items-center justify-center">
-              <ProviderTable
-                columns={columns}
-                providers={providers}
-                loading={configuredProvidersLoading}
-              />
+              <ConfigureProviderContext.Provider
+                value={{configureProvider: configureProvider}}>
```
chance-wnb, line 189: if we have the code above, this will be `value={value}`

chance-wnb (approved):
A minor nit is left. Thanks for the change!

PR #4094, opened 2025-04-09
chore(weave): Create long-live storage clients per instance

tssweeney:
This is a pure refactor of the file storage mechanics such that the trace server client only has instantiates a single connection to the bucket. As implemented today, a client connection is minted for each request which is not as performant as it could be. 

Furthermore, it correctly decouples the loading of an environment-configured client from something that could be generated at runtime. This is the first step to moving to the better auth system.

### DIFF
--- a/weave/trace_server/file_storage.py
+++ b/weave/trace_server/file_storage.py
@@ -1,5 +1,5 @@
 import logging
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from typing import Any, Callable, Union, cast
 
 import boto3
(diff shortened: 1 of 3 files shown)


weave/trace_server/file_storage.py
```diff
@@ -219,23 +253,40 @@ def store(self, uri: AzureFileStorageURI, data: bytes) -> None:
...
+        )
         client = self._get_client(uri.account)
         container_client = client.get_container_client(uri.container)
         blob_client = container_client.get_blob_client(uri.path)
         stream = blob_client.download_blob()
         return stream.readall()
 
 
-def _get_storage_client(uri: FileStorageURI) -> FileStorageClient:
+def get_storage_client_from_env() -> Optional[FileStorageClient]:
```
gtarpenning, line 266: slightly confusing name given that this is an optional

weave/trace_server/file_storage.py
```diff
@@ -211,6 +242,9 @@ def _get_client(self, account: str) -> BlobServiceClient:
     @create_retry_decorator("azure_storage")
     def store(self, uri: AzureFileStorageURI, data: bytes) -> None:
         """Store data in Azure container with automatic retries on failure."""
+        assert isinstance(uri, AzureFileStorageURI) and uri.to_uri_str().startswith(
```
gtarpenning, line 245: why do we have to assert the uri bit?

weave/trace_server/file_storage.py
```diff
@@ -181,9 +208,13 @@ class AzureStorageClient(FileStorageClient):
     """Azure Blob Storage implementation supporting both connection string and account credentials."""
 
     def __init__(
-        self, credentials: Union[AzureConnectionCredentials, AzureAccountCredentials]
+        self,
+        base_uri: FileStorageURI,
+        credentials: Union[AzureConnectionCredentials, AzureAccountCredentials],
     ):
         """Initialize Azure client with either connection string or account credentials."""
+        assert isinstance(base_uri, AzureFileStorageURI)
```
gtarpenning, line 216: interesting, we prefer this over typing `base_uri`? or can't?

gtarpenning (approved):
lgtm just questions. avoiding minting a client every time feels right

weave/trace_server/file_storage.py
gtarpenning: because we aren't guaranteed these uri's are for the given type, surprised we don't just rely on the type hints from the parse fn but I get it

weave/trace_server/file_storage.py
tssweeney: Fixed

weave/trace_server/file_storage.py
tssweeney: It violates the `Liskov Substitution Principle`: "Method arguments being contravariant not only means that we can widen their types, but also that we cannot make their types more narrow." https://madewithlove.com/blog/liskov-substitution-principle-explained/

gtarpenning: gah silly Liskov! yeah

weave/trace_server/file_storage.py
tssweeney: I want to be really careful here in the case that a url stored points to storage X, but then the client is storage Y and we blindly use the path of X.

PR #4103, opened 2025-04-10
chore(weave): crewai release broken on 3.10

jamie-rasmussen:
## Description

crewai release 0.114.0 is broken on Python 3.10. See:
https://github.com/crewAIInc/crewAI/issues/2543
https://github.com/crewAIInc/crewAI/pull/2553

This is temporarily pinning the version to be not greater than the previous release, 0.108.0 to fix CI.


gtarpenning (approved):
and a lint!

PR #4112, opened 2025-04-10
chore(app): Update NumberInput to new designs

ericakdiaz:
## Description

This updates the NumberInput to have an option for the newest designs with a +/- stepper. Since the wider width may not be compatible with older usages, I added an optional prop so that it's opt-in as older ones are migrated. 

See designs here: 
https://www.figma.com/design/BsoKhzDe7Q8OX08owa9DuM/Media-panel-improvements?node-id=5257-141396&m=dev

## Testing

How was this PR tested?

Tested in https://github.com/wandb/core/pull/28695. It should only be enabled in that one place, which is currently behind a ramp flag.


peytonsarmiento (approved):
🙌

PR #4114, opened 2025-04-11
perf(weave): add indexed datetime filtering to calls query

gtarpenning:
## Description

[WB-24385](https://wandb.atlassian.net/browse/WB-24385)

This pr: 
- adds a new column `sortable_datetime` to the `calls_merged` table with default `coalesce(started_at, ended_at, NOW())`
- adds that column to the calls_merged_view as `coalesce(started_at, ended_at, call_parts.created_at)`
- adds and materialized a minmax index on that column
- refactors the sql optimization paradigm in `calls_query_builder` to be more of a mapper pattern
- adds a `SortableDatetimeOptimizationProcessor` class that adds datetime filtering pre-where to the calls query
- add a `NotContext` to track if the current filter has a `not` prepended. We need to do this so we can apply the buffer in the correct direction. 
- uses configurable timestamp buffer, defaults to 5 minutes. This means that in our pre-groupby filter we include more calls than necessary, and rely on the outer filter in the group by to pair down the results to be exact. 

NOTE:
- technically using this column as a heuristic is actually incorrect because we will occasionally filter out call_ends that are outside of the time window while the call_start is valid and within bounds. This will result in wonky results on the margins of the date filter. One easy way to lessen the impact of this is to make the initial date filter less strict (by adding a fuzzing/buffer to the initial timestamp filter). Because this is being implemented as an optimization, we still retain the original filter. 

Example:
```sql
SELECT count() as c
FROM
(
    SELECT calls_merged.id AS id
    FROM calls_merged
    WHERE calls_merged.project_id = 'project_id'
         AND calls_merged.sortable_datetime > '2025-04-01 03:25:58.0000'  --<<--------
    HAVING (any(calls_merged.started_at) > _CAST(1742472000, 'UInt64')) AND (NOT (any(calls_merged.started_at) > _CAST(1742475600, 'UInt64'))) AND (any(calls_merged.deleted_at) IS NULL) AND (NOT (any(calls_merged.started_at) IS NULL))
)
```

## Testing

Updates and adds unit and client end to end tests 

Replica tests (stacked on `op_name` improvement already merged [here](https://github.com/wandb/weave/pull/4116))
| Environment | Num Runs | Avg Mem     | Rows Read | p50 Latency | p90 Latency | p99 Latency |
|-------------|----------|-------------|-----------|-------------|-------------|-------------|
| prod        | 5        | 9.24 GiB    | 8,271,114 | 2.895s      | 5.212s      | 5.212s      |
| branch      | 6        | 195.52 MiB  | 104,851   | 0.629s      | 1.213s      | 1.213s      |

With a slightly better order (datetime first, then op_name) compared to current prod (no op_name optimization)
| Environment | Num Runs | Avg Mem     | Rows Read | p50 Latency | p90 Latency | p99 Latency |
|-------------|----------|-------------|-----------|-------------|-------------|-------------|
| prod        | 4        | 22.42 GiB   | 8,271,114 | 5.494s      | 5.597s      | 5.597s      |
| branch      | 5        | 183.06 MiB  | 104,851   | 0.549s      | 0.625s      | 0.625s      |

Migration tests in prod replica take `<100ms`

[WB-24385]: https://wandb.atlassian.net/browse/WB-24385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/migrations/012_add_sortable_timestamp.up.sql
```diff
@@ -0,0 +1,35 @@
+-- Create non-nullable Datetime field in the call_merged table
+ALTER TABLE calls_merged 
+    ADD COLUMN sortable_datetime Datetime(6) DEFAULT coalesce(started_at, ended_at, NOW());
```
tssweeney, line 3: when we did this together, we used the keyword `MATERIALIZED` vs `DEFAULT`. was this switch intentional? If so, can you share why?

gtarpenning: Yes! You can't insert into a column that has been `MATERIALIZED`. Like at all. We were able to query from it fine, but any inserts, even from the materialized view, fail horrible with errors like:

```
2025.04.11 19:31:33.462313 [ 801 ] {18445af9-cffc-4176-8df9-222fbca2f5e6} <Error> PushingToViews: Failed to push block to view default.calls_merged_view (52823336-5ef9-48aa-a9a3-7750e121d926), Cannot insert column sortable_datetime, because it is MATERIALIZED column
2025.04.11 19:31:33.463877 [ 801 ] {18445af9-cffc-4176-8df9-222fbca2f5e6} <Error> executeQuery: Code: 44. DB::Exception: Cannot insert column sortable_datetime, because it is MATERIALIZED column. (ILLEGAL_COLUMN) (version 24.6.2.17 (official build)) (from 192.168.65.1:58981) (in query: INSERT INTO `call_parts` (`started_at`, `output_dump`, `deleted_at`, `attributes_dump`, `wb_user_id`, `parent_id`, `wb_run_id`, `display_name`, `ended_at`, `op_name`, `project_id`, `inputs_dump`, `id`, `input_refs`, `output_refs`, `trace_id`, `summary_dump`, `exception`) FORMAT Native ), Stack trace (when copying this message, always include the lines below):```

File: weave/trace_server/calls_query_builder/optimization_builder.py [hunk 2]
```diff
@@ -0,0 +1,572 @@
...
+        return None
+
+    # convert timestamp to datetime_str
+    timestamp = int(literal_value)
+
+    # Apply buffer in appropriate direction based on context
+    buffer_seconds = int(DATETIME_BUFFER_TIME_SECONDS)
+    if NotContext.is_in_not_context():
+        # For NOT context, add buffer to make filter more permissive
+        timestamp += buffer_seconds
```
tssweeney, line 562: doesn't this buffer direction depend on the direction of the operation?

gtarpenning: yeah, we only support gt and gte rn. that could be made direction agnostic easily. If we ever support LT ;)

tssweeney: If we add lt / lte (which we absolutely should!) does that break anything? Do we have protections against this?

tssweeney: We should add a restriction on the operator itself to ensure that this constraint is asserted

gtarpenning: Yo we can only get here when we are in a `process_gt` or `process_gte` context. you would have to explicitly add a `process_lt` to support LT

gtarpenning: in which case you would refactor this optimization function

gtarpenning: if you add `$lt` support tomorrow everything will still work, you won't get the optimization until you also implement the `process_lt` fn, but it won't break

gtarpenning: Its also typed, would be hard to accidentally do it.

```python
def _create_datetime_optimization_sql(
    operation: Union[tsi_query.GtOperation, tsi_query.GteOperation],
```

tssweeney (approved):

PR #4116, opened 2025-04-11
perf(weave): filter op_name pre group by 

gtarpenning:
## Description

[WB-24386](https://wandb.atlassian.net/browse/WB-24386)

In this pr we handle the `op_name` separately from the other having conditions. So we can pull it before the group by. Because ALL call starts are guaranteed to have an `op_name` we can always filter by `op_name LIKE <> or op_name is NULL` to grab all possible rows that match. 

Example:
```sql
SELECT calls_merged.id AS id
   FROM calls_merged
   WHERE calls_merged.project_id = {pb_263_2:String}
     -------------------------
     AND ((calls_merged.op_name LIKE {pb_263_1:String}) OR (calls_merged.op_name IS NULL))
     -------------------------
   GROUP BY (calls_merged.project_id,
             calls_merged.id)
   HAVING (((any(calls_merged.started_at) > {pb_263_0:Float64}))
           AND ((any(calls_merged.deleted_at) IS NULL))
   LIMIT 50
   OFFSET 0
```

## Testing

- update unit tests

Replica tests:
| Environment | Num Runs | Avg Mem     | Rows Read | p50 Latency | p90 Latency | p99 Latency |
|-------------|----------|-------------|-----------|-------------|-------------|-------------|
| prod        | 7        | 3.47 GiB    | 8,451,140 | 2.294s      | 4.534s      | 4.534s      |
| branch      | 9        | 204.58 MiB  | 8,451,140 | 1.366s      | 1.643s      | 1.643s      |

[WB-24386]: https://wandb.atlassian.net/browse/WB-24386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace_server/test_calls_query_builder.py
+++ b/tests/trace_server/test_calls_query_builder.py
@@ -114,12 +114,13 @@ def test_query_heavy_column_simple_filter() -> None:
                 calls_merged.id AS id
             FROM calls_merged
             WHERE calls_merged.project_id = {pb_1:String}
+                AND ((calls_merged.op_name IN {pb_0:Array(String)})
+                    OR (calls_merged.op_name IS NULL))
             GROUP BY (calls_merged.project_id, calls_merged.id)
             HAVING (
                 ((any(calls_merged.deleted_at) IS NULL))
                 AND ((NOT ((any(calls_merged.started_at) IS NULL))))
-            AND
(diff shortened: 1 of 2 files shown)


neutralino1 (approved):
Thank you!

PR #4123, opened 2025-04-14
chore(weave): Add grok-3 to playground

jwlee64:
## Description
Screenshot 2025-04-14 at 10 16 35 AM

## Testing

How was this PR tested?


weave/trace_server/model_providers/model_providers.json
```diff
@@ -347,6 +347,18 @@
     "litellm_provider": "azure",
     "api_key_name": "AZURE_API_KEY"
   },
```
jwlee64, line 349: this is just generation from running make update_model_providers

weave/trace_server/llm_completion.py
```diff
@@ -37,6 +37,9 @@ def lite_llm_completion(
     # XAI models don't support response_format
     elif provider == "xai":
         inputs.response_format = None
+        if "grok-3-mini" in inputs.model:
```
jwlee64, line 40: doing one offs like this is probably unsustainable and I'll have to come up with a better way to specify which params are supported for different models, but for now this works

jamie-rasmussen (approved):

PR #4124, opened 2025-04-14
chore(weave): increase filter bar debounce + fix character reappearing

gtarpenning:
## Description

[WB-24415](https://wandb.atlassian.net/browse/WB-24415)

This pr: 
- increases the debounce time from 700ms to 1s
- fixes an issue with the string filter where rapidly deleting the filter results in the last character reappearing
   - Now we explicitly keep track of incomplete filters. We only submit filters that are complete, and keep in local state the rest. This is only when a filter row has been modified from complete to incomplete. Instead of removing the entire row, which causes a large UI change and the filter to disappear, we keep it around in local state and render it with the value field as empty (which is what you expect as a user). 
   - This should feel exactly the same as the current flow, without the reappearing character bug, _and_ now when the filter is incomplete we query the backend as if the filter was not present. 

## Testing

prod:
filter-debounce-prod-2

branch:
filter-debounce-branch

[WB-24415]: https://wandb.atlassian.net/browse/WB-24415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
@@ -224,7 +224,7 @@ export const FilterBar = ({
         debouncedSetFilterModel(newItemsModel);
       }
     },
-    [localFilterModel, debouncedSetFilterModel, applyCompletedFilters]
+    [localFilterModel, debouncedSetFilterModel]
   );
 
   const onRemoveFilter = useCallback(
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
```diff
@@ -30,7 +30,7 @@ import {combineRangeFilters, getNextFilterId} from './filterUtils';
 import {GroupedOption, SelectFieldOption} from './SelectField';
 import {VariableChildrenDisplay} from './VariableChildrenDisplayer';
 
-const DEBOUNCE_MS = 700;
+const DEBOUNCE_MS = 1_000;
```
zbirenbaum, line 33: Is the implication here that if for some reason a state update takes longer than 1 second (as opposed to 700ms) the old state may still appear?

It seems like there were some logic changes in this PR but it's not immediately apparent to me what the debounce vs the filter state changes address respectively

gtarpenning: updated the pr description for clarity on the functional changes.

zbirenbaum (approved):

PR #4129, opened 2025-04-14
explicit adjusting filter loading state -- still bugged

gtarpenning:
## Description

TODO:
- how can it be possible that call.loading returns false but callsResult is empty?? tableData computation? When exactly does this happen, on filter updates? on interrupted filter updates? 

## Testing

How was this PR tested?


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
```diff
@@ -194,11 +194,10 @@ export const useCallsForQuery = (
       storageSizeResults,
       loading: calls.loading,
       // Return faster calls query results until cost query finishes
-      result: calls.loading
```
gtarpenning, line 197: not needed

PR #4130, opened 2025-04-14
feat(weave): Support emitting storage size in object details view (trace_server)

chance-wnb:
## Summary
This PR adds support for displaying storage size information in the object details view and table statistics. It consists of mostly the backend changes, and minimum frontend changes to fit the interface changes related to object storage size only.

TLDR, I changed the following two APIs to include storage size information

```
/objs/query  ---> this will be used to retrieve each individual object version's storage size,
/table/query_stats_batch ---> this will be used to retrieve each table's storage size from all of its rows.
```

After this PR is merged. The internal PR https://github.com/wandb/core/pull/28783 should merge soon.

As a reference, the Frontend change that is going to utilize the API endpoint in this PR can be reviewed [here](https://github.com/wandb/weave/pull/4142).

### Backend Changes
1. Enhanced object querying:
   - Added `include_storage_size` option to `ObjQueryReq`
   - Updated `ObjectMetadataQueryBuilder` to handle storage size inclusion
   - Modified query result formatting to include storage size when requested
   -
2. In order to prevent breaking existing clients (FE+SDK) due to tempering the existing endpoint of `table_stats_query`:
   - Created a new endpoint `table_stats_query_batch` which takes plural form parameter `digests`
   - Added `include_storage_size` flag to control storage size retrieval
   - Created new `make_table_stats_query_with_storage_size` function for efficient storage size queries

### Frontend Changes
   - Updated TypeScript interfaces to include new storage size fields for object queries

### Testing
1. Added new test cases:
   - `test_obj_query_with_storage_size_clickhouse` to verify object storage size functionality
   - `test_table_query_stats_with_storage_size` to verify table storage size retrieval
   - Updated existing tests in `test_table_query.py` suite to test against the `table_stats_query_batch` api.

2. Added SQL query tests:
   - Fixed breakage in the `test_objects_query_builder.py` with storage size related tests
   - Added SQL formatting assertions for better query validation

## Future Improvements
- Hook up the frontend.
- Remove the deprecated `table/query_stats` endpoint, and use the batch API `table/query_stats_batch` (Far future when the conditions meet)

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -30,6 +30,7 @@ from weave.trace.context.weave_client_context import (
     get_weave_client,
     set_weave_client_global,
 )
+from weave.trace.refs import parse_uri
 from weave.trace.vals import MissingSelfInstanceError
 from weave.trace.weave_client import sanitize_object_name
 from weave.trace_server import trace_server_interface as tsi
@@ -3691,6 +3692,58 @@ def test_calls_hydrated(client):
     assert calls[1].inputs["input_ref"]["hi"]["there"]["foo"] == "bar"
     assert calls[2].output["woahhhh"]["hi"]["there"]["foo"] == "bar"
     assert calls[2].inputs["input_ref"]["hi"]["there"]["foo"] == "bar"
+
+
+def test_obj_query_with_storage_size_clickhouse(client):
+    """Test querying objects with storage size information"""
+    if client_is_sqlite(client):
+        pytest.skip("Skipping test for sqlite clients")
+
+    # Create a test object with some data to ensure it has size
+    dataset = weave.Dataset(name="test_dataset", rows=[{"key": "value" * 1000}])
+    weave.publish(dataset)
+
+    # Query the object with storage size included
+    res = client.server.objs_query(
+        tsi.ObjQueryReq(
+            project_id=get_client_project_id(client),
+            include_storage_size=True,
+            filter={"object_ids": ["test_dataset"]},
+        )
+    )
+
+    assert len(res.objs) == 1
+    queried_obj = res.objs[0]
+
+    # Verify that storage size is present
+    assert queried_obj.size_bytes is not None
+    assert queried_obj.size_bytes == 270  # Should have some size due to the test data
+
+    # Test that a table is created and its size is correct
+    table_ref = parse_uri(queried_obj.val["rows"])
+    res = client.server.table_query_stats_batch(
+        tsi.TableQueryStatsBatchReq(
+            project_id=client._project_id(),
+            digests=[table_ref.digest],
+            include_storage_size=True,
+        )
+    )
+
+    assert res.tables[0].storage_size_bytes == 5011
+
+    # Query without storage size (default behavior)
+    res_without_size = client.server.objs_query(
+        tsi.ObjQueryReq(
+            project_id=get_client_project_id(client),
+            filter={"object_ids": ["test_dataset"]},
+        )
+    )
+
+    assert len(res_without_size.objs) == 1
+    queried_obj_without_size = res_without_size.objs[0]
+
+    # Verify that storage size is not included when not requested
+    assert queried_obj_without_size.size_bytes is None
 
 
 def test_call_query_stream_with_costs_and_storage_size(client, clickhouse_client):
--- a/tests/trace/test_table_query.py
+++ b/tests/trace/test_table_query.py
@@ -1,6 +1,11 @@
 import random
 from collections.abc import Iterator
 
+import pytest
+
+from tests.trace.util import (
+    client_is_sqlite,
+)
 from weave.trace.weave_client import WeaveClient
 from weave.trace_server import trace_server_interface as tsi
 
@@ -265,40 +270,40 @@ def test_table_query_multiple_sort_criteria(client: WeaveClient):
 def test_table_query_stats(client: WeaveClient):
     digest, row_digests, data = generate_table_data(client, 10, 10)
 
-    stats_res = client.server.table_query_stats(
-        tsi.TableQueryStatsReq(
+    stats_res = client.server.table_query_stats_batch(
+        tsi.TableQueryStatsBatchReq(
             project_id=client._project_id(),
-            digest=digest,
+            digests=[digest],
         )
     )
 
-    assert stats_res.count == len(data)
+    assert stats_res.tables[0].count == len(data)
 
 
 def test_table_query_stats_empty(client: WeaveClient):
     digest, row_digests, data = generate_table_data(client, 0, 0)
 
-    stats_res = client.server.table_query_stats(
-        tsi.TableQueryStatsReq(
+    stats_res = client.server.table_query_stats_batch(
+        tsi.TableQueryStatsBatchReq(
             project_id=client._project_id(),
-            digest=digest,
+            digests=[digest],
         )
     )
 
-    assert stats_res.count == len(data)
+    assert stats_res.tables[0].count == len(data)
 
 
 def test_table_query_stats_missing(client: WeaveClient):
     digest, row_digests, data = generate_table_data(client, 10, 10)
 
-    stats_res = client.server.table_query_stats(
-        tsi.TableQueryStatsReq(
+    stats_res = client.server.table_query_stats_batch(
+        tsi.TableQueryStatsBatchReq(
             project_id=client._project_id(),
-            digest="missing",
+            digests=["missing"],
         )
     )
 
-    assert stats_res.count == 0
+    assert len(stats_res.tables) == 0
 
 
 def generate_duplication_simple_table_data(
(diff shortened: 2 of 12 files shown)


neutralino1 (approved):
Thank you.

weave/trace_server/trace_server_interface.py
```diff
@@ -749,14 +754,33 @@ class TableQueryStatsReq(BaseModel):
...
 
 
-class TableQueryStatsRes(BaseModel):
+class TableStatsResRow(BaseModel):
     count: int
+    digest: str
+    storage_size_bytes: Optional[int] = None
+
+
+class TableQueryStatsRes(BaseModel):
```
tssweeney, line 782: this introduces a breaking change to the existing service layer. Specifically: this will change the http spec. there are two problems here:
1. practically, from a deployment perspective, the gap between deploying the new frontend and backend will have a broken UI
2. any users of this stats endpoint directly will break. There is low likelihood that end users are directly targeting this, so it might be a risk we are willing to take, but certainly not best practice.

Either: a) we should make the changes such that the are not breaking; or b) introduce a batch version of the endpoint

chance-wnb: Yes, all agreed.

On
`any users of this stats endpoint directly will break. There is low likelihood that end users are directly targeting this, so it might be a risk we are willing to take, but certainly not best practice.`

Just FYI, this API endpoint is used in the dataset read operation, which actually has a high impact.

tests/trace/test_table_query.py
```diff
@@ -265,40 +270,40 @@ def test_table_query_multiple_sort_criteria(client: WeaveClient):
 def test_table_query_stats(client: WeaveClient):
     digest, row_digests, data = generate_table_data(client, 10, 10)
 
-    stats_res = client.server.table_query_stats(
-        tsi.TableQueryStatsReq(
+    stats_res = client.server.table_query_stats_batch(
```
chance-wnb, line 273: So I am changing this suite to test `TableQueryStatsBatchReq` instead of `TableQueryStatsReq` and regard `TableQueryStatsReq` as obsolete.

tests/trace/test_client_trace.py
```diff
@@ -3691,3 +3691,45 @@ def nest(input_ref: str):
...
+def test_obj_query_with_storage_size_clickhouse(client):
+    """Test querying objects with storage size information"""
+    if client_is_sqlite(client):
+        pytest.skip("Skipping test for sqlite clients")
+
+    # Create a test object with some data to ensure it has size
+    dataset = weave.Dataset(name="test_dataset", rows=[{"key": "value" * 1000}])
+    weave.publish(dataset)
+
+    print(dataset)
```
tssweeney, line 3705: dev tracks

weave/trace_server/table_query_builder.py
```diff
@@ -113,3 +113,25 @@ def make_standard_table_query(
...
+    digest_ids = pb.add_param(table_digests)
+
+    query = f"""
+    SELECT tb_digest, any(length), sum(size_bytes) FROM
+    (
+        SELECT digest as tb_digest, length(row_digests) as length, row_digests
+        FROM tables
+        WHERE project_id = {{{project_id_name}: String}} AND digest in {{{digest_ids}: Array(String)}}
+    ) ARRAY JOIN row_digests as row_digest
+    LEFT JOIN table_rows_stats on table_rows_stats.digest = row_digest
```
tssweeney, line 134: it might improve perf to have a project_id in this part of the left join as well

weave/trace_server/objects_query_builder.py
```diff
@@ -243,6 +252,16 @@ def set_include_deleted(self, include_deleted: bool) -> None:
...
         columns = ",\n    ".join(OBJECT_METADATA_COLUMNS)
+
+        main_table_alias = "main"
+
+        if self.include_storage_size:
+            columns += ",\n   size_bytes"
+
+        join_clause = ""
+        if self.include_storage_size:
+            join_clause = f"LEFT JOIN object_versions_stats ON object_versions_stats.digest = {main_table_alias}.digest"
```
tssweeney, line 263: might want to include project id as well

tests/trace/test_client_trace.py
gtarpenning: the prints are only collected on test failure, I sort of like them in there. If the test fails you get more info, if it passes no harm.

weave/trace_server/objects_query_builder.py
gtarpenning: we can run an `EXPLAIN indexes=1` to see if that's required, but yeah doesn't hurt.

tests/trace/test_client_trace.py
chance-wnb: Yeah, thanks, my bad, and I don't think the logging for the created dataset is that useful. I will just remove.

weave/trace_server/objects_query_builder.py
```diff
@@ -243,6 +252,16 @@ def set_include_deleted(self, include_deleted: bool) -> None:
 
     def make_metadata_query(self) -> str:
         columns = ",\n    ".join(OBJECT_METADATA_COLUMNS)
+
+        main_table_alias = "main"
+
+        if self.include_storage_size:
```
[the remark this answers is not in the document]
chance-wnb: Yeah, much better. Made it that way, thanks!

weave/trace_server/objects_query_builder.py
chance-wnb: Thanks for the suggestion. I made it to
```sql
LEFT JOIN object_versions_stats ON object_versions_stats.digest = {main_table_alias}.digest
            AND object_versions_stats.project_id = {{project_id: String}}
```

however, maybe what you suggested is 

```sql
LEFT JOIN (
        SELECT * FROM object_versions_stats WHERE object_versions_stats.project_id =  {{project_id: String}}
    ) as filtered_object_versions_stats ON filtered_object_versions_stats.digest = main.digest
```

let me know if the latter is preferred.

chance-wnb:
> Very lovely implementation! Nice job (:
> 
> Tactically, we probably want:
> 
> 1. A quick eye from @gtarpenning regarding the query builder
> 2. Split the python and typescript since the typescript can't be landed before the server picks up the changes

Thank you so much! I have removed the TS changes from this PR.

weave/trace_server/objects_query_builder.py
chance-wnb: I adopted the 2nd approach(sub-query) now.

gtarpenning (approved):
lgtm, thanks for the query change. Let's make sure to test in a QA environment before deploying just to be sure that the shape of real data doesn't cause any issues. I don't expect it to, but this is a sufficiently large change as to make me slightly cautious.

tests/trace/test_client_trace.py
```diff
@@ -3693,6 +3693,46 @@ def nest(input_ref: str):
     assert calls[2].inputs["input_ref"]["hi"]["there"]["foo"] == "bar"
 
 
+def test_obj_query_with_storage_size_clickhouse(client):
```
tssweeney, line 3696: this test seems like a prime opportunity to also assert on the table query size

tests/trace/test_table_query.py
```diff
@@ -446,3 +451,21 @@ def test_duplicate_table_with_identical_rows(client: WeaveClient):
...
+        tsi.TableQueryStatsBatchReq(
+            project_id=client._project_id(),
+            digests=[digest],
+            include_storage_size=True,
+        )
+    )
+
+    assert stats_res.tables[0].count == len(data)
+    assert stats_res.tables[0].storage_size_bytes > 0
```
tssweeney, line 471: why not have this exact?

tssweeney (approved):
Approving with some lightweight comments on additional tests

tssweeney:
Also, i would normally like to break this into 2 prs since there are 2 distinct functionalities produced and tested. Ok to leave as is, but in the future consider breaking up more.

chance-wnb:
> Also, i would normally like to break this into 2 prs since there are 2 distinct functionalities produced and tested. Ok to leave as is, but in the future consider breaking up more.

That's a completely valid point and I am actually aware of that. Due to leadership's urgent expectations for wrapping up the project, and potential back-n-forth during code reviews if splitting, I had to resort to this desperate measure. 

I will absolutely try to minimize the occurrences like this in the future.

tests/trace/test_table_query.py
chance-wnb: `generate_table_data(client, 10, 10)` generates random data. so there is no exact number to test against. `generate_table_data()` is pretty good for other test cases, so I decide to not temper it. 

Considering that in the `test_client_trace` unit test I also added logic to verify the table size, and that test scenario already tests the exact number. I am not over-complicating things here.

tests/trace/test_client_trace.py
chance-wnb: Yep, added that. Thanks for noticing the opportunity!

chance-wnb:
> lgtm, thanks for the query change. Let's make sure to test in a QA environment before deploying just to be sure that the shape of real data doesn't cause any issues. I don't expect it to, but this is a sufficiently large change as to make me slightly cautious.

I ran the requests against the QA "weave-team/types-chance" project, the current scenarios all work fine. I also simulated requests as if they need to request storage size (since currently UI is not hooked up), the requests all return quickly.

gtarpenning:
🚢🚢🚢

PR #4131, opened 2025-04-14
fix(app): Fix image mask visibility hierarchy

ericakdiaz:
## Description

https://weightsandbiases.slack.com/archives/C085WBN5FPV/p1744673754916729 for context. We want to be consistent in having the "lower" level override the higher one.

## Testing

How was this PR tested?

https://github.com/user-attachments/assets/c1600c81-699c-4101-a814-35e5f09f55de

### DIFF
--- a/weave-js/src/components/Panel2/ImageWithOverlays.tsx
+++ b/weave-js/src/components/Panel2/ImageWithOverlays.tsx
@@ -220,7 +220,7 @@ const OverlayCanvas = styled.canvas`
 interface SegmentationCanvasProps {
   segmentation: ImageData;
   classState: Record<string, ClassState>;
-  classOverlay: Record<string, Controls.OverlayClassState>;
+  classOverlay: Partial<Record<string, Controls.OverlayClassState>>;
 }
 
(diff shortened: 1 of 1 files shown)


weave-js/src/components/Panel2/ImageWithOverlays.tsx
```diff
@@ -220,7 +220,7 @@ const OverlayCanvas = styled.canvas`
 interface SegmentationCanvasProps {
   segmentation: ImageData;
   classState: Record<string, ClassState>;
-  classOverlay: Record<string, Controls.OverlayClassState>;
+  classOverlay: Partial<Record<string, Controls.OverlayClassState>>;
```
ericakdiaz, line 223: without the `Partial` it tells Typescript that every single `string` key will be in the Object. so line 240 didn't make sense to me at first because Typescript said `classOverlay.all` is always defined. but it's actually not always defined.

KyleGoyette (approved):

jwlee64 (approved):

PR #4133, opened 2025-04-15
chore(weave): Cleanup errors from empty storage URI

tssweeney:


gtarpenning (approved):
nit: `if not file_storage_uri`

PR #4134, opened 2025-04-15
feat(weave): load json strings and parse dict to array

zbirenbaum:
## Description

Parse JSON strings as objects for openinference and convert prompt/completion fields for openllmetry to arrays


zbirenbaum:
Closing because this is done in #4143

PR #4136, opened 2025-04-15
perf(weave): filter `trace_id` pre groupby

gtarpenning:
## Description

[WB-24427](https://wandb.atlassian.net/browse/WB-24427)

This pr pulls out the `trace_id` filter construction in the calls query, and puts it before the group by using a non-aggregating condition. We want to do this because the more we can reduce the number of traces that hit the groupby, the less peak memory is consumed. Why is this safe? Crucially, **all** valid call starts have a `trace_id`. This means that we can always compare trace ids to any trace id filter, while also accepting **all** rows where `trace_id` is NULL as these **must** be non-call start rows. This pr also adds an additional minor performance enhancement where we treat the special case of exactly one `trace_id` differently, comparing with an `=` instead of the `IN` array operator. 

### Example

Prod:
```sql
SELECT
    calls_merged.id AS id
FROM calls_merged
WHERE calls_merged.project_id = 'project'
    AND ((any(calls_merged.trace_id) IN ['trace-id']))
```

Branch:
```sql
SELECT
    calls_merged.id AS id
FROM calls_merged
WHERE calls_merged.project_id = 'project'
    AND (calls_merged.trace_id = 'trace-id' OR calls_merged.trace_id IS NULL)
```

## Testing

- unit tests

### Benchmarking

| Query  | Latency | Rows read | Result Size | Memory usage |
|--------|---------|-----------|-------------|---------------|
| Prod   | 13.296s | 293,398   | 164.43 MB   | 942.87 MiB    |
| Branch | 6.223s  | 290,616   | 164.43 MB   | 965.37 MiB    |

[WB-24427]: https://wandb.atlassian.net/browse/WB-24427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1109,6 +1116,37 @@ def process_op_name_filter_to_conditions(
...
+
+    # If there's only one trace_id, use an equality condition for performance
+    if len(trace_ids) == 1:
+        trace_cond = f"{trace_id_field_sql} = {param_slot(param_builder.add_param(trace_ids[0]), 'String')}"
+    elif len(trace_ids) > 1:
+        trace_cond = f"{trace_id_field_sql} IN {param_slot(param_builder.add_param(trace_ids), 'Array(String)')}"
+    else:
+        return ""
+
+    return f" AND ({trace_cond} OR {trace_id_field_sql} IS NULL)"
```
chance-wnb, line 1147: Adding `OR {trace_id_field_sql} IS NULL)"` seems making the logic different than what's been removed from line 1141-1146. Would you mind sharing your thoughts on this?

chance-wnb (commented):
This PR looks great. Thanks! 

I have a few questions:

1. why we need to take `trace_id is NULL` , what scenarios are we guarding against?
2. Do you have plans to test the performance boost with this change ? (happy to off load or help in any way possible)
3. On what occasions do we expect `trace_ids` filters being used?

weave/trace_server/calls_query_builder/calls_query_builder.py
tssweeney: I agree - I think you still need 1141-1146 since you can now over-select

gtarpenning: discussion moved [here](https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1744754060675289?thread_ts=1744748084.998869&cid=C03BSTEBD7F). 

To your point chance: We are moving this filter handling from after the `GROUP BY (project_id, id)` to directly in the `WHERE`. This means we are operating on raw rows, rather than grouped rows, so we need to handle more cases. See the slack thread for a breakdown of cases.

gtarpenning: I could be wrong, its good to point out and discuss

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -715,6 +715,12 @@ def _as_sql_base_format(
             table_alias,
         )
 
+        trace_id_sql = process_trace_id_filter_to_conditions(
```
gtarpenning, line 718: Add comment explaining the other use for the `started_at is NULL` condition

chance-wnb (approved):
Thank you for the change!

PR #4141, opened 2025-04-16
chore(weave): Fixes pydantic serialization issues

tssweeney:
1. When serializing pydantic objects we: 1) should respect the alias when serializing and 2) respect field exclusion
2. It should be trivial to load a registered object with nested pydantic fields.

Specific Changes:
1. The serialization procedure correctly uses the pydantic alias when extracting field names AND respects excluded fields (customer ask actually)
2. All Weave wrapper classes now support `.unwrap` (WeaveObject, WeaveTable, WeaveList, WeaveDict, ObjectRecord)


docs/docs/reference/gen_notebooks/leaderboard_quickstart.md
```diff
@@ -306,7 +306,7 @@ We can also generate a leaderboard directly from Python:
 
 ```python
 from weave.flow import leaderboard
-from weave.trace.weave_client import get_ref
+from weave.trace.ref_util import get_ref
```
tssweeney, line 309: many of the files changed are due to this symbol move

weave/trace/ref_util.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
```
tssweeney, line 1: this file is jus an extraction to avoid circular deps

tests/integrations/google_genai/test_google_genai.py
```diff
@@ -34,12 +34,12 @@ def test_content_generation_sync(client):
     trace_name = op_name_from_ref(call.op_name)
     assert trace_name == "google.genai.models.Models.generate_content"
     assert call.output is not None
-    assert call.output.usage_metadata.candidates_token_count > 0
-    assert call.output.usage_metadata.prompt_token_count > 0
+    assert call.output.usageMetadata.candidatesTokenCount > 0
```
tssweeney, line 37: this whole file's changes are the result of using aliases correctly now

neutralino1 (approved):
This works for my branch.
Does it help @jamie-rasmussen ?

PR #4143, opened 2025-04-16
feat(weave): Refactor otel parsing to standardize fields

zbirenbaum:
## Description
- Fixes WB-24513
- Fixes WB-24514

This adjusts the parsing logic to make it simple to generate and keep docs in line with the parser code. This adjusts the parser such that a set of standard fields are created for each of input, output, summary and attributes. Each field is assigned a list of keys which are checked until a valid entry is discovered or the possible key entries are exhausted.

Also includes cost calculation fix

PRs into master which should be reviewed and merged before reviewing this:
- #4177

PRs into this branch which should be reviewed and merged before reviewing this:
- #4181


weave/trace_server/opentelemetry/python_spans.py
```diff
@@ -268,14 +279,38 @@ def as_dict(self) -> dict[str, Any]:
...
+
+        # We failed to load any of the Weave attributes, dump all attributes
+        if not has_attributes and not has_inputs and not has_outputs and not has_usage:
+            attributes = to_json_serializable(self.attributes)
+
+        attributes["otel_span"] = self.as_dict()
 
-        # Options: set
         start_call = tsi.StartedCallSchemaForInsert(
             project_id=project_id,
```
parambharat, line 315: @zacharyblasczyk : I came across this PR as I was implementing the `pydantic_ai` otel integration, one quick request I had was if it was possible to set the `display_name: Optional[str] = None` in `StartedCallSchemaForInsert`. We can map it to an attribute like `weave.display_name` from the attributes dict of the span. This can help with having nice display names for the calls in the UI.

zacharyblasczyk: I assume you meant @zbirenbaum.

zbirenbaum: @parambharat That's possible but I think the better route would be to try and set display name based on the attribute and fall back to the field we currently use to determine the name if it isn't there. Otherwise we would end up with traces that have a perfectly fine name to use being unnamed if that attribute isn't set

weave/trace_server/opentelemetry/attributes.py
```diff
@@ -313,199 +308,90 @@ def unflatten_key_values(
...
+        and "total_tokens" not in usage
+    ):
+        usage["total_tokens"] = usage["prompt_tokens"] + usage["completion_tokens"]
+    return usage
 
-        if extra:
-            attributes.update(extra)
 
-        return to_json_serializable(attributes)
+def get_weave_inputs(_: list[SpanEvent], attributes: dict[str, Any]) -> dict[str, Any]:
```
zbirenbaum, line 392: Passing the events here even though they are unused because there are a few exporters which make use of event attributes. We don't support this yet and should add it as a followup in the next release.

weave/trace_server/opentelemetry/attributes.py
```diff
@@ -313,199 +308,90 @@ def unflatten_key_values(
...
+            value = get_attribute(attributes, attribute_key)
+            if value:
+                if attribute_key in KEY_HANDLERS:
+                    try:
+                        value = KEY_HANDLERS[attribute_key](value)
+                    except:
+                        pass
+                result[key] = value
+                break
+    return result
```
zbirenbaum, line 367: `to_json_serializable` here just to be safe. Haven't been able to trigger a json serialization error yet but would be bad if somehow got triggered.

weave/trace_server/opentelemetry/python_spans.py
```diff
@@ -268,18 +281,48 @@ def as_dict(self) -> dict[str, Any]:
...
-        inputs = self.attributes.get_weave_inputs()
-        outputs = self.attributes.get_weave_outputs()
-        attributes = self.attributes.get_weave_attributes(
-            extra={"otel_span": self.as_dict()}
+        events = [SpanEvent(e.as_dict()) for e in self.events]
+        usage = get_weave_usage(self.attributes) or {}
+        inputs = get_weave_inputs(events, self.attributes) or {}
+        outputs = get_weave_outputs(events, self.attributes) or {}
+        attributes = get_weave_attributes(self.attributes) or {}
+        llm_usage = tsi.LLMUsageSchema(
```
tssweeney, line 289: not to future devs: llm_usage is not auto rolled up here like it is in python clients which will create different UI experiences.

weave/trace_server/opentelemetry/attributes.py
```diff
@@ -313,199 +308,90 @@ def unflatten_key_values(
...
-        total_tokens = self.get_attribute_value(oi.SpanAttributes.LLM_TOKEN_COUNT_TOTAL)
-        return LLMUsageSchema(
-            prompt_tokens=int(prompt_tokens) if prompt_tokens else None,
-            completion_tokens=int(completion_tokens) if completion_tokens else None,
-            total_tokens=int(total_tokens) if total_tokens else None,
-        )
 
-    def get_weave_summary(self) -> SummaryInsertMap:
-        return SummaryInsertMap(usage={"usage": self.get_weave_usage()})
+KEY_HANDLERS = {
```
tssweeney, line 331: for clarity it would be nice to colocate your mappings with the handlers

weave/trace_server/opentelemetry/constants.py
```diff
@@ -0,0 +1,56 @@
+INPUT_KEYS = {
+    "value": [
+        "weave.input",  # Weave
```
tssweeney, line 3: hmmm - how does weave get in here?

weave/trace_server/opentelemetry/constants.py
```diff
@@ -0,0 +1,56 @@
+INPUT_KEYS = {
```
tssweeney, line 1: I would document this file - this seems to be your ultimate source of truth for mapping

weave/trace_server/opentelemetry/constants.py
zbirenbaum: This is addressed in https://github.com/wandb/weave/pull/4181 I'd review it first

weave/trace_server/opentelemetry/constants.py
zbirenbaum: Lets merge in https://github.com/wandb/weave/pull/4181 and https://github.com/wandb/weave/pull/4182/files since it changes some stuff and then I'll document it

weave/trace_server/opentelemetry/constants.py
```diff
@@ -0,0 +1,47 @@
+INPUT_KEYS = [
```
zbirenbaum, line 1: Comment this file

tssweeney (approved):

weave/trace_server/opentelemetry/attributes.py
zbirenbaum: I'll make a followup ticket for this. I definitely agree but right now it will cause a circular dependency to put this in constants.py and it's much clearer to have constants.py isolated.

I'll open a ticket for a followup to move the helper functions which include the handlers used here to `helpers.py`. I'd rather not obscure the functional changes in this diff further by just moving around large blocks of code.

zbirenbaum: I ended up just doing it so no need to worry about follow up

ayulockin:
Hey @zbirenbaum and @tssweeney when are we anticipating merging this PR and deploy the trace server with updates?

zbirenbaum:
> Hey @zbirenbaum and @tssweeney when are we anticipating merging this PR and deploy the trace server with updates?

It was ideally going to be merged friday but it got stuck in CI due to some change in dspy integration tests. It should be merged today

weave/trace_server/opentelemetry/constants.py
```diff
@@ -0,0 +1,121 @@
...
+}
+
+# ATTRIBUTE_KEYS: Maps common LLM call metadata attributes to the types of attributes expected in weave traces
+# This is used to populate the `attributes_dump` column in clickhouse
+# Prior to dumping, the full span is dumped under another key not listed here called `otel_span`
+ATTRIBUTE_KEYS = {
+    # System prompt/instructions
+    "system": [
+        "gen_ai.system",  # OpenTelemetry AI
+        "llm.system",  # OpenInference
```
vanpelt, line 99: Likely not a big deal, but in our Input / Output keys OpenInference is before OT AI

zbirenbaum: I'll switch it real quick for consistency but it shouldn't actually make a difference here, the same information is recorded in each key so even if they have both it will work properly.

weave/trace_server/opentelemetry/helpers.py
```diff
@@ -0,0 +1,440 @@
...
+        return str(value)
+    elif isinstance(value, Enum):
+        return value.value
+    elif hasattr(value, "__dataclass_fields__"):  # Handle dataclasses
+        return {
+            k: to_json_serializable(getattr(value, k))
+            for k in value.__dataclass_fields__
+        }
+    else:
+        raise ValueError(f"Unsupported type for JSON serialization: {type(value)}")
```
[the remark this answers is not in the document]
zbirenbaum:
```
    elif isinstance(value, (bytes, bytearray)):
        return base64.b64encode(value).decode('ascii')
```

Thanks for pointing those out, this should make it a bit more robust. Only note: is this the way we want to handle bytes values?

PR #4148, opened 2025-04-16
fix(weave_ts): Make credential handling match the wandb sdk

vanpelt:
## Description

- Fixes WB-22092

This fixes broken credential handling in weave TS.  Before we were assuming netrc would always exist and we just read the last entry in it which is REALLY broken.  This respects WANDB_BASE_URL or the ~/.config/wandb/settings file and bombs out if it can't write the netrc (which happens by default when running an environment like Deno).

## Testing

I wrote (or I should say Cursor wrote) a bevy of amazing tests.


sdks/node/src/wandb/settings.ts
```diff
@@ -1,45 +1,78 @@
...
+    const configPath = path.join(os.homedir(), '.config', 'wandb', 'settings');
+    if (fs.existsSync(configPath)) {
+      const config = ini.parse(fs.readFileSync(configPath, 'utf-8'));
+      if (config.default?.wandb_base_url) {
+        return config.default.wandb_base_url;
+      }
+    }
+  } catch (error) {
+    // Silently fail if we can't read the config file
+    // This could happen in restricted environments like Deno
```
andrewtruong, line 26: Why fail silently?

sdks/node/src/wandb/settings.ts
```diff
@@ -1,45 +1,78 @@
...
+}
+
+function getApiKeyFromNetrc(host: string): string | undefined {
+  try {
     const netrc = new Netrc();
-    apiKey = netrc.entries.get(host)?.password;
+    return netrc.entries.get(host)?.password;
+  } catch (error) {
+    // Silently fail if we can't read the netrc file
+    // This could happen in restricted environments like Deno
```
andrewtruong, line 38: Why fail silently?

sdks/node/src/__tests__/wandb/settings.test.ts
```diff
@@ -1,82 +1,127 @@
...
 
-    it('returns API key from netrc file', () => {
-      MockedNetrc.prototype.entries = new Map([
-        [
-          'api.wandb.ai',
-          {machine: 'api.wandb.ai', login: 'user', password: 'netrc-api-key'},
-        ],
-      ]);
+    test('returns API key from netrc when environment variable is not set', () => {
+      delete process.env.WANDB_API_KEY;  // Ensure env var is not set
```
andrewtruong, line 40: isn't this already guaranteed  above?

sdks/node/src/__tests__/wandb/settings.test.ts
```diff
@@ -1,82 +1,127 @@
...
       expect(getApiKey('api.wandb.ai')).toBe('netrc-api-key');
     });
 
-    it('throws error when no API key is found', () => {
-      MockedNetrc.prototype.entries = new Map();
-      expect(() => getApiKey('api.wandb.ai')).toThrow(
-        'wandb API key not found'
-      );
+    test('throws error when no API key is found', () => {
+      delete process.env.WANDB_API_KEY;  // Ensure env var is not set
```
andrewtruong, line 52: isn't this already guaranteed above?

sdks/node/src/__tests__/wandb/settings.test.ts
```diff
@@ -1,82 +1,127 @@
...
+      const mockNetrc = {
+        entries: new Map(),
+      };
+      (Netrc as jest.Mock).mockImplementation(() => mockNetrc);
+
+      expect(() => getApiKey('api.wandb.ai')).toThrow('wandb API key not found');
+    });
+
+    test('handles netrc read errors gracefully', () => {
+      delete process.env.WANDB_API_KEY;  // Ensure env var is not set
```
andrewtruong, line 62: isn't this already guaranteed above?

sdks/node/src/__tests__/wandb/settings.test.ts
```diff
@@ -1,82 +1,127 @@
...
+        resolvedHost: 'custom.wandb.ai',
+        domain: 'custom.wandb.ai',
       });
     });
 
-    it('throws error when no netrc entry is found', () => {
-      // Mock netrc with no entries
-      MockedNetrc.prototype.getLastEntry = jest.fn().mockReturnValue(null);
+    test('reads base URL from config file', () => {
+      delete process.env.WANDB_BASE_URL;  // Ensure env var is not set
```
andrewtruong, line 87: isn't this already guaranteed above?

sdks/node/src/__tests__/wandb/settings.test.ts
```diff
@@ -1,82 +1,127 @@
...
+      (fs.existsSync as jest.Mock).mockReturnValue(true);
+      (fs.readFileSync as jest.Mock).mockReturnValue(mockConfig);
+      process.env.WANDB_API_KEY = 'test-api-key';
+
+      const configs = getWandbConfigs();
+      expect(configs.resolvedHost).toBe('custom.wandb.ai');
+    });
+
+    test('falls back to default host when no custom URL is configured', () => {
+      delete process.env.WANDB_BASE_URL;  // Ensure env var is not set
```
andrewtruong, line 101: isn't this already guaranteed above?

sdks/node/src/__tests__/wandb/settings.test.ts
```diff
@@ -1,82 +1,127 @@
...
+        domain: 'wandb.ai',
+      });
     });
 
-    it('throws error when netrc throws error', () => {
-      // Mock netrc throwing error
-      MockedNetrc.prototype.getLastEntry = jest.fn().mockImplementation(() => {
-        throw new Error('Failed to read netrc');
+    test('handles config file read errors gracefully', () => {
+      delete process.env.WANDB_BASE_URL;  // Ensure env var is not set
```
andrewtruong, line 116: isn't this already guaranteed above?

sdks/node/src/__tests__/wandb/settings.test.ts
vanpelt: Yeah, good old cursor code gen was being extra careful 😉 .  I'll remove

sdks/node/src/wandb/settings.ts
vanpelt: The failure case is if the file exists but we're unable to parse it.  This should be exceedingly rare, but since we can't really do anything about it from this SDK (wandb login --host=xxxx sets it), I figure we just pretend it doesn't exist and default to api.wandb.ai

sdks/node/src/wandb/settings.ts
vanpelt: Again the failure mode is either the file is corrupt or I don't have read permissions for the file.  Since I can't really tell the user how to fix it, failing silently here will ultimately end up in failing with a no credentials error if the user hasn't set an api key in their environment.

sdks/node/src/wandb/settings.ts
```diff
@@ -1,45 +1,78 @@
...
     
     You can either:
     
-    1. Set the WANDB_API_KEY environment variable.
+    1. Set the WANDB_API_KEY environment variable
     2. Log in using weave.login()
-    3. Add your API key to your .netrc file, in a stanza like this:
-        machine ${domain}
-            login user
-            password <your-wandb-api-key>
```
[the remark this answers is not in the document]
vanpelt: In what world would we want users hand editing their `.netrc` files?  This just reeks of shitty UX.  Ultimately we should likely provide a weave cli command that let's them run `weave login` from the command line to write it.  Also, this message is worthless to most devs.  What's a `.netrc` file?  Where do I find it? etc.

andrewtruong (approved):

PR #4150, opened 2025-04-16
chore(weave): add max_execution_time to ch client

gtarpenning:
## Description

Are there any reasonable database queries that we expect to return in over 60 seconds? I think there is an argument to be made that well before this mark we need to be handling the query in a different way.

When requests do take longer than 60 seconds, it's very likely they are consuming high amounts of database memory and CPU, and users are probably not waiting around for it to complete. There are also frequent cases where users kick off an initial query, realize that it's not finishing, and start another, perhaps with a different filter/query. In this case, we probably want to have some reasonable max limit on the execution of that query. 

This will increase the number of queries that error! We should make sure to improve the error handling in both clients to accurately convey to the user that the query timed out. 

Here is an example of one or two rogue queries completely pegging one of the CPU nodes on the db cluster:
Screenshot 2025-06-03 at 10 16 05 AM

This pr sets a default to 1 minute, with the ability to override, which may be important for specific instances that have long queries.


weave/trace_server/environment.py
```diff
@@ -38,6 +38,17 @@ def wf_clickhouse_max_memory_usage() -> Optional[int]:
...
 
+def wf_clickhouse_max_execution_time() -> Optional[int]:
+    """The maximum execution time for the clickhouse server."""
+    time = os.environ.get("WF_CLICKHOUSE_MAX_EXECUTION_TIME")
+    if time is None:
+        return None
+    try:
+        return int(time)
+    except ValueError:
+        return None
```
neutralino1, line 49: Do we want to silence incorrect values?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -190,10 +190,15 @@
...
+# https://clickhouse.com/docs/operations/settings/settings#max_memory_usage
 DEFAULT_MAX_MEMORY_USAGE = 16 * 1024 * 1024 * 1024  # 16 GiB
+# https://clickhouse.com/docs/operations/settings/settings#max_execution_time
+DEFAULT_MAX_EXECUTION_TIME = 60 * 1  # 1 minute
 CLICKHOUSE_DEFAULT_QUERY_SETTINGS = {
     "max_memory_usage": wf_env.wf_clickhouse_max_memory_usage()
-    or DEFAULT_MAX_MEMORY_USAGE
+    or DEFAULT_MAX_MEMORY_USAGE,
+    "max_execution_time": wf_env.wf_clickhouse_max_execution_time()
+    or DEFAULT_MAX_EXECUTION_TIME,
```
neutralino1, line 201: Does this bring performance improvements or is this just a guardrail to avoid rogue queries?

neutralino1 (commented):
A couple of comments.

weave/trace_server/environment.py
gtarpenning: I think silently failing here is better than crashing the pod if the var is incorrect. It's better to not apply the fix than brick I think here...

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: just a guardrail, there are some 5+ minute queries that i'm SURE users are not sitting around waiting for....

neutralino1 (approved):

tssweeney (approved):

PR #4162, opened 2025-04-17
chore(weave): Add server-side respect of pydantic aliases

tssweeney:
Ensures we correctly support pydantic aliases for server-side object validation


tests/trace/test_base_object_classes.py
```diff
@@ -53,15 +53,15 @@ def test_pythonic_creation(client: WeaveClient):
...
+        nested_base_model=TestOnlyNestedBaseModel(a=2, aliased_property_alias=3),
         nested_base_object=weave.publish(nested_obj).uri(),
     )
     ref = weave.publish(top_obj)
 
     top_obj_gotten = weave.ref(ref.uri()).get()
 
     assert isinstance(top_obj_gotten, base_objects.TestOnlyExample)
-    assert top_obj_gotten.model_dump() == top_obj.model_dump()
+    assert top_obj_gotten.model_dump(by_alias=True) == top_obj.model_dump(by_alias=True)
```
jamie-rasmussen, line 64: We could consider this trick to avoid having to `by_alias` so much:
https://github.com/pydantic/pydantic/discussions/8725#discussioncomment-11722435

weave/trace_server/object_class_util.py
```diff
@@ -128,8 +129,9 @@ def dump_object(val: BaseModel) -> dict:
     # Order matters here due to the way we calculate the digest!
     # This matches the client
     dump["_type"] = cls_name
-    for k in val.model_fields:
-        dump[k] = _general_dump(getattr(val, k))
+    d = pydantic_asdict_one_level(val)
```
jamie-rasmussen, line 132: It would be good to have an inline comment here explaining as neither pydantic_asdict_one_level nor _general_dump has a docstring.

Seeing the comment above, are there any consequences of this change in terms of digest calculation?

jamie-rasmussen (approved):

weave/trace_server/object_class_util.py
tssweeney: in fact - this is important (: After adding the new field to the test with the alias, the digests had mismatch! so this brings the server in line with the client (:

PR #4166, opened 2025-04-17
feat(weave): Imperative Evaluation APIs

andrewtruong:
This is the SDK portion of the larger Framework-agnostic Eval API improvements
https://github.com/wandb/weave/pull/4005

### DIFF
--- a/weave/__init__.py
+++ b/weave/__init__.py
@@ -11,6 +11,7 @@ from weave.flow.agent import AgentState as AgentState
 from weave.flow.annotation_spec import AnnotationSpec
 from weave.flow.dataset import Dataset
 from weave.flow.eval import Evaluation
+from weave.flow.eval_imperative import ImperativeEvaluationLogger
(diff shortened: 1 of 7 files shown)


weave/__init__.py
```diff
@@ -37,6 +38,7 @@
     attributes,
     # Re-exported from flow module
     Object,
+    ImperativeEvaluationLogger,
```
tssweeney, line 41: maybe we exclude this from this PR (effectively keeping it hidden). Then brining it to the top level is the "release moment".

tssweeney: I also suspect Adrian/Scott/Shawn will have opinions about the name "ImparativeEvaluationLogger" lol

andrewtruong: we can just call it `EvaluationLogger` -- that is implicitly imperative I think

tssweeney (approved):

PR #4177, opened 2025-04-18
fix(weave): Fix trace export error when op_name is too long

zbirenbaum:
## Description
- Fixes WB-24498

Adds a `shorten_name` helper and validates `op_name` is not too long for otel traces

## Testing

Added unit tests for introduced helpers and added regression test for modified version of the failing payload

### DIFF
--- a/weave/trace_server/opentelemetry/helpers.py
+++ b/weave/trace_server/opentelemetry/helpers.py
@@ -1,0 +1,55 @@
+import re
+
+# TODO: Move shared helpers into this file
+# This logic doesn't really belong in any existing files, but there is logic in existing files that belongs here.
+
+def capture_parts(
+    s: str,
+    delimiters: list[str]=[",", ";", "|", " ", "/", "?", "."]
+) -> list[str]:
+    """
+    # Split a string based on multiple patterns, including the split characters in the result.
+    """
+
+    # Escape special regex characters and join with | for regex alternation
+    capture = '|'.join(map(re.escape, delimiters))
+    pattern = f"({(capture)})"
+
+    # Use re.split with capturing groups to include the delimiters
+    parts = re.split(pattern, s)
+
+    # Filter out empty strings that might result from the split
+    result = [part for part in parts if part != '']
+
+    result = list(filter(lambda x: len(x) > 0, parts))
+    # If no split occurred, return the original string in a list
+    if len(result) == 0:
+        return [s]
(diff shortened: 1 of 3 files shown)


weave/trace_server/opentelemetry/helpers.py
```diff
@@ -0,0 +1,72 @@
...
+
+    result = list(filter(lambda x: len(x) > 0, parts))
+    # If no split occurred, return the original string in a list
+    if len(result) == 0:
+        return [s]
+
+    return result
+
+
+def shorten_name(name: str, max_len: int, abbrv: str = "...") -> str:
```
tssweeney, line 30: a code comment on the intended functionality would be nice.

While not required, we have a much stricter sanitization for object names here: sanitize_object_name (this is more restrictive than is required for op names).

Moreover - probably don't need to do this right now, but FYI: the 128 is sort of out of thin air - we could increase if it makes sense.

One more thought / optional: if you truncate, that leaves open the possibility of making two distinct things indistinguishable. Consider hashing the entire name then adding a 4 character suffix at the end if it is truncated.

tssweeney (approved):
Approving for expediency, but please consider my comment

weave/trace_server/opentelemetry/helpers.py
zbirenbaum: Yeah a comment would be good, I'll add one here.
I like the hashing recommendation. Which one of these is closer to what you were thinking
1.  the shortening logic here but instead of doing the next delimiter plus `...` we do `:hash(full_name)[:4]`
2. remove the shortening logic. Just return `f"{hash(name)[:max_len-5]}:{uuid.uuid4()[:4]"`

weave/trace_server/opentelemetry/python_spans.py
```diff
@@ -278,7 +279,15 @@ def to_call(
         )
         op_name = self.name
         if len(op_name) >= MAX_OP_NAME_LENGTH:
-            op_name = shorten_name(op_name, MAX_OP_NAME_LENGTH)
+            # Since op_name will typically be what is displayed, we don't want to just truncate
+            # Create an identifier abbreviation so similar long names can be distinguished
+            identifier = str(uuid.uuid4())[:4]
```
tssweeney, line 284: you do not want this to be random - you want it to be deterministic based on the content

PR #4178, opened 2025-04-18
chore(weave): LLMStructuredCompletionModel built in object mvp and types

jwlee64:
## Description
Creates a built in object definition for saved llm model instances
runs type generation
leaves predict and json_schema unimplemented for now

## Testing

How was this PR tested?


weave/trace_server/object_class_util.py
```diff
@@ -109,7 +109,7 @@ def process_incoming_object_val(
                 )
             return ProcessIncomingObjectResult(
                 val=dict_val,
-                base_object_class=new_val_object_classes["base_object_class"],
+                base_object_class=new_val_object_classes["object_class"],
```
tssweeney, line 112: this is not quite correct. indeed we actually want to save both - the base_object_class and the object_class. But if we override this here, then this does not look like a "Model" to the api

tssweeney: drat - you are hitting the edges of the system.

weave/trace/serialization/serialize.py
```diff
@@ -304,16 +301,18 @@ def from_json(obj: Any, project_id: str, server: TraceServerInterface) -> Any:
...
-            isinstance(val_type, str)
-            and obj.get("_class_name") == val_type
-            and (builtin_object_class := BUILTIN_OBJECT_REGISTRY.get(val_type))
-        ):
-            return builtin_object_class.model_validate(obj)
-        else:
-            return ObjectRecord(
-                {k: from_json(v, project_id, server) for k, v in obj.items()}
+        elif isinstance(val_type, str) and obj.get("_class_name") == val_type:
+            from weave.trace_server.interface.builtin_object_classes.builtin_object_registry import (
```
tssweeney, line 305: this is a smell - why did it become circular?

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
```diff
@@ -0,0 +1,51 @@
...
+
+class LLMStructuredCompletionModelDefaultParams(BaseModel):
+    temperature: Optional[float] = None
+    top_p: Optional[float] = None
+    max_tokens: Optional[int] = None
+    presence_penalty: Optional[float] = None
+    frequency_penalty: Optional[float] = None
+    stop: Optional[list[str]] = None
+    n_times: Optional[int] = None
+    response_format: Optional[ResponseFormat] = None
```
tssweeney, line 27: response format is twice? Once here and once in the parent?

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
```diff
@@ -0,0 +1,51 @@
...
+    functions: Optional[list[dict]] = None
+
+
+class LLMStructuredCompletionModel(Model):
+    # <provider>/<model> or ref to a provider model
+    llm_model_id: Union[str, base_object_def.RefStr]
+
+    # Could use Prompt objects for the message template
+    messages_template: list[dict]
```
tssweeney, line 36: I think it is critical to document the expected interplay of messages_template and the inputs

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
```diff
@@ -0,0 +1,51 @@
...
+    max_tokens: Optional[int] = None
+    presence_penalty: Optional[float] = None
+    frequency_penalty: Optional[float] = None
+    stop: Optional[list[str]] = None
+    n_times: Optional[int] = None
+    response_format: Optional[ResponseFormat] = None
+    functions: Optional[list[dict]] = None
+
+
+class LLMStructuredCompletionModel(Model):
```
tssweeney, line 31: Inheriting from Model will not work on the server since you can't import outside of `weave.trace_server` from here. This edges on the whole: how do we use the weave library in the server conversation

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # TODO: Currently not used. Fast follow up with json_schema
+    response_format_schema: dict
+
+    default_params: LLMStructuredCompletionModelDefaultParams = Field(
+        default_factory=LLMStructuredCompletionModelDefaultParams
+    )
+
+    @weave.op()
+    async def predict(self, **inputs: dict[str, Any]) -> dict:
```
tssweeney, line 46: what do you expect inputs to be here?

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
```diff
@@ -0,0 +1,65 @@
...
+    name: Optional[str]
+    function_call: Optional[dict]
+    tool_call_id: Optional[str]
+
+
+class LLMStructuredCompletionModelDefaultParams(BaseModel):
+    # Could use Prompt objects for the message template
+    # This is a list of Messages, loosely following litellm's message format
+    # https://docs.litellm.ai/docs/completion/input#properties-of-messages
+    messages_template: list[Message]
```
tssweeney, line 40: yes! Love this model. I think there is a slight iteration i think might generalize well. Let me start with the problem statement: given a new user input and some history, what is the correct way to construct the messages sent to the llm? I _think_ the abstraction we want is:

```
messages_sent = fill_template(config) + possibly_truncate(history) + user_template(new input)
```

firstly, does that make sense? If not, let's chat. If so, I think we need to possibly modify this slightly.

Another socratic way to get at this: what is your code that consumes this object and how does it construct the messages?

tssweeney (approved):

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    role: str
+    content: Optional[Union[str, list[dict]]]
+    name: Optional[str]
+    function_call: Optional[dict]
+    tool_call_id: Optional[str]
+
+
+class LLMStructuredCompletionModelDefaultParams(BaseModel):
```
jwlee64, line 36: This structure is intended to be used in a future predict function similar in form to
`def predict(self, user_input: list[Message] | Message, config: dict) -> Message:`

PR #4181, opened 2025-04-18
feat(weave): use otel attribute keys as input and output subfields

zbirenbaum:
## Description

- Fixes WB-24515

Required for #4143

Example for OpenInference (input.value and output.value are the attributes in the OTEL span)
image

Note: While it the key display is a bit wonky since input.value is parsed as two layers of headers, that's for the frontend to handle not the backend.

### DIFF
--- a/tests/trace_server/test_opentelemetry.py
+++ b/tests/trace_server/test_opentelemetry.py
@@ -426,7 +426,6 @@ class TestSemanticConventionParsing:
 
         # Test get_weave_attributes
         extracted = get_weave_attributes(attributes)
-        print(extracted)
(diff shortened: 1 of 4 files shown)


zbirenbaum:
Before merging there's one open question:
Since we have decided to be true to the otel attribute keys in the backend for input and output, should we be doing the same for attributes?

2 Options:
1. (current implementation): Choose a few select fields (invocation parameters, model etc) which map to our current usage of attributes for the weave sdk 
2. Dump all attributes to a dict under their attribute key as we do here and let the user choose their filters to hide them

It's pretty trivial to switch over to 2 and I'm leaning towards it but open to thoughts

PR #4182, opened 2025-04-18
feat(weave): allow setting display name through weave attribute

zbirenbaum:
## Description

- Fixes WB-24499

creates new `wandb.display_name` check for otel attributes to allow setting `display_name` for call through traces

## Testing

added unit tests

### DIFF
--- a/weave/trace_server/opentelemetry/constants.py
+++ b/weave/trace_server/opentelemetry/constants.py
@@ -40,3 +40,14 @@ ATTRIBUTE_KEYS = {
     ],
     "model_parameters": ["gen_ai.request", "llm.invocation_parameters"],
 }
+
+# Wandb/Weave specific call attributes
+WB_KEYS = {
+    "wb_user_id": ["wandb.user_id"],
+    "wb_run_id": ["wandb.run_id"],
+    "display_name": ["wandb.display_name"],
+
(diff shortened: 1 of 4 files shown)


zbirenbaum:
Should we also add wb_run_id to this in a followup?

weave/trace_server/opentelemetry/constants.py
```diff
@@ -40,3 +40,9 @@
     ],
     "model_parameters": ["gen_ai.request", "llm.invocation_parameters"],
 }
+
+# Wandb/Weave specific call attributes
+WB_KEYS = {
+    "display_name": ["wandb.display_name"],
+    "project_id": ["wandb.project_id"],
```
tssweeney, line 47: this seems like a potential auth issue. cannot blindly trust this

zbirenbaum: project_id is totally unused at the moment. I added it here because it's related to a followup PR I am working through.

tssweeney: please remove this

zbirenbaum: removed in https://github.com/wandb/weave/tree/zach/no-synthetic-io-keys

PR #4190, opened 2025-04-18
chore(weave): dump option click filter values to string

gtarpenning:
## Description

[WB-24519](https://wandb.atlassian.net/browse/WB-24519)

"option click" filters use the raw value, which is fine for the initial filter, until you change the operator. Lets just cast to string, which should also always work (also tested for bool)

Prevents the weird class of errors highlighted in the ticket. It is still possible to create these errors through the API, not exactly sure how to deal with that...

## Testing

Prod: 
filter-op-value-prod-1

Branch:
filter-op-value-branch-1

[WB-24519]: https://wandb.atlassian.net/browse/WB-24519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterRow.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterRow.tsx
@@ -10,6 +10,7 @@ import {
   FilterId,
   getFieldType,
   getGroupedOperatorOptions,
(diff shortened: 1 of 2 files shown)


neutralino1 (approved):
Thank you.

PR #4191, opened 2025-04-19
fix(weave): actually drop too-large calls and error to sentry

gtarpenning:
## Description

Errors look like [this](https://weights-biases.sentry.io/issues/6551857568/?project=4507019311251456&query=%22single%20call%20size%22&referrer=issue-stream&stream_index=0)

## Testing

👀 
Screenshot 2025-04-18 at 5 09 22 PM

### DIFF
--- a/weave/trace_server_bindings/remote_http_trace_server.py
+++ b/weave/trace_server_bindings/remote_http_trace_server.py
@@ -4,6 +4,7 @@ import logging
 from collections.abc import Iterator
 from typing import Any, Optional, Union, cast
 
+import sentry_sdk
 from pydantic import BaseModel
 
 from weave.trace.env import weave_trace_server_url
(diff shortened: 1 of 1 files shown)


adrnswanberg (approved):

tssweeney (changes requested):
I don't think we should do this (specifically: we should not be raising a value error)

jamie-rasmussen:
> I don't think we should do this (specifically: we should not be raising a value error)

I think as a general rule we shouldn't catch the overly broad `Exception`. It also seems unnecessary to raise and immediately catch the error, we could just create a `ValueError` object if we want one for logging. `ValueError` seems ok to me for exceeding a size limit, we could also use a more specific subclass.

Did you have a different objection?

tssweeney:
> > I don't think we should do this (specifically: we should not be raising a value error)
> 
> I think as a general rule we shouldn't catch the overly broad `Exception`. It also seems unnecessary to raise and immediately catch the error, we could just create a `ValueError` object if we want one for logging. `ValueError` seems ok to me for exceeding a size limit, we could also use a more specific subclass.
> 
> Did you have a different objection?

I agree with you Jamie - there is no need to raise-catch here. Moreover, we should not stop control flow. In this PR we return early. We should continue to attempt to send the data over. This is more of a warning since the max call size is an estimate and based on the server configuration.

PR #4192, opened 2025-04-19
perf(weave): address data loading perf issue on eval compare

chance-wnb:
## Description

This addresses a part of the evaluation comparison performance issue. No the customer scenario will no longer crash. 

UI wise, there is no noticeable behavior change.

Here I attach [a video(internal)](https://www.loom.com/share/e07ee35dcd464094bc06aebf97fb8962?sid=691778a9-6156-4b2f-884a-095f3a78d29c) to explain the change.

## Testing

This PR is manually tested against the customer scenario and locally.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/ExampleCompareSection.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/ExampleCompareSection.tsx
@@ -1,5 +1,6 @@
 import {Box, Tooltip} from '@material-ui/core';
 import {WarningAmberOutlined} from '@mui/icons-material';
+import {LoadingDots} from '@wandb/weave/components/LoadingDots';
 import _ from 'lodash';
 import React, {useCallback, useEffect, useMemo, useRef} from 'react';
 import styled from 'styled-components';
@@ -50,6 +51,7 @@ import {
 } from '../ScorecardSection/ScorecardSection';
 import {
   PivotedRow,
+  useExampleCompareData,
   useFilteredAggregateRows,
 } from './exampleCompareSectionUtil';
 
@@ -225,6 +227,16 @@ export const ExampleCompareSection: React.FC<{
     return filteredRows[targetIndex];
   }, [filteredRows, targetIndex]);
 
+  const {targetRowValue, loading: loadingInputValue} = useExampleCompareData(
+    props.state,
+    filteredRows,
+    targetIndex
+  );
+
+  const inputColumnKeys = useMemo(() => {
+    return Object.keys(targetRowValue ?? {});
+  }, [targetRowValue]);
+
   const [selectedTrials, setSelectedTrials] = React.useState<{
     [evalCallId: string]: number;
   }>({});
@@ -251,9 +263,8 @@ export const ExampleCompareSection: React.FC<{
   );
 
   const inputRef = parseRef(target.inputRef) as WeaveObjectRef;
-  const inputColumnKeys = Object.keys(target.input);
-  const numInputProps = inputColumnKeys.length;
-  const numOutputKeys = outputColumnKeys.length;
+  const numInputProps = inputColumnKeys?.length ?? 0;
+  const numOutputKeys = outputColumnKeys?.length ?? 0;
 
   const numTrials = orderedCallIds.map(leafId => {
     return target.originalRows.filter(row => row.evaluationCallId === leafId)
@@ -448,7 +459,7 @@ export const ExampleCompareSection: React.FC<{
 
   const inputPropValComp = (inputPropIndex: number) => {
     return (
-      <ICValueView value={target.input[inputColumnKeys[inputPropIndex]]} />
+      <ICValueView value={targetRowValue?.[inputColumnKeys[inputPropIndex]]} />
     );
   };
 
@@ -759,16 +770,29 @@ export const ExampleCompareSection: React.FC<{
             </React.Fragment>
           )}
           {/* INPUT ROWS */}
-          {_.range(numInputProps).map(inputPropIndex => {
-            return (
-              <React.Fragment key={inputPropMapKey(inputPropIndex)}>
-                <GridCell style={{...stickySidebarStyleMixin}}>
-                  {inputPropKeyComp(inputPropIndex)}
-                </GridCell>
-                <GridCell>{inputPropValComp(inputPropIndex)}</GridCell>
-              </React.Fragment>
-            );
-          })}
+          {loadingInputValue || targetRowValue === undefined ? (
+            <React.Fragment key={'loading'}>
+              <GridCell style={{...stickySidebarStyleMixin}}>
+                <LoadingDots />
+              </GridCell>
+              <GridCell>
+                <LoadingDots />
+              </GridCell>
+            </React.Fragment>
+          ) : (
+            <>
+              {_.range(numInputProps).map(inputPropIndex => {
+                return (
+                  <React.Fragment key={inputPropMapKey(inputPropIndex)}>
+                    <GridCell style={{...stickySidebarStyleMixin}}>
+                      {inputPropKeyComp(inputPropIndex)}
+                    </GridCell>
+                    <GridCell>{inputPropValComp(inputPropIndex)}</GridCell>
+                  </React.Fragment>
+                );
+              })}
+            </>
+          )}
         </GridCellSubgrid>
         {/* OUTPUT SECTION */}
         <GridCellSubgrid
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
@@ -1,7 +1,19 @@
-import _ from 'lodash';
-import {useMemo} from 'react';
+import {WeaveObjectRef} from '@wandb/weave/react';
+import {parseRef} from '@wandb/weave/react';
+import _, {isEmpty} from 'lodash';
+import {
+  MutableRefObject,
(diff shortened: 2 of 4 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -323,9 +302,181 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
       outputColumnKeys,
       leafDims,
     };
-  }, [filteredRows, inputColumnKeys, leafDims, outputColumnKeys]);
+  }, [filteredRows, leafDims, outputColumnKeys]);
 };
+
+// Get the table digest used in the dataset of the first evaluation,
+// which prepares a request for actually fetching the table rows later
+async function getPartialTableRequest(
```
gtarpenning, line 313: super nit, maybe we can call this `makePartialTableReq` so that its clear this is generating a req payload.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -323,9 +302,181 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
+
+  const [loading, setLoading] = useState<boolean>(false);
+
+  const targetRowValue = useMemo(() => {
+    if (isEmpty(filteredRows)) {
+      return undefined;
+    }
+    const digest = filteredRows[targetIndex].inputDigest;
+    return flattenObjectPreservingWeaveTypes(cachedRowData.current[digest]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
gtarpenning, line 398: I think when we override this specific rule it makes sense to add a comment to explain exactly why we need to do that.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -323,9 +302,181 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
+  }, [mutationCounter, filteredRows, targetIndex]);
+
+  useEffect(() => {
+    (async () => {
```
gtarpenning, line 402: does this have to be an anonymous function? might be more readable if we named it.

gtarpenning: also because this is performance sensitive, anonymous functions are so much more annoying to debug in the javascript console.

gtarpenning (commented):
I played around with it and its really nice, I did notice one small bug which is that if you click the next example button rapidly, we get into a state where the query fires but the loading state if permanently stuck. I think this has to do with the mutation counter but i'm not sure. A very edge case, but because we don't have a way of going to a specific example (which I think we should eventually add btw), I could imagine a user spamming the next example button to get to the one they know is interesting. We might be able to get away with just loading 5 ahead or something 🤷🏻 
small-eval-perf-example-bug

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
gtarpenning: nvm, this lgtm

chance-wnb:
> I played around with it and its really nice, I did notice one small bug which is that if you click the next example button rapidly, we get into a state where the query fires but the loading state if permanently stuck. I think this has to do with the mutation counter but i'm not sure. A very edge case, but because we don't have a way of going to a specific example (which I think we should eventually add btw), I could imagine a user spamming the next example button to get to the one they know is interesting. We might be able to get away with just loading 5 ahead or something 🤷🏻

Great job on discovering this bug, it is a solid bug. This is because in the `loadRowDataIntoCache()` I was passing `cachedRowData.current` which is the at-the-time value of the cache, which does not reflect the latest value during race conditions. My fix is to change the parameter to pass the refs (the boxes object), and the issue went away.

gtarpenning (approved):
lgtm, love the code removal. I think the cache handling is neat. I trust that the mutation counter is common practice, and after watching your video it makes sense; @tssweeney might be curious about it!

chance-wnb:
> lgtm, love the code removal. I think the cache handling is neat. I trust that the mutation counter is common practice, and after watching your video it makes sense; @tssweeney might be curious about it!

In regard to the `mutation counter` thingy, I now feel that semantically giving it a better name seems to justify its usage better. now I renamed it to `cacheVersion`.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksEvaluationComparison.ts
```diff
@@ -499,30 +499,6 @@ const fetchEvaluationComparisonResults = async (
       });
     });
 
-  // 3.5 Populate the inputs
```
tssweeney, line 502: Given this, you are no long populating `inputs` in 
```
export type EvaluationComparisonResults = {
  // Inputs are the intersection of all inputs used in the evaluations.
  // Note, we are able to "merge" the same input digest even if it is
  // used in different evaluations.
  inputs: {
    [rowDigest: string]: DatasetRow;
  };
```

as such, I would either:
a) populate this field using other forms (ex. the predict_and_score input) or
b) remove this field from the type definition and force resolution of any typing issues caused from this

tssweeney: i think (b) corresponds with this PR

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -130,50 +141,40 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
-              });
-            }
+            const output = predictAndScoreRes._rawPredictTraceData?.output;
+            rows.push({
+              id: predictAndScoreRes.callId,
+              evaluationCallId: predictAndScoreRes.evaluationCallId,
+              inputDigest: predictAndScoreRes.rowDigest,
+              inputRef: predictAndScoreRes.exampleRef,
+              output: flattenObjectPreservingWeaveTypes({output}),
+              scores: Object.fromEntries(
```
tssweeney, line 151: in @andrewtruong's upcoming PR, we can't count that the data comes from a dataset (gasp!) it might be good to at least leave a comment here as this would be a possible location to record the raw predict_and_score inputs as the presumed data row.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/ExampleCompareSection.tsx
```diff
@@ -225,6 +227,16 @@ export const ExampleCompareSection: React.FC<{
     return filteredRows[targetIndex];
   }, [filteredRows, targetIndex]);
 
+  const {targetRowValue, loading: loadingInputValue} = useExampleCompareData(
```
tssweeney, line 230: my reading of the implementation of `useExampleCompareData` is that filteredRows is only used to get the index using `filteredRows[targetIndex]`. It might be cleaner to just pass `filteredRows[targetIndex]` to `useExampleCompareData`, simplifying the params and internal logic of `useExampleCompareData`

chance-wnb: filteredRows should be needed to invalidate the cache in case the filter condition changes.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -323,9 +309,183 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
+  const targetRowValue = useMemo(() => {
+    if (isEmpty(filteredRows)) {
+      return undefined;
+    }
+    const digest = filteredRows[targetIndex].inputDigest;
+    return flattenObjectPreservingWeaveTypes(cachedRowData.current[digest]);
+    // Including `cacheVersion` in the dependency array ensures the memo recalculates
+    // when it changes, even though it's not directly used in the calculation.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [cacheVersion, filteredRows, targetIndex]);
```
tssweeney, line 408: I am not the biggest fan of this pattern (in particular, explicitly depending on `cacheVersion` that is not used in the hook). But I would agree it is hard to get around this one. Approving, but do think there might be a cleaner way.

tssweeney (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooksEvaluationComparison.ts
chance-wnb: Thanks for catching this. It was an oversight, it should be removed.

PR #4203, opened 2025-04-21
chore(weave): fix api key name for gemini

jwlee64:
## Description

we call it google_api_key in the trace server
this causes users to add a wrongly named api key when configuring from the drawer in playground

## Testing

How was this PR tested?


gtarpenning (approved):
It looks like we use `GEMINI_API_KEY` in the integration tests too

jwlee64:
Hmm actually I think I approached this wrong, I think the correct decision is ther other way around and to change the backend to match the current standard of using GEMINI_API_KEY

PR #4204, opened 2025-04-21
fix(ui): inverted photo in filter tag

jamie-rasmussen:
## Description

This is a little ugly. For context, night mode inverts everything with a CSS filter, and then an additional rule for images normally inverts that (not perfectly, but that's besides the point here). However, tags, including the Filters tag, use a `useTagClasses` hook which applies `night-aware` which overrides this. The solution in this PR is to apply night-aware to just the user filter tag contents to invert that yet again, making the profile picture look right.

A better solution would be to figure out why tags were set to be night aware not just for themselves but for their children as well. However, that would require a larger code audit than I think we want to take on right now.

Before:
Screenshot 2025-04-21 at 1 29 13 PM

After:
Screenshot 2025-04-21 at 1 29 32 PM

## Testing

How was this PR tested?


gtarpenning (approved):
a _little_ ugly?? more like terrifying

PR #4210, opened 2025-04-21
chore(weave_query): clarify behaviour of cache bucketing

nicholaspun-wandb:
## Description

Update incorrect example in docstring and verify example behaviour in unit tests

--- 

<details>
<summary>PR was updated to only change the docstring and unit tests, here's the original PR description</summary>

This PR corrects the behaviour of our cache bucketing method to what I believe is the intended implementation.

---

Currently, timestamps starting at the current interval and up to the next interval will be assigned into the same bucket. Concretely, given an interval of 7 days and the days: [day 0, day 1, ..., day 6]. All 7 days will be assigned the "day 7" bucket. Consequently, the following unit test will fail: 

```
  # self.DAY_IN_SECONDS = 24 * 60 * 60
   def test_bucket_should_advance_when_time_passes(self):
        INTERVAL_DAYS = 7
        INITIAL_TIME = 100
        with mock.patch("time.time", return_value=100):
            initial_bucket = int(cache.bucket_timestamp(INTERVAL_DAYS), 10)
            assert initial_bucket == INTERVAL_DAYS * self.DAY_IN_SECONDS

        with mock.patch("time.time", return_value=INITIAL_TIME + self.DAY_IN_SECONDS):
            print("INITIAL BUCKET: ", initial_bucket)
            print("NEW BUCKET: ", cache.bucket_timestamp(INTERVAL_DAYS))
            assert (
                int(cache.bucket_timestamp(INTERVAL_DAYS), 10)
                == initial_bucket + self.DAY_IN_SECONDS
            )
```
with the following buckets returned:
```
INITIAL BUCKET:  604800
NEW BUCKET:  604800
```

---

This differs from the explanation provided in the comment, which provides the following example:
```
system time             bucketed time
11:22 Jan 1 1970       00:00 Jan 8 1970
12:00 Jan 3 2024       00:00 Jan 10 2024
```
So as time advances by the day, we should see that the bucket also advances. Concretely, back to our example with an interval of 7 days and the days: [day 0, day 1, ... day 6], the buckets should be: [day 0 + 7, day 1 + 7, ..., day 6 + 7]. With the changes to the method in this PR, the unit test above passes, and we can see that the buckets are different.
```
INITIAL BUCKET:  604800
NEW BUCKET:  691200
```
The new bucket being one day later than the initial since the mock timestamps differ by one day.

</details>

## Testing

Unit Tests

### DIFF
--- a/weave_query/tests/test_cache.py
+++ b/weave_query/tests/test_cache.py
@@ -1,6 +1,7 @@
 import datetime
 import os
 import time
+from unittest import mock
 
 from weave_query import cache, environment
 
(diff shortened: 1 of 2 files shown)


jwlee64 (approved):
the original implementation was so that 

days 1, days 2, days 3, ... , day 7 all end up one bucket
even if we start day 3 the bucket doesnt change

ie the buckets are all already preset, no matter the start point

day 8 - day 14 all end up one bucket regardless of when it started

Though switching the implementation is welcome :)

nicholaspun-wandb:
Ohh, wait so in that docstring, should the bucketed times be the same?
```
system time             bucketed time
00:00 Jan 8 1970       00:00 Jan 15 1970
16:00 Jan 12 1970      00:00 Jan 19 1970
```

jwlee64:
@nicholaspun-wandb Oh yes that docstring is very much wrong, so sorry about that Jan 3,1970 and Jan 4,1970 would both point to Jan 7th,1970 bucket, since the bucket is calculated from epoch start

domphan-wandb (approved):
Approving to unblock because the bucketing has been improved with testing despite a slightly different intended implementation.

I think this new implementation provides more granular cache invalidation, so lgtm! 👍

PR #4212, opened 2025-04-21
chore(weave): improve loading state management for default date filters

gtarpenning:
## Description

This pr:
- refactors `GridFilterItem` and `GridFilterModel` to `ExtendedGridFilterItem` and `ExtendedGridFilterModel` so that we can track more information on the filter. Specifically, we want to keep track of if the filter was applied by the user or was a default filter. Currently we are using a heuristic, but that breaks when we want to more aggressively condition on it. 
- Prolongs the `callsTable` loading state to wait until we are done adjusting the datetime filter.
- Slightly changes the `useCallsNoExpansion` loading state management, moving from a ref to explicitly use state hooks. This allows us to set the `loading=false` state after setting `callRes` which leads to one fewer render where the `loading` and `results` states are out of sync. 

This pr does not **solve** the issue, but reduces the prevalence of this specific race condition. The root of the issue comes from inconsistent results from the hook, where the hook returns "all good, done loading" but the results are empty. This happens most consistently when repeatedly refreshing the hook, but because we are caching our datetime filter selection, this should not be an issue. 

## Testing

Prod

Prod (caching disabled)

Branch (caching disabled)

[WB-24243]: https://wandb.atlassian.net/browse/WB-24243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
@@ -2,7 +2,6 @@ import {ApolloProvider} from '@apollo/client';
 import {Box, Drawer} from '@mui/material';
 import {
   GridColumnVisibilityModel,
-  GridFilterModel,
   GridPaginationModel,
   GridPinnedColumnFields,
   GridSortModel,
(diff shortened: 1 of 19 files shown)


gtarpenning:
closing in favor of: https://github.com/wandb/weave/pull/4237

PR #4216, opened 2025-04-22
fix(weave): Fix Weave self-managed docs

flamarion:
## Description

This change fixes the docs to prevent users from deploying the latest version of the Bitnami helm chart, which wasn't tested and doesn't work with the configuration from this document.

- Remove image reference (breaks the config if not using Bitnami image)
- Fix the persistence volume size
- Fix the cache size
- Fix the install command
- Remove namespace references.

## Testing

Running the docs system locally.

### DIFF
--- a/docs/docs/guides/platform/weave-self-managed.md
+++ b/docs/docs/guides/platform/weave-self-managed.md
@@ -66,12 +66,6 @@ Modify the following parameters:
 W&B recommends keeping the `clusterName` value in `values.yaml` set to `weave_cluster`.  This is the expected cluster name when W&B Weave runs the database migration. If you need to use a different name, see the [Setting `clusterName`](#setting-clustername) section for more information.
 
 ```yaml
-# Stable version
-image:
-  registry: docker.io
-  repository: bitnami/clickhouse
-  tag: 24.8
-
 ## @param clusterName ClickHouse cluster name
 clusterName: weave_cluster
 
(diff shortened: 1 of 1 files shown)


docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -66,12 +66,6 @@ Modify the following parameters:
 W&B recommends keeping the `clusterName` value in `values.yaml` set to `weave_cluster`.  This is the expected cluster name when W&B Weave runs the database migration. If you need to use a different name, see the [Setting `clusterName`](#setting-clustername) section for more information.
 
 ```yaml
-# Stable version
-image:
-  registry: docker.io
-  repository: bitnami/clickhouse
-  tag: 24.8
```
raubitsj, line 73: Where did specifying the clickhouse image go to?

raubitsj (approved):

docs/docs/guides/platform/weave-self-managed.md
flamarion: I tried to test using the stable version, but it seems the Bitnami image has something else that prevents the stable version from working correctly.
That's the reason I removed it.
The image is set in the chart version `8.0.10` is this:

```
image:
  registry: docker.io
  repository: bitnami/clickhouse
  tag: 25.3.2-debian-12-r3  
```

PR #4217, opened 2025-04-22
fix(ui): react 18 createRoot warning

jamie-rasmussen:
## Description

Fixes this error:
Screenshot 2025-04-21 at 9 40 40 AM

This was also the root cause behind a Saved Views bug with filters.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/VariableChildrenDisplayer.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/VariableChildrenDisplayer.tsx
@@ -5,7 +5,11 @@
 
 import _ from 'lodash';
 import React, {isValidElement, ReactChild, useEffect, useState} from 'react';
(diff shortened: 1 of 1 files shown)


neutralino1 (commented):
How much time to fix this the right way?

We need to be able to keep upgrading smoothly and not be stuck on one version.

Maybe we can hold off on this fix and you can debug the tests after Saved Views are merged?

jamie-rasmussen:
> How much time to fix this the right way?
> 
> We need to be able to keep upgrading smoothly and not be stuck on one version.
> 
> Maybe we can hold off on this fix and you can debug the tests after Saved Views are merged?

I don't have a time estimate, I'm not familiar with either the failing tests or the code they are testing.

neutralino1 (approved):
Can we create a JIRA ticket to take care of this soon?

jamie-rasmussen:
> Can we create a JIRA ticket to take care of this soon?

Created https://wandb.atlassian.net/browse/WB-24568

PR #4222, opened 2025-04-22
chore(weave): add error and span logging for inserts

gtarpenning:
## Description

This pr: 
- logs error information (`table`, `data_len`, `data_bytes`, `example_data` (first row), `column_names`, `settings`) when an insert fails
- adds the count of stripped payloads to the dd trace in `_strip_large_values`

## Testing

- really hard to test the error reporting, i'm not sure how to cause the error, thats why we need the additional logging.


neutralino1 (approved):
Thank you.

PR #4228, opened 2025-04-22
chore(weave): Fix nulls returning from cost query

jwlee64:
## Description
Fixes issue where unfinished calls returned nulls in the cost query

2 errors
1. if we did default to the proper empty usage json, it was not properly escaped and the query would error returning null
fixed default usage not being properly escaped
2. we needed to account for the default usage being {}

https://github.com/user-attachments/assets/2a8eaa17-5a9b-4db0-a473-6b96a40f1452

## Testing

updated test
check the fetching works for finished, unfinished and failed calls

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -9,6 +9,7 @@ from collections import defaultdict, namedtuple
 from contextlib import contextmanager
 from contextvars import copy_context
 from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


tssweeney (commented):
Please write a unit test that sends only a start call and then makes the requests, demonstrating the correct end-user behavior

neutralino1 (approved):
Thank you.

PR #4234, opened 2025-04-22
feat(weave): Communicate trace endpoint errors to users

chance-wnb:
## Description

- Fixes [WB-24053](https://wandb.atlassian.net/browse/WB-24053)

Communicate trace endpoint errors to users

In case of a trace call loading error due to network or auth issues, user will see this:

image

Ok, based on [the latest feedback](https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1745524789840019?thread_ts=1745451631.673699&cid=C03BSTEBD7F): 

I have decided to show StatusChip in place of the costs/storage size errors, which would look like this:

image

when a user hovers it, we will see:

image

## Testing

Manually tested with auth error and a forged bad server response

[WB-24053]: https://wandb.atlassian.net/browse/WB-24053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -26,6 +26,7 @@ import {
 import {MOON_200, TEAL_300} from '@wandb/weave/common/css/color.styles';
 import {Button} from '@wandb/weave/components/Button';
 import {Checkbox} from '@wandb/weave/components/Checkbox/Checkbox';
+import {ErrorPanel} from '@wandb/weave/components/ErrorPanel';
 import {
   Icon,
   IconNotVisible,
@@ -464,9 +465,11 @@ export const CallsTable: FC<{
     allowedColumnPatterns,
     onAddFilter,
     calls.costsLoading,
+    !!calls.costsError,
     shouldIncludeTotalStorageSize,
     shouldIncludeTotalStorageSize ? calls.storageSizeResults : null,
-    shouldIncludeTotalStorageSize && calls.storageSizeLoading
+    shouldIncludeTotalStorageSize && calls.storageSizeLoading,
+    !!calls.storageSizeError
   );
 
   // This contains columns which are suitable for selection and raw data
@@ -794,6 +797,41 @@ export const CallsTable: FC<{
     [callsLoading, setPaginationModel]
   );
 
+  const noRowsOverlay = useCallback(() => {
+    if (calls.primaryError) {
+      return (
+        <ErrorPanel
+          title="Oh no! Unable to load traces!..."
+          error={calls.primaryError}
+        />
+      );
+    }
+    return (
+      <CallsTableNoRowsOverlay
+        entity={entity}
+        project={project}
+        callsLoading={callsLoading}
+        callsResult={callsResult}
+        isEvaluateTable={isEvaluateTable}
+        effectiveFilter={effectiveFilter}
+        filterModelResolved={filterModelResolved}
+        clearFilters={clearFilters}
+        setFilterModel={setFilterModel}
+      />
+    );
+  }, [
+    calls.primaryError,
+    callsLoading,
+    callsResult,
+    clearFilters,
+    effectiveFilter,
+    filterModelResolved,
+    isEvaluateTable,
+    setFilterModel,
+    entity,
+    project,
+  ]);
+
   // CPR (Tim) - (GeneralRefactoring): Pull out different inline-properties and create them above
   return (
     <FilterLayoutTemplate
@@ -1032,7 +1070,7 @@ export const CallsTable: FC<{
         // SORT SECTION END
         // PAGINATION SECTION START
         pagination
-        rowCount={callsTotal}
+        rowCount={calls.primaryError ? 0 : callsTotal}
         paginationMode="server"
         paginationModel={paginationModel}
         onPaginationModelChange={onPaginationModelChange}
@@ -1066,19 +1104,7 @@ export const CallsTable: FC<{
           },
         }}
         slots={{
-          noRowsOverlay: () => (
-            <CallsTableNoRowsOverlay
-              entity={entity}
-              project={project}
-              callsLoading={callsLoading}
-              callsResult={callsResult}
-              isEvaluateTable={isEvaluateTable}
-              effectiveFilter={effectiveFilter}
-              filterModelResolved={filterModelResolved}
-              clearFilters={clearFilters}
-              setFilterModel={setFilterModel}
-            />
-          ),
+          noRowsOverlay,
           columnMenu: CallsCustomColumnMenu,
           pagination: () => <PaginationButtons hideControls={hideControls} />,
           columnMenuSortDescendingIcon: IconSortDescending,
@@ -1089,6 +1115,7 @@ export const CallsTable: FC<{
           ),
           columnMenuPinRightIcon: IconPinToRight,
         }}
+        className="tw-style"
       />
     </FilterLayoutTemplate>
   );
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -801,7 +801,7 @@ export const CallsTable: FC<{
     if (calls.primaryError) {
       return (
         <ErrorPanel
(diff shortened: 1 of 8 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -1009,7 +1044,7 @@ export const CallsTable: FC<{
         // SORT SECTION END
         // PAGINATION SECTION START
         pagination
-        rowCount={callsTotal}
+        rowCount={calls.error ? 0 : callsTotal}
```
chance-wnb, line 1047: This is necessary because of MUI only shows the overlay when rowCount is 0.

https://github.com/mui/mui-x/blob/f7b122d45ac103f421f4569ec372e8c88919394d/packages/x-data-grid/src/hooks/features/overlays/useGridOverlays.tsx#L30C41-L30C47

gtarpenning: This sort of brings up a deeper philosophical point, what happens when we can return 99% of a call stream, but one of the calls errors. Should we show nothing? Should we show the calls that we were able to return? Should be Show as many as we can, with a warning to the user that there was an error and the data may be incomplete? I'm not sure that we should always do this... Maybe we need a partial error state, perhaps this is where the red toast can come in!

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -771,6 +772,38 @@ export const CallsTable: FC<{
...
   );
 
+  const noRowsOverlay = useCallback(() => {
+    if (calls.error) {
+      return (
+        <ErrorPanel title="Trace endpoint hit an error" error={calls.error} />
+      );
+    }
+    return (
+      <CallsTableNoRowsOverlay
```
gtarpenning, line 782: nit: might be cleaner (or not) to move this handling inside `CallsTableNoRowsOverlay` and add the error prop. We are still displaying the `noRowsOverlay`, it just looks different because of the error.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
```diff
@@ -154,6 +155,10 @@ export const useCallsForQuery = (
     }
   );
 
+  const error = useMemo(() => {
+    return calls.error || costs.error || storageSize.error || null;
```
gtarpenning, line 159: this is another more philosophical decision. I personally think we probably should not show the full error page when one of the auxiliary queries fails! The cost query can be more computationally expensive, so there are cases where the trace table loads no problem, but when we fetch costs the database can't handle it, and it errors. I think we should still show the traces, but somehow convey to the user that cost data won't be coming. Idk... maybe another example of this was we had an incident yesterday where the cost query was slightly bugged and returned `null` when a call was in progress. This caused the whole page to load forever and many people wrote in about the issue. With this change, it feels similar, now there is another way we can fail to show data to the user. Maybe this is just an edge case, we shouldn't have any erroring queries anyway.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -307,9 +308,11 @@ const useCallsNoExpansion = (
     const onError = (e: any) => {
       loadingRef.current = false;
       console.error(e);
+      setError(e);
       setCallRes({calls: []});
     };
-    getTsClient().callsStreamQuery(req).then(onSuccess).catch(onError);
+    setError(null);
+    getTsClient().callsStreamQuery(req).then(onSuccess, onError);
```
gtarpenning, line 315: I have literally never seen `.then(x, y)`. cool

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -771,6 +772,38 @@ export const CallsTable: FC<{
     [callsLoading, setPaginationModel]
   );
 
+  const noRowsOverlay = useCallback(() => {
+    if (calls.error) {
+      return (
+        <ErrorPanel title="Trace endpoint hit an error" error={calls.error} />
```
gtarpenning, line 778: I think this copy should get another pass, "trace endpoint" feels like a leaky abstraction. We probably want something more folksy, like "Oh no! Unable to load traces!..." Or something less technical.

chance-wnb: Yeah, very good suggestion!

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
chance-wnb: > Maybe this is just an edge case, we shouldn't have any erroring queries anyway.

I am lean towards failing entirely in case of any error. Because I think a partially succeed query is also a failed query.
It might increase the system complexity when we need to explain to the customer what failed what not. Both engineering-wise to us and UI-wise to the users.

If there is an error we should fix, and we should make the error evident. We definitely don't want the customers to be happy with 8 traces where she is supposed to have 10 right? :slightly_smiling_face:

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
chance-wnb: I tried to work towards this direction then found that the logic inside `CallsTableNoRowsOverlay` is quite complicated. It has some hooks which loads data. And since hooks are not conditional, I have to involuntarily go through those hooks for the error case despite I have no use of them. To conditionally avoid those hooks, I can branch off from a parent component, but that simply boils down to what I have right now. 

So keeping it as is might be most economical.

chance-wnb:
* **#4234** <a href="https://app.graphite.dev/github/pr/wandb/weave/4234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4234?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
chance-wnb: Yeah, thanks for the insight! Through some internal discussion, I have decided to show the primary call results if those are present, and inline the error display in the table cells for costs and storage size load errors.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
chance-wnb: Thanks, the latest resolution is updated in the description. Thanks for the discussion, I appreciate it!

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
chance-wnb: Thanks, the latest resolution is updated in the description. Thanks for the discussion, I appreciate it!

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -794,6 +797,41 @@ export const CallsTable: FC<{
     [callsLoading, setPaginationModel]
   );
 
+  const noRowsOverlay = useCallback(() => {
+    if (calls.primaryError) {
+      return (
+        <ErrorPanel
+          title="Oh no! Unable to load traces!..."
```
gtarpenning, line 804:
```suggestion
          title="Oh no! Unable to load traces..."
```

gtarpenning (approved):
LGTM!

PR #4235, opened 2025-04-22
chore(weave): fix analytics event to be less noisy

gtarpenning:
## Description

Fix the analytics on datetime filter bar changes. Currently noisily logs all interactions, but we really only want to log when the user changes something for the first time. 

## Testing

👀


chance-wnb (approved):
LGTM. Thanks!

PR #4236, opened 2025-04-23
fix(app): Fix bounding box filter

ericakdiaz:
## Description

This is similar to https://github.com/wandb/weave/pull/4131 - for bounding boxes as well on point clouds and images, we should respect the same rules. 

## Testing

How was this PR tested?

Tested in https://github.com/wandb/core/pull/28975


weave-js/src/common/util/PointCloudFiltering.ts
```diff
@@ -70,7 +83,10 @@ export type ClassIdToLabelMap = Map<number, string>;
 // transformation was tricky/more error prone than this way.
 // TODO: merge this and isBoundingBoxHidden.
 const isBoundingBoxVisible = (box: SceneBox, filter: Filter) => {
-  if (filter.hideAllBoxes) {
+  if (
+    filter.hideAllBoxes &&
+    (box.label == null || !filter.shownBoundingBoxLabels.includes(box.label))
+  ) {
```
ericakdiaz, line 89: boxes that are manually shown after all are hidden should be still shown. this respects the consistent rule of the most specific setting taking precedence.

weave-js/src/components/Panel2/ImageWithOverlays.tsx
```diff
@@ -409,7 +409,7 @@ const isBoundingBoxHidden = (
   const allDisabled =
     classOverlayStates.all?.disabled ?? DEFAULT_ALL_MASK_CONTROL.disabled;
 
-  if (allDisabled) {
+  if (allDisabled && classOverlayStates?.[box.class_id]?.disabled !== false) {
```
ericakdiaz, line 412: same here but for images

weave-js/src/common/util/PointCloudFiltering.ts
```diff
@@ -33,28 +35,39 @@ export const getFilterFromBBoxConfig = (
...
+      } else {
+        shownBoundingBoxLabelsSoFar.push(label);
       }
 
-      resultSoFar.push(label);
-      return resultSoFar;
+      return [hiddenBoundingBoxLabelsSoFar, shownBoundingBoxLabelsSoFar];
     },
-    []
+    [[], []]
```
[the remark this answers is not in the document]
ericakdiaz: ahh just noticed PointCloudFiltering already has unit tests! i added one in for this case :)

weave-js/src/common/util/PointCloudFiltering.ts
ericakdiaz: added it in to be more clear!

domphan-wandb (approved):

PR #4237, opened 2025-04-23
fix(weave): fix useCalls hook to return correct loading state

gtarpenning:
## Description

The useCallsNoExpansion hook does not work properly when the request parameters change before the initial result has returned. This edge case now occurs on first page load with the default time filter when the project doesn't have enough calls to render the page in the last week. We learn from the stats query that there isn't enough, normally before the stream query returns, and then change the filter accordingly. However, the hook returns `loading=false` when the _first_ result returns, which is usually some hundreds of milliseconds (can be more) before the second query using the same hook returns. During this time we show an ugly "no result rows" text block to the user on the trace table.

The fix is in the hook: hash the request, and only update the loading and call state if the current request matches what we expect. When there is a mismatch, we know we are trying to set the state for an out of date request!

## Testing

Master (caching turned off):
master-no-cache-1

Branch (caching turned off):
branch-no-cache

Branch:
branch-cache

[WB-24243]: https://wandb.atlassian.net/browse/WB-24243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -313,40 +309,58 @@ const useCallsNoExpansion = (
...
+  }, [makeRequest]);
+
+  // Keep track of the request hash we're waiting for, so that we
+  // can ignore requests that are superceded by more recent reqs
+  const expectedRequestHashRef = useRef(currentRequestHash);
+
+  const doFetch = useCallback(() => {
+    if (opts?.skip) {
+      return;
+    }
```
neutralino1, line 338: Redundant with line 379.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -313,40 +309,58 @@ const useCallsNoExpansion = (
...
     columns,
-    getTsClient,
+    opts?.includeCosts,
+    opts?.includeFeedback,
+    opts?.includeTotalStorageSize,
   ]);
 
+  // Create a hash of the current request parameters
+  const currentRequestHash = useMemo(() => {
+    return JSON.stringify(makeRequest());
```
neutralino1, line 328: I'm guessing `JSON.stringify` sorts keys?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -313,40 +309,58 @@ const useCallsNoExpansion = (
...
+      if (expectedRequestHashRef.current === currentRequestHash) {
+        loadingRef.current = false;
+        setCallRes(res);
+      }
+    };
+    const onError = (e: any) => {
+      // Only update state if this response matches our current request
+      if (expectedRequestHashRef.current === currentRequestHash) {
+        loadingRef.current = false;
+        console.error(e);
```
neutralino1, line 357: Unsure if it's good practice to log errors. The console prints failing requests anyway. RIght?

neutralino1 (changes requested):
A couple of nits.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
gtarpenning: This was here before, i'm happy to remove, but in general this should only happen when something is wrong, so I don't hate getting the users attention.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
gtarpenning: you are generating a new object each time, try:
```javascript
const payload = {"a": 1, "b": 2}
JSON.stringify(payload)
JSON.stringify(payload)
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
chance-wnb: > The console prints failing requests anyway. RIght? 

Yes, but the error bubbles up several levels, it could be possible that the error was reprocessed into something else other than the raw HTTP failure.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
gtarpenning: oh, but because I DRYed it out with the `makeRequest` function we are actually assigning twice. good catch

gtarpenning: Hmm, well because its in the `useMemo` hook it should only ever fire when one of the makeRequest dependencies changes: 
```
entity,
    project,
    deepFilter,
    limit,
    offset,
    sortBy,
    query,
    columns,
    opts?.includeCosts,
    opts?.includeFeedback,
    opts?.includeTotalStorageSize,
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
chance-wnb: BTW, in my PR https://github.com/wandb/weave/pull/4234/files I collect this error and make it copy-paste-able from a UI copy button. So user can give us this error through support channel.

It is not relevant to this PR, it's just a FYI.

gtarpenning: Chance do you think I should remove the console log, if we are going to display it to the user anyway?

gtarpenning: Chance do you think I should remove the console log, if we are going to display it to the user anyway?

chance-wnb: I will vote for keeping it. Simply because I don't think its a big deal. Seldomly users look at the console log. And it helps do the development and notice issues in production.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -313,40 +308,51 @@ const useCallsNoExpansion = (
...
+
+  const doFetch = useCallback(() => {
+    setCallRes(null);
+    loadingRef.current = true;
+    expectedRequestStrRef.current = currentRequestStr;
+
+    const req = makeQueryReq();
+    const onSuccess = (res: traceServerTypes.TraceCallsQueryRes) => {
+      // Only update state if this response matches our current request
+      if (expectedRequestStrRef.current === currentRequestStr) {
```
chance-wnb, line 340: How does this compare to lodash `isEqual`? It might be more performant than serialization + string compare.

I got inspired by seeing the implementation of `useDeepMemo`. that's what uses `isEqual`.

gtarpenning: good idea, performance is comparable, but its simpler now.

chance-wnb (approved):
Thank you Griffin!

neutralino1 (approved):
Thank you!

PR #4240, opened 2025-04-23
feat(ui): Saved Views

jamie-rasmussen:
## Description

Internal Jira: https://wandb.atlassian.net/browse/WB-21701

Adds "Saved Views" to the UI and SDK. Views are saved as Weave objects. Table settings on the Traces and Evaluations pages like filters, sorting, page size, column visibility, and column pinning are stored in the view object.

New views can be saved, views can be updated, renamed, and deleted.

### DIFF
--- a/tests/trace/test_saved_view.py
+++ b/tests/trace/test_saved_view.py
@@ -1,0 +1,379 @@
+import datetime
+
+import pytest
+
+import weave
+from weave.flow.saved_view import (
+    Filter,
+    Filters,
+    filters_to_query,
+    query_to_filters,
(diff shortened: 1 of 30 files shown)


tssweeney (approved):
In order to unblock, pre-approving conditioned on consideration of above comments.

jamie-rasmussen:
I was not properly forbidding the rename action on the default view. (And wasn't checking the readonly permissions correctly for that either.) Renaming the default view caused numerous downstream problems.

I added a pill to the switch panel to indicate the default view.

PR #4245, opened 2025-04-23
fix(weave): remove duplicate project_id condition

gtarpenning:
## Description

We explicitly add a `project_id` condition when joining on the feedback table, but this is not necessary.  As far as I can tell, it is impossible to generate a feedback join without the existing `project_id` condition, making this redundant and safe to remove. 

Introduced in [this pr](https://github.com/wandb/weave/pull/2865). 

## Testing

Updated unit tests.


neutralino1 (approved):
Thank you!

PR #4250, opened 2025-04-23
perf(weave): add purpose built endpoint for project data

gtarpenning:
## Description

[WB-24603](https://wandb.atlassian.net/browse/WB-24603)

Adds a purpose built endpoint for checking if a project has traces. Query `project_check` with a request payload of the `project_id` and get back a `has_data` bool. 

Fast follow in core: https://github.com/wandb/core/pull/28989

## Testing

- Unit test
- Manual performance testing, in high trace environment:

| Environment | Elapsed Time | Rows Read  | Data Read | Peak Memory   | Error |
|-------------|--------------|------------|-----------|----------------|--------|
| Prod        | 3.5s         | 53,890,051 | –         | 31.94 GiB      | True   |
| Branch      | 0.008s       | 15,631     | 1.24 MB   | 126.63 MiB     | False  |

In prod, this query is run 1.2 million times a week, so even modest improvements in speed might actually have an impact on clickhouse performance. 

| Env | Runs | Errors | Avg. Read Rows | Avg. Memory Usage | p5 | p90 | p99 |
|-------------|------------------|--------|----------------|-------------------|-------------|-------------|-------------|
| Prod        | 1,169,821        | 0      | 30,333.25      | 95.86 MiB         | 0.019s      | 0.04s       | 0.541s      |

[WB-24603]: https://wandb.atlassian.net/browse/WB-24603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/test_trace_server.py
+++ b/tests/trace/test_trace_server.py
@@ -2,6 +2,7 @@ import urllib
 
 import pytest
 
+import weave
 from weave.trace_server import trace_server_interface as tsi
 
 
@@ -80,3 +81,26 @@ def test_robust_to_url_sensitive_chars(client):
         )
     )
     assert read_res.vals[0] == bad_val[bad_key]
+
+
+def test_project_check(client):
+    # real project, no data
+    proj = client._project_id()
+    res = client.server.project_check(tsi.ProjectCheckReq(project_id=proj))
+    assert not res.has_data
+
+    # fake project
+    res = client.server.project_check(
+        tsi.ProjectCheckReq(project_id="shawn/proj-does-not-exist")
+    )
+    assert not res.has_data
+
+    # real project w/ data
+    @weave.op
+    def log():
+        return "a"
+
+    log()
+
+    res = client.server.project_check(tsi.ProjectCheckReq(project_id=proj))
+    assert res.has_data
--- a/tests/trace/test_trace_server.py
+++ b/tests/trace/test_trace_server.py
@@ -95,7 +95,7 @@ def test_project_check(client):
     )
     assert not res.has_data
 
-    # real project w/ data
+    # real project w/ just calls data
     @weave.op
     def log():
         return "a"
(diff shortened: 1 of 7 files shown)


neutralino1 (commented):
Can you explain the use-case for this endpoint?

gtarpenning:
> Can you explain the use-case for this endpoint?

@neutralino1 

The frontend pr should clarify: https://github.com/wandb/weave/pull/4269

We use it for determining whether or not to show the weave UI when rendering the core models UI.

neutralino1:
I see. Is there any chance we could roll this into the `/calls/stats` endpoint? Like maybe with a special query arg...

gtarpenning:
> I see. Is there any chance we could roll this into the `/calls/stats` endpoint? Like maybe with a special query arg...

1. yes, reasonable request. I initially wanted to do this with like a `LIMIT` param. You could hit the stats endpoint with a limit, which would be useful in more than just this case. For example we could show in the UI `100+` for large expensive queries instead of returning `102,301` or whatever. 
2. However, fundamentally this query is different. It checks both calls and **objects**, and uses a very custom query construction. I think there are arguments to be made that it should be separate. Because of the objects bit it doesn't exactly fit under `calls/stats`
3. I think we will eventually want a query that returns the lastest runs in the `x` last projects. This means that we will probably eventually want a `project/` endpoint anyway. It might be `project/latest` etc. This query right now is still too expensive to work on every project though, otherwise I would have gone forward with that.

I can write up a doc if we think its an invasive enough change. In general though we are only exposing this at the service layer, so it won't be easily accessible to users, so if/when we change this endpoint it won't require back-compat etc.

neutralino1:
Re: calls+object – what if we modified both `/calls/query_stats` and `/objects/query_stats` and queried both in parallel?

Although I just realized we don't have `/objects/query_stats`. 🙃

gtarpenning:
> Re: calls+object – what if we modified both `/calls/query_stats` and `/objects/query_stats` and queried both in parallel?
> 
> Although I just realized we don't have `/objects/query_stats`. 🙃

@neutralino1 

we could do this, totally an option. One additional wrinkle though is that for `calls/query_stats` I will still be implementing custom handling for `LIMIT 1`, or adding a special flag for this query. The existing calls stats query does not work for mega large projects even with `LIMIT 1`. 

So, the options are: 

Path 1 (this pr):
- custom endpoint for just this very specific case

Path 2:
- add LIMIT to `calls/stats`, which will be useful in other cases
- add special handling for `LIMIT 1` or special flag for this specific case
- add `objects/stats` endpoint with `LIMIT` handling and maybe also special case for `LIMIT 1`

Honestly I could go either way, both seem annoying.

neutralino1:
I understand Path 2 is a little more work. My goal here is to limit custom endpoints and instead have powerful and flexible endpoints operating on primitives (calls, objects, etc.).
We will have to enhance the stats endpoint anyway for the purpose of trace plots and dashboarding. So maybe this work would align with that.

gtarpenning:
closing in favor of https://github.com/wandb/weave/pull/4331

PR #4252, opened 2025-04-23
feat(weave): Dataset.add_rows helper for efficient append to an existing dataset

bcsherma:
## Description

https://wandb.atlassian.net/browse/WB-24473

Adds public `add_rows` method on the `Dataset` class. This method takes rows to be appended to the dataset as input and returns a new dataset derived from appending the provided rows to the existing dataset. The new dataset returned by the helper has already been published to the server.

To use:

```python
import weave

weave.init("bcanfieldsherman/tldr")

dataset = weave.ref(
    "weave:///bcanfieldsherman/tldr/object/annotated_loops:33yoXWgOPrFNqhXhrDkiaoGMXPeaAjQEkocRmeRjCpk"
).get()


new_dataset = dataset.add_rows(
    [
        {
            "output": "The pull request introduces a really mid method that does not work. Whoever wrote this should be fired.",
            "reactions": ["👎", "😠", "💩"],
            "notes": ["This is really mean", "Please never do this again", "Not cool"],
        },
        {
            "output": "Whoever wrote this is a genius. I love this method.",
            "reactions": ["👍", "😊", "🚀"],
            "notes": [
                "Really good summary",
                "I love this method",
                "This is a masterpiece",
            ],
        },
    ]
)
```

### DIFF
--- a/tests/trace/test_dataset.py
+++ b/tests/trace/test_dataset.py
@@ -141,3 +141,31 @@ def test_dataset_caching(client):
 
     with raise_on_captured_errors():
         assert len(ds2) == 200
+
+
+def test_add_rows(client):
+    ds = weave.Dataset(name="test", rows=[{"a": i} for i in range(10)])
+    ref = weave.publish(ds)
+
+    ds = ref.get()
+    ds2 = ds.add_rows([{"a": 10}])
+
+    assert len(ds2) == 11
+    assert ds2.rows[10]["a"] == 10
+
+    ds3 = ds2.add_rows([{"a": 11}, {"a": 12}, {"a": 13}])
+    assert len(ds3) == 14
+    assert ds3.rows[12]["a"] == 12
+    assert ds3.rows[11]["a"] == 11
+    assert ds3.rows[10]["a"] == 10
+
+    # Verify that publishing an already published dataset doesn't
+    # do anything.
+    ds4 = weave.publish(ds3).get()
+    assert ds3.rows == ds4.rows
+
+
+def test_add_rows_to_unsaved_dataset(client):
+    ds = weave.Dataset(rows=[{"a": i} for i in range(10)])
+    with pytest.raises(TypeError):
+        ds.add_rows([{"a": 10}])
--- a/weave/flow/dataset.py
+++ b/weave/flow/dataset.py
@@ -7,10 +7,13 @@ from typing_extensions import Self
 import weave
 from weave.flow.obj import Object
 from weave.flow.util import short_str
+from weave.trace.context.weave_client_context import require_weave_client
 from weave.trace.isinstance import weave_isinstance
 from weave.trace.objectify import register_object
 from weave.trace.vals import WeaveObject, WeaveTable
-from weave.trace.weave_client import Call
+from weave.trace.weave_client import (
+    Call,
+)
 
 if TYPE_CHECKING:
     import pandas as pd
@@ -70,6 +73,48 @@ class Dataset(Object):
             raise ImportError("pandas is required to use this method")
 
         return pd.DataFrame(self.rows)
+
+    def add_rows(self, rows: Iterable[dict]) -> "Dataset":
+        """Create a new dataset version by appending rows to the existing dataset.
+
+        This is useful for adding examples to large datasets without having to
+        load the entire dataset into memory.
+
+        Args:
+            rows: The rows to add to the dataset.
+
+        Returns:
+            The updated dataset.
+        """
+        client = require_weave_client()
+        if not isinstance(self.rows, WeaveTable) or not self.rows.table_ref:
+            raise TypeError(
+                "This dataset is not saved to weave. Call weave.publish(dataset) "
+                "to save the dataset before adding rows."
+            )
+        if self.rows.table_ref.project != client.project:
+            raise ValueError(
+                "This dataset is not saved to the same project as the current weave client. "
+                "Client is in project:  "
+                + client.project
+                + " but dataset is in project: "
+                + self.rows.table_ref.project
+            )
+        if self.rows.table_ref.entity != client.entity:
+            raise ValueError(
+                "This dataset is not saved to the same entity as the current weave client. "
+                "Client is in entity: "
+                + client.entity
+                + " but dataset is in entity: "
+                + self.rows.table_ref.entity
+            )
+
+        new_table = client._append_to_table(self.rows.table_ref.digest, list(rows))
+        new_dataset = Dataset(
+            name=self.name, description=self.description, rows=new_table
+        )
+        weave.publish(new_dataset, name=self.name)
+        return new_dataset
 
     @field_validator("rows", mode="before")
     def convert_to_table(cls, rows: Any) -> Union[weave.Table, WeaveTable]:
--- a/weave/trace/vals.py
+++ b/weave/trace/vals.py
@@ -286,7 +286,7 @@ class WeaveObject(Traceable):
 
 
 class WeaveTable(Traceable):
-    filter: TableRowFilter
+    filter: Optional[TableRowFilter] = None
     _known_length: Optional[int] = None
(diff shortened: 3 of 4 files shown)


bcsherma:
* **#4252** <a href="https://app.graphite.dev/github/pr/wandb/weave/4252?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4252?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace/refs.py
```diff
@@ -317,6 +317,10 @@ def parse_object_uri(uri: str) -> ObjectRef:
     return parsed
 
 
+def format_table_uri(entity: str, project: str, digest: str) -> str:
```
chance-wnb, line 320: I think the `TableRef` struct is designed for this.

chance-wnb (approved):
LGTM! Please also check the TableRef struct

weave/flow/dataset.py
```diff
@@ -71,6 +75,64 @@ def to_pandas(self) -> "pd.DataFrame":
...
+            rows: The rows to add to the dataset.
+
+        Returns:
+            The updated dataset.
+        """
+        client = get_weave_client()
+        if client is None:
+            raise ValueError("Weave is not initialized, call weave.init().")
+        if not isinstance(self.rows, WeaveTable) or not self.rows.table_ref:
+            raise TypeError(
```
gtarpenning, line 94: we can probably try to unit test each of these errors, free testcov

weave/flow/dataset.py
```diff
@@ -71,6 +75,64 @@ def to_pandas(self) -> "pd.DataFrame":
...
+            self.rows.table_ref.digest, list(rows)
+        )
+        new_table_ref = TableRef(
+            entity=client.entity,
+            project=client.project,
+            _digest=table_update_res.digest,
+        ).uri()
+        return client._save_object(
+            {
+                # str() to prevent weave from capturing values as refs
```
gtarpenning, line 124: I don't understand, what does this mean... example?

weave/flow/dataset.py
```diff
@@ -71,6 +75,64 @@ def to_pandas(self) -> "pd.DataFrame":
...
+                "name": str(self.name),
+                "rows": new_table_ref,
+                **({"description": str(self.description)} if self.description else {}),
+                # Class metadata
+                "_type": "Dataset",
```
gtarpenning, line 129: Can we use client.publish? or like is there a helper? I _feel_ like we should be able to call `make_dataset` which calls `make_base_object[T] Dataset` and does the right validation + adds the private class metadata stuff. This is deep in the pydantic serialization world that @neutralino1 Tim and Jamie have been battling, I wonder if they have any ideas.

gtarpenning: oh client.publish doesn't exist, thats `weave.publish`

weave/flow/dataset.py
bcsherma: Screenshot 2025-04-24 at 10 13 56 AM

The name field and description fields will be shown in the oject versions table as an attr reference, instead of the plain string value

weave/flow/dataset.py
bcsherma: Yeah and unfortunately the `Dataset` object cannot currently requires `rows` to be a list of dict when it gets inited. I think there is probably way to do what you are describing and not hard code the class metadata but it will require more complex changes to the class.

neutralino1: Can we modify the `.publish` logic to detect whether `rows` is a list of dicts or a `TableRef`?

Let's explore what it'd take to do the right thing here.

neutralino1: For posterity.
Screenshot 2025-04-24 at 10 59 38 AM

bcsherma: I've updated the PR so that the client helper for appending to a table now returns a properly formed `WeaveTable`. I had to make a few fields of the constructor and attributes of `WeaveTable` optional, but the uses of those fields seem to treat them as optional anyway so I don't think this is an issue. Now the dataset helper looks much cleaner and we publish using `weave.publish` directly instead of calling internal client functions from the dataset class

weave/flow/dataset.py
bcsherma: I added tests for the no client and unpublished dataset error cases

gtarpenning (approved):
can @andrewtruong just take a sneaky peek? I think it looks good!

tests/trace/test_dataset.py
```diff
@@ -141,3 +141,32 @@ def test_dataset_caching(client):
...
+    ds2 = ds.add_rows([{"a": 10}])
+
+    assert len(ds2) == 11
+    assert ds2.rows[10]["a"] == 10
+
+    ds3 = ds2.add_rows([{"a": 11}, {"a": 12}, {"a": 13}])
+    assert len(ds3) == 14
+    assert ds3.rows[12]["a"] == 12
+    assert ds3.rows[11]["a"] == 11
+    assert ds3.rows[10]["a"] == 10
```
andrewtruong, line 160: You should additionally publish `ds2`, get it back as `ds3`, then do your assertions.

weave/flow/dataset.py
```diff
@@ -71,6 +74,50 @@ def to_pandas(self) -> "pd.DataFrame":
...
+
+        Args:
+            rows: The rows to add to the dataset.
+
+        Returns:
+            The updated dataset.
+        """
+        client = get_weave_client()
+        if client is None:
+            raise ValueError("Weave is not initialized, call weave.init().")
```
andrewtruong, line 91: nit: use `require_weave_client` which will do this for you

weave/flow/dataset.py
```diff
@@ -71,6 +74,50 @@ def to_pandas(self) -> "pd.DataFrame":
...
+                "to save the dataset before adding rows."
+            )
+        if self.rows.table_ref.project != client.project:
+            raise ValueError(
+                "This dataset is not saved to the same project as the current weave client. "
+                "Client is in project:  "
+                + client.project
+                + " but dataset is in project: "
+                + self.rows.table_ref.project
+            )
```
andrewtruong, line 104: Why does this matter?

weave/flow/dataset.py
```diff
@@ -71,6 +74,50 @@ def to_pandas(self) -> "pd.DataFrame":
...
+        if self.rows.table_ref.entity != client.entity:
+            raise ValueError(
+                "This dataset is not saved to the same entity as the current weave client. "
+                "Client is in entity: "
+                + client.entity
+                + " but dataset is in entity: "
+                + self.rows.table_ref.entity
+            )
```
andrewtruong, line 112: same here

andrewtruong (changes requested):
Please update this test:
https://github.com/wandb/weave/pull/4252/files#r2059332890

tests/trace/test_dataset.py
bcsherma: The add rows function publishes them

weave/flow/dataset.py
bcsherma: We cannot create a dataset that points to a table in a different project

tests/trace/test_dataset.py
andrewtruong: In that case I think you can still add a publish to the test and note that it won't do anything.

The case I'm trying to cover is where the user:
1. Adds rows to a ref-gotten dataset; and
2. Publishes and tries to get that dataset back

bcsherma: I added a check that calling weave.publish(...).get() doesn't change the rows

andrewtruong (approved):

tests/trace/test_dataset.py
andrewtruong: thanks!

PR #4254, opened 2025-04-24
fix(app): define default controls for masks from provided layers

nicholaspun-wandb:
## Description

- Fixes [WB-24353](https://wandb.atlassian.net/browse/WB-24353)

## Testing

https://github.com/user-attachments/assets/c54e5d81-8741-42ca-90e5-387f98dc7457


domphan-wandb (approved):
🔥

PR #4256, opened 2025-04-24
fix(ui): show less confusing empty page for object versions

jamie-rasmussen:
## Description

When you delete all versions of an object, we show "No objects yet" messaging, which is confusing. This is a quick fix to display a more correct message, though we should probably think about what we actually want the UX to be. (E.g. return to objects list page?)

Before:
image

After:

image

## Testing

wf versions example, deleted object versions.


gtarpenning (approved):
lgtm, tagging @chance-wnb because he is currently thinking through UI error / empty states

PR #4257, opened 2025-04-24
feat(weave): Expose name config and read-only props on EvaluationLogger

andrewtruong:
Now you can pass `EvaluationLogger(name="my_evaluation")` to get this:
image

You can also now get the UI URL directly from the EvaluationLogger:
```py
ev = weave.EvaluationLogger(...)
ev.ui_url  # link to your evaluation call
```

### DIFF
--- a/weave/flow/eval_imperative.py
+++ b/weave/flow/eval_imperative.py
@@ -128,6 +128,8 @@ def _cast_to_cls(type_: type[T]) -> Callable[[str | T], T]:
             return cast(T, cls())
 
         elif isinstance(value, type_):
+            if hasattr(value, "name") and not value.name:
+                value.name = _default_memorable_class_name(value)
             return value
 
         raise TypeError("Unsupported type for casting")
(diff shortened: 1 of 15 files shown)


weave/flow/eval_imperative.py
```diff
@@ -323,6 +331,16 @@ class EvaluationLogger(BaseModel):
     _evaluate_call: Call | None = PrivateAttr(None)
     _pseudo_evaluation: Evaluation = PrivateAttr()
 
+    # These properties forward useful properties of the Evaluate call onto
```
jamie-rasmussen, line 334: A little confused by plural properties, is this referring to something besides `ui_url`?

jamie-rasmussen (approved):

weave/flow/eval_imperative.py
andrewtruong: It originally included another property, but I removed it.

I'll rename

PR #4260, opened 2025-04-24
chore(weave): option click filter always uses strings, fixed

gtarpenning:
## Description

re-do of https://github.com/wandb/weave/pull/4190, with special conditioning for strings. This time I actually tested it on more than just float/number types 🤦🏻 . 

## Testing

All option clicked into the filter:
Screenshot 2025-04-24 at 8 40 13 AM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -387,6 +387,10 @@ export const CallsTable: FC<{
             field = expandedRef.field;
           }
           const op = operator ? operator : getDefaultOperatorForValue(value);
+
+          // All values added to the filter model should be strings, we
+          // only allow text-field input in the filter bar (even for numeric)
+          // They are converted during the backend mongo-style filter creation
           let strVal: string;
           if (typeof value !== 'string') {
             strVal = JSON.stringify(value);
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -388,13 +388,23 @@ export const CallsTable: FC<{
...
+          }
+
           // Check if there is an exact match for field, operator, and value in filterModel.items
           // If an exact match exists, remove it instead of adding a duplicate.
           const existingFullMatchIndex = filterModel.items.findIndex(
             item =>
               item.field === field &&
               item.operator === op &&
-              JSON.stringify(item.value) === JSON.stringify(value)
+              JSON.stringify(item.value) === strVal
```
jamie-rasmussen, line 407: I think you need to change this too, otherwise the additional quotes added on a string will prevent you from toggling it off.

gtarpenning: Good catch. fixed
remove-filter-option

gtarpenning:
> Trying this out with the wf values example it appears there are still issues with option+click on e.g. long markdown strings, dict values, etc., but I also suspect those might have been not working before this.

@jamie-rasmussen 

Yes, there are issues with Tuples too I have uncovered, but non-uniquet to this pr. I'll ticket

PR #4262, opened 2025-04-24
chore(ui): update styling for protected aliases

amusipatla-wandb:
## Description

- Fixes [WB-23977](https://wandb.atlassian.net/browse/WB-23977)

This PR updates the protected alias tag design in the (deprecated) tag component. While the component is deprecated, it is still used in various locations for artifacts. switching them out to the current component would require heavier work, so for now we're still updating this component & plan to switch to the current component later.

protected aliases design: 
- [notion link](https://www.notion.so/wandbai/Protected-alias-tag-1c8e2f5c7ef38078b75dfecdb5e80a9f)
- [figma link](https://www.figma.com/design/vXjdSVJ18xjU3T2pKnLuVd/QA-for-Workspaces?node-id=3033-26667&t=exnXegNXf7GUJyZ3-11)

## Testing
the screenshots below also include tags since those icons were also affected by this PR

protected aliases in collection card:
Screenshot 2025-04-28 at 3 43 03 PM

deletable protected aliases on versions page:
Screenshot 2025-04-28 at 3 57 57 PM

not deletable protected aliases on versions page:
Screenshot 2025-04-28 at 3 43 53 PM

[WB-23977]: https://wandb.atlassian.net/browse/WB-23977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/common/components/Tags.styles.ts
+++ b/weave-js/src/common/components/Tags.styles.ts
@@ -1,49 +1,35 @@
-import {WBIcon} from '@wandb/ui';
 import {Icon} from '@wandb/weave/components/Icon';
 import styled, {css} from 'styled-components';
 
 const IconVariants = {
   small: css`
-    font-size: 12px;
+    width: 12px;
+    height: 12px;
   `,
   medium: css`
-    font-size: 14px;
+    width: 14px;
+    height: 14px;
   `,
   large: css`
-    font-size: 16px;
+    width: 16px;
+    height: 16px;
   `,
 };
 
-export const StyledIcon = styled(WBIcon)<{
+export const StyledIcon = styled(Icon)<{
   size: keyof typeof IconVariants;
   $pos: string;
   $opacity?: number;
   $cursor?: string;
 }>`
   ${props => IconVariants[props.size]};
-  margin: 4px 4px 4px 4px;
+  margin: auto 4px;
   ${props =>
     props.$pos === 'left' ? `margin-left: -4px;` : `margin-right: -4px;`}
   display: flex;
   align-items: center;
+  justify-content: center;
+  flex-shrink: 0;
   opacity: ${props => props.$opacity};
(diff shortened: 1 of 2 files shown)


weave-js/src/common/components/Tags.styles.ts
```diff
@@ -28,3 +29,21 @@ export const Icon = styled(WBIcon)<{
   opacity: ${props => props.$opacity};
   ${props => (props.$cursor ? `cursor: ${props.$cursor};` : '')}
 `;
+
+export const ProtectedAliasIcon = styled(Icon)<{
```
amusipatla-wandb, line 33: unfortunately, `WBIcon` does not support the `lock-closed` icon we want to use, and is not backwards compatible with `tag-latest`. `tag` and `email-at` work, but the spacing/sizing is off and would require some specific custom styling as well

ssisk: note for posterity, I chatted with amulya, and directed her to: https://congenial-broccoli-daa12ae2.pages.github.io/?path=/docs/common-icons--docs for adding a new icon

ssisk (approved):
I assume that your change here does not change the dimensions of tags? if it does change the dimensions, we should do more checking of other non-registry locations tags are used (but I assume it doesn't change dimensions)

amusipatla-wandb:
> I assume that your change here does not change the dimensions of tags? if it does change the dimensions, we should do more checking of other non-registry locations tags are used (but I assume it doesn't change dimensions)

yup it doesn't change the dimensions, the height is defined by `run tag (small/medium/large)` in [Tags.less](https://github.com/wandb/core/blob/d8a1eab91b866472c114cc2f5a5bdac798686382/frontends/app/src/css/Tags.less) and I also tested with all the size options to make sure the expected style was maintained.

PR #4264, opened 2025-04-24
chore(weave): Add scoring metadata for imperative evals

andrewtruong:
Adds metadata to identify that the call is a score and comes from the imperative eval suite


neutralino1 (approved):
Thank you.
Can you add a quick PR description please?

PR #4265, opened 2025-04-24
chore(weave): allow non-admins to delete object versions

gtarpenning:
## Description

Unify object delete permissions across the different paths.

## Testing

How was this PR tested?


gtarpenning:
jamie beat me to it

PR #4269, opened 2025-04-24
chore(weave): frontend to use limited stats query

gtarpenning:
## Description

Depends on https://github.com/wandb/weave/pull/4250

Frontend to add a limit to the stats query. This limit has a special case when `limit=1` so this should be super speedy. Because of our basic call query construction, even limited queries without filters are extremely problematic, causing memory exceeded errors. 

## Testing

Prod:

We use a stream query with limit: 1
Screenshot 2025-04-30 at 9 57 02 AM

Branch:
We use the new stats query limit: 1
image


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/traceServerClientTypes.ts
```diff
@@ -420,3 +420,11 @@ export type TableCreateRes = {
   digest: string;
   row_digests: string[];
 };
+
+export type ProjectCheckReq = {
+  project_id: string;
+};
+
+export type ProjectCheckRes = {
+  has_data: boolean;
```
zbirenbaum, line 429: I'm guessing that the purpose of this endpoint is that checking if data exists is faster than actually pulling the project data and sending it over the wire?

zbirenbaum (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
```diff
@@ -46,39 +46,16 @@ export const useProjectHasTraceServerData = (
...
-    undefined,
-    undefined,
-    columns,
-    undefined,
-    {
-      skip: !hasTraceServer,
-    }
-  );
-  const loading = objs.loading || calls.loading;
+  const projectId = useMemo(() => `${entity}/${project}`, [entity, project]);
```
chance-wnb, line 49: There is a `projectIdFromParts` function for this.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
```diff
@@ -46,39 +46,16 @@ export const useProjectHasTraceServerData = (
...
+  const hasCalls = tsWFDataModelHooks.useProjectCheck(projectId, {
+    skip: !hasTraceServer,
+  });
+  const hasData = hasCalls.result?.has_data ?? false;
```
chance-wnb, line 53: nit: shorter: `!!hasCalls.result?.has_data`

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/context.tsx
gtarpenning: true, nice

gtarpenning: true, nice

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/traceServerDirectClient.ts
```diff
@@ -361,6 +363,13 @@ export class DirectTraceServerClient {
     }
   }
 
+  public projectCheck(req: ProjectCheckReq): Promise<ProjectCheckRes> {
+    return this.makeRequest<ProjectCheckReq, ProjectCheckRes>(
+      '/project/check',
```
chance-wnb, line 368: Is this based on a new endpoint? I can't seem to be able to find it in the code base, and neither from https://trace.wandb.ai/docs

gtarpenning: yes, based on https://github.com/wandb/weave/pull/4250 and https://github.com/wandb/core/pull/28989, but Emmanuel raised a couple points so we might be going in a slightly different route.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -475,6 +475,7 @@ const useCallsStats = (
   project: string,
   filter: CallFilter,
   query?: Query,
+  limit?: number,
```
jamie-rasmussen, line 478: I think this is kind of a counterintuitive parameter, I think it would be worth adding a comment explaining its purpose to trigger the optimized query, and maybe consider a wrapper method named something like `useCallsExist`.

weave-js/src/common/components/TraceCountDisplay.tsx
```diff
@@ -17,6 +17,7 @@ export const TraceCountDisplay = ({
     project,
     {},
     undefined,
+    undefined,
```
jamie-rasmussen, line 20: Even thought it is more verbose, I think it could be nicer to do something like `const limit = undefined;` or just add an in-line comment so future devs won't have to resort to hovering to see what these parameters mean.

jamie-rasmussen (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
gtarpenning: interesting, I agree it's counter intuitive. I can add that hook, but I also forsee adding limit to other stats queries. For example, for a string filter query, most use cases include either "show me some examples of this case" or "i know exactly what i'm looking for", both don't really benefit from the exact "497 matches"; I think we can get away with "100+" for example. This should unlock performance down the road. Happy to add comments around it though.

PR #4270, opened 2025-04-24
fix(weave): fixes an issue that impacts product

chance-wnb:
## Description

Fixes a production issue impact some of the users.

## Testing

TBD


gtarpenning (approved):
testing in prod, all results _should_ be the same, this is safe

PR #4273, opened 2025-04-24
fix(weave): Uncomment validation check and fix table_query_stats query

chance-wnb:
## Description

- Uncomments the error check in `table_query_stats` to ensure exactly one result is returned. The commenting was an expedited measure to quickly put prod errors out.

- Modifies the SQL query in `table_query_stats_batch` to:
  - Use `any(length(row_digests))` to pick one result instead of just `length(row_digests)`
  - Add a `GROUP BY digest` clause to properly aggregate results

- I have validated that the advanced form query generated by `make_table_stats_query_with_storage_size` does not return duplicated results.

## Testing

This change was tested by running the produced query against the impacted production data.

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -1047,8 +1047,8 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
 
         res = self.table_query_stats_batch(batch_req)
 
-        # if len(res.tables) != 1:
-        #     raise RuntimeError("Unexpected number of results", res)
+        if len(res.tables) != 1:
+            raise RuntimeError("Unexpected number of results", res)
 
         count = res.tables[0].count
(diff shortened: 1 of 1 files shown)


chance-wnb:
* **#4273** <a href="https://app.graphite.dev/github/pr/wandb/weave/4273?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4273?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

gtarpenning (approved):
I think we shouldn't raise, lets use log an exception `logger.error` or something. Even in the very worst case of getting unexpected table results, I think that would just be outdated data. In that case its probably better to show outdated data than to crash the page. Logging an error would let us triage and fix the issue but still show the user something in the meantime.... The query change looks good, up to you to change the exception if you want.

PR #4276, opened 2025-04-25
feat(weave): End-to-end online monitoring

neutralino1:
## Description

This PR adds support for online evaluations. Users can set up monitors to automatically score incoming calls.

There are N main components:
- The asynchronous scoring worker in `weave/workers/weave_scorer.py`. This is where incoming calls are scorer based on user configuration stored in monitors.
- The trace server enqueues "end call" events for the scoring worker to consume.
- The UI enables users to create and manage monitors
- Traces cal be filtered by which were scored by a particular monitor.

## Testing

How was this PR tested?

- Some unit tests, more needed
- Manual testing


weave-js/src/components/Form/TextArea.tsx
```diff
@@ -0,0 +1,35 @@
+/**
```
neutralino1, line 1: Ported from `core`.

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,445 @@
...
+from weave.trace.objectify import maybe_objectify
+from weave.trace.op import _default_on_input_handler
+from weave.trace.serialization.serialize import to_json
+from weave.trace.weave_client import (
+    RUNNABLE_FEEDBACK_TYPE_PREFIX,
+    Call,
+    FeedbackCreateReq,
+    ObjectRef,
+    from_json,
+)
```
neutralino1, line 32: All these require the SDK code to be included in the scorer Docker image, which means a dedicated image needs to be built, instead of reusing the server image, which would be ideal (lighter, simpler build process, simpler Helm charts).

Therefore, either we are ok with the server image containing SDK code, or we could refactor the SDK to isolate components that are required both in the SDK and the worker (i.e. the above imports).
For example, we could have `weave.lib` module that contains primitives (e.g. `Call`, `Monitor`, `Scorer`, etc.) and could be harmlessly importer in both the SDK and the workers.
Then a single image would be sufficient for all backend processes (server, worker(s), etc.). Much cleaner.

weave/trace_server/objects_query_builder.py
```diff
@@ -293,7 +293,7 @@ def make_metadata_query(self) -> str:
         ) as version_count,
         row_number() OVER (
             PARTITION BY project_id, kind, object_id
-            ORDER BY (deleted_at IS NULL) DESC, created_at DESC
```
neutralino1, line 296: This actually caused a bug, returning incorrect objects.

weave/workers/weave_scorer.py
neutralino1: I actually ended up copying the entire SDK code into the server image so that I can resuse for the worker. The build process and helm charts are much simpler now. See https://github.com/wandb/core/pull/29039

If separation of client and backend code is very important, isolate `weave.lib` as state above.

neutralino1:
@tssweeney would love your eyes on this PR when you get the chance. 🙏 

Many things can be improved as follow-ups. For example:
- better separation of client/backend code
- a more formal client for workers that mirrors SDK APIs to access/create/mutate objects
- abstract dequeuing infrastructure to be reused in other async jobs.

tests/trace/test_objects_query_builder.py
```diff
@@ -174,7 +174,7 @@ def test_object_query_builder_sort():
         ) as version_count,
         row_number() OVER (
             PARTITION BY project_id, kind, object_id
-            ORDER BY (deleted_at IS NULL) DESC, created_at DESC
+            ORDER BY created_at DESC
```
tssweeney, line 177: I would imagine this deleted clause was important. Can you share why it was removed here?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/tabularListViews/operators.ts
```diff
@@ -41,6 +41,13 @@ export const operationConverter = (
         substr: {$literal: item.value},
       },
     };
+  } else if (item.operator === '(string): by') {
```
tssweeney, line 44: i presume this operator is for `monitored "by" XYZ`. Seems weird to have the `(string)` prefix if so

weave-js/src/components/FancyPage/useProjectSidebar.ts
```diff
@@ -161,7 +161,11 @@ export const useProjectSidebar = (
           {
             type: 'button' as const,
             name: 'Evals',
-            slug: 'weave/evaluations',
+            slug: 'weave/evaluations/evals',
```
tssweeney, line 164: `/evaluations/evals` seems odd to me I would try to be specific like `evaluations/runs` or `evaluations/batch`.

weave-js/src/components/FancyPage/useProjectSidebar.ts
```diff
@@ -161,7 +161,11 @@ export const useProjectSidebar = (
...
+            additionalSlugs: [
+              'weave/evaluations/monitors',
```
tssweeney, line 166: In general i think it is good to have the URL scheme match the application architecture. However, in this case, I am not sure that `monitors` will like under `evaluations` for the foreseeable future. I might opt to have `monitors` be a top-level path allowing us to iterate a bit more on application architecture before committing to this nesting.

tests/workers/test_weave_scorer.py
```diff
@@ -0,0 +1,117 @@
...
+                created_at=datetime.now(),
+                version_index=1,
+                is_latest=1,
+                kind="object",
+                base_object_class="Monitor",
+                val={
+                    "name": "test-monitor-1",
+                    "description": "Test Monitor 1",
+                    "sampling_rate": 1.0,
+                    "scorers": ["scorer1", "scorer2"],
```
tssweeney, line 41: maybe this is just example data, but it seems sus to me that scorers can be non-refs

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -411,9 +414,12 @@ const Browse3ProjectRoot: FC<{
         <Route path={`${projectRoot}/calls/:itemName`}>
           <CallPageBinding />
         </Route>
-        <Route path={`${projectRoot}/:tab(evaluations|traces|calls)`}>
+        <Route path={`${projectRoot}/:tab(traces|calls)`}>
           <CallsPageBinding />
         </Route>
+        <Route path={`${projectRoot}/evaluations/:tab(evals|monitors)`}>
```
tssweeney, line 420: what happens when someone navigates to the old `/evaluations` route? should we default to evals?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -0,0 +1,488 @@
...
+      monitorName.length > 0 &&
+      selectedOpVersionOption.length > 0 &&
+      scorers.length > 0 &&
+      nameError === null
+    );
+  }, [monitorName, selectedOpVersionOption, scorers, nameError]);
+
+  const objCreate = useObjCreate();
+
+  const createScorer = useCallback(
```
tssweeney, line 153: maybe i'll find out below - but this does not seem to do anything other than create a scorer of a class & name - no properties - maybe this is just for testing?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -0,0 +1,488 @@
...
+      let scorerDigest: string;
+      if (refUri) {
+        // If the scorer already exists (this is an edit) we do not recreate it.
+        scorerDigest = parseRef(refUri).artifactVersion;
+      } else {
+        // If refUri is undefined the scorer needs to be create.
+        const scorerClassName = scorerName;
+        scorerName = `${scorerName}-${uuidv4().split('-')[0]}`;
+
+        const scorerObj = {
```
tssweeney, line 170: this structure is implicitly bound to the payload that the python client creates. It would be nice for that to be extracted into an specific module so that it is not buried inside component implementation. For example: `ref: None` is a BUG in python! Fixing this should be centralized.

tssweeney (commented):
Review Phase 1: High-level UI code analysis
* Looks pretty good. Left a few comments where appropriate. I think the only "discussion" point might be the URL scheme. 

Next:
* [ ] High level python code review
* [ ] Usability review

weave/flow/scorer.py
```diff
@@ -35,6 +37,17 @@ def summarize(self, score_rows: list) -> Optional[dict]:
         return auto_summarize(score_rows)
 
 
+class BuiltInScorer(Scorer):
```
tssweeney, line 40: This seems odd. Searching the current code base, I see this pattern exists elsewhere, but it certainly seems odd to me that we need to explicitly do this. Adding an intermediate class definition with a general-purpose deserializer?? Why is this not just a class method for on the `Object` class? Maybe question for @andrewtruong?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
neutralino1: At this time, this does not support parametrizing scorers from the UI. This is supported via the SDK though.

In order to support parametrizing scorers in the UI, we will need to do custom work for each scorer (custom form fields). So it would be good to define prioritization. I selected `ValidJSONScorer` as first to support since it doesn't require any parameters. I will add `ValidXMLScorer` too.
Then we can start `LLMAsAJudgeScorer` for parametrizable scorers.

weave/trace/weave_client.py
```diff
@@ -1,7 +1,6 @@
 from __future__ import annotations
```
neutralino1, line 1: These changes should be undone.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -213,8 +217,11 @@ def as_sql(
         cast: Optional[tsi_query.CastTo] = None,
     ) -> str:
         inner = super().as_sql(pb, "feedback")
-        param_name = pb.add_param(self.feedback_type)
-        res = f"anyIf({inner}, feedback.feedback_type = {param_slot(param_name, 'String')})"
+        if self.feedback_type == "*":
```
tssweeney, line 220: would be nice to not in the interface definition that `feedback_type` can be a "*".

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -319,6 +320,9 @@ def call_end(self, req: tsi.CallEndReq) -> tsi.CallEndRes:
         # the call does not already exist
         self._insert_call(ch_call)
 
+        kafka_producer = KafkaProducer.from_env()
```
tssweeney, line 323: idk what the overhead is of creating a new producer & connection is for each call, but this might be better suited as a property of the server class itself.

weave/trace_server/kafka.py
```diff
@@ -0,0 +1,43 @@
...
+            "client.id": socket.gethostname(),
+        }
+        return cls(conf)
+
+    def produce_call_end(self, call_end: tsi.EndedCallSchemaForInsert) -> None:
+        self.produce(
+            topic=CALL_ENDED_TOPIC,
+            value=call_end.model_dump_json(),
+        )
+        self.flush()
```
tssweeney, line 26: any perf considerations with flushing immediately here?

weave/trace_server/calls_query_builder/calls_query_builder.py
neutralino1: Good point. Where is the interface defined though? I can add a comment on line 179 where the field is defined.

tests/trace/test_objects_query_builder.py
neutralino1: For some reason this didn't work when I was developing. I recovered the clause now.

weave/trace_server/clickhouse_trace_server_batched.py
neutralino1: Fixed.

weave/trace_server/calls_query_builder/calls_query_builder.py
tssweeney: ahh - i see it is not really made explicit anywhere. maybe add it to 380 of `weave/trace_server/trace_server_interface.py`

weave/workers/weave_scorer.py
tssweeney: agreed - I think this is the way to go for now. We will probably want to split things up in the future

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,445 @@
...
+    if _TRACE_SERVER is None:
+        _TRACE_SERVER = ClickHouseTraceServer.from_env()
+
+    return _TRACE_SERVER
+
+
+# This should be cached to avoid hitting ClickHouse for each ended call.
+def get_active_monitors(
+    project_id: str,
+) -> list[tuple[Monitor, InternalObjectRef, Optional[str]]]:
```
tssweeney, line 62: I would prefer a TypedDict or helper class over tuples

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,445 @@
...
+            )
+        else:
+            logger.exception("Kafka error: %s", msg.error())
+
+        return False
+
+    return True
+
+
+async def cleanup_tasks(call_processors: set[asyncio.Task]) -> set[asyncio.Task]:
```
tssweeney, line 392: this function name seems misleading. reading the code, it is more like: `waitAndPopFirstComleted`

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,445 @@
...
+async def process_kafka_message(
+    msg: Message, consumer: KafkaConsumer, call_processors: set[asyncio.Task]
+) -> bool:
+    """Process a single Kafka message and create a task for it."""
+    try:
+        ended_call = tsi.EndedCallSchemaForInsert.model_validate_json(
+            msg.value().decode("utf-8")
+        )
+        logger.info("Processing ended call %s", ended_call.id)
+
```
tssweeney, line 363: there is a critical item here: we cannot be sure that the start event is indeed inserted before the end event. There are a few ways to handle this. One simple method could be to re-queue the item with some backoff bookkeeping

tssweeney: the way I had proposed during scoping is actually to have another queue (completed calls) that contains calls that are fully matched up. This would allow for other downstream applications to operate on already-matched calls.

neutralino1: Do we need to care that the start event was inserted?
All we need is the end event, which contains the Op output. No?

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,445 @@
...
+        return
+
+    for scorer in monitor.scorers:
+        logger.info("Applying scorer %s to call %s", scorer.name, call.id)
+        await apply_scorer(
+            monitor_internal_ref, scorer, call, ended_call.project_id, wb_user_id
+        )
+
+
+def _do_score_call(scorer: Scorer, call: Call, project_id: str) -> tuple[str, Any]:
```
tssweeney, line 237: This method (and a few others) are very similar to what is already in the client. Was it possible to reuse that code / refactor it to do so? If not, a very clear comment and explanation would be useful as they will need to be kept in sync.

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,445 @@
...
+    global _TRACE_SERVER
+
+    if _TRACE_SERVER is None:
+        _TRACE_SERVER = ClickHouseTraceServer.from_env()
+
+    return _TRACE_SERVER
+
+
+# This should be cached to avoid hitting ClickHouse for each ended call.
+def get_active_monitors(
```
tssweeney, line 68: I think you pretty much have to cache this now, else we are going to needlessly slam clickhouse. I'd recommend LRU with TTL of 1 min

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,458 @@
...
+    # - if that returns > 0, we know the call exists, so we can proceed with the filter
+
+    count_req = tsi.CallsQueryStatsReq(
+        project_id=ended_call.project_id,
+        filter=tsi.CallsFilter(call_ids=[ended_call.id]),
+    )
+
+    count = server.calls_query_stats(count_req).count
+
+    if count == 0:
```
tssweeney, line 163: assuming that this will return 0 when the start event is not yet written, then this will effectively serve as a "wait" for the start event

weave/workers/weave_scorer.py
neutralino1: Do we have an LRU with TTL implementation somewhere already?
`functools.lru_cache` does not have TTL.

weave/workers/weave_scorer.py
tssweeney: this wait however locks up the worker thread until complete and possibly rejects before the start comes through

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,458 @@
...
+            call_ids=[ended_call.id],
+            query=query,
+            op_names=op_names,
+        ),
+    )
+
+    calls = server.calls_query(req).calls
+
+    if len(calls) == 0:
+        logger.warning("No matching calls found for call id %s", ended_call.id)
```
tssweeney, line 178: this shouldn't be a warning - my reading is that this will happen nearly all the time because most calls will not match a monitor

tssweeney (commented):
Review Part 2: Python Review. I think the main things are:
1. A decent amount of copied code from the client risks falling out of sync
2. I am not sure that we have a robust way to ensure that we wait for a complete (start + end) event
3. It seems like we are are going to be bashing clickhouse with current implementation. Would be great to cache and/or group (maybe pull off a chunk of 100 calls and process project groups). This could dramatically reduce the number of calls against the DB.

Next: UX Review

tssweeney (commented):
First UI test. Totally new project, create simple monitor.

Screenshot 2025-05-13 at 15 01 52
Screenshot 2025-05-13 at 15 00 36

```
react.tsx:563 Uncaught TypeError: Failed to construct 'URL': Invalid URL
    at parseRef (react.tsx:563:15)
    at Object.renderCell (MonitorsPage.tsx:102:39)
    at GridCell2 (GridCell.js:286:23)
    at renderWithHooks (react-dom.development.js:15486:18)
    at updateForwardRef (react-dom.development.js:19245:20)
    at beginWork (react-dom.development.js:21675:16)
    at HTMLUnknownElement.callCallback2 (react-dom.development.js:4164:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:16)
    at invokeGuardedCallback (react-dom.development.js:4277:31)
    at beginWork$1 (react-dom.development.js:27490:7)
```

tssweeney:
> First UI test. Totally new project, create simple monitor.
> 
> Screenshot 2025-05-13 at 15 01 52 Screenshot 2025-05-13 at 15 00 36
> 
> ```
> react.tsx:563 Uncaught TypeError: Failed to construct 'URL': Invalid URL
>     at parseRef (react.tsx:563:15)
>     at Object.renderCell (MonitorsPage.tsx:102:39)
>     at GridCell2 (GridCell.js:286:23)
>     at renderWithHooks (react-dom.development.js:15486:18)
>     at updateForwardRef (react-dom.development.js:19245:20)
>     at beginWork (react-dom.development.js:21675:16)
>     at HTMLUnknownElement.callCallback2 (react-dom.development.js:4164:14)
>     at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:16)
>     at invokeGuardedCallback (react-dom.development.js:4277:31)
>     at beginWork$1 (react-dom.development.js:27490:7)
> ```

Probably because I just chose "all ops" from the op filter: 
```
export const ALL_TRACES_OR_CALLS_REF_KEY = '__NO_REF__';
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -0,0 +1,492 @@
...
+      refUri?: string;
+    }): Promise<WeaveObjectVersionKey> => {
+      let scorerDigest: string;
+      if (refUri) {
+        // If the scorer already exists (this is an edit) we do not recreate it.
+        scorerDigest = parseRef(refUri).artifactVersion;
+      } else {
+        // If refUri is undefined the scorer needs to be create.
+        const scorerClassName = scorerName;
+        scorerName = `${scorerName}-${uuidv4().split('-')[0]}`;
```
tssweeney, line 168: this seems odd to just slap on an arbitrary id to the name here. What is the purpose?

Screenshot 2025-05-13 at 16 03 40

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorPage.tsx
```diff
@@ -0,0 +1,275 @@
...
+                    )}
+                  </span>
+                  <Button
+                    variant={ButtonVariants.Secondary}
+                    icon={IconNames.Table}
+                    onClick={onGoToTableClick}>
+                    Go to table
+                  </Button>
+                </Box>
+                <CallsTable
```
tssweeney, line 261: Screenshot 2025-05-13 at 16 07 21

* Would be nice if the attributes were hidden by default like they are in the main table
* Would be nice to take up 100% of allowed height

weave/trace_server/kafka.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class KafkaConsumer(ConfluentKafkaConsumer):
+    @classmethod
+    def from_env(cls) -> "KafkaConsumer":
+        conf = {
+            "bootstrap.servers": _make_broker_host(),
+            "client.id": socket.gethostname(),
+            "group.id": "weave-worker-scorer",
```
tssweeney, line 35: this `group.id` is super critical (in that once we deploy cannot change). Suggest hoisting to const similar to the topic

tests/workers/test_weave_scorer.py
```diff
@@ -0,0 +1,117 @@
+from datetime import datetime
```
tssweeney, line 1: I think tests for this change are sorely lacking. It would be great to have a test suite that actually tests the end to end functionality: create a monitor, log calls, query and get their scores. I think this primarily comes down to getting a queue into CI

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
neutralino1: I'll move this out to `tsDataModelHooks.ts` in a follow-up PR.

Re: `ref: None`, I tried publishing an `Evaluation` and a `Scorer` from the Python SDK and they also had `ref: None` in the `val` field. 🤔

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorPage.tsx
neutralino1: I'll address this in a follow-up PR.

weave/trace_server/kafka.py
neutralino1: I will find out during load testing.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorsPage.tsx
```diff
@@ -0,0 +1,209 @@
...
+const CallCountCell = ({
+  monitorRef,
+  entity,
+  project,
+}: {
+  monitorRef: string;
+  entity: string;
+  project: string;
+}) => {
+  const {useCallsStats} = useWFHooks();
```
chance-wnb, line 179: For new component, we should be able to directly export it instead of using a context.

chance-wnb (commented):
Thanks for the change. I helped reviewed the Frontend piece only.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorsPage.tsx
chance-wnb: I guess this is not yet refactored?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorPage.tsx
```diff
@@ -0,0 +1,277 @@
...
+];
+const typographyStyle = {fontFamily: 'Source Sans Pro'};
+
+export const MonitorPage = (props: {
+  entity: string;
+  project: string;
+  objectName: string;
+  version: string;
+}) => {
+  const {useRootObjectVersions} = useWFHooks();
```
chance-wnb, line 44: Can do directly import instead of grabbing from the context.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorPage.tsx
```diff
@@ -0,0 +1,277 @@
...
+    const url = baseRouter.callsUIUrl(
+      entity,
+      project,
+      undefined,
+      callsFilterModel
+    );
+    history.push(url);
+  }, [baseRouter, history, callsFilterModel, entity, project]);
+
+  const {useCallsStats} = useWFHooks();
```
chance-wnb, line 116: Can do directly import instead of grabbing from the context.

chance-wnb (commented):
It might be worth double checking if the way of directly importing hooks  is adopted.

chance-wnb (approved):
Great work! Thank you!

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,489 @@
...
+        scorer.__dict__["internal_ref"] = scorer_ref
+
+    return scorers  # type: ignore
+
+
+class CallNotWrittenError(Exception):
+    pass
+
+
+# We do the following retry logic because ClickHouse does not guarantee that the call will be available immediately.
```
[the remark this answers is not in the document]
neutralino1: Not just that I think. I think CH does not guarantee write->read consistency.

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,489 @@
...
+
+    monitor: Monitor
+    internal_ref: InternalObjectRef
+    wb_user_id: Optional[str]
+
+
+class MonitorsCache:
+    """Cache for active monitors."""
+
+    _cache: OrderedDict[str, tuple[datetime, list[ActiveMonitor]]] = OrderedDict()
```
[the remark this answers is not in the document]
neutralino1: The cache has a capacity of 100 monitor. When `set` is called, we pop older monitors so no risk of memory leak here.

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,489 @@
...
+        payload=payload,
+        runnable_ref=scorer.__dict__["internal_ref"].uri(),
+        call_ref=score_call_ref.uri(),
+        wb_user_id=wb_user_id,
+        trigger_ref=monitor_internal_ref.uri(),
+    )
+
+    logger.info("Creating feedback for scorer %s and call %s", scorer.name, call.id)
+    server = get_trace_server()
+    server.feedback_create(feedback_req)
```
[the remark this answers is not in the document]
neutralino1: Yes that is true. For every call we scorer, we write 2 rows to the `call_parts` table and 1 row to the `feedbacks` table. The writing to `call_parts` is batched but not to `feedbacks`. That solves half the problem. At this time I believe the main performance hangup are these non-batched writes. The reason I did not do it is that batching call writes is already supported by our CH code, but not batching feebdack writes. SO I need to do some refactoring there as a follow-up.

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/FilterBar.tsx
```diff
@@ -170,6 +174,12 @@ export const FilterBar = ({
     value: 'id',
     label: 'Call ID',
   });
+  if (isWandbAdmin) {
+    (options[0] as GroupedOption).options.push({
+      value: 'feedback.[*].trigger_ref',
```
tssweeney: So this filters to all calls that are monitored? In the event that we add other trigger types, this will over-select. Might want to just add a note

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
```diff
@@ -52,6 +52,7 @@ export const FIELD_LABELS: Record<string, string> = {
   started_at: 'Called',
   wb_run_id: 'Run',
   wb_user_id: 'User',
+  'feedback.[*].trigger_ref': 'Monitored',
```
tssweeney, line 55: given how popular this `feedback.[*].trigger_ref` string is (not only as a key, but also a query path) it would be good to move it out of a magic string and into a const

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/OpSelector.tsx
```diff
@@ -0,0 +1,154 @@
+import {
```
tssweeney, line 1: <3 love these re-org refactors

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -0,0 +1,496 @@
...
+      monitorName.length > 0 &&
+      selectedOpVersionOption.length > 0 &&
+      scorers.length > 0 &&
+      nameError === null
+    );
+  }, [monitorName, selectedOpVersionOption, scorers, nameError]);
+
+  const objCreate = useObjCreate();
+
+  const createScorer = useCallback(
```
tssweeney: Given that you have chosen to punt on the comments below relating to this createScorer, can you please leave a comment in code describing the outstanding dev tasks to be followed up.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorPage.tsx
tssweeney: In cases like this where you elect to postpone addressing a reviewer's comment, can you please at least add a code comment indicating the followup TODO and intended action. I think this makes for a more friendly reviewer/reviewee pattern and ensure we don't forget about the task

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorsPage.tsx
```diff
@@ -0,0 +1,208 @@
...
+}: {
+  monitorRef: string;
+  entity: string;
+  project: string;
+}) => {
+  const query: Query = useMemo(() => {
+    return {
+      $expr: {
+        $contains: {
+          input: {$getField: 'feedback.[*].trigger_ref'},
```
tssweeney: reminder - another place for that magic string - probably a search and replace would be good.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ObjectsPage/ObjectVersionsTable.tsx
```diff
@@ -83,6 +82,12 @@ export const ObjectVersionsTable: React.FC<{
         // solution here in the future. Maybe exclude table refs?
         val = _.omit(val, 'rows');
       }
+      // This is a dirty hack to enable keeping the nested query value for Monitors
```
tssweeney, line 85: I had to really think hard about what this hack is doing. Do you mind adding a comment here describing what is going on? In particular, i think the summary is: `keepNestedVal` allows you to specify top-level fields from the underlying vals which are not flattened into nested columns.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -182,7 +182,7 @@ const useMakeTraceServerEndpoint = <
   return traceServerRequest;
 };
 
-const useCall = (params: UseCallParams): Loadable<CallSchema | null> => {
+export const useCall = (params: UseCallParams): Loadable<CallSchema | null> => {
```
tssweeney, line 186: Nice - I actually like that you started exporting these. Adding this for context sharing: normally, everywhere you would do something like `const {useCall} = useWFHooks();`, then call `useCall`. This is because when we first started Weave, we had an adapter layer for our two backends (yes, we had a FastAPI backend AND a QueryService backend). The query service one can/should be deprecated which means the adapter can go away, which means we can just use these methods directly.

weave-js/src/components/PagePanelComponents/Home/Browse3/smallRef/SmallRef.tsx
```diff
@@ -6,17 +6,32 @@ import {
...
 export const SmallRef: FC<{
   objRef: ObjectRef;
   iconOnly?: boolean;
   noLink?: boolean;
 }> = ({objRef, iconOnly = false, noLink = false}) => {
   if (isWeaveObjectRef(objRef)) {
-    return (
-      <SmallWeaveRef objRef={objRef} iconOnly={iconOnly} noLink={noLink} />
-    );
+    if (objRef.artifactVersion === '*') {
```
tssweeney: nit: I would move this switch logic to `SmallWeaveRef`. Given that these conditionals are specific only to the Weave-style refs. Other similar switching happens in `SmallWeaveRef`

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -244,6 +246,13 @@ def file_storage_client(self) -> Optional[FileStorageClient]:
         self._file_storage_client = maybe_get_storage_client_from_env()
         return self._file_storage_client
 
+    @property
+    def kafka_producer(self) -> KafkaProducer:
+        if self._kafka_producer is not None:
+            return self._kafka_producer
+        self._kafka_producer = KafkaProducer.from_env()
```
tssweeney, line 253: Is this going to have problems for environments that don't have kafka setup? should this be a conditional return?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -311,7 +320,7 @@ def call_start(self, req: tsi.CallStartReq) -> tsi.CallStartRes:
             trace_id=ch_call.trace_id,
         )
 
-    def call_end(self, req: tsi.CallEndReq) -> tsi.CallEndRes:
+    def call_end(self, req: tsi.CallEndReq, publish: bool = True) -> tsi.CallEndRes:
```
tssweeney, line 323: What does this `publish` control? It is very rare (if at all) that we deviate from the req/res pattern. I am curious why we need this and where it is being called from

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,489 @@
...
+
+    finally:
+        if tasks:
+            await asyncio.gather(*tasks)
+        consumer.close()
+
+
+def init_sentry() -> None:
+    sentry_sdk.init(
+        dsn="https://8ed9a67d68481736b1bc4e815a2a8901@o151352.ingest.us.sentry.io/4509210797277185",
```
tssweeney, line 491: Probably want to make this an env var so we don't get slammed by an attacker

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,489 @@
...
+            summary={},
+        )
+    )
+
+    server.call_end(call_end_req, publish=False)
+
+    return call_start_res.id, result
+
+
+async def apply_scorer(
```
tssweeney, line 301: i'd slap datadog tracing on all these primary methods

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,489 @@
...
+            call_ids,
+            monitor.op_names,
+            monitor.query,
+        )
+
+        with get_trace_server().call_batch():
+            for call in calls:
+                for scorer in monitor.scorers:
+                    logger.info("Applying scorer %s to call %s", scorer.name, call.id)
+                    await apply_scorer(
```
tssweeney: highly recommend asyncio gather over these coroutines, otherwise you are going to serially process each call

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,489 @@
...
+        monitor = active_monitor["monitor"]
+        monitor_internal_ref = active_monitor["internal_ref"]
+        wb_user_id = active_monitor["wb_user_id"]
+
+        # Here we potentially query the same calls multiple times.
+        # The reasons is that each monitor has its own call filter, so we need to query the DB for each monitor.
+        # We could merge the call filters and do a single query, but then
+        # we would need to figure out which call match what monitor filter.
+        # Currently we have no way to apply filters outside the DB
+        calls = await get_filtered_calls(
```
tssweeney, line 411: if you wanted to break this up further, you could parallelize processing each monitor and then, each call for that monitor using asynio gather

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,489 @@
...
+        with get_trace_server().call_batch():
+            for call in calls:
+                for scorer in monitor.scorers:
+                    logger.info("Applying scorer %s to call %s", scorer.name, call.id)
+                    await apply_scorer(
+                        monitor_internal_ref, scorer, call, project_id, wb_user_id
+                    )
+
+
+async def run_consumer() -> None:
```
tssweeney, line 434: I think there might be a logical bug here if the finish event is processed in the queue before the start event is written to db. I think we discussed this in person or in writing somewhere but I can't quite find it. Anyhow... maybe something to note? While an edge case, i don't think it is too rare. We could make a query that finds how often this happens in prod. Probably not blocking for first land, but something to be aware of.

weave/workers/weave_scorer.py
neutralino1: There is already a public Sentry URL here: https://github.com/wandb/weave/blob/b033ca6652177f186aa7bee6b2b3e2a8c72bf3fb/weave/trace/trace_sentry.py#L30

weave/trace_server/clickhouse_trace_server_batched.py
neutralino1: This will only be called if `WF_ENABLE_ONLINE_EVAL` is `true`.

weave/trace_server/clickhouse_trace_server_batched.py
neutralino1: This is called with `publish=False` from the scoring worker. The goal is to reduce noise.
Scoring calls creates new calls (the `Scorer.score` call), which would get published to the Bufstream topic again. Of course those calls are dismissed by the worker as they do no match any monitor, but it still doubles the volume of messages received for every call to actually score.

Would you rather I add `publish` as a member of `CallEndReq`?

weave/workers/weave_scorer.py
neutralino1: It actually does not matter at this time since `apply_scorer` is actually synchronous. It should actually not have an `async` on it. Right now it assumes all `Scorer.score` calls are sync. Of course when we introduce `LLMAsAJudgeScorer` the `score` call will be async. At that point there will be a benefit in making `apply_scorer` async.
I made the suggested change anyway in anticipation for a truly async `apply_scorer`.

weave/workers/weave_scorer.py
neutralino1: I added a `ended_at IS NOT NULL` clause to the count query to make sure we only count calls for which the end call was written.

weave/workers/weave_scorer.py
neutralino1: Done.

weave/workers/weave_scorer.py
tssweeney: that one is a client sdk project and this one is internal, but i hear you.

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: no, i think this is fine, but it would be nice to explain this param in code so that we understand why it deviates from the standard external interface pattern

weave/workers/weave_scorer.py
```diff
@@ -0,0 +1,504 @@
...
+    query: Optional[tsi.Query],
+) -> list[Call]:
+    """This function is used to get the calls that match the filter."""
+    logger.info("Attempting to get %s calls", len(call_ids))
+    server = get_trace_server()
+    count_req = tsi.CallsQueryStatsReq(
+        project_id=project_id,
+        filter=tsi.CallsFilter(call_ids=call_ids),
+        # We want to make sure all end_call events have been written to the DB
+        query={
```
tssweeney, line 205: since `call_end` events are added to kafka queue after they are adde to CH, i think this filter is going to have very little effect. The comment I made before was that the corresponding `call_start` event may be logged from the client sometime after `call_end`. In those cases, you are going to hit the `raise CallNotWrittenError("Some calls not yet written to DB")` error below - which (while an edge case) can still happen. I think the `CallNotWrittenError` error is pretty harsh. It is going to crash the entire process & drop all the calls currently being processed in the current batch

tssweeney (commented):
Nice Work! I think this is a phenomenal enhancement to the product! Great work pushing through! (: 🔥 

I do believe the comment here illustrates a pretty gnarly bug, but approving none the less to unblock and move forward: https://github.com/wandb/weave/pull/4276/files#r2127896961

weave/workers/weave_scorer.py
tssweeney: Private convo pointed out that this will retry for a bit, then drop that batch, but not terminate the process

PR #4277, opened 2025-04-25
fix(weave): duplicate dataset references cause wild queries

gtarpenning:
## Description

[WB-24616](https://wandb.atlassian.net/browse/WB-24616)

As we iterate through refs and construct query builder conditions, if we have already seen the ref before just continue. We already have handling for if the ref doesn't yield a value (`root_val_cache.get(make_root_ref_cache_key(ref), None)`) so I think this is safe and will deduplicate conditions.

These are related to insane 10+ minute queries:
Screenshot 2025-04-24 at 6 42 32 PM

## Testing

This is hard to test, we need to create a trace that has a ref that has many refs to the same dataset. 

[WB-24616]: https://wandb.atlassian.net/browse/WB-24616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1185,6 +1185,7 @@ def get_object_refs_root_val(
                 parameters[object_id_param_key] = ref.name
                 parameters[version_param_key] = ref.version
                 ref_digests.add(ref.version)
+                root_val_cache[cache_key] = None
```
gtarpenning, line 1188: this is the real change

weave/trace_server/sqlite_trace_server.py
```diff
@@ -641,6 +641,8 @@ def _expand_refs(
 
             derefed_val = self.refs_read_batch(tsi.RefsReadBatchReq(refs=[val])).vals[0]
             set_nested_key(data, col, derefed_val)
+            ref_col = f"{col}._ref"
+            set_nested_key(data, ref_col, val)
```
gtarpenning, line 645: bringing sqlite to parity with batched

neutralino1 (approved):
Thank you.

PR #4278, opened 2025-04-25
chore(weave): Add new testing shards

andrewtruong:
Moves tests around into new shards to speed up tests


tests/trace/test_client_trace.py
```diff
@@ -122,20 +122,6 @@ def my_op(a: int) -> int:
     )
 
 
-def test_dataset(client):
```
gtarpenning, line 125: we don't need this?

gtarpenning (approved):

PR #4289, opened 2025-04-25
chore(weave): Update noxfile

andrewtruong:
Cleans up the noxfile and starts testing shards that are now py313 compatible


zbirenbaum (approved):
LGTM

PR #4290, opened 2025-04-25
feat(ui): add bulk object delete functionality

jamie-rasmussen:
## Description

Adds checkboxes and a delete button to the Object Versions page, allowing you to quickly delete an entire object (not just one version of it) or multiple selected objects.

image
image

This UI follows the current backend permissions checks, which allow anyone with write access to the project to delete, even if they are not an admin / not the original creator of the version.

@Yangyi-wandb @m-rgba - design consult, do we want more of a warning about this irreversible action?

Fixes some additional bugs:
* Delete button only getting shown for site admins on version list page
* Row selection state not getting cleared when object versions are deleted through peek drawer.

## Testing

Manual through `wf` versions example.


gtarpenning (approved):
LGTM, tests and works great. nice.

PR #4291, opened 2025-04-25
feat(ui): add buttons to control line wrapping

jamie-rasmussen:
## Description

Internal Slack: https://weightsandbiases.slack.com/archives/C01T8BLDHKP/p1745584425628329

Adds two buttons to toggle between horizontal scroll or wrapping of long lines. Design reviewed by @Yangyi-wandb 

Screenshot 2025-04-25 at 4 28 39 PM
Screenshot 2025-04-25 at 4 28 45 PM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse2/CellValueString.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse2/CellValueString.tsx
@@ -206,17 +206,18 @@ const CellValueStringWithPopup = ({value}: CellValueStringProps) => {
                       variant="ghost"
                       active={!wrapLines}
                       icon="pan-tool-1"
-                      tooltip="Don't wrap long lines"
+                      tooltip="Don't wrap code"
                       onClick={() => {
                         setWrapLines(false);
                       }}
                     />
                     <Button
+                      className="mr-4"
                       size="small"
                       variant="ghost"
                       active={wrapLines}
                       icon="wrap"
-                      tooltip="Wrap long lines"
+                      tooltip="Wrap code"
                       onClick={() => {
                         setWrapLines(true);
                       }}


neutralino1 (commented):
That's super helpful.
Should we set the two toggle buttons inside a button group to express the either-or behavior?
Or maybe use a toggle component?

jamie-rasmussen:
> Should we set the two toggle buttons inside a button group to express the either-or behavior? Or maybe use a toggle component?

I think a toggle isn't ideal - without iconography you'd need to have a label or something to indicate what it was for, which would take up more space and be more visually inconsistent with the rest of the toolbar.

We do have a ToggleButtonGroup component but it is pretty new, not used in many places, and appears to be buggy - rather than take the time to debug it I implemented it this way, but I could look more closely if you prefer.

neutralino1:
cc @m-rgba can we get your input here?

m-rgba:
So we've got two button groups, and two independent panel controls.
- I'd look at adding some spacing between the button group on the right and the panel controls.
- The positioning of the auxiliary (code) button group is ok, since having two next to each other is awkward.
- I try to minimize pop-in as much as possible - it adds a bit more to the controls, but if we just disabled those and enabled them when code was selected it can help minimize drawing the eye to it.
- We can simplify the copy a bit to `Wrap / Un-wrap code` to help teach the user which mode they need to be in.

I prototyped those changes below -- and can add a commit / PR if everyone agrees:

popover

neutralino1:
@m-rgba maybe we should hide the wrap/unwrap buttons altogether if we are not on the code view?

m-rgba:
@neutralino1 I don't have a strong feeling about it, since they aren't in use at all in those other modes.
But generally, popping in controls can feel a little awkward.

I think the `mr-8` on the right button would solve the issue of separating them from the panel controls.

neutralino1:
Sounds good 👍

jamie-rasmussen:
> @neutralino1 I don't have a strong feeling about it, since they aren't in use at all in those other modes. But generally, popping in controls can feel a little awkward.

Yes, but I feel it's less awkward overall than having disabled controls that don't apply to your current mode.

> I think the `mr-4` on the right button (wrap) would solve the issue of separating them from the panel controls.

Added the margin. I also shortened the tooltip text, though I didn't like the term "Un-wrap code" which seemed confusing because this is more about the absence of an action rather than taking one - I don't the "un-wrap" terminology is common for this use case.

weave-js/src/components/CodeEditor.tsx
```diff
@@ -84,6 +86,7 @@ export const CodeEditor = ({
       handleMouseWheel: handleMouseWheel ?? false,
       alwaysConsumeMouseWheel: alwaysConsumeMouseWheel ?? true,
     },
+    wordWrap: wrapLines ? ('on' as const) : ('off' as const),
```
gtarpenning, line 89: why the `as const` here?

jamie-rasmussen: Short answer is that `tsc` wasn't able to determine that `on | off` is a subset of `on | off | wordWrapColumn | bounded`

neutralino1 (approved):
Thank you.

PR #4298, opened 2025-04-25
fix(ui): don't show overview in nav sidebar twice

jamie-rasmussen:
## Description

In the case of an empty project or one that has both Models and Weave data we are showing the Overview page twice, once at the top of the nav bar and once in the Models overflow menu.

Before:
Screenshot 2025-04-25 at 4 02 23 PM

After:
Screenshot 2025-04-25 at 4 02 36 PM


jwlee64 (approved):

neutralino1 (approved):
Thank you.

Issue #3494, opened 2025-01-26
Same function Defininition after @weave.op decorator


kiansierra:
![Image](https://github.com/user-attachments/assets/b3b11971-445c-4e03-8b07-52e71ad0c1a6)
It would be a nice quality of life have the decorated function return the same signature as the original one when using the intelisense in the IDE

jamie-rasmussen:
We agree! There is an open ticket for this https://github.com/wandb/weave/issues/3423 - the author of that posted a workaround you might be interested in.

Issue #3423, opened 2025-01-16
`@weave.op` decorator destroys type information


pokey:
If I have the following code:

```py
@weave.op()
def get_first_draft(
    previous_episodes: list[EpisodeCore],
    repository_data: dict,
) -> str:
```

VSCode type checking stops checking for valid args when I call the function, and autocomplete breaks for the args. The function is just interpreted as an `Op` object. See eg the hover

<img width="312" alt="Image" src="https://github.com/user-attachments/assets/8276fd2d-9cc0-409a-b114-bec7dcfa44d7" />

This is a showstopper, as I rely heavily on autocomplete and type checking. Notice how I forgot to add an arg

pokey:
For now, this is my workaround:

```py
from collections.abc import Callable
from typing import TypeVar, cast

import weave

C = TypeVar("C", bound=Callable)


def weave_op():
    def ret(fn: C) -> C:
        return cast(C, weave.op()(fn))

    return ret
```

I don't support any of the `@weave.op()` arguments, and I don't expose any typing information about what I can use on an `Op`, but it's good enough for me for now. I think the correct thing would be to make the `Op` class have a generic type argument

jamie-rasmussen:
Nice workaround! Glad you are unblocked. Will open an internal ticket to track this.

michael-linnane-lrn:
I see [chore(weave): Don't destroy type information when using op decorator #4305](https://github.com/wandb/weave/pull/4305) has been merged but this error is still on the latest release. Any updates here?

PR #4305, opened 2025-04-25
chore(weave): Don't destroy type information when using op decorator

andrewtruong:
This adds better type hinting support for editors.

Resolves:
- https://github.com/wandb/weave/issues/3423
- https://github.com/wandb/weave/issues/3494

Before:
image

After:
image
image

### DIFF
--- a/weave/flow/eval.py
+++ b/weave/flow/eval.py
@@ -137,7 +137,7 @@ class Evaluation(Object):
         if self.name is None and self.dataset.name is not None:
             self.name = self.dataset.name + "-evaluation"  # type: ignore
 
-    @weave.op()
+    @weave.op
     async def predict_and_score(self, model: Union[Op, Model], example: dict) -> dict:
         apply_model_result = await apply_model_async(
             model, example, self.preprocess_model_input
--- a/weave/scorers/bedrock_guardrails.py
+++ b/weave/scorers/bedrock_guardrails.py
@@ -60,7 +60,7 @@ class BedrockGuardrailScorer(weave.Scorer):
         return {"source": self.source, "content": [{"text": {"text": output}}]}
 
     @weave.op
-    def score(self, output: str) -> WeaveScorerResult:
+    def score(self, *, output: str, **kwargs: Any) -> WeaveScorerResult:
         if self._bedrock_runtime is None:
             raise ValueError("Bedrock runtime client is not initialized")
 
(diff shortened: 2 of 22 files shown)


weave/flow/eval.py
```diff
@@ -137,7 +137,7 @@ def model_post_init(self, __context: Any) -> None:
         if self.name is None and self.dataset.name is not None:
             self.name = self.dataset.name + "-evaluation"  # type: ignore
 
-    @weave.op()
```
zbirenbaum, line 140: Is there a functional reason this was this changed from `@weave.op()` to `weave.op`? Some of the others were left as calls so it's not immediately apparent to me

andrewtruong: just stylistic -- it's cleaner without paren imo

zbirenbaum (approved):
Really awesome work here!

weave/flow/eval.py
zbirenbaum: Honestly agree

weave/scorers/bedrock_guardrails.py
```diff
@@ -60,7 +60,7 @@ def format_content(self, output: str) -> dict[str, Any]:
         return {"source": self.source, "content": [{"text": {"text": output}}]}
 
     @weave.op
-    def score(self, output: str) -> WeaveScorerResult:
+    def score(self, *, output: str, **kwargs: Any) -> WeaveScorerResult:
```
zbirenbaum, line 63: Since these args are wildcarded now, does that mean that you will have to call `score(output=some_val)` instead of just being able to call `score(some_val)`? It looks like the tests were adjusted in this way. We may want to give a notice that this is potentially a breaking change.

zbirenbaum (approved):
Really awesome work here!

weave/scorers/bedrock_guardrails.py
andrewtruong: Yep.  These were always supposed to be wildcarded -- the fact that they weren't initially is a bug that we didn't catch because the types weren't good enough

zbirenbaum: That's pretty entertaining. I remember when I tried this one out I made similar changes to op.py and then saw all the type errors across the codebase and assumed something was wrong. Thanks for picking this up, huge QOL improvement.

PR #4308, opened 2025-04-27
fix(weave): Resolve datetime import conflict causing Pydantic error

wangbooth:
In `weave/flow/eval_imperative.py`, the use of `from datetime import datetime` (implicitly via `datetime.datetime.now()` in `_default_dataset_name`) caused the name `datetime` in the module scope to refer to the `datetime.datetime` class instead of the `datetime` module.

This led to an `AttributeError: type object 'datetime.datetime' has no attribute 'datetime'` when Pydantic attempted to parse the `ScoreLogger` model, specifically when resolving type annotations potentially involving the `datetime` module (likely within the `Call` type or its dependencies).

## Description

- Fixes #4307 

What does the PR do? Include a concise description of the PR contents.

This commit changes the import to `import datetime` at the top level and updates the usage in `_default_dataset_name` to `datetime.datetime.now()` to ensure `datetime` consistently refers to the module, resolving the Pydantic type resolution error.

## Testing

How was this PR tested?

Y

PR #4309, opened 2025-04-27
fix(weave): fix datetime attribute error

zbirenbaum:
## Description

Currently, whenever running pytest or trying to run trace server I encounter the following error:

`AttributeError: type object 'datetime.datetime' has no attribute 'datetime'`

By importing datetime directly and using `datetime.datetime.now()` instead, the issue seems to be fixed, although it is very unclear why.

### DIFF
--- a/weave/flow/eval_imperative.py
+++ b/weave/flow/eval_imperative.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
 import atexit
+import datetime
 import logging
 from collections.abc import Callable, Iterator
 from contextlib import contextmanager
 from contextvars import ContextVar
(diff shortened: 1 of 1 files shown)


zbirenbaum:
Ran into this issue while working on something else, and it seems to be affecting master as well. It's very unclear to me why this fix does anything at all since these two should be equivalent:

```
# Does not work
from datetime import datetime
datetime.now()
```

```
# Works
import datetime
datetime.datetime.now()
```

I think this might be indicative of another bug being present in evals

zbirenbaum:
Seems like I duplicated #4308 so I'll close this out

PR #4198, opened 2025-04-21
perf(weave): add pre group by ref filtering

gtarpenning:
## Description

[WB-24547](https://wandb.atlassian.net/browse/WB-24547)

Add a ref filtering condition pre-group by. 

Technically we do not need `AND (hasAny(array_concat_agg(calls_merged.input_refs), ['ref'])))` as we are guaranteed that the one call start row contain any refs, so any call starts included at the group by will be valid. Lets include it for now just to be safe, it should not add meaningful latency. 

```sql
SELECT calls_merged.id AS id,
FROM calls_merged
WHERE calls_merged.project_id = 'UHJvamVjdEludGVybmFsSWQ6NDAwMDAyNDg='
        AND ((hasAny(calls_merged.output_refs, ['ref'])) OR length(calls_merged.output_refs) = 0)
        ----------------------------------------------------------------------------------------
        AND ((hasAny(calls_merged.input_refs, ['ref'])) OR length(calls_merged.input_refs) = 0)
        -------------------------------------------------------------------------------------
GROUP BY (calls_merged.project_id, calls_merged.id)
HAVING (((any(calls_merged.deleted_at) IS NULL))
        AND ((NOT ((any(calls_merged.started_at) IS NULL))))
        AND (hasAny(array_concat_agg(calls_merged.output_refs), ['ref'])))
        AND (hasAny(array_concat_agg(calls_merged.input_refs), ['ref'])))
ORDER BY any(calls_merged.started_at) ASC
```

Without this change we get frequent memory exceeded errors, trace link in ticket. This query is the single most errored query in prod the last week, with over 50 error queries (2.2% failure rate). 

Screenshot 2025-04-23 at 6 00 50 PM

## Testing

- query unit tests
- client tests
- manual prod test, on a query that repeatedly **fails** with memory exceeded errors:
(prod, then branch)
Screenshot 2025-04-21 at 10 01 26 AM

[WB-24547]: https://wandb.atlassian.net/browse/WB-24547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4323, opened 2025-04-28
revert(weave): add pre group by ref filtering

gtarpenning:
Reverts wandb/weave#4198


gtarpenning:
no need, fixed forward

PR #4329, opened 2025-04-28
chore(ui): Eval compare: metrics sort

m-rgba:
## Description

| Before | After |
| -------- | ------- |
| before-1 | after-2 |
| before-2 | after-1 |

Creates a consistent sort order (alphabetical) for metrics in eval compare (reported by customer).


m-rgba:
Updated Featurebee:
https://beta.wandb.ai/?betaVersion=81ab756988182268db5e2cb26b38908c07e217b2

m-rgba:
*The lint was a little aggressive because of an indent-level change - if it helps with the diff we can back it out for review and then re-apply it

PR #4331, opened 2025-04-28
chore(weave): add limit to stats query, and special case 

gtarpenning:
## Description

[WB-24603](https://wandb.atlassian.net/browse/WB-24603)

Adds a `limit` to the `calls/stats` endpoint which limits the count query. In the special case where there are no filters, query, and a limit = 1, then we a super optimized check that is guaranteed to be fast.

Frontend prs:
- https://github.com/wandb/weave/pull/4269 
- https://github.com/wandb/core/pull/29093

## Testing

- Unit test
- Manual performance testing, in high trace environment:

| Environment | Elapsed Time | Rows Read  | Data Read | Peak Memory   | Error |
|-------------|--------------|------------|-----------|----------------|--------|
| Prod        | 3.5s         | 53,890,051 | –         | 31.94 GiB      | True   |
| Branch      | 0.008s       | 15,631     | 1.24 MB   | 126.63 MiB     | False  |

In prod, this query is run 1.2 million times a week, so even modest improvements in speed might actually have an impact on clickhouse performance. 

| Env | Runs | Errors | Avg. Read Rows | Avg. Memory Usage | p5 | p90 | p99 |
|-------------|------------------|--------|----------------|-------------------|-------------|-------------|-------------|
| Prod        | 1,169,821        | 0      | 30,333.25      | 95.86 MiB         | 0.019s      | 0.04s       | 0.541s      |

[WB-24603]: https://wandb.atlassian.net/browse/WB-24603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


neutralino1 (commented):
Yeah this is nice.
In an ideal world, we wouldn't have to special-case the query but oh well.

Wanna write a test and a PR description?

gtarpenning:
> Yeah this is nice. In an ideal world, we wouldn't have to special-case the query but oh well.
> 
> Wanna write a test and a PR description?

@neutralino1. 

Yeah, next I'm going to take a stab at upgrading the stats query function, but for now this hard coding is guaranteed to be the fastest query possible. 

Will also break it up into two prs, so we can land the backend change first.

neutralino1 (approved):
Thank you.

PR #4333, opened 2025-04-29
chore(weave): Fix cost query again

jwlee64:
## Description

my previous fix for the cost query introduced a regression while fixing in progress evals returning null with the cost query

2 fixes 
moves the where for the join to prices table to inside the join

old (this filters out rows that have no price in the price table in them, was returning nulls)
"""
LEFT JOIN llm_token_prices ON (llm_usage.llm_id = llm_token_prices.llm_id)
                WHERE ((llm_token_prices.pricing_level_id = {pb_2:String})
                    OR (llm_token_prices.pricing_level_id = {pb_3:String})
                    OR (llm_token_prices.pricing_level_id = {pb_4:String})))
"""

new (leaves the values in, by moving the filter into the join) 
"""
LEFT JOIN llm_token_prices ON ((llm_usage.llm_id = llm_token_prices.llm_id) AND ((llm_token_prices.pricing_level_id = {pb_2:String})
                    OR (llm_token_prices.pricing_level_id = {pb_3:String})
                    OR (llm_token_prices.pricing_level_id = {pb_4:String}))) )
"""

2nd change
"""
# If no cost was found dont add a costs object
    summary_dump_snippet = f"""
    if( any(llm_id) = '{DUMMY_LLM_ID}' or any(llm_token_prices.id) == '',
"""

this was returning costs objects with cost==0 if we had usage but no cost in the system, this was misleading users, who might think their cost was 0, instead it being that we did not have a cost in the db

## Testing

Added another test to ensure we dont return null from un added cost
Updated the original test to test this as well


neutralino1 (approved):
Thank you.

PR #4335, opened 2025-04-29
feat(weave): Support video input and output in SDK

zbirenbaum:
## Description

- Towards WB-22961

This is a scoped down iteration that introduces Video handling to the SDK.
- Image handling has been used as a reference implementation.
- Like the image type handler supports only pillow images, this adds support for moviepy.VideoClip and all subclasses (the video library used by models)

Points to note:
There are *a lot* of edge cases here:
1. Moviepy does not seem particularly stable
2. Attempting to handle more than just VideoFileClips (videos which are loaded from valid files with a known extension) required extra attributes to be set in tests. This may just be a problem with the constructed ones missing metadata that would normally be present
3. I included some fairly extensive test cases, but some strange backend issues with webm encoding still seem to be present
4. Other features like Image preview will come in followup PRs. They were working fine, but there's plenty to review here already and having the load op return a type other than what was directly passed in breaks some of our API promises to the client. The way around this would be to include something like the VideoWrapper in the test cases which the client imports, but @neutralino1 and I discussed avoiding wrapped types so this requires a bit more thought

## Testing
Pretty extensive tests were added. I may cut out some of them, as there's so much going on it got to the point that claude wanted to make a README for the test DIR


tests/trace/type_handlers/Video/README.md
```diff
@@ -0,0 +1,37 @@
+# Video Type Handler Tests
```
neutralino1, line 1: Any reason to capitalize the directory name?

weave/type_handlers/Video/__init__.py
neutralino1, line 1: Same here re: directory name capitalization.

weave/type_handlers/Video/video.py
```diff
@@ -0,0 +1,135 @@
...
+else:
+    dependencies_met = True
+
+if TYPE_CHECKING:
+    from moviepy.editor import (
+        VideoClip,
+        VideoFileClip,
+    )
+
+SUPPORTED_FORMATS = ["gif", "mp4", "webm"]
```
neutralino1, line 28: Maybe use an enum here?

weave/type_handlers/Video/video.py
```diff
@@ -0,0 +1,135 @@
...
+        obj: The VideoClip or VideoWithPreview to save
+        artifact: The artifact to save to
+        name: Ignored, see comment below
+    """
+    is_video_file = isinstance(obj, VideoFileClip)
+
+    if is_video_file:
+        video_format = get_format_from_filename(obj.filename) or DEFAULT_VIDEO_FORMAT
+    else:
+        video_format = getattr(obj, "format", DEFAULT_VIDEO_FORMAT)
```
neutralino1, line 68: I'm not sure I understand the default treatment.
According to your code, if we don't recognize the video format, we assume it is `gif`. So we will write an unknown video format to a GIF file. Do we have any guarantee that playback will work?

neutralino1 (changes requested):
Thank you. A few questions.

tests/trace/type_handlers/Video/video_edge_cases_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+import weave
+
+"""This module tests edge cases in the Video type handler."""
+
+
+@pytest.fixture
+def sample_mp4_path():
+    """Create a sample MP4 file and return its path."""
+    with tempfile.NamedTemporaryFile(suffix=".mp4", delete=False) as tmp:
```
andrewtruong, line 16: this kinda defeats the purpose of a temporary file

It's better to keep `delete=True` and then `yield tmp_path` below.  Then at the end of the test, the temporary file will be automatically cleaned up

weave/type_handlers/Video/video.py
zbirenbaum: The write_gif method should do the encoding with ffmpeg behind the scenes. If we think audio will often be a part of the clip we can do MP4 instead

tests/trace/type_handlers/Video/README.md
zbirenbaum: Convention, all the other type handlers have capital names for their directories. I guess because it refers to a type 'Video' and types are capitalized

weave/type_handlers/Video/video.py
zbirenbaum: I made an enum and constructed the supported formats from its values. Should make it easier/clearer when adding or removing formats

tests/trace/type_handlers/Video/video_edge_cases_test.py
zbirenbaum: Good catch, fixed. Quick question though, what's the functional difference between yielding and returning from the test?

andrewtruong: When you `return`, you just get the value.

When you `yield`, you can put cleanup code afterwards.  In this case you don't explicitly have to write any cleanup because the `TemporaryFile` contextmanager already does it for you, so you can just do

```py
with NamedTemporaryFile(...) as tmp:
    clip = ...
    yield tmp.name

# then the file will get cleaned up automatically when it gets here
```

tests/trace/type_handlers/Video/video_edge_cases_test.py
```diff
@@ -22,26 +23,26 @@ def sample_mp4_path():
         tmp_path, codec="libx264", audio=False, verbose=False, logger=None
     )
 
-    return tmp_path
+    yield tmp_path
```
andrewtruong, line 26: you have to yield inside the context manager, otherwise the file will get cleaned up before it gets to the test

tests/trace/type_handlers/Video/video_edge_cases_test.py
zbirenbaum: Oh now I see what you mean, this would be pretty useful in some other scenarios, thank you for bringing it up

tests/trace/type_handlers/Video/video_edge_cases_test.py
zbirenbaum: I wonder why the tests are still passing with it like this. That might indicate it's not working how we think it is

zbirenbaum: Fixed, the file was being written after the tempfile was closed. I put both the writing and the yield inside and now it works properly. Nice catch.

neutralino1 (approved):
Thank you. One naming nit.

weave/type_handlers/Video/video.py
```diff
@@ -0,0 +1,193 @@
...
+    # Assume there can only be 1 video in the artifact
+    for filename in artifact.path_contents:
+        path = artifact.path(filename)
+        if filename.startswith("video."):
+            return VideoFileClip(path)
+
+    raise ValueError("No video or found for artifact")
+
+
+def is_instance(obj: Any) -> bool:
```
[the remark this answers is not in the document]
zbirenbaum: That is a bit clearer, fixed

PR #4339, opened 2025-04-29
feat(ui): Create WB logo loader component

connieelee:
## Description

https://wandb.atlassian.net/browse/WB-24726

Creates a new loader component with fancy animations. It looks like the W&B logo, and can display a rotation of loading texts. We're going to use this as the new loading screen on workspaces.

https://github.com/user-attachments/assets/951f7c99-251e-479c-90a6-85b502916814

## Testing

Used in core -- see recording above


weave-js/src/components/Loaders/WBLogoLoader.less
connieelee, line 1: this component is heavy on the animations and i had to define several keyframes, so i decided using a .less file would be the most appropriate styling solution here (rather than tailwind)

jo-fang: might be worth adding comment into file for future us to know

jo-fang (approved):

PR #4340, opened 2025-04-29
fix(weave): display name updates in call details page

gtarpenning:
## Description

[WB-24629](https://wandb.atlassian.net/browse/WB-24629)

This pr:
- adds a `refetchOnRename` opt to the `useCall` hook
- invalidates the cache to fetch the new `displayName` only on rename

## Testing

Prod:
prod-display-name1

Branch:
branch-display-name

[WB-24629]: https://wandb.atlassian.net/browse/WB-24629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
@@ -166,7 +166,11 @@ const useMakeTraceServerEndpoint = <
 
 const useCall = (
   key: CallKey | null,
-  opts?: {includeCosts?: boolean; includeTotalStorageSize?: boolean}
+  opts?: {
+    includeCosts?: boolean;
+    includeTotalStorageSize?: boolean;
+    refetchOnRename?: boolean;
+  }
 ): Loadable<CallSchema | null> => {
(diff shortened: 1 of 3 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -187,33 +191,48 @@ const useCall = (
...
-            : null),
-        })
-        .then(res => {
-          loadingRef.current = false;
-          setCallRes(res);
-        });
-    }
-  }, [deepKey, getTsClient, opts?.includeCosts, opts?.includeTotalStorageSize]);
+  const doFetch = useCallback(
+    (invalidateCache = false) => {
```
bcsherma, line 195: I think this is cleaner with `invalidateCache` as a named argument, eg
```suggestion
    ({invalidateCache = false}: {invalidateCache?: boolean} = {}) => {
```

and then called as `doFetch({invalidateCache: true})` or `doFetch({invalidateCache: false})`

bcsherma (approved):

PR #4341, opened 2025-04-29
chore(weave): break out trace count display

gtarpenning:
## Description

Trace count display component so we don't have to use the hook in the frontend.

## Testing

Screenshot 2025-04-29 at 2 16 01 PM


jwlee64 (approved):
ah makes sense

jamie-rasmussen (approved):

PR #4345, opened 2025-04-29
chore(ui): Annotations: user issue - annotations to show

m-rgba:
## Description

image

Annotation display and filtering has a few bugs which this addresses:

1. Annotation items are not filtered for the current user - if any user fills in annotation feedback it will populate the feedback annotation form drawer (this defaults to populating the last annotation).
2. The annotation feedback in the calls table shows the last annotation entered by any user (not the current user, similar to the above).
    - In a perfect world we'd probably have some form of summary columns for all user responses, that's more of an enhancement.
3. Filtering on an annotation column seems to return any of the feedback saved by any user (not the current user).

Solution:

- This adds current user filtering for the calls table and the annotation drawer.
  - Tested adding and changing annotations between multiple users.
  - Tested filtering on success and failure cases across the types with different user situations.


m-rgba:
Updated Featurebee:
https://beta.wandb.ai/?betaVersion=8754dce333f9db9f8aa311c1a94050e69ffcf50d

PR #4356, opened 2025-04-30
chore(weave): make compare button primary + tooltip

gtarpenning:
## Description

The compare view is so great, it deserves some love. Lets try it as primary for a bit. Also add some tooltip text. 

## Testing

Before: 
Screenshot 2025-04-30 at 11 38 16 AM

After:
Screenshot 2025-04-30 at 11 37 55 AM

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -900,6 +900,7 @@ export const CallsTable: FC<{
               </div>
               {isEvaluateTable ? (
                 <CompareEvaluationsTableButton
+                  tooltipText="Compare metrics and examples for selected evaluations"
                   onClick={() => {
                     history.push(
                       router.compareEvaluationsUri(
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTableButtons.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTableButtons.tsx
@@ -437,7 +437,7 @@ export const CompareEvaluationsTableButton: FC<{
     }}>
     <Button
       size="medium"
-      variant="ghost"
+      variant="primary"
       disabled={disabled}
       onClick={onClick}
       icon="chart-scatterplot"


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -900,6 +900,7 @@ export const CallsTable: FC<{
               </div>
               {isEvaluateTable ? (
                 <CompareEvaluationsTableButton
+                  tooltipText="Compare metrics and examples for selected evaluations"
```
J2-D2-3PO, line 903: Since comparison also applies to traces, can we make this text more generic (e.g. "... selected objects"), or dynamic on tab context (e.g. "... selected evals" in the Evals tab and "... selected traces" in the Traces tab)?

Also, since I'm not familiar with UI stuff, is it possible to add markdown links like `[available comparison operations](https://weave-docs.wandb.ai/guides/tools/comparison/#available-actions)`? If so, I think it would be helpful to add this link in, so that users can look at the available actions in the Compare view, etc.

gtarpenning: The buttons have different treatments in the code, the traces compare button still has no tooltip: 
Screenshot 2025-04-30 at 11 46 31 AM

J2-D2-3PO: Example:

```
Compare one or more objects. For usage information, see https://weave-docs.wandb.ai/guides/tools/comparison#available-actions.
```

gtarpenning: You can't add a link to a tooltip because you can't click on it, its only on hover

J2-D2-3PO: ^ ah great point 

Could we add the link into the Comparison view, maybe somewhere near the top?  I still feel like a link out to the docs would help somewhere in the flow

m-rgba: - @J2-D2-3PO since this is an action area we probably don't need docs specifically in here.
- On evals we do have an empty state with docs.
- For both here and Traces we should probably have a more accessible area to access docs which is persistent (after first onboard). Will wire this up in another PR.

m-rgba (approved):

jamie-rasmussen (approved):

PR #4357, opened 2025-04-30
chore(weave): dont allow limit=0 in calls and calls stats query

gtarpenning:
## Description

`limit=0` doesn't make sense, now we error if you try `limit=0`

## Testing

updates test


neutralino1 (commented):
What if `limit` is computed programmatically on the client side and happens to be `0`. Should we not just skip the SQL query and just return `{count: 0}` by default?

PR #4358, opened 2025-04-30
chore(weave): restrict name filter type to string

gtarpenning:
## Description

[WB-24755](https://wandb.atlassian.net/browse/WB-24755)

Restrict the name filter to just string operations. 

## Testing

Prod:
Screenshot 2025-04-30 at 11 58 11 AM

Branch:

Screenshot 2025-04-30 at 11 57 54 AM

[WB-24755]: https://wandb.atlassian.net/browse/WB-24755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
@@ -74,9 +74,11 @@ export const getFieldLabel = (field: string): string => {
   return FIELD_LABELS[field] ?? field;
 };
 
+const STRING_LIKE = 'string-like';
 export const FIELD_TYPE: Record<string, string> = {
-  id: 'id',
+  id: STRING_LIKE,
   'summary.weave.status': 'status',
+  'summary.weave.trace_name': STRING_LIKE,
   wb_user_id: 'user',
   started_at: 'datetime',
 };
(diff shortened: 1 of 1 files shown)


bcsherma (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
```diff
@@ -247,7 +249,7 @@ export const getOperatorValueType = (operatorValue: string): string => {
 
 export const getOperatorOptions = (field: string): SelectOperatorOption[] => {
   const fieldType = getFieldType(field);
-  if ('id' === fieldType) {
+  if (STRING_LIKE === fieldType) {
```
jamie-rasmussen, line 252: Something isn't right here - this doesn't have `contains`

jamie-rasmussen: I think maybe we want to keep `id` as its own special fieldType - I'm not sure e.g. "does not contain" on an id is useful.

jamie-rasmussen (approved):

PR #4361, opened 2025-04-30
chore(weave): add call part size warning

gtarpenning:
## Description

Add warning when logging very large inputs/output.

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -807,6 +807,7 @@ class AttributesDict(dict):
 
 
 BACKGROUND_PARALLELISM_MIX = 0.5
+MAX_TRACE_PAYLOAD_SIZE = 3.5 * 1024 * 1024  # 3.5 MB
 
 
 class WeaveClient:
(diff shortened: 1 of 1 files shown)


weave/trace/weave_client.py
```diff
@@ -906,6 +906,7 @@ def __repr__(self) -> str:
 
 
 BACKGROUND_PARALLELISM_MIX = 0.5
+MAX_TRACE_PAYLOAD_SIZE = int(3.5 * 1024 * 1024)  # 3.5 MB
```
jamie-rasmussen: MB => MiB to be precise. Also wondering if there is a comment or reference that can be added about why this is the number.

weave/trace/weave_client.py
```diff
@@ -1306,23 +1307,30 @@ def send_start_call() -> bool:
...
+                    wb_run_id=current_wb_run_id,
+                    wb_run_step=current_wb_run_step,
                 )
             )
+
+            bytes_size = len(call_start_req.model_dump_json())
+            if bytes_size > MAX_TRACE_PAYLOAD_SIZE:
+                logger.warning(
+                    f"Trace input size ({bytes_size} bytes) exceeds the maximum allowed size of {MAX_TRACE_PAYLOAD_SIZE} bytes."
+                    "Inputs may be dropped."
```
jamie-rasmussen, line 1332: Just for my own education, this is a "may be" and not a "will be"?

jamie-rasmussen (approved):

weave/trace/weave_client.py
gtarpenning: The `3.5MiB` is a rough estimate, and this occurs due to some obscure and undocumented clickhouse insert error, so its possible they patch it or change it in the future.

PR #4363, opened 2025-04-30
feat(weave): Add mp3 support

zbirenbaum:
## Description
PR for Backend MP3 support in Weave:

- Fixes WB-24097

pydub is a good library for audiofiles. However, like the other libraries I looked into it doesn't preserve the filename on the object when you load it. Unfortunately this means:

1. The user will always have to wait for the file to be encoded
2. The file may upload in a lower quality than the original

To address this shortcoming, I created a wrapper class for file paths which just checks if the extension is on our supported extension list. If not, it raises an error. Otherwise, export will just copy the file into the artifact.

#### Example usage for weave.Audio:

```python
import weave

@weave.op
def return_weave_audio(file: str) -> weave.Audio:
    audio = weave.Audio(file)
    return audio

return_weave_audio(file='./audio/test.mp3')
```

In a followup this wrapper class should just be automatically loaded when the user annotates a value like so:
```python
Annotated[str, weave.Audio]
```

#### Example usage for pydub.AudioSegment:

```python
from pydub import AudioSegment
@weave.op
def read_audio_file(file: str) -> AudioSegment:
    segment = AudioSegment.from_mp3(file)
    return segment

read_audio_file(file='./audio/test.mp3')
```

### File Load:

Currently loading requires pydub to be installed for anything that isn't a WAV file. It may be alright to return a weave.Audio object with a URI that the user can pull the file from, since that would be the equivalent of what we do during saving. Would be great to discuss if anyone has input or thoughts on this.

## Testing
Test cases still required ~+ confirmation that import handling works so you can still export wav files without having pydub~


zbirenbaum:
It looks like ffmpeg isn't installed on the test runner. May need to adjust the image?

weave/type_handlers/Audio/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # If there's no dot or it's the last character, return None
+    if last_dot == -1 or last_dot == len(filename) - 1:
+        return AudioFormat.UNSUPPORTED
+
+    # Get the extension without the dot
+    return AudioFormat(filename[last_dot + 1 :])
+
+
+class AudioFile:
```
andrewtruong, line 57: What is the point of this class?

weave/type_handlers/Audio/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    MP3 = "mp3"
+    WAV = "wav"
+    UNSUPPORTED = "unsupported"
+
+    def __str__(self) -> str:
+        return self.value
+
+    @classmethod
+    def _missing_(cls, value: Any) -> "AudioFormat":
```
andrewtruong, line 30: nit: return `Self` type

weave/type_handlers/Audio/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    These are NOT the list of formats we accept from the user
+    Rather, these are the list of formats we can save to weave servers
+    If we detect that the file is in these formats, we copy it over directly
+    Otherwise, we encode it to one of these formats using ffmpeg (mp3 by default)
+    """
+
+    MP3 = "mp3"
+    WAV = "wav"
+    UNSUPPORTED = "unsupported"
```
andrewtruong, line 24: Do you need this?  I think if you don't have this it will raise `ValueError` anyways

weave/type_handlers/Audio/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __str__(self) -> str:
+        return self.value
+
+    @classmethod
+    def _missing_(cls, value: Any) -> "AudioFormat":
+        return cls.UNSUPPORTED
```
andrewtruong, line 31: Same here

weave/type_handlers/Audio/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    If we detect that the file is in these formats, we copy it over directly
+    Otherwise, we encode it to one of these formats using ffmpeg (mp3 by default)
+    """
+
+    MP3 = "mp3"
+    WAV = "wav"
+    UNSUPPORTED = "unsupported"
+
+    def __str__(self) -> str:
+        return self.value
```
andrewtruong, line 27: If you are defining this too you may as well just make it a `AudioFormat: Literal["mp3", "wav"]`

weave/type_handlers/Audio/audio.py
```diff
@@ -25,10 +47,84 @@ def save(obj: wave.Wave_read, artifact: MemTraceFilesArtifact, name: str) -> Non
...
+    elif has_pydub and isinstance(obj, pydub.AudioSegment):
+        with artifact.writeable_file_path(f"audio.{DEFAULT_VIDEO_FORMAT}") as fp:
+            obj.export(
+                fp,
+                format=DEFAULT_VIDEO_FORMAT.value,
+            )
+    elif isinstance(obj, wave.Wave_read):
+        # Object is a wave.Wave_read object
+        save_wave(obj, artifact, name)
+    else:
```
andrewtruong, line 79: this seems like a smell

tests/trace/type_handlers/Audio/audio_segment_test.py
```diff
@@ -0,0 +1,128 @@
...
+
+def test_weave_audio_mp3_as_call_io(client: WeaveClient, tmp_path: Path) -> None:
+    @weave.op
+    def weave_audio_as_input_and_output_part(in_audio: weave.Audio) -> dict:
+        return {"out_audio": in_audio}
+
+    client.project = "test_audio_as_call_io"
+    temp_file = str(tmp_path / "audio.mp3")
+    # Create a temporary audio file
+    make_audio_file(temp_file)
```
andrewtruong, line 82: This setup code should probably be your fixture

tests/trace/type_handlers/Audio/audio_segment_test.py
```diff
@@ -0,0 +1,128 @@
...
+    filename = str(tmp_path / "audio.mp3")
+    duration = 1
+    make_audio_file(filename, duration)
+
+    # Goes in as wrapper around path
+    audio = weave.Audio(filename)
+    weave.publish(audio)
+    ref = get_ref(audio)
+    assert ref is not None
+    weave.ref(ref.uri()).get()
```
andrewtruong, line 53: remove

tests/trace/type_handlers/Audio/audio_segment_test.py
```diff
@@ -0,0 +1,128 @@
...
+    # Comes out as an AudioSegment object
+    gotten_audio: AudioSegment = weave.ref(ref.uri()).get()
```
andrewtruong, line 55: why do you need to `weave.ref(ref.uri())`?  This should be the same as just `ref.get()`

tests/trace/type_handlers/Audio/audio_segment_test.py
```diff
@@ -0,0 +1,128 @@
...
+    source_audio = AudioSegment.from_mp3(temp_file)
+
+    weave_audio_as_input_and_output_part(weave_audio)
+
+    # Load op returns an AudioSegment, so we don't need to .from_mp3 them
+    input_output_part_call = weave_audio_as_input_and_output_part.calls()[0]
+
+    # Load op returns a AudioSegment
+    in_audio = input_output_part_call.inputs["in_audio"]
+    assert in_audio[:5] == source_audio[:5]
```
andrewtruong, line 95: What does this do?

weave/__init__.py
```diff
@@ -21,6 +21,7 @@
 from weave.initialization import *
 from weave.trace.util import Thread as Thread
 from weave.trace.util import ThreadPoolExecutor as ThreadPoolExecutor
+from weave.type_handlers.Audio.audio import AudioFile as Audio
```
andrewtruong, line 24: Why not just call the class `Audio`?

weave/type_handlers/Audio/utils.py
zbirenbaum: I made a similar enum for Video support since  @neutralino1 recommended it in the review over there.  If we think just defining the literals is better we can change it for both, but I'm not sure we actually want to do that. I'm not totally opposed to removing this but I want to point out this flow:

The idea of unsupported is:
- User has `.flac` file
- User instantiates a weave.Audio object on the flac file
- File is detected as being `UNSUPPORTED`
- Instead of directly copying, we can use pydub/ffmpeg backend to load it and export it as `DEFAULT_AUDIO_FORMAT`

This isn't actually an error, it's an expected code path. We could do some try catch handling but I thought this makes it look a bit more obvious.

I see two options
1. Introduce conversion as part of this PR (pretty easy)
2. make it a literal as you suggested below and adjust video support to be similar in a followup.

weave/type_handlers/Audio/audio.py
zbirenbaum: Are you referring to the flow being in a different function or WAV files being treated differently than all others?

Wave is supported by the standard library while all other formats are not. If we change this to use the pydub/ffmpeg backend currently functional code will break.

weave/__init__.py
zbirenbaum: updated

tests/trace/type_handlers/Audio/audio_segment_test.py
zbirenbaum: I'll update it, this should be changed in the existing tests as well.

zbirenbaum:
@andrewtruong @neutralino1 This is ready for another pass.

## Overview:

### Changes from last version:

#### `pydub.AudioSegement`
Pydub has been totally removed
1. Tons of dependency issues with it 
2. Requiring object instances from specific libraries isn't a great experience for us or users...
3. I'm ok with `wave` since it's standard library but we should not play whack-a-mole with library support

#### `weave.Audio` class 
 - intended which for use in pre/post process functions and annotations
 - standardizes various common parameter inputs to a single validated object
 - mirrors the format of a typical payload received from a gen_ai
 - Should work great with Pre/Post processing functions for the Op

#### Backwards Compatibility
 - Retained ability to load `wav` files uploaded as `wave.Wave_read` objects via a metadata file (I couldn't get the metadata field on the artifact to persist)
 - Removed a test for maximum allowed filesize which totally didn't do what it claimed, and took a very long time to run once fixed

 
### Followup:
 - Extensive Testing for `weave.Audio` should probably be added, but i've validated it generally works (This PR)
 - Annotation parsing from within weave.Op (Followup PR)
 -  I'm still learning Pydantic but it feels like I could use it for some more elegant parameter validation in `weave.Audio` (Probably a followup but open to doing here)

## Usage

### Direct initialization in Op pre or post-process function:

#### Filepath with ext:

```python
weave.Audio(path='some_file.mp3')
```

#### Filepath without ext + format:

```python
weave.Audio(path='some_file', fmt='mp3')
```

#### Base64 encoded bytes (Like what LLM generations return) + format:

```python
weave.Audio(data=base64_str_or_bytes, fmt='mp3')
```

#### Raw decoded audio bytes + format:

```python
with open('some_file.mp3', 'rb') as f:
    raw_audio_bytes = f.read()
weave.Audio(data=raw_audio_bytes, fmt='mp3')
```

### Annotated initialization

#### From filepath without format annotation
```python
# Weave input treated as 'str' in while output is treated as weave.Audio
def read_example(path_to_mp3: str) -> Annotated[str, weave.Audio]:
    return path_to_mp3
```

#### From bytes with format annotation
```python
# Weave input treated as 'str' in while output is treated as weave.Audio
def read_example(path_to_mp3: str) -> Annotated[bytes, weave.Audio[Literal["mp3"]]]:
    with open(path_to_mp3, "rb") as f:
        raw_audio_bytes = f.read()
    return raw_audio_bytes
```

#### From base64 encoded str returned by LLM
```python
def gen_audio(prompt: str) -> Annotated[str, weave.Audio[Literal["mp3"]]]:
    completion = client.chat.completions.create(
        model="gpt-4o-audio-preview",
        modalities=["text", "audio"],
        audio={"voice": "alloy", "format": "mp3"},
        messages=[
            {
                "role": "user",
                "content": "Is a golden retriever a good family dog?"
            }
        ]
    )

    return completion.choices[0].message.audio.data
```

tests/trace/type_handlers/Audio/audio_test.py
```diff
@@ -77,27 +137,3 @@ def test_audio_as_call_io(client: WeaveClient, tmp_path: Path) -> None:
 
     assert input_output_part_call.inputs["in_audio"].readframes(5) == exp_bytes
     assert input_output_part_call.output["out_audio"].readframes(5) == exp_bytes
-
-
-def test_audio_with_max_nframes(client: WeaveClient, tmp_path: Path) -> None:
```
zbirenbaum, line 82: This test totally didn't do what it was claiming, and just kinda froze when I fixed it, so I went ahead and removed it. 

I don't think file size is really something we need to test for here since limits aren't applied by any of the `type_handler` surface area.

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+def load(artifact: MemTraceFilesArtifact, name: str) -> "wave.Wave_read | Audio":
+    pytype = None
+    if artifact.path_contents.get("metadata.json"):
+        with open(artifact.path("metadata.json")) as f:
+            pytype = json.load(f).get("_type")
+
+    for filename in artifact.path_contents:
+        path = artifact.path(filename)
+        if filename.startswith("audio."):
+            if (not pytype and filename.endswith(".wav")) or pytype == "wave.Wave_read":
```
andrewtruong, line 202: nit: Better to make explicit `pytype is None` check here

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+
+def load(artifact: MemTraceFilesArtifact, name: str) -> "wave.Wave_read | Audio":
+    pytype = None
+    if artifact.path_contents.get("metadata.json"):
+        with open(artifact.path("metadata.json")) as f:
+            pytype = json.load(f).get("_type")
+
+    for filename in artifact.path_contents:
+        path = artifact.path(filename)
+        if filename.startswith("audio."):
```
andrewtruong, line 201: The `audio.` magic string appears in a lot of places.  Better to pull it out as a const at the top

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+                # which bumps into the 4GB max file size when creating the wave.Wave_write
+                # header on close.
+                wav_file.setframerate(params.framerate)
+                wav_file.setnchannels(params.nchannels)
+                wav_file.setsampwidth(params.sampwidth)
+                wav_file.setcomptype(params.comptype, params.compname)
+                wav_file.writeframes(frames)
+        # Rewind to the original position
+        obj.setpos(original_frame_position)
+    else:
```
andrewtruong, line 188: could also explicit return in the block above so you can avoid this else

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+
+    # Base64 encoded audio data
+    data: str
+
+    def __init__(
+        self,
+        path: Union[str, bytes, Path, os.PathLike, None] = None,
+        data: Union[bytes, str, None] = None,
+        fmt: Union[F, None] = None,
+    ):
```
andrewtruong, line 119: This constructor is very complex.  Consider breaking it out into a few methods like
```py
@classmethod
def from_file(path): ...
@classmethod
def from_data(data, fmt): ...
```

The init can be the most common of the above methods

tests/trace/type_handlers/Audio/audio_test.py
andrewtruong, line 1: Where are the mp3 tests?

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+        )
+
+        return completion.choices[0].message.audio.data
+    """
+
+    # File Format
+    fmt: str
+
+    # Base64 encoded audio data
+    data: str
```
andrewtruong, line 112: do you always need to store the data as a b64 string (e.g. is that required in the mp3 path??  Is it more efficient to store this as the raw bytes and then have a property that converts to b64 on-demand?

zbirenbaum: I went back and forth on this. The main motivation for having it as a b64 string is that when sending it over the wire as json it needs to be a base64 string, so this matches the data fields returned by OpenAI and other generators. I think that you are probably correct that we can just store this as raw bytes though and have a `to_json` or something

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+    Audio class to handle audio data.
+    Can be initialized with a file path or raw audio data with a format
+
+    Direct initialization in Op pre or post-process function:
+
+    From a file with ext:
+    weave.Audio(path='some_file.mp3')
+
+    Filename without ext + format:
+    weave.Audio(path='some_file', fmt='mp3')
```
neutralino1, line 71: Maybe `format` instead of `fmt` for clarity/explicitness?

weave/type_handlers/Audio/audio.py
neutralino1: Do we not have APIs to send over raw bytes?

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+                # Exclude nframes param, it is often set as the maximum number of frames
+                # which bumps into the 4GB max file size when creating the wave.Wave_write
+                # header on close.
+                wav_file.setframerate(params.framerate)
+                wav_file.setnchannels(params.nchannels)
+                wav_file.setsampwidth(params.sampwidth)
+                wav_file.setcomptype(params.comptype, params.compname)
+                wav_file.writeframes(frames)
+        # Rewind to the original position
+        obj.setpos(original_frame_position)
```
neutralino1, line 187: Maybe break out into standalone function for bette readability.

neutralino1 (commented):
Nice work. A few notes.

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+        raw_audio_bytes = f.read()
+    weave.Audio(data=raw_audio_bytes, fmt='mp3')
+
+
+    Annotated initialization performed by SDK:
+
+    def read_example(path_to_mp3: str) -> Annotated[str, weave.Audio]:
+        return path_to_mp3
+
+    def read_example(path_to_mp3: str) -> Annotated[bytes, weave.Audio[Literal["mp3"]]]:
```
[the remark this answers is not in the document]
zbirenbaum: I really wanted to do this but I can't figure out how to get it to work. I've been fighting with the Literal stuff all morning because it's impossible to go from str -> Literal["some_str"]

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
+
+    return fmt
+
+
+def is_base64(data: Union[str, bytes, None]) -> bool:
+    # check if a string is base64 encoded.
+    if not data:
+        return False
+
+    pattern = "^(?:[a-za-z0-9+/]{4})*(?:[a-za-z0-9+/]{2}==|[a-za-z0-9+/]{3}=)?$"
```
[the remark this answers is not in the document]
zbirenbaum: Removed in favor of just attempting a decode in validation mode, it's much more reliable

weave/type_handlers/Audio/audio.py
```diff
@@ -1,34 +1,216 @@
...
-            # which bumps into the 4GB max file size when creating the wave.Wave_write
-            # header on close.
-            wav_file.setframerate(params.framerate)
-            wav_file.setnchannels(params.nchannels)
-            wav_file.setsampwidth(params.sampwidth)
-            wav_file.setcomptype(params.comptype, params.compname)
-            wav_file.writeframes(frames)
-    # Rewind to the original position
-    obj.setpos(original_frame_position)
+def get_format_from_filename(filename: str) -> Union[str, None]:
```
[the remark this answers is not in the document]
zbirenbaum: Thanks this is much better

neutralino1 (approved):
Looks really good, congrats!

PR #4365, opened 2025-04-30
docs(weave): fix docstring with outdated "call" method signature

jamie-rasmussen:
## Description

Actual changes are to weave/trace/context/call_context.py - making the example match the `call` method and updating a reference to a different deprecated `call` method. Everything else is codegen from running `make docs`


zbirenbaum (approved):
LGTM

PR #4369, opened 2025-05-01
chore(weave): use tmp_path builtin and remove redundant test in audio_test.py

zbirenbaum:
## Description

`audio_test.py` currently has two tests doing the same thing and uses tempfile with the delete=False option. This PR removes the redundant test and revises all tests to use the built in temporary test directory fixture provided by pytest.

### DIFF
--- a/tests/trace/type_handlers/Audio/audio_test.py
+++ b/tests/trace/type_handlers/Audio/audio_test.py
@@ -1,6 +1,6 @@
 import math
-import tempfile
 import wave
+from pathlib import Path
 from typing import Optional
 
 import weave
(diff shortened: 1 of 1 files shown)


tests/trace/type_handlers/Audio/audio_test.py
```diff
@@ -25,92 +25,79 @@ def sound_wave(frequency, num_seconds):
...
     client.project = "test_audio_as_dataset_cell"
-    with tempfile.NamedTemporaryFile(delete=False, suffix=".wav") as temp_file:
-        make_audio_file(temp_file.name)
-        audio = wave.open(temp_file.name, "rb")
-        dataset = weave.Dataset(rows=[{"audio": audio}])
-        weave.publish(dataset)
+    temp_file = str(tmp_path / "audio.wav")
+    make_audio_file(temp_file)
+    audio = wave.open(temp_file, "rb")
+    dataset = weave.Dataset(rows=weave.Table([{"audio": audio}]))
```
chance-wnb, line 46: it seems to me that there is an auto-conversion mechanism applied for the `rows` field, so we don't have to specify `weave.Table` explicitly. 

There is definitely nothing need to be changed at all. But I am just curious: did you do this because mypy generated an error?

chance-wnb (approved):
Awesome work. Thank you!

tests/trace/type_handlers/Audio/audio_test.py
zbirenbaum: There is, I just did this because a type error is displayed in my editor and it annoys me. Its functionally the same.

PR #4381, opened 2025-05-01
chore(weave): add better test for latency filtering

gtarpenning:
## Description

This pr:
- adds a better test for latency filtering/sorting
- cast the latency to float in sqlite

## Testing
- updates test


bcsherma (approved):
🙏🏻

PR #4382, opened 2025-05-01
style(weave): Allows 2+ chats, card formatted chats

jwlee64:
## Description
Recommend reviewing with hide whitespace on

beta link
http://beta.wandb.ai/?betaVersion=6d9e6c4e0501c493b78a6dc65378dbb0a9bca9e0

implements first row of designs
https://www.figma.com/design/NfPUcLZ4wgv7sXPyAP8OIw/Playground--Save-model?node-id=17-18343&t=mbV7WsI73X7hQnCZ-0

Adds card styling
Makes chat horizontally scrollable
Removes 2 chat limit

https://github.com/user-attachments/assets/d3c9e9e8-e2fa-43c0-ab12-21aea93c0cd2

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundPage.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundPage.tsx
@@ -1,12 +1,21 @@
 import {Box} from '@mui/material';
 import {WeaveLoader} from '@wandb/weave/common/components/WeaveLoader';
-import React, {useCallback, useEffect, useMemo, useState} from 'react';
+import {Button} from '@wandb/weave/components/Button';
+import React, {
+  Dispatch,
+  SetStateAction,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 
 import {SimplePageLayoutWithHeader} from '../common/SimplePageLayout';
 import {useWFHooks} from '../wfReactInterface/context';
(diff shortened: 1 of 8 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundPage.tsx
```diff
@@ -20,7 +29,29 @@ export type PlaygroundPageProps = {
   callId: string;
 };
 
+type PlaygroundPageInnerProps = PlaygroundPageProps & {
+  setPlaygroundStates: Dispatch<SetStateAction<PlaygroundState[]>>;
```
chance-wnb, line 33: To reduce clutter, you can consider using `& ReturnType<typeof usePlaygroundState>`

chance-wnb (approved):
Thanks for the change. My generic feedback is that:
1. We have a mixed usage of MUI `sx=` and tailwind classes. Can we lean towards classes and adopt tailwind when possible, because they are more efficient (browser level styling, without js computations) 

2. There are intensive usage of `Box`es, but I think Box is an expensive component than mere `div`, often times `div`s might be sufficient. 

If you do decide to refactor, I am happy to look at this PR again. otherwise I have the PR approved, you can also leave it as is, since it is a matter of personal preference to some degree.

m-rgba:
playground

Small async issue. If we create a new chat while the current chat is loading, it clones the processing state over but never completes leaving the card hanging.

PR #4383, opened 2025-05-01
fix(ui): min size and truncation of call name

jamie-rasmussen:
## Description

Fix https://wandb.atlassian.net/browse/WB-24780

Before - as width available for title gets small the height gets huge
Screenshot 2025-05-01 at 2 51 18 PM

After:
We start truncating with ellipsis after 4 (arbitrarily chosen) lines of text. Full text is in tooltip.
Screenshot 2025-05-01 at 2 51 31 PM

Set a minimum width so that name is still at least somewhat visible. This will cause horizontal scrolling but seems better than alternative?
Screenshot 2025-05-01 at 2 51 41 PM

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallOverview.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallOverview.tsx
@@ -20,13 +20,13 @@ export const Overview = styled.div`
 `;
 Overview.displayName = 'S.Overview';
 
-export const CallName = styled.div<{$isEditing?: boolean}>`
+export const CallName = styled.div`
   font-family: Source Sans Pro;
   font-size: 16px;
   font-weight: 600;
   text-align: left;
   word-break: break-all;
-  width: ${props => (props.$isEditing ? '100%' : 'max-content%')};
+  width: 100%;
 `;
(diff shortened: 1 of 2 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallPage/CallOverview.tsx
```diff
@@ -18,13 +18,13 @@ export const Overview = styled.div`
...
 
-export const CallName = styled.div<{$isEditing?: boolean}>`
+export const CallName = styled.div`
   font-family: Source Sans Pro;
   font-size: 16px;
   font-weight: 600;
   text-align: left;
   word-break: break-all;
-  width: ${props => (props.$isEditing ? '100%' : 'max-content%')};
+  width: 100%;
```
jamie-rasmussen, line 27: @m-rgba - I think the `max-content` won't work inside the flex container regardless but the trailing `%` on `max-content` makes that value invalid and ignored anyway.

m-rgba:
Updated Featurebee:
https://beta.wandb.ai/?betaVersion=24bdfe49166b818ed68d9267ff8e5ae30eb90d06

tssweeney (approved):

PR #4384, opened 2025-05-01
fix(ui): do not expand weave objects when moving from schema mapping to confirm step

bcsherma:
## Description

1. Don't flatten dataset columns if the nested data is a weave object, like an image. 
2. Include fields from the flattened objects in the resulting dataset if they have an included field as a prefix of their name.
3. Put a `CustomWeaveTypeProjectContext` in the add to dataset drawer scope so weave objects will render.

https://github.com/user-attachments/assets/037463d4-6ffa-4c8d-a7b0-76930de7ce59

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
@@ -9,6 +9,7 @@ import {useWeaveflowRouteContext} from '../context';
 import {ResizableDrawer} from '../pages/common/ResizableDrawer';
 import {useWFHooks} from '../pages/wfReactInterface/context';
 import {useClientSideCallRefExpansion} from '../pages/wfReactInterface/tsDataModelHooksCallRefExpansion';
+import {CustomWeaveTypeProjectContext} from '../typeViews/CustomWeaveTypeDispatcher';
 import {
   ACTION_TYPES,
   DatasetDrawerProvider,
@@ -39,7 +40,13 @@ export const AddToDatasetDrawer: React.FC<AddToDatasetDrawerProps> = props => {
       onClose={props.onClose}
       entity={props.entity}
       project={props.project}>
-      <AddToDatasetDrawerInner {...props} />
+      <CustomWeaveTypeProjectContext.Provider
+        value={{
+          entity: props.entity,
+          project: props.project,
+        }}>
+        <AddToDatasetDrawerInner {...props} />
+      </CustomWeaveTypeProjectContext.Provider>
     </DatasetDrawerProvider>
   );
 };
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/schemaUtils.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/schemaUtils.ts
@@ -44,6 +44,7 @@ export const FIELD_NAMES = {
   IS_NEW: 'isNew',
   SERVER_VALUE: 'serverValue',
   SELF: 'self',
+  MODEL: 'model',
 };
 
 // Weave metadata namespace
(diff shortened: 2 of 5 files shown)


bcsherma:
* **#4384** <a href="https://app.graphite.dev/github/pr/wandb/weave/4384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/schemaUtils.ts
```diff
@@ -547,11 +547,12 @@ export const mapCallsToDatasetRows = (
...
       const newPrefix = prefix ? `${prefix}.${key}` : key;
 
-      // If it's an object and not an array, recurse and merge results
+      // If it's an object and not an array or weave object, recurse and merge results
       if (
         typeof value === 'object' &&
         value !== null &&
-        !Array.isArray(value)
+        !Array.isArray(value) &&
+        value?._type === undefined
```
bcsherma, line 555: prevent expansion of our objects where these are defined

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/schemaUtils.ts
```diff
@@ -589,21 +588,16 @@ export const mapCallsToDatasetRows = (
...
           sourceValue !== null &&
           !Array.isArray(sourceValue)
         ) {
-          // Flatten nested objects into path-based keys
-          const flattenedValues = flattenObject(sourceValue);
-          Object.keys(flattenedValues).forEach(key => {
-            const fullKey = `${mapping.targetField}.${key}`;
-            rowData[fullKey] = flattenedValues[key];
+          Object.entries(flattenObject(sourceValue)).forEach(([key, value]) => {
+            rowData[`${mapping.targetField}.${key}`] = value;
```
bcsherma, line 598: Check for prefixes in the schema, eg include `inputs.examples.row` if `inputs.example` was mapped

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/schemaUtils.ts
```diff
@@ -639,7 +633,12 @@ export function filterRowsForNewDataset(
 
         const {[WEAVE_NAMESPACE]: weaveData, ...rest} = row;
         const filteredData = Object.fromEntries(
-          Object.entries(rest).filter(([key]) => targetFields.has(key))
+          Object.entries(rest).filter(([key]) =>
+            Array.from(targetFields).some(
+              schemaField =>
+                key === schemaField || key.startsWith(`${schemaField}.`)
+            )
+          )
```
bcsherma, line 641: Include fields from the call if a parent field is included, eg included `inputs.example.row` if `inputs.example` was set

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/AddToDatasetDrawer.tsx
```diff
@@ -39,7 +40,14 @@ export const AddToDatasetDrawer: React.FC<AddToDatasetDrawerProps> = props => {
       onClose={props.onClose}
       entity={props.entity}
       project={props.project}>
-      <AddToDatasetDrawerInner {...props} />
+      <CustomWeaveTypeProjectContext.Provider
+        value={{
+          entity: props.entity,
+          project: props.project,
+          mode: 'dataset_drawer',
```
jamie-rasmussen, line 47: You could omit `mode` if there is no behavior that is dependent on rendering differently when inside a dataset_drawer.

weave-js/src/components/PagePanelComponents/Home/Browse3/datasets/schemaUtils.ts
```diff
@@ -639,7 +633,12 @@ export function filterRowsForNewDataset(
 
         const {[WEAVE_NAMESPACE]: weaveData, ...rest} = row;
         const filteredData = Object.fromEntries(
-          Object.entries(rest).filter(([key]) => targetFields.has(key))
+          Object.entries(rest).filter(([key]) =>
```
jamie-rasmussen, line 638: I don't see where `filterRowsForNewDataset` is getting used?

bcsherma: Oh you're right I ripped that out because we are doing it within `createProcessedRowsMap`, I will just delete this function

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/customWeaveType.types.ts
```diff
@@ -38,9 +38,11 @@ export const isCustomWeaveTypePayload = (
     if (data.load_op == null) {
       return false;
     }
+    // Load op can be a ref, so we need to unwrap it.
+    const load_op = data.load_op?.__val__ ?? data.load_op;
```
jamie-rasmussen, line 42: Does this need to be `?.__val__` - at this point since `data.load_op != null`?

jamie-rasmussen: This line of code is also a little confusing because below it seems like we could still be in the case that `data.load_op` is a string.

jamie-rasmussen (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/customWeaveType.types.ts
bcsherma: Good call, the correct fix here was to ensure that we resolve those `__ref__/__val__` patterns in the `mapCallsToDatasetRows` function in the case where an op has one output and its a custom weave type. I'm removing this change.

PR #4387, opened 2025-05-01
fix(ui): Saved view column set not always getting applied

jamie-rasmussen:
## Description

Fixes https://wandb.atlassian.net/browse/WB-24740 

Could use a closer review on this one since the logic here is complicated and I'm not super familiar.

When you switched views using the panel, you could get an old set of calls grid columns. We have a lot of logic around calculating what the set of columns to show is to keep them stable while new data is loading, especially if that incoming data doesn't require changing the set of columns shown.

In this PR, changing the view causes a refresh of the column calculation to fix the problem. I changed the implementation of `useAllDynamicColumnNames` - its use of `useEffect` meant that you could have renders where the computed dynamic columns lagged behind the data change which I found pretty confusing.

## Testing

Adam's project, confirm that ref expansion is still working.


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
```diff
@@ -124,10 +127,12 @@ export const useCallsTableColumns = (
     [columnsWithRefs]
   );
 
+  // If either of these values has changed we'll reset the dynamic columns.
+  const resetDep = {effectiveFilter, currentViewId};
```
tssweeney, line 131: this will be reset for every render since the dict construction is inline. i think you might want to memo this?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableColumns.tsx
```diff
@@ -814,39 +819,36 @@ const useAllDynamicColumnNames = (
...
-      tableData.forEach(row => {
-        Object.keys(row).forEach(key => {
-          const keyAsPath = stringToPath(key);
-          if (isDynamicCallColumn(keyAsPath)) {
-            nextAsPaths = insertPath(nextAsPaths, stringToPath(key));
-          }
-        });
-      });
+  const prevColumnsRef = useRef<string[]>([]);
+  const prevResetDepRef = useRef<any>(resetDep);
```
tssweeney, line 823: hunch: `prevResetDepRef` should be set to `memoedDep`

tssweeney (approved):

PR #4390, opened 2025-05-02
chore(ui): replace MUI Chip with RemovableTag

jamie-rasmussen:
## Description

Replace the UI for the parentId filter value, from one using a Material UI Chip to one using a W&B common component removable tag.

Besides a more consistent look with the rest of the UI, changes/improvements are:
* Use the display name of the parent call, which might have been changed
* Use the "trailing 4 char" form of id abbreviation instead of an older approach.
* The component automatically has a tooltip if the call name is truncated.

Before:
Screenshot 2025-05-02 at 2 53 47 PM

After:
Screenshot 2025-05-02 at 2 54 11 PM
Screenshot 2025-05-02 at 2 57 11 PM

@Yangyi-wandb @m-rgba - if we go with this we should also replace the Chips used for input/output filters but wanted to get your opinions on it first.

----

Update - have replace Chip component for the input/output object link too.

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
@@ -6,13 +6,7 @@
  *    * (BackendExpansion) Move Expansion to Backend, and support filter/sort
  */
 
-import {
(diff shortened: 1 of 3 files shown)


Yangyi-wandb:
I agree we should replace the MUI chips used elsewhere with the tag

PR #4395, opened 2025-05-02
fix(weave): do not retry call inserts that fail with a non retryable status code

bcsherma:
Modifies the retry logic in `_flush_calls` to prevent infinite retries of non-retyable errors. Previously we requeued all batches that failed. Now we requeue batches that fail with a retryable exception (ie 4XX). If a batch failsfails with with 4XX, we `logger.error` the response code, text, and reason, the dropped start and end event ids that were dropped, and proceed.

Example script that would infinitely retry 400:

```python
import weave

weave.init("testing")

@weave.op()
def get_artifact_url() -> str:  # Malformed ref in output causes a 400 error on request validation
    return "weave:///artifact/bcanfieldsherman/testing/toxic-extra-path/run-4cinli05-test.py:v0"

get_artifact_url()
```

Previously running this generated an infinite loop of:
```
Batch failed after max retries, requeueing batch with len(batch)=2 for later processing
```

and now the output is:
```
Logged in as Weights & Biases user: bcanfieldsherman.
View Weave data at https://wandb.ai/bcanfieldsherman/testing/weave
Error sending batch of 1 call events to server
dropped call end ids: ['01969356-77a1-7441-a4f4-4180c4335465']
status code: 400
reason: Bad Request
text: {"reason":"1 validation error for CallEndCHInsertable\noutput_refs\n  Value error, Unknown ref kind: toxic-extra-path [type=value_error, input_value=['weave-trace-internal://...un-4cinli05-test.py:v0'], input_type=list]\n    For further information visit https://errors.pydantic.dev/2.7/v/value_error"}
```

### DIFF
--- a/weave/trace_server_bindings/remote_http_trace_server.py
+++ b/weave/trace_server_bindings/remote_http_trace_server.py
@@ -11,7 +11,7 @@ from weave.trace.settings import max_calls_queue_size
 from weave.trace_server import requests
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server_bindings.async_batch_processor import AsyncBatchProcessor
-from weave.utils.retry import with_retry
+from weave.utils.retry import _is_retryable_exception, with_retry
 from weave.wandb_interface import project_creator
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


bcsherma:
* **#4395** <a href="https://app.graphite.dev/github/pr/wandb/weave/4395?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4395?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -148,22 +171,25 @@ def _flush_calls(
...
         try:
             self._send_batch_to_server(encoded_data)
-        except Exception:
-            # Add items back to the queue for later processing
-            logger.warning(
-                f"Batch failed after max retries, requeueing batch with {len(batch)=} for later processing",
-            )
+        except Exception as e:
+            if not _is_retryable_exception(e):
+                _log_dropped_call_batch(batch, e)
```
andrewtruong, line 176: Suppose a batch contains 10 calls, and only 1 call raises an non-retry-able error.  Won't the code as written drop all 10 calls?  Shouldn't we just drop the 1 call that is not retry-able?

bcsherma: Yes but the server doesn't tell us which call made the request bad. We could bisect the batch and try each half but that will mean n log n server calls to potentially upload zero calls in the worst case, if all the calls in the batch are bad

bcsherma: Also this PR does not modify that behavior- currently if you have a batch of 10 calls the behavior is we retry the batch forever and drop all the calls because every attempt 400s

andrewtruong: hmm okay I think this is still better than what we have today, but please make a note to revisit this

andrewtruong (approved):

PR #4396, opened 2025-05-03
chore(weave): trace server does not 400 for malformed refs in call inputs or output

bcsherma:
## Description

https://wandb.atlassian.net/browse/WB-24747

If these are included in the `input_refs` and `output_refs` field for a `CHInsertable`, it will lead to a validation error on the pydantic class and then a 400 bad request back to the client. This PR modifies `extract_refs_from_values` to attempt a parse of each of our special uri prefixes before we construct the `CHInsertable` and perform the validation there:
- `weave:///`
- `weave-trace-internal:///`
- `wandb-artifact:///`

If the parse fails, we omit the value from the refs list, since this will lead to a downstream validation error and 400. The result is that our backend treats these malformed uris in call i/o and objects as strings and our queries will not attempt to resolve those refs.

### DIFF
--- a/weave/trace_server/trace_server_interface_util.py
+++ b/weave/trace_server/trace_server_interface_util.py
@@ -28,8 +28,7 @@ def _order_dict(dictionary: dict) -> dict:
     }
 
 
-valid_schemes = [
-    TRACE_REF_SCHEME,
+valid_internal_schemes = [
     ARTIFACT_REF_SCHEME,
     refs_internal.WEAVE_INTERNAL_SCHEME,
 ]
@@ -47,10 +46,16 @@ def extract_refs_from_values(
         elif isinstance(val, list):
             for v in val:
                 _visit(v)
-        elif isinstance(val, str) and any(
-            val.startswith(scheme + "://") for scheme in valid_schemes
-        ):
-            refs.append(val)
+        elif isinstance(val, str):
+            if any(
+                val.startswith(scheme + ":///") for scheme in valid_internal_schemes
+            ):
+                try:
+                    parsed = refs_internal.parse_internal_uri(val)
+                    if parsed.uri() == val:
+                        refs.append(val)
+                except Exception:
+                    pass
 
     _visit(vals)
     return refs
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -51,6 +51,21 @@ from weave.trace_server.trace_server_interface_util import (
 
 ## Hacky interface compatibility helpers
 
+
+def extract_weave_refs_from_value(value):
+    """Extract all strings that start with 'weave:///' from a value."""
+    refs = []
+    if isinstance(value, str) and value.startswith("weave:///"):
+        refs.append(value)
+    elif isinstance(value, dict):
+        for v in value.values():
+            refs.extend(extract_weave_refs_from_value(v))
+    elif isinstance(value, list):
+        for v in value:
+            refs.extend(extract_weave_refs_from_value(v))
+    return refs
+
+
 ClientType = weave_client.WeaveClient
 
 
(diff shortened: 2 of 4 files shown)


bcsherma:
* **#4396** <a href="https://app.graphite.dev/github/pr/wandb/weave/4396?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4396?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/validation.py
```diff
@@ -83,7 +83,12 @@ def object_id_validator(s: str) -> str:
 
 
 def refs_list_validator(s: list[str]) -> list[str]:
-    return [validation_util.require_internal_ref_uri(ref) for ref in s]
+    # Drop any malformed refs
+    return [
+        result
+        for string in s
+        if (result := validation_util.require_internal_ref_uri(string)) is not None
+    ]
```
neutralino1, line 91: Is it preferred behavior to drop malformed inputs or to raise an error?
Don't we risk dropping data (i.e. not inserting)? Don't we also risk missing client bugs that generate malformed inputs?

neutralino1 (commented):
One question.

weave/trace_server/validation.py
bcsherma: Currently we error and drop all the call events in the batch we are inserting, but I believe it would be better to capture the calls and let the user know that the inputs/output contain malformed refs. Dropping data is a cardinal sin in our business and I don't think allowing malformed ref strings in the inputs/output breaks anything (could use more eyes on that hypothesis).

In the referenced code above we are not actually dropping the inputs per se- the inputs will appear in the app exactly as the user logged them. Here are we are only dropping them from the `input_refs` and `output_refs` keys of a few clickhouse columns. (Everything here https://github.com/wandb/weave/blob/c9ee987348f833f3d6cd333e6a9c8e2a60ea979d/weave/trace_server/clickhouse_schema.py that uses `refs_list_validator`.  I believe the effect is that malformed refs are treated by our backend as strings and will be treated by our frontend as refs that do not resolve and display with strikethrough.

Screenshot 2025-05-05 at 9 31 00 AM

We should still be able to detect if our internal ref usage breaks (eg evals start creating malformed refs) because the eval ui will break, and we won't crash user code that returns a malformed ref string. We can also validate the refs client side and warn the user if malformed refs are detected.

neutralino1: Ah ok, that makes sense. That's exactly what I was suggesting on the JIRA ticket. Perfect.

neutralino1 (approved):
Thank you!

tssweeney (changes requested):
Can you please author some tests that exercise this new code path? This would be helpful in understanding exactly what cases you are now trying to protect against. My intuition here is that the change is in the incorrect spot and/or should be changed. Why is "bad" data even getting into the validators in the first place? This seems like something that should be fixed upstream

bcsherma:
@tssweeney I have update the pr to instead attempt a parse of each string with a ref schema we encounter in the object or call i/o. If the parse fails we omit the ref from the list provided for the `CHInsertable*`. I also added two tests which run ops that return malformed weave and artifact URIs and verify that the calls make it to the server and can be fetched.

weave/trace_server/trace_server_interface_util.py
```diff
@@ -47,10 +47,22 @@ def _visit(val: Any) -> Any:
...
             for v in val:
                 _visit(v)
-        elif isinstance(val, str) and any(
-            val.startswith(scheme + "://") for scheme in valid_schemes
-        ):
-            refs.append(val)
+        elif isinstance(val, str):
+            if val.startswith(f"{TRACE_REF_SCHEME}:///"):
+                try:
+                    parse_uri(val)
```
tssweeney, line 53: any reason you skipped the `parsed.uri() == val` in this block?

tssweeney (approved):

PR #4398, opened 2025-05-03
feat(weave): Video display in weave

zbirenbaum:
## Description

Adds a video panel to trace view and a link which opens a modal to the trace table cells containing a video ref

- Fixes WB-24814

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/moviepy.video.VideoClip/VideoPlayer.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/moviepy.video.VideoClip/VideoPlayer.tsx
@@ -1,6 +1,4 @@
 import * as Dialog from '@wandb/weave/components/Dialog/Dialog';
-import {Icon} from '@wandb/weave/components/Icon';
-import {IconNames} from '@wandb/weave/components/Icon/types';
 import {LoadingDots} from '@wandb/weave/components/LoadingDots';
 import {NotApplicable} from '@wandb/weave/components/PagePanelComponents/Home/Browse3/NotApplicable';
 import {CustomLink} from '@wandb/weave/components/PagePanelComponents/Home/Browse3/pages/common/Links';
@@ -101,7 +99,6 @@ const VideoPlayerWithSize: React.FC<VideoPlayerWithSizeProps> = ({
             text={videoText}
             fontWeight={400}
             onClick={() => setShowPopup(true)}
-            icon={<Icon name={IconNames.Play} />}
           />
         </div>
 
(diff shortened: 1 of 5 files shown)


zbirenbaum:
Added Views:
Trace Table
image

On-click `MP4 Video` link in trace table
image

Trace View
image

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/moviepy.video.VideoClip/VideoPlayer.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    return (
+      <>
+        <div
+          style={{
+            display: 'flex',
+            justifyContent: 'flex-start',
+            width: '100%',
+            margin: 'auto',
+            padding: '6px',
```
neutralino1, line 104: You use tailwind below, can you use it here as well?

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/moviepy.video.VideoClip/VideoPlayer.tsx
```diff
@@ -0,0 +1,188 @@
...
+          }}>
+          <CustomLink text={videoText} onClick={() => setShowPopup(true)} />
```
neutralino1, line 106: Any chance we could make the link NOT bold? To differentiate from the call name link at the beginning of the row?

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/moviepy.video.VideoClip/VideoPlayer.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  if (!videoKey) {
+    return <NotApplicable />;
+  }
+
+  const fileExt = VIDEO_TYPES[videoKey];
+  const title = data.custom_name || entity.split('-')[0];
+
+  // Link mode (default view)
+  if (!mode || mode !== 'object_viewer') {
```
neutralino1, line 93: I wonder if it would be wiser to have two components, one for link view and one for player view, instead of one with a `if`.
It seems the link needs much less code, so most of this component would be unused, which could become heavy in a long list of calls.

neutralino1 (changes requested):
A couple of notes.

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/moviepy.video.VideoClip/VideoPlayer.tsx
zbirenbaum: That would be ideal. Unfortunately it's a bit difficult. The code is set up s.t. there's a single component which is used to render the CustomWeaveType. The only way we really have of controlling how this renders is by checking the `mode`.

This context already existed even though most components don't use it, so someone has definitely run into this issue before...

zbirenbaum:
I addressed the comments with the exception of the separate components one since I'm not sure that can be easily done. I also added a play button but I'm not sure how I feel about it...
image

PR #4399, opened 2025-05-05
feat(weave): add google adk otel keys to otel server

parambharat:
## Description

Add otel parsing for google adk otel keys


weave/trace_server/opentelemetry/constants.py
```diff
@@ -53,6 +53,8 @@
     "mlflow.spanInputs",  # From MLFlow's tracking format
     "traceloop.entity.input"  # From Traceloop's conventions
     "input",  # Generic fallback for Pydantic models - lowest priority
+    "gcp.vertex.agent.llm_request",
+    "gcp.vertex.agent.tool_call_args",
```
zbirenbaum, line 57: Let's put these above `input`. It's probably fine in this scenario but `input` should be lowest priority.

Just to confirm:
Once `gcp.vertex.agent.llm_request` is found `gcp.vertex.agent.tool_call_args` won't be checked for.

If both are present in the call this will require some changes so that we can parse out both. Is there a scenario when we would need to parse out both of these values for a single Span?

weave/trace_server/opentelemetry/constants.py
```diff
@@ -66,6 +68,8 @@
     "gen_ai.content.completion",  # From OpenLit project's format
     "traceloop.entity.output",  # From Traceloop's conventions
     "output",  # Generic fallback for Pydantic models - lowest priority
+    "gcp.vertex.agent.llm_response",
+    "gcp.vertex.agent.tool_response",
```
zbirenbaum, line 72: Same here

zbirenbaum (changes requested):
Thanks for adding this! Small ordering change so that the comments remain accurate and these fields don't get overridden by the fallback.

If you happen to have a working example a screenshot would be cool to confirm it wor

weave/trace_server/opentelemetry/constants.py
parambharat: I've added these before the input, also changed the ordering. In my tests I didn't come across a scenario where both need to be parsed

zbirenbaum (approved):
LGTM, thanks for adding this!

parambharat:
hi @zbirenbaum thanks for the quick review. 

> If you happen to have a working example a screenshot would be cool to confirm it works.

How do test this ? Should I be running the weave server locally ?

zbirenbaum:
> How do I test this ? Should I be running the weave server locally ?

Yes, just run the server locally with Governor or Tilt. If you have already confirmed the fields it's probably fine in this scenario, but if you plan on working on OTEL more I would definitely recommend setting it up.

You just need to point the OTEL export to the trace server instance you run locally

PR #4400, opened 2025-05-05
feat(weave): Adds Dataset.map and progress bars

tcapelle:
- Modifies `async_foreach` to support rich progressbar
- Enables `weave.Evaluation` with progress bar, also `verbose=False` to silence printing per row
- Adds `Dataset.map` that uses `async_foreach` to process each example
- Same `inspect` convention as with evaluations.

```py
ds = weave.Dataset(rows=[
    {"id": i, "val": i} for i in range(1000)
])

@weave.op
def double_value(val):
    time.sleep(.001)  # Simulate work
    return {"new": val * 2}

new_ds1 = ds.map(double_value)
```

### DIFF
--- a/tests/trace/test_dataset.py
+++ b/tests/trace/test_dataset.py
@@ -1,4 +1,5 @@
 import pytest
+import asyncio
 
 import weave
 from tests.trace.test_evaluate import Dataset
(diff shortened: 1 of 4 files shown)


tests/trace/test_dataset.py
```diff
@@ -222,3 +224,388 @@ def test_add_rows_to_unsaved_dataset(client):
...
+
+    # Lambda that takes specific params and returns a dictionary
+    mapped_ds = ds.map(lambda a, b: {"sum": a + b})
+    assert len(mapped_ds) == 5
+    assert mapped_ds[0] == {"a": 0, "b": 0, "sum": 0}
+    assert mapped_ds[2] == {"a": 2, "b": 4, "sum": 6}
+
+    # Lambda with a scalar return value should raise TypeError
+    with pytest.raises(TypeError, match="Function must return a dictionary"):
+        ds.map(lambda a, b: a * b)
```
andrewtruong, line 431: is it too magical to intercept a lambda returning a scalar and instead returning `{"result": a *b}`?

weave/flow/dataset.py
```diff
@@ -181,3 +199,178 @@ def select(self, indices: Iterable[int]) -> Self:
...
+    @weave.op
+    def map(
+        self, func: Callable, num_procs: int = get_weave_parallelism()
+    ) -> "Dataset":
+        """
+        Apply a function to each row of the dataset in parallel and return a new dataset.
+
+        The `map` method transforms each row in the dataset by applying `func` to it, adding or
+        updating columns based on the dictionary returned by the function. This is the primary
+        way to transform or enrich datasets.
```
andrewtruong, line 212: What do you mean by enrich datasets here?

weave/flow/dataset.py
```diff
@@ -181,3 +199,178 @@ def select(self, indices: Iterable[int]) -> Self:
...
+                        f"which columns to add or update."
+                    )
+            except Exception as e:
+                # Log the error and propagate it to stop the map operation
+                print(f"Error processing row with map function: {short_str(row)}")
+                traceback.print_exc()
+                raise e
+            else:
+                # Update the original row with the results
+                row_copy = row.copy()
```
andrewtruong, line 349: This is a shallow copy -- is that intentional?

weave/flow/dataset.py
```diff
@@ -181,3 +199,178 @@ def select(self, indices: Iterable[int]) -> Self:
...
+            TextColumn("[progress.description]{task.description}"),
+            MofNCompleteColumn(),
+            TaskProgressColumn(),
+            BarColumn(),
+            IterationSpeedColumn(),
+            TimeElapsedColumn(),
+            console=console,  # Use the instantiated console
+            transient=True,  # Remove progress bar when done
+        ) as progress:
+            return asyncio.run(main(progress))
```
andrewtruong, line 376: Is this safe in all contexts?  e.g. does it work in a notebook, existing async event loop, etc?

weave/flow/eval.py
```diff
@@ -192,7 +202,9 @@ async def summarize(self, eval_table: EvaluationResults) -> dict:
                     summary[name] = model_output_summary
         return summary
 
-    async def get_eval_results(self, model: Union[Op, Model]) -> EvaluationResults:
+    async def get_eval_results(
+        self, model: Union[Op, Model], verbose: bool = False
```
andrewtruong, line 206: Should verbosity be controlled here or in settings?

tests/trace/test_dataset.py
tcapelle: it was like that early, but I don't know rick, if they already have key called `results` we would magically overwrite it?

weave/flow/dataset.py
```diff
@@ -181,3 +199,178 @@ def select(self, indices: Iterable[int]) -> Self:
...
+        processed_rows = []
+
+        # Inspect the function signature
+        sig = inspect.signature(func)
+        param_names = list(sig.parameters.keys())
+
+        # Check if a lambda function is being used and warn about traceability
+        if func.__name__ == "<lambda>":
+            warnings.warn(
+                "Lambda functions in Dataset.map reduce traceability in Weave. "
```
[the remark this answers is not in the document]
tcapelle: It should be more explicit, don't use lambda

weave/flow/dataset.py
tcapelle: nope!

weave/flow/dataset.py
tcapelle: haven't tests =P

weave/flow/eval.py
tcapelle: seems practical here

tcapelle:
@andrewtruong as I added a method to dataset, is it normal that the hashes changes? (and thus the failing tests)

andrewtruong:
Yes.  Since you added a new `op`, the digest is expected to change

tcapelle:
> Yes. Since you added a new `op`, the digest is expected to change

How do I fix the tests then?

weave/flow/dataset.py
```diff
@@ -1,22 +1,40 @@
...
+
+# from weave.flow.eval import console # REMOVED due to circular import
+from rich.progress import (
+    BarColumn,
+    MofNCompleteColumn,
+    Progress,
+    TaskProgressColumn,
+    TextColumn,
+    TimeElapsedColumn,
+)
```
andrewtruong, line 20: Rich is now optional -- please use the viewer system

weave/flow/dataset.py
```diff
@@ -236,3 +254,169 @@ def select(self, indices: Iterable[int]) -> Self:
...
+                        kwargs[name] = row[name]
+                    else:
+                        # Skip this parameter or use a default if available
+                        if sig.parameters[name].default is not inspect.Parameter.empty:
+                            kwargs[name] = sig.parameters[name].default
+                        else:
+                            raise ValueError(
+                                f"Function expects parameter '{name}' but this column "
+                                f"is not in the dataset row: {short_str(row)}"
+                            )
```
andrewtruong, line 375: Is this basically saying:

"Select the subset of row keys that are relevant for this func"?

weave/flow/dataset.py
```diff
@@ -236,3 +254,169 @@ def select(self, indices: Iterable[int]) -> Self:
...
+                        f"from function '{func_name}'. Use a dictionary to specify "
+                        f"which columns to add or update."
+                    )
+            except Exception as e:
+                # Log the error and propagate it to stop the map operation
+                print(f"Error processing row with map function: {short_str(row)}")
+                traceback.print_exc()
+                raise e
+            else:
+                row_copy = copy.deepcopy(row)
```
andrewtruong, line 393: Do you need to do this?  Deepcopy can be expensive for large rows

weave/flow/dataset.py
tcapelle: yeap

weave/flow/dataset.py
tcapelle: I don't want to modify the underlying dataset, I return a new dataset

weave/flow/dataset.py
tcapelle: what is that?

PR #4401, opened 2025-05-05
chore(weave): Allow get to operate on both uris and refs

andrewtruong:
Allows `get` to operate on both uris and refs


weave/trace/api.py
```diff
@@ -214,10 +214,12 @@ def get(uri: str) -> Any:
     dataset2 = weave.get(ref.uri())  # same as dataset!
     ```
     """
+    if isinstance(uri, ObjectRef):
```
jamie-rasmussen, line 217: Should be able to simplify the example code in the docstring now

jamie-rasmussen (approved):

PR #4403, opened 2025-05-05
style(weave): Fix playground add chat animation

jwlee64:
## Description

animation buffer looks like lag in playground
adds option to not use buffer

before

https://github.com/user-attachments/assets/278bff62-d58f-42df-a23c-9c7b59123652

after

https://github.com/user-attachments/assets/ee547968-15c2-4ec0-bc40-9513e1e6ed0d

## Testing

How was this PR tested?


gtarpenning (approved):
why do we need the animation buffer? Drawer bouncing?

jwlee64:
@gtarpenning Lol you added it which is why i requested you 
https://github.com/wandb/weave/pull/2821
the bounce when opening the drawer though

PR #4411, opened 2025-05-06
feat(weave): Mistral chat

tcapelle:
Add hooks for MistralAI chat rendering on the UI.

Screenshot 2025-05-06 at 19 27 50


jwlee64:
How does it work with playground? can you add completions with a support llm top the chat view? want to make sure try in playground works as well, even just for the simple case. 
To pass lint run `yarn lint-fix` in /weave-js/

jwlee64 (approved):

PR #4412, opened 2025-05-06
feat(ui): Add generic file handling and PDF viewer support

jamie-rasmussen:
## Description

Fixes https://wandb.atlassian.net/browse/WB-24856 and https://wandb.atlassian.net/browse/WB-24857

Screenshot 2025-05-06 at 2 27 34 PM

Screenshot 2025-05-06 at 2 28 04 PM

Screenshot 2025-05-06 at 2 30 11 PM

## Testing

* Added PDF and CSV file examples to `wf`.
* Single-page PDF - renders without counter, thumbnails, arrows
* PDF with images - OK, screenshot above
* 26MB, 50+ page PDF - OK, lazy rendering of pages, loading indicators

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/File/FileView.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/File/FileView.tsx
@@ -1,0 +1,251 @@
+import {Button} from '@wandb/weave/components/Button';
+import {Tooltip} from '@wandb/weave/components/Tooltip';
+import React, {useCallback, useEffect, useState} from 'react';
+
+import {convertBytes} from '../../../../../../util';
+import {Icon, IconName} from '../../../../../Icon';
+import {LoadingDots} from '../../../../../LoadingDots';
+import {TailwindContents} from '../../../../../Tailwind';
+import {CustomLink} from '../../pages/common/Links';
+import {useWFHooks} from '../../pages/wfReactInterface/context';
+import {CustomWeaveTypePayload} from '../customWeaveType.types';
+import {PDFView} from './PDFView';
+
+const ICON_MAP: Record<string, IconName> = {
+  'application/json': 'job-program-code',
+  'text/csv': 'table',
+  'text/html': 'job-program-code',
+  'text/xml': 'job-program-code',
+};
+
+const getIconName = (mimetype: string): IconName => {
+  const iconName = ICON_MAP[mimetype];
+  if (iconName) {
+    return iconName;
+  }
+
+  const [type] = mimetype.split('/', 2);
+  if (type === 'image') {
+    return 'photo';
+  } else if (type === 'audio') {
+    return 'music-audio';
+  } else if (type === 'video') {
+    return 'video-play';
+  } else if (type === 'text') {
+    return 'document';
+  }
+
+  return 'document';
+};
+
+type FileTypePayload = CustomWeaveTypePayload<
+  'weave.type_handlers.File.file.File',
+  {file: string; 'metadata.json': string}
+>;
+
+type FileViewProps = {
+  entity: string;
+  project: string;
+  mode?: string;
+  data: FileTypePayload;
+};
+
+type FileMetadata = {
+  original_path: string;
+  mimetype: string;
+  size: number;
+};
+
+export const FileView = ({entity, project, data}: FileViewProps) => {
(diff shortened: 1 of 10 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/File/FileView.tsx
```diff
@@ -0,0 +1,251 @@
...
+import {CustomLink} from '../../pages/common/Links';
+import {useWFHooks} from '../../pages/wfReactInterface/context';
+import {CustomWeaveTypePayload} from '../customWeaveType.types';
+import {PDFView} from './PDFView';
+
+const ICON_MAP: Record<string, IconName> = {
+  'application/json': 'job-program-code',
+  'text/csv': 'table',
+  'text/html': 'job-program-code',
+  'text/xml': 'job-program-code',
```
neutralino1, line 18: Maybe use the `IconNames` const values instead?

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/File/FileView.tsx
```diff
@@ -0,0 +1,251 @@
...
+    return 'photo';
+  } else if (type === 'audio') {
+    return 'music-audio';
+  } else if (type === 'video') {
+    return 'video-play';
+  } else if (type === 'text') {
+    return 'document';
+  }
+
+  return 'document';
```
neutralino1, line 38: Same here. `IconNames.Document`, `IconNames.Photo`, etc.

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/File/FileView.tsx
```diff
@@ -0,0 +1,251 @@
...
+    const a = document.createElement('a');
+    a.href = url;
+    a.download = filename;
+    document.body.appendChild(a);
+    a.click();
+    setTimeout(() => {
+      document.body.removeChild(a);
+      window.URL.revokeObjectURL(url);
+    });
+  }, [fileResult, filename]);
```
neutralino1, line 161: I don't understand what this does.

neutralino1 (commented):
Looking solid. A couple of minor nits.

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/File/FileView.tsx
jamie-rasmussen: Is there a reason you prefer the more verbose style?

jamie-rasmussen: Not any safe type-wise because its a `Record<string, IconName>` but I will change if you think it improves readability.

neutralino1: Yes, I realize `IconName` is just as safe. As you wish, no obligation. I personally don't like loose strings 😁

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/File/FileView.tsx
jamie-rasmussen: This method downloads the blob that is in fileResult in a way that works across browsers.

neutralino1: I see. Can we extract it from the component as a generic function and make `doSave` call it:
```typescript
  const doSave = useCallback(() => {
    const url = URL.createObjectURL(fileResult);
    downloadBlob(url);
  }, [fileResult, filename]);
```
Then it can be reused elsewhere if needed. Maybe also call `doSave` differently, like `doDownload` or something.

jamie-rasmussen: I extracted a `saveBlob` function and added a comment about it. I think "Save" is more accurate than "Download" because at this point the data has already been downloaded from the server, and what you are really doing is saving it from memory to the downloads folder.

Longer term I'd recommend extending the server APIs to include one that sets `Content-Disposition: attachment`.

neutralino1 (approved):
Thank you!

PR #4413, opened 2025-05-06
chore(weave): Add support for tracing generators

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-24713

Adds basic sync and async generator tracing support

See example for code:
```py
from typing import Generator


@weave.op
def basic_gen(x: int) -> Generator[int, None, None]:
    yield from range(x)


@weave.op
def inner(x: int) -> int:
    return x + 1


@weave.op
def nested_generator(x: int) -> Generator[int, None, None]:
    for i in range(x):
        yield inner(i)


@weave.op
def deeply_nested_generator(x: int) -> Generator[int, None, None]:
    for i in range(x):
        for j in nested_generator(i):
            yield j


res = deeply_nested_generator(4)
list(res)
```

image


weave/trace/op.py
```diff
@@ -535,6 +633,428 @@ def on_output(output: Any) -> Any:
...
+    func = op.resolve_fn
+    call = placeholder_call()
+
+    # Handle all of the possible cases where we would skip tracing.
+    if should_skip_tracing_for_op(op):
+        gen = func(*args, **kwargs)
+        call.output = gen
+        return gen, call
+
+    if _should_sample_traces(op):
```
neutralino1, line 652: I don't really understand this block. If `_should_sample_traces` is True, why do we disable tracing?

weave/trace/op.py
```diff
@@ -535,6 +633,428 @@ def on_output(output: Any) -> Any:
...
+            return
+        has_finished = True
+
+        try:
+            # Apply any post-processing to the accumulated state if needed
+            try:
+                if (
+                    hasattr(op, "_on_finish_post_processor")
+                    and op._on_finish_post_processor
+                ):
```
neutralino1, line 697:
```suggestion
                if getattr(op, "_on_finish_post_processor", None) is not None:
```

weave/trace/op.py
```diff
@@ -535,6 +633,428 @@ def on_output(output: Any) -> Any:
     return res, call
 
 
+def _call_sync_gen(
```
neutralino1, line 636: This seems to duplicate a lot of the `_call_sync_func` logic. Is there a chance we could DRY?
Beyond compactness of the code, it's easier to fix bugs when there is only one place to look.

I realize it may not be possible to reuse `_call_sync_func` with additional arguments, the logic may differ a little too much. However, can some of the shared logic be abstracted into small function that could be called by both `_call_sync_func` and `_call_sync_gen`?

weave/trace/op.py
```diff
@@ -535,6 +633,428 @@ def on_output(output: Any) -> Any:
...
+            call.output = gen
+            return gen, call
+
+    # Proceed with tracing
+    try:
+        call = _create_call(
+            op,
+            *args,
+            __weave=__weave,
+            extra_attributes={"is_generator": True},
```
neutralino1, line 664: Why not add this to `__weave`?

neutralino1 (commented):
Amazing work!
A few nits, and a DRY comment.

weave/trace/op.py
```diff
@@ -670,15 +1194,34 @@ def create_wrapper(func: Callable[P, R]) -> Op[P, R]:
                 @wraps(func)
                 async def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:  # pyright: ignore[reportRedeclaration]
                     res, _ = await _call_async_func(
-                        cast(Op, wrapper), *args, __should_raise=True, **kwargs
+                        cast(Op[P, R], wrapper), *args, __should_raise=True, **kwargs
                     )
                     return cast(R, res)
+            elif is_generator:
```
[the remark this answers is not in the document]
andrewtruong: will update for clarity, but note that your comment is not true in python!
https://peps.python.org/pep-0525/

image

weave/trace/op.py
andrewtruong: Good catch, there should be a sampling param here that is missing

andrewtruong: I guess it reads better as `if not _should_sample_traces(op)` -- then that function should be inverted.

The intent here is to say "if we are sampling, then dont always log everything"

weave/trace/op.py
andrewtruong: I think it's worth decoupling at least for now.  I'm concerned about the complexity of fixing bugs in one type of function affecting another type (e.g. a sync gen change affects regular sync func).

If in future we find that it's unnecessary I'm happy to merge.

weave/trace/op.py
andrewtruong: good point

weave/trace/op.py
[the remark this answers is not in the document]
neutralino1: That actually makes sense.

weave/trace/op.py
```diff
@@ -187,6 +205,8 @@ class Op(Protocol[P, R]):
     call: Callable[..., Any]
     calls: Callable[..., CallsIter]
 
+    _accumulator: Callable[[Any | None, Any], Any] | None
```
chance-wnb, line 208: Is the Accumulator concept a parallel new feature? Or is it something necessary to make the generator flow working?

weave/trace/op.py
chance-wnb: I also feel that there is an opportunity to make it "DRY"er. I focus on the context management part. Maybe we can have a templated function that does the common context management logic, and have the differences of sync and async specialized logic as template fillers. I think it might improve readability, but if it cannot, it's not worth doing it.

You can make a call on when and whether to do it.

chance-wnb (approved):
Fantastic work! Thank you!

weave/trace/op.py
andrewtruong: This is something to make accumulating generators work

This lets the user specify a reducer func over the generator, e.g. to accumulate streaming chat completion results

weave/trace/op.py
neutralino1: We now have four functions doing similar things with slight differences: `_call_sync_func`, `_call_async_func`, `_call_sync_gen`, `_call_async_gen`. That's not so great. Any chance we can rationalize that? Could be in a follow-up PR if you're itching to merge this soon for Perplexity.

neutralino1 (approved):
:shipit:

PR #4429, opened 2025-05-07
perf(weave): add parent id filtering pre-group by

gtarpenning:
## Description

The very last of the low hanging fruit to do filtering pre-group by on conditions that exist on the call starts. It might actually be slightly slower for already very fast queries, but has a big impact on larger queries, especially on peak memory usage. 

Before:
```sql
SELECT
    calls_merged.id AS id
FROM calls_merged
WHERE calls_merged.project_id = {pb_2:String}
    AND ((NOT ((any(calls_merged.started_at) IS NULL))
    AND (any(calls_merged.parent_id) IN {pb_0:Array(String)}))
```

Now:
```sql
SELECT
    calls_merged.id AS id
FROM calls_merged
WHERE calls_merged.project_id = {pb_2:String}
    AND (calls_merged.parent_id IN {pb_1:Array(String)}
        OR calls_merged.parent_id IS NULL)
    AND ((NOT ((any(calls_merged.started_at) IS NULL))
    AND (any(calls_merged.parent_id) IN {pb_0:Array(String)}))
```

## Testing

- unit test
- lots of existing parent_id end to end coverage

For a heavy query: 
|        | Elapsed | Rows Read   | Data Read | Peak Memory |
|--------|---------|-------------|-----------|-------------|
| prod   | 4.958s  | 69,805,742  | 19.86 GB  | 9.35 GiB    |
| opt    | 3.384s  | 69,804,304  | 19.44 GB  | 636.6 MiB   |

For a light query:
|        | Elapsed | Rows Read | Data Read | Peak Memory  |
|--------|---------|-----------|-----------|--------------|
| prod   | 0.040s  | 302,031   | 79.34 MB  | 129.81 MiB   |
| opt    | 0.048s  | 302,314   | 51.72 MB  | 113.09 MiB   |


bcsherma (approved):
I think it may warrant a comment that we are conditioning on whether a parent id is provided in two places in different ways when constructing the query, but otherwise LGTM.

PR #4436, opened 2025-05-07
docs(weave): Document mp4 support

J2-D2-3PO:
## Description

- Fixes https://wandb.atlassian.net/browse/DOCS-1486

Documents mp4 support

## Testing

1. yarn start on local
2. Tested and updated original code sample in a python venv with `weave` and `moviepy==1.0.3`


docs/docs/guides/tracking/video.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. Initializes a Weave project `video-test`.
+2. Defines a `get_video` function tracked as a `weave.op` that extracts a 5-second subclip of the loaded `VideoFileClip` as a `VideoClip`.
+3. Uploads and tracks the clip in Weave.
+4. Automatically generates a dummy MP4 video if none is found.
+
+Before you can use the code sample, complete the prerequisites:
+
+1. Install `weave` and `moviepy==1.0.3`.
+2. Create a W&B account
```
J2-D2-3PO, line 46:
```suggestion
2. Create a W&B account.
```

docs/docs/guides/tracking/video.md
```diff
@@ -0,0 +1,89 @@
...
+weave.init('video-test')
+
+@weave.op
+def get_video(clip: VideoFileClip) -> VideoClip:
+    """Process a video by path rather than by passing the clip directly.
+
+    This ensures that the VideoFileClip is created and managed within the
+    Weave op's thread context, avoiding thread-safety issues.
+    """
+    new_clip = clip.subclipped(0, 5)
```
zbirenbaum, line 67: Lets make this `clip.subclipped(0, 1)` so that the dummy clip and original clip differ

zbirenbaum (changes requested):
One small note on the code and also it would be good to show the video player in the trace view in the screenshot rather than just the trace table

docs/docs/guides/tracking/video.md
```diff
@@ -0,0 +1,91 @@
...
+weave.init('video-test')
+
+@weave.op
+def get_video(clip: VideoFileClip) -> VideoClip:
+    """Process a video by path rather than by passing the clip directly.
+
+    This ensures that the VideoFileClip is created and managed within the
+    Weave op's thread context, avoiding thread-safety issues.
+    """
+    new_clip = clip.subclipped(0, 1)
```
J2-D2-3PO, line 67:
```suggestion
    new_clip = clip.subclip(0, 1)
```

docs/docs/guides/tracking/video.md
J2-D2-3PO: Updated, also realized that the code was using the function from 2.1.2 (subclipped) from my testing, updated it to use the 1.0.3. function (subclip) https://github.com/wandb/weave/pull/4436/files#diff-ea9928fd306380baddb60acc9456f702ff0a27b4b439ed3033819dc504fdc50fR67

zbirenbaum:
It looks like something is going on with the Video Player in the trace view, I'm looking into it

jamie-rasmussen (commented):
Thoughts on having this as a separate page vs. with audio and images on https://weave-docs.wandb.ai/guides/core-types/media ?

J2-D2-3PO:
> Thoughts on having this as a separate page vs. with audio and images on https://weave-docs.wandb.ai/guides/core-types/media ?

I get the idea, but I feel like that page may get bloated quickly, since the video support page is it's own full page worth of contents and it sounds like we are going to have more audio support soon. I also want to make sure users can find what they are looking for (e.g. tracing video info from the Tracing subsection) 

If anything, in the future we might want to break "Logging Media" into a grouping of multiple pages (Audio, Video, Images) each with their own intro page, examples, etc.. Which sounds like it's getting into IA refactor territory again... 😅 so I vote let's just ship this as is for now

J2-D2-3PO:
@zbirenbaum I think I've addressed everything here, lmk if / when this is good to merge

zbirenbaum (approved):
Looks awesome, thank you!

PR #4441, opened 2025-05-08
chore(weave): Zero out call stack before each test

andrewtruong:
This helps resolves some CI issues where a straggling test that incorrectly cleaned up could cause a cascade of test failures.

We should still be on the look out for those stragglers


zbirenbaum (approved):
LGTM

PR #4443, opened 2025-05-08
feat(weave): automatically parse audio based on annotations

zbirenbaum:
## Description

Basically just an auto-applied pre/post processor based on type annotations for audio.


zbirenbaum:
There are some sanitization TODOs around where to place the specific files but this officially works

image

```python
import weave
import base64
from typing import Annotated, Literal

weave.init('audio_annotation_test')

@weave.op
def with_annotation(path_str: Annotated[str, weave.Audio]) -> Annotated[str, weave.Audio[Literal["mp3"]]]:
    binary_data = open(path_str, 'rb').read()
    return base64.b64encode(binary_data).decode('ascii')

@weave.op
def without_annotation(path_str: str) -> str:
    binary_data = open(path_str, 'rb').read()
    return base64.b64encode(binary_data).decode('ascii')

def main():
    fpath = './audio/gen.mp3'
    with_annotation(fpath)
    without_annotation(fpath)


if __name__ == "__main__":
    main()
```

Issue #4444, opened 2025-05-08
"unsupported locale setting"


RoniFinTech3:
I pip installed the latest Weave version (0.51.45) and I started seeing this error:

```
[ERROR] Error: unsupported locale setting
Traceback (most recent call last):
  File "/var/lang/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
.
.
.
  File "/var/lang/lib/python3.12/site-packages/weave/__init__.py", line 19, in <module>
    from weave.flow.saved_view import SavedView
  File "/var/lang/lib/python3.12/site-packages/weave/flow/saved_view.py", line 13, in <module>
    from weave.trace.grid import Grid
  File "/var/lang/lib/python3.12/site-packages/weave/trace/grid.py", line 25, in <module>
    locale.setlocale(locale.LC_ALL, "")
  File "/var/lang/lib/python3.12/locale.py", line 615, in setlocale
    return _setlocale(category, locale)
```

Could be an issue with this version?

jamie-rasmussen:
Thank you for reporting this - I wonder if you might have an unexpected value set for the environment variable `LANG` or `LC_ALL`? I should have a fix ready shortly, downgrading to `0.51.44` should also work.

jamie-rasmussen:
We have released v0.51.46 with a fix for this. If you are able to verify that release resolves the issue for you it would be appreciated. Thank you!

andrewtruong:
Closing as resolved

PR #4449, opened 2025-05-08
fix(weave): Grid locale can error

jamie-rasmussen:
## Description

Fix for https://github.com/wandb/weave/issues/4444

## Testing

I haven't been able to reproduce the problem on my machine when passing `''`. I can get the locale error if I do:
`import locale; locale.setlocale(locale.LC_ALL, 'invalid_locale')`

I thought it might be because I have `LANG=en_US.UTF-8` in my environment, but it doesn't error for me even if I unset that, so maybe this is OS or Python version specific? In any case, I believe this should resolve the error for the user.

### DIFF
--- a/weave/trace/grid.py
+++ b/weave/trace/grid.py
@@ -22,7 +22,11 @@ if TYPE_CHECKING:
 
 
 # Set locale to user's default setting
-locale.setlocale(locale.LC_ALL, "")
+try:
+    locale.setlocale(locale.LC_ALL, "")
+except locale.Error:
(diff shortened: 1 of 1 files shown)


gtarpenning (approved):
Interesting... I wonder if we need a fallback for the fallback

jamie-rasmussen:
> Interesting... I wonder if we need a fallback for the fallback

Yeah, I was having trouble figuring that out from the docs - seems like it could be an issue with a more esoteric runtime? Figure we'd cross that bridge if we get a complaint.

PR #4452, opened 2025-05-08
chore(weave_query): Use previous increment paths to handle artifact ordering

domphan-wandb:
## Description

https://wandb.atlassian.net/browse/WB-20520

This PR changes the implementation for loading up incremental tables to handle the scenario where incremental artifacts are received by the backend in different orders.

The artifacts system will be eventually consistent in the sense that all table increments will be present in the final version of the artifact.

Based on that eventual consistency, on the SDK, we store the artifact urls for the previous increment paths. which guarantees the contents and the order of the increments for the history row at the time the increment was logged.

Query engine uses the `previous_increments_paths` list to load up the incremental table with its data in the correct order.

## Testing

How was this PR tested?
locally invoker.ini

### DIFF
--- a/weave_query/weave_query/ops_domain/table.py
+++ b/weave_query/weave_query/ops_domain/table.py
@@ -20,7 +20,6 @@ from weave_query import weave_types as types
 from weave_query.api import op, weave_class
 from weave_query.ops_domain import trace_tree, wbmedia
 
-
 @dataclasses.dataclass(frozen=True)
 class TableType(types.ObjectType):
     name = "table"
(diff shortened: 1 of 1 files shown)


weave_query/weave_query/ops_domain/table.py
```diff
@@ -687,33 +686,33 @@ def _get_incremental_table_awl_from_file(
 ) -> ops_arrow.ArrowWeaveList:
     from weave_query.ops_domain.wb_util import escape_artifact_path, _filesystem_artifact_file_from_artifact_path
     all_awls: list[ops_arrow.ArrowWeaveList] = []
-    increment_dir = file.artifact.path_info("")
-    if isinstance(increment_dir, artifact_fs.FilesystemArtifactDir):
```
nicholaspun-wandb, line 691: When a run completes, will the summary metrics point to the final increment file for the table or the directory? If we do somehow come across a directory, is that an error or should we keep this code path?

weave_query/weave_query/ops_domain/table.py
```diff
@@ -687,33 +686,33 @@ def _get_incremental_table_awl_from_file(
...
-            
-        object_type = types.union(*object_types)
-
-        for rows, file in zip(rrows, files.values()):
-            all_awls.append(_get_table_awl_from_rows_object_type(rows, object_type, file))
+    files = {}
+    if "previous_increments_paths" in data:
+        # Get the last 100 increments (including the current file)
+        # If we have more than 100, otherwise use all increments
+        all_increment_paths = data["previous_increments_paths"]
```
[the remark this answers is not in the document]
domphan-wandb: Thanks for this comment, I added more handling for run resumption: https://github.com/wandb/wandb/pull/9834

The behavior that I added is that it'll look at the summary and if the type at that key is an incremental-table-file, it'll assign the `previous_increment_paths` and append `artifact_path.`

So the resumed table have all the previous increments paths unless they log different types to the same key.

weave_query/weave_query/ops_domain/table.py
domphan-wandb: The summary metric `artifact_path` will always point to an ArtifactEntry, so a file. It shouldn't ever be a directory.

nicholaspun-wandb (approved):

PR #4462, opened 2025-05-09
feat(weave): MP3 support on frontend

zbirenbaum:
## Description

Requires: #4363

- Fixes WB-24097

Adds MP3 Support for frontend

image

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
@@ -2,7 +2,7 @@ import {MOON_350, TEAL_500} from '@wandb/weave/common/css/color.styles';
 import {formatDurationWithColons} from '@wandb/weave/common/util/time';
 import {Button} from '@wandb/weave/components/Button';
 import {Tailwind} from '@wandb/weave/components/Tailwind';
-import React, {FC, useEffect, useMemo,useRef, useState} from 'react';
+import React, {FC, useEffect, useMemo, useRef, useState} from 'react';
 import WaveSurfer from 'wavesurfer.js';
 
 import {LoadingDots} from '../../../../../LoadingDots';
@@ -29,11 +29,14 @@ export const AudioPlayer: FC<{
   );
 
   // Memoize the params to prevent unnecessary re-queries
-  const fileContentParams = useMemo(() => ({
-    entity,
-    project,
-    digest: audioFile ? data.files[audioFile] : '',
-  }), [entity, project, audioFile, data.files]);
+  const fileContentParams = useMemo(
+    () => ({
+      entity,
+      project,
+      digest: audioFile ? data.files[audioFile] : '',
+    }),
+    [entity, project, audioFile, data.files]
+  );
 
   const audioBinary = useFileContent(fileContentParams);
 
@@ -57,8 +60,9 @@ export const AudioPlayer: FC<{
 
   const downloadFile = () => {
     if (!audioUrl) {
-      return console.error('Audio URL is not available');
-    };
+      console.error('Audio URL is not available');
+      return;
+    }
     const a = document.createElement('a');
     a.href = audioUrl;
     const date = new Date().toISOString().split('T')[0];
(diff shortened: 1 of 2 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
```diff
@@ -21,12 +21,25 @@ export const AudioPlayer: FC<{
...
+  // Find the first audio file in the files object
+  const audioFile = Object.keys(data.files).find(fileName =>
+    fileName.startsWith('audio.')
+  );
+
   const audioBinary = useFileContent({
     entity,
     project,
-    digest: data.files['audio.wav'],
+    digest: audioFile ? data.files[audioFile] : '',
```
gtarpenning, line 34: I think we might want to memo this. If the component gets rendered again it might re-query depending on how data.files is defined

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
```diff
@@ -40,10 +53,11 @@ export const AudioPlayer: FC<{
   }
 
   const downloadFile = () => {
+    if (!audioUrl || !audioFileName) return;
     const a = document.createElement('a');
     a.href = audioUrl;
     const date = new Date().toISOString().split('T')[0];
-    a.download = `${project}_${date}_audio.wav`;
+    a.download = `${project}_${date}_${entity}_${audioFileName}`;
```
gtarpenning, line 60: confirming audioFileName includes the media type right?

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
```diff
@@ -40,10 +53,11 @@ export const AudioPlayer: FC<{
   }
 
   const downloadFile = () => {
+    if (!audioUrl || !audioFileName) return;
```
gtarpenning, line 56: I wonder if we should error? either with a toast or even just to the browser console?

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
zbirenbaum: Good call, I changed it to Memoize the fileContentParams

I notice for traces with a lot of audio files there's still a bit of pop-in but I think it was pre-existing...

gtarpenning (approved):
lgtm, tested on a project with existing wave audio files and works.

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
zbirenbaum: I think a console error is fine, I removed the file name check since the url won't exist without the name.

Download button doesn't exist until button until the URL has loaded in and the player rendered so neither should ever occur.

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
zbirenbaum: Yup, audioFileName includes the extension 
image

weave-js/src/components/PagePanelComponents/Home/Browse3/typeViews/Audio/AudioPlayer.tsx
```diff
@@ -56,7 +56,9 @@ export const AudioPlayer: FC<{
   }
 
   const downloadFile = () => {
-    if (!audioUrl || !audioFileName) return;
+    if (!audioUrl) {
+      return console.error('Audio URL is not available');
```
gtarpenning, line 60: what happens when you return a console error? shouldn't you error, then return None?

zbirenbaum: console.error is a void function so it shouldn't do anything but I'll change it just in case

gtarpenning: yeah ig, just looks weird

PR #4465, opened 2025-05-09
chore(weave): Decompose useExampleCompareData for Eval Compare Page

tssweeney:
Recently @chance-wnb did some great work to improve the loading and caching of our eval compare page. I am reusing some of these methods in an upcoming UI enhancement. In particular, rending a list-view instead of a detail view. 

The current implementation of useExampleCompareData couples data caching, fetching, and prefetching into a single hook. While i wanted to reuse this logic, the current interface was not well suited for my new use case.

This PR is a non-functional change, but decouples the different functionalities and makes it a bit easier to reuse. Specifically:

`useExampleCompareData` is renamed to `useExampleCompareDataAndPrefetch` and:

```
/**
 * The following functions are used to fetch data rows (and cache them in the state - Note: caching them in
 * the state is a little hacky since it is not reactive, but it does work for our use case)
 *
 * Anyway, there are the following methods:
 * * `useExampleCompareDataAndPrefetch` - This is used to fetch a target row and prefetch the adjacent rows.
 *   * This is the primary exported hook from this section.
 * * `useExampleCompareData` - This is used to fetch a single row.
 * * `loadMissingRowDataIntoCache` - This is used to asynchronously load rows that are not yet in our cache.
 * * Then there are a few helper functions that are used to make the above hooks work:
 *   * `makePartialTableReq` - This is used to make a partial table request for a dataset.
 *   * `usePartialTableRequest` - This is used to wrap the `makePartialTableReq` hook and provide a reactive partial table request.
 *   * `loadRowDataIntoCache` - This is used to directly load rows into our cache.
 *   * `getCachedRowData` - This is used to directly get a row from our cache.
 */

```

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
@@ -412,8 +412,7 @@ function useExampleCompareData(
       partialTableRequest
     ).then(rows => {
       if (mounted) {
-        const data = rows[0]?.val;
-        // If the value is already loaded, don't fetch again
+        const data = rows[0];
         if (data != null) {
           setTargetRowValue(flattenObjectPreservingWeaveTypes(data));
         }
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
@@ -1,17 +1,11 @@
-import {parseRef, parseRefMaybe, WeaveObjectRef} from '@wandb/weave/react';
-import _, {isEmpty} from 'lodash';
-import {
-  MutableRefObject,
-  useCallback,
-  useEffect,
-  useMemo,
-  useRef,
-  useState,
-} from 'react';
+import {isWeaveObjectRef, parseRefMaybe} from '@wandb/weave/react';
+import _ from 'lodash';
+import {useEffect, useMemo, useState} from 'react';
 
 import {flattenObjectPreservingWeaveTypes} from '../../../../flattenObject';
 import {TraceServerClient} from '../../../wfReactInterface/traceServerClient';
 import {useGetTraceServerClientContext} from '../../../wfReactInterface/traceServerClientContext';
+import {TraceTableQueryReq} from '../../../wfReactInterface/traceServerClientTypes';
 import {projectIdFromParts} from '../../../wfReactInterface/tsDataModelHooks';
 import {
   buildCompositeMetricsMap,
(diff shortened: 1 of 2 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -316,189 +310,243 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
-        cachedRowData.current[selectedRowDigest] = rawDataRow;
-        increaseCacheVersion();
-        return;
-      }
+// Primary method for fetching and caching a single row
+function useExampleCompareData(
+  state: EvaluationComparisonState,
+  targetDigest: string
+) {
+  const initialValue = getCachedRowData(state, targetDigest);
```
chance-wnb, line 384: Nit: I don't think you have to run this `getCachedRowData` on every render, you could move it to useEffect and only call it the first time if it is not set.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -316,189 +310,243 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
+    }
+    // immediately fetch the current row
+    setLoading(true);
+
+    loadRowDataIntoCache(
+      client,
+      state,
+      [targetDigest],
+      partialTableRequest
+    ).then(rows => {
```
chance-wnb, line 413: It is possible to use async function in useEffect
```
useEffect(() => {
    (async() => {
       ...
    })();
}, ...) 
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -316,189 +310,243 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
+// Helper to fetch and store row data in our state cache
+async function loadRowDataIntoCache(
+  client: TraceServerClient,
+  state: EvaluationComparisonState,
+  rowDigests: string[],
+  partialTableRequest: PartialTableRequestType
+): Promise<any[]> {
+  const setCachedRowData = (digest: string, data: any) => {
+    const currentRow =
+      state.loadableComparisonResults.result?.resultRows?.[digest];
```
chance-wnb, line 530: This is controversial, but I think that relying on the `EvaluationComparisonState` to store the cache is fragile -- the `state` is ephemeral and up to the change of the parent, the cache might be randomly gone. The fact that it is working right now is just a coincidence. 

We can leave it as is, but I suggest we look into https://jotai.org/ for robustness in the future. I know your goal is to make this info accessible for higher level. An atom is global but not owned by a specific component.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -316,189 +310,243 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
+      if (mounted) {
+        const data = rows[0];
+        if (data != null) {
+          setTargetRowValue(flattenObjectPreservingWeaveTypes(data));
+        }
+        setLoading(false);
       }
+    });
+    return () => {
+      mounted = false;
```
chance-wnb, line 423: Maybe this cleanup function doesn't achieve any goal.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
tssweeney: `getCachedRowData` is a simple key-value lookup to to a memory stable object. Why add these additional layers of complexity?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
tssweeney: i think this is stylistic preference. Based on your comment here and in my other PR, you prefer the `async/await` patterns. Other than style, is there a benefit to this approach?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
tssweeney: it is actually quite important. If you have promises / async effects which have deferred state update, they could finish after the component is unmounted resulting in a react warning saying that the program is attempting to update state after unmount. This pattern (toggling the mounted flag) is used in a number of places in our app to prevent this edge case.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -316,189 +310,243 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
+  const [partialTableRequest, setPartialTableRequest] =
+    useState<PartialTableRequestType | null>(null);
+
+  const datasetRef = useMemo(() => {
+    return Object.values(state.summary.evaluations)[0].datasetRef as string;
+  }, [state.summary.evaluations]);
+
+  useEffect(() => {
+    let mounted = true;
+    makePartialTableReq(client, datasetRef).then(res => {
```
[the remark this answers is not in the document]
tssweeney: See comment above - let's discuss if this pattern is unclear. The happy path is
```
1. Component Mount
2. Effect Runs
3. Request started
4. Request finishes
5. state updated
6. component unmount
```

However, if component unmount happens between start and finish of request we avhe
```
1. Component Mount
2. Effect Runs
3. Request started
4. component unmount
5. Request finishes
6. state updated (this is invalid!)
```
Therefore the mount guard is indeed useful as authored

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
chance-wnb: It is recommended in our style guide: https://www.notion.so/wandbai/Frontend-Code-Best-Practices-d4995e2b47ba42d9abf01b134ad9c6b9?pvs=4#18be2f5c7ef38161b0dcd16aef052ada

My personal take is that Ecmascript 2017 natively supports async/await, so they are no longer always the syntax sugar for `.then()` I believe now they have an essential performance implications in modern version of ecmascript. 

Other reasons are purely style-related.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
chance-wnb: I see. Thanks for the explanation.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
chance-wnb: Yes, delegating the state mutation to parent is a great approach too. So parent needs to be responsible for the integrity of the data.

Judging by the name of `setCachedRowData and getCachedRowData`, you might want to do `useState`, that will always trigger re-rendering. My previous `useRef` approach is for avoiding unnecessary re-rendering. But if it is regarded as hacky, I am ok with aiming for readability.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
chance-wnb: Yes, that's true. We don't have adopt it if it doesn't make sense.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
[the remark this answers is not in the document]
chance-wnb: Thanks for the explanation. There is no issue here now :)

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
chance-wnb: I guess I never cared about this situation. The framework should automatically swallow such state update for an unmounted component, because it would almost never incur any unwanted side effect. 

Just curious, does it also trigger a warning even for production build?

chance-wnb (approved):
There is no outstanding non-style related issues in this PR. Approving now for efficiency!

Thanks a lot for helping out on the refactoring!

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
tssweeney: I went ahead and implemented the ore simple context approach

chance-wnb: Thanks. LGTM.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CompareEvaluationsPage/sections/ExampleCompareSection/exampleCompareSectionUtil.ts
```diff
@@ -316,189 +310,243 @@ export const useFilteredAggregateRows = (state: EvaluationComparisonState) => {
...
-    cachedRowData.current[row.digest] = row.val;
-  }
-}
-
-type UseExampleCompareDataParams = Parameters<typeof useExampleCompareData>;
-
-export function useExampleCompareData(
+/**
+ * The following functions are used to fetch data rows (and cache them in the state - Note: caching them in
+ * the state is a little hacky since it is not reactive, but it does work for our use case)
```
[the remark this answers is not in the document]
tssweeney: finished removing this hackiness

Issue #4487, opened 2025-05-14
Node SDK gives undecipherable errors when calling trace server


pjlsergeant:
When calls to the trace server fail, the user gets an unhelpful error:

```
node:internal/process/promises:392
      new UnhandledPromiseRejection(reason);
      ^

UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<_Response>".
    at throwUnhandledRejectionsMode (node:internal/process/promises:392:7)
    at processPromiseRejections (node:internal/process/promises:475:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:106:32) {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

This is because `swagger-typescript-api` ends up throwing the response error itself rather than an error object, and [src/weaveClient.ts]() doesn't wrap all of its calls to TraceServerApi in proper error handling; the response object itself gets bubbled up (which doesn't serialize well) and doesn't have a stack.

**This makes the Node SDK essentially unusable; if you trip this error and don't spend the several hours I just have combing through code trying to find the source, you have absolutely no idea why it's blowing up.**

PR #4488, opened 2025-05-14
extract meaningful errors when calls to the traceServerApi fail (#4487)

pjlsergeant:
# Description

Adds debugging messages of last-resort to HTTP Responses thrown as errors inside [src/weaveClient.ts](). This is obviously a band-aid patch: the better fix would be to lobby the `swagger-typescript-api` not to throw a Response object, and to also add proper error handling to each of the call sites in [src/weaveClient.ts](). HOWEVER, the Node SDK is essentially unusable without this in its current form (see: #4487).

# Testing

I ran the test suite, it passed.

### DIFF
--- a/sdks/node/src/weaveClient.ts
+++ b/sdks/node/src/weaveClient.ts
@@ -236,12 +236,12 @@ export class WeaveClient {
     limit: number = 1000
   ): AsyncIterableIterator<CallSchema> {
     const resp =
-      await this.traceServerApi.calls.callsQueryStreamCallsStreamQueryPost({
+      await debugResponse( () => this.traceServerApi.calls.callsQueryStreamCallsStreamQueryPost({
         project_id: this.projectId,
         filter,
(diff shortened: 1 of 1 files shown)


pjlsergeant:

chance-wnb:
hi @pjlsergeant 

Thanks for making the change. 

I think a more universal approach is to directly log an error in the `traceServerApi.ts`. It looks like `swagger-typescript-api` supports generating a template first, having the template manually modified, then generate the eventual ts file using the modified template. We should be able to modify the template to add logging. That seems to be the cleanest approach available.

Please let me know if you are willing to go this route. Otherwise, feel free to leave this to our team(Weave) to handle it. 

I know debugging the unhandled rejection is such a pain. Have you tried capturing unhandled rejections on the process level?

```
process.on('unhandledRejection',  callback);
```

would it be used as a temporary solution at the moment?

PR #4497, opened 2025-05-14
chore(weave): Refactor ChatView hooks into seperate files

zbirenbaum:
## Description

ChatView's hooks file is getting a bit too verbose and it's very difficult to track program flow. This is step one in a refactor to make chat schema handling more robust.

Questions raised for followups:
 - Do we really need the model parameter on the ChatCompletion?
 - Since we have decided not to display anything if either the input or output is invalid, do we even need to seperate normalization?

## Testing

It was mostly just moving stuff around and the code compiles so it should be fine.


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
```diff
@@ -359,6 +82,22 @@ export const getChatFormat = (call: CallSchema): ChatFormat => {
   return ChatFormat.None;
 };
 
+export const normalizeChatCompletion = (
```
zbirenbaum, line 85: The fact that we have to pass a request alongside the completion in order to normalize this indicates to me that either:
1. We should not have separate logic for handling these at all
2. We should omit the model from the completion

I think something like this makes sense:
```
type NormalizedChat {
  model: string, // Don't make this optional, just give it a placeholder like 'unknown' if we can't parse from anywhere
  request: ChatRequest,
  response: Omit<ChatResponse, 'model'>,
}
```

then just something like: 
```
const extractChatFromCall = (call: TraceCallSchema): NormalizedChat | null => {
  ...
}
 ```

File: weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/ChatFormats/anthropic.ts [hunk 2]
```diff
@@ -0,0 +1,117 @@
+import _ from 'lodash';
```
zbirenbaum, line 1: Next step for all of these `ChatFormats` files is to use zod schema validation to automatically parse out which provider handling we should do

zbirenbaum:
I'm undecided on whether to merge the changes here incrementally or do a full rewrite

I think just doing the full refactor and merging in one go makes sense but we need good tests for that.

neutralino1 (approved):
Thank you.

PR #4498, opened 2025-05-14
Draft: modularize and reuse SDK logic for worker logic

neutralino1:
https://github.com/wandb/weave/pull/4276 duplicates the scoring logic present in the SDK The main reason is that the SDK talks to the trace server API, while the worker talks to the database directly. Plus since the worker only needs to support built-in scorers (not arbitrary ops), the tracking and scoring logic can be much simplified.

This is not ideal as it duplicates logic, leading to multiple places to fix bugs. This PR attempts to modularize the SDK code to accept a different client than `WeaveClient`.


weave/trace/op.py
```diff
@@ -276,11 +280,35 @@ def _default_on_input_handler(func: Op, args: tuple, kwargs: dict) -> ProcessedI
...
+        output: Any,
+        exception: BaseException | None = None,
+        *_: Any,
+        **__: Any,
+    ) -> None: ...
+
+
 def _create_call(
-    func: Op, *args: Any, __weave: WeaveKwargs | None = None, **kwargs: Any
+    client: ClientInterface,
```
neutralino1, line 306: This is what allow the worker to pass its own client.

weave/trace/op.py
```diff
@@ -359,13 +387,14 @@ def _call_sync_func(
...
     __require_explicit_finish: bool = False,
+    client: ClientInterface | None = None,
     **kwargs: Any,
 ) -> tuple[Any, Call]:
     func = op.resolve_fn
     call = placeholder_call()
 
     # Handle all of the possible cases where we would skip tracing.
-    if is_tracing_setting_disabled() or should_skip_tracing_for_op(op):
+    if False:  # is_tracing_setting_disabled() or should_skip_tracing_for_op(op):
```
neutralino1, line 397: This is still an outstanding problem. Unsure how to solve this.
The global context has no `WeaveClient` so the code things that tracing is disabled. I cannot use `set_weave_client_global` as each incoming call has potentially a different `project_id` so the client needs to be set per-call. And since they are processed asynchronously, using `set_weave_client_global` leads to race conditions and thread-unsafety.

weave/workers/weave_scorer.py
```diff
@@ -264,56 +267,92 @@ async def process_monitor(
...
-            summary={},
+    def _async_call_op(
+        op: Op, *args: Any, **kwargs: Any
+    ) -> Coroutine[Any, Any, tuple[Any, Call]]:
+        return asyncio.to_thread(
+            lambda: _call_sync_func(op, *args, client=client, **kwargs)
         )
+
+    apply_scorer_result = await apply_scorer_async(
+        scorer, example, output, _async_call_op
```
neutralino1, line 349: This is where we reuse SDK logic instead of duplicating it. We need to customize `async_call_op` in order to inject our custom client as seen above.

weave/workers/weave_scorer.py
```diff
@@ -264,56 +267,92 @@ async def process_monitor(
...
+        self,
+        op: str | Op,
+        inputs: dict,
+        parent: Call | None = None,
+        attributes: dict | None = None,
+        display_name: str | Callable[[Call], str] | None = None,
+        *,
+        use_stack: bool = True,
+    ) -> Call:
+        assert isinstance(op, Op)  # necessary for type narrowing
```
tssweeney, line 285: Would be good to add message to help with debugging

PR #4499, opened 2025-05-15
fix(weave): only perform trace to chat conversion when underlying data changes

zbirenbaum:
## Description

- Fixes WB-25009

Currently normalizing the call trace into a Chat occurs on every keypress when the user types into the playground prompt box and multiple times every time the chat is rendered in playground and trace view.

With this change it is only re-parsed when the underlying data changes.

## Testing

tested in playground

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import React from 'react';
 
 import {isWeaveRef} from '../../filters/common';
 import {mapObject, traverse, TraverseContext} from '../CallPage/traverse';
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
```diff
@@ -561,30 +562,43 @@ export const useCallAsChat = (
 ): {
   loading: boolean;
 } & Chat => {
-  // Traverse the data and find all ref URIs.
-  const refs = getRefs(call);
+  // Memoize the call data processing to prevent unnecessary recalculations
+  // when the component re-renders but the call data hasn't changed
+  const refs = React.useMemo(
```
neutralino1, line 567: Can you import `useMemo` explicitly like we do in the rest of the app? Just for style consistency.

neutralino1 (changes requested):
One minor nit.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
zbirenbaum: Done

PR #4502, opened 2025-05-15
fix(weave): Dont crash in langchain if client is not init'd

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-25005

Previously, if `weave.init` was called in a submodule, patching would apply everywhere and could affect modules where `weave.init` was not called, causing a crash.

Now, patching still applies everywhere, but it simply no-ops if there is no client available.

### DIFF
--- a/weave/integrations/langchain/langchain.py
+++ b/weave/integrations/langchain/langchain.py
@@ -31,10 +31,12 @@ This approach allows for more flexible runtime configuration while still respect
 
 import datetime
 import json
+import logging
 from contextlib import contextmanager
 from contextvars import ContextVar
 from uuid import UUID
 
+from weave.flow.util import warn_once
 from weave.integrations.integration_utilities import (
     make_pythonic_function_name,
     truncate_op_name,
@@ -59,6 +61,8 @@ from collections.abc import Generator
 from typing import Any, Optional
 
 RUNNABLE_SEQUENCE_NAME = "RunnableSequence"
+
+logger = logging.getLogger(__name__)
 
 if not import_failed:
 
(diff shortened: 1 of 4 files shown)


weave/integrations/langchain/langchain.py
```diff
@@ -94,17 +98,31 @@ class WeaveTracer(BaseTracer):  # pyright: ignore[reportRedeclaration]
         run_inline: bool = True
 
         def __init__(self, **kwargs: Any) -> None:
+            self.gc = None
+            # self.gc = weave_client_context.require_weave_client()
```
neutralino1, line 102: 🔪

weave/integrations/openai_agents/openai_agents.py
```diff
@@ -71,6 +73,9 @@ def __init__(self) -> None:
 
     def on_trace_start(self, trace: tracing.Trace) -> None:
         """Called when a trace starts."""
+        if not (wc := get_weave_client()):
```
neutralino1, line 76:
```suggestion
        if (weave_client := get_weave_client()) is None:
```
More explicit naming and explicit `None` check. Same below.

neutralino1 (changes requested):
Thank you. A few nits.

weave/integrations/langchain/langchain.py
```diff
@@ -94,17 +98,31 @@ class WeaveTracer(BaseTracer):  # pyright: ignore[reportRedeclaration]
         run_inline: bool = True
 
         def __init__(self, **kwargs: Any) -> None:
+            self.gc = None
```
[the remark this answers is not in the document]
andrewtruong: It used to be called `GraphClient`, so `gc`.  I think we can just rename to `wc` -- it's descriptive and short, and the type hint will immediately tell you that it's the `WeaveClient`

weave/integrations/langchain/langchain.py
```diff
@@ -94,17 +98,30 @@ class WeaveTracer(BaseTracer):  # pyright: ignore[reportRedeclaration]
         run_inline: bool = True
 
         def __init__(self, **kwargs: Any) -> None:
+            self.wc = None
+            if gc := weave_client_context.get_weave_client():
+                self.wc = gc
```
neutralino1, line 103:
```suggestion
            if wc := weave_client_context.get_weave_client():
                self.wc = wc
```

neutralino1 (approved):
Missing one renaming 😉  Otherwise good to go.

PR #4503, opened 2025-05-15
chore(weave): Add top level get_client function

andrewtruong:
This is helpful for users who do:
```py
weave.init(project)
```

but then need to retrieve the client later (perhaps in a different context).  Now they can:
```py
client = weave.get_client()
```


neutralino1 (approved):
Thank you.

PR #4505, opened 2025-05-15
fix(ui): fix full screen babylon fails when media too far down on page

KyleGoyette:
## Description

- https://wandb.atlassian.net/browse/WB-24788

Moves the created fullscreen element off of the original deeply nested element and onto the main document body. This fixes the issue where when we would full screen, the original container element would go far enough off screen that it would vanish and full screen mode would exit.

Instead now, we have a brief reload time for all object3d tiles when exiting full screen, not ideal but as far as I know, chrome always scrolls to the top when full screen mode is requested.

Before:

https://github.com/user-attachments/assets/3ef357e4-67b9-47e2-a153-6085726bf714

After:

https://github.com/user-attachments/assets/3761e872-9463-42f5-902b-926fcd01b19f

## Testing

How was this PR tested?


ericakdiaz (approved):
i found https://issues.chromium.org/issues/40260339 from 2016, that claims it's fixed in general, i wonder why we only see it in chrome? maybe it's a weird interaction with the virtualization library?

KyleGoyette:
>i found https://issues.chromium.org/issues/40260339 from 2016, that claims it's fixed in general, i wonder why we only see it in chrome? maybe it's a weird interaction with the virtualization library?

Based on my reading, they fixed returning to the original scroll position, no?

PR #4507, opened 2025-05-15
feat(ui): Improve run links and filtering

jamie-rasmussen:
## Description

* When the user hovers over a value in the Run column we show a little popup summary.
* Option+Click to add a filter still works, but is less hacky because we can query for the project's internal ID.
* We don't have to show the confusing run name in the filter bar, we can show the display name and give it the same hover treatment as the cell value.
* Users can select Run as the field in the filter popup, and a new Select widget allows them to pick one.

A follow up PR in the core repo will link from the Run Overview to the filtered traces page using code in this PR.

I also copied the `RunState` component from the core repo so we could use it in Weave. Follow up PRs will point core uses at this location and then delete the core version to finish that migration.

Before:

Screenshot 2025-05-15 at 2 23 56 PM

After:

Hover over the run display name shows more information
Screenshot 2025-05-15 at 2 21 41 PM

Filter bar shows display name instead of name
Screenshot 2025-05-15 at 2 21 54 PM

Users can now easily switch between runs using the filter dialog
Screenshot 2025-05-15 at 2 22 03 PM

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
@@ -37,7 +37,6 @@ export const UNFILTERABLE_FIELDS = [
   'tokens',
   'cost',
   'wb_user_id', // Option+Click works
-  'wb_run_id', // Option+Click works
   'total_storage_size_bytes',
 ];
 
@@ -79,6 +78,7 @@ export const FIELD_TYPE: Record<string, string> = {
   id: 'id',
   'summary.weave.status': 'status',
   'summary.weave.trace_name': 'string',
+  wb_run_id: 'run',
   wb_user_id: 'user',
   started_at: 'datetime',
 };
@@ -289,6 +289,15 @@ export const getOperatorOptions = (field: string): SelectOperatorOption[] => {
       {
         value: '(string): in',
         label: 'in',
+        group: 'string',
+      },
+    ];
+  }
+  if ('run' === fieldType) {
+    return [
+      {
+        value: '(string): equals',
+        label: 'equals',
         group: 'string',
       },
     ];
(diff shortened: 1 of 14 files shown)


weave-js/src/components/RunState.tsx
```diff
@@ -0,0 +1,140 @@
+/**
```
jamie-rasmussen, line 1: Note that this file is brought over unchanged from the core repo so that we can use it in Weave.

weave-js/src/components/PagePanelComponents/Home/Browse3/filters/common.ts
```diff
@@ -293,6 +293,15 @@ export const getOperatorOptions = (field: string): SelectOperatorOption[] => {
       },
     ];
   }
+  if ('run' === fieldType) {
+    return [
+      {
+        value: '(string): equals',
+        label: 'equals',
```
neutralino1, line 300: Should we make it say "Run is super-silent-1" instead of "equals"?

weave-js/src/components/RunLink.tsx
```diff
@@ -0,0 +1,217 @@
...
+  runName: string
+) => {
+  const apolloClient = useApolloClient();
+
+  const [runData, setRunData] = useState<RunResult>('load');
+  useEffect(() => {
+    let mounted = true;
+    setRunData('loading');
+    apolloClient
+      .query({
```
neutralino1, line 150: More N+1 queries 😢 
Can we query runs in bulk once per page?

neutralino1 (commented):
Thank you. A couple of notes.

weave-js/src/components/RunLink.tsx
jamie-rasmussen: Agree it is 😢 but N+1 is how it already works, not something changed in this PR. Bulk querying runs information would be a more complicated change that I don't think we should block this improvement on. We do also have client-side caching so it's not quite as bad as it seems, especially since a common case will be only viewing the calls for the selected run.

neutralino1 (approved):
Thank you!

PR #4509, opened 2025-05-15
chore(weave): Adds "WEAVE_LOG_LEVEL" setting and consolidates terminal printing into common module

tssweeney:
This PR does the following:
1. Adds lint rules disallowing raw `print` calls in our user-facing codebase
2. Replaces all applicable `print`s with `logger.info` to standardize writes
3. Modifies the root logger (`weave`) such that it respects global log level setting, support prefixing.
4. Adds WEAVE_LOG_LEVEL setting


neutralino1 (commented):
Is there a way to do this using the native `logging` API?

For example, we could define a centralized logger that has all the right formatting set up. Then all call sites would import the logger from the central location and call it using the native `logging` API.

neutralino1 (commented):
We could set the default logging level based on the value of `should_be_silent()`. For example if `True`, default log level is `WARNING`, if not it's `INFO`. Then call sites call `logger.info` (where `logger` is imported from a central module) and it will get printed or not based on the setting.

My greater point is that I don't like wrapping basic APIs, instead of configuring said APIs and letting call sites use the native logger API. This means no custom magical `weave_print`, just standard APIs.

tssweeney:
> We could set the default logging level based on the value of `should_be_silent()`. For example if `True`, default log level is `WARNING`, if not it's `INFO`. Then call sites call `logger.info` (where `logger` is imported from a central module) and it will get printed or not based on the setting.
> 
> 
> 
> My greater point is that I don't like wrapping basic APIs, instead of configuring said APIs and letting call sites use the native logger API. This means no custom magical `weave_print`, just standard APIs.

Yeah I think you are right. I'll rework it

neutralino1 (approved):
Beautiful 👌

PR #4514, opened 2025-05-16
feat(weave): TS SDK: allows call handle to be returned

chance-wnb:
## Description

Adds a new `invoke` method to ops that returns both the result and a Call object, allowing users to update call metadata after execution. The Call object provides a `setDisplayName` method to change the display name of a call even after it has completed.

## Testing

Tested by creating ops, calling them with the new `invoke` method, and verifying that the returned Call object can be used to update the display name both during and after execution.

Some unit tests scenarios are authored as well.

### DIFF
--- a/sdks/node/src/opType.ts
+++ b/sdks/node/src/opType.ts
@@ -11,7 +11,7 @@ export type Op<T extends (...args: any[]) => any> = {
   __name: string;
   __savedRef?: OpRef | Promise<OpRef>;
   __parameterNames?: ParameterNamesOption;
-  call: CallMethod<T>;
+  invoke: CallMethod<T>;
 } & T &
   ((
     ...args: Parameters<T>
--- a/sdks/node/src/opType.ts
+++ b/sdks/node/src/opType.ts
@@ -1,3 +1,4 @@
+import {Call} from './call';
 import {getGlobalDomain} from './urls';
 import {WeaveObject} from './weaveObject';
 
@@ -10,6 +11,7 @@ export type Op<T extends (...args: any[]) => any> = {
   __name: string;
   __savedRef?: OpRef | Promise<OpRef>;
   __parameterNames?: ParameterNamesOption;
+  call: CallMethod<T>;
 } & T &
   ((
     ...args: Parameters<T>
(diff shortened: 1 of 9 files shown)


chance-wnb:
* **#4514** <a href="https://app.graphite.dev/github/pr/wandb/weave/4514?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#4515](https://github.com/wandb/weave/pull/4515) <a href="https://app.graphite.dev/github/pr/wandb/weave/4515?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4516](https://github.com/wandb/weave/pull/4516) <a href="https://app.graphite.dev/github/pr/wandb/weave/4516?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/opType.ts
```diff
@@ -10,6 +11,7 @@ export type Op<T extends (...args: any[]) => any> = {
   __name: string;
   __savedRef?: OpRef | Promise<OpRef>;
   __parameterNames?: ParameterNamesOption;
+  invoke: CallMethod<T>;
```
chance-wnb, line 14: The reason why I chose `invoke` instead of `call` like the python side does is because:

JavaScript's functions natively have a [`call()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) method. Its first parameter is `this` which helps you to redirect the `this` context of the function. 

Our op.call() signature doesn't enforce user to put (this) as the first parameter (it is more convenient if user doesn't have to always specify it). So this `op.call()` method is fundamentally and functionally different than JS's native `call` method.

Plus, I have a hard time making the TS typing right. TS always assume `call` comes from a function object, and must have certain shape. Despite that I have overwritten it in the runtime by completely replacing the implementation of `call`, it still doesn't want to conform to the new typing I assigned to it. 

Lastly, I feel that user might have some use for the original `call()` method. It might not be a good idea to overwrite it in a irreversible fashion. So I decide to keep it and use a new `invoke()` method to implement what should be for `op.call()`.

Despite that runtime-wise everything works. The typing part (compile time) is far from ideal. I am still researching it, the better result might land in the next PR.

neutralino1: I agree, let's not override native APIs.

neutralino1 (approved):
Thank you!

PR #4515, opened 2025-05-16
limit the display name size

chance-wnb:
## Description

- Limits display names to 1000 characters to prevent excessively long names
- Adds elision logic to truncate long display names and append "..." when they exceed the limit
- Adds a console warning when display names are truncated

## Testing

Tested by verifying that display names longer than 1000 characters are properly truncated with an ellipsis and that a warning is logged to the console.

PR #4516, opened 2025-05-16
Unit tests for the call handle feature

chance-wnb:
## Description

- Adds a new `invoke` method to the `Op` type that returns both the result and the call object
- Replaces the existing `call` method with `invoke` in the op wrapper
- Adds tests for retrieving call objects from standalone functions and class methods
- Adds tests for call manipulation (setting display name)
- Simplifies mock implementations in tests to use rest parameters

## Testing

Tested with new unit tests that verify:
- Getting call objects from standalone function calls
- Getting call objects from class method calls
- Using call handles to manipulate calls (e.g., setting display names)

### DIFF
--- a/sdks/node/src/opType.ts
+++ b/sdks/node/src/opType.ts
@@ -11,7 +11,7 @@ export type Op<T extends (...args: any[]) => any> = {
   __name: string;
   __savedRef?: OpRef | Promise<OpRef>;
   __parameterNames?: ParameterNamesOption;
-  call: CallMethod<T>;
+  invoke: CallMethod<T>;
 } & T &
   ((
     ...args: Parameters<T>
--- a/sdks/node/src/__tests__/call.manipulation.test.ts
+++ b/sdks/node/src/__tests__/call.manipulation.test.ts
@@ -1,0 +1,54 @@
+import * as weave from 'weave';
+import {op, Op} from 'weave';
+import {CallState, InternalCall} from 'weave/call';
+import {getGlobalClient} from 'weave/clientApi';
+
+const mockCallId = 'test-call-id';
+const mockProjectId = 'test-project-id';
+
+const mockUpdateCall = jest.fn();
+
+jest.mock('weave/clientApi', () => ({
+  getGlobalClient: jest.fn(() => {
+    const weaveClient = new weave.WeaveClient(
+      null as any,
+      null as any,
+      mockProjectId
+    );
+
+    Object.assign(weaveClient, {
+      getCall: async (callId: string) => {
+        const internalCall = new InternalCall();
+        internalCall.updateWithCallSchemaData({
+          id: callId,
+        });
+
+        internalCall.state = CallState.finished;
+
+        return internalCall.proxy;
+      },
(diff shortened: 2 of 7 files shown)


chance-wnb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/wandb/weave/4516?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4516** <a href="https://app.graphite.dev/github/pr/wandb/weave/4516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4516?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4514** <a href="https://app.graphite.dev/github/pr/wandb/weave/4514?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#4515](https://github.com/wandb/weave/pull/4515) <a href="https://app.graphite.dev/github/pr/wandb/weave/4515?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

sdks/node/src/opType.ts
```diff
@@ -11,7 +11,7 @@ export type Op<T extends (...args: any[]) => any> = {
   __name: string;
   __savedRef?: OpRef | Promise<OpRef>;
   __parameterNames?: ParameterNamesOption;
-  call: CallMethod<T>;
+  invoke: CallMethod<T>;
```
neutralino1, line 14: I think it's called `call` on the Python side. Why rename?

neutralino1 (commented):
One question.

sdks/node/src/opType.ts
chance-wnb: This is a great question. 

JavaScript's functions natively have a [`call()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) method. Its first parameter is `this` which helps you to redirect the `this` context of the function. 

Our op.call() signature doesn't enforce user to put (this) as the first parameter (it is more convenient if user doesn't have to always specify it). So this `op.call()` method is fundamentally and functionally different than JS's native `call` method.

Plus, I have a hard time making the TS typing right. TS always assume `call` comes from a function object, and must have certain shape. Despite that I have overwritten it in the runtime by completely replacing the implementation of `call`, it still doesn't want to conform to the new typing I assigned to it. 

Lastly, I feel that user might have some use for the original `call()` method. It might not be a good idea to overwrite it in a irreversible fashion. So I decide to keep it and use a new `invoke()` method to implement what should be for `op.call()`.

Despite that runtime-wise everything works. The typing part (compile time) is far from ideal. I am still researching it, the better result might land in the next PR.

chance-wnb:
The PR is closed in favor of https://github.com/wandb/weave/pull/4514 (I merged the two, otherwise the base PR will fail some unit tests)

PR #4526, opened 2025-05-19
fix(weave): Fixes inf loop when expanding inputs & outputs

tssweeney:
Specifically, there was a bug when:
1. User expanded first level of input/output ref
2. User expands the first child ref of that recently expanded parent ref.

Low level: fixes `useRefsReadBatch` to not return stale data.

---

Some technical details... the pattern that was happening was:
1. user expands Ref X
2. useRefsReadBatch is called with [X]
3. user expands Y
4. useRefsReadBatch is called with [Y] -> returns immediately with stale X data
5. React component puts X's data inside Y, creating an inf recursion!

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
@@ -1347,8 +1347,14 @@ const useObjectDeleteFunc = () => {
 
 const useRefsReadBatch = (params: UseRefsReadBatchParams) => {
   const getTsClient = useGetTraceServerClientContext();
-  const [refsRes, setRefsRes] =
-    useState<traceServerTypes.TraceRefsReadBatchRes | null>(null);
+  // Here we keep track of the result and the uris we fetched for. This is
+  // because repeated calls to this hook with different uris will return
+  // could result in stale results if we just returned the result from the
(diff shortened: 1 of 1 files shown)


adrnswanberg (approved):

neutralino1 (commented):
Any chance of memory leak here?
If the user leaves the tab open for a long time and keeps expanding many refs, we never clear the state.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -1347,8 +1347,14 @@ const useObjectDeleteFunc = () => {
 
 const useRefsReadBatch = (params: UseRefsReadBatchParams) => {
   const getTsClient = useGetTraceServerClientContext();
-  const [refsRes, setRefsRes] =
-    useState<traceServerTypes.TraceRefsReadBatchRes | null>(null);
+  // Here we keep track of the result and the uris we fetched for. This is
+  // because repeated calls to this hook with different uris will return
```
zbirenbaum, line 1351: Remove extra `will return`

Might be helpful to have a comment detailing what `forUris` contains as well

zbirenbaum (approved):
One nit on existing comment and optional suggestion for another comment. Otherwise looks good, thank you!

neutralino1 (approved):

PR #4534, opened 2025-05-20
fix(weave): User sentry not respected depending on import order

zbirenbaum:
## Description

- Fixes WB-25064

1. Added dual Sentry reporting to both user and Weave hubs
2. Added tests for the Sentry functionality
3. Added function to check if user has already configured Sentry
4. Modified initialization to respect user's Sentry configuration

## Testing

Added integration tests with mocked endpoints to verify sentry works properly

### DIFF
--- a/weave/trace/trace_sentry.py
+++ b/weave/trace/trace_sentry.py
@@ -32,6 +32,15 @@ SENTRY_DEFAULT_DSN = "https://99697cf8ca5158250d3dd6cb23cca9b0@o151352.ingest.us
 SessionStatus = Literal["ok", "exited", "crashed", "abnormal"]
 
 
+# Function to check if Sentry SDK has already been initialized by the user
+def _is_sentry_configured() -> bool:
+    """Check if the Sentry SDK has already been configured with a client."""
+    # Check if the current hub has a client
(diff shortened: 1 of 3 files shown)


weave/trace/trace_sentry.py
```diff
@@ -245,6 +340,23 @@ def _is_local_dev_install(module: Any) -> bool:
...
+    when this module is imported. This ensures that if the user has already
+    initialized Sentry, we'll properly respect that configuration while
+    also reporting to Weave's Sentry endpoint.
+    """
+    global_trace_sentry.setup()
+    global_trace_sentry.configure_scope()
+
+
+# For backward compatibility, initialize if not already using a user-configured hub
+if not _is_sentry_configured():
```
tssweeney, line 361: why would we only init our own if user doesn't have one?

weave/trace/trace_sentry.py
```diff
@@ -58,7 +67,9 @@ def __init__(self) -> None:
 
         self.dsn = SENTRY_DEFAULT_DSN
 
-        self.hub: sentry_sdk.hub.Hub | None = None
+        self.hub: sentry_sdk.Hub | None = None
```
tssweeney, line 70: this is really weird to me that we need to manage both - can you explain the approach and reasoning behind this PR in the description?

PR #4535, opened 2025-05-20
fix(weave): run trace server bindings tests

zbirenbaum:
## Description

- Fixes WB-24830

1 . Adds trace_server_bindings to nox shard list so tests run again
2. Some changes to oversized item handling so that the item that is too large is not attempted
3. In the tests:
 - Fixed the assertion in test_requeue_after_max_retries to match the actual log message.
 - Modified test_post_timeout to check for warning logs instead of error logs.
 - @pytest.mark.disable_logging_error_check decorator for test_oversized_item_will_error_without_sending to prevent it from failing due to the log message.
 - Reset the mock properly in test_post_timeout and used a new server instance.

## Testing

runs in CI and tests pass now


neutralino1 (approved):
Thank you.

PR #4536, opened 2025-05-20
fix(weave): Hide "python" function type in the UI and move to weave-managed namespace

tssweeney:
Moves function type bookkeeping to the weave namespace and hides in the UI


neutralino1 (approved):
Thank you.

PR #4538, opened 2025-05-20
fix(weave): Evaluation progress updates should not block on long-running rows

tssweeney:
As currently implemented, the Evaluation.evaluate routine will print "Evaluated X of Y" results once index X is complete - which means long-running X's will block the printing and result in bursty printing. However, the processing itself is done eagerly - creating a misperception of slowness and lack of parallelism. This fixes that, printing `X` once `X` are completed in any order (matching the true processing).

This is implemented by updating the contract of `async_foreach` to yield results out-of-order, but include an index for the caller to use if desired.


neutralino1 (approved):
Thank you.

PR #4540, opened 2025-05-20
chore(weave): add icons missing from wandb/ui

onx2:
## Description

Adds a few missing (reports-related) icons needed for the transition off of `@wandb/ui`


jamie-rasmussen (changes requested):
Please run the `addicon` tool from devtools, it will run checks (e.g. your icons aren't at the standard 24x24) and generate the code changes for you.
https://github.com/wandb/devtools/blob/main/docs/addicon.md

jamie-rasmussen:
It looks like these still have `viewBox="0 0 32 32"` instead of `viewBox="0 0 24 24"` - for consistency with the rest of the icon set you may want to check with the design team about how they are exporting them.

onx2:
> It looks like these still have `viewBox="0 0 32 32"` instead of `viewBox="0 0 24 24"` - for consistency with the rest of the icon set you may want to check with the design team about how they are exporting them.

These are super old icons that are used in reports so there aren't [new versions](https://www.figma.com/design/01KWBdMZg5QM9SRS1pQq0z/Design-System----Robot-Styles?node-id=3-510&t=DY3hOEoAjFifwk73-0) 😭 the viewbox is just the relative sizing for the coordinates when it is drawn so as along as the width & height are correct it should be fine _(aside from it not being consistent in the codebase)_. B/c of that and to get off of older libraries I'd like to push on this going in. I can circle back with design later to get official support for reports icons (I think we are missing a few as well in _all_ libraries).

jamie-rasmussen:
Yeah, it's just that addicon will now complain about them every time it's run but I can live with it.

jamie-rasmussen (approved):

onx2:
> Yeah, it's just that addicon will now complain about them every time it's run but I can live with it.

I messaged Lauralee to see if this is something we can easily fix 😄

PR #4543, opened 2025-05-21
fix(weave): Speed up EvaluationLogger

tssweeney:
The current EvaluationLogger suffers from a number of problems:
1. The `predict_and_score` virtualized call accidentally blocks on uploading the Scorer
2. Each time you log a score, a new scorer is minted, meaning it uploads a new one every time!

This PR implements the following improvements:
1. Modifies the score sending logic to be non-blocking on the main process and backgrounded. This will only block therefore if the user process is about to quit and we need to upload the data.
2. Maintains an id-based LRU cache of scorers so we don't mint new ones!


adrnswanberg (approved):
🔥

PR #4544, opened 2025-05-21
chore(weave): Re-organize eval compare page closer to long-term design

tssweeney:
This PR basically just splits the eval compare page into 2 tabs and makes a bunch of minor style adjustments to match the aesthetics of the updated ux from design team


adrnswanberg (approved):
My nit on the design is that the tabs are not obvious / kind of easy to miss. Not sure if there's an easy way to highlight them more.

PR #4549, opened 2025-05-21
feat(weave): Fix call attribute mutability and allow summary edits

tssweeney:
This PR was generated by Code, task id: `task_b_682e2653a0448331b5eefeabfae86200`

Original Instructions:

Weave allows users to get the current call using `get_current_call` in `weave/trace/context/call_context.py`. We notice that users have been attempting to set `attributes` on this call object, but report this as a bug, since that data is not persisted to the database. In fact, modifying `attributes` after the call is created should be disallowed! Instead, users should be able to set values on a `summary` payload, which is merged with additional calculated summary values when finishing the call in `finish_call` inside of `weave/trace/weave_client.py`.

Tasks:
1. Disallow edits to the `attributes` payload after the call has started
2. Allow users to edit `summary` and ensure that those edits are successfully included in the eventual data write.

Make sure you author tests to exercise these cases.

--- Codex Generated Summary:

## Summary
- make `AttributesDict` immutable once a call starts
- default `Call.summary` to an empty dict
- merge user defined summary with computed values on finish
- keep attributes read-only when loaded from server
- add regression tests for current call behaviour

## Testing
- `ruff check tests/trace/test_current_call.py weave/trace/weave_client.py`
- `ruff format weave/trace/weave_client.py tests/trace/test_current_call.py`
- `python -m pip list | head -n 20` *(fails: `pytest` not installed)*

--- Copy of Internal Codex Summary:

## Notes

* Tests could not be executed because pytest is not installed in the environment.

## Summary

* Introduced a freeze mechanism in AttributesDict to prevent modification of call attributes after a call is created
* Defaulted Call.summary to an empty dictionary for easier runtime updates
* Ensured attributes are frozen when a call is created and when calls are loaded from the server
* Merged user-provided summaries with computed values during finish_call
* Added tests validating attribute immutability and summary mutability via get_current_call


chance-wnb (approved):
Looks great. Thank you!

PR #4554, opened 2025-05-22
feat(weave): support auto-intecept commonjs modules

chance-wnb:
## Description

This PR adds automatic instrumentation for the OpenAI SDK in Node.js. It implements a module loader system that can intercept and patch module imports, allowing Weave to automatically wrap the OpenAI client with our existing functionality without requiring explicit user configuration.

Key changes:
- Implements a CommonJS module loader that intercepts require() calls
- Creates an instrumentation system to register hooks for specific modules
- Adds OpenAI instrumentation that automatically wraps the OpenAI client
- Included some key dependencies that helps implements the module loading hooks

## Some background to understand this PR better:
(You can also read a more detailed version in this [design doc](https://www.notion.so/wandbai/Weave-TypeScrip-SDK-Competitive-Analysis-1eee2f5c7ef38077ac3fe055a6c8d5ba?pvs=4#1fbe2f5c7ef380889219fef69ad285f1) )

*Module Interception in Node.js: A Deep Dive*

#### CommonJS

- **What it is:**  
  CommonJS is the traditional module system used in Node.js. It uses `require()` to load modules synchronously. When you call `require('module')`, Node.js resolves the module’s file path, executes the file, and returns its `module.exports`.

- **How it works:**  
  Node.js maintains a cache (`require.cache`) to avoid re-executing the same module multiple times. The resolution process involves checking the cache, resolving the file path (using `Module._resolveFilename`), and then executing the file in a new module context.

#### ESM (ECMAScript Modules)

- **What it is:**  
  ESM is the newer, standardized module system in JavaScript. It uses `import` and `export` statements and is designed to be asynchronous.

- **How it works:**  
  ESM modules are loaded asynchronously. The resolution process is handled by the loader, which can be customized using loader hooks. Unlike CommonJS, ESM does not have a built-in synchronous interception mechanism.

#### CommonJS Interception

- **Approach:**  
  In CommonJS, you can intercept module loading by patching the `require` function. This is done by replacing `Module.prototype.require` with a custom function that wraps the original `require` call.

- **How it works:**  
  The patched `require` function can inspect the module’s file path, check if an instrumentation hook is registered, and apply the hook to the module’s exports before returning them.

#### ESM Interception

- **Approach:**  
  ESM interception is more complex because it is asynchronous. However, it can be achieved using libraries like `import-in-the-middle`.

- **How it works:**  
  The `import-in-the-middle` library allows you to intercept and transform ESM module loading. It provides hooks that can be used to modify the module’s exports or behavior before they are used in the application.

#### Overview

The current code change introduces a mechanism to intercept CommonJS module loading, specifically for the `openai` module. It uses a patched `require` function to apply instrumentation hooks to the module’s exports.

#### Key Components

1. **`commonJSLoader.ts`:**  
   - This file defines a patched `require` function that intercepts module loading.
   - It checks if an instrumentation hook is registered for the module and applies the hook to the module’s exports.

2. **`instrumentations.ts`:**  
   - This file maintains a global map of registered instrumentation hooks, keyed by the module name and sub-path.
   - It provides a function `addInstrumentation` to register new hooks.

3. **`openai.ts`:**  
   - This file registers an instrumentation hook for the `openai` module.
   - The hook wraps the `OpenAI` class to add logging and other functionality.

#### How It Works

- **Patching `require`:**  
  The `commonJSLoader.ts` file patches `Module.prototype.require` to intercept module loading. When a module is required, it checks if an instrumentation hook is registered for that module.

- **Applying Hooks:**  
  If a hook is registered, it is applied to the module’s exports. This allows for dynamic modification of the module’s behavior, such as logging or wrapping methods.

- **Caching and Pass-Through:**  
  The patched `require` function also caches the patched exports to avoid re-applying hooks and handles pass-through for modules that do not have registered hooks.

#### Example

In the test file `openaiRequire.fork.test.ts`, a fake `openai` module is created, and an instrumentation hook is registered to log the parameters and results of the `chat.completions.create` method. The test verifies that the hook is called and that the expected output is logged.

## Testing

Tested by verifying that OpenAI client instances are automatically wrapped with Weave's functionality when imported, without requiring explicit calls to `wrapOpenAI()`. Verified that all existing OpenAI integration tests continue to pass with the new automatic instrumentation.

Unit test TBD


chance-wnb:
* **#4554** <a href="https://app.graphite.dev/github/pr/wandb/weave/4554?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4554?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/__tests__/integrations/openaiRequire.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    `;
+    fs.writeFileSync(path.join(modDir, 'index.js'), file);
+  });
+
+  afterEach(() => {
+    reset();
+    fs.rmSync(tmpDir, {recursive: true, force: true});
+  });
+
+  test('patchedRequire is called in child process', done => {
```
chance-wnb, line 36: It is pretty hard to isolate the test scenarios from the enclosing test runtime. So I launch a fake project and a fake openai library in a child process.

sdks/node/src/integrations/openai.ts
```diff
@@ -236,3 +237,27 @@ export function wrapOpenAI<T extends OpenAIAPI>(openai: T): T {
     },
   });
 }
+
+export function instrumentOpenAI() {
+  addInstrumentation({
+    moduleName: 'openai',
+    subPath: 'index.js',
+    version: '>= 4.0.0',
```
neutralino1, line 245: Maybe a comment to explain why this version constraint?

neutralino1 (approved):
This is pretty cool, congrats.

PR #4555, opened 2025-05-22
fix(weave): exclude null ref from object serialization

tssweeney:
`ref:null` was getting logged all over the place in our app. This PR fixes that for objects in a safe way moving forward.

## Summary
- don't serialize `ref` field on `ObjectRecord`
- check that serialization drops null refs
- ensure model publish/query excludes ref field

## Testing
- `nox --no-install -e lint`
- `.nox/tests-3-12-shard-trace/bin/pytest tests/trace/test_serialize.py::test_to_json_object_excludes_ref tests/trace/test_objs_query.py::test_publish_model_query_no_ref tests/trace/test_op_decorator_behaviour.py::test_sync_method_call tests/trace/test_op_decorator_behaviour.py::test_async_method_call`


neutralino1:
Should this PR get closed?

gtarpenning (approved):
There is one place we explicitly attach a ref, when we expand on the backend, but those should be valid and relatively useful.

PR #4559, opened 2025-05-22
fix(app): fixes icon shrinking problem

onx2:
## Description

Prevents icons from shrinking in flex containers. See https://wandb.ai/authorize for a reference to this problem.

image

image

### DIFF
--- a/weave-js/src/components/Icon/Icon.tsx
+++ b/weave-js/src/components/Icon/Icon.tsx
@@ -290,8 +290,11 @@ const updateIconProps = (props: SVGIconProps) => {
   return {
     width: 20,
     height: 20,
-    flexShrink: 0,
     ...props,
+    style: {
+      ...props.style,
+      flexShrink: 0,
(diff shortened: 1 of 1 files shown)


weave-js/src/components/Icon/Icon.tsx
```diff
@@ -291,6 +291,10 @@ const updateIconProps = (props: SVGIconProps) => {
     width: 20,
     height: 20,
     ...props,
+    style: {
+      ...props.style,
```
jamie-rasmussen, line 295: I don't really think it will come up, but do you think props.style should take precedence?

jamie-rasmussen (approved):

weave-js/src/components/Icon/Icon.tsx
onx2: hmm fair point. that would align with the current expectations for weave components so i'll update to make it an overrideable default

PR #4563, opened 2025-05-22
fix(weave): Streaming mode is broken for mistral endpoints

chance-wnb:
## Description

Fix [WB-25044](https://wandb.atlassian.net/browse/WB-25044)

- Fixes issue with stream_options being set for non-OpenAI endpoints
### What problem does it fix?

`stream_options` currently only happen for Completion series API, because that's the only occasion we would attach {stream: true}

For non-OpenAI endpoints, eg mistral, the `stream_options` config is not supported and will be rejected. 

So we can't just add `stream_options` indiscriminately.

This PR modifies the `_add_stream_options` function to only set `stream_options` when the API endpoint is specifically targeting OpenAI's official API (`api.openai.com`). It checks the hostname from the client's base URL before applying the stream options, preventing these options from being incorrectly applied to other compatible APIs.

## Testing

added tests in [openai_sdk.py](https://github.com/wandb/weave/pull/4563/files#diff-4ca40709ac141a5367973a7d99334b63f415ebaf67f8af4cc096a41080b667d2)

[WB-25044]: https://wandb.atlassian.net/browse/WB-25044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/integrations/openai/openai_sdk.py
+++ b/weave/integrations/openai/openai_sdk.py
@@ -4,6 +4,7 @@ import importlib
 import logging
 from functools import wraps
 from typing import TYPE_CHECKING, Any, Callable
+from urllib.parse import urlparse
 
 import weave
 from weave.integrations.patcher import MultiPatcher, NoOpPatcher, SymbolPatcher
@@ -329,7 +330,11 @@ def create_wrapper_sync(settings: OpSettings) -> Callable[[Callable], Callable]:
             @wraps(fn)
             def _wrapper(*args: Any, **kwargs: Any) -> Any:
                 if kwargs.get("stream") and kwargs.get("stream_options") is None:
-                    kwargs["stream_options"] = {"include_usage": True}
+                    completion = args[0]
+                    base_url = str(completion._client._base_url)
+                    # Only set stream_options if it targets the OpenAI endpoints
+                    if urlparse(base_url).hostname == "api.openai.com":
+                        kwargs["stream_options"] = {"include_usage": True}
                 return fn(*args, **kwargs)
 
             return _wrapper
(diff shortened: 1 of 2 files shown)


chance-wnb:
* **#4563** <a href="https://app.graphite.dev/github/pr/wandb/weave/4563?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4563?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/integrations/openai/openai_sdk.py
```diff
@@ -329,7 +330,11 @@ def _add_stream_options(fn: Callable) -> Callable:
             @wraps(fn)
             def _wrapper(*args: Any, **kwargs: Any) -> Any:
                 if kwargs.get("stream") and kwargs.get("stream_options") is None:
-                    kwargs["stream_options"] = {"include_usage": True}
+                    completion = args[0]
```
neutralino1, line 333: Are we guaranteed `args` has an item `0`?

neutralino1 (commented):
Does setting `stream_options` make other queries fail? What does this PR fix?

weave/integrations/openai/openai_sdk.py
chance-wnb: Yes, the args[0] is `self`. In the patched scenarios, it is always a class's method. so there is always `self`

weave/integrations/openai/openai_sdk.py
```diff
@@ -329,7 +330,11 @@ def _add_stream_options(fn: Callable) -> Callable:
...
+                    base_url = str(completion._client._base_url)
+                    # Only set stream_options if it targets the OpenAI endpoints
+                    if urlparse(base_url).hostname == "api.openai.com":
```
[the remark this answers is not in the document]
chance-wnb: That's a great question. I didn't think of that. 

I actually copied this logic from LiteLLM, that's how they do it.
https://github.com/BerriAI/litellm/blob/2c90ca0189557ec0054c5204e6c86a2aeda4ecc8/litellm/llms/openai/openai.py#L1043

Since the OpenAI and mistral endpoints are so close, there might be no other options to rely on. This might be the best one can do.

weave/integrations/openai/openai_sdk.py
chance-wnb: Great sugestion!

PR #4577, opened 2025-05-24
chore(ui): Eval compare design enhancements

m-rgba:
## Description

Latest updated Featurebee:
https://beta.wandb.ai/?betaVersion=d45e5e0da0d2ba389f06b7bdc4dcd4b830e5a0db

| List view | Detail view | Detail + list |
| -------- | ------- | ------- |
| image | image | image |

### User story
User story we're designing for:
- User comes to view the compare / expanded table.
- They review lines for the ones which are important to them.
- They did into those lines for further review.

Changes made:
- Start off at the table view to allow users to glance to review the example line-items.
- Started with the rows in a single line state.
- When the user needs to dig in for more details, they can:
  - Expand rows to get more lines to review.
    - Tweaked formatting and line-clamps.
  - Open the example details popover.
    - Made nearly full-screen - since reviewing needs more space, it's very busy.
  - Open the call peek-drawer.
    - Made call links available from example list.

---

### Larger updates
- Added call ID pills to the table.
- Updated the example detail card to be nearly full-width (needs more space for reviewing in that format).
  - Updated example detail card to make it resizeable.
- Added logic to make the example detail card full width when the calls peek-drawer is open when they're open at the same time (and then revert when the peek-drawer is closed).
- Adjusted row-size logic to add line-clamps (adding space adds lines and then we truncate as needed).
- Added the ability to Show/Hide (toggle) evals from header area. 
- Added the ability to show and hide columns (/ sets of columns).

### Smaller updates
- Using filled-circle indicator instead of model icon with color.
  - Does add some visual noise, but we'd take that trade-off to allow the user to visually understand which eval is mapped to which column quickly (interpreting the change in icon color is more difficult). 
Make hit points better for opening calls / example detail view
- Loads with example detail pane being closed.
- Loads with a single-row row-size per dataset example.
- Moved header row to footer area.
  - Space saving until we add filtering section above.
  - Will probably shift some of that back up when we have a filter / column adjustment section.
  - Hid "Open Details" button for simplicity / allowing user to select from rows.
- Adjusted header heights (more compact).
- Increased scorer column size (probably need to normalize the change percentages and values).
- Added trials tooltips for expand/contract icons.
- Updated tooltips to use design system (vs. MUI).
- Small design tweaks
  - Font sizes on summary page
  - Table footer pagination (size / font / colors)
  - Detail view header design tweaks (size, gap, color)
  - Double border table header
  - Pivot icon

### ToDo
ToDos:
- [ ] Use better popover display (from dataset cell renderer)
- [ ] Use small medium large sizing for height (similar to models)
- [ ] Add some nice stuff like search for column, cleanup "Calls" showing multiple times
- [ ] Make resize design nicer

Design feedback:
- [ ] Chart numbers - adjust sizes
- [ ] Try removing selected evals when in table view
- [ ] Add animations to detail drawer
- [ ] More aggressive drop-shadow detail drawer
- [ ] Full-screen drawer
- [ ] Close button tooltip remove

Additional feedback:
- [ ] Show hide all buttons for columns
- [ ] Multi-sort call-out
- [ ] Ctrl+click count on rows

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse2/CellValueString.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse2/CellValueString.tsx
@@ -42,7 +42,6 @@ type CellValueStringProps = {
 const Collapsed = styled.div`
   overflow: hidden;
   text-overflow: ellipsis;
-  white-space: nowrap;
   cursor: pointer;
 `;
 Collapsed.displayName = 'S.Collapsed';
@@ -87,7 +86,7 @@ Spacer.displayName = 'S.Spacer';
 const CellValueStringWithPopup = ({value, style}: CellValueStringProps) => {
   const ref = useRef<HTMLDivElement>(null);
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
-  const onClick = (event: React.MouseEvent<HTMLElement>) => {
+  const onClick = (_event: React.MouseEvent<HTMLElement>) => {
     setAnchorEl(anchorEl ? null : ref.current);
   };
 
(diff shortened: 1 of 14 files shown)


m-rgba:
Updated Featurebee:
https://beta.wandb.ai/?betaVersion=a99ee612e8ae66158f15fe7149ffadc2a4db3e1a

PR #4579, opened 2025-05-25
fix(weave): Don't cause EvaluationLogger instances to stomp on each other when logging in parallel

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-25510

This PR prevents instances from stomping on each other when logging in parallel by explicitly not using the global stack

This snippet will generate the following topology:
```py
def log_evals():
    ev = weave.EvaluationLogger()
    for x in range(3):
        pred = ev.log_prediction(inputs={"a": x, "b": x+1}, output=x+2)
        pred.log_score(scorer="scorer1", score=x+2)
        pred.log_score(scorer="scorer2", score=x+3)

for x in range(3):
    log_evals()
```

Before:
image

After:
image

### DIFF
--- a/weave/flow/eval_imperative.py
+++ b/weave/flow/eval_imperative.py
@@ -463,9 +463,9 @@ class EvaluationLogger(BaseModel):
 
         self._cleanup_predictions()
 
-        assert self._evaluate_call is not None, (
-            "Evaluation call should exist for finalization"
-        )
+        assert (
+            self._evaluate_call is not None
+        ), "Evaluation call should exist for finalization"
 
         # Finish the evaluation call
         wc = require_weave_client()
@@ -547,9 +547,9 @@ class EvaluationLogger(BaseModel):
             final_summary = {**final_summary, **summary}
 
         # Call the summarize op
-        assert self._evaluate_call is not None, (
-            "Evaluation call should exist for summary"
-        )
+        assert (
+            self._evaluate_call is not None
+        ), "Evaluation call should exist for summary"
 
         # Use set_call_stack to temporarily set the evaluation as the parent
         with call_context.set_call_stack([self._evaluate_call]):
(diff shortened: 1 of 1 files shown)


weave/flow/eval_imperative.py
```diff
@@ -478,45 +478,41 @@ def _finalize_evaluation(self, output: Any = None) -> None:
...
-        with _set_current_output(output):
-            with weave.attributes(IMPERATIVE_EVAL_MARKER):
-                _, predict_and_score_call = (
-                    self._pseudo_evaluation.predict_and_score.call(
-                        self._pseudo_evaluation,
-                        self.model,
-                        inputs,
-                        __require_explicit_finish=True,
+        # Use set_call_stack to temporarily set the evaluation as the parent
+        assert self._evaluate_call is not None
```
neutralino1, line 489: Maybe add an error message to help debugging if this happens.
```python
assert self._evaluate_call is not None, "<error message>"
```

weave/flow/eval_imperative.py
```diff
@@ -478,45 +478,41 @@ def _finalize_evaluation(self, output: Any = None) -> None:
...
-                )
 
-        # Get the predict_call from the context variable
-        predict_call = current_predict_call.get()
-        if predict_call is None:
-            raise ValueError("predict_call should not be None")
+            # Get the predict_call from the context variable
+            predict_call = current_predict_call.get()
+            if predict_call is None:
+                raise ValueError("predict_call should not be None")
```
neutralino1, line 507: Maybe use an `assert` like you did above.

neutralino1 (approved):
Approved with a couple of minor non-blocking nits. Thank you for this.

neutralino1:
@andrewtruong merge or close?

PR #4581, opened 2025-05-26
chore(weave): denest inputs.inputs and output.output

zbirenbaum:
## Description

Copies the following fields to `inputs` and `output` directly:
1. inputs.input
2. inputs.inputs
3. output.output
4. output.outputs

Note:
For inputs this will occur only if input.inputs is a dict or a list:
e.g.
`inputs.input: "hello"` will be unaffected and show up under a nested header in the ui
`inputs.input: { value: "hello"}` will show up under inputs.value in the ui


neutralino1 (approved):
Looks good, thank you.

weave/trace_server/opentelemetry/python_spans.py
```diff
@@ -287,8 +287,27 @@ def to_call(
     ) -> tuple[tsi.StartedCallSchemaForInsert, tsi.EndedCallSchemaForInsert]:
         events = [SpanEvent(e.as_dict()) for e in self.events]
         usage = get_weave_usage(self.attributes) or {}
+
         inputs = get_weave_inputs(events, self.attributes) or {}
+        # Only de-nest if we have one key
+        if len(inputs) == 1:
```
zbirenbaum, line 293: I'm open to discussing this but I added this check for if there's multiple keys. It feels like we shouldn't just unpack and drop the other keys.

PR #4582, opened 2025-05-26
chore(weave): Default json structured output to text properly in playground

jwlee64:
## Description
We dont support structured outputs yet, until we do default to text 

after
Screenshot 2025-05-25 at 8 23 42 PM

before
Screenshot 2025-05-25 at 8 23 35 PM

## Testing

How was this PR tested?

### DIFF
--- a/tests/integrations/openai/cassettes/openai_test/test_openai_responses_tool_calling_async_stream.yaml
+++ b/tests/integrations/openai/cassettes/openai_test/test_openai_responses_tool_calling_async_stream.yaml
@@ -1,796 +1,796 @@
 interactions:
-- request:
-    body: '{"input":"What was a positive news story from today?","model":"gpt-4o-2024-08-06","stream":true,"tools":[{"type":"web_search_preview"}]}'
-    headers:
-      accept:
-      - application/json
-      accept-encoding:
-      - gzip, deflate, zstd
-      connection:
-      - keep-alive
-      content-length:
-      - '136'
-      content-type:
-      - application/json
-      host:
-      - api.openai.com
-      user-agent:
-      - AsyncOpenAI/Python 1.66.3
-      x-stainless-arch:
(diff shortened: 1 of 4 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/usePlaygroundState.ts
```diff
@@ -139,7 +139,10 @@ export const usePlaygroundState = () => {
           }
         }
         if (inputs.response_format) {
-          newState.responseFormat = inputs.response_format.type;
+          newState.responseFormat =
+            inputs.response_format.type in PlaygroundResponseFormats
+              ? inputs.response_format.type
+              : PlaygroundResponseFormats.Text;
```
neutralino1, line 145: What is `inputs.response_format.type` in the case of structured output?
We could detect it and return `PlaygroundResponseFormats.JsonObject` instead. Would that display it nicely?

neutralino1 (approved):
One comment. But we can ship this first.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/usePlaygroundState.ts
jwlee64: Screenshot 2025-05-25 at 8 42 05 PM

it can really be anything, type json_object and type text will default, anything else in this field will go to text

before because the field had something here, we were not defaulting properly

neutralino1: Oh wow, didn't know they returned the whole schema of the response. Your treatment is the best option at this time then.

PR #4586, opened 2025-05-27
feat(weave): Chat View for OpenAI responses

zbirenbaum:
## Description

- WB-25453

Display OpenAI Responses calls in Chat View

## Testing
Validated for normal conversations and conversations involving tool calls

image

image


neutralino1 (approved):
Thank you.

PR #4594, opened 2025-05-27
fix(weave): Updating smollagents test for new version

neutralino1:
## Description

`smolagents` got a new version (1.17.0). Which changes its internal logic a bit so that we don't get exactly the same number of runs.


weave/integrations/smolagents/smolagents_sdk.py
```diff
@@ -107,6 +107,7 @@ def get_smolagents_patcher(
             # Patch model-related classes
             get_symbol_patcher("smolagents", "TransformersModel.__call__", base),
             get_symbol_patcher("smolagents", "HfApiModel.__call__", base),
+            get_symbol_patcher("smolagents", "InferenceClientModel.__call__", base),
```
neutralino1, line 110: `HfApiModel` is deprecated in favor of `InferenceClientModel`. Leaving the former for backwards compatibility.

gtarpenning (approved):
🙏🏻

PR #4596, opened 2025-05-28
chore(weave): fix pill icon alignment in trace navigator

gtarpenning:
## Description

[This](https://github.com/wandb/weave/pull/4559) recent pr now causes an alignment issue with the pill. There may be a better way of fixing this, but this restores previous functionality. 

## Testing

Before:
Screenshot 2025-05-28 at 9 42 01 AM

After:
Screenshot 2025-05-28 at 9 41 21 AM

### DIFF
--- a/weave-js/src/components/Tag/Pill.tsx
+++ b/weave-js/src/components/Tag/Pill.tsx
@@ -53,17 +53,24 @@ export type IconOnlyPillProps = {
   icon: IconName;
   color?: TagColorName;
   isInteractive?: boolean;
+  style?: React.CSSProperties;
 };
 export const IconOnlyPill: FC<IconOnlyPillProps> = ({
   icon,
   color,
   isInteractive,
+  style,
 }) => {
   const classes = useTagClasses({color, isInteractive});
   return (
     <Tailwind>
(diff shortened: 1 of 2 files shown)


weave-js/src/components/Tag/Pill.tsx
```diff
@@ -53,17 +53,24 @@ export type IconOnlyPillProps = {
...
   const classes = useTagClasses({color, isInteractive});
   return (
     <Tailwind>
       <div className={twMerge(classes, 'rounded-2xl', 'max-w-[22px]')}>
-        <Icon role="presentation" className="m-4 h-14 w-14" name={icon} />
+        <Icon
+          role="presentation"
+          className="m-4 h-14 w-14"
+          name={icon}
+          style={style}
```
neutralino1, line 72: What is the risk of hard-coding the style here instead of passing it in as a prop?

gtarpenning: there are a few other call sites, including in the Models product which I'm not sure I can test correctly.

neutralino1 (approved):

chance-wnb (approved):
Thanks for the change!

PR #4597, opened 2025-05-28
perf(weave): feedback table join with project_id

gtarpenning:
## Description

The _simplest_ way of actually using the primary index on the feedback table when joining calls and feedback. We add a condition on the project_id, which means we don't join the entire table! This is **much** better, but still not perfect. Really, we should be adding a new `weave_ref_id` to the primary keys of the feedback table, enabling `primary_key, id` primary key usage. 

The more invasive but also more future-proof way of fixing this would be to do: https://github.com/wandb/weave/pull/4598

Master index usage:
Screenshot 2025-05-28 at 12 48 56 PM

Prod:
Screenshot 2025-05-28 at 1 00 18 PM

Branch index usage:
Screenshot 2025-05-28 at 12 48 37 PM

## Testing

Replica testing:

Time to migrate: `0.126s`

Query tested: [prod dd link](https://us5.datadoghq.com/logs?query=span_id%3A196428918209355957&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AwAAAZcYz2MwCyHVZgAAABhBWmNZejJvaEFBQlRlTThzM1FnS05RQVgAAAAkMDE5NzE4ZDItOWVjMy00ZGY5LWI3NDEtNjYyMDIwNzkwZmE0AAdH0A&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=desc&viz=stream&from_ts=1748467920408&to_ts=1748467935689&live=false)

|                         | Elapsed | Rows Read | Data Read | Peak Memory |
|-------------------------|---------|-----------|-----------|-------------|
| prod             | 8.922s  | 7,802,016 | 3.82 GB   | 4.06 GiB    |
| **branch**   | 2.560s  | 7,802,016 | 1.24 GB   | 56.99 MiB   |
| full table migration ([linked pr](https://github.com/wandb/weave/pull/4598))  | 2.247s  | 7,802,016 | 385.32 MB | 72.96 MiB   |


weave/trace_server/migrations/013_add_call_id_to_feedback.up.sql
```diff
@@ -0,0 +1,3 @@
+-- Add column and backfill call_id for all rows that have a call_ref, the id is everything after the last '/'
+-- example: weave-trace-internal:///UHJvamVjdEludGVybmFsSWQ6MTA0MzQwOA==/call/0194f9bc-3347-7920-970e-580ac9ad21ed --> 0194f9bc-3347-7920-970e-580ac9ad21ed 
+ALTER TABLE feedback ADD COLUMN weave_ref_id String MATERIALIZED coalesce(reverse(substring(reverse(weave_ref), 1, position(reverse(weave_ref), '/') - 1)), '');
```
gtarpenning, line 3: this is nasty, but flexible to any id type

neutralino1 (approved):
Thank you!

weave/trace_server/migrations/014_add_call_id_to_feedback.up.sql
```diff
@@ -0,0 +1,3 @@
+-- Add column and backfill weave_ref_id for all rows that have a weave_ref, the id is everything after the last '/'
+-- example: weave-trace-internal:///UHJvamVjdEludGVybmFsSWQ6MTA0MzQwOA==/call/0194f9bc-3347-7920-970e-580ac9ad21ed --> 0194f9bc-3347-7920-970e-580ac9ad21ed 
+ALTER TABLE feedback ADD COLUMN weave_ref_id String MATERIALIZED coalesce(regexpExtract(weave_ref, '/([^/]+)$'), '');
```
tssweeney: We should note that this removes an implied, but not explicit property that you can leave feedback on properties of objects/calls. For example: comment on a row of a dataset. While I think this is probably OK, we should acknowledge this limitation.

gtarpenning: We can massage the regex to try to work in as many cases as possible, do you happen to know what those refs look like?

tssweeney (approved):

PR #4602, opened 2025-05-29
fix(weave): Do not attempt to read .netrc when API key is set in env

zbirenbaum:
## Description

- Fixes 25242

Attempting to read an unavailable netrc file was reported to cause a large delay in execution.

This changes the code to just return the env key if found (which is prioritized anyways) which is how the node SDK already handles this.

### DIFF
--- a/weave_query/weave_query/environment.py
+++ b/weave_query/weave_query/environment.py
@@ -252,8 +252,7 @@ def weave_wandb_api_key() -> typing.Optional[str]:
     if env_api_key is not None:
         return env_api_key
 
-    netrc_api_key = _wandb_api_key_via_netrc()
-    return env_api_key or netrc_api_key
+    return _wandb_api_key_via_netrc()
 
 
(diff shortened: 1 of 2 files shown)


chance-wnb (approved):
Awesome. Thanks!

weave_query/weave_query/environment.py
```diff
@@ -249,17 +249,10 @@ def _wandb_api_key_via_netrc_file(filepath: str) -> typing.Optional[str]:
...
-        if wandb_production():
-            raise errors.WeaveConfigurationError(
-                "WANDB_API_KEY should not be set in both ~/.netrc and the environment."
-            )
-        elif env_api_key != netrc_api_key:
-            print(
-                "There are different credentials in the netrc file and the environment. Using the environment value."
-            )
-    return env_api_key or netrc_api_key
+    if env_api_key is not None:
```
tssweeney, line 252: this change effects the weave query service. cc @domphan-wandb

PR #4603, opened 2025-05-29
chore(ui): add motion library

jamie-rasmussen:
## Description

[Motion](https://motion.dev/) is an animation library I'll be using - handles things like fading out, smoothly transitioning position, etc.

## Testing

How was this PR tested?


gtarpenning (approved):
🏃🏻

PR #4606, opened 2025-05-29
chore(weave): add health check thread to async processor

gtarpenning:
## Description

If the thread dies, we have no way of restarting the queue. Use a seperate health check thread to keep track of processor health and revive if dead.

## Testing

tod

### DIFF
--- a/weave/trace_server_bindings/async_batch_processor.py
+++ b/weave/trace_server_bindings/async_batch_processor.py
@@ -5,10 +5,14 @@ from queue import Empty, Full, Queue
 from threading import Event, Lock, Thread
 from typing import Callable, Generic, TypeVar
 
+import sentry_sdk
+
 from weave.trace.context.tests_context import get_raise_on_captured_errors
 
 T = TypeVar("T")
 logger = logging.getLogger(__name__)
+
+HEALTH_CHECK_INTERVAL = 5.0  # seconds
 
 
 class AsyncBatchProcessor(Generic[T]):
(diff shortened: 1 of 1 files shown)


tests/trace_server_bindings/test_async_batch_processor.py
```diff
@@ -60,3 +60,195 @@ def test_wait_until_all_processed():
...
+
+        # Test 2: Health check revival on full queue
+        mock_logger.reset_mock()
+        with patch.object(
+            processor.health_check_thread, "is_alive", return_value=False
+        ):
+            processor.enqueue([1])  # Fills the queue
+            processor.enqueue(
+                [2]
+            )  # Should trigger Full exception and health check revival
```
andrewtruong: Then why is there no `with pytest.raises` here?

tests/trace_server_bindings/test_async_batch_processor.py
```diff
@@ -60,3 +60,195 @@ def test_wait_until_all_processed():
...
+
+            mock_logger.warning.assert_any_call(
+                "Health check thread died, attempting to revive it"
+            )
+            mock_logger.info.assert_any_call("Health check thread successfully revived")
```
andrewtruong: wouldn't it be better to check if the the thread is literally alive?

tests/trace_server_bindings/test_async_batch_processor.py
gtarpenning: Dumb comment, we catch on `Full` and log the warning, which gets asserted on in other tests in `test_weave_client.py`

tests/trace_server_bindings/test_async_batch_processor.py
gtarpenning: yup

PR #4610, opened 2025-05-30
feat(weave): First version of `LLMAsAJudgeScorer`

neutralino1:
## Description

This version simply supports `llmlite` provides. The purpose is to converge on the scorer's API.

Follow-up work:
- Support `PlaygroundModel`s

### Example usage

```python
from weave.scorers import LLMAsAJudgeScorer
import asyncio

SYSTEM_PROMPT = """
You are a judge that scores the quality of the response to a question.

You will be given an input and an output.

You will need to evaluate whether this output is a true or false statement based on the input.

The response format should be a JSON object with the following fields:
- "is_true": boolean
- "reasoning": string
"""

SCORING_PROMPT = """
Here is the input: {input}

Here is the output: {output}
"""

async def main():
    scorer = LLMAsAJudgeScorer(
        model="anthropic/claude-3-7-sonnet-20250219",
        system_prompt=SYSTEM_PROMPT,
        scorer_prompt=SCORING_PROMPT,
        response_format={"type": "object", "properties": {"is_true": {"type": "boolean"}, "reasoning": {"type": "string"}}}
    )

    score = await scorer.score(
        input="My cat's name is Fluffy.",
        output="The user's cat name is Bobby.",
    )

    print(score)
```
Output:
```
{'is_true': False, 'reasoning': "The input clearly states that the cat's name is Fluffy, but the output incorrectly claims that the cat's name is Bobby. This is a direct contradiction of the information provided in the input, making the output statement false."}
```

## Testing

- Unit tests.
- Manual tests.


tests/scorers/test_llm_as_a_judge_scorer.py
```diff
@@ -0,0 +1,81 @@
+from unittest.mock import AsyncMock, patch
```
tssweeney, line 1: I think the important test that is missing here is the ability to publish, get, then use this model successfully

weave/scorers/llm_as_a_judge_scorer.py
```diff
@@ -0,0 +1,85 @@
...
+from typing import Any, Type, Union
+
+import jsonschema
+from pydantic import BaseModel, Field
+
+from weave.scorers.scorer_types import LLMScorer
+from weave.trace.op import op
+
+
+class LLMAsAJudgeScorer(LLMScorer):
```
tssweeney, line 11: Inheriting from `LLMScorer` seems like overkill. The LLMScorer was written by growth team and just binds litellm methods to the class (and a model_id). Curious why this was chosen

tests/scorers/test_llm_as_a_judge_scorer.py
neutralino1: I didn't write this test as the class inherits from `Scorer` with no exotic attributes. I assumed `Scorer` had sufficient tests in that area. Happy to add more for safety.

weave/scorers/llm_as_a_judge_scorer.py
neutralino1: I assumed this was the logical inheritance chain, but in hindsight, we actually don't want to bind litellm on init as we only want to use it if `model` is a `str`. I'll remove the inheritance.

weave/scorers/llm_as_a_judge_scorer.py
```diff
@@ -0,0 +1,85 @@
...
+    to be scored. The LLM's response is then parsed according to a specified
+    `response_format` (either a JSON schema or a Pydantic model) to extract
+    the score and any other relevant information.
+
+    Attributes:
+        system_prompt: The system prompt to guide the behavior of the scoring LLM.
+        scorer_prompt: The user prompt template for the scoring LLM. This prompt
+                       should include a placeholder for the `output` to be scored,
+                       e.g., "Please score the following output: {output}".
+        model: The identifier of the LLM to be used for scoring (e.g., "gpt-3.5-turbo").
```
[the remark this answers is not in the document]
neutralino1: Yes, the purpose was for `model` to support both `str` for llmlite model slug and `PlaygroundLLM` for a saved model (once Josiah's work lands). I'll remove the inheritance anyway.

tests/scorers/test_llm_as_a_judge_scorer.py
tssweeney: Well it actually inherits from `LLMScorer`. I would **not** make the assumption that any child class of Scorer "just works" in terms of reloading. And this is the whole point if this class, so I would recommend testing it.

neutralino1: Test added.

PR #4615, opened 2025-05-31
fix(weave): Collect the usage info from Langchain integration

chance-wnb:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?

### DIFF
--- a/weave/integrations/langchain/langchain.py
+++ b/weave/integrations/langchain/langchain.py
@@ -32,6 +32,7 @@ This approach allows for more flexible runtime configuration while still respect
 import datetime
 import json
 import logging
+from collections import defaultdict
 from contextlib import contextmanager
 from contextvars import ContextVar
 from uuid import UUID
@@ -44,6 +45,7 @@ from weave.integrations.integration_utilities import (
 from weave.integrations.patcher import Patcher
 from weave.trace.context import call_context
 from weave.trace.context import weave_client_context as weave_client_context
+from weave.trace.op import Op
 from weave.trace.weave_client import Call
 
 import_failed = False
@@ -241,7 +243,7 @@ if not import_failed:
                 # Finish the call.
                 call = self._call_map.pop(run_id)
                 run_dict = _run_to_dict(run, as_input=False)
-                self.wc.finish_call(call, run_dict)
+                self.wc.finish_call(call, run_dict, op=MockOpForFormatting())
 
         def _update_run_error(self, run: Run) -> None:
             call = self._call_map.pop(str(run.id), None)
@@ -365,6 +367,50 @@ else:
 
     class WeaveTracer:  # type: ignore
         pass
+
+
+class MockOpForFormatting(Op):
+    _tracing_enabled = True
+    call_display_name = "langchain.llm"
+    ref = None
+    resolve_fn = None
+    postprocess_inputs = None
+    postprocess_output = None
+    accumulator = None
+
+    def _on_finish_handler(
+        self, call: Call, output: Any, exception: Optional[BaseException] = None
+    ) -> None:
+        """Handle the finish event of a call.
+
+        Args:
+            call: The call that finished
+            output: The output of the call
+        """
+        usage = defaultdict(lambda: defaultdict(int))
+        if output is not None:
+            if "outputs" in output:
+                for output in output["outputs"]:
+                    for generation_list in output["generations"]:
+                        for generation in generation_list:
+                            response_metadata = generation["message"]["kwargs"][
+                                "response_metadata"
+                            ]
+                            model = response_metadata["model_name"]
+                            usage_metadata = response_metadata["usage_metadata"]
+
+                            usage[model]["prompt_tokens"] += usage_metadata.get(
+                                "prompt_token_count", 0
+                            )
+                            usage[model]["completion_tokens"] += usage_metadata.get(
+                                "candidates_token_count", 0
+                            )
+                            usage[model]["total_tokens"] += usage_metadata.get(
+                                "total_token_count", 0
+                            )
+
+        if call.summary is not None:
+            call.summary.update({"usage": usage})
 
 
 weave_tracing_callback_var: ContextVar[Optional[WeaveTracer]] = ContextVar(


chance-wnb:
* **#4615** <a href="https://app.graphite.dev/github/pr/wandb/weave/4615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace/weave_client.py
```diff
@@ -1364,8 +1366,15 @@ def finish_call(
...
-            op._on_finish_handler(call, original_output, exception)
+        if on_finish_handler is not None:
+            _on_finish_handler = on_finish_handler
+        elif op is not None and op._on_finish_handler:
+            _on_finish_handler = op._on_finish_handler
+        else:
+            _on_finish_handler = None
+
+        if _on_finish_handler is not None:
+            _on_finish_handler(call, original_output, exception)
```
neutralino1: Should we instead attach the handler to the op in the langchain integration (or wrap the existing one)? Then it wouldn't require this API change.

chance-wnb: hmm, thats a good idea. I will give it a try

chance-wnb: Oh, I recall why I didn't go that way:

1. in [`summary = {}`](https://github.com/wandb/weave/blob/b668e484ecbab7dbd9cab827c2bf6ad9246a65f7/weave/trace/weave_client.py#L1319), we reconstruct the summary field in the `finish_call()` procedure, so any prior work to store information into `call.summary` is lost. 

2. It might be possible to attach some info onto that [`original_output`](https://github.com/wandb/weave/blob/b668e484ecbab7dbd9cab827c2bf6ad9246a65f7/weave/trace/weave_client.py#L1323C24-L1323C39), so `finish_call()` could automatically collect the usage, and we don't have to introduce a new parameter onto `finish_call()`. However,  `original_output` is called  `original_output` for a reason, I would like to avoid tempering that if possible. 

3. I believe this idea comes from other integrations for example [`op._set_on_finish_handler(google_genai_gemini_on_finish)
`](https://github.com/wandb/weave/blob/b668e484ecbab7dbd9cab827c2bf6ad9246a65f7/weave/integrations/google_genai/gemini_utils.py#L116), on_finish_handler seems to be an API extension point especially reserved for 3rd party integrations. 

4. Maybe what we can do instead is to change `summary = {}` into `summary = call.summary` so the previous work attached to call.summary is not lost. This is however a very implicit integration point than explicitly declare `on_finish_handler` 

Please let me know what you think?

tests/integrations/langchain/cassettes/langchain_test/test_langchain_anthropic_usage.yaml
```diff
@@ -0,0 +1,611 @@
...
+      accept-encoding:
+      - gzip, deflate, zstd
+      connection:
+      - keep-alive
+      host:
+      - raw.githubusercontent.com
+      user-agent:
+      - python-httpx/0.28.1
+    method: GET
+    uri: https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json
```
tssweeney: this litellm block shows up in every single recording (because litellm auto downloads it as the first thing it does on import). it is best practice to attempt to remove it to save space and see if tests still pass. Please do this here and elsewhere.

weave/trace/weave_client.py
tssweeney: @chance-wnb - I think the problem you identified is with summary handling. I have a PR here: https://github.com/wandb/weave/pull/4549 that implements #4. Perhaps you can review that and we can build this PR off that one and I think it will solve the problem?. I think we should take that approach (allow summary to be updated instead of overridden) as opposed to adding a workaround param here.

weave/integrations/langchain/langchain.py
```diff
@@ -367,6 +371,131 @@ class WeaveTracer:  # type: ignore
...
+            and safe_get(generations[0][0], ["generation_info", "usage_metadata"])
+        ):
+            # google genai response format
+            usage = _extract_google_genai_usage(output)
+
+    if call.summary is not None and usage is not None:
+        call.summary.update({"usage": usage})
+
+
+def _extract_openai_usage(output: dict) -> dict:
```
tssweeney: i'd recommend typing these extractions as `LLMUsageSchema` as defined in `trace_server_schema` to ensure they conform to the expected payload

weave/integrations/langchain/langchain.py
```diff
@@ -367,6 +371,131 @@ class WeaveTracer:  # type: ignore
...
+                usage[model]["prompt_tokens"] += usage_metadata.get("input_tokens", 0)
+                usage[model]["completion_tokens"] += usage_metadata.get(
+                    "output_tokens", 0
+                )
+                usage[model]["total_tokens"] += usage_metadata.get("total_tokens", 0)
+
+    return convert_defaultdict_to_dict(usage)
+
+
+def _extract_chat_message_usage(response_metadata: dict) -> dict:
```
tssweeney: these processors make me realize we might want to have a similar extraction of logic from call_finish, for the general case (optional, but seems nice)
```
            # Handle the cases where we are emitting an object instead of a pre-serialized dict
            # In fact, this is going to become the more common case
            model = original_output.model
            usage = original_output.usage
            if isinstance(usage, pydantic.BaseModel):
                usage = usage.model_dump(exclude_unset=True)
            if isinstance(usage, dict) and isinstance(model, str):
                computed_summary[RESERVED_SUMMARY_USAGE_KEY] = {}
                computed_summary[RESERVED_SUMMARY_USAGE_KEY][model] = {
                    "requests": 1,
                    **usage,
                }
```

chance-wnb: Hmm, I am not sure what you refer to. Would you mind explaining in more details?

FWIW, for the several test scenarios I had with LangChain, with different LLM providers. I have never seen the output being an object instead of a dictionary.

weave/trace/weave_client.py
chance-wnb: Sounds cool. I have rebased your PR and removed the extra API contract.

tests/integrations/langchain/cassettes/langchain_test/test_langchain_anthropic_usage.yaml
chance-wnb: I wondered that too. Thanks for the feedback. Yeah, removing them got the tests passed too.

weave/integrations/langchain/langchain.py
tssweeney: Oh, i was just pointing out that the snippet above is the general-purpose way to extract usage and you are basically just creating some special cases for langchain in particular. It would be nice / useful potentially to extract this snippet. Totally optional

chance-wnb: Ok, I see what you mean now. I guess we can consider the extracts* in the langchain integration separate, lets put them into a grave and cover dirt on them. Likely we don't have to see them for some time. :) (I don't treat the integrations an organic part of the core system)

I am thinking that if I extract this general-purpose function somewhere, I would have a hard time to decide where to put it. Especially considering that I am pretending that langchain's extracts are invisible. With them out of the picture, it might diminish the need of extracting this logic and make things fragment. So I would prefer only doing such if there is a need coming from the perspective of weave_client alone.

tssweeney: 👍

PR #4620, opened 2025-06-02
perf(weave): reorder optimization conditions in calls query 

gtarpenning:
## Description

Sql ordering does sometimes have a real effect on the performance of the query, we should be filtering by the most restrictive first. Trace ids have much higher cardinality and are much more specific than `op_name` and `parent_id` so do that first. When ref ids are present, they too are much more specific than op names, and should happen before. 

also renames optimizations like the str and datetime sql strings to include "opt" in the name, as they _require_ additional filtering and are only optimizations. 

## Testing

Updates tests


weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -763,7 +763,7 @@ def _as_sql_base_format(
             pb,
             table_alias,
         )
-        sortable_datetime_sql = (
+        sortable_datetime_opt_sql = (
```
bcsherma, line 766: why are you renaming these?

gtarpenning: Now, you can _easily_ tell which of the filters are optimizations, requiring an additional condition after the `HAVING` block, and which **replace** the having block entirely.

bcsherma (approved):

gtarpenning:
nah

PR #4621, opened 2025-06-02
chore(weave): add unprocessable item handling in batch processor

gtarpenning:
## Description

[WB-25436](https://wandb.atlassian.net/browse/WB-25436)

If somehow there is a queue item that is unprocessable, for whatever reason, there is no recourse for handling that broken item. This pr implements a handling scheme that will attempt to write unprocessable items to a log jsonl file. The max size of the log file is set at 512MB before writing to a new file, with a max of 3. They get rotated when the third file is full. This should make the queue more robust. Because this processor batches requests, but unprocessable queue items will likely be individual, now when a batch fails we iterate through the batch manually to identify the broken item. This _should_ happen very rarely, but might be a potential performance bottleneck if there are somehow many unprocessable items. The errors raised in this case are intentionally quite obnoxious, this should be a rare and troublesome error case. 

## Testing

- test for unprocessable entity handling
- specific test for log rotate, end to end + unit 

[WB-25436]: https://wandb.atlassian.net/browse/WB-25436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server_bindings/async_batch_processor.py
```diff
@@ -96,13 +128,17 @@ def _process_batches(self) -> None:
...
                     self.processor_fn(current_batch)
+                    for item in current_batch:
+                        self.queue.task_done()
                 except Exception as e:
                     if get_raise_on_captured_errors():
                         raise
-                    logger.exception(f"Error processing batch: {e}")
-                else:
-                    for _ in current_batch:
-                        self.queue.task_done()
```
andrewtruong: why did you change this?

weave/trace_server_bindings/async_batch_processor.py
```diff
@@ -111,25 +147,107 @@ def _process_batches(self) -> None:
...
-        # Start a new processing thread
-        self.processing_thread = start_thread(self._process_batches)
-        # Restart health check thread
-        self.health_check_thread = start_thread(self._health_check)
+    def _process_batch_individually(self, batch: list[T]) -> None:
+        """Process each item in a batch individually, handling failures appropriately."""
+        for item in batch:
+            try:
+                self.processor_fn([item])
+                self.queue.task_done()
```
andrewtruong: this can be in a finally block

andrewtruong (approved):

weave/trace_server_bindings/async_batch_processor.py
gtarpenning: I probably thought it was more readable, i'm never exactly sure what the try/except/else stuff does. I'll change it back though now I get it

tests/trace_server_bindings/test_async_batch_processor.py
```diff
@@ -243,3 +246,209 @@ def tracking_processor_fn(batch):
...
+        nonlocal successful_items
+        processed = []
+        for item in batch:
+            if item == poison_const:
+                raise RuntimeError("Poison pill error")
+            elif item == batch_killer_const:
+                raise RuntimeError("Batch processing error")
+            processed.append(item)
+
+        print(f"Processed: {processed}")
```
andrewtruong: track

gtarpenning: I like the prints in tests, but okay

PR #4629, opened 2025-06-02
chore(weave): Fix double stream loading

jwlee64:
## Description
Separate the loading states
Before even if one chat finished it wouldnt show the whole text because the other was still loading

Before

https://github.com/user-attachments/assets/db82a54a-32c1-49df-b8f4-053564e60ff9

after

https://github.com/user-attachments/assets/b3888198-719c-4f11-9465-8cec5831ccab


zbirenbaum (approved):
LGTM

PR #4636, opened 2025-06-03
chore(weave): add leaf_object_class to object_versions table

jwlee64:
## Description
Adds a leaf_object_class field to the object versions table
This enables saved playground models to inherit from model and be base_object_class model as well as a leaf_object class of LLMStructuredCompletionModel

## Testing

How was this PR tested?


weave/trace_server/migrations/014_add_leaf_object_class.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE object_versions
```
tssweeney, line 1: may want to update object_versions_stats_view as well

weave/trace_server/trace_server_interface.py
```diff
@@ -509,6 +511,11 @@ class ObjectVersionFilter(BaseModel):
         description="Filter objects by their base classes",
         examples=[["Model"], ["Dataset"]],
     )
+    leaf_object_classes: Optional[list[str]] = Field(
```
tssweeney, line 513: can you please add some tests that:
1. showcase filtering by leaf object classes (particularly when they differ from the base)
2. showcase inserting where leaf object class is extracted from payload
3. showcase inserting where leaf object class is provided as a field and checked

weave/trace_server/migrations/014_add_leaf_object_class.up.sql
jwlee64: makes sense

weave/trace_server/trace_server_interface.py
jwlee64: Ok, i added tests for varying object hierarchies and filtering states

weave/trace_server/migrations/014_add_leaf_object_class.up.sql
tssweeney: i think you might need to also update `object_versions_stats_view`. it looks like the last update of `object_versions_stats` may have forgot this

weave/trace_server/trace_server_interface.py
```diff
@@ -210,6 +211,7 @@ class ObjSchemaForInsert(BaseModel):
     object_id: str
     val: Any
     builtin_object_class: Optional[str] = None
+    leaf_object_class: Optional[str] = None
```
tssweeney: this might be a good opportunity to add field descriptions here and above (possibly to the related other object class as well

tests/trace/test_objs_query.py
```diff
@@ -322,3 +374,540 @@ def test_objs_query_delete_and_add_new_versions(client: WeaveClient):
     )
     assert len(res.objs) == 3
     assert all(obj.val["i"] in [4, 5, 6] for obj in res.objs)
+
+
+def test_objs_query_filter_leaf_object_class_basic(client: WeaveClient):
```
tssweeney: this is looking really good! Nice work. I do think there are a few more tests that are not captured here (see `tests/trace/test_base_object_classes.py` testing of `builtin_object_class`).
1. Setting a leaf object class on the create side and having the backend validate it and support deserialization
2. Setting a leaf object class on the create side and having an incorrect payload resulting in a validation error

weave/trace_server/trace_server_interface.py
tssweeney: ohhh actually the more I am reading this, the more I am wondering why we need `leaf_object_class` here (specifically on the insert). Should't `builtin_object_class` be treated as the leaf? Under what conditions do you have a valid leaf class that is not a builtin!?

tests/trace/test_objs_query.py
tssweeney: Oh, actually, this comment below pairs well: https://github.com/wandb/weave/pull/4636/files#r2125618034 ... i think when setting the builtin_object_class, it should also set the leaf correctly!

tssweeney (commented):
Ok - summarizing the key points:
a) everything looks really nice! Well done 👍 

I think the following comment should just result in a little tweak of the code and some more tests. First, let's define the different terms:
* `base_object_class`: for subclasses of `weave.Object` or `weave.BaseObject`, the *fist child* class of the instance
* `builtin_object_class`: Any subclass of `weave.Object` or `weave.BaseObject` which is explicitly registered as a "builtin". Before this PR, this would be the `weave.BaseObject`s that are explicitly shared between server and client. 
* (new in this PR) `leaf_object_class`: for subclasses of `weave.Object` or `weave.BaseObject`, the *last descendent* class of the instance

Now, I believe there are a few characteristics that we want to hold:
1. Ability to query by `leaf_object_class` (and have avail on the object itself)
2. Ability to extract `leaf_object_class` when available in the object create value payload.
3. Correctly set `leaf_object_class` when using the `builtin_object_class` query param.

* Your PR correctly tests and implements 1 & 2.

However, there is a gap in 3:
1. Adding `leaf_object_class` to the `ObjSchemaForInsert` is unnecessary (I think). Because by definition, anything you set there, should also be set for the sibling `builtin_object_class`. And when you set the `builtin_object_class` - that literally is the leaf_object_class
2. We don't have any tests for the lower level usage of this field (all your added tests focus on 1 and 2). There are examples in `tests/trace/test_base_object_classes.py`. Essentially, I think all you really need to test is that when you set the `builtin_object_class`, the `leaf_object_class` is set on the stored object correctly.

So, i think this is just a little removal of the `leaf_object_class` setter, and a few lines of additional assertions in the other test file.

However, you might find that the test style you are using for the other tests will fail to correctly work since `SimpleModel` and it's siblings are not defined on the server and registered as a builtin. I got around this by defining `TestOnlyNestedBaseModel` which you might want to create something similar.

jwlee64:
Okay I moved the tests to test base object classes, and added a test object called testOnlyInheritedBaseObject.
Changed the tests to use only built ins and updated the migration, Should be good to go @tssweeney. Let me know if i missed anything

tssweeney (approved):
Nice!

PR #4640, opened 2025-06-05
chore(dev): workspaces for the trace table

bcsherma:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?


bcsherma:
* **#4640** <a href="https://app.graphite.dev/github/pr/wandb/weave/4640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4640?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/aggregation.ts
```diff
@@ -0,0 +1,351 @@
+/*
```
zbirenbaum, line 1: Do tests belong in the same directory as the src?

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/extractData.test.ts
```diff
@@ -0,0 +1,559 @@
+/**
```
zbirenbaum: Same here, do tests belong here?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -1084,99 +1154,264 @@ export const CallsTable: FC<{
           </div>
         </TailwindContents>
       }>
-      {isMetricsChecked && (
-        <CallsCharts
-          entity={entity}
-          project={project}
-          filter={filter}
-          filterModelProp={filterModelResolved}
+      <style>
```
tssweeney: this style block seems unnecessary: 
1. redundant with other style inlining
2. everything can be put directly into the style blocks. 

We already mix direct `style` tags, MUI's `sx` tag, and Tailwind, it would be nice to not have another style method layered in

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -1084,99 +1154,264 @@ export const CallsTable: FC<{
...
+            style={{
+              width: `${tableWidthFraction * 100}%`,
+              minWidth: '300px',
+              height: '100%',
+              overflow: 'hidden',
+              flexShrink: 0,
+              flexGrow: 0,
+              boxSizing: 'border-box',
+            }}>
+            <StyledDataGrid
```
tssweeney: is this StyledDataGrid and the next one exactly the same? if so, it should be defined above and then included in each of your blocks to keep it DRY

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -662,6 +664,74 @@ export const CallsTable: FC<{
     setSelectedCalls([]);
   }, [setSelectedCalls]);
 
+  // Detect peek drawer state to flip proportions
+  const peekLocation = usePeekLocation();
+  const isPeekOpen = peekLocation != null;
+
+  // Default proportions: 60% table, 40% charts
+  // When peek drawer opens: 40% table, 60% charts
+  const baseTableWidthFraction = isPeekOpen ? 0.4 : 0.6;
```
tssweeney: please move to constants

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -662,6 +664,74 @@ export const CallsTable: FC<{
...
+      if (!isDragging) return;
+
+      const container = document.querySelector('.calls-table-layout');
+      if (!container) return;
+
+      const containerRect = container.getBoundingClientRect();
+      const newWidth = e.clientX - containerRect.left;
+
+      // Set min and max width constraints
+      const minWidth = 300;
```
tssweeney: please move to constants

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -662,6 +664,74 @@ export const CallsTable: FC<{
...
+  const handleMouseDown = useCallback((e: React.MouseEvent) => {
+    setIsDragging(true);
+    e.preventDefault();
+  }, []);
+
+  const handleMouseMove = useCallback(
+    (e: MouseEvent) => {
+      if (!isDragging) return;
+
+      const container = document.querySelector('.calls-table-layout');
```
tssweeney: using selector pattern makes it prohibitive to have this component appear twice on the page. Please use an element ref pattern.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -662,6 +664,74 @@ export const CallsTable: FC<{
     setSelectedCalls([]);
   }, [setSelectedCalls]);
 
+  // Detect peek drawer state to flip proportions
```
tssweeney: These hooks pair with 

```
<div
            className={`resize-handle ${isDragging ? 'dragging' : ''}`}
            onMouseDown={handleMouseDown}
          />
```

I would challenge you to make a generic `Draggable` esk component that hides these implementation details. Then the CallsTable just includes that with some params

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/extractData.ts
```diff
@@ -0,0 +1,957 @@
...
+    axisFields.push(axisField);
+  });
+
+  return axisFields;
+}
+
+export function getInputOutputFieldValue(
+  extractedData: ExtractedCallData,
+  fieldKey: string
+): any {
```
tssweeney: Here, and in many places, I think you want to use `unknown` and not `any`. The reason is because once something is an `any` it can be used anywhere, without type guards. Opposite: `unknown` basically can't be used anywhere until it is type checked. This will require more type checking in consumption code, but that is generally better so that you avoid unexpected types at runtime.

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/persistence.ts
```diff
@@ -0,0 +1,67 @@
...
+        xAxis: 'prompt_tokens',
+        yAxis: 'completion_tokens',
+        plotType: 'scatter',
+      },
+    ],
+    openDrawerChartId: null,
+  };
+}
+
+export function loadChartsConfig(
```
tssweeney: even thought these are not async right now - once they are backed by backend data storage, they will need to be. Making them async now could help reduce the migration cost later by forcing your to promisify/hookify parts of the app now

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/types.ts
```diff
@@ -0,0 +1,144 @@
...
+  annotations: {[key: string]: any};
+  /** Scorer values keyed by scorer name (e.g., { toxicity: 0.2, coherence: 0.9 }) */
+  scorers: {[key: string]: any};
+  /** Array of note strings */
+  notes: string[];
+  /** Array of reaction emojis */
+  reactions: string[];
+};
+
+export type ExtractedCallData = {
```
tssweeney, line 25: this seems nearly identical to other `Call` definitions - any reason they are re-defined instead of extended?

tssweeney:
Screenshot 2025-06-30 at 11 05 21

It would be nice if the `Add Plot` drawer had an exit/cancel/close button

tssweeney:
Screenshot 2025-06-30 at 11 07 15

overlow headers is odd

tssweeney:
Screenshot 2025-06-30 at 11 07 53

P95 chart is really easy to "zoom" into no data, and when that happens it is just completely blank

tssweeney:
Screenshot 2025-06-30 at 11 12 16

The axis labels get truncated when they are longer. This one is hard... i know.

zbirenbaum:
Couple of Strange Behaviors With Grouping Option:
1. Open `Create Plot Menu`:
image

2. Select Display Name Grouping:
image

3. Clearing Grouping does not restore previous state:
image

3a. Selecting Op Name grouping changes the plot but still shows "No Grouping"
image

zbirenbaum:
Not sure `Sum` is being calculated correctly. I would expect that doing a `Sum cost` on the Y-axis with `started at` on the X-axis would exclusively increase, but instead I get this plot.

image

zbirenbaum:
If there is no data prior to a certain point on the X Axis, the chart should automatically re-size so that the first data point is located at the origin.

Maybe recalculating X & Y dims on each option select would fix this
image

bcsherma:
> If there is no data prior to a certain point on the X Axis, the chart should automatically re-size so that the first data point is located at the origin.
> 
> Maybe recalculating X & Y dims on each option select would fix this image

@zbirenbaum the sum is the sum of values within a particular time bin, not meant to be a cumulative across bins. That is why you don't see a monotonically increasing plot.

zbirenbaum:
> > If there is no data prior to a certain point on the X Axis, the chart should automatically re-size so that the first data point is located at the origin.
> > Maybe recalculating X & Y dims on each option select would fix this image
> 
> @zbirenbaum the sum is the sum of values within a particular time bin, not meant to be a cumulative across bins. That is why you don't see a monotonically increasing plot.

This makes sense, I do think the stuff around the grouping option might want to be addressed. I know this is a huge PR though and Tim already gave approval so I definitely wouldn't be opposed to making it a followup

bcsherma:
> If there is no data prior to a certain point on the X Axis, the chart should automatically re-size so that the first data point is located at the origin.
> 
> Maybe recalculating X & Y dims on each option select would fix this image

This is the current behavior - I think the issue with the plot you shared is the left end has a single point that doesn't connect to any other points. the point should show if you hover over the bin. I think I should have series of length 1 always show their single point

weave-js/src/components/Button/Button.tsx
```diff
@@ -55,6 +55,9 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
     },
     ref
   ) => {
+    if (active) {
```
zbirenbaum: Remove

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/ChartModal.tsx
```diff
@@ -0,0 +1,700 @@
...
+
+    if (reactionFields.length > 0) {
+      groups.push({
+        label: 'Reactions',
+        options: reactionFields,
+      });
+    }
+
+    return groups;
+  }, [isScatterPlot, localConfig.plotType, callData]);
```
zbirenbaum: Redundant.

Add this helper:
```
const groupFields = (fields: ChartAxisField[]) => {
  const groups: Record<string, { label: string, options: FieldOption[] }> = {
    general: { label: 'General', options: [] },
    inputs: { label: 'Inputs', options: [] },
    outputs: { label: 'Outputs', options: [] },
    annotations: { label: 'Annotations', options: [] },
    scores: { label: 'Scores', options: [] },
    reactions: { label: 'Reactions', options: [] },
  };

  // Sometimes forEach can cause issues with duplication. May want to map here instead
  fields.forEach(field => {
    const option = { value: field.key, label: field.label };
    if (field.key.startsWith('input.')) {
      option.label = field.label.replace('Input: ', '');
      groups.inputs.options.push(option);
    } else if (field.key.startsWith('output.')) {
      option.label = field.label.replace('Output: ', '');
      groups.outputs.options.push(option);
    } else if (field.key.startsWith('annotations.')) {
      groups.annotations.options.push(option);
    } else if (field.key.startsWith('scores.')) {
      groups.scores.options.push(option);
    } else if (field.key.startsWith('reactions.')) {
      groups.reactions.options.push(option);
    } else {
      groups.general.options.push(option);
    }
  });

  return Object.values(groups).filter(g => g.options.length > 0);
};
```

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/ChartModal.tsx
```diff
@@ -0,0 +1,700 @@
...
+
+    if (reactionFields.length > 0) {
+      groups.push({
+        label: 'Reactions',
+        options: reactionFields,
+      });
+    }
+  }
+
+  return groups;
```
zbirenbaum: Redundant

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/BarChart.tsx
```diff
@@ -0,0 +1,546 @@
...
+                        group: value.group,
+                        color: value.color,
+                      },
+                      binStart,
+                      binEnd,
+                    });
+                  }}
+                  onValueMouseOut={() => setHintValue(null)}
+                />
+              ))}
```
zbirenbaum: Could we pre-calculate this for each bin rather than recalculating on mouseover? With enough data this could be pretty performance intensive

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/ChartsContext.tsx
```diff
@@ -0,0 +1,133 @@
...
+    case 'ADD_CHART': {
+      const needsBinning =
+        action.payload?.plotType === 'line' ||
+        action.payload?.plotType === 'bar';
+      const newChart: ChartConfig = {
+        id: uuidv4(),
+        xAxis: action.payload?.xAxis || defaultXAxis,
+        yAxis: action.payload?.yAxis || defaultYAxis,
+        plotType: action.payload?.plotType || 'scatter',
+        ...(needsBinning ? {binCount: 20, aggregation: 'average'} : {}),
```
zbirenbaum: Any particular reason binning is set to 20 by default? Maybe make a constant called DEFAULT_BIN_COUNT?

zbirenbaum (approved):
Couple of comments, mostly optional

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/ChartModal.tsx
zbirenbaum: Call it here like this:
```
const createGroupedAxisOptions = (
  baseFields: typeof chartAxisFields,
  callData: ExtractedCallData[],
  filterFn?: (field: ChartAxisField) => boolean
): GroupedFieldOptions[] => {
  let allFields = [...baseFields];
  if (callData.length > 0) {
    const schema = extractDynamicFields(callData);
    allFields.push(...convertSchemaToAxisFields(schema));
  }
  const filteredFields = allFields.filter(field => !filterFn || filterFn(field));
  return groupFields(filteredFields);
};
```

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/ChartsContext.tsx
bcsherma: No it's very arbitrary, moving to a constant though

weave-js/package.json
```diff
@@ -135,6 +135,7 @@
     "react-vega": "^7.4.1",
     "react-virtualized": "^9.10.1",
     "react-virtualized-auto-sizer": "^1.0.6",
+    "react-vis": "^1.12.1",
```
zbirenbaum, line 138: react-vis hasn't been meaningfully updated in a few years. Recharts and Nivo are decent options. Maybe one of them is fairly compatible with minimal effort to change. Totally optional though

bcsherma: Yeah I went with react viz because we use it in models and it was the easiest to match the style of the designs and the models plots. I messed with other things and if I had to start over I would probably do recharts. The data rendering is easy to port but we getting the interaction behavior to match will get trickier as we add more interaction behaviors.

weave-js/src/components/PagePanelComponents/Home/Browse3/charts/ChartModal.tsx
bcsherma: There is some logic in there to identify which of these fields are categorical, but I was able to highly simplify and eliminate like 135 lines, ty

PR #4645, opened 2025-06-05
chore(app): Updating tag to support icon only display

jtulk:
- gap spacing preferred to explicit since it automatically handles different numbers of items
- setting explicit height so it doesn't require line-height from `truncatedText`
- removing spacing cheat since it unbalances no-text versions

## Description
I require a single icon version of a tag:
image

This looks awful in existing implementations:
Screenshot 2025-06-05 at 1 17 44 PM

So I made it not awful:
Screenshot 2025-06-05 at 1 18 36 PM

### DIFF
--- a/weave-js/src/components/Tag/Tag.tsx
+++ b/weave-js/src/components/Tag/Tag.tsx
@@ -43,7 +43,7 @@ export function useTagClasses({
     () =>
       classNames(
         'night-aware',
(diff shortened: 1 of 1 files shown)


jtulk:
@onx2 I just one-offed the element in the `RunSelectorIssuesCell` to avoid making changes to the design system that might not be fully blessed. I'll close this.

PR #4646, opened 2025-06-05
chore(weave): only show feedback hover with reactions

gtarpenning:
## Description

Only show the popup when hovered or actually has existing reactions, not all feedback (which is now more broad). 

## Testing

Prod
Screenshot 2025-06-05 at 2 20 05 PM

Branch
image

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/ReactionsLoaded.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/ReactionsLoaded.tsx
@@ -138,7 +138,10 @@ export const ReactionsLoaded = ({
 
   const onMouseEnter = () => setIsMouseOver(true);
   const onMouseLeave = () => setIsMouseOver(false);
-  const showContent = forceVisible || isMouseOver || reactions.length > 0;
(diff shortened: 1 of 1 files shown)


jamie-rasmussen (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/feedback/ReactionsLoaded.tsx
```diff
@@ -273,3 +276,10 @@ export const ReactionsLoaded = ({
     </Tailwind>
   );
 };
+
+function feedbackIsLegacy(feedbackType: string) {
+  return (
+    feedbackType.startsWith('wandb.note') ||
+    feedbackType.startsWith('wandb.reaction')
```
neutralino1, line 283: Why are these legacy? I still see emoji reactions and notes persisted with those feedback types.

gtarpenning: it's the legacy way of defining feedback, before we added the more complex ref fields. I just used "legacy" because it was consistent with assignments in that file. Reactions and notes are supported of course, maybe "default" would have been a better name.

PR #4652, opened 2025-06-06
chore(weave): wait after flush for clickhouse to be ready in tests

gtarpenning:
## Description

Check if a query in a test is a call read query, if so optimize the table before reading. 

## Testing

test!

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import time
 import base64
 import contextlib
 import logging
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -438,7 +439,21 @@ def __getattribute__(self, name):
...
+                    )._next_trace_server._internal_trace_server
+                    # Sleep to allow inserts to become available, when flush did something
+                    if (
+                        isinstance(
+                            server,
+                            clickhouse_trace_server_batched.ClickHouseTraceServer,
+                        )
+                        and has_pending_jobs
+                    ):
+                        time.sleep(0.01)
```
chance-wnb: I think maybe do a `SELECT count(*) FROM calls_merged FINAL` before the sleep will help trigger Materialized View aggregation immediately.

tssweeney (changes requested):
I would be concerned about this one. We would not be running this in production and could lead us to writing code that depends on finalized tables.

PR #4653, opened 2025-06-06
chore(weave): dspy 22 -> 32 eval calls

gtarpenning:
## Description

Version bump? [This test](https://github.com/wandb/weave/actions/runs/15498991640/job/43642396204?pr=4652) is now not passing


gtarpenning:
???

PR #4655, opened 2025-06-06
feat(weave): OpenAI API endpoint tracking

chance-wnb:
## Description

- Captures the OpenAI API endpoint URL in call attributes when using the OpenAI SDK
- Extracts the base URL from the completion client and adds it to the call input parameter `self`.

## Testing

Tested by verifying that the API endpoint is correctly captured in the call input when making OpenAI API calls through the SDK.

Preview:

image

### DIFF
--- a/weave/integrations/openai/openai_sdk.py
+++ b/weave/integrations/openai/openai_sdk.py
@@ -9,7 +9,12 @@ from urllib.parse import urlparse
 import weave
 from weave.integrations.patcher import MultiPatcher, NoOpPatcher, SymbolPatcher
 from weave.trace.autopatch import IntegrationSettings, OpSettings
-from weave.trace.op import Op, ProcessedInputs, _add_accumulator
+from weave.trace.op import (
+    Op,
+    ProcessedInputs,
+    _add_accumulator,
+    _default_on_input_handler,
+)
 
 if TYPE_CHECKING:
     from openai.types.chat import ChatCompletionChunk
@@ -300,26 +305,58 @@ def should_use_accumulator(inputs: dict) -> bool:
     )
 
 
+def convert_completion_to_dict(obj: Any) -> dict:
+    return {
+        "client": {
+            "base_url": str(obj._client._base_url),
+            "version": obj._client._version,
+        },
+    }
+
+
+def completion_instance_check(obj: Any) -> bool:
+    return (
+        hasattr(obj, "messages")
+        and hasattr(obj, "_client")
+        and hasattr(obj._client, "_base_url")
+        and hasattr(obj._client, "_version")
+    )
+
+
 def openai_on_input_handler(
     func: Op, args: tuple, kwargs: dict
 ) -> ProcessedInputs | None:
+    original_args = args
+    original_kwargs = kwargs
+
+    processed_inputs = _default_on_input_handler(func, tuple(args), kwargs)
+    inputs = processed_inputs.inputs
+
+    args = list(args)  # type: ignore[assignment]
+    if len(args) > 0 and completion_instance_check(args[0]):
+        # This will be the `self` argument to the function, convert it to a dict
+        args[0] = convert_completion_to_dict(args[0])  # type: ignore[index]
+        inputs.update({"self": args[0]})
+
     if len(args) == 2 and isinstance(args[1], weave.EasyPrompt):
         original_args = args
         original_kwargs = kwargs
         prompt = args[1]
         args = args[:-1]
         kwargs.update(prompt.as_dict())
-        inputs = {
-            "prompt": prompt,
-        }
-        return ProcessedInputs(
-            original_args=original_args,
-            original_kwargs=original_kwargs,
-            args=args,
-            kwargs=kwargs,
-            inputs=inputs,
+        inputs.update(
+            {
+                "prompt": prompt,
+            }
         )
-    return None
+
+    return ProcessedInputs(
+        original_args=original_args,
+        original_kwargs=original_kwargs,
+        args=tuple(args),
+        kwargs=kwargs,
+        inputs=inputs,
+    )
 
 
 def create_wrapper_sync(settings: OpSettings) -> Callable[[Callable], Callable]:
--- a/tests/integrations/openai/test_openai_sdk.py
+++ b/tests/integrations/openai/test_openai_sdk.py
@@ -1,20 +1,161 @@
+from unittest.mock import Mock
+
 import pytest
 
+import weave
 from weave.integrations.openai.openai_sdk import (
     create_wrapper_async,
     create_wrapper_sync,
+    openai_on_input_handler,
 )
 from weave.trace.autopatch import OpSettings
 
 
 class DummyClient:
-    def __init__(self, base_url: str):
+    def __init__(self, base_url: str, version: str = "1.0.0"):
         self._base_url = base_url
+        self._version = version
 
 
 class DummyCompletion:
-    def __init__(self, base_url: str):
-        self._client = DummyClient(base_url)
+    def __init__(self, base_url: str, version: str = "1.0.0"):
+        self._client = DummyClient(base_url, version)
+        self.messages = []
+
+
+class NonCompletion:
+    """An object that doesn't match completion_instance_check requirements"""
+
+    def __init__(self):
+        self.data = "not a completion"
+
+
+def test_openai_on_input_handler_with_completion_instance():
+    """Test that openai_on_input_handler processes completion instances correctly"""
+    completion = DummyCompletion("https://api.openai.com", "1.2.3")
+    mock_op = Mock()
+    mock_op.name = "test_op"
+
+    args = (completion, {"model": "gpt-4", "messages": []})
+    kwargs = {"temperature": 0.7}
+
+    result = openai_on_input_handler(mock_op, args, kwargs)
+
+    # Should return ProcessedInputs
+    assert result is not None
+    assert hasattr(result, "inputs")
+    assert hasattr(result, "args")
+    assert hasattr(result, "kwargs")
+
+    # Should have converted completion to dict in inputs
+    assert "self" in result.inputs
+    expected_self = {
+        "client": {
+            "base_url": "https://api.openai.com",
+            "version": "1.2.3",
+        }
+    }
+    assert result.inputs["self"] == expected_self
+
+    # Args should be modified (completion converted to dict)
+    assert result.args[0] == expected_self
+    assert result.args[1] == {"model": "gpt-4", "messages": []}
+
+    # Kwargs should remain unchanged
+    assert result.kwargs == {"temperature": 0.7}
+
+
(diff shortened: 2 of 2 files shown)


chance-wnb:
* **#4655** <a href="https://app.graphite.dev/github/pr/wandb/weave/4655?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4655?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

neutralino1 (approved):

weave/integrations/openai/openai_sdk.py
```diff
@@ -303,6 +304,14 @@ def should_use_accumulator(inputs: dict) -> bool:
 def openai_on_input_handler(
     func: Op, args: tuple, kwargs: dict
 ) -> ProcessedInputs | None:
+    attrs = call_context.call_attributes.get()
+
+    completion = args[0]
+    if hasattr(completion, "_client") and hasattr(completion._client, "_base_url"):
+        base_url = str(completion._client._base_url)
+        attrs["api_endpoint"] = base_url
```
tssweeney: `api_endpoint` as a top-level attribute needs to be carefully considered.
1. Once you add this here, there is going to be an expectation to match this elsewhere. It is going to end up having queries against it etc... this is almost as important as thinking about a DB field. 
2. We would typically put weave-specific values under the `weave` sub-key. However in this case, it is closer to user-defined so it seems OK to put here.
3. Do we want to call it `client_base_url`? Why change the name from the source?

weave/integrations/openai/openai_sdk.py
```diff
@@ -303,6 +304,14 @@ def should_use_accumulator(inputs: dict) -> bool:
 def openai_on_input_handler(
     func: Op, args: tuple, kwargs: dict
 ) -> ProcessedInputs | None:
+    attrs = call_context.call_attributes.get()
```
tssweeney: This is not quite right. As an implementation detail, we happen to be calling `input_handler` while the call is still on the stack, but that is circumstantial. It really seems like we should be able to extract the base_url as part of the serialization procedure for the client. I think that would be a better path.

I think instead you should define a custom serializer for `class OpenAI` and `class AsyncOpenAI`. That way `completion._client` is serialized to a payload representative of what it is.

weave/integrations/openai/openai_sdk.py
chance-wnb: Thanks for the tip. 

On number 3, I have a reason but I don't feel strongly. This info is both a configuration of the client as (`client_base_url`) and the destination the requests will be sent to (`api_endpoint`). In the business standpoint that initiated the work, we care more about which API endpoint served the requests. So I feel `api_endpoint` is easier to get a Aha than `client_base_url`.

weave/integrations/openai/openai_sdk.py
chance-wnb: Thanks for the feedback. May I ask what `a custom serializer` refers to? do we have a precedence?

chance-wnb: Hmm, I think this `openai_on_input_handler` is already the input serializer you talked about, `_on_input_handler` will decide what get logged as input to the database, but it doesn't influence what inputs are passed to the original function, which will not break things.

However, is artificially tempering the original input a good idea? (It is not what the method got called with, the original openai.`create()` function does not have the _client as a parameter). This essentially moves the info from the `attributes` column to the `inputs` column. 

Can you help confirm that you indeed want to go this route, and why?

tssweeney: Sure. I am certainly not recommending to tamper with the original values. Let me be more clear:

1. `on_input_handler` is intended to allow users to override how _input_ data is serialized before being sent over the wire. As you mention, this is non-destructive and does not mess with the call path.
2. However, using `on_input_hander` to modify attributes is a pattern we don't do anywhere else. More importantly, as implemented seems fragile due to the use of `call_attributes` context. There is no explicit contract that says it is safe to mutate call_attributes in the input handler. The only reason this works is because `attributes = dictify(call_attributes.get())` happens to be after `pargs = func._on_input_handler(func, args, kwargs)` in `def _create_call(`. This implicit execution order dependency seems circumstantial and not contractual. We are taking something that is intended to be a pure transformation and intruding side-effects. 

Therefore, option 1: if you want to use this patten, I would either:
1. Write an explicit enforce that contract (updating `call_attributes` inside of an op's `on_input_handler` does indeed persist those attributes)
2. Modify `OnInputHandlerType` to allow a return of attributes as well.
3. Add an equivalent `_get_derived_attributes` callback that explicitly does this.

Option 1 is the easiest, but leaves in the implicit contract that the `call_attributes` context is correct during execution of `on_input_hander`.

---

Now, all of this is only needed if you think `api_endpoint` should indeed be an `attribute`. Right now, `attributes` are "things we know about the context of the call" ... and in this case expands that a bit to be: "derived properties of the call". While subtle, I think it is an important note. Furthermore, `api_endpoint` is indeed a property of the inputs to the call. It is `self._client._base_url` (as you explicitly extract). Currently the `self` arg is a `openai.resources.chat.completions.completions.Completions` object. Since Weave does not know how to serialize this object, we just get `<openai.resources.chat.completions.completions.Completions object at 0x7e0a656853d0>` (see screenshot) as the input value. This is not really useful at all since it is just the class name and a memory address that changes on every call. In a better world, it would be a serialized object describing the client. Ideally, we create a custom serializer that generates a dict representation like
```
{
   _type: ...,
   val: {
     base_url: ...
     # ... any other important properties about the client itself
   }
}
```

See `weave/type_handlers/Markdown/markdown.py` for an example of how to do this.

This would then ensure that anytime a `openai.resources.chat.completions.completions.Completions` is serialized across the application, an actually useful payload is created.

Screenshot 2025-06-09 at 10 58 42

chance-wnb: This is great. Thanks for the detailed explanation. 

I went to the non-attribute approach. ~~However there is one issue: if we zoom in to [`_save_nested_objects`](https://github.com/wandb/weave/blob/5d16712a4ea2ccd8b0734dfd1dfeceed69b05f51/weave/trace/weave_client.py#L1863), the `self` parameter (of type `openai.resources.chat.completions.completions.Completions`) is not any of the types checked in that function.~~

~~Even the `if isinstance(obj, Object):` doesn't capture this object, because the `Object` is from Pydantic module and it requires the var to be a Pydantic model.~~

~~Therefore our serialization framework cannot be triggered for this type.~~

~~I think there is a reason why we are particularly interested in certain types in this `saving` procedure, so I didn't go any further in this direction.~~

If the goal is to put down some meaningful info for the `self` parameter, and maybe there is never a need to reversely re-hydrate/instantiate this input parameter, I think directly converting it in the input handler is a way cleaner approach (also we keep an OpenAI issue within the OpenAI intergration). Hence the new iteration.

Please let me know if you agree with this approach. I will add some unit tests later after everything is confirmed.

weave/integrations/openai/openai_sdk.py
```diff
@@ -300,26 +305,65 @@ def should_use_accumulator(inputs: dict) -> bool:
     )
 
 
+def convert_completion_to_dict(obj: Any) -> dict:
+    return {
+        "client": {
```
chance-wnb, line 310: I stripped off the `_` prefix, or the UI doesn't display them.

weave/integrations/openai/openai_sdk.py
chance-wnb: The above comment is not completely correct, the serialization mechanism has nothing to do with [_save_nested_objects](https://github.com/wandb/weave/blob/5d16712a4ea2ccd8b0734dfd1dfeceed69b05f51/weave/trace/weave_client.py#L1863), but it is overseen by [to_json](https://github.com/wandb/weave/blob/afd128d919fab14047a88fa9ac54eef25f015f0d/weave/trace/serialization/serialize.py#L46)

I could get the custom serializer working, but with a lot of efforts. I implemented it in https://github.com/wandb/weave/pull/4676

1. The main issue is that a frontend viewer need to be implemented for the single parameter of `self`. It is no longer a compound value that could participle in the root ObjectViewer hierarchy organically. 
2. The serialization mechanism involves preparations for future deserialization. but in our case it doesn't seem to be needed. 

I still think a conversion originated from the integration itself is a more optimal and cleaner solution.

weave/integrations/openai/openai_sdk.py
```diff
@@ -300,26 +305,65 @@ def should_use_accumulator(inputs: dict) -> bool:
...
+            "base_url": str(obj._client._base_url),
+            "version": obj._client._version,
+        },
+    }
+
+
+def completion_instance_check(obj: Any) -> bool:
```
tssweeney, line 317: this is an odd check here IMO. why not verify the nested properties your are extracting (obj._client._base_url and obj._client._version)?

tssweeney (approved):
Approving this as it is an incremental improvement, but I do think we should think about how to get something like https://github.com/wandb/weave/pull/4676 to work. You are correct that the mechanics are not quite right for viewing/filtering in the table or the detail view... however, I feel like that is a shortcoming of the UI and we should solve this generally. Imagine a case where the user says: I have internally defined class X, how do i have weave serialize that properly? Then have it become a dictionary for display/filter reasons. This fix as is def feels like a specific patch rather than a general solution

chance-wnb:
> I do think we should think about how to get something like #4676 to work.
> that is a shortcoming of the UI and we should solve this generally
> Imagine a case where the user says: I have internally defined class X, how do i have weave serialize that properly?

100% agreed, I feel that it is more like a UI related issue. 

1. I do agree that if user's custom type is like a plot or something, we should just do it, and it should be well supported today.
2. If it is just a nested dictionary, until we have a mechanism to globally absorb each "val" field from each CustomType and put them into the root ObjectViewer, it might be super ugly today. :) (Like the current state of [PR 4676](https://github.com/wandb/weave/pull/4676) ) And I do believe after we "solve this generally", there is room to improve the work in this PR as well.
3. However the current approach as is does support both displaying and filtering well (it is organically already a part of the input dump dict), I am not sure why it is still concerning?

(See below for illustration)

image


On the contrary, if we store the input dump as 

```
{"self": {"_type": "CustomWeaveType", "weave_type": {"type": "weave.type_handlers.OpenAI.completion.OpenAICompletion"}, "load_op": "weave-trace-internal:///UHJvamVjdEludGVybmFsSWQ6NDEyOTkyODc=/op/load_weave.type_handlers.OpenAI.completion.OpenAICompletion:CbzWcD70VH38bNpETh1TlMwcJyGzWCyRh3tXabDvHeM", "val": {"client": {"base_url": "https://api.openai.com/v1/"}}}, "messages": [{"role": "user", "content": "Hello, world!"}], "model": "gpt-4o"} 
```

that would make it harder to filter, since it is deeply nested with a propertory format.

PR #4664, opened 2025-06-09
perf(weave): hardcap light filter with heavy filter

gtarpenning:
## Description

**Dangerous** optimization that moves string filter conditions into the light filter with a hardcap on the number of items returned. This means that for projects with many unmerged call parts, this query is no longer guaranteed to be correct.

Before:
```sql
WITH filtered_calls AS
  (SELECT calls_merged.id AS id
   FROM calls_merged
   WHERE calls_merged.project_id = {pb_247_4:String}
     AND ((calls_merged.op_name LIKE {pb_247_2:String}) OR (calls_merged.op_name IS NULL))
   GROUP BY (calls_merged.project_id,  calls_merged.id)
           AND ((NOT ((any(calls_merged.started_at) IS NULL))))
)
SELECT ...
FROM calls_merged
WHERE calls_merged.project_id = {pb_247_4:String}
  AND ((calls_merged.output_dump LIKE {pb_247_3:String} OR calls_merged.output_dump IS NULL))
HAVING position(JSON_VALUE(any(calls_merged.output_dump), {pb_247_0:String}), {pb_247_1:String}) > 0
LIMIT 50
OFFSET 0
```

After:
```sql
WITH filtered_calls AS
  (SELECT calls_merged.id AS id
   FROM calls_merged
   WHERE calls_merged.project_id = {pb_247_4:String}
     AND ((calls_merged.op_name LIKE {pb_247_2:String}) OR (calls_merged.op_name IS NULL))
     AND ((calls_merged.output_dump LIKE {pb_247_3:String} OR calls_merged.output_dump IS NULL))
   GROUP BY (calls_merged.project_id,  calls_merged.id)
           AND ((NOT ((any(calls_merged.started_at) IS NULL))))
           AND (position(JSON_VALUE(any(calls_merged.output_dump), {pb_247_0:String}), {pb_247_1:String}) > 0))
   LIMIT 500  --<<<<<<<
)
SELECT ...
FROM calls_merged
WHERE calls_merged.project_id = {pb_247_4:String}
  AND ((calls_merged.output_dump LIKE {pb_247_3:String} OR calls_merged.output_dump IS NULL))
HAVING position(JSON_VALUE(any(calls_merged.output_dump), {pb_247_0:String}), {pb_247_1:String}) > 0
LIMIT 50
OFFSET 0
```

## Testing

How was this PR tested?


tssweeney (changes requested):
I think this is buggy in a few ways:
1. the order by clause is now after the limit - meaning your inner query is going to be a random selection of rows, resulting in non-deterministic results
2. As a biproduct of that, any pages after the 10 * limit will result in 0 as well.
3. if you move the heavy filters inside the next layer of the query, you don't really need to do any filtering at the heavy filter side.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -650,25 +650,17 @@ def as_sql(self, pb: ParamBuilder, table_alias: str = "calls_merged") -> str:
...
-            outer_query.order_fields = self.order_fields
-        else:
-            outer_query.order_fields = self.order_fields
-            outer_query.limit = self.limit
-            outer_query.offset = self.offset
+        filter_query.order_fields = self.order_fields
+        filter_query.limit = self.limit
+        filter_query.offset = self.offset
+        # SUPER IMPORTANT: still need to re-sort the final query
+        outer_query.order_fields = self.order_fields
```
gtarpenning: rename to `select_query`?

PR #4671, opened 2025-06-09
chore(weave): update client types to allow primatives

gtarpenning:
## Description

[WB-25421](https://wandb.atlassian.net/browse/WB-25421)

Fix typing for `client.get_calls` (and others) to allow type hints to validate primetives. 

Example:

```python
import weave

client = weave.init("prod-evals-aug")
calls = client.get_calls(
    filter={"trace_roots_only": True},
    sort_by=[{"field":"started_at","direction":"desc"}],
)
```

## Testing

Existing test coverage should be sufficient, tweaked one test explicitly.

### DIFF
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -587,7 +587,7 @@ def test_calls_delete(client):
 
     assert len(list(client.get_calls())) == 4
 
-    result = list(client.get_calls(filter=tsi.CallsFilter(op_names=[call0.op_name])))
+    result = list(client.get_calls(filter={"op_names": [call0.op_name]}))
     assert len(result) == 3
 
     # should deleted call0_child1, _call0_child2, call1, but not call0
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -31,6 +31,7 @@ import pydantic
 from requests import HTTPError
 
 from weave import version
+from weave.flow.casting import CallsFilterLike, QueryLike, SortByLike
 from weave.trace import trace_sentry, urls
 from weave.trace.concurrent.futures import FutureExecutor
 from weave.trace.context import call_context
@@ -1036,14 +1037,15 @@ class WeaveClient:
     ################ Query API ################
 
     @trace_sentry.global_trace_sentry.watch()
+    @pydantic.validate_call
     def get_calls(
         self,
         *,
-        filter: CallsFilter | None = None,
+        filter: CallsFilterLike | None = None,
         limit: int | None = None,
         offset: int | None = None,
-        sort_by: list[SortBy] | None = None,
-        query: Query | None = None,
+        sort_by: list[SortByLike] | None = None,
+        query: QueryLike | None = None,
         include_costs: bool = False,
         include_feedback: bool = False,
         columns: list[str] | None = None,
@@ -1501,9 +1503,10 @@ class WeaveClient:
             )
         )
 
+    @pydantic.validate_call
     def get_feedback(
         self,
-        query: Query | str | None = None,
+        query: QueryLike | None = None,
         *,
         reaction: str | None = None,
         offset: int = 0,
@@ -1577,9 +1580,10 @@ class WeaveClient:
         )
 
     @deprecated(new_name="get_feedback")
+    @pydantic.validate_call
     def feedback(
         self,
-        query: Query | str | None = None,
+        query: QueryLike | None = None,
         *,
         reaction: str | None = None,
         offset: int = 0,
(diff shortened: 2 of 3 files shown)


tests/trace/test_weave_client.py
```diff
@@ -587,7 +587,7 @@ def test_calls_delete(client):
 
     assert len(list(client.get_calls())) == 4
 
-    result = list(client.get_calls(filter=tsi.CallsFilter(op_names=[call0.op_name])))
+    result = list(client.get_calls(filter={"op_names": [call0.op_name]}))
```
andrewtruong, line 590: I think you want to test both

weave/trace/weave_client.py
andrewtruong, line 1: Please call this file `weave/trace/casting.py`

weave/trace/weave_client.py
```diff
@@ -889,6 +891,47 @@ def __repr__(self) -> str:
...
 
+def cast_to_calls_filter(obj: Any) -> CallsFilter:
+    if isinstance(obj, CallsFilter):
+        return obj
+
+    if isinstance(obj, dict):
+        return CallsFilter(**obj)
+
+    if obj is None:
+        return CallsFilter()
```
andrewtruong: Does coercing None here make sense?

weave/trace/weave_client.py
```diff
@@ -889,6 +891,47 @@ def __repr__(self) -> str:
...
+
+def cast_to_query(obj: Any) -> Query | None:
+    if isinstance(obj, Query):
+        return obj
+
+    if isinstance(obj, dict):
+        return Query(**obj)
+
+    if obj is None:
+        return None
```
andrewtruong: same here

tests/trace/test_weave_client.py
gtarpenning: there are tons of existing examples for the other way

weave/trace/weave_client.py
gtarpenning: Hmm, the query allows "None" in place of query. I guess i'm not exactly sure how the pydantic.validate_call handles cases like that. If it sees None, will it ignore the cast function? If not, then we need these little escape hatches.

andrewtruong: but in that case your type should be
```py
CallsFilterLike = Annotated[CallsFilter | None, ...]
```

The alternative is to keep `CallsFilterLike` as you have currently defined, and then add explicit `None` handling in the function body

gtarpenning: I'm confused by exactly what the two options are. We have a function def like: 
```python
    @pydantic.validate_call
    def get_calls(
        self,
        filter: CallsFilterLike | None = None,
        sort_by: list[SortByLike] | None = None,
        query: QueryLike | None = None
    )
```

Option 1:
```python
def cast_to_query(obj: Any) -> Query | None:
    if isinstance(obj, Query):
        return obj

    if isinstance(obj, dict):
        return Query(**obj)

    raise TypeError("Unable to cast to Query")

QueryLike = Annotated[Query | None, pydantic.BeforeValidator(cast_to_query)]
```

Option 2:
```python
def cast_to_query(obj: Any) -> Query:
    if obj is None:
        return None

    if isinstance(obj, Query):
        return obj

    if isinstance(obj, dict):
        return Query(**obj)

    raise TypeError("Unable to cast to Query")

QueryLike = Annotated[Query, pydantic.BeforeValidator(cast_to_query)]
```

This doesn't make sense to me, what am I missing? @andrewtruong

andrewtruong: The options are:
```py
def cast_to_query(obj: Any) -> Query:
    if isinstance(obj, Query):
        return obj
    if isinstance(obj, dict):
        return Query(**obj)
    raise TypeError(f"Expected Query or dict, got {type(obj)}")


QueryLike = Annotated[Query, pydantic.BeforeValidator(cast_to_query)]

def get_calls(self, query: QueryLike | None = None):
    if query is None:
        query = ...
```

Then define what happens to `None` inside the `get_calls` function body (in this case we just pass `None` all the way down, but in other cases we may convert `None` into a sane default)

---

Or:
```py

def cast_to_query(obj: Any) -> Query:
    if isinstance(obj, Query):
        return obj
    if isinstance(obj, dict):
        return Query(**obj)
    if isinstance(obj, None):
        return Query()
    raise TypeError(f"Expected Query or dict, got {type(obj)}")


QueryLike = Annotated[Query | None, pydantic.BeforeValidator(cast_to_query)]

def get_calls(self, query: QueryLike = None): ...
```

Basically if you think `None` should always map to `Query()` then the second option is better.  If you think you might want to customize what `None` maps to (e.g. there might be another function that would prefer `Query(expr=...)`, then the first option is better.

gtarpenning: I think we should do the first option, fixing.

weave/trace/casting.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def cast_to_sort_by(obj: Any) -> SortBy:
+    if isinstance(obj, SortBy):
+        return obj
+
+    if isinstance(obj, dict):
+        return SortBy(**obj)
+
+    raise TypeError(f"Unable to cast to SortBy: {obj}")
```
andrewtruong, line 25: why is this different from the others (it has obj)?

gtarpenning: hmm good point, adding obj to all errors

PR #4648, opened 2025-06-06
feat(weave): Support anthropic SDK

chance-wnb:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

## Testing

How was this PR tested?

PR #4673, opened 2025-06-10
feat(weave): Support Responses api in openAI SDK

chance-wnb:
## Description

- Adds support for OpenAI's `responses.create` endpoint in the Node SDK
- Implements streaming and non-streaming response handling
- Properly processes response chunks including deltas and completed messages

## Testing

- Added test cases for both streaming and non-streaming responses
- Includes tests for different response formats
- Verified proper handling of delta chunks and final text output

### DIFF
--- a/sdks/node/src/integrations/openai.ts
+++ b/sdks/node/src/integrations/openai.ts
@@ -66,6 +66,9 @@ export const openAIStreamReducer = {
 
     return state;
   },
+  finalizeFn: () => {
+    // no-op
+  },
 };
 
 export function makeOpenAIChatCompletionsOp(originalCreate: any, name: string) {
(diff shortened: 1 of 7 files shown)


chance-wnb:
* **#4673** <a href="https://app.graphite.dev/github/pr/wandb/weave/4673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4673?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4648** <a href="https://app.graphite.dev/github/pr/wandb/weave/4648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sdks/node/src/integrations/openai.ts
```diff
@@ -129,6 +129,276 @@ export function makeOpenAIImagesGenerateOp(originalGenerate: any) {
...
+  metadata: any;
+};
+
+type StreamChunk =
+  | {type: 'response.created'; sequence_number: number; response: Response}
+  | {
+      type: 'response.in_progress';
+      sequence_number: number;
+      response: Response;
+    }
```
andrewtruong, line 171: Is it better to pull these from OpenAI instead?

sdks/node/src/integrations/openai.ts
```diff
@@ -129,6 +129,276 @@ export function makeOpenAIImagesGenerateOp(originalGenerate: any) {
...
+            }
+            outputItem.content[chunk.content_index] = chunk.part;
+          }
+        }
+        break;
+
+      case 'response.output_text.delta':
+        // Skip individual deltas since we'll get the complete text in response.output_text.done
+        // This avoids duplication as requested by the user
+        break;
```
andrewtruong: I think this works in most cases, but if the stream breaks before an `output_text.done` for any reason then you wouldn't be able to log it

sdks/node/src/integrations/openai.ts
chance-wnb: Pulling them from OpenAI would involve declaring `openai` as a dependency, which I would like to avoid. 

(I know that currently we are already doing that, but I would like to look into any opportunity to get it off unless absolutely necessary.)

sdks/node/src/integrations/openai.ts
chance-wnb: Thanks for the feedback. Now I am putting deltas in the staging area and use them to form the final message in case the stream accidentally interrupted.

sdks/node/src/integrations/openai.ts
```diff
@@ -129,6 +132,309 @@ export function makeOpenAIImagesGenerateOp(originalGenerate: any) {
...
+          }
+        }
+        break;
+
+      case 'response.output_text.delta':
+        if (!state._outputStaging) {
+          state._outputStaging = [];
+        }
+        // To prevent output stream interruption, we also temporarily store the delta in a staging area
+        state._outputStaging.push(chunk.delta);
```
chance-wnb, line 300: I noticed that the code above uses `state.choices[0].message.content += choice.delta.content;` which might have a performance impact related to repeated string concatenation. 

If this array.push() approach makes more sense and is acknowledged, I can fix the code above in a separately PR.

andrewtruong (approved):

PR #4691, opened 2025-06-10
fix(weave): get_feedback accepts custom queries in mongo format

bcsherma:
## Description

Remove the `.dict` attribute read when using a custom user query- this was causing the final query to always fail pydantic validation. The `expr_` key stores the query expression in a usable format. Also `.dict` is deprecated.

Also adds an example to the docstring for `get_feedback` and some test cases to `test_feedback.py` to verify that these queries work.

### DIFF
--- a/tests/trace/test_feedback.py
+++ b/tests/trace/test_feedback.py
@@ -8,6 +8,7 @@ from weave import AnnotationSpec
 from weave.trace.weave_client import WeaveClient, get_ref
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.errors import InvalidRequest
+from weave.trace_server.interface.query import Query
 from weave.trace_server.trace_server_interface import (
     FeedbackCreateReq,
     FeedbackQueryReq,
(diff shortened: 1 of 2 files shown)


bcsherma:
* **#4691** <a href="https://app.graphite.dev/github/pr/wandb/weave/4691?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4691?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

tests/trace/test_feedback.py
```diff
@@ -628,3 +629,85 @@ def test_feedback_replace(client) -> None:
...
+
+def test_get_feedback_with_dict_query(client) -> None:
+    """Test that get_feedback works with dict queries as shown in the docstring example."""
+    # Create some test feedback using the server API
+    project_id = client._project_id()
+
+    # Create a call to attach feedback to
+    call = client.create_call("test_op", {"input": "test"})
+    client.finish_call(call, "output")
+    call_ref_uri = f"weave:///{project_id}/calls/{call.id}"
```
chance-wnb: Just curious, without actually creating a real call, would it still work? (faking up a call id for example)

chance-wnb (approved):
Thank you!

tests/trace/test_feedback.py
bcsherma: It does actually! Seems we don't do any verification that the ref resolves. I'm going to remove the call creation from the test, good eye 👁️

PR #4703, opened 2025-06-11
fix(weave): Fixes flushing mechanics when using server cache

tssweeney:
Our WeaveClient has some hacky code that allows it to access the `call_processor` of the internal server when that server is of the `RemoteHTTPTraceServer` class. This is a violation of interface rules (since the interface does not define such a member - and nor should it!). With the caching enabled, we wrap the server in a caching layer that makes the instance check to perform the above hack invalid. This PR simply exposes the `call_processor` through the caching layer. Frankly, I don't love this. The correct solution would be for us to finally bite the bullet and split up our trace interface. But let's do the quick thing to unblock deploys.

### DIFF
--- a/weave/scripts/prerelease_dry_run.py
+++ b/weave/scripts/prerelease_dry_run.py
@@ -24,7 +24,7 @@ def main() -> None:
     client = weave.init("test-project")
     res = func(42)
 
-    client._flush()
+    client.flush()
     calls = func.calls()
 
     assert len(calls) == 1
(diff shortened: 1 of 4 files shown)


weave/scripts/prerelease_dry_run.py
```diff
@@ -24,7 +24,7 @@ def main() -> None:
     client = weave.init("test-project")
     res = func(42)
 
-    client._flush()
+    client.flush()
```
tssweeney, line 27: this is non-important - `flush` has been a public method for a while now

PR #4706, opened 2025-06-11
feat(weave): Online evals with LLM-as-a-judge scorers

neutralino1:
**Note to reviewers**
- @m-rgba @Yangyi-wandb design updates discussed this morning are not yet applied
- @tssweeney changes to the `score` function signature and prompt interpolation logic are not yet applied
- Tests are not yet passing due to the GCP outage (images fail to fetch, etc.)

Please review anyway so that this can merge as soon as possible. The feature is currently gated for admins only anyway.

## Description 
This PR adds the ability to run an LLM-as-a-judge scorer in the online monitoring worker.

Scorers can now be configured from the UI as well.

One note:
Completion calls for the scoring itself can be arbitrarily long so it needs to be done asynchronously so that it doesn't hold up the scoring worker. This means that the regular Weave client cannot be used since all `requests` calls are synchronous.
For this purpose, the worker implements a custom completion call function using `aiohttp`. All the logic to prepare and parse the response are mutualized with the baseline logic in `LLMStructuredCompletionModel`.


weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -1013,3 +1013,12 @@ const PlaygroundPageBinding = () => {
...
 };
+
+export const useEntityProject = (): {
+  entity: string;
+  project: string;
+  projectId: string;
+} => {
+  const {entity, project} = useParamsDecoded<Browse3ProjectParams>();
+  return {entity, project, projectId: `${entity}/${project}`};
+};
```
neutralino1: Tired of carrying `entity`/`project` up and down the call stack. Using this to access when needed.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/FormComponents.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export const typographyStyle = {fontFamily: 'Source Sans Pro'};
+
+export const FieldName = ({name}: {name: string}) => {
+  return (
+    <Typography sx={typographyStyle} className="mb-8 font-semibold">
+      {name}
+    </Typography>
+  );
+};
```
neutralino1, line 12: Extracted from `CreateMonitorDrawer.tsx` to avoid circular dep.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorsPage.tsx
```diff
@@ -215,3 +212,11 @@ const CallCountCell = ({
...
   );
 };
+
+export const SafeOpRef = ({opRef}: {opRef: string}) => {
+  try {
+    return <SmallRef objRef={parseRef(opRef) as WeaveObjectRef} />;
+  } catch (e) {
+    return <span>Incorrect op ref: {opRef}</span>;
+  }
+};
```
neutralino1: There is a minor flaw in our API: the `Monitor` class expects an `op_names` array, which actually expects an op ref, not just name. This doesn't matter when monitors are created via the UI, but is slightly inconsistent when creating via the SDK (`op_name` should be called `op_refs`). I will add validation to `Monitor` to make sure this doesn't happen.

weave/flow/scorer.py
```diff
@@ -36,8 +36,6 @@ def score(self, *, output: Any, **kwargs: Any) -> Any:
     def summarize(self, score_rows: list) -> Optional[dict]:
         return auto_summarize(score_rows)
 
-
-class BuiltInScorer(Scorer):
```
neutralino1, line 40: As mentioned by @tssweeney, this was not really necessary.

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
```diff
@@ -156,6 +156,47 @@ def predict(
                 "You must call `weave.init(<project_name>)` first, to predict with a LLMStructuredCompletionModel"
```
neutralino1, line 159: The changes in this file are made so that the logic is more modular and can be reused in the scoring worker.

Why not just use `model.predict` in the scoring worker?
To support high volume of calls we need the completion calls to be made asynchronously. `current_client.server.completion_create` is currently sync.
A better implementation in the worker would be to simply patch the model completion logic. At this time I do it the other way around, the scoring worker uses the modular logic here to replicate the logic.

weave/workers/helpers.py
```diff
@@ -0,0 +1,202 @@
...
+                # This is required for the trace server to accept the request
+                "Authorization": "Basic Og==",
+            },
+            ssl=False,
+        ) as response:
+            return CompletionsCreateRes.model_validate(await response.json())
+
+
+# EVERYTHING BELOW THIS LINE IS FOR TESTING PURPOSES ONLY
+# WILL BE YANKED BEFORE MERGING
```
neutralino1: Everything below this will be removed before merging.

weave/workers/helpers.py
```diff
@@ -0,0 +1,202 @@
+import logging
```
neutralino1: This file contains a few functions to query Gorilla (to convert internal IDs to external) and the completion endpoint of the trace server. This code should be organized into separate modules. Will do after launch.

weave/workers/weave_scorer.py
```diff
@@ -68,51 +78,13 @@ def get_trace_server() -> ClickHouseTraceServer:
...
+# Encapsulates a few data that are received from the server and need to be passed around
 class ActiveMonitor(TypedDict):
-    """Type definition for the output of get_active_monitors."""
-
     monitor: Monitor
     internal_ref: InternalObjectRef
     wb_user_id: Optional[str]
 
 
-class MonitorsCache:
```
neutralino1, line 79: Removing in favor of 
```python
from cachetools import TLRUCache, cached

@cached(cache=TLRUCache(maxsize=1000, ttu=lambda key, value, now: now + 60))
def ...
```

weave/workers/weave_scorer.py
```diff
@@ -201,10 +193,21 @@ async def get_filtered_calls(
...
+                        "$not": [
+                            {"$eq": [{"$getField": "ended_at"}, {"$literal": None}]}
+                        ]
+                    },
+                    {
+                        "$not": [
+                            {"$eq": [{"$getField": "started_at"}, {"$literal": None}]}
+                        ]
+                    },
+                ]
```
neutralino1: This is to make sure all call parts have been merged into the merged table.

I have ideas for how to make this more efficient and reliable, will implement after launch.

weave/workers/weave_scorer.py
```diff
@@ -297,6 +307,33 @@ def _do_score_call(scorer: Scorer, call: Call, project_id: str) -> tuple[str, An
     return call_start_res.id, result
 
 
+async def _do_score_call_llm_as_a_judge(
```
neutralino1: This is where I reuse the modularized logic from `LLMStructuredCompletionModel` and replace the completion call logic to make it async.

weave/workers/helpers.py
```diff
@@ -0,0 +1,202 @@
...
+    request: CompletionsCreateReq, user_id: str
+) -> CompletionsCreateRes:
+    username = get_username_from_user_id(user_id)
+    try:
+        trace_server_base_url = os.environ["WEAVE_TRACE_SERVER_BASE_URL"]
+    except KeyError:
+        raise KeyError("WEAVE_TRACE_SERVER_BASE_URL is not set")
+
+    async with aiohttp.ClientSession() as session:
+        async with session.post(
```
neutralino1: Using `aiohttp` instead of `requests` to make calls asynchronous.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -22,80 +22,136 @@ import {
...
+]);
+
+export const MonitorDrawerRouter = (props: CreateMonitorDrawerProps) => {
+  if (props.monitor) {
+    return (
+      <CreateMonitorDrawerWithScorers {...props} monitor={props.monitor} /> // Repeating monitor to appease type checking
+    );
+  }
+  return <CreateMonitorDrawer {...props} />;
+};
```
neutralino1: This pattern is to enable fetching scorers from the server when editing an existing monitor (vs. creating a new one from scratch).

neutralino1:
@cursor-com first bug was fixed.

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
tssweeney: if you want to do this, please use a context var. I really try to reduce URL-based state management as much as possible to ensure portability

tssweeney: Fundamental design theory: pretty early in the component hierarchy should be a URL <> state conversion, and ideally all children should no know about URLs.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorsPage.tsx
tssweeney: nit: try/catch should go around the method, not component render. Also, you mihgt want to use `parseRefMaybe`

weave/workers/helpers.py
```diff
@@ -0,0 +1,203 @@
...
+    except KeyError:
+        raise KeyError("WANDB_INTERNAL_SERVICE_TOKEN is not set")
+
+    return f"{secret_name}:{internal_service_token}"
+
+
+INTERNAL_SERVICE_TOKEN_PREFIX = "X-Wandb-Internal-Service"
+
+
+def get_authenticated_client(impersonate_as: Optional[str] = None) -> Client:
```
tssweeney: for security reasons, I would define a context var that is a Client type that is expected to be set externally. Then the worker setup in our private repository does the work to extract and initialize the authenticated client. Seems a bit risky to put internal authentication details in OSS

weave/workers/helpers.py
```diff
@@ -0,0 +1,203 @@
...
+    try:
+        trace_server_base_url = os.environ["WEAVE_TRACE_SERVER_BASE_URL"]
+    except KeyError:
+        raise KeyError("WEAVE_TRACE_SERVER_BASE_URL is not set")
+
+    async with aiohttp.ClientSession() as session:
+        async with session.post(
+            f"{trace_server_base_url}/completions/create",
+            json=request.model_dump(),
+            headers={
```
tssweeney: yeah, this stuff really should be dependency injected

weave/workers/weave_scorer.py
tssweeney: `callsQuery` will always filter out items without start

weave/workers/weave_scorer.py
```diff
@@ -415,21 +450,41 @@ async def _process_monitor(active_monitor: ActiveMonitor) -> None:
...
+
         with get_trace_server().call_batch():
-            await asyncio.gather(
-                *[
-                    apply_scorer(
-                        monitor_internal_ref, scorer, call, project_id, wb_user_id
+            scoring_tasks = []
+            for call in calls:
+                # Apply sampling rate
+                if random.random() > monitor.sampling_rate:
```
tssweeney: isn't this supposed to be `<=` ?

neutralino1: I don't think so.
If your rate is `0.1`, any random pick above 0.1 (i.e. 90% of picks) should result in the call being skipped.

weave/workers/helpers.py
neutralino1: We currently have no worker setup in the private repo. This seems like a bit too much of a lift for launch.

What if I made `INTERNAL_SERVICE_TOKEN_PREFIX` a secret env var, so that it doesn't appear here?
Then as a fast-follow I can set up a worker shell in the private repo to inject this client?

weave/workers/weave_scorer.py
tssweeney: 👍

weave/workers/helpers.py
tssweeney: I would at least do that. IMO, this is really not great to have floating around OSS. Nothing in the OSS repo deals with auth / gorilla interactions. Even as is (now in a public commit) it is outed, merging this to master pretty much locks it in for good in history. 

In this PR: https://github.com/wandb/core/pull/29039/files you are setting calling the `wevae_scorer` entry point directly. I don't see why `main` can't be parameterized to support this injection and in core you have a small script that does the injection.

If you feel this is an ok sec risk, I would like a second opinion.

neutralino1: No I agree, this is leaking too much. I'll figure something out.

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -257,96 +258,98 @@ const MainPeekingLayout: FC = () => {
...
-          sx={{
-            flex: '1 1 auto',
-            width: '100%',
-            height: '100%',
-            display: 'flex',
-            overflow: 'hidden',
-            flexDirection: 'row',
-            alignContent: 'stretch',
-          }}>
+    <EntityProjectProvider entity={params.entity!} project={params.project!}>
```
neutralino1, line 261: This is the only change. Everything below is just indentation.

chance-wnb, line 261: If I am not mistaken, the code base as its currently state, uses `useParamsDecoded()` to get the project name and entity name, which has many prior usages. 

This hook should be accessible from any hierarchy level. So having a provider for this doesn't seem to have any superior benefits.

However you could still have a version of `useEntityProject` that does not use a context.  

You can make a call.

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -733,7 +736,7 @@ const CallPageBinding = () => {
 
 const MonitorsPageBinding = () => {
   const {entity, project} = useParamsDecoded<Browse3TabParams>();
```
chance-wnb: This line seems obsolete

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -22,80 +23,134 @@ import {
...
+
+const CreateMonitorDrawerWithScorers = (
```
chance-wnb: It seems this component is active when an existing monitor is selected, if so, should it named more like `EditMonitorDrawerWithScorers` ? 

If this makes sense to you, you might also want to resolve the names like `CreateMonitorDrawerProps' -> 'MonitorDrawerProps` (Props), `CreateMonitorDrawer` -> 'MonitorDrawer` (Component) and the file name as well.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -22,80 +23,134 @@ import {
...
+  });
+  const [filterModel, setFilterModel] = useState<GridFilterModel>(
+    queryToGridFilterModel(monitor?.val['query']) || {
+      items: [],
+    }
+  );
   const {useCalls} = useWFHooks();
-  const [scorers, setScorers] = useState<{name: string; refUri?: string}[]>([]);
-  const [active, setActive] = useState<boolean>(false);
+  const [scorers, setScorers] = useState<ObjectVersionSchema[]>(
```
chance-wnb: Just FYI, I previously introduced https://github.com/streamich/react-use/blob/master/docs/useList.md to the repository. 

which might help reduce some clutter and make the code more semantical.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -400,9 +516,11 @@ export const CreateMonitorDrawer = ({
                       <FieldName name="Scorers" />
                       <Autocomplete
                         multiple
-                        options={Object.values(BuiltInScorers)}
+                        options={Array.from(SCORER_FORMS.keys())}
```
chance-wnb, line 510: This value seems static. Repeated calculate it adds some burden to Autocomplete component.

You could get around it with inline useMemo:
```
options={useMemo(Array.from(SCORER_FORMS.keys()), [])}
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -400,9 +516,11 @@ export const CreateMonitorDrawer = ({
...
                         sx={{width: '100%'}}
-                        value={scorers.map(scorer => scorer.name)}
+                        value={scorers.map(
```
chance-wnb, line 512: Again, scorers.map creates a new instance each time component is rendered. You can use useMemo()

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -400,9 +516,11 @@ export const CreateMonitorDrawer = ({
...
+                          scorer => scorer.objectId || scorer.val['_type']
+                        )}
```
chance-wnb, line 514: Also I notice that reselecting the same scorer cancel the selection. It might be by design for now, but I wonder how you are going to add the same scorer multiple times in the future. (The underlying idea of your code seem to suggest multiple same LLM scorers would be supported)

Just a friendly reminder to confirm, nothing need to be changed for now.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
```diff
@@ -22,80 +23,134 @@ import {
...
-    setActive(monitor?.val['active'] || false);
-    setFilterModel(
-      queryToGridFilterModel(monitor?.val['query']) || {
-        items: [],
-      }
-    );
-  }, [monitor, setMonitorName]);
+
+  useEffect(() => {
+    setScorerValids(currentValids => currentValids.slice(0, scorers.length));
```
chance-wnb: Does it make sense to do this in the Autocomplete onChange event?

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
```diff
@@ -156,6 +156,47 @@ def predict(
...
+        req = self.prepare_completion_request(
+            project_id=f"{current_client.entity}/{current_client.project}",
+            user_input=user_input,
+            config=config,
+            **template_vars,
+        )
+
+        # 5. Call the LLM API
+        try:
+            api_response = current_client.server.completions_create(req=req)
```
chance-wnb, line 171: In the PR description you said "This means that the regular Weave client cannot be used since all requests calls are synchronous"

Isn't the `current_client` the weave client? I am a bit confused.

chance-wnb (commented):
Overall looks great! I don't think there are many blockings issues. 

I can help review again after you read my comments.

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
tssweeney: I asked for this change actually. I am always trying to prevent URL-based implementations as much as possible. There should only be one layer that decodes the URL (and ideally very few URL modifiers) such that views can more easily be embedded into different parts of the application that do not rely on their mount location. 

For example: let's say we want to embed weave components into reports, we don't want to rely on the url scheme as much as possible because reports can be mounted outside the entity/project pattern.

IMO anytime a component interacts with the URL is a point of unfortunate coupling that should be minimized

weave/scorers/llm_as_a_judge_scorer.py
```diff
@@ -0,0 +1,22 @@
...
+from weave.flow.scorer import Scorer
+from weave.trace.objectify import register_object
+from weave.trace.op import op
+from weave.trace_server.interface.builtin_object_classes.llm_structured_model import (
+    LLMStructuredCompletionModel,
+)
+
+
+@register_object
+class LLMAsAJudgeScorer(Scorer):
```
tssweeney, line 12: <3 love this

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
chance-wnb: Yeah, that sounds good.

neutralino1: @tssweeney said he was not comfortable using `useParamsDecoded` because it relies on URL params. 
Ideally URL params are read once then saved as a state for the rest of the app. This is what this provider does. The input parameters (`params.entity` and `params.project`) are extracted using `useParamsDecoded`, then stored as a state within `EntityProjectProvider`.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
neutralino1: I don't see any mentions of `useList` anywhere in the codebase 🤔

chance-wnb: True, the `react-use` library was added as a npm dependency, but we don't have much usage on that yet. (So far I have only used useAsync ).

Just a friendly reminder that it is available, you can make a call whether to use it or not. ( I am pretty sure you know about it in the `A***T**` development before :) )

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/CreateMonitorDrawer.tsx
neutralino1: We're removing the autocomplete for now, hard-coding to single scorer.

weave/trace_server/interface/builtin_object_classes/llm_structured_model.py
neutralino1: Yes, this is in the SDK class. In the worker, we replace this a custom logic. See https://github.com/wandb/core/pull/30189 if you're interested.

PR #4711, opened 2025-06-11
chore(ui): Add line-clamp to playground settings

m-rgba:
## Description

| Before | After |
| -------- | ------- |
| image | image |

- Added a line-clamp to the settings header in Playground.
- Note: The `[&>span]:` adds the line-clamp to the tooltip's span which I found was necessary to clamp it properly (ellipses, not only overflow-hidden) - since we don't have styling props on the tooltip itself and didn't want to modify it for this small-ish use case.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundSettings/PlaygroundSettings.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundSettings/PlaygroundSettings.tsx
@@ -59,11 +59,11 @@ export const PlaygroundSettings: React.FC<PlaygroundSettingsProps> = ({
     <div className="relative flex h-full w-[320px] shrink-0 flex-col border-l border-moon-250 pb-4">
       <div className="flex items-center justify-between gap-8 border-b border-moon-250 px-16 py-8">
         {/* Header */}
-        {/* [&>span]: adds line-clamp to the child span from the tooltip*/}
-        <div className="line-clamp-1 flex items-center gap-8 [&>span]:line-clamp-1">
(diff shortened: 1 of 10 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/PlaygroundSettings/PlaygroundSettings.tsx
```diff
@@ -59,7 +59,8 @@ export const PlaygroundSettings: React.FC<PlaygroundSettingsProps> = ({
     <div className="relative flex h-full w-[320px] shrink-0 flex-col border-l border-moon-250 pb-4">
       <div className="flex items-center justify-between gap-8 border-b border-moon-250 px-16 py-8">
         {/* Header */}
-        <div className="flex items-center gap-8">
+        {/* [&>span]: adds line-clamp to the child span from the tooltip*/}
+        <div className="line-clamp-1 flex items-center gap-8 [&>span]:line-clamp-1">
```
jamie-rasmussen: I think there might be a slightly simpler approach here by specifying `noTriggerWrap` on the `<Tooltip>` and `w-full` on this flex container.

m-rgba: Yep that works and is much simpler - ty updated 👍

m-rgba:
> Can probably also get rid of the line-clamp-1 since you have nowrap

So I noticed without the line-clamp the dismiss settings button is being pushed away:
without:
image
with:
image

I think it's specifically the overflow hidden which is stopping that from happening - just updated it to `overflow-hidden` instead so that it's more explicit in what's happening.

PR #4715, opened 2025-06-11
chore(ui): Leaderboard: Edit buttons in peek / delete controls

m-rgba:
## Description

image

image

Mostly quality-of-life changes around editing and deletingLeaderboards.

- Added delete controls for the Leaderboard objects.
- Added Edit / Delete buttons to both peek and expanded Leaderboard page (quality-of-life: saw users having problems understanding to open to full screen before they could edit their Leaderboard).


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/LeaderboardPage/LeaderboardDeleteButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import {
+  useClosePeek,
+  useWeaveflowRouteContext,
+  WeaveflowPeekContext,
+} from '../../context';
+import {DeleteModal} from '../common/DeleteModal';
+import {useWFHooks} from '../wfReactInterface/context';
+
+export const DeleteLeaderboardButton: React.FC<{
```
tssweeney, line 13: can we please remove the delete button? This is such a destructive operation and could very easily be accidentally clicked and erase a bunch of config work. We can separate adding a delete button from the edit button in peak

m-rgba: I see what you mean, I pulled it off of all of the happy-path areas.

How would you feel about keeping one of them on the edit expanded page in the bottom left corner of the form? The button also uses our confirm confirmation modal.
I feel like we should have a delete button somewhere in this expanded area (so the user doesn't need to go through objects to remove their leaderboards), but I'm ok taking it out of this PR if you'd like to chat through it more.

tssweeney: if it has a confirmation modal, then I am fine with it. In general, i think this action should be very challenging to do and hidden. But as long as it is not a single click, i am fine

PR #4716, opened 2025-06-11
chore(weave): clarify role of model id and idPlayground

jamie-rasmussen:
## Description

Moving to being more explicit about the meaning of different fields in our model catalog data and using them consistently. This fixes a bug Deok reported with the wrong value showing up in the Playground select widget, and I think should get the correct model string to the Trace Server for it to pass on to the Inference Service.

## Testing

How was this PR tested?


hhuangwb (approved):
lgtm

PR #4720, opened 2025-06-11
chore(weave): a couple more datetime filter tweaks

gtarpenning:
## Description

- reset invalid filter state when a correct filter is selected from dropdown
- don't reload the filter if calendar date is the same as current filter
- refactor cancel debounce timer into update function

## Testing

branch
datetime-tweak-branch


gtarpenning:
> Looks good!
> 
> Totally not required but the double negative for `setIsInvalid(false)` makes it slightly harder to follow. `setIsValidDate(true)` might be a bit clearer

I'm going to punt on this but that makes perfect sense. The reason it's a double negative is because the action state is when its "invalid". We are specifically tracking "invalid" states, that's when we want to do something, in this case show an error. But that feedback is noted and when I do another pass on this file I will try to work that in.

PR #4730, opened 2025-06-12
chore(weave): update model catalog data

jamie-rasmussen:
## Description

Update to latest model catalog data from Deok's spreadsheet. Small tweaks to show active/total parameter count tiles on detail page instead of like and download counts (which are harder to keep updated).

## Testing

How was this PR tested?


weave-js/src/components/PagePanelComponents/Home/Browse3/inference/modelsFinal.json
```diff
@@ -97,87 +121,20 @@
...
-      "downloadsHuggingFace": 3590321
+      "modalities": [
+        "Text"
+      ],
+      "likesHuggingFace": 501,
+      "downloadsHuggingFace": 255012,
+      "license": "mit"
     }
   ]
-}
```
gtarpenning, line 183: should this have a newline?

jamie-rasmussen: Won't break anything without but I've fixed it

gtarpenning (approved):
the util functions make this nice to review, if we continue to build off them we might want docstrings at some point.

PR #4732, opened 2025-06-12
fix(weave): sanitize memory addresses to prevent dupe ops

gtarpenning:
## Description

[WB-25498](https://wandb.atlassian.net/browse/WB-25498)

Turn changing memory addresses into stable `0x000000000` so multiple instantiations of the same function at different memory addresses doesn't create multiple versions of ops. This impacts all `litellm` calls! 

Turn `<Function object at 0x10c349010>`  --> `<Function object at 0x000000000>`

_We could also move these two regex operations into one:_ `_try_replace_memory_address` 

## Testing

added test (fails without my change)

[WB-25498]: https://wandb.atlassian.net/browse/WB-25498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace/serialization/op_type.py
+++ b/weave/trace/serialization/op_type.py
@@ -34,7 +34,8 @@ logger = logging.getLogger(__name__)
 WEAVE_OP_PATTERN = re.compile(r"@weave\.op(\(\))?")
 WEAVE_OP_NO_PAREN_PATTERN = re.compile(r"@weave\.op(?!\()")
 
-MEMORY_ADDRESS_PATTERN = re.compile(r"0x[0-9a-fA-F]{9}>", re.ASCII)
+# Memory address with at least 4 charaters (decrease false positives)
+MEMORY_ADDRESS_PATTERN = re.compile(r"0x[0-9a-fA-F]{4,}>", re.ASCII)
 
 CODE_DEP_ERROR_SENTINEL = "<error>"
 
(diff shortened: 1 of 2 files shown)


weave/trace/serialization/op_type.py
```diff
@@ -466,6 +468,18 @@ def _get_code_deps(
     return {"import_code": import_code, "code": code, "warnings": warnings}
 
 
+MEMORY_ADDRESS_PATTERN = re.compile(r"0x[0-9a-fA-F]{9}>")
```
andrewtruong: please move const to top of file

andrewtruong: also nit: you can optimize this by passing the `re.ASCII` flag

weave/trace/serialization/op_type.py
```diff
@@ -466,6 +471,15 @@ def _get_code_deps(
     return {"import_code": import_code, "code": code, "warnings": warnings}
 
 
+def _has_memory_address(obj: Any) -> bool:
+    return isinstance(obj, str) and MEMORY_ADDRESS_PATTERN.search(obj) is not None
+
+
+def _replace_memory_address(json_val: str) -> str:
+    """Turn <Function object at 0x10c349010> into <Function object at 0x000000000>"""
+    return MEMORY_ADDRESS_PATTERN.sub("0x000000000>", json_val)
```
jamie-rasmussen: Optional - I think you could use a lambda for the replacement to ensure the replacement address was the same length as the input one.

gtarpenning: oh, yeah that's a nice idea thx.

PR #4733, opened 2025-06-12
fix(weave): media panel image night mode

onx2:
# WB-16745

## Description

Renders the image in an iframe to avoid night mode


onx2:
CSS filter applies at the compositing stage of rendering — after layout and painting, but before the final pixels are rendered to screen. `ShadowDOM` and `<iframe>` encapsulate DOM and CSS, but not pixel-level rendering. Filters affect the visual output, not the underlying document structure.

**DOM Construction** - HTML is parsed into the DOM.
**Style Calculation** -CSS is parsed and computed into a style tree.
**Layout** - The positions and sizes of elements are calculated.
**Painting** - Elements are turned into individual paint instructions.
**Compositing** - Painted layers are combined, often on the GPU, into a single final image (bitmap) before it's displayed.

CSS filter is a compositing-time effect. That means it operates on the visual layer output, applying after everything is painted, including content from iframes or shadow DOM.

**TL;DR;** We literally can't do anything about night mode unless we do some major refactoring so we'll have to live with inaccurate "revert night mode" calculations.

PR #4736, opened 2025-06-12
chore(weave): add model id copy button to details page

jamie-rasmussen:
## Description

Previously added model ID copy button to model tile, design requested adding it to details page too.
Internal slack: https://weightsandbiases.slack.com/archives/C08RU04P36G/p1749746145399799?thread_ts=1749683637.172539&cid=C08RU04P36G

Recommend "Hide Whitespace". I also put in some better conditional rendering if `idPlayground` is missing, which is allowed by current type definitions but shouldn't come up in practice with our starting data.

Screenshot 2025-06-12 at 3 53 27 PM

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/inference/ModelDetailsLoaded.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/inference/ModelDetailsLoaded.tsx
@@ -176,7 +176,12 @@ export const ModelDetailsLoaded = ({
     codeExample = codeExample.replace('{model_id}', model.idPlayground);
   }
 
-  const onClickCopy = useCallback(() => {
+  const onClickCopyModelId = useCallback(() => {
+    copyToClipboard(model.idPlayground ?? '');
+    toast('Copied to clipboard');
+  }, [model.idPlayground]);
+
+  const onClickCopyCode = useCallback(() => {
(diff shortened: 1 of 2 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/inference/ModelDetailsLoaded.tsx
```diff
@@ -198,6 +203,15 @@ export const ModelDetailsLoaded = ({
           </div>
         </div>
         <div className="flex items-center gap-8">
+          {model.idPlayground && (
+            <Button
+              size="large"
+              icon="copy"
+              variant="secondary"
+              onClick={onClickCopyModelId}
+              tooltip="Copy ID for API use to clipboard"
```
gtarpenning, line 212: re-reading this, the "use" is sort of awkward because of this word's other more common use. ;)

gtarpenning: youz vs yoos

jamie-rasmussen: I was having trouble coming up with a non-awkward tooltip string, I'll note it and push the decision to design.

PR #4740, opened 2025-06-12
chore(dev): if run id verification fails for a trace do not dr op it, drop the run field

bcsherma:
## Description

In `call_start` if the `wb_run_id` field is passed but the run does not exist or is in a project other than the call's, the server issues a 500 error. This PR modifies the behavior so that instead of dropping the call we drop the run id and step from the call and insert it. The validation error is logged and will be captured by dd.

### DIFF
--- a/weave/trace_server/external_to_internal_trace_server_adapter.py
+++ b/weave/trace_server/external_to_internal_trace_server_adapter.py
@@ -1,4 +1,5 @@
 import abc
+import logging
 import typing
 from collections.abc import Iterator
 from typing import Callable, TypeVar
@@ -38,6 +39,8 @@ class IdConverter:
 
 A = TypeVar("A")
 B = TypeVar("B")
+
+logger = logging.getLogger(__name__)
 
 
 class ExternalTraceServer(tsi.TraceServerInterface):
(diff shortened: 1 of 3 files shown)


bcsherma:
* **#4740** <a href="https://app.graphite.dev/github/pr/wandb/weave/4740?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -108,7 +111,12 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
     def call_start(self, req: tsi.CallStartReq) -> tsi.CallStartRes:
         req.start.project_id = self._idc.ext_to_int_project_id(req.start.project_id)
         if req.start.wb_run_id is not None:
-            req.start.wb_run_id = self._idc.ext_to_int_run_id(req.start.wb_run_id)
+            try:
+                req.start.wb_run_id = self._idc.ext_to_int_run_id(req.start.wb_run_id)
+            except Exception as e:
```
gtarpenning, line 117: can we catch the exact exception? Should be only one right?

PR #4753, opened 2025-06-13
chore(weave): When initializing client in a new project, flush the existing client first

andrewtruong:
Resolves: https://wandb.atlassian.net/browse/WB-25497

This prevents calls from getting lost in case the first client gets cleaned up


gtarpenning (approved):
Might be able to unit test this

PR #4767, opened 2025-06-13
chore(weave): update prod inference service hostname

jamie-rasmussen:
## Description

Now that DNS is setup, use the preferred hostnames in Trace Server.

## Testing

How was this PR tested?


gtarpenning (approved):
Nice

PR #4773, opened 2025-06-13
fix(weave): clear weave client context when notebook cells are interrupted

bcsherma:
## Description
This PR updates the op code to apply the same handling we do for `Exception` to all instances of `BaseException`.

Error handling for ops only captures `Exception`s, but we need to also handle `KeyboardInterrupt`, `SystemExit`, and `GeneratorExit` which, like `Exception`, inherit from `BaseException`. Traces experiencing these would be left in a running state forever, and if the client context is re-used, eg in a notebook, all new calls will inherit the call stack from the point of failure.

### DIFF
--- a/weave/trace/context/call_context.py
+++ b/weave/trace/context/call_context.py
@@ -22,8 +22,67 @@ logger = logging.getLogger(__name__)
 
 _tracing_enabled = contextvars.ContextVar("tracing_enabled", default=True)
 
+# Interactive session detection (cached for performance)
+_is_interactive_session: bool | None = None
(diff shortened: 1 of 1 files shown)


bcsherma:
* **#4773** <a href="https://app.graphite.dev/github/pr/wandb/weave/4773?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4773?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #4774, opened 2025-06-14
fix(weave): handle `KeyboardInterrupt`, `SystemExit` within ops

bcsherma:
## Description

Error handling for ops only captures Exceptions, but we need to also handle KeyboardInterrupt, SystemExit which, like Exception, inherit from BaseException. Traces experiencing these would be left in a running state forever, and if the client context is re-used, eg in a notebook, all new calls will inherit the call stack from the point of failure.

Changing the blanket catch from `Exception` to `BaseException` will catch the things we want but will also catch `GeneratorExit`, which gets raised in healthy generator exits, kind of a weird implementation detail of python.

### DIFF
--- a/weave/trace/op.py
+++ b/weave/trace/op.py
@@ -511,6 +511,9 @@ def _call_sync_func(
         if __should_raise:
             raise
         return None, call
+    except (SystemExit, KeyboardInterrupt) as e:
+        finish(exception=e)
+        raise
 
     res = box.box(res)
     try:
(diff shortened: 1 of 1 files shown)


bcsherma:
* **#4774** <a href="https://app.graphite.dev/github/pr/wandb/weave/4774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

chance-wnb (approved):
Thank you!

PR #4782, opened 2025-06-16
chore(weave): raise error on order by storage field

gtarpenning:
## Description

[WB-25518](https://wandb.atlassian.net/browse/WB-25518)

We already prevent filtering by storage fields, now prevent ordering. 

## Testing

- adds unit test

### DIFF
--- a/weave/trace_server/calls_query_builder/calls_query_builder.py
+++ b/weave/trace_server/calls_query_builder/calls_query_builder.py
@@ -461,6 +461,8 @@ class CallsQuery(BaseModel):
         return self
 
     def add_order(self, field: str, direction: str) -> "CallsQuery":
+        if field in DISALLOWED_FILTERING_FIELDS:
+            raise ValueError(f"Field {field} is not allowed in ORDER BY")
         direction = direction.upper()
         if direction not in ("ASC", "DESC"):
             raise ValueError(f"Direction {direction} is not allowed")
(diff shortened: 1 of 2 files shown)


weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -461,6 +461,8 @@ def set_hardcoded_filter(self, filter: HardCodedFilter) -> "CallsQuery":
         return self
 
     def add_order(self, field: str, direction: str) -> "CallsQuery":
+        if field in DISALLOWED_FILTERING_FIELDS:
+            raise ValueError(f"Field {field} is not allowed in ORDER BY")
```
neutralino1, line 465: Is this error message propagated back to SDK users?

gtarpenning: we don't handle _any_ errors correctly on the streaming endpoint... my personal vendetta bug, but it's a big lift to fix.

neutralino1: So here users will simply see some sort of 40x failure?

gtarpenning: hmm, yeah...

neutralino1 (approved):
Can you add a JIRA ticket to properly propagate error messages to users in the SDK?

gtarpenning:
> Can you add a JIRA ticket to properly propagate error messages to users in the SDK?

this should fall under a ticket that Chance currently has assigned to him

PR #4783, opened 2025-06-16
fix(weave): Use last known state when useFile skip is true

zbirenbaum:
## Description

Currently even if a file has been loaded the result will be null if skip is passed as true. This just modifies the hook to return the last known state (which will be the same if the file has never been loaded). This makes it a bit easier to manage state for components using this hook.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
@@ -989,7 +989,11 @@ const useFileContent = (
   return useMemo(() => {
     // Just submit the last known or initial state
     if (params.skip) {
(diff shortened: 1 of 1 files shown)


zbirenbaum:
This is the definition of seems benign but could have side effects so don't merge until after FC.

neutralino1 (approved):

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/tsDataModelHooks.ts
```diff
@@ -987,8 +987,13 @@ const useFileContent = (
   }, [getTsClient, params.entity, params.project, params.digest, params.skip]);
```
zbirenbaum, line 987: I think params.skip here might mess things up and cause it to recalculate when you skip based on if the data is already loading or not. Could someone else confirm?

neutralino1:
@zbirenbaum Merge or close?

PR #4787, opened 2025-06-16
feat(weave): Add generic content to weave

zbirenbaum:
## Description

1. Removes `weave.File` and adds the more general `weave.Content` in its place
2. Regex parsing for annotated values

## Testing

Tested fairly extensively

# Content API Specification

### Overview

The Content class is a generic content handler that wraps binary data with metadata for different file types and sources.

TLDR - It provides a unified interface (which serializers can convert objects to and from) and enables stuff like this:
```
# Output and mp3_path will both show up as an MP3 object in the UI
@weave.op
def read_mp3_file(mp3_path: Annotated[str, Content]) -> Annotated[bytes, Content]:
    return Path(path).read_bytes()
file_bytes = read_mp3_file # This is bytes
```

### Content[T]

Generic factory class for handling various content inputs (files, bytes, base64, objects).

Constructor

Content(input: Any, type_hint: str | None = None, /, **kwargs)

Parameters:

 - input: Source data (Path, str, bytes, or object)
 - type_hint: Optional MIME type or file extension hint
 - **kwargs: Additional optional fields and metadata (filename, path, extension, mimetype, size, extra)

Input Types:

 - Path/File path string: Loads file from filesystem
 - Base64 string: Decodes base64 data
 - Bytes: Raw binary data
 - Objects (Expermental): Uses to_bytes methods or custom handler

Properties

 - data: bytes - Raw binary content
 - metadata: dict[str, Any] - All metadata excluding data (includes all fields set in `kwargs['extra']`)
 - size: int - Content size in bytes
 - filename: str - File name
 - extension: str - File extension
 - mimetype: str - MIME type
 - original_path: str | None - Source file path (if applicable)

Methods

 - open() -> bool - Open file with system default application (Only for files with a path currently, universal support in followup)
 - save(dest: str | Path) -> None - Save content to specified path

Factory Functions
  - create_bytes_content(bytes, **kwargs) -> BaseContentHandler
  - create_file_content(str | Path, **kwargs) -> BaseContentHandler
  - create_b64_content(str | bytes, **kwargs) -> BaseContentHandler

# Usage Examples

### From file
```python
# Filename will be parsed as file.jpg
# Path will be parsed as file.jpg
# mimetype will be parsed from filename
content = Content("path/to/file.jpg")
```

### From base64
```python
# Filename will be a generated datetime string
# mimetype will be parsed from decoded bytes by libmagic
content = Content("path/to/file.jpg")
content = Content(base64_string)
```

### From bytes with metadata
```
# Filename will be "image.png"
# mimetype will be accepted as is
content = Content(data, filename="image.png", mimetype="image/png")
```

### Access properties
```
print(content.size, content.mimetype, content.filename)
content.save("output.jpg")
```

### From bytes with extra metadata
```
content = Content(data, extra={ "resolution": "1200x1000" })
print(content.metadata.get("resolution")
```

### Infer the mime_type from the data using libmagic
```
@weave.op
def test_annotated_without_hint(path: Annotated[str, Content]) -> Annotated[bytes, Content]:
    return Path(path).read_bytes()
```

### Pass an extension or a mimetype to the 'type_hint' parameter through the annotation
```
@weave.op
def test_annotated_with_hint(path: Annotated[str, Content[Literal["pdf"]]]) -> Annotated[bytes, Content[Literal["application/pdf"]]]:
    return Path(path).read_bytes()
```


weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,250 @@
...
+            self.content_handler = create_bytes_content(input, **content_args)
+        else:
+            raise TypeError(f"Unsupported input type: {type(input)}")
+
+    @property
+    def metadata(self) -> dict[str, Any]:
+        return self.content_handler.metadata
+
+    @property
+    def input_type(self) -> str | None:
```
zbirenbaum: If there's anything in this PR that I would guess is up for debate it's how `input_type` and `input_category` are being handled. I made a design choice here after trying it a few different ways and finding this one worked best, but I'm open to input.

bcsherma (commented):
Is it necessary to remove `File`? This would break user code depending on `File` if they upgrade their SDK. I think it would be safer to deprecate `File` and point people to `Content`.

zbirenbaum:
> Is it necessary to remove `File`? This would break user code depending on `File` if they upgrade their SDK. I think it would be safer to deprecate `File` and point people to `Content`.

We never actually exposed `File` via docs or announcement. Unless a user was reading through our codebase they would never have any idea it exists. I already discussed this with Jamie and we thought it was safe to remove.

.github/workflows/nightly-tests.yaml
```diff
@@ -85,6 +85,8 @@ jobs:
       - name: Install SQLite dev package
         if: steps.filter_shards.outputs.skip != 'true'
         run: sudo apt update && sudo apt install -y libsqlite3-dev
+      - name: Install Libmagic dependency package
+        run: sudo apt update && sudo apt install -y libmagic1
```
andrewtruong: why not combine with above?

.github/workflows/test.yaml
```diff
@@ -327,6 +327,8 @@ jobs:
         run: echo "ACTIONS_STEP_DEBUG=true" >> $GITHUB_ENV
       - name: Install SQLite dev package
         run: sudo apt update && sudo apt install -y libsqlite3-dev
+      - name: Install Libmagic dependency package
+        run: sudo apt update && sudo apt install -y libmagic1
```
andrewtruong: Why not combine with above?

pyproject.toml
```diff
@@ -47,6 +47,7 @@ dependencies = [
   "jsonschema>=4.23.0",      # Used by scorers for field validation
   "diskcache==5.6.3",        # Used for data caching
   "nest-asyncio==1.6.0",     # Used to support nested event loops in asyncio
+  "python-magic",            # Used for content type detection
```
andrewtruong: Is it ok to use this?  I think `libmagic` is GPLv3.

Also, will this work on Windows?

tests/trace/type_handlers/Content/content_test.py
```diff
@@ -0,0 +1,308 @@
...
+from weave.type_wrappers.Content.content import Content
+
+TEST_FILE_DIR = os.path.join(os.path.dirname(__file__), "examples")
+
+TEST_FILES = [
+    ("file", "png", "image/png"),
+    ("file", "mp3", "audio/mpeg"),
+    ("file", "mp4", "video/mp4"),
+    ("file", "pdf", "application/pdf"),
+]
```
andrewtruong: Why not use https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

andrewtruong: For other tests we are generating e.g. the images on the fly.  Do you think it's better to have actual mp3, mp4, etc in git?

weave/__init__.py
```diff
@@ -22,8 +22,8 @@
 from weave.trace.util import Thread as Thread
 from weave.trace.util import ThreadPoolExecutor as ThreadPoolExecutor
 from weave.type_handlers.Audio.audio import Audio
-from weave.type_handlers.File.file import File
```
andrewtruong, line 25: There are probably few/no people using this, but in case there are it might be nice to have a deprecation message

weave/trace/annotation_parser.py
```diff
@@ -0,0 +1,151 @@
...
+    # - r"\]": Matches the final closing square bracket for Annotated.
+
+    pattern_with_format = re.compile(
+        r"typing\.Annotated\["
+        r"(.+?),\s*"  # Group 1: Base type (non-greedy)
+        rf"{re.escape(CONTENT_CLASS_NAME)}\["
+        r"typing\.Literal\[['\"](.+?)['\"]\]"  # Group 2: Any literal, non-greedy
+        r"\]"  # Closing bracket for Content[...]
+        r"\]"  # Closing bracket for Annotated[...]
+    )
```
andrewtruong: This should happen outside of the function body so you don't re-compile on each invocation

weave/trace/annotation_parser.py
```diff
@@ -0,0 +1,151 @@
...
+    # - r"(.+?)": Group 1. Captures the base type (non-greedy).
+    # - r",\s*": Matches a comma followed by optional whitespace.
+    # - fr"<class '{re.escape(content_class_name)}'>": Matches the class representation.
+    # - r"\]": Matches the closing square bracket for Annotated.
+    pattern_without_format = re.compile(
+        r"typing\.Annotated\["
+        r"(.+?),\s*"  # Group 1: Base type (non-greedy)
+        rf"<class '{re.escape(CONTENT_CLASS_NAME)}'>"
+        r"\]"  # Closing bracket for Annotated[...]
+    )
```
andrewtruong: same here and the other places

weave/trace/annotation_parser.py
```diff
@@ -0,0 +1,151 @@
...
+    parsed_annotations = {}
+
+    for param_name, param in sig.parameters.items():
+        if not param.annotation:
+            continue
+
+        parse_result = parse_content_annotation(str(param.annotation))
+
+        if not parse_result:
+            continue
```
andrewtruong: nit: could be walrus

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+    data: bytes
+    extra: dict
+    encoding: str
+    path: str | None = None
+    model_config = ConfigDict(extra="allow")
+
+    def __init__(self, data: bytes, /, **values: Any):
+        # Default here so when a factory calls a factory we can safely do the existence check
+        values["encoding"] = values["encoding"] or "utf-8"
+        super().__init__(data=data, **values)
```
andrewtruong: In general you should not override the `__init__` for pydantic models.  Why do you need to do this, and is there a better way?

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+        return self.model_dump(exclude={"data"})
+
+
+def create_bytes_content(
+    input: bytes, /, **values: Unpack[ContentArgs]
+) -> BaseContentHandler:
+    values["size"] = values["size"] or len(input)
+    values["extra"]["input_type"] = values["extra"].get("input_type") or str(
+        type(input)
+    )
```
andrewtruong:
```suggestion
values["extra"]["input_type"] = values["extra"].get("input_type", str(type(input)))
```

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+    values["extra"]["input_category"] = values["extra"].get("input_category") or "data"
```
andrewtruong:
```suggestion
    values["extra"]["input_category"] = values["extra"].get("input_category", "data")
```

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+    # Raw binary data has no encoding unless explicitly provided
+    if values["mimetype"] is None or values["extension"] is None:
+        mimetype, extension = get_mime_and_extension(
+            buffer=input[:2048],
```
andrewtruong: clarify this magic number

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+            filename=values["filename"],
+            extension=values["extension"],
+            mimetype=values["mimetype"],
+        )
+        values["mimetype"] = mimetype
+        values["extension"] = extension
+
+    values["filename"] = values["filename"] or default_filename(
+        str(values["extension"])
+    )
```
andrewtruong: see suggestion above

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+        raise ValueError(f"Input {input} is not a valid file path.")
+
+    path = Path(input)
+    values["path"] = str(path)
+    values["size"] = path.stat().st_size
+    values["filename"] = path.name
+    values["extra"]["original_path"] = values["extra"].get("original_path") or str(path)
+    values["extra"]["input_type"] = values["extra"].get("input_type") or str(
+        type(input)
+    )
```
andrewtruong: see suggestions above

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+    input: str | bytes, /, **values: Unpack[ContentArgs]
+) -> BaseContentHandler:
+    try:
+        data = base64.b64decode(input, validate=True)
+        values["extra"]["input_type"] = values["extra"].get("input_type") or str(
+            type(input)
+        )
+        values["extra"]["input_category"] = (
+            values["extra"].get("input_category") or "base64"
+        )
```
andrewtruong: see suggestions above

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+        # Set encoding to 'base64' for base64 encoded data
+        values["encoding"] = "base64"
+        result = create_bytes_content(data, **values)
+    except Exception as e:
+        raise ValueError(f"Invalid base64 string: {e}") from e
+
+    return result
+
+
+T = TypeVar("T", bound=str)
```
andrewtruong: move to top of file

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+    benefit from the organization of the content handlers without having to
+    manage the object recognition for the weave serializers
+    """
+
+    content_handler: BaseContentHandler
+    _last_saved_path: str | None = None
+
+    def __init__(
+        self,
+        input: Any,
```
andrewtruong: All constructors should have a specific input type.  Instead of switching, make the most common input type the default init.  In this case it might be `path`

Then add classmethods for the alternate types, e.g. `from_bytes`

andrewtruong: Define the actual alternate constructor in the classmethod, not in the main init body

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+    def extension(self) -> str:
+        return self.content_handler.extension
+
+    @property
+    def mimetype(self) -> str:
+        return self.content_handler.mimetype
+
+    @property
+    def encoding(self) -> str:
+        return self.content_handler.encoding or "utf-8"
```
andrewtruong: why or `utf-8`?

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+
+    @property
+    def path(self) -> str | None:
+        return self.content_handler.path
+
+    def as_string(self) -> str:
+        return self.data.decode(self.encoding)
+
+    def open(self) -> bool:
```
andrewtruong, line 331: how do you close?

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,296 @@
...
+    def save(self, dest: str | Path) -> None:
+        """Copy the file to the specified destination path.
+        Updates the filename and the path of the content to reflect the last saved copy
+
+        Args:
+            dest: Destination path where the file will be copied to (string or pathlib.Path)
+                  The destination path can be a file or a directory.
+        """
+        path = Path(dest) if isinstance(dest, str) else dest
+        os.makedirs(path.parent, exist_ok=True)
```
andrewtruong: just use pathlib, don't mix in the `os` level stuff

weave/type_wrappers/Content/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    import magic
+
+try:
+    import magic
+
+    _MAGIC_INSTANCE_FOR_CHECK = magic.Magic(mime=True)
+    MAGIC_LIB_AVAILABLE = True
+    del _MAGIC_INSTANCE_FOR_CHECK  # Clean up
+except Exception as e:
+    MAGIC_LIB_AVAILABLE = False
```
andrewtruong: isn't the import sufficient?

weave/type_wrappers/Content/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    extension: str | None
+    mimetype: str | None
+    filename: str | None
+    path: str | None
+    size: int | None
+    encoding: str | None
+    extra: dict[str, Any]
+
+
+class ContentKeywordArgs(TypedDict, total=False):
```
andrewtruong: why `total=False`?

weave/type_wrappers/Content/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    encoding: str
+    extra: dict[str, Any]
+
+
+def is_valid_path(input: str | Path) -> bool:
+    if isinstance(input, str):
+        input = Path(input)
+    try:
+        return input.exists() and input.is_file()
+    except Exception as _:
```
andrewtruong: why `as _`?

weave/type_wrappers/Content/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    path: str
+    size: int
+    encoding: str
+    extra: dict[str, Any]
+
+
+def is_valid_path(input: str | Path) -> bool:
+    if isinstance(input, str):
+        input = Path(input)
+    try:
```
andrewtruong, line 72: do you even need this try?

weave/type_wrappers/Content/content.py
zbirenbaum: You would just close the triggered application. `open` here doesn't represent `open` like opening for reading and writing, it refers essentially to xdg-open or your platform equivalent. For a PDF this would open your PDF viewer application in a desktop window. This is just something Jamie wrote for the File API so I figured I would include it here.

At least for OSX and Windows it is actually impossible to keep a hook to the created process, as the call exits with no status. I tried really hard to make it more controllable but it ultimately wasn't working and wasn't worth the effort to keep pushing.

weave/type_wrappers/Content/utils.py
zbirenbaum: Not quite. Having a valid import doesn't mean they have `libmagic` installed. In that case, they would be able to import but once they actually try to use the library the code would crash.

pyproject.toml
zbirenbaum: There was a solution for windows but the GPL thing got me looking around (since this repo is open source but including GPL libs is definitely not something to do carelessly)

After looking around a bit I moved over to `polyfile` which is a pure python implementation of libmagic with no outside deps.

weave/trace/annotation_parser.py
zbirenbaum: Fixed. I'm going to leave the continue above since it looks cleaner than the other ways of checking.

.github/workflows/test.yaml
zbirenbaum: It's removed now

.github/workflows/nightly-tests.yaml
zbirenbaum: removed when swapping to polyfile

weave/__init__.py
zbirenbaum: Added it back and it prints a deprecation message with migration instructions on File init now. It also saves any new files as content instead.

weave/type_wrappers/Content/content.py
zbirenbaum: So I wrote it like this in my first pass but it doesn't work because `.get` only defaults when the key is totally unset.

By doing `.get(...) or ...` we can default to the righthand side when the result of `.get` is `None`

```
>>> example = {"hello": None}
>>> print(example.get("hello", "goodbye"))
None
```

weave/type_wrappers/Content/content.py
zbirenbaum: I think there's a better way of doing this, but for context here was the rationale:

Scenario 1:
- User has a Base64 encoded JSON document (not really necessary but just for the sake of example):
- The user passes in encoding='utf-8'
- Expected behavior is `content.as_string()` they get back a valid json string

Scenario 2:
- User has a Base64 encoded string representing a MP4 Video:
- User does not pass encoding
- Expected behavior is to get back a base64 string when calling `content.as_string()`

If we set the default to UTF-8 prior to initialization, by the time we get to the base_64 factory we no longer know if it was user-provided or defaulted.

weave/type_wrappers/Content/content.py
zbirenbaum: Leftover, but we could do the None check here instead of overriding it inside the Pydantic class

weave/type_wrappers/Content/utils.py
zbirenbaum: Pretty certain I added the try after encountering an error when it attempted to check if a bunch of raw data was a valid path on the system. If a non-pathlike string is passed (which we want to do) then this check raises an error

weave/type_wrappers/Content/content.py
zbirenbaum: Moved to utils constant and added documentation

weave/type_wrappers/Content/content.py
andrewtruong: Why do you need to open the application?

weave/type_wrappers/Content/utils.py
andrewtruong: Makes sense

weave/type_wrappers/Content/content.py
zbirenbaum: I guess to view a photo? It was in File and I wanted to keep the API consistent as a superset

pyproject.toml
zbirenbaum: For now I'm going with `puremagic`. I don't think it's quite as robust as the other solutions and returns slightly different (although valid) results like audio/x-wav instead of audio/wav but it seems to work for common files and the browser shouldn't care.

Going to mark this as resolved for now. We can revisit later if necessary.

tests/trace/type_handlers/Content/content_test.py
zbirenbaum: I added generation and removed these files. The video generation was really iffy so I had an LLM construct minimal bytes with markers which the detector should show as mp4. Tests should be good now.

zbirenbaum:
Something I realized while looking through tables for hack-week:

With these changes it won't be possible to do aggregations by media type - I believe this is the reason we called old style media things like: `video.mp4` or `audio.mp3` which gives some information about the type of the content

This is a pretty important feature for metrics. Ideally we would have a separate column for mimetypes. We can pretty easily write a migration for the old files, or we can change content to name the file based on the resolved mimetype and extension.

I'm not a huge fan of storing file information inside of the file name itself as it adds another layer of complexity. Open to input here.

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,338 @@
...
+        Field(
+            description="Extra metadata to associate with the content",
+            examples=[{"number of cats": 1}],
+        ),
+    ] = None
+    encoding: str | None = "utf-8"
+    path: str | None = None
+    extension: str | None = None
+
+    def __init__(
```
zbirenbaum: I feel like it's more convenient for this to delegate to the guess constructor, but your call @andrewtruong

tests/trace/type_handlers/Content/content_test.py
```diff
@@ -0,0 +1,356 @@
...
+def pdf_file(tmp_path_factory) -> Path:
+    file = generate_media("PDF")
+    fn = tmp_path_factory.mktemp("data") / "file.pdf"
+    file.save(fn)
+    return fn
+
+
+@pytest.fixture(scope="session")
+def audio_file(tmp_path_factory) -> Path:
+    file = "file.wav"
```
andrewtruong: remove

weave/trace/serialization/custom_objs.py
```diff
@@ -124,7 +125,11 @@ def _decode_custom_files_obj(
     load_instance_op._tracing_enabled = False  # type: ignore
     art = MemTraceFilesArtifact(encoded_path_contents, metadata={})
     res = load_instance_op(art, "obj")
-    res.art = art
+    # Pydantic models don't allow attribute setting like this
```
andrewtruong: Why do you have to do this?

weave/type_handlers/Content/content.py
```diff
@@ -0,0 +1,42 @@
...
+
+from weave.trace.serialization import serializer
+from weave.trace.serialization.custom_objs import MemTraceFilesArtifact
+from weave.type_wrappers import Content
+from weave.type_wrappers.Content.content import (
+    ResolvedContentArgs,
+    ResolvedContentArgsWithoutData,
+)
+
+logger = logging.getLogger(__name__)
```
andrewtruong: unused

weave/type_handlers/File/file.py
```diff
@@ -11,12 +11,24 @@
...
+from weave.type_wrappers.Content.content import Content
 
 logger = logging.getLogger(__name__)
 
+DEPRECATION_MESSAGE = """
+File is deprecated and scheduled to be removed in the next release\n
+Use the new Content API which supports all File methods instead:\n
+File(path) -> Content.from_path(path)\n
+File(path, mimetype) -> Content.from_path(path, mimetype)\n
+"""
```
andrewtruong, line 26: it may be helpful to write a fixit rule for this as well

weave/type_wrappers/Content/content.py
andrewtruong: What's the alternative?

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,339 @@
...
+    digest: str
+    filename: str
+    content_type: Literal["bytes", "text", "base64", "file", "url"]
+    input_type: str
+
+    # Optional Fields
+    extra: dict[str, Any] | None
+    path: str | None
+    extension: str | None
+    encoding: str | None
```
andrewtruong: For typed dict, optional fields are represented with `NotRequired`

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,339 @@
...
+# This is what is saved to the 'metadata.json' file by serialization layer
+# It is used to 'restore' an existing content object
+class ResolvedContentArgsWithoutData(TypedDict):
+    # Required Fields
+    id: str
+    size: int
+    mimetype: str
+    digest: str
+    filename: str
+    content_type: Literal["bytes", "text", "base64", "file", "url"]
```
andrewtruong: Is this supposed to be different from the `ContentType` above?

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,339 @@
...
+    def save(self, dest: str | Path) -> None:
+        """Copy the file to the specified destination path.
+        Updates the filename and the path of the content to reflect the last saved copy
+
+        Args:
+            dest: Destination path where the file will be copied to (string or pathlib.Path)
+                  The destination path can be a file or a directory.
+                  If dest has no file extension (e.g. .txt), destination will be considered a directory.
+        """
+        path = Path(dest) if isinstance(dest, str) else dest
```
andrewtruong: You can just call `path = Path(dest)` for both cases

weave/type_wrappers/Content/utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    import puremagic
+try:
+    import puremagic
+
+    MAGIC_LIB_AVAILABLE = True
+except Exception as e:
```
andrewtruong: should be (ImportError, ModuleNotFoundError)

andrewtruong: Or something more precise than Exception

weave/type_wrappers/Content/utils.py
andrewtruong: You can just do `except Exception`

tests/trace/type_handlers/Content/content_test.py
```diff
@@ -0,0 +1,356 @@
...
+        file_path = request.getfixturevalue(fixture_name)
+
+        # Test Content() with Path object
+        content = Content(file_path)
+        assert content is not None
+        assert content.extension == extension
+        assert content.mimetype == mimetype
+        assert content.filename == file_path.name
+        assert content.path == str(file_path.resolve())
+        # input_type can change based on library version so we don't check it here
```
andrewtruong: What do you mean?

tests/trace/type_handlers/Content/content_test.py
```diff
@@ -0,0 +1,356 @@
...
+        assert content is not None
+        assert content.extension == ".txt"
+        assert content.mimetype == "text/plain"
+        assert content.data == text_data.encode("utf-8")
+        assert content.encoding == "utf-8"
+        assert content.input_type == "<class 'str'>"
+        assert content.content_type == "text"
+
+        # Test with custom encoding
+        content2 = Content.from_text(text_data, extension=".txt", encoding="utf-16")
```
andrewtruong, line 212: is there a difference between `.txt` and `txt`?

andrewtruong: Seems like no

andrewtruong: Maybe worth some documentation in a README

weave/trace/annotation_parser.py
```diff
@@ -0,0 +1,166 @@
...
+import logging
+import re
+from inspect import Signature
+
+from pydantic import BaseModel
+
+logger = logging.getLogger(__name__)
+
+CONTENT_CLASS_NAME = "weave.type_wrappers.Content.content.Content"
+
```
andrewtruong, line 14: Instead of hardcoding, you can instead get this by importing the object getting the fully qualified name yourself.

weave/trace/annotation_parser.py
```diff
@@ -0,0 +1,166 @@
...
+# - r"\]": Matches the final closing bracket of the Annotated type.
+
+PATTERN_WITH_TYPE_HINT = re.compile(
+    r"typing\.Annotated\["
+    r"(.+?),\s*"  # Group 1: Base type (non-greedy)
+    rf"<class '{re.escape(CONTENT_CLASS_NAME)}\["
+    r"[^\]]*Literal\[['\"](.+?)['\"]\]"  # Group 2: Any literal, non-greedy
+    r"\]'>"  # Closing bracket for Content[...] and class
+    r"\]"  # Closing bracket for Annotated[...]
+)
```
andrewtruong: will this work e.g. if the user writes `Annotated[...]` instead of `typing.Annotated[...]`?

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,339 @@
...
+    def open(self) -> bool:
+        """Open the file using the operating system's default application.
+
+        This method uses the platform-specific mechanism to open the file with
+        the default application associated with the file's type.
+
+        Returns:
+            bool: True if the file was successfully opened, False otherwise.
+        """
+        path = self._last_saved_path or self.path
```
andrewtruong, line 340: Is it possible for `self._last_saved_path` to be undefined here?

weave/type_wrappers/Content/content.py
zbirenbaum: The alternative is to allow the full set of parameters or a union of valid parameters and then delegate to this method:

```
    @classmethod
    def _from_guess(
        cls: type[Self],
        input: bytes | str | Path,
        /,
        extension: str | None = None,
        mimetype: str | None = None,
    ) -> Self:
        # First check if it is a path, we only check validity for str scenario
        # because we have dedicated error message for invalid path
        if isinstance(input, Path) or (isinstance(input, str) and is_valid_path(input)):
            return cls.from_path(input, mimetype=mimetype)

        # Then check if it is base64
        elif isinstance(input, (bytes, str)) and is_valid_b64(input):
            return cls.from_base64(input, mimetype=mimetype, extension=extension)

        # If it is still a str - treat as raw text
        elif isinstance(input, str):
            return cls.from_text(input, mimetype=mimetype, extension=extension)

        return cls.from_bytes(input, mimetype=mimetype, extension=extension)
```

Personally I feel it's more ergonomic.

weave/trace/serialization/custom_objs.py
zbirenbaum: I can give you the technical reason but I'm not sure why this is done for the first place and spent quite a bit looking into it.

The technical reason:
- Content now subclasses BaseModel
- Pydantic models have strict schema checking so setting res.art doesn't work.
- We could do a `copy_model` and create a new instance with this field set
- That can get expensive since large data is likely to be copied and I couldn't find a reason that this field is even necessary.

Ideally this is a Tim question but he is out.

Since it doesn't affect functionality and it's only done at the de-serialization layer this can be changed later without side effects if necessary.

tests/trace/type_handlers/Content/content_test.py
zbirenbaum: E.g. In CI it's `pathlib._local.PosixPath`
Locally it's `pathlib.PosixPath`
I think I've also gotten `pathlib.Path` before with some versions

It's very messy and prone to break CI to just have a list to check so this should be ignored.

tests/trace/type_handlers/Content/content_test.py
zbirenbaum: I'll document it. I changed it so that internally we use the conventional extension formatting which is to have a `.` leading. For the user they can omit or include.

weave/type_wrappers/Content/content.py
zbirenbaum: Messaged you on this.

weave/type_wrappers/Content/content.py
zbirenbaum: Fixed

weave/type_wrappers/Content/utils.py
zbirenbaum: Fixed

weave/trace/annotation_parser.py
zbirenbaum: fixed so it will match any library to import from.

weave/type_wrappers/Content/content.py
zbirenbaum: Fixed

weave/type_handlers/File/file.py
zbirenbaum: Added to rules/legacy_methods.py

weave/trace/annotation_parser.py
zbirenbaum: Fixed

weave/type_wrappers/Content/content.py
zbirenbaum: Fixed

weave/trace/serialization/custom_objs.py
andrewtruong: I think you can resolve this by adding art as a PrivateAttr?

tests/trace/type_handlers/Content/content_test.py
andrewtruong: Can you do an `in` check here?  Otherwise I would leave a comment in the code about this

rules/legacy_methods.py
```diff
@@ -57,3 +59,99 @@ def visit_Call(self, node: libcst.Call) -> None:
...
+
+                # Report the violation and provide the fix.
+                self.report(
+                    node,
+                    "The `File` class is deprecated. Use `Content.from_path` instead. "
+                    "You may need to add 'from weave import Content'.",
+                    replacement=replacement_node,
+                )
+        except Exception:
+            # Avoid crashing the linting process if metadata isn't available for any reason.
```
andrewtruong: When would this happen?

weave/type_wrappers/Content/content.py
andrewtruong: in that case maybe we just don't allow the user to use the normal constructor and force them to use a classmethod?  Might be more obvious

rules/legacy_methods.py
zbirenbaum: I think it was more relevant before I added the check for Collection instance. I can remove the try-catch now

weave/type_wrappers/Content/content.py
```diff
@@ -0,0 +1,359 @@
...
+    # Required Fields
+    id: str
+    size: int
+    mimetype: str
+    digest: str
+    filename: str
+    content_type: ContentType
+    input_type: str
+
+    # Optional fields - can be omitted or None
```
andrewtruong: can't be none

andrewtruong (approved):
lgtm, a few nits

tests/trace/type_handlers/Content/content_test.py
zbirenbaum: Yeah one minor change I made was to make all extensions have a `.` since other libraries do the same. User input is still free to omit it but internally we should use the conventional representation.

pyproject.toml
```diff
@@ -47,6 +47,10 @@ dependencies = [
   "diskcache==5.6.3",         # Used for data caching
   "nest-asyncio==1.6.0",      # Used to support nested event loops in asyncio
 
+  # Used for content type detection
+  # Remove once license content filtering is upstreamed
+  "polyfile @ git+https://github.com/zbirenbaum/polyfile",
```
andrewtruong: please pin a specific sha

weave/type_wrappers/Content/content.py
```diff
@@ -80,21 +45,29 @@ class Content(BaseModel, Generic[T]):
...
     path: str | None = None
     extension: str | None = None
 
     _last_saved_path: Annotated[
         str | None, Field(description="Last path the file was saved to")
     ] = None
 
+    # These fields are set by serialization layer when it picks up a pydantic class
+    # We define them here so they can be set without doing `extra=allow`
+    _ref: Annotated[Any, PrivateAttr] = PrivateAttr(None)
```
andrewtruong: This should be Ref | None

weave/type_wrappers/Content/content.py
```diff
@@ -80,21 +45,29 @@ class Content(BaseModel, Generic[T]):
...
+    _art: Annotated[Any, PrivateAttr] = PrivateAttr(None)
```
andrewtruong: This should be whatever the Art type is or None

weave/type_wrappers/Content/utils.py
```diff
@@ -69,25 +89,42 @@ def default_filename(
...
 
-    if res := next(MagicMatcher.DEFAULT_INSTANCE.match(buffer)).mimetypes[0]:
-        return res
-    return None
+    try:
+        # Lazily import polyfile only when needed.
+        from polyfile.magic import MagicMatcher
+
+        matcher = cast("MagicMatcher", MagicMatcher.DEFAULT_INSTANCE)
+        return next(matcher.match(buffer)).mimetypes[0]
```
andrewtruong: Move this into a separate try block

weave/type_wrappers/Content/utils.py
```diff
@@ -69,25 +89,42 @@ def default_filename(
...
+        logger.warning(
+            "Failed to determine MIME type from file extension and cannot infer from data\n"
+            "MIME type detection from raw data requires the polyfile library\n"
+            "Install it by running: `pip install polyfile `\n"
+            "See: https://pypi.org/project/polyfile for detailed instructions"
+        )
+        return None
+    except IndexError:
+        # This occurs if polyfile is installed but finds no match.
+        return None
```
andrewtruong, line 124: same block

weave/type_wrappers/Content/content.py
zbirenbaum: Fixed

weave/type_wrappers/Content/utils.py
zbirenbaum: Fixed

weave/type_wrappers/Content/utils.py
zbirenbaum: Fixed

pyproject.toml
zbirenbaum: Very good catch

weave/type_wrappers/Content/content.py
zbirenbaum: Fixed. I was worried about circular imports but turns out its fine

PR #4789, opened 2025-06-17
fix(weave): Fix call compare view hang

chance-wnb:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CompareGrid.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CompareGrid.tsx
@@ -77,66 +77,17 @@ export const CompareGrid = ({
       ...(baselineEnabled ? [objectIds[0]] : []),
     ],
   };
-  const columns: GridColDef[] = [];
-  if (mode === 'unified' && objectIds.length === 2) {
-    columns.push({
-      field: 'value',
-      headerName: 'Value',
-      flex: 1,
-      display: 'flex',
-      sortable: false,
-      renderCell: cellParams => {
-        const objId = objectIds[1];
-        const compareIdx = baselineEnabled
-          ? 0
-          : Math.max(0, objectIds.indexOf(objId) - 1);
-        const compareId = objectIds[compareIdx];
-        const compareValue = cellParams.row.values[compareId];
-        const compareValueType = cellParams.row.types[compareId];
-        const value = cellParams.row.values[objId];
-        const valueType = cellParams.row.types[objId];
-        const rowChangeType = cellParams.row.changeType;
-        return (
-          <div className="w-full p-8">
-            <CompareGridCell
-              path={cellParams.row.path}
-              displayType="both"
-              value={value}
-              valueType={valueType}
-              compareValue={compareValue}
-              compareValueType={compareValueType}
-              rowChangeType={rowChangeType}
-            />
-          </div>
-        );
-      },
-    });
-  } else {
-    const versionCols: GridColDef[] = objectIds
-      .slice(0, MAX_OBJECT_COLS)
-      .map(objId => ({
-        field: objId,
-        headerName: objId,
+  const columns: GridColDef[] = useMemo(() => {
+    const columns: GridColDef[] = [];
+    if (mode === 'unified' && objectIds.length === 2) {
+      columns.push({
+        field: 'value',
+        headerName: 'Value',
         flex: 1,
         display: 'flex',
-        width: 500,
         sortable: false,
-        valueGetter: (unused: any, row: any) => {
-          return row.values[objId];
-        },
-        renderHeader: (params: any) => {
-          if (objectType === 'call') {
-            // TODO: Make this a peek drawer link
-            return objId;
-          }
-          const idx = objectIds.indexOf(objId);
-          const objVersion = objects[idx];
-          const objRef = objectVersionSchemaToRef(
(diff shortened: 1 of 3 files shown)


chance-wnb:
* **#4789** <a href="https://app.graphite.dev/github/pr/wandb/weave/4789?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4789?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CompareGrid.tsx
```diff
@@ -77,66 +77,17 @@ export const CompareGrid = ({
...
-        );
-      },
-    });
-  } else {
-    const versionCols: GridColDef[] = objectIds
-      .slice(0, MAX_OBJECT_COLS)
-      .map(objId => ({
-        field: objId,
-        headerName: objId,
+  const columns: GridColDef[] = useMemo(() => {
```
chance-wnb, line 80: The diffing is weird, but I only added a useMemo.

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CompareGridCell.tsx
```diff
@@ -24,6 +25,37 @@ type CompareGridCellProps = {
...
+  }
+
+  // Check for data URL format: data:image/[type];base64,[data]
+  if (str.startsWith('data:image/') && str.includes(';base64,')) {
+    return true;
+  }
+
+  // Check for 'base64:' prefix format
+  if (str.startsWith('base64:')) {
+    return true;
```
gtarpenning, line 54: totally optional: could consider regex.test, should be slightly more performant and we can make both conditions in one

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CompareGridCell.tsx
```diff
@@ -45,12 +77,15 @@ export const CompareGridCell = ({
...
+    if (isBase64Image(compareValue) || isBase64Image(value)) {
+      return <ImageCompareViewer value={value} compareValue={compareValue} />;
+    }
     // TODO: Need to figure out how to override font to be 'Source Sans Pro'
     return (
       <div className="m-[-6px] flex items-start text-xs">
         <ReactDiffViewer
-          oldValue={compareValue}
-          newValue={value}
+          oldValue={limitStringLength(compareValue, 10000)}
```
gtarpenning: I like how this is configurable, 10,000 is very generous, I wonder if we should make this a constant so its easy to go in later and reduce it if we still see problems?

gtarpenning (approved):
image
Small nit, maybe the arrow can go in between the images? Lets tag @m-rgba for a design once-over. :) LGTM !

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ImageCompareViewer.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  // Show side-by-side comparison for different images
+  return (
+    <div className="flex items-start gap-4">
+      <div className="flex flex-col items-center">
+        <div className="text-gray-500 mb-2 text-xs">Before</div>
+        <ValueViewImage value={processedCompareValue} />
+      </div>
+      <div className="text-gray-400 flex items-center">→</div>
```
m-rgba: - We might want to do `items-center` on the container here so the icon is between the two images.
- Could we also use the `IconForwardNext` icon from the design system for the arrow?
Either than that LGTM!

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CompareGridCell.tsx
chance-wnb: Yeah, thanks for the feedback. I will leave it as it for now to go for readability.

Whether regex is a more performant might be controversial: 
1. It does require compilation, which takes extra overhead. otherwise:
2. I can put the regex outside of the function to the global scope, that takes memory. 

I am not super clear TBH, it is very possible that regex has better performance in this case and you are completely correct. But I feel that the performance difference might be very subtle.

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/CompareGridCell.tsx
chance-wnb: Thanks. added a constant now.

chance-wnb:
Thanks guys, I have now made the arrow to appear in the middle and I am using `<IconForwardNext />` as well. It now looks like this:

image

weave-js/src/components/PagePanelComponents/Home/Browse3/compare/ImageCompareViewer.tsx
chance-wnb: Yep, made it that way. Thanks!

gtarpenning:
> Thanks guys, I have now made the arrow to appear in the middle and I am using `<IconForwardNext />` as well. It now looks like this:
> 
> image

awesome

m-rgba (approved):

PR #4796, opened 2025-06-17
chore(artifacts): Add tag to indicate reference artifacts cannot be downloaded

ruhiparvatam:
## Description

- Fixes WB-25478

Reference artifacts cannot be downloaded via the UI. This PR replaces the download icon with a tag and popup that indicates to the user that this is a reference artifact and cannot be downloaded.

Screenshot 2025-06-17 at 11 50 11 AM

## Testing

Testing locally on SaaS.

### DIFF
--- a/weave-js/src/components/Panel2/PanelDir/Component.tsx
+++ b/weave-js/src/components/Panel2/PanelDir/Component.tsx
@@ -16,6 +16,8 @@ import React, {useState} from 'react';
 import {Icon, Pagination, Table} from 'semantic-ui-react';
 
 import * as LLReact from '../../../react';
+import {Tag} from '../../Tag';
+import {Tooltip} from '../../Tooltip';
 import * as Panel2 from '../panel';
 import {Panel2Loader} from '../PanelComp';
 import {inputType} from './common';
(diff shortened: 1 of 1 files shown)


weave-js/src/components/Panel2/PanelDir/Component.tsx
```diff
@@ -193,14 +195,27 @@ const SubfileRow: React.FC<SubfileRowProps> = props => {
...
+            <Tooltip
+              content={
+                <>
+                  This artifact exists externally and is <br />
+                  only referenced in the code. It cannot
+                  <br />
+                  be downloaded via the UI.
+                </>
+              }
+              trigger={<Tag label="Referenced" color={'moon'} />}
```
jamie-rasmussen: Can write `color="moon"`

weave-js/src/components/Panel2/PanelDir/Component.tsx
```diff
@@ -193,14 +195,27 @@ const SubfileRow: React.FC<SubfileRowProps> = props => {
...
-        )}
+        {isDownloadable &&
+          (file.ref ? (
+            <Tooltip
+              content={
+                <>
+                  This artifact exists externally and is <br />
+                  only referenced in the code. It cannot
+                  <br />
+                  be downloaded via the UI.
```
jamie-rasmussen, line 207: It's a little confusing if `isDownloadable` is true for something that isn't downloadable - could we update the logic to be `if (isDownloadable)... else if (isExternalRef)...` or something like that?

ruhiparvatam: `isDownloadable` has more to do with if the user permissions to download the file, so I updated the variable name to reflect that. Technically, `isDownloadable` files can still be downloaded, just not via UI.

ibindlish (approved):
lgtm

PR #4797, opened 2025-06-17
fix(weave): support eval ref get for scorer class

### LINKED ISSUE #4793: Bug: 'TypeError: Unable to cast to Scorer' when retrieving an Evaluation that uses a class-based scorer
#### Summary  
An `Evaluation` that contains a scorer **class instance** cannot be retrieved with  
`weave.ref(<uri>).get()`.  
The same logic implemented as an **`@weave.op` function** works fine.

---

#### Repro steps

1. Run **Script A** (class-based scorer) ⬇️ 
2. Observe that the evaluation runs successfully, but the final line throws  
   `TypeError: Unable to cast to Scorer` when we dereference the URI.  
3. Run **Script B** (function scorer) ⬇️  
4. Observe that the script completes without error.

<details>
<summary>Script A – scorer <em>class</em> (fails when dereferencing URI)</summary>

```python
import weave
from weave import Model, Evaluation, Scorer
import asyncio

WEAVE_TEAM    = "<sample_team>"
WEAVE_PROJECT = "<sample_project>"

client = weave.init(f"{WEAVE_TEAM}/{WEAVE_PROJECT}")

# -- dataset -----------------------------------------------------------
rows = [
    {"prompt": "sample prompt 1", "answer": "sample answer 1"},
    {"prompt": "sample prompt 2", "answer": "sample answer 2"},
    {"prompt": "sample prompt 3", "answer": "sample answer 3"},
    {"prompt": "sample prompt 4", "answer": "sample answer 4"},
    {"prompt": "sample prompt 5", "answer": "sample answer 5"},
]
local_ds = weave.Dataset(name="simple_qa_demo", rows=rows)
weave.publish(local_ds)
ds = weave.ref("simple_qa_demo").get()

# -- scorer ------------------------------------------------------------
class SampleScorer(Scorer):
    @weave.op()
    def score(self, output: dict | None, answer: str):
        return {"correct": output["generation"] == answer}

    @weave.op()
    def summarize(self, rows: list):
        truthy = [r["correct"] for r in rows if r.get("correct") is not None]
        return {"correct": truthy.count(True)}

eval = Evaluation(dataset=ds, scorers=[SampleScorer()])
weave.publish(eval, "sample_eval")

# -- dummy model -------------------------------------------------------
class SampleModel(Model):
    @weave.op()
    def predict(self, prompt: str):
        return {"generation": "generic"}

asyncio.run(eval.evaluate(SampleModel()))

uri = eval.ref.uri()
print("Evaluation URI:", uri)

# 💥 fails here
weave.ref(uri).get()
```
</details>

<details>
<summary>Script B – scorer <em>function</em> (works)</summary>

```python
import weave
from weave import Model, Evaluation
import asyncio

WEAVE_TEAM    = "<sample_team>"
WEAVE_PROJECT = "<sample_project>"

client = weave.init(f"{WEAVE_TEAM}/{WEAVE_PROJECT}")

# -- dataset -----------------------------------------------------------
rows = [
    {"prompt": "sample prompt 1", "answer": "sample answer 1"},
    {"prompt": "sample prompt 2", "answer": "sample answer 2"},
    {"prompt": "sample prompt 3", "answer": "sample answer 3"},
    {"prompt": "sample prompt 4", "answer": "sample answer 4"},
    {"prompt": "sample prompt 5", "answer": "sample answer 5"},
]
local_ds = weave.Dataset(name="simple_qa_demo", rows=rows)
weave.publish(local_ds)
ds = weave.ref("simple_qa_demo").get()

# -- scorer ------------------------------------------------------------
@weave.op()
def sample_scorer(output: dict | None, answer: str):
    return {"correct": output["generation"] == answer}

eval = Evaluation(dataset=ds, scorers=[sample_scorer])
weave.publish(eval, "sample_eval")

# -- dummy model -------------------------------------------------------
class SampleModel(Model):
    @weave.op()
    def predict(self, prompt: str):
        return {"generation": "generic"}

asyncio.run(eval.evaluate(SampleModel()))

uri = eval.ref.uri()
print("Evaluation URI:", uri)

# ✅ succeeds
weave.ref(uri).get()
```
</details>

---

#### Expected behaviour  
Both scripts should reload the `Evaluation` without raising an exception.

#### Actual behaviour  
Only the function-based scorer reloads; the class-based scorer raises:

```
TypeError: Unable to cast to Scorer
```

#### Environment  
* Weave version: **v0.51.54** (same behaviour on v0.51.53)
* Python  : **3.13.3** (venv)  
* OS      : macOS 14 (Apple Silicon)

#### Notes / possible cause  
During reload, the scorer instance arrives at `weave/flow/casting.py::cast_to_scorer` as an `ObjectRef`, which is not handled, so the final `else` path raises the error.  
An `@weave.op` function, in contrast, is restored as an `OpRef`, which *is* handled.

---

Thanks!

gtarpenning:
## Description

[WB-25574](https://wandb.atlassian.net/browse/WB-25574)

`eval_results.ref.get()` breaks when defining eval scorers as classes (but works for functions). 

Fixes: https://github.com/wandb/weave/issues/4793

## Testing

adds test that's breaking

[WB-25574]: https://wandb.atlassian.net/browse/WB-25574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/flow/casting.py
+++ b/weave/flow/casting.py
@@ -42,6 +42,8 @@ def cast_to_scorer(obj: Any) -> Scorer | Op:
         res = weave.op(obj)
     elif isinstance(obj, OpRef):
         res = obj.get()
+    elif isinstance(obj, ObjectRef):
+        res = Scorer.from_obj(obj.get())
     else:
         raise TypeError("Unable to cast to Scorer")
 
--- a/tests/trace/test_evaluate.py
+++ b/tests/trace/test_evaluate.py
@@ -537,3 +537,7 @@ def test_evaluate_get_ref(client):
     assert retrieved_eval2 is not None
     assert len(retrieved_eval1.scorers) == 1
     assert len(retrieved_eval2.scorers) == 1
+
+    # Type assertions
+    assert isinstance(retrieved_eval1.scorers[0], TestScorer)
+    retrieved_eval1 = weave.ref(ref1.uri()).get()
--- a/tests/trace/test_evaluate.py
+++ b/tests/trace/test_evaluate.py
@@ -489,3 +489,51 @@ def test_evaluate_table_order(client):
     scores = [c.output["scores"]["score_simple"] for c in predict_and_score_calls]
     assert all(scores)
     assert len(scores) == 5
+
+
+def test_evaluate_get_ref(client):
+    """Test that we can get refs for evaluations with both class and function scorers."""
+    # Simple dataset
+    rows = [
+        {"prompt": "1 + 1", "answer": "2"},
+        {"prompt": "2 + 2", "answer": "4"},
+    ]
+    dataset = weave.Dataset(rows=rows)
+
+    # Class-based scorer
+    class TestScorer(weave.Scorer):
+        @weave.op()
+        def score(self, output, answer):
+            return {"correct": output == answer}
+
+    # Function-based scorer
+    @weave.op()
+    def test_scorer_func(output, answer):
+        return {"correct": output == answer}
+
+    # Create evaluations
+    eval_with_class = Evaluation(dataset=rows, scorers=[TestScorer()])
+    eval_with_func = Evaluation(dataset=rows, scorers=[test_scorer_func])
+
+    # Publish evaluations
+    ref1 = weave.publish(eval_with_class, "eval_class")
+    ref2 = weave.publish(eval_with_func, "eval_func")
+
+    # Simple model
+    @weave.op()
+    def simple_model(prompt):
+        return "2" if "1 + 1" in prompt else "4"
+
+    # Run evaluations to ensure they work
+    result1 = asyncio.run(eval_with_class.evaluate(simple_model))
+    result2 = asyncio.run(eval_with_func.evaluate(simple_model))
+
+    # Verify we can get refs for both evaluations
+    retrieved_eval1 = weave.ref(ref1.uri()).get()
+    retrieved_eval2 = weave.ref(ref2.uri()).get()
+
+    # Basic assertions to ensure they work
+    assert retrieved_eval1 is not None
+    assert retrieved_eval2 is not None
+    assert len(retrieved_eval1.scorers) == 1
+    assert len(retrieved_eval2.scorers) == 1


neutralino1 (approved):
Thank you.

jwlee64:
Just had a user hit this at FC

weave/flow/casting.py
```diff
@@ -42,6 +42,8 @@ def cast_to_scorer(obj: Any) -> Scorer | Op:
         res = weave.op(obj)
     elif isinstance(obj, OpRef):
         res = obj.get()
+    elif isinstance(obj, ObjectRef):
+        res = Scorer.from_obj(obj.get())
```
andrewtruong, line 46: Shouldn't this just be `obj.get()`?

gtarpenning: That doesn't work!

KensingtonOscupant:
Hi again, thanks for looking into this so quickly!  

I looked into your suggested fix too and I think there is another problem now/still:

---

### Minimal repro (modify end of previous script I posted for class-based scorer)

```python
# ... identical setup as before ...

uri = eval.ref.uri()
eval_based_on_uri = weave.ref(uri).get()

print("original:", eval.scorers[0])
print("round-trip:", eval_based_on_uri.scorers[0])
```

**Observed output**

```
original: SampleScorer(name=None, description=None, column_map=None)
round-trip: Scorer(name=None, description=None, column_map=None)
```

---

### Where it breaks

1. `weave.ref(uri).get()` deserialises the `Evaluation`.
2. Inside `cast_to_scorer` the scorer arrives as an `ObjectRef`, so we now hit the suggested fix.
3. `Scorer.from_obj(obj.get())` instantiates via `cls(**field_values)` where `cls` is **always `Scorer`**.

Hope this helps!

weave/flow/casting.py
gtarpenning:
```
 TypeError: WeaveObject(ObjectRecord({'name': None, 'description': None, 'column_map': None, 'score': OpRef(entity='shawn', project='test-project', name='TestScorer.score', _digest='qJcpCGsmfXS0MxQXVTRQxsfzbgKYroIzo3jJ3fNMpE8', _extra=()), 'summarize': OpRef(entity='shawn', project='test-project', name='Scorer.summarize', _digest='POnP8qYSdK8dO5KWzwnC2SCyEvdiekoISUTevkJqr6U', _extra=()), '_class_name': 'TestScorer', '_bases': ['Scorer', 'Object', 'BaseModel']})) is not a callable object
```

KensingtonOscupant: I looked into this a little more - using ```obj.get()``` in itself works as intended, it returns:

- A `WeaveObject` wrapper containing an `ObjectRecord`
- The `ObjectRecord` contains all the original data
- The `_class_name` field is preserved as `"SampleScorer"`

The real problem is in the deserialization, specifically in the `from_json()` function in `serialize.py`. From my understanding, the deserialization system has two paths:

1. **Builtin objects**: Uses `BUILTIN_OBJECT_REGISTRY` to reconstruct the original class
2. **Custom objects**: Falls back to creating a generic `ObjectRecord`

However, there is no path for custom classes like `SampleScorer` that inherit from builtin classes:

```python
# In from_json()
elif isinstance(val_type, str) and obj.get("_class_name") == val_type:
    cls = BUILTIN_OBJECT_REGISTRY.get(val_type)  # "SampleScorer" not in registry
    if cls:
        # This branch was never taken
        return cls.model_validate(obj_data)
    
    # Falls through to generic ObjectRecord creation
    return ObjectRecord({...})
```

Since `SampleScorer` isn't in the `BUILTIN_OBJECT_REGISTRY`, it would create a generic `ObjectRecord` instead of reconstructing the original class. A fix would be to just use dynamic import logic to handle custom classes - this of course isn't suitable for production (and also doesn't allow users to pull classes they can't import locally) but might help for illustration:

```python
else:
    # Handle custom classes that aren't in the builtin registry
    class_name = obj.get("_class_name")
    if class_name:
        try:
            if class_name == "SampleScorer":
                from sample_scorer import SampleScorer
                cls = SampleScorer
            
            if cls:
                # Filter out metadata fields and method URIs before validation
                obj_data = {
                    k: v for k, v in obj.items() 
                    if k not in ("_class_name", "_bases", "score", "summarize")
                }
                return cls.model_validate(obj_data)
        except ImportError:
            pass  # Fall back to ObjectRecord
```

Also to give more context on why I was even looking into this: I need to run "URI-based" evals for my use case. You can extend the script I shared in the issue by adding

```
evaluation_based_on_uri = weave.ref(uri).get()
asyncio.run(evaluation_based_on_uri.evaluate(model))
```

Using ```Scorer.from_obj(obj.get())```, the last line will raise a ```NotImplementedError``` (has to do with scorer.py), but when changing to ```obj.get()```and modifying ```serialize.py``` like above, that URI-based eval will succeed.

gtarpenning:

PR #4799, opened 2025-06-17
fix(ui): Leaderboard: rendering fix / prevent loop

m-rgba:
## Description

- Race condition / react state updating issue.
  - `defaultSortModel` was recalculated on render 
  - Effect would trigger when `defaultSortModel` changed.
  - Setting `sortModel` caused a render.

Added a `sortModelInitialized` flag so that sort would only run once.
Intermittent, doesn't happen every load, usually only when the page is loading slowly.

This is an example Leaderboard which consistently sees the issue:
https://wandb.ai/wandb-smle/weave-cookboook-demo/weave/evaluations?view=evaluations_default&peekPath=%2Fwandb-smle%2Fweave-cookboook-demo%2Fobjects%2Fgen_eval_dataset-evaluation%2Fversions%2FhiEbttos6nKKYfJpBzctPhDf4qYPp0dJ1p94iOfSpsY%3F%26

(Uses an older version of Leaderboard which loads on top of the Evaluations table? Used by SMLE though)


m-rgba:
Updated:
https://beta.wandb.ai/?betaVersion=176867ddb076bfd046a3eb110cb371d35513bae4

and this for the leaderboard above:
[https://beta.wandb.ai/wandb-smle/weave-cookboook-demo/weave/evaluations?view=evalua[…]2Fversions%2FhiEbttos6nKKYfJpBzctPhDf4qYPp0dJ1p94iOfSpsY%3F%26](https://beta.wandb.ai/wandb-smle/weave-cookboook-demo/weave/evaluations?view=evaluations_default&peekPath=%2Fwandb-smle%2Fweave-cookboook-demo%2Fobjects%2Fgen_eval_dataset-evaluation%2Fversions%2FhiEbttos6nKKYfJpBzctPhDf4qYPp0dJ1p94iOfSpsY%3F%26)

weave-js/src/components/PagePanelComponents/Home/Browse3/views/Leaderboard/LeaderboardGrid.tsx
```diff
@@ -540,6 +540,8 @@ export const LeaderboardGrid: React.FC<LeaderboardGridProps> = ({
     });
   }, [columns, columnOrder]);
 
+  const [sortModelInitialized, setSortModelInitialized] = useState(false);
```
jamie-rasmussen: Since you don't want setting this value to trigger a render `useRef(false)` could be a better approach

m-rgba: Gotcha. Yeah, cause it's just a background flag, not the actual state itself. Will update.

jamie-rasmussen (approved):

PR #4800, opened 2025-06-17
feat(weave): Remove all ops option from monitor ops dropdown

neutralino1:
Removing "All Ops" option from the monitor ops dropdown.


jwlee64 (approved):
lgtm

PR #4804, opened 2025-06-17
chore(app): Enable using non-canvas bounding box and labels for images

ericakdiaz:
## Description

- https://wandb.atlassian.net/browse/WB-5663
- https://wandb.atlassian.net/browse/WB-6453
- https://wandb.atlassian.net/browse/WB-2313

## Testing

How was this PR tested?

By using a div (for the box) and tooltip (for the label) instead of drawing on a canvas, we don't end up having super thick box lines as well as badly scaled up, pixelated text if the image is displayed larger. Furthermore, we can rely on Radix's tooltip placement to avoid it being cut off the canvas without having to do the algorithm ourselves.

A lot of customers were also asking to hide the label unless we hover, so this also does that.

This is only enabled if the `cardSize` is passed into `ImageWithOverlays` because we need to know the size the image will be displayed at.

See https://github.com/wandb/core/pull/30412 for testing


weave-js/src/components/Panel2/ImageWithOverlays.tsx
```diff
@@ -482,9 +486,101 @@ export const BoundingBoxesCanvas: FC<BoundingBoxCanvasProps> = ({
...
+            left = position.minX * widthMultiplier;
+            top = position.minY * heightMultiplier;
+            width = (position.maxX - position.minX) * widthMultiplier;
+            height = (position.maxY - position.minY) * heightMultiplier;
+          } else {
+            left = (position.middle[0] - position.width / 2) * widthMultiplier;
+            top = (position.middle[1] - position.height / 2) * heightMultiplier;
+            width = position.width * widthMultiplier;
+            height = position.height * heightMultiplier;
+          }
```
ericakdiaz: this is all the same as in `drawBox` https://github.com/wandb/weave/blob/f8a1da887dd5b37cbc448b0723db592eb96846dc/weave-js/src/components/Panel2/ImageWithOverlays.tsx#L538-L567

weave-js/src/components/Panel2/ImageWithOverlays.tsx
```diff
@@ -479,9 +486,120 @@ export const BoundingBoxesCanvas: FC<BoundingBoxCanvasProps> = ({
...
     sliderControls,
+    cardSize,
   ]);
 
-  return <OverlayCanvas {...mediaSize} ref={canvasRef} />;
+  const collisionBoundary = useRef<HTMLDivElement>(null);
+
+  // Scale the image to fit the card size
+  const scale = Math.min(
+    (cardSize?.width ?? 0) / mediaSize.width,
```
peytonsarmiento, line 496: Could `mediaSize.width` or `mediaSize.height` be 0?

ericakdiaz: hmm i think it shouldn't be but might as well handle it haha

ericakdiaz: ok handled it below by doing `Number.isFinite(scale)` since anything `/ 0` = NaN which returns false for isFinite

peytonsarmiento: ok sweet, thanks!

PR #4808, opened 2025-06-18
feat(weave): Threads Query API

chance-wnb:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?


chance-wnb:
* **#4889** <a href="https://app.graphite.dev/github/pr/wandb/weave/4889?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#4969](https://github.com/wandb/weave/pull/4969) <a href="https://app.graphite.dev/github/pr/wandb/weave/4969?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4975](https://github.com/wandb/weave/pull/4975) <a href="https://app.graphite.dev/github/pr/wandb/weave/4975?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#4824** <a href="https://app.graphite.dev/github/pr/wandb/weave/4824?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4823** <a href="https://app.graphite.dev/github/pr/wandb/weave/4823?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4808** <a href="https://app.graphite.dev/github/pr/wandb/weave/4808?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4808?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4888** <a href="https://app.graphite.dev/github/pr/wandb/weave/4888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1179,6 +1180,46 @@ def _default_true(val: Union[bool, None]) -> bool:
...
+                continue
+
+            threads.append(
+                tsi.ThreadSchema(
+                    thread_id=thread_id,
+                    trace_count=trace_count,
+                    start_time=start_time_with_tz,
+                    last_updated=last_updated_with_tz,
+                )
+            )
```
neutralino1: What will the UI do with this? Will it just display the thread info? Or will it query calls per threads to show the nested calls?

weave/trace_server/threads_query_builder.py
```diff
@@ -0,0 +1,234 @@
...
+            id,                              -- Call identifier
+            trace_id,                        -- Trace identifier
+            any(thread_id) as thread_id,     -- Get any non-null thread_id for this call
+                                            -- (all non-null values should be identical)
+            min(started_at) as start_time,   -- Earliest start time for this call
+            max(ended_at) as last_updated    -- Latest end time for this call
+        FROM calls_merged
+        WHERE project_id = {{{project_id_param}: String}}
+            {sortable_datetime_filter_clause}
+        GROUP BY id, trace_id               -- Group by call to merge partial rows
```
neutralino1:
```suggestion
        GROUP BY id, trace_id               -- Group by call id to merge partial rows
```

weave/trace_server/clickhouse_trace_server_batched.py
chance-wnb: Good question. I should have included more context with this PR.

There will be a dedicated page to list threads first, then there will be a subsequent call(or calls) to grab the calls of each thread, with some frontend grouping to get some additional information of each thread (eg, token usage and first and last input/ouput). 

And yes, query calls per thread is also by design. it will be used in the drawer view.

The UI will look like this:

image


https://www.figma.com/design/NSK3Ls0YoDCRFMec3BiThg/Threads-view---group-by?node-id=177-12642&m=dev

weave/trace_server/trace_server_interface.py
```diff
@@ -1042,6 +1042,69 @@ class ProjectStatsRes(BaseModel):
...
+    limit: Optional[int] = Field(
+        default=None, description="Maximum number of threads to return"
+    )
+    offset: Optional[int] = Field(default=None, description="Number of threads to skip")
+    sort_by: Optional[list[SortBy]] = Field(
+        default=None,
+        description="Sorting criteria for the threads. Supported fields: 'thread_id', 'turn_count', 'start_time', 'last_updated', 'p50_call_duration_ms', 'p99_call_duration_ms'.",
+        examples=[[SortBy(field="last_updated", direction="desc")]],
+    )
+    sortable_datetime_after: Optional[datetime.datetime] = Field(
```
tssweeney: Let's make 2 changes here:
1. the `sortable` part of `sortable_datetime_after` leaks implementation details. As a user-facing API, I think you just want `after_datetime` and `before_datetime`. 
2. The field description leaks too much implementation detail end users should not know or need to know about database optimizations etc... 

As an extension to #1, we currently have 2 filter patterns in the calls query:
1. A `filter` parameter that accepts an opinionated set of filters (which is where the `*_datetime`might more reasonable go.
2. A `query` param that is the mongo-style general filter. (we can leave this out for now).

In sum, I would recommend changing this to a `filter` param that accepts a `ThreadsQueryFilter` which in itself has the two filter fields. This will more closely match our other API patterns

weave/trace_server/trace_server_interface.py
```diff
@@ -1120,3 +1183,6 @@ def completions_create_stream(
 
     # Project statistics API
     def project_stats(self, req: ProjectStatsReq) -> ProjectStatsRes: ...
+
+    # Thread API
+    def threads_query(self, req: ThreadsQueryReq) -> ThreadsQueryRes: ...
```
tssweeney: this might be fine for now, but prior experience has shown us that a streaming model is more effective for queries against calls. If not now, i expect we will soon need `threads_query_stream` modeling the `calls_query_stream` above.

weave/trace_server/trace_server_interface.py
```diff
@@ -1042,6 +1042,69 @@ class ProjectStatsRes(BaseModel):
...
+    last_updated: datetime.datetime = Field(
+        description="Latest end time of turn calls in this thread"
+    )
+    first_call_id: Optional[str] = Field(
+        description="Call ID of the first call in this thread (earliest start_time)"
+    )
+    latest_call_id: Optional[str] = Field(
+        description="Call ID of the latest call in this thread (latest end_time)"
+    )
+    p50_call_duration_ms: Optional[float] = Field(
```
tssweeney: I am curious why you decided to eagerly add `p50_call_duration_ms` and other aggregate duration. It seems like these are hand-picked aggregates that: a) might add to query time; b) are a subset of all aggregations. How did you pick these and why do you feel they are appropriate to commit to calculating for every query?

chance-wnb: Thanks for asking about it.

The aggregation fields are what need to be displayed on the UI. (lets review the UI in https://www.figma.com/design/NSK3Ls0YoDCRFMec3BiThg/Threads-view---group-by?node-id=177-12784&m=dev )

Suppose we don't do them here, then I have to make a separate call later and grab all the turn calls of a thread and do aggregation on the frontend. You are right on `a) might add to query time; `, however if I don't do it, with separate calls which do the querying again, I estimate the overall time would be longer. 

I still need to make a second server request, however this set up allows me to escape querying a group of turn calls of a thread, and only focus on getting calls of `first_call_id` and `last_call_id`. Call ids are well indexed on calls table so, doing this is quite manageable. And the reason why I decided to make a separate call instead of table joining with the call table is due to worrying about excess memory usage from the joining input + output, then we later only project into two calls per thread. It might not be worth it.

Separately, the reason that `min(call_start_time) as start_time` and `max(call_end_time) as last_updated` are added in the SQL is because I want those two fields sortable, which I think would be useful for the user. If there is no sortable requirement, I could have retrieved those fields from `first_call_id` and `last_call_id`.

tssweeney: `The aggregation fields are what need to be displayed on the UI.` -> this explanation is always worrisome for me. We should be careful adding something to our public interface that we are therefore committed to maintaining just because it showed up in the UI. Are we even sure that this is a useful metric? I doubt p50/90 for all children calls is really useful for users. My suspicion is that the UI sprinkle will end up going away, then we will be left maintaining a property that is not used and adds complexity. (for example: i suspect the user would more care about turn timing, not call timing). I guess the generalized comment is:
1. Adding properties to our interface means we must maintain it moving forward - therefore we should be thoughtful when doing so.
2. Just because something appears in a UI design, does not mean it was critically thought out in terms of long term value. It very well could have been that the mockup wanted something analytical to put in that space and chose this as a quick space filler.

... I agree with everything else you said in your comment here. I think my main point is when looking at that design, I would have just excluded the timing aggregates from MVP and focus on the core functionality, to avoid the risk described above.

chance-wnb: Sounds good. I will take it out. 

> i suspect the user would more care about turn timing, not call timing

I am somewhat confused on this comment: the percentiles calculated are actually turn timing, not just call timing. In that, a turn contains all its descendant calls, so the timing for the turn call is actually measuring all the time spent in that turn.

tssweeney: This isn't a hard blocker comment, but I want us to be "defensive" when making these decisions. I'd much rather have followups to add such functionality rather than have to maintain something we may want to change / remove in the future

chance-wnb: I am not a fan of having too many features either. I am very happy to take it out then add it later if needs arise. :)

tssweeney: If you want to include it, I would certainly rename the field to `*turn*` instead of `*call*` as that is what lead to my confusion. Looking at the code, i see that the implementation is turn-only.

weave/trace_server/trace_server_interface.py
chance-wnb: Thanks for helping thoroughly inspecting the interfaces. I have implemented all except the `query` param.

weave/trace_server/trace_server_interface.py
chance-wnb: Yes, that makes sense. I have now implemented the *_stream interface instead.

weave/trace_server/trace_server_interface.py
chance-wnb: Sounds good, the new iteration now uses the `*turn*` names.

chance-wnb:
> * chore(weave): Unit tests for the /threads/query API #4824

Please review the new iteration, due to interface and name changes, the unit tests in https://github.com/wandb/weave/pull/4824 are updated as well.

PR #4812, opened 2025-06-18
chore(ui): Monitor: Scorer flow enhancements

m-rgba:
## Description

Makes a few flow changes to the first version of the Monitor area:
- Hides "Active Monitor" switch - defaults to Active. Until we have the edit mode for Monitors, creating an inactive monitor would put the user in a bad state.
- Made LLM Model Configuration read-only:
  - The user can select a model, and we can auto-generate the model details for them (such as the system prompt), and then create a new model.
  - If the user uses a custom model from playground we will load the details from that model into the flow.
    - In the future, we can have an integrated form-flow for setting up custom models in the creation flow.
- Added some example scoring prompts which can be inserted in the scoring prompt area.
  - Increased the size of the scoring prompt area (since this space will be used for composing the prompt information).


m-rgba:
Closed for this PR:
https://github.com/wandb/weave/pull/4834

PR #4813, opened 2025-06-18
chore(ui): Monitor: Transform names (vs. validate and block)

m-rgba:
## Description

https://github.com/user-attachments/assets/2f20c4b6-5acf-4572-b992-49f77a6e2920

Transformation improvements:
- Adds a `transformToValidName` function which will automatically convert the user's nice name they're entering into a valid name for the Weave system.
- Gives the user feedback as their typing to the change which is going to be made to their entry.
- Replaces previous validation which blocked the user from submitting names which were invalid.

Error handling interaction improvements:
-


m-rgba:
Updated Featurebee:
https://beta.wandb.ai/?betaVersion=60fea1732af5dd4827e7438839070855daa9f7de

m-rgba:
Replaced with a better version here: https://github.com/wandb/weave/pull/4832

PR #4817, opened 2025-06-18
chore(ui): Playground: Thinking support in message streaming

m-rgba:
## Description

https://github.com/user-attachments/assets/c9742fbe-bd1c-438e-9aac-eaecee4b9f3a

Adds thinking indicator, and then collapse and expand area for and tags in streaming LLM responses and then the following response in chat view.

- If the completion starts with `<think>` or `<thinking>` we can put the block in a box with moon styling and italics.
  - We'll omit the thinking tags from the chat view.
- Content `trim()` is to reduce whitespace around the thinking box and completion.
- The user can expand and collapse the thinking box.

Follow up PR for scroll into view making it hard to read incoming response:
https://github.com/wandb/weave/pull/4818


m-rgba:
Updated FeatureBee:
https://beta.wandb.ai/?betaVersion=ae7c6b33de07c82d8e3866a38297c8d63cb30b04

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
```diff
@@ -14,11 +15,52 @@ type MessagePanelPartProps = {
   showCursor?: boolean;
 };
 
+const parseThinkingBlocks = (text: string) => {
```
chance-wnb: I am not sure whether using arrow function is preferred over a named function in our style guide. 

However in this case, I personally prefer a named function `function parseThinkingBlocks(text: string) {} `. 

- It is more debug friendly, developer tools will show a function name 
- It has hoisting, so a function can be consumed even before its defined.

If the arrow function is an agreed contract in the style guide, I really wonder why.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
```diff
@@ -14,11 +15,52 @@ type MessagePanelPartProps = {
...
+  const parts: Array<{type: 'thinking' | 'text'; content: string}> = [];
+  const thinkingRegex =
+    /<(think|thinking)>([\s\S]*?)(<\/(think|thinking)>|$)/gi;
+  let lastIndex = 0;
+  let match;
+
+  while ((match = thinkingRegex.exec(text)) !== null) {
```
chance-wnb, line 26: Is it the case that the `text` input will contain at most one `<think>...</think>` section?

Also is it the case that we will receive an increasing data payload as the input `text`? I haven't checked the upstream code, but I guess receiving deltas makes more sense. This `text` seem to be able to include all the history so far. (Are we using streaming API but we glue everything together manually? )

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
```diff
@@ -29,6 +71,10 @@ export const MessagePanelPart = ({
     // if (isLikelyMarkdown(value)) {
     //   return <Markdown content={value} />;
     // }
+
+    const parts = parseThinkingBlocks(value);
```
chance-wnb: Minor: it might deserve using a `useMemo`.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
```diff
@@ -43,10 +89,47 @@ export const MessagePanelPart = ({
           `}
         </style>
         <div>
-          <span className="whitespace-break-spaces">{value}</span>
-          {showCursor && (
-            <span className="cursor-blink -mb-[2px] ml-[4px] inline-block h-[12px] w-[12px] rounded-full bg-gold-500" />
-          )}
+          {parts.map((part, index) => {
```
chance-wnb, line 167: nit: The large curry function, and each of the if branch can be componentized. But I don't think it's a big deal.

chance-wnb (approved):
Thank you!

neutralino1 (commented):
In Cursor, the thinking block auto-collapses when the final response appears. We may want to do that too?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
m-rgba: That makes sense, good with that style for the utility function - updated 👍

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
m-rgba: So for the delta question - yep, we're accumulating and concatenating the deltas and when we pass it through with text it can look at the whole ch

For the `<think>` and `<thinking>` blocks - the happy path is that there will be one thinking block per message.
Theoretically we _could_ have multiple tags which generate multiple thinking blocks.
I don't think that it's a problem if we roll those up in thinking areas (cause they're technically still thinking) -- if we run into any bad cases we can add a filter to only take the first match though.

What may be a problem is that we don't filter out `user` messages where they've typed out `<thinking>`.
We roll those up as well, and we probably shouldn't since the style looks a bit weird.
It's a bit of an edge case, but I'm going to look at passing the role down to the `MessagePanelPart` so we can filter out user messages.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
m-rgba: Yeah this is fair, there's a lot of stuff that could trigger the regex with streaming. Will update.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
m-rgba: Nope, that's fair, we can at least chunk it out to have the thinking, text, and cursor as components. Will update

m-rgba:
> In Cursor, the thinking block auto-collapses when the final response appears. We may want to do that too?

👍 We actually do that by default.
- Based on the `isExpanded` logic in the thinking part we'll default it to open if the cursor is out, and then default to closed if the cursor is done streaming.
- If the user interacts with the expand/contract during streaming though we'll keep that as the state since the user has explicitly set it.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
chance-wnb: Thanks for the explanation, I think we can look at the edge case when it happens.

chance-wnb (approved):
Thank you Martin, the PR still LGTM.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/MessagePanelPart.tsx
```diff
@@ -12,13 +13,133 @@ type MessagePanelPartProps = {
...
+  role,
 }: MessagePanelPartProps) => {
+  const [expandedThinking, setExpandedThinking] = useState<{
+    [key: number]: boolean;
+  }>({});
+
+  // Only parse thinking blocks for assistant messages, not user messages
+  const shouldParseThinking = role !== 'user' && typeof value === 'string';
+  const parts = useMemo(
+    () =>
```
chance-wnb: If you don't have to stick to a one-liner, `shouldParseThinking` can be put inside of the useMemo.

m-rgba: Yep, that's cleaner, updated

PR #4818, opened 2025-06-18
chore(ui): Playground: Only trigger scrollIntoView on new messages (not while streaming)

m-rgba:
## Description

Problem:
- `scrollIntoView` currently triggers during streaming, causing a jerky behavior which makes it hard to read as the generation is occurring.

Solution:
- Only trigger when new messages are added.
- Added a counter so that we won't trigger when messages are deleted.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/ChatView.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/ChatView.tsx
@@ -57,7 +57,7 @@ export const ChatView = ({chat}: ChatViewProps) => {
     ) {
       outputRef.current.scrollIntoView();
     }
-  }, [chatResult]);
+  }, [chat.request?.messages?.length]);
 
   return (
     <div className="flex flex-col pb-32">
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/ChatView.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/ChatView.tsx
@@ -15,6 +15,7 @@ type ChatViewProps = {
 
 export const ChatView = ({chat}: ChatViewProps) => {
   const outputRef = useRef<HTMLDivElement>(null);
+  const prevMessageCountRef = useRef(0);
   const playgroundContext = usePlaygroundContext();
 
   const chatResult = useDeepMemo(chat.result);
@@ -49,7 +50,11 @@ export const ChatView = ({chat}: ChatViewProps) => {
   );
 
   useEffect(() => {
+    const currentMessageCount = chat.request?.messages?.length || 0;
+    const hasNewMessage = currentMessageCount > prevMessageCountRef.current;
+    
     if (
+      hasNewMessage &&
       outputRef.current &&
       chatResult &&
       'choices' in chatResult &&
@@ -57,7 +62,9 @@ export const ChatView = ({chat}: ChatViewProps) => {
     ) {
       outputRef.current.scrollIntoView();
     }
-  }, [chat.request?.messages?.length]);
+    
+    prevMessageCountRef.current = currentMessageCount;
+  }, [chatResult, chat.request?.messages]);
 
   return (
     <div className="flex flex-col pb-32">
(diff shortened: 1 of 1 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/ChatView.tsx
```diff
@@ -49,15 +50,22 @@ export const ChatView = ({chat}: ChatViewProps) => {
...
+        outputRef.current &&
+        chatResult &&
+        'choices' in chatResult &&
+        chatResult.choices
+      ) {
+        outputRef.current.scrollIntoView();
+      }
     }
-  }, [chatResult]);
+  }, [chat.request?.messages, chatResult]);
```
chance-wnb: It seems to me , if we depend on `[chat.request?.messages?.length, chatResult]` we will automatically trigger this useEffect at the proper time. That saves us from planting an extra useRef().

chance-wnb (commented):
Thanks! Proposed one idea to try on. But the current state is also good.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/ChatView.tsx
m-rgba: So for these, we actually slightly over-trigger since each of the changes the the messages - deletions, edits - will also bump the user to the bottom of the page.
That's why we added the `hasNewMessage ` part with the new ref, to add a counter to only push the user down to the bottom of the view if they've added a new message.

m-rgba: With fresh eyes we could actually probably reduce the effect to `[chat.request?.messages?.length]`, removing the `chatResult` cause we're not using that at all for adjusting the position.

chance-wnb:
Yes, if `[chat.request?.messages?.length]` alone can work, that would be the best. From what I see - your `hasNewMessage` takes the value of whether the count of `chat.request?.messages?.length` has changed (current > previous), that is equivalent to React's useEffect dependency tracking, which means simply using `[chat.request?.messages?.length]` should represent the same logic. This is because the length, if I am not mistaken, is monotonically increasing. 

Therefore there is no need to use `useRef` at all.

m-rgba:
> Yes, if `[chat.request?.messages?.length]` alone can work, that would be the best. From what I see - your `hasNewMessage` takes the value of whether the count of `chat.request?.messages?.length` has changed (current > previous), that is equivalent to React's useEffect dependency tracking, which means simply using `[chat.request?.messages?.length]` should represent the same logic. This is because the length, if I am not mistaken, is monotonically increasing.
> 
> Therefore there is no need to use `useRef` at all.

`chat.request?.messages?.length` will work, but it will still trigger when deleting an old message further up the stack (which can be a kind of weird UX since you're deleting something up-top and then being pushed to the bottom).
It's not a breaking UX change, just a little weird - it would be good to keep the ref, but if we feel for simplicity/perf we should remove it I'm good with that as well.

chance-wnb:
> > Yes, if `[chat.request?.messages?.length]` alone can work, that would be the best. From what I see - your `hasNewMessage` takes the value of whether the count of `chat.request?.messages?.length` has changed (current > previous), that is equivalent to React's useEffect dependency tracking, which means simply using `[chat.request?.messages?.length]` should represent the same logic. This is because the length, if I am not mistaken, is monotonically increasing.
> > Therefore there is no need to use `useRef` at all.
> 
> `chat.request?.messages?.length` will work, but it will still trigger when deleting an old message further up the stack (which can be a kind of weird UX since you're deleting something up-top and then being pushed to the bottom). It's not a breaking UX change, just a little weird - it would be good to keep the ref, but if we feel for simplicity/perf we should remove it I'm good with that as well.

I see. Sorry I didn't know that the length is not monotonically increasing, I think your implementation is good. 

Using useRef is totally fine. For the sake of sharing information -- previously I introduced the react-use library to the code base, and there is [`usePrevious`](https://github.com/streamich/react-use/blob/master/docs/usePrevious.md) available now. Feel free to make a call whether to adopt that or keeping using useRef please.

m-rgba:
> Using useRef is totally fine. For the sake of sharing information -- previously I introduced the react-use library to the code base, and there is [usePrevious](https://github.com/streamich/react-use/blob/master/docs/usePrevious.md) available now. Feel free to make a call whether to adopt that or keeping using useRef please.

`usePrevious` looks a little cleaner, will update, we can use that since we've got it in there

PR #4823, opened 2025-06-18
feat(weave): SDK support for using threads

chance-wnb:
## Description

- Adds thread context support to Weave, allowing users to group related operations under a common thread ID
- Implements `weave.thread()` context manager for setting thread IDs on calls
- Adds turn_id tracking to identify top-level operations within a thread
- Introduces ThreadContext class that provides access to thread_id and turn_id
- Adds thread_id and turn_id fields to Call objects and ensures proper inheritance

## Design Doc:

Please review this doc for the overall plan
https://www.notion.so/wandbai/Threads-API-Design-Document-PRD-21ee2f5c7ef3806f93ecfdb38b8982fb

## Testing

Added comprehensive test cases that verify:
- Thread context management with the weave.thread() API
- Thread ID capture in call records
- Thread ID inheritance by child calls
- Error handling within thread contexts
- Turn ID functionality for tracking top-level operations
- Nested thread contexts and thread disabling
- Auto-generation of thread IDs

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -4580,3 +4580,556 @@ def test_calls_query_with_descendant_error(client):
         )
 
         assert len(calls) == count
+
+
+def test_thread_context_with_weave_api(client):
+    """Test thread context using the weave.thread() API with ThreadContext."""
+    import weave
+    from weave.trace.context import call_context
+
+    # Test default thread_id is None (using internal context for verification)
+    assert call_context.get_thread_id() is None
+
+    # Test using weave.thread context manager with ThreadContext
+    with weave.thread("api_thread_1") as t:
+        # Use ThreadContext object to access thread_id
+        assert t.thread_id == "api_thread_1"
+        assert (
+            call_context.get_thread_id() == "api_thread_1"
+        )  # Verify internal consistency
+
+        # Test nested context with ThreadContext
+        with weave.thread("api_thread_2") as inner_t:
+            assert inner_t.thread_id == "api_thread_2"
+            assert t.thread_id == "api_thread_1"  # Outer context unchanged
+            assert call_context.get_thread_id() == "api_thread_2"
+
+        # Should revert to parent context
+        assert t.thread_id == "api_thread_1"
+        assert call_context.get_thread_id() == "api_thread_1"
+
+    # Should revert to None
+    assert call_context.get_thread_id() is None
+
+
+def test_thread_id_in_calls(client):
+    """Test that thread_id is properly captured in call records, including auto-generation."""
+    import weave
+
+    @weave.op
+    def test_op_with_thread(x: int) -> int:
+        return x * 2
+
+    @weave.op
+    def test_op_without_thread(x: int) -> int:
+        return x * 3
+
+    @weave.op
+    def test_op_auto_thread(x: int) -> int:
+        return x * 4
+
+    # Call without thread context
+    result1 = test_op_without_thread(5)
+    assert result1 == 15
+
+    # Call with explicit thread context
+    with weave.thread("test_thread_id") as t:
+        assert t.thread_id == "test_thread_id"
+        result2 = test_op_with_thread(5)
+        assert result2 == 10
+
+    # Call with auto-generated thread_id
+    auto_thread_id = None
+    with weave.thread() as t:
+        auto_thread_id = t.thread_id
+        assert auto_thread_id is not None
+        assert isinstance(auto_thread_id, str)
+        assert len(auto_thread_id) > 20  # Should be UUID v7
+        result3 = test_op_auto_thread(5)
+        assert result3 == 20
+
+    # Get the calls to verify thread_id
+    calls = client.get_calls()
+
+    # Find our calls (most recent first)
+    auto_call = None
+    thread_call = None
+    no_thread_call = None
+
+    for call in calls:
+        if "test_op_auto_thread" in call.op_name:
+            auto_call = call
+        elif "test_op_with_thread" in call.op_name:
+            thread_call = call
+        elif "test_op_without_thread" in call.op_name:
+            no_thread_call = call
+
+    assert auto_call is not None
+    assert thread_call is not None
+    assert no_thread_call is not None
+
+    # Verify thread_id values
+    assert auto_call.thread_id == auto_thread_id  # Should match auto-generated ID
+    assert thread_call.thread_id == "test_thread_id"
+    assert no_thread_call.thread_id is None
+
+
+def test_thread_id_inheritance(client):
+    """Test that thread_id is inherited by child calls, demonstrated via ThreadContext."""
+    import weave
+
+    @weave.op
+    def child_op(x: int) -> int:
+        return x + 1
+
+    @weave.op
+    def parent_op(x: int) -> int:
+        return child_op(x) * 2
+
+    # Call with thread context, using ThreadContext to demonstrate inheritance
+    inherited_thread_id = None
+    with weave.thread("inherited_thread") as t:
+        inherited_thread_id = t.thread_id
+        assert inherited_thread_id == "inherited_thread"
+
+        # Both parent and child calls should inherit this thread_id
+        result = parent_op(10)
+        assert result == 22  # (10 + 1) * 2
+
+        # ThreadContext shows the current thread_id throughout execution
+        assert t.thread_id == "inherited_thread"
+
+    # Get the calls to verify thread_id inheritance
+    calls = client.get_calls()
+
+    # Find our calls
+    parent_call = None
+    child_call = None
+
+    for call in calls:
+        if "parent_op" in call.op_name:
+            parent_call = call
+        elif "child_op" in call.op_name:
+            child_call = call
+
+    assert parent_call is not None
+    assert child_call is not None
+
+    # Both should have the same thread_id that was shown in ThreadContext
+    assert parent_call.thread_id == inherited_thread_id
+    assert child_call.thread_id == inherited_thread_id
+    assert parent_call.thread_id == "inherited_thread"
+    assert child_call.thread_id == "inherited_thread"
+
+
+def test_thread_context_error_handling(client):
+    """Test that ThreadContext is properly managed even when exceptions occur."""
+    import weave
+    from weave.trace.context import call_context
+
+    @weave.op
+    def failing_op(should_fail: bool) -> str:
+        if should_fail:
+            raise ValueError("Test exception")
+        return "success"
+
+    # Test that thread context is properly restored after exception
+    assert call_context.get_thread_id() is None
+
+    # Test ThreadContext behavior during exception
+    exception_thread_context = None
+    try:
+        with weave.thread("exception_thread") as t:
+            exception_thread_context = t
+            assert t.thread_id == "exception_thread"
+            assert call_context.get_thread_id() == "exception_thread"
+            failing_op(True)  # This will raise an exception
+    except ValueError:
+        pass  # Expected
+
+    # Verify ThreadContext maintained its state even after exception
+    assert exception_thread_context.thread_id == "exception_thread"
+
+    # Thread context should be restored to None after exiting context
+    assert call_context.get_thread_id() is None
+
+    # Test successful call after exception with auto-generated thread
+    recovery_thread_id = None
+    with weave.thread() as t:  # Use auto-generation
+        recovery_thread_id = t.thread_id
+        assert recovery_thread_id is not None
+        assert len(recovery_thread_id) > 20  # Should be auto-generated UUID v7
+
+        result = failing_op(False)
+        assert result == "success"
+        assert t.thread_id == recovery_thread_id  # ThreadContext consistent
+
+    assert call_context.get_thread_id() is None
+
+
+def test_turn_id_functionality(client):
+    """Test that turn_id is properly assigned for turn calls and descendants."""
+    import weave
+
+    @weave.op
+    def child_op(x: int) -> int:
+        return x + 1
+
+    @weave.op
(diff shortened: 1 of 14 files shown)


chance-wnb:
* **#4889** <a href="https://app.graphite.dev/github/pr/wandb/weave/4889?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#4969](https://github.com/wandb/weave/pull/4969) <a href="https://app.graphite.dev/github/pr/wandb/weave/4969?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4975](https://github.com/wandb/weave/pull/4975) <a href="https://app.graphite.dev/github/pr/wandb/weave/4975?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#4824** <a href="https://app.graphite.dev/github/pr/wandb/weave/4824?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4823** <a href="https://app.graphite.dev/github/pr/wandb/weave/4823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/4823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4808** <a href="https://app.graphite.dev/github/pr/wandb/weave/4808?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4888** <a href="https://app.graphite.dev/github/pr/wandb/weave/4888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

chance-wnb:
> Is my understanding correct that here threads have nothing to do with execution threads? They are essentially arbitrary IDs that users set to group traces together?

Yes, it doesn't have anything to do with execution threads. Actually, multiple execution thread by default share the same "thread_id" unless especially set up to use different ones. 

The ID doesn't even have to the UUIDs. It can even be a label in natural language.

> As a user, I would love this to be done by default automatically per actual invocation/execution of my code. 

Not sure I follow. What should be done by default please?

neutralino1:
> Not sure I follow. What should be done by default please?

I'm not suggesting to change anything in this PR. Just suggesting a better default behavior.

I presume that if the user does not use `with thread(...`, there will be no `thread_id` set on the calls. When I first used Weave, I found that puzzling. It was hard to tell which traces came from which execution/run. I would love to see a grouping by default, based on invocations of my code. Every time I run my program, a thread ID is set by default.
Then if users want to control this (e.g. an agent that runs on a loop and serves multiple users, should have a different thread ID per user), they can use this API to set the thread ID.

chance-wnb:
> > Not sure I follow. What should be done by default please?
> 
> I'm not suggesting to change anything in this PR. Just suggesting a better default behavior.
> 
> I presume that if the user does not use `with thread(...`, there will be no `thread_id` set on the calls. When I first used Weave, I found that puzzling. It was hard to tell which traces came from which execution/run. I would love to see a grouping by default, based on invocations of my code. Every time I run my program, a thread ID is set by default. Then if users want to control this (e.g. an agent that runs on a loop and serves multiple users, should have a different thread ID per user), they can use this API to set the thread ID.

I see, so maybe such a default thread id should be per python script execution (session) based? 

Yeah, that's actually a good idea. If there are consensus on it, I can implement it later.

weave/__init__.py
```diff
@@ -71,6 +73,7 @@
     "EvaluationLogger",
     "File",
     "finish",
+    "generate_id",
```
tssweeney: Why are we export `generate_id`? Do we expect a user to call this directly?

chance-wnb: Yes, I would like to recommend using this ID as the thread ID in the doc. Though user could have other options. 

Do you have any concerns on exporting?

tests/trace/test_client_trace.py
```diff
@@ -4579,3 +4579,139 @@ def parent_op(val: int):
...
+            parent_call = call
+        elif "child_op" in call.op_name:
+            child_call = call
+
+    assert parent_call is not None
+    assert child_call is not None
+
+    # Both should have the same thread_id
+    assert parent_call.thread_id == "inherited_thread"
+    assert child_call.thread_id == "inherited_thread"
```
tssweeney, line 4722: I actually think this is NOT how it should work. I think we only want immediate children of the context handler to get the thread id. for example:

```
with weave.thread("inherited_thread"):
    result = parent_op(10)
    result = parent_op(10)
```

this should result in 2 calls with thread_id set, and the children ops should NOT have it set. 

why?

Well, if we assign this to all children, then we don't know which op is the entry-point of the thread participant. you would have to do a very costly recursive query to find the "head". If only the immediate children of the context get the thread_id, then you can safely query for all calls with a given thread ID to reconstruct the thread itself

tssweeney (changes requested):
Let's discuss the recursive thread_id - I think we should change behavior here.

weave/__init__.py
tssweeney: I think we should try to limit the re-exporting of symbols as much as possible. This is primarily intended to surface the most important symbols to avoid deeper imports. I would instead make `weave thread() as t` yield a "Thread" class that has an "id" property containing a generated ID if one is not passed. This would allow the user to store that thread id if needed, but not need to have extra code / imports

tests/trace/test_client_trace.py
chance-wnb: I see, the scenario is trickier than I thought. 

Previously, I thought the the entry-points of the thread participant can only be Traces(root calls). Now I realize that a user might skip Traces, but start to inject Thread id for some intermediate ops. 

I think this imposes more challenges:

1. How can I efficiently enumerate all the descendants of an intermediate op? if it is not a trace? The descendants will be used to display the tree view. 

2. We have a column called "Traces" in the threads view, what should that become now? Is it the number of entry-point ops? or all calls in the subtree? It seems we are introducing a new type of entity
image


Alternatively, we can probably think about back filling the ancestry ops with the thread by inspecting the callstack and attach thread ids to the stack, so we have a chance to send it during `call_end` event. Therefore we make everything Trace-centric. 

As a second thought: regarding `you would have to do a very costly recursive query to find the "head"`, I might be able to work around it by assembling the tree on the frontend. not ideal though.

chance-wnb:
> Let's discuss the recursive thread_id - I think we should change behavior here.

Yes we should. There is no code change yet, but I left a comment in the discussion thread.

tests/trace/test_client_trace.py
tssweeney: I think the solution to this does not require additional complexity:
1. Only associate "trace_id" to the immediate children of the context manager. This will require "resetting" the context inside of a call such that it only gets associated with the first layer.
2. The thread list is a list of unique thread ids. The "first input" and "last output" are simply the first and last call associated with that ID.
3. the "Traces" column here really should be something like "steps" or "calls" which counts the number of calls with that trace_id. This should be exactly the same complexity as you had considered before.
4. `How can I efficiently enumerate all the descendants of an intermediate op? if it is not a trace? The descendants will be used to display the tree view.` ... When you link to a call view it is automatically truncated to itself (even if it is not the trace root).

chance-wnb: Yes, this should work. Thanks for the tip.

chance-wnb: I took a look at how we truncate the tree view to the call itself in TraceNavigator.tsx. It involves getting the entire trace first, then extract the subtree. 

I might have to repeat this process to extract the subtree when I need to find out which sub calls contain the chat messages to display the chat log view as illustrated in this figma https://www.figma.com/design/NSK3Ls0YoDCRFMec3BiThg/Threads-view---group-by?node-id=177-12995&m=dev

And I might have to do this for each of the entry-point call. This process might not be very efficient.

So this makes the algorithm to find the head calls easier, but makes discovering the chat message harder. 

On contrary, if we mark all the calls in a thread with thread_id, it will make it easier to discover calls with chat messages (might be possible even on the backend), while for discovering the head nodes, I can re-assemble all of the calls on the frontend, and I only need to do it once. So it seems lessor of the two evils if we paint the entire thread tree with a thread id. 

What do you think? @tssweeney

tssweeney: I think you raise a few good points, but there is still a bit more to consider here. A few notes:

1. You are correct that if the primary user experience requires looking "down" the tree to find descendant calls, then it is more effective to label all the calls and do the query once, assembling later. (Agreed)
2. I think the UX designs assume the user wants to look down the tree and find all the LLM calls to surface as intermediate steps. However, I think another very popular pattern is going to be where each "turn" in a thread is a single instance of a call with that thread id. The "user" message is the input to that thread, and then the output is the agent message. 

Let's assume we have the following abstract agent loop:

```python
@op
def do_agent_loop(result_from_thread_id = None):
   history = load_history(result_from_thread_id)
   while True:
      # Each loop of this `while` is a "turn"
      user_input = get_user_input()
      response = do_agent_turn(user_input, history)
      show_response_to_user(response)

@op
def do_agent_turn(user_input, history):
    history.append(user_input)
    response = None
    while not response:
       # Each loop of this `while` is a "step" of a "turn"
       generation = do_agent_step(history)
       history.append(generation)
       if generation.tool_call:
           # `do_tool` could contain sub-agent llm calls!
           tool_result = do_tool_call(generation.tool_call)
           history.append(tool_result)
       else:
           response = generation.content
     return response

@op
def do_agent_step(history):
    # this is often a traced llm vendor call
    return llm.complete(history)
```

Certainly there are tons of ways to do this, but this is a basic scaffolding. I would argue that the user might want to see a "chat sequence" that is just the user_inputs + responses, whereas other view might be all the lower level llm.completions.

Ultimately, we are missing some data. Let's define some terms. We need to be able to:
1. (turn-root) Identify the root call of a given "turn"
2. (turn-tree) Identify all the calls belonging to a given "turn"
3. (thread-sequence) Identify the sequence of "turn roots" belonging to a given "thread"
4. (agent-sequence) Identify the sequence of internal LLM calls for a given thread.

---

It would be helpful to understand how we plan on instrumenting code like the above and how we can effectively query for the different possible data views. I think this entire thing could alternatively be solved by enforcing the constraint you mentioned above: assume that each turn-root must also be a trace-root. In this case, we can quickly lookup or re-construct any of the views. This would preclude having `@op` on the equivalent to `do_agent_loop` however. 

---

I need to send this message as it is getting late, food for thought - lets chat tomorrow

tssweeney:
> > Not sure I follow. What should be done by default please?
> 
> I'm not suggesting to change anything in this PR. Just suggesting a better default behavior.
> 
> I presume that if the user does not use `with thread(...`, there will be no `thread_id` set on the calls. When I first used Weave, I found that puzzling. It was hard to tell which traces came from which execution/run. I would love to see a grouping by default, based on invocations of my code. Every time I run my program, a thread ID is set by default. Then if users want to control this (e.g. an agent that runs on a loop and serves multiple users, should have a different thread ID per user), they can use this API to set the thread ID.

We should note that "threads" are distinct from "executions". I agree it would be nice to have an "execution" id that could be used to group traces that belong to the same execution... however I think many times this is not what we mean by threads

tests/trace/test_client_trace.py
chance-wnb: Thank you for clearly summarizing the scenarios.  

Regarding the assumption that *"each turn-root must also be a trace-root"*, on a second thought, I believe this approach presents some challenges:  

1. **Usability Concerns**:  
   - This requirement imposes an undue burden on users, as identifying the root operation and wrapping it in a thread context is not always doable depends on customer's scenario.  
   - It also restricts flexibility by disallowing nested threads, which are useful for distinguishing between framework, infrastructure, and business-level threads.  

2. **Implementation Challenges**:  
   - In nested thread scenarios, intermediate threads cannot propagate upward to modify their ancestors' thread IDs once those operations have already been assigned distinct, meaningful thread IDs.  
   - Enforcing this would lead to convoluted code and unintuitive behavior, negatively impacting user experience.  

Given these trade-offs, we can probably look at the following approach:  

Instead, we could allow an entry-point to be any arbitrary intermediate operation. Assuming we also agree to assign a thread ID to all calls within the thread, we could proceed as follows (reordered for clarity):  

3- **(Thread-Sequence)** Identify the sequence of "turn roots" belonging to a given thread.  
   - Reconstruct the call forest for the entire thread, with unparented tree roots serving as turn roots.  (I don't even have to properly reconstruct the calls as trees, as long as I can identify the roots)

2- **(Turn-Tree)** Identify all calls belonging to a given "turn."  
   - During reconstruction, associate all calls in a tree (from step 3) with their respective turn root.  

1- **(Turn-Root)** Identify the root call of a given "turn."  
   - This is naturally derived from step 3.  

4- **(Agent-Sequence)** Identify the sequence of internal LLM calls for a given thread.  
   - Using an O(n) scan, we can filter for LLM calls and display them chronologically, with each call already mapped to its respective turn.  

This approach balances implementation feasibility with an optimal user experience. Please let me know your thoughts.

chance-wnb:
> Let's discuss the recursive thread_id - I think we should change behavior here.

Hi @tssweeney 


**I've added a `turn_id` column to the calls in the latest iteration.**  

- For calls within a thread context, this value will always be non-`None`.  
- **Entry-point calls** (the initial calls of a thread) will have `turn_id == call.id`.  
- **Descendant calls** (subsequent calls in the thread) will inherit `turn_id == entry_point_call.id`.  

This means filtering for `call.id = call.turn_id` will identify all entry-point calls, while the `turn_id` on descendant calls will correctly indicate their associated turn.  

For implementation details, please refer to the related prerequisite PRs in the stack.  
https://github.com/wandb/weave/pull/4888
https://github.com/wandb/weave/pull/4808

PR #4828, opened 2025-06-20
feat(weave): Read inference service base URL from env var

neutralino1:
## Description

Reads the inference service base URL from an env var instead of inferring from request hostname.

- Fixes WB-25624

## Testing

Tested on QA.

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -2845,18 +2845,20 @@ def _setup_completion_model_info(
     elif model_name.startswith("coreweave/"):
         # See https://docs.litellm.ai/docs/providers/openai_compatible
         # but ignore the bit about omitting the /v1 because it is actually necessary
-        req.inputs.model = "openai/" + model_name.replace("coreweave/", "", 1)
+        req.inputs.model = model_name.replace("coreweave/", "openai/", 1)
+
         provider = "custom"
-        base_url = "https://api.inference.wandb.ai/v1"
+
+        base_url = wf_env.inference_service_base_url()
+
         # The API key should have been passed in as an extra header.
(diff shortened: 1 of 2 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2845,18 +2845,20 @@ def _setup_completion_model_info(
     elif model_name.startswith("coreweave/"):
         # See https://docs.litellm.ai/docs/providers/openai_compatible
         # but ignore the bit about omitting the /v1 because it is actually necessary
-        req.inputs.model = "openai/" + model_name.replace("coreweave/", "", 1)
+        req.inputs.model = model_name.replace("coreweave/", "openai/", 1)
+
         provider = "custom"
-        base_url = "https://api.inference.wandb.ai/v1"
+
+        base_url = wf_env.inference_service_base_url()
```
jamie-rasmussen, line 2852: What will happen if base_url is None?

jamie-rasmussen (approved):

weave/trace_server/clickhouse_trace_server_batched.py
neutralino1: It will fail. Let me default to prod.

neutralino1: Added a default value set to the prod URL.

weave/trace_server/environment.py
```diff
@@ -195,3 +195,13 @@ def wf_file_storage_project_ramp_pct() -> Optional[int]:
         )
 
     return pct
+
+
+# Inference Service Settings
+
+
+def inference_service_base_url() -> Optional[str]:
```
jamie-rasmussen: Return type just `str` now

PR #4833, opened 2025-06-20
chore(ui): Monitor: Active toggle - convert to switch / on edit only

m-rgba:
## Description

image

- Converts the active toggle to be enabled by default.
- Switches it to use our `<Switch>` component from the design system.
- Hides it on the create new monitor pages (only showing it in the edit monitor state - to be developed).


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorFormDrawer.tsx
```diff
@@ -424,17 +427,17 @@ export const MonitorFormDrawer = ({
                       onChange={e => setDescription(e.target.value)}
                     />
                   </Box>
-                  <Box>
+                  {monitor && (
```
neutralino1: Why hide on creation form?

m-rgba: I'll pull that out - I didn't think that the Monitor edit view was wired up, but on looking around I noticed it's in the list overflow area.
Going to look at adding the edit control to the expanded view for the monitor as well (similar to datasets).

neutralino1: Ah yes, I've been meaning to do this, thank you.

Also, maybe in the future we want to have the active/inactive toggle be displayed and actionable in the monitors table directly, i.e. no need to open the edit panel to just switch something on and off.

m-rgba: Removed the monitor only criteria.
Added another PR over here which adds the edit button and a loader state to the monitor expanded page:
https://github.com/wandb/weave/pull/4854

m-rgba: > Also, maybe in the future we want to have the active/inactive toggle be displayed and actionable in the monitors table directly, i.e. no need to open the edit panel to just switch something on and off.

Yep we could pull that out to the table view as well - probably just need a confirmation as the user enables / disables since we don't have too many inline edits like that. Think I may also bring the delete out inline in the disclosure with edit button as well.

neutralino1 (approved):

PR #4842, opened 2025-06-23
feat(weave): filter by object references using CTE

gtarpenning:
## Description

Frontend portion of this pr will be broken out into https://github.com/wandb/weave/pull/5062 after approval/ final testing

Filter calls by properties of referenced objects. Should work for any level of nested objects. This will mostly affect the UI, as users will be able to sort and filter by traces that have nested objects, which is a frequent ask.

Note on **performance**. I believe that this solution is performant enough to release. For the ultra large p99 customers, especially those with vast quantities of objects, this may not be good enough. However, for most users, the main performance bottleneck will remain the calls query itself, so focusing on optimizing that will also improve performance here. The order by performance is considerably worse, we have to JOIN in order to propagate up values from nested objects; thus the pathological case is for users with lots of objects, and who want to order by deeply nested fields.

Example query, filtering by a nested ref:
```python
client.get_calls(
   query={"$expr": {"$and": [{"$eq": [{"$getField": "inputs.worker_config.temp.val"}, {"$literal":1}]}]}]},
   expand_columns=['inputs.worker_config', 'inputs.worker_config.temp']
)
```

```sql
-- EXPLAIN indexes=1
WITH obj_filter_0 AS (
    SELECT 
        digest,
        CONCAT('weave-trace-internal:///', project_id, '/object/', object_id, ':', digest) AS full_ref
    FROM object_versions
    WHERE project_id = 'UHJvamVjdEludGVybmFsSWQ6NDE5NjExMDA='
      AND toFloat64OrNull(JSON_VALUE(val_dump, '$.val')) = 1
    GROUP BY project_id, object_id, digest

   UNION ALL

    SELECT 
       digest, 
       digest as ref
    FROM table_rows
    WHERE project_id = {pb_0:String}
         AND toFloat64OrNull(JSON_VALUE(val_dump, '$.val')) = 1
     GROUP BY project_id, digest
),
obj_filter_1 AS (
    SELECT 
        digest, 
        CONCAT('weave-trace-internal:///', project_id, '/object/', object_id, ':', digest) AS full_ref
    FROM object_versions
    WHERE project_id = 'UHJvamVjdEludGVybmFsSWQ6NDE5NjExMDA='
       AND length(refs) > 0
       AND JSON_VALUE(val_dump, '$.temp') IN (SELECT full_ref FROM obj_filter_0)
    GROUP BY project_id, object_id, digest
)
SELECT 
  id, ...
FROM calls_merged
WHERE project_id = 'UHJvamVjdEludGVybmFsSWQ6NDE5NjExMDA='
  -- potential optimization --
  -- AND length(calls_merged.input_refs) > 0
GROUP BY project_id, id
HAVING (((JSON_VALUE(any(calls_merged.inputs_dump), '$."worker_config"') IN
        (SELECT ref
         FROM obj_filter_0)
     OR regexpExtract(JSON_VALUE(any(calls_merged.inputs_dump), '$."worker_config"'), '/([^/]+)$', 1) IN
        (SELECT ref
         FROM obj_filter_0)))
     AND ((any(calls_merged.deleted_at) IS NULL))
ORDER BY any(started_at) DESC
LIMIT 50
```

Handles trivial case for evals
filter-by-ref-eval

Should propagate to export view as well, for both filter and order: 
image

Performance of this query in a project with 150,000+ heavy calls, and 10,000 objects:
Latency: `~1s`
Rows read: `581,133`
Peak memory: `3.52 GiB` (mostly because of the calls)

Performance should scale linearly with # of objects and calls: `O(n + m)`. Uses `project_id` primary key where possible:
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -135,8 +135,8 @@ langchain_nvidia_ai_endpoints = [
   "langchain-nvidia-ai-endpoints",
 ]
 litellm = ["litellm>=1.36.1"]
-# temporary max pin b/c 0.12.0 changes call structure and therefor tests
-llamaindex = ["llama-index>=0.10.35", "openai<1.92.0"]
+# temporary max pin b/c 0.12.50 changes call structure and therefor tests
+llamaindex = ["llama-index>=0.10.35,<0.12.50", "llama-index-core<0.12.50", "openai<1.92.0"]
 mistral = ["mistralai>=1.0.0"]
 scorers = [
   "numpy>1.21.0",


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -422,11 +422,26 @@ export const CallsTable: FC<{
             field,
             rowId
           );
+          const op = operator ? operator : getDefaultOperatorForValue(value);
```
neutralino1:
```suggestion
          const op = operator ?? getDefaultOperatorForValue(value);
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -422,11 +422,26 @@ export const CallsTable: FC<{
...
+            // 1. the ref
+            // 2. the value of the ref
+            // This is because the ref is not a real column, and we need to filter by the
+            // value of the ref.
+            // filterModel.items.push({
+            //   field: expandedRef.field,
+            //   operator: op,
+            //   value: expandedRef.value,
+            // });
+            console.log('setting expanded ref cols', expandedRef.field);
```
neutralino1: 🔪

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/callsTableQuery.ts
```diff
@@ -145,7 +146,7 @@ export const useCallsForQuery = (
     limit: effectiveLimit,
     columns: storageSizeCols,
     expandedRefColumns: expandedColumns,
-    skip: calls.loading || noCalls || !options?.includeTotalStorageSize,
+    skip: calls.loading || noCalls,
```
neutralino1: Intentional?

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -352,19 +342,80 @@ def as_sql(self, pb: ParamBuilder, table_alias: str) -> str:
...
+            expand_columns
+            and field_to_object_join_alias_map
+            and has_object_ref_field(self.raw_field_path, expand_columns)
+        ):
+            order_condition = ObjectRefOrderCondition(
+                field_path=self.raw_field_path,
+                expand_columns=expand_columns,
+            )
+            cte_alias = field_to_object_join_alias_map.get(order_condition.unique_key)
+            if cte_alias:
```
neutralino1, line 354: Nit
```suggestion

            if cte_alias := field_to_object_join_alias_map.get(order_condition.unique_key):
```

weave/trace_server/calls_query_builder/object_ref_query_builder.py
```diff
@@ -0,0 +1,930 @@
...
+        expand_columns: list[str],
+        case_insensitive: bool = False,
+        conversion_type: Optional[
+            Literal["double", "string", "int", "bool", "exists"]
+        ] = None,
+    ):
+        self.field_path = field_path
+        self.expand_columns = expand_columns
+        self.case_insensitive = case_insensitive
+        self.conversion_type = conversion_type
```
neutralino1: Pydantic?

tests/trace/test_weave_client.py
```diff
@@ -3569,3 +3569,489 @@ def test_sum_dict_leaves_deep_nested(client):
         },
         "top": [10, "mixed", 20],
     }
+
```
tssweeney: 2 things:
1. I am not seeing a test where you have a mix of rows (some of those rows have refs at the ref locations, where others just have data payloads). I am wondering if your query works in this case.
2. (less important) Would be nice to just have a new test file to isolate this component of the query layer (this test file is huge)

tests/trace_server/test_calls_query_builder.py
```diff
@@ -2283,6 +2283,20 @@ def test_filter_length_validation():
...
+    complex_conditions = {
+        "$and": [
+            {"$eq": [{"$getField": f"inputs.field{i}"}, {"$literal": i}]}
+            for i in range(20)  # Exceeds MAX_CTES_PER_QUERY
+        ]
+    }
+    cq.add_field("id")
+    cq.add_condition(complex_conditions)
+    cq.set_expand_columns(["inputs"])
+    with pytest.raises(ValueError, match="Too many object reference conditions"):
```
tssweeney, line 2297: smart

tests/trace_server/test_calls_query_builder.py
```diff
@@ -2461,3 +2475,696 @@ def test_thread_id_and_turn_id_filter_combined():
...
+                "$eq": [
+                    {"$getField": "inputs.model.temperature"},
+                    {"$literal": 1},
+                ]
+            }
+        )
+    )
+    cq.add_order("started_at", "desc")
+    cq.set_expand_columns(["inputs.model"])
+    assert_sql(
```
tssweeney: reading this example, I am starting to imply that your design is:
1. split all filter paths by the expand columns
2. introduce intermediate queries for each "subpath" of the filter.

If this implication is correct, i think there is a feature constraint here that that expand columns in themselves become filters (effectively: must have refs at these paths). For example, given two rows:

```
{key_a: {key_b: val}}
```

and

```
{key_a: REF}
REF = {key_b: val}
```

then:
```
filter $.key_a.key_b == val
```

will match the first row only, and if you specify an expand column of `key_a` then it will only match the second row.

---

I do think this is better than what we have, but admittedly a gap in the most pure sense of the feature.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
```diff
@@ -432,11 +432,16 @@ export const CallsTable: FC<{
             field,
             rowId
           );
+          const op = operator ?? getDefaultOperatorForValue(value);
           if (expandedRef != null) {
-            value = expandedRef.value;
-            field = expandedRef.field;
+            setExpandedRefCols(prevState => {
```
tssweeney: this seems counter intuitive to me. Why do we remove the inputs.example when updating the expandedRef?

weave/trace_server/trace_server_interface.py
```diff
@@ -423,6 +423,13 @@ class CallsQueryReq(BaseModel):
         examples=[["inputs.self.message", "inputs.model.prompt"]],
         description="Columns to expand, i.e. refs to other objects",
     )
+    # Internal flag used for playing nice with the recursive object ref handling
```
tssweeney: this isn't really internal-only as it is used by the UI client and potentially python clients in the future. You basically are saying we have two modes:
1. expand_columns is only used to specify filter/sort expansion
2. expand_columns does 1 AND returns these expanded values.

where today we have: 3: expand the values.

I think you might want to call this: `return_expanded_column_values` or something that more clearly describes the purpose of the field

weave/trace_server/calls_query_builder/object_ref_query_builder.py
```diff
@@ -0,0 +1,914 @@
...
+
+    field_path: str
+    expand_columns: list[str]
+    case_insensitive: bool = False
+    conversion_type: Optional[tsi_query.CastTo] = None
+
+    @property
+    def is_table_rows_condition(self) -> bool:
+        # TODO: is this the only time this is an actual table join condition??
+        if (
```
tssweeney: no! this is circumstantial it is perfectly possible for a table row to have a different path && for inputs.example to NOT be a table row

tssweeney: you only know if something is a table by expanding it until you "hit" a table. 

doing this here would probably work for 99% of cases, but is certainly an incorrect assumption and something that would easily break if someone decided to call a param something different

tssweeney (commented):
There are 2 major considerations I can see:

1. We are excluding results which lie on or beyond an expanded path that are not mediated by a ref at that location
2. The table detection is very happenstance and not robust to different op patterns

tests/trace_server/test_calls_query_builder.py
gtarpenning: If i'm understanding you correctly, it should be the exact same paradigm as the existing "expand_columns" where you _must_ pass in every ref. This current implementation works by going bottom up, starting at the leaf nodes of the tree (The filter values) which are objects or table_rows, and then walking back up to the calls. The advantage of this is that by the time we have joined with the calls table we should be able to select only the calls where the specific condition is met. However, as you have identified, it's not exactly flexible, you can't have either a ref or not a ref and meet the same condition. To achieve this we would need to do a forward pass from the calls as root, which has considerably worse performance, as designed currently (if it's even possible to massage a one-time sql query to be that flexible). I honestly can't even convince myself it's possible to do a forward pass. 

To your specific concern, this would be a problem in one case: A user logs an object as a ref, then later dumps that object to dict (or vice versa) and wants to filter on all call by a nested path in that object/dict. In this case, the user would either expand the ref using the header in the table and only get back reffed objects, or not expand the header and only filter on the raw fields. I think this is probably an edge case, but we could probably mitigate this by including an additional OR condition on the raw field (do ref filter stuff OR json_value(raw path)). This would only work for one layer of nesting, but probably would cover the lionshare of the edge case.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/CallsPage/CallsTable.tsx
gtarpenning: hmm, probably don't need to, but in this case its _always_ present which was messing up ref expansion, at least in an early version. i'll fix that.

weave/trace_server/calls_query_builder/object_ref_query_builder.py
gtarpenning: > you only know if something is a table by expanding it until you "hit" a table.

That is surprisingly non-trivial to do. The reason is fundamentally because in the trace table (and obviously payloads) we allow different objects in the same field. For example, I can log a config called `Config1` at key "config" but then later log a new object `ConfigFormatted` in the same key. This ends up meaning that we can't use object_ids when filtering for objects, we must look at the actual values of the object. If we are looking at values, we don't know the ref structure, and thus we don't know if a given ref is to a table or not. Moreover, what if a user logs a reference to a dataset row under key "test1" and then later logs an object under that same key? We now have two different kinds of refs in that key, but in the calls table we will happily unroll them and the user would expect to be able to filter on them... Yikes... This just got gnarly. The only way I can see to making this work would be to always check both the table_rows table AND the objects table for any given ref, doubling the number of CTE's required. That's probably not too bad for most queries, but ensures that the performance of deeply nested stuff is going to be crippled.

weave/trace_server/calls_query_builder/object_ref_query_builder.py
```diff
@@ -0,0 +1,909 @@
...
+            UNION ALL
+
+            SELECT
+                digest,
+                {val_dump_select}
+                digest as ref
+            FROM table_rows
+            WHERE project_id = {param_slot(project_param, "String")}
+            {val_condition_sql}
+            GROUP BY project_id, digest
```
gtarpenning: leaf objects can be in either `table_rows` or `object_versions`

weave/trace_server/trace_server_interface.py
```diff
@@ -435,6 +455,7 @@ class CallsQueryStatsReq(BaseModel):
     query: Optional[Query] = None
     limit: Optional[int] = None
     include_total_storage_size: Optional[bool] = False
+    expand_columns: Optional[list[str]] = None
```
tssweeney: what does this do?

gtarpenning: pretty much the same as in the calls query, i will add a comment and nice type. ALthough we don't get the columns **back**, we still need to tell the query engine which columns are refs. In order to get an accurate count of a filtered column, we pass in where the refs are in the columns that we are filtering by.

weave/trace_server/trace_server_interface.py
```diff
@@ -417,12 +417,32 @@ class CallsQueryReq(BaseModel):
...
         description="Columns to expand, i.e. refs to other objects",
     )
+    # Controls whether or not to return expanded ref columns. In most clients,
+    # refs are resolved recursively by making additional api calls, either for
+    # performance or convenience reasons. In that case, we do not want to return
+    # resolved refs. However, expand_columns still must contain paths to all
+    # refs when filtering or sorting. Set this value to false to filter/order
+    # by refs but rely on client methods for actually resolving the values. The
+    # default is to resolve and return expanded values when expand_columns is set.
+    return_expanded_column_values: Optional[bool] = Field(
```
tssweeney, line 440: Do we expect users to use this field, or is it something that is more internal?

gtarpenning: internal, the default to true should be what 99.9% user queries want.

gtarpenning:
> Are you going to split this into backend and frontend? I am ready to approve the backend so we can get that in and deployed

@tssweeney  Frontend will be removed from this pr and moved here: https://github.com/wandb/weave/pull/5062

tssweeney (approved):

PR #4843, opened 2025-06-23
feat(weave): Revert "Pick up dedicated feature flag from core (#4839)"

neutralino1:
This reverts commit 49cd67dc77ebfb85880e36abc6c4a368f34c8a41.

This is actually not necessary as `envIsLocal` is already included in `isWandbAdmin`.

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
@@ -2,7 +2,6 @@ import {ApolloProvider} from '@apollo/client';
 import {Box, Drawer} from '@mui/material';
 import {LicenseInfo} from '@mui/x-license';
 import {makeGorillaApolloClient} from '@wandb/weave/apollo';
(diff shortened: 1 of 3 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -420,11 +419,9 @@ const Browse3ProjectRoot: FC<{
...
           <CallsPageBinding />
         </Route>
-        {!getConfig().IS_DEDICATED && (
-          <Route path={`${projectRoot}/monitors`}>
-            <MonitorsPageBinding />
-          </Route>
-        )}
+        <Route path={`${projectRoot}/monitors`}>
+          <MonitorsPageBinding />
+        </Route>
```
gtarpenning, line 424: do we not want to gate this?

neutralino1: Not a big deal IMO.

PR #4852, opened 2025-06-24
fix(flow): Allow MessagesPrompt to have nested dict messages

aduverger:
## Description

This PR allow `MessagesPrompt` to format nested dict messages.
Currently, if `MessagesPrompt` is similar to:

```python
prompt = MessagesPrompt(
    [
        {"role": "system", "content": "You are a pirate."},
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": "Tell us your thoughts on {topic}.",
                }
            ],
        },
    ]
)
```
Then `{topic}` will not be replaced when using `prompt.format(topic="airplanes")`.

This is because the `.format()` method considers that `content`'s value is a `str`.

Yet, LLM client do allow `content` to be a `list[dict]`, see [this example](https://docs.mistral.ai/capabilities/OCR/document_qna/) with Mistral OCR:
```python
messages = [
    {
        "role": "user",
        "content": [
            {
                "type": "text",
                "text": "what is the last sentence in the document"
            },
            {
                "type": "document_url",
                "document_url": "https://arxiv.org/pdf/1805.04770"
            }
        ]
    }
]
```

## Testing

Added a unit test to cover this use case.

### DIFF
--- a/weave/flow/prompt/prompt.py
+++ b/weave/flow/prompt/prompt.py
@@ -23,7 +23,7 @@ from weave.trace.vals import WeaveObject
 
 class Message(TypedDict):
     role: str
-    content: str
+    content: str | list[dict]
 
 
 def maybe_dedent(content: str, dedent: bool) -> str:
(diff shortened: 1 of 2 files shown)


aduverger:

aduverger:
Not sure how this PR might not get lost into the 333 opened PR 😂
If you have time to have a look @andrewtruong @gtarpenning @jamie-rasmussen 🙏🏻

jamie-rasmussen:
Thank you for the PR! It is on my TODO list for today.

weave/flow/prompt/prompt.py
```diff
@@ -23,7 +23,7 @@
 
 class Message(TypedDict):
     role: str
-    content: str
+    content: str | list[dict]
```
jamie-rasmussen:
```suggestion
    content: Union[str, list[dict]]
```

jamie-rasmussen: Needed for Python 3.9 which we still support

jamie-rasmussen:
Thank you for the contribution!

PR #4858, opened 2025-06-24
fix(weave): disable auto retries in gcs boto causing long busy waits

gtarpenning:
## Description

[WB-25670](https://wandb.atlassian.net/browse/WB-25670)

I believe the issue was that the gcp client automatically retries "retryable" exceptions, per [this](https://cloud.google.com/python/docs/reference/storage/latest/retry_timeout) doc. So we were actually retrying like 5 x 3 times! Disable the auto retries, rely on our own handling.

This probably also explains why the timeouts were longer, and there are no logs for the retries! 

## Testing

Adds test

[WB-25670]: https://wandb.atlassian.net/browse/WB-25670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/test_server_file_storage.py
+++ b/tests/trace/test_server_file_storage.py
@@ -181,12 +181,12 @@ class TestGCSStorage:
         # In-memory storage for all blobs
         blob_data = {}
 
-        def mock_upload_from_string(data, timeout=None):
+        def mock_upload_from_string(data, timeout=None, **kwargs):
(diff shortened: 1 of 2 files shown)


zbirenbaum (approved):
Looks good to me!

PR #4863, opened 2025-06-24
fix(weave): Include entity in weave.init in Python code sample

jamie-rasmussen:
## Description

Internal Jira: https://wandb.atlassian.net/browse/WB-25727
Internal Slack: https://weightsandbiases.slack.com/archives/C01T8BLDHKP/p1750792719999149

We should scope with the entity when doing a `weave.init` in the Python export code sample - if your default entity is not the one you are looking at, the code sample could confusingly create a different project with the same name and the query would return 0 results.

Note that the curl code sample does already include the entity.

## Testing

How was this PR tested?


gtarpenning (approved):
lgtm, should have been the same between curl and python anyway. thx

PR #4872, opened 2025-06-24
feat(weave): Frontend implementation for universal content

zbirenbaum:
## Description

 - Adds Content weave type handling to the frontend
 - Unifies all existing content to use the same display component as their content counterpart for consistency
 - Minimal changes (except for consistency) to all existing media displays

Includes custom content displays for the following media types:
 - Audio
 - Video
 - Images
 - PDF

Includes a fallback component which displays non-bolded text that shows metadata on hover

image


neutralino1 (commented):
DO I understand correctly that this is moving code around with minimal changes?

zbirenbaum:
> Do I understand correctly that this is moving code around with minimal changes?

Essentially yes

There were some minor changes to allow for reusing the same components across the new and old media types, as well as some consistency changes like which tooltips are used and what the files are called when downloaded but overall the impact and surface area is a drastic step down from my initial attempt at this.

neutralino1 (approved):

zbirenbaum:
> > Do I understand correctly that this is moving code around with minimal changes?
> 
> Essentially yes
> 
> There were some minor changes to allow for reusing the same components across the new and old media types, as well as some consistency changes like which tooltips are used and what the files are called when downloaded but overall the impact and surface area is a drastic step down from my initial attempt at this.

I will point out that all of the stuff in ContentView and the components it calls (except pdf) are new, but the underlying components they delegate to are moved.

PR #4877, opened 2025-06-25
chore(weave): Add T20 lint rule

andrewtruong:
https://docs.astral.sh/ruff/rules/#flake8-print-t20


jamie-rasmussen (approved):
I wonder if this one will end up being annoying during local development when you might want to run ruff before you are ready to push.

PR #4879, opened 2025-06-25
chore(weave): Add PT006 lint rule

andrewtruong:
https://docs.astral.sh/ruff/rules/pytest-parametrize-names-wrong-type/


tests/trace/test_file_storage_uris.py
```diff
@@ -188,7 +188,7 @@ def test_empty_required_components(storage_class, bad_args):
 
 
 @pytest.mark.parametrize(
-    "original_uri",
+    ("original_uri"),
```
jamie-rasmussen: Are these parens expected?

andrewtruong: fixed

PR #4883, opened 2025-06-25
feat(weave): Update docs links for online evals

neutralino1:
Update documentation links for online evals.


J2-D2-3PO (approved):
Links are correct, LGTM. Per Jamie's point, not sure if we need to update for the analytics tho

neutralino1:
Ah! How do we configure those link? @jamie-rasmussen

jamie-rasmussen:
> Ah! How do we configure those link? @jamie-rasmussen

Sent some info in DM. Stamping to unblock in case there's trouble getting that setup.

PR #4884, opened 2025-06-25
chore(weave): Add PT023 lint rule

andrewtruong:
https://docs.astral.sh/ruff/rules/pytest-incorrect-mark-parentheses-style/


pyproject.toml
```diff
@@ -292,6 +294,9 @@ exclude = ["weave_query"]
 [tool.ruff.lint.isort]
 known-third-party = ["wandb", "weave_query"]
 
+[tool.ruff.lint.flake8-pytest-style]
+mark-parentheses = false
```
jamie-rasmussen, line 299: Seems like this is the expected default, are you just being explicit? https://docs.astral.sh/ruff/settings/#lint_flake8-pytest-style_mark-parentheses

jamie-rasmussen (approved):

PR #4890, opened 2025-06-26
fix(weave): Anthropic system message and tool call missing

zbirenbaum:
## Description

- Fixes WB-25757

Two issues fixed here:
1. System prompt not displaying properly
2. Anthropic style tool calls were not being translated which caused rendering issues

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
@@ -147,13 +147,11 @@ export const normalizeChatRequest = (request: any): ChatRequest => {
   if (isTraceCallChatFormatOAIResponsesRequest(request)) {
     return normalizeOAIResponsesRequest(request);
   }
-  if (isAnthropicCompletionFormat(request)) {
-    return normalizeAnthropicChatRequest(request);
-  }
   if (isTraceCallChatFormatOTEL(request)) {
     return normalizeOTELChatRequest(request);
   }
-  return request as ChatRequest;
+  // Anthropic is close but system and nested tool calls handled differently
+  return normalizeAnthropicChatRequest(request) as ChatRequest;
 };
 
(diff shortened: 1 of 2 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
```diff
@@ -147,13 +147,11 @@ export const normalizeChatRequest = (request: any): ChatRequest => {
...
   }
-  if (isAnthropicCompletionFormat(request)) {
-    return normalizeAnthropicChatRequest(request);
-  }
   if (isTraceCallChatFormatOTEL(request)) {
     return normalizeOTELChatRequest(request);
   }
-  return request as ChatRequest;
+  // Anthropic is close but system and nested tool calls handled differently
+  return normalizeAnthropicChatRequest(request) as ChatRequest;
```
neutralino1: I don't understand this change. Don't we want to test `isAnthropicCompletionFormat` before we call `normalizeAnthropicChatRequest`?

zbirenbaum: `isAnthropicCompletionFormat` tests the outputs while this function processes the inputs. This was a bug. The line actually should have been: `isAnthropicRequestFormat` 

There isn't actually an `isAnthropicRequestFormat` because previously the only difference in how we handled it was that the system prompt was a top level field rather than an actual message.

E.g. If you pretend there was a function called `isOpenAIRequestFormat` (this is our default) it would return true since it's really hard to differentiate.

There are also some slight differences in how tool call messages are shown in the inputs which are only detectable by iterating over the full data anyways, which means we may as well just process it in place rather than detecting first.

I think the whole way we do detection needs to be redone, but that's a pretty extensive refactor


Let me know if that makes sense. I know it's a little counterintuitive

neutralino1 (approved):
Thank you. Please fix merge conflict and merge.

PR #4892, opened 2025-06-26
Switch default hub for Local Models from artifacts to hf hub

morganmcg1:


morganmcg1:
Final test passing depends on this PR being merged:

https://github.com/wandb/weave/pull/4941

tests/scorers/test_bias_scorer.py
```diff
@@ -17,14 +17,14 @@ def weave_bias_scorer():
 def test_bias_scorer_simple(weave_bias_scorer):
     """Tests that the scorer can handle a basic string."""
     output = "This is a balanced statement with no bias."
-    result = weave_bias_scorer.score(output)
-    assert not result.passed
+    result = weave_bias_scorer.score(output=output)
+    assert result.passed
```
andrewtruong, line 22: Why is this the opposite now?

tests/scorers/test_coherence_scorer.py
```diff
@@ -21,7 +21,7 @@ def test_score_messages(weave_coherence_scorer):
     result = weave_coherence_scorer._score_messages(query, output)
     # Check that the pydantic model has the expected attributes.
     assert result.metadata is not None
-    assert result.metadata["coherence_label"] == "A Little Incoherent"
+    assert result.metadata["coherence_label"] == "Perfectly Coherent"
```
andrewtruong, line 24: Why did this change?

tests/scorers/test_utils.py
```diff
@@ -19,6 +24,11 @@
     "llamaguard_scorer": "c-metrics/weave-scorers/llamaguard_scorer_tiny:latest",
 }
```
andrewtruong, line 24: Didn't you remove these?  Why are they still here?

weave/scorers/utils.py
```diff
@@ -65,14 +65,33 @@ def ensure_hf_imports() -> None:
...
 ) -> str:
     """Load the local model weights for a Hugging Face model.
 
-    If model_name_or_path is a directory, it is assumed to be the local model weights path.
-    If model_name_or_path is provided (non-empty), it is used to download the model using the existing download_model function.
-    If no model_name_or_path is provided, and a default_model is supplied, it downloads the default model.
+    If a default_model is supplied, it downloads the default model from the
+    Hugging Face Hub. Otherwise, if model_name_or_path is a directory, it is
+    assumed to be the local model weights path. If model_name_or_path is provided
+    and is not a local directory, it will be downloaded from wandb artifacts.
```
andrewtruong: Dont couple the implementations.  Write a third function that will do this dispatch

tests/scorers/test_bias_scorer.py
morganmcg1: looks like it was previously using one of the TINY models, these are the same architecture but have most of their middle layers taken out so they're lightweight. I don't think we should have been using these for tests where we're asserting that the model evaluates some text correctly.

I've updated the model to use the full scorer, which will correctly return `result.passed == True`

tests/scorers/test_coherence_scorer.py
morganmcg1: Same as above, using the full model weights (and looking at the text given to the model), the correct category should be "Perfectly Coherent". looks like the previous assert was in place to just match the (flawed) output of the TINY model.

tests/scorers/test_utils.py
morganmcg1: not sure why they're still there, must be a hangover from the llamaguard implementation, removed

morganmcg1:
@andrewtruong comments addressed. Make a few more fixes, mostly adding the full weave local scorers model weights to the other tests instead of the tiny model weights which means we're doing proper evaluation of the outputs (and not just ensuring that they run).

PR #4896, opened 2025-06-26
perf(weave): slight refactor to storage query, big perf improvement

gtarpenning:
## Description

[WB-25774](https://wandb.atlassian.net/browse/WB-25774)

Refactors for performance aggregation of call storage. This query in prod fails for large projects. Due to not grouping by call_id, this new query is technically slightly different than prod. While the # of calls is consistent, the bytes is off by <1%. I wonder if this is a big deal....

[Failing example dd trace.](https://us5.datadoghq.com/logs?query=trace_id%3A685d8f9400000000c16e4b1cdbfbd301&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZeteO2o_kro-AAAABhBWmV0ZVFNY0FBREJ0c1hIZm03S0dBQUoAAAAkMDE5N2FkNzktMjA5Ni00YmVlLTkwMmUtNzYzZGNjMjFhY2M2AAA58Q&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=desc&viz=stream&from_ts=1750961068851&to_ts=1750963073745&live=false)

## Testing

|   env   | duration | rows       | memory      | # calls | bytes |
|---------|----------|------------|-------------| ------- | ----- |
| prod    | 8.114s   | 131259491  | 29.25  GiB  | 65517099 | 731193577615 |
| branch  | 2.637s   | 131259797  | 583.01 MiB  | 65517099 | 731228264910 |

[WB-25774]: https://wandb.atlassian.net/browse/WB-25774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/calls_query_builder/calls_query_builder.py
+++ b/weave/trace_server/calls_query_builder/calls_query_builder.py
@@ -818,7 +818,8 @@ class CallsQuery(BaseModel):
                 FROM calls_merged_stats
                 WHERE project_id = {param_slot(project_param, "String")}
                 GROUP BY id
-            ) as {STORAGE_SIZE_TABLE_NAME} USING id
+            ) as {STORAGE_SIZE_TABLE_NAME}
+                ON calls_merged.id = {STORAGE_SIZE_TABLE_NAME}.id
             """
 
         total_storage_size_sql = ""
@@ -837,7 +838,8 @@ class CallsQuery(BaseModel):
                 FROM calls_merged_stats
                 WHERE project_id = {param_slot(project_param, "String")}
                 GROUP BY trace_id
-            ) as {ROLLED_UP_CALL_MERGED_STATS_TABLE_NAME} USING trace_id
+            ) as {ROLLED_UP_CALL_MERGED_STATS_TABLE_NAME}
+                ON calls_merged.trace_id = {ROLLED_UP_CALL_MERGED_STATS_TABLE_NAME}.trace_id
             """
 
         raw_sql = f"""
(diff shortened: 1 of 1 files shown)


gtarpenning:
Actually performant query:

```sql

SELECT
    countIf(deleted_at IS NULL AND started_at IS NOT NULL) AS calls,
    sumIf(total_storage_size_bytes,
          parent_id IS NULL
          AND deleted_at IS NULL
          AND started_at IS NOT NULL) AS bytes
FROM calls_merged

-- /* light one-to-one join kept in RAM once */
ANY LEFT JOIN
(
    SELECT
        trace_id,
        sum(
            coalesce(attributes_size_bytes, 0) +
            coalesce(inputs_size_bytes,     0) +
            coalesce(output_size_bytes,     0) +
            coalesce(summary_size_bytes,    0)
        ) AS total_storage_size_bytes
    FROM calls_merged_stats
    WHERE project_id = 'UHJvamVjdEludGVybmFsSWQ6NDEzNzg0NTY='
    GROUP BY trace_id
) AS rolled USING trace_id
WHERE project_id = 'UHJvamVjdEludGVybmFsSWQ6NDEzNzg0NTY='
SETTINGS enable_filesystem_cache=0;
```

We would have to refactor the SELECT considerably, way out of scope. Prefer frontend only change: https://github.com/wandb/weave/pull/4898

PR #4901, opened 2025-06-26
feat(weave): add inference service chat client

jamie-rasmussen:
## Description

Adds model listing and chat completion APIs to the client returned by `weave.init`. These APIs are intended to be mostly compatible with OpenAI to allow quickly swapping them in.

Making these APIs available in the Weave client has a number of advantages vs. our currently recommended approach of using the OpenAI Python SDK:

* Don't have to add a dependency on `openai` if you don't already have one.
* Simpler code. User doesn't have to specify `base_url` or their W&B `api_key` when constructing the client. Also don't have to specify the scoped project ID a second time.

The client allows either talking directly to the Inference Service or going through the Trace Server Playground APIs (the default), controlled by the `endpoint` argument. An advantage of going through the Playground APIs is that you are not limited to the CoreWeave hosted models but could use any that are available in the playground.

My idea is to ship this without fanfare, let internal people try it out for a little and then when we are happy with it add to Weave docs and update the Model Catalog examples UI to have this as the recommended usage (while maintaining a tab showing how to access hosted models through `openai`).

Note that most of the code in the `types` folder is lightly edited versions of files from the OpenAI SDK (which uses the same open source license we do.)

### List models
Note, this only returns information about hosted models. Probably we should move information about other available models to the backend and have a Trace Server endpoint that returns it.
```
import weave
client = weave.init("wandb/2025-06-25_inference")

for model in client.inference_models.list():
    print(model.id)
```

### Completion:
```
import weave
client = weave.init("wandb/2025-06-25_inference")
    model="coreweave/meta-llama/Llama-3.1-8B-Instruct",
    messages=[
        {"role": "user", "content": "Tell me a joke."}
    ],
)
print(response.choices[0].message.content)
```

### No tracing
```
import weave
client = weave.init("wandb/2025-06-25_inference")
    model="coreweave/meta-llama/Llama-3.1-8B-Instruct",
    messages=[
        {"role": "user", "content": "Tell me a joke."}
    ],
    track_llm_call=False,
)
print(response.choices[0].message.content)
```

### Talk directly to the Inference Service instead of going through the Playground
```
import weave
client = weave.init("wandb/2025-06-25_inference")
    model="coreweave/meta-llama/Llama-3.1-8B-Instruct",
    messages=[
        {"role": "user", "content": "Tell me a joke."}
    ],
    endpoint="inference",
)
print(response.choices[0].message.content)
```

### Streaming completion
```
import weave
client = weave.init("wandb/2025-06-25_inference")
    model="coreweave/meta-llama/Llama-3.1-8B-Instruct",
    messages=[
        {"role": "user", "content": "Tell me a joke."}
    ],
    stream=True,
)
for chunk in response:
    content = chunk.choices[0].delta.content
    if content:
        print(content, end="", flush=True)
```

### Using a non-hosted model
```
import weave
client = weave.init("wandb/2025-06-25_inference")
    model="claude-sonnet-4-20250514",
    messages=[
        {"role": "user", "content": "Tell me a joke."}
    ],
)
print(response.choices[0].message.content)
```

## Testing

How was this PR tested?


weave/trace/weave_client.py
```diff
@@ -968,6 +970,8 @@ def __init__(
         if hasattr(self.server, "get_call_processor"):
             self._server_call_processor = self.server.get_call_processor()
         self.send_file_cache = WeaveClientSendFileCache()
+        self.models = Models(entity, project)
```
tssweeney: this is only going to list models hosted by coreweave. I feel like this is going to be misleading and potentially overloading our own Weave Models. I wonder if we should be calling this FoundationModels or HostedModels or even CoreWeaveModels something that prevents the namespace and product collision with Weave Models. Do we eventually expect this to return the known models from FM vendors?

weave/chat/completions.py
```diff
@@ -0,0 +1,263 @@
...
+            data["stream"] = stream
+        if stream_options is not NOT_GIVEN:
+            data["stream_options"] = stream_options
+        if temperature is not NOT_GIVEN:
+            data["temperature"] = temperature
+        if top_p is not NOT_GIVEN:
+            data["top_p"] = top_p
+
+        auth = None
+        if endpoint == "playground":
```
tssweeney, line 221: other than for development debugging, I am not sure why the user would want to switch between these endpoint modes. Is this intended to be user-facing?

weave/chat/chat.py
```diff
@@ -0,0 +1,6 @@
+from weave.chat.completions import Completions
+
+
+class Chat:
+    def __init__(self, entity: str, project: str):
```
neutralino1: Maybe a minimal docstring? (why is this here?)

weave/chat/common.py
```diff
@@ -0,0 +1,4 @@
+INFERENCE_HOST = "api.inference.wandb.ai"
+
+# Keep this in sync with weave.trace_server.constants which is what the Playground uses.
```
neutralino1: Why not re-use that constant?

weave/chat/completions.py
```diff
@@ -0,0 +1,263 @@
...
+        self.project = project
+
+    def create(
+        self,
+        *,
+        endpoint: Endpoint = "playground",
+        track_llm_call: bool = True,
+        messages: Iterable,
+        model: str,
+        frequency_penalty: float | None | NotGiven = NOT_GIVEN,
```
neutralino1, line 52: Is the `NOT_GIVEN` pattern something prescribed by the OpenAI SDK, or is it something new we're introducing?

weave/chat/completions.py
neutralino1: Playground offers tracking but low perf, inference is not tracked but better perf?

It seems relevant to be able to switch. And given that the logic is different, we kinda have to support the switch at this level (as opposed to simply letting users switch `base_url`)

neutralino1: There is a Hackathon project to offer Weave tracking by default for the inference endpoints. If this ever lands, then maybe we don't need to support it here?

weave/chat/models.py
```diff
@@ -0,0 +1,43 @@
...
+    ModelsResponseError,
+    ModelsResponseSuccess,
+)
+from weave.wandb_interface.wandb_api import get_wandb_api_context
+
+
+class Models:
+    def __init__(self, entity: str, project: str):
+        self.entity = entity
+        self.project = project
```
neutralino1: Does this break compatibility with the OpenAI SDK? Should we pick those up from the context instead?

weave/chat/models.py
```diff
@@ -0,0 +1,43 @@
...
+        headers = {
+            "Authorization": f"Bearer {api_key}",
+            "OpenAI-Project": f"{self.entity}/{self.project}",
+            "Content-Type": "application/json",
+        }
+        url = f"https://{INFERENCE_HOST}/v1/models"
+        response = requests.post(url, headers=headers)
+        d = response.json()
+
+        if "error" in d:
```
neutralino1: Any better way to detect an error? like `response.status_code`?

weave/trace/weave_client.py
jamie-rasmussen: Great point. There's a question of what we want the user to be able to do, and then a subsequent one around naming. I think users should be able to enumerate all of the models that are available to the project through the playground. This interface was designed to be a drop-in replacement for OpenAI's, but likely we would want to extend it to provide additional information, e.g. we could potentially return information about models that aren't available for this project, but would be if you configured an API key, and whether to include models with that status could be a parameter to the list method.

The issue we have today is that that set of models is configured in the UI, we don't have a nice way to get it. So this is only talking to the inference service and getting inference service models. :-( 

The naming conflict between models here and Weave models is also definitely a concern. Not having a way to list objects in our SDK is a gap. It's not clear to me if we would have SDK-side methods for listing each of the subtypes of Object. Maybe there is just a generic `list_objects` with a type filter? Maybe some of the key user facing things like models or datasets get `list_*` methods but not all subtypes do?

Overall, I think the "drop-in compatibility" is going to be way more useful for completions that models, so maybe not worth the additional confusion to make it work here.

weave/chat/completions.py
jamie-rasmussen: Debugging (including through asking the user to try it) was the primary motivation, I was also anticipating a performance boost if people communicate directly with the inference service, but haven't yet measured how much of a difference it makes. (But having the parameter there for now will help with that!) We could potentially change the name to be underscore-prefixed.

jamie-rasmussen: > There is a Hackathon project to offer Weave tracking by default for the inference endpoints. If this ever lands, then maybe we don't need to support it here?

It's not clear to me that having the inference service endpoint itself log a trace is architecturally desirable and especially considering you usually want the LLM call logged in the context of a whole trace tree.

jamie-rasmussen:
> I am sure there is a strong reason here - can you share why it is beneficial to copy the OpenAI type shapes and have our own completion implementation as opposed to using the OpenAI SDK itself?

The overall motivation is in the PR description, but for types specifically:
* Avoiding adding a dependency (one that is constantly changing)
* Inconsistencies (current and future) between OpenAI SDK and what Inference Service returns.

weave/chat/completions.py
jamie-rasmussen: It's from the OpenAI SDK

weave/trace/weave_client.py
tssweeney: Maybe an easy compromise is that we change the name from `models` to something like `llm_models` (up to you on the name). Then once we have all 3:
1. CoreWeave Models
2. Configured Models (similar to the dropdown available in UI)
3. Playground Models (those saved with pre-canned settings)

We could consider duplicating the symbol at the more general `models` location as it would more completely describe the set of possible models?

tssweeney: the service layer has a way to query for this - it has not been hoisted up to the `WeaveClient` level

weave/chat/common.py
jamie-rasmussen: I missed that it was defined in the weave SDK and not the Trace Server, I'll reuse.

weave/trace/weave_client.py
jamie-rasmussen: I changed it to `inference_models` for now.

weave/chat/models.py
jamie-rasmussen: Made it more compatible

PR #4915, opened 2025-06-26
chore(weave): Add F811 lint rule

andrewtruong:
https://docs.astral.sh/ruff/rules/redefined-while-unused/


tests/integrations/openai/openai_test.py
```diff
@@ -10,54 +10,6 @@
 model = "gpt-4o"
 
 
-@pytest.mark.skip_clickhouse_client  # TODO:VCR recording does not seem to allow us to make requests to the clickhouse db in non-recording mode
```
andrewtruong, line 13: This test is duplicated below

PR #4917, opened 2025-06-26
chore(weave): Add F822 lint rule

andrewtruong:
https://docs.astral.sh/ruff/rules/undefined-export/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -255,6 +255,7 @@ select = [
   "F6",      # https://docs.astral.sh/ruff/rules/multi-value-repeated-key-literal/
   "F811",    # https://docs.astral.sh/ruff/rules/redefined-while-unused/
   "F821",    # https://docs.astral.sh/ruff/rules/undefined-name/
+  "F822",    # https://docs.astral.sh/ruff/rules/undefined-export/
   "TID252",  # https://docs.astral.sh/ruff/rules/relative-imports/#relative-imports-tid252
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -252,6 +252,11 @@ select = [
   "I",       # https://docs.astral.sh/ruff/rules/#isort-i
   "W",       # https://docs.astral.sh/ruff/rules/#warning-w
   "F401",    # https://docs.astral.sh/ruff/rules/unused-import/
+  "F402",    # https://docs.astral.sh/ruff/rules/import-shadowed-by-loop-var/
+  "F6",      # https://docs.astral.sh/ruff/rules/multi-value-repeated-key-literal/
+  "F811",    # https://docs.astral.sh/ruff/rules/redefined-while-unused/
+  "F821",    # https://docs.astral.sh/ruff/rules/undefined-name/
+  "F822",    # https://docs.astral.sh/ruff/rules/undefined-export/
```
tssweeney, line 258: Seems like this does more than F822?

zbirenbaum: I checked locally and confirmed none of the removed symbols are importable currently. LSP also gives a warning about them not being present.

I'm guessing that the file these were initially defined in got modified, then maybe something auto-fixed the undefined imports but didn't pick up the warning that the symbols were being exported?

Little confused how this happened in the first place but this looks good.

PR #4932, opened 2025-06-27
chore(weave): Adds `watch-code-check-cache` as yarn script

tssweeney:
Now we can run `yarn watch-code-check-cache` in the `weave-js` root to have a cached, live lint-fixer.

Normally I dev a bunch, then run linter before committing. It takes forever, typically takes a few runs, etc... Super slow. Now I just run this in a shell and everything live fixes itself super quickly.


zbirenbaum (approved):
Can't wait to try this out!

PR #4933, opened 2025-06-27
feat(weave): LangChain ChatView

zbirenbaum:
## Description

- Fixes WB-25805

Enables ChatView for langchain when used with Anthropic, Gemini, and OpenAI providers. Also adds error handling which hides the chat view if an error is encountered while processing a detected format.

## Testing

Tested with various scripts and forced an error to be thrown which resulted in the chat view being hidden.

image
image

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
@@ -21,7 +21,7 @@ import {
   normalizeGeminiChatRequest,
 } from './ChatFormats/gemini';
 import {
-    isChatRequestFormatLangchain,
+  isChatRequestFormatLangchain,
   isTraceCallChatFormatLangchain,
   normalizeLangchainChatCompletion,
   normalizeLangchainChatRequest,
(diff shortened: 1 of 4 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/ChatView/hooks.ts
```diff
@@ -150,9 +194,16 @@ export const normalizeChatRequest = (request: any): ChatRequest => {
   if (isAnthropicCompletionFormat(request)) {
     return normalizeAnthropicChatRequest(request);
   }
-  if (isTraceCallChatFormatOTEL(request)) {
```
zbirenbaum, line 153: This was an unrelated bug that was causing issues with the error handling

zbirenbaum:
> Looks good! I will say it was not clear to me what the behavior change was from the PR. I would summarize the current behavior as we only enable the chat view for autopatched calls to an llm api, which are always leaves of the trace tree. This PR introduces a change so that langchain wrappers of those leaves also show the chat view. It occurs to me that there may be other cases where we have an op which is simply wrapping calls to a provider where we could potentially do some similar stiching together so the user can see a complete chat view even when the chat is broken across multiple calls.

This seems pretty correct

The changes are as follows:
1. Langchain wraps multiple providers
2. A small subset of the provider schema is used in the nested fields, but not enough to recursively resolve it
3. ChatView is totally disabled when the normalization would break (we never encountered this previously, but can encounter it with unsupported langchain providers)

The stitching together would be complicated and I think other things take priority right now, but I've been thinking about this. For the moment this is stable and works so I'll merge

PR #4937, opened 2025-06-27
fix(weave): fix scroll not reaching bottom of code

zbirenbaum:
## Description

- Fixes WB-25598

Added a parent container for the code viewer with some styling for overflow handling

## Testing
Bug was reproduced: In the view accessible by clicking on the op+version from the summary tab the code tab is opened by default and does not reach the bottom.
After this change, even when first opened the code will scroll to the bottom

image


neutralino1 (approved):
Thank you.

PR #4941, opened 2025-06-28
fix(weave): Make scorer column mapping work with scorers that use kwargs

andrewtruong:
Previously we did not skip 'kwargs' which would cause column mappings to fail

### DIFF
--- a/weave/flow/scorer.py
+++ b/weave/flow/scorer.py
@@ -249,7 +249,7 @@ def preparer_scorer_op_args(
 
         # Build arguments dictionary using column mapping
         for arg in score_arg_names:
-            if arg == "output" or arg == "model_output":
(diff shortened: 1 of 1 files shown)


weave/flow/scorer.py
```diff
@@ -196,7 +196,7 @@ class ApplyScorerSuccess:
 ApplyScorerResult = ApplyScorerSuccess
 
 
-def preparer_scorer_op_args(
+def prepare_scorer_op_args(
```
andrewtruong, line 255: previously, this func was tres francais

andrewtruong:
The actual fix was upstreamed in another PR.  This adds more tests

neutralino1 (approved):

PR #4945, opened 2025-06-30
fix(weave): Fetch correct scorer in monitor form

neutralino1:
Before this PR, the monitor form would always fetch the latest object named after the scorer object. Those names are not unique across object types.
In the case where the user names the scorer the same as the monitor, the fetched scorer would actually be the monitor object, resulting in the form not showing the scorer sub-form. Then upon saving, the scorer is removed from the monitor.

This PR uses the exact scorer version key to fetch it from the server.


gtarpenning (approved):
lgtm

PR #4947, opened 2025-07-01
RAU Track: chore(weave): Non User Facing - Adds `run_model` API to the TraceServerInterface

tssweeney:
## Add `run_model` Server API with Process-Isolated Execution

Most simply, this PR adds `run_model` to the server API, allowing a caller to directly execute a model (limited to the `LLMStructuredCompletionModel` class) by calling the server. Note: this is not yet exposed as an HTTP API, so technically it is not user facing yet.

This allows for calls like:
```python
await trace_server.run_model(
    RunModelReq.model_validate(
        {
            "project_id": project_id,
            "model_ref": model_ref_uri,
            "inputs": {"user_input": "Hello!"},
        }
    )
)
```

### Why This Matters

This PR introduces the foundation for server-side execution of user code, starting with model inference. The key innovation is that we can now safely run user models on the server while maintaining complete isolation between different users and projects.

---

However, this is facilitated by a fairly involved execution pattern:

1. **The `ClickhouseTraceServer` can now call methods on a new class called `RunAsUser`**. For example:
```python
await RunAsUser(
    internal_trace_server=self,
    project_id=req.project_id,
    wb_user_id=req.wb_user_id,
).run_model(req)
```

2. **What this will do is setup a new child process with an auto-initialized `WeaveClient`**. This special Weave client can communicate with the parent (current) `ClickhouseTraceServer` via multiprocessing queues, intermediated by some thin layers that:
   - Solve the id-swapping/ref issue (converting internal refs to external format)
   - Enforce that the client code can only operate on the original project & user
   - Prevent any cross-contamination of object references between users

3. **The real magic is in the process boundary**. Because WeaveClient mutates local objects when it attaches refs to them, having multiple authenticated clients in the same process would be unsafe. By using separate processes, we guarantee complete memory isolation.

4. **The benefit of this infrastructure is that we can author scripts that look exactly like normal Weave scripts** (eg. `weave/trace_server/execution_runner/user_scripts/run_model.py`). These scripts:
   - Can be safely run on the server, on behalf of the calling user
   - Log equivalent data as if it was run by the client itself
   - Keep all ref-mutated objects hidden behind the process boundary
   - Have zero chance of cross-project / cross-user ref mixing

### Architecture Highlights

- **Three-layer security model**: Process isolation → Project validation → User injection
- **Queue-based communication**: Parent and child communicate only through multiprocessing queues
- **Transparent API**: User scripts use standard WeaveClient APIs without modification
- **60-second timeout**: Protects against runaway processes

### What's Next

Ultimately, this mechanic will be used to execute scorers as well (and evals). However, evals will be executed by a backend worker, allowing for long-running evaluation jobs without blocking the main server.

This PR lays the groundwork for a new class of server-side features where we can safely execute user-provided logic in a controlled environment, opening up possibilities for automated scoring, server-side inference, and more sophisticated evaluation pipelines.

In the future, I would like to setup 1 runner per user/project pair that is held in memory in an LRU on the server, so that we don't pay the process initialization cost for every request when there are many requests. However, that is an activity for a future self. **Currently each request has about 2 second overhead for process initialization.**

### DIFF
--- a/tests/trace_server/execution_runner/test_run_as_user.py
+++ b/tests/trace_server/execution_runner/test_run_as_user.py
@@ -31,7 +31,7 @@ from tests.trace_server.execution_runner.test_functions import (
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.execution_runner.run_as_user import (
     RunAsUser,
-    RunAsUserException,
+    RunAsUserError,
 )
 
 
@@ -82,7 +82,7 @@ class TestRunAsUser:
 
         req = TestRequest(value="test_error")
         # The child process will crash when the exception is raised
-        with pytest.raises(RunAsUserException, match="exit code"):
+        with pytest.raises(RunAsUserError, match="exit code"):
             await runner._run_user_scoped_function(
                 failing_function,
                 req,
@@ -105,7 +105,7 @@ class TestRunAsUser:
         )
 
         req = TestRequest(value="timeout_test", sleep_time=2.0)
-        with pytest.raises(RunAsUserException, match="timed out after 0.5 seconds"):
+        with pytest.raises(RunAsUserError, match="timed out after 0.5 seconds"):
             await runner._run_user_scoped_function(
                 timeout_function,
                 req,
@@ -126,7 +126,7 @@ class TestRunAsUser:
         )
 
         req = TestRequest(value="exit_test", exit_code=42)
-        with pytest.raises(RunAsUserException, match="exit code: 42"):
+        with pytest.raises(RunAsUserError, match="exit code: 42"):
             await runner._run_user_scoped_function(
                 exit_code_function,
                 req,


.github/workflows/test.yaml
```diff
@@ -375,6 +375,7 @@ jobs:
           WF_CLICKHOUSE_HOST: localhost
           WEAVE_SERVER_DISABLE_ECOSYSTEM: 1
           DD_TRACE_ENABLED: false
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
gtarpenning: interesting, this feels like a big change, we are actually going to be hitting the API in CI?

neutralino1 (commented):
Impressive work, Tim!

I'm a little concerned with a few things in this design:

- **Overall complexity**: as you point out yourself in the PR description, this execution pattern is quite complex. This means brittle and difficult to maintain.
- **Arbitrarily long processes in the API server**: you set a timeout of 60 seconds, but that still means the server is held up from serving request for up to 60 seconds. This can become a performance hog over time. We currently have 32 server workers. This means if at least 32 of these queries come in, nobody else can use the API server.
- **Misalignment with scoring worker code**: The scoring worker also has to perform actions on behalf of users. It seems like we should unify these patterns and build a "server-side client" that is able to do this in a thread-safe manner without needing to be isolated in a separate process.

I wonder if a more asynchronous design would work better: for example, `run_model` requests could be `POST /v2/feedbacks` which would enqueue a bufstream message and a worker could consume them. Then we can batch more easily and not worry about holding up the API server for long periods of time. Then we can mutualize code and execution patterns across workers (scoring worker, `run_model` worker, and all other future workers).

On the client side (UI), we can poll for completion, or in the long term use websockets or such.

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+# Add any context variables here that need to be available in the worker thread
+# Example: If you have a custom context var, add it to this list:
+#   from mymodule import my_context_var
+#   CONTEXT_VARS_TO_PROPAGATE.append(my_context_var)
+CONTEXT_VARS_TO_PROPAGATE: list[contextvars.ContextVar] = [
+    _secret_fetcher_context,
+]
+
+# Default timeout for request/response cycles
+TIMEOUT_SECONDS = 60.0
```
gtarpenning: Definitely going to want to make this configurable. Maybe even on a per-user basis. I wonder how many bitcoin you can mine in 60 seconds 🤔

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+    A trace server implementation that delegates operations via queues.
+
+    This class acts as a proxy that sends requests through a request queue
+    and receives responses through a response queue. It doesn't know about
+    the actual trace server implementation - it just handles the
+    communication protocol.
+
+    The actual trace server runs in a separate thread (in the parent process)
+    and this class can be used from child processes to communicate with it.
+    """
```
gtarpenning: couple-line example instantiation would make this description more useful.

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+        """
+        Handle responses from the worker thread.
+
+        This method runs in a separate thread and continuously polls the
+        response queue for incoming responses, matching them with pending
+        requests.
+        """
+        while not self._shutdown:
+            try:
+                response = self._response_queue.get(timeout=0.1)
```
gtarpenning: very tight timeout, interesting. What happens under infra load? I'm thinking about the case when all workers are in use. Child threads are probably fine, and I see we are catching this error with the blanket exception... Maybe want to catch specifically, and logger.warn though.

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+
+    def _handle_responses(self) -> None:
+        """
+        Handle responses from the worker thread.
+
+        This method runs in a separate thread and continuously polls the
+        response queue for incoming responses, matching them with pending
+        requests.
+        """
+        while not self._shutdown:
```
gtarpenning: do we need a polling interval here? how frequent is this?

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+    Args:
+        trace_server: The trace server instance to run in the worker thread (stays in parent)
+
+    Returns:
+        A dictionary containing the queues to pass to the child process
+    """
+    # Get the multiprocessing context
+    ctx = multiprocessing.get_context()
+
+    # Create multiprocessing queues (these CAN be serialized)
```
gtarpenning: and are safe to write to between processes with impunity. no need to take a lock

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+
+
+def _trace_server_worker_loop_with_context(
+    request_queue: multiprocessing.Queue,
+    response_queue: multiprocessing.Queue,
+    trace_server: TraceServerInterface,
+    context_values: dict[contextvars.ContextVar, Any],
+) -> None:
+    """
+    Worker loop that handles requests using the provided trace server with context.
```
gtarpenning: this is non trivial, maybe a bit more description here....

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+        with self._lock:
+            response = self._responses.pop(request_id)
+            del self._pending_requests[request_id]
+
+        if response["error"]:
+            raise SendRequestException(f"Worker thread error: {response['error']}")
+
+        return response["result"]
+
+    def shutdown(self) -> None:
```
gtarpenning: does this get called automatically on cleanup? It's not immediately clear under what conditions the cross process trace server gets cleaned up. Like, when the `_run_user_scoped_function` finishes?

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+                response = ResponseWrapper(
+                    request_id=wrapped_request["request_id"], result=None, error=str(e)
+                )
+
+            # Send the response
+            response_queue.put(response)
+
+        except Exception as e:
+            # Critical error in worker loop
+            logger.exception(f"Critical error in worker loop: {e}")
```
gtarpenning: nit: the double try/except is a bit gnarly, suggest making a helper fn

weave/trace_server/execution_runner/run_as_user.py
```diff
@@ -0,0 +1,346 @@
...
+        # Check if process is still alive (timeout occurred)
+        if process.is_alive():
+            # Terminate the process since it exceeded timeout
+            process.terminate()
+            # Give it a moment to terminate gracefully
+            process.join(timeout=1)
+            if process.is_alive():
+                # Force kill if still running
+                process.kill()
+                process.join()
```
gtarpenning: do we need another timeout here? Is process.kill guaranteed to be successful?

weave/trace_server/execution_runner/README.md
```diff
@@ -0,0 +1,284 @@
...
+    class USC_DESC,CTS_DESC,REQ_DESC,ETS_DESC,CH_DESC descBox
+```
+
+### Multiprocessing Communication Flow
+
+This diagram shows how processes communicate during a `run_model` execution:
+
+```mermaid
+sequenceDiagram
+    participant MP as Main Process<br/>(ClickHouse Server)
```
gtarpenning, line 208: technically isn't there also a queue between the main process and the forked child? I realize its more of an implementation detail rather htan an architectural point, but the layers of queues gets sort of confusing to reason about.

weave/trace_server/execution_runner/trace_server_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+
+    Raises:
+        ValueError: If any reference contains a different project ID
+    """
+
+    def convert_project_id(internal_project_id: str) -> str:
+        if project_id != internal_project_id:
+            raise ValueError(
+                f"Project ID mismatch: {project_id} != {internal_project_id}. "
+                "This is a security issue."
```
gtarpenning: nit: more actionably error message, or omit

weave/trace_server/execution_runner/user_scripts/run_model.py
```diff
@@ -0,0 +1,89 @@
...
+    Raises:
+        TypeError: If the model reference doesn't point to a valid LLMStructuredCompletionModel
+                  or if the inputs are not a dictionary when resolved
+        ValueError: If the input type is neither 'value' nor 'ref'
+    """
+    # Get the scoped client from context (set up by parent process)
+    client = require_weave_client()
+
+    # Load the model from its reference
+    loaded_model = client.get(parse_uri(req.model_ref))
```
gtarpenning, line 54: can this be empty/none?

tssweeney:
> Impressive work, Tim!
> 
> I'm a little concerned with a few things in this design:
> 
> * **Overall complexity**: as you point out yourself in the PR description, this execution pattern is quite complex. This means brittle and difficult to maintain.
> * **Arbitrarily long processes in the API server**: you set a timeout of 60 seconds, but that still means the server is held up from serving request for up to 60 seconds. This can become a performance hog over time. We currently have 32 server workers. This means if at least 32 of these queries come in, nobody else can use the API server.
> * **Misalignment with scoring worker code**: The scoring worker also has to perform actions on behalf of users. It seems like we should unify these patterns and build a "server-side client" that is able to do this in a thread-safe manner without needing to be isolated in a separate process.
> 
> I wonder if a more asynchronous design would work better: for example, `run_model` requests could be `POST /v2/feedbacks` which would enqueue a bufstream message and a worker could consume them. Then we can batch more easily and not worry about holding up the API server for long periods of time. Then we can mutualize code and execution patterns across workers (scoring worker, `run_model` worker, and all other future workers).
> 
> On the client side (UI), we can poll for completion, or in the long term use websockets or such.

I agree with all these points. A few followup notes
1. Complexity: I agree here. The alternative is to completely rewrite the client to support multiple, simultaneous, differently-authed clients at the same time. This seems like a heavy lift that if done later on, would allow us to remove the isolation layers, but keep the other implementation code written here. I think the current implementation nicely separates the isolation business logic from the actual code we are running for the user. Do you have an alternative suggestion here?
2. Long Process: I am more than happy to reduce the timeout to something shorter like 5 seconds. This should be no worse than our current playground completion endpoints (as they are just making a single llm network call). Indeed, these are actually designed to be our first async methods on the server. This should allow us to leverage the benefits of FastAPI's async pattern, which would result in non-blocking calls, allowing for more requests at a time. The more async-y we make the user code, the better this gets as well. However, in principle, I agree - having long-running requests is not a good pattern. This brings me to the next bullet: Evals (Model run and score are on the order of LLM completion), but Evals are inherently much longer:
3. Misalignment with Scorer Code: I would like to adapt the pattern for Evals to be more aligned with Scorer code (in that we use the queuing system to handle the workload asynchronously). However, in contrast to the Scorer code today, I think it is beneficial to be able to author plain "user scripts" and have the underlying client handle the complexity (which is what this PR does). The client/server pair hides the complexity of id swapping/refs/etc... and allows the code to be executed normally. So it think in a way, I am doing what you suggest here?

gtarpenning (approved):
Very cool, two pr-wide comments, otherwise just asked a few questions and nits for my edification and for future readers. 

1. Will we expect all consumers of this method (in the future) to only interact with this endpoint using the async pattern? I sort of think we might want to have two versions of the client, one that is async and one that is sync, users can decide when they instantiate the client. Way out of scope for this pr, but just thinking about it because this is the first trace server endpoint that is fully async. 
2. The memory handling in this pr, while well documented and broken up to be quite readable given the complexity, could be delicate. We should watch the performance and memory impact on individual pods when this goes live. I also think it might be nice to include some memory profiling in this initial internal version, so we can get some insight into how resources are being utilized. I am especially interested in how fastapi interacts with the child process forking and subsequent cleanup. There are some pretty lightweight python libs, not exactly sure where the best place to wire them up would be but could be useful, if only to prove that this is airtight.

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: Good call, I'll make it tunable. Remember, this is not real user-code, so no mining could occur.

.github/workflows/test.yaml
tssweeney: it just needs any value to be there 🤦 ... we mock the response, but it still checks for key existence. (FWIW, we do this in the sqlite tests - which to your point, is bad form imo. I will replace with a dummy key)

andrewtruong: Agree with Griffin -- recommend putting an explicit dummy api key with comment

tests/trace_server/conftest.py
```diff
@@ -99,3 +111,14 @@ def trace_server(
...
+
+
+@pytest.fixture
+def ch_only_trace_server(
+    request, trace_server
+) -> TestOnlyUserInjectingExternalTraceServer:
+    trace_server_flag = get_trace_server_flag(request)
+    if trace_server_flag == "clickhouse":
+        return trace_server
+    else:
```
andrewtruong: nit: dont need else

tests/trace_server/execution_runner/test_run_as_user.py
```diff
@@ -0,0 +1,317 @@
...
+ensuring the system can handle stuck processes.
+"""
+
+import pytest
+
+from tests.trace_server.conftest_lib.trace_server_external_adapter import (
+    TestOnlyUserInjectingExternalTraceServer,
+)
+
+# Import test functions from separate module to ensure they're picklable
```
andrewtruong, line 20: Why are you doing this?

tests/trace_server/execution_runner/test_run_as_user.py
```diff
@@ -0,0 +1,317 @@
...
+    ExternalTraceServer,
+)
+
+
+def get_internal_trace_server(
+    trace_server: TestOnlyUserInjectingExternalTraceServer,
+) -> tsi.TraceServerInterface:
+    if isinstance(trace_server, ExternalTraceServer):
+        return trace_server._internal_trace_server
+    else:
```
andrewtruong: nit: don't need this else

weave/trace_server/execution_runner/run_as_user.py
```diff
@@ -0,0 +1,346 @@
...
+            RunAsUserException: If process execution fails or times out
+        """
+        if project_id != self.project_id:
+            raise ValueError(
+                f"Provided project_id '{project_id}' does not match instance value '{self.project_id}'"
+            )
+        if wb_user_id != self.wb_user_id:
+            raise ValueError(
+                f"Provided wb_user_id '{wb_user_id}' does not match instance value '{self.wb_user_id}'"
+            )
```
andrewtruong, line 269: Why do you need these checks?  You defined the ids in the init already, why do you need to explicitly pass the ids again and check they are the same here?

weave/trace_server/execution_runner/trace_server_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.trace_server_converter import universal_int_to_ext_ref_converter
+
+# Special placeholder used to indicate server-managed entities
+# This prevents clients from specifying arbitrary entities
+SERVER_SIDE_ENTITY_PLACEHOLDER = "__SERVER__"
+SERVER_SIDE_PROJECT_ID_PREFIX = SERVER_SIDE_ENTITY_PLACEHOLDER + "/"
+
+
+def externalize_trace_server(
```
andrewtruong: nit: prefer these "optional wrap" functions to start with `with`, which makes it easier to skim what's going on, e.g.

```py
server = with_feature1(with_feature2(...))
```

tests/trace_server/execution_runner/test_run_as_user.py
```diff
@@ -0,0 +1,317 @@
...
+            req,
+            "test_project",
+            "test_user",
+            TestResponse,
+        )
+
+        assert result.result == "Success: test_value"
+        assert result.process_id is not None
+        # Verify it ran in a different process
+        import os
```
andrewtruong: move to top

tests/trace_server/test_evaluation_execution_api.py
```diff
@@ -0,0 +1,167 @@
...
+            ObjCreateReq.model_validate(
+                {
+                    "obj": {
+                        "project_id": project_id,
+                        "object_id": model_object_id,
+                        "val": llm_model_val,
+                        "builtin_object_class": "LLMStructuredCompletionModel",
+                    }
+                }
+            )
```
andrewtruong, line 54: why not just construct the object normally in pydantic?

tests/trace_server/test_evaluation_execution_api.py
```diff
@@ -0,0 +1,167 @@
...
+        run_mode_local: bool = False,
+    ) -> RunModelRes:
+        project_id = f"{entity}/{project}"
+        req = RunModelReq.model_validate(
+            {
+                "project_id": project_id,
+                "model_ref": model_ref_uri,
+                "inputs": {"user_input": user_input},
+            }
+        )
```
andrewtruong, line 81: same here

tests/trace_server/test_evaluation_execution_api.py
```diff
@@ -0,0 +1,167 @@
...
+
+        assert model_run_res.output == expected_output
+
+        calls_res = ch_only_trace_server.calls_query(
+            CallsQueryReq.model_validate(
+                {
+                    "project_id": project_id,
+                }
+            )
+        )
```
andrewtruong, line 113: same here (and below)

weave/trace/weave_client.py
```diff
@@ -1345,7 +1354,7 @@ def send_start_call() -> bool:
                     started_at=started_at,
                     parent_id=parent_id,
                     inputs=inputs_json,
-                    attributes=attributes_dict,
+                    attributes=attributes_dict.unwrap(),
```
andrewtruong, line 1357: Why do you need this?

weave/trace_server/execution_runner/cross_process_trace_server.py
```diff
@@ -0,0 +1,540 @@
...
+import contextvars
+import logging
+import multiprocessing
+import queue
+import threading
+from collections.abc import Iterator
+from typing import Any, Optional, TypedDict
+
+from weave.trace_server.secret_fetcher_context import _secret_fetcher_context
+from weave.trace_server.trace_server_interface import *
```
andrewtruong: please don't use star import outside of `__init__.py`.  We already have a convention here for convenience:

```py
```

tssweeney:
> Very cool, two pr-wide comments, otherwise just asked a few questions and nits for my edification and for future readers.
> 
> 1. Will we expect all consumers of this method (in the future) to only interact with this endpoint using the async pattern? I sort of think we might want to have two versions of the client, one that is async and one that is sync, users can decide when they instantiate the client. Way out of scope for this pr, but just thinking about it because this is the first trace server endpoint that is fully async.

- The fact that the `remote_http_trace_server` has the same api as the backend trace server implementations doesn't add a lot of value (beyond testing). In the beginning it was useful to have the same interface, but in reality, there is no real good reason that these need to be the same. I don't expect users to call this directly. However, if they do, then I agree, in principle, we should have a sync and async version of every endpoint.

> 2. The memory handling in this pr, while well documented and broken up to be quite readable given the complexity, could be delicate. We should watch the performance and memory impact on individual pods when this goes live. I also think it might be nice to include some memory profiling in this initial internal version, so we can get some insight into how resources are being utilized. I am especially interested in how fastapi interacts with the child process forking and subsequent cleanup. There are some pretty lightweight python libs, not exactly sure where the best place to wire them up would be but could be useful, if only to prove that this is airtight.

I'll make sure to add good logging before making anything live

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: added

weave/trace_server/execution_runner/run_as_user.py
```diff
@@ -0,0 +1,346 @@
...
+
+    # Execute the function within a user-scoped client context
+    with user_scoped_client(wrapper_context.project_id, safe_trace_server):
+        res = asyncio.run(func(req))
+
+        # Convert the Pydantic response to a dict and send back
+        wrapper_context.result_queue.put(res.model_dump())
+
+
+class RunAsUserException(Exception):
```
andrewtruong: nit: In python, prefer calling specific errors `Error`, so this would be `RunAsUserError`

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: This timeout should be fairly short as it controls the frequency of the loop. A longer timeout will result in a longer time to shutdown, but yes, less loops. I pushed to 0.5 I would not expect this portion of the system to be bad. The reason i have a catch all is b/c if not, then this could result in a broken child thread that doesn't process future requests.

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: the timeout is what controls the frequency

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: correct - the queue mechanism is designed to handle multiple process writing to the queue

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: You are correct, this was an oversight - i failed to call this at program completion. fixed

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: this is not a double try/except. maybe I am missing something?

weave/trace_server/execution_runner/run_as_user.py
tssweeney: added timeout for safety

weave/trace_server/execution_runner/trace_server_adapter.py
tssweeney: modified

weave/trace_server/execution_runner/cross_process_trace_server.py
gtarpenning: I guess I meant nested. Which it is no?

weave/trace_server/execution_runner/user_scripts/run_model.py
tssweeney: not allowed to be None

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: added

tests/trace_server/test_evaluation_execution_api.py
tssweeney: I got into the habit of doing this to avoid needing to specifically import any nested objects - I'll clean this up for readability

tssweeney: actually, i already don't like this. 
```
 ObjCreateReq.model_validate(
                {
                    "obj": {
                        "project_id": project_id,
                        "object_id": model_object_id,
                        "val": llm_model_val,
                        "builtin_object_class": "LLMStructuredCompletionModel",
                    }
                }
            )
        )
```

this requires importing `ObjSchemaForInsert` for obj value. I like treating this like a pure dictionary

weave/trace/weave_client.py
tssweeney: ugg - reconstruction of the attributes over the queue results in the `weave` key throwing an error b/c it thinks it is getting edited.

tests/trace_server/conftest.py
tssweeney: fixed

tests/trace_server/execution_runner/test_run_as_user.py
tssweeney: it simulates the import pattern for the actual implementation. If a local function references a local variable that is not picklable, then the process executor fails

tests/trace_server/execution_runner/test_run_as_user.py
tssweeney: 👍

weave/trace_server/execution_runner/run_as_user.py
tssweeney: this is more of a safety check that the caller does not accidentally construct a runner for one project/user pair and then execute it for a different request

weave/trace_server/execution_runner/trace_server_adapter.py
tssweeney: changed

tests/trace_server/execution_runner/test_run_as_user.py
tssweeney: done

tests/trace_server/test_evaluation_execution_api.py
tssweeney: I want to test the interface layer - importantly: without python client code. As close to service-level testing as possible

tests/trace_server/test_evaluation_execution_api.py
tssweeney: this makes it closer to server-level testing

tests/trace_server/test_evaluation_execution_api.py
tssweeney: see above

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: fixed

weave/trace_server/execution_runner/run_as_user.py
tssweeney: done

weave/trace_server/execution_runner/cross_process_trace_server.py
tssweeney: this should be fixed now

weave/trace_server/execution_runner/README.md
tssweeney: added

PR #4952, opened 2025-07-01
fix(weave): contains operator work with json arrays

gtarpenning:
## Description

[WB-25645](https://wandb.atlassian.net/browse/WB-25645)

Now the filter bar supports "contains" operations on json keys to complex values like lists. 

We do this by adding a setting to the calls query, with the `SETTINGS` block appendix. 

## Testing

Prod
Screenshot 2025-07-01 at 1 02 16 PM

Branch
Screenshot 2025-07-01 at 1 02 41 PM

### DIFF
--- a/weave/trace_server/calls_query_builder/optimization_builder.py
+++ b/weave/trace_server/calls_query_builder/optimization_builder.py
@@ -423,12 +423,7 @@ def _create_like_optimized_eq_condition(
         # Empty string is not a valid value for LIKE optimization
         return None
 
-    # Boolean literals are not wrapped in quotes in JSON payloads
-    if literal_value in ("true", "false"):
-        like_pattern = f"%{literal_value}%"
-    else:
-        like_pattern = f'%"{literal_value}"%'
-
+    like_pattern = f"%{literal_value}%"
     like_condition = _create_like_condition(field, like_pattern, pb, table_alias)
     if _field_requires_null_check(field):
         return f"({like_condition} OR {table_alias}.{field} IS NULL)"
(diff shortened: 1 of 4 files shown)


weave/trace_server/calls_query_builder/optimization_builder.py
```diff
@@ -423,12 +423,7 @@ def _create_like_optimized_eq_condition(
         # Empty string is not a valid value for LIKE optimization
         return None
 
-    # Boolean literals are not wrapped in quotes in JSON payloads
-    if literal_value in ("true", "false"):
-        like_pattern = f"%{literal_value}%"
-    else:
-        like_pattern = f'%"{literal_value}"%'
```
neutralino1: This was added to support filtering on booleans. No longer necessary?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -479,6 +479,8 @@ def calls_query_stream(self, req: tsi.CallsQueryReq) -> Iterator[tsi.CallSchema]
             pb.get_params(),
         )
 
+        print("parameters", pb.get_params())
```
neutralino1: 🔪

weave/trace_server/calls_query_builder/optimization_builder.py
gtarpenning: previously we had quotes, which is an extra optimization. now we have no quotes, which should work for both normal strings and bools. slightly less optimized but whatever

neutralino1 (approved):
Thank you.

weave/trace_server/calls_query_builder/optimization_builder.py
gtarpenning: i made this pr way better

PR #4953, opened 2025-07-01
chore(weave): Enables direct testing of trace server

tssweeney:
**Summary:** This PR introduces a new fixture called `trace_server` which allows tests to specifically test the trace server. Over time, we should split the client tests from the trace server tests so we can isolate the code. In the meantime, this allows for more focused testing.

No more `client.server...`. Just `trace_server` and test!

**Background:**
Currently, we pretty much author most server tests as an end-to-end test where we rely on constructing a tracing client, then write and read with higher level APIs. Occasionally we do something like `client.server` to reach into the server and directly hit the server, but this practice is not great because the different forms a server can take makes for invalid typing and nasty checks. Furthermore, this pattern means our backend tests are coupled to our client implementation - making the eventual code split harder and harder with each test. 

This PR introduces the following technical changes:
1. Changes the `weave-server` flag to `trace-server` for alignment with code semantics
2. In CI, splits the tests into 3 segments, ordered as follows:
   * Non Trace Server (all tests that do not depend on a running Trace Server)
   * Clickhouse Trace Server (tests that depend on Trace Server, with Clickhouse backend)
   * Sqlite Trace Server (tests that depend on Trace Server, with Sqlite backend)
3. Made Clickhouse backend the default for all tests
4. Moved Trace Server specific fixtures out of `tests/conftest.py` in to `tests/trace_server/conftest.py`. Simplified caller code. Similarly moved `tests/conftest_lib/*` to `tests/trace_server/conftest_lib/*`.
5. Inside `tests/trace_server/conftest.py`, improved the fixture architecture in order to have cleaner separation of concerns. Now, we can simply use `trace_server` fixture for tests against the trace server, while still using `client` for tests against the end to end client.
6. Added a simple test of the Ci infra in `tests/trace_server/test_trace_server_conftest.py`


tests/trace_server/conftest_lib/clickhouse_server.py
```diff
@@ -103,7 +103,7 @@ def ensure_clickhouse_db_process_running(
 
     The fixture handles cleanup by stopping the process when the test session ends.
     """
-    server_up = check_server_up(host, port, 0)
+    server_up = check_server_up(host, port, 1)
```
gtarpenning, line 106: why this change?

tssweeney: 0 was a bug - 0 doesn't wait at all, 1 tries once

gtarpenning (approved):
Might be nice to have an example test of using the new flow? or refactoring an existing test?

tssweeney:
> Might be nice to have an example test of using the new flow? or refactoring an existing test?

Coming up in the next pr (:

PR #4409, opened 2025-05-06
chore(ui): Instrument onboarding spaces

m-rgba:
## Description

Added and tested telemetry events for the following:

- View empty states: trace and eval view
- Link clicks to documentation
- Link clicks to colab onboarding notebooks

PR #4954, opened 2025-07-01
chore(weave): Adding telemetry to monitors forms

neutralino1:
Adding three new Segment events to track online monitoring usage.


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/MonitorsPage/MonitorsPage.tsx
```diff
@@ -215,13 +216,18 @@ const CallCountCell = ({
...
+  const {entity, project} = useEntityProject();
+
   return (
     <Tailwind>
       <div className="mr-16 flex gap-8">
         <Button
           icon="add-new"
           variant="ghost"
-          onClick={onCreateMonitor}
+          onClick={() => {
```
zbirenbaum: This works for now but I think an even nicer way to do this might be to have a component Called TelemetryButton or something which takes the monitor info as props. That way we can just update all our buttons easily to integrate telemetry.

Also is the telemetry sdk something you have been working on or pre-existing

neutralino1: Hehe that's funny, I made the same comment on @m-rgba 's PR here: https://github.com/wandb/weave/pull/4409/files#r2178228330

zbirenbaum: Haha, probably a sign to make a ticket then

neutralino1: Or to implement it. Just added `TrackedButton` to the PR.

PR #4967, opened 2025-07-02
chore(ui): add alternate get example for models and other objects

jamie-rasmussen:
## Description

Same change as in https://github.com/wandb/weave/pull/4944 but for models and other Weave objects. Adds an example to the Use tabs showing how to access with a name and version index instead of complete URI.

Screenshot 2025-07-02 at 2 36 59 PM

## Testing

Checked copied code samples in ipython


neutralino1 (approved):
Should we only display the short version?

jamie-rasmussen:
> Should we only display the short version?

Both versions are useful - the full URI works without a separate weave.init

PR #4976, opened 2025-07-03
chore(weave): Parallelize trace tests

andrewtruong:
Parallelize trace tests by splitting them into shards.  Speeds up tests from ~17 min to ~6.5min (62%)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -260,7 +260,10 @@ jobs:
             #
           ]
         nox-shard: [
-            "trace",
+            "trace1",
+            "trace2",
+            "trace3",
+            "trace4",
             "flow",
             "trace_server",
             "trace_server_bindings",
--- a/noxfile.py
+++ b/noxfile.py
@@ -16,6 +16,7 @@ PY313_INCOMPATIBLE_SHARDS = [
     "crewai",
 ]
 PY39_INCOMPATIBLE_SHARDS = ["crewai", "google_genai", "mcp", "smolagents", "dspy"]
+NUM_TRACE_SHARDS = 4
 
 
 @nox.session
@@ -35,6 +36,9 @@ def lint(session):
         session.run("pre-commit", "run", "--hook-stage=pre-push", "--all-files")
 
 
+trace_shards = [f"trace{i}" for i in range(1, NUM_TRACE_SHARDS + 1)]
+
+
 @nox.session(python=SUPPORTED_PYTHON_VERSIONS)
 @nox.parametrize(
     "shard",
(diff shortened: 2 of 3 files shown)


noxfile.py
```diff
@@ -16,6 +16,7 @@
     "crewai",
 ]
 PY39_INCOMPATIBLE_SHARDS = ["crewai", "google_genai", "mcp", "smolagents", "dspy"]
+NUM_TRACE_SHARDS = 4
```
zbirenbaum, line 19: Why 4 shards?

noxfile.py
```diff
@@ -122,12 +127,13 @@ def tests(session, shard):
...
     test_dirs_dict = {
         "custom": [],
-        "trace": ["trace/"],
         "flow": ["flow/"],
         "trace_server": ["trace_server/"],
         "trace_server_bindings": ["trace_server_bindings"],
         "mistral": ["integrations/mistral/"],
         "scorers": ["scorers/"],
+        "trace": ["trace/"],
+        **{shard: ["trace/"] for shard in trace_shards},
```
zbirenbaum, line 136: I know based on your metrics this isn't the case, but at first glance this looks like it would run the full trace suite for each shard. Could you add a comment clarifying what this is doing?

.github/workflows/test.yaml
```diff
@@ -260,7 +260,10 @@ jobs:
             #
           ]
         nox-shard: [
-            "trace",
+            "trace1",
```
zbirenbaum, line 263: Can we add 'shard' to the name here to be more descriptive? like trace-shard-1 or trace_shard_1?

noxfile.py
```diff
@@ -136,13 +142,25 @@ def tests(session, shard):
...
+    pytest_args = [
         "pytest",
         "--durations=20",
         "--strict-markers",
         "--cov=weave",
         "--cov-report=html",
         "--cov-branch",
+    ]
+
+    # Handle trace sharding: run every 3rd test starting at different offsets
```
zbirenbaum, line 155: I'm not sure this comment is accurate. Shouldn't it be every `NUM_TRACE_SHARDS` tests? 

If there are 4 shards indexed from 0 every third would mean 
shard 1 runs: test0, test3
Shard2 runs: test1, test4
Shard3 runs: test2, test5
Shard4 runs: test3, test6

test3 is run twice

noxfile.py
```diff
@@ -136,13 +142,25 @@ def tests(session, shard):
...
+    if shard in trace_shards:
+        shard_id = int(shard[-1]) - 1
```
zbirenbaum, line 157: Why not just index the shard names from 0 so you don't have to subtract here and add in the list comprehension above?

PR #4979, opened 2025-07-03
chore(weave): Make numpy an optional dependency

andrewtruong:
Provides alternative non-numpy impls in cases that it's not available.  Removing numpy slims down package install by about 40mb


gtarpenning (approved):
lgtm

PR #4982, opened 2025-07-03
chore(weave): Make wandb an optional dependency

andrewtruong:
https://wandb.atlassian.net/browse/WB-25995

Provides alternative non-wandb impls in cases that it's not available.  Removing wandb slims down package install by about 140mb


pyproject.toml
```diff
@@ -127,7 +128,7 @@ langchain_nvidia_ai_endpoints = [
   "langchain-core>=0.2.1",
   "langchain-nvidia-ai-endpoints",
 ]
-litellm = ["litellm>=1.36.1"]
+litellm = ["litellm>=1.36.1,<=1.73.6"]
```
andrewtruong: pinning to resolve failing tests from recent litellm update

neutralino1: Can you add a ticket to fix tests so we can unpin?

neutralino1 (approved):
Thank you.

I wonder if there is a way to keep some of this shared code in sync. I don't really see one.

andrewtruong:
> I tried `pip install -e` in a clean venv and I'm getting
> 
> ```
> File ~/source/wandb/core/services/weave-python/weave-public/weave/trace/weave_client.py:2483, in safe_current_wb_run_id()
>    2480 try:
>    2481     import wandb
> -> 2483     wandb_run = wandb.run
>    2484     if wandb_run is None:
>    2485         return None
> 
> AttributeError: module 'wandb' has no attribute 'run'
> ```

Added a fix for this.  I think it has to do with an interaction with the existence of any dir called `wandb`.

tests/trace/test_client_trace.py
```diff
@@ -364,12 +363,8 @@ def liner(m: Number, b, x) -> Number:
 
     num_calls = 5
     run_calls = 0
-    if init_wandb:
```
jamie-rasmussen, line 367: It looks like we could now remove the `init_wandb` argument from `simple_line_call_bootstrap`

weave/trace/init_message.py
```diff
@@ -29,36 +37,47 @@ def _parse_version(version: str) -> packaging.version.Version:
...
 
+    wandb_messages = check_available(wandb.__version__, "wandb")
+    if not wandb_messages:
+        return
+
+    # Don't print the upgrade message, only the delete or yank message
+    use_message = wandb_messages.get("delete_message") or wandb_messages.get(
+        "yank_message"
+    )  #  or wandb_messages.get("upgrade_message")
+    if not use_message:
```
jamie-rasmussen: I don't quite understand why the negation changed here - seems to conflict with the comment

weave/trace/weave_client.py
```diff
@@ -2476,33 +2477,36 @@ def get_parallelism_settings() -> tuple[int | None, int | None]:
...
         import wandb
+    except (ImportError, ModuleNotFoundError):
+        return None
 
+    try:
         wandb_run = wandb.run
-        if wandb_run is None:
-            return None
-    except ImportError:
+    except AttributeError:
```
jamie-rasmussen, line 2493: Could use a comment explaining this condition

weave/wandb_thin/util.py
```diff
@@ -0,0 +1,25 @@
...
+def get_app_url() -> Optional[str]:
+    return os.getenv("WANDB_APP_URL")
+
+
+def app_url(api_url: str) -> str:
+    """Return the frontend app url without a trailing slash."""
+    # TODO: move me to settings
+    app_url = get_app_url()
+    if app_url is not None:
+        return str(app_url.strip("/"))
```
jamie-rasmussen: `str()` seems unnecessary here

weave/wandb_interface/wandb_api.py
```diff
@@ -455,6 +463,41 @@ def username(self) -> Optional[str]:
...
+        self,
+        project: str,
+        id: Optional[str] = None,
+        description: Optional[str] = None,
+        entity: Optional[str] = None,
+    ) -> dict[str, Any]:
+        """Create a new project.
+
+        Args:
+            project (str): The project to create
```
jamie-rasmussen, line 488: Missing `id` in list

weave/wandb_thin/termlog.py
```diff
@@ -0,0 +1,74 @@
...
+    newline: bool = True,
+    repeat: bool = True,
+    prefix: bool = True,
+) -> None:
+    """Log to standard error with formatting.
+
+    Arguments:
+        string (str, optional): The string to print
+        newline (bool, optional): Print a newline at the end of the string
+        repeat (bool, optional): If set to False only prints the string once per process
```
jamie-rasmussen, line 28: prefix is missing

weave/trace/weave_client.py
andrewtruong: added

weave/trace/init_message.py
andrewtruong: oops, thanks for catching

tests/trace/test_client_trace.py
andrewtruong: done

weave/wandb_thin/util.py
andrewtruong: You're right.  This is a copy from `wandb`.  I'll remove the `str`

weave/wandb_thin/termlog.py
andrewtruong: fixed

weave/wandb_interface/wandb_api.py
andrewtruong: this is not a required arg and now removed

jamie-rasmussen: Should it be removed from `UPSERT_PROJECT_MUTATION` then

andrewtruong: I think it could be.  I'm keeping it for now as it's a copy from the wandb lib

PR #4985, opened 2025-07-03
chore(ui): add error boundary around trace navigator

jamie-rasmussen:
## Description

Wrap the Trace navigator in an ErrorBoundary to avoid a whole page crash on error. (Hide whitespace)

## Testing

Tried with the issue in https://github.com/wandb/weave/pull/4983 before that fix.


gtarpenning (approved):
nice, there are probably other similar assumptions this will catch

PR #4986, opened 2025-07-03
feat(weave): Adding prompt support to TS SDK

neutralino1:
This PR adds support for persisting and fetching `StringPrompt` and `MessagePrompt` object to/from the trace server using the Typescript SDK.


sdks/node/src/weaveClient.ts
```diff
@@ -295,14 +296,49 @@ export class WeaveClient {
...
+      // Avoid circular dependency
+      const {Dataset} = await import('./dataset');
+
+      const {description, rows} = val;
+
       let obj = new Dataset({
-        id: _baseParameters.id,
-        description: _baseParameters.description,
+        id: dataObj.id,
+        description: description,
```
neutralino1: Serialization/deserialization should not rely on `_baseParameters`. This property is not available on objects serialized by the Python SDK.

sdks/node/src/prompt.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  format(values: Record<string, any> = {}): Record<string, any>[] {
+    return this.messages.map(message => {
+      return this.formatMessage(message, values);
+    });
+  }
+}
+
+function formatString(str: string, kwargs: Record<string, any> = {}): string {
```
chance-wnb, line 67: A lodash alternative:
```
_.template("this is {a}", {interpolate: /{([\s\S]+?)}/g})(
   {a: "1"}
)
```

I think it is more organic. But feel free to make a call on whether to adopt.

sdks/node/src/weaveClient.ts
chance-wnb: abbreviation available for "description ~~: description~~"

sdks/node/src/weaveObject.ts
```diff
@@ -57,25 +57,13 @@ export class WeaveObject {
...
-      }
-    }
-
-    // Then ops
-    for (const key of nonUnderscoreKeys) {
-      // @ts-ignore
-      const value: any = this[key];
-      if (isOp(value)) {
-        attrs[key] = value;
+      if (!key.startsWith('_')) {
```
chance-wnb, line 61: It seems the logic is not exactly the same after your refactoring:

The previous logic would take in the non-function members whose key is prefixed with an underscore. 

Are we now intentionally dropping those?

sdks/node/src/prompt.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export class MessagesPrompt extends Prompt {
+  messages: Record<string, any>[];
+
+  constructor(parameters: MessagesPromptParameters) {
+    super(parameters);
+    this.messages = parameters.messages;
+  }
+
+  formatMessage(
```
chance-wnb: Would you mind adding a unit test to demonstrate how a MessagesPrompt can be used by users? what would formatMessage() do?

chance-wnb (commented):
A few suggestions and a question.

sdks/node/src/prompt.ts
chance-wnb: Reference: https://lodash.com/docs/4.17.15#template

sdks/node/src/weaveObject.ts
neutralino1: Yes, I believe this was a roundabout way to let `_baseParameters` go through. I believe it should not. But I'm happy to revert this change.

sdks/node/src/prompt.ts
neutralino1: Discussed offline: `lodash` is not a current dependency of the SDK so we will stick with the manual implementation.

sdks/node/src/weaveObject.ts
chance-wnb: Cool. It looks like not having it is fine.

sdks/node/src/__tests__/live/prompt.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+describe('MessagesPrompt', () => {
+  test('should format a messages prompt', async () => {
+    const prompt = new MessagesPrompt({
+      messages: [{role: 'user', content: 'Hello, {name}!'}],
+    });
+    const formatted = prompt.format({name: 'John'});
+    expect(formatted).toEqual([{role: 'user', content: 'Hello, John!'}]);
+  });
+});
```
neutralino1, line 28: Added this test @chance-wnb

sdks/node/src/prompt.ts
neutralino1: Done.

chance-wnb: I see. so this is a private method. Maybe mark it as such.

chance-wnb (approved):
Thank you.

 I guess maybe we want to mark formatMessages as a private method.

PR #4988, opened 2025-07-04
chore(weave): Move emoji dependency to trace server

andrewtruong:
https://wandb.atlassian.net/browse/WB-25999

Removing emoji from the SDK slims down package install by about 8.5mb


pyproject.toml
```diff
@@ -76,6 +75,8 @@ trace_server = [
   "opentelemetry-proto>=1.12.0",
   "opentelemetry-semantic-conventions-ai>=0.4.3",
   "openinference-semantic-conventions>=0.1.17",
+  # For emoji shortcode support in Feedback
+  "emoji>=2.12.1",
```
neutralino1, line 79: Is this not needed on the SDK side to decode emoji feedbacks?

andrewtruong: afaict, no.  cc: @jamie-rasmussen

jamie-rasmussen: > afaict, no. cc: @jamie-rasmussen

If I remember correctly, the emoji library was needed for the sqlite backend.

andrewtruong: But it's not required for the client, right?  If users need the sqlite backend, they can install `emoji` separately

jamie-rasmussen: > But it's not required for the client, right? If users need the sqlite backend, they can install `emoji` separately

I think it should not be required for the client and I'm not opposed to removing the dependency, just wanted to flag that there could be some corner cases with running the tests against sqlite or people (maybe just Shawn?) using the sqlite backend.

PR #5000, opened 2025-07-07
chore(dev): add workflow to notify wandb/core when branches are deleted

blalor:
Creates a new workflow that notifies wandb/core when a branch is deleted. This is the counterpart to the existing `.github/workflows/notify-wandb-core.yaml` workflow.  The intent is to keep wandb/core from sprouting unbounded branches.

See also https://github.com/wandb/core/pull/31187.


neutralino1 (approved):

blalor:

PR #5007, opened 2025-07-08
perf(weave_query): optimize union type deduplication from O(n²) to O(n) for typedDicts

dannygoldstein:
Improves the performance of the `union()` function when deduplicating typedDict types from O(n²) to O(n) complexity.

The previous implementation used `_.uniqWith()` which compared every type against every other type, resulting in quadratic time complexity. This became a bottleneck when processing unions with many typedDict members. The new implementation uses a hash-based approach with a `getTypedDictKey()` function that creates canonical string representations of typedDict types for fast deduplication via a Set.

The optimization maintains the original behavior including order preservation of union members. Non-typedDict types continue to use the original comparison logic to ensure correct assignability checking. Also includes a fix for notRequiredKeys assignability to ensure that required fields in one type cannot be satisfied by optional fields in another type.

PR #5006, opened 2025-07-08
test(weave_query): add comprehensive unit tests for union type deduplication

dannygoldstein:
Adds comprehensive test coverage for the `union()` function in TypeScript to ensure correct behavior before performance optimizations.

The tests cover basic functionality including empty arrays, single members, and simple type unions. They also test typedDict deduplication with identical properties, different properties, complex nested types, and notRequiredKeys. Mixed type deduplication is tested with combinations of typedDicts and other types. Tagged value handling tests include flattening, merging by tag, and grouping. Edge cases like self-referential structures and order preservation are also covered.

Additionally includes tests for notRequiredKeys assignability logic in both TypeScript and Python to ensure type compatibility works correctly when optional fields are involved.

PR #5010, opened 2025-07-08
perf(weave_query): memoize typedDictPropertyTypes for performance

dannygoldstein:
This PR adds memoization to the `typedDictPropertyTypes` function using the memize library. This function is called frequently during type checking operations and computes complex union type properties. By caching results, we avoid redundant calculations for the same type inputs, improving frontend performance especially when dealing with complex nested types.


dannygoldstein:
* **#5021** <a href="https://app.graphite.dev/github/pr/wandb/weave/5021?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5010** <a href="https://app.graphite.dev/github/pr/wandb/weave/5010?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/5010?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5007** <a href="https://app.graphite.dev/github/pr/wandb/weave/5007?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5006** <a href="https://app.graphite.dev/github/pr/wandb/weave/5006?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave-js/src/core/model/helpers.ts
```diff
@@ -1245,62 +1246,64 @@ export const nullableTaggableValue = (type: Type): Type => {
...
-          if (prev[existingProp] != null) {
-            const prevProp = prev[existingProp];
-            if (
-              unwrappedCurr !== 'none' &&
-              unwrappedCurr.propertyTypes[existingProp]
-            ) {
-              const currProp = unwrappedCurr.propertyTypes[existingProp];
-              if (currProp) {
-                newVal[existingProp] = union([prevProp, currProp]);
+export const typedDictPropertyTypes = memize(
```
dannygoldstein, line 1249: this line is the only substantive  change in this pr

nicholaspun-wandb: Was doing a pass at how the package does equality checks between two incoming arguments and it [looks like](https://github.com/aduth/memize/blob/master/index.js#L70-L76) it's just `===`. Since `type` is an object, I think `===` only returns `true` if the memory locations match. I'm surprised that we're getting such a huge speedup since it implies that we're calling this function on the same object (and not just the same type) quite often 😅

dannygoldstein: yep exactly!

PR #5012, opened 2025-07-08
fix(weave): Align serialization payloads between Typescript and Python

neutralino1:
## Description

The current serialization payload for TS objects looks like this:

```
{
  ...
  val: {
    "_type": "<Type>",
    "_class_name": "<ClassName>",
    "_bases": [...],
    "_baseParameters": {
      "id": "<id>",
      "description": "<description>"
    }
  }
}
```

This is misaligned with payloads for Python objects. This PR changes it to match as such:

```
{
  ...
  val: {
    "_type": "<Type>",
    "_class_name": "<ClassName>",
    "_bases": [...],
    "name": "<name>",
    "description": "<description>"
  }
}
```

- Fixes WB-26068

## Testing

- unit tests

### DIFF
--- a/sdks/node/src/__tests__/live/prompt.test.ts
+++ b/sdks/node/src/__tests__/live/prompt.test.ts
@@ -66,23 +66,27 @@ describe('Prompt persistence', () => {
 
     const prompt = new StringPrompt({
       content: 'Hello, {name}!',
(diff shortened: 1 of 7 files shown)


andrewtruong (approved):
Thanks Emmanuel!

PR #5032, opened 2025-07-09
feat(weave): Document view for Langchain and ChromaDB

zbirenbaum:
## Description

- Fixes WB-25881

Adds document view with cards to display document contents and metadata to the trace view sidebar.

Introduces new schema parsing and validation logic which improves upon the way we handle other schemas such as chats. Eventually I'd like to use the new `ParseResult` type and Zod schemas/translation for Chat parsing as well. What is introduced in this PR is very ergonomic, simple to maintain and enables easy caching of parsed data for a trace.

Supports:
- Langchain Documents
- ChromaDB Documents

In order to adhere to the design for the cards a number of optional parameters have been added to the ObjectViewer

## Testing

Tested with scripts that input/return documents for both chromadb and langchain.

Preview:
https://www.loom.com/share/5469198c83ce483eb0324aa00617520d


weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/DocumentView.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }, [doc]);
+
+  return (
+    <Box
+      border={1}
+      borderColor={MOON_250}
+      borderRadius={'8px'}
+      paddingX={'20px'}
+      paddingY={'16px'}>
+      <div style={{display: 'flex', flexDirection: 'column', gap: 8}}>
```
neutralino1: Tailwind please.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/DocumentView.tsx
```diff
@@ -0,0 +1,102 @@
...
+        height: 'auto',
+      }}>
+      <div style={{display: 'flex', marginBottom: '4px'}}>
+        <Header>{title}</Header>
+      </div>
+      <div style={{display: 'flex', flexDirection: 'column', gap: 16}}>
+        {documents.map((doc, index) => (
+          <div
+            key={index}
+            style={{display: 'flex', flexDirection: 'column', gap: 16}}>
```
neutralino1: Tailwind.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
```diff
@@ -0,0 +1,242 @@
...
+      ) {
+        return; // Suppress this specific error
+      }
+      originalError.apply(console, args);
+    };
+
+    return () => {
+      console.error = originalError;
+    };
+  }, []);
```
neutralino1: This seems pretty bad lol
I don't have a better idea. Maybe @chance-wnb or @jamie-rasmussen have ideas?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
```diff
@@ -0,0 +1,242 @@
...
+const MetadataViewerSectionInner = ({
+  title,
+  data,
+  open,
+}: MetadataViewerSectionProps) => {
+  const apiRef = useGridApiRef();
+  // Update this when we change the state to hidden
+  // That way it restores the last state when uncollapsed
+  const [isOpen, setIsOpen] = useState(open ?? false);
+  const [mode, setMode] = useState('collapsed');
```
neutralino1: Should we use an enum for this?

neutralino1 (commented):
Nice work. Some comment.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
zbirenbaum: This was taken from the ObjectViewerSection code. If we change it here we should do it there as well for consistency.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/DocumentView.tsx
zbirenbaum: Fixed

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/DocumentView.tsx
zbirenbaum: Fixed

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
zbirenbaum: For context here is the full error:
```
hook.js:608 MUI X: useResizeContainer - The parent DOM element of the data grid has an empty height.
Please make sure that this element has an intrinsic height.
The grid displays with a height of 0px.
```

I've tried adding a small min height to everywhere from the collapse (while collapsed), to the containing divs. Unfortunately it seems that unless you do a conditional render (which makes the table jumpy and the dropdown less smooth) there's no way to fix this going off of posts I found while researching

chance-wnb: I could fix it by doing 

```
       <Collapse in={isOpen} sx={{
        '& .MuiCollapse-wrapperInner': {
          minHeight: '50px',
          paddingTop: '8px',
        }
      }}>
        {body}
      </Collapse>
```

feel free to give it a try.

BTW, there are console throws here and there every day. I wouldn't bother fixing it as long as it doesn't impact UX. patching console.error is not worth it IMHO :)

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
```diff
@@ -0,0 +1,242 @@
...
+        active={mode === 'json'}
+        onClick={() => setMode('json')}
+        tooltip="View as JSON"
+      />
+    </div>
+  );
+
+  const headerSection = (
+    <div>
+      <div style={{display: 'flex', minHeight: '32px'}}>
```
chance-wnb: I think you have an opportunity to use tailwind across this file.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
zbirenbaum: This seemed to work!! Thank you so much @chance-wnb! For future reference how did you figure this out? Would like to be able to do it myself next time.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
zbirenbaum: I updated to an enum in both spots. Resolving

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
zbirenbaum: Just updated it

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
chance-wnb: Great news! 

I was tracing the ancestor nodes of the Data Grid Pro, and see why they can't have a height. I saw that you added "height: 100%" everywhere, but by inspecting the DOM tree, I noticed that Collapse added 3 levels of divs without respecting nor inheriting the height settings of the parent. Then I removed `{body}`, and noticed that when there is no child, the parent container `.MuiCollapse-wrapperInner` indeed has a 0 height. Hence I deduced that solution.

At any rate, this isn't a simple problem. I also figured it out painstakingly.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/parser.ts
```diff
@@ -0,0 +1,91 @@
...
+import {
+  ParsedCall,
+  ParseResult,
+  SCHEMA_PARSERS,
+  TraceCallMinimalSchema,
+  WeaveDocument,
+} from './schemas';
+
+// Cached call parses
+const parsedCallCache = new Map<string, ParsedCall<WeaveDocument>>();
```
chance-wnb: I think this persists memory forever and would cause memory leak. Is there any way we can leverage WeakMap to get stale entries auto recycled?

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/MetadataViewerSection.tsx
zbirenbaum: ~After conversion I ran into an issue with the line:
`<div style="" class="tw-style min-h-32 flex">`~

~The tailwind min-h is not respected, leading to a height of 26 while collapsed and 32 when the buttons show up while expanded. I looked around and there are a number of threads on it, which just say to reinstall tailwind.~ 

I got it by using h-32 instead of min-h-32. Tailwind is a bit weird with heights.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/DocumentView/parser.ts
zbirenbaum: The keys would have to be the have to be the actual traceCall which the ID corresponds to. I'll have to think more about this. I would rather not use the full traceCall as the cache keys for two reasons

- Its unclear how persistent references to the trace call might be, this might be hard to test
- I really like that this parser only requires the ID, input, and output, which decouples it from the full trace call a bit.

I think the simpler solution for now would be to just remove the caching. This was mostly a bonus feature and we don't do it elsewhere for more intensive operations.

chance-wnb: Yes, I'd go forward with just removing the caching.

PR #5033, opened 2025-07-09
chore(weave): Add better error message for missing imports in scorers

andrewtruong:
https://wandb.atlassian.net/browse/WB-26086

image


tssweeney (approved):
This is a good start, but I challenge you to do better (: 

I think the `wandb` module has some cool patterns. it would be nice if there was not a scary stack trace like this.

Could we have a method `load_optional_module` or something that could allow developers to more cleanly load optional modules with clear instructions and controlled error output

andrewtruong:
I intentionally made the stack trace like this so users can easily tell what the problem is.  We can make the stack trace tighter by showing just the `pip install weave[scorers]` part.

image


RE: wandb-style, the main drawback with something like a `load_optional_module` is that it messes with the type checker and auto-complete.  I haven't been able to find a way around this yet.  If we can find a better way then I'm happy to switch to that

PR #5034, opened 2025-07-09
perf(weave): dont query for individual feedbacks

gtarpenning:
## Description

Requires that one line backend change. 

Don't query for feedback for every call, just use the feedback that we already attach to the calls query response! 

## Testing

prod
lots-of-feedback-prod

branch
lots-of-feedback-branch


gtarpenning:
already did this

PR #5038, opened 2025-07-10
chore(weave): Adds the ability for remote_http_trace_server to accept additional headers

tssweeney:
Prep work for remote evals:

Allows the auth and extra_headers properties to be set by the constructor. Small refactor to make this nicer in terms of calling

### DIFF
--- a/weave/trace_server_bindings/remote_http_trace_server.py
+++ b/weave/trace_server_bindings/remote_http_trace_server.py
@@ -78,6 +78,8 @@ class RemoteHTTPTraceServer(tsi.TraceServerInterface):
         should_batch: bool = False,
         *,
         remote_request_bytes_limit: int = REMOTE_REQUEST_BYTES_LIMIT,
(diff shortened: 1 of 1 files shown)


weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -268,8 +276,7 @@ def _generic_stream_request(
     @with_retry
     def server_info(self) -> ServerInfoRes:
         r = requests.get(
-            self.trace_server_url + "/server_info",
-            # timeout=DEFAULT_TIMEOUT,
+            "/server_info",
```
jamie-rasmussen, line 286: Did you mean to introduce a `self.get`?

tssweeney: nice catch - thank you - adding

jamie-rasmussen (approved):

PR #5040, opened 2025-07-10
chore(weave): Ensure that LLM-based evaluations can be serialized and deserialized correctly

tssweeney:
Our serialization/deserialization strategy is sorely in need of a re-write. I attempted it here, but it is a big change with wide-spread blast radius. Instead, I just made a very pointed hack to make sure we can deserialize things correctly. I learned a lot though, and am tempted to fix it once and for all soon.

### DIFF
--- a/weave/flow/eval.py
+++ b/weave/flow/eval.py
@@ -29,7 +29,7 @@ from weave.flow.scorer import (
 )
 from weave.flow.util import make_memorable_name, transpose
 from weave.trace.env import get_weave_parallelism
-from weave.trace.objectify import register_object
+from weave.trace.objectify import maybe_objectify, register_object
 from weave.trace.op import CallDisplayNameFunc, Op, OpCallError, as_op, is_op
 from weave.trace.vals import WeaveObject
 from weave.trace.weave_client import Call, get_ref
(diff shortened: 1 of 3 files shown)


weave/flow/eval.py
```diff
@@ -118,6 +118,26 @@ def from_obj(cls, obj: WeaveObject) -> Self:
             if hasattr(obj, field_name):
                 field_values[field_name] = getattr(obj, field_name)
 
+        # Start mega-hack
+        # This is a very bad hack. Our deserialization/portability logic
+        # if totally broken. It will require a complete re-write of our
+        # deserialziation layer to fix. In the meantime, this is such a
```
zbirenbaum: nit: `deserialziation` -> `deserialization`

weave/flow/eval.py
```diff
@@ -118,6 +118,26 @@ def from_obj(cls, obj: WeaveObject) -> Self:
...
+        # common pattern, that I am going to fix it here. If you are a
+        # future dev and you actually fix the serialization stuff, that
+        # may or may not break this. That is OK! please feel free to remove
+        # this as we have tests that validate the end-user experience.
+        if orig_scorers := field_values.get("scorers"):
```
zbirenbaum, line 132: It might be worthwhile to give a bit of info on exactly what is wrong with the serialization stuff for the future dev that decides to fix it 😉 .

It seems like this hack addresses the possibility that the scorers could be either deserialized class instances or still serialized WeaveObjects (probably due to differences between versions?) and we need to try and coerce them all to class instances.

tssweeney:
Thanks for the review @zbirenbaum! I've addressed both of your comments:

1. **Fixed the typo**: `deserialziation` → `deserialization`

2. **Added more context about the serialization issue**: Updated the comment to explain that the specific problem is that our deserialization code does not recursively deserialize custom objects (in this case scorers) and therefore needs to be done manually.

The mega-hack now has clearer documentation for future developers who might need to fix the broader serialization system.

PR #5048, opened 2025-07-10
chore(weave): Add minimal replacement for the `wandb` package in `weave`

andrewtruong:
https://wandb.atlassian.net/browse/WB-26097

This PR implements the `login` method from wandb to allow users to login using a W&B api key even if they don't have the `wandb` SDK installed


weave/wandb_thin/util.py
andrewtruong, line 1: This file is unused

andrewtruong:
Some code is copied, and some is a minimal reimplementation

weave/trace/weave_init.py
```diff
@@ -107,13 +109,17 @@ def init_weave(
...
             "Please login to Weights & Biases (https://wandb.ai/) to continue..."
         )
         wandb_termlog_patch.ensure_patched()
-        wandb.login(anonymous="never", force=True)  # type: ignore
+        try:
+            import wandb
+        except (ImportError, ModuleNotFoundError):
+            from weave import wandb_thin as wandb  # type: ignore[no-redef]
+        finally:
+            wandb.login(anonymous="never", force=True, host=host)  # type: ignore
```
tssweeney: a few things:
1. Let's not do this try/catch here. Let's create a module (`wandb_thin` maybe) that exposes methods which do the switching under the hood. Let's not put it on the caller to do this dance.
2. adding `host` here implies that we will not pick up the `WANDB_BASE_URL` env var. I think that is a miss. We should have the thin version behave as identically to the real version as possible.

tssweeney: I'd also like to see some tests for different hosts. there is a pretty complicated switch in `wandb` to handle the different host patterns

weave/wandb_interface/context.py
andrewtruong, line 1: This file contains all the context-specific parts of the now deleted `wandb_api.py`

weave/compat/wandb/__init__.py
andrewtruong, line 1: The thin wrapper should export the same symbols as regular `wandb`.

The only differences are `Api` and `ApiAsync` which just include our stripped down versions

bcsherma:
This is a bit odd for testing because `wandb` is still a required dependency (edit jk it's not but for some reason pip install -e . installed wandb again). So to try it out you need to install weave then manually `pip uninstall wandb`. When I do that and then try to log a trace, I see:
```
  File "/Users/ben/core/services/weave-python/weave-public/test.py", line 3, in <module>
    weave.init("bcanfieldsherman/test")
  File "/Users/ben/core/services/weave-python/weave-public/weave/trace/api.py", line 98, in init
    initialized_client = weave_init.init_weave(
                         ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ben/core/services/weave-python/weave-public/weave/trace/weave_init.py", line 120, in init_weave
    wandb_run_id = weave_client.safe_current_wb_run_id()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ben/core/services/weave-python/weave-public/weave/trace/weave_client.py", line 2311, in safe_current_wb_run_id
    wandb_run = _safe_get_wandb_run()
                ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ben/core/services/weave-python/weave-public/weave/trace/weave_client.py", line 2306, in _safe_get_wandb_run
    return wandb.run
           ^^^^^^^^^
AttributeError: module 'wandb' has no attribute 'run'
```

Edit: I removed the `wandb` directory from workdir and the error went away. I think if you check `wandb.__file__` after importing wandb you should be able to tell if we are importing a site package or doing a relative import.

PR #5050, opened 2025-07-10
chore(weave): add request id correlation

gtarpenning:
## Description

Add `retry_id` to client requests for correlation on retry attempts. 

## Testing

todo qa

### DIFF
--- a/weave/trace_server_bindings/remote_http_trace_server.py
+++ b/weave/trace_server_bindings/remote_http_trace_server.py
@@ -11,7 +11,7 @@ from weave.trace.settings import max_calls_queue_size, should_enable_disk_fallba
 from weave.trace_server import requests
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server_bindings.async_batch_processor import AsyncBatchProcessor
-from weave.utils.retry import _is_retryable_exception, with_retry
+from weave.utils.retry import _is_retryable_exception, get_current_retry_id, with_retry
 from weave.wandb_interface import project_creator
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -114,18 +114,27 @@ def set_auth(self, auth: tuple[str, str]) -> None:
         self._auth = auth
 
     def get(self, url: str, *args: Any, **kwargs: Any) -> requests.Response:
+        headers = dict(self._extra_headers) if self._extra_headers else {}
```
andrewtruong, line 118: shouldn't this be in the init?

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -114,18 +114,27 @@ def set_auth(self, auth: tuple[str, str]) -> None:
...
+
         return requests.get(
             self.trace_server_url + url,
             *args,
+            headers=headers,
             **kwargs,
         )
 
     def post(self, url: str, *args: Any, **kwargs: Any) -> requests.Response:
+        headers = dict(self._extra_headers) if self._extra_headers else {}
```
andrewtruong: same here

weave/utils/retry.py
```diff
@@ -33,11 +40,28 @@ def wrapper(*args: Any, **kwargs: Any) -> T:
             retry_error_callback=_log_failure,
             reraise=True,
         )
-        return retry(lambda: func(*args, **kwargs))
+
+        # Generate a retry ID for this request
+        retry_id = str(uuid.uuid4())
```
andrewtruong: for consistency I think this should be uuid7 which includes a timestamp component

weave/utils/retry.py
```diff
@@ -33,11 +40,28 @@ def wrapper(*args: Any, **kwargs: Any) -> T:
...
+
+        # Set the retry ID in context and execute the function
+        def run_with_retry_id() -> T:
+            _retry_id.set(retry_id)
+            return func(*args, **kwargs)
```
andrewtruong: we also need to unset, no?

weave/utils/retry.py
```diff
@@ -14,13 +16,18 @@
 
 T = TypeVar("T")
 
+# Context variable to store retry ID for correlation
+_retry_id: contextvars.ContextVar[str] = contextvars.ContextVar("retry_id")
```
andrewtruong, line 20: There's only 1 var here -- can only 1 thing be retried at a time per thread?

gtarpenning: yeah... but I think that each thread should always only be doing one thing right? 1 thread per conn. I think...

weave/trace_server_bindings/remote_http_trace_server.py
gtarpenning: I don't think it can be because the retry_id is dynamic right?

weave/utils/retry.py
gtarpenning: hmm, correct, i think it worked before but refactored to make this explicit. using a token + id now.

PR #5056, opened 2025-07-10
chore(weave): Introduce IsolatedClientExecutor for secure cross-process execution

tssweeney:
This PR introduces the `IsolatedClientExecutor ` utility that executes user functions in isolated processes with proper WeaveClient lifecycle management. This is essential for backend workers running evals safely. The implementation provides process isolation and error handling. This will also be used by trace-server level execution where we want to execute fast workloads as if we are the user. In order to execute code on a users behalf (either on the trace server directly or in a worker), we need to run it in an isolated manner. This is because the WeaveClient attaches refs to ops in a non-memory safe way.

Most simple example:

```python
def create_client():
    return WeaveClient(entity="my-entity", project="my-project")

runner = IsolatedClientExecutor(client_factory=create_client)
try:
    result = await runner.execute(my_function, request)
finally:
    runner.stop()
```


weave/trace_server/run_as_user/run_as_user.py
```diff
@@ -0,0 +1,358 @@
...
+                        elif signal == "EXEC":
+                            try:
+                                # Execute the function
+                                result = func(request)
+
+                                # Handle async functions - check for both Coroutine and Awaitable
+                                if isinstance(result, Coroutine):
+                                    result = asyncio.run(result)
+                                elif hasattr(result, "__await__"):
+                                    # Handle other awaitable types
```
gtarpenning: I feel like this quadruple try/except can be decomposed at least slightly for readability

weave/trace_server/run_as_user/run_as_user.py
```diff
@@ -0,0 +1,358 @@
...
+    ) -> None:
+        """
+        Main loop for the worker process.
+
+        Handles both synchronous and asynchronous function execution.
+        """
+        try:
+            # Create and initialize the client
+            if get_weave_client() is not None:
+                raise RunAsUserError("Unsafe to run as user with existing weave client")
```
gtarpenning: shouldn't this happen out of the try/except? this is programmer error mostly right?

weave/trace_server/run_as_user/run_as_user.py
```diff
@@ -0,0 +1,358 @@
...
+                            break
+                        elif signal == "EXEC":
+                            try:
+                                # Execute the function
+                                result = func(request)
+
+                                # Handle async functions - check for both Coroutine and Awaitable
+                                if isinstance(result, Coroutine):
+                                    result = asyncio.run(result)
+                                elif hasattr(result, "__await__"):
```
gtarpenning: ultra nit: should be one condition, andrew has a lint rule for something similar (nested conditions without elses), this feels basically the same.

tests/trace_server/run_as_user/cross_process_trace_server.py
```diff
@@ -0,0 +1,616 @@
...
+        """
+        while not self._stop_event.is_set():
+            try:
+                # Get request with timeout so we can check stop event periodically
+                try:
+                    request_item = self.request_queue.get(
+                        timeout=WORKER_LOOP_TIMEOUT_SECONDS
+                    )
+                except Empty:
+                    # Timeout, check stop event and continue
```
gtarpenning: weird comments here, also does the timeout actually happen in the get if empty?

tests/trace_server/run_as_user/cross_process_trace_server.py
```diff
@@ -0,0 +1,616 @@
...
+                # Check for stop signal
+                if request_item.request_id == STOP_SIGNAL:
+                    break
+
+                try:
+                    # Dispatch to the appropriate method on the trace server
+                    method_name = request_item.method
+                    method = getattr(self.trace_server, method_name)
+
+                    if method_name.endswith("_stream"):
```
gtarpenning: woah, we should make sure that we comment, or ideally enforce, this standard somewhere

tests/trace_server/run_as_user/cross_process_trace_server.py
```diff
@@ -0,0 +1,616 @@
...
+        1. Gets requests from the queue (with timeout to check stop event)
+        2. Dispatches to the appropriate trace server method
+        3. Handles both regular and streaming methods
+        4. Sends responses back via the response queue
+        5. Handles errors and continues processing
+
+        The loop terminates when a "STOP" signal is received.
+        """
+        while not self._stop_event.is_set():
+            try:
```
gtarpenning: these nested trys are pretty annoying to reason about, having a _worker_process helper that runs in the loop might make it cleaner tioli

tests/trace_server/run_as_user/cross_process_trace_server.py
```diff
@@ -0,0 +1,616 @@
...
+        response_queue: multiprocessing.Queue,
+        lock: multiprocessing.Lock,
+    ):
+        self.request_queue = request_queue
+        self.response_queue = response_queue
+        self._request_id_counter = 0
+        # Cache for responses that arrive before we're waiting for them
+        # TODO: Add periodic cleanup of old entries to prevent memory leaks
+        # TODO: Consider adding maximum cache size with LRU eviction
+        self._pending_requests: dict[str, ResponseQueueItem] = {}
```
gtarpenning: if we are worried about this, maybe there would be a convenient place to log to datadog the state of the queues? we can log or attach it to a span? might help catch errors before they blow up

weave/trace_server/run_as_user/run_as_user.py
tssweeney: sure - that is fair moving

weave/trace_server/run_as_user/run_as_user.py
tssweeney: fair - snagged

weave/trace_server/run_as_user/run_as_user.py
tssweeney: yeah, i'll make it cleaner

tests/trace_server/run_as_user/cross_process_trace_server.py
tssweeney: fwiw, this entire file is just for testing right now, but yes, if we want to use this in production, then I agree it would be good to make this more robust (as mentioned in the header).

tests/trace_server/run_as_user/cross_process_trace_server.py
tssweeney: fixed the comments. `get` will timeout if the queue is empty for timeout time - which we want to simply re-enter the loop

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
```diff
@@ -0,0 +1,601 @@
...
+        try:
+            yield
+        finally:
+            self._lock.release()
+
+    def _generate_request_id(self) -> str:
+        """Generate a unique request ID for tracking purposes."""
+        with self.lock():
+            self._request_id_counter += 1
+            return f"req_{self._request_id_counter}_{uuid.uuid4().hex[:8]}"
```
gtarpenning, line 132: i like that this includes the `_request_id_counter`

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
```diff
@@ -0,0 +1,601 @@
...
+        except Empty as e:
+            # Check if response was cached while we were waiting
+            with self.lock():
+                response_item = self._pending_requests.pop(request_id, None)
+            if response_item is None:
+                raise TimeoutError(f"Timeout waiting for response to {method}") from e
+
+        if response_item.request_id == request_id:
+            return response_item
+        else:
```
gtarpenning: style question: do you prefer if/else if both branches return? while this is easier to read from a conditional perspective, I find myself omitting the `else` to sort of mean "if this, return early" then continue execution

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
```diff
@@ -0,0 +1,601 @@
...
+            # Out-of-order response - cache it for later
+            with self.lock():
+                self._pending_requests[response_item.request_id] = response_item
+            # Recursively wait for the correct response
+            return self._wait_for_response(request_id, method, timeout)
+
+    def _send_request(self, method: str, payload: BaseModel) -> BaseModel:
```
gtarpenning, line 169: interesting typing here, why is it just `BaseModel`?

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
```diff
@@ -0,0 +1,601 @@
...
+        # This method has a different signature, so we need to create a payload
+        raise NotImplementedError("ensure_project_exists is not implemented")
+
+    # Regular method implementations (reduced duplication)
+    def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
+        """Export OTEL traces."""
+        return self._send_request("otel_export", req)
+
+    def call_start(self, req: tsi.CallStartReq) -> tsi.CallStartRes:
+        """Start a call."""
```
gtarpenning, line 284: it would be nice if we could like generate these interfaces, or dynamically create them based on the base interface class. way out of scope, just noting

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
```diff
@@ -0,0 +1,601 @@
...
+
+                # Check for stop signal
+                if request_item.request_id == STOP_SIGNAL:
+                    break
+
+                try:
+                    # Call the method and handle whatever it returns
+                    method_name = request_item.method
+                    method = getattr(self.trace_server, method_name)
+                    result = method(request_item.payload)
```
gtarpenning, line 526: this is where we actually hit the trace server, so this could be a blocking multi-second operation like a query? maybe worth a comment?

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
```diff
@@ -0,0 +1,601 @@
...
+                            error=STREAM_END_SIGNAL,
+                        )
+                        self.response_queue.put(response_item)
+                    else:
+                        # Handle regular response
+                        response_item = ResponseQueueItem(
+                            request_id=request_item.request_id,
+                            payload=result,
+                        )
+                        self.response_queue.put(response_item)
```
gtarpenning, line 552: again ulta nitty but I think this could be a helper: `_handle_trace_server_response`, just keeps the top level logical flow a bit more distinct.

gtarpenning: although the error handling might be harder... tioli

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
```diff
@@ -0,0 +1,601 @@
...
+        Returns:
+            CrossProcessTraceServerSender: Configured to send requests to this receiver
+        """
+        # Create a new lock for this sender instance through the manager
+        sender_lock = self.manager.Lock()
+        return CrossProcessTraceServerSender(
+            self.request_queue, self.response_queue, sender_lock
+        )
+
+    def stop(self) -> None:
```
gtarpenning, line 594: in the other stop process method we are more careful to handle cases where the process doesn't want to stop, and have contingency like killing etc, what's the difference? could the worker die and become unresponsive? maybe not, just curious if you thought about that

weave/utils/attributes_dict.py
```diff
@@ -63,3 +66,9 @@ def _set_weave_item(self, subkey: str, value: Any) -> None:
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({super().__repr__()})"
+
+    def unwrap(self) -> dict:
```
gtarpenning, line 70: i forget why we have to do this now?

weave/trace_server/isolated_client_executor.py
```diff
@@ -0,0 +1,453 @@
...
+                    result, error = _execute_function(func, request)
+                    response_queue.put((result, error))
+                else:
+                    raise ValueError(f"Unknown signal: {signal}")
+            except KeyboardInterrupt:
+                break
+            except Exception as e:
+                logger.error(f"Error in worker loop: {e}", exc_info=True)
+                response_queue.put((None, e))
+                break
```
gtarpenning, line 382: very clean

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
tssweeney: i suppose i do internally prefer the else branch, but can't construct a rational reason. modifying.

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
tssweeney: This is a serializable interface

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
tssweeney: i agree - the number of boilerplate items here is rough

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
tssweeney: noted

tests/trace_server/run_as_user/cross_process_trace_server.py
tssweeney: reworked completely

tests/trace_server/isolated_client_executor/cross_process_trace_server.py
tssweeney: I just didn't put as much time into this guy b/c it was just for testing

weave/utils/attributes_dict.py
tssweeney: the attributes dict does not like being serialized ....

PR #5059, opened 2025-07-11
chore(weave): Implement Evaluation status

tssweeney:
Implements the Evaluation status method for both backends


tests/trace_server/test_trace_server_evaluation_apis.py
```diff
@@ -0,0 +1,64 @@
...
+    eval_call_id = generate_id()
+
+    def get_status():
+        return client.server.evaluation_status(
+            EvaluationStatusReq(project_id=client._project_id(), call_id=eval_call_id)
+        ).status
+
+    @weave.op
+    def model(a: int) -> int:
+        assert get_status() == EvaluationStatusRunning(
```
jwlee64, line 26: I thought we switched this value from status to code? though I guess this succeeds cause it defaults the code value

PR #5063, opened 2025-07-11
feat(weave): Let Thread API to support filter by thread id

chance-wnb:
## Description

- Adds thread_id filtering capability to the threads_query API
- Implements thread_id filtering in both ClickHouse and SQLite query builders
- Updates the ThreadsQueryFilter model to include the new thread_id field
- Adds comprehensive tests for thread_id filtering, including combinations with other filters

This changed is needed for refactoring of the Thread details page in https://github.com/wandb/weave/pull/5064

## Testing

- Added unit tests for thread_id filtering in both ClickHouse and SQLite query builders
- Added integration tests that verify thread_id filtering works correctly in the API
- Tested combinations of thread_id with other filters (date ranges, sorting, pagination)

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -5062,33 +5062,47 @@ def test_threads_query_endpoint(client):
     # Should include our threads since they're before this future time
     assert len(before_filter_res) >= 3
 
-    # Test thread_id filtering
-    # Test filtering by specific thread_id
+    # Test thread_ids filtering
+    # Test filtering by specific thread_ids (single thread)
     for test_thread_id in thread_ids:
-        thread_id_filter_res = list(
+        thread_ids_filter_res = list(
             client.server.threads_query_stream(
                 tsi.ThreadsQueryReq(
                     project_id=get_client_project_id(client),
-                    filter=tsi.ThreadsQueryFilter(thread_id=test_thread_id),
+                    filter=tsi.ThreadsQueryFilter(thread_ids=[test_thread_id]),
                 )
             )
         )
         # Should find exactly one thread with the specified thread_id
-        assert len(thread_id_filter_res) == 1
-        assert thread_id_filter_res[0].thread_id == test_thread_id
+        assert len(thread_ids_filter_res) == 1
+        assert thread_ids_filter_res[0].thread_id == test_thread_id
 
-    # Test filtering by non-existent thread_id
+    # Test filtering by multiple thread_ids
+    multiple_thread_ids_filter_res = list(
+        client.server.threads_query_stream(
+            tsi.ThreadsQueryReq(
+                project_id=get_client_project_id(client),
+                filter=tsi.ThreadsQueryFilter(thread_ids=thread_ids[:2]),
+            )
+        )
+    )
+    # Should find exactly two threads
+    assert len(multiple_thread_ids_filter_res) == 2
+    found_thread_ids = {t.thread_id for t in multiple_thread_ids_filter_res}
+    assert found_thread_ids == set(thread_ids[:2])
+
+    # Test filtering by non-existent thread_ids
     nonexistent_filter_res = list(
         client.server.threads_query_stream(
             tsi.ThreadsQueryReq(
                 project_id=get_client_project_id(client),
-                filter=tsi.ThreadsQueryFilter(thread_id="nonexistent_thread_id"),
+                filter=tsi.ThreadsQueryFilter(thread_ids=["nonexistent_thread_id"]),
             )
         )
     )
     assert len(nonexistent_filter_res) == 0
 
-    # Test combining thread_id filter with other filters
+    # Test combining thread_ids filter with other filters
     combo_thread_filter_res = list(
         client.server.threads_query_stream(
             tsi.ThreadsQueryReq(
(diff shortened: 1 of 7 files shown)


chance-wnb:
* **#5064** <a href="https://app.graphite.dev/github/pr/wandb/weave/5064?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5063** <a href="https://app.graphite.dev/github/pr/wandb/weave/5063?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/5063?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/traceServerClientTypes.ts
```diff
@@ -553,6 +553,7 @@ export type ThreadSchema = {
 export type ThreadsQueryFilter = {
   after_datetime?: string;
   before_datetime?: string;
+  thread_id?: string;
```
tssweeney: since this is a query and not a direct read/lookup, i recommend making this a list of thread ids

tests/trace/test_client_trace.py
```diff
@@ -4988,6 +4988,51 @@ def multi_call_op(thread_name: str) -> list[str]:
...
+    # Test combining thread_id filter with other filters
+    combo_thread_filter_res = list(
+        client.server.threads_query_stream(
+            tsi.ThreadsQueryReq(
+                project_id=get_client_project_id(client),
+                limit=1,
+                sort_by=[tsi.SortBy(field="turn_count", direction="desc")],
+                filter=tsi.ThreadsQueryFilter(
+                    thread_id="analytics_thread",
+                    after_datetime=middle_time,
```
gtarpenning, line 5114: just curious, what's the reason to use `after_datetime` here instead of the mongo construction and getField `started_at` pattern?

chance-wnb: hmm, so far I haven't made the thread API to support the `query` parameter yet.

chance-wnb: Also I believe this parameter is supposed to operate on the `sortable_datetime` field.

weave/trace_server/threads_query_builder.py
```diff
@@ -108,7 +116,7 @@ def make_threads_query(
         WHERE project_id = {{{project_id_param}: String}}
             {sortable_datetime_filter_clause}
         GROUP BY id                         -- Group by call id to merge partial rows
```
gtarpenning: not strictly necessary but we usually do `GROUP BY (project_id, id)` as this is the primary key on the table. The query planner does the right thing in this case, but just fyi this now looks different than the main calls query

tests/trace_server/test_threads_query_builder.py
```diff
@@ -761,6 +761,360 @@ def test_turn_filtering_explanation():
...
+                CASE
+                    WHEN call_end_time IS NOT NULL AND call_start_time IS NOT NULL
+                    THEN dateDiff('millisecond', call_start_time, call_end_time)
+                    ELSE NULL
+                END as call_duration
+            FROM calls_merged
+            WHERE project_id = {pb_0: String}
+                AND sortable_datetime > {pb_1: String} AND sortable_datetime < {pb_2: String}
+            GROUP BY id                         -- Group by call id to merge partial rows
+            HAVING thread_id IS NOT NULL AND thread_id != '' AND id = any(turn_id) AND thread_id = {pb_3: String}  -- Filter to turn calls only
```
gtarpenning: why do we have `thread_id != ''` and `thread_id = {pb_3: String}`? Shouldn't we only have the latter?

gtarpenning: if this HAVING clause gets large, as in the number of results that we have to group by here is large, we may want to include a WHERE optimization that filter down the calls as much as possible before grouping. Might be worth making a note here

chance-wnb: Good call. will fix!

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/traceServerClientTypes.ts
chance-wnb: Yeah, that makes sense. I am using `thread_ids` (plural) now.

weave/trace_server/threads_query_builder.py
chance-wnb: Love this comment. Very insightful! 

I have changed to using the full primary key `(project_id, id)`. Thanks!

tests/trace_server/test_threads_query_builder.py
chance-wnb: Fixed by combining the filter condition when possible

chance-wnb: Yes, I just implemented the WHERE filtering as well.

chance-wnb:
PR is reworked to have the merge conflicts resolved.

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/wfReactInterface/traceServerClientTypes.ts
```diff
@@ -553,6 +553,7 @@ export type ThreadSchema = {
 export type ThreadsQueryFilter = {
   after_datetime?: string;
   before_datetime?: string;
+  thread_ids?: string[];
```
gtarpenning: this file doesn't exist anymore! this change needs to be made in a core pr

chance-wnb: Yes! Thanks for reminding!

PR #5072, opened 2025-07-11
chore(weave): Implement Evaluate Model Dispatcher

tssweeney:
This PR sets up the pattern for enabling the trace server to execute evaluation workloads. The patter developer here is:
1. `weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py` defines a `EvaluateModelDispatcher` abstract class as well as the actual `evaluate_model` implementation. The `EvaluateModelDispatcher` is an adapter that can be provided to either the `ClickhouseTraceServer` or the `SqliteTraceServer`. If present, then the `evaluate_model` method will "dispatch" the request to this class for processing. This is useful because it allow us to test the flow locally, but then provide a kafka-based dispatcher in prod. The `evaluate_model` is written with pure user-code for evaluating a model
2. A `EvaluateModelTestDispatcher` is used at test time to circumvent the need for complex kafka setup.
3. A test suite tests both the normal script as well as the end-to-end flow.

As implemented, there are no user-facing changes, but allows us to move onto the next step of configuring the procedure to work in a production setting


weave/trace/op.py
```diff
@@ -334,6 +336,7 @@ def _create_call(
         # Very important for `call_time_display_name` to take precedence over `func.call_display_name`
         display_name=call_time_display_name or func.call_display_name,
         attributes=attributes,
+        call_id=preferred_call_id,
```
gtarpenning: Are we now explicitly letting users to control this in the happy path? This feels dangerous / or at least sub optimal. obviously users can always fork the client and go crazy, but this opens the door for any users to just check the api spec and start logging calls with uuid4 for example.

tssweeney: Users already can already set the call ID from the API. This results in non uuid7 from otel. In this case, the param is used by the worker to force an id minted by the producing server. Would you prefer a different API that is even more hidden?

tests/trace_server/completions_util.py
```diff
@@ -0,0 +1,64 @@
+"""This module contains test utilities that are useful for mocking litellm completions responses (used by the playground and completion endpoints)"""
+
+from contextlib import contextmanager
```
zbirenbaum, line 3: Should we VCR this instead? Just a bit worried tests will still pass while prod is broken if litellm updates their schema.

tests/trace_server/test_trace_server_evaluation_apis.py
```diff
@@ -68,3 +79,308 @@ def generate_id_side_effect():
...
+
+    def create_dataset() -> str:
+        """Create a test dataset and return its reference URI."""
+        dataset_table_val = [
+            {"user_input": "How are you?", "expected": "I'm doing well, thank you!"},
+            # {"user_input": "What's 2+2?", "expected": "4"},
+            # {
+            #     "user_input": "Tell me a joke",
+            #     "expected": "Why did the chicken cross the road?",
+            # },
```
zbirenbaum: Remove or uncomment

tests/trace_server/test_trace_server_evaluation_apis.py
```diff
@@ -68,3 +79,308 @@ def generate_id_side_effect():
...
+
+    # evaluate
+    # predict_and_score
+    #    predict
+    #       complete
+    #    score
+    #       predict
+    #          complete
+    # summary
+    #    scorer summary
```
zbirenbaum, line 335: What is this comment?

weave/trace_server/workers/__init__.py
zbirenbaum, line 1: Is this necessary?

weave/trace_server/workers/evaluate_model_worker/__init__.py
zbirenbaum, line 1: Same here

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -0,0 +1,63 @@
...
+    LLMStructuredCompletionModel,
+)
+
+
+class EvaluateModelArgs(BaseModel):
+    project_id: str
+    evaluation_ref: str
+    model_ref: str
+    wb_user_id: str
+    evaluation_call_id: str
```
zbirenbaum, line 20: Is there a reason to differentiate `evaluation_call_id` from `call_id` when the same generated ID is shared at initialization above?

weave/trace_server/sqlite_trace_server.py
```diff
@@ -1531,9 +1540,21 @@ def threads_query_stream(
             )
 
     def evaluate_model(self, req: tsi.EvaluateModelReq) -> tsi.EvaluateModelRes:
-        raise NotImplementedError(
-            "evaluate_model is not implemented for SQLite trace server"
+        if self._evaluate_model_dispatcher is None:
+            raise ValueError("Evaluate model dispatcher is not set")
+        if req.wb_user_id is None:
+            raise ValueError("wb_user_id is required")
```
zbirenbaum, line 1549: Little confused on the wb_user_id check

Seems like the only reason it would be optional on EvaluateModelReq would be if we had automated dispatching which isn't attributed to a user, but maybe a project or team.

if that's the case though, why is is required here?

Could we just make this field non optional?

zbirenbaum (commented):
Couple of comments and questions

tests/trace_server/completions_util.py
tssweeney: Hmmm, I don't think we really care about the low level implementation of completions in this case. VCR could be used by the `tests/trace_server/test_llm_completion.py` file if we wanted to assert litellm behavior behind completions. Mocking is a bit clearer IMO in this case

weave/trace/op.py
gtarpenning: Sorry for belated reply, but yeah, having this top level makes it seem like we are encouraging users to supply their own ids. Maybe @andrewtruong has a different take though.

gtarpenning: honestly even just naming it something like `_call_id_override` would satisfy me

tests/trace_server/test_trace_server_evaluation_apis.py
tssweeney: done

tests/trace_server/test_trace_server_evaluation_apis.py
tssweeney: these are the 9 calls. i will comment

weave/trace/op.py
tssweeney: changed!

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
tssweeney: honestly just for clarity. this is the expected evaluation call id

weave/trace_server/workers/evaluate_model_worker/__init__.py
tssweeney: removed

weave/trace_server/workers/__init__.py
tssweeney: removed

weave/trace_server/sqlite_trace_server.py
tssweeney: this is just a stupid artifact of our interface design. all of the requests which have `wb_user_id` are optional because at the service layer we don't want them to be provided, but we do want them in the implementation (since they are injected by auth layer). you are right that this is odd, but this pattern is odd for a few cases and needs to be fixed once we split the trace server api def from the impl def

PR #5076, opened 2025-07-14
chore(weave): Add TRY401 lint rule

andrewtruong:
https://docs.astral.sh/ruff/rules/verbose-log-message/

### DIFF
--- a/weave/integrations/autogen/autogen_sdk.py
+++ b/weave/integrations/autogen/autogen_sdk.py
@@ -483,7 +483,7 @@ def _get_class_and_subclass_patchers(
                     patchers.append(patcher)
 
     except AttributeError as e:
-        logger.warning(f"Class {class_name} not found in {module_path}, skipping")
+        logger.warning(f"Class {class_name} not found in {module_path}, skipping: {e}")
         return patchers
     except Exception as e:
         logger.exception(
--- a/weave/integrations/autogen/autogen_sdk.py
+++ b/weave/integrations/autogen/autogen_sdk.py
@@ -228,7 +228,7 @@ def _create_wrapper_async_generator(
                     yield value
             except Exception as e:
                 logger.exception(
-                    f"{e}\nError in autogen async gen wrapper for {fn.__name__}",
+                    f"Error in autogen async gen wrapper for {fn.__name__}",
                     stacklevel=2,
                 )
                 # Fall back to the original function if our instrumentation fails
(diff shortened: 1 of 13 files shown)


weave/integrations/autogen/autogen_sdk.py
```diff
@@ -483,11 +483,11 @@ def collect_subclasses(cls: type) -> None:
                     patchers.append(patcher)
 
     except AttributeError as e:
-        logger.warning(f"Class {class_name} not found in {module_path}, skipping: {e}")
+        logger.warning(f"Class {class_name} not found in {module_path}, skipping")
```
jamie-rasmussen: This is a `.warning` not a `.exception`

weave/trace_server/environment.py
```diff
@@ -71,9 +71,7 @@ def wf_clickhouse_max_memory_usage() -> Optional[int]:
     try:
         return int(mem)
     except ValueError as e:
-        logger.exception(
-            f"WF_CLICKHOUSE_MAX_MEMORY_USAGE value '{mem}' is not a valid. Error: {str(e)}"
-        )
+        logger.exception(f"WF_CLICKHOUSE_MAX_MEMORY_USAGE value '{mem}' is not a valid")
```
jamie-rasmussen: I see it was like this before, but "not a valid" should be "not valid"?

weave/trace_server/environment.py
```diff
@@ -86,7 +84,7 @@ def wf_clickhouse_max_execution_time() -> Optional[int]:
         return int(time)
     except ValueError as e:
         logger.exception(
-            f"WF_CLICKHOUSE_MAX_EXECUTION_TIME value '{time}' is not a valid. Error: {str(e)}"
+            f"WF_CLICKHOUSE_MAX_EXECUTION_TIME value '{time}' is not a valid"
```
jamie-rasmussen: Same as above

weave/integrations/autogen/autogen_sdk.py
andrewtruong: fixed

weave/trace_server/environment.py
andrewtruong: fixed

weave/trace_server/environment.py
andrewtruong: fixed

PR #5081, opened 2025-07-14
docs(weave): Add pricing info to FAQs page

J2-D2-3PO:
## Description

- Fixes https://wandb.atlassian.net/browse/DOCS-1633

Per request from sales team, adding info from https://wandb.ai/site/pricing/ to https://weave-docs.wandb.ai/guides/tracking/faqs/

## Testing

How was this PR tested?

### DIFF
--- a/docs/docs/guides/tracking/faqs.md
+++ b/docs/docs/guides/tracking/faqs.md
@@ -148,3 +148,40 @@ pref_scorer = PreferenceScorer(other_model=model_b)
 evaluation = Evaluation(dataset=dataset, scorers=[pref_scorer])
 evaluation.evaluate(model_a)
 ```
+
+## Pricing FAQs
+
+The following section provides answers to pricing FAQs. For the latest information, see the [pricing page](https://wandb.ai/site/pricing/).
+
+### How is Weave data ingestion calculated?
+We define ingested bytes as bytes that we receive, process, and store on your behalf. This includes trace metadata, LLM inputs/outputs, and any other information you explicitly log to Weave, but does not include communication overhead (e.g., HTTP headers) or any other data that is not placed in long term storage. We count bytes as "ingested" only once at the time they are received and stored.
+
+### How much Weave data ingestion do I need?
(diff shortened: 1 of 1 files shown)


docs/docs/guides/tracking/faqs.md
```diff
@@ -148,3 +148,40 @@ pref_scorer = PreferenceScorer(other_model=model_b)
 evaluation = Evaluation(dataset=dataset, scorers=[pref_scorer])
 evaluation.evaluate(model_a)
 ```
+
+## Pricing FAQs
+
+The following section provides answers to pricing FAQs. For the latest information, see the [pricing page](https://wandb.ai/site/pricing/).
+
+### How is Weave data ingestion calculated?
+We define ingested bytes as bytes that we receive, process, and store on your behalf. This includes trace metadata, LLM inputs/outputs, and any other information you explicitly log to Weave, but does not include communication overhead (e.g., HTTP headers) or any other data that is not placed in long term storage. We count bytes as "ingested" only once at the time they are received and stored.
```
neutralino1, line 157: Some customers think that deleting data will reduce their bill. It may be good to clarify that we bill on "ingested" and not "stored". So deletion does not change change their bill.

neutralino1 (approved):
One comment.

docs/docs/guides/tracking/faqs.md
J2-D2-3PO: @neutralino1 this makes sense to clarify this in pricing info. However, I think we should just push this as is, since this the exact language approved by marketing atm, and documenting pricing info in the docs can get a little tricky for that reason - I'd be concerned about having two even slightly different answers RE a pricing question (your suggested wording vs. what's on https://wandb.ai/site/pricing/)

Short-term, let's push this as is. Long-term, I would advocate for unifying all pricing info on https://wandb.ai/site/pricing/, and linking there from the docs as need be. Reasoning is being discussed here with sales https://weightsandbiases.slack.com/archives/C01DX8LRZEJ/p1752518283958629?thread_ts=1751571576.613069&cid=C01DX8LRZEJ

PR #5082, opened 2025-07-14
feat(ui): Add prompt creation and editing feature.

jamie-rasmussen:
## Description

Adds the ability to create and edit prompt objects through the UI. Largely cloned from dataset creation, there are opportunities for cleaning up common code. 

Screenshot 2025-07-14 at 2 01 41 PM
Screenshot 2025-07-14 at 2 01 31 PM

## Testing

How was this PR tested?

### DIFF
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
@@ -28,6 +28,7 @@ import {
   PEEK_PARAM,
   SHOW_FEEDBACK_PARAM,
   useClosePeek,
+  useEntityProject,
   usePeekLocation,
   useWeaveflowCurrentRouteContext,
   useWeaveflowRouteContext,
@@ -822,7 +823,7 @@ const ObjectVersionsPageBinding = () => {
 };
 
 const PromptsPageBinding = () => {
-  const {entity, project} = useParamsDecoded<Browse3TabParams>();
+  const {entity, project} = useEntityProject();
   const query = useURLSearchParamsDict();
   const filters = useMemo(() => {
     if (query.filter === undefined) {
@@ -847,8 +848,6 @@ const PromptsPageBinding = () => {
 
   return (
     <PromptsPage
-      entity={entity}
-      project={project}
       initialFilter={filters}
       onFilterUpdate={onFilterUpdate}
     />
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
@@ -847,10 +847,7 @@ const PromptsPageBinding = () => {
   );
 
   return (
-    <PromptsPage
-      initialFilter={filters}
-      onFilterUpdate={onFilterUpdate}
-    />
+    <PromptsPage initialFilter={filters} onFilterUpdate={onFilterUpdate} />
   );
 };
 
--- a/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
+++ b/weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
@@ -55,6 +55,7 @@ import {OpsPage} from './Browse3/pages/OpsPage/OpsPage';
 import {OpVersionPage} from './Browse3/pages/OpsPage/OpVersionPage';
 import {OpVersionsPage} from './Browse3/pages/OpsPage/OpVersionsPage';
 import {PlaygroundPage} from './Browse3/pages/PlaygroundPage/PlaygroundPage';
+import {PromptsPage} from './Browse3/pages/PromptsPage/PromptsPage';
 import {ScorersPage} from './Browse3/pages/ScorersPage/ScorersPage';
 import {TablePage} from './Browse3/pages/TablePage';
 import {TablesPage} from './Browse3/pages/TablesPage';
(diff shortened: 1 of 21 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
```diff
@@ -817,6 +821,40 @@ const ObjectVersionsPageBinding = () => {
   );
 };
 
+const PromptsPageBinding = () => {
+  const {entity, project} = useParamsDecoded<Browse3TabParams>();
```
neutralino1: Maybe use the centralized
```typescript
const {entity, project} = useEntityProject();
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PromptsPage/PromptsPage.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+const PROMPT_TYPE = 'Prompt' as const;
+
+export const PromptsPage: React.FC<{
+  entity: string;
+  project: string;
+  initialFilter?: PromptFilter;
+  onFilterUpdate?: (filter: PromptFilter) => void;
+}> = props => {
+  const {entity, project} = props;
```
neutralino1: Same here, can we use the context instead of passing those down the stack?
```suggestion
  const {entity, project} = useEntityProject();
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PromptsPage/PromptsPage.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  const [selectedVersions, setSelectedVersions] = useState<string[]>([]);
+
+  const onCompare = () => {
+    history.push(router.compareObjectsUri(entity, project, selectedVersions));
+  };
+
+  const title = useMemo(() => {
+    if (filter.objectName) {
+      return 'Versions of ' + filter.objectName;
```
neutralino1:
```suggestion
      return `Versions of ${filter.objName}`;
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/EmptyContent.tsx
```diff
@@ -6,10 +6,41 @@ import {TargetBlank} from '../../../../../../common/util/links';
...
 import {useDatasetSaving} from '../../datasets/useDatasetSaving';
+import {CreatePromptDrawer} from '../../prompts/CreatePromptDrawer';
+import {usePromptSaving} from '../../prompts/usePromptSaving';
 import {MonitorDrawerRouter} from '../MonitorsPage/MonitorFormDrawer';
 import {EmptyProps} from './Empty';
 import {Link} from './Links';
 
+const NewPromptButton: React.FC = () => {
+  const [isDrawerOpen, setIsDrawerOpen] = useState(false);
+  const {entity, project} = useParams<{entity: string; project: string}>();
```
neutralino1:
```suggestion
  const {entity, project} = useEntityProject();
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/common/EmptyContent.tsx
```diff
@@ -160,6 +191,9 @@ export const EMPTY_PROPS_PROMPTS: EmptyProps = {
         prompt basics
       </TargetBlank>
       .
+      <Box sx={{mt: 2}}>
```
neutralino1: Tailwind?

weave-js/src/components/PagePanelComponents/Home/Browse3/prompts/CreatePromptDrawer.tsx
```diff
@@ -0,0 +1,262 @@
...
+          sx={{
+            display: 'flex',
+            justifyContent: 'space-between',
+            alignItems: 'center',
+            height: 44,
+            minHeight: 44,
+            pl: 2,
+            pr: 1,
+            borderBottom: '1px solid',
+            borderColor: 'divider',
```
neutralino1: Tailwind.

Core has lint rules to prevent those. They will not apply to Weave code at first as Ben is porting it as a separate module still, but they will eventually.

weave-js/src/components/PagePanelComponents/Home/Browse3/prompts/PromptVersionPage.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const getTsClient = useGetTraceServerClientContext();
+  const tsClient = getTsClient();
+
+  const [isEditing, setIsEditing] = useState(false);
+
+  const entityName = objectVersion.entity;
+  const projectName = objectVersion.project;
+  const objectName = objectVersion.objectId;
+  const objectVersionIndex = objectVersion.versionIndex;
+  const {createdAtMs} = objectVersion;
```
neutralino1: Why not
```suggestion
  const {
    entity: entityName,
    project: projectName,
    objectId: objectName,
    versionIndex: objectVersionIndex,
    createdAtMs
  } = objectVersion;
```

weave-js/src/components/PagePanelComponents/Home/Browse3/pages/PromptsPage/PromptsPage.tsx
zbirenbaum: I think the idea is that by passing them as props if either one changes it will force the page to re-render immediately rather than waiting until the next cycle for the hook. I made the change and it seems to work though.

Similar to above, since every other page does this I don't see a problem with continuing the pattern here and then bulk refactoring if we decide its unnecessary.

weave-js/src/components/PagePanelComponents/Home/Browse3.tsx
zbirenbaum: I actually think it is correct to use this here since every other page definition does `useParamsDecoded` and passes the entity + project as props. I'm not entirely clear on the difference between it and `useEntityProject` but it seemed to work without side effects when I made the switch.

If we think this is tech-debt, updating it is a simple search and replace as long as we keep things consistent.

PR #5098, opened 2025-07-15
Async Python Package from WandB

zamborg:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?


weave/trace/async_client/async_http_trace_server.py
```diff
@@ -0,0 +1,470 @@
...
+import logging
+from collections.abc import AsyncIterator
+from typing import Any, Optional
+
+try:
+    import httpx
+except ImportError:
+    raise ImportError(
+        "The async WeaveClient requires httpx. Install it with: pip install httpx"
+    )
```
andrewtruong, line 15: I think it's safe to make this a dep

weave/trace/async_client/async_http_trace_server.py
```diff
@@ -0,0 +1,470 @@
...
+
+class AsyncRemoteHTTPTraceServer(tsi.TraceServerInterface):
+    """Async HTTP implementation of TraceServerInterface."""
+
+    def __init__(
+        self,
+        trace_server_url: str,
+        *,
+        remote_request_bytes_limit: int = REMOTE_REQUEST_BYTES_LIMIT,
+        auth: Optional[tuple[str, str]] = None,
```
andrewtruong, line 38: if importing `from __future__ import annotations`, use it!

This should be `tuple[str, str] | None` for example

PR #5105, opened 2025-07-15
docs(weave): Group by threads

J2-D2-3PO:
## Description

- Fixes https://wandb.atlassian.net/browse/DOCS-1644

1. Documents Group by Threads feature in new page `https://weave-docs.wandb.ai/guides/tracking/threads`
2. Adds `/guides/tracking/threads` to sidebar under 🔄 Iteration > Tracing & Debugging

## Testing

yarn start on local


adrnswanberg (approved):

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+A _Thread_ is a logical grouping of related calls that share a common conversational context. A Thread:
+
+- Has a unique `thread_id`
+- Contains one or more _turns_
+- Maintains context across calls
+- Represent complete user sessions or interaction flows
+
+### Turn
+
+A _Turn_ is a high-level operation within a Thread. Turns are top-level calls where `call.id == turn_id`. Each Turn:
```
chance-wnb: > Turns are top-level calls where `call.id == turn_id`.

I think this is too implementation-detailed. We might not want to expose it to the user, or you might have to also explain what `turn_id` is . 

I highly recommend phrasing it in a different way ( something like "turns are the calls of the immediate children ops under a thread" ). If not possible, removing it is also fine.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+### Call
+
+A _Call_ is any `@weave.op`-decorated function execution in your application.
+
+- _Turn calls_ are top-level operations that start new turns
+- _Nested calls_ are lower-level operations within a turn
+
+### Trace
+
+A _Trace_ shows the full call stack of a single operation, including nested calls. Threads are built on top of traces by grouping a specific type of call (turns) that share the same `thread_id`. For more information on Traces, see the [Tracing overview](tracing.mdx).
```
chance-wnb: `Threads are built on top of traces`

This might feel a little bit confusing to the users. In your opinion what relationship does a thread have with a trace?

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+
+## UI overview
+
+In the Weave sidebar, select **Threads** (located below **Scorers** and above **More**) to access the [Threads list view](#threads-list-view).
```
chance-wnb: super nit: `located below **Scorers** and above **More**` can we just mention that it can be found in the sidebar? I expect that we might adjust the position of , or completely remove the icon in the near future, and there is a good chance that we will forget updating the doc accordingly. So it might make some sense to word it in a future-proof way.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+- Lists recent threads in your project
+- Columns include number of turns, start time, and last updated
+- Click a row to open its [detail drawer](#threads-detail-drawer)
+
+![The Threads list view](imgs/threads-list.png)
+
+### Threads detail drawer
+
+- Click any row to open the detail drawer for that row
+- Shows all turns within a thread, ordered by execution
```
chance-wnb: `ordered by execution` might lack some precision. It might be useful to note that this is ordered by the "start time" of the turns. (maybe a user friendly way to say is "ordered by the time sequence of the executions") It is not ordered by the duration of the executions nor its ordered by ended_time/last_updated time of the turn calls.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+- Includes call-level metadata (latency, inputs, outputs)
+- Optionally shows message content or structured data if logged
```
chance-wnb, line 77: I don't think the messages feature is emphasized enough. which is the primary value we want to deliver to the customers. I would suggest adding this info:

1. Explain what are the messages displayed in the chat view. (They are the interactions with LLM providers)
2. What is the criteria to have things show up in the chat view, what ops/calls will not be displayed in the chat view.
3. What will happen if a turn doesn't emit any chat messages?
4. What interactions are available between the turn list view and the chat view (select to scroll, highlighting the current during scroll).
5. The ability to navigate to the trace view and back.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+
+![The Threads drawer view](imgs/threads-drawer.png)
+
+## Programmatic usage
+
+### API specification
```
chance-wnb: The API usage is not important, we could redirect users who are interested to read the auto-gened API references.
At any rate, the SDK usage is way more important. I would even suggest putting the SDK instructions first, even before explaining the UI interactions.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+# Alternative: Use explicit thread_id for session tracking
+session_id = "user_session_123"
+with weave.thread(session_id) as thread_ctx:
+    print(f"Session Thread ID: {thread_ctx.thread_id}")  # "user_session_123"
+
+    agent.process_user_message("Continue our previous conversation")  # Turn 1 in this session
+    agent.process_user_message("Can you summarize what we discussed?") # Turn 2 in this session
+```
+
+#### OpenAI integration
```
chance-wnb: I am sorry, but I think my design doc is misleading -- we are not yet patching the OpenAI agents yet, so this example will not create any threads automatically. We can remove this section.

The design doc talks about an approach we would adopt when we implement the OpenAI integration in order to prove that the concept of the entire project is feasible. However we have not yet had the OpenAI integration to support threads yet.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+# Get threads started in the last 24 hours
+yesterday = datetime.now() - timedelta(days=1)
+response = client.threads_query(ThreadsQueryReq(
+    project_id="my-project",
+    sortable_datetime_after=yesterday,
+    sort_by=[SortBy(field="start_time", direction="desc")]
+))
+```
+
+### Advanced examples
```
chance-wnb: If we put the SDK usage first, then these might be just `basic examples` :) not advanced examples.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+
+Each example in this section demonstrates a different strategy for organizing turns and threads in your application. For most examples, you should provide your own LLM call or system behavior inside the stub functions.
+
+- Use `weave.thread()` to create or resume a thread.
```
chance-wnb: I think this is the core concept of SDK usage. Putting a detailed overview would help.

I would suggest adding:

1. The parameter of `weave.thread(thread_id)`, what will happen if a parameter is specified vs omitted?
2. What is the return value of  `weave.thread(thread_id)`, what can it do?
3. How would nested thread context work? how could a user achieve desired thread inheritance vs new thread fork. Would forking a new thread cause the old thread to discontinue?

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,569 @@
...
+
+### Basic examples
+
+#### Query recent active threads
+
+This example fetches the 50 most recently updated threads. Replace `my-project` with your actual project ID.
+
+```python
+# Get most recently active threads
+response = client.threads_query(ThreadsQueryReq(
```
chance-wnb: I hope we can render an impression that user should prioritize using SDKs to create threads, and UI to preview them.  Programatic consumption of the threads should be a rare use case.

We can put this section last

J2-D2-3PO:
Note: Don't merge until we get the green light from eng, probably next week https://wandb.atlassian.net/browse/DOCS-1644?focusedCommentId=105839

J2-D2-3PO:
@chance-wnb I made some updates to address your feedback, with some open questions. Updates made:

* Explained the difference between passing vs omitting a `thread_id` in `weave.thread()`
* Documented the return value of `weave.thread()` (`ThreadContext.thread_id`)
* Clarified how nested thread contexts work and when a new thread is forked
* Noted that turn-level traces can be opened from the UI detail drawer

**Unknowns**
For the following, I wasn't able to find info in the design doc, and the demo Threads project doesn't have any Threads in it so I can't confirm these questions. Can you help me confirm the following?

1. What qualifies as a message in the chat view?

   * Are these only calls to LLM APIs (e.g., `openai.ChatCompletion.create`)?
   * Are there specific criteria for message extraction?

2. What happens if a turn has no messages?

   * Is the chat pane empty?
   * Are there any UI fallbacks?

3. Turn list ↔ chat pane interaction:

   * Does clicking a turn scroll the chat pane?
   * Does scrolling the chat highlight the active turn?

4. Trace navigation:

   * Can the user return from the trace view back to the thread detail drawer?
   * Is state preserved across that transition?

chance-wnb:
Thanks @J2-D2-3PO here are the answers to your questions:

1. What qualifies as a message in the chat view?

   Are these only calls to LLM APIs (e.g., openai.ChatCompletion.create)?
- Essentially yes. They are ~100% times the calls from our automatically patched 3rd part integrations. Please see the next answer for more details.

   Are there specific criteria for message extraction?
- Yes, the algorithm works by looking at the leaf calls from a turn, and see who have `usage` attribute on them. The calls with `usage` attribute who are also leaf calls will undergo message extraction. The `usage` attribute means the call has invoked LLM endpoints. And the reason that we only care about the leaf calls is because we roll up and sum `usage` across the LLM calls ancestor levels.

2. What happens if a turn has no messages?
   Is the chat pane empty?
- The chat pane show messages for all turns, so if a turn doesn't have any messages, some other turns might. So it is often times not empty.

   Are there any UI fallbacks?

- We are going to show an empty section for exactly that turn without any message. The UI is illustrated in https://www.figma.com/design/NSK3Ls0YoDCRFMec3BiThg/Threads-view---group-by?node-id=343-3973&m=dev

3. Turn list ↔ chat pane interaction:

  Does clicking a turn scroll the chat pane?
- Yes, it will pin to the location of that turn.

  Does scrolling the chat highlight the active turn?
- Yes

4. Trace navigation:

  Can the user return from the trace view back to the thread detail drawer?
- Yes, there will be a back button at the top left corner

image

  Is state preserved across that transition?
- Unfortunately no.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,523 @@
...
+- Contains one or more _turns_
+- Maintains context across calls
+- Represent complete user sessions or interaction flows
+
+### Turn
+
+A _Turn_ is a high-level operation within a Thread, displayed in the UI as individual rows in a thread view. Each Turn:
+
+- Represents one logical step in a conversation or workflow
+- Turns are the direct children of a thread context and are
```
chance-wnb: `and are May` might suggest the `are` part is extra

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,523 @@
...
+Each example in this section demonstrates a different strategy for organizing turns and threads in your application. For most examples, you should provide your own LLM call or system behavior inside the stub functions.
+
+- To track a session or conversation, use the `weave.thread()` context manager.
+- Decorate logical operations with `@weave.op` to track them as turns or nested calls.
+- If you pass a `thread_id`, Weave uses it to group all operations in that block under the same thread. If you omit the `thread_id`, Weave auto-generates a unique one for you.
+
+The return value from `weave.thread()` is a `ThreadContext` object with a `thread_id` property, which you can log, reuse, or pass to other systems.
+
+Nested `weave.thread()` contexts always start a new thread unless the same `thread_id` is reused. Ending a child context does not interrupt or overwrite the parent context. This allows for forked thread structures or layered thread orchestration, depending on your app logic.
+
```
chance-wnb, line 123: This section is really well written. Love it. Thanks!

docs/docs/guides/tracking/threads.md
chance-wnb: Again, I recommend putting `Query recent active threads`, `Query threads by activity level` and `Query recent threads only` to the last and group them with `## API specification` section.

We should put the `Thread creation` examples first, because that's the scenarios which we expect the users to use the most.

Threads retrievals through API is an edge use case. In the extreme cases, if user still want to do that, they should use our SDK wrappers instead of directly interacting with REST endpoint.

docs/docs/guides/tracking/threads.md
```diff
@@ -0,0 +1,575 @@
...
+The chat pane displays structured message data extracted from LLM calls made during each turn. This view gives you a conversational-style rendering of the interaction.
+
+![Chat view](imgs/threads-chat-view.png)
+
+#### What qualifies as a message?
+
+Messages are extracted from child calls within a turn that:
+
+- Have the `usage` attribute (indicating an LLM API call occurred)
```
chance-wnb: This is an implementation detail, which I exposed merely to explain things to you earlier. 

There is nowhere in the UI that we expose this attribute directly (it is indirectly as LLM cost in the Call Summary tab), so I feel that mentioning it here might be confusing to users. 

Is it possible that we don't mention it and still make the paragraph coherent?

J2-D2-3PO: Ah got it, yes, I've made an update here https://github.com/wandb/weave/pull/5105/commits/84fae974824b79d4ad9d999906f7b3c9ad6ce2a5

PR #5115, opened 2025-07-16
chore(weave): Add fixit and basic weave.op transform rule

andrewtruong:


rules/op_deco.py
andrewtruong, line 1: To create a new lint rule, simply define a python file and make a class that subclasses `LintRule`.  Then match on the relevant [libcst.Node](https://libcst.readthedocs.io/en/latest/nodes.html#) and optionally offer a `replacement`

PR #5116, opened 2025-07-17
fix(weave): Make EvaluationLogger work with Leaderboards

andrewtruong:
Modifies the leaderboard UI code to support `EvaluationLogger` (imperative-style) evals

See example code where we only log summary data:
```py
import weave
import random

weave.init('testing1009')

for k in range(2):
    ev = weave.EvaluationLogger(model=f"abc{k}", dataset="def")
    ev.log_summary({f"metric{i}": random.random() for i in range(4)})
```

image

### DIFF
--- a/weave/flow/eval_imperative.py
+++ b/weave/flow/eval_imperative.py
@@ -546,7 +546,7 @@ class EvaluationLogger(BaseModel):
         if summary_data:
             final_summary = summary_data
         if summary is not None:
-            final_summary = {**final_summary, **summary}
+            final_summary = {**final_summary, "output": summary}
 
         # Call the summarize op
         assert self._evaluate_call is not None, (
--- a/tests/flow/test_evaluation_imperative.py
+++ b/tests/flow/test_evaluation_imperative.py
@@ -65,10 +65,12 @@ def test_basic_evaluation(
     assert evaluate_call.inputs["self"]._class_name == "Evaluation"
     assert evaluate_call.inputs["model"]._class_name == "Model"
     assert evaluate_call.output == {
-        "avg_score": 1.0,
-        "total_examples": 3,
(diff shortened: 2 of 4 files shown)


weave-js/src/components/PagePanelComponents/Home/Browse3/views/Leaderboard/LeaderboardGrid.tsx
```diff
@@ -164,6 +164,7 @@ export const LeaderboardGrid: React.FC<LeaderboardGridProps> = ({
 
   const rows: RowData[] = useMemo(() => {
     const rowData: RowData[] = [];
+    console.log('data', data);
```
neutralino1: 🔪

weave/flow/eval_imperative.py
```diff
@@ -546,7 +546,7 @@ def log_summary(
         if summary_data:
             final_summary = summary_data
         if summary is not None:
-            final_summary = {**final_summary, **summary}
+            final_summary = {**final_summary, "output": summary}
```
neutralino1, line 549: Intentional?

This needs an SDK release I presume.

neutralino1 (commented):
Beautiful 😻 

A couple of comments.

weave/flow/eval_imperative.py
andrewtruong: Yes, this corrects the API to match the normal Evaluation, which puts summary stats under an `output` key

weave-js/src/components/PagePanelComponents/Home/Browse3/views/Leaderboard/LeaderboardGrid.tsx
andrewtruong: fixed

PR #5119, opened 2025-07-17
chore(weave): Add E266 lint rule

andrewtruong:
https://docs.astral.sh/ruff/rules/multiple-leading-hashes-for-block-comment/


docs/notebooks/audio_with_weave.ipynb
```diff
@@ -477,7 +477,7 @@
     "    ERROR = \"error\"\n",
     "\n",
     "\n",
-    "#### Session Update\n",
+    "# Session Update\n",
```
neutralino1, line 480: I think those are markdown headings. Should probably be conserved.

tests/integrations/openai_agents/openai_agents_test.py
```diff
@@ -122,7 +122,7 @@ async def homework_guardrail(ctx, agent, input_data):
         result = await Runner.run(triage_agent, "what is life")
 
     #####################
-    ### Result1 Block ###
+    # Result1 Block ###
```
neutralino1, line 125: Should be kept.

PR #5127, opened 2025-07-18
fix(weave): LlamaIndex test

ayulockin:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

The `test_llamaindex_quick_start` started failing with the new version of LlamaIndex -- this PR fixes the same.

The current behaviour is correct where the exact class name is seen as display name instead of the base class names.

## Testing

How was this PR tested?

PR #5128, opened 2025-07-18
chore(weave): Temporarily pin `azure-storage-blob`

andrewtruong:
A new version of the API is not compatible with Azurite.  I tried to skip the API version check in Azurite, but wasn't able to get it to work.

Pinning here is safe because it's only used server side

See related issue:
https://github.com/azure/azure-sdk-for-python/issues/42077

---

This PR also includes commits from https://github.com/wandb/weave/pull/5127 to unblock CI


gtarpenning (approved):
what are the image changes?

PR #5138, opened 2025-07-21
chore(dev): remove weave-js

bcsherma:
## Description

Removes weave-js from weave as this code now lives in the private monorepo.


neutralino1:
Was it merged in core?
Should we keep this around for a few days in case of 💥 ?

neutralino1:
@bcsherma Why close? 🤔

PR #5143, opened 2025-07-22
Griffin/weave error registry

gtarpenning:
## Description

This pr:
- adds an error registry so we only define errors in one place and they are handled across the trace server
- updates error handling in the client to give more detailed messages
- catches errors in the streaming iterator and re-raises them manually, which fastapi handlers can't do
- streaming handler returns **both** content chunks _and_ a status_code, which allows for dynamic status code header assignment. in practice this means that a stream can be 200 green and then error and turn into a 500 error.

## Testing

todo: 
- lots of manual testing

### DIFF
--- a/weave/trace_server/errors.py
+++ b/weave/trace_server/errors.py
@@ -85,7 +85,7 @@ class ErrorRegistry:
 
     def __init__(self) -> None:
         self._definitions: dict[type, ErrorDefinition] = {}
-        self._setup_builtin_errors()
+        self._setup_common_errors()
 
     def register(
         self,
@@ -129,27 +129,33 @@ class ErrorRegistry:
         except json.JSONDecodeError:
             return {"reason": exc_str}
 
-    def _setup_builtin_errors(self) -> None:
-        """Register all built-in error handlers."""
-        # Import here to avoid circular imports
-        from src import id_converters as idc
-
-        from weave.trace_server import clickhouse_trace_server_batched as cts
-        from weave.trace_server.trace_server_converter import InvalidExternalRef
-
-        # InsertTooLarge (413)
+    def _setup_common_errors(self) -> None:
+        """Register common/standard library errors that don't depend on domain-specific modules."""
+        # Our own error types (413)
         self.register(InsertTooLarge, 413, self._format_insert_too_large)
-
-        # RequestTooLarge (413)
         self.register(RequestTooLarge, 413, lambda exc: {"reason": "Request too large"})
 
-        # QueryMemoryLimitExceeded (502)
+        # Our own error types (502)
         self.register(
             QueryMemoryLimitExceeded,
             502,
             lambda exc: {"reason": "Query memory limit exceeded"},
         )
 
+        # Our own error types (400)
+        self.register(MissingLLMApiKeyError, 400, self._format_missing_llm_api_key)
+
+        # Our own error types (403)
+        self.register(InvalidFieldError, 403, self._default_json_formatter)
+
+        # Our own error types (404)
+        self.register(NotFoundError, 404, self._default_json_formatter)
+        self.register(ObjectDeletedError, 404, self._format_object_deleted_error)
+
+        # Standard library exceptions
+        self.register(ValueError, 400, self._default_json_formatter)
+
+        # Third-party library exceptions that are commonly used
         # Timeout errors (504)
         self.register(
             requests.exceptions.ReadTimeout, 504, lambda exc: {"reason": "Read timeout"}
@@ -172,19 +178,8 @@ class ErrorRegistry:
             lambda exc: {"reason": "Temporary backend error"},
         )
 
-        # Client errors (400)
-        self.register(ValueError, 400, self._default_json_formatter)
-        self.register(InvalidExternalRef, 400, self._default_json_formatter)
-        self.register(MissingLLMApiKeyError, 400, self._format_missing_llm_api_key)
-
-        # Permission errors (403)
-        self.register(InvalidFieldError, 403, self._default_json_formatter)
+        # GraphQL transport errors (403)
         self.register(TransportQueryError, 403, self._format_transport_query_error)
-
-        # Not found errors (404)
-        self.register(cts.NotFoundError, 404, self._default_json_formatter)
-        self.register(idc.ProjectNotFound, 404, self._default_json_formatter)
-        self.register(ObjectDeletedError, 404, self._format_object_deleted_error)
 
     def _format_insert_too_large(self, exc: Exception) -> dict[str, Any]:
         """Format InsertTooLarge exception."""
--- a/weave/trace_server/errors.py
+++ b/weave/trace_server/errors.py
@@ -1,4 +1,10 @@
 import datetime
+import json
+from typing import Any, Callable, Optional
+
+import clickhouse_connect
+import requests
+from gql.transport.exceptions import TransportQueryError
 
 
 class Error(Exception):
@@ -21,6 +27,12 @@ class InvalidRequest(Error):
 
 class InsertTooLarge(Error):
     """Raised when a single insert is too large."""
+
+    pass
+
+
+class QueryMemoryLimitExceeded(Error):
+    """Raised when a query memory limit is exceeded."""
 
     pass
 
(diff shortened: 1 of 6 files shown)


weave/trace_server/errors.py
```diff
@@ -39,15 +65,245 @@ def __init__(self, message: str, api_key_name: str):
...
+def register_error(
+    exception_class: type,
+    status_code: int,
+    formatter: Optional[Callable[[Exception], dict[str, Any]]] = None,
+) -> None:
+    """Convenience function to register an error with the global registry."""
+    get_error_registry().register(exception_class, status_code, formatter)
+
+
+def error_handler(
```
neutralino1: Where is this used?

weave/trace_server/errors.py
```diff
@@ -39,15 +65,245 @@ def __init__(self, message: str, api_key_name: str):
...
+
+    def _format_error_to_json_with_extra(
+        self, exc: Exception, extra_fields: dict[str, Any]
+    ) -> dict[str, Any]:
+        """Helper to format exception as JSON or fallback to reason, always adding extra fields."""
+        exc_str = str(exc)
+        try:
+            result = json.loads(exc_str)
+        except json.JSONDecodeError:
+            result = {"reason": exc_str}
```
neutralino1:
```suggestion
        result = self._default_json_formatter(exc)
```

weave/trace_server/errors.py
```diff
@@ -39,15 +65,245 @@ def __init__(self, message: str, api_key_name: str):
...
+        # GraphQL transport errors (403)
+        self.register(TransportQueryError, 403, self._format_transport_query_error)
+
+    def _format_error_to_json(self, exc: Exception) -> dict[str, Any]:
+        """Helper to format exception as JSON or fallback to reason field."""
+        exc_str = str(exc)
+        try:
+            return json.loads(exc_str)
+        except json.JSONDecodeError:
+            return {"reason": exc_str}
```
neutralino1: Same logic as `_default_json_formatter`, remove?

weave/trace_server/errors.py
```diff
@@ -39,15 +65,245 @@ def __init__(self, message: str, api_key_name: str):
...
+
+    def _setup_common_errors(self) -> None:
+        """Register common/standard library errors that don't depend on domain-specific modules."""
+        # Our own error types
+        self.register(RequestTooLarge, 413, lambda exc: {"reason": "Request too large"})
+        self.register(InvalidRequest, 400, self._default_json_formatter)
+        self.register(InsertTooLarge, 413, self._format_insert_too_large)
```
neutralino1: Why not simply use `self._default_json_formatter`?

weave/trace_server/errors.py
```diff
@@ -39,15 +65,245 @@ def __init__(self, message: str, api_key_name: str):
...
+            result = json.loads(exc_str)
+        except json.JSONDecodeError:
+            result = {"reason": exc_str}
+
+        result.update(extra_fields)
+        return result
+
+    def _format_insert_too_large(self, exc: Exception) -> dict[str, Any]:
+        """Format InsertTooLarge exception."""
+        return self._format_error_to_json(exc)
```
neutralino1: Unnecessary.

weave/trace_server/errors.py
```diff
@@ -39,15 +65,245 @@ def __init__(self, message: str, api_key_name: str):
...
+            for error in transport_exc.errors:
+                if error.get("extensions", {}).get("code") == "PERMISSION_ERROR":
+                    return {"reason": "Forbidden"}
+                if error.get("message") == "project not found" and error.get(
+                    "path"
+                ) == ["upsertBucket"]:
+                    # This seems counter intuitive, but gorilla returns this error when the project exists
+                    # but the user does not have access to it. This seems like a bug on Gorilla's side.
+                    return {"reason": "Forbidden"}
+        return {"reason": "Forbidden"}
```
neutralino1: In all cases you return `{"reason": "Forbidden"}`. What is the purpose of this?

weave/trace_server/errors.py
```diff
@@ -39,15 +65,245 @@ def __init__(self, message: str, api_key_name: str):
...
+
+    def _format_missing_llm_api_key(self, exc: Exception) -> dict[str, Any]:
+        """Format MissingLLMApiKeyError with api_key field."""
+        if isinstance(exc, MissingLLMApiKeyError):
+            return self._format_error_to_json_with_extra(
+                exc, {"api_key": exc.api_key_name}
+            )
+        return self._format_error_to_json(exc)
```
neutralino1: Make the default formatter accept extra fields and default them to `{}` if `isinstance(exc, MissingLLMApiKeyError)` is false. Then you don't need `format_error_to_json_with_extra`.

neutralino1 (commented):
Was this written by AI? 🤔

weave/trace_server/errors.py
gtarpenning: nowhere, but I thought it might be more convenient if someone had an error in another module they wanted to keep there, but still register. happy to remove if you think it's unnecessary.

weave/trace_server/errors.py
gtarpenning: yeah trax, it used to have more specific handling, but I eventually removed all of it.

gtarpenning:
> Was this written by AI? 🤔

haha what isn't? But unfortunately no this was just two separate PRs that got merged together poorly. I feel like I would have more faith in the AI ;) one moment

weave/trace_server/errors.py
gtarpenning: this is just the exact logic that was implemented before. i don't really get it either but I was trying to change as little as possible

weave/trace_server/errors.py
```diff
@@ -39,15 +65,245 @@ def __init__(self, message: str, api_key_name: str):
...
+
+    def get_definition(self, exception_class: type) -> Optional[ErrorDefinition]:
+        """Get error definition for an exception class."""
+        return self._definitions.get(exception_class)
+
+    def get_all_definitions(self) -> dict[type, ErrorDefinition]:
+        """Get all registered error definitions."""
+        return self._definitions.copy()
+
+    def handle_exception(self, exc: Exception) -> tuple[dict[str, Any], int]:
```
neutralino1: I presume this the API used by endpoints to treat a particular exception? I don't see it used anywhere.

weave/trace_server/errors.py
neutralino1: It looks sweet but in the spirit of YAGNI, let's remove if not necessary.

weave/trace_server/errors.py
gtarpenning: yeah, here for example: https://github.com/wandb/core/pull/32523/files#r2248846853

gtarpenning: and here, where we register the fastapi error handlers: https://github.com/wandb/core/pull/32523/files#r2248849736

weave/trace_server/errors.py
```diff
@@ -39,15 +76,227 @@ def __init__(self, message: str, api_key_name: str):
...
+        return {"reason": exc_str}
+
+
+def _format_error_to_json_with_extra(
+    exc: Exception, extra_fields: dict[str, Any]
+) -> dict[str, Any]:
+    """Helper to format exception as JSON or fallback to reason, always adding extra fields."""
+    result = _format_error_to_json(exc)
+    result.update(extra_fields)
+    return result
```
neutralino1: Merge those two into a single function that accepts an optional `extra_fields` arg.

weave/trace_server/errors.py
```diff
@@ -39,15 +76,227 @@ def __init__(self, message: str, api_key_name: str):
...
+        for error in transport_exc.errors:
+            if error.get("extensions", {}).get("code") == "PERMISSION_ERROR":
+                return {"reason": "Forbidden"}
+            if error.get("message") == "project not found" and error.get("path") == [
+                "upsertBucket"
+            ]:
+                # This seems counter intuitive, but gorilla returns this error when the project exists
+                # but the user does not have access to it. This seems like a bug on Gorilla's side.
+                return {"reason": "Forbidden"}
+    return {"reason": "Forbidden"}
```
neutralino1: I know this is not a new function in the PR, but this seems like useless code since it always returns `{"reason": "Forbidden"}`. Let's yank it.

neutralino1 (approved):
A couple of nits.

pyproject.toml
```diff
@@ -87,6 +87,8 @@ trace_server = [
   "openinference-semantic-conventions>=0.1.17",
   # For emoji shortcode support in Feedback
   "emoji>=2.12.1",
+  # ClickHouse support
+  "clickhouse_connect==0.7.0",
```
gtarpenning, line 91: this was added because the docs weren't building correctly, due to a very long stack trace that ends in the errors module being loaded, which has clickhouse specific errors

weave/trace_server/errors.py
```diff
@@ -19,18 +30,44 @@ class InvalidRequest(Error):
     pass
 
 
+# Clickhouse errors
+class QueryMemoryLimitExceeded(Error):
```
andrewtruong: Errors should always end in `Error`

weave/trace_server/errors.py
```diff
@@ -39,15 +76,211 @@ def __init__(self, message: str, api_key_name: str):
...
+
+    def get_definition(self, exception_class: type) -> Optional[ErrorDefinition]:
+        """Get error definition for an exception class."""
+        return self._definitions.get(exception_class)
+
+    def get_all_definitions(self) -> dict[type, ErrorDefinition]:
+        """Get all registered error definitions."""
+        return self._definitions.copy()
+
+    def handle_exception(self, exc: Exception) -> tuple[dict[str, Any], int]:
```
andrewtruong: Prefer a dataclass here instead of tuple.  It will be easier to read

pyproject.toml
andrewtruong: I don't understand this comment

PR #5144, opened 2025-07-22
chore(weave): display error message when provided in client

gtarpenning:
## Description

Partially addresses: [WB-25859](https://wandb.atlassian.net/browse/WB-25859)

When the backend provides an error message to the client, use it! This should always be safe, because any sensitive errors should never make it out of the server...

## Testing

Prod
Screenshot 2025-07-22 at 11 50 09 AM

Branch
image

### DIFF
--- a/weave/trace_server_bindings/http_utils.py
+++ b/weave/trace_server_bindings/http_utils.py
@@ -1,0 +1,73 @@
+import json
+import logging
+from typing import TYPE_CHECKING, Union
+
+from weave.trace_server import requests
+
+if TYPE_CHECKING:
+    from weave.trace_server_bindings.models import EndBatchItem, StartBatchItem
+
+logger = logging.getLogger(__name__)
+
+
+def log_dropped_call_batch(
+    batch: list[Union["StartBatchItem", "EndBatchItem"]], e: Exception
+) -> None:
+    """Log details about a dropped call batch for debugging purposes."""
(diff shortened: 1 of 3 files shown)


bcsherma (approved):
I think it would be nice if we could maintain the error message that is associated with the code, eg Forbidden for URL in the case of 403. Not super informative if the user understands the codes already but feels cleaner to me. So in your example if the last line were
```
requests.exceptions.HTTPError: 403 Client Error: forbidden for url: Field BROKENHAHA is no allowed
```
or maybe break it up over two lines or something. Like

```
requests.exceptions.HTTPError: 403 Client Error: forbidden for url
response: Field BROKENHAHA is no allowed
```
take it or leave it 🤷

gtarpenning:
> I think it would be nice if we could maintain the error message that is associated with the code, eg Forbidden for URL in the case of 403. Not super informative if the user understands the codes already but feels cleaner to me. So in your example if the last line were
> 
> ```
> requests.exceptions.HTTPError: 403 Client Error: forbidden for url: Field BROKENHAHA is no allowed
> ```
> 
> or maybe break it up over two lines or something. Like
> 
> ```
> requests.exceptions.HTTPError: 403 Client Error: forbidden for url
> response: Field BROKENHAHA is no allowed
> ```
> 
> take it or leave it 🤷

done!

bcsherma (approved):

PR #5145, opened 2025-07-22
fix(weave): actually do validation on client side call queries

gtarpenning:
## Description

Partially solves: [WB-25859](https://wandb.atlassian.net/browse/WB-25859)

Currently we allow extra keys in both the `CallsFilter`, `SortBy`, and `Query` portions of the `client.get_calls` query, which basically renders the existing validation worthless! Lets be stricter, and tell users if they have typed in something wrong...

This predictably also exposes a few unrelated errors for incorrect usages of these types!

## Testing

- add explicit tests
- fix broken tests that weren't being validated

[WB-25859]: https://wandb.atlassian.net/browse/WB-25859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/test_publish_round_trip.py
+++ b/tests/trace/test_publish_round_trip.py
@@ -19,6 +19,8 @@ def test_publish_round_trip_query_object(client) -> None:
     ref = weave.publish(query)
     res = client.server.refs_read_batch(RefsReadBatchReq(refs=[ref.uri()]))
 
+    # remove the extra python class information that round tripping generates
+    # specifically: _type, _class_name, _bases
     clean = {"$expr": res.vals[0]["$expr"]}
     query_2 = Query.model_validate(clean)
     assert query_2 == query
--- a/tests/trace/test_publish_round_trip.py
+++ b/tests/trace/test_publish_round_trip.py
@@ -18,7 +18,9 @@ def test_publish_round_trip_query_object(client) -> None:
     query = Query(**query_raw)
     ref = weave.publish(query)
     res = client.server.refs_read_batch(RefsReadBatchReq(refs=[ref.uri()]))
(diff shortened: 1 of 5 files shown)


tests/trace/test_publish_round_trip.py
```diff
@@ -18,7 +18,9 @@ def test_publish_round_trip_query_object(client) -> None:
     query = Query(**query_raw)
     ref = weave.publish(query)
     res = client.server.refs_read_batch(RefsReadBatchReq(refs=[ref.uri()]))
-    query_2 = Query.model_validate(res.vals[0])
+
+    clean = {"$expr": res.vals[0]["$expr"]}
```
andrewtruong, line 24: what does `clean` mean?

gtarpenning: without `_type` and `_base_classes` and the other stuff that round tripping generates apparently

andrewtruong: can you add that as a comment

gtarpenning: yep

weave/trace_server/trace_server_interface.py
```diff
@@ -363,6 +363,8 @@ class CompletionsCreateRes(BaseModel):
 
 
 class CallsFilter(BaseModel):
+    model_config = ConfigDict(extra="forbid")
```
gtarpenning, line 366: this is the change thats affecting you

neutralino1: Noted, thanks.

PR #5147, opened 2025-07-22
chore(weave): Add clear self-contained benchmarks for weave perf

andrewtruong:


scripts/benchmark.py
```diff
@@ -0,0 +1,341 @@
...
+
+A typer app for selecting and running benchmarks from the benchmarks/ directory.
+"""
+#!/usr/bin/env -S uv run --script
+# /// script
+# dependencies = [
+#     "rich==14.0.0",
+#     "typer==0.16.0",
+# ]
+# ///
```
zbirenbaum, line 12: Is this necessary?

scripts/benchmarks/basic_openai_logging.py
```diff
@@ -0,0 +1,364 @@
+"""Benchmark Weave overhead on OpenAI API calls."""
+
+#!/usr/bin/env -S uv run --script
+# /// script
+# dependencies = [
+#     "openai==1.97.1",
+#     "weave==0.51.56",
+#     "rich==14.0.0",
+# ]
+# ///
```
zbirenbaum, line 10: Same here

scripts/benchmarks/weave_setup_time.py
```diff
@@ -0,0 +1,322 @@
+"""Benchmark Weave import and initialization time."""
+
+#!/usr/bin/env -S uv run --script
+# /// script
+# dependencies = [
+#     "weave==0.51.56",
+#     "rich==14.0.0",
+# ]
+# ///
```
zbirenbaum, line 9: Here too

tests/trace/test_import.py
```diff
@@ -1,13 +1,18 @@
...
     root_dir = Path.cwd().parent.parent  # root dir of repo
     monkeypatch.setenv("PYTHONPATH", str(root_dir))
 
-    from scripts.benchmark_import import run_single_import
-
-    import_time = run_single_import()
+    import_time = time_weave_import()
 
     # Ideally the import takes < 1s, but in CI it can take up to 7s.
     assert import_time < 7, f"Import time was {import_time} seconds"
```
zbirenbaum: I changed this to 8 so may need to rebase and fix conflict

scripts/benchmarks/weave_setup_time.py
```diff
@@ -0,0 +1,322 @@
...
+    console.print("[bold]Weave Setup Time Benchmark[/bold]")
+    console.print(f"Running {args.iterations} iterations...\n")
+
+    # Run the benchmark
+    results = run_benchmark_iterations(args.iterations)
+
+    # Write results to CSV in the same directory as the script
+    script_dir = os.path.dirname(os.path.abspath(__file__))
+    csv_filename = os.path.join(script_dir, "weave_setup_time_results.csv")
+    write_results_to_csv(results, csv_filename)
```
zbirenbaum: Might want to make `-w` `--write` a cli arg with a default of false which can be a boolean or a string value. if string, name the file the input, if set with no name, default to the existing one.

scripts/benchmarks/weave_setup_time.py
```diff
@@ -0,0 +1,322 @@
...
+
+    # Read results back from CSV and display
+    csv_data = read_results_from_csv(csv_filename)
```
zbirenbaum: Could we do this in memory without making a persistent file?
normally a file output is optional for scripts like these.

If that's too much work then I'd say make sure to gitignore the csv

zbirenbaum: Also, might be good to have a --read flag which displays the results from a file to terminal, otherwise re-runs the test

zbirenbaum (commented):
Awesome util, was wishing for something like this.

Couple of questions and recommendations.

scripts/benchmarks/basic_openai_logging.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+    for _ in range(num_warmup):
+        try:
+            make_openai_call()
+        except Exception:
+            # Ignore warm-up errors
+            pass
+
+
+def time_function_calls(func, iterations: int = 10, warmup: int = 3) -> list[float]:
```
chance-wnb: It feels that this function (added with a `wram_up_func` parameter to customize) and the ones below (calculate_stats, run_multiple_rounds and create_results_table_from_csv) might belong to some common utility set instead of belong to this specific benchmark.

Since you are pioneering this work, you might want to set a paradigm for the future developers.

scripts/benchmarks/weave_setup_time.py
```diff
@@ -0,0 +1,322 @@
...
+        else:
+            console.print(f"[red]Full test failed: {result.stderr}[/red]")
+            return 0.0, 0.0
+    finally:
+        # Clean up temp file
+        if os.path.exists(temp_file):
+            os.remove(temp_file)
+
+
+def run_benchmark_iterations(iterations: int) -> dict[str, list[float]]:
```
chance-wnb, line 131: Just a reminder that these reappears in the other benchmark

scripts/benchmark.py
andrewtruong: Yes, this is what makes it a self-contained uv script.
See: https://peps.python.org/pep-0723/

scripts/benchmarks/weave_setup_time.py
chance-wnb: +1

And [io.BytesIO](https://docs.python.org/3/library/io.html#binary-i-o) might work the same way as an on-disk file.

chance-wnb (commented):
Overall LGTM. Great work!

I will leave the final approval to Zach.

scripts/benchmarks/basic_openai_logging.py
andrewtruong: fixed

scripts/benchmarks/weave_setup_time.py
andrewtruong: fixed

scripts/benchmarks/weave_setup_time.py
andrewtruong: added

andrewtruong: added

scripts/benchmarks/weave_setup_time.py
andrewtruong: added

tests/trace/test_import.py
andrewtruong: fixed

PR #5151, opened 2025-07-23
chore(weave): Reorder pre-commits to have most relevant checks run first

andrewtruong:
This PR:
1. Moves the Fixit command first as it is a codemod and will affect all downstream checks
2. Moves pyright up before mypy because it's faster
3. Moves mypy up before the other checks which are not specifically code related


gtarpenning (approved):
Nice, much better

PR #5152, opened 2025-07-23
fix(weave): Don't overwrite user defined Model predict method

andrewtruong:
- **test**
- **test**

### DIFF
--- a/weave/flow/eval_imperative.py
+++ b/weave/flow/eval_imperative.py
@@ -22,7 +22,7 @@ from pydantic import (
 import weave
 from weave.flow.dataset import Dataset
 from weave.flow.eval import Evaluation, default_evaluation_display_name
-from weave.flow.model import Model
+from weave.flow.model import MissingInferenceMethodError, Model
 from weave.flow.scorer import Scorer
 from weave.flow.scorer import auto_summarize as auto_summarize_fn
 from weave.flow.util import make_memorable_name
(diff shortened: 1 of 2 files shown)


weave/flow/eval_imperative.py
```diff
@@ -389,12 +389,23 @@ def model_post_init(self, __context: Any) -> None:
...
-            # Get the output from the context variable
-            return current_output.get()
+        # Only modify the predict method if the model doesn't already have one
+        try:
+            assert isinstance(self.model, Model)
+            self.model.get_infer_method()
+        except MissingInferenceMethodError:
+            has_infer_method = False
+        else:
+            has_infer_method = True
```
neutralino1:
```suggestion
        try:
            assert isinstance(self.model, Model)
            self.model.get_infer_method()
            has_infer_method = True
        except MissingInferenceMethodError:
            has_infer_method = False
```

weave/flow/eval_imperative.py
```diff
@@ -389,12 +389,23 @@ def model_post_init(self, __context: Any) -> None:
...
+
+        if not has_infer_method:
+
+            @weave.op(name="Model.predict", enable_code_capture=False)
+            def predict(self: Model, inputs: dict) -> Any:
+                # Get the output from the context variable
+                return current_output.get()
 
-        self.model.__dict__["predict"] = MethodType(predict, self.model)
+            self.model.__dict__["predict"] = MethodType(predict, self.model)
```
neutralino1, line 403: Why not do this directly in the `except` block and get rid of `has_infer_method`?

andrewtruong: good point

weave/flow/eval_imperative.py
```diff
@@ -389,12 +389,18 @@ def model_post_init(self, __context: Any) -> None:
         # objects that "look right" to our object saving system.
 
         # --- Setup the model object ---
-        @weave.op(name="Model.predict", enable_code_capture=False)
-        def predict(self: Model, inputs: dict) -> Any:
-            # Get the output from the context variable
-            return current_output.get()
+        # Only modify the predict method if the model doesn't already have one
+        try:
+            assert isinstance(self.model, Model)
```
neutralino1, line 394: I think this is going to raise an `AssertionError` that will not be caught by your `except` clause.

neutralino1 (commented):
One thing I missed.

weave/flow/eval_imperative.py
andrewtruong: It doesn't need to be caught.   At this stage of the code, `self.model` is always `Model`.  This check is just to satisfy the type checker, which doesn't know because of a limitation with pydantic

PR #5153, opened 2025-07-23
docs(weave): Inference service docs followups

J2-D2-3PO:
## Description

Fixes https://wandb.atlassian.net/browse/DOCS-1666

- In models info table, update display names to match internal spreadsheet
- Update some screenshots to show latest model tiles
- Update Huggingface string id list

## Testing

yarn start on local


docs/docs/guides/integrations/inference.md
```diff
@@ -115,6 +115,9 @@ To create a chat completion, you will need:
   - `deepseek-ai/DeepSeek-R1-0528`
   - `meta-llama/Llama-4-Scout-17B-16E-Instruct`
   - `microsoft/Phi-4-mini-instruct`
+  - `moonshotai/Kimi-K2-Instruct`
+  - `Qwen/Qwen3-235B-A22B-Instruct-2507`
+  - `Qwen/Qwen3-Coder-480B-A35B-Instruct`
```
jamie-rasmussen, line 120: We also need to update the Python example below (line 154)

jamie-rasmussen (approved):
I'm making some of these display name changes in the UI as well - would be nice to update the screenshots to show those as well, though personally I think the next model release probably isn't far away and we could wait for that.

docs/docs/guides/integrations/inference.md
J2-D2-3PO: Ah totally missed that, thanks for catching, updated in https://github.com/wandb/weave/pull/5153/commits/503625bdeeb042f5ecfecf8d76ee76aa937ac152

J2-D2-3PO:
> I'm making some of these display name changes in the UI as well - would be nice to update the screenshots to show those as well, though personally I think the next model release probably isn't far away and we could wait for that.

I vote let's wait for that then. Never been a huge fan of screenshots in docs in general unless they add clear value, and the UI for our product changes pretty regularly

PR #5154, opened 2025-07-23
fix(weave): indeterminate ordering in callsquery w/ identical started_at

gtarpenning:
## Description

[WB-26077](https://wandb.atlassian.net/browse/WB-26077)

The solution is to always order by `started_at, id` instead of just `started_at`. 

Actually, we should probably be adding order by id after ALL non-null sorts. The user should not be responsible for figuring that out...

## Testing

- adds `test_trace_call_query_timings` which fails in prod

[WB-26077]: https://wandb.atlassian.net/browse/WB-26077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -471,8 +471,13 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
         if req.sort_by is not None:
             for sort_by in req.sort_by:
                 cq.add_order(sort_by.field, sort_by.direction)
+            # If the only sort field is "started_at", add "id" as secondary sort for consistency
+            if len(req.sort_by) == 1 and req.sort_by[0].field == "started_at":
+                cq.add_order("id", "asc")
         else:
+            # Default sorting: started_at with id as secondary sort for consistency
             cq.add_order("started_at", "asc")
+            cq.add_order("id", "asc")
         if req.limit is not None:
             cq.set_limit(req.limit)
         if req.offset is not None:
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -471,8 +471,10 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
         if req.sort_by is not None:
             for sort_by in req.sort_by:
                 cq.add_order(sort_by.field, sort_by.direction)
-            # If the only sort field is "started_at", add "id" as secondary sort for consistency
-            if len(req.sort_by) == 1 and req.sort_by[0].field == "started_at":
+            # If user isn't already sorting by id, add id as secondary sort for consistency
+            if req.sort_by and not any(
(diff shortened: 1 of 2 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -471,8 +471,13 @@ def calls_query_stream(self, req: tsi.CallsQueryReq) -> Iterator[tsi.CallSchema]
         if req.sort_by is not None:
             for sort_by in req.sort_by:
                 cq.add_order(sort_by.field, sort_by.direction)
+            # If the only sort field is "started_at", add "id" as secondary sort for consistency
+            if len(req.sort_by) == 1 and req.sort_by[0].field == "started_at":
```
zamborg: I know our issue was duplicate `started_at` times but this issue will occur if the user provides a filter that *also* doesn't have a tiebreaker for duplicates. EG: `completed_at`. 

Should we just *always* append `id` as a tiebreaker to prevent page-duplicates?

gtarpenning: literally just added that to the description and pushing

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -471,8 +471,13 @@ def calls_query_stream(self, req: tsi.CallsQueryReq) -> Iterator[tsi.CallSchema]
...
+                cq.add_order("id", "asc")
```
bcsherma: mightn't this be an issue for `ended_at` as well?

bcsherma (approved):
One comment but lgtm, iiuc there is no/minimal perf overhead because we only use the second order key in the case of a collisions in the first order key. And in that hopefully rare case we just tiebreak by id real fast.

weave/trace_server/clickhouse_trace_server_batched.py
zamborg: @gtarpenning resolving see my comment [here](https://github.com/wandb/weave/pull/5154#discussion_r2226280795)

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -471,8 +471,10 @@ def calls_query_stream(self, req: tsi.CallsQueryReq) -> Iterator[tsi.CallSchema]
         if req.sort_by is not None:
             for sort_by in req.sort_by:
                 cq.add_order(sort_by.field, sort_by.direction)
-            # If the only sort field is "started_at", add "id" as secondary sort for consistency
-            if len(req.sort_by) == 1 and req.sort_by[0].field == "started_at":
+            # If user isn't already sorting by id, add id as secondary sort for consistency
+            if req.sort_by and not any(
```
zamborg: you don't need the `if req.sort_by` for this section since you're guaranteed its `not None` inside of this if statement. 

But also it makes the code look nicer so I don't mind.

PR #5156, opened 2025-07-23
chore(weave): threads query builder style cleanup

gtarpenning:
## Description

Tiny style nits on the threads query builder: 
- use capitals for SQL terms like `AS` and `ON`
- move comments from the query itself, these get logged to dd and imo actually make it harder to understand the query.

## Testing

updates tests


neutralino1 (approved):

chance-wnb (approved):
Beautiful! Thanks!

PR #5163, opened 2025-07-24
feat(weave): performant call descendants query in db and client

gtarpenning:
## Description

This pr:

creates a new clickhouse database function that queries for descendents using a recursive CTE query
exposes endpoint on the client so that users can ask for descendents
Usage:

```
client = weave.init("project")

calls = client.get_calls()
first_root_call = calls[0]
first_root_call_descendents = client.get_call_descendants(first_root_call)
for descendant in first_root_call_descendents:
   print(descendant.display_name)

# Or, in batch
all_root_descendants = client.get_call_descendants(call_ids=[c.id for c in calls])

# Or with the class method
first_root_depth_2 = first_root_call.descendants(depth=2) # only children + grandchildren
```

## Testing

adds tests

### DIFF
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -3597,6 +3597,50 @@ def test_sum_dict_leaves_mixed_types(client):
     }
 
 
+def test_call_descendants(client):
+    @weave.op
+    def parent_op(x):
+        child_op(x + 1)
+        return x * 2
+
+    @weave.op
+    def child_op(x):
+        grandchild_op(x + 1)
+        return x * 3
+
+    @weave.op
+    def grandchild_op(x):
+        return x * 4
+
+    # Run the parent operation
+    result = parent_op(5)
+
+    # Get all calls
+    calls = list(client.get_calls())
+    parent_call = None
+    for call in calls:
+        if call.op_name == "parent_op":
+            parent_call = call
+            break
+
+    assert parent_call is not None
+
+    # Test the calls_descendants functionality
+    descendants = list(
+        client.server.calls_descendants(
+            tsi.CallsDescendantsReq(
+                project_id=client._project_id(), call_id=parent_call.id
+            )
+        )
+    )
+
+    # Should have the child and grandchild calls
+    descendant_op_names = [call.op_name for call in descendants]
+    assert "child_op" in descendant_op_names
+    assert "grandchild_op" in descendant_op_names
+    assert len(descendants) == 2
+
+
 def test_sum_dict_leaves_deep_nested(client):
     """Test that sum_dict_leaves correctly handles deeply nested dictionaries (2 levels) with mixed types."""
     dicts = [
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -603,6 +603,113 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
                         set_nested_key(calls[i], col, val)
 
     @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched.calls_delete")
+    def call_descendants(self, req: tsi.CallDescendantsReq) -> tsi.CallDescendantsRes:
+        """
+        Get all descendants of the specified parent calls using a recursive CTE query.
+
+        Args:
+            req: Request containing parent_call_id or parent_call_ids, along with optional
+                 filters like limit, depth, and column specifications.
+
+        Returns:
+            CallDescendantsRes containing the list of descendant calls.
+        """
+        # Validate input - must have either parent_call_id or parent_call_ids
+        if not req.parent_call_id and not req.parent_call_ids:
+            raise ValueError("Must specify either parent_call_id or parent_call_ids")
+
+        if req.parent_call_id and req.parent_call_ids:
+            raise ValueError("Cannot specify both parent_call_id and parent_call_ids")
+
+        parent_ids = (
+            [req.parent_call_id] if req.parent_call_id else (req.parent_call_ids or [])
+        )
+        if not parent_ids:
+            return tsi.CallDescendantsRes(calls=[])
+
+        pb = ParamBuilder()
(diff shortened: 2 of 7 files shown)


weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1613,6 +1661,60 @@ def optimized_project_contains_call_query(
...
+    # Prevent infinite recursion by limiting depth
+    recursive_case_sql = f"""
+    SELECT c.id, d.depth + 1 AS depth
+    FROM calls_merged c
+    INNER JOIN {cte_name} d ON c.parent_id = d.id
+    WHERE c.project_id = {param_slot(project_id_param, "String")}
+        AND d.depth < {param_slot(max_depth_param, "UInt64")}
+    """
+
+    descendant_sql = f"""RECURSIVE {cte_name} AS (
```
gtarpenning: this is the recursion

tests/trace/test_weave_client.py
```diff
@@ -3597,6 +3598,168 @@ def test_sum_dict_leaves_mixed_types(client):
...
+    # should return same as depth=2 since we only have 2 levels of descendants
+    descendants_depth_3 = client.get_calls_descendants(
+        parent_call_ids=[parent_call.id], depth=10
+    )
+    assert len(descendants_depth_3) == 2
+    assert op_name_from_ref(descendants_depth_3[0].op_name) == "child_op"
+    assert op_name_from_ref(descendants_depth_3[1].op_name) == "grandchild_op"
+
+
+def test_calls_descendants_complex(client):
```
gtarpenning, line 3691: add test for passing multiple descendant ids

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -602,7 +574,99 @@ def _expand_call_refs(
                             val["_ref"] = ref.uri()
                         set_nested_key(calls[i], col, val)
 
-    @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched.calls_delete")
+    @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched.calls_descendants")
+    def calls_descendants(
```
neutralino1, line 578: Should this be a new endpoint or a new way to query the existing `/calls/stream_query` endpoint. For example with a `ancestor_ids` argument?

I presume this would de-dup some of the logic below?

neutralino1 (changes requested):
Has this been urgently requested by users/customers?

I wonder if we should do this "the right way", by starting to persist lists of ancestors on each calls (the accumulation can be done at the SDK level), and then amending the regular call listing endpoint to support an `ancestor_ids` argument and use the call's ancestor list.

Yes, this means the feature would only work for new calls (hence my asking about urgency). `depth` may also be tricky to support by maybe that's ok.

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: could totally be rolled into that, this is how I thought about the current structure:

Pros: 
- doesn't increase complexity on our single most important endpoint, which is already overloaded imo.
- easier to change in the future if we want to go a different route 
- we don't have to support stuff like storage or other flags that might be impossibly hard to compute but are on the streaming endpoint.

Cons:
- some duplicate code
- another endpoint, which is specialized and not very flexible

gtarpenning:
> Has this been urgently requested by users/customers?
> 
> I wonder if we should do this "the right way", by starting to persist lists of ancestors on each calls (the accumulation can be done at the SDK level), and then amending the regular call listing endpoint to support an `ancestor_ids` argument and use the call's ancestor list.
> 
> Yes, this means the feature would only work for new calls (hence my asking about urgency). `depth` may also be tricky to support by maybe that's ok.

@neutralino1 Great point, I thought about this and I think that even if end up doing "the right way", this path will end up being useful for calls that don't have the hierarchy of ids. Then, all we have to do is check to see if a call has that field, and we can use the more performant lookup, otherwise use this method. So while this might be the primary method now, it could become the backup method for older calls. 

I will say, however, that because we are doing only ID lookups this query is surprisingly performant. I think for 99% of users this will absolutely be sufficient forever. It's also >10x faster than grabbing all the calls and doing filtering in memory, and >100x faster than iteratively calling `call.children()` in the client. If the "right way" is entirely about performance, I can also benchmark against some larger projects to get a feel for it.

neutralino1:
@gtarpenning 
> calls that don't have the hierarchy of ids

Can you give example of such cases?

> If the "right way" is entirely about performance

It is indeed about that, but also about keeping complexity and tech debt under control, plus the satisfaction of doing things correctly. If a candidate writes an O(N) algorithm when there exists an O(1) solution, we typically turn them away. Yet, in our own code base we are ok with smells such as recursive CTEs.

neutralino1:
@gtarpenning actually, is it possible that the "materialized path" option would be slower than recursive CTE because querying on IDs is faster than on a new `ancestor_ids` array column?

gtarpenning:
@neutralino1 
 
> calls that don't have the hierarchy of ids

I meant that in reference to the other solution, so to answer your question: all calls currently. 

>  If a candidate writes an O(N) algorithm when there exists an O(1) solution, we typically turn them away. Yet, in our own code base we are ok with smells such as recursive CTEs.

Haha fair. This was mostly a Friday afternoon experiment to revive that old branch. FWIW I do think there's another way of looking at this though. You say that this is has added complexity (which I agree with), but I would argue so does the alternative. That requires a migration and some way of playing nice with new client versions that try to make use of the new query path (we would have to error, which is sort of frustrating for users with evals they have already run). Changing the write path is easy in the server, but in the client we would need to make sure we collect all the ids for many generations of children, across threads and processes and i'm not 100% sure thats trivial... Would we return the ancestor Ids to the user? We currently return everything on a call to the user, but this feels a bit like an internal accounting, or index, and might be confusing.

Anyway, the only reason i'm pushing back slightly is because I don't think this is ever going to be a primary query path, merely a convenience, so I'm a bit more tolerant of db hacks.

PR #5164, opened 2025-07-24
chore(weave): Update strip exif script to use uv

andrewtruong:
Converts the script into a self-contained `uv` script


neutralino1 (approved):
@neutralino1 approved this PR from Slack with [Graphite](https://app.graphite.dev/github/pr/wandb/weave/5164)

PR #5171, opened 2025-07-25
feat(weave): add custom eval attributes to EvaluationLogger

DanielPolatajko:
## Description

This PR adds the ability to customise the attributes of an imperative EvaluationLogger. This is useful in order to write metadata columns to the Evaluations tab in the UI when using the EvaluationLogger.

## Testing

Added a unit test - I have also tried out previously in a test environment on my own [project](https://github.com/DanielPolatajko/inspect_weave/blob/add-metadata/inspect_weave/custom_evaluation_logger.py)


DanielPolatajko:

tests/flow/test_evaluation_imperative.py
```diff
@@ -580,3 +580,15 @@ def invoke(self, query: dict, **options) -> str:
...
+def test_evaluation_logger_with_custom_attributes(client):
+    ev = weave.EvaluationLogger(eval_attributes={"custom_attribute": "value"})
+    pred = ev.log_prediction(inputs={"a": 1, "b": 2}, output=3)
+    ev.log_summary(auto_summarize=False)
+    ev.finish()
+    client.flush()
+
+    calls = client.get_calls()
+    for call in calls:
+        assert call.attributes["custom_attribute"] == "value"
```
andrewtruong: Is this what we should expect?  Why do you want to apply a custom attribute to all calls instead of just the root call?

DanielPolatajko: For my use case, I probably would want to have all (maybe most) of the same attributes on the eval and its child calls, mainly because it makes filtering traces at different levels a bit easier in the UI

But I can see that the approach here is potentially limiting flexibility - I will reduce it to just the root call as you suggest

andrewtruong:
This was merged in the PR above!

PR #5179, opened 2025-07-28
chore(weave): Add client methods to get evaluations and scores

andrewtruong:
https://wandb.atlassian.net/browse/WB-26478

Adds client methods for getting evaluations and scores.

Currently only works for imperative evals!


weave/trace/weave_client.py
```diff
@@ -927,6 +921,44 @@ def get(self, ref: ObjectRef, *, objectify: bool = True) -> Any:
...
+                    }
+                }
+            }
+        )
+
+    def get_scores(self) -> CallsIter:
+        """Gets all score calls in this project.
+
+        Currently only works for imperative evals"""
+        score_json_path = "attributes._weave_eval_meta.score"
```
gtarpenning: we cool with this magic string living here?

weave/trace/weave_client.py
```diff
@@ -927,6 +921,44 @@ def get(self, ref: ObjectRef, *, objectify: bool = True) -> Any:
...
+                            "$convert": {
+                                "input": {"$getField": score_json_path},
+                                "to": "string",
+                            },
+                        },
+                        {"$literal": "true"},
+                    ]
+                }
+            },
+            expand_columns=[score_json_path],
```
gtarpenning: I think this won't work, the ref is `_weave_eval_meta` right? so this path needs to be `attributes._weave_eval_meta`

tests/trace/test_weave_client.py
```diff
@@ -3624,3 +3625,32 @@ def test_sum_dict_leaves_deep_nested(client):
...
+    assert len(evals) == 1
+    ev2 = evals[0]
+    assert ev2.output["score"]["mean"] == 3
+    assert ev2.output["score2"]["mean"] == 4
+    assert ev2.output["output"]["y"] == 5
+    assert ev2.inputs["self"].name == "def-evaluation"
+    assert ev2.inputs["model"].name == "abc"
+
+
+@pytest.mark.skipif(os.getenv("CI"), reason="fails in CI for some reason")
```
gtarpenning: this shouldn't fail in ci

tests/integrations/pandas-test/test_evals_to_pandas.py
```diff
@@ -0,0 +1,28 @@
...
+    assert evals_df.loc[0, "output.score.mean"] == 3
+    assert evals_df.loc[0, "output.score2.mean"] == 4
+    assert evals_df.loc[0, "output.output.y"] == 5
+    assert evals_df.loc[0, "inputs.self"].name == "def-evaluation"
+    assert evals_df.loc[0, "inputs.model"].name == "abc"
+
+
+def test_get_scores_to_pandas(client, make_evals):
+    if client_is_sqlite(client):
+        return pytest.skip("skipping for sqlite")
```
neutralino1: Why skip here but not in `test_get_evaluations_to_pandas`?

weave/trace/weave_client.py
andrewtruong: This does seem to work for me

tests/trace/test_weave_client.py
neutralino1: Also, if something fails, we should not skip, we should investigate and fix.

andrewtruong: I get this error.  Can you help me understand why?
https://github.com/wandb/weave/actions/runs/16598048466/job/46949676118

weave/trace/weave_client.py
andrewtruong: What's the alternative?

weave/trace/weave_client.py
gtarpenning: In what project are you testing this? its possible the filter is just being ignored and its returning everything

tests/trace/test_weave_client.py
gtarpenning: oh this is actually very interesting. i think the error is occuring during the client unraveling step, which might be getting messed up by the addition of the expand columns. Can you try passing `return_expanded_column_values=False`?

gtarpenning: it will need to make its way into the `client.server.calls_query_stream`

weave/trace/weave_client.py
gtarpenning: fair

weave/trace/weave_client.py
```diff
@@ -927,6 +921,44 @@ def get(self, ref: ObjectRef, *, objectify: bool = True) -> Any:
 
     ################ Query API ################
 
+    def get_evaluations(self) -> CallsIter:
```
neutralino1: Should this be called `get_evaluations_calls`? And `get_evaluations` should return the `Evaluation` objects?

weave/trace/weave_client.py
```diff
@@ -927,6 +921,44 @@ def get(self, ref: ObjectRef, *, objectify: bool = True) -> Any:
...
+        """Gets all root Evaluation.evaluate calls in this project."""
```
neutralino1: Needs more detailed docstring since this is a public API. It should describe the output type and provide a code example of how to use this API.

weave/trace/weave_client.py
```diff
@@ -927,6 +921,44 @@ def get(self, ref: ObjectRef, *, objectify: bool = True) -> Any:
...
+                "$expr": {
+                    "$contains": {
+                        "input": {"$getField": "op_name"},
+                        "substr": {"$literal": evaluate_op_name},
+                    }
+                }
+            }
+        )
+
+    def get_scores(self) -> CallsIter:
```
neutralino1: I would expect `get_scores` to return the score payloads themselves (i.e. the calls output), or the function to be calls `get_score_calls`.

Also should this either take an evaluation object as input, or be defined on the `Evaluation` class? Does this fetch ALL score calls for ALL evals?

weave/trace/weave_client.py
neutralino1: Actually `get_evaluation_calls` without parametrization does not make a lot of sense. Users likely want calls for a specific evaluation. Should this be defined on the `Evaluation` object or expect an `evaluation_id` or sorts?

tests/trace/test_weave_client.py
andrewtruong: @gtarpenning am I doing this right?  It seems to have the same issue: https://github.com/wandb/weave/actions/runs/16603013893/job/46967545552?pr=5179

gtarpenning: hmm, okay that wasn't it. you can't reproduce this locally? My intuition is then that there is something wrong with the conftest eval construction, where in CI where the tests are run in parallel and the env might be dirty...

andrewtruong: hmm, it's possible the env is dirty and we're hiding it somehow... but identifying and fixing that is probably outside the scope of this PR

weave/trace/weave_client.py
```diff
@@ -927,6 +925,107 @@ def get(self, ref: ObjectRef, *, objectify: bool = True) -> Any:
...
+                            "$convert": {
+                                "input": {"$getField": score_json_path},
+                                "to": "string",
+                            },
+                        },
+                        {"$literal": "true"},
+                    ]
+                }
+            },
+            "expand_columns": [score_json_path],
```
gtarpenning: this is unnecessary `attributes._weave_eval_meta.score` is accessible directly, without ref resolution

weave/flow/eval.py
```diff
@@ -268,6 +280,121 @@ async def evaluate(self, model: Union[Op, Model]) -> dict:
...
+            ValueError: If the evaluation has no ref (hasn't been saved/run yet).
+
+        Examples:
+            >>> evaluation = Evaluation(dataset=examples, scorers=[scorer])
+            >>> await evaluation.evaluate(model)  # Run evaluation first
+            >>> calls = evaluation.get_evaluate_calls()
+            >>> for call in calls:
+            ...     print(f"Evaluation run: {call.id} at {call.started_at}")
+        """
+        evaluate_op_name = "Evaluation.evaluate"
```
gtarpenning: will this work with calls logged through the TS sdk?

weave/flow/eval.py
```diff
@@ -268,6 +280,121 @@ async def evaluate(self, model: Union[Op, Model]) -> dict:
...
+        return client.get_calls(
+            query={
+                "$expr": {
+                    "$contains": {
+                        "input": {"$getField": "op_name"},
+                        "substr": {"$literal": evaluate_op_name},
+                    }
+                }
+            },
+            filter=CallsFilter(input_refs=[self.ref.uri()]),
```
gtarpenning: I understand the query portion of this as: get me the evaluate call, but then the filter is "get me the calls that consume the eval as input". I would sort of expect those to be mutually exclusive? Clearly thats wrong, so maybe a comment?

weave/flow/eval.py
```diff
@@ -268,6 +280,121 @@ async def evaluate(self, model: Union[Op, Model]) -> dict:
 
         return summary
 
+    def get_evaluate_calls(self) -> CallsIter:
```
gtarpenning, line 283: this name is a bit confusing, in my mind its more like: `get_all_root_evaluation_calls` thats probably too verbose, but this caused me to misunderstand the get_scores function...

weave/flow/eval.py
```diff
@@ -268,6 +280,121 @@ async def evaluate(self, model: Union[Op, Model]) -> dict:
...
+            >>> score_calls = evaluation.get_score_calls()
+            >>> for trace_id, calls in score_calls.items():
+            ...     print(f"Trace {trace_id}: {len(calls)} scorer calls")
+            ...     for call in calls:
+            ...         scorer_name = call.summary.get("weave", {}).get("trace_name")
+            ...         print(f"  Scorer: {scorer_name}, Output: {call.output}")
+        """
+        d = {}
+        client = require_weave_client()
+        for call in self.get_evaluate_calls():
```
gtarpenning: maybe `for eval_call in ...`

weave/flow/eval.py
```diff
@@ -268,6 +280,121 @@ async def evaluate(self, model: Union[Op, Model]) -> dict:
...
+            descendents = list(client.get_calls(filter={"trace_ids": [call.trace_id]}))
+            summary_call = list(descendents)[-1]
+            scorer_names = {
+                k for k in summary_call.output if k not in ["output", "model_latency"]
+            }
+            score_calls = [
+                call
+                for call in descendents
+                if call.summary.get("weave", {}).get("trace_name") in scorer_names
```
gtarpenning, line 350: before we use this, can we check that we can't just use the `op_name` or `display_name`? its possible we can't, in which case maybe add a TODO for a cheeky helper, this looks like a hack

weave/flow/eval.py
```diff
@@ -268,6 +280,121 @@ async def evaluate(self, model: Union[Op, Model]) -> dict:
...
+                    "$contains": {
+                        "input": {"$getField": "op_name"},
+                        "substr": {"$literal": evaluate_op_name},
+                    }
+                }
+            },
+            filter=CallsFilter(input_refs=[self.ref.uri()]),
+        )
+
+    def get_score_calls(self) -> dict[str, list[Call]]:
```
gtarpenning, line 318: this is a bit surprising that calling this one method gets **all** scores for **all** evaluations. Not sure when I would ever call this. I think I would more frequently use the get all scores for a **given** evaluation (pass in a call or call_id?).

gtarpenning: I guess my confusion stems from the differentiation between an `Evaluation` object and an `evaluation.Evaluate` _run_...

andrewtruong: I thought about this, but I don't think there's an easy way for a user to know the exact trace_id of the a given evaluation.

weave/flow/eval.py
```diff
@@ -268,6 +280,121 @@ async def evaluate(self, model: Union[Op, Model]) -> dict:
...
+        if not self.ref:
+            raise ValueError("Evaluation has no ref, please run the evaluation first!")
+
+        client = require_weave_client()
+        return client.get_calls(
+            query={
+                "$expr": {
+                    "$contains": {
+                        "input": {"$getField": "op_name"},
+                        "substr": {"$literal": evaluate_op_name},
```
gtarpenning: consider using:
```
eval_op_ref = f"weave:///{client._project_id()}/op/{evaluate_op_name}:*"
query={"$expr": {"$eq": [{"$getField": "op_name"}, {"$literal": eval_op_ref}]}}
```

gtarpenning: should be more performant and a bit tighter

weave/flow/eval.py
andrewtruong: What did you think it meant?

weave/flow/eval.py
gtarpenning: wait actually, this is what we want:

```
eval_op_ref = f"weave:///{client._project_id()}/op/{evaluate_op_name}:*"
client.get_calls(filter={"op_names": [eval_op_ref]})
```

andrewtruong: Why is this better?

weave/flow/eval.py
gtarpenning: The idea of an `Evaluation` object itself was sort of confusing, where you have multiple runs of the same object, but producing different evals with different trace_ids. The imperative "evaluate" in the name suggests action, "get" is an action, "evaluate" is an action, idk...

gtarpenning: its more like `get_evaluation_runs`

gtarpenning: but obviously thats not quite right

weave/flow/eval.py
gtarpenning: Its both more performant in the db and is easier to read as a user

gtarpenning: specifically, we do a special pre-group by filter for hardcoded conditions (in the filter as opposed to query)

weave/flow/eval.py
gtarpenning: well, we might want to just make it optional, if they happen to have it from a previous query?

gtarpenning: I find myself doing a get all root traces query much of the time, that would end up in a list of evals with trace_ids, then i might want scores for the last 5 of them or whatever

neutralino1: Maybe this should expect a run as input, i.e. an `evaluate` call.

Longer term, should we add a `Run` class, that would essentially wrap around `Evaluation.evaluate` calls? Then we could define APIs such as `evaluation.get_runs()` and `run.get_scores()`.

Could look like
```python
class Run:
    id: str # root_call.trace_id
    root_call: Call # The Evaluation.evaluate call
```

weave/flow/eval.py
neutralino1: I think `get_evaluate_calls` is the best option here. Yes, `evaluate` calls represent runs, but since we don't have a concept of an evaluation run, this is the closest correct option.

weave/trace/weave_client.py
```diff
@@ -800,6 +801,71 @@ def get(self, ref: ObjectRef, *, objectify: bool = True) -> Any:
 
     ################ Query API ################
 
+    def get_evaluation(self, uri: str) -> Evaluation:
```
neutralino1, line 806: I wonder if this API is useful since users can readily do `weave.get(uri)`...

weave/flow/eval.py
andrewtruong: Yeah our nomenclature here needs some work.  Personally I think `Evaluation` should really be called `Task`, and `Evaluation` is this new "run" concept

weave/trace/weave_client.py
andrewtruong: I think it's a nice convenience method.  It will feel more complete when the client has equivalent `get_dataset`, `get_{thing}` methods

weave/flow/eval.py
andrewtruong: I'll make an issue to note the nomenclature change

weave/flow/casting.py
```diff
@@ -32,6 +33,8 @@ def cast_to_scorer(obj: Any) -> Scorer | Op:
     res: Scorer | Op
     if isinstance(obj, Scorer):
         res = obj
+    elif weave_isinstance(obj, Scorer):
+        res = Scorer.from_obj(obj)
```
gtarpenning, line 37: i think this doesn't work for classes? we instantiate the base class but not any extra scoring bits, so you can't actually re-score using this

weave/flow/eval.py
```diff
@@ -268,6 +280,119 @@ async def evaluate(self, model: Union[Op, Model]) -> dict:
...
+            >>> # }
+        """
+        score_calls = self.get_score_calls()
+        d: dict[str, dict[str, list[Any]]] = {}
+        for trace_id, calls in score_calls.items():
+            d[trace_id] = {}
+            for call in calls:
+                if call.summary is None:
+                    continue
+                scorer_name = call.summary.get("weave", {}).get("trace_name")
```
gtarpenning, line 390: todo: we should add a helper on the `Call` class for this

gtarpenning (approved):

PR #5182, opened 2025-07-28
docs(weave): Rework thread handling in OpenAI and Anthropic API calls; Add session resumption example

chance-wnb:
## Description

1. The previous example of OpenAI and Anthropic provider example uses  `weave.thread()` context manager in the leaf weave.op functions, but when there is no weave ops underneath a thread, nothing will be grouped. This doc fixe lifts the location of `weave.thread()` usage 1 level up. This ensures thread context is established at the appropriate boundary.

2. Adds a new section on resuming previous sessions with examples of how to properly handle thread IDs when implementing optional thread resumption. The new documentation emphasizes the importance of always providing a valid thread ID rather than using `None`.

### DIFF
--- a/docs/docs/guides/tracking/threads.md
+++ b/docs/docs/guides/tracking/threads.md
@@ -353,12 +353,12 @@ When no `thread_id` is specified, Weave's internal implementation automatically 
 
 ```python
 import weave
-import uuid
+import uuidv7
 import argparse
 
 def generate_id():
     """Generate a unique thread ID using UUID v7."""
-    return str(uuid.uuid4())
+    return str(uuidv7.uuidv7())
 
 @weave.op
 def load_history(session_id):
--- a/docs/docs/guides/tracking/threads.md
+++ b/docs/docs/guides/tracking/threads.md
@@ -232,18 +232,17 @@ class OpenAIProvider:
         # Input validation, routing logic, basic preprocessing
         print(f"  L1: Routing to OpenAI for: {user_input}")
 
-        # Call Level 2 directly - this creates the call chain depth
-        return self.execute_openai_call(user_input, thread_id)
-
-    @weave.op
-    def execute_openai_call(self, user_input: str, thread_id: str) -> str:
-        """Level 2: TURN BOUNDARY - Execute OpenAI API call"""
-        print(f"    L2: Executing OpenAI API call")
-
         # This is the turn boundary - wrap with thread context
         with weave.thread(thread_id):
-            response = f"OpenAI GPT-4 response: {user_input}"
-            return response
+            # Call Level 2 directly - this creates the call chain depth
+            return self.execute_openai_call(user_input)
+
+    @weave.op
+    def execute_openai_call(self, user_input: str) -> str:
+        """Level 2: TURN BOUNDARY - Execute OpenAI API call"""
+        print(f"    L2: Executing OpenAI API call")
+        response = f"OpenAI GPT-4 response: {user_input}"
+        return response
 
 
 class AnthropicProvider:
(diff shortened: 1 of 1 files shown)


docs/docs/guides/tracking/threads.md
```diff
@@ -343,6 +341,53 @@ if __name__ == "__main__":
 # - Supporting operations in call chain tracked as nested calls, not turns
 ```
 
+### Resume a previous session
+
+Sometimes you need to resume a previously started session and continue adding calls to the same thread. In other cases, you may not be able to resume an existing session and must start a new thread instead.
+
+When implementing optional thread resumption, **never** leave the `thread_id` parameter as `None`, as this will disable thread grouping entirely. Instead, always provide a valid thread ID. If you need to create a new thread, generate a unique identifier using a function like `generate_id()`.
```
neutralino1, line 348:
```suggestion
To enable thread resumption, pass the thread ID of the existing thread to `weave.thread(...)`. If no thread ID is provided, Weave will automatically generate a new one using UUIDv7.
```

docs/docs/guides/tracking/threads.md
```diff
@@ -343,6 +341,53 @@ if __name__ == "__main__":
...
 
+### Resume a previous session
+
+Sometimes you need to resume a previously started session and continue adding calls to the same thread. In other cases, you may not be able to resume an existing session and must start a new thread instead.
+
+When implementing optional thread resumption, **never** leave the `thread_id` parameter as `None`, as this will disable thread grouping entirely. Instead, always provide a valid thread ID. If you need to create a new thread, generate a unique identifier using a function like `generate_id()`. 
+
+:::info
+When no `thread_id` is specified, Weave's internal implementation automatically generates a random UUID v7. You can replicate this behavior in your own `generate_id()` function or use any unique string value you prefer.
+:::
```
neutralino1, line 352: Unnecessary if above suggestion is applied.

docs/docs/guides/tracking/threads.md
```diff
@@ -343,6 +341,53 @@ if __name__ == "__main__":
...
+
+```python
+import weave
+import uuid
+import argparse
+
+def generate_id():
+    """Generate a unique thread ID using UUID v7."""
+    return str(uuid.uuid4())
```
neutralino1:
```suggestion
    return str(uuid.uuid7())
```

neutralino1 (commented):
A couple of comments to streamline the docs.

docs/docs/guides/tracking/threads.md
chance-wnb: Thanks for the feedback, unfortunately that misses the point of this doc change. From what Scott reported:
https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1753440962441659

He doesn't want to do this:

```
if thread_id:
    thread_ctx_mgr = weave.thread(thread_id)
else:
    thread_ctx_mgr = weave.thread()
    
with thread_ctx_mgr as thread_ctx:
```
in that sometimes we provision a parameter, sometimes not, it is kind awkward. 

So to keep using `weave.thread(thread_id)` uniformly, I proposed him a way to transform the value of thread_id instead.

neutralino1 (approved):

chance-wnb:
* **#5182** <a href="https://app.graphite.dev/github/pr/wandb/weave/5182?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/5182?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

docs/docs/guides/tracking/threads.md
chance-wnb: uuid doesn't have the uuid7, but `uuidv7` package has. Updated document accordingly.

PR #5187, opened 2025-07-29
feat(weave): allow omitting coreweave/ prefix on hosted model strings

jamie-rasmussen:
## Description

Updates our script that creates the `models_providers.json` file to also pull in information about CoreWeave hosted models.
This gives us a way to know which model strings should result in delegation to the Inference Service without having to prefix the model strings with `coreweave/`

The `gemini/` changes are unrelated update from re-running `make update_model_providers`

## Testing

How was this PR tested?

### DIFF
--- a/weave/trace_server/model_providers/model_providers.py
+++ b/weave/trace_server/model_providers/model_providers.py
@@ -1,7 +1,6 @@
 import json
 import logging
 import os
-from pathlib import Path
 from typing import TypedDict
 
 import requests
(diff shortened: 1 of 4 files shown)


weave/trace_server/model_providers/model_providers.py
```diff
@@ -1,20 +1,31 @@
...
 
 import requests
 
 model_providers_url = "https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json"
 MODEL_PROVIDER_INFO_FILE = "model_providers.json"
 
+HOSTED_MODEL_INFO_DIR = (
+    Path(__file__)
+    .parent.joinpath(
+        "../../../../../../frontends/weave/src/components/PagePanelComponents/Home/Browse3/inference"
```
neutralino1: Should we create a local symlink?

jwlee64: idk if i like that we are pulling from a frontend weave relative path so far away, seems like something easy to break since files are not collocated, but i can also understand wanting a single source of truth

jwlee64 (approved):
small comment, but lgtm

weave/trace_server/model_providers/model_providers.py
jamie-rasmussen: I know, I don't like it either, planning to file a ticket to track cleaning this up a bit.

PR #5191, opened 2025-07-30
fix(weave): update langchain google-genai usage extraction logic

parambharat:
## Description

- Fixes [WB-26251](https://wandb.atlassian.net/browse/WB-26251)

## Testing

Added a new helpers module and unit tests for the module.. All existing tests pass. 

[WB-26251]: https://wandb.atlassian.net/browse/WB-26251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chance-wnb (commented):
Hi @parambharat 

I love to see this change. I believe the logic is correct. 

However, for future-proof purpose, can I recommend a way to universally look for usage data among all keys? It seems regardless of the response format, the key is always called `usage_metadata`. so we can:

1. Flatten the dict,
2. See which keys contain keyword `usage_metadata`
3. Does the value of that key have our desired shape? {"input_tokens"/"prompt_tokens": ..., "output_tokens"/"completion_tokens":.., "total_tokens": ..}

If all criteria match, then that is the usage field. Therefore if google's format changes again, we are more resilient. And we also have good backward compatibility.

For reference, I recently implemented something similar in https://github.com/wandb/core/blob/master/frontends/weave/src/components/PagePanelComponents/Home/Browse3/pages/ThreadsPage/common.ts#L48

Let me know if you want to go that way. If not, I am happy to approve this PR as well.

parambharat:
Hey @chance-wnb, thank you for the feedback and review comments. I've refactored the code to a new helpers module to handle usage more consistently. Please take a look and let me know your thoughts.

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+
+        _extract_usage_data(call, output)
+
+        assert call.summary is not None
+        assert "usage" in call.summary
+        usage = call.summary["usage"]
+        assert "gpt-4o-mini" in usage
+        assert usage["gpt-4o-mini"]["prompt_tokens"] == 10
+        assert usage["gpt-4o-mini"]["completion_tokens"] == 5
+        assert usage["gpt-4o-mini"]["total_tokens"] == 15
```
andrewtruong: better to assert on the whole dict.  You will get a nicer error in pytest

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+
+        _extract_usage_data(call, output)
+
+        assert call.summary is not None
+        assert "usage" in call.summary
+        usage = call.summary["usage"]
+        assert "gemini-1.5-pro" in usage
+        assert usage["gemini-1.5-pro"]["prompt_tokens"] == 20
+        assert usage["gemini-1.5-pro"]["completion_tokens"] == 7
+        assert usage["gemini-1.5-pro"]["total_tokens"] == 27
```
andrewtruong: same here and below

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+
+        assert call.summary is not None
+        assert "usage" in call.summary
+        usage = call.summary["usage"]
+        assert "claude-opus-4-20250514" in usage
+        assert usage["claude-opus-4-20250514"]["prompt_tokens"] == 25
+        assert usage["claude-opus-4-20250514"]["completion_tokens"] == 11
+        assert usage["claude-opus-4-20250514"]["total_tokens"] == 36
+
+    def test_batch_operations_aggregation(self):
```
andrewtruong: Cool you have this -- do we support the batching endpoints for our other first party integrations?  If not, we should

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+        assert call.summary is not None
+        assert "usage" in call.summary
+        usage = call.summary["usage"]
+        assert "gpt-4o-mini" in usage
+        # Should aggregate both requests
+        assert usage["gpt-4o-mini"]["prompt_tokens"] == 22  # 10 + 12
+        assert usage["gpt-4o-mini"]["completion_tokens"] == 13  # 5 + 8
+        assert usage["gpt-4o-mini"]["total_tokens"] == 35  # 15 + 20
+
+    def test_tuple_outputs_normalized(self):
```
andrewtruong: When would this happen?

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+            "outputs.0.generations.0.0.message.kwargs.usage_metadata.output_tokens": 7,
+            "outputs.0.generations.0.0.message.kwargs.usage_metadata.total_tokens": 27,
+        }
+        base_path = "outputs.0.generations.0.0.message.kwargs.usage_metadata"
+
+        result = _reconstruct_usage_data_from_flattened(flattened, base_path)
+
+        assert result == {"input_tokens": 20, "output_tokens": 7, "total_tokens": 27}
+
+    def test_ignores_nested_fields(self):
```
andrewtruong: when would this happen?

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+            "outputs.0.generations.0.0.message.kwargs.token_usage.nested.field": "ignored",
+            "outputs.0.generations.0.0.message.kwargs.token_usage.completion_tokens": 5,
+        }
+        base_path = "outputs.0.generations.0.0.message.kwargs.token_usage"
+
+        result = _reconstruct_usage_data_from_flattened(flattened, base_path)
+
+        assert result == {"prompt_tokens": 10, "completion_tokens": 5}
+
+    def test_converts_float_to_int(self):
```
andrewtruong: when would this ever be float?

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+            "outputs.0.generations.0.0.message.kwargs.token_usage.prompt_tokens": 10.0,
+            "outputs.0.generations.0.0.message.kwargs.token_usage.completion_tokens": 5.5,
+        }
+        base_path = "outputs.0.generations.0.0.message.kwargs.token_usage"
+
+        result = _reconstruct_usage_data_from_flattened(flattened, base_path)
+
+        assert result == {"prompt_tokens": 10, "completion_tokens": 5}
+
+    def test_ignores_non_numeric_values(self):
```
andrewtruong: why would you want this?

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+        assert result == {"prompt_tokens": 10, "completion_tokens": 5}
+
+
+class TestValidateUsageShape:
+    """Test usage data validation for known provider formats."""
+
+    def test_valid_openai_format(self):
+        """Test OpenAI format validation."""
+        usage_data = {"prompt_tokens": 10, "completion_tokens": 5, "total_tokens": 15}
+        assert _validate_usage_shape(usage_data) is True
```
andrewtruong: nit: just assert, don't need `is True`

weave/integrations/langchain/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+            # Extract the token field name (e.g., "input_tokens" from "base_path.input_tokens")
+            field_name = key[len(base_prefix) :]
+            # Only include direct token fields, not nested structures
+            if "." not in field_name and isinstance(value, (int, float)):
+                usage_data[field_name] = int(value)
+
+    return usage_data
+
+
+def _validate_usage_shape(usage_data: dict) -> bool:
```
andrewtruong: May be better named `_is_valid_usage_shape`

weave/integrations/langchain/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+
+    # Google Vertex AI format
+    has_vertex_tokens = (
+        "prompt_token_count" in usage_data and "candidates_token_count" in usage_data
+    )
+
+    return has_openai_tokens or has_genai_tokens or has_vertex_tokens
+
+
+def _normalize_token_counts(usage_data: dict) -> tuple[int, int, int]:
```
andrewtruong: Better to return dataclass with named fields here

tests/integrations/langchain/helpers_test.py
```diff
@@ -0,0 +1,597 @@
...
+        assert _validate_usage_shape(usage_data) is False
+
+    def test_invalid_non_dict(self):
+        """Test non-dictionary input is invalid."""
+        assert _validate_usage_shape("not a dict") is False
+        assert _validate_usage_shape(None) is False
+        assert _validate_usage_shape(123) is False
+
+
+class TestNormalizeTokenCounts:
```
andrewtruong: This suite of tests is probably better handled with a `pytest.mark.parametrize`

andrewtruong: (same with most of the other test classes)

weave/integrations/langchain/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+        Google GenAI: input_tokens → prompt_tokens, output_tokens → completion_tokens
+        Google Vertex AI: prompt_token_count → prompt_tokens, candidates_token_count → completion_tokens
+    """
+    prompt_tokens = 0
+    completion_tokens = 0
+    total_tokens = 0
+
+    # OpenAI format: prompt_tokens, completion_tokens, total_tokens
+    if "prompt_tokens" in usage_data and "completion_tokens" in usage_data:
+        prompt_tokens = usage_data.get("prompt_tokens", 0) or 0
```
andrewtruong: the `get(..., 0)` here are redundant

weave/integrations/langchain/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+        # response_metadata patterns
+        search_patterns.append(
+            ".".join(base_path + ["response_metadata", "model_name"])
+        )
+
+    for pattern in search_patterns:
+        if pattern in flattened:
+            model_name = flattened[pattern]
+            if isinstance(model_name, str) and model_name:
+                return model_name
```
andrewtruong: what is going on here?

weave/integrations/langchain/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+    for prefix_length in range(len(path_parts) - 2, 0, -1):
+        prefix = ".".join(path_parts[:prefix_length])
+        for key, value in flattened.items():
+            if (
+                key.startswith(prefix)
+                and "model_name" in key
+                and isinstance(value, str)
+                and value
+            ):
+                return value
```
andrewtruong: Also here

weave/integrations/langchain/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+            usage_data = _reconstruct_usage_data_from_flattened(flattened, base_path)
+            if not _validate_usage_shape(usage_data):
+                continue
+
+            model = _extract_model_from_flattened_path(flattened, base_path)
+            prompt_tokens, completion_tokens, total_tokens = _normalize_token_counts(
+                usage_data
+            )
+
+            usage_dict[model]["prompt_tokens"] += prompt_tokens
```
chance-wnb: I see that things get somewhat tricky here:

Previously I was looping through `generations` (a designated location), for example:

```python
for output_item in output["outputs"]
        for generation_list in output_item["generations"]:
            for generation in generation_list:
```

That means if there are multiple generations from a response (it generated multi-part content), it makes sense to add the usage tokens together. 

However with your new implementation, for cases of vertex AI

https://wandb.ai/wandb/wb-21343/weave/traces?view=traces_default&peekPath=%2Fwandb%2Fwb-21343%2Fcalls%2F0197411d-a49b-7a20-8ee5-c20c03841044%3FdescendentCallId%3D0197411d-a65e-7e71-aeac-e84a2d2217dc%26hideTraceTree%3D0&pageSize=50

The usage data reappear in both the 
`outputs.0.generations.0.0.generation_info.usage_metadata` section
and
`outputs.0.generations.0.0.mesage.kwargs.usage_metadata` section.

If you are simply summing up all occurrences, you might double-count some usage data.

ideally, for each generation, we should only take 1.


I don't have a good solution, my subpar solution would be to identify each generation first using some heuristic (probably something similar as identifying the usage_metadata), grab its sub tree, then find all the usages like you are doing now within the subtree, finally only take 1 per subtree.

weave/integrations/langchain/helpers.py
parambharat: renamed as suggested

weave/integrations/langchain/helpers.py
parambharat: fixed the redundancy

weave/integrations/langchain/helpers.py
parambharat: refactored to return a dataclass instead

weave/integrations/langchain/helpers.py
parambharat: This searches for model names in predictable locations relative to usage metadata paths. We generate candidate paths by truncating the usage path and adding known model name locations like "generation_info.model_name" and "response_metadata.model_name".

weave/integrations/langchain/helpers.py
parambharat: This is fallback logic for edge cases where providers use non-standard nesting. It progressively searches broader prefixes of the usage path to catch model names that don't follow the common patterns above.

weave/integrations/langchain/helpers.py
parambharat: Added generation-scoped extraction with `_find_generation_paths()` and `_extract_usage_from_generation()`. Now we identify each generation boundary first, then extract usage data only once per generation with priority ordering (generation_info > message.kwargs > other locations) to prevent double-counting.

tests/integrations/langchain/helpers_test.py
parambharat: resolved. Testing the whole dict here and everywhere else

tests/integrations/langchain/helpers_test.py
parambharat: resolved. Testing the whole dict here and everywhere else

tests/integrations/langchain/helpers_test.py
parambharat: I'm not sure if we do in the other integrations, but, langchain has a batch method in their llm.invoke, that's why I have this tested

tests/integrations/langchain/helpers_test.py
parambharat: this was in place due to a simple bug. resolved now.

tests/integrations/langchain/helpers_test.py
parambharat: removed, this is handled already and is not necessary anymore

tests/integrations/langchain/helpers_test.py
parambharat: removed

tests/integrations/langchain/helpers_test.py
parambharat: removed

tests/integrations/langchain/helpers_test.py
parambharat: resolved

tests/integrations/langchain/helpers_test.py
parambharat: resolved: converted to use pytest functionality instead.

parambharat:
@andrewtruong and @chance-wnb, thank you again. I appreciate the eye for detail and thorough review. I was able to fix and resolve all the comments. Let me know what you think now.

weave/integrations/langchain/helpers.py
andrewtruong: how brittle is this?  I don't really understand what `range(1, min(len(path_parts), 6))` is supposed to mean -- did you just make that up?

weave/integrations/langchain/helpers.py
andrewtruong: Similar question as above -- wondering how brittle this is?

It's not intuitive to me what I'm looking at when I read this block.

tests/integrations/langchain/helpers_test.py
```diff
@@ -7,591 +7,880 @@
...
-                            }
+                                    },
+                                    "generation_info": {"model_name": "gpt-4o-mini"},
+                                }
+                            ]
                         ]
-                    ]
+                    }
+                ]
+            },
```
andrewtruong: please make these into fixtures!

andrewtruong: The fixtures will be reusable in tests below as well

weave/integrations/langchain/helpers.py
```diff
@@ -21,13 +22,90 @@
...
+        """
+        token_counts = _normalize_token_counts(usage_data)
+        return cls(
+            prompt_tokens=token_counts.prompt_tokens,
+            completion_tokens=token_counts.completion_tokens,
+            total_tokens=token_counts.total_tokens,
+            model_name=model_name,
+        )
+
+    def is_valid(self) -> bool:
```
andrewtruong: If you have this, shouldn't it be a check in the constructors?

weave/integrations/langchain/helpers.py
```diff
@@ -0,0 +1,491 @@
...
+
+    Returns:
+        Model name or "unknown" if not found
+
+    Common patterns:
+        usage at "outputs.0.usage_metadata" → model at "outputs.0.generation_info.model_name"
+    """
+    path_parts = usage_key_path.split(".")
+
+    # Generate specific candidate paths based on observed provider patterns.
```
chance-wnb: Would this algorithm work?

```python
keys = flattened.keys()
keys = keys.filter (key => key.startWith(usage_key_path))
keys = keys.filter( key => key.endWith("model_name")) # this is a string literal

return flattened[keys[0]]

```

What are some edges that this wouldn't work?

weave/integrations/langchain/helpers.py
parambharat: 6 was based on the observed dictionaries is among most langchain providers. I chose it to balance coverage and performance, assuming most provider response paths won’t be deeper than 6 segments from the usage metadata. But you are right, it could be brittle and break things. So. I'll change to do a full check instead.

weave/integrations/langchain/helpers.py
parambharat: this was also a heuristic. I was trying broader prefixes by progressively shortening the path, but the -2 offset is arbitrary. Changing this to be checking all prefixes instead.

weave/integrations/langchain/helpers.py
parambharat: good catch, moved it to the post_init instead

weave/integrations/langchain/helpers.py
parambharat: added a new simplified fast path first. But this doesn't handle edge cases so we fallback to the more complex check

parambharat:
@chance-wnb  and @andrewtruong . I went back to the drawing board on this one. I've simplified the usage extraction logic to leverage existing metadata that LangChain already maintains. I've confirmed the new changes work with the existing LLM providers as well as other providers. Additionally, I observed that Google Genai tests were failing in the 3.9 shard due to HTTP gzip encoding errors when cassettes are written in Python 3.9. I've removed langchain tests from the 3.9 shard as a workaround. Please take a look and let me know what you think.

tests/integrations/langchain/helpers_test.py
parambharat: made all tests parameterized

tests/integrations/langchain/langchain_test.py
```diff
@@ -759,7 +759,39 @@ def test_langchain_google_genai_usage(client: WeaveClient) -> None:
...
+    assert "usage" in call.summary
+    assert "gemini-1.5-pro-002" in call.summary["usage"].unwrap()
+
+
+@pytest.mark.skip_clickhouse_client
+@pytest.mark.vcr(
+    filter_headers=["authorization", "x-api-key"],
+    allowed_hosts=["api.wandb.ai", "localhost", "trace.wandb.ai"],
+    before_record_request=filter_body,
+    match_on=["method", "scheme", "path", "query"],
```
andrewtruong, line 786: why do you need this?

weave/integrations/langchain/helpers.py
```diff
@@ -11,481 +11,144 @@
...
-        return False
-
-    # OpenAI format
-    has_openai_tokens = (
-        "prompt_tokens" in usage_data and "completion_tokens" in usage_data
-    )
-
-    # Google GenAI format
-    has_genai_tokens = "input_tokens" in usage_data and "output_tokens" in usage_data
+        Tuple of (prompt_tokens, completion_tokens, total_tokens)
```
andrewtruong, line 37: not a tuple

weave/integrations/langchain/helpers.py
```diff
@@ -11,481 +11,144 @@
...
-                (
-                    ".input_tokens",
-                    ".output_tokens",
-                    ".total_tokens",
-                    ".prompt_tokens",
-                    ".completion_tokens",
-                    ".prompt_token_count",
-                    ".candidates_token_count",
-                    ".total_token_count",
+def _extract_usage_data(call: Call, output: Any) -> None:
```
andrewtruong, line 105: refine type for output

chance-wnb (approved):
Thanks Bharat! LGTM. Approving.

chance-wnb:
Sorry, I didn't expect the PR auto-merged upon approval. Please help also address Andrew's comment as well.

PR #5195, opened 2025-07-31
feat: handle error states in imperative eval logger

DanielPolatajko:
## Description

When using the imperative eval logger, it is possible that we might encounter an error state. In such cases, the eval logger cannot currently make the eval as failed, which gives the appearance of hanging evals in the UI. This PR adds the option to pass an exception to the finish method, which allows the eval logger to gracefully handle error states.

## Testing

Unit tested + tried it in a test project

### DIFF
--- a/weave/flow/eval_imperative.py
+++ b/weave/flow/eval_imperative.py
@@ -463,7 +463,7 @@ class EvaluationLogger(BaseModel):
                 # This is best effort.  If we fail, just swallow the error.
                 pass
 
-    def _finalize_evaluation(self, output: Any = None) -> None:
+    def _finalize_evaluation(self, output: Any = None, exception: BaseException | None = None) -> None:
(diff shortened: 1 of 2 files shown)


weave/flow/eval_imperative.py
```diff
@@ -571,7 +571,7 @@ def log_summary(
 
         self._finalize_evaluation(output=final_summary)
 
-    def finish(self) -> None:
+    def finish(self, exception: BaseException | None = None) -> None:
```
andrewtruong, line 574: Should we instead create a new method called `fail`?  What do you think?

DanielPolatajko: i think having a convenience method `fail` that calls `finish` as written makes sense and is least repetitive - see what you think

andrewtruong:
This was merged in the PR above!

PR #5197, opened 2025-07-31
chore(weave): Improve evaluation status resolver logic

tssweeney:
Fixes a bug identified when building evals backend. In particular, the return type of the completed status was incorrect. Fixes test to catch it correctly.

### DIFF
--- a/weave/trace_server/methods/evaluation_status.py
+++ b/weave/trace_server/methods/evaluation_status.py
@@ -30,7 +30,7 @@ def evaluation_status(
         return tsi.EvaluationStatusRes(status=tsi.EvaluationStatusFailed())
 
     if eval_call.call.ended_at is not None:
-        return tsi.EvaluationStatusRes(status=tsi.EvaluationStatusComplete())
+        return tsi.EvaluationStatusRes(status=tsi.EvaluationStatusComplete(output=eval_call.call.output))
 
     # determine completed rows (children complete)
     children_calls = server.calls_query(
@@ -42,7 +42,8 @@ def evaluation_status(
             ),
             columns=["id", "ended_at"],
         )
-    )
+    ).calls
+
     completed_children_calls = [
         call.id for call in children_calls if call.ended_at is not None
     ]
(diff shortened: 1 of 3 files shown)


bcsherma (approved):

weave/trace_server/methods/evaluation_status.py
```diff
@@ -87,7 +90,7 @@ def get_total_rows() -> int:
         table_rows_stats = server.table_query_stats(
             tsi.TableQueryStatsReq(
```
gtarpenning, line 100: i'm going to put up a fast follow to mark all of our req types with `model_config = ConfigDict(extra="forbid")` that would have caught this....

bcsherma:
I stamped but looks like the sqlite tests are failing (edit: I think it's a flake)

weave/trace_server/methods/evaluation_status.py
```diff
@@ -30,7 +30,9 @@ def evaluation_status(
         return tsi.EvaluationStatusRes(status=tsi.EvaluationStatusFailed())
 
     if eval_call.call.ended_at is not None:
-        return tsi.EvaluationStatusRes(status=tsi.EvaluationStatusComplete())
+        return tsi.EvaluationStatusRes(
+            status=tsi.EvaluationStatusComplete(output=eval_call.call.output)
+        )
 
     # determine completed rows (children complete)
     children_calls = server.calls_query(
```
gtarpenning: why are we using `calls_query` instead of the stream?

gtarpenning (approved):
Interesting... lgtm, i think strict typing on the pydantic models would have caught this, or at least one of the issues...

weave/trace_server/methods/evaluation_status.py
tssweeney: 👍 yup good idea

weave/trace_server/methods/evaluation_status.py
tssweeney: i actually modified this to use the stats query which is faster

PR #5198, opened 2025-07-31
chore(weave): strict request and filter types in trace server interface

gtarpenning:
## Description

[WB-26574](https://wandb.atlassian.net/browse/WB-26574)

No requests to the trace server should have extra keys, explicitly forbid them in pydantic, classes should inherit from this.  

## Testing

existing coverage should suffice, as we have made the typing strictly stricter. 

also fixes various mis-uses of the types. 

[WB-26574]: https://wandb.atlassian.net/browse/WB-26574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/trace_server_interface.py
```diff
@@ -237,7 +243,7 @@ class TableSchemaForInsert(BaseModel):
     rows: list[dict[str, Any]]
 
 
-class OtelExportReq(BaseModel):
+class OtelExportReq(BaseModelStrict):
     model_config = ConfigDict(arbitrary_types_allowed=True)
```
andrewtruong: Wont this override your model config declared earlier?

andrewtruong (commented):
Is `forbid` too harsh?  What happens as the API evolves?  Would the default `ignore` and a helpful warning be more useful?

weave/trace_server/trace_server_interface.py
gtarpenning: nice, yeah probably, fixing

gtarpenning:
> Is `forbid` too harsh? What happens as the API evolves? Would the default `ignore` and a helpful warning be more useful?

If a new version of the client comes out with updated types, i think we should error with the old request type because the server will be using the new type, and the old request body will presumably result will be misconfigured. Thats mostly a hypothetical because I can't see removing/deprecating params from requests very frequently.

andrewtruong (approved):

PR #5199, opened 2025-07-31
feat(weave): Enable auth on Kafka clients for dedicated deployments

neutralino1:
## Description

Enables auth for bufstream clients. This is necessary to support online evals in dedicated.

## Testing

- QA azure (bufstream with auth)
- QA MTSaaS (bufstream without auth)


weave/trace_server/environment.py
```diff
@@ -7,14 +7,24 @@
 # Kafka Settings
 
 
-def wf_kafka_broker_host() -> str:
+def kafka_broker_host() -> str:
     """The host of the kafka broker."""
-    return os.environ.get("WF_KAFKA_BROKER_HOST", "localhost")
+    return os.environ.get("KAFKA_BROKER_HOST", "localhost")
```
neutralino1, line 12: Renaming the env vars so that they match the kafka config map in dedicated deployments.

weave/trace_server/environment.py
```diff
@@ -82,7 +92,7 @@ def wf_clickhouse_max_execution_time() -> Optional[int]:
         return None
     try:
         return int(time)
-    except ValueError as e:
+    except ValueError:
```
neutralino1, line 95: These are automated linting changes I believe.

weave/trace_server/environment.py
tssweeney: do we need to change current configuration in MTSaaS charts?

neutralino1: https://github.com/wandb/core/pull/32537/files

neutralino1:
@tssweeney That doesn't seem necessary as long as I can merge https://github.com/wandb/core/pull/32537 at the same time.

I could even wait until later hours to make sure no deploys go through in the meantime.

But that also brings up another point, kafka conf should not be in weave, all of this should live in core. Follow-up work for me.

tssweeney:
> @tssweeney That doesn't seem necessary as long as I can merge [wandb/core#32537](https://github.com/wandb/core/pull/32537) at the same time.
> 
> I could even wait until later hours to make sure no deploys go through in the meantime.
> 
> But that also brings up another point, kafka conf should not be in weave, all of this should live in core. Follow-up work for me.

Sure, but core generally introduces gaps. I would love to keep the train moving here as my stuff is blocked if possible. If you don't want to update the code, we can just have a deploy freeze on TS until core lands

PR #5201, opened 2025-08-01
feat(weave): add custom eval attributes to EvaluationLogger

andrewtruong:
https://github.com/wandb/weave/pull/5171

## Description

This PR adds the ability to customise the attributes of an imperative EvaluationLogger. This is useful in order to write metadata columns to the Evaluations tab in the UI when using the EvaluationLogger.

## Testing

Added a unit test - I have also tried out previously in a test environment on my own [project](https://github.com/DanielPolatajko/inspect_weave/blob/add-metadata/inspect_weave/custom_evaluation_logger.py)


weave/flow/eval_imperative.py
```diff
@@ -355,6 +355,15 @@ class EvaluationLogger(BaseModel):
         ),
     ]
 
+    eval_attributes: Annotated[
+        dict[str, Any],
+        Field(
+            default_factory=dict,
+            description="(Optional): A dictionary of attributes to add to the evaluation call."
+            "These attributes can be used to add additional metadata columns to the Evaluation",
```
gtarpenning:
```suggestion
            "These attributes can be used to add additional metadata columns to the Evaluation.",
```

weave/flow/eval_imperative.py
```diff
@@ -369,6 +378,10 @@ def ui_url(self) -> str | None:
             return None
         return self._evaluate_call.ui_url
 
+    @property
+    def attributes(self) -> dict[str, Any]:
+        return IMPERATIVE_EVAL_MARKER | self.eval_attributes
```
gtarpenning: will our marker get clobbered? shouldn't this be the other way around?

gtarpenning (approved):

weave/flow/eval_imperative.py
andrewtruong: good point

PR #5203, opened 2025-08-01
feat(weave): add OTEL parsing for thread_id and handling for is_turn

zbirenbaum:
## Description

- Fixes WB-26584

Add parsing and handling for these two attributes:
wandb.thread_id
wandb.is_turn

When wandb.is_turn is set to true, sets turn_id to the span.id (which is used for call.id)
When wandb.is_turn is set but thread_id is unset, ignore, is_turn is ignored

## Testing

Added detailed test cases for the different scenarios

### DIFF
--- a/weave/trace_server/opentelemetry/python_spans.py
+++ b/weave/trace_server/opentelemetry/python_spans.py
@@ -352,9 +352,7 @@ class Span:
 
         thread_id = wandb_attributes.get("thread_id") or None
         if thread_id is not None and (wandb_attributes.get("is_turn")):
-            from uuid import uuid4
-
-            turn_id = str(uuid4())
+            turn_id = self.span_id
         else:
             turn_id = None
 
--- a/weave/trace_server/opentelemetry/python_spans.py
+++ b/weave/trace_server/opentelemetry/python_spans.py
@@ -353,6 +353,7 @@ class Span:
         thread_id = wandb_attributes.get("thread_id") or None
         if thread_id is not None and (wandb_attributes.get("is_turn")):
             from uuid import uuid4
+
             turn_id = str(uuid4())
         else:
             turn_id = None
--- a/weave/trace_server/opentelemetry/python_spans.py
+++ b/weave/trace_server/opentelemetry/python_spans.py
@@ -11,7 +11,6 @@ from collections.abc import Iterator
 from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any, Optional
-from uuid import uuid4
 
 from opentelemetry.proto.common.v1.common_pb2 import InstrumentationScope
 from opentelemetry.proto.resource.v1.resource_pb2 import Resource as PbResource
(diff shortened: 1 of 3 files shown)


zbirenbaum:
* **#5203** <a href="https://app.graphite.dev/github/pr/wandb/weave/5203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/5203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/opentelemetry/python_spans.py
```diff
@@ -350,6 +350,16 @@ def to_call(
         if display_name and len(display_name) >= MAX_DISPLAY_NAME_LENGTH:
             display_name = shorten_name(display_name, MAX_DISPLAY_NAME_LENGTH)
 
+        thread_id = wandb_attributes.get("thread_id") or None
+        if thread_id is not None and (wandb_attributes.get("is_turn")):
+            from uuid import uuid4
+            turn_id = str(uuid4())
```
chance-wnb: For turn calls, we require that turn_id equals call.id. Otherwise a call will not be recognized as turns.

chance-wnb (changes requested):

weave/trace_server/opentelemetry/python_spans.py
zbirenbaum: I made this fix but I think there's probably a more straightforward way to handle this knowing that.

For OTEL call.id = span.id

We could probably just propogate span_id to turn_id directly, but either way results in the same. I think this is fine for now and we can update later if it's more convenient or we ship a lib/sdk util to do this for the user

chance-wnb: > but either way results in the same

In this scenario:
thread 
-> parent call 1
----> child call 1
----> child call 2
-> parent call 2
----> child call 3
----> child call 4

If we just assign span.id(aka call id) -> turn_id. 

Then parent call 1, child call 1, child call 2, parent call 2, child call 3, child call 4 will all become turns of the thread. The parent-child relationship becomes siblings. 

Therefore only marking `parent call 1` and `parent call 2` with `is_turn` is ideal. And you current approach is correct.

weave/trace_server/opentelemetry/python_spans.py
```diff
@@ -350,6 +350,15 @@ def to_call(
         if display_name and len(display_name) >= MAX_DISPLAY_NAME_LENGTH:
             display_name = shorten_name(display_name, MAX_DISPLAY_NAME_LENGTH)
 
+        thread_id = wandb_attributes.get("thread_id") or None
+        if thread_id is not None and (wandb_attributes.get("is_turn")):
```
neutralino1, line 354:
```suggestion
        if thread_id is not None and wandb_attributes.get("is_turn") is True:
```

zbirenbaum: We do not want to do this. Having is_turn be any truthy value lets me reuse the same key for is_turn and thread_id for the ADK instance.

This check would make it much more difficult to implement

PR #5210, opened 2025-08-05
chore(weave): sort call inserts for grouping

gtarpenning:
## Description

Call inserts have a random order, making grouping in ch cloud annoying. Lets sort these so they all appear under the same query.  also fix an empty set operation

Example:
Screenshot 2025-08-05 at 9 12 41 AM


gtarpenning:
closing to prefer full settings refactor

PR #5230, opened 2025-08-06
chore(deps): Move sentry-sdk to optional dependency

samad-wisdomlabs:
The sentry-sdk dependency has been moved from required to optional dependencies. This change reduces the required dependencies for users who don't need Sentry integration.

Changes:
- Moved sentry-sdk from main dependencies to optional dependencies in pyproject.toml
- Added conditional imports with SENTRY_AVAILABLE flag in async_batch_processor.py
- Updated trace_sentry.py to handle missing sentry-sdk gracefully
- All Sentry API calls are now guarded with availability checks

Users can now install with Sentry support using: pip install weave[sentry]

## Description

Reduces the number of required dependencies for Weave
Allows users to opt-in to Sentry integration only when needed
Follows the same pattern already established for other optional dependencies like wandb

## Testing
Regular CI checks
TODO

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,6 @@ dependencies = [
 
   # Used for content type detection
   # Remove once license content filtering is upstreamed
-  "polyfile-weave",
 
   # This dependency will be updated each time we regenerate the trace server bindings.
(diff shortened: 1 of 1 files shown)


samad-wisdomlabs:

PR #5232, opened 2025-08-06
chore(weave): Update codegen to use Stainless CLI

andrewtruong:
Updates codegen to use the Stainless CLI tool instead of a previously included node bundle


tools/codegen/generate.py
```diff
@@ -14,10 +14,21 @@
...
-from __future__ import annotations
+#!/usr/bin/env -S uv run --script
+# /// script
+# dependencies = [
+#     "click==8.2.1",
+#     "httpx==0.28.1",
+#     "tomlkit==0.13.3",
+#     "PyYAML==6.0.2",
+#     "rich==14.0.0",
+# ]
```
neutralino1, line 25: Is this a uv-specific syntax?

tools/codegen/generate.py
```diff
@@ -593,5 +608,11 @@ def _load_config(config_path: str | Path) -> dict[str, Any]:
         return cfg
 
 
+def _random_temp_dir() -> str:
+    ascii_letters_and_digits = string.ascii_letters + string.digits
+    random_string = "".join(random.choices(ascii_letters_and_digits, k=16))
+    return f"tmp/{random_string}"
```
neutralino1: Why not use `tempfile`?

```python
import tempfile

# Create a temporary directory
with tempfile.TemporaryDirectory() as temp_dir:
    print(f"Temporary directory: {temp_dir}")
    # Use the directory here
    # It gets automatically deleted when exiting the with block

# Or create one that persists:
temp_dir = tempfile.mkdtemp()
print(f"Persistent temp directory: {temp_dir}")
```

neutralino1 (commented):
One comment.

tools/codegen/generate.py
andrewtruong: This is PEP723:
https://peps.python.org/pep-0723/

tools/codegen/generate.py
andrewtruong: Will check with Stainless -- this is what they suggested for the `branch` flag

andrewtruong: this was mislabeled -- it's just a branch name, not a temp dir

neutralino1 (approved):

PR #5237, opened 2025-08-07
chore(weave): Add drop-in replacement for remote_http_trace_server

andrewtruong:


pyproject.toml
andrewtruong, line 1: this change makes the stainless client an optional extra during the opt-in period

weave/trace/weave_init.py
```diff
@@ -207,8 +208,15 @@ def init_weave_disabled() -> InitializedClient:
 def init_weave_get_server(
     api_key: str | None = None,
     should_batch: bool = True,
-) -> remote_http_trace_server.RemoteHTTPTraceServer:
-    res = remote_http_trace_server.RemoteHTTPTraceServer.from_env(should_batch)
+) -> (
+    remote_http_trace_server.RemoteHTTPTraceServer
+    | stainless_trace_server.RemoteHTTPTraceServer
```
neutralino1, line 209: Shouldn't we abstract the interface as an abstract base class? I guess essentially `TraceServerInterface`?

weave/trace_server_bindings/stainless_trace_server.py
```diff
@@ -0,0 +1,838 @@
...
+)
+from weave.utils.retry import _is_retryable_exception, get_current_retry_id, with_retry
+from weave.wandb_interface import project_creator
+
+REMOTE_REQUEST_BYTES_LIMIT = (
+    (32 - 1) * 1024 * 1024
+)  # 32 MiB (real limit) - 1 MiB (buffer)
+
+
+class RemoteHTTPTraceServer(tsi.TraceServerInterface):
```
neutralino1, line 41: There seems to be a lot of duplicated logic with the baseline class. Can we subclass and just override the parts that differ, e.g. setting `_client` to `WeaveTrace`...

neutralino1 (commented):
Main comment is that maybe we could subclass from the existing class and simply override the parts that differ instead of replicating a lot of logic? Is that possible?

weave/trace_server_bindings/stainless_trace_server.py
andrewtruong: The interface is just a protocol, so you have to implement everything anyways.  

The implementation of `stainless_trace_server.py` is intended to look like https://github.com/wandb/weave/blame/andrew/stainless-bindings/weave/trace_server_bindings/remote_http_trace_server.py#L39, just using the Stainless client under the hood

andrewtruong: We could refactor the `RemoteHTTPTraceServer` to share code with this, but seeing as the intent is to remove it in the near future I don't think it's necessary

weave/trace/weave_init.py
andrewtruong: See comment below -- the type here is really a variant of `TraceServerInterface`, but I don't want to change the impl of `RemoteHTTPTraceServer` when we plan on removing it soon.

neutralino1 (approved):

PR #5245, opened 2025-08-08
chore(weave): Update codegen to include new endpoints

andrewtruong:
Adds OTEL, completions, and threads endpoints.

Additionally exposes the `completions` endpoint which was previously hidden


gtarpenning (approved):
This is autogen right?

andrewtruong:
`openapi.json` is generated, the other two are manual

PR #5249, opened 2025-08-09
chore(weave): Update stainless generator to automatically create a branch in the `weave_server_sdk` repo that mirrors the current branch

andrewtruong:
Adds a step in the generator to automatically create a branch in the `weave_server_sdk` repo that mirrors the current branch of the `weave` repo.

### DIFF
--- a/tools/codegen/generate.py
+++ b/tools/codegen/generate.py
@@ -228,7 +228,7 @@ def update_pyproject(
     if release:
         version = _get_package_version(python_output)
         _update_pyproject_toml(package_name, version, True)
-        info(f"Updated {package_name} dependency to version: {version}")
+        info(f"Updated {package_name} in [stainless] extra to version: {version}")
     else:
         repo_info = _get_repo_info(python_output)
         remote_url = repo_info.remote_url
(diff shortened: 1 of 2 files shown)


tools/codegen/generate.py
```diff
@@ -237,7 +237,204 @@ def update_pyproject(
             error(f"Failed to get git SHA (got: {sha=})")
             sys.exit(1)
         _update_pyproject_toml(package_name, f"{remote_url}@{sha}", False)
-        info(f"Updated {package_name} dependency to SHA: {sha}")
+        info(f"Updated {package_name} in [stainless] extra to SHA: {sha}")
+
+
+@app.command()
+def merge_generated_code(
```
gtarpenning, line 244: i'm worried this is not going to work in all cases, git state can get pretty complex... just something to keep an eye on I think

tools/codegen/generate.py
```diff
@@ -273,13 +470,21 @@ def all(
         bool | None,
         Option("--release", help="Update to the latest version"),
     ] = None,
+    auto_merge: Annotated[
+        bool,
+        Option(
+            "--auto-merge",
+            help="Automatically create a branch with generated code after generation",
+        ),
+    ] = True,
```
gtarpenning, line 479: this is on by default?

gtarpenning: oh i see, we are allowing for _not_ auto-merge in this case

PR #5251, opened 2025-08-10
chore(weave): Run linting on all files in CI

andrewtruong:
Also fixes some lint checks that were previously missed


weave/trace_server/reference/generate.py
```diff
@@ -131,11 +131,11 @@ def read_root(
         return service.read_root()
 
     @router.post("/otel/v1/trace", tags=[OTEL_TAG_NAME])
-    def export_trace(
+    def otel_export(
```
jamie-rasmussen: Is this change related?

jamie-rasmussen (approved):

PR #5258, opened 2025-08-11
chore(weave): Improve evaluation logger data collection

tssweeney:
Allows more configuration of Kafka consumers and adds some more logging to the system


weave/trace_server/kafka.py
```diff
@@ -17,12 +17,18 @@
 
 class KafkaProducer(ConfluentKafkaProducer):
     @classmethod
-    def from_env(cls) -> "KafkaProducer":
+    def from_env(
+        cls, additional_config: Optional[dict[str, Any]] = None
+    ) -> "KafkaProducer":
```
andrewtruong, line 29: This should be `-> Self`

weave/trace_server/kafka.py
```diff
@@ -41,14 +47,20 @@ def produce_call_end(
 
 class KafkaConsumer(ConfluentKafkaConsumer):
     @classmethod
-    def from_env(cls, group_id: str) -> "KafkaConsumer":
+    def from_env(
+        cls, group_id: str, additional_config: Optional[dict[str, Any]] = None
+    ) -> "KafkaConsumer":
```
andrewtruong, line 67: here too

weave/trace_server/kafka.py
```diff
@@ -17,12 +17,18 @@
 
 class KafkaProducer(ConfluentKafkaProducer):
     @classmethod
-    def from_env(cls) -> "KafkaProducer":
+    def from_env(
+        cls, additional_config: Optional[dict[str, Any]] = None
```
andrewtruong: nit: We should add ruff to check for this, but I believe the trace server is on py311 so this should be annotated as `dict[str, Any] | None`

weave/trace_server/kafka.py
```diff
@@ -17,12 +17,18 @@
...
+    ) -> "KafkaProducer":
+        if additional_config is None:
+            additional_config = {}
+
         config = {
             "bootstrap.servers": _make_broker_host(),
             "client.id": socket.gethostname(),
             "message.timeout.ms": 500,
             **_make_auth_config(),
+            **additional_config,
```
andrewtruong: Where are these configs coming from?  Can we add a comment?

weave/trace_server/kafka.py
tssweeney: done

weave/trace_server/kafka.py
tssweeney: done

weave/trace_server/kafka.py
tssweeney: this conforms to the current lint rules of this module. Happy to change once new rules are set

PR #5291, opened 2025-08-18
chore(weave): Pin openai to unblock CI

andrewtruong:
`openai==1.100.0` changes imports which causes CI to fail.

Note that this is happening increasingly frequently!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,6 +94,7 @@ trace_server = [
 trace_server_tests = [
   # BYOB - S3
   "moto[s3]>=5.0.0",
+  "openai<1.100.0",
 ]
 docs = [
   "playwright",


pyproject.toml
```diff
@@ -107,13 +108,17 @@ anthropic = ["anthropic>=0.18.0"]
 cerebras = ["cerebras-cloud-sdk"]
 cohere = ["cohere>=5.9.1,<5.9.3"]
 # Pin until https://github.com/crewAIInc/crewAI/pull/2553 is released
-crewai = ["crewai>=0.100.1,<=0.108.0", "crewai-tools>=0.38.0", "openai>=1.99.9"]
-dspy = ["dspy>=2.6.27", "litellm[proxy]", "openai>=1.99.9"]
+crewai = [
+  "crewai>=0.100.1,<=0.108.0",
+  "crewai-tools>=0.38.0",
+  "openai>=1.99.9,<1.100.0",
```
chance-wnb, line 114: I think this equates to `openai ==1.99.9` 

there is no version in between

andrewtruong: Yes, that is the required version effectively

chance-wnb: Shall we just write `openai ==1.99.9` for simplicity and clarity?

PR #5293, opened 2025-08-18
chore(weave): Ban weave client SDK from importing from `trace_server` module

andrewtruong:
https://wandb.atlassian.net/browse/WB-27521

This PR adds an enforcement check to ensure that client code never imports from the `trace_server` dir because that's unsafe.

Unfortunately, our code is a bit tangled atm.  I'm adding this to prevent further tangling while we work to untangle!


weave/trace/refs.py
```diff
@@ -8,8 +8,10 @@
 
 from typing_extensions import Self
 
-from weave.trace_server import refs_internal
-from weave.trace_server.errors import ObjectDeletedError
+from weave.trace_server import refs_internal  # noqa: TID251
+from weave.trace_server.errors import (  # noqa: TID251 (this might be a real problem)
```
andrewtruong, line 12: @gtarpenning I think you added this.  Can you remind me why you need to get the specific trace server error?

weave/trace/vals.py
```diff
@@ -30,8 +30,10 @@
 )
 from weave.trace.serialization.serialize import from_json
 from weave.trace.table import Table
-from weave.trace_server.errors import ObjectDeletedError
-from weave.trace_server.trace_server_interface import (
+from weave.trace_server.errors import (  # noqa: TID251 (I think this might be a real problem)
```
andrewtruong, line 32: @gtarpenning same here

weave/trace/refs.py
gtarpenning: this is the exact error raised by the server layer, so we can easily identify and catch and manipulate the actual error. This callsite it less important, the `vals.py` callsite is where we catch the error.

weave/trace/vals.py
gtarpenning: why do you think "this might be a real problem"?

andrewtruong: Because we should never import from the `trace_server` dir inside the client

neutralino1:
@andrewtruong merge or close?

PR #5294, opened 2025-08-18
fix(weave): defer clickhouse imports until registered, in server

gtarpenning:
## Description

[WB-26999](https://wandb.atlassian.net/browse/WB-26999)

Fixes sdk dependency on `clickhouse_connect`

## Testing

existing coverage

[WB-26999]: https://wandb.atlassian.net/browse/WB-26999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,7 @@ trace_server = [
 trace_server_tests = [
   # BYOB - S3
   "moto[s3]>=5.0.0",
+  "openai<1.100.0",
 ]
 docs = [
   "playwright",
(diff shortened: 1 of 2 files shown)


gtarpenning:
> Why does this work? Is it because we only do an isinstance check in the client?

@andrewtruong We are importing the `ObjectDeletedError` in the client which does not require clickhouse. This pr defers importing `clickhouse_connect` to where we register all errors, which only happens in the server.

andrewtruong:
Won't this still error if the user e.g. hits this line?
https://github.com/wandb/weave/blame/master/weave/trace/vals.py#L780

gtarpenning:
> Won't this still error if the user e.g. hits this line? https://github.com/wandb/weave/blame/master/weave/trace/vals.py#L780

@andrewtruong  Why exactly would it error? The `ObjectDeletedError` does not require clickhouse_connect. None of the errors defined in the top level of the `errors.py` file require clickhouse_connect. The `ErrorRegistry` class calls `register` in the server (not client) which is where we make use of the clickhouse_connect imports.

andrewtruong:
Oh I see, I think I misread the diff.

I think we should do a better job separating "shared" vs "server-only" code, but that can left for another PR

PR #5306, opened 2025-08-19
chore(weave): Add support for gql>=4.0.0 (pt2)

andrewtruong:
Re-adds support for gql>=4.0.0 which was previously added then accidentally removed in https://github.com/wandb/weave/commit/e70cfe4af2eb0e3371ee48608c918d2595e67738#diff-a106ba8e7cd28ff9ea9bd95a182b7397daa2c2fa24394faf35cca07202be9953


jamie-rasmussen (approved):
Approving change previously stamped in https://github.com/wandb/weave/pull/5290

PR #5311, opened 2025-08-19
chore(weave): dynamically strip payloads on insert, slightly higher cap

gtarpenning:
## Description

[WB-27048](https://wandb.atlassian.net/browse/WB-27048)

Instead of stripping any payloads that are larger than 1mb, keep the whole sum under our cap by iteratively removing the largest json field, if the sum data size is over the limit. 

## Testing

updates existing test

Manually able to log much larger payloads, as long as the total size it beneath the strict limit. I was actually able to log up to a `8.1MB` payload! Let's bump up the limit.
Screenshot 2025-08-20 at 10 37 48 AM

Worst case, when you go over, we display a real error to the client:
```
weave: Trace input size (6292353 bytes) exceeds the maximum allowed size of 3670016 bytes.Inputs may be dropped.
weave: Trace input size (6292353 bytes) exceeds the maximum allowed size of 3670016 bytes.Inputs may be dropped.
weave: 🍩 https://wandb.ai/griffin_wb/log-limits/r/call/0198c8aa-07e1-70a5-a969-ed385f55d21b
weave: 🍩 https://wandb.ai/griffin_wb/log-limits/r/call/0198c8aa-07e2-7f82-8402-66886faedc5a
weave: Error sending batch of 4 call events to server
weave: dropped call start ids: ['0198c8aa-07e1-70a5-a969-ed385f55d21b', '0198c8aa-07e2-7f82-8402-66886faedc5a']
weave: dropped call end ids: ['0198c8aa-07e1-70a5-a969-ed385f55d21b', '0198c8aa-07e2-7f82-8402-66886faedc5a']
weave: status code: 413
weave: reason: Request Entity Too Large
weave: text: {"reason":"Database insertion failed. Record too large. A likely cause is that a single row or cell exceeded the limit. If logging images, save them as `Image.PIL`."}
```

[WB-27048]: https://wandb.atlassian.net/browse/WB-27048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2317,46 +2318,66 @@ def _flush_calls(self) -> None:
...
-            logger.info("Retrying with large objects stripped.")
-            batch = self._strip_large_values(self._call_batch)
-            self._insert_call_batch(batch)
+            logger.info(
+                "Retrying with large objects stripped and inserting one by one."
+            )
+            stripped_batch = self._strip_large_values(self._call_batch)
+            # Insert rows one at a time after stripping large values
+            for row in stripped_batch:
+                self._insert(
```
chance-wnb: (Optional). Do you want to properly wrap this into a `_insert_call_single()` (or a better name) method. I see that there is some extra logic in _insert_call_batch that does some DD and CH settings.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2317,46 +2318,66 @@ def _flush_calls(self) -> None:
...
-                        > ch_settings.CLICKHOUSE_SINGLE_VALUE_BYTES_LIMIT
-                    ):
-                        stripped_item += [ch_settings.ENTITY_TOO_LARGE_PAYLOAD]
-                        stripped_count += 1
-                    else:
-                        stripped_item += [value]
-                final_batch.append(stripped_item)
-            else:
+            # Calculate only JSON dump bytes
+            json_sizes = [
```
chance-wnb: nit; it might slight improve readability if you rename this to `json_idx_size_pairs` .

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2317,46 +2318,66 @@ def _flush_calls(self) -> None:
...
+                (i, _num_bytes(item[i])) for i in ALL_CALL_JSON_COLUMN_INDICES
```
chance-wnb: Nit: do you expect any other functions to reuse ALL_CALL_JSON_COLUMN_INDICES ? If not, then I guess calculate it within this function base off ALL_CALL_JSON_COLUMN on the fly might reveal what it is better and improves readability.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2317,46 +2318,66 @@ def _flush_calls(self) -> None:
...
+            ]
+            total_json_bytes = sum(size for _, size in json_sizes)
+
+            # If JSON columns are under limit, keep as is
+            if total_json_bytes <= ch_settings.CLICKHOUSE_SINGLE_ROW_INSERT_BYTES_LIMIT:
```
chance-wnb: nit: I think you might be doing this for performance reasons, so we save the need to run a sort. Considering the array is short, sorting might not mean much. Then you should be able to combine this check into line 2366, because this line is just a special case of that. In that way we make the function shorter and potentially more readable.

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: actually we can just use `_insert_call` without batching, good call

chance-wnb: If you use _insert_call, maybe you need to ensure self._flush_immediately is true? or maybe having a dedicated method is easier.

gtarpenning: it should do that automatically as long as its not in the context

chance-wnb (approved):
LGTM!

PR #5314, opened 2025-08-19
chore(weave): Add a data marker for evaluation worker evaluations

tssweeney:
Adds metadata to the attributes so we know which evals were run by the worker


weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -13,6 +14,8 @@
     LLMStructuredCompletionModel,
 )
 
+EVALUATE_MODEL_WORKER_MARKER = {"_weave_eval_meta": {"evaluate_model_worker": True}}
```
gtarpenning, line 17: might want to define `_weave_eval_meta` in some common place, now that we have multiple uses for it. 
could punt to followup 

@andrewtruong

PR #5315, opened 2025-08-19
feat(weave): Add time to first token for OpenAI streaming endpoints

andrewtruong:
https://wandb.atlassian.net/browse/WB-27046

This tacks a `time_to_first_token` key onto the output of OpenAI calls.

Ideally this would be in the summary!

image


weave/integrations/openai/openai_sdk.py
```diff
@@ -15,6 +16,13 @@
     _add_accumulator,
     _default_on_input_handler,
 )
+from weave.utils.stream_metrics import (
+    WEAVE_STREAM_START_TIME,
```
zbirenbaum, line 20: nit & totally not required but for style consistency you could add a function or two which perform the operations you do with this const to keep the logic flow a bit cleaner in here. I think it's fine as is personally

zbirenbaum (approved):
Looks good, one small thing that came to mind but not requred

PR #5323, opened 2025-08-20
feat(weave): Add cw costs, to cost checkpoint

jwlee64:
## Description

Fixes https://wandb.atlassian.net/browse/WB-26160

Updates script to add costs
runs make update_costs

Screenshot 2025-08-20 at 1 27 32 PM

## Testing

manually

### DIFF
--- a/weave/trace_server/costs/cost_checkpoint.json
+++ b/weave/trace_server/costs/cost_checkpoint.json
@@ -9556,5 +9556,165 @@
       "output": 4.4e-06,
       "created_at": "2025-08-19 20:10:24"
     }
+  ],
+  "together_ai/Qwen/Qwen3-235B-A22B-Instruct-2507-tput": [
+    {
+      "provider": "together_ai",
+      "input": 2e-07,
+      "output": 6e-06,
+      "created_at": "2025-08-20 13:33:54"
+    }
+  ],
+  "together_ai/Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8": [
+    {
+      "provider": "together_ai",
+      "input": 2e-06,
+      "output": 2e-06,
+      "created_at": "2025-08-20 13:33:54"
+    }
+  ],
+  "together_ai/Qwen/Qwen3-235B-A22B-Thinking-2507": [
+    {
+      "provider": "together_ai",
+      "input": 6.5e-07,
+      "output": 3e-06,
+      "created_at": "2025-08-20 13:33:54"
+    }
+  ],
+  "together_ai/Qwen/Qwen3-235B-A22B-fp8-tput": [
+    {
+      "provider": "together_ai",
+      "input": 2e-07,
+      "output": 6e-07,
+      "created_at": "2025-08-20 13:33:54"
+    }
+  ],
+  "together_ai/deepseek-ai/DeepSeek-R1-0528-tput": [
+    {
+      "provider": "together_ai",
+      "input": 5.5e-07,
+      "output": 2.19e-06,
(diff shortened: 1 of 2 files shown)


weave/trace_server/costs/cost_checkpoint.json
```diff
@@ -9556,5 +9556,261 @@
       "output": 4.4e-06,
       "created_at": "2025-08-19 20:10:24"
     }
+  ],
+  "together_ai/Qwen/Qwen3-235B-A22B-Instruct-2507-tput": [
```
jwlee64, line 9560: Looks like these together ai ones got updated in litellm so they were updated as well when running make update_costs

weave/trace_server/costs/update_costs.py
```diff
@@ -77,6 +80,73 @@ def fetch_new_costs() -> dict[str, CostDetails]:
...
+
+    Returns:
+        dict[str, CostDetails]: Dictionary of model costs keyed by model ID.
+
+    Examples:
+        >>> costs = fetch_models_begin_costs()
+        >>> len(costs) > 0
+        True
+    """
+    models_begin_path = os.path.join(os.path.dirname(__file__), MODELS_BEGIN_FILE)
```
jamie-rasmussen: I'd use `os.path.abspath` or pathlib resolve so the missing check below doesn't print something like `costs/../model_providers`

weave/trace_server/costs/update_costs.py
```diff
@@ -77,6 +80,73 @@ def fetch_new_costs() -> dict[str, CostDetails]:
...
+
+    if not os.path.exists(models_begin_path):
+        print(f"modelsBegin.json not found at {models_begin_path}, skipping")
+        return {}
```
jamie-rasmussen: This feels like maybe it should halt execution and `sys.exit(1)` instead?

weave/trace_server/costs/update_costs.py
```diff
@@ -77,6 +80,73 @@ def fetch_new_costs() -> dict[str, CostDetails]:
     return costs
 
 
+def fetch_models_begin_costs() -> dict[str, CostDetails]:
+    """
+    Fetches costs from modelsBegin.json and converts them to the expected format.
+
+    Converts from cents per billion tokens to cost per token:
```
jamie-rasmussen: I'd explicitly say dollars per token

weave/trace_server/costs/update_costs.py
```diff
@@ -77,6 +80,73 @@ def fetch_new_costs() -> dict[str, CostDetails]:
...
+    - priceCentsPerBillionTokensInput ÷ 100,000,000,000 = input cost per token
```
jamie-rasmussen: I might use `/` instead of `÷` - that looked like a plus to my bad eyes.

weave/trace_server/costs/update_costs.py
```diff
@@ -77,6 +80,73 @@ def fetch_new_costs() -> dict[str, CostDetails]:
...
+            models_data = json.load(f)
+    except (json.JSONDecodeError, OSError) as e:
+        print(f"Failed to read modelsBegin.json: {e}")
+        return {}
+
+    costs: dict[str, CostDetails] = {}
+    current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+
+    for model in models_data:
+        if not isinstance(model, dict):
```
jamie-rasmussen, line 119: Probably worth at least logging warnings if these JSON shape checks fail instead of silently skipping - suggests something is very bad with the data file.

PR #5326, opened 2025-08-21
chore(weave): Add more efficient future handling for FutureExecutor

andrewtruong:
Changes the `_active_futures` container from list to set because we can get O(1) operations on set instead of O(N) operations on list


andrewtruong:
What do you mean duplicate identical futures?

Issue #5228, opened 2025-08-06
Making Sentry dependency optional


samad-wisdomlabs:
In my project, I have a custom version of the sentry sdk installed. This is creating conflicts because weave relies on sentry_sdk >= 2.0, causing conflicts with my project dependencies. Is there any way we can make the dependency on sentry optional? Thank you!
Created a rough PR for this (https://github.com/wandb/weave/pull/5230) as I plan to fork this for my project, but sharing here incase you are open to it.

andrewtruong:
Marking as complete from the PR above!

PR #5333, opened 2025-08-21
chore(weave): Dont crash if sentry is not installed

andrewtruong:
https://github.com/wandb/weave/issues/5228

We still have sentry on by default, but allow users to uninstall if they want

### DIFF
--- a/weave/telemetry/trace_sentry.py
+++ b/weave/telemetry/trace_sentry.py
@@ -17,11 +17,6 @@ import site
 import sys
 from typing import TYPE_CHECKING, Any, Callable, Literal
 
-if TYPE_CHECKING:
-    from sentry_sdk._types import Event, ExcInfo
-    from sentry_sdk.hub import Hub
-
-# Try to import sentry_sdk, but make it optional
 try:
     import sentry_sdk  # type: ignore
     import sentry_sdk.utils  # type: ignore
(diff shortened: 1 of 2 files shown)


weave/telemetry/trace_sentry.py
```diff
@@ -21,11 +21,25 @@
...
+except ImportError:
+    SENTRY_AVAILABLE = False
+    sentry_sdk = None  # type: ignore
+
+    # Create dummy types for when sentry is not available
+    class DummyHub:
+        pass
+
+    class DummyClient:
+        pass
```
bcsherma: Is this used?

weave/telemetry/trace_sentry.py
```diff
@@ -58,9 +72,12 @@ def __init__(self) -> None:
 
         self.dsn = SENTRY_DEFAULT_DSN
 
-        self.hub: sentry_sdk.hub.Hub | None = None
```
bcsherma, line 61: why be more permissive with the typing here?

weave/telemetry/trace_sentry.py
```diff
@@ -106,6 +126,9 @@ def exception(
         status: SessionStatus | None = None,
     ) -> None:
         """Log an exception to Sentry."""
+        if not SENTRY_AVAILABLE or self.hub is None:
```
chance-wnb: when SENTRY_AVAILABLE `_safe_noop` should have this converted to noop already. Is there still need to do a check at every method?

weave/telemetry/trace_sentry.py
andrewtruong: nope

weave/telemetry/trace_sentry.py
andrewtruong: hmm fair point.  I don't think it hurts though

weave/telemetry/trace_sentry.py
andrewtruong: Was trying to avoid type issues if sentry was not installed but I think it should be ok with postponed evaluation of annotations: https://peps.python.org/pep-0563/

Issue #1502, opened 2024-04-09
Add parallelism argument to Evaluation.evaluate


shawnlewis:
We currently have an environment variable WEAVE_PARALLELISM to control parallelism of Weave Evaluations. Let's add an argument to Evaluate.evaluate to do the same thing.

bdytx5:
Just wanted to revive this as I strongly agree it should be an argument. Additionally, if we could pass an argument like max_tokens_per_minute to prevent rate limiting, I would definitely use it. This would require the model to have a tokenizer, so possibly supplying a default llama tokenizer that would give a good rough estimate of token counts. Additionally adding some sort of built in retry logic to the Weave model class could be helpful. I've probably ran 20-30 different evals and have encountered rate limiting issues probably 75% of time.

PR #5341, opened 2025-08-21
feat(weave): Add parallelism argument to weave.Evaluation

andrewtruong:
https://github.com/wandb/weave/issues/1502

Exposes an argument to control parallelism


gtarpenning (approved):
A bit weird that it clobbers the env var, we might consider a warning if both are set

PR #5347, opened 2025-08-22
perf(weave): upload table in parallel with `table_create_from_digests`

gtarpenning:
## Description

[WB-27126](https://wandb.atlassian.net/browse/WB-27126)

This pr adds: 
- a `table_create_from_digests` endpoint that accepts `row_digests` instead of `rows` to create a table. A table is just a collection of row digests anyway, the endpoint implementation is trivially simple. Rows must already be inserted in the database, probably through the existing `table_create` endpoint. 
- method to create tables in parallel, then create one super-table with the new endpoint by passing in all the rows. 
- adds a settings/env var to turn this on

Must be back compatible, so clients with old servers are still functional. 

TODO:
- removes the ability to just hit the server.table_create endpoint had have it work for any size table, is that an issue? 

## Testing

adds test 

[WB-27126]: https://wandb.atlassian.net/browse/WB-27126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server_bindings/http_utils.py
+++ b/weave/trace_server_bindings/http_utils.py
@@ -269,7 +269,6 @@ def check_endpoint_exists(
     try:
         # Try calling the function with test request
         func(test_req)
-        endpoint_exists = True
     except Exception as e:
         # Check if this is a 404 (method not found)
         response = getattr(e, "response", None)
@@ -278,6 +277,8 @@ def check_endpoint_exists(
             endpoint_exists = status_code != 404
         else:
             endpoint_exists = False
+    else:
+        endpoint_exists = True
 
     if endpoint_exists:
         _ENDPOINT_CACHE.add(cache_key)


weave/trace/table_upload_chunking.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        if not rows:
+            return []
+
+        chunks = []
+        current_chunk: list[RowItemType] = []
+        current_chunk_bytes = 0
+
+        for row in rows:
+            row_bytes = self.calculate_row_bytes(row)
```
gtarpenning, line 55: inefficient

andrewtruong:
This seems like a pretty significant change -- I think it would be good to:
1. Keep the original implementation as backup;
2. Allow users to opt-in to this implementation via a setting, at least for now

weave/trace_server_bindings/http_utils.py
```diff
@@ -231,3 +238,46 @@ def handle_response_error(response: requests.Response, url: str) -> None:
         message = f"{response.status_code} Error for url {url}: Request failed"
 
     raise requests.HTTPError(message, response=response)
+
+
+def check_endpoint_exists(
```
andrewtruong, line 243: Can you just check the `openapi.json` that FastAPI exposes instead?

Checking if the endpoint exists could get you throttled

weave/trace/weave_client.py
```diff
@@ -1976,6 +1995,132 @@ def send_table_create() -> TableCreateRes:
 
         return table_ref
 
+    def _should_use_chunking(self, table: Table | WeaveTable) -> tuple[bool, bool]:
```
andrewtruong: nit: don't return a tuple, it's hard to understand.  Return a dataclass or at least a typeddict

weave/trace/weave_client.py
```diff
@@ -1976,6 +1995,132 @@ def send_table_create() -> TableCreateRes:
...
+        """Determine if we should use chunking and parallel chunks for a table."""
+        remote_request_bytes_limit = getattr(
+            self.server, "remote_request_bytes_limit", REMOTE_REQUEST_BYTES_LIMIT
+        )
+
+        # Primary heuristic: row count
+        use_chunking = len(table.rows) > ROW_COUNT_CHUNKING_THRESHOLD
```
andrewtruong, line 1628: may be better to have an early return for `len(table.rows) == 0`

weave/trace/weave_client.py
```diff
@@ -1976,6 +1995,132 @@ def send_table_create() -> TableCreateRes:
...
+                )
+
+            use_parallel_chunks = check_endpoint_exists(
+                test_func, test_req, "table_create_from_digests"
+            )
+
+        return use_chunking, use_parallel_chunks
+
+    def _create_table_with_parallel_chunks(
+        self, table: Table | WeaveTable
```
andrewtruong, line 1663: why is this `Table | WeaveTable`?

weave/trace_server_bindings/http_utils.py
gtarpenning: we can't because what if we have an updated client, but are hitting an old trace server, like in dedicated

weave/trace/weave_client.py
```diff
@@ -1976,6 +1995,132 @@ def send_table_create() -> TableCreateRes:
...
+                update_result = self.server.table_update(update_req)
+                current_digest = update_result.digest
+                all_row_digests.extend(update_result.updated_row_digests)
+
+            return TableCreateRes(
+                digest=current_digest,
+                row_digests=all_row_digests,
+            )
+
+        # Chain the sequential processing after the base table is created
```
andrewtruong, line 1746: Is the chaining here absolutely necessary?  Seems like it would be an obvious bottleneck for long tables

gtarpenning: this is the legacy implementation. and yes, its an obvious bottleneck

weave/trace_server_bindings/http_utils.py
andrewtruong: I see you're caching it -- I guess this works too

weave/trace_server_bindings/http_utils.py
```diff
@@ -11,6 +11,13 @@
 
 logger = logging.getLogger(__name__)
 
+# Global cache for endpoint availability to avoid repeated checks
+_ENDPOINT_CACHE: dict[str, bool] = {}
```
andrewtruong: this can just be a `set`.  Existence in the set is the same as `key: True`

gtarpenning:
The existing implementation is in the `remote_http_trace_server.py` `table_create` method, which is awkward because I need to call that exact endpoint for the new updated implementation. There is an existing legacy implementation, but I agree toggling this with a env var would be safer.

weave/trace/weave_client.py
gtarpenning: follows the existing implementation of _save_table

weave/trace_server_bindings/http_utils.py
```diff
@@ -231,3 +238,47 @@ def handle_response_error(response: requests.Response, url: str) -> None:
...
+        cache_key = str(id(func))
+
+    # Check cache first
+    if cache_key in _ENDPOINT_CACHE:
+        return True
+
+    try:
+        # Try calling the function with test request
+        func(test_req)
+        endpoint_exists = True
```
andrewtruong: nit: move to else block

andrewtruong:
Yay perf!

PR #5350, opened 2025-08-22
docs(weave): added clickhouse altinity doc

LalitSahuWnB:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?

### DIFF
--- a/docs/docs/guides/platform/weave-self-managed.md
+++ b/docs/docs/guides/platform/weave-self-managed.md
@@ -17,7 +17,7 @@ Although the deployment process sets up a fully functional ClickHouseDB instance
 ## Requirements
 
 - W&B Platform installed. For more information, see the [Self-Managed Deployment Guide](https://docs.wandb.ai/guides/hosting/hosting-options/self-managed/).
-- [Bitnami's ClickHouse Helm Chart](https://github.com/bitnami/charts/tree/main/bitnami/clickhouse).
+- [Altinity Kubernetes Operator for ClickHouse](https://docs.altinity.com/altinitykubernetesoperator).
 - An S3 bucket pre-configured for ClickHouse storage. For configuration details, see [Provide S3 Credentials](#provide-s3-credentials).
 - Kubernetes Cluster Nodes with the following specifications:
   - CPU: 8 cores  
@@ -29,13 +29,13 @@ Although the deployment process sets up a fully functional ClickHouseDB instance
 For a detailed reference architecture, see [https://docs.wandb.ai/guides/hosting/self-managed/ref-arch/](https://docs.wandb.ai/guides/hosting/self-managed/ref-arch/#models-and-weave).
 :::
 
-## 1. Configure ClickHouse
+## 1. Deploy ClickHouse
 
-The ClickHouse deployment in this document uses the [Bitnami ClickHouse](https://bitnami.com/stack/clickhouse) package.
+The ClickHouse deployment in this document uses the [Altinity Kubernetes Operator for ClickHouse](https://docs.altinity.com/altinitykubernetesoperator).
 
-The Bitnami Helm chart provides good support for basic ClickHouse functionalities, particularly the use of [ClickHouse Keeper](https://clickhouse.com/docs/en/guides/sre/keeper/clickhouse-keeper).
+Altinity uses operator to install, configure, and run a ClickHouse cluster in a Kubernetes cluster, including how to set up persistent storage and replication. Be aware that the GitHub repo for the project (https://github.com/Altinity/clickhouse-operator) has in-depth technical details (https://github.com/Altinity/clickhouse-operator/tree/master/docs#table-of-contents) for a number of advanced configuration.
 
-To configure Clickhouse, complete the following steps:
+To Install Altinity Operator for clickhouse follow the below steps:
 
 1. [Configure the Helm repository](#configure-helm-repository)
 2. [Create Helm Configuration](#create-helm-configuration)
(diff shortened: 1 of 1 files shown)


docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -21,282 +21,383 @@ Although the deployment process sets up a fully functional ClickHouseDB instance
...
+          # affinity:
+          #   podAntiAffinity:
+          #     requiredDuringSchedulingIgnoredDuringExecution:
+          #       - labelSelector:
+          #           matchExpressions:
+          #             - key: "app"
+          #               operator: In
+          #               values:
+          #                 - clickhouse-keeper
+          #         topologyKey: "kubernetes.io/hostname"
```
flamarion: I recommend uncommenting this. It will ensure the pods are spread across the cluster and not in the same node

docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -21,282 +21,383 @@ Although the deployment process sets up a fully functional ClickHouseDB instance
...
+          # affinity:
+          #   podAntiAffinity:
+          #     requiredDuringSchedulingIgnoredDuringExecution:
+          #       - labelSelector:
+          #           matchExpressions:
+          #             - key: "app"
+          #               operator: In
+          #               values:
+          #                 - clickhouse-server
+          #         topologyKey: "kubernetes.io/hostname"
```
flamarion: Same here to spread the pods across the nodes.

docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -21,282 +21,383 @@ Although the deployment process sets up a fully functional ClickHouseDB instance
...
+  configuration:
+    # Refer to: https://clickhouse.com/docs/operations/server-configuration-parameters/settings#zookeeper
+    zookeeper:
+      nodes:
+        - host: chk-wandb-keeper-0-0.clickhouse.svc.cluster.local
+          port: 2181
+        - host: chk-wandb-keeper-0-1.clickhouse.svc.cluster.local
+          port: 2181
+        - host: chk-wandb-keeper-0-2.clickhouse.svc.cluster.local
+          port: 2181
```
flamarion: These names depend on the name assigned to the keepr deployment, so we could generalize them or add a comment to it.

docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -21,282 +21,383 @@ Although the deployment process sets up a fully functional ClickHouseDB instance
...
+                  <host>chi-wandb-clickhouse-0-0.clickhouse.svc.cluster.local</host>
+                  <port>9000</port>
+                </replica>
+                <replica>
+                  <host>chi-wandb-clickhouse-0-1.clickhouse.svc.cluster.local</host>
+                  <port>9000</port>
+                </replica>
+                <replica>
+                  <host>chi-wandb-clickhouse-0-2.clickhouse.svc.cluster.local</host>
+                  <port>9000</port>
```
flamarion: This also depends on the name assigned to the deployment; maybe it would be good to generalize or explain why these names are like this.

docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -21,282 +21,383 @@ Although the deployment process sets up a fully functional ClickHouseDB instance
...
+            </weave_cluster>
+          </remote_servers>
+        </clickhouse>
+      config.d/storage_configuration.xml: |
+        <clickhouse>
+            <storage_configuration>
+                <disks>
+                    <s3_disk>
+                        <type>s3</type>
+                        <endpoint>https://storage.googleapis.com/wandbench-ch-1/s3_disk/{replica}</endpoint>
```
flamarion: Remove the `s3_disk` from the path; this will only add more complexity. 
Using the root path is simpler, and each replica will create its own folder.

docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -307,15 +408,10 @@ clickhouse-shard0-0                  1/1     Running   0          9m59s
 clickhouse-shard0-1                  1/1     Running   0          10m
 clickhouse-shard0-2                  1/1     Running   0          10m
```
flamarion: Are these still the pod names created by Altinity?
I don't remember, to be honest.

flamarion (changes requested):
I think the first iteration is okay. I'll review the dev branch and provide more comments.
Thank you for putting this together.

docs/docs/guides/platform/weave-self-managed.md
LalitSahuWnB: uncommented as requested.

docs/docs/guides/platform/weave-self-managed.md
LalitSahuWnB: uncommented as requested.

docs/docs/guides/platform/weave-self-managed.md
LalitSahuWnB: updated

docs/docs/guides/platform/weave-self-managed.md
LalitSahuWnB: updated

docs/docs/guides/platform/weave-self-managed.md
LalitSahuWnB: updated with the correct pods that i got created.

docs/docs/guides/platform/weave-self-managed.md
LalitSahuWnB: added this section with comments.

dbrian57 (approved):
Edits made and approved.

docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -1,321 +1,424 @@
...
+  configuration:
+    # Refer to: https://clickhouse.com/docs/operations/server-configuration-parameters/settings#zookeeper
+    zookeeper:
+      #nodes:
+      #  - host: chk-wandb-keeper-0-0-0.clickhouse.svc.cluster.local # Set zookeeper node endpoints
+      #    port: 2181
+      #  - host: chk-wandb-keeper-0-1-0.clickhouse.svc.cluster.local # Set zookeeper node endpoints
+      #    port: 2181
+      #  - host: chk-wandb-keeper-0-2-0.clickhouse.svc.cluster.local # Set zookeeper node endpoints
+      #    port: 2181
```
flamarion: Although it's the Zookeeper section, this will actually configure Keeper, so it must be uncommented.
Please refer to the config file I created and Jeff uploaded to Notion

flamarion (changes requested):

docs/docs/guides/platform/weave-self-managed.md
```diff
@@ -367,13 +470,13 @@ When using more than one replica (W&B recommend a least 3 replicas), ensure to h
 extraEnv:
   WF_CLICKHOUSE_REPLICATED: "true"
 ```
-This has the same effect of `replicated: true` which in preview.
+This has the same effect as `replicated: true`.
 :::
 
 
-3. Set the `clusterName` in `values.yaml` to `weave_cluster`. If it is not, the database migration will fail.  
+3. Set the `clusterName` in `values.yaml` to `weave_cluster`. The database migration fails if you do not set this value correctly.
```
flamarion, line 477: In this case, the cluster name will be defined by this [line](https://github.com/wandb/weave/pull/5350/files#diff-488e133752a995c0a14c732d0b4514b73c05ef2a5a2191ad80452e3b2f683201R317)

flamarion (changes requested):

nickpenaranda (approved):

docs/docs/guides/platform/weave-self-managed.md
LalitSahuWnB: uncommented

PR #5359, opened 2025-08-25
chore(weave): Track memory consumption in tests

andrewtruong:


noxfile.py
```diff
@@ -169,6 +169,15 @@ def tests(session, shard):
         "--cov-branch",
     ]
 
+    # Memray not working with trace_server shard atm
```
gtarpenning, line 172:
```suggestion
    # Memory not working with trace_server shard atm
```

PR #5364, opened 2025-08-25
chore(weave): Add entity/project init style similar to wandb

andrewtruong:
https://wandb.atlassian.net/browse/WB-27237

Supports passing an explicit entity and project


adrnswanberg (approved):
Fully backwards compatible right?

neutralino1:
@andrewtruong do we want to merge or close this?

PR #5366, opened 2025-08-25
feat(weave): OpenAI realtime support

zbirenbaum:
## Description

- Fixes WB-26886

Adds SDK support for OpenAI realtime API

Includes a number of autopatchers for common websockets libraries as well as the ability to instrument your own websocket and disable the autopatcher.

Couple of notes before this goes into review:
1. Types - Any is used a lot, I have a good set of types, they need to be integrated before this goes live
2. Testing - Testing this in CI is a bit difficult. I have a large number of integration tests but they don't always accurately reflect live behavior
3. Edge cases - Event order is implementation specific, the best solution I could come up with to get transcripts as much as possible was to add a debounce. There are a couple of other scenarios like this which may come up
4. Sessions.py is too long, probably want to split this up somehow

### DIFF
--- a/weave/integrations/openai_realtime/__init__.py
+++ b/weave/integrations/openai_realtime/__init__.py
@@ -1,0 +1,16 @@
+# OpenAI Realtime API integration for Weave
+# This module automatically patches websocket connections to track sessions and conversations
+
+from .openai_realtime_sdk import (
+    OpenAIRealtimeSettings,
+    wrap_websocket_async,
+    wrap_websocket_connect,
+    wrap_websocket_sync,
+)
+
+__all__ = [
+    "OpenAIRealtimeSettings",
+    "wrap_websocket_async",
+    "wrap_websocket_connect",
+    "wrap_websocket_sync",
+]
--- a/weave/trace/autopatch.py
+++ b/weave/trace/autopatch.py
@@ -50,6 +50,7 @@ class AutopatchSettings(BaseModel):
     notdiamond: IntegrationSettings = Field(default_factory=IntegrationSettings)
     openai: IntegrationSettings = Field(default_factory=IntegrationSettings)
     openai_agents: IntegrationSettings = Field(default_factory=IntegrationSettings)
+    openai_realtime: IntegrationSettings = Field(default_factory=IntegrationSettings)
     vertexai: IntegrationSettings = Field(default_factory=IntegrationSettings)
     chatnvidia: IntegrationSettings = Field(default_factory=IntegrationSettings)
     smolagents: IntegrationSettings = Field(default_factory=IntegrationSettings)
@@ -89,6 +90,9 @@ def autopatch(settings: Optional[AutopatchSettings] = None) -> None:
     from weave.integrations.notdiamond.tracing import get_notdiamond_patcher
     from weave.integrations.openai.openai_sdk import get_openai_patcher
     from weave.integrations.openai_agents.openai_agents import get_openai_agents_patcher
+    from weave.integrations.openai_realtime.openai_realtime_sdk import (
+        get_websocket_patcher,
+    )
     from weave.integrations.smolagents.smolagents_sdk import get_smolagents_patcher
     from weave.integrations.verdict.verdict_sdk import get_verdict_patcher
     from weave.integrations.vertexai.vertexai_sdk import get_vertexai_patcher
@@ -112,6 +116,7 @@ def autopatch(settings: Optional[AutopatchSettings] = None) -> None:
     get_huggingface_patcher(settings.huggingface).attempt_patch()
     get_smolagents_patcher(settings.smolagents).attempt_patch()
     get_openai_agents_patcher(settings.openai_agents).attempt_patch()
+    get_websocket_patcher(settings.openai_realtime).attempt_patch()
     get_verdict_patcher(settings.verdict).attempt_patch()
 
     langchain_patcher.attempt_patch()
@@ -145,6 +150,9 @@ def reset_autopatch() -> None:
     from weave.integrations.notdiamond.tracing import get_notdiamond_patcher
     from weave.integrations.openai.openai_sdk import get_openai_patcher
     from weave.integrations.openai_agents.openai_agents import get_openai_agents_patcher
+    from weave.integrations.openai_realtime.openai_realtime_sdk import (
+        get_websocket_patcher,
+    )
     from weave.integrations.smolagents.smolagents_sdk import get_smolagents_patcher
     from weave.integrations.verdict.verdict_sdk import get_verdict_patcher
     from weave.integrations.vertexai.vertexai_sdk import get_vertexai_patcher
@@ -168,6 +176,7 @@ def reset_autopatch() -> None:
     get_huggingface_patcher().undo_patch()
     get_smolagents_patcher().undo_patch()
     get_openai_agents_patcher().undo_patch()
+    get_websocket_patcher().undo_patch()
     get_verdict_patcher().undo_patch()
 
     langchain_patcher.undo_patch()
--- a/weave/integrations/openai_realtime/connection.py
+++ b/weave/integrations/openai_realtime/connection.py
@@ -1,0 +1,261 @@
+import json
+import uuid
+from typing import Any
+
+import websocket
+
+import weave
+
+from .sessions import SessionManager
+
+try:
+    from aiohttp import ClientWebSocketResponse, WSMsgType
+except ImportError:
+    ClientWebSocketResponse = None
+    WSMsgType = None
+
+
+def _try_json_load(data: Any) -> Any:
+    if isinstance(data, str):
+        try:
+            return json.loads(data)
+        except json.JSONDecodeError:
+            return data
+    if isinstance(data, bytes):
+        try:
+            return json.loads(data.decode())
+        except (UnicodeDecodeError, json.JSONDecodeError):
+            return data
+    return data
+
+
+class WeaveMediaConnection:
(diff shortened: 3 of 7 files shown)


weave/integrations/openai_realtime/__init__.py
```diff
@@ -0,0 +1,16 @@
+# OpenAI Realtime API integration for Weave
+# This module automatically patches websocket connections to track sessions and conversations
+
+from .openai_realtime_sdk import (
```
neutralino1: Full import path please?

neutralino1 (commented):
Amazing work.

You mention you have tests but I don't see them.

weave/trace/autopatch.py
```diff
@@ -89,6 +90,9 @@ def autopatch(settings: Optional[AutopatchSettings] = None) -> None:
     from weave.integrations.notdiamond.tracing import get_notdiamond_patcher
     from weave.integrations.openai.openai_sdk import get_openai_patcher
     from weave.integrations.openai_agents.openai_agents import get_openai_agents_patcher
+    from weave.integrations.openai_realtime.openai_realtime_sdk import (
+        get_websocket_patcher,
```
andrewtruong: this will change soon, but for now I recommend calling this specifically the `openai_realtime_websocket_patcher`

weave/integrations/openai_realtime/audio_buffer.py
```diff
@@ -0,0 +1,51 @@
...
+
+    sample_rate_hz: int = 24000
+    bits_per_sample: int = 16
+    channels: int = 1
+    buffer: bytearray = field(default_factory=bytearray)
+
+    def bytes_per_sample(self) -> int:
+        return (self.bits_per_sample // 8) * self.channels
+
+    def append_base64(self, b64: str) -> None:
```
andrewtruong: shouldn't this be `extend`?

weave/trace/autopatch.py
zbirenbaum: Just to clarify:

~You mean to change `get_websocket_patcher` -> `openai_realtime_websocket_patcher` correct?~
Changed to 
```
from weave.integrations.openai_realtime import get_openai_realtime_websocket_patcher
```

weave/integrations/openai_realtime/audio_buffer.py
zbirenbaum: Good call, updated

weave/integrations/openai_realtime/models.py
```diff
@@ -0,0 +1,931 @@
...
+SessionID = Annotated[str, AfterValidator(is_sess_id)]
+EventID = Annotated[str, AfterValidator(is_event_id)]
+ItemID = Annotated[str, AfterValidator(is_item_id)]
+ResponseID = Annotated[str, AfterValidator(is_resp_id)]
+CallID = Annotated[str, AfterValidator(is_call_id)]
+ConversationID = Annotated[str, AfterValidator(is_conv_id)]
+
+
+Voice = Literal["alloy", "shimmer", "echo"]
+AudioFormat = Literal["pcm16", "g711-ulaw", "g711-alaw"]
```
andrewtruong: please add docs for where these values came from!

weave/integrations/openai_realtime/models.py
andrewtruong, line 1: Are many of these types coming from OpenAI?  Why not use their actual types?

weave/integrations/openai_realtime/models.py
```diff
@@ -0,0 +1,931 @@
...
+    "response.audio.done": ResponseAudioDoneMessage,
+    "response.function_call_arguments.delta": ResponseFunctionCallArgumentsDeltaMessage,
+    "response.function_call_arguments.done": ResponseFunctionCallArgumentsDoneMessage,
+    "rate_limits.updated": RateLimitsUpdatedMessage,
+}
+
+MessageType = Union[UserMessageType, ServerMessageType]
+
+
+def create_user_message_from_dict(data: dict) -> UserMessageType:
```
andrewtruong: wouldn't it be easier to use a discriminated union for all of these classes?
https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators

weave/integrations/openai_realtime/models.py
```diff
@@ -0,0 +1,931 @@
...
+    return UnknownClientMessage(**data)
+
+
+# Whenever we get a response we have to acquire a lock and associate it with the last conversation.item.recated item
+# Whenever we get a delta event we have to acquire a lock
+
+
+def has_item_id(message: MessageType) -> bool:
+    """Check if a message has an item ID, directly or nested on item.id."""
+    if hasattr(message, "item_id") and message.item_id is not None:
```
andrewtruong: nit: This can be a `if getattr(message, "item_id", None) is not None`

weave/integrations/openai_realtime/openai_realtime_websocket_patcher.py
```diff
@@ -0,0 +1,192 @@
...
+import logging
+from functools import wraps
+from typing import TYPE_CHECKING, Any, Callable
+
+from weave.integrations.openai_realtime import connection
+from weave.integrations.patcher import MultiPatcher, NoOpPatcher, SymbolPatcher
+from weave.trace.autopatch import IntegrationSettings
+
+if TYPE_CHECKING:
+    pass
```
andrewtruong: remove

weave/integrations/openai_realtime/models.py
zbirenbaum: These functions were removed since they aren't really necessary

weave/integrations/openai_realtime/models.py
zbirenbaum: I attempted this, it's definitely not possible - Realtime is in super active development and OpenAI changes the types slightly on a weekly basis.

~~Essentially if we use official types we are praying that all of these are 100% up to date:~~
~~1. The OpenAI Realtime Server Endpoint~~
~~2. The user's client messages~~
~~3. The version of the SDK we got those types from~~

~~The user has no obligation to use the SDK with websockets and many clients will be written without it. Even if they have the package installed their messages likely will not 100% match the types in the SDK~~

Edit: OpenAI finally got around to updating their docs (they didn't when they took realtime out of beta) and it turns out the reason I couldn't get their types to work is because of a header flag that was previously required to function and now is not:

With Flag: You get messages matching beta API
Without Flag: You get messages matching the release API

I'll have to change this but for now lets get this in since it works for the beta release.

weave/integrations/openai_realtime/models.py
zbirenbaum: This would be a followup. I'm going to hold off on non-functional changes until this goes in.

weave/integrations/openai_realtime/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+# This module automatically patches websocket connections to track sessions and conversations
+from weave.integrations.openai_realtime.openai_realtime_websocket_patcher import (
+    OpenAIRealtimeSettings,
+    get_openai_realtime_websocket_patcher,
+    wrap_websocket_async,
+    wrap_websocket_connect,
+    wrap_websocket_sync,
+)
+
+__all__ = [
```
tssweeney, line 11: are all of these truly "external" APIs? for example `wrap_*`?

weave/integrations/openai_realtime/audio_buffer.py
```diff
@@ -0,0 +1,50 @@
...
+    buffer: bytearray = field(default_factory=bytearray)
+
+    def bytes_per_sample(self) -> int:
+        return (self.bits_per_sample // 8) * self.channels
+
+    def extend_base64(self, b64: str) -> None:
+        """Decode base64 audio and extend buffer. Ignores invalid payloads gracefully."""
+        try:
+            self.buffer.extend(base64.b64decode(b64))
+        except Exception as e:
```
tssweeney: bare exceptions are risky. If you truly don't know the exception you are looking for, please just wrap the decode in the try-catch to limit the over-exposure

weave/integrations/openai_realtime/models.py
```diff
@@ -0,0 +1,902 @@
+from __future__ import annotations
```
tssweeney, line 1: i would add a comment to the header of this file describing why you had to do this and probably mentioning that it should not change / stay in sync with openai. i'd also recommend not calling it `models` due to the overloaded nature of that word in our app

weave/integrations/openai_realtime/models.py
zbirenbaum: Added docs citation and updated voice list

weave/integrations/openai_realtime/__init__.py
zbirenbaum: Yes, those are the methods for manual patching
```
        import websockets
        from weave.integrations.openai_realtime import wrap_websocket_async

        async def connect():
            ws = await websockets.connect(url)
            tracked_ws = wrap_websocket_async(ws)
            return tracked_ws
```

weave/integrations/openai_realtime/audio_buffer.py
zbirenbaum: addressed

weave/integrations/openai_realtime/models.py
zbirenbaum: It's likely to change/be removed at some point. Especially once OpenAI manages to stabilize their API types and fix some stuff. I'm going to leave it as is for now but I'll add a comment

PR #5372, opened 2025-08-26
chore(weave): Respect WF_TRACE_SERVER_URL when printing links

andrewtruong:
https://wandb.atlassian.net/browse/WB-18776

This will show correct call links when this env var is set

### DIFF
--- a/weave/trace/urls.py
+++ b/weave/trace/urls.py
@@ -1,4 +1,5 @@
-from urllib.parse import quote
+import os
+from urllib.parse import quote, urlparse
 
 from weave.trace import env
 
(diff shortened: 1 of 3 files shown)


weave/trace/urls.py
```diff
@@ -32,4 +33,21 @@ def leaderboard_path(entity_name: str, project_name: str, object_name: str) -> s
...
+        # Default to localhost:9000 if parsing fails or no host is found
+        if parsed.hostname:
+            base_url = f"{parsed.scheme or 'http'}://{parsed.hostname}"
+            if parsed.port:
+                base_url += f":{parsed.port}"
+        else:
+            # Fallback to localhost:9000 if we can't parse the URL properly
+            base_url = "http://localhost:9000"
+        return f"{base_url}/{entity_name}/{quote(project_name)}/{WEAVE_SLUG}/r/call/{call_id}"
+
```
neutralino1: I see this
```python
def remote_project_root_url(entity_name: str, project_name: str) -> str:
    return f"{wandb.app_url(env.wandb_base_url())}/{entity_name}/{quote(project_name)}"
```

why isn't this override done inside `env.wandb_base_url()`?

andrewtruong: I don't think that's part of this change.  Per skimming the code, I think it's been set like this for about a year.

I can take a closer look

neutralino1:
@jamie-rasmussen do you still want to push this through? Or close it?

weave/trace/urls.py
```diff
@@ -1,18 +1,27 @@
...
 WEAVE_SLUG = "weave"
 
 
-def remote_project_root_url(entity_name: str, project_name: str) -> str:
-    return f"{wandb.app_url(env.wandb_base_url())}/{entity_name}/{quote(project_name)}"
+def project_weave_root_url(entity: str, project: str) -> str:
+    """Get the frontend URL for viewing a Weave project.
 
+    Args:
+        entity: The entity/organization name
```
jamie-rasmussen, line 13: I don't think organization name would work, just entity.

jamie-rasmussen (approved):

neutralino1:
@andrewtruong do you still want to push this through? Or close it?

PR #5373, opened 2025-08-26
fix(weave): Don't allow project to be empty in init

andrewtruong:
https://wandb.atlassian.net/browse/WB-24532

Disallows empty string


jamie-rasmussen:
Passing an empty string should probably be an error, but I think we should consider having `weave.init()` (with a default `None` for entity/project) use the user's default entity and a hardcoded project name, just to allow new users to get started with something a tiny bit simpler. This is the path we're going with Inference.

PR #5376, opened 2025-08-26
perf(weave): Add explicit patching

andrewtruong:
Adds a system to remove auto-patching and explicitly enable patching (opt-in) for specific integrations.  Previously, we let the user opt-out by default, which eagerly loaded a lot of dependencies and significantly slowed down both import and init time

A fast-follow PR adds back implicit patching, in a slightly different way, that maintains these perf improvements.
https://github.com/wandb/weave/pull/5377

## Benchmarking
script:
```py
import weave
weave.init('testing')
```

## Before
image

## After
image


examples/tutorial_scripts/01_quickstart.py
```diff
@@ -26,5 +26,9 @@ def extract_fruit(sentence: str) -> dict:
 
 
 weave.init("intro-example")
+
+# Enable OpenAI integration tracing
+weave.integrations.patch_openai()
```
gtarpenning, line 31: this is mandatory?

PR #5377, opened 2025-08-26
feat(weave): Implicit patching

andrewtruong:
This adds back implicit patching in a more performant-by-default way.

By default, all libraries are implicitly patched.
1. You can disable implicit patching entirely with `WEAVE_IMPLICITLY_PATCH_INTEGRATIONS=false`
2. Unlike before, you CANNOT "implicitly patch all except XYZ".
    1. If you want this behaviour, you can just explicitly patch the packages you want.


weave/integrations/patch.py
```diff
@@ -201,18 +233,219 @@ def patch_autogen(settings: Optional[IntegrationSettings] = None) -> None:
...
+    """Check sys.modules and automatically patch any already-imported integrations.
+
+    This function is called during weave.init() to enable implicit patching.
+    If a library is already imported when weave.init() is called, we automatically
+    patch it without requiring an explicit patch_X() call.
+
+    This respects the implicitly_patch_integrations setting - if disabled, no automatic
+    patching will occur.
+    """
+    from weave.trace.settings import should_implicitly_patch_integrations
```
gtarpenning, line 397: circular dependency otherwise?

gtarpenning (approved):
lgtm, good one to test thoroughly though

PR #5379, opened 2025-08-26
chore(weave): Move Op protocol into its own file

andrewtruong:
This helps improve import times


gtarpenning (approved):
i'm okay with this, but it does make the sdk a bit more complex... Did you consider defining `Op` protocol like `Op_` to make it obvious it's something different? is `from weave.trace.op_protocol import Op` the same as `from weave import Op`? Sort of a big change. Does this only affect internal code or would this affect user scripts that have imported `Op`?

andrewtruong:
User code should not import `Op`.  There are convenience methods for interacting with this protocol at a higher level in the form of `is_op` and `as_op`

PR #5380, opened 2025-08-26
chore(weave): Dont import from top level weave in submodules

andrewtruong:
- **test**
- **evaluation**
- **imperative_eval**
- **scorer**


weave/dataset/dataset.py
```diff
@@ -140,6 +140,8 @@ def add_rows(self, rows: Iterable[dict]) -> "Dataset":
         Returns:
             The updated dataset.
         """
+        import weave
```
gtarpenning, line 143: why do we need to import here?

andrewtruong: weave.publish is at the top level -- doing this for convenience while I refactor the other parts

PR #5385, opened 2025-08-27
fix(weave): Don't mutate inputs when a postprocessor is specified

andrewtruong:
https://wandb.atlassian.net/browse/WB-27193


neutralino1:
@andrewtruong merge or close?

PR #5388, opened 2025-08-27
chore(weave): Fix code capture for import as xyz case

andrewtruong:
https://wandb.atlassian.net/browse/WB-19579

Previously code would show up as
```py
from module import nameas n
```

This fixes it to
```py
from module import name as n
```


tests/trace/test_op_versioning.py
```diff
@@ -553,3 +553,57 @@ def t(text: str):
...
+    def func(data: dict) -> str:
+        return js.dumps(data)
+
+    func({"a": 1})
+
+    ref = as_op(func).ref
+    assert ref is not None
+
+    saved_code = get_saved_code(client, ref)
+    print(saved_code)
```
jamie-rasmussen, line 580: Did you want to leave these `print` statements in?

andrewtruong: That seems to be how the other tests are written

PR #5391, opened 2025-08-28
feat(weave): Add wandb integration

andrewtruong:
https://wandb.atlassian.net/browse/WB-27151

If a wandb run is active, init weave to that project on import


weave/integrations/wandb/wandb.py
```diff
@@ -0,0 +1,26 @@
...
+from weave.trace.api import init
+
+logger = logging.getLogger(__name__)
+
+
+def wandb_init_hook() -> None:
+    # Try to get the active run path from wandb if we can
+    try:
+        from wandb.integration.weave import active_run_path
+    except Exception:
```
timoffex: Do we expect anything other than an `ImportError`?

weave/integrations/wandb/wandb.py
```diff
@@ -0,0 +1,26 @@
...
+        return
+    if not (run_path := active_run_path()):
+        return
+    if not (entity := run_path.entity):
+        return
+    if not (project := run_path.project):
+        return
```
timoffex: Hmm, let me improve the `wandb` side here. Let's have it be so that if `active_run_path()` returns a truthy value, then `entity` and `project` are both nonempty.

weave/integrations/wandb/wandb.py
```diff
@@ -0,0 +1,26 @@
...
+
+    project_name = f"{entity}/{project}"
+    logger.info(
+        f"Active wandb run detected.  Using project name from wandb: {project_name}"
```
timoffex: nit: not sure about the convention you use in the repo, but I generally expect a single space rather than two between sentences

tests/integrations/wandb/wandb_init_hook_test.py
```diff
@@ -0,0 +1,75 @@
...
+            monkeypatch.setitem(sys.modules, name, mod)
+
+        # Patch weave.integrations.wandb.wandb.init to record calls
+        target_module = importlib.import_module("weave.integrations.wandb.wandb")
+        calls = []
+
+        def fake_init(project_name=None, **_):
+            calls.append(project_name)
+
+        monkeypatch.setattr(target_module, "init", fake_init, raising=True)
```
timoffex: I believe you can just do

```py
monkeypatch.setattr("weave.integrations.wandb.wandb.init", fake_init)
```

(also, `raising` defaults to True)

But using a qualified import inside the module and monkeypatching the real symbol is better practice:

```py
monkeypatch.setattr("weave.trace.api.init", fake_init)
```

weave/integrations/wandb/wandb.py
andrewtruong: The most specific are probably `(ImportError, AttributeError)`, but since this in the package import path, it should never error.

weave/integrations/wandb/wandb.py
andrewtruong: fixed

weave/integrations/wandb/wandb.py
timoffex: Maybe an `AttributeError` is possible, not sure

tests/integrations/wandb/wandb_init_hook_test.py
```diff
@@ -0,0 +1,74 @@
...
+        ),
+        TestCase(
+            name="active_run",
+            active_run=RunPath(entity="ent", project="proj"),
+            expected_calls=["ent/proj"],
+        ),
+        # TODO: What if there are multiple active runs?
+    ],
+    ids=lambda tc: tc.name,
+)
```
timoffex: fyi: you can parametrize multiple parameters at the same time

```py
@pytest.mark.parametrize(
    "name,active_run,expected_calls",
    (
        ("no_active_run", None, []),
        ...,
    ),
)
```

That way you don't need a TestCase class

andrewtruong: yeah, but I find the tuple style less readable

weave/integrations/wandb/wandb.py
andrewtruong: Added more specific handling

jamie-rasmussen (approved):

neutralino1:
This is nice.

Can we have something similar so that users do not need to call `init` on both libraries? Even a if a run is not active yet.

If a user has already called `wandb.init`, they should not need to call `weave.init`.

PR #5395, opened 2025-08-28
chore(weave): Add nightly tests

andrewtruong:
Adds tests for all supported versions to the nightly suite


andrewtruong:
@gtarpenning https://github.com/wandb/weave/actions/workflows/nightly-tests.yaml

The scorers have been failing for a while.  We should setup a notification for failing shards nightly

PR #5403, opened 2025-08-28
fix(weave): Patch non-beta OpenAI methods

jamie-rasmussen:
## Description

In OpenAI [v1.92.0](https://github.com/openai/openai-python/releases/tag/v1.92.0) the `parse` method was [moved out of beta](https://github.com/openai/openai-python/commit/0e358ed66b317038705fb38958a449d284f3cb88).

Our OpenAI integration was only checking for this parse method where it no longer existed, so parse calls were not getting tracked at all.

After:

Trace exists with token usage.
image

## Testing

How was this PR tested?

### DIFF
--- a/weave/integrations/openai/openai_sdk.py
+++ b/weave/integrations/openai/openai_sdk.py
@@ -752,10 +752,10 @@ def get_openai_patcher(
         update={"name": base.name or "openai.chat.completions.create"}
     )
     completions_parse_settings = base.model_copy(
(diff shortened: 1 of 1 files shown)


gtarpenning (approved):
do we need back compat?

jamie-rasmussen:
> do we need back compat?

I did not delete the attempts to patch it in the previous beta location, so that should be fine. (When symbols are not found our patching just silently moves onwards.)

PR #5405, opened 2025-08-29
feat(weave): auto convert base64 to content

zbirenbaum:
## Description

- Fixes WB-27306

This PR injects handling into the trace server to automatically detect base64 strings and URIs and converts them to content. This provides a few advantages. 
1. No more large base64 strings in ClickHouse. Everything will go into bucket storage when available.
2. Provides a path forward for easy multimedia upload and response support in Playground
3. Customers should never see 'trace too large' again due to a media response from LLMs

This also adds a method to convert content back to the original base64 string/URI but it is currently unused since it is not entirely clear that this is the best place to perform that conversion for efficiency reasons. Open to discussion on how to address this

## Testing

Tested with a script. Needs a more rigorous integration test.

### DIFF
--- a/weave/trace_server/base64_content_conversion.py
+++ b/weave/trace_server/base64_content_conversion.py
@@ -297,62 +297,80 @@ def reconstruct_base64_from_refs(
     Returns:
         Processed value with content references replaced by base64 where applicable
     """
+    import json
 
     def _visit(val: Any) -> Any:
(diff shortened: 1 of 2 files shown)


neutralino1 (commented):
Is this a duplicate of #5410 ?

zbirenbaum:
Improved and merged via #5410

PR #5413, opened 2025-09-02
fix(weave): Fix code saving when weave is imported as an alias

andrewtruong:
https://wandb.atlassian.net/browse/WB-19577

Code saving now handles `weave` import aliases properly

## After
```py
import weave as wv

@wv.op
def func(): ...
```

## Before
```py
import weave as wv

@weave.op
@wv.op
def func(): ...
```


weave/trace/serialization/op_type.py
```diff
@@ -548,18 +548,50 @@ def save_instance(obj: Op, artifact: MemTraceFilesArtifact, name: str) -> None:
...
+    if has_expected_decorator:
+        # Normalize existing decorator to have parentheses
+        op_function_code = re.sub(
+            rf"@{re.escape(decorator_prefix)}\.op(?!\()",
+            f"@{decorator_prefix}.op()",
+            op_function_code,
+        )
+    elif weave_alias and WEAVE_OP_PATTERN.search(op_function_code):
+        # Has @weave.op but import uses alias - normalize to @weave.op()
         op_function_code = WEAVE_OP_NO_PAREN_PATTERN.sub(
```
chance-wnb: Is this doing the same thing as `# Normalize existing decorator to have parentheses` above?

andrewtruong: Added a comment to clarify

chance-wnb: Then I believe it is doing the same thing. I think line 579 to 585 can be removed. Please correct me if this is wrong.

tests/trace/test_anonymous_ops.py
andrewtruong, line 1: These test hash changes are related to formatting

Previously the code saved was

```python
import weave


@weave.op
def func(): ...
```

Now it's just 1 newline

```py
import weave

@weave.op
def func(): ...
```

weave/trace/serialization/op_type.py
andrewtruong: Yep makes sense

PR #5432, opened 2025-09-05
feat(weave): Add WEAVE_SILENT flag similar to wandb

andrewtruong:


neutralino1:
@andrewtruong merge or close?

PR #5435, opened 2025-09-08
fix(weave): Always use user's supplied output for imperative eval, even if model has a predict method

andrewtruong:
https://wandb.atlassian.net/browse/WB-27128

Previously, if the `Model` had a defined predict method, that actual method would be called instead of using the user's provided output.  This change fixes that.


gtarpenning (approved):
interesting, maybe we should note an error or warning too? that we are ignoring the predict?

neutralino1 (approved):

PR #5439, opened 2025-09-08
feat(weave): Add method to get predict calls from evaluation

andrewtruong:
https://wandb.atlassian.net/browse/WB-27517
https://wandb.atlassian.net/browse/WB-27503

Adds a method to get the `predict` calls from an evaluation

### DIFF
--- a/weave/evaluation/eval.py
+++ b/weave/evaluation/eval.py
@@ -385,6 +385,8 @@ class Evaluation(Object):
                     print(f"  Output: {call.output}")
             ```
         """
+        from weave.flow.model import INFER_METHOD_NAMES
+
         d = {}
         client = require_weave_client()
         for evaluate_call in self.get_evaluate_calls():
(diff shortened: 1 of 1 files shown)


weave/evaluation/eval.py
```diff
@@ -323,6 +323,44 @@ def get_evaluate_calls(self) -> CallsIter:
             ),
         )
 
+    def get_predict_calls(self) -> dict[str, list[Call]]:
```
neutralino1: Should this be called `get_predict_and_score_calls`?

andrewtruong: fixed -- I made separate predict and predict_and_score funcs

weave/evaluation/eval.py
```diff
@@ -323,6 +323,109 @@ def get_evaluate_calls(self) -> CallsIter:
...
+                print(f"Trace {trace_id}: {len(calls)} predict_and_score calls")
+                for call in calls:
+                    print(f"  Output: {call.output.get('output')}")
+                    print(f"  Scores: {call.output.get('scores')}")
+                    print(f"  Latency: {call.output.get('model_latency')}")
+            ```
+        """
+        d = {}
+        client = require_weave_client()
+        for evaluate_call in self.get_evaluate_calls():
```
neutralino1, line 350: Is there a way to do this without the N+1 queries? Like can we directly query `Evaluation.predict_and_score` calls?

weave/evaluation/eval.py
```diff
@@ -323,6 +323,109 @@ def get_evaluate_calls(self) -> CallsIter:
...
+            descendents = list(
+                client.get_calls(filter={"trace_ids": [evaluate_call.trace_id]})
+            )
+            predict_and_score_calls = [
+                call
+                for call in descendents
+                if call.summary.get("weave", {}).get("trace_name")
+                == "Evaluation.predict_and_score"
```
neutralino1, line 358: If the `Evaluation` object is not named "Evaluation" (i.e. the user specified a custom name), will this work?

weave/evaluation/eval.py
```diff
@@ -323,6 +323,109 @@ def get_evaluate_calls(self) -> CallsIter:
...
+            )
+
+            # Find calls that match common model inference method names
+            # We check both the trace_name (cleaner) and op_name (for compatibility)
+            predict_calls = []
+            for call in descendents:
+                # Check trace_name first (cleaner and more reliable)
+                trace_name = call.summary.get("weave", {}).get("trace_name", "")
+                if trace_name and any(
+                    trace_name.endswith(f".{method_name}") or trace_name == method_name
```
neutralino1, line 405: This seems safer than full equality.

andrewtruong:
Yeah, that's a good idea.  We already do that for the imperative evals.  I think that can be a separate PR

weave/evaluation/eval.py
gtarpenning: yeah can we add `op_name` to the filter? 

```
py_op_name_ref = f"weave:///{entity}/{project}/op/Evaluation.predict_and_score:*"
js_op_name_ref = ....
imperative_op_name_ref = ...
client.get_calls(filter={"trace_ids": [evaluate_call.trace_id], "op_names": [op_name_ref]})
```

weave/evaluation/eval.py
```diff
@@ -323,6 +323,109 @@ def get_evaluate_calls(self) -> CallsIter:
...
+                # Check if op_name contains the method name (handles weave URIs)
+                if call.op_name:
+                    for method_name in INFER_METHOD_NAMES:
+                        # Check various patterns in op_name
+                        if (
+                            f".{method_name}:" in call.op_name
+                            or f"/{method_name}:" in call.op_name
+                            or call.op_name.endswith(f".{method_name}")
+                            or call.op_name == method_name
+                        ):
```
gtarpenning, line 421: helper?

PR #5440, opened 2025-09-08
feat(weave): add metrics table for tracking alert metrics

gtarpenning:
wip


weave/trace_server/kafka.py
```diff
@@ -99,8 +102,11 @@ def produce_call_end(
                     }
                 )
 
+        # Use project_id as key to ensure all messages for a project
+        # go to the same partition for sequential processing by one worker
         self.produce(
             topic=CALL_ENDED_TOPIC,
+            key=call_end.project_id.encode("utf-8"),
```
gtarpenning: @neutralino1 making a note to chat with you about this change. Will be critical for the alert worker logic and might have downstream consequences on the existing scoring worker flow. I do think this is the right way to go though. Pros: fewer db queries because calls are sorted into projects, Cons: one very large project load will hotspot a worker.

gtarpenning: https://www.confluent.io/learn/kafka-message-key/#when-no-key-is-provided

> Imagine a Kafka topic with three partitions: Partition 0, Partition 1, and Partition 2. When a producer sends messages with different keys (e.g., user1, user2, and user3), Kafka will hash these keys and assign each one to a partition based on the hash result:

> Messages with the key user1 might be assigned to Partition 0.
Messages with the key user2 might be assigned to Partition 1.
Messages with the key user3 might be assigned to Partition 2.
Now, every time the producer sends a message with the key user1, it will always go to Partition 0. This consistency in partitioning ensures that messages related to user1 are kept together and processed in the correct order within the same partition.

neutralino1: That seems like a good idea.
We already batch by project_id on the worker, but different calls from the same project could be spread across multiple projects.
This might be an issue in dedicated deployments though, where most calls hit a single project (e.g. tmob), meaning only one worker would do all the work.

gtarpenning: Yeah good callout, more investigation required. We could also query gorilla for previously fired alerts, but there will always be an annoying race condition...

tests/trace/test_client_trace.py
```diff
@@ -5814,3 +5814,97 @@ def op1(extra_message: Optional[str] = None):
     #     )
     # )
     # assert len(calls) == 1
+
+
+def test_alert_metrics_create_query(client):
+    # Create single metric
+    print(f"Client server type: {type(client.server)}")
```
gtarpenning: trax

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2063,6 +2063,109 @@ def evaluation_status(
...
+                "metric_value",
+                "call_id",
+            ],
+        )
+
+        return tsi.AlertMetricsCreateRes(ids=metric_ids)
+
+    def alert_metrics_query(
+        self, req: tsi.AlertMetricsQueryReq
+    ) -> tsi.AlertMetricsQueryRes:
```
gtarpenning: todo make this nice, and add querying stuff

weave/trace_server/trace_server_interface.py
```diff
@@ -1218,6 +1218,53 @@ class EvaluationStatusRes(BaseModel):
...
+    call_id: str
+    created_at: datetime.datetime
+
+
+class AlertMetricsCreateReq(BaseModelStrict):
+    """Request to create multiple alert metrics in batch."""
+
+    project_id: str
+    metrics: list[AlertMetricItem]
+    wb_user_id: Optional[str] = Field(None, description=WB_USER_ID_DESCRIPTION)
```
gtarpenning: technically many ppl can be writing to the same project, this needs to be per-metric item

weave/trace_server/kafka.py
```diff
@@ -129,6 +135,9 @@ def from_env(
             "group.id": group_id,
             "auto.offset.reset": "earliest",
             "enable.auto.commit": False,
+            # Use cooperative-sticky assignment to minimize rebalances
+            # and keep partitions with the same consumer when possible
+            "partition.assignment.strategy": "cooperative-sticky",
```
gtarpenning, line 143: https://medium.com/apache-kafka-from-zero-to-hero/apache-kafka-guide-16-partition-rebalance-static-group-membership-1a5af31269b8

weave/trace_server/kafka.py
gtarpenning: I think I have an OK solution. The alert worker actually only cares about ops, not projects. We can reasonably fire two different alerts at the exact same time for the same project, if they are from two different alerts on different ops. If we move the cache key to project_id + op, we thus can maintain our alert worker separation while also breaking up large projects for the other workers. Now the only hotspotting issue would be for projects that have a very high throughput on one op. Most real prod mon scenarios are probably tracing a lot more than just one op, even the largest single instance prodmon projects have a pretty good distribution of ops.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2063,6 +2063,109 @@ def evaluation_status(
     ) -> tsi.EvaluationStatusRes:
         return evaluation_status(self, req)
 
+    # Alert API
+    def alert_metrics_create(
```
gtarpenning, line 2231: make async

weave/trace_server/kafka.py
gtarpenning: wont work. we dont know the op_name if its just a `call_end`

gtarpenning:
Closing - stale, alert metrics direction superseded.

PR #5446, opened 2025-09-08
chore(weave): Increase minimum python version to 3.10 and upgrade to py310 syntax

andrewtruong:
This PR increases weave's minimum python version to `3.10`.

Python 3.9 reached end of life in October 2025
https://devguide.python.org/versions/


neutralino1 (approved):
Thank you. Can we add this early graceful prompt to upgrade Python versions if the user is on <= 3.9 ?

andrewtruong:
Added

PR #5449, opened 2025-09-09
fix(weave): drop calls from bufstream produce after max (10k)

gtarpenning:
## Description

[WB-27534](https://wandb.atlassian.net/browse/WB-27534)

Max buffer size on the trace server kafka queue. Warn if at 50%. Drop call from queue push at 100%. Call will still be inserted, but not eligible for online scoring. 

## Testing

how do we test scoring worker stuff? 

[WB-27534]: https://wandb.atlassian.net/browse/WB-27534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


neutralino1 (commented):
Thank you

PR #5450, opened 2025-09-09
fix log_summary function

Esther-Guo:
## Description
  1. Fixed `summary` None check: summary None check logic does not always get executed
  2. Eliminated duplication: When `auto_summarize=False`, user summary only appears under the specified key -> note that this is imcompatible with old version, but it makes more sense to remove duplication?
  3. Added customization: New `summary_key` parameter (default: "output") allows custom key names (otherwise weave ui shows 'output' under 'output')

## Testing

How was this PR tested?
Modified the related tests


Esther-Guo:

PR #5457, opened 2025-09-09
chore(weave): Defer numpy import in scorer

andrewtruong:


weave/flow/scorer.py
```diff
@@ -18,12 +18,14 @@
...
 from weave.trace.weave_client import sanitize_object_name
 
-try:
-    import numpy as np
-except ImportError:
-    _NUMPY_AVAILABLE = False
-else:
-    _NUMPY_AVAILABLE = True
+
+def _numpy() -> Optional[Any]:
```
neutralino1:
```suggestion
def _import_numpy() -> Optional[Any]:
```

neutralino1 (approved):

PR #5458, opened 2025-09-09
chore(weave): Replace requests with httpx

andrewtruong:
- **test**
- **test**


weave/utils/http_requests.py
```diff
@@ -7,10 +7,39 @@
 from typing import Any, Optional, Union
 
 import httpx
+from httpx import HTTPStatusError
 from httpx import Request, Response
 
 from weave.trace.display.display import Console, Text
 
+
+class HTTPError(HTTPStatusError):
```
andrewtruong, line 15: I think we specifically need this because the trace server has not switched over to httpx yet.

PR #5468, opened 2025-09-11
feat(weave): add image support to trace server

jwlee64:
## Description

Add image creation path to trace server
- saves base64 and urls as content objs, and returns a content_ref to the frontend

## Testing

manually


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2049,6 +2053,121 @@ def _single_error_iter(e: Exception) -> Iterator[dict[str, str]]:
...
+        Create images using LLM image generation.
+
+        Args:
+            req (tsi.ImageGenerationCreateReq): The image generation request.
+
+        Returns:
+            tsi.ImageGenerationCreateRes: The image generation response.
+        """
+        # Validate input parameters
+        if not req.inputs.model:
```
neutralino1: Can you use explicit conditions, e.g.
```suggestion
        if req.inputs.model is None:
```

Truthiness is a 🦶 🔫 .

neutralino1: Same for all the conditions below.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2049,6 +2053,121 @@ def _single_error_iter(e: Exception) -> Iterator[dict[str, str]]:
...
+        if not req.inputs.prompt:
+            return tsi.ImageGenerationCreateRes(
+                response={"error": "No prompt specified in request"}
+            )
+
+        # Get API key from secret fetcher
+        secret_fetcher = _secret_fetcher_context.get()
+        if not secret_fetcher:
+            return tsi.ImageGenerationCreateRes(
+                response={"error": "No secret fetcher available"}
```
neutralino1: Are these error messages displayed to users? If yes, is this message leaking implementation details?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2049,6 +2053,121 @@ def _single_error_iter(e: Exception) -> Iterator[dict[str, str]]:
...
+        )
+        start_call = _start_call_for_insert_to_ch_insertable_start_call(start)
+
+        end = tsi.EndedCallSchemaForInsert(
+            project_id=req.project_id,
+            id=start_call.id,
+            ended_at=end_time,
+            output=res.response,
+            summary={},
+        )
```
neutralino1, line 2147: Why do we need to log an op call for this?

weave/trace_server/image_completion.py
neutralino1, line 1: Is this the module that is duplicated with @zbirenbaum 's work?

zbirenbaum:
@jwlee64 Lets talk, this redoes a bunch of stuff I think I tried in a different PR and found some issues with. We have to be super careful on this

weave/trace_server/clickhouse_trace_server_batched.py
jwlee64: update error message

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2052,6 +2056,123 @@ def _single_error_iter(e: Exception) -> Iterator[dict[str, str]]:
...
+                response={
+                    "error": "Unable to access required credentials for image generation"
+                }
+            )
+
+        api_key = (
+            secret_fetcher.fetch("OPENAI_API_KEY")
+            .get("secrets", {})
+            .get("OPENAI_API_KEY")
+        )
```
neutralino1, line 2095: Are OpenAI models the only ones supporting image generation? Should we validate that the requested model is indeed an OpenAI model?

jwlee64: I scoped this down a while back to only include 3 image models that were all openai ones
litellm only really supports a small subset of image generation models. Though in theory someone could hit this endpoint with a curl request and an incorrect model, and it would simply fail in the litellm validation

weave/trace_server/image_completion.py
```diff
@@ -0,0 +1,215 @@
...
+
+        try:
+            # gpt-image-1 doesn't support response_format parameter
+            if model_name == "gpt-image-1":
+                res = image_generation(**image_params)
+            else:
+                res = image_generation(response_format="url", **image_params)
+        except Exception as e:
+            raise
+
```
zbirenbaum, line 165: I think these parameters should be sent in the request and validated by the client. We should throw an error if the params are not valid (can probably just propagate whatever the API sends back to the user

weave/trace_server/image_completion.py
```diff
@@ -0,0 +1,215 @@
...
+        if model_name == "dall-e-3":
+            image_params["style"] = "natural"  # Default style
+
+        # Set default size for models that need it
+        if model_name == "dall-e-3":
+            image_params["size"] = "1024x1024"
+        elif model_name == "gpt-image-1":
+            image_params["size"] = "1024x1024"
+        else:
+            image_params["size"] = "1024x1024"
```
zbirenbaum, line 153: Make all of these optional properties of a `params` field or similar on ImageCreateReq. You can use these as defaults. Ideally you'd have discriminated unions based on the model field to concretely know what params are valid though the types.

We should throw an error if the params are not valid (can probably just propagate whatever the API sends back to the user

weave/trace_server/image_completion.py
```diff
@@ -0,0 +1,215 @@
...
+                            )
+
+                            # If we created a Content object, add it directly to content
+                            if "content" in processed_item:
+                                content_items.append(
+                                    {
+                                        "type": "image",
+                                        "image": processed_item["content"],
+                                    }
+                                )
```
zbirenbaum: Lets make response_data look like this to match our existing image message logic. content should not be top level, should be output here.

```
{
  ...,
  "output": [{
    role: "assistant",
    content: [{
      type: "output_image", # image is also fine but openai is moving to using input_image and output_image
      image: ${Content Object}
    }]
  }]
}
```

zbirenbaum (changes requested):
A few things. Going to be adamant on the schema change, lets keep things as consistent as possible.

The parameter comments are a bit more optional for this PR. I do think they should be addressed but it might take a while and it doesn't affect our data model or storage so I'm alright with leaving for a followup.

weave/trace_server/image_completion.py
jwlee64: everything returned here is the output

jwlee64: Screenshot 2025-09-19 at 11 46 32 AM


down to switch to output image though

weave/trace_server/image_completion.py
```diff
@@ -0,0 +1,215 @@
...
+        # Handle base64-encoded images
+        elif (
+            "b64_json" in data_item
+            and data_item["b64_json"]
+            and trace_server
+            and project_id
+        ):
+            b64_data = data_item["b64_json"]
+            # Use Content.from_base64() to handle decoding and content creation
+            content_obj = Content.from_base64(
```
zbirenbaum, line 65: Do we know the mimetype of this image? I couldn't find it in the litellm docs. If so it would be best if we manually construct the data url and let it autoconvert or do from_data_url for clarity and efficiency reasons.

weave/trace_server/image_completion.py
jwlee64: currently we have no way in the UI to set these fields which is why they are simply defaulted

weave/trace_server/image_completion.py
jwlee64: it should just be png, ill convert it to a data url and add the mime type

jwlee64: going to still manually convert, because we only auto convert when using call_start() whiuch we do not use here

weave/trace_server/image_completion.py
```diff
@@ -0,0 +1,206 @@
...
+
+            content_obj = Content.from_base64(
+                b64_data,
+                mimetype="image/png",
+                metadata={"source_index": index, "_original_schema": "b64_json"},
+            )
+            content_dict = store_content_object(content_obj, project_id, trace_server)
+            processed_item["b64_json"] = content_dict
+
+    except Exception as e:
```
zbirenbaum, line 73: Raw exceptions are dangerous. Could we narrow this down?

weave/trace_server/image_completion.py
```diff
@@ -0,0 +1,206 @@
...
+        # Don't raise - just log the error and return the original item
+        processed_item["processing_error"] = str(e)
```
zbirenbaum:
```suggestion
        processed_item["error"] = str(e)
```

weave/trace_server/image_completion.py
```diff
@@ -0,0 +1,206 @@
...
+                            processed_item = _process_image_data_item(
+                                data_item, index, trace_server, project_id, wb_user_id
+                            )
+
+                            if "b64_json" in processed_item:
+                                data_item["b64_json"] = processed_item["b64_json"]
+                            if "url" in processed_item:
+                                data_item["url"] = processed_item["url"]
+
+            except Exception as e:
```
zbirenbaum: You already have a try in _process_image_data_item, this one isn't necessary

weave/trace_server/image_completion.py
```diff
@@ -0,0 +1,206 @@
...
+                # Continue without failing - the response will still contain the original data
+                pass
+
+        return tsi.ImageGenerationCreateRes(response=response_data)
+    except Exception as e:
+        error_message = str(e)
+        error_message = error_message.replace("litellm.", "")
+        return tsi.ImageGenerationCreateRes(response={"error": error_message})
```
zbirenbaum, line 202: Shouldn't we also use the error field of the object returned by
`_process_image_data_item` to return an error? It would be better to return an error if content conversion fails because that's a critical code path.

weave/trace_server/image_completion.py
jwlee64: hmm there could be. number of errors that arise
data format errors
http errors
image processing errors, which is why i left it as a raw catch

weave/trace_server/image_completion.py
jwlee64: will propagate the error up

weave/trace_server/image_completion.py
zbirenbaum: We should probably make a content error so we can catch all of those easily, fair enough for now though

PR #5469, opened 2025-09-11
fix(weave): use delimiter and custom:: prefix for custom providers

jwlee64:
## Description

Allows the usage of '/' and ':' in provider and custom model names unblocking ollama and openrotuer custom model usage
Changes to accept a custom:: prefix from the frontend on custom provider requests instead of just guessing at it

## Testing

created and hit local ollama models and sent reqeusts to openrouter

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -2915,15 +2915,11 @@ def _create_tracked_stream_wrapper(
 
             # Prepare summary and end call
             summary: dict[str, Any] = {}
-            if (
-                aggregated_output
-                and "usage" in aggregated_output
-                and model_name is not None
-            ):
-                summary["usage"] = {model_name: aggregated_output["usage"]}
-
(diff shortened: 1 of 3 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2903,9 +2915,16 @@ def _stream_wrapper() -> Iterator[dict[str, Any]]:
...
-            if aggregated_output and "usage" in aggregated_output:
+            if (
+                aggregated_output
+                and "usage" in aggregated_output
+                and model_name is not None
+            ):
                 summary["usage"] = {model_name: aggregated_output["usage"]}
 
+            if aggregated_output is not None and model_name is not None:
+                aggregated_output["model"] = model_name
```
neutralino1: Nit
```suggestion
            if (
                aggregated_output is not None
                and model_name is not None
            ):
                aggregated_output["model"] = model_name
                if "usage" in aggregated_output:
                     summary["usage"] = {model_name: aggregated_output["usage"]}
```

PR #5470, opened 2025-09-11
fix(weave): add explicit thread cleanup/gc with weakref in ch batched

gtarpenning:
## Description

Attempts to fix potential memory leak in threaded impl of ch batched.

## Testing

How was this PR tested?


gtarpenning:
THis is messed up, we create one health check thread per sync worker thread i think... closing

PR #5472, opened 2025-09-11
chore(weave): Add server availability check

andrewtruong:
Adds a check to confirm if the server is available (it may not be if weave is not enabled on the customer's instance)

### DIFF
--- a/tests/trace/test_weave_init_availability.py
+++ b/tests/trace/test_weave_init_availability.py
@@ -8,7 +8,7 @@ from weave.trace_server_bindings import remote_http_trace_server
 
 
 def test_weave_is_available_json_decode_error():
-    """Test that _weave_is_available returns False when server_info() raises JSONDecodeError."""
+    """Test that _weave_is_available returns False server is not available."""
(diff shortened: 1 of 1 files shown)


weave/trace/weave_init.py
```diff
@@ -73,6 +74,19 @@ def get_entity_project_from_project_name(project_name: str) -> tuple[str, str]:
...
 
+def _weave_is_available(server: remote_http_trace_server.RemoteHTTPTraceServer) -> bool:
+    try:
+        server.server_info()
+    except JSONDecodeError:
+        return False
+    except Exception:
+        logger.warning(
+            "Unexpected error when checking if Weave is available on the server.  Please contact support."
+        )
```
neutralino1, line 85: Doesn't this double print with the exception below?

andrewtruong: Yes, but this is the case where an unexpected error happened, so it's something for us to investigate.  The regular case is the `JSONDecodeError`

andrewtruong:
### Merge activity

* **Sep 11, 9:52 PM UTC**: @andrewtruong merged this pull request with [Graphite](https://app.graphite.dev/github/pr/wandb/weave/5472).

PR #5475, opened 2025-09-11
feat(weave): from url for content

zbirenbaum:
## Description

Adds a from_url classmethod for content

## Testing

added unit tests


tests/trace/type_handlers/Content/content_test.py
neutralino1, line 1: The usual naming pattern is `test_content.py`.

zbirenbaum: I'll rename in followup posted right after this, if I do it here it will mess up the diff I think

zbirenbaum: https://github.com/wandb/weave/pull/5477

PR #5486, opened 2025-09-16
fix(weave): Arbitrary strings being converted to content

zbirenbaum:
## Description

- Fixes WB-27317

Any word with length mod 4 equal to 0 is technically valid base64 so this changes the conversion to only occur on data urls

## Testing

updated tests


gtarpenning (approved):
will this change the de-serialization of existing objects? b64 strings that we converted will still be served right?

zbirenbaum:
> will this change the de-serialization of existing objects? b64 strings that we converted will still be served right?

No, nothing happens on the way out. This has only affected things on the way in

PR #5488, opened 2025-09-17
chore(weave): Put trace tests under 1 big worker

andrewtruong:


gtarpenning (approved):
probably should break this up into multiple prs, but each change lgtm

PR #5491, opened 2025-09-17
fix(weave): update cohere integration definitions for new sdk version

parambharat:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,7 +104,7 @@ docs = [
 video_support = ["moviepy<=1.0.3"]
 anthropic = ["anthropic>=0.18.0"]
 cerebras = ["cerebras-cloud-sdk"]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -104,7 +104,7 @@ docs = [
 video_support = ["moviepy<=1.0.3"]
 anthropic = ["anthropic>=0.18.0"]
 cerebras = ["cerebras-cloud-sdk"]
-cohere = ["cohere>=5.9.1,<5.9.3"]
+cohere = ["cohere>=5.9.1,<5.19.0"]
```
ayulockin: do we need a `<` pin here? 

I think we should also update the `>=` pin to a release from early this year.

PR #5494, opened 2025-09-17
fix(weave): OTEL summary missing fields

zbirenbaum:
## Description

- Fixes WB-27756

Fixes OTEL to properly set summary fields used by various queries. Additionally changes the default unset status code to map to a completed weave status so we always get a real status.

## Testing

Tested manually


gtarpenning (approved):
This enables sorting latency?

zbirenbaum:
> This enables sorting latency?

No but it is still an issue. Haven't tracked down the reason for that yet

PR #5496, opened 2025-09-18
chore(weave): attributes on eval, optional scorers on imperative logger

gtarpenning:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?

### DIFF
--- a/tests/flow/test_evaluation_imperative.py
+++ b/tests/flow/test_evaluation_imperative.py
@@ -655,3 +655,45 @@ def test_evaluation_invalid_model_name_fixable(model_name):
 def test_evaluation_invalid_model_name_not_fixable(model_name):
     with pytest.raises(ValueError):
         weave.EvaluationLogger(model=model_name)
+
+
+def test_evaluation_logger_with_predefined_scorers(client, caplog):
+    """Test that EvaluationLogger can track predefined scorers and warn when using unlisted ones."""
+    import logging
+
+    # Create evaluation with predefined scorer names
+    ev = EvaluationLogger(
+        model="test_model",
+        dataset=[{"input": 1}],
+        scorers=["accuracy", "precision"],  # List of allowed scorer names
+    )
+
+    with caplog.at_level(logging.WARNING):
+        pred = ev.log_prediction({"input": 1}, 1)
+
+        # These should not warn (in the predefined list)
+        pred.log_score("accuracy", 0.9)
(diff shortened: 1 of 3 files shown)


neutralino1 (changes requested):
I'm not sure what the purpose of this change is. How does it address the issue described in the JIRA ticket?

gtarpenning:
> I'm not sure what the purpose of this change is. How does it address the issue described in the JIRA ticket?

Now, specify the optional `scorers` to actually create new versions of the Evaluation object, which will cause the UI to pick up all available scorers.

tests/flow/test_evaluation_imperative.py
```diff
@@ -655,3 +655,45 @@ def test_evaluation_invalid_model_name_fixable(model_name):
...
+
+def test_evaluation_logger_with_predefined_scorers(client, caplog):
+    """Test that EvaluationLogger can track predefined scorers and warn when using unlisted ones."""
+    import logging
+
+    # Create evaluation with predefined scorer names
+    ev = EvaluationLogger(
+        model="test_model",
+        dataset=[{"input": 1}],
+        scorers=["accuracy", "precision"],  # List of allowed scorer names
```
neutralino1, line 668: I think `scorers` is an awkward name here, since there are no scorers with `EvaluationLogger`.

Two options discussed with @gtarpenning :
1. rename to `scores` or `metrics`
2. instead of this awkward list of strings, let users specify a version number manually.

Option 2 is the most elegant and gives users the most control over versioning. The problem is that this version number would be out of sync with `version_index`. One way to get around that would be to call it `version_seed` or something, to signify that changing it bumps the version, but it's not great. It makes option 1 more compelling.

WDYT @andrewtruong ?

andrewtruong: I'm ok with the name `scorers`.

If this field is specified, it might be good to give a warning to the user if they end up logging a score with a name not in the provided list.

EDIT: And I see you have already done that -- great!

andrewtruong: @neutralino1 changing the `scorers` field here will change `Evaluation.attributes`, which will trigger the version bump.  I think that's the expected behaviour, no?

One nit is on the name `attributes`.  I'm not sure if this is more appropriate or maybe something like `metadata`

PR #5499, opened 2025-09-18
chore(weave): Remove nest_asyncio dep

andrewtruong:


gtarpenning (approved):
I think this makes sense, but i'm not sure of the exact magic that the other package was doing... extensively tested?

PR #5502, opened 2025-09-18
chore(weave): Use httpx

andrewtruong:


jwlee64 (approved):
The updates to the scripts lgtm as well as the completions.
Maybe cc @jamie-rasmussen just as heads up
chat/completions is for inference

PR #5510, opened 2025-09-23
feat(weave): user code can define custom content to display on a call page

bcsherma:
## Description

Adds `weave.set_view` to the primary api. Adds `EvaluationLogger.set_view` to the imperative eval logger api.

The parameter and return signature is the same for both. They accept a view `name` and `weave.Content` to display.

Content is persisted to the file backend and the json manifest is dumped into `summary.weave.views.<view_name>`.

This frontend PR adds rendering for these custom views to a new default tab for the call page: https://github.com/wandb/core/pull/34732

## Testing

Added unit tests to verify the summary is populated with the `Content` json dump when called in an op or the imperative eval method, passed either by a `Content` object or a string with an extension or mimetype specified.

### DIFF
--- a/weave/evaluation/eval_imperative.py
+++ b/weave/evaluation/eval_imperative.py
@@ -33,10 +33,11 @@ from weave.trace.api import attributes
 from weave.trace.call import Call
 from weave.trace.context import call_context
 from weave.trace.context.weave_client_context import require_weave_client
-from weave.trace.op import op
-from weave.trace.op_protocol import Op
+from weave.trace.op import Op, op
 from weave.trace.table import Table
 from weave.trace.util import Thread
+from weave.trace.view_utils import set_call_view
+from weave.type_wrappers.Content.content import Content
 
 T = TypeVar("T")
 ID = str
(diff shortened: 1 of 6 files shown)


bcsherma:
* **#5510** <a href="https://app.graphite.dev/github/pr/wandb/weave/5510?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/5510?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace/view_utils.py
```diff
@@ -0,0 +1,133 @@
...
+        TypeError: If ``content`` is neither a string nor a ``weave.Content``.
+
+    Examples:
+        >>> resolve_view_content("# Title", extension="md").mimetype
+        'text/markdown'
+    """
+    if isinstance(content, Content):
+        return content
+
+    if isinstance(content, str):
```
zbirenbaum: Mimetype can't be easily resolved for text type contents so this is going to typically show up as text/plain if the default args are used just a heads up.

weave/trace/view_utils.py
```diff
@@ -0,0 +1,133 @@
...
+        return Content.from_text(
```
zbirenbaum: Also is there a reason this is exclusively for text as an alternative input?

You could pretty easily use Content._from_guess and allow using bytes, string, filepath etc. Not sure if it makes sense for this context though.

bcsherma: Ah I missed this, I will take a look at this, also wrt your other comment

weave/evaluation/eval_imperative.py
```diff
@@ -672,6 +672,62 @@ def log_summary(
...
+
+        Returns:
+            None
+
+        Examples:
+            >>> import weave
+            >>> ev = weave.EvaluationLogger()
+            >>> ev.set_view("report", "# Report", extension="md")
+        """
+        if not name or not isinstance(name, str):
```
neutralino1: Please don't use truthiness.
```suggestion
        if not isinstance(name, str) or len(name) == 0:
```
(also swapping to test `str` instance first)

weave/evaluation/eval_imperative.py
```diff
@@ -672,6 +672,62 @@ def log_summary(
...
+            raise ValueError("`name` must be a non-empty string")
+
+        if self._evaluate_call is None:
+            raise RuntimeError(
+                "Evaluation call not initialized; cannot add view before evaluation starts"
+            )
+
+        wc = require_weave_client()
+
+        from weave.trace.view_utils import set_call_view
```
neutralino1: Why not at the top? Dynamic imports are a 👃.

weave/trace/api.py
```diff
@@ -250,6 +248,58 @@ def attributes(attributes: dict[str, Any]) -> Iterator:
...
+    Examples:
+        >>> import weave
+        >>> weave.init("proj")
+        >>> @weave.op
+        ... def foo():
+        ...     weave.set_view("readme", "# Hello", extension="md")
+        ...     return 1
+        >>> foo()
+    """
+    if not name or not isinstance(name, str):
```
neutralino1: Same.

weave/trace/api.py
```diff
@@ -250,6 +248,58 @@ def attributes(attributes: dict[str, Any]) -> Iterator:
...
+        raise ValueError("`name` must be a non-empty string")
+
+    call = require_current_call()
+    client = weave_client_context.require_weave_client()
+
+    from weave.trace.view_utils import set_call_view
```
neutralino1: Same.

weave/trace/view_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    else:
+        raise TypeError("`content` must be a weave.Content instance or a string")
+
+    updates: dict[str, Any] = {}
+    if metadata is not None:
+        updates["metadata"] = metadata
+    if encoding is not None:
+        updates["encoding"] = encoding
+
+    if updates:
```
neutralino1: Explicit condition please.

weave/trace/view_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        >>> call = Call(None, "proj", None, {})  # doctest: +SKIP
+        >>> set_call_view(  # doctest: +SKIP
+        ...     call=call,
+        ...     client=client,
+        ...     name="report",
+        ...     content="<h1>Hello</h1>",
+        ...     extension="html",
+        ... )
+    """
+    if not name or not isinstance(name, str):
```
neutralino1: Same.

neutralino1: Are the validations in the upstream functions necessary if it is done here?

neutralino1 (approved):
I actually would have kept the centralized validation and would have gotten rid of the other two ones, but as you wish ✅

bcsherma:
Yeah this is not DRY. I think it is nice to do this type validation directly in the user facing function so that we can assume type compliance for callers for internal apis and the stack trace for the user/caller if they fail to comply is super transparent ⚖️

bcsherma:
What do you think of that rationale? Would you recommend I switch to the central validation?

PR #5511, opened 2025-09-23
fix(weave): Add partial success handling and descriptive errors for invalid otel attributes

zbirenbaum:
## Description

- Fixes WB-27789

- Validates and returns descriptive errors instead of returning a 500 for the following scenario and similar:
- Partial success handling means valid traces will still be logged and only invalid ones will fail

1. The user set some key like "gen_ai.prompt" to “True” and "gen_ai.prompt.content" to “Hello”
2. We go to unflatten the object:
3. We first handle "gen_ai.prompt" since it is the parent key
4. We determine that gen_ai.prompt is a boolean value and assign it to be true
5. We find "gen_ai.prompt.content"
6. We implicitly determine that the type of gen_ai.prompt should be a dict (not a boolean)
7. When we try and set ‘content’ in the unflattened dict we get an assignment error
8. There isn’t a way to expand the key into a dict format (which is required to turn it to a valid structured json). A similar error can be triggered if you set gen_ai.prompt.0.* to a value, which would be argument of type 'bool' is not iterable

## Testing

Added unit tests

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -1021,8 +1021,29 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
         sort_fields = []
         if req.sort_by:
             for sort in req.sort_by:
+                # Validate sort field to prevent empty JSON paths
+                if not sort.field or not sort.field.strip():
+                    raise InvalidRequest("Sort field cannot be empty")
+
+                # Check for invalid dot patterns that would create malformed JSON paths
+                if (
+                    sort.field.startswith(".")
+                    or sort.field.endswith(".")
+                    or ".." in sort.field
+                ):
+                    raise InvalidRequest(
+                        f"Invalid sort field '{sort.field}': field names cannot start/end with dots or contain consecutive dots"
+                    )
+
                 # TODO: better splitting of escaped dots (.) in field names
                 extra_path = sort.field.split(".")
+
+                # Additional validation: ensure no empty path components
+                if any(not component.strip() for component in extra_path):
+                    raise InvalidRequest(
+                        f"Invalid sort field '{sort.field}': field path components cannot be empty"
+                    )
+
                 field = OrderField(
                     field=QueryBuilderDynamicField(
                         field=VAL_DUMP_COLUMN_NAME, extra_path=extra_path
--- a/docs/docs/guides/core-types/env-vars.md
+++ b/docs/docs/guides/core-types/env-vars.md
@@ -22,7 +22,7 @@ os.environ["WEAVE_PRINT_CALL_LINK"] = "false"
 |----------|------|---------|-------------|
 | `WANDB_API_KEY` | `string` | `None` | If set, automatically log into W&B Weave without being prompted for your API key. To generate an API key, log in to your W&B account and go to [https://wandb.ai/authorize](https://wandb.ai/authorize). |
 | `WEAVE_DISABLED` | `bool` | `false` | When set to `true`, disables all Weave tracing. Weave ops will behave like regular functions. |
-| `WEAVE_PRINT_CALL_LINK` | `bool` | `true` | Controls whether to print a link to the Weave UI when calling a Weave op. |
+| `WEAVE_PRINT_CALL_LINK` | `bool` | `true` | Controls whether to print a link to the Weave UI when calling a Weave op. You can also set this directly in your code by configuring the `settings` argurment for `weave.init()` like this: `weave.init("your-project-name", settings={"print_call_link": False})` |
 | `WEAVE_LOG_LEVEL` | `str` | `INFO` | Controls the log level of the weave logger.
 | `WEAVE_CAPTURE_CODE` | `bool` | `true` | Controls whether to save code for ops so they can be reloaded for later use. |
 | `WEAVE_DEBUG_HTTP` | `bool` | `false` | When set to `true`, turns on HTTP request and response logging for debugging. |
--- a/docs/docs/guides/integrations/agno.md
+++ b/docs/docs/guides/integrations/agno.md
@@ -1,5 +1,7 @@
 # Agno
 
+import DefaultEntityNote from '../../../src/components/DefaultEntityNote.mdx';
+
 You can trace [Agno](https://docs.agno.com/) agent and tool calls in Weave using [OpenTelemetry (OTEL)](https://opentelemetry.io/). Agno is a Python framework for building multi-agent systems with shared memory, knowledge, and reasoning. It's designed to be lightweight, model-agnostic, and highly performant, supporting multi-modal capabilities including text, image, audio, and video processing.
 
 This guide explains how to trace Agno agent and tool calls using OTEL, and visualize those traces in Weave. You'll learn how to install the required dependencies, configure an OTEL tracer to send data to Weave, and instrument your Agno agents and tools.
@@ -37,7 +39,7 @@ It is recommended that you store sensitive environment variables like your API k
 - **Endpoint:** `https://trace.wandb.ai/otel/v1/traces`. If you are using a dedicated Weave instance, the URL follows this pattern instead: `{YOUR_WEAVE_HOST}/traces/otel/v1/traces`
 - **Headers:**
   - `Authorization`: Basic auth using your W&B API key
-  - `project_id`: Your W&B entity/project name (e.g., `myteam/myproject`)
+  - `project_id`: Your W&B entity/project name (e.g., `myteam/myproject`). <DefaultEntityNote />
 
 ## Send OTEL traces from Agno to Weave
 
--- a/docs/docs/guides/integrations/anthropic.md
+++ b/docs/docs/guides/integrations/anthropic.md
@@ -1,5 +1,7 @@
 # Anthropic
 
+import DefaultEntityNote from '../../../src/components/DefaultEntityNote.mdx';
+
 <a target="_blank" href="https://colab.research.google.com/github/wandb/examples/blob/master/weave/docs/quickstart_anthropic.ipynb">
   <img src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open In Colab"/>
 </a>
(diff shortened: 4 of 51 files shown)


weave/trace_server/opentelemetry/helpers.py
```diff
@@ -123,19 +123,89 @@ def _get_value_from_nested_dict(d: dict[str, Any], key: str) -> Any:
     return current
 
 
+class AttributePathConflictError(ValueError):
```
andrewtruong: is this better as a `TypeError`?

weave/trace_server/opentelemetry/helpers.py
```diff
@@ -123,19 +123,89 @@ def _get_value_from_nested_dict(d: dict[str, Any], key: str) -> Any:
...
         d[key] = value
         return
 
     parts = key.split(".")
     current = d
-    for _, part in enumerate(parts[:-1]):
+    # Walk down to the parent map, creating dicts as needed, but fail if we need
+    # to descend into a non-dict value.
+    for i, part in enumerate(parts[:-1]):
+        parent_path = ".".join(parts[: i + 1])
```
andrewtruong, line 188: this took me a bit to read.  Can you provide a simple comment with the example of what this is doing?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -217,26 +218,53 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
...
+                        error_messages.append(f"Rejected span ({span_ident}): {e!s}")
         # TODO: Actually populate the error fields if call_start_batch fails
         self.call_start_batch(tsi.CallCreateBatchReq(batch=calls))
+        if rejected_spans > 0:
+            return tsi.OtelExportRes(
+                partial_success=tsi.ExportTracePartialSuccess(
+                    rejected_spans=rejected_spans,
+                    error_message=(
+                        "; ".join(error_messages[:20])
+                        + ("; ..." if len(error_messages) > 20 else "")
```
andrewtruong: please factor out into a descriptive var

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -217,26 +218,53 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
...
-                    )
+        calls: list[dict[str, object]] = []
+        rejected_spans = 0
+        error_messages: list[str] = []
+        for proto_resource_spans in req.traces.resource_spans:
+            resource = Resource.from_proto(proto_resource_spans.resource)
+            for proto_scope_spans in proto_resource_spans.scope_spans:
+                for proto_span in proto_scope_spans.spans:
+                    try:
+                        span = Span.from_proto(proto_span, resource)
```
andrewtruong, line 229: I think this is the one line that can raise `AttributePathConflictError`, is that correct?  If yes, please tighten the try-except block to just this line

tests/trace_server/test_opentelemetry.py
```diff
@@ -713,6 +715,37 @@ def test_flatten_attributes(self):
...
+    def test_expand_attributes_conflict_parent_then_child(self):
+        """Setting a parent primitive then a nested subkey should raise a clear error."""
+        flat_attrs = [
+            ("gen_ai.prompt", True),
+            ("gen_ai.prompt.content", "Hello"),
+        ]
+
+        try:
+            expand_attributes(flat_attrs)
+            raise AssertionError("Expected AttributePathConflictError")
```
andrewtruong, line 727: I dont understand what is happening in this test

tests/trace_server/test_opentelemetry.py
```diff
@@ -713,6 +715,37 @@ def test_flatten_attributes(self):
...
+    def test_expand_attributes_conflict_child_then_parent(self):
+        """Setting nested subkeys then a parent primitive should raise a clear error."""
+        flat_attrs = [
+            ("gen_ai.prompt.content", "Hello"),
+            ("gen_ai.prompt", True),
+        ]
+
+        try:
+            expand_attributes(flat_attrs)
+            raise AssertionError("Expected AttributePathConflictError")
```
andrewtruong, line 743: I dont understand what is happening in this test

weave/trace_server/opentelemetry/helpers.py
```diff
@@ -123,19 +123,89 @@ def _get_value_from_nested_dict(d: dict[str, Any], key: str) -> Any:
...
-    """Set a value in a nested dictionary using a dot-separated key."""
+    """Set a value in a nested dictionary using a dot-separated key.
+
+    Raises AttributePathConflictError when a parent path already holds a non-dict value
+    and a nested subkey is being set or when attempting to overwrite an existing mapping
+    with a primitive while subkeys exist.
+    """
     if "." not in key:
+        # If there is already a mapping at this key, assigning a primitive would drop subkeys
+        existing = d.get(key, None)
```
andrewtruong: nit: the second arg to `get` is default `None`

weave/trace_server/opentelemetry/helpers.py
```diff
@@ -123,19 +123,89 @@ def _get_value_from_nested_dict(d: dict[str, Any], key: str) -> Any:
...
+            raise AttributePathConflictError(
+                parent_key=parent_path,
+                attempted_subkey=parts[i + 1],
+                existing_type=type(current[part]),
+            )
         current = current[part]
-    current[parts[-1]] = value
+
+    last = parts[-1]
+    existing_leaf = current.get(last, None)
```
andrewtruong: same here

tests/trace_server/test_opentelemetry.py
zbirenbaum: the flat_attrs are invalid:

gen_ai.prompt -> set to True (type is boolean)
gen_ai.prompt.content -> Implied type of gen_ai.prompt is dict (e.g. gen_ai.prompt['content'] == "hello")

So if `expand_attributes` succeeds there's a problem. We expect it to throw an error. The below is just checking the error message.

tests/trace_server/test_opentelemetry.py
zbirenbaum: This tests the same as the above but is an order of operations test ensuring we handle the case that the subkey is checked before the primary.

tests/trace_server/test_opentelemetry.py
andrewtruong: Shouldn't `expand_attributes` already raise?  Why are you additionally raising in the test?

weave/trace_server/opentelemetry/helpers.py
zbirenbaum: Its more like a ValueError. The types themselves aren't invalid, they've just set their attribute values in an invalid way (e.g. gen_ai.prompt should not be set only if gen_ai.prompt.0 is set).  I'm not opposed to changing it I think theres an argument for each.

weave/trace_server/opentelemetry/helpers.py
zbirenbaum: Yeah, reading through this is a bit difficult. I did it this way for efficiency (1 pass instead of 2) but I think performing validation in a separate step might make things clearer. I don't love how much this obfuscates the tree logic.

tests/trace_server/test_opentelemetry.py
zbirenbaum: If I don't raise the test will pass even if it does not raise. I guess I could just assert false here instead? I think I'm missing something

weave/trace_server/opentelemetry/helpers.py
zbirenbaum: Fixed

weave/trace_server/opentelemetry/helpers.py
zbirenbaum: Separated it out into a helper and added descriptive examples + scenarios

weave/trace_server/opentelemetry/helpers.py
zbirenbaum: Fixed

weave/trace_server/opentelemetry/helpers.py
zbirenbaum: Switched to a type error since it can go either way.

weave/trace_server/clickhouse_trace_server_batched.py
zbirenbaum: Done

gtarpenning:
@zbirenbaum  Status here? This is our number one error in the trace server rn

zbirenbaum:
> @zbirenbaum Status here? This is our number one error in the trace server rn

I fixed it up and requested re-review just waiting on a stamp.

tests/trace_server/test_opentelemetry.py
andrewtruong: oh I see

I would put the raise in an else block for clarity

tests/trace_server/test_opentelemetry.py
andrewtruong: same here

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -217,15 +218,34 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
...
+                        rejected_spans += 1
+                        # Use data available on the proto span for context
+                        try:
+                            trace_id = proto_span.trace_id.hex()
+                            span_id = proto_span.span_id.hex()
+                            name = getattr(proto_span, "name", "")
+                        except Exception:
+                            trace_id = ""
+                            span_id = ""
+                            name = ""
```
andrewtruong, line 241: Is it possible that `trace_id.hex()` is valid but `span_id.hex()` is not?  I wonder if you should have separate try/excepts to capture the most data

zbirenbaum: I don't think this should ever occur. Trace and span ID should always be defined, but in the case that it does instead of crashing we just get something non descriptive. I'm pretty defensive with OTEL stuff because people have such varied setups.

zbirenbaum: I'm going to go ahead and merge and if I think it requires a second pass do so, but this is a pretty big issue and  is required before  we can add failed spans to the UI

PR #5512, opened 2025-09-24
fix(weave): better feedback field validation

gtarpenning:
## Description

[WB-27642](https://wandb.atlassian.net/browse/WB-27642)

Fixes bug in field validation logic that was allowing any param with a column substring to hit the db. 

To do so we needed to add a way of bypassing the `_transform_external_field_to_internal_field` which gets called on more than just external fields it turns out. All of the cost/token query fields are hitting this function, which really makes zero sense!

## Testing

add test 

and tested locally, costs appear fine:
Screenshot 2025-09-25 at 9 13 34 AM

[WB-27642]: https://wandb.atlassian.net/browse/WB-27642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/orm.py
+++ b/weave/trace_server/orm.py
@@ -223,7 +223,10 @@ class Select:
     action: Action
 
     _project_id: typing.Optional[str]
+    # Fields from the user that must be transformed to internal field names
+    # like "inputs.my_field" or "payload.value"
     _fields: typing.Optional[list[str]]
+    # Fields that we have constructed internally, like cost query fields
     _raw_sql_fields: typing.Optional[list[str]]
     _query: typing.Optional[tsi.Query]
     _order_by: typing.Optional[list[tsi.SortBy]]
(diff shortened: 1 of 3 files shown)


weave/trace_server/orm.py
```diff
@@ -223,7 +223,11 @@ class Select:
     action: Action
 
     _project_id: typing.Optional[str]
+    # Fields from the user that must be transformed to internal field names
+    # like "inputs.my_field" or "payload.value"
     _fields: typing.Optional[list[str]]
+    # Fields that we have constructed internally, like cost query fields
+    _raw_sql_fields: typing.Optional[list[str]]
```
gtarpenning, line 230: @jwlee64 more than this?

weave/trace_server/orm.py
```diff
@@ -260,6 +265,11 @@ def fields(self, fields: typing.Optional[list[str]]) -> "Select":
         self._fields = fields
         return self
 
+    def raw_sql_fields(self, raw_fields: typing.Optional[list[str]]) -> "Select":
```
jwlee64, line 268: i was thinking at least here, and maybe a comment about how we dont want to transform the cost query fields

PR #5516, opened 2025-09-25
total hack job but informative

gtarpenning:
wip

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -94,6 +94,9 @@ from weave.trace_server.interface.feedback_types import (
     runnable_feedback_runnable_ref_selector,
 )
 from weave.trace_server.trace_server_interface import (
+    CallBatchEndMode,
+    CallBatchStartMode,
+    CallCreateBatchReq,
     CallEndReq,
     CallsDeleteReq,
     CallsFilter,
@@ -345,6 +348,9 @@ class WeaveClient:
         if hasattr(self.server, "get_feedback_processor"):
             self._server_feedback_processor = self.server.get_feedback_processor()
         self.send_file_cache = WeaveClientSendFileCache()
+
+        # Cache to store call starts so they can be sent with call ends
+        self._call_start_cache: dict[str, CallStartReq] = {}
 
     ################ High Level Convenience Methods ################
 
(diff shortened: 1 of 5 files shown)


weave/trace/weave_client.py
```diff
@@ -346,6 +350,12 @@ def __init__(
             self._server_feedback_processor = self.server.get_feedback_processor()
         self.send_file_cache = WeaveClientSendFileCache()
 
+        # Cache to store call starts so they can be sent with call ends
+        self._call_start_cache: TTLCache[str, CallStartReq] = TTLCache(
```
andrewtruong: Why is this a TTLCache?  What if the call lasts longer than the TTL time?

weave/trace/weave_client.py
```diff
@@ -960,9 +972,23 @@ def send_end_call() -> None:
...
+            if cached_start is not None:
+                # Send complete call (start + end)
+                batch_req = CallCreateBatchReq(
+                    batch=[
+                        CallBatchStartMode(req=cached_start),
+                        CallBatchEndMode(req=call_end_req),
+                    ]
+                )
+                self.server.call_start_batch(batch_req)
+                del self._call_start_cache[call.id]
```
andrewtruong: it's more idiomatic to do a `cached_start = self._call_start_cache.pop(call.id)` here (and no `del` below), though I'm not sure if the cache impl you're using supports it

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -82,6 +82,9 @@ def as_sql(
     def as_select_sql(self, pb: ParamBuilder, table_alias: str) -> str:
         return f"{self.as_sql(pb, table_alias)} AS {self.field}"
 
+    def is_heavy(self) -> bool:
```
andrewtruong, line 85: is this for something in the future?

weave/trace/weave_client.py
gtarpenning: I wanted max-size control, and in general I think the vast vast majority of calls are finishing in an hour. Exceptions being long eval runs. 

Because i'm committing to support all old clients forever, there will be a fallback read before write in the trace server when it sees call ends without call starts. The cache should cover 99% of cases, read-before-write handles the rest. 

TTLCache will also clean up broken runs that fail and would otherwise just clog up and cause OOM for longrunning clients.

weave/trace/weave_client.py
gtarpenning: yeah thats ai slop, I am generally very wary of pythons `del` statement

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: nah my linter says that we need that field because of the class def and LLMs insist on implementing. This is an existing paradigm

PR #5526, opened 2025-10-03
chore(weave): Added `wb_run_step_end` to DB, interface, query, and write layer.

tssweeney:
Currently, we track `wb_run_step` on the call - but this is the run step at the _start_ of the call. This means we lose information about what steps if any happened during the call. For example:

```
@weave.op
def inner():
   wandb.log({"a": 2})

@weave.op
def outer():
   wandb.log({"a": 1})
   inner()
   wandb.log({"a": 3})

outer()
```

In this case, we will know what `outer` started with step 0 and `inner` started with step 1, but we don't know that step 2 happened inside of outer. When constructing the historical data for a session, we want to know the range of steps for an op's execution. This is critical for reconstructing a true timeline of events as we merge Models and Weave.

Note: the current field is called `wb_run_step` and i added `wb_run_step_end`. I wish we called the original `wb_run_step_start`, but i wanted to keep any migration/renaming a separate activity.

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -309,6 +309,8 @@ def simple_line_call_bootstrap() -> OpCallSpec:
                 part_sequence.append(("end", name))
                 return res
 
+            wrapped.__name__ = fn.__name__
+
             return wrapped
 
         return wrapper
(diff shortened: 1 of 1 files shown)


tssweeney:
> LGTM. I agree that the name change is a separate issue, although by adding the changes in `weave/trace/call.py` are we not making it harder to fix? I guess it will require some back compat anyway...
> 
> Aside: we never surface the run step information to the user, so I can assume this change is entirely for querying? As in, run step info is not selectable, so we will never return it from a calls query.

This is setting up the storage and query layer in anticipation for some UI mechanics that leverage it

PR #5527, opened 2025-10-03
chore(weave): Update Imperative Eval API to support new objects endpoints

andrewtruong:
https://wandb.atlassian.net/browse/WB-28283

Updates the EvaluationLogger API to support newly added objects endpoints


weave/evaluation/eval_imperative.py
andrewtruong, line 1: It looks like a lot has changed, but the contents of this file have just been moved to eval_imperative_v1, and new routing logic is added here

PR #5528, opened 2025-10-03
Griffin/custom wb stats query

gtarpenning:
## Description

[WB-28058](https://wandb.atlassian.net/browse/WB-28058)

This is technically not correct, and will not filter out deleted rows that have not been properly aggregated in the background by clickhouse. However, because this is already just a rough heuristic, I think it's way better to be fast and right 99% of the time instead of always slow. 

## Testing

adds test

### DIFF
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -3923,3 +3923,33 @@ def test_table_create_from_digests(network_proxy_client):
     assert shuffled_res.digest != original_digest, (
         f"Different row order should produce different digest: {shuffled_res.digest} vs {original_digest}"
     )
+
+
+def test_calls_query_with_wb_run_id_not_null(client, monkeypatch):
+    """Test optimized stats query for wb_run_id not null."""
+    # Mock wandb to simulate a run
+    import weave.trace.weave_client as wc
+
+    mock_run_id = "entity/project/test_run_123"
+    monkeypatch.setattr(wc, "safe_current_wb_run_id", lambda: mock_run_id)
+
+    @weave.op
+    def test_op(x: int) -> int:
+        return x * 2
+
+    # Create a call with wb_run_id
+    test_op(5)
+    client.flush()
+
+    # Query for calls with wb_run_id not null using limit=1 to trigger optimization
+    query = tsi.Query(
+        **{
+            "$expr": {
+                "$not": [{"$eq": [{"$getField": "wb_run_id"}, {"$literal": None}]}]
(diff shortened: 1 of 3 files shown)


weave/trace_server/sqlite_trace_server.py
```diff
@@ -370,8 +370,14 @@ def process_operation(operation: tsi_query.Operation) -> str:
...
-                    rhs_part = process_operand(operation.eq_[1])
-                    cond = f"({lhs_part} = {rhs_part})"
+                    if (
+                        isinstance(operation.eq_[1], tsi_query.LiteralOperation)
+                        and operation.eq_[1].literal_ is None
+                    ):
+                        cond = f"({lhs_part} IS NULL)"
+                    else:
+                        rhs_part = process_operand(operation.eq_[1])
+                        cond = f"({lhs_part} = {rhs_part})"
```
gtarpenning, line 383: we have not been handling `IS NOT NULL` conditions correctly in sqlite... forever... The test exposed the issue

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1618,33 +1618,35 @@ def process_calls_filter_to_conditions(
...
+
+    # Try optimized special case queries first
+    if opt_query := _try_optimized_stats_query(req, param_builder):
+        return (opt_query, aggregated_columns.keys())
+
+    # Fall back to general query builder
     cq = CallsQuery(
         project_id=req.project_id,
-        include_total_storage_size=req.include_total_storage_size,
+        include_total_storage_size=req.include_total_storage_size or False,
```
gtarpenning, line 1650: linter was complaining about this, i think it's fine either way

tests/trace/test_weave_client.py
```diff
@@ -3923,3 +3923,44 @@ def test_table_create_from_digests(network_proxy_client):
...
+        }
+    )
+    calls = list(
+        client.server.calls_query(
+            tsi.CallsQueryReq(project_id=client._project_id(), query=query, limit=1)
+        ).calls
+    )
+
+    assert len(calls) == 1
+    assert calls[0].wb_run_id == mock_run_id
```
chance-wnb, line 3966: with or without your optimization, this test case both passes right? so what is the point of the testing?

Is testing from a query builder perspective easier?

gtarpenning: we previously were not actually testing this path!

PR #5531, opened 2025-10-03
chore(weave): Claude setup in weave

tssweeney:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,0 +1,1 @@
+Refer to @AGENTS.md for agentic LLM instructions.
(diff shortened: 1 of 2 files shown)


.claude/CLAUDE.md
andrewtruong, line 1: this can be a symlink

tssweeney: it is recommended by anthropic docs to use a reference like this in case there are additional claude-specific instructions to put here.

PR #5534, opened 2025-10-07
chore(weave): Pin pydantic to unblock CI

andrewtruong:
Pins `pydantic<2.12.0` as the latest version seems to break a lot of things


chance-wnb (approved):
Thanks!

PR #5540, opened 2025-10-08
feat(weave): Add new CRUD endpoints for interacting with Ops

andrewtruong:
https://wandb.atlassian.net/browse/WB-28278

This PR adds CRUD endpoints for interacting with Ops as part of the V2 Evals API.

Pairs with https://github.com/wandb/core/pull/35081

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -22,10 +22,17 @@ from clickhouse_connect.driver.summary import QuerySummary
 from opentelemetry.proto.collector.trace.v1.trace_service_pb2 import (
     ExportTraceServiceRequest,
 )
+from tenacity import (
+    retry,
+    retry_if_exception_type,
+    stop_after_attempt,
+    wait_exponential,
+)
 
 from weave.trace_server import clickhouse_trace_server_migrator as wf_migrator
 from weave.trace_server import clickhouse_trace_server_settings as ch_settings
 from weave.trace_server import environment as wf_env
+from weave.trace_server import object_creation_utils
 from weave.trace_server import refs_internal as ri
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.actions_worker.dispatcher import execute_batch
@@ -157,7 +164,7 @@ logger.setLevel(logging.INFO)
 _CH_POOL_MANAGER = get_pool_manager(maxsize=50, num_pools=2)
 
 
-class ClickHouseTraceServer(tsi.TraceServerInterface):
+class ClickHouseTraceServer(tsi.FullTraceServerInterface):
     def __init__(
         self,
         *,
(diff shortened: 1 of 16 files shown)


weave/trace_server/reference/generate.py
```diff
@@ -494,4 +495,61 @@ def threads_query_stream(
...
+
+    @router.post("/v2/op/delete", tags=[V2_OPS_TAG_NAME])
+    def op_delete_v2(
+        req: tsi.OpDeleteV2Req,
+        service: weave.trace_server.trace_service.TraceService = Depends(get_service),  # noqa: B008
+    ) -> tsi.OpDeleteV2Res:
+        """Delete an op object."""
+        return service.trace_server_interface.op_delete_v2(req)
+
     return router
```
neutralino1, line 597: As posted here https://github.com/wandb/core/pull/34978#discussion_r2415124117

The typical RESTful endpoints usually look like

```
# Create
POST /ops

# Get one
GET /ops/<id>

# List
GET /ops

# Update
PUT /ops/<id>

# Delete
DELETE /ops/<id>
```
i.e. plural resource name, and the action is contained in the HTTP method, not in the path.

Your doc seemed to also adopt this standard. Why the change now?

andrewtruong: Oops, good point!

weave/trace_server/trace_server_interface.py
```diff
@@ -1248,6 +1243,102 @@ class EvaluationStatusRes(BaseModel):
...
+
+class OpCreateV2Req(OpCreateV2Body):
+    """Request model for creating an Op object.
+
+    Extends OpCreateV2Body by adding project_id for internal API usage.
+    """
+
+    project_id: str = Field(
+        ..., description="The project where this object will be saved"
+    )
```
bcsherma, line 1269: why are you breaking this into hierarchical models for body and whole request? why not just have a type for the request like all other requests?

andrewtruong: This is because we need the `project_id` in the path for this API.  If you also include the `project_id` in the request body, then it becomes redundant and confusing -- what happens if the body `project_id` is different?

See how it's used here:
https://github.com/wandb/weave/pull/5540/files#diff-becc31f22ba0b0d9e7b7adb0c5e4d91f1d12da3f4089a98065c5e248b9957a8eR498-R505

weave/trace_server/reference/generate.py
```diff
@@ -494,4 +495,102 @@ def threads_query_stream(
             media_type="application/jsonl",
         )
 
+    @router.post(
+        "/v2/{entity}/{project}/ops",
+        tags=[V2_OPS_TAG_NAME],
+    )
+    def op_create_v2(
```
neutralino1: Thoughts about namespacing the module, i.e. storing these endpoints into a `v2/` directory, but keep the function names canonical such as `def create_op`?

Also, a general best practice is for Python function to start with a verb (i.e. the action performed), meaning `create_op` is more canonical than `op_create`.

weave/trace_server/reference/generate.py
```diff
@@ -494,4 +495,102 @@ def threads_query_stream(
...
+        """Create an op object."""
+        project_id = f"{entity}/{project}"
+        req = tsi.OpCreateV2Req(project_id=project_id, **body.model_dump())
+        return service.trace_server_interface.op_create_v2(req)
+
+    @router.get(
+        "/v2/{entity}/{project}/ops/{object_id}/{digest}",
+        tags=[V2_OPS_TAG_NAME],
+    )
+    def op_read_v2(
```
neutralino1:
```suggestion
    def read_op(
```

weave/trace_server/reference/generate.py
```diff
@@ -494,4 +495,102 @@ def threads_query_stream(
...
+        entity: str,
+        project: str,
+        object_id: str,
+        digest: str,
+        service: weave.trace_server.trace_service.TraceService = Depends(get_service),  # noqa: B008
+    ) -> tsi.OpReadEagerV2Res:
+        """Get an op object with resolved code (eager loading)."""
+        project_id = f"{entity}/{project}"
+        req = tsi.OpReadV2Req(project_id=project_id, object_id=object_id, digest=digest)
+        return service.trace_server_interface.op_read_eager_v2(req)
```
neutralino1: Can you explain what eager-loading is here?

weave/trace_server/reference/generate.py
```diff
@@ -494,4 +495,102 @@ def threads_query_stream(
...
+                        "schema": {
+                            "type": "array",
+                            "items": {"$ref": "#/components/schemas/Schema"},
+                        }
+                    }
+                },
+            }
+        },
+    )
+    def op_list_v2(
```
neutralino1:
```suggestion
    def list_ops(
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1228,6 +1255,278 @@ def threads_query_stream(
...
+        object_query_builder.add_is_op_condition(True)
+        object_query_builder.add_object_ids_condition([req.object_id])
+        object_query_builder.add_digests_conditions(req.digest)
+        object_query_builder.set_include_deleted(include_deleted=True)
+        objs = self._select_objs_query(object_query_builder)
+        if len(objs) == 0:
+            raise NotFoundError(f"Op {req.object_id}:{req.digest} not found")
+
+        # There should not be multiple ops returned, but in case there are, just
+        # return the first one.
```
neutralino1, line 1337: Should we raise here?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1228,6 +1255,278 @@ def threads_query_stream(
...
+            created_at=_ensure_datetimes_have_tz(obj.created_at),
+            code_ref=code_ref,
+        )
+
+    def op_read_eager_v2(self, req: tsi.OpReadV2Req) -> tsi.OpReadEagerV2Res:
+        """Get a specific op object with resolved code.
+
+        This is similar to op_read_v2 but makes the additional API call to
+        resolve the code reference and return the actual source code.
+        """
```
neutralino1: Ok, why do we call this eager? Is it because otherwise the client would have to make a second request to get the code?

Do we need this endpoint?

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -460,3 +460,27 @@ def evaluation_status(
...
+        req.project_id = self._idc.ext_to_int_project_id(original_project_id)
+        return self._ref_apply(self._internal_trace_server.op_read_eager_v2, req)
+
+    def op_list_v2(self, req: tsi.OpListV2Req) -> Iterator[tsi.OpReadV2Res]:
+        req.project_id = self._idc.ext_to_int_project_id(req.project_id)
+        return self._stream_ref_apply(self._internal_trace_server.op_list_v2, req)
+
+    def op_delete_v2(self, req: tsi.OpDeleteV2Req) -> tsi.OpDeleteV2Res:
+        req.project_id = self._idc.ext_to_int_project_id(req.project_id)
+        return self._ref_apply(self._internal_trace_server.op_delete_v2, req)
```
neutralino1, line 483: I wish we didn't need to write this silly boilerplate, can we do this automatically as some sort of middleware for all requests?

weave/trace_server/object_creation_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+def get_placeholder_op_source() -> str:
+    """Get placeholder source code for an Op.
+
+    Returns:
+        Python source code for a placeholder op function
+    """
+    return """def func(*args, **kwargs):
+    ... # Code-capture unavailable for this op
+"""
```
neutralino1: Why not a constant?

weave/trace_server_bindings/caching_middleware_trace_server.py
```diff
@@ -531,6 +531,31 @@ def evaluation_status(
...
+        return self._with_cache_pydantic(
+            self._next_trace_server.op_read_eager_v2, req, tsi.OpReadEagerV2Res
+        )
+
+    def op_list_v2(self, req: tsi.OpListV2Req) -> Iterator[tsi.OpReadV2Res]:
+        return self._next_trace_server.op_list_v2(req)
+
+    def op_delete_v2(self, req: tsi.OpDeleteV2Req) -> tsi.OpDeleteV2Res:
+        return self._next_trace_server.op_delete_v2(req)
+
```
neutralino1, line 551: Same here, it's silly that we need this boilerplate. We should just add a decorator on the cacheable endpoint functions and that's it.

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -694,6 +749,79 @@ def evaluation_status(
     ) -> tsi.EvaluationStatusRes:
         raise NotImplementedError("evaluation_status is not implemented")
 
+    # === V2 APIs ===
```
neutralino1, line 752: Should these be generated by Stainless?

weave/trace_server/reference/generate.py
andrewtruong: That was my initial approach, but I stopped short because it involved refactoring a bunch of code to share utils.

If you think it's cleaner, we can try again with a dedicated v2 file that contains just the v2 router and endpoints

andrewtruong: RE: naming, I think it's easier to scan when the first few chars are the same, e.g. `op_{action}`, but I understand that the prose is a bit clunkier

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: No, I think that's overly defensive.

weave/trace_server/reference/generate.py
andrewtruong: Eager loading means we make all of the internal calls in one endpoint.  This avoids multiple network hops

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: It's an optimization I'm considering -- for some objects, there will be multiple calls to the network that could otherwise be made in just 1 call.

If you imagine a heavy object loading case, like Evaluation, loading it today involves several round trip network requests:
1. Get the Evaluation object
2. (3x) Get the Evaluation object's ops
3. (3x) Get each op's source code

An eager (or maybe "heavy") endpoint would make all of those internal requests and just return the final payload in one network request

weave/trace_server/external_to_internal_trace_server_adapter.py
andrewtruong: I think the only distinction is delegating based on whether the endpoint is an iterator or not.  I think that can be a separate PR.

weave/trace_server/object_creation_utils.py
andrewtruong: It could be

weave/trace_server_bindings/caching_middleware_trace_server.py
andrewtruong: Agreed, but I think that's also outside the scope of this PR

weave/trace_server/clickhouse_trace_server_batched.py
neutralino1: Could this be enabled with a param in the URL instead of a separate endpoint? E.g.
```
GET /v2/foo/bar/ops/my_op/abc123?load_code=true
```
of `=false` depending on the default behavior we choose.

andrewtruong: So the payload is variable depending on how you set that param?

I wonder if it makes sense to have a payload like this (open question, not stating an opinion):
```py
class Evaluation(BaseModel):
    predict_and_score: Ref | OpReadV2Res
```

neutralino1: It's not unusual to have a return payload depend on query params.

I feel like it would be easier to manage than two different endpoints.

weave/trace_server/object_creation_utils.py
neutralino1: Can you make it one?

weave/trace_server_bindings/caching_middleware_trace_server.py
neutralino1: Agreed.

weave/trace_server/trace_server_interface.py
```diff
@@ -1248,6 +1243,102 @@ class EvaluationStatusRes(BaseModel):
...
+
+
+class OpDeleteV2Res(BaseModel):
+    num_deleted: int = Field(
+        ..., description="Number of op versions deleted from this op"
+    )
+
+
 class TraceServerInterface(Protocol):
```
chance-wnb, line 1328: Why don't we create TraceServerInterfaceV2 ? And make a server that implements TraceServerInterfaceV2. On the client side (TS and SDK), we could switch between the V1 server and V2 server as needed.

So the method names of `op_*_v2` can be just `op_*`. I think that's more organic and cleaner.

I hope we don't have to carry the historic burden and can delete the old code in the future once and for all.

andrewtruong: Added a `TraceServerInterfaceV2` and migrated existing impls to use a new `FullTraceServerInterface` that implements both V1 and V2

chance-wnb: Thanks Andrew. 


Is it possible, for example, in weave_init.py, based on a feature flag(eg, local env var), we selectively start a `TraceServerInterfaceV2` only server instance.

So 
```
    remote_server = init_weave_get_server(api_key)
    server: FullTraceServerInterface = remote_server
```
becomes
```
    if flag:
        remote_server = init_weave_get_server_v2(api_key)
        server: TraceServerInterfaceV2 = remote_server
    else:
        remote_server = init_weave_get_server(api_key)
        server: TraceServerInterface = remote_server
```

then
1. we can have a dedicated `clickhouse_trace_server_batched_v2.py` that focuses on implementing the `TraceServerInterfaceV2` interface. I think since you already designed the new `reference` server, why don't you just leverage that and start the new set clean?
2. We will have the freedom to name `op_*_v2` to be just `op_*` in the TraceServerInterfaceV2 interface.

Can you help me understand why do you have to combine the methods into a whole `FullTraceServerInterface`?

weave/trace_server/sqlite_trace_server.py
```diff
@@ -1613,6 +1614,150 @@ def evaluation_status(
     ) -> tsi.EvaluationStatusRes:
         return evaluation_status(self, req)
 
+    def op_create_v2(self, req: tsi.OpCreateV2Req) -> tsi.OpCreateV2Res:
```
chance-wnb, line 1649: For clientside, if we detect that the backend is a SQLite server that we talk to, we can choose to use the V1 implementation (v1 server_client). 

Can we take this chance to get rid of implementing towards future SQLite feature for V2? Maintaining the SQLIte thingy which no one actually uses is a nightmare. 

1. In the future falling back to V1 FE /sdk in case of SQLite might be acceptable. 

2. For those who are really enthusiastic about SQLite backend, let them pin to the old version SDK.

PR #5542, opened 2025-10-08
feat(weave): ImperativeEval in TS SDK Part 1

chance-wnb:
# Imperative Evaluation Logger for TypeScript SDK

## JIRA:

https://wandb.atlassian.net/browse/WB-28160

## Summary
This PR implements **Imperative Evaluation Logger** for the TypeScript SDK, providing an alternative to the traditional batch-oriented `Evaluation` class. This feature allows incremental logging of predictions and scores as they happen, without requiring a fixed dataset or batch processing.

This implementation achieves feature parity with the Python SDK's `eval_imperative.py`.

## Motivation

**This feature already exists in the Python SDK but was missing from the TypeScript SDK.** This PR brings feature parity between the two SDKs.

Traditional evaluation in Weave requires:
- Upfront dataset definition
- Batch processing of all examples
- All scorers defined before evaluation starts

However, many real-world use cases need:
- Streaming data sources (e.g., live production traffic, user sessions)
- Dynamic/conditional scoring (different examples may use different scorers)
- Incremental evaluation results
- Flexible evaluation workflows

**Imperative Evaluation Logger** solves these problems by allowing users to log predictions one at a time, add scores dynamically, and generate summaries on demand.

### Core Components

1. **`EvaluationLogger`** - Main entry point for imperative evaluation
   - Creates root `evaluate` call on construction
   - Manages prediction lifecycle
   - Generates auto-summary or accepts custom summary
   - Tracks all predictions for summary calculation

2. **`ScoreLogger`** - Manages scoring for a single prediction
   - Returned from `logPrediction()`
   - Allows multiple scores to be logged incrementally
   - Creates scorer calls with feedback attachment
   - Finalizes prediction with `finish()`

3. **WeaveObject Classes** (Internal)
   - `Evaluation` - Root evaluation object
   - `Model` - Model wrapper object
   - `Scorer` - Dynamically created per scorer name

4. **Built-in Ops** (Internal)
   - `Evaluation.evaluate` - Root evaluation op
   - `Evaluation.predict_and_score` - Prediction container op
   - `Evaluation.summarize` - Summary generation op
   - `Model.predict` - Prediction op
   - `{scorer_name}.score` - Scorer ops (dynamically created)

### Call Hierarchy
```
evaluate (root)
├── predict_and_score (per prediction)
│   ├── predict (finished immediately)
│   └── {scorer_name}.score (per score, with feedback attachment)
├── predict_and_score (next prediction)
│   └── ...
└── summarize (finalization)
```

## API Usage

```typescript
import {EvaluationLogger} from '@wandb/weave';

// Create evaluation logger
const ev = new EvaluationLogger({
  name: 'my-eval',
  description: 'Evaluating my model',
  dataset: 'my-dataset', // optional
  scorers: ['accuracy', 'f1'], // optional, for warnings only
  model: myModel, // optional
  attributes: {trial_id: 'abc123'} // optional custom attributes
});

// Log predictions incrementally
for (const example of streamingData) {
  const output = await myModel.predict(example);

  // Log prediction and get scorer
  const pred = await ev.logPrediction(example, output);

  // Conditionally add scores
  if (shouldScore(output)) {
    await pred.logScore('accuracy', calculateAccuracy(output));
    await pred.logScore('f1', calculateF1(output));
  }

  // Finalize prediction
  await pred.finish();
}

// Generate summary (auto-calculates mean/true_fraction)
await ev.logSummary();

// Or provide custom summary
await ev.logSummary({
  accuracy: {mean: 0.95},
  custom_metric: 'custom_value'
});
```

#### 1. **Direct Call Management vs. Op System Coordination**

**The Critical Choice:**

The Python SDK achieves imperative evaluation by coordinating the Op system (`__require_explicit_finish`, call stack manipulation, context variables). We **rejected this approach** in favor of **direct call management** using the existing `createCall()`/`finishCall()` API.

**Why Direct Call Management?**
- **40% faster implementation** (6-9 days vs 10-15 days) (Chance: Well the actual time took about 2 days)
- **Zero changes to Op system** - no risk of breaking existing functionality
- **Explicit parent-child relationships** via `createCall(..., parentEntry)`
- **Simple metadata tracking** instead of complex stack manipulation
- **Better debugging** - clear call hierarchy in code, not implicit in framework

**Trade-off:** Requires manual call lifecycle management instead of automatic Op coordination, but this is actually **simpler** and more maintainable.

#### 2. **Metadata-Driven Architecture**

Instead of coordinating Ops, we track call relationships in lightweight metadata:

```typescript
class PredictAndScoreCallMetadata {
  predictAndScoreCall: InternalCall;       // The container call
  predictAndScoreEntry: CallStackEntry;    // Call stack entry
  evaluateEntry: CallStackEntry;           // Parent reference
  predictCallId: string;                   // For feedback attachment
  scores: Record<string, any> = {};        // Accumulated scores
  output: any;                             // Prediction output
  isFinished: boolean = false;             // Lifecycle state
}
```

**Why this matters:**
- **Explicit state tracking** - no hidden framework state
- **Clear parent-child relationships** - set in `createCall()` not via stack manipulation
- **Flexible timing** - finish calls whenever ready, not tied to Op execution
- **Easy debugging** - inspect metadata to see all call relationships

#### 3. **Built-in SDK Ops (Not Dynamic Op Creation)**

We define imperative evaluation Ops as **built-in SDK operations** that are never executed:

```typescript
const evaluationEvaluate = op(
  async function evaluate(evaluation: any, model: any): Promise<any> {
    // Never executed - only for UI display and call structure
    throw new Error('Use EvaluationLogger instead of calling directly');
  },
  {name: 'Evaluation.evaluate'}
);
```

**Why this approach?**
- **Auto-saved by createCall()** - no manual save/publish needed
- **Shared across all instances** - zero per-instance overhead
- **Source code captured** - displayed in Weave UI for documentation
- **Type-safe** - proper TypeScript signatures
- **Internal to module** - not exposed in public API

**Alternative rejected:** Dynamic Op creation per EvaluationLogger instance (expensive, complex lifecycle).

#### 4. **Internal WeaveObject Classes**

Created dedicated `Evaluation`, `Model`, `Scorer` classes **internal to evaluationLogger**:

```typescript
class Evaluation extends WeaveObject {
  dataset?: Dataset<any> | string;
  scorers: string[];
  // Auto-serializes with class_name: 'Evaluation'
}
```

**Why internal classes?**
- **Class name = serialized name** - `class Evaluation` → `class_name: 'Evaluation'` in trace
- **No naming conflicts** - doesn't interfere with existing `Evaluation` class in batch eval
- **Auto-saved by SDK** - `saveWeaveValues()` recursively saves all WeaveObjects in call inputs
- **Flexible attributes** - copy any custom fields (trial_id, etc.) into object

**Key insight:** The SDK's existing `saveWeaveValues()` infrastructure handles all object persistence - we just need to pass WeaveObjects in call inputs.

#### 5. **Scorer Feedback via Finished Predict Call**

Scorers attach results as **feedback** to the already-finished `predict` call:

```typescript
// In logPrediction(): finish predict immediately
await createAndFinishCall(..., modelPredict, ...);
const predictCallId = predictEntry.callId;

// In logScore(): attach feedback to finished predict call
await client.addFeedback(
  predictCallId,        // Target: finished predict call
  scorerEntry.callId,   // Source: scorer call
  scorerRefUri,
  score
);
```

**Why finish predict immediately?**
- **Captures accurate timing** - predict duration reflects actual prediction time
- **Enables feedback attachment** - can only attach feedback to finished calls
- **Matches Python behavior** - scorer results are feedback, not child calls

**Alternative rejected:** Keep predict call open until finish() - would show incorrect durations.

#### 6. **Deferred Finishing for predict_and_score**

The `predict_and_score` call starts in `logPrediction()` but finishes in `finish()`:

```typescript
// In logPrediction():
const startPromise = client.createCall(...);  // Start predict_and_score
await startPromise;
// Return ScoreLogger - call still open

// In finish():
await client.finishCall(..., startPromise);   // Finish with accumulated scores
```

**Why defer finishing?**
- **Accumulate scores** - multiple `logScore()` calls add to `scores` map
- **Correct output** - finish with `{output, scores}` tuple
- **User controls timing** - finish when ready, not tied to framework

**Critical detail:** Must save `startPromise` from `createCall()` and pass to `finishCall()` - required by SDK's call matching logic.

#### 7. **Auto-Summary Generation**

When `logSummary()` called without arguments, auto-generates summary from accumulated scores:

```typescript
private generateAutoSummary(): Record<string, any> {
  // For numeric scores: calculate mean
  if (values.every(v => typeof v === 'number')) {
    summary[scoreName] = {mean: sum / values.length};
  }
  // For boolean scores: calculate true_count and true_fraction
  else if (values.every(v => typeof v === 'boolean')) {
    summary[scoreName] = {
      true_count: trueCount,
      true_fraction: trueCount / values.length
    };
  }
}
```

**Why auto-generate?**
- **Convenience** - users don't need to compute manually
- **Python parity** - matches Python SDK behavior
- **Flexibility** - users can override with custom summary

#### 8. **Async Safety Warning**

Warn if `logSummary()` called with unfinished predictions:

```typescript
const unfinishedCount = this.predictions.filter(p => !p.isFinished).length;
if (unfinishedCount > 0) {
  console.warn(
    `logSummary() called with ${unfinishedCount} unfinished prediction(s). ` +
    `Make sure to call finish() on all ScoreLoggers before calling logSummary().`
  );
}
```

**Why this matters:**
- **Catches user errors** - forgetting to await finish()
- **Prevents incomplete summaries** - scores from unfinished predictions excluded
- **Async timing issues** - JavaScript is single-threaded but async can cause race conditions

### Changes to Existing Code

**`weaveClient.ts`** (sdks/node/src/weaveClient.ts:574-647)
- Added `addFeedback()` method for attaching scorer feedback to predict calls
- Exports `CallStackEntry` type for internal use by evaluationLogger

**`inMemoryTraceServer.ts`** (sdks/node/src/inMemoryTraceServer.ts:173-182)
- Added `feedback` API stub for test infrastructure
- Returns mock feedback IDs for testing

**`index.ts`** (sdks/node/src/index.ts)
- Exports `EvaluationLogger` and `ScoreLogger` classes

**`package.json`**
- Version bump for release

## Testing

Created comprehensive test suite with **10 tests across 5 categories**:

1. **Basic Functionality** (2 tests)
   - Complete evaluation lifecycle
   - Full configuration options with custom attributes

2. **Call Hierarchy** (1 test)
   - Verifies correct parent-child relationships in call tree

3. **Attribute Markers** (2 tests)
   - `IMPERATIVE_EVAL_MARKER` on evaluation calls
   - `IMPERATIVE_SCORE_MARKER` on scorer calls

4. **Summary Generation** (3 tests)
   - Auto-summary calculates `mean` for numeric scores
   - Auto-summary calculates `true_count` and `true_fraction` for boolean scores
   - Custom summary overrides auto-summary

5. **Edge Cases** (3 tests)
   - Prediction with no scores
   - Summary with no predictions
   - Multiple predictions sequentially

**Test Infrastructure:**
- Uses `InMemoryTraceServer` for mocking
- Helper function `getCalls()` with built-in `waitForPendingOperations()`
- All tests pass ✅

## Files Changed

**New Files:**
- `sdks/node/src/evaluationLogger.ts` - 749 lines - Core implementation
- `sdks/node/src/__tests__/evaluationLogger.test.ts` - 303 lines - Test suite

**Modified Files:**
- `sdks/node/src/weaveClient.ts` - Added `addFeedback()` method
- `sdks/node/src/inMemoryTraceServer.ts` - Added feedback API stub
- `sdks/node/src/index.ts` - Export new classes
- `sdks/node/src/weaveObject.ts` - Removed unused code
- `sdks/node/package.json` - Version bump

**Total:** +1,129 lines, -15 lines

### What to Focus On

1. **Async correctness**: Verify the async flow in `logPrediction()` and `logSummary()`
2. **Call hierarchy**: Check that parent-child relationships are correct
3. **Feedback attachment**: Ensure scorer feedback properly attached to predict calls
4. **Auto-summary logic**: Review mean/true_fraction calculations
5. **API ergonomics**: Does the API feel natural for TypeScript users?

### Design Tradeoffs

- **No lock protection on `this.predictions` array**: Safe because JavaScript is single-threaded, but relies on users following correct async patterns
- **Warning-only scorers validation**: Unlike strict enforcement, we only warn about unrecognized scorers (matches Python SDK)
- **No Dataset creation in tests**: Simplified to use string dataset names to avoid testing Dataset functionality

## Related Work

- Python SDK: `weave/evaluation/eval_imperative.py`
- Design doc: [Internal design doc](https://www.notion.so/wandbai/Imperative-Evaluation-Logger-TypeScript-SDK-Design-Document-284e2f5c7ef3804baf18c34b3a7532ee?pvs=12)

## Checklist

- [x] Implementation complete with all core features
- [x] Test suite with 10 comprehensive tests (all passing)
- [x] Async safety warnings added
- [x] TypeScript idioms followed (`this` vs `self`)
- [x] Code cleanup (utility functions, removed redundancies)
- [x] Documentation in code comments
- [x] Exported from main SDK index

### DIFF
--- a/sdks/node/src/__tests__/evaluationLogger.test.ts
+++ b/sdks/node/src/__tests__/evaluationLogger.test.ts
@@ -37,8 +37,8 @@ describe('EvaluationLogger - Basic Functionality', () => {
       'output'
     );
 
-    await scoreLogger.logScore('accuracy', 0.95);
-    await scoreLogger.logScore('f1', 0.88);
+    scoreLogger.logScore('accuracy', 0.95);
+    scoreLogger.logScore('f1', 0.88);
     await scoreLogger.finish();
     await evalLogger.logSummary();
 
@@ -86,7 +86,7 @@ describe('EvaluationLogger - Call Hierarchy', () => {
       {input: 'test'},
       'output'
     );
-    await scoreLogger.logScore('accuracy', 0.95);
+    scoreLogger.logScore('accuracy', 0.95);
     await scoreLogger.finish();
     await evalLogger.logSummary();
 
@@ -147,7 +147,7 @@ describe('EvaluationLogger - Attribute Markers', () => {
       {input: 'test'},
       'output'
     );
-    await scoreLogger.logScore('accuracy', 0.95);
+    scoreLogger.logScore('accuracy', 0.95);
     await scoreLogger.finish();
     await evalLogger.logSummary();
 
@@ -176,7 +176,7 @@ describe('EvaluationLogger - Summary Generation', () => {
         {input: `test${i}`},
         `output${i}`
       );
-      await scoreLogger.logScore('accuracy', 0.9 + i * 0.05);
+      scoreLogger.logScore('accuracy', 0.9 + i * 0.05);
       await scoreLogger.finish();
     }
 
@@ -193,15 +193,15 @@ describe('EvaluationLogger - Summary Generation', () => {
     const evalLogger = new EvaluationLogger({name: 'test-eval'});
 
     const scoreLogger1 = await evalLogger.logPrediction({input: '1'}, 'out1');
-    await scoreLogger1.logScore('passed', true);
+    scoreLogger1.logScore('passed', true);
     await scoreLogger1.finish();
 
     const scoreLogger2 = await evalLogger.logPrediction({input: '2'}, 'out2');
-    await scoreLogger2.logScore('passed', false);
+    scoreLogger2.logScore('passed', false);
     await scoreLogger2.finish();
 
     const scoreLogger3 = await evalLogger.logPrediction({input: '3'}, 'out3');
-    await scoreLogger3.logScore('passed', true);
+    scoreLogger3.logScore('passed', true);
     await scoreLogger3.finish();
 
     await evalLogger.logSummary();
@@ -221,7 +221,7 @@ describe('EvaluationLogger - Summary Generation', () => {
       {input: 'test'},
       'output'
     );
-    await scoreLogger.logScore('accuracy', 0.95);
+    scoreLogger.logScore('accuracy', 0.95);
     await scoreLogger.finish();
 
     await evalLogger.logSummary({
@@ -287,7 +287,7 @@ describe('EvaluationLogger - Edge Cases', () => {
         {input: `test${i}`},
         `output${i}`
       );
-      await scoreLogger.logScore('accuracy', 0.9);
+      scoreLogger.logScore('accuracy', 0.9);
       await scoreLogger.finish();
     }
 
--- a/sdks/node/src/evaluationLogger.ts
+++ b/sdks/node/src/evaluationLogger.ts
@@ -19,7 +19,7 @@
 import {WeaveObject, WeaveObjectParameters} from './weaveObject';
 import {Dataset} from './dataset';
 import {op} from './op';
-import {getGlobalClient} from './clientApi';
+import {requireGlobalClient} from './clientApi';
 import {InternalCall} from './call';
 import {CallStackEntry} from './weaveClient';
 import {uuidv7} from 'uuidv7';
@@ -78,8 +78,6 @@ async function createAndFinishCall(
     displayName,
     attributes
   );
-
-  await startPromise;
 
   const endTime = new Date();
   await client.finishCall(
@@ -206,9 +204,7 @@ class Scorer extends WeaveObject {
 
 const evaluationEvaluate = op(
   async function evaluate(evaluation: any, model: any): Promise<any> {
-    // Note: This Op is never executed in imperative eval.
-    // Its source code is captured and displayed in the UI.
-    throw new Error('Use EvaluationLogger instead of calling directly');
+    // internal function
   },
   {name: 'Evaluation.evaluate'}
 );
@@ -218,15 +214,15 @@ const evaluationPredictAndScore = op(
     evaluation: any,
     model: any,
     example: Record<string, any>
-  ): Promise<{output: any; scores: Record<string, any>}> {
-    throw new Error('Use EvaluationLogger.logPrediction() instead');
+  ): Promise<any> {
+    // internal function
   },
   {name: 'Evaluation.predict_and_score'}
 );
 
 const evaluationSummarize = op(
-  async function summarize(evaluation: any): Promise<Record<string, any>> {
-    throw new Error('Use EvaluationLogger.logSummary() instead');
+  async function summarize(evaluation: any): Promise<any> {
+    // internal function
   },
   {name: 'Evaluation.summarize'}
 );
@@ -236,7 +232,7 @@ const modelPredict = op(
     model: any,
     inputs: Record<string, any>
   ): Promise<any> {
-    throw new Error('Use EvaluationLogger.logPrediction() instead');
+    // internal function
   },
   {name: 'Model.predict'}
 );
@@ -244,7 +240,7 @@ const modelPredict = op(
 const scorerScoreFactory = (scorerName: string) =>
   op(
     async function score(scorer: any, output: any, target?: any): Promise<any> {
-      throw new Error('Use ScoreLogger.logScore() instead');
+      // internal function
     },
     {name: `${scorerName}.score`}
   );
@@ -303,6 +299,7 @@ class PredictAndScoreCallMetadata {
  */
 export class ScoreLogger {
   private predMeta: PredictAndScoreCallMetadata;
+  private _scoringProcesses: Array<Promise<void>> = [];
 
   constructor(predMeta: PredictAndScoreCallMetadata) {
     this.predMeta = predMeta;
@@ -316,79 +313,80 @@ export class ScoreLogger {
    * @param score - The score value
    */
   async logScore(scorerName: string, score: any): Promise<void> {
-    if (this.predMeta.isFinished) {
-      throw new Error('Cannot log score after prediction is finished');
-    }
-
-    const client = getGlobalClient();
-    if (!client) {
-      console.warn('Weave not initialized, skipping score logging');
-      return;
-    }
-
-    // Dynamically create a Scorer class with the scorer name
-    // Override the class name so constructor.name === scorerName for proper serialization
-    class DynamicScorer extends Scorer {
-      constructor(parameters: ScorerParameters) {
-        super({
-          ...parameters,
-          name: scorerName,
-          scorerName: scorerName,
-        });
+    // Collect this logScore call in a promise so that we can await it later in finish()
+    const asyncProcess = (async () => {
+      if (this.predMeta.isFinished) {
+        throw new Error('Cannot log score after prediction is finished');
       }
-    }
 
-    // // Override the constructor name to be the scorer name
-    Object.defineProperty(DynamicScorer, 'name', {
-      value: scorerName,
-      writable: false,
-      configurable: true,
-    });
+      const client = requireGlobalClient();
 
-    // Create an instance of the dynamic scorer class
-    const scorer = new DynamicScorer({});
+      // Dynamically create a Scorer class with the scorer name
+      // Override the class name so constructor.name === scorerName for proper serialization
+      class DynamicScorer extends Scorer {
+        constructor(parameters: ScorerParameters) {
+          super({
+            ...parameters,
+            name: scorerName,
+            scorerName: scorerName,
+          });
+        }
+      }
 
-    // Save the scorer to get its URI for feedback attachment
-    const scorerRef = await client.publish(scorer);
-    const scorerRefUri = scorerRef.uri();
+      // Override the constructor name to be the scorer name
+      Object.defineProperty(DynamicScorer, 'name', {
+        value: scorerName,
+        writable: false,
+        configurable: true,
+      });
 
-    // Create and finish scorer call (already have the score)
-    const scorerCall = new InternalCall();
-    const scorerEntry: CallStackEntry = {
-      callId: uuidv7(),
-      traceId: this.predMeta.predictAndScoreEntry.traceId,
-      childSummary: {},
-    };
+      // Create an instance of the dynamic scorer class
+      const scorer = new DynamicScorer({});
 
-    await createAndFinishCall(
-      client,
-      scorerCall,
-      scorerScoreFactory(scorerName),
-      [{this: scorer, output: this.predMeta.output}],
-      'useParam0Object',
-      scorer,
-      scorerEntry,
-      this.predMeta.predictAndScoreEntry,
-      new Date(),
-      score,
-      `${scorerName}.score`,
-      IMPERATIVE_SCORE_MARKER
-    );
+      // Save the scorer to get its URI for feedback attachment
+      const scorerRef = await client.publish(scorer);
+      const scorerRefUri = scorerRef.uri();
 
-    // Attach the score as feedback to the predict call
-    await client.addFeedback(
-      this.predMeta.predictCallId,
-      scorerEntry.callId,
-      scorerRefUri,
-      score
-    );
+      // Create and finish scorer call (already have the score)
+      const scorerCall = new InternalCall();
+      const scorerEntry: CallStackEntry = {
+        callId: uuidv7(),
+        traceId: this.predMeta.predictAndScoreEntry.traceId,
+        childSummary: {},
+      };
 
-    // Store score in metadata
-    this.predMeta.scores[scorerName] = score;
+      await createAndFinishCall(
+        client,
+        scorerCall,
+        scorerScoreFactory(scorerName),
+        [{self: scorer, output: this.predMeta.output}],
+        'useParam0Object',
+        scorer,
+        scorerEntry,
+        this.predMeta.predictAndScoreEntry,
+        new Date(),
+        score,
+        `${scorerName}.score`,
+        IMPERATIVE_SCORE_MARKER
+      );
+
+      // Attach the score as feedback to the predict call
+      await client.addScore(
+        this.predMeta.predictCallId,
+        scorerEntry.callId,
+        scorerRefUri,
+        score
+      );
+
+      // Store score in metadata
+      this.predMeta.scores[scorerName] = score;
+    })();
+    this._scoringProcesses.push(asyncProcess);
+    await asyncProcess;
   }
 
   /**
-   * Finish this prediction.
+   * Finish the scoring process for the prediction.
    * Finalizes the predict_and_score call with accumulated scores.
    * Note: The predict call is finished by EvaluationLogger, not here.
    */
@@ -397,11 +395,9 @@ export class ScoreLogger {
       return; // Already finished
     }
 
-    const client = getGlobalClient();
-    if (!client) {
-      console.warn('Weave not initialized, skipping prediction finish');
-      return;
-    }
+    const client = requireGlobalClient();
+
+    await Promise.all(this._scoringProcesses);
 
     const endTime = new Date();
 
@@ -516,11 +512,7 @@ export class EvaluationLogger {
    * Start the evaluate call (root of the call hierarchy).
    */
   private async startEvaluate(): Promise<void> {
-    const client = getGlobalClient();
-    if (!client) {
-      console.warn('Weave not initialized, skipping evaluate call creation');
-      return;
-    }
+    const client = requireGlobalClient();
 
     // Create evaluate call
     const evaluateCall = new InternalCall();
@@ -534,7 +526,7 @@ export class EvaluationLogger {
     const evaluateStartPromise = client.createCall(
       evaluateCall,
       evaluationEvaluate,
-      [{this: this.evaluation, model: this.model}],
+      [{self: this.evaluation, model: this.model}],
       'useParam0Object',
       undefined,
       evaluateEntry,
@@ -560,7 +552,7 @@ export class EvaluationLogger {
     output: any
   ): Promise<ScoreLogger> {
     await this.initPromise;
-    const client = getGlobalClient();
+    const client = requireGlobalClient();
     if (!client || !this.evaluateEntry) {
       console.warn('Weave not initialized or evaluate not started');
       // Return a no-op ScoreLogger
@@ -588,7 +580,7 @@ export class EvaluationLogger {
     const predictAndScoreStartPromise = client.createCall(
       predictAndScoreCall,
       evaluationPredictAndScore,
-      [{this: this.evaluation, model: this.model, example: inputs}],
+      [{self: this.evaluation, model: this.model, example: inputs}],
       'useParam0Object',
       undefined,
       predictAndScoreEntry,
@@ -644,7 +636,7 @@ export class EvaluationLogger {
    */
   async logSummary(summary?: Record<string, any>): Promise<void> {
     await this.initPromise;
-    const client = getGlobalClient();
+    const client = requireGlobalClient();
     if (!client || !this.evaluateEntry) {
       console.warn('Weave not initialized or evaluate not started');
       return;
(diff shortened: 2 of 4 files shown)


chance-wnb:
* **#5542** <a href="https://app.graphite.dev/github/pr/wandb/weave/5542?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#5543](https://github.com/wandb/weave/pull/5543) <a href="https://app.graphite.dev/github/pr/wandb/weave/5543?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#5555](https://github.com/wandb/weave/pull/5555) <a href="https://app.graphite.dev/github/pr/wandb/weave/5555?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/5542?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/weaveObject.ts
```diff
@@ -50,10 +50,6 @@ export class WeaveObject {
 
   constructor(protected _baseParameters: WeaveObjectParameters) {}
 
-  className() {
```
chance-wnb, line 53: No one calls.

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+      constructor(parameters: ScorerParameters) {
+        super({
+          ...parameters,
+          name: scorerName,
+          scorerName: scorerName,
+        });
+      }
+    }
+
+    // // Override the constructor name to be the scorer name
```
andrewtruong: nit: double comment

sdks/node/src/weaveClient.ts
```diff
@@ -644,9 +644,9 @@ export class WeaveClient {
   ) {
     let inputs: Record<string, any> = {};
 
-    // Add 'self' first if thisArg is a WeaveObject
-    if (thisArg instanceof WeaveObject) {
-      inputs['self'] = thisArg;
+    // Add 'this' first if thisArg is a WeaveObject
+    if (thisArg instanceof WeaveObject && !('this' in inputs)) {
+      inputs['this'] = thisArg;
```
andrewtruong: Does this work?  Does the UI use the `self` key for anything?

sdks/node/src/weaveClient.ts
```diff
@@ -706,7 +706,8 @@ export class WeaveClient {
     currentCall: CallStackEntry,
     parentCall: CallStackEntry | undefined,
     startTime: Date,
-    displayName?: string
+    displayName?: string,
+    customAttributes?: Record<string, any>
```
andrewtruong: nit: This should be called `evalAttributes` for consistency

sdks/node/src/weaveClient.ts
```diff
@@ -821,6 +827,54 @@ export class WeaveClient {
...
+    predictCallId: string,
+    scorerCallId: string,
+    runnableRefUri: string,
+    scorerOutput: any
+  ): Promise<string> {
+    // Parse entity and project from projectId (format: "entity/project")
+    const [entity, project] = this.projectId.split('/');
+
+    // Build call URIs in weave:/// format
+    const predictCallUri = `weave:///${entity}/${project}/call/${predictCallId}`;
```
andrewtruong: maybe this should be factored into a helper

sdks/node/src/weaveClient.ts
```diff
@@ -821,6 +827,54 @@ export class WeaveClient {
...
+  /**
+   * Add feedback (e.g., scorer output) to a call.
+   * Used in imperative evaluation to attach scorer results to predict calls.
+   *
+   * @param predictCallId - ID of the predict call to attach feedback to
+   * @param scorerCallId - ID of the scorer call that generated the feedback
+   * @param runnableRefUri - URI of the scorer (Op or Object ref)
+   * @param scorerOutput - Output of the scorer
+   */
+  public async addFeedback(
```
andrewtruong: I would rename this to `addScore` as we already have a concept called `Feedback`, and this is specifically for scoring eval predictions

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+    parameterNames,
+    thisArg,
+    currentEntry,
+    parentEntry,
+    startTime,
+    displayName,
+    attributes
+  );
+
+  await startPromise;
```
andrewtruong: Do you have to block here?

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+/**
+ * Evaluation WeaveObject for imperative evaluation logging.
+ * Internal to evaluationLogger - not exported from main SDK.
+ *
+ * Named "Evaluation" so that constructor.name serializes to 'Evaluation'.
+ */
+interface EvaluationParameters extends WeaveObjectParameters {
+  dataset?: Dataset<any> | string;
+  scorers?: string[];
+  [key: string]: any; // Allow arbitrary attributes (trial_id, run_id, etc.)
```
andrewtruong, line 109: Why not explicitly define the params?

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+// ============================================================================
+
+/**
+ * Evaluation WeaveObject for imperative evaluation logging.
+ * Internal to evaluationLogger - not exported from main SDK.
+ *
+ * Named "Evaluation" so that constructor.name serializes to 'Evaluation'.
+ */
+interface EvaluationParameters extends WeaveObjectParameters {
+  dataset?: Dataset<any> | string;
```
andrewtruong, line 107: OK for MVP, but you might want to parametrize the types similar to `Evaluation`

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+  scorers?: string[];
+  [key: string]: any; // Allow arbitrary attributes (trial_id, run_id, etc.)
+}
+
+class Evaluation extends WeaveObject {
```
andrewtruong, line 112: I think we already have an `Evaluation` inside `evaluation.ts`, why redefine it?

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+   *
+   * @param scorerName - Name of the scorer (e.g., "accuracy", "f1_score")
+   * @param score - The score value
+   */
+  async logScore(scorerName: string, score: any): Promise<void> {
+    if (this.predMeta.isFinished) {
+      throw new Error('Cannot log score after prediction is finished');
+    }
+
+    const client = getGlobalClient();
```
andrewtruong: we have `requireGlobalClient` for this

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+      scorerRefUri,
+      score
+    );
+
+    // Store score in metadata
+    this.predMeta.scores[scorerName] = score;
+  }
+
+  /**
+   * Finish this prediction.
```
andrewtruong: You mean scorer?

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+   * Finalizes the predict_and_score call with accumulated scores.
+   * Note: The predict call is finished by EvaluationLogger, not here.
+   */
+  async finish(): Promise<void> {
+    if (this.predMeta.isFinished) {
+      return; // Already finished
+    }
+
+    const client = getGlobalClient();
```
andrewtruong: use `requireGlobalClient`

sdks/node/src/evaluationLogger.ts
```diff
@@ -0,0 +1,749 @@
...
+      undefined,
+      evaluateEntry,
+      undefined, // No parent (root call)
+      startTime,
+      undefined, // displayName
+      {...IMPERATIVE_EVAL_MARKER, ...this.evalAttributes} // Merge eval attributes
+    );
+    await evaluateStartPromise;
+
+    this.evaluateCall = evaluateCall;
```
andrewtruong, line 540: What is the difference between this and `evaluateStartPromise`?

sdks/node/src/__tests__/evaluationLogger.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    const evalLogger = new EvaluationLogger({name: 'test-eval'});
+
+    const scoreLogger = await evalLogger.logPrediction(
+      {input: 'test'},
+      'output'
+    );
+
+    await scoreLogger.logScore('accuracy', 0.95);
+    await scoreLogger.logScore('f1', 0.88);
+    await scoreLogger.finish();
```
andrewtruong: Should the user need to await all of these?  I think it makes more sense that you just need to wait for the group, not individually

sdks/node/src/weaveClient.ts
chance-wnb: I haven't seen anything broken UI-wise yet.

sdks/node/src/weaveClient.ts
chance-wnb: Hmm, I see what you mean. 

`createCall()` theoretically is capable of serving consumers other than the eval scenarios. So I think having a natural name here is fine. Since this is nit, I will keep it as is.

sdks/node/src/evaluationLogger.ts
chance-wnb: User has the freedom to specify any parameter as they desire. The parameters will be displayed as metadata in the UI. User will eventually decide what metadata they want to see. So I guess we don't need to dictate anything here.

sdks/node/src/evaluationLogger.ts
chance-wnb: Maybe I don't follow, but I think the parameters are flexible parameters. Can you show a code example on what you expect it to be?

sdks/node/src/weaveClient.ts
andrewtruong: Oh sorry, I didn't see that this was the generic `createCall` -- in that case I think this should just be called `attributes`

sdks/node/src/evaluationLogger.ts
chance-wnb: Great question. 

The new classes are for serialization purpose only, I don't want to pull in any other behaviors of a class. According to the logic of [`saveObject()`](https://github.com/wandb/weave/blob/e75e8fcff4f211330afce3a7ecdf6781d512d91b/sdks/node/src/weaveClient.ts#L422) it only looks at the class name and its enumerable properties. For cleanness, I want to have a simple class that is away from any potential intricacies that might come from the classic Evaluation class. 

These object are automatically serialized when they appear in the input of a call. Having this natural serialization process is ideal, that's why dummy classes are created.

sdks/node/src/evaluationLogger.ts
chance-wnb: The evaluateCall (InternalCall) is a wrapper of CallSchema, which aims to add some behaviors. 
`evaluateStartPromise` is just a signal to gauge the timing on `createCall()`.

Since `evaluateStartPromise` does not live inside evaluateCall, so I need to keep track of them separately.

sdks/node/src/__tests__/evaluationLogger.test.ts
chance-wnb: That's a great insight! No they shouldn't need to wait on these. We should strive for fire-n-forget. 

It does make things a bit trickier: since we need to ensure the scorings are all done before finishing the `predict_and_call`. Added a promise capturing for that. 

This part has been redone, please kindly help review again! Thanks!

sdks/node/src/evaluationLogger.ts
chance-wnb: yeah, reworded a bit for clarity.

sdks/node/src/evaluationLogger.ts
chance-wnb: Nope, removed. Thanks!

sdks/node/src/weaveClient.ts
chance-wnb: Indeed. renamed as such.

sdks/node/src/weaveClient.ts
chance-wnb: I added a CallRef class for that, please let me know if the shape is good.

sdks/node/src/weaveClient.ts
chance-wnb: Yep, renamed to `attributes` for now.

sdks/node/src/weaveClient.ts
chance-wnb: Sorry Andrew, somehow after the new iteration I start to see something broken. I have to have this PR to fix (core side): https://github.com/wandb/core/pull/35201

Do you feel this is risky? I so, I am happy to revert back to using `self`.

andrewtruong: I think the client would break on older servers.  The safest option would be to use `self`, then switch back to `this` next month (to align with the next server release cycle`)

chance-wnb: Indeed, I have reverted back to use self.

sdks/node/src/evaluationLogger.ts
andrewtruong: For example, the regular `Evaluation` is parametrized with `Evaluation<R, E, M>`

Since this is just metadata I don't think this is a huge issue.

chance-wnb: Ah, I see what you mean. I envision these part to be written after integrating your Eval V2 API. I will address these then.

sdks/node/src/__tests__/evaluationLogger.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    scoreLogger1.logScore('passed', true);
+    await scoreLogger1.finish();
+
+    const scoreLogger2 = await evalLogger.logPrediction({input: '2'}, 'out2');
+    scoreLogger2.logScore('passed', false);
+    await scoreLogger2.finish();
+
+    const scoreLogger3 = await evalLogger.logPrediction({input: '3'}, 'out3');
+    scoreLogger3.logScore('passed', true);
+    await scoreLogger3.finish();
```
andrewtruong, line 205: I don't think the user should need to call await here, except maybe at the very end.

sdks/node/src/__tests__/evaluationLogger.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    const evalLogger = new EvaluationLogger({name: 'test-eval'});
+
+    for (let i = 0; i < 3; i++) {
+      const scoreLogger = await evalLogger.logPrediction(
+        {input: `test${i}`},
+        `output${i}`
+      );
+      scoreLogger.logScore('accuracy', 0.9 + i * 0.05);
+      await scoreLogger.finish();
+    }
```
andrewtruong, line 181: Why do you need to block here?

sdks/node/src/__tests__/evaluationLogger.test.ts
```diff
@@ -0,0 +1,303 @@
...
+  beforeEach(() => {
+    traceServer = new InMemoryTraceServer();
+    initWithCustomTraceServer(projectId, traceServer);
+  });
+
+  // Test: Basic flow - logPrediction → logScore → finish → logSummary
+  test('complete evaluation lifecycle', async () => {
+    const evalLogger = new EvaluationLogger({name: 'test-eval'});
+
+    const scoreLogger = await evalLogger.logPrediction(
```
andrewtruong, line 35: do you need to await here?

andrewtruong:
I think this is good for an MVP.  There are a few places where I think we don't need to be awaiting anything, but we can fix that in a future iteration.

PR #5545, opened 2025-10-09
fix(weave): Allow markdown to be stored in files

tssweeney:
Currently our custom type serialization layer suffers from a fractured design between "inline" and "file" based payloads. This implementation has a number of problems:
1. "inline" objects typically has referred to when we put data "inline" as opposed to in saved objects - this use of "inline" is a different concept
2. There is no reason that a serializer can't use both files and values (in fact, we want to do in many cases where there is a bigger file and a smaller metadata payload).

This Pr implements a simplified pattern that combines these approaches, allowing serializers to choose if they want to use files, metadata, or both. This simplifies the code path and supports backwards compatibility.

Further, it makes Markdown use this new approach. Relies on https://github.com/wandb/core/pull/35135 being shipped first to support this new layout.

### DIFF
--- a/weave/trace/serialization/custom_objs.py
+++ b/weave/trace/serialization/custom_objs.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from collections.abc import Mapping
 from typing import Any, Literal, TypedDict
 
@@ -22,6 +23,8 @@ from weave.trace_server.trace_server_interface import (
     FileContentReadReq,
     TraceServerInterface,
 )
+
+logger = logging.getLogger(__name__)
 
 
 class WeaveTypeDict(TypedDict):
@@ -135,6 +138,7 @@ def maybe_attach_art(res: Any, art: MemTraceFilesArtifact) -> None:
     try:
         res.art = art
     except Exception:
+        logger.debug("Failed to attach artifact to result", exc_info=True)
         pass
 
 


weave/trace/serialization/custom_objs.py
```diff
@@ -56,80 +61,81 @@ def encode_custom_obj(obj: Any) -> dict | None:
...
         "load_op": load_op_uri,
     }
 
-    # If the save method just takes one argument, it is an inline serializer
-    if is_inline_save(serializer.save):
-        encoded["val"] = serializer.save(obj)
-    elif is_file_save(serializer.save):
-        art = MemTraceFilesArtifact()
-        serializer.save(obj, art, "obj")
+    art = MemTraceFilesArtifact()
```
neutralino1, line 106:
```suggestion
    artifact = MemTraceFilesArtifact()
```

weave/trace/serialization/custom_objs.py
```diff
@@ -56,80 +61,81 @@ def encode_custom_obj(obj: Any) -> dict | None:
...
-        raise ValueError(
-            f"Failed to load op needed to decode object of type `{type_}`. See logs above for more information."
-        )
+def decode_custom_obj(encoded: dict) -> Any:
+    return _decode_custom_obj(
+        encoded["weave_type"],
+        encoded.get("files", {}),
+        encoded.get("val"),
+        encoded.get("load_op"),
+    )
```
neutralino1: Can we merge `decode_custom_obj` and `_decode_custom_obj`?

zbirenbaum:
There is a scenario in which this causes a breaking change for a large number of users that I reproduced. It seems like you attempted to handle similar scenarios with the legacy stuff but it didn't work since KNOWN_OBJECTS go through a different code path.

1. User A publishes object inside a trace using SDK version with this patch
2. User B running previous SDK version attempts to retrieve object
3. All KNOWN_TYPE functions have signature changes so the sdk no longer can properly retrieve the object
4. User B runs get_calls and errors when they hit that trace 

@tssweeney I think this change is OK but we need to do another patch before merging or put up a docs page so that that we can notify users why this error is appearing. 

We ideally should keep the old serialize and deserialize versions, print a deprecation when they are used, and then have a try, except block which falls back to them in order to avoid the error.

tssweeney:
> There is a scenario in which this causes a breaking change for a large number of users that I reproduced. It seems like you attempted to handle similar scenarios with the legacy stuff but it didn't work since KNOWN_OBJECTS go through a different code path.
> 
> 1. User A publishes object inside a trace using SDK version with this patch
> 2. User B running previous SDK version attempts to retrieve object
> 3. All KNOWN_TYPE functions have signature changes so the sdk no longer can properly retrieve the object
> 4. User B runs get_calls and errors when they hit that trace
> 
> @tssweeney I think this change is OK but we need to do another patch before merging or put up a docs page so that that we can notify users why this error is appearing.
> 
> We ideally should keep the old serialize and deserialize versions, print a deprecation when they are used, and then have a try, except block which falls back to them in order to avoid the error.

Hey Zach - thanks for the feedback. A few questions/notes:
1. Can you share what repros you are using?
2. I think the case of: old SDK consuming new SDK objects is certainly something we want to work, but if a breaking change must be made, this is the acceptable break as the fix is to update
3. There should not need to be a docs page or error we should always be able to handle the error or recommend an upgrade. all of this stuff is implementation details and a user should never need to know how we specifically serialize stuff

weave/trace/serialization/custom_objs.py
```diff
@@ -56,80 +82,95 @@ def encode_custom_obj(obj: Any) -> dict | None:
...
+
+def maybe_attach_art(res: Any, art: MemTraceFilesArtifact) -> None:
+    """Attach the artifact to the result if possible.
+    This is a best-effort approach to attach the artifact to the result.
+
+    I am pretty sure this is just used in a legacy code capture case
+    """
+    try:
+        res.art = art
+    except Exception:
```
zbirenbaum, line 140: Could we log the error here when one occurs? This line actually caused a massive headache while working on content and was quite a pain to track down.

PR #5546, opened 2025-10-09
feat(weave): Add new CRUD endpoints for interacting with Evaluations

andrewtruong:
https://wandb.atlassian.net/browse/WB-28281

This PR adds CRUD endpoints for interacting with Evaluations as part of the V2 Evals API.

Notably, this PR does not implement endpoints for interacting with the Evaluation's calls.  That feature will come later!

Pairs with https://github.com/wandb/core/pull/35136

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -1763,6 +1763,178 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         result = self.obj_delete(obj_delete_req)
         return tsi.ScorerDeleteV2Res(num_deleted=result.num_deleted)
 
+    def evaluation_create_v2(
+        self, req: tsi.EvaluationCreateV2Req
+    ) -> tsi.EvaluationCreateV2Res:
+        """Create an evaluation object.
+
+        Creates placeholder ops for evaluate, predict_and_score, and summarize methods.
+        """
+        # Create a safe ID for the evaluation
+        if req.name is None:
+            evaluation_id = "Evaluation"
+        else:
+            evaluation_id = object_creation_utils.make_safe_name(req.name)
+
+        # Create placeholder evaluate op
+        evaluate_op_req = tsi.OpCreateV2Req(
+            project_id=req.project_id,
+            name=f"{evaluation_id}.evaluate",
+            source_code=object_creation_utils.PLACEHOLDER_EVALUATE_OP_SOURCE,
+        )
+        evaluate_op_res = self.op_create_v2(evaluate_op_req)
+        evaluate_ref = evaluate_op_res.digest
+
+        # Create placeholder predict_and_score op
+        predict_and_score_op_req = tsi.OpCreateV2Req(
+            project_id=req.project_id,
+            name=f"{evaluation_id}.predict_and_score",
+            source_code=object_creation_utils.PLACEHOLDER_PREDICT_AND_SCORE_OP_SOURCE,
+        )
+        predict_and_score_op_res = self.op_create_v2(predict_and_score_op_req)
+        predict_and_score_ref = predict_and_score_op_res.digest
+
+        # Create placeholder summarize op
+        summarize_op_req = tsi.OpCreateV2Req(
+            project_id=req.project_id,
+            name=f"{evaluation_id}.summarize",
+            source_code=object_creation_utils.PLACEHOLDER_EVALUATION_SUMMARIZE_OP_SOURCE,
+        )
+        summarize_op_res = self.op_create_v2(summarize_op_req)
+        summarize_ref = summarize_op_res.digest
+
+        # Create the evaluation object
+        evaluation_val = object_creation_utils.build_evaluation_val(
+            name=req.name,
+            dataset_ref=req.dataset,
+            trials=req.trials,
+            description=req.description,
+            scorer_refs=req.scorers,
+            evaluation_name=req.evaluation_name,
+            metadata=None,
+            preprocess_model_input=None,
+            eval_attributes=req.eval_attributes,
+            evaluate_ref=evaluate_ref,
+            predict_and_score_ref=predict_and_score_ref,
+            summarize_ref=summarize_ref,
+        )
+        obj_req = tsi.ObjCreateReq(
+            obj=tsi.ObjSchemaForInsert(
+                project_id=req.project_id,
+                object_id=evaluation_id,
+                val=evaluation_val,
+                wb_user_id=None,
+            )
+        )
+        obj_result = self.obj_create(obj_req)
+
+        # Query the object back to get its version index (this may not be
+        # immediately available, so we retry a few times)
+        obj_read_req = tsi.ObjReadReq(
+            project_id=req.project_id,
+            object_id=evaluation_id,
+            digest=obj_result.digest,
+        )
+        obj_read_res = self._obj_read_with_retry(obj_read_req)
+
+        # Get the ref and return the create result
+        evaluation_ref = ri.InternalObjectRef(
+            project_id=req.project_id,
+            name=evaluation_id,
+            version=obj_result.digest,
+        ).uri()
+        return tsi.EvaluationCreateV2Res(
+            digest=obj_result.digest,
+            object_id=evaluation_id,
+            version_index=obj_read_res.obj.version_index,
+            evaluation_ref=evaluation_ref,
+        )
+
+    def evaluation_read_v2(
+        self, req: tsi.EvaluationReadV2Req
+    ) -> tsi.EvaluationReadV2Res:
+        """Get an evaluation object by delegating to obj_read with retry logic."""
+        obj_req = tsi.ObjReadReq(
+            project_id=req.project_id,
+            object_id=req.object_id,
+            digest=req.digest,
+        )
+        result = self._obj_read_with_retry(obj_req)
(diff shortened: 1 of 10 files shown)


weave/trace_server/reference/generate.py
```diff
@@ -769,4 +770,69 @@ def scorer_delete_v2(
         )
         return service.trace_server_interface.scorer_delete_v2(req)
 
+    @router.post("/v2/{entity}/{project}/evaluation", tags=[V2_EVALUATIONS_TAG_NAME])
```
neutralino1:
```suggestion
    @router.post("/v2/{entity}/{project}/evaluations", tags=[V2_EVALUATIONS_TAG_NAME])
```

In general, resource names are pluralized in RESTful APIs.

weave/trace_server/reference/generate.py
```diff
@@ -769,4 +770,69 @@ def scorer_delete_v2(
...
+    def evaluation_create_v2(
+        req: tsi.EvaluationCreateV2Req,
+        service: weave.trace_server.trace_service.TraceService = Depends(get_service),  # noqa: B008
+    ) -> tsi.EvaluationCreateV2Res:
+        """Create an evaluation object."""
+        return service.trace_server_interface.evaluation_create_v2(req)
+
+    @router.get(
+        "/v2/{entity}/{project}/evaluation/{object_id}/{digest}",
```
neutralino1:
```suggestion
        "/v2/{entity}/{project}/evaluations/{object_id}/{digest}",
```

weave/trace_server/reference/generate.py
```diff
@@ -769,4 +770,69 @@ def scorer_delete_v2(
...
+        service: weave.trace_server.trace_service.TraceService = Depends(get_service),  # noqa: B008
+    ) -> tsi.EvaluationReadV2Res:
+        """Get an evaluation object."""
+        req = tsi.EvaluationReadV2Req(
+            project_id=project_id, object_id=object_id, digest=digest
+        )
+        return service.trace_server_interface.evaluation_read_v2(req)
+
+    @router.get(
+        "/v2/{entity}/{project}/evaluation",
```
neutralino1:
```suggestion
        "/v2/{entity}/{project}/evaluations",
```

weave/trace_server/reference/generate.py
neutralino1: Also, I wonder if this endpoint should be
```
/v2/{entity}/{project}/evaluations/{object_id}/versions/{digest}
```
This is more semantically correct.

Then we can have 
```
/v2/{entity}/{project}/evaluations/{object_id}
```
to return all versions.

weave/trace_server/reference/generate.py
```diff
@@ -769,4 +770,69 @@ def scorer_delete_v2(
...
+        service: weave.trace_server.trace_service.TraceService = Depends(get_service),  # noqa: B008
+    ) -> StreamingResponse:
+        """List evaluation objects."""
+        req = tsi.EvaluationListV2Req(project_id=project_id, limit=limit, offset=offset)
+        return StreamingResponse(
+            service.trace_server_interface.evaluation_list_v2(req),
+            media_type="application/jsonl",
+        )
+
+    @router.post(
```
neutralino1:
```suggestion
    @router.delete(
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1801,6 +1801,169 @@ def scorer_delete_v2(self, req: tsi.ScorerDeleteV2Req) -> tsi.ScorerDeleteV2Res:
...
+    def evaluation_create_v2(
+        self, req: tsi.EvaluationCreateV2Req
+    ) -> tsi.EvaluationCreateV2Res:
+        """Create an evaluation object.
+
+        Creates placeholder ops for evaluate, predict_and_score, and summarize methods.
+        """
+        # Create a safe ID for the evaluation
+        safe_name = object_creation_utils.make_safe_name(req.name)
+        evaluation_id = f"evaluation_{safe_name}"
```
neutralino1: Is this the current standard? I don't remember seeing this elsewhere.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1801,6 +1801,169 @@ def scorer_delete_v2(self, req: tsi.ScorerDeleteV2Req) -> tsi.ScorerDeleteV2Res:
...
+
+        # Create placeholder evaluate op
+        evaluate_op_req = tsi.OpCreateV2Req(
+            project_id=req.project_id,
+            name=f"{evaluation_id}_evaluate",
```
neutralino1: Isn't it supposed to be `MyEvaluation.evaluate`?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1801,6 +1801,169 @@ def scorer_delete_v2(self, req: tsi.ScorerDeleteV2Req) -> tsi.ScorerDeleteV2Res:
...
+            source_code=object_creation_utils.get_placeholder_evaluate_source(),
```
neutralino1: Also generally, why does this op need to be a placeholder? Shouldn't the Op creation endpoint have been called prior to this one, and the request payload should contain an Op ID (name+digest)?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1801,6 +1801,169 @@ def scorer_delete_v2(self, req: tsi.ScorerDeleteV2Req) -> tsi.ScorerDeleteV2Res:
...
+        predict_and_score_ref = predict_and_score_op_res.digest
+
+        # Create placeholder summarize op
+        summarize_op_req = tsi.OpCreateV2Req(
+            project_id=req.project_id,
+            name=f"{evaluation_id}_summarize",
+            source_code=object_creation_utils.get_placeholder_summarize_source(),
+        )
+        summarize_op_res = self.op_create_v2(summarize_op_req)
+        summarize_ref = summarize_op_res.digest
```
neutralino1, line 1804: All of these should be created ahead of time using the `ops` endpoint. Otherwise, when are they going to be set to their actual values?

Or maybe you are being proactive, since the eventual data model will no contain those op refs?

The problem is that this will yield some dummy ops in the UI, won't it?

weave/trace_server/object_creation_utils.py
```diff
@@ -81,6 +81,51 @@ def get_default_summarize_op_source() -> str:
 """
 
 
+def get_placeholder_evaluate_source() -> str:
+    """Get placeholder source code for the evaluate method.
+
+    Returns:
+        Python source code for a placeholder evaluate function
+    """
+    return """import weave
```
neutralino1:
```suggestion
PLACEHOLDER_EVALUATION_SOURCE = """import weave
```

neutralino1: Same below.

weave/trace_server/trace_server_interface.py
```diff
@@ -1509,6 +1509,115 @@ class ScorerDeleteV2Res(BaseModel):
...
+    )
+    description: Optional[str] = Field(
+        None,
+        description="A description of this evaluation",
+    )
+
+    dataset_ref: str = Field(..., description="Reference to the dataset (weave:// URI)")
+    scorer_refs: Optional[list[str]] = Field(
+        None, description="List of scorer references (weave:// URIs)"
+    )
```
neutralino1: Here we could expect an optional 
```python
    op_refs: Optional[dict["evaluate" | "summarize" | "predict_and_score", str]]
```

It could be optional so that when we switch to the new data model, we don't expect these anymore from the SDK.

neutralino1 (commented):
Nice work. A few notes.

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: fixed

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: fixed

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: These are created ahead of time, but using the internal op APIs (since this is the server). 

Eventually I think we can move to a model without ops, but I'm doing this to make sure the v2 API internally generates the same data as the v1 API.

weave/trace_server/object_creation_utils.py
andrewtruong: fixed

weave/trace_server/reference/generate.py
```diff
@@ -363,6 +364,85 @@ def scorer_delete_v2(
...
+        """List evaluation objects."""
+        project_id = f"{entity}/{project}"
+        req = tsi.EvaluationListV2Req(project_id=project_id, limit=limit, offset=offset)
+        return StreamingResponse(
+            service.trace_server_interface.evaluation_list_v2(req),
+            media_type="application/jsonl",
+        )
+
+    @router.delete(
+        "{entity}/{project}/evaluation/{object_id}", tags=[V2_EVALUATIONS_TAG_NAME]
```
neutralino1:
```suggestion
        "{entity}/{project}/evaluations/{object_id}", tags=[V2_EVALUATIONS_TAG_NAME]
```

PR #5547, opened 2025-10-09
feat(weave): Add new CRUD endpoints for interacting with EvaluationRuns, Predictions, Scores

andrewtruong:
https://wandb.atlassian.net/browse/WB-28282
https://wandb.atlassian.net/browse/DOCS-1873

This PR adds CRUD endpoints for interacting with EvaluationRuns, Predictions, and Scores as part of the V2 Evals API.

For this POC, the implementation simply delegates to the existing Calls endpoints.  There is some logic to generate the same payloads as the client would normally create.

Pairs with https://github.com/wandb/core/pull/35137

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -2931,6 +2931,19 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         )
 
         # Start a call to represent the score
+        score_attributes = {
+            constants.WEAVE_ATTRIBUTES_NAMESPACE: {
+                constants.SCORE_ATTR_KEY: "true",
+                constants.SCORE_PREDICTION_ID_ATTR_KEY: req.prediction_id,
+                constants.SCORE_SCORER_ATTR_KEY: req.scorer,
+            }
+        }
+        # Store evaluation_run_id as attribute if provided
+        if req.evaluation_run_id:
+            score_attributes[constants.WEAVE_ATTRIBUTES_NAMESPACE][
+                constants.SCORE_EVALUATION_RUN_ID_ATTR_KEY
+            ] = req.evaluation_run_id
+
         call_start_req = tsi.CallStartReq(
             start=tsi.StartedCallSchemaForInsert(
                 project_id=req.project_id,
@@ -2939,13 +2952,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
                 parent_id=parent_id,
                 op_name=score_op_ref.uri(),
                 started_at=datetime.datetime.now(datetime.timezone.utc),
-                attributes={
-                    constants.WEAVE_ATTRIBUTES_NAMESPACE: {
-                        constants.SCORE_ATTR_KEY: "true",
-                        constants.SCORE_PREDICTION_ID_ATTR_KEY: req.prediction_id,
-                        constants.SCORE_SCORER_ATTR_KEY: req.scorer,
-                    }
-                },
+                attributes=score_attributes,
                 inputs={
                     "self": req.scorer,
                     "inputs": prediction_inputs,
@@ -3022,9 +3029,10 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         # The output is stored directly as the numeric value
         value = call.output if call.output is not None else 0.0
 
-        # If the parent is a predict_and_score call, get the evaluation_run_id from its parent
-        evaluation_run_id = None
-        if call.parent_id:
+        # Get evaluation_run_id from attributes (preferred), fallback to parent traversal for backwards compatibility
+        evaluation_run_id = attributes.get(constants.SCORE_EVALUATION_RUN_ID_ATTR_KEY)
+        if evaluation_run_id is None and call.parent_id:
+            # Fallback: If the parent is a predict_and_score call, get the evaluation_run_id from its parent
             parent_read_req = tsi.CallReadReq(
                 project_id=req.project_id,
                 id=call.parent_id,
@@ -3068,6 +3076,19 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
             )
         )
 
+        # Filter by evaluation_run_id if provided
+        if req.evaluation_run_id:
+            conditions.append(
+                tsi_query.EqOperation(
+                    eq_=[
+                        tsi_query.GetFieldOperator(
+                            get_field_=f"attributes.{constants.WEAVE_ATTRIBUTES_NAMESPACE}.{constants.SCORE_EVALUATION_RUN_ID_ATTR_KEY}"
+                        ),
+                        tsi_query.LiteralOperation(literal_=req.evaluation_run_id),
+                    ]
+                )
+            )
+
         # Combine all conditions with AND (or use single condition if only one)
         if len(conditions) == 1:
             query = tsi.Query(expr_=conditions[0])
@@ -3087,9 +3108,12 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
             attributes = call.attributes.get(constants.WEAVE_ATTRIBUTES_NAMESPACE, {})
             value = call.output if call.output is not None else 0.0
 
-            # If the parent is a predict_and_score call, get the evaluation_run_id from its parent
-            evaluation_run_id = None
-            if call.parent_id:
+            # Get evaluation_run_id from attributes (preferred), fallback to parent traversal for backwards compatibility
+            evaluation_run_id = attributes.get(
+                constants.SCORE_EVALUATION_RUN_ID_ATTR_KEY
+            )
+            if evaluation_run_id is None and call.parent_id:
+                # Fallback: If the parent is a predict_and_score call, get the evaluation_run_id from its parent
                 parent_read_req = tsi.CallReadReq(
                     project_id=req.project_id,
                     id=call.parent_id,
@@ -3100,10 +3124,6 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
                     constants.EVALUATION_RUN_PREDICTION_AND_SCORE_OP_NAME,
                 ):
                     evaluation_run_id = parent_res.call.parent_id
-
-            # Skip if filtering by evaluation_run_id and this doesn't match
-            if req.evaluation_run_id and evaluation_run_id != req.evaluation_run_id:
-                continue
 
             yield tsi.ScoreReadV2Res(
                 score_id=call.id,
--- a/weave/trace_server/constants.py
+++ b/weave/trace_server/constants.py
@@ -32,6 +32,7 @@ PREDICTION_EVALUATION_RUN_ID_ATTR_KEY = "evaluation_run_id"
 SCORE_ATTR_KEY = "score"
 SCORE_PREDICTION_ID_ATTR_KEY = "prediction_id"
 SCORE_SCORER_ATTR_KEY = "scorer"
+SCORE_EVALUATION_RUN_ID_ATTR_KEY = "evaluation_run_id"
 
 # Weave attributes namespace
 WEAVE_ATTRIBUTES_NAMESPACE = "weave"
--- a/weave/trace_server/sqlite_trace_server.py
+++ b/weave/trace_server/sqlite_trace_server.py
@@ -3226,6 +3226,19 @@ class SqliteTraceServer(tsi.FullTraceServerInterface):
         )
 
         # Start a call to represent the score
+        score_attributes = {
+            constants.WEAVE_ATTRIBUTES_NAMESPACE: {
+                constants.SCORE_ATTR_KEY: "true",
+                constants.SCORE_PREDICTION_ID_ATTR_KEY: req.prediction_id,
+                constants.SCORE_SCORER_ATTR_KEY: req.scorer,
+            }
+        }
+        # Store evaluation_run_id as attribute if provided
+        if req.evaluation_run_id:
+            score_attributes[constants.WEAVE_ATTRIBUTES_NAMESPACE][
+                constants.SCORE_EVALUATION_RUN_ID_ATTR_KEY
+            ] = req.evaluation_run_id
+
         call_start_req = tsi.CallStartReq(
             start=tsi.StartedCallSchemaForInsert(
                 project_id=req.project_id,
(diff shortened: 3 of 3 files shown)


weave/trace_server/reference/generate.py
```diff
@@ -443,6 +444,175 @@ def evaluation_delete_v2(
...
+                },
+            }
+        },
+    )
+    def evaluation_run_list_v2(
+        entity: str,
+        project: str,
+        limit: int | None = None,
+        offset: int | None = None,
+        evaluation_refs: str | None = None,
```
chance-wnb: ~why don't you design evaluation_refs and model_refs as lists I wonder?~

Ok. maybe these fields should go to body or query strings?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1935,6 +1935,483 @@ def evaluation_delete_v2(
...
+            finished_at=call.ended_at,
+            summary=call.summary,
+        )
+
+    def evaluation_run_list_v2(
+        self, req: tsi.EvaluationRunListV2Req
+    ) -> Iterator[tsi.EvaluationRunReadV2Res]:
+        """List evaluation runs by querying calls with evaluation_run attribute."""
+        # Build the filter
+        calls_filter = tsi.CallsFilter()
```
chance-wnb: This is a default filter right? Do we miss some actual filtering conditions?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1935,6 +1935,483 @@ def evaluation_delete_v2(
...
+        call_res = self.call_read(call_read_req)
+        evaluation_ref = ""
+        if call_res.call is not None:
+            attributes = call_res.call.attributes.get(
+                constants.WEAVE_ATTRIBUTES_NAMESPACE, {}
+            )
+            evaluation_ref = attributes.get(
+                constants.EVALUATION_RUN_EVALUATION_ATTR_KEY, ""
+            )
+
```
chance-wnb, line 2433: what happens if the evaluation.evalute() is None here? (not created)

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1935,6 +1935,483 @@ def evaluation_delete_v2(
...
+                        constants.EVALUATION_RUN_MODEL_ATTR_KEY: req.model,
+                    }
+                },
+                inputs=model_predict_inputs,
+            )
+        )
+        self.call_start(model_predict_start_req)
+
+        # End the Model.predict call with the output
+        model_predict_end_req = tsi.CallEndReq(
```
chance-wnb: I think we should look into an opportunity to combine start and end together into one CallParts record. I think that's the direction Griffin's work is going towards. Maybe you guys should sync?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1935,6 +1935,483 @@ def evaluation_delete_v2(
...
+
+        # Build predict_and_score output with structured data
+        predict_and_score_output = {
+            "output": req.output,
+            "scores": [],  # Will be populated by score calls
+            "model_latency": 0,  # Placeholder for now
+        }
+
+        # End the predict_and_score call with structured output
+        predict_and_score_end_req = tsi.CallEndReq(
```
chance-wnb: Are we ending this predict_and_score call too early? the scoring calls have not yet happened.

I am curious on how you are going to deal with calculating `model_latency` and updating `predict_and_score_output.scores` AFAIK, my recent work on the FE depends on these info. 

Maybe you need a separate predict_and_score finish API.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1935,6 +1935,483 @@ def evaluation_delete_v2(
...
+                        constants.EVALUATION_RUN_SCORER_ATTR_KEY: req.scorer,
+                    }
+                },
+                inputs=score_inputs,
+            )
+        )
+        self.call_start(call_start_req)
+
+        # End the call immediately with the score
+        call_end_req = tsi.CallEndReq(
```
chance-wnb, line 2967: Same thing, if we could have an opportunity to combine start and end, we should seize such an opportunity to create a single combined record. 

We can deal with it in a follow up if you prefer.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1935,6 +1935,483 @@ def evaluation_delete_v2(
...
+                inputs=summarize_inputs,
+            )
+        )
+        self.call_start(summarize_start_req)
+
+        # The summarize output is the summary data
+        summarize_output = req.summary or {}
+
+        # End the Evaluation.summarize call with the summary
+        summarize_end_req = tsi.CallEndReq(
```
chance-wnb: Same thing, if we could have an opportunity to combine start and end, we should seize such an opportunity to create a single combined record.

We can deal with it in a follow up if you prefer.

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: What are you thinking about?  cc: @gtarpenning

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: I think it's a relic of the tracing structure.  Ideally the user (or the client) just supplies their own latencies and we can calculate it in the backend for them

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: Do you mean the `Evaluation` object?  It should already exist, because it's required for the `EvaluationRun` to exist

weave/trace_server/clickhouse_trace_server_batched.py
chance-wnb: however the scorings haven't yet occurred. if we conclude the call, when can we backfill this data of  `predict_and_score_output.scores`?

If you have a separate predict_and_score finish API, we could end the call then, then the latency is naturally captured.

weave/trace_server/clickhouse_trace_server_batched.py
chance-wnb: Sorry I wasn't clear. 

On line 2471 (evaluation_run_log_score_v2) you have:
```
   if predict_call_res.call is None:
            raise ValueError(f"Predict call {req.predict_call_id} not found")
```

to guard against the case that the upstream call is missing.

Maybe you want to have something similar here as well.

```
        if call_res.call is not None:
              raise  ...
```

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: Hmm good point.  I'm not certain we want a "finish" API, but maybe it is required in this intermediate state...

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2123,6 +2124,860 @@ def model_delete_v2(self, req: tsi.ModelDeleteV2Req) -> tsi.ModelDeleteV2Res:
...
+        evaluation_run_call = self.call_read(evaluation_run_read_req).call
+        evaluation_ref = None
+        if evaluation_run_call and evaluation_run_call.inputs:
+            evaluation_ref = evaluation_run_call.inputs.get("self")
+
+        # Query all predict_and_score children to compute means
+        # (Do this first so we can use the same data for both summarize and evaluation_run)
+        calls_query_req = tsi.CallsQueryReq(
+            project_id=req.project_id,
+            filter=tsi.CallsFilter(
```
chance-wnb, line 2326: nit: Maybe you want to add a `columns` filter, the subsequent logic seem to only care about `output`. Otherwise I fear we might run into memory explosion like we did on the frontend. Especially predict_and_score_calls might be too many.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2123,6 +2124,860 @@ def model_delete_v2(self, req: tsi.ModelDeleteV2Req) -> tsi.ModelDeleteV2Res:
...
+                parent_ids=[req.evaluation_run_id],
+                op_names=[constants.EVALUATION_RUN_PREDICTION_AND_SCORE_OP_NAME],
+            ),
+        )
+        predict_and_score_calls = self.calls_query(calls_query_req).calls
```
chance-wnb: You might want to use the `calls_query_stream()` version. `calls_query` serializes the stream.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2123,6 +2124,860 @@ def model_delete_v2(self, req: tsi.ModelDeleteV2Req) -> tsi.ModelDeleteV2Res:
...
+
+        # Query for calls that have the prediction attribute
+        calls_query_req = tsi.CallsQueryReq(
+            project_id=req.project_id,
+            query=query,
+            limit=req.limit,
+            offset=req.offset,
+        )
+
+        calls_res = self.calls_query(calls_query_req)
```
chance-wnb: calls_query_stream() might be better.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2123,6 +2124,860 @@ def model_delete_v2(self, req: tsi.ModelDeleteV2Req) -> tsi.ModelDeleteV2Res:
...
+                # Build the scores dict by querying all score children of predict_and_score
+                scores_dict = {}
+                calls_query_req = tsi.CallsQueryReq(
+                    project_id=req.project_id,
+                    filter=tsi.CallsFilter(
+                        parent_ids=[prediction_res.call.parent_id],
+                        op_names=[constants.SCORER_SCORE_OP_NAME],
+                    ),
+                )
+                score_calls = self.calls_query(calls_query_req).calls
```
chance-wnb: calls_query_stream

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: Fixed

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: Fixed

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: fixed

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: fixed

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: Let's do that in a future PR

weave/trace_server/interface/query.py
andrewtruong, line 1: This allows us to construct queries using the actual objects in addition to the "mongo-dict" style

tests/trace/test_trace_server_v2_api.py
andrewtruong, line 1: This file tests the basic CRUD operations

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: fixed

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2123,6 +2127,987 @@ def model_delete_v2(self, req: tsi.ModelDeleteV2Req) -> tsi.ModelDeleteV2Res:
...
+                    ),
+                    tsi_query.LiteralOperation(literal_="true"),
+                ]
+            )
+        )
+
+        # Combine all conditions with AND (or use single condition if only one)
+        if len(conditions) == 1:
+            query = tsi.Query(expr_=conditions[0])
+        else:
```
chance-wnb, line 2673: How could the conditions have a size other than 1 here?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2123,6 +2127,987 @@ def model_delete_v2(self, req: tsi.ModelDeleteV2Req) -> tsi.ModelDeleteV2Res:
...
+                )
+                parent_res = self.call_read(parent_read_req)
+                if parent_res.call and tsc.op_name_matches(
+                    parent_res.call.op_name,
+                    constants.EVALUATION_RUN_PREDICTION_AND_SCORE_OP_NAME,
+                ):
+                    evaluation_run_id = parent_res.call.parent_id
+
+            # Skip if filtering by evaluation_run_id and this doesn't match
+            if req.evaluation_run_id and evaluation_run_id != req.evaluation_run_id:
```
chance-wnb: I guess this query would be too expensive, it is going to walk through all the prediction calls of the project, read parent's parent for each, then filter by evaluation_run_id. 

Maybe you want to attach `evaluation_run_id` as one of the attributes of the predication calls. then you can directly filter.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2123,6 +2127,987 @@ def model_delete_v2(self, req: tsi.ModelDeleteV2Req) -> tsi.ModelDeleteV2Res:
...
```
chance-wnb: Same thing here: this query would be too expensive, it is going to walk through all the scoring calls of the project, read parent's parent for each, then filter by evaluation_run_id.

Maybe you want to attach evaluation_run_id as one of the attributes of the scoring calls. then you can directly filter.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2123,6 +2127,987 @@ def model_delete_v2(self, req: tsi.ModelDeleteV2Req) -> tsi.ModelDeleteV2Res:
...
+                    ),
+                    tsi_query.LiteralOperation(literal_="true"),
+                ]
+            )
+        )
+
+        # Combine all conditions with AND (or use single condition if only one)
+        if len(conditions) == 1:
+            query = tsi.Query(expr_=conditions[0])
+        else:
```
chance-wnb, line 3095: It seems to me the conditions always have size 1.

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: good catch, fixed

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: fixed

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: This will be more than 1 now that we use attr filtering :)

weave/trace_server/clickhouse_trace_server_batched.py
andrewtruong: same as above

andrewtruong:
Thanks @chance-wnb -- I addressed the items that made sense for this PR.

RE: Perf -- agree, but I think that can be done in a follow up PR.  Eventually we will want to move EvaluationRun and Prediction away from the Calls underlying model and into their own thing.  At that point we can interact with the DB directly!

PR #5557, opened 2025-10-10
fix(weave): bump size for autoconversion

zbirenbaum:
## Description

- Fixes WB-28266

Anything 1KB or under isn't worth the network latency to upload and actually can increase the storage size in DB to store the json blob to represent it.

### DIFF
--- a/weave/trace_server/base64_content_conversion.py
+++ b/weave/trace_server/base64_content_conversion.py
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 DATA_URI_PATTERN = re.compile(r"^data:([^;]+);base64,([A-Za-z0-9+/=]+)$", re.IGNORECASE)
 
 # Minimum size to create a file (to avoid making more data than what the original is)
-AUTO_CONVERSION_MIN_SIZE = 1024 # 1 KiB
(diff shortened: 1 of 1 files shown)


weave/trace_server/base64_content_conversion.py
```diff
@@ -23,13 +23,8 @@
 # Format: data:[content-type];base64,[base64_data]
 DATA_URI_PATTERN = re.compile(r"^data:([^;]+);base64,([A-Za-z0-9+/=]+)$", re.IGNORECASE)
 
-# Maximum size for base64 content to be processed (to avoid memory issues)
-MAX_BASE64_SIZE = 100 * 1024 * 1024  # 100 MiB
-
-# Minimum size for standalone base64 (to avoid false positives)
-MIN_BASE64_SIZE = 100  # 100 bytes
-
-MIN_TEXT_SIZE = 50 * 1024  # 50 KiB
```
neutralino1, line 32: Don't we need these constants anymore?

zbirenbaum: No, they were unused and got left over after removing raw base64 and text conversion. Data urls converted 100% of the time.

PR #5558, opened 2025-10-10
batch all file creates before flushing

gtarpenning:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

## Testing

How was this PR tested?

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -176,6 +176,8 @@ class ClickHouseTraceServer(tsi.TraceServerInterface):
         self._database = database
         self._flush_immediately = True
         self._call_batch: list[list[Any]] = []
+        self._file_batch: list[tuple[tsi.FileCreateReq, str, bool]] = []
+        self._file_batch_seen: set[tuple[str, str]] = set()
         self._use_async_insert = use_async_insert
         self._model_to_provider_info_map = read_model_to_provider_info_map()
         self._file_storage_client: Optional[FileStorageClient] = None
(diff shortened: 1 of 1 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1534,11 +1540,61 @@ def resolve_extra(extra: list[str], val: Any) -> PartialRefResult:
...
+            if client is not None and use_file_storage:
+                bucket_files.append((req, digest))
+            else:
+                clickhouse_files.append((req, digest))
+
+        # Write to bucket storage with fallback to ClickHouse
+        for req, digest in bucket_files:
+            try:
+                self._file_create_bucket(req, digest, client)
+            except FileStorageWriteError:
```
zbirenbaum, line 1569: What happens if this fails?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1534,11 +1540,61 @@ def resolve_extra(extra: list[str], val: Any) -> PartialRefResult:
...
+            total_bytes += len(req.content)
+            if client is not None and use_file_storage:
+                bucket_files.append((req, digest))
+            else:
+                clickhouse_files.append((req, digest))
+
+        # Write to bucket storage with fallback to ClickHouse
+        for req, digest in bucket_files:
+            try:
+                self._file_create_bucket(req, digest, client)
```
zbirenbaum, line 1568: We could use coroutines or threads to do these all at the same time. If a huge file is being uploaded before a bunch of small ones everything else has to wait.

gtarpenning: everything has to wait already. adding threads opens a whole new can of worms, and we can always do that in the future

PR #5559, opened 2025-10-10
fix(weave): Batch file writes

tssweeney:
[WB-27711](https://wandb.atlassian.net/browse/WB-27711)

batches file inserts

[WB-27711]: https://wandb.atlassian.net/browse/WB-27711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace_server/test_clickhouse_batching.py
+++ b/tests/trace_server/test_clickhouse_batching.py
@@ -105,7 +105,7 @@ def test_clickhouse_batching():
         # Execute the batch
         trace_server.call_start_batch(batch_req)
 
-        # THE KEY ASSERTION: 
+        # THE KEY ASSERTION:
         # Verify that there are exactly 2 inserts:
         # 1 call that is the batch of call inserts
         # 1 call that is the batch of file inserts


weave/trace_server/clickhouse_schema.py
```diff
@@ -206,3 +206,18 @@ class SelectableCHObjSchema(BaseModel):
...
+    digest: str
+    chunk_index: int
+    n_chunks: int
+    name: str
+    val_bytes: bytes
+    bytes_stored: int
+    file_storage_uri: Optional[str]
+
+
+ALL_FILE_CHUNK_INSERT_COLUMNS = list(FileChunkCreateCHInsertable.model_fields.keys())
```
gtarpenning: this should be `sorted` too for a consist column ordering

tssweeney: 👍

tests/trace_server/test_clickhouse_batching.py
```diff
@@ -0,0 +1,120 @@
...
+            ]
+        )
+
+        # Execute the batch
+        trace_server.call_start_batch(batch_req)
+
+        # THE KEY ASSERTION: Verify that 3 calls resulted in exactly 1 ClickHouse insert
+        # This confirms the batching behavior is working correctly
+        insert_call_count = mock_ch_client.insert.call_count
+        assert insert_call_count == 2, (
```
gtarpenning: maybe fix comment, explain why there are 2 inserts not 1?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -275,10 +278,13 @@ def call_batch(self) -> Iterator[None]:
         self._flush_immediately = False
         try:
             yield
+            self._flush_immediately = True
             self._flush_calls()
             self.kafka_producer.flush()
+            self._flush_file_chunks()
```
gtarpenning: Are we sure this is the correct order? This might make calls available in the UI before the content is actually present in the db. Also if online monitors have to use the content (audio?) then we should do the queue produce last.

gtarpenning: so maybe:
```python
self._flush_file_chunks()
self._flush_calls()
self.kafka_producer.flush()
```

not totally sure though. Would love to test on some large payloads in qa or something.

tssweeney: that is fair, i'll move to the top

PR #5562, opened 2025-10-13
chore(weave): Add Windows tests

andrewtruong:
https://wandb.atlassian.net/browse/WB-28277
https://wandb.atlassian.net/browse/WB-28242
https://wandb.atlassian.net/browse/WB-27498

Adds tests for Windows and fixes an issue where Op deserialization did not work on Windows

### DIFF
--- a/tests/trace/data_serialization/test_cases/media_cases.py
+++ b/tests/trace/data_serialization/test_cases/media_cases.py
@@ -471,7 +471,7 @@ media_cases = [
             "weave_type": {"type": "weave.type_wrappers.Content.content.Content"},
             "files": {
                 "content": "xfOhnNfgQxRzgWZ6DC1QEGt9vrJWcathymKPPZQmmIw",
-                "metadata.json": "v5gYkptifYVGd0JYbW3U0jYQPSKkLzEI3eamc0cnCG8"
+                "metadata.json": "0tY4LYkQE9BXzCQDItzaUoFLd3lesQ0RkuHNMuXQJIk"
                 if sys.platform == "win32"
                 else "0tY4LYkQE9BXzCQDItzaUoFLd3lesQ0RkuHNMuXQJIk",
             },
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -189,16 +189,6 @@ def tests(session, shard):
         "--cov-branch",
     ]
 
-    # Memray profiling is optional and controlled via environment variable
-    # (not working with trace_server shard atm)
-    if os.getenv("WEAVE_USE_MEMRAY") == "1" and shard != "trace_server":
```
andrewtruong, line 194: Memray profiling was not working with Windows.  This was also causing flake issues on linux, so removing for now

tests/trace/test_client_server_caching.py
```diff
@@ -200,7 +200,7 @@ def get_latency_for_server(server: TraceServerInterface, count: int):
     added_latency = latency_with_cache - latency_without_cache
     print(f"Added latency: {added_latency}")
 
-    assert added_latency < 0.001
+    assert added_latency < 0.002
```
andrewtruong, line 203: This test is slightly slower on Windows for some reason

tests/conftest.py
```diff
@@ -556,9 +556,13 @@ def post(url, data=None, json=None, **kwargs):
 @pytest.fixture(autouse=True)
 def caching_client_isolation(monkeypatch, tmp_path):
     """Isolate cache directories for each test to prevent cross-test contamination."""
-    test_specific_cache_dir = (
-        tmp_path / f"weave_cache_{get_test_name().replace('/', '_').replace('::', '_')}"
-    )
+    # Replace characters that are invalid in Windows paths
+    # Windows disallows: < > : " / \ | ? *
+    test_name = get_test_name()
+    for char in ["/", "\\", ":", "*", "?", '"', "<", ">", "|"]:
```
chance-wnb, line 562: Do you need to do this basing on OS type?

eg 
```
if sys.platform == "win32":

```

tests/trace/test_client_server_caching.py
chance-wnb: Maybe windows VM is slow? :)

PR #5567, opened 2025-10-14
chore(weave): Unpin pydantic

andrewtruong:


jtschoonhoven:
+1 this tripped me up

weave/utils/dict_utils.py
```diff
@@ -109,7 +109,7 @@ def sum_dict_leaves(dicts: list[dict]) -> dict:
     for k in nested_dicts.keys():
         result[k] = sum_dict_leaves(nested_dicts[k])
 
-    return result
+    return convert_defaultdict_to_dict(result)
```
andrewtruong, line 112: This is required due to behaviour change in pydantic==2.12.0
https://github.com/pydantic/pydantic/issues/12376

PR #5571, opened 2025-10-14
chore(weave): Add serialization assertions for the remaining data types: Media Types (1)

tssweeney:
Adds test assertions for the shape of all Media Types (and some test file refactoring)

A few logic changes to ensure the same data is always serialized the same way:
* removes the id from Content. I am not sure why this is used at all and makes each content type unique, resulting in massive duplication in the backend
* Removes the local path from serialization for similar reasons as the above

### DIFF
--- a/tests/trace/data_serialization/test_cases/media_cases.py
+++ b/tests/trace/data_serialization/test_cases/media_cases.py
@@ -221,7 +221,7 @@ media_cases = [
             "weave_type": {"type": "weave.type_wrappers.Content.content.Content"},
             "files": {
                 "content": "xfOhnNfgQxRzgWZ6DC1QEGt9vrJWcathymKPPZQmmIw",
-                "metadata.json": "5G0s5f53CdX1LbhCwq3i3ue5M7SOXcKlv7YFTdz6Lho",
+                "metadata.json": "0tY4LYkQE9BXzCQDItzaUoFLd3lesQ0RkuHNMuXQJIk",
             },
             "load_op": "weave:///shawn/test-project/op/load_weave.type_wrappers.Content.content.Content:pHK80K8ec7lRHbNwjR2LjWQXRxAV8n5BmkixcTamz2k",
         },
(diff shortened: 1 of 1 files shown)


weave/type_wrappers/Content/content_types.py
```diff
@@ -17,7 +17,7 @@
 # It is used to 'restore' an existing content object
 class ResolvedContentArgsWithoutData(TypedDict):
     # Required Fields
-    id: str
+    # id: str
```
zbirenbaum: Is this supposed to be commented?

weave/type_wrappers/Content/utils.py
```diff
@@ -93,10 +93,10 @@ def is_valid_path(input: str | Path) -> bool:
 
 def default_filename(
     extension: str | None,
-    mimetype: str,
+    mimetype: str | None,
```
zbirenbaum: Mimetype should never be None, `application/octet-stream` is the default

tssweeney: removed

weave/type_wrappers/Content/content_types.py
tssweeney: removed

tests/trace/data_serialization/test_cases/media_cases.py
```diff
@@ -0,0 +1,255 @@
...
+            },
+            {
+                "digest": "Aoxws9QUryX0YiZ8ScTAyi4YzX2SO5QHTsLsYABBMjc",
+                "exp_content": VIDEO_BYTES,
+            },
+        ],
+        equality_check=lambda a, b: a.duration
+        == b.duration,  # could do better, but this is a good start
```
zbirenbaum, line 210: I think a duration check here is fine. We already test this elsewhere so if it was re-encoding and downscaling quality we'd know, no reason to check frame by frame except to make the test take longer. Can probably remove this comment.

PR #5576, opened 2025-10-15
chore(weave): Add serialization assertions for the remaining data types: Container Types (2)

tssweeney:
1. Adds container types to the test suite


andrewtruong (approved):
I think you should also test the nested case

PR #5577, opened 2025-10-15
chore(weave): Add serialization assertions for the remaining data types: Core Types (3)

tssweeney:
Adds support for core weave types in the serialization tests

### DIFF
--- a/tests/trace/data_serialization/test_cases/media_cases.py
+++ b/tests/trace/data_serialization/test_cases/media_cases.py
@@ -173,6 +173,7 @@ media_cases = [
         ],
         equality_check=lambda a, b: a.markup == b.markup
         and a.code_theme == b.code_theme,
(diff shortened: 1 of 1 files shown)


tests/trace/data_serialization/test_cases/core_type_factory.py
```diff
@@ -0,0 +1,23 @@
...
+@register_object
+class CustomObject(weave.Object):
+    my_name: str
+
+    @weave.op
+    def say_hello(self):
+        return "hello " + self.my_name
+
+    @classmethod
+    def from_obj(cls, obj: weave.Object):
```
andrewtruong: The `obj` here is actually a `WeaveObject`, not `weave.Object` (yes, the nomenclature could use work...)

tssweeney: 👍

PR #5578, opened 2025-10-15
chore(weave): Add serialization assertions for the remaining data types: Evaluation Library Types (4)

tssweeney:
Adds Evaluation library objects

### DIFF
--- a/tests/trace/data_serialization/test_serialization_correctness.py
+++ b/tests/trace/data_serialization/test_serialization_correctness.py
@@ -33,16 +33,25 @@ independent of the actual code that is used to serialize the data.
 """
 
 
+@pytest.fixture
+def set_weave_logger_to_debug():
+    logger = logging.getLogger("weave")
+    current_level = logger.level
+    logger.setLevel(logging.DEBUG)
+    try:
+        yield
+    finally:
+        logger.setLevel(current_level)
+
+
 @pytest.mark.parametrize(
     "case",
     cases,
     ids=lambda case: case.id,
 )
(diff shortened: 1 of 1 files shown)


tests/trace/data_serialization/test_cases/library_cases.py
```diff
@@ -0,0 +1,323 @@
...
+            },
+        ],
+        # Sad ... equality is really a pain to assert here (and is broken)
+        # TODO: Write a good equality check and make it work
+        equality_check=lambda a, b: True,
```
andrewtruong, line 274: Maybe this should be fixed before merging

andrewtruong (commented):
We should test the library objects individually as well as composites

tests/trace/data_serialization/test_cases/library_cases.py
tssweeney: I agree. However, I am really trying to just get the current expectations fixed so that I can modify for a customer.

andrewtruong: Maybe I am missing something -- isn't the assert the expectation?  How can you be sure unless the test actually passes?

andrewtruong (approved):
Approving despite lack of asserts :)

tests/trace/data_serialization/test_cases/library_cases.py
tssweeney: oh, the equality check is just one of many assertions. we do a bunch of validation on the actual payloads as well.

PR #5579, opened 2025-10-15
chore(weave): Add serialization assertions for the remaining data types: Config Types (5)

tssweeney:
Adds the config types


tssweeney:
> where are the asserts?

They are in the test harness

PR #5581, opened 2025-10-15
feat(weave): Warn users when setting WEAVE_* env vars after init

andrewtruong:
## Summary

- Adds environment variable monitoring to detect when users attempt to set or modify `WEAVE_*` environment variables after Weave initialization
- Issues warnings to help users understand that environment variables should be set before `weave.init()` is called
- Includes comprehensive test coverage and updated documentation

## Motivation

Many Weave settings are read from environment variables during initialization. Setting these variables after `weave.init()` has been called typically has no effect, which can lead to user confusion and debugging challenges. This PR adds runtime warnings to alert users immediately when they make this mistake.

## Implementation Details

1. **env_monitor module** (`weave/trace/env_monitor.py`):
   - Tracks initialization state
   - Wraps `os.environ` to intercept environment variable modifications
   - Emits warnings when `WEAVE_*` variables are set/modified after init
   - Avoids duplicate warnings for the same variable

2. **Integration** (`weave/trace/weave_init.py`):
   - Automatically installs the monitor during `init_weave()`
   - Captures initial state of all `WEAVE_*` variables
   - Marks Weave as initialized to enable monitoring

3. **Tests** (`tests/trace/test_env_monitor.py`):
   - Covers various scenarios (late setting, modification, deletion)
   - Tests that pre-init variables don't trigger warnings
   - Verifies non-`WEAVE_*` variables are ignored
   - Tests different `os.environ` mutation methods (`update`, `setdefault`, etc.)

4. **Documentation** (`docs/docs/guides/core-types/env-vars.md`):
   - Adds prominent warning about timing requirements
   - Updates code examples to show proper usage
   - Clarifies when variables take effect

## Test plan

- [x] Unit tests pass for env_monitor module
- [x] Manual testing confirms warnings are issued correctly
- [x] Existing environment variables set before init continue to work without warnings
- [x] Non-`WEAVE_*` variables don't trigger false positives
- [ ] Run full test suite to ensure no regressions

### DIFF
--- a/docs/docs/guides/core-types/env-vars.md
+++ b/docs/docs/guides/core-types/env-vars.md
@@ -2,6 +2,10 @@
 
 W&B Weave provides a set of environment variables to configure and optimize its behavior. You can set these variables in your shell or within scripts to control specific functionality.
 
+:::warning
+Environment variables must be set **before** calling `weave.init()`. Setting or modifying `WEAVE_*` environment variables after initialization may not have any effect, as most settings are read during initialization. Starting from the latest version, Weave will warn you if you attempt to set a `WEAVE_*` environment variable after initialization.
+:::
+
 ```bash
 # Example of setting environment variables in the shell
 export WEAVE_PARALLELISM=10  # Controls the number of parallel workers
@@ -10,10 +14,15 @@ export WEAVE_PRINT_CALL_LINK=false  # Disables call link output
 
 ```python
 # Example of setting environment variables in Python
+# IMPORTANT: Set these BEFORE calling weave.init()
 import os
 
 os.environ["WEAVE_PARALLELISM"] = "10"
 os.environ["WEAVE_PRINT_CALL_LINK"] = "false"
+
+# Now initialize Weave
+import weave
+weave.init("my-project")
 ```
 
 ## Available Environment Variables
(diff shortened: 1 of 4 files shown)


docs/docs/guides/core-types/env-vars.md
neutralino1, line 1: cc @dbrian57

weave/trace/env_monitor.py
```diff
@@ -0,0 +1,147 @@
...
+
+def _check_for_late_env_vars() -> None:
+    """Check if any WEAVE_* env vars have been added or modified after initialization."""
+    if not _weave_initialized:
+        return
+
+    # Get current WEAVE_* env vars
+    current_weave_env_vars = {
+        key: value for key, value in os.environ.items() if key.startswith("WEAVE_")
+    }
```
neutralino1, line 51: Nit: maybe DRY with line 33.

weave/trace/env_monitor.py
```diff
@@ -0,0 +1,147 @@
...
+import warnings
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+# Track when weave has been initialized
+_weave_initialized = False
+
+# Store the initial state of WEAVE_* environment variables
+_initial_weave_env_vars: dict[str, str | None] = {}
```
neutralino1, line 21: I thought this would be used as source of truth, but it's only used to monitor and compare. Should this be merged with the code path that actually reads and stores settings?

neutralino1 (commented):
A couple of comments.

PR #5586, opened 2025-10-15
feat(weave): Support context propagation in EvaluationLogger

andrewtruong:
This changes `EvaluationLogger` to defer finishing calls so summary stats can be accurately computed.  It also introduces a new context manager API for nesting calls under a given `predict` or `score` call

See before and after results for the snippet below:
```py
import openai
import weave

client = weave.init("nested-imperative-evaluation")
oai = openai.OpenAI()

ev = weave.EvaluationLogger()

user_prompt = "Tell me a joke"
with ev.log_prediction(inputs={"user_prompt": user_prompt}) as pred:
    result = oai.chat.completions.create(
        model="gpt-5-mini",
        messages=[{"role": "user", "content": user_prompt}],
    )

    # Set the output of the "predict" call
    pred.output = result.choices[0].message.content

    # Log scores
    pred.log_score("correctness", 1.0)
    pred.log_score("ambiguity", 0.3)
    with pred.log_score("llm_judge") as score:
        result = oai.chat.completions.create(
            model="gpt-4o-mini",
            messages=[
                {"role": "system", "content": "Rate how funny the joke is from 1-5"},
                {"role": "user", "content": pred.output},
            ],
        )
        score.value = result.choices[0].message.content

ev.log_summary({"avg_score": 1.0})

```

## Before
image

## After
image

### DIFF
--- a/weave/evaluation/eval_imperative.py
+++ b/weave/evaluation/eval_imperative.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import atexit
+import contextlib
 import datetime
 import json
 import keyword
@@ -12,7 +13,10 @@ from contextlib import contextmanager
 from contextvars import ContextVar
 from threading import Lock
 from types import MethodType
-from typing import Annotated, Any, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Annotated, Any, TypeVar, Union, cast, overload
+
+if TYPE_CHECKING:
+    from weave.trace.call import Call
 
 from pydantic import (
     BaseModel,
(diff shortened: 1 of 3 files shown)


adrnswanberg (approved):

weave/evaluation/eval_imperative.py
```diff
@@ -232,8 +237,172 @@ def get_scorer(
...
+
+    @property
+    def output(self) -> Any:
+        """Get the current output value."""
+        return self.score_logger._predict_output
+
+    @output.setter
+    def output(self, value: Any) -> None:
+        """Set the output value that will be used when finishing."""
+        self.score_logger._predict_output = value
```
chance-wnb, line 257: This is a really cool idea. Now I think we could have a way to record the timing for prediction. (Something to consider, no need for this PR)

PR #5587, opened 2025-10-16
calls complete with project switching -- combo poc

gtarpenning:
## Description

Instead of doing a backfill, just keep track of what projects are using the new table and which are legacy. 

Pros: 
- No mandatory backfill, new clients write to new projects and they are fast. We can still backfill, but now its not required in order to maintain existing write performance, and can be done as slow as necessary. 
- Lower risk of messing up existing data paths, all existing queries still run exactly as before, existing users should be unaffected.

Cons: 
- Maintain two separate write paths, two separate table schemas etc. Until backfill. 
- Old clients can't write to new projects, which will definenly burn some users. 

Outstanding question:
- what happens for super long running calls, like hypothetically a week, and the call start falls out of the client cache? We have no way of ingesting this call, as we need the **complete** call according to our api contract. Might be prudent to add a "slow call end manual" endpoint that can be called with the call end information, which will cause a read-before-write in the server in order to generate the completed payload. This could also be used internally to clean up "running" calls that were started but never actually completed, run as a scheduler process internal to the trace server. 

Uses core branch: https://github.com/wandb/core/pull/35414
## Testing

How was this PR tested?

### DIFF
--- a/weave/trace_server/calls_query_builder/calls_query_builder.py
+++ b/weave/trace_server/calls_query_builder/calls_query_builder.py
@@ -1,34 +1,52 @@
-"""This module builds on the orm.py module to provide a more hard-coded optimized
-query builder specifically for the "Calls" table (which is the `calls_merged`
-underlying table).
+"""Optimized query builder for calls table queries.
 
-The `CallsQuery` class is the main entry point for building a query. These
-optimizations are hand-tuned for Clickhouse - and in particular attempt to
-perform predicate pushdown where possible and delay loading expensive fields as
-much as possible. In testing, Clickhouse performance is dominated by the amount
-of data loaded into memory.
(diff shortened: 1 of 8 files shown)


weave/trace/weave_client.py
```diff
@@ -965,9 +987,85 @@ def send_end_call() -> None:
...
+                    end=EndedCallSchemaForInsert(
+                        project_id=project_id,
+                        id=call.id,
+                        ended_at=ended_at,
+                        output=output_json,
+                        summary=merged_summary,  # type: ignore[arg-type]
+                        exception=exception_str,
+                    )
+                )
+                self.server.call_end(call_end_req)
```
gtarpenning: I don't think this works. Where will this go? How would we ever get this end back? We may need a special endpoint for this...

PR #5590, opened 2025-10-17
fix(weave): Better context manager support for EvaluationLogger

andrewtruong:
Fixes an issue where the `EvaluationLogger` would not work as expected if the user specified `output` when using the contextmanager form of `log_prediction`:
```py
with log_prediction(inputs=..., output=...):
```


weave/evaluation/eval_imperative.py
```diff
@@ -237,74 +237,6 @@ def get_scorer(
 global_scorer_cache = ScorerCache()
 
 
-class _LogPredictionContext:
```
andrewtruong, line 240: The details of this are moved to `ScoreLogger`

Issue #5591, opened 2025-10-17
Dependency Conflict: `'graphql' has no attribute 'DocumentNode'`


jtschoonhoven:
TL;DR: This package requires `gql>=3.0.0` but it can be installed with `gql<3`.

Versions:
`python==3.13.2`
`weave==0.52.10`
`wandb==0.22.2`
`build-backend = "poetry.core.masonry.api"`

Attempting to follow the quickstart, I set up a new project with `poetry init` and `poetry add wandb weave`. Attempting to run the code, I immediately faced this error:

```
  File "~/src/weave/weave/compat/wandb/wandb_thin/internal_api.py", line 20, in Api
    def query(self, query: graphql.DocumentNode, **kwargs: Any) -> Any:
                           ^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'graphql' has no attribute 'DocumentNode'
```

That's just a type annotation, but even if you wrap the annotations in quotes to suppress that error, then you still get this one:

```
  File "~/src/weave/weave/compat/wandb/wandb_thin/internal_api.py", line 47, in query
    session = client.connect_sync()  # type: ignore
              ^^^^^^^^^^^^^^^^^^^
AttributeError: 'Client' object has no attribute 'connect_sync'
```

These attributes were all added in gql v3, so `pyproject.toml` needs to be updated with that version requirement. I'm making a PR.

PR #5592, opened 2025-10-17
fix(weave): require gql >= 3.0.0 in pyproject.toml

jtschoonhoven:
## Description

This package depends on features that are only available in gql v3 and above, but it can be installed with gql v2 and below. E.g. when trying to follow the quickstart with `poetry install wandb weave`, the default behavior seems to be that gql v2 will be installed and the example code breaks.

- Fixes https://github.com/wandb/weave/issues/5591

Requires `gql>=3.0.0` in pyproject.toml

## Testing

Works on my machine.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -387,3 +387,137 @@ jobs:
         uses: re-actors/alls-green@release/v1
         with:
           jobs: ${{ toJSON(needs) }}
+
+  # ==== Windows Jobs ====
+  windows_trace_no_server:
+    name: Windows Trace non-trace server tests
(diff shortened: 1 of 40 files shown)


jtschoonhoven:

jtschoonhoven:
recheck

jtschoonhoven:
recheck

jtschoonhoven:
@andrewtruong CI is stuck?

andrewtruong:
@jtschoonhoven our CI has trouble when an external user opens a PR, so we have to make a copy of this PR to kick it off.  I'll cherry-pick your commits into a new PR and it should work

PR #5600, opened 2025-10-22
fix(weave): prevent deferred import bug in openai postprocessor

bcsherma:
## Description

This PR removes all lazy imports from the openai autopatch integration and moves the imports to the module level. The module `openai_sdk.py` is only imported by the autopatcher and only runs if `openai` is found in `sys.modules`, so this does not inadvertently make `openai` a required dependency of `weave`. I tested this by importing `weave` and running `weave.init` in a python environment without `openai` installed.

This is also a win because the wrapper functions don't need to run that import over and over again for every completion.

Also fixes a yet unexplained bug. Lazy imports of some types from the openai sdk fail in our postprocessor when stremaing is enabled and weave is imported after openai in the user code.

Eg
```
import openai
import weave

weave.init(...)
client = openai.Client()

    stream=True,
    model="gpt-4o-mini",
    messages=[
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "Tell me a joke."},
    ],
)
```

will fail with
```
weave: Error closing iterator, call data may be incomplete:
weave: Traceback (most recent call last):
weave:   File "/Users/ben/core/services/weave-python/weave-public/weave/trace/op.py", line 1429, in _call_on_close_once
weave:     self._on_close()  # type: ignore
weave:     ^^^^^^^^^^^^^^^^
weave:   File "/Users/ben/core/services/weave-python/weave-public/weave/trace/op.py", line 1637, in on_close
weave:     on_finish(acc.get_state(), None)
weave:   File "/Users/ben/core/services/weave-python/weave-public/weave/trace/op.py", line 1681, in wrapped_on_finish
weave:     value = on_finish_post_processor(value)
weave:             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
weave:   File "/Users/ben/core/services/weave-python/weave-public/weave/integrations/openai/openai_sdk.py", line 68, in openai_on_finish_post_processor
weave:     from openai.types.chat import ChatCompletion, ChatCompletionChunk
weave: ModuleNotFoundError: import of openai.types.chat halted; None in sys.modules
weave:  (subsequent messages of this type will be suppressed)
```

Somehow these conditions lead to the lazy imports failing- and the `sys.modules` for all `openai.types.chat.*` are observably `None` in the runtime. This PR prevents the bug from occurring, but the root cause of the bug is unclear.

### DIFF
--- a/weave/integrations/openai/openai_sdk.py
+++ b/weave/integrations/openai/openai_sdk.py
@@ -7,6 +7,21 @@ from functools import wraps
 from typing import TYPE_CHECKING, Any, Callable
 from urllib.parse import urlparse
 
+# Do not import this module directly, it should only be invoked
+from openai._legacy_response import LegacyAPIResponse
+from openai._response import APIResponse
+from openai.types.chat import ChatCompletion, ChatCompletionChunk
+from openai.types.chat.chat_completion_chunk import (
+    ChoiceDeltaFunctionCall,
+    ChoiceDeltaToolCall,
+    ChoiceDeltaToolCallFunction,
+)
+from openai.types.chat.chat_completion_message import FunctionCall
+from openai.types.chat.chat_completion_message_tool_call import (
+    ChatCompletionMessageToolCall,
+    Function,
+)
+
 import weave
 from weave.integrations.patcher import MultiPatcher, NoOpPatcher, SymbolPatcher
 from weave.trace.autopatch import IntegrationSettings, OpSettings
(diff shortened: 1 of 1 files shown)


bcsherma:
* **#5600** <a href="https://app.graphite.dev/github/pr/wandb/weave/5600?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/5600?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #5604, opened 2025-10-22
chore(weave): force round robin in kafka producer

gtarpenning:
## Description

[WB-28652](https://wandb.atlassian.net/browse/WB-28652)

Rather than do sticky batch assignment, see if we can get more even partition sizing with an explicitly random strategy

## Testing

Before:
image

After:
todo

[WB-28652]: https://wandb.atlassian.net/browse/WB-28652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -240,8 +240,7 @@ jobs:
             "13",
             #
           ]
-        nox-shard:
-          [
+        nox-shard: [
             "trace1",
             "trace2",
             "trace3",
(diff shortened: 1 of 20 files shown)


weave/trace_server/kafka.py
```diff
@@ -60,6 +60,7 @@ def from_env(
             "message.timeout.ms": total_timeout_ms,
             "retries": 1,
             "retry.backoff.ms": request_retry_backoff_ms,
+            "partitioner": "murmur2_random",  # Explicit round robin
```
neutralino1, line 63: That's a random name indeed...

PR #5621, opened 2025-10-24
feat(weave): Add new CRUD endpoints for interacting with Models

andrewtruong:
https://wandb.atlassian.net/browse/WB-28668

This PR adds CRUD endpoints for interacting with Models as part of the V2 Evals API.

Pairs with https://github.com/wandb/core/pull/35735

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -1984,19 +1984,17 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         )
         obj_read_res = self._obj_read_with_retry(obj_read_req)
 
-        # Build model reference
-        entity, project = req.project_id.split("/")
-        model_ref = object_creation_utils.build_object_ref(
-            entity=entity,
-            project=project,
-            object_id=object_id,
-            digest=obj_result.digest,
-        )
+        # Build model reference - external adapter will convert to external format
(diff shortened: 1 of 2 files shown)


weave/trace_server/reference/generate.py
```diff
@@ -443,6 +444,90 @@ def evaluation_delete_v2(
...
         return service.trace_server_interface.evaluation_delete_v2(req)
 
+    @router.post(
+        "{entity}/{project}/models",
+        tags=[V2_MODELS_TAG_NAME],
+    )
+    def model_create_v2(
+        entity: str,
+        project: str,
+        body: tsi.ModelCreateV2Body,
```
chance-wnb, line 454: May I ask why the method signatures in `generate_routes_v2` are different than the ones in TraceServerInterfaceV2 ? (method names are the same, but the parameters here seem to have been expanded) 

How are these `generate_routes_v2` routes going to be used in the future?

andrewtruong: Great question.

In the interface definition, we have the expected `Req/Res` pairs that fully describe what the expected inputs and outputs are.

In the reference server, we want the endpoint to take a certain shape (e.g. `{entity}/{project}/models`) and FastAPI doesn't have a way to destructure this info out of the class.  Instead we need to make `entity` and `project` arguments to the resolver.  Then to avoid duplicating these arguments in the body, we have to exclude them there, which results in the pattern you see in `generate.py`.

PR #5625, opened 2025-10-25
chore(weave): Link to the Eval page instead of the Eval's call

andrewtruong:
Previously we linked to the Eval's call page.
image

Now we link to the Eval page, which is more relevant
image

### DIFF
--- a/weave/evaluation/eval_imperative.py
+++ b/weave/evaluation/eval_imperative.py
@@ -929,6 +929,45 @@ class EvaluationLogger(BaseModel):
 
         return pred
 
+    def log_example(
+        self, inputs: dict[str, Any], output: Any, scores: dict[str, ScoreType]
+    ) -> None:
+        """Log a complete example with inputs, output, and scores.
+
+        This is a convenience method that combines log_prediction and log_score
+        for when you have all the data upfront.
+
+        Args:
+            inputs: The input data for the prediction
+            output: The output value
+            scores: Dictionary mapping scorer names to score values
+
+        Example:
+        ```python
+        ev = EvaluationLogger()
+        ev.log_example(
+            inputs={'q': 'What is 2+2?'},
+            output='4',
+            scores={'correctness': 1.0, 'fluency': 0.9}
+        )
+        ```
+        """
+        if self._is_finalized:
+            raise ValueError(
+                "Cannot log example after evaluation has been finalized. "
+                "Call log_example before calling finish() or log_summary()."
+            )
+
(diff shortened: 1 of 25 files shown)


weave/evaluation/eval_imperative.py
```diff
@@ -708,7 +708,17 @@ def ui_url(self) -> str | None:
         # at init time.
         if self._evaluate_call is None:
             return None
-        return self._evaluate_call.ui_url
+        if not self._evaluate_call.id:
+            return None
+
+        from weave.trace import urls
```
tssweeney, line 714: why import in the middle. this is a smell of unchecked claude

weave/trace/urls.py
```diff
@@ -33,3 +33,10 @@ def leaderboard_path(entity_name: str, project_name: str, object_name: str) -> s
 
 def redirect_call(entity_name: str, project_name: str, call_id: str) -> str:
     return f"{remote_project_root_url(entity_name, project_name)}/r/call/{call_id}"
+
+
+def evaluation_call_url(entity_name: str, project_name: str, call_id: str) -> str:
+    """Generate a URL for an evaluation call that links to the evaluations page."""
+    project_id = f"{entity_name}/{project_name}"
+    peek_path = f"/{project_id}/calls/{call_id}?hideTraceTree%3D1"
+    return f"{project_weave_root_url(entity_name, project_name)}/evaluations?view=evaluations_default&peekPath={peek_path}"
```
tssweeney, line 42: we should not do this - encoding a peek path inside the dumped URL is quite ugly. The `/r/call` path above was invented exactly for this... we should just change what the "(r)edirect" is

weave/evaluation/eval_imperative.py
andrewtruong: fixed

weave/trace/urls.py
andrewtruong: Can you show me what that would look like here?  I'm not sure I fully understand

tssweeney: In core, `frontends/app/src/pages/RedirectPage/RedirectCall.tsx` controls where the `/r/call/id` gets routed. let's change that in the frontend so it does not create a link dependency between client and ui

andrewtruong: I'm not sure how to get this to redirect to the evaluation tab though

andrewtruong: Does it require a core-side change?

PR #5630, opened 2025-10-27
feat(weave): Support properly retrieving a dataset back in TS SDK

chance-wnb:
## Description

- Implements loading for Dataset and Table objects in the TypeScript SDK
- Adds URI parsing utilities for Weave references

This PR adds support for loading of Dataset and Table objects, allowing them to be retrieved from the server on demand. It includes a new URI parser for handling different types of Weave references 

The implementation ensures that dataset rows can be accessed synchronously after loading.

The PR also extends the InMemoryTraceServer to support table operations for testing scenarios. 

## Testing

- Added comprehensive tests for Dataset creation and retrieval
- Implemented round-trip testing to verify data integrity
- Extended InMemoryTraceServer to support table operations for testing

## Docs:
https://github.com/wandb/docs/pull/1809

### DIFF
--- a/sdks/node/src/inMemoryTraceServer.ts
+++ b/sdks/node/src/inMemoryTraceServer.ts
@@ -44,10 +44,17 @@ interface File {
   content: Blob;
 }
 
+interface Table {
+  project_id: string;
+  digest: string;
+  rows: Array<{digest: string; val: any}>;
+}
+
 export class InMemoryTraceServer {
   private _calls: Call[] = [];
   private _objs: Obj[] = [];
   private _files: File[] = [];
+  private _tables: Table[] = [];
   private _lastCallCount: number = 0;
   private _lastChangeTime: number = Date.now();
 
@@ -144,6 +151,80 @@ export class InMemoryTraceServer {
       return {
         data: {
           digest: digest,
+        },
+      };
+    },
+
+    objReadObjReadPost: async (req: {
+      project_id: string;
+      object_id: string;
+      digest?: string;
+    }) => {
+      const obj = this._objs.find(
+        o =>
+          o.project_id === req.project_id &&
+          o.object_id === req.object_id &&
+          (req.digest ? o.digest === req.digest : o.is_latest === 1)
+      );
+
+      if (!obj) {
+        throw new Error(
+          `Object not found: ${req.project_id}/${req.object_id}${req.digest ? ':' + req.digest : ''}`
+        );
+      }
+
+      return {
+        data: {
+          obj: obj,
+        },
+      };
+    },
+  };
+
+  table = {
+    tableCreateTableCreatePost: async (req: {
+      table: {project_id: string; rows: any[]};
+    }) => {
+      const digest = this.generateDigest(req.table.rows);
+
+      // Create row entries with individual digests
+      const rows = req.table.rows.map(rowVal => ({
+        digest: this.generateDigest(rowVal),
+        val: rowVal,
+      }));
+
+      const newTable: Table = {
+        project_id: req.table.project_id,
+        digest: digest,
+        rows: rows,
+      };
+
+      this._tables.push(newTable);
+
+      return {
+        data: {
+          digest: digest,
+        },
+      };
+    },
+
+    tableQueryTableQueryPost: async (req: {
+      project_id: string;
+      digest: string;
+    }) => {
+      const table = this._tables.find(
+        t => t.project_id === req.project_id && t.digest === req.digest
+      );
+
+      if (!table) {
+        throw new Error(
+          `Table not found: ${req.project_id}/table/${req.digest}`
+        );
+      }
+
+      return {
+        data: {
+          rows: table.rows,
         },
       };
     },
--- a/sdks/node/src/uriParser.ts
+++ b/sdks/node/src/uriParser.ts
@@ -1,0 +1,124 @@
+/**
+ * Utility for parsing Weave ref URIs.
+ *
+ * Supports parsing different types of Weave refs:
+ * - Table refs: weave:///entity/project/table/digest
+ * - Object refs: weave:///entity/project/object/name:digest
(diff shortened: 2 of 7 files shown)


chance-wnb:
* **#5630** <a href="https://app.graphite.dev/github/pr/wandb/weave/5630?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/wandb/weave/5630?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/inMemoryTraceServer.ts
```diff
@@ -147,6 +154,80 @@ export class InMemoryTraceServer {
         },
       };
     },
+
+    objReadObjReadPost: async (req: {
```
neutralino1, line 158: The name of these functions is strange. It duplicated `objRead`, can we do better?

sdks/node/src/inMemoryTraceServer.ts
```diff
@@ -147,6 +154,80 @@ export class InMemoryTraceServer {
...
+      return {
+        data: {
+          obj: obj,
+        },
+      };
+    },
+  };
+
+  table = {
+    tableCreateTableCreatePost: async (req: {
```
neutralino1, line 185: Same.

sdks/node/src/inMemoryTraceServer.ts
```diff
@@ -147,6 +154,80 @@ export class InMemoryTraceServer {
...
+      table: {project_id: string; rows: any[]};
+    }) => {
+      const digest = this.generateDigest(req.table.rows);
+
+      // Create row entries with individual digests
+      const rows = req.table.rows.map(rowVal => ({
+        digest: this.generateDigest(rowVal),
+        val: rowVal,
+      }));
```
neutralino1, line 194: Any performance concerns here?

sdks/node/src/inMemoryTraceServer.ts
```diff
@@ -147,6 +154,80 @@ export class InMemoryTraceServer {
...
+      this._tables.push(newTable);
+
+      return {
+        data: {
+          digest: digest,
+        },
+      };
+    },
+
+    tableQueryTableQueryPost: async (req: {
```
neutralino1, line 211: `strangeFunctionStrangeFunctionName`

sdks/node/src/uriParser.ts
```diff
@@ -0,0 +1,124 @@
...
+      project: string;
+      name: string;
+      digest: string;
+    }
+  | {
+      type: 'call';
+      entity: string;
+      project: string;
+      id: string;
+    };
```
neutralino1, line 37: Could this be
```typescript
{
  type: 'table' | 'object' | 'op' | 'call';
  entity: string;
  project: string;
  name?: string;
  digest: string;
}
```
also calls have names too no?

sdks/node/src/uriParser.ts
```diff
@@ -0,0 +1,124 @@
...
+ * @returns Parsed URI components or null if invalid
+ *
+ * @example
+ * ```typescript
+ * const parsed = parseWeaveUri('weave:///wandb/project/table/abc123...');
+ * if (parsed && parsed.type === 'table') {
+ *   console.log(parsed.digest);  // 'abc123...'
+ * }
+ * ```
+ */
```
neutralino1, line 52: The definitive mark of vide-coded functions 😆

sdks/node/src/inMemoryTraceServer.ts
chance-wnb: This is a test harness file. it will only be used in test cases. So I guess it should be fine.

sdks/node/src/inMemoryTraceServer.ts
chance-wnb: While you are saying it, I am shocked too. 

But all these come from the auto-gen`ed http client stubs: https://github.com/wandb/weave/blob/66b158cbc69c9af3d5aeeebdb9bcb37b40e8bd07/sdks/node/src/generated/traceServerApi.ts#L1541

And those have existed at least for a year. 

I will ask Andrew tomorrow about why things are this way. but most likely it might be the way it is.

sdks/node/src/inMemoryTraceServer.ts
chance-wnb: lol. I will investigate, but likely we might have to leave it as is.

sdks/node/src/uriParser.ts
chance-wnb: Yeah, according to the current consumption pattern the current pattern might be the best.

Because different ref types really don't have the same components beside entity or project. I think the current type unions is easier to consume despite more verbose. (so the consumers don't have to prepare for a case that a field might be optional)  

In case of call ref, it is used only when we need to attach a feedback to a call. And it is very internal (users almost have no way to consume one). https://github.com/wandb/weave/blob/66b158cbc69c9af3d5aeeebdb9bcb37b40e8bd07/weave/trace_server/trace_server_interface.py#L921

So no, call refs actually don't have names.

sdks/node/src/uriParser.ts
chance-wnb: Haha, yes. want me to change anything?

sdks/node/src/inMemoryTraceServer.ts
chance-wnb: I investigated a bit. it turned out that we can fix that by putting some annotations on the server side FastAPI level. 

Details are explained in https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1761668495870889

I'd suggest we tackle this as a tech debt item separately.

chance-wnb:
The failed test case is known and irrelevant. merging.

PR #5635, opened 2025-10-28
chore(weave): Improve agent-based testing

tssweeney:
Testing in Claude/Cursor tends to be fragile and requires many iterations. I worked with Claude to revise the Agents.md so that it actually one-shots tests now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,7 @@ langchain = [
   "langchain-cohere>=0.4.5",
   "langchain-litellm>=0.2.2",
   "chromadb>=0.5.0",
-  "pysqlite3; platform_system != 'Windows'",
+  "pysqlite3; platform_system == 'Linux'",
   "opentelemetry-exporter-otlp",
   "openai>=1.99.9,<1.100.0",
 ]
--- a/tests/trace/data_serialization/test_serialization_correctness.py
+++ b/tests/trace/data_serialization/test_serialization_correctness.py
@@ -52,6 +52,10 @@ def set_weave_logger_to_debug():
 def test_serialization_correctness(
     client, case: SerializationTestCase, set_weave_logger_to_debug
 ):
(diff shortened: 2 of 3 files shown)


tests/trace/data_serialization/test_serialization_correctness.py
```diff
@@ -52,6 +52,10 @@ def set_weave_logger_to_debug():
 def test_serialization_correctness(
     client, case: SerializationTestCase, set_weave_logger_to_debug
 ):
+    # Skip image test on macOS - PIL/Pillow produces different PNG encoding on macOS vs Linux
+    # resulting in different file digests. This is expected behavior, not a bug.
+    if sys.platform == "darwin" and case.id == "image":
```
andrewtruong, line 57: Why only skip this for mac?

pyproject.toml
```diff
@@ -131,7 +131,7 @@ langchain = [
   "langchain-cohere>=0.4.5",
   "langchain-litellm>=0.2.2",
   "chromadb>=0.5.0",
-  "pysqlite3; platform_system != 'Windows'",
+  "pysqlite3; platform_system == 'Linux'",
```
andrewtruong, line 134: is this correct?

tests/trace/data_serialization/test_serialization_correctness.py
tssweeney: this always fails on mac because the written file is not the same between the OSs

pyproject.toml
tssweeney: I believe so. It is not needed on Mac as it causes issues

tests/trace/data_serialization/test_serialization_correctness.py
andrewtruong: Not critical, but I think it would be more correct to record the Mac-specific hashes as we do for Windows.

PR #5640, opened 2025-10-30
fix(weave): Fix wb_run_id schema and add wb_run_step

zbirenbaum:
## Description

- Fixes WB-28773
- todo:  https://wandb.atlassian.net/browse/DOCS-1876

 - Fixes invalid schema for wb_run_id
 - Allows passing wb_run_id from header as well as attributes
 - Introduces parsing for wandb.wb_run_step and wandb.wb_run_step_end
 - Modifies wb_run_id to only be read from wandb.wb_run_id for consistency

### Header usage example:
```
import os
import base64
import wandb
import openai
from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
from opentelemetry.sdk import trace as trace_sdk
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry import trace
from openinference.instrumentation.openai import OpenAIInstrumentor

ENTITY = "your_entity"
PROJECT = "your_project"
WANDB_API_KEY = "your_wandb_api_key"

run = wandb.init(entity=ENTITY, project=PROJECT)

WANDB_BASE_URL = "https://trace.wandb.ai"
PROJECT_ID = f"{ENTITY}/{PROJECT}"
AUTH = base64.b64encode(f"api:{WANDB_API_KEY}".encode()).decode()
OTEL_EXPORTER_OTLP_ENDPOINT = f"{WANDB_BASE_URL}/otel/v1/traces"
headers = {
    "authorization": os.environ["TRACE_SERVER_AUTH"],
    "project_id": f"{PROJECT_ID}",
    "wb_run_id": f"{PROJECT_ID}/{run.id}" 
}

tracer_provider = trace_sdk.TracerProvider()
tracer_provider.add_span_processor(
    SimpleSpanProcessor(
        OTLPSpanExporter(endpoint=OTEL_EXPORTER_OTLP_ENDPOINT, headers=headers)
    )
)

"""Set the tracer provider globally so any instrumentations that default to
the global provider will export correctly."""
trace.set_tracer_provider(tracer_provider)

# Instrument OpenAI with the configured provider
OpenAIInstrumentor().instrument(tracer_provider=tracer_provider)

def main():
    client = openai.OpenAI()
        model="gpt-3.5-turbo",
        messages=[{"role": "user", "content": "Describe OTEL in a long sentence."}],
        max_tokens=20,
        stream=True,
        stream_options={"include_usage": True},
    )
    for i, chunk in enumerate(response):
        if chunk.choices and (content := chunk.choices[0].delta.content):
            run.log({f"chunk_content_{i}": content})
            print(content, end="")

if __name__ == "__main__":
    main()
    wandb.finish()
```

### Attributes usage example:
```
import os
import base64
import wandb
import openai
from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
from opentelemetry.sdk import trace as trace_sdk
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry import trace
from openinference.instrumentation.openai import OpenAIInstrumentor
from wandb.sdk.wandb_run import Run


WANDB_API_KEY = os.environ.get("WANDB_API_KEY")
ENTITY = "your_entity"
PROJECT = "your_project"

run = wandb.init(entity=ENTITY, project=PROJECT)

WF_TRACE_SERVER_URL="https://trace.wandb.ai"
PROJECT_ID = f"{ENTITY}/{PROJECT}"
AUTH = base64.b64encode(f"api:{WANDB_API_KEY}".encode()).decode()
OTEL_EXPORTER_OTLP_ENDPOINT = f"{WF_TRACE_SERVER_URL}/otel/v1/traces"
headers = {
    "authorization": f"Basic {AUTH}",
    "project_id": f"{PROJECT_ID}",
}

class RunIdAttributeProcessor(trace_sdk.SpanProcessor):
    """A lightweight processor that injects the W&B run id on span start.

    This processor does NOT handle exporting. Keep exporting separate using
    a normal SimpleSpanProcessor hooked up to your exporter.
    """
    def __init__(self, run: Run) -> None:
        self.run = run

    def on_start(self, span, parent_context=None) -> None:  # type: ignore[override]
        if self.run.id:
            span.set_attributes(attributes={"wandb.wb_run_id": run.id})

tracer_provider = trace_sdk.TracerProvider()

tracer_provider.add_span_processor(RunIdAttributeProcessor(run=run))
tracer_provider.add_span_processor(
    SimpleSpanProcessor(
        OTLPSpanExporter(endpoint=OTEL_EXPORTER_OTLP_ENDPOINT, headers=headers)
    )
)

trace.set_tracer_provider(tracer_provider)

# Instrument OpenAI with the configured provider
OpenAIInstrumentor().instrument(tracer_provider=tracer_provider)

def main():
    client = openai.OpenAI()
    tracer_provider.get_tracer(
    for i in range(2):
            model="gpt-3.5-turbo",
            messages=[{"role": "user", "content": "Describe OTEL in a long sentence."}],
            max_tokens=20,
            stream=True,
            stream_options={"include_usage": True},
        )
        for j, chunk in enumerate(response):
            if chunk.choices and (content := chunk.choices[0].delta.content):
                run.log({f"chunk_content_{i}_{j}": content})
                print(content, end="")

if __name__ == "__main__":
    main()
    wandb.finish()
```

## Testing

Tested with above scripts and others

### DIFF
--- a/weave/trace_server/opentelemetry/python_spans.py
+++ b/weave/trace_server/opentelemetry/python_spans.py
@@ -360,9 +360,9 @@ class Span:
         # If it contains a '/' then it is a malformed conversion in adapting layer
         # If it doesn't contain a ':' while this is defined from headers it is a malformed conversion in adapting layer
         # If it does contain a ':' while this is not defined and the attribute field is defined, again we have a malformed value
-        if not wb_run_id: # headers not defined
+        if not wb_run_id:  # headers not defined
             attr_wb_run_id = wandb_attributes.get("wb_run_id") or None
-            if attr_wb_run_id and not ('/' in attr_wb_run_id or ":" in attr_wb_run_id):
+            if attr_wb_run_id and not ("/" in attr_wb_run_id or ":" in attr_wb_run_id):
                 wb_run_id = f"{project_id}:{attr_wb_run_id}"
         ### END HACK
 


weave/trace_server/opentelemetry/python_spans.py
```diff
@@ -352,7 +355,16 @@ def to_call(
         else:
             turn_id = None
 
-        wb_run_id = wandb_attributes.get("wb_run_id") or None
+        wb_run_id = wb_run_id or wandb_attributes.get("wb_run_id") or None
+
+        if wb_run_id:
+            if wb_run_id.find(":") == -1:
```
tssweeney: The condition where ":" is not in the wb_run_id implies that it was not correctly converted on the way in. Can you share why you decided to add this condition? As is, it seems like it might be hiding a bug

zbirenbaum: TLDR: 
wandb run ids don't suffer from the same 'internal' vs 'external' difficulties that our other ids do. The ID the client gets is the same as the ID that the server needs (just without the internal project id).

Full Explanation: 

 - external run id looks like this: `{entity}/{project}/{run_id}`
 - internal run id looks like this: `{internal_project_id}:{run_id)`

run_id is the same in both scenarios and we already compute the internal_project_id on the way in

the wb_run_id in clickhouse must be an internal run id, but how we compute does not matter.

When passed in as a header, the adapting API requires an "external" run ID (although we could bypass this by removing the validation)

When set as an attribute, we only have the run ID and use the internal_project_id we already compute to construct the internal representation

```
run = wandb.init(entity, project)
span.set_attribute("wandb.wb_run_id", run.id)
```

This allows a user to manage multiple runs in a single session, while doing the same with headers gets very tricky fast.

The ':' missing scenario means that we detected the run_id from the attributes. If ':' is present, then we got the id from the header and it is already converted on the way in.

Advanced users should prefer setting wb_run_id attribute (not converted on the way in) while quickstart will usually make use of the "wb_run_id" header which is converted.

tssweeney: Thank you for this clear explanation. This makes sense. I think it is worth pointing out that the low level trace server doesn't really "know" anything about W&B or how these ids are constructed. This change blurs these lines and changes that contract. I would recommend that:
1. At the service network interface, `wb_run_id` is always the `<entity>/<project>/<run>`. Regardless of if this is set via header/attributes/sent via python/the service/etc... This keeps the contract consistent.
2. The `trace_server.py` (FastAPI wrapper around the low level ClickhouseTraceServer) should be responsible for doing the conversion of external to internal (and visa versa). Again, this matches all patterns.
3. The `clickhouse_trace_server` defined here should be totally agnostic to the construction and would never need to handle a situation where there is not a project prefix

The reason i started this thread/comment is that I saw #3 was not true in this implementation, which implies something deviated in the interface contracts.

---

Now, I _think_ the convenience you are pointing out is that doing something like `span.set_attribute("wandb.wb_run_id", run.id)` is easier than constructing the fully qualified `<entity>/<project>/<run>`. **Assuming this is your implication:** I think that is a nice convenience. If you want to support that, I would change the `trace_server.py` (in core) to prepend the project components such that the low level trace server always has the same experience.

tssweeney (approved):

PR #5646, opened 2025-10-31
feat(weave): add key for root level attributes to filter on

zbirenbaum:
## Description

keys under wandb.attributes will be populated to the top level

image


gtarpenning (approved):
Epic

PR #5650, opened 2025-10-31
fix(weave): fix DSPy handling of BaseModel based modules and context propagation

ayulockin:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

This PR do the following:
- fix a bug where `serialized_pred` was not initialized to None which if a custom dspy Module was used with, threw not able to access local variable error.
- Handle pydantic BaseModel as potential output return type
- Use the new context propagation capability to tie the program execution trace with evaluation trace's `Model.Predict`.

docs are not required

## Testing

How was this PR tested?

### DIFF
--- a/weave/integrations/dspy/dspy_sdk.py
+++ b/weave/integrations/dspy/dspy_sdk.py
@@ -117,7 +117,11 @@ class DSPyPatcher(MultiPatcher):
                 }
 
                 # prepare dataset for the evaluation logger
-                dataset = [dictify(ex.toDict()) for ex in devset] if devset is not None else []
+                dataset = (
+                    [dictify(ex.toDict()) for ex in devset]
+                    if devset is not None
+                    else []
+                )
 
                 ev = EvaluationLogger(
                     name=f"dspy_eval_{model_name}",


weave/integrations/dspy/dspy_sdk.py
```diff
@@ -112,13 +116,15 @@ def _wrapped_call(
                     "provide_traceback": provide_traceback,
                 }
 
+                # prepare dataset for the evaluation logger
+                dataset = [dictify(ex.inputs()) for ex in devset if devset is not None]
```
andrewtruong: Is this correct?

ayulockin: Nice catch. I referred back to `dspy.Example` and changed `.inputs` to `.toDict`.

andrewtruong: no I mean the `if devset is not None` check -- are you sure that's correct?  I think this will result in a list of a bunch of None values

andrewtruong: Maybe instead you want:
```py
dataset = [dictify(ex.toDict()) for ex in devset] if devset is not None else []
```

ayulockin: yes yes, this is fixed now.

PR #5660, opened 2025-10-31
chore(weave): Simplify op signature

andrewtruong:
Uses `Unpack` to simplify the signature of our ops


neutralino1 (approved):
Sweet! TIL `Unpack`.

PR #5667, opened 2025-11-03
calls complete -- noop SWITCHER (Step 2)

gtarpenning:
## Description

NOOP pr to test the impact of stats query determining which project to switch to. 

## Testing

Should have no impact other than a small latency impact.


gtarpenning:
closing in favor of graphite stack

PR #5675, opened 2025-11-04
chore(weave): OTEL Op optimizations + obj_create_batch

zbirenbaum:
## Description

- Fixes WB-28898

- Modifies otel_export flow to first query all of the objects that would be generated as ops for the span list
- uses query to determine which spans should have ops created and which can be linked to an existing one
- uses placeholder to pre-calculate code digest and fully bypass file creation when a placeholder code file has already been generated for the project
- Adds obj_create_batch helper to trace server and adjusts the existing obj_create response to include the required information to make use of a batched result.

This is an enormous perf bump.

## Testing

Tested using performance script for OTEL:
Before:
 - 100 unique span names:
image

- 300 unique span names times out

After:
 - 100 unique span names:
image

 - 300 unique span names:
image


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -293,34 +294,58 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
             ),
         ).objs
 
+        # FILE BYPASS START
```
gtarpenning, line 296: Instead of `# FILE BYPASS START` we might want to comment what this is actually doing: "Instead of allowing op creation to generate files, for perf manually create our placeholder otel op file and skip file creation when generating ops later" etc.

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning, line 410: hmm, I think we need this

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -293,34 +294,58 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
...
+        if len(existing_objects) == 0:
+            # We don't know if any existing ops have been created
```
neutralino1, line 303: Is this comment correct? Isn't it more like
```suggestion
            # None of the ops already exist, we create one, same for all, since no code capture available for OTel
```
?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -4277,6 +4306,66 @@ def with_new_client(self) -> Iterator[None]:
...
+        obj_results = []
+        ch_insert_batch = []
+        if batch:
+            settings = {}
+            if self._use_async_insert:
+                settings["async_insert"] = 1
+                # https://clickhouse.com/docs/en/optimize/asynchronous-inserts#enabling-asynchronous-inserts
+                # Setting wait_for_async_insert = 0 does not guarantee that insert errors
+                # are caught, reverting to default behavior.
+                settings["wait_for_async_insert"] = 1
```
gtarpenning, line 4336: we dont need this anymore

zbirenbaum: The whole block? Or what specifically do we not need

weave/trace_server/clickhouse_trace_server_batched.py
zbirenbaum: Not sure where that came from, might have been a relic of improper rebase?

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: the highlighted lines: `4325-4330`

weave/trace_server/clickhouse_trace_server_batched.py
zbirenbaum: It's more like
`# None of the ops already exist, we create one source file, same for all, since no code capture available for OTel`
this is clearer though so I'll adjust

zbirenbaum:
Closing in favor of https://github.com/wandb/weave/pull/5689

PR #5676, opened 2025-11-04
feat(weave): implement obj_create_batch

zbirenbaum:
## Description

- Fixes WB-28898
- DOCS-1880

Implements batch object creation on the trace server and associated bindings for access.

## Testing

Adds unit tests

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -747,8 +747,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
     def obj_create_batch(
         self, batch: list[tsi.ObjSchemaForInsert]
     ) -> list[tsi.ObjCreateRes]:
-        """
-        This method is for the special case where all objects are known to use a placeholder.
+        """This method is for the special case where all objects are known to use a placeholder.
         We lose any knowledge of what version the created object is in return for an enormous
         performance increase for operations like OTel ingest.
 


weave/trace_server_bindings/caching_middleware_trace_server.py
```diff
@@ -288,6 +288,15 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
             self._next_trace_server.obj_create, req, tsi.ObjCreateRes
         )
 
+    def obj_create_batch(self, req: tsi.ObjCreateBatchReq) -> tsi.ObjCreateBatchRes:
+        """Pass-through batch create.
+
+        We currently do not add caching for the batch endpoint. Implementing
+        it as a passthrough ensures functional correctness while keeping
+        consistency with single `obj_create` behavior.
+        """
```
gtarpenning: 🔪

tests/trace/test_trace_server.py
```diff
@@ -81,3 +81,100 @@ def test_robust_to_url_sensitive_chars(client):
         )
     )
     assert read_res.vals[0] == bad_val[bad_key]
+
+
+def test_save_object_batch(client):
```
gtarpenning: I think we should add a bunch more tests:
- what happens when you batch upload two objects with the same object_id, different hash? same hash, different object id? identical for both? 
- batch upload 4 versions of the same objects in one batch, then confirm the read path works
- delete an object version for an object with multiple versions after batch uploading, make sure it is returned correctly and the versions are intact
- try to batch upload to different projects in the same batch, we should throw an error

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -736,7 +737,62 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
             column_names=list(ch_obj.model_fields.keys()),
         )
-        return tsi.ObjCreateRes(digest=digest)
+
+        return tsi.ObjCreateRes(
+            digest=digest,
+            object_id=req.obj.object_id,
+        )
+
+    @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched._create_obj_batch")
```
gtarpenning:
```suggestion
    @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched.create_obj_batch")
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -736,7 +737,62 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
+                {
+                    "clickhouse_trace_server_batched._create_obj_batch.count": str(
+                        len(req.batch)
+                    )
+                }
+            )
+
+        obj_results = []
+        ch_insert_batch = []
+        if req.batch:
```
gtarpenning: i realize this is mirroring the other path but we really should be returning early here instead of all the batch handling under the conditional

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -736,7 +737,62 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
+            for obj in req.batch:
```
gtarpenning: we might want to do some de-duping here, we dont need to insert identical objects

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -250,6 +250,15 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
             req.obj.wb_user_id = self._idc.ext_to_int_user_id(req.obj.wb_user_id)
         return self._ref_apply(self._internal_trace_server.obj_create, req)
 
+    def obj_create_batch(self, req: tsi.ObjCreateBatchReq) -> tsi.ObjCreateBatchRes:
+        # Convert project and user IDs for each object in the batch
+        for obj in req.batch:
+            obj.project_id = self._idc.ext_to_int_project_id(obj.project_id)
+            if obj.wb_user_id is not None:
+                obj.wb_user_id = self._idc.ext_to_int_user_id(obj.wb_user_id)
+        # Apply internal method and convert any refs in the response
```
gtarpenning: 🔪

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -250,6 +250,15 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
             req.obj.wb_user_id = self._idc.ext_to_int_user_id(req.obj.wb_user_id)
         return self._ref_apply(self._internal_trace_server.obj_create, req)
 
+    def obj_create_batch(self, req: tsi.ObjCreateBatchReq) -> tsi.ObjCreateBatchRes:
+        # Convert project and user IDs for each object in the batch
```
gtarpenning: 🔪

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -736,7 +737,62 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
+            digest=digest,
+            object_id=req.obj.object_id,
+        )
+
+    @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched._create_obj_batch")
+    def obj_create_batch(self, req: tsi.ObjCreateBatchReq) -> tsi.ObjCreateBatchRes:
+        if root_span := ddtrace.tracer.current_span():
+            root_span.set_tags(
+                {
+                    "clickhouse_trace_server_batched._create_obj_batch.count": str(
```
gtarpenning:
```suggestion
                    "clickhouse_trace_server_batched.create_obj_batch.count": str(
```

tests/trace/test_trace_server.py
zbirenbaum: > what happens when you batch upload two objects with the same object_id, different hash? same hash, different object id? identical for both?

Added tests:
1. Same ID different hash: two objects with different versions
2. Different ID same hash: two objects with different IDs
3. Identical for both: only one object is created

> batch upload 4 versions of the same objects in one batch, then confirm the read path works

I tested this but the versioning is random since. I think this revealed a bug, we have to read the digests from the insert summary rather than going in order.

Should the version_index stay identical or decrement if a previous version was deleted?


> try to batch upload to different projects in the same batch, we should throw an error

It would actually be pretty cool if we allowed this as long as you have write access to all the projects. For now I agree, but I think this should occur in the external trace server. The internal trace server shouldn't really care about validating something like this

tests/trace_server/test_clickhouse_batching.py
```diff
@@ -124,3 +130,234 @@ def test_clickhouse_batching():
...
+    return base64.b64encode(b"test_project").decode()
+
+
+def _internal_wb_user_id() -> str:
+    return base64.b64encode(b"test_user").decode()
+
+
+def test_obj_create_batch_internal_clickhouse_insert_once():
+    """Test internal obj_create_batch performs a single ClickHouse insert."""
+    from weave.trace_server.clickhouse_trace_server_batched import ClickHouseTraceServer
```
gtarpenning: move

tests/trace_server/test_clickhouse_batching.py
```diff
@@ -124,3 +130,234 @@ def test_clickhouse_batching():
...
+        ClickHouseTraceServer, "_mint_client", return_value=mock_ch_client
+    ):
+        server = ClickHouseTraceServer(host="test_host")
+        pid = _internal_pid()
+        wb_user_id = _internal_wb_user_id()
+        batch = [
+            _mk_obj(pid, "obj_a", wb_user_id, {"v": 1}),
+            _mk_obj(pid, "obj_b", wb_user_id, {"v": 2}),
+            _mk_obj(pid, "obj_c", wb_user_id, {"v": 3}),
+        ]
```
gtarpenning: why are we mocking this but later we aren't? lets just not mock if we don't need to

tests/trace_server/test_clickhouse_batching.py
```diff
@@ -124,3 +130,234 @@ def test_clickhouse_batching():
...
+            project_id=pid,
+            filter=tsi.ObjectVersionFilter(object_ids=[obj_id], latest_only=False),
+        )
+    )
+    assert len(res.objs) == 1
+    assert res.objs[0].digest == str_digest(json.dumps(val))
+
+
+def test_obj_batch_four_versions_and_read_path(trace_server):
+    """Batch upload 4 versions and verify reads over all and latest work."""
```
gtarpenning, line 238: probably dont need the first test if we are doing this

tests/trace_server/test_clickhouse_batching.py
```diff
@@ -124,3 +130,234 @@ def test_clickhouse_batching():
...
+    server = trace_server._internal_trace_server
+    pid1 = _internal_pid()
+    wb_user_id = _internal_wb_user_id()
+    pid2 = "cHJvamVjdF8y"  # base64("project_2")
+    batch = [
+        _mk_obj(pid1, "p1_obj", wb_user_id, {"a": 1}),
+        _mk_obj(pid2, "p2_obj", wb_user_id, {"b": 2}),
+    ]
+
+    with pytest.raises(InvalidRequest):
```
gtarpenning: can we check the error message? I could imagine this changing, InvalidRequest is pretty generic

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -736,7 +737,75 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
         )
-        return tsi.ObjCreateRes(digest=digest)
+
+        return tsi.ObjCreateRes(
+            digest=digest,
+            object_id=req.obj.object_id,
+        )
+
+    @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched.create_obj_batch")
+    def obj_create_batch(self, req: tsi.ObjCreateBatchReq) -> tsi.ObjCreateBatchRes:
```
gtarpenning: because this is internal, I think using an interface class doesn't really make sense. Can we just parameterize it? We use the request objects for actual requests but in this case we only use it internally in clickhouse batched right?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -736,7 +737,75 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
+        ch_insert_batch = []
+        if (
+            not req or not req.batch
+        ):  # Note: Mirrors check in obj_create. Would req ever be None here?
+            return tsi.ObjCreateBatchRes(results=[])
+
+        unique_projects = {obj.project_id for obj in req.batch}
+        if len(unique_projects) > 1:
+            # Note(zach): Is this really necessary?
+            # Cross project updates should be fine if external validation is correct
```
gtarpenning: for consistency we do not allow this on any endpoint, I think its a good idea.

weave/trace_server/trace_server_interface.py
```diff
@@ -554,7 +554,16 @@ class ObjCreateReq(BaseModelStrict):
...
+    digest: str
+    object_id: str
+
+
+class ObjCreateBatchReq(BaseModelStrict):
+    batch: list[ObjSchemaForInsert]
+
+
+class ObjCreateBatchRes(BaseModel):
+    results: list[ObjCreateRes]
```
gtarpenning: I think these classes should only be used for endpoint code, at least for now we shouldn't put these here.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -736,7 +737,75 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
+                    )
+                }
+            )
+
+        obj_results = []
+        ch_insert_batch = []
+        if (
+            not req or not req.batch
+        ):  # Note: Mirrors check in obj_create. Would req ever be None here?
+            return tsi.ObjCreateBatchRes(results=[])
```
gtarpenning: if we parameterize then we can remove this.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -736,7 +737,74 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
+                    "clickhouse_trace_server_batched.create_obj_batch.count": str(
+                        len(batch)
+                    )
+                }
+            )
+
+        obj_results = []
+        ch_insert_batch = []
+        if not batch:
+            # Note: Mirrors check in obj_create. Would req ever be None here?
```
gtarpenning: 🔪

PR #5678, opened 2025-11-04
refactor(weave): no-op refactor to improve calls query organization

gtarpenning:
[WB-28919](https://wandb.atlassian.net/browse/WB-28919)

## Description

Refactor the calls query builder and moves tests together. This pr: 
- groups CTE handling (costs, id filter, object ref handling)
- groups table joins
- groups calls query WHERE filter handling
- moves the tests into their own folder

## Testing

existing test coverage

and dev: 

image

[WB-28919]: https://wandb.atlassian.net/browse/WB-28919?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -54,29 +54,29 @@ Focus on these primary test shards:
 #### Basic Test Commands
 
 1. Run all tests in a specific shard: `nox --no-install -e "tests-3.12(shard='trace')"`
-2. Run a specific test by appending `-- [test]` like so: `nox --no-install -e "tests-3.12(shard='trace')" -- trace/test_client_trace.py::test_simple_op`
+2. Run a specific test by appending `-- [test]` like so: `nox --no-install -e "tests-3.12(shard='trace')" -- tests/trace/test_client_trace.py::test_simple_op`
 3. Run linting: `nox --no-install -e lint` (Note: This will modify files)
 
 _Important:_ Since you don't have internet access, you must run `nox` with `--no-install`. We have pre-installed the requirements on the above shards.
 
 #### Critical Path Information
 
-**Test paths must be relative to the `tests/` directory**, not the repository root. When nox runs, it automatically changes to the `tests/` directory before executing pytest.
+**Test paths must be relative to the repository root**, not the `tests/` directory.
 
 Examples:
-- ✅ CORRECT: `-- trace/test_dataset.py::test_basic_dataset_lifecycle`
-- ❌ WRONG: `-- tests/trace/test_dataset.py::test_basic_dataset_lifecycle`
+- ✅ CORRECT: `-- tests/trace/test_dataset.py::test_basic_dataset_lifecycle`
+- ❌ WRONG: `-- trace/test_dataset.py::test_basic_dataset_lifecycle`
 
 #### Backend Selection
 
 **SQLite (Default/Recommended for Development):**
 ```bash
-nox --no-install -e "tests-3.12(shard='trace')" -- trace/test_client_trace.py::test_simple_op --trace-server=sqlite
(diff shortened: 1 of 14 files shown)


jwlee64 (approved):
The cost cte refactors look good to me.
Cleaner overall, thank you!

Approving from my end, though a more in depth look at the calls query builder might be warranted

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -737,131 +809,107 @@ def as_sql(self, pb: ParamBuilder, table_alias: str = "calls_merged") -> str:
...
-        needs_feedback = False
-        select_fields_sql = ", ".join(
-            field.as_select_sql(pb, table_alias) for field in self.select_fields
+        return select_query._as_sql_base_format(
+            pb,
+            table_alias,
+            id_subquery_name="filtered_calls",
+            field_to_object_join_alias_map=field_to_object_join_alias_map,
+            expand_columns=self.expand_columns,
         )
```
gtarpenning: this could be cleaner 🤔

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -720,9 +787,14 @@ def as_sql(self, pb: ParamBuilder, table_alias: str = "calls_merged") -> str:
...
         # Build CTEs for object reference filtering and ordering
-        object_join_cte, field_to_object_join_alias_map = build_object_ref_ctes(
+        object_ref_cte_list, field_to_object_join_alias_map = build_object_ref_ctes(
             pb, self.project_id, object_ref_conditions
         )
+        ctes = CTECollection()
+
+        # Add object reference CTEs to collection
+        for cte in object_ref_cte_list:
+            ctes.add_cte(cte.name, cte.sql)
```
neutralino1: Why doesn't `build_object_ref_ctes` already return a `CTECollection`?

weave/trace_server/calls_query_builder/cte.py
```diff
@@ -0,0 +1,42 @@
...
+
+    name: str
+    sql: str
+
+    def to_sql(self) -> str:
+        """Format this CTE for inclusion in a WITH clause."""
+        return f"{self.name} AS ({self.sql})"
+
+
+class CTECollection(BaseModel):
```
neutralino1, line 17: Thoughts on inheriting this from `list[CTE]` directly?

weave/trace_server/calls_query_builder/object_ref_query_builder.py
```diff
@@ -686,7 +687,7 @@ def build_object_ref_ctes(
         )
 
     project_param = pb.add_param(project_id)
-    cte_parts = []
+    cte_list: list[CTE] = []
```
neutralino1:
```suggestion
    cte_list = CTECollection([])
```

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: good call

weave/trace_server/calls_query_builder/cte.py
gtarpenning: Yeah let me try that

gtarpenning: I think this will end up being less straightforward, requiring consumers of the class to cast to CTE and .append, rather than `add_cte` which I like...

neutralino1 (approved):
Nice work!

PR #5682, opened 2025-11-04
chore(weave): fix llm completions test ch batched

gtarpenning:
## Description

another one, i think this is right? 

## Testing

test


jwlee64 (approved):
wow ty, i fixed like none of these tests

PR #5683, opened 2025-11-05
fix(weave): run all trace server tests in CI

gtarpenning:
## Description

This probably needs to be tuned, but currently we are deselecting 230 tests, mostly errantly. 

## Testing

this prs ci runs

### DIFF
--- a/tests/trace_server/conftest.py
+++ b/tests/trace_server/conftest.py
@@ -55,10 +55,7 @@ def pytest_collection_modifyitems(config, items):
     # 2. All tests that use the trace_server fixture (for tests outside this directory)
     for item in items:
         # Check if the test is in the trace_server directory by checking parent directories
-        test_path = item.path if hasattr(item, "path") else item.fspath
-        is_in_trace_server_dir = "trace_server" in test_path.parts
-
-        if is_in_trace_server_dir:
+        if "trace_server" in item.path.parts:
             item.add_marker(pytest.mark.trace_server)
         # Also mark tests that use the trace_server fixture (for tests outside this dir)
         elif "trace_server" in item.fixturenames:


tests/trace_server/conftest.py
```diff
@@ -50,9 +50,16 @@ def pytest_addoption(parser):
...
 def pytest_collection_modifyitems(config, items):
-    # Add the trace_server marker to all tests that have a client fixture
+    # Add the trace_server marker to:
+    # 1. All tests in the trace_server directory (regardless of fixture usage)
+    # 2. All tests that use the trace_server fixture (for tests outside this directory)
     for item in items:
-        if "trace_server" in item.fixturenames:
+        # Check if the test is in the trace_server directory
+        test_file_path = str(item.fspath)
+        if "tests/trace_server" in test_file_path or "/trace_server/" in test_file_path:
```
andrewtruong: Why both `tests/trace_server` and `/trace_server/`?  Intuitively this seems wrong

tests/trace_server/conftest.py
```diff
@@ -50,9 +50,16 @@ def pytest_addoption(parser):
...
 
 def pytest_collection_modifyitems(config, items):
-    # Add the trace_server marker to all tests that have a client fixture
+    # Add the trace_server marker to:
+    # 1. All tests in the trace_server directory (regardless of fixture usage)
+    # 2. All tests that use the trace_server fixture (for tests outside this directory)
     for item in items:
-        if "trace_server" in item.fixturenames:
+        # Check if the test is in the trace_server directory
+        test_file_path = str(item.fspath)
```
andrewtruong: Just use an actual path, `fspath` is deprecated

tests/trace_server/conftest.py
gtarpenning: fixed

andrewtruong:
you might need to pick your other changes into this branch

tests/trace_server/conftest.py
```diff
@@ -50,9 +50,18 @@ def pytest_addoption(parser):
...
 
 def pytest_collection_modifyitems(config, items):
-    # Add the trace_server marker to all tests that have a client fixture
+    # Add the trace_server marker to:
+    # 1. All tests in the trace_server directory (regardless of fixture usage)
+    # 2. All tests that use the trace_server fixture (for tests outside this directory)
     for item in items:
-        if "trace_server" in item.fixturenames:
+        # Check if the test is in the trace_server directory by checking parent directories
+        test_path = item.path if hasattr(item, "path") else item.fspath
```
andrewtruong: Why not just `item.path`?  Do you really need `fspath`?

gtarpenning: pytest uses is, its typed that way... I can remove it but it might not work in some envs...

gtarpenning: okay fine they deprecated it a year ago you are right

PR #5686, opened 2025-11-05
chore(weave): content test fix for length incraese

gtarpenning:
## Description

Adjust the test to use the actual content conversion min length

### DIFF
--- a/tests/trace_server/test_base64_content_conversion.py
+++ b/tests/trace_server/test_base64_content_conversion.py
@@ -22,6 +22,9 @@ from weave.trace_server.trace_server_interface import (
 )
 from weave.type_wrappers.Content.content import Content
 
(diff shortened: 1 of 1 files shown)


tests/trace_server/test_base64_content_conversion.py
```diff
@@ -105,7 +106,8 @@ def test_replace_data_uri_in_dict_only(self):
             ]
         )
 
-        test_data = b"a" * 100
+        # Use data larger than AUTO_CONVERSION_MIN_SIZE to trigger conversion
+        test_data = b"a" * (AUTO_CONVERSION_MIN_SIZE + 10)
```
andrewtruong: nit: would extract `AUTO_CONVERSION_MIN_SIZE+10` to const and avoid comments

PR #5689, opened 2025-11-05
feat(weave): create ops for OTEL spans

zbirenbaum:
## Description

- Fixes WB-26557
- Fixes WB-25891

Creates OPs for OTel spans enabling all features which work with normal OPs such as datasets, evals, and trace selection. Docs are not required

## Testing

heavily tested and performance benchmarked

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -228,7 +228,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         return self._kafka_producer
 
     def _get_existing_ops_from_spans(
-        self, seen_ids: set[str], project_id: str, limit=None
+        self, seen_ids: set[str], project_id: str, limit: Optional[int] = None
     ) -> list[tsi.ObjSchema]:
         obj_version_filter = tsi.ObjectVersionFilter(
             object_ids=list(seen_ids),
@@ -245,7 +245,9 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
             ),
         ).objs
 
-    def _create_or_get_placeholder_ops_digest(self, project_id: str, create: bool) -> str:
+    def _create_or_get_placeholder_ops_digest(
+        self, project_id: str, create: bool
+    ) -> str:
         if not create:
             return bytes_digest(
                 (object_creation_utils.PLACEHOLDER_OP_SOURCE).encode("utf-8")


neutralino1 (approved):
Great work, thank you!

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -259,21 +264,107 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
...
+        )
+
+        existing_objects = self.objs_query(
+            tsi.ObjQueryReq(
+                project_id=req.project_id,
+                filter=obj_version_filter,
+                metadata_only=True,
+                limit=limit,
+            ),
+        ).objs
```
gtarpenning: this can all be a helper like `_get_existing_ops_from_spans`

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -259,21 +264,107 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
...
+            source_file_req = tsi.FileCreateReq(
+                project_id=req.project_id,
+                name=object_creation_utils.OP_SOURCE_FILE_NAME,
+                content=source_code.encode("utf-8"),
+            )
+            digest = self.file_create(source_file_req).digest
+        else:
+            digest = bytes_digest(
+                (object_creation_utils.PLACEHOLDER_OP_SOURCE).encode("utf-8")
+            )
```
gtarpenning: this can also be a helper like `_build_placeholder_op_digest` or something

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -259,21 +264,107 @@ def otel_export(self, req: tsi.OtelExportReq) -> tsi.OtelExportRes:
...
+        res = self.obj_create_batch(obj_creation_batch)
+
+        for result in res:
+            op_ref_uri = ri.InternalOpRef(
+                project_id=req.project_id,
+                name=result.object_id,
+                version=result.digest,
+            ).uri()
+            for idx in obj_id_idx_map[result.object_id]:
+                calls[idx][0].op_name = op_ref_uri
```
gtarpenning, line 364: ideally this would also be a helper, although its a bit more annoying because you are updating in place. still possible obviously

weave/trace_server/clickhouse_trace_server_batched.py
zbirenbaum: Like an inline function? I think that just adds lines of code

weave/trace_server/clickhouse_trace_server_batched.py
zbirenbaum: Yeah I think putting this in a helper will obscure a very important piece. I updated the other two though

PR #5692, opened 2025-11-06
chore(weave): Simplify Stainless codegen tooling

andrewtruong:
Simplifies the `generate` command


tools/codegen/generate.py
```diff
@@ -1,915 +1,396 @@
 """Weave SDK Code Generator.
```
tssweeney, line 1: Any idea what this notation is? `100755 -> 100644`

Image

PR #5694, opened 2025-11-06
chore(weave): analytics on batch start/end breakdown

gtarpenning:
## Description

How many call starts end up in batches without call ends? potentially critical info for calls v2. 

## Testing

Even for larger batches (100 calls) this adds only 100 microseconds of latency
image


gtarpenning:
@neutralino1  > Any SDK performance impact from doing this?

impact is as long as it takes python to iterate through a list, microseconds probably

neutralino1:
Network calls too no? `set_tag`?

neutralino1 (approved):

PR #5697, opened 2025-11-07
chore(weave): Update default names for V2 objects

andrewtruong:
Updates the default names for V2 objects.

Previously, it would show as `datasets_{safe_name}`, but now it will either be `Dataset` or `{safe_name}`

### DIFF
--- a/weave/trace_server/object_creation_utils.py
+++ b/weave/trace_server/object_creation_utils.py
@@ -91,113 +91,26 @@ def make_safe_name(name: str | None) -> str:
     return name.replace(" ", "_").replace("/", "_").lower()
 
 
-def make_dataset_id(name: str | None) -> str:
-    """Create a dataset object ID from a dataset name.
+def make_object_id(name: str | None, default: str) -> str:
+    """Create an object ID from a name with a default fallback.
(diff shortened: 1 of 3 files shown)


weave/trace_server/object_creation_utils.py
```diff
@@ -91,6 +91,116 @@ def make_safe_name(name: str | None) -> str:
...
+        'Op'
+        >>> make_op_id("My Op")
+        'my_op'
+        >>> make_op_id("user/op")
+        'user_op'
+    """
+    if name is None:
+        return "Op"
+    return make_safe_name(name)
+
```
neutralino1, line 115: You could DRY this whole module with
```python
def _make_safe_name_or_default(name: str | None, default: str) -> str:
    if name is None:
        return default

    return make_safe_name(name)
```

neutralino1 (commented):
One comment.

weave/trace_server/object_creation_utils.py
andrewtruong: fixed

neutralino1 (approved):

PR #5699, opened 2025-11-07
chore(weave): Add StainlessRemoteHTTPTraceServer as opt-in backend

andrewtruong:
Adds a `RemoteHTTPTraceServer` backend using the Stainless client that users can opt in to using the `use_stainless_server` setting or `WEAVE_USE_STAINLESS_SERVER` env var


weave/trace_server/trace_server_interface.py
```diff
@@ -2177,3 +2180,7 @@ class FullTraceServerInterface(TraceServerInterface, TraceServerInterfaceV2, Pro
     """
 
     pass
+
+
+class FullTraceServerWithInfo(FullTraceServerInterface, Protocol):
```
tssweeney: I would move this definition to the bindings and call it "TraceServerClientInterface" or something to start clearly separating concerns

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -0,0 +1,2065 @@
...
+        username = None
+        password = None
+        if auth:
+            username, password = auth
+
+        default_headers = dict(extra_headers) if extra_headers else {}
+        if retry_id := get_current_retry_id():
+            default_headers["X-Weave-Retry-Id"] = retry_id
+
+        self._client = WeaveTraceClient(
```
tssweeney: can we use something other than "_client"? feels very overloaded

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -0,0 +1,2065 @@
+import datetime
```
tssweeney, line 3: it seems like a huge amount of this is copy/pasta. did you consider creating a more reusable adapter pattern?

tssweeney: i am seeing now that this is not really true. leaving comment, but i see there are differences

weave/trace_server_bindings/stainless_remote_http_trace_server.py
andrewtruong: renamed to `_stainless_client`

weave/trace_server/trace_server_interface.py
andrewtruong: done

weave/trace/weave_init.py
```diff
@@ -230,8 +229,20 @@ def init_weave_disabled() -> weave_client.WeaveClient:
...
+    if should_use_stainless_server():
+        from weave.trace_server_bindings.stainless_remote_http_trace_server import (
+            StainlessRemoteHTTPTraceServer,
+        )
+
+        res = StainlessRemoteHTTPTraceServer.from_env(should_batch)
+    else:
+        from weave.trace_server_bindings.remote_http_trace_server import (
+            RemoteHTTPTraceServer,
+        )
```
neutralino1: Any reason not to import those at the top?

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -0,0 +1,2064 @@
...
+        """
+        if isinstance(req, dict):
+            req = tsi.ObjCreateReq.model_validate(req)
+
+        req = cast(tsi.ObjCreateReq, req)
+
+        self._update_client_headers()
+        req_dict = req.model_dump(by_alias=True)
+        response = self._stainless_client.objects.create(**req_dict)
+        return tsi.ObjCreateRes.model_validate(response.model_dump())
```
neutralino1: To @tssweeney 's point, you could DRY a lot of this with
```python
    def _stainless_request(
        self,
        req: T | dict[str, Any],
        req_type: T,
        res_type: BaseModel,
        stainless_api: Callable
    ):
        if isinstance(req, dict):
            req = req_type.model_validate(req)

        req = cast(req_type, req)

        self._update_client_headers()
        req_dict = req.model_dump(by_alias=True)
        response = stainless_api(**req_dict)
        return res_type.model_validate(response.model_dump())


    def obj_create(self,
        req: req: Union[tsi.ObjCreateReq, dict[str, Any]
    ) -> tsi.ObjCreateRes:
        return self._stainless_request(
            req,
            tsi.ObjCreateReq,
            tsi.ObjCreateRes,
            self._stainless_client.objects.create
        )
```

weave/trace/weave_init.py
andrewtruong: If the user does not have the stainless SDK installed, this will crash.  We defer import here to prevent that crash

tssweeney: this is probably true for `weave.trace_server_bindings.stainless_remote_http_trace_server`, but not the RemoteHTTPTraceServer

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -0,0 +1,2045 @@
...
+    ServerInfoRes,
+    StartBatchItem,
+)
+from weave.utils.retry import get_current_retry_id
+from weave.wandb_interface import project_creator
+
+logger = logging.getLogger(__name__)
+
+
+class StainlessRemoteHTTPTraceServer(tsi.FullTraceServerInterface):
```
tssweeney: isn't this supposed to inherit from the new Client interface?

weave/trace_server_bindings/client_interface.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class TraceServerClientInterface(FullTraceServerInterface, Protocol):
+    """Interface for trace server client implementations that support server_info.
+
+    This protocol extends FullTraceServerInterface to include the server_info method,
+    which allows clients to query server capabilities and version information.
+    """
+
+    def server_info(self) -> "ServerInfoRes": ...
```
tssweeney: seems like there are methods like `from_env`, `set_auth`, etc that you are explicitly defining in children classes which you would want to put here

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -0,0 +1,2045 @@
...
+        Args:
+            req: Scorer list V2 request.
+
+        Yields:
+            ScorerReadV2Res instances.
+        """
+        if isinstance(req, dict):
+            req = tsi.ScorerListV2Req.model_validate(req)
+        req = cast(tsi.ScorerListV2Req, req)
+        entity, project = req.project_id.split("/", 1)
```
tssweeney: why are we doing this splitting everywhere and then untyped kwargs. seems messy

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -0,0 +1,2045 @@
...
+        """
+        return self._stainless_request_v2(
+            req,
+            tsi.DatasetCreateV2Req,
+            tsi.DatasetCreateV2Res,
+            self._stainless_client.v2.datasets.create,
+            exclude={"project_id"},
+        )
+
+    def dataset_read_v2(
```
tssweeney: I really don't like all this "v2" terminology slipping into apis - it is messy and overly tied to implementation details

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -0,0 +1,2045 @@
...
+        if exclude:
+            exclude_set.update(exclude)
+        if exclude_set:
+            dump_kwargs["exclude"] = exclude_set
+
+        req_dict = req.model_dump(**dump_kwargs)
+        response = stainless_api(**req_dict, **extra_kwargs)
+        return res_type.model_validate(response.model_dump())
+
+    def _stainless_request_v2(
```
tssweeney: the only difference here is the project id splitting? seems like it could be much more dry

weave/trace_server_bindings/stainless_remote_http_trace_server.py
andrewtruong: Resolved in another PR

weave/trace_server_bindings/stainless_remote_http_trace_server.py
andrewtruong: Now just calling the methods directly

weave/trace_server_bindings/client_interface.py
andrewtruong: Added the relevant methods

weave/trace_server_bindings/stainless_remote_http_trace_server.py
andrewtruong: Yep

tests/trace_server/conftest.py
```diff
@@ -25,7 +25,7 @@ def pytest_addoption(parser):
             "--trace-server",
             action="store",
             default="clickhouse",
-            help="Specify the client object to use: sqlite or clickhouse",
+            help="Specify the client object to use: sqlite, clickhouse, or stainless",
```
tssweeney: I don't think this is controlling the correct thing. This flag changes the server engine. However, you have used it in the `remote_http_trace_server` fixture which is used in the `trace_server` fixture which was added recently and overloads the existing `trace_server` fixture. I think we should:
1. Rename the fixtures so they are not in conflict
2. Don't re-use this flag - instead create a new one that doesn't overload existing stuff

noxfile.py
```diff
@@ -190,6 +194,7 @@ def tests(session, shard):
         "verifiers_test": ["tests/integrations/verifiers/"],
         "trace": ["tests/trace/"],
         "trace_no_server": ["tests/trace/"],
+        "stainless": ["tests/trace_server_bindings/"],
```
tssweeney: I think we are mixing sharding with matrixing. "stainless" should be a matrix option rather than a new shard

andrewtruong: I have it as a shard to isolate from other shards that do not require the stainless sdk installed

tssweeney (approved):

PR #5701, opened 2025-11-07
chore(weave): calls_complete migration file

gtarpenning:
## Description

[WB-29382](https://wandb.atlassian.net/browse/WB-29382%5D(https://wandb.atlassian.net/browse/WB-29382))

Adds migrations for the new `calls_complete` table

Just the migrations themselves should have zero impact on the application.

Lets go crazy with skip indices, its super easy to drop them if we don't derive benefit from them or they get too large. That being said, I have carefully scoped them to their intended use case and data type so I don't think this should be an issue.

## Testing

This query will hit every skip index:

```
EXPLAIN indexes=1
SELECT *
FROM calls_complete
WHERE
  started_at >= toDateTime('2025-11-01 00:00:00')
  AND ended_at >= toDateTime('2025-11-01 00:00:00')
  AND trace_id  IN ('tr_1','tr_2')
  AND parent_id IN ('par_9')
  AND wb_run_id IN ('run_42')
  AND thread_id = 'thread_007'
  AND inputs_dump LIKE '%Explain%'
  AND output_dump LIKE '%success%'
  AND op_name LIKE '%gen%'
```

[WB-29382]: https://wandb.atlassian.net/browse/WB-29382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5702, opened 2025-11-07
chore(weave): calls complete only project switcher

gtarpenning:
## Description

[WB-29383](https://wandb.atlassian.net/browse/WB-29383)

Adds project switching logic to determine when a project should be served from the old table (`calls_merged`​) or the new table `calls_complete`​. Because the write path is not deployed, there should be zero cases where we switch to the new project. Furthermore, because we don't even have the read path, this pr adds **noop** project version queries at key places in the query layer where we will need to do this in the future. There are no places that actually condition on this information, it is entirely in place to determine the performance cost.

## Testing

Adds quite a few tests checking the core functionality of both the clickhouse source of truth implementation and the in memory repr.

QA:
image

Screenshot 2025-12-01 at 1 01 20 PM

[WB-29383]: https://wandb.atlassian.net/browse/WB-29383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -195,15 +195,40 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         self._user = user
         self._password = password
         self._database = database
-        self._flush_immediately = True
-        self._call_batch: list[list[Any]] = []
-        self._file_batch: list[FileChunkCreateCHInsertable] = []
         self._use_async_insert = use_async_insert
         self._model_to_provider_info_map = read_model_to_provider_info_map()
         self._file_storage_client: FileStorageClient | None = None
         self._kafka_producer: KafkaProducer | None = None
         self._evaluate_model_dispatcher = evaluate_model_dispatcher
         self._table_routing_resolver: TableRoutingResolver | None = None
+
+    @property
+    def _flush_immediately(self) -> bool:
+        return getattr(self._thread_local, "flush_immediately", True)
+
+    @_flush_immediately.setter
+    def _flush_immediately(self, value: bool) -> None:
+        self._thread_local.flush_immediately = value
+
+    @property
+    def _call_batch(self) -> list[list[Any]]:
+        if not hasattr(self._thread_local, "call_batch"):
+            self._thread_local.call_batch = []
+        return self._thread_local.call_batch
+
+    @_call_batch.setter
+    def _call_batch(self, value: list[list[Any]]) -> None:
+        self._thread_local.call_batch = value
+
+    @property
+    def _file_batch(self) -> list[FileChunkCreateCHInsertable]:
+        if not hasattr(self._thread_local, "file_batch"):
+            self._thread_local.file_batch = []
+        return self._thread_local.file_batch
+
+    @_file_batch.setter
+    def _file_batch(self, value: list[FileChunkCreateCHInsertable]) -> None:
+        self._thread_local.file_batch = value
 
     @classmethod
     def from_env(
--- a/weave/trace_server/project_version/project_version.py
+++ b/weave/trace_server/project_version/project_version.py
@@ -1,4 +1,5 @@
 import logging
+import threading
 
 import ddtrace
 from cachetools import LRUCache
@@ -22,6 +23,7 @@ PROJECT_RESIDENCE_CACHE_SIZE = 10_000
 _project_residence_cache: LRUCache[str, ProjectDataResidence] = LRUCache(
     maxsize=PROJECT_RESIDENCE_CACHE_SIZE
 )
+_project_residence_cache_lock = threading.Lock()
 
 
 class TableRoutingResolver:
@@ -36,7 +38,8 @@ class TableRoutingResolver:
     def _get_residence(
         self, project_id: str, ch_client: CHClient
     ) -> ProjectDataResidence:
-        cached = _project_residence_cache.get(project_id)
+        with _project_residence_cache_lock:
+            cached = _project_residence_cache.get(project_id)
         if cached is not None:
             return cached
 
@@ -44,7 +47,8 @@ class TableRoutingResolver:
 
         # Don't cache if project is empty, we could write to either table.
         if residence != ProjectDataResidence.EMPTY:
-            _project_residence_cache[project_id] = residence
+            with _project_residence_cache_lock:
+                _project_residence_cache[project_id] = residence
 
         # TODO: remove me, this is temporary to guage cache size impact
         if root_span := ddtrace.tracer.current_root_span():
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -73,6 +73,23 @@ def reset_serializer_load_refs():
 
 
 @pytest.fixture(autouse=True)
+def reset_project_residence_cache():
+    """Reset the project residence cache between tests.
+
+    The project residence cache stores the residence (merged/complete/both) of a project's data.
+    This needs to be cleared between tests to prevent state leakage, especially when tests
+    create projects with the same name but different data characteristics.
+    """
+    from weave.trace_server.project_version.project_version import (
+        _project_residence_cache,
+    )
+
+    _project_residence_cache.clear()
+    yield
+    _project_residence_cache.clear()
+
+
+@pytest.fixture(autouse=True)
 def disable_datadog():
     """Disables Datadog logging and tracing for tests.
 


gtarpenning:
* **#5704** <a href="https://app.graphite.com/github/pr/wandb/weave/5704?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#5724](https://github.com/wandb/weave/pull/5724) <a href="https://app.graphite.com/github/pr/wandb/weave/5724?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#5725](https://github.com/wandb/weave/pull/5725) <a href="https://app.graphite.com/github/pr/wandb/weave/5725?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#5703** <a href="https://app.graphite.com/github/pr/wandb/weave/5703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5702** <a href="https://app.graphite.com/github/pr/wandb/weave/5702?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/5702?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/project_version/providers/clickhouse_provider.py
```diff
@@ -0,0 +1,91 @@
...
+            This allows each thread to get its own thread-local client.
+
+    Examples:
+        >>> ch_provider = ClickHouseProjectVersionProvider(ch_client_factory=lambda: get_ch_client())
+        >>> version = ch_provider.get_project_version_sync("proj-123")
+        >>> assert version in (ProjectVersion.CALLS_MERGED_VERSION, ProjectVersion.CALLS_COMPLETE_VERSION, ProjectVersion.EMPTY_PROJECT)
+    """
+
+    def __init__(self, ch_client_factory: Callable[[], Any]):
+        self._get_ch_client = ch_client_factory
```
neutralino1: Why not pass the client instance directly?

weave/trace_server/project_version/providers/clickhouse_provider.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @ddtrace.tracer.wrap(name="clickhouse_project_version_provider.get_project_version")
+    def get_project_version_sync(
```
neutralino1: Since this class has only one method, and init only expects one argument, why not make this simply a function?
```python
def get_project_calls_table_version_sync(project_id: str, is_write: bool = False) -> CallsTableVersion:
    ch_client = get_ch_client()
    ...
```

weave/trace_server/project_version/providers/memory_cache_provider.py
neutralino1, line 1: why call this module a provider?

weave/trace_server/project_version/providers/memory_cache_provider.py
```diff
@@ -0,0 +1,69 @@
...
+from cachetools import LRUCache
+
+from weave.trace_server.project_version.types import ProjectVersion
+
+logger = logging.getLogger(__name__)
+
+PER_REPLICA_CACHE_SIZE = 10_000
+
+
+class InMemoryCache:
```
neutralino1: This a wrapper with essentially zero added business logic. Is this really necessary? Why not simply instantiate an `LRUCache` at the call site?

weave/trace_server/project_version/project_version.py
```diff
@@ -0,0 +1,184 @@
...
+        >>> resolver = ProjectVersionResolver(ch_client_factory=lambda: get_ch_client())
+        >>> version = await resolver.get_project_version_async("my-project")
+    """
+
+    def __init__(
+        self,
+        ch_client_factory: Callable[[], Any],
+        cache_size: int = PER_REPLICA_CACHE_SIZE,
+    ):
+        self._cache = InMemoryCache(maxsize=cache_size)
```
neutralino1:
```suggestion
        self._cache = LRUCache(maxsize=cache_size)
```

weave/trace_server/project_version/project_version.py
```diff
@@ -0,0 +1,184 @@
...
+            ProjectVersion enum indicating which table to use.
+        """
+        if self._mode == ProjectVersionMode.OFF:
+            return ProjectVersion.CALLS_MERGED_VERSION
+
+        if self._mode == ProjectVersionMode.CALLS_MERGED_READ and not is_write:
+            return ProjectVersion.CALLS_MERGED_VERSION
+
+        version = self._resolve_version_sync(project_id)
+        return self._apply_mode(version, is_write)
```
neutralino1: Why not skip the CH calls if `self._mode == ProjectVersionMode.CALLS_MERGED` (what `_apply_mode` does)?
In this case the output of the CH calls are disregarded.

weave/trace_server/project_version/project_version.py
```diff
@@ -0,0 +1,184 @@
...
+        self, resolved_version: ProjectVersion, is_write: bool
+    ) -> ProjectVersion:
+        """Apply mode-specific overrides to resolved version.
+
+        Modes that skip DB queries return early in get_project_version_sync.
+        This method only handles modes that override the resolved result.
+        """
+        if self._mode == ProjectVersionMode.CALLS_MERGED:
+            return ProjectVersion.CALLS_MERGED_VERSION
+        return resolved_version
```
neutralino1: As mentioned below, this seems unnecessary. Simply incorporate this condition into the resolution logic alongside the other conditions on `self._mode`.

weave/trace_server/project_version/types.py
```diff
@@ -0,0 +1,96 @@
...
+            return cls(mode_str)
+        except ValueError:
+            logger.warning(
+                f"Invalid PROJECT_VERSION_MODE '{mode_str}', defaulting to 'auto'. "
+                f"Valid options: {', '.join([m.value for m in cls])}"
+            )
+            return cls.AUTO
+
+
+class Provider(Protocol):
```
neutralino1: Overkill.

neutralino1 (commented):
This seems overly complex. A few notes.

weave/trace_server/project_version/providers/clickhouse_provider.py
gtarpenning: Client is reinstantiated for each request, we can't persist a client here. The cache must persist across requests, but the ch client must be minted each time.

weave/trace_server/project_version/providers/memory_cache_provider.py
gtarpenning: Hrm...

weave/trace_server/project_version/providers/memory_cache_provider.py
gtarpenning: Fair criticism, the reasons were to allow extensibility in the future, like overloading a base class and using Redis, and for ease of testing. But I'm happy to make it simpler.

weave/trace_server/project_version/project_version.py
gtarpenning: The modes here are subtle, I think in this case we still want to query the database so that we know the added latency of the query, but never actually want to use the result. When OFF, we short circuit completely. That being said, it's clearly confusing, so I will refactor.

weave/trace_server/project_version/types.py
gtarpenning: But but but oop! Yeah okay

weave/trace_server/project_version/clickhouse_project_version.py
```diff
@@ -0,0 +1,73 @@
...
+
+        root_span = ddtrace.tracer.current_root_span()
+        if root_span:
+            root_span.set_tags({"has_complete": has_complete, "has_merged": has_merged})
+
+        if has_complete and has_merged:
+            logger.warning(
+                "Project has traces in both calls_complete and calls_merged tables.",
+                extra={"project_id": project_id},
+            )
```
tssweeney: super nit: I would recommend returning here explicitly rather than waiting for the next `if has_merged` branch - as you are basically saying: if there is both, defer to merged

weave/trace_server/project_version/types.py
```diff
@@ -0,0 +1,69 @@
...
+    """New schema using calls_complete table."""
+    CALLS_COMPLETE_VERSION = 1
+
+
+class ProjectVersionMode(str, Enum):
+    """Modes for controlling project version resolution behavior.
+
+    AUTO: Default behavior - uses table to determine project version.
+    CALLS_MERGED: Forces all reads/writes to calls_merged table (queries DB for perf measurement).
+    CALLS_MERGED_READ: Forces reads to calls_merged table, writes use determined version.
```
tssweeney: under what conditions do we want `CALLS_MERGED_READ`?

weave/trace_server/project_version/project_version.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Args:
+            ch_client_factory: Callable that returns a ClickHouse client.
+            cache_size: Size of the in-memory cache (defaults to 10,000).
+
+        Returns:
+            The global singleton ProjectVersionResolver instance.
+        """
+        global _global_resolver
+        with _global_resolver_lock:
```
tssweeney: sort of odd but worth noting that calling this multiple times with diff params will always select the first param set

gtarpenning: oh, true, we should just be using the class properties.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -231,6 +236,22 @@ def kafka_producer(self) -> KafkaProducer:
...
+    @property
+    def project_version_resolver(self) -> ProjectVersionResolver:
+        if self._project_version_resolver is not None:
+            return self._project_version_resolver
+
+        init_resolver(ch_client_factory=lambda: self.ch_client)
+        self._project_version_resolver = ProjectVersionResolver.get_instance()
+        return self._project_version_resolver
+
+    def _noop_project_version_latency_test(self, project_id: str) -> None:
```
tssweeney, line 270: you should add a DD trace to this so that you can measure latency

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -4676,6 +4705,10 @@ def _insert_call(self, ch_call: CallCHInsertable) -> None:
     @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched._flush_calls")
     def _flush_calls(self) -> None:
         self._analyze_call_batch_breakdown()
+        if len(self._call_batch) > 0:
+            project_id_idx = ALL_CALL_INSERT_COLUMNS.index("project_id")
+            project_id = self._call_batch[0][project_id_idx]
```
tssweeney, line 4734: while true in practice, there is nothing forcing all calls to have the same project id - I would consider explicitly checking the set

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: the underlying `get_project_version_sync` is traced, which is technically the actual function we want to test the latency of :)

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: This is implemented in less of a hacky way in the actual write path pr.

weave/trace_server/project_version/types.py
```diff
@@ -0,0 +1,67 @@
+"""Project version types for routing between calls tables."""
+
+import logging
+import os
+from enum import Enum
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_SERVER_MODE = "force_legacy"
```
tssweeney, line 9: nit: should be `CallsStorageServerMode.FORCE_LEGACY`

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -231,6 +235,16 @@ def kafka_producer(self) -> KafkaProducer:
         self._kafka_producer = KafkaProducer.from_env()
         return self._kafka_producer
 
+    def _noop_project_version_latency_test(self, project_id: str) -> None:
+        # NOOP for testing latency impact of project switcher
+        try:
+            init_resolver(ch_client_factory=lambda: self.ch_client)
```
tssweeney: this global singleton is a weird pattern. why not just pass self.ch_client to your `resolve_read_table` method?

gtarpenning: It's so we can keep the cache consistent between requests, but I completely agree it is not my favorite pattern. Ideas?

weave/trace_server/project_version/project_version.py
```diff
@@ -0,0 +1,136 @@
...
+
+def resolve_write_target(project_id: str) -> WriteTarget:
+    if _resolver is None:
+        raise RuntimeError(
+            "init_resolver() must be called before resolve_write_target()"
+        )
+    return _resolver.resolve_write_target(project_id)
+
+
+class TableRoutingResolver:
```
tssweeney: Great! this Router contains the logic i was hoping for. a clear delineation of concerns

weave/trace_server/project_version/project_version.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Call this once at application startup. Subsequent calls are ignored
+    (the first initialization wins).
+
+    Args:
+        ch_client_factory: Callable that returns a ClickHouse client.
+        cache_size: Size of the in-memory cache (defaults to 10,000).
+    """
+    global _resolver
+    with _resolver_lock:
```
tssweeney: this seems sus. - why not just pass along the ch_client. why need a whole factory and global state

gtarpenning: its for the in-memory cache

weave/trace_server/clickhouse_trace_server_batched.py
tssweeney: well two options:

1. just have the router be a member of the ClickhouseTraceServer (this is a singleton in the server process)
2. OR, if you really wanted to force it, just put the `self._cache: LRUCache[str, ProjectDataResidence] = LRUCache(maxsize=cache_size)` in the global space

weave/trace_server/project_version/project_version.py
```diff
@@ -0,0 +1,136 @@
...
+
+        residence = self._get_residence(project_id)
+
+        if self._mode == CallsStorageServerMode.FORCE_LEGACY:
+            return WriteTarget.CALLS_MERGED
+
+        if self._mode == CallsStorageServerMode.AUTO:
+            if residence in (
+                ProjectDataResidence.COMPLETE_ONLY,
+                ProjectDataResidence.BOTH,
```
tssweeney, line 130: this tuple should contain `EMPTY`

weave/trace_server/project_version/project_version.py
```diff
@@ -0,0 +1,136 @@
...
+            ):
+                return WriteTarget.CALLS_COMPLETE
+            return WriteTarget.CALLS_MERGED
+
+        if self._mode == CallsStorageServerMode.DUAL_WRITE:
+            return WriteTarget.BOTH
```
tssweeney: this needs to condition on residency. if only in calls merged, don't do both

tssweeney (approved):
NICE

PR #5703, opened 2025-11-07
chore(weave): calls complete only write path

gtarpenning:
## Description

[https://wandb.atlassian.net/browse/WB](https://wandb.atlassian.net/browse/WB-29712)

## Testing

How was this PR tested?

### DIFF
--- a/tests/trace/test_calls_complete.py
+++ b/tests/trace/test_calls_complete.py
@@ -132,6 +132,116 @@ MODE_PARAMS = [
     ("force_legacy_mode", ProjectDataResidence.BOTH, False, True, "sdk_calls_complete"),
 ]
 
+# Old SDK mode parameters (call_start/call_end endpoints)
+# Format: (mode_fixture_name, ProjectDataResidence, should_succeed, writes_to_calls_complete, writes_to_calls_merged)
+# Old SDK uses CallSource.SDK_CALLS_MERGED and gets rejected if write_target is CALLS_COMPLETE or BOTH
+MODE_PARAMS_OLD_SDK = [
+    # EMPTY projects (new projects with no data)
+    (
+        "auto_mode",
+        ProjectDataResidence.EMPTY,
+        True,
+        False,
+        True,
+    ),  # Fixed by resolve_write_target check
+    (
+        "dual_write_read_merged_mode",
+        ProjectDataResidence.EMPTY,
+        True,
+        False,
+        True,
+    ),  # Allowed: old SDK writes to merged only, doesn't initiate dual-write
+    (
+        "dual_write_read_complete_mode",
+        ProjectDataResidence.EMPTY,
+        True,
+        False,
+        True,
+    ),  # Allowed: old SDK writes to merged only, doesn't initiate dual-write
+    (
+        "force_legacy_mode",
+        ProjectDataResidence.EMPTY,
+        True,
+        False,
+        True,
+    ),  # Allowed: writes to merged only
+    # MERGED_ONLY projects (legacy projects with only calls_merged data)
+    ("auto_mode", ProjectDataResidence.MERGED_ONLY, True, False, True),
+    (
+        "dual_write_read_merged_mode",
+        ProjectDataResidence.MERGED_ONLY,
+        True,
+        False,
+        True,
+    ),
+    (
+        "dual_write_read_complete_mode",
+        ProjectDataResidence.MERGED_ONLY,
+        True,
+        False,
+        True,
+    ),
+    ("force_legacy_mode", ProjectDataResidence.MERGED_ONLY, True, False, True),
+    # COMPLETE_ONLY projects (should never accept old SDK writes)
+    (
+        "auto_mode",
+        ProjectDataResidence.COMPLETE_ONLY,
+        False,
+        False,
+        False,
+    ),  # Rejected: write_target is CALLS_COMPLETE
+    (
+        "dual_write_read_merged_mode",
+        ProjectDataResidence.COMPLETE_ONLY,
+        False,
+        False,
+        False,
+    ),  # Rejected: edge case
+    (
+        "dual_write_read_complete_mode",
+        ProjectDataResidence.COMPLETE_ONLY,
+        False,
+        False,
+        False,
+    ),  # Rejected: edge case
+    (
+        "force_legacy_mode",
+        ProjectDataResidence.COMPLETE_ONLY,
+        True,
+        False,
+        True,
+    ),  # Allowed: force_legacy always writes merged
+    # BOTH projects (dual-write projects with data in both tables)
+    (
+        "auto_mode",
+        ProjectDataResidence.BOTH,
+        False,
+        False,
+        False,
+    ),  # Rejected: write_target is CALLS_COMPLETE
+    (
+        "dual_write_read_merged_mode",
+        ProjectDataResidence.BOTH,
+        False,
+        False,
+        False,
+    ),  # Rejected: write_target is BOTH
+    (
+        "dual_write_read_complete_mode",
+        ProjectDataResidence.BOTH,
+        False,
+        False,
+        False,
+    ),  # Rejected: write_target is BOTH
+    (
+        "force_legacy_mode",
+        ProjectDataResidence.BOTH,
+        True,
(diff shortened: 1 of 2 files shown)


gtarpenning:
* **#5704** <a href="https://app.graphite.com/github/pr/wandb/weave/5704?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#5724](https://github.com/wandb/weave/pull/5724) <a href="https://app.graphite.com/github/pr/wandb/weave/5724?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#5725](https://github.com/wandb/weave/pull/5725) <a href="https://app.graphite.com/github/pr/wandb/weave/5725?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#5703** <a href="https://app.graphite.com/github/pr/wandb/weave/5703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/5703?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5702** <a href="https://app.graphite.com/github/pr/wandb/weave/5702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/llm_completion.py
```diff
@@ -269,7 +269,9 @@ def lite_llm_completion(
                 # Handle different return types if needed in the future
                 pass
 
-            return tsi.CompletionsCreateRes(response=res.model_dump())
+            return tsi.CompletionsCreateRes(
```
gtarpenning, line 272: why do we need this??

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1962,3 +1966,320 @@ def _is_minimal_filter(filter: tsi.CallsFilter | None) -> bool:
...
+    UPDATE calls_complete
+    SET
+        ended_at = CASE {format_cases(field_cases["ended_at"])} ELSE ended_at END,
+        output_dump = CASE {format_cases(field_cases["output_dump"])} ELSE output_dump END,
+        output_refs = CASE {format_cases(field_cases["output_refs"])} ELSE output_refs END,
+        summary_dump = CASE {format_cases(field_cases["summary_dump"])} ELSE summary_dump END,
+        exception = CASE {format_cases(field_cases["exception"])} ELSE exception END,
+        wb_run_step_end = CASE {format_cases(field_cases["wb_run_step_end"])} ELSE wb_run_step_end END,
+        updated_at = now64(3)
+    WHERE project_id = {project_id_param}
```
neutralino1, line 2189: We can have thousands of calls per batch right? So the `CASE ...` statements could have thousands of `WHEN` parts?

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1962,3 +1966,320 @@ def _is_minimal_filter(filter: tsi.CallsFilter | None) -> bool:
...
+      AND id IN ({", ".join(where_params)})
+    """
+
+    return safely_format_sql(raw_sql, logger)
+
+
+def _build_calls_complete_update_query(
```
neutralino1, line 2196: The difference between this one and the previous is that here all updated calls get the same values?

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1962,3 +1966,320 @@ def _is_minimal_filter(filter: tsi.CallsFilter | None) -> bool:
...
+    This uses ClickHouse's lightweight DELETE with parameterized IN clause.
+    """
+    return _build_calls_complete_update_query(
+        project_id=project_id,
+        call_ids=call_ids,
+        update_fields={"deleted_at": (deleted_at, "DateTime64(3)")},
+        wb_user_id=wb_user_id,
+        updated_at=updated_at,
+        pb=pb,
+    )
```
neutralino1, line 2285: These are to support batch renaming and deletes right?

weave/trace_server/opentelemetry/python_spans.py
neutralino1, line 1: Are these changes necessary or just ergonomics?

weave/trace_server/trace_server_interface.py
```diff
@@ -2125,6 +2246,11 @@ def score_read(self, req: ScoreReadReq) -> ScoreReadRes: ...
     def score_list(self, req: ScoreListReq) -> Iterator[ScoreReadRes]: ...
     def score_delete(self, req: ScoreDeleteReq) -> ScoreDeleteRes: ...
 
+    # Calls V2 - Batch Endpoints
+    # TODO: this is start AND complete endpoint, we should name it something better
+    def calls_start_batch(self, req: CallsStartBatchReq) -> CallsStartBatchRes: ...
+    def calls_end_batch(self, req: CallsEndBatchReq) -> CallsEndBatchRes: ...
+
```
neutralino1, line 2490: Since we now support regular UPDATEs, should we split the endpoints as
```
POST /v2/calls  # create call
PUT /v2/calls   # update call
```
to make them more RESTful?

weave/trace_server/opentelemetry/python_spans.py
gtarpenning: refactors the logic to allow `to_complete`. mostly ergonomics

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: yeah, although should be extremely rare / abuse if true. I was planning on adding a limit, like... 100? Good reminder to do this here.

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: this is specifically for "updates" which in the existing schema only pertains to `display_name` and `deleted_at` but could be extended.

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: yes for delete, technically we do not allow batch renaming currently.

gtarpenning: this mirrors the existing `calls_merged` implementation

weave/trace_server/trace_server_interface.py
gtarpenning: after agonizing over this, i think it doesn't matter. PUT suggests inserting a complete object, which isn't exactly what we want. PATCH is more like what we want but still doesn't convey exactly the goal. We have 3 kinds of objects: starts, ends, and completes... Either way, this is the internal definition, so I think we want it to convey meaning closer to the implementation details. In the fastapi trace server definition we can define a different `/v2/...` external endpoint def, so i'm punting. I think the current impl is at least mostly clear.

unless we can think of a better name for "start". "insert" isn't too bad -> `calls_insert_batch`... idk.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -521,6 +551,200 @@ def call_end(
...
+            ]
+            self._insert_ch_insertable_calls_to_complete_table(
+                ch_insertable_starts + ch_insertable_completes,
+                do_sync_insert=do_sync_insert,
+            )
+
+        if ends:
+            self._update_calls_complete_batch(ends)
+
+    def _dual_write_calls_batch(
```
gtarpenning, line 737: We need to always dual write if the project residence is both. We should error if we can't dual write. When we cut over and stop dual writing, we can't have this constraint though.

gtarpenning: If there is any data in calls_complete, there must be an exact copy in calls_merged. If we ever break this contract, like where is more data in calls_merged, when we cut over data will be lots. If we have dual writes turned onto OTEL for new projects, then old clients can write to ONLY calls_merged and never end up in calls_complete.

gtarpenning: and the reverse

PR #5713, opened 2025-11-10
chore(weave): Narrow V2 API ObjectVersionFilter to ops or not-ops

andrewtruong:
The previous setting would query for all "objects" instead of just non-op objects


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1728,7 +1728,9 @@ def dataset_list_v2(
         Returns the rows reference as a string.
         """
         # Query the objects
-        dataset_filter = tsi.ObjectVersionFilter(base_object_classes=["Dataset"])
+        dataset_filter = tsi.ObjectVersionFilter(
+            base_object_classes=["Dataset"], is_op=False
```
gtarpenning, line 1730: might want to enum these magic strings at some point

gtarpenning (approved):

PR #5718, opened 2025-11-10
fix(weave): sort calls query with costs by complex fields 

gtarpenning:
[WB-29172](https://wandb.atlassian.net/browse/WB-29172)

## Description

Fixes a longstanding issue where it was impossible to sort by a constructed field with `include_costs=True` in the calls query. A frontend change made this much more obvious by (errantly) passing in sort fields to the cost query. 

## Testing

adds a bunch of unit tests and end to end tests that sort by complex constructed fields like specific dynamic fields and feedback join fields. 

[WB-29172]: https://wandb.atlassian.net/browse/WB-29172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/token_costs.py
+++ b/weave/trace_server/token_costs.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from typing import TYPE_CHECKING
 
 from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.calls_query_builder.cte import CTE
 from weave.trace_server.clickhouse_schema import SelectableCHCallSchema
 from weave.trace_server.errors import InvalidRequest
 from weave.trace_server.orm import (
@@ -17,7 +18,6 @@ from weave.trace_server.validation import (
     validate_purge_req_multiple,
     validate_purge_req_one,
 )
-from weave.trace_server.calls_query_builder.cte import CTE
 
 if TYPE_CHECKING:
     from weave.trace_server.calls_query_builder.calls_query_builder import OrderField


weave/trace_server/token_costs.py
```diff
@@ -335,15 +336,25 @@ def final_call_select_with_cost(
...
     select_fields: list[str],
-    order_fields: list[tsi.SortBy],
+    order_fields: list["OrderField"],
+    project_id: str,
 ) -> PreparedSelect:
     # We filter out summary_dump, because we add costs to summary dump in the select statement
     final_select_fields = [field for field in select_fields if field != "summary_dump"]
 
+    # circular dependency
+    from weave.trace_server.calls_query_builder.calls_query_builder import (
```
jwlee64: Dont love thats this is circular, but I can see why we'd import it

gtarpenning: yeah we should pull out all the types into a types file, but that's more work than i'm willing to do rn

weave/trace_server/token_costs.py
```diff
@@ -442,12 +452,55 @@ def final_call_select_with_cost(
...
+    )
+    if needs_feedback:
+        project_param = param_builder.add_param(project_id)
+        feedback_join = f"""
+        LEFT JOIN (
+            SELECT * FROM feedback WHERE feedback.project_id = {{{project_param}:String}}
+        ) AS feedback ON (
+            feedback.weave_ref = concat('weave-trace-internal:///', {{{project_param}:String}}, '/call/', {price_table_alias}.id))
+        """
+        # Insert the join after the FROM clause
```
gtarpenning: insane

weave/trace_server/token_costs.py
jwlee64, line 308: does this comment still make sense in this spot, since we moved the final select, might just be a delete, I think I have a similar or exact comment in the final cost query function, describing what the final select is supposed to do

jwlee64 (approved):
The new changes look good to me
Makes the cost query a lot cleaner.
I like the increased test coverage as well.

My main concern would be that by abstracting a lot of the logic out of the cte building makes the final query slightly harder to wrap your head around in its fullness, but honestly was quite tough already, and it simplifies the code from a lot of string manipulation to more functional

ty for the refactor and the fix

PR #5720, opened 2025-11-11
chore(weave): Rename V2 APIs to Object APIs

andrewtruong:
This is a chunky PR that removes the "V2" nomenclature from our APIs and instead calls them the "Object" APIs.

There are 3 core parts of this PR:
1. The V2 API definition is moved to its own file called `object_interface.py`, which defines an `ObjectInterface`.
    1. All models for this API are moved to this file as well
    3. The `TraceServerInterface` has been modified to include a new property `object_interface: ObjectInterface`, which contains the interface definition above.
2. All instances of the "V2" name have been removed in favour of just the regular name, e.g.
    1. `scorer_list_v2 -> scorer_list`;
    2. `ScorerReadV2Res -> ScorerReadRes`; etc
3. Various updates to the implementations to follow the new design described above

### DIFF
--- a/tests/trace_server/test_scorer_v2_api.py
+++ b/tests/trace_server/test_scorer_v2_api.py
@@ -25,9 +25,7 @@ def score(output: str, target: str) -> dict:
         description="A simple exact match scorer",
         op_source_code=op_source_code,
     )
-    create_res = trace_server.object_interface.object_interface.scorer_create(
-        create_req
-    )
+    create_res = trace_server.object_interface.scorer_create(create_req)
 
     assert create_res.digest is not None
     assert create_res.object_id == "exact_match"


weave/trace_server/trace_server_interface.py
```diff
@@ -2085,95 +1350,13 @@ def evaluate_model(self, req: EvaluateModelReq) -> EvaluateModelRes: ...
...
 
     This protocol represents a trace server implementation that supports the full
-    set of APIs - both legacy V1 endpoints and modern V2 endpoints. Use this type
+    set of APIs - both legacy V1 endpoints and modern Object endpoints. Use this type
     for implementations that need to support both API versions.
     """
 
-    pass
+    @cached_property
+    def object_interface(self) -> oi.ObjectInterface: ...
```
tssweeney, line 1362: curious why you chose to go this route as opposed to the mixin

andrewtruong: mainly because there is overlap in the namespace with `op_create` and `op_read` and mixin will clobber

tssweeney: hmmm. well `op_*` is not even implemented... i mean, i'm not going to die on this hill, but seems like unnecessary redirection

andrewtruong: `op_read` is implemented
https://github.com/wandb/weave/blame/master/weave/trace_server/clickhouse_trace_server_batched.py#L787

andrewtruong: But it seems like we don't expose this on the actual trace server... I wonder why this was implemented at all?  cc: @gtarpenning

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -465,215 +467,269 @@ def evaluation_status(
         req.project_id = self._idc.ext_to_int_project_id(req.project_id)
         return self._ref_apply(self._internal_trace_server.evaluation_status, req)
 
-    # === V2 APIs ===
+    @cached_property
+    def object_interface(self) -> oi.ObjectInterface:
```
tssweeney, line 471: another smell - you have to have a 2 way binding between these objects (.object_interface and ._server) ...

weave/trace_server/trace_server_interface.py
tssweeney: That PR was just a refactor of prior code (looking at the git history). the `op_*` methods were never exposed. see the server:

```
# Op API

# @prefix_router.post("/op/create", tags=[OPS_TAG_NAME])
# def op_create(req: tsi.OpCreateReq) -> tsi.OpCreateRes:
#     api = get_server_api()
#     assert auth_enforcement.can_write_project_scope(req.op_obj.entity, req.op_obj.project)
#     return api.op_create(req)


# @prefix_router.post("/op/read", tags=[OPS_TAG_NAME])
# def op_read(req: tsi.OpReadReq) -> tsi.OpReadRes:
#     api = get_server_api()
#     assert auth_enforcement.can_read_project_scope(req.entity, req.project)
#     return api.op_read(req)


# @prefix_router.post("/ops/query", tags=[OPS_TAG_NAME])
# def ops_query(req: tsi.OpQueryReq) -> tsi.OpQueryRes:
#     api = get_server_api()
#     assert auth_enforcement.can_read_project_scope(req.entity, req.project)
#     return api.ops_query(req)


# # Obj API

```

This goes back to the first implementation

tssweeney: If you are really trying to keep this simple and clean, I would either: rename the `op_*` methods to some underscore for legacy purposes or just delete them. They are not invoked anywhere in the code base.

tssweeney: it doesn't seem to me that this would justify the extra indirection

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -9,6 +9,7 @@
 from collections import defaultdict
 from collections.abc import Iterator, Sequence
 from contextlib import contextmanager
+from functools import cached_property
```
tssweeney, line 12: why does this file have so many LoC change? it seems like you used some automated tool that moved a bunch of things around

PR #5721, opened 2025-11-11
feat(weave): usage stats api

bcsherma:
## Summary

This PR + https://github.com/wandb/core/pull/36322 introduces a new `/calls/stats` endpoint that provides time-binned statistics for calls, supporting both LLM usage metrics (tokens/cost) and general call metrics (latency, counts).

Note that the query supports filtering by `tsi.CallsFilter` which includes op version, name, input refs, etc. but does not include heavy and computed fields that are available via the mongo-db-like Query parameter (see `trace_server/interface/query.py` if you don't know what I am talking about).

**Warning:** Because token usage in the summary is cumulatively rolled up the trace tree, failing to set `trace_roots_only=true` or a single value for `op_names` filter can result in double+ counting of real tokens and cost. For example, call A has children B and C. B and C have usage, A has the sum of their usages. If you specify `trace_roots_only` or filter to only the op of A then you get the correct rolled up count, otherwise you would get double the actual usage. This rollup behavior is also specific to the Python client—OTEL ingested calls or calls submitted via the REST API will not have such rollups applied before being written to ClickHouse.

The API allows users to request time-bucketed statistics for calls. The user provides a project, date range, time bucketing options, and two optional lists of metrics:
- **`usage_metrics`**: Token and cost metrics (always grouped by model)
- **`call_metrics`**: Latency, call count, error count (not grouped by model)

See the additions to `trace_server_interface.py` for exact request and response shapes.

### Request Example (CallStatsReq)
```json
{
    "project_id": "entity/project",
    "start": "2024-01-01T00:00:00Z",
    "end": "2024-01-02T00:00:00Z",
    "granularity": 3600,
    
    "usage_metrics": [
        {
            "metric": "input_tokens",
            "aggregations": ["sum", "avg"],
            "percentiles": [50, 95, 99]
        },
        {
            "metric": "total_cost",
            "aggregations": ["sum"]
        }
    ],
    
    "call_metrics": [
        {
            "metric": "latency_ms",
            "aggregations": ["avg", "max"],
            "percentiles": [95, 99]
        },
        {
            "metric": "call_count",
            "aggregations": ["sum"]
        },
        {
            "metric": "error_count",
            "aggregations": ["sum"]
        }
    ],
    
    "filter": {
        "op_names": ["openai.chat.completions.*"]
    },
    
    "timezone": "America/New_York"
}
```

### Response Example (CallStatsRes)
```json
{
    "start": "2024-01-01T00:00:00Z",
    "end": "2024-01-02T00:00:00Z",
    "granularity": 3600,
    "timezone": "America/New_York",
    
    "usage_buckets": [
        {
            "timestamp": "2024-01-01T00:00:00",
            "model": "gpt-4o",
            "sum_input_tokens": 15000,
            "avg_input_tokens": 150.5,
            "p95_input_tokens": 245,
            "sum_total_cost": 0.45,
            "count": 100
        }
    ],
    
    "call_buckets": [
        {
            "timestamp": "2024-01-01T00:00:00",
            "avg_latency_ms": 234.5,
            "max_latency_ms": 1250.0,
            "p95_latency_ms": 890.0,
            "sum_call_count": 100,
            "sum_error_count": 3
        }
    ]
}
```

### Design Decisions

**Two metric types with different grouping:**
- **Usage metrics** (`input_tokens`, `output_tokens`, `total_tokens`, `input_cost`, `output_cost`, `total_cost`) are always grouped by model name. The model comes from keys in `summary.usage` (e.g., `gpt-4o`, `claude-3-opus`). Model grouping is necessary to join on costs in the backend.
- **Call metrics** (`latency_ms`, `call_count`, `error_count`) are NOT grouped by model because we don't know how to attribute some latency or error to one model vs another, and a call can have usage for multiple models.

**Cost calculation:** Cost metrics (`input_cost`, `output_cost`, `total_cost`) are computed by joining token counts with the `llm_token_prices` table, not extracted directly from call data.

**Integration Tests** (`tests/trace_server/test_call_stats.py`):
End-to-end tests that:
- Create calls with known usage data via `call_start`/`call_end`
- Query the `/calls/stats` endpoint
- Verify aggregated results match expected values
- Cover both usage metrics (tokens per model) and call metrics (latency, counts)


bcsherma:
* **#5721** <a href="https://app.graphite.com/github/pr/wandb/weave/5721?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/5721?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5892** <a href="https://app.graphite.com/github/pr/wandb/weave/5892?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/trace_server_interface.py
```diff
@@ -2132,6 +2132,7 @@ def calls_query(self, req: CallsQueryReq) -> CallsQueryRes: ...
     def calls_query_stream(self, req: CallsQueryReq) -> Iterator[CallSchema]: ...
     def calls_delete(self, req: CallsDeleteReq) -> CallsDeleteRes: ...
     def calls_query_stats(self, req: CallsQueryStatsReq) -> CallsQueryStatsRes: ...
+    def call_stats(self, req: "CallStatsReq") -> "CallStatsRes": ...
```
chance-wnb, line 2256: You might need to take care of `/weave/trace_server/sqlite_trace_server.py` as well

chance-wnb (changes requested):
Thanks for the change. I suggest a division of the API scope.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -609,6 +618,228 @@ def calls_query_stats(self, req: tsi.CallsQueryStatsReq) -> tsi.CallsQueryStatsR
...
+                    if completion_cost
+                    else 0.0,
+                }
+        except Exception:
+            # If price query fails, return empty prices (costs will be 0)
+            return {}
+        else:
+            return prices
+
+    def _compute_costs_for_buckets(
```
chance-wnb, line 845: I think the benefits of server side processing is that it has:  
  
1\. Strong aggregation capability from a large scale of data  
2\. Strong bucketing capability from a large scale of data  
  
These are not something frontend can easily achieve. So this process should be done on the server side.   
  
However I don't think such a `price * unit` calculation need to happen on the server side. FE should be more suitable to carry out these tasks so the backend procedures can finish sooner. It also reduces network transportation volume.   
  
So I would suggest consider putting `_get_prices_for_models` as its own endpoint (or maybe we already have an endpoint to emit model prices?) and assembly the cost info purely on the frontend.   
  
At the end of the day, a less integrated and more atomic API might be more flexible and useful. For maintainability purpose, I am seeking all opportunities to simplify things.  
  
I am `soft` putting this PR as `request changes` , but I am totally open to negotiation if you think otherwise. 🍻

chance-wnb (changes requested):

weave/trace_server/trace_server_interface.py
bcsherma: I configured the end to end test to skip when the sqlite server is used 🤫 😁

weave/trace_server/clickhouse_trace_server_batched.py
bcsherma: > However I don't think such a price * unit calculation need to happen on the server side. FE should be more suitable to carry out these tasks so the backend procedures can finish sooner. It also reduces network transportation volume.

Agreed that the cost calculation does not need to happen on the server side. We do have an endpoint for reading model prices! While it will certainly reduce network transport volume for the `/call/stats` query we will then be issuing an additional query, so depending on the number of providers, number of buckets, etc it may end up being more transport when split up.

> At the end of the day, a less integrated and more atomic API might be more flexible and useful. For maintainability purpose, I am seeking all opportunities to simplify things.

I think being able to get costs back directly from the stats query is useful and makes the api more flexible.

Do you mean atomic in the sense that each endpoint should perform at most one query against the clickhouse database? I atomicity in that sense makes it easier for us as engineers to reason about the trace server API and our clickhouse queries as if they are 1-1. 

In this case we probably **can** make it a single query.  It should be possible for me to move this cost join that currently happens in the api into the clickhouse query layer. I opted for this route with a follow up query to try and keep the main query on `calls_merged` as simple as I could. I am going to give that a shot.

chance-wnb: > it may end up being more transport when split up.

In terms of transportation cost, FE only need to retrieve the cost information once and caching it. So the subsequent interaction: zooming, panning, re-bucketing don't need to refetch it again. We are going to save the compute on the server side, FE compute is cheap, we only need to pay a development cost. 

Also in your current implementation, you transport the cost information per query per bucket. After the refactoring the cost payload will be per model exactly once. This should save significantly.

By atomicity, I mean that an endpoint does one focused task at a time. That simply means the `/call/stats` do not include the cost information. I see that you have a request flag to turn this feature on and off. I am thinking that the API doesn't even need to worry about the potential of calculating cost. So the freedom of composition is on caller side, not the API side. 

> at most one query against the clickhouse database?

Nope, that doesn't really matter. backend implementation takes as many queries as needed. that's totally fine.

weave/trace_server/trace_server_interface.py
chance-wnb: sqlite_trace_server also implements TSI. As an implementor, without properly implementing the interface, wouldn't it error out?

weave/trace_server/clickhouse_trace_server_batched.py
chance-wnb: > move this cost join that currently happens in the api into the clickhouse query layer

Honestly speaking I am somewhat against the idea. That adds complexity to the query and might make the future work more difficult :) I observed that the SQL join for the calls_query adds a lot complexity and enforces as to use alias whenever we need to optimize the performance. That's part of the reason why I believe atomicity reduces complexity and is more flexible. FE might be a better role to orchestrate the workflow.

bcsherma: > In terms of transportation cost, FE only need to retrieve the cost information once and caching it. So the subsequent interaction: zooming, panning, re-bucketing don't need to refetch it again.

Unless the set of used models changes when the time window changes. But agreed that these results are highly cacheable. But if we have a chart zoom/pan case, our API already supports making the optimization you are proposing- the frontend can just request token metrics and read costs and do the logic itself. I don't think that means we should remove the ability to get cost stats directly.

> By atomicity, I mean that an endpoint does one focused task at a time. That simply means the /call/stats do not include the cost information. I see that you have a request flag to turn this feature on and off. I am thinking that the API doesn't even need to worry about the potential of calculating cost. So the freedom of composition is on caller side, not the API side.

I guess I don't understand how to determine whether our queries are performing "one focused task at a time" or not. I don't really know what task or focus mean in this precise context 

> Honestly speaking I am somewhat against the idea. That adds complexity to the query and might make the future work more difficult :)

My thoughts exactly

I am understanding your comment as recommendation that the api should not send redundant data. Some clear examples where we do this now:
- the calls query reads costs and populates the call summary with token AND cost information
- calls have started_at and ended_at fields but we also take their difference and pack into `summary.weave.latency_ms` 

I agree that we can compress these redundancies out of the api, but I question the value of doing so and to me it seems like it clearly makes the api less flexible and useful if you remove optional capabilities from it.

Is the change you are requesting that I remove the ability to get cost stats directly from this api?

PR #5722, opened 2025-11-11
fix(weave): allow escaped periods to designate non-delimiters

gtarpenning:
## Description

[WB-22547](https://wandb.atlassian.net/browse/WB-22547)

Backend to support filtering by keys that have periods in them that are not actually nested. Comprehensive examples in tests, but trivially this looks like:
```python
@weave.op
def log():
   return {"nested.key": "hi"}
```

Now, using this change, you can filter by these items with:
```python
client.get_calls(query={"$expr": {"$eq": [{"$getField": "output.nested\\.key"}, {"$literal": "hi"}]}})
```

## Testing

- query builder unit tests
- end to end sdk tests

[WB-22547]: https://wandb.atlassian.net/browse/WB-22547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/calls_query_builder/utils.py
+++ b/weave/trace_server/calls_query_builder/utils.py
@@ -6,11 +6,7 @@ from typing import Optional
 import sqlparse
 
 from weave.trace_server.interface import query as tsi_query
-from weave.trace_server.orm import (
-    ParamBuilder,
-    clickhouse_cast,
-    quote_json_path_parts,
-)
+from weave.trace_server.orm import ParamBuilder, clickhouse_cast, quote_json_path_parts
 
 
 def param_slot(param_name: str, param_type: str) -> str:


weave/trace_server/orm.py
```diff
@@ -528,14 +528,59 @@ def clickhouse_cast(
         raise ValueError(f"Unknown cast: {cast}")
 
 
+def split_escaped_field_path(path: str) -> list[str]:
```
neutralino1, line 532: Could this be done with a regex and match groups?

gtarpenning: yeah, I can do that.

neutralino1 (approved):

PR #5723, opened 2025-11-11
chore(weave): lint some files

gtarpenning:
## Description

These files can be linted and are annoying.


jamie-rasmussen:
seems like ruff disagrees :-) version issue?

gtarpenning:
local version mismatch, closing ty @jamie-rasmussen

PR #5704, opened 2025-11-07
chore(weave): calls complete only read path

gtarpenning:
## Description

Read path for reading from the new calls_complete table. 

Old read path should be unaffected, but we should now automatically be able to generate queries against the new table. 

## Testing

- adds a new shard that runs in `dual_write_read_calls_complete` mode, writing to both tables and reading all call information from the new table. 
- does not modify any of the existing unit tests that generate calls queries

PR #5725, opened 2025-11-11
feat(weave): sdk to enable v2 call writes

gtarpenning:
## Description

Implement a call_processor which specially handles sending traces to the new v2 calls endpoints. Handles cases like ends getting logged first, flushes interrupting sending, call_start_delay grouping starts and ends...

## Testing

manual, along with basic unit tests

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -766,6 +766,7 @@ class WeaveClient:
             current_wb_run_step = None
 
         started_at = datetime.datetime.now(tz=datetime.timezone.utc)
+        call.started_at = started_at
         project_id = self._project_id()
 
         should_print_call_link_ = should_print_call_link()
@@ -954,6 +955,7 @@ class WeaveClient:
                 end=EndedCallSchemaForInsert(
                     project_id=project_id,
                     id=call.id,
+                    started_at=call.started_at,
                     ended_at=ended_at,
                     output=output_json,
                     summary=merged_summary,
--- a/weave/trace_server_bindings/call_batch_processor.py
+++ b/weave/trace_server_bindings/call_batch_processor.py
@@ -2,13 +2,14 @@
 
 This module provides CallBatchProcessor, which extends AsyncBatchProcessor
 to maximize complete calls by pairing starts with ends at enqueue time.
+
+Handles the race condition where ends may arrive before their starts in
+concurrent execution by buffering both starts and ends until they can be paired.
 """
 
 import logging
 import time
 from collections.abc import Callable
-
-logger = logging.getLogger(__name__)
 
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.constants import EVALUATION_RUN_OP_NAME
@@ -22,30 +23,19 @@ from weave.trace_server_bindings.models import (
     StartBatchItem,
 )
 
-# Type alias for batch items
+logger = logging.getLogger(__name__)
+
 BatchItem = StartBatchItem | EndBatchItem | CompleteBatchItem
 
-# Maximum delay (in seconds) for start_hold_timeout
 MAX_CALL_START_DELAY = 600.0  # 10 minutes
+FLUSH_POLL_INTERVAL = 0.1  # 100ms
 
 
 class CallBatchProcessor(AsyncBatchProcessor[BatchItem]):
     """Batch processor that pairs starts with ends to maximize complete calls.
 
-    Extends AsyncBatchProcessor with call-specific pairing behavior:
-
-    1. **Buffers starts**: StartBatchItems are held in a dict
-    2. **Pairs on end arrival**: EndBatchItems are paired with matching starts
-       to create CompleteBatchItems
-    3. **Timeout-based flush**: Starts without ends after `start_hold_timeout`
-       are flushed as unpaired starts
-
-    Example flow:
-        1. start_a arrives → buffered in pending_starts
-        2. start_b arrives → buffered in pending_starts
-        3. end_a arrives → paired with start_a → CompleteBatchItem queued
-        4. timeout expires → start_b flushed as unpaired start
-        5. batch sent: [complete_a, start_b]
+    Buffers both starts and ends to handle race conditions in concurrent execution
+    where ends may arrive before their corresponding starts.
     """
 
     def __init__(
@@ -62,11 +52,12 @@ class CallBatchProcessor(AsyncBatchProcessor[BatchItem]):
 
         Args:
             processor_fn: Function to process batches of items.
-            max_batch_size: Maximum items per batch. Defaults to 100.
-            min_batch_interval: Minimum seconds between processing. Defaults to 1.0.
-            max_queue_size: Maximum queue size. Defaults to 10,000.
-            start_hold_timeout: Seconds to hold starts waiting for ends.
-                Use -1 for infinite wait, 0 for no buffering.
+            max_batch_size: Maximum items per batch.
+            min_batch_interval: Minimum seconds between processing.
+            max_queue_size: Maximum queue size.
+            start_hold_timeout: Seconds to hold items waiting for pairing.
+                Use -1 for infinite wait (capped at MAX_CALL_START_DELAY during flush),
+                0 for no buffering.
             enable_disk_fallback: Write dropped items to disk if True.
             disk_fallback_path: Path for dropped items log file.
         """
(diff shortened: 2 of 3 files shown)


gtarpenning:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/wandb/weave/5725?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5887** <a href="https://app.graphite.com/github/pr/wandb/weave/5887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5725** <a href="https://app.graphite.com/github/pr/wandb/weave/5725?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/5725?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5704** <a href="https://app.graphite.com/github/pr/wandb/weave/5704?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#5724](https://github.com/wandb/weave/pull/5724) <a href="https://app.graphite.com/github/pr/wandb/weave/5724?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#5703** <a href="https://app.graphite.com/github/pr/wandb/weave/5703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5702** <a href="https://app.graphite.com/github/pr/wandb/weave/5702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

tests/trace/test_weave_client.py
```diff
@@ -4217,3 +4218,81 @@ def test_evaluate_with_llm_completion_model_and_prompt_template_vars(client):
...
         {"role": "user", "content": "Additional context"}
     ]
+
+
+def test_call_start_delay_modes(client):
+    """Test different call start delay modes and their behavior."""
+    # Test 1: delay=0 sends immediately (call visible during execution)
+    with patch.object(settings, "call_start_delay", return_value=0.0):
+
+        @weave.op
```
andrewtruong: This seems weird.  Shouldn't both `immediate_op` and `delayed_op` have the same impl?  The difference is when you call `get_calls`, not the op def itself as these sections seem to imply.

tests/trace_server_bindings/conftest.py
```diff
@@ -91,22 +94,32 @@ def server(request, server_class):
     """Common server fixture that uses server_class based on the CLI flag."""
     server_ = server_class("http://example.com", should_batch=True)
 
+    shared_mock = MagicMock()
```
andrewtruong: Why shared?

tests/trace_server_bindings/test_trace_server_bindings.py
```diff
@@ -32,11 +33,61 @@
...
+
+    The test fixture sets both _send_calls_start_batch_to_server and
+    _send_calls_end_batch_to_server to the same mock object, so we need
+    to check for that to avoid double-counting.
+    """
+    if (
+        server._send_calls_start_batch_to_server
+        is server._send_calls_end_batch_to_server
+    ):
+        return server._send_calls_start_batch_to_server.call_count
```
andrewtruong: Can you add a comment for when we would hit each case?

tests/trace_server_bindings/test_trace_server_bindings.py
```diff
@@ -32,11 +33,61 @@
...
+    Handles deduplication when both mocks point to the same object.
+    """
+    total_items_sent = 0
+    seen_calls = set()
+    for mock_method in [
+        server._send_calls_start_batch_to_server,
+        server._send_calls_end_batch_to_server,
+    ]:
+        for call in mock_method.call_args_list:
+            call_id = id(call)
```
andrewtruong: Shouldn't this be the `call.id`?

tests/trace_server_bindings/test_trace_server_bindings.py
```diff
@@ -32,11 +33,61 @@
...
+            if call_id not in seen_calls:
+                seen_calls.add(call_id)
+                called_data = call[0][2]
```
andrewtruong: This seems kinda magical.  What does [0][2] mean?

tests/trace_server_bindings/test_trace_server_bindings.py
```diff
@@ -32,11 +33,61 @@
...
+    RemoteHTTPTraceServer consolidates matching start/end pairs into complete calls,
+    so N pairs = N items. StainlessRemoteHTTPTraceServer sends starts and ends
+    separately, so N pairs = 2N items.
+    """
+    return num_pairs if server_class == RemoteHTTPTraceServer else num_pairs * 2
+
+
 @pytest.mark.parametrize("server", ["small_limit"], indirect=True)
-def test_large_batch_is_split_into_multiple_smaller_batches(server):
+def test_large_batch_is_split_into_multiple_smaller_batches(server, server_class):
```
andrewtruong: What is `server_class` and why was it added here?

tests/trace_server_bindings/test_trace_server_bindings.py
```diff
@@ -68,9 +115,11 @@ def test_small_batch_is_sent_in_one_request(server):
...
     server._flush_calls(batch)
 
-    # Verify _send_batch_to_server was called once with the entire batch
-    assert server._send_batch_to_server.call_count == 1
-    called_data = server._send_batch_to_server.call_args[0][0]
+    # Small batch should be sent in one request without splitting
+    assert get_total_mock_calls(server) == 1
+
+    # Verify the single item was sent
+    called_data = server._send_calls_start_batch_to_server.call_args[0][2]
```
andrewtruong: Same, a bit magical

weave/trace/settings.py
```diff
@@ -206,6 +206,17 @@ class UserSettings(BaseModel):
     Can be overridden with the environment variable `WEAVE_USE_STAINLESS_SERVER`
     """
 
+    call_start_delay: float = 5.0
```
andrewtruong, line 209: Worth calling out in the comment that the max is 600s per the const in the next file

weave/trace/weave_client.py
```diff
@@ -801,18 +813,57 @@ def send_start_call() -> bool:
...
+        else:
+
+            def on_complete(f: Future) -> None:
+                try:
+                    root_call_did_not_error = f.result() and not current_call
+                    if root_call_did_not_error and should_print_call_link_:
+                        print_call_link(call)
+                except Exception:
+                    # Ensure the event is set even if call_start failed
+                    start_enqueued_event.set()
```
andrewtruong: These feel like they should be in a `finally` block somewhere

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -131,55 +131,231 @@ def delete(self, url: str, *args: Any, **kwargs: Any) -> httpx.Response:
...
+                consolidated.append(end_item)
+
+        # Add pre-existing completes
+        consolidated.extend(completes)
+
+        return consolidated
+
+    def _extract_entity_project(
+        self, batch: list[StartBatchItem | EndBatchItem | CompleteBatchItem]
+    ) -> tuple[str, str, str]:
```
andrewtruong: Try not to return tuple if you can.  It's harder to glance what is actually being returned

andrewtruong: I don't think this is in a hot path so returning a dataclass or similar here is better

tests/trace/test_weave_client.py
gtarpenning: delayed op waits 0.6 seconds IN the op, so the start gets sent before the end completes. We call the get_calls DURING the op to ensure each state we expect; in the delayed op case have an additional "running" state before the end hits the db

tests/trace_server_bindings/conftest.py
gtarpenning: just shared between `_send_calls_start_batch_to_server` and `_send_calls_end_batch_to_server`

tests/trace_server_bindings/test_trace_server_bindings.py
gtarpenning: thats cleaner, doesn't matter in this case but yeah it should be.

tests/trace_server_bindings/test_trace_server_bindings.py
gtarpenning: hmm, clarifying

tests/trace_server_bindings/test_trace_server_bindings.py
gtarpenning: there is different behavior in the stainless server setup because we have special handling in remote_http_trace_server

tests/trace/test_weave_client.py
andrewtruong: I see, thanks for clarifying

tests/trace_server_bindings/conftest.py
```diff
@@ -92,21 +95,29 @@ def server(request, server_class):
     server_ = server_class("http://example.com", should_batch=True)
 
     if request.param == "normal":
-        server_._send_batch_to_server = MagicMock()
+        server_._send_calls_start_batch_to_server = MagicMock()
+        server_._send_calls_end_batch_to_server = MagicMock()
```
jtschoonhoven, line 99: Nit: I would typically use `patch.object(server_, "_send_calls_start_batch_to_server")` for this rather than overwriting the property entirely. Though possibly there's a good reason for this I'm missing.

weave/trace/settings.py
```diff
@@ -336,6 +350,14 @@ def should_use_stainless_server() -> bool:
     return _should("use_stainless_server")
 
 
+def call_start_delay() -> float:
+    """Returns the delay in seconds before sending a call start."""
+    delay = _optional_float("call_start_delay")
+    if delay is None:
+        return 5.0
```
jtschoonhoven, line 357: Nit: maybe there should be a constant `DEFAULT_CALL_START_DELAY = 5.0` so that this value can't come out of sync with the `call_start_delay` assigned above.

weave/trace_server_bindings/call_batch_processor.py
```diff
@@ -0,0 +1,246 @@
...
+    EndBatchItem,
+    StartBatchItem,
+)
+
+logger = logging.getLogger(__name__)
+
+BatchItem = StartBatchItem | EndBatchItem | CompleteBatchItem
+
+MAX_CALL_START_DELAY = 600.0  # 10 minutes
+FLUSH_POLL_INTERVAL = 0.1  # 100ms
```
jtschoonhoven, line 31: Tiny nit: could you use `timedelta(minutes=10)` so the units/magnitude are obvious?

weave/trace_server_bindings/call_batch_processor.py
```diff
@@ -0,0 +1,246 @@
...
+            wb_run_id=start.req.start.wb_run_id,
+            wb_run_step=start.req.start.wb_run_step,
+            wb_run_step_end=end.req.end.wb_run_step_end,
+        )
+        return CompleteBatchItem(req=tsi.CallCompleteReq(complete=complete_call))
+
+    def _queue_item(self, item: BatchItem) -> None:
+        try:
+            self.queue.put_nowait(item)
+        except Exception:
```
jtschoonhoven, line 207: Nit
```suggestion
        except Full:
```

weave/trace_server_bindings/call_batch_processor.py
```diff
@@ -0,0 +1,246 @@
...
+            self._dropped_item_count += 1
+            error_message = (
+                f"Ready queue full. Dropping item. Max queue size: {self.queue.maxsize}"
+            )
+            if self._dropped_item_count % 1000 == 1:
+                logger.warning(
+                    f"{error_message}. Total dropped items: {self._dropped_item_count}"
```
jtschoonhoven, line 214: Tiny nit/suggestion:
```suggestion
                    f"{error_message}. Total dropped items: {self._dropped_item_count}",
                    exc_info=True
```

weave/trace_server_bindings/models.py
```diff
@@ -13,10 +13,23 @@ class EndBatchItem(BaseModel):
     req: tsi.CallEndReq
 
 
+class CompleteBatchItem(BaseModel):
+    mode: str = "complete"
```
jtschoonhoven, line 17: Can this type be more constrained than any `str`?

weave/trace_server_bindings/call_batch_processor.py
gtarpenning: Cool, didn't know about this

weave/trace_server_bindings/models.py
gtarpenning: I mean... `Literal` but no caller would ever set this value themselves, i think it's fine.

PR #5726, opened 2025-11-11
fix: Improve __dirname handling for CommonJS and ESM

mscolnick:
This changes the `__dirname` logic to handle commonjs and esm.

For context, I wanted to get this running in a cloudflare worker, which led to this error:

```
  Uncaught ReferenceError: __dirname is not defined at null.<anonymous>
```

### DIFF
--- a/sdks/node/src/utils/userAgent.ts
+++ b/sdks/node/src/utils/userAgent.ts
@@ -1,10 +1,21 @@
 import {existsSync, readFileSync} from 'fs';
-import {join} from 'path';
+import {dirname, join} from 'path';
+import {fileURLToPath} from 'url';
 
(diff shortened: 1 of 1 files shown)


mscolnick:
actually, i don't think this is the only incompatibility. so this change along won't unblock esm.

are there any plans support to have an esm and commonjs build? e.g. `.cjs` and `.ejs`

neutralino1:
cc @chance-wnb 👆

chance-wnb:
I have been successful using weave in a esm project before, despite the weave part is in commonjs. 

What issues did you run into?

Seperately, we don't have plan yet to compile duel-version of cjs and ejs for weave.

PR #5733, opened 2025-11-12
chore(weave): Make digest calculation stable across client and server by stripping refs of non-content data

tssweeney:
This PR changes how we compute digests for stored values. That’s a big deal: after this lands, a lot of previously-stored objects will get **new digests** even when the underlying content hasn’t meaningfully changed. In practice that shows up as “version bumps”/dedupe misses for older data. We think the one-time disruption is worth it because it unlocks a much simpler and faster ingest path.

### Why this change?

Today the client can’t safely compute the same digest the server will store. The root cause is refs:

- Clients embed **external refs** like `weave:///entity/project/...`
- The server rewrites those into an internal form for storage (stable across renames/migrations)

That mutation means the JSON payload the client sees is not the same as the JSON payload the server hashes, so the digest diverges. The client has to wait for the server response, which forces complicated queueing / dependency chains on the sender and blocks cleaner client→sender→server architectures.

### Alternatives considered

We could give the client project IDs up front and have it construct the internal ref format. That’s workable but has real downsides:
- Requires a pre-flight network exchange before logging (hurts offline/future modes)
- Exposes internal storage details to the client API surface

### What we’re doing instead

We make digest calculation **ref-aware** and **self-contained**:

- Add a single entrypoint `safe_digest(...)` used by both client and server.
- When hashing structured values, it normalizes ref strings so the “entity/project” portion doesn’t affect identity. Concretely we rewrite both external and internal refs into a common anonymous form (`weave-anonymous:///_/_/...`) before hashing.
- Digests are stable to dict insertion order (recursively).
- Sets are rejected (unordered); callers must convert to list/tuple.
- `safe_digest` also supports hashing raw `bytes` (file contents) and `str` (source/code capture), so callers don’t need separate APIs.

We also keep `table_digest_from_row_digests` as the table-specific digest (order-dependent), and route all trace-server object/table/file digests through `safe_digest`.

### Impact

- **Expected**: existing objects in the DB may appear as new versions / new digests after this change.
- **Benefit**: clients can compute digests locally without knowing server internals, enabling simpler and more performant ingest pipelines.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,14 @@ repos:
     rev: v2.1.0
     hooks:
       - id: fixit-fix
+  - repo: local
+    hooks:
+      - id: ty
+        name: ty (type checker)
+        entry: uv run ty check
+        language: python
+        always_run: true
+        pass_filenames: false
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.12.3
     hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -381,6 +381,7 @@ dev = [
   # linting
   "pre-commit",
   "jupyter",
+  "ty>=0.0.2",
 ]
 trace_server = [
   # Same dependencies as the trace_server optional dependency, but as a dependency group for tests
@@ -490,3 +491,33 @@ conflicts = [
 [tool.coverage.report]
 exclude_also = ["if TYPE_CHECKING:", "@abstractmethod"]
 omit = ["weave/agent/**"]
+
+[tool.ty.src]
+include = ["weave"]
+exclude = ["tests", "examples", "pyproject.toml"]
+
+[tool.ty.environment]
+python-version= "3.10"
+
+[tool.ty.rules]
+# These are ignored for the sake of getting the type checker added.
+# We should be dilligent in removing these ignores over time.
+invalid-assignment = "ignore"
+possibly-missing-attribute = "ignore"
+unresolved-import = "ignore"
+call-non-callable = "ignore"
+unresolved-attribute = "ignore"
+invalid-argument-type = "ignore"
+redundant-cast = "ignore"
+invalid-method-override = "ignore"
+deprecated = "ignore"
+missing-typed-dict-key = "ignore"
+invalid-return-type = "ignore"
+unknown-argument = "ignore"
+missing-argument = "ignore"
+non-subscriptable = "ignore"
+unsupported-operator = "ignore"
+no-matching-overload = "ignore"
+possibly-missing-import = "ignore"
+too-many-positional-arguments = "ignore"
+not-iterable = "ignore"
--- a/weave/evaluation/eval_imperative.py
+++ b/weave/evaluation/eval_imperative.py
@@ -13,18 +13,7 @@ from contextlib import contextmanager
 from contextvars import ContextVar
 from threading import Lock
 from types import MethodType
-from typing import TYPE_CHECKING, Annotated, Any, TypeVar, cast, overload
-
-if TYPE_CHECKING:
-    from weave.trace.call import Call
-
-from pydantic import (
-    BaseModel,
-    BeforeValidator,
-    ConfigDict,
-    Field,
-    PrivateAttr,
-)
+from typing import TYPE_CHECKING, Any, TypeVar, cast, overload
 
 from weave.dataset.dataset import Dataset
 from weave.evaluation.eval import Evaluation, default_evaluation_display_name
@@ -43,6 +32,9 @@ from weave.trace.util import Thread
 from weave.trace.view_utils import set_call_view
 from weave.type_wrappers.Content.content import Content
 from weave.utils.sentinel import NOT_SET, _NotSetType
+
+if TYPE_CHECKING:
+    from weave.trace.call import Call
 
 T = TypeVar("T")
 ID = str
@@ -309,7 +301,7 @@ class _LogScoreContext:
                 self._call_stack_context.__exit__(exc_type, exc_val, exc_tb)
 
 
-class ScoreLogger(BaseModel):
+class ScoreLogger:
     """Interface for logging scores and managing prediction outputs.
 
     This class is returned by `EvaluationLogger.log_prediction()` and can be used
@@ -346,20 +338,24 @@ class ScoreLogger(BaseModel):
     ```
     """
 
-    # model_id: ID
-    predict_and_score_call: Call
-    evaluate_call: Call
-    predict_call: Call
-    predefined_scorers: list[str] | None = None
+    def __init__(
+        self,
+        predict_and_score_call: Call,
+        evaluate_call: Call,
+        predict_call: Call,
+        predefined_scorers: list[str] | None = None,
+    ) -> None:
+        self.predict_and_score_call = predict_and_score_call
+        self.evaluate_call = evaluate_call
+        self.predict_call = predict_call
+        self.predefined_scorers = predefined_scorers
 
-    _captured_scores: dict[str, ScoreType] = PrivateAttr(default_factory=dict)
-    _has_finished: bool = PrivateAttr(False)
-    _predict_output: Any = PrivateAttr(None)
-    _call_stack_context: contextlib.AbstractContextManager[list[Call]] | None = (
-        PrivateAttr(None)
-    )
-
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+        self._captured_scores: dict[str, ScoreType] = {}
+        self._has_finished: bool = False
+        self._predict_output: Any = None
+        self._call_stack_context: (
+            contextlib.AbstractContextManager[list[Call]] | None
+        ) = None
 
     def finish(self, output: Any | None = None) -> None:
         """Finish the prediction and log all scores.
@@ -544,14 +540,7 @@ class ScoreLogger(BaseModel):
 
     async def alog_score(
         self,
-        scorer: Annotated[
-            Scorer | dict | str,
-            Field(
-                description="A metadata-only scorer used for comparisons."
-                "Alternatively, you can pass a dict of attributes or just a string"
-                "representing the ID of your scorer."
-            ),
-        ],
+        scorer: Scorer | dict | str,
         score: ScoreType,
     ) -> None:
         if self._has_finished:
@@ -605,7 +594,7 @@ class ScoreLogger(BaseModel):
                 self._call_stack_context.__exit__(exc_type, exc_val, exc_tb)
 
 
-class EvaluationLogger(BaseModel):
+class EvaluationLogger:
     """This class provides an imperative interface for logging evaluations.
 
     An evaluation is started automatically when the first prediction is logged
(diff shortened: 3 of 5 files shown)


weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+        ref (str): The input reference string.
+
+    Returns:
+        str: The stabilized reference string.
+
+    Examples:
+        >>> _stabilize_ref_str("weave:///entity/proj/object/Foo:abc123/attr/bar")
+        'weave-anonymous:///_/_/object/Foo:abc123/attr/bar'
+        >>> _stabilize_ref_str("weave-trace-internal:///proj_id/object/Foo:abc123/attr/bar")
+        'weave-anonymous:///_/_/object/Foo:abc123/attr/bar'
```
neutralino1, line 173: Is the `_/_/` part really necessary? We could actually simply strip everything before `/object`. No need for a prefix since this is only in-memory to compute a digest.
We could simply call this the "ref path" like URI path that strips scheme and domain.

```
        >>> _get_ref_path("weave:///entity/proj/object/Foo:abc123/attr/bar")
        '/object/Foo:abc123/attr/bar'
        >>> _get_ref_path("weave-trace-internal:///proj_id/object/Foo:abc123/attr/bar")
        '/object/Foo:abc123/attr/bar'
```

neutralino1: Additional question, this introduces cross-project collisions. Now two objects from two different projects but with the same value and name will get the same digest. Is that a problem?

neutralino1: `_get_ref_path` could actually be entirely implemented as a regex.

tests/trace_server/test_digest_builder.py
```diff
@@ -0,0 +1,143 @@
...
+
+def test_bytes_digest_is_deterministic_and_urlsafeish():
+    d1 = safe_digest(b"abc")
+    d2 = safe_digest(b"abc")
+    d3 = safe_digest(b"abcd")
+
+    assert d1 == d2
+    assert d1 != d3
+
+    # Our encoding strips '=' padding and replaces '-'/'_' with 'X'/'Y'.
```
andrewtruong, line 35: why?

tests/trace_server/test_digest_builder.py
```diff
@@ -0,0 +1,143 @@
...
+    v2 = {
+        "outer": {
+            "a": [{"k1": 1, "k2": 2}, {"y": 2, "x": 1}],
+            "b": {"a": 2, "z": 1},
+        }
+    }
+    assert safe_digest(v1) == safe_digest(v2)
+
+
+# def test_safe_digest_raises_on_non_string_dict_keys():
```
andrewtruong, line 113: Should this be commented?

tests/trace_server/test_digest_builder.py
```diff
@@ -0,0 +1,143 @@
...
+    payload_digest = safe_digest({"x": 1, "y": 2})
+    ext = {"ref": f"weave:///entity/proj/object/Foo:{payload_digest}/attr/bar"}
+    intr = {
+        "ref": f"weave-trace-internal:///proj_id/object/Foo:{payload_digest}/attr/bar"
+    }
+
+    assert safe_digest(ext) == safe_digest(intr)
+
+
+def test_safe_digest_raises_on_sets():
```
andrewtruong, line 128: shouldn't this raise on anything that is not hashable?

The corollary: we should be able to do this for anything that is hashable

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+from collections.abc import Callable
+from typing import Any, TypeVar, cast
+
+_EXTERNAL_REF_PREFIX = "weave:///"
+_INTERNAL_REF_PREFIX = "weave-trace-internal:///"
+# NOTE: This string becomes part of the stabilized value that is hashed. Changing it
+# changes digests and should be treated as a migration-level change *once shipped*.
+_STABILIZED_REF_URI_PREFIX = "weave-anonymous:///_/_/"
+
+__all__ = ["safe_digest", "table_digest_from_row_digests"]
```
andrewtruong, line 29: nit: should be at bottom by convention

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+    # IMPORTANT: This must stay stable over time. We explicitly sort dict keys so that
+    # semantically-identical JSON objects (that differ only in insertion order) hash the same.
+    #
+    # NOTE: We intentionally do *not* change json.dumps separators/whitespace settings here,
+    # because that would change all digests. Keep encoding defaults unless there's a strong
+    # reason to change them (and a migration plan).
+    return _str_digest(json.dumps(json_val, sort_keys=True))
+
+
+E = TypeVar("E")
```
andrewtruong, line 102: nit: by convention, `T`

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def _map_val(obj: E, func: Callable[[E], E]) -> E:
+    if isinstance(obj, dict):
```
andrewtruong, line 106: should this work for any Mapping?

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+        return cast(E, {k: _map_val(v, func) for k, v in obj.items()})
+    if isinstance(obj, list):
```
andrewtruong, line 108: should this work for any Sequence?

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+        return cast(E, [_map_val(v, func) for v in obj])
+    if isinstance(obj, tuple):
```
andrewtruong, line 110: do we have tuple-specific handling in the rest of our code?  IIRC tuple is basically treated like list once serialized

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+    if isinstance(obj, set):
+        # Sets are not supported anywhere in safe_digest; error early and avoid iterating
+        # over an unordered collection.
+        raise TypeError(
+            "Cannot hash set values in digest_builder; convert to a list/tuple first."
+        )
+    return func(obj)
+
+
+def _canonicalize_json_like(obj: Any) -> Any:
```
andrewtruong, line 121: same here

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+
+    This is intentionally ref-unaware: it only deals with deterministic encoding of
+    containers (dict/list/tuple/set) so that stable hashing is possible.
+    """
+    if isinstance(obj, dict):
+        return {k: _canonicalize_json_like(obj[k]) for k in sorted(obj.keys())}
+    if isinstance(obj, list):
+        return [_canonicalize_json_like(v) for v in obj]
+    if isinstance(obj, tuple):
+        return [_canonicalize_json_like(v) for v in obj]
```
andrewtruong, line 132: I see this is specifically a list not a tuple!

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+        raise TypeError(
+            "Cannot hash set values in digest_builder; convert to a list/tuple first."
+        )
+    return obj
+
+
+def _stabilize_refs_in_json_like(val: Any) -> Any:
+    """Return a JSON-like value with Weave ref strings normalized."""
+
+    def mapper(v: Any) -> Any:
```
andrewtruong, line 145: This is not a closure, so it's better to put this at the top level.  Otherwise you are redefining this def (expensive) in a likely hot path of the code

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+    base64_encoded_hash = base64.urlsafe_b64encode(hash_bytes).decode("utf-8")
+    return base64_encoded_hash.replace("-", "X").replace("_", "Y").rstrip("=")
+
+
+def _str_digest(str_val: str) -> str:
+    """Digest a string by hashing its UTF-8 bytes (private helper)."""
+    return _bytes_digest(str_val.encode())
+
+
+def safe_digest(val: Any) -> str:
```
andrewtruong, line 46: maybe not a huge deal, but I think function has the property `safe_digest("x") != safe_digest(json.dumps("x"))`, which is not very intuitive to me (this happens because calling `json.dumps` on a str adds quotes to it.

tssweeney: you are right that this is indeed a true property: `safe_digest('x') != safe_digest('"X"')`. Not sure what you would instead imagine here, as it correctly produces two different digests

weave/trace_server/client_server_common/digest_builder.py
tssweeney: 1. the _/_ is not necessary, but i did like that it maintained consistency with the current patterns. You are correct however that it might just be erroneous. I do like keeping some prefix so that it does not collide with the probably rare circumstance that someone has injected "just a path"
2. collisions: there is nothing wrong with this given our current pattern (you always query by project/digest tuple). However, now that you bring it up, i think that in a possibly future world where objects would be deduped by digest alone in terms of storage, then there would be a problem. I'll think about this

tests/trace_server/test_digest_builder.py
tssweeney: legacy reasons - this is how it works today. i think the original reason was to be closer with artifacts, but i am not sure. Will ask Shawn

tests/trace_server/test_digest_builder.py
tssweeney: yes, i don't like this constraint. I can remove the code itself

weave/trace_server/client_server_common/digest_builder.py
```diff
@@ -0,0 +1,195 @@
...
+
+import base64
+import hashlib
+import json
+from collections.abc import Callable
+from typing import Any, TypeVar, cast
+
+_EXTERNAL_REF_PREFIX = "weave:///"
+_INTERNAL_REF_PREFIX = "weave-trace-internal:///"
+# NOTE: This string becomes part of the stabilized value that is hashed. Changing it
```
zbirenbaum, line 25: The fact that this is part of the hashed value really confuses me.

@tssweeney If I'm reading this code correctly you are hashing everything in the ref except for the entity and project?

 Lets say a user wants to upload DeepSeekR1 weights as an object and they name it MyModelTypo
- if they try to upload the weights as MyModelTypo again, the client can compute it, check it exists and skip
- If I'm reading the code right, if they try and upload the exact same weights but under the name MyModel the client will not see this object exists and will reupload instead of just making a new pointer.
It seems strange to me to have metadata determining the hash used to identify an underlying object.

What are we gaining by including the prefix and all the non entity/project metadata in the hashed value?

Edit: non entity/project lol

tssweeney: I think you raise a good point, (actually two points):
1. you are correct that as implemented, the "name" part of the ref is included. I sort of think that is incorrect (as you imply), and we should remove the name itself.
2. `What are we gaining by including the prefix and all the entity/project metadata in the hashed value?` -> we are explicitly excluding the entity/project, so this is seems to be a misinterpretation

zbirenbaum: sorry that was a typo while editing. I meant to say all the non entity/project metadata. But the question stands as long as any part of the metadata is included in the hash

PR #5736, opened 2025-11-13
chore(weave): Update codegen to support typescript

andrewtruong:
Adds typescript generation support


tools/codegen/generate.py
```diff
@@ -48,7 +48,6 @@ class Config:
     python_output: Path
     package_name: str
     openapi_output: Path
-    node_output: Path | None = None
```
andrewtruong, line 51: Node is deprecated, TypeScript is the new one

PR #5738, opened 2025-11-13
chore(weave): Moves errors to the top of the file so it's more obvious what errors exist

andrewtruong:
Purely cosmetic change, but this should make it easier to know which errors exist.


zbirenbaum (approved):
This is better but this also feels like something which could be a shared file somewhere as well.

andrewtruong:
We could have an errors file

PR #5740, opened 2025-11-13
chore(weave): Validate remote http trace server with pydantic `@validate_call` decorator

andrewtruong:
Simplifies validation by deferring to pydantic.  We are already doing this manually anyways


zbirenbaum (approved):
Such a nice cleanup

PR #5744, opened 2025-11-13
chore(weave): Ensure remaining tests run

tssweeney:
Adds accidentally missed tests dir to the correct shard


andrewtruong:
Thanks.  Why did you remove the mistral tests?  I think we should add those back

tssweeney:
> Thanks. Why did you remove the mistral tests? I think we should add those back

they are captured by the integrations fallback?

andrewtruong:
lgtm, thanks!

PR #5745, opened 2025-11-14
chore(weave): Update stainless tooling

andrewtruong:
This PR updates some of the tooling to generate Stainless SDKs.

We update resources to remove the "V2" nomenclature, and simplify the generate command


tools/codegen/generate.py
andrewtruong, line 1: Simplified -- don't need to checkout main

tools/codegen/openapi.json
andrewtruong, line 1: This is an update following the renaming to remove "V2" from the names, but keeping it as a route

weave/trace_server/reference/server.py
```diff
@@ -26,7 +26,7 @@ def authenticate(
 
 server_dependency = ServiceDependency(service_factory=noop_trace_server_factory)
 trace_service_router = generate_routes(APIRouter(), server_dependency)
-v2_router = generate_v2_routes(APIRouter(prefix="/object"), server_dependency)
+v2_router = generate_v2_routes(APIRouter(prefix="/v2"), server_dependency)
```
andrewtruong, line 29: oops, fixed!

neutralino1: Is this supposed to have `v2` as a prefix? I thought you had removed that.

tools/codegen/generate.py
```diff
@@ -282,22 +282,12 @@ def manage_git_branch(config: Config) -> RepoInfo:
         error(f"Python output directory does not exist: {repo_path}")
 
     run_git(repo_path, ["rev-parse", "--git-dir"])  # Verify git repo
-    run_git(repo_path, ["fetch", "origin", "main"], check=False)
 
-    # Ensure main branch exists
-    if not branch_exists(repo_path, "main"):
-        if not branch_exists(repo_path, "main", remote=True):
```
neutralino1, line 289: Is `branch_exists` still needed? Can it me removed too?

andrewtruong:
It is supposed to be v2, yes.

tools/codegen/generate.py
andrewtruong: Yes, we use it in the creating-a-mirror-branch flow

neutralino1 (approved):

PR #5752, opened 2025-11-14
chore(weave): Remove _set_* handlers from op protocol

andrewtruong:
Just set the fields directly


weave/trace/op.py
```diff
@@ -220,24 +217,6 @@ def setup_dunder_weave_dict(d: WeaveKwargs | None = None) -> WeaveKwargs:
...
-def _set_on_output_handler(func: Op, on_output: OnOutputHandlerType) -> None:
-    if func._on_output_handler is not None:
-        raise ValueError("Cannot set on_output_handler multiple times")
-    func._on_output_handler = on_output
-
-
-def _set_on_finish_handler(func: Op, on_finish: OnFinishHandlerType) -> None:
-    if func._on_finish_handler is not None:
-        raise ValueError("Cannot set on_finish_handler multiple times")
-    func._on_finish_handler = on_finish
```
neutralino1, line 238: Don't we want to keep these safety checks?

andrewtruong: hmm yeah I guess it doesn't hurt to keep it.  We can just remove them from the protocol and make this an implementation detail

PR #5759, opened 2025-11-17
chore(weave): Remove examples as they have moved to the docs site

andrewtruong:
These have moved to the docs site


neutralino1 (commented):
Can we keep an md file in `examples/` with a link to the docs examples directory?
This repo is the entry point for some folks looking for Weave docs, it would be good to keep a path to the examples.

andrewtruong:
We have the `DOCS.md` at root level

PR #5762, opened 2025-11-17
chore(weave): all inserts use async inserts, every 1000ms (not 2s)

gtarpenning:
## Description

This pr 
- moves all clickhouse inserts to be async inserts. `call_parts` is already async and it had a wonderful impact on the db, let's move all inserts to use the buffer flush.
     - includes: `table_rows`, `tables`, `object_versions`, and `feedback`
- set `async_insert_busy_timeout_ms = 1000` (from `2000`)

## Testing

object insert:
image

calls insert:
Screenshot 2025-11-17 at 1 30 31 PM

db impact: positive, hard to know the total impact without seeing larger traffic like in prod.
Screenshot 2025-11-17 at 1 42 29 PM

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -3947,7 +3947,13 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         processed_payload = process_feedback_payload(req)
         row = format_feedback_to_row(req, processed_payload)
         prepared = TABLE_FEEDBACK.insert(row).prepare(database_type="clickhouse")
-        self._insert(TABLE_FEEDBACK.name, prepared.data, prepared.column_names)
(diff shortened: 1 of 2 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -4554,11 +4544,24 @@ def _insert(
...
+            root_span.set_tags(
+                {
+                    "clickhouse_trace_server_batched._insert.table": table,
+                }
+            )
+
+        if self._use_async_insert:
+            if settings is None:
+                settings = {}
+            settings.update(ch_settings.CLICKHOUSE_ASYNC_INSERT_SETTINGS)
```
tssweeney: small nit: i think you should let the user-defined settings override these (in case there is some reason a caller wants to modify the async insert settings)

PR #5764, opened 2025-11-18
feat(weave): Start tracking x-request-id for openai responses api

abraham-leal:
## Description

Start tracking x-request-id for openai responses api

image
image

## Testing

Locally - Weave proj: https://wandb.ai/wandb-smle/test-openai-responses-patching/weave/traces?view=traces_default
Code: https://colab.research.google.com/drive/1RGvUqV4T2bHxGT7rqjpCPgoM2U-WL_y5?usp=sharing

Since the response return is now a weave dict on return, changed test cases

## Docs

This is a minor change that does not require docs.

### DIFF
--- a/tests/integrations/openai/openai_test.py
+++ b/tests/integrations/openai/openai_test.py
@@ -1211,7 +1211,7 @@ def test_openai_responses_tool_calling(client: WeaveClient) -> None:
 
     output = call.output
     assert output["model"] == "gpt-4o-2024-08-06"
-    assert output["object"] == "response" #
+    assert output["object"] == "response"
 
     web_search_call = output["output"][0]
     assert web_search_call["status"] == "completed"


andrewtruong (approved):

tests/integrations/openai/openai_test.py
```diff
@@ -1210,25 +1210,10 @@ def test_openai_responses_tool_calling(client: WeaveClient) -> None:
...
-    response_output_message = output.output[1]
-    search_results = response_output_message.content[0].annotations
-    assert len(search_results) > 0
-
-    tools = call.output.tools
-    web_search_tool = tools[0]
-    assert web_search_tool.type == "web_search_preview"
-    assert web_search_tool.search_context_size == "medium"
-    assert web_search_tool.user_location.type == "approximate"
-    assert web_search_tool.user_location.country == "US"
```
ayulockin, line 1229: LGTM!

My only concern is if we really need to delete these assertions? I don't think they are connected to the request_id and should be checked for?

abraham-leal: thanks, fixed!

PR #5773, opened 2025-11-20
feat(weave): Save and use prompts and template vars on LLMStructuredCompletionModels

jwlee64:
## Description

Adds support for MessagesPrompt objects in LLMStructuredCompletionModel by introducing a new `prompt` field in LLMStructuredCompletionModelDefaultParams. This allows models to reference a published MessagesPrompt object instead of using inline message templates.

When a prompt is specified, it takes precedence over messages_template, and the model passes the prompt reference and template variables to the completions endpoint for resolution. This enables more flexible and reusable prompt templates in evaluation workflows.

## Testing

This PR was tested with:

- Unit tests for LLMStructuredCompletionModel with prompt references
- Tests verifying prompt precedence over messages_template
- Integration tests for evaluation workflows using LLM models with prompt templates
- Serialization tests for library objects including Evaluation, Scorer, and LLMStructuredCompletionModel

PR #5774, opened 2025-11-20
feat(weave): Add prompts to LLMAsAJudgeScorer -> Monitors

jwlee64:
## Description

This PR enhances the `LLMAsAJudgeScorer` to support using a reference to a `MessagesPrompt` object instead of just an inline string prompt. This allows for more complex, reusable prompts with multiple messages and roles.

Key changes:

- Added `scoring_prompt_ref` field to `LLMAsAJudgeScorer` that can accept either a URI string or a resolved `MessagesPrompt` object
- Implemented `build_scoring_messages()` method to handle both prompt types
- Enhanced the `score()` method to use the new prompt reference functionality
- Added comprehensive tests for the new functionality
- Updated `MessagesPrompt` and `StringPrompt` to properly pass kwargs to parent class

This change maintains backward compatibility with existing string-based prompts while enabling more sophisticated prompt structures.

## Testing

The changes were tested by updating test cases to reflect the new functionality, including updating digests and expected values in the serialization tests.


jwlee64:
* **#5774** <a href="https://app.graphite.com/github/pr/wandb/weave/5774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/5774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5773** <a href="https://app.graphite.com/github/pr/wandb/weave/5773?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/prompt/prompt.py
```diff
@@ -148,8 +148,8 @@ def format_message_with_template_vars(message: dict, **kwargs: Any) -> dict:
 class MessagesPrompt(Prompt):
     messages: list[dict] = Field(default_factory=list)
 
-    def __init__(self, messages: list[dict]):
```
jwlee64, line 151: Was getting an error `TypeError: MessagesPrompt.__init__() got an unexpected keyword argument 'name'` passing in name is a field on `Prompt` accepting kwargs fixes it

tests/trace/test_evaluation_ref_get.py
```diff
@@ -51,3 +52,58 @@ def test_publish_and_load_evaluation(client):
...
+    # Publish and retrieve the scorer - weave.get() auto-resolves the ref
+    published_scorer_ref = weave.publish(scorer)
+    gotten_scorer = weave.get(published_scorer_ref.uri())
+
+    # Verify the round-trip works - scoring_prompt_ref is now the resolved MessagesPrompt
+    assert isinstance(gotten_scorer, LLMAsAJudgeScorer)
+    assert isinstance(gotten_scorer.scoring_prompt_ref, MessagesPrompt)
+    assert gotten_scorer.scoring_prompt is None
+    # Verify the resolved prompt has the expected messages
+    assert len(gotten_scorer.scoring_prompt_ref.messages) == 1
```
tssweeney: the icing on the cake here would be if you could actually invoke the scorer with vars here! Then you could just call that directly from the scoring worker

weave/scorers/llm_as_a_judge_scorer.py
```diff
@@ -1,22 +1,148 @@
...
+        scoring_prompt: A prompt string template with {variable} placeholders (optional if scoring_prompt_ref is provided)
+        scoring_prompt_ref: A reference to a MessagesPrompt object, either as a URI string
+            or the resolved MessagesPrompt object (optional, takes precedence over scoring_prompt)
+    """
+
     model: LLMStructuredCompletionModel
-    scoring_prompt: str
+    scoring_prompt: str | None = None
+    # Accept both string (ref URI) and MessagesPrompt (resolved ref from weave.get())
+    scoring_prompt_ref: str | MessagesPrompt | None = None
```
tssweeney, line 83: having both `scoring_prompt` and `scoring_prompt_ref` is very odd and different than existing patterns. In other places, the property would be a `MessagesPrompt` type - then at serialization time it would turn into a ref and deserialization it would turn back into a MessagesPrompt. the fact that you are dealing with this here is a sign that something else is more fundamentally wrong. I think you should be able to pull this off by just having the scoring_prompt field be a union of string | MessagesPrompt. Let's chat/pair if this is not working easily.

weave/scorers/llm_as_a_judge_scorer.py
```diff
@@ -1,22 +1,148 @@
...
+                prompt_obj = self.scoring_prompt_ref
+            elif server is not None:
+                # Direct server access - used by the scoring worker to avoid
+                # reinitializing a client for each project being scored
+                prompt_obj = _load_prompt_from_ref(self.scoring_prompt_ref, server)
+            else:
+                # Client-based resolution for normal SDK usage
+                client = get_weave_client()
+                if client is None:
+                    raise ValueError(
```
tssweeney, line 115: this is a smell. we should not need to initialize a tracing client (traces ops) in order to resolve this ref

PR #5775, opened 2025-11-20
chore(weave): Add attribute tracking to node sdk

tssweeney:
100% vibe coded implementation of `attribute` support in node sdk

Prompt 1: 
```
Notice how the weave python SDK supports a context manager adding attributes to calls with @api.py (227-245)  which is consumed during call creation here: @op.py (322-335) . Please add similar support for the typescript version, which already has an attributes param here: @weaveClient.ts (707-718) . Probably need to add the instrumentation here: @op.ts (151-162).
```

Prompt 2:
```
One thing: won't the async nature of this mess things up? can the attrbiutes clobber eachother? Please add a test
```


tssweeney:
closing in favor of: https://github.com/wandb/weave/pull/5780

PR #5776, opened 2025-11-21
chore(weave): Fix OpenAI Agents SDK rollups

andrewtruong:
Fixes an issue with rollups.  Now if a guardrail is triggered, the rollups will be correct

## After
image

## Before
image

### DIFF
--- a/weave/integrations/openai_agents/openai_agents.py
+++ b/weave/integrations/openai_agents/openai_agents.py
@@ -67,8 +67,6 @@ class WeaveTracingProcessor(TracingProcessor):  # pyright: ignore[reportGeneralT
         self._trace_calls: dict[str, Call] = {}
         self._span_calls: dict[str, Call] = {}
         self._ended_traces: set[str] = set()
-        self._span_parents: dict[str, str] = {}
-        self._pushed_calls: set[str] = set()
 
     def on_trace_start(self, trace: Trace) -> None:
         """Called when a trace starts."""
@@ -96,7 +94,6 @@ class WeaveTracingProcessor(TracingProcessor):  # pyright: ignore[reportGeneralT
         # Push the trace call onto the call context stack so nested spans
         # and OpenAI calls can find it as their parent
         call_context.push_call(trace_call)
-        self._pushed_calls.add(trace_call.id)
 
     def on_trace_end(self, trace: Trace) -> None:
         """Called when a trace ends."""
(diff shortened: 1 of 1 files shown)


zbirenbaum:
@andrewtruong Can you provide some context on exactly what the issue being fixed was and what the repro case is?

andrewtruong:
@zbirenbaum this creates the Agents calls at span start (instead of span end) and keeps track of them in the tracing processor.  We do this so that child calls from the Responses API always know who their parents are.  Previously, they could be orphaned.

There is also more specific handling for exceptions so that they show up as errors instead of successes.  I think this is more faithful to what users expect to see

zbirenbaum:
> @zbirenbaum this creates the Agents calls at span start (instead of span end) and keeps track of them in the tracing processor. We do this so that child calls from the Responses API always know who their parents are. Previously, they could be orphaned.

This makes sense. I feel like I brought this up last time we investigated and there was a reason we didn't do this in the first place, since it seems like the natural solution

Regardless if it works fine then take it out of draft mode, I'll try it out and approve

andrewtruong:
I'm not sure why we didn't do this originally.  Absent that detail I think this is more faithful to what the user expects!

weave/integrations/openai_agents/openai_agents.py
```diff
@@ -303,6 +310,38 @@ def on_span_start(self, span: Span) -> None:
         span_type = _call_type(span)
         parent_call = self._get_parent_call(span)
 
+        # For Response spans, create the call immediately (even without input data)
+        # so it's on the stack when the OpenAI call happens. We'll update the inputs
+        # in on_span_end when we have the full data.
```
zbirenbaum, line 315: How do you update inputs from on_span_end?

Our start call is immutable once stored. This was a major issue while working on realtime.

weave/integrations/openai_agents/openai_agents.py
```diff
@@ -372,14 +381,20 @@ def on_span_end(self, span: Span) -> None:
...
+            exception = RuntimeError(error_message)
         elif log_data["error"]:
             output["error"] = log_data["error"]
 
-        # Finish the call with the collected data
-        wc.finish_call(span_call, output=output)
+        # Finish the call with the collected data, even if interrupted/cancelled
+        # This ensures all calls are properly finalized
+        # Note: finish_call automatically pops the call from the stack
+        wc.finish_call(span_call, output=output, exception=exception)
```
zbirenbaum, line 397: I don't see anything here that would update the inputs

PR #5779, opened 2025-11-21
chore(weave): Add missing test shards

andrewtruong:
Previously these shards were not running -- not good!


.github/workflows/nightly-tests.yaml
```diff
@@ -41,13 +41,19 @@ jobs:
             "dspy",
             "groq",
             "google_genai",
+            # "huggingface",
```
tssweeney, line 44: why commented out?

PR #5780, opened 2025-11-21
feat(weave): typescript sdk withAttributes

shawnlewis:
This PR adds a new api to the Typescript SDK: `withAttributes`. This allows the user to specify key-value pairs that are added to the attributes of all children calls within a given context. 

Docs: https://wandb.atlassian.net/browse/DOCS-1915

### DIFF
--- a/sdks/node/src/settings.ts
+++ b/sdks/node/src/settings.ts
@@ -1,11 +1,7 @@
-export type SettingsInit =
-  | Settings
-  | {
-      printCallLink?: boolean;
-      globalAttributes?: Record<string, any>;
-    }
-  | undefined;
-
+export interface SettingsInit {
+  printCallLink?: boolean;
+  globalAttributes?: Record<string, any>;
+}
 export class Settings {
   constructor(
     private printCallLink: boolean = true,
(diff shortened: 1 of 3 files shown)


neutralino1:
@dbrian57 We're adding this API to the Typescript SDK, can you amend these docs to include a TS snippet?
https://docs.wandb.ai/weave/guides/tools/attributes#define-and-log-attributes

Note that this API only supports call attributes, not global attributes.

tssweeney:
Linted, added doc link, added summary - should be good for merge on CI complete

sdks/node/src/settings.ts
```diff
@@ -1,5 +1,16 @@
+export type SettingsInit =
+  | Settings
+  | {
+      printCallLink?: boolean;
+      globalAttributes?: Record<string, any>;
+    }
```
neutralino1: Isn't this redundant? That's the same signature as `Settings` below.

shawnlewis: This is just a bare object, so its not required to have the additional methods the Settings object has. The type checker won't allow to pass {globalAttributes: ...} to weave.init without this. You'd have to do (new Settings(...)) which we don't export and also doesn't have named arguments.

shawnlewis: But there's probably a slightly better way to do this. Hang on...

PR #5784, opened 2025-11-23
fix(weave): fix type signature for override method

zbirenbaum:
Minor fix because the pydantic version change modified the signature of these methods.


andrewtruong:
What changed?

PR #5788, opened 2025-11-25
feat(weave): parse Bedrock OTEL events for agent context

NiWaRe:
## Description

WS AgentCore’s docs explicitly recommend wiring agents into OpenTelemetry via the AWS Distro for OpenTelemetry (ADOT). The [“Configure observability” section of the AgentCore developer guide](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/observability-configure.html?utm_source=openai) walks through enabling OTEL exporters for AgentCore runtimes so every agent/tool call is traced via the same opentelemetry-instrument/ADOT pipeline - **this PR makes Weave capable of parsing all of these OTEL spans!** 

* AWS Bedrock agent traces arrive via the Python auto-instrumentation path ([opentelemetry-python-contrib](https://github.com/open-telemetry/opentelemetry-python-contrib)’s `opentelemetry-instrument` CLI, plus the AWS distro defaults) so user prompts, assistant responses, and tool metadata are emitted as OTEL events like `gen_ai.user.message` and `gen_ai.choice`
* We already persisted those events under `attributes.otel_span`, but `inputs_dump`/`output_dump` ignored them. 
* This change folds OTEL span events into an `events.*` namespace and extends our constant maps so Bedrock/AWS spans parsed via [opentelemetry-python-contrib](https://github.com/open-telemetry/opentelemetry-python-contrib) + aws-otel-python now surface full context in the Weave UI.

## Testing
Based on [this sample trace](https://wandb.ai/wandb-smle/bedrock-agentcore-demo/weave/traces?view=traces_default&sort=%5B%7B%22field%22%3A%22started_at%22%2C%22sort%22%3A%22desc%22%7D%5D&peekPath=%2Fwandb-smle%2Fbedrock-agentcore-demo%2Fcalls%2Fb9088ebfe9a233da%3FdescendentCallId%3Db9088ebfe9a233da). 

1. **Replay the sample trace**  
   ```bash
   uv run python - <<'PY'
   import weave
   client = weave.init("wandb-smle/bedrock-agentcore-demo")
   call = client.get_call("b9088ebfe9a233da")
   print(call.inputs)   # should now include events.gen_ai.user.message
   print(call.outputs)  # should now include events.gen_ai.choice
   print(call.attributes["system"])
   PY
   ```
   You should see the user prompt, assistant reply, and agent/tool metadata in the resulting dumps.

2. **Sanity-check constants**  
   Confirm that the new keys appear in `weave/trace_server/opentelemetry/constants.py` (`events.gen_ai.*`, `system_prompt`, `gen_ai.agent.*`, etc.) so future OTEL spans will hit the new codepaths.

3. **Run the focused unit test** 
   ```bash
   uv run pytest tests/trace_server/test_opentelemetry.py -k bedrock_agent_event_input_output_and_tools
   ```

I have done everything except step 3.

### DIFF
--- a/tests/trace_server/test_opentelemetry.py
+++ b/tests/trace_server/test_opentelemetry.py
@@ -1112,8 +1112,10 @@ class TestSemanticConventionParsing:
 
         usage = get_weave_usage(attributes) or {}
 
+        # Verify individual tokens are parsed
         assert usage.get("input_tokens") == 10
         assert usage.get("output_tokens") == 20
+        # Verify total_tokens is calculated when not provided
         assert usage.get("total_tokens") == 30
 
     def test_opentelemetry_usage_output_tokens_with_explicit_total(self):
@@ -1128,8 +1130,10 @@ class TestSemanticConventionParsing:
 
         usage = get_weave_usage(attributes) or {}
 
+        # Verify individual tokens are parsed
         assert usage.get("input_tokens") == 10
         assert usage.get("output_tokens") == 20
+        # Verify explicit total_tokens is used instead of calculated value
         assert usage.get("total_tokens") == 35
 
     def test_opentelemetry_cost_calculation(self, client: weave_client.WeaveClient):
--- a/weave/trace_server/opentelemetry/attributes.py
+++ b/weave/trace_server/opentelemetry/attributes.py
@@ -38,37 +38,19 @@ def _insert_event(events_tree: dict[str, Any], name: str, attributes: dict[str, 
     leaf = parts[-1]
     existing = current.get(leaf)
     if existing is None:
-        current[leaf] = [attributes]
+        current[leaf] = [copy.deepcopy(attributes)]
     elif isinstance(existing, list):
-        existing.append(attributes)
+        existing.append(copy.deepcopy(attributes))
     elif isinstance(existing, dict):
-        existing.setdefault("__events__", []).append(attributes)
+        existing.setdefault("__events__", []).append(copy.deepcopy(attributes))
     else:
-        current[leaf] = [attributes]
+        current[leaf] = [copy.deepcopy(attributes)]
 
 
-def _merge_event_trees(dest: dict[str, Any], src: dict[str, Any]) -> dict[str, Any]:
-    """Recursively merge two event trees."""
-    for key, value in src.items():
-        if key not in dest:
-            dest[key] = copy.deepcopy(value)
-            continue
-        if isinstance(value, dict) and isinstance(dest[key], dict):
-            _merge_event_trees(dest[key], value)
-            continue
-        if isinstance(value, list) and isinstance(dest[key], list):
-            dest[key].extend(copy.deepcopy(value))
-            continue
-        dest[key] = copy.deepcopy(value)
-    return dest
-
-
-def _augment_attributes_with_events(
-    attributes: dict[str, Any], events: list[SpanEvent]
-) -> dict[str, Any]:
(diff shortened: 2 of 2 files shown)


tests/trace_server/test_opentelemetry.py
```diff
@@ -1103,31 +1103,87 @@ def test_opentelemetry_usage_output_tokens_extraction(self):
...
+                "llm.usage.total_tokens": 35,
             }
         )
 
         usage = get_weave_usage(attributes) or {}
 
-        # Verify individual tokens are parsed
         assert usage.get("input_tokens") == 10
         assert usage.get("output_tokens") == 20
-        # Verify explicit total_tokens is used instead of calculated value
```
zbirenbaum, line 1137: I actually think this comment and the ones above are useful. It's a bit confusing otherwise what the intended behavior is.

weave/trace_server/opentelemetry/attributes.py
```diff
@@ -24,6 +25,75 @@ class SpanEvent(dict):
...
 
+def _insert_event(events_tree: dict[str, Any], name: str, attributes: dict[str, Any]) -> None:
+    """Insert an OTEL span event into the nested events tree."""
+    parts = name.split(".")
+    current = events_tree
+    for part in parts[:-1]:
+        next_value = current.get(part)
+        if not isinstance(next_value, dict):
+            next_value = {}
+            current[part] = next_value
```
zbirenbaum, line 36: I think you can use defaultdict here to set it if it doesn't exist

weave/trace_server/opentelemetry/attributes.py
```diff
@@ -101,14 +171,16 @@ def get_weave_usage(attributes: dict[str, Any]) -> dict[str, Any]:
...
+    return parse_weave_values(augmented_attributes, INPUT_KEYS)
 
 
-# Pass events here even though they are unused because some libraries put output in event attributes
-def get_weave_outputs(_: list[SpanEvent], attributes: dict[str, Any]) -> dict[str, Any]:
-    return parse_weave_values(attributes, OUTPUT_KEYS)
+# Some providers (e.g. Bedrock agents) encode outputs inside span events
+def get_weave_outputs(events: list[SpanEvent], attributes: dict[str, Any]) -> dict[str, Any]:
+    augmented_attributes = _augment_attributes_with_events(attributes, events)
+    return parse_weave_values(augmented_attributes, OUTPUT_KEYS)
```
zbirenbaum: This isn't the place to do that I think. By doing it in both places I think you risk duplication since this modifies in place. Let me think about this, you are repeating a lot of work. The reason I passed the span events as an `_` was because I wanted to create a mapping of events to inputs and outputs.

We probably want to pass the events to 'parse_weave_values'. That way if something like `gen_ai.prompt` shows up in events instead of the attributes (I think there are cases of this between different specs/providers) we are able to parse it for free instead of having to add 'events.*' to everything

We definitely don't want to modify the attributes from the original payload either when storing

tests/trace_server/test_opentelemetry.py
NiWaRe: Yes that makes sense overlooked that edit from Cursor - back in!

weave/trace_server/opentelemetry/attributes.py
NiWaRe: Yes true better not to mutate the attributes and directly pass events to parse_weave_values. @zbirenbaum what do you think about the updated code? If this is too big of a change for fast checks we can also only add the contants for now and leave the events parsing for another PR?

PR #5790, opened 2025-11-25
chore(weave): Reraise 4xx errors

andrewtruong:
Re-raises 4xx errors from the backend instead of showing 500.  This is more clear to the end user what is happening

### DIFF
--- a/weave/trace_server/errors.py
+++ b/weave/trace_server/errors.py
@@ -377,4 +377,4 @@ def _get_transport_server_error_status_code(exc: Exception) -> int:
 
 
 def _get_transport_server_error_message(exc: TransportServerError) -> dict[str, Any]:
-    return {"reason": f"{exc.code} error"}
+    return {"reason": f"{exc.code} Error"}


weave/trace_server/errors.py
```diff
@@ -251,9 +260,18 @@ def _setup_common_errors(self) -> None:
...
         self.register(TransportQueryError, 403, lambda exc: {"reason": "Forbidden"})
+        self.register(
+            TransportServerError,
+            lambda exc: (
+                exc.code
+                if isinstance(exc, TransportServerError)
+                and exc.code
+                and 400 <= exc.code < 500
+                else 500
+            ),
```
gtarpenning: could probably make this a helper

weave/trace_server/errors.py
```diff
@@ -183,7 +201,7 @@ def handle_exception(self, exc: Exception) -> ErrorWithStatus:
         if definition:
             error_content = definition.formatter(exc)
```
gtarpenning, line 202: won't this leak the actual error?

gtarpenning: we want the error code, but in TransportError case i'm not sure we should be sending the error text as well. Reason being its totally opaque, some random gql issue. We could reformat it / clean it up I guess, if we are willing to bake in some custom error handling here

weave/trace_server/errors.py
andrewtruong: done

weave/trace_server/errors.py
andrewtruong: We could, but I think it's outside of the scope of this PR.

Why did you write it this way originally?

gtarpenning: wdym? Each error is registered with a handler, for errors we don't want to leak to the user we override it with `lambda exc: {"reason": "Temporary backend error"}` for example. 

In this pr we do: 
`lambda exc: {"reason": str(exc)},` which will actually send out the real error message. We don't need to do that, we could for example do: 

```python
lambda exc: {"reason": str(_get_transport_server_error_status_code(exc)) + " error"},
```

andrewtruong: Oh I see, sorry I misunderstood from the line that you highlighted.  Yes that makes sense -- fixed.

PR #5801, opened 2025-12-02
chore(weave): Dont template assistant messages

jwlee64:
## Description

Dont template assistant messages, sometimes they return json

## Testing

How was this PR tested?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -37,7 +37,7 @@ def patch_kafka_producer():
     If a test needs to test the Kafka producer, they should orride this patch explicitly.
     """
     with patch(
-        "weave.trace_server.clickhouse_trace_server_batched.KafkaProducer.from_env",
+        "weave.trace_server.kafka.KafkaProducer.from_env",
         return_value=MagicMock(),
     ):
         yield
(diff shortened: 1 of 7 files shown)


tests/trace_server/test_llm_completion.py
```diff
@@ -1260,6 +1260,54 @@ def mock_obj_read(req):
...
 
+    def test_resolve_and_apply_prompt_skips_assistant_messages(self):
+        """Test that template variable substitution is skipped for assistant messages."""
+        from weave.trace_server.llm_completion import resolve_and_apply_prompt
+
+        def mock_obj_read(req):
+            raise NotImplementedError("Should not be called")
+
+        # Messages including an assistant message with template-like content
+        user_messages = [
```
bcsherma: I feel like this should just be called messages

tests/trace_server/test_llm_completion.py
```diff
@@ -1260,6 +1260,54 @@ def mock_obj_read(req):
...
+            {"role": "system", "content": "You are {assistant_name}."},
+            {"role": "user", "content": "Hello {user_name}!"},
+            {"role": "assistant", "content": "Hello! I see you mentioned {user_name}."},
```
bcsherma: the message we are filtering out here doesn't contain json. I am confused about why this constraint applies to assistant messages but not other messages

weave/trace_server/llm_completion.py
```diff
@@ -210,7 +210,11 @@ def resolve_and_apply_prompt(
     # Step 3: Apply template variable substitution (if provided)
     if template_vars and combined_messages:
         combined_messages = [
-            format_message_with_template_vars(msg, **template_vars)
+            (
+                format_message_with_template_vars(msg, **template_vars)
+                if msg.get("role") != "assistant"
+                else msg
+            )
```
bcsherma, line 219: this may merit a comment explaining why the assistant messages need to be filtered out

bcsherma (commented):
I have some qs 🕵🏻

tests/trace_server/test_llm_completion.py
jwlee64: this is more for the test.
But assistant messages dont return parameters, just normal text, can clarify it though

PR #5803, opened 2025-12-02
refactor(weave): add dd set tags helpers

gtarpenning:
## Description

Add helpers for setting datadog tags on root and current spans. 

## Testing

qa

### DIFF
--- a/weave/trace_server/datadog.py
+++ b/weave/trace_server/datadog.py
@@ -1,0 +1,47 @@
+"""Datadog integration utilities for trace server."""
+
+import logging
+
+import ddtrace
+
+logger = logging.getLogger(__name__)
+
+
+def set_root_span_dd_tags(tags: dict[str, str | float | int]) -> None:
+    """Set tags on the current root span for Datadog tracing.
+
+    Args:
+        tags: Dictionary of tags to set on the root span.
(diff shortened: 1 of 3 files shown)


weave/trace_server/datadog.py
```diff
@@ -0,0 +1,47 @@
...
+        >>> set_root_span_dd_tags({"cache.hit": "true", "cache.key": "abc123"})
+    """
+    root_span = ddtrace.tracer.current_root_span()
+    if root_span is None:
+        logger.debug("No root span")
+    else:
+        root_span.set_tags(tags)
+
+
+def set_current_span_dd_tags(tags: dict[str, str | float | int]) -> None:
```
chance-wnb, line 31: Does it make sense to also use this here: https://github.com/wandb/core/blob/7729176db9e65113cc2f6f6038e51db2e52145f5/services/weave-trace/src/request_validators.py#L54

gtarpenning: yeah but unfortunately that's in the part of the trace server that's not in the python package, probably want to keep it separate.

chance-wnb: Ok, sounds good. Thanks for checking.

PR #5805, opened 2025-12-03
fix(weave): Fix clickhouse import issue

andrewtruong:
Defers imports to avoid the client taking a dependency on clickhouse-connect

This is unfortunate coupling and was what the imports looked like previously.  It should be fixed at some point.  This hotfix is just for expediency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -329,7 +329,7 @@ disallow_untyped_defs = true
 disallow_untyped_calls = true
 
 [tool.bumpversion]
-current_version = "0.52.21-dev0"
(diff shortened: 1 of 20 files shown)


weave/trace_server/errors.py
```diff
@@ -252,6 +250,14 @@ def _setup_common_errors(self) -> None:
         )
 
         # ClickHouse errors
+        # It's unfortunate we have to defer imports here because the client also imports from this file.
```
tssweeney, line 253: very sus. i tracked down the usage and I think that is a bug. we should be breaking these dep chains instead

PR #5809, opened 2025-12-04
feat(weave): Update Leaderboard schema to support overhaul features

zbirenbaum:
Schema for new editable leaderboards

 - Towards [WB-29573](https://wandb.atlassian.net/browse/WB-29573)
 
docs are not required

[WB-29573]: https://wandb.atlassian.net/browse/WB-29573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/interface/builtin_object_classes/generated/generated_builtin_object_class_schemas.json
+++ b/weave/trace_server/interface/builtin_object_classes/generated/generated_builtin_object_class_schemas.json
@@ -803,6 +803,18 @@
           "default": null,
           "title": "Scorer Name"
         },
+        "display_name": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "title": "Display Name"
+        },
         "summary_metric_path": {
           "anyOf": [
             {
--- a/weave/trace_server/interface/builtin_object_classes/saved_view.py
+++ b/weave/trace_server/interface/builtin_object_classes/saved_view.py
@@ -53,6 +53,7 @@ class ObjectConfig(BaseModel):
 class DynamicLeaderboardColumnConfig(BaseModel):
     evaluation_object_ref: base_object_def.RefStr | None = Field(default=None)
     scorer_name: str | None = Field(default=None)
+    display_name: str | None = Field(default=None)
     summary_metric_path: str | None = Field(default=None)
     should_minimize: bool | None = Field(default=None)
     deselected: bool | None = Field(


weave/trace_server/interface/builtin_object_classes/dynamic_leaderboard.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class AggregationMethod(str, Enum):
+    LATEST = "latest"
+    AVERAGE = "average"
+
+
+class ObjectVersionGroup(BaseModel):
+    label: str  # label for the combination of the groups
+    versionRefs: list[str]  # Digests of the versions to group
```
tssweeney: Nit: use snake case in python

weave/trace_server/interface/builtin_object_classes/dynamic_leaderboard.py
```diff
@@ -0,0 +1,40 @@
...
+class DynamicLeaderboardColumnConfig(BaseModel):
+    evaluation_object_ref: base_object_def.RefStr
+    scorer_name: str
+    summary_metric_path: str
+    should_minimize: bool
+    deselected: bool  # If True, this metric is excluded from the leaderboard
+
+
+class DynamicLeaderboard(base_object_def.BaseObject):
+    calls_query: str
```
tssweeney: Why is this a string and not a calls query type object

weave/trace_server/interface/builtin_object_classes/dynamic_leaderboard.py
tssweeney: Also are these refs or digests? If digests, don't use the term ref

weave/trace_server/interface/builtin_object_classes/dynamic_leaderboard.py
zbirenbaum: It's a combination of GridFilterModel and ExpandColRefs from the frontend. Can update this, but just to check does the type you are referring to include expand cols?

weave/trace_server/interface/builtin_object_classes/dynamic_leaderboard.py
zbirenbaum: Initially they were going to be refs but now they are either a list of digests or "*" to wildcard. Fixed the snake case as well. Added a base_ref field to this for the unversioned object

PR #5814, opened 2025-12-04
fix(weave): correct object deleted error handling

gtarpenning:
## Description

fixes nasty dependency chain between the trace server and the sdk. 

## Testing

updates tests

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -419,16 +419,18 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
             for idx in obj_id_idx_map[result.object_id]:
                 calls[idx][0].op_name = op_ref_uri
 
-        batch = [
-            item
-            for start_call, end_call in calls
-            for item in (
-                tsi.CallBatchStartMode(req=tsi.CallStartReq(start=start_call)),
-                tsi.CallBatchEndMode(req=tsi.CallEndReq(end=end_call)),
+        # Convert calls to CH insertable format and then to rows for batch insertion
+        batch_rows = []
+        for start_call, end_call in calls:
+            ch_start = _start_call_for_insert_to_ch_insertable_start_call(
+                start_call, self
             )
-        ]
-        # TODO: Actually populate the error fields if call_start_batch fails
-        self.call_start_batch(tsi.CallCreateBatchReq(batch=batch))
+            ch_end = _end_call_for_insert_to_ch_insertable_end_call(end_call, self)
+            batch_rows.append(_ch_call_to_row(ch_start))
+            batch_rows.append(_ch_call_to_row(ch_end))
+
+        # Insert directly without async_insert for OTEL calls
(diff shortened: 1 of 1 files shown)


weave/trace/vals.py
```diff
@@ -819,7 +826,7 @@ def make_trace_obj(
             parent=parent,
         )
 
-    if extra:
+    if extra and not isinstance(val, DeletedRef):
```
gtarpenning: fixes a type error

weave/trace/vals.py
```diff
@@ -819,7 +828,7 @@ def make_trace_obj(
             parent=parent,
         )
 
-    if extra:
+    if extra and not isinstance(val, (DeletedRef, ObjectRef, OpRef, CallRef)):
```
gtarpenning: can't index into val if its one of these... why was this working before?

weave/trace/vals.py
```diff
@@ -780,10 +781,16 @@ def make_trace_obj(
                 )
             )
             val = from_json(read_res.obj.val, project_id, server)
-        except ObjectDeletedError as e:
-            # encountered a deleted object, return DeletedRef, warn and continue
-            val = DeletedRef(ref=new_ref, deleted_at=e.deleted_at, error=e)
-            logger.warning(f"Could not read deleted object: {new_ref}")
+        except httpx.HTTPStatusError as e:
```
tssweeney, line 784: the fact that `server.obj_read` raises an `httpx.HTTPStatusError` exception is an implementation detail of the trace server itself and the way that the remotehttptraceserver works. I'd recommend either:
1. putting this try catch inside the remotehttptraceserver and returning a synthetic objectschema with `deleted_at`; or
2. raising a specific error in the server that can be caught here.

weave/trace/refs.py
```diff
@@ -327,7 +325,7 @@ def parse_uri(uri: str) -> CallRef:
 class DeletedRef(Ref):
     ref: Ref
     deleted_at: datetime
-    error: ObjectDeletedError
+    error: httpx.HTTPStatusError
```
tssweeney: similar to the comment below, this should not be an http error, but a custom error. honestly, sort of odd for the deletedref to have an error anyway

weave/trace/vals.py
gtarpenning: okay bad choice to use `httpx.HTTPStatusError`, but this is basically option 2 right? The trace server is raising a specific error, that we are trying to identify here. We can blanket catch `Exception` and do the existing logic, I just thought that was less specific than at least narrowing the exception to an http exception. I do understand what you are saying though. 

Option 1 feels even worse tbh, i'm not 100% sure I understand the philosophical purpose of remotehttptraceserver but I think I would be surprised if I saw type conversion to a synthetic object schema in there. I guess if it's purpose is as broad as general middlewear to process from the request layer down to our client it's appropriate...

weave/trace/refs.py
gtarpenning: yeah this is a yagni situation. removing

PR #5822, opened 2025-12-10
chore(weave): Remove BaseModel from imperative evals

andrewtruong:
We don't need to use BaseModel here.  Using a regular class is fine, and less expensive


andrewtruong:
It does, but it's overkill imo.  We don't need it.  If you feel safer, we can add guards in the init

PR #5847, opened 2025-12-15
feat(weave): Add option to show storage size in get_calls

andrewtruong:
Adds new params to `get_calls` to expose how much storage is being used.

```py
calls = client.get_calls(include_total_storage_size=True, include_storage_size=True)
```

To get just the storage data:
```py
calls = client.get_calls(
    include_total_storage_size=True,
    include_storage_size=True,
    columns=["id", "storage_size_bytes", "total_storage_size_bytes"],
)
```

### DIFF
--- a/weave/trace/call.py
+++ b/weave/trace/call.py
@@ -75,6 +75,11 @@ class Call:
     wb_run_step: int | None = None
     wb_run_step_end: int | None = None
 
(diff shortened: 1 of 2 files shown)


andrewtruong:
dupe of https://github.com/wandb/weave/pull/5852/changes

PR #5852, opened 2025-12-16
chore(weave): Expose storage parameters in weave_client.get_calls()

mscavezze-cw:
## Description

Expose a couple of existing parameters from `weave_client.server.calls_query_stream()` in `weave_client.get_calls()`:
- `include_storage_size`
- `include_total_storage_size`

Fixes WB-29716

## Testing

Wrote and executed new tests in `test_weave_client.py`, which performs end-to-end integration testing.

### DIFF
--- a/tests/trace/test_weave_client.py
+++ b/tests/trace/test_weave_client.py
@@ -1744,7 +1744,7 @@ def test_get_calls_storage_size_with_filter(client):
         client.get_calls(
             filter=tsi.CallsFilter(op_names=[call0.op_name]),
             include_storage_size=True,
-            include_total_storage_size=True
+            include_total_storage_size=True,
         )
     )
     assert len(calls_filtered) == 2
@@ -1830,7 +1830,7 @@ def test_get_calls_storage_size_values(client, clickhouse_client):
                 == client_call.total_storage_size_bytes
             )
             assert server_call.storage_size_bytes is not None
-            
+
             # total_storage_size_bytes is only set for root calls (parent_id is None)
             # For child calls, it is intentionally None
             expect_total_storage_size_bytes = server_call.parent_id is None


mscavezze-cw:

tests/trace/test_weave_client.py
```diff
@@ -1720,6 +1720,140 @@ def test_summary_tokens_cost_sqlite(client):
...
+    expected_calls = _setup_calls_for_storage_size_test(client)
+    call0 = expected_calls[0]
+
+    # Test that parameters can be combined with other parameters
+    calls_filtered = list(
+        client.get_calls(
+            filter=tsi.CallsFilter(op_names=[call0.op_name]),
+            include_storage_size=True,
+            include_total_storage_size=True,
+            limit=2,
```
chance-wnb, line 1763: If we are filtering by call0.op_name, even if we don't put `limit=2` it also returns 2 calls right? so it does not prove the calls query is limited.

Super nit: It is not that clear enough to say 2 is a number that can be used to limit the query, because I don't know what the expected_count is unless I check.  Adding assert expected_count > 2 might help.

tests/trace/test_weave_client.py
```diff
@@ -1720,6 +1720,140 @@ def test_summary_tokens_cost_sqlite(client):
...
+    # Verify storage size fields and compare values between server and client calls
+    for server_call in server_calls:
+        # Verify storage size fields are present on server calls
+        assert hasattr(server_call, "storage_size_bytes")
+        assert hasattr(server_call, "total_storage_size_bytes")
+
+        # Verify that storage size values match between server and client calls
+        if server_call.id and server_call.id in client_calls_by_id:
+            client_call = client_calls_by_id[server_call.id]
+            assert server_call.storage_size_bytes == client_call.storage_size_bytes
```
chance-wnb, line 1827: I would recommend adding `server_call.storage_size_bytes is not None` and `server_call.total_storage_size_bytes is not None` check as well

tests/trace/test_weave_client.py
```diff
@@ -1720,6 +1720,140 @@ def test_summary_tokens_cost_sqlite(client):
...
+        pytest.skip("Skipping test for sqlite clients")
+
+    _setup_calls_for_storage_size_test(client)
+
+    # This is a best effort to achieve consistency in the calls_merged_stats table.
+    # calls_merged_stats is an AggregatingMergeTree table populated by a materialized view.
+    # ClickHouse merges data asynchronously, so queries may see unmerged data.
+    # OPTIMIZE TABLE ... FINAL forces an immediate merge to ensure consistency for tests.
+    if clickhouse_client:
+        clickhouse_client.command("OPTIMIZE TABLE calls_merged_stats FINAL")
```
chance-wnb, line 1788: Have you observed test flakiness if you don't do this?

tests/trace/test_weave_client.py
```diff
@@ -1720,6 +1720,140 @@ def test_summary_tokens_cost_sqlite(client):
...
+        client.get_calls(include_storage_size=False, include_total_storage_size=False)
+    )
+    baseline_count = len(baseline_calls)
+
+    # Verify baseline matches expected count
+    assert baseline_count == expected_count
+
+    # Get all storage parameter permutations (excluding baseline)
+    storage_permutations = [
+        (False, True),
```
chance-wnb: nit: I don't thinking any permutation is going to change the number of returned calls. It is a good sanity check, but the purpose of the test doesn't feel very strong for me.  line 1830 feels already sufficient.

mscavezze-cw: Good call - I've removed that

tests/trace/test_weave_client.py
mscavezze-cw: > even if we don't put limit=2 it also returns 2 calls right? so it does not prove the calls query is limited.
You are correct. I've split the test into two versions, one with filtering and one with limiting.

> Adding assert expected_count > 2 might help.
Added for both of the tests

tests/trace/test_weave_client.py
mscavezze-cw: I haven't personally, but this follows a pattern in test_client_trace.py.

tests/trace/test_weave_client.py
mscavezze-cw: I tried adding both of these, and the test failed. It looks like `server_call.total_storage_size_bytes` is `None`. What do you think would cause this? Do you think this indicates a problem with the code, a problem with test, or an incorrect assumption?

mscavezze-cw: total_storage_size_bytes is only set for root calls (parent_id is None)

I've added a comment and assert

chance-wnb: yes, total_storage_size_bytes is only set for root calls.

PR #5855, opened 2025-12-17
feat(weave): add timestamp override support for retroactive call logging

vanpelt:
## Description

The claude session logger I made for hack week allows users to import historic claude sessions.  The UI looks much better if the start and end times of the calls reflect what they were when that claude session ran.  We need to consider if doing this will mess with data locality and our special time ordered uuids.

## Testing

How was this PR tested?


andrewtruong:
looks good, can you please add a test

PR #5862, opened 2025-12-18
feat(weave): Add support for op kinds and colors

andrewtruong:
This PR adds support for better organizing traces in the UI by extending the op data model to include 2 new paths:
1. `attributes.weave.kind`
2. `attributes.weave.color`

To add a kind or color, simply define it in the op def:
```py
@weave.op(kind="llm", color="blue")
def func(): ...
```

Pair with: https://github.com/wandb/core/pull/37460

<div>
    <a href="https://www.loom.com/share/0d7acf0081f14c5ebf8bbf11d2e44b02">
      <p>Introducing the New Op.Kind Indicator for Enhanced Trace Tree Organization - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/0d7acf0081f14c5ebf8bbf11d2e44b02">
      
    </a>
  </div>


weave/trace/op_protocol.py
```diff
@@ -19,6 +20,12 @@
 P = ParamSpec("P")
 R = TypeVar("R")
 
+# Each Op kind has an associated icon in the UI.
+OpKind = Literal["agent", "llm", "tool", "search"]
+
+# Basic colors for additional organization.
+OpColor = Literal["red", "orange", "yellow", "green", "blue", "purple"]
+
```
neutralino1, line 28: Should these be Enums? So that users can use the Enum directly in there code? Cleaner than loose strings.

andrewtruong: I think this is cleaner and you will get autocomplete
image

PR #5865, opened 2025-12-18
fix(weave): Costs query not handling input and output tokens

zbirenbaum:
## Description

- Fixes WB-29947

Bug reported by customer regarding OTEL costs not appearing. Query was not accounting for input and output tokens. Unfortunately parent calls which have usage information but no valid model do not display tokens in summary since we do not generate a cost summary without a model to check costs for. We can fix this after code freeze on front end.

docs are not required

## Testing

Added new test to OTEL for cost calculation

image

### DIFF
--- a/tests/trace_server/test_opentelemetry.py
+++ b/tests/trace_server/test_opentelemetry.py
@@ -1187,18 +1187,14 @@ class TestSemanticConventionParsing:
 
         call_with_cost = res_with_cost.calls[0]
         call_no_cost = res_no_cost.calls[0]
+        assert call_with_cost.summary is not None
 
-        # Verify cost information is present when requested
-        assert "weave" in call_with_cost.summary
-        assert "costs" in call_with_cost.summary["weave"]
-        assert "gpt-4" in call_with_cost.summary["weave"]["costs"]
+        # Verify model cost information is present when requested
+        assert "gpt-4" in call_with_cost.summary["weave"]["costs"]  # type: ignore
 
-        gpt4_cost = call_with_cost.summary["weave"]["costs"]["gpt-4"]
-
-        # Delete timestamp fields that vary
-        del gpt4_cost["effective_date"]
-        del gpt4_cost["created_at"]
-
+        gpt4_cost = call_with_cost.summary["weave"]["costs"]["gpt-4"]  # type: ignore
+        del gpt4_cost["effective_date"]  # type: ignore
+        del gpt4_cost["created_at"]  # type: ignore
         # Verify cost calculation matches expected values
         assert (
             gpt4_cost
@@ -1221,7 +1217,7 @@ class TestSemanticConventionParsing:
         )
 
         # Verify no cost information when not requested
-        assert "costs" not in call_no_cost.summary.get("weave", {})
+        assert "costs" not in call_no_cost.summary.get("weave", {})  # type: ignore
 
 
 class TestHelpers:


weave/trace_server/token_costs.py
```diff
@@ -143,10 +143,10 @@ def get_llm_usage(param_builder: ParamBuilder, table_alias: str) -> PreparedSele
             ) AS kv"""
     llm_id = "kv.1 AS llm_id"
     requests = "JSONExtractInt(kv.2, 'requests') AS requests"
-    # Some libraries return prompt_tokens and completion_tokens, others prompt_tokens and completion_tokens
-    # prompt_tokens is not present, we use input_tokens
-    prompt_tokens = """if(JSONHas(kv.2, 'prompt_tokens'), JSONExtractInt(kv.2, 'prompt_tokens'), JSONExtractInt(kv.2, 'input_tokens')) AS prompt_tokens"""
-    completion_tokens = """if(JSONHas(kv.2, 'completion_tokens'), JSONExtractInt(kv.2, 'completion_tokens'), JSONExtractInt(kv.2, 'output_tokens')) AS completion_tokens"""
+    # Some libraries return prompt_tokens and completion_tokens, others input_tokens and output_tokens
+    # If both are present, add them together; if only one is present, use that one
```
gtarpenning, line 147: I wonder if there is ever a case where we double count because of this 🤔 probably not.

tests/trace_server/test_opentelemetry.py
```diff
@@ -1128,6 +1129,100 @@ def test_opentelemetry_usage_output_tokens_with_explicit_total(self):
...
+        # Verify cost information is present when requested
+        assert "weave" in call_with_cost.summary
+        assert "costs" in call_with_cost.summary["weave"]
+        assert "gpt-4" in call_with_cost.summary["weave"]["costs"]
+
+        gpt4_cost = call_with_cost.summary["weave"]["costs"]["gpt-4"]
+
+        # Delete timestamp fields that vary
+        del gpt4_cost["effective_date"]
+        del gpt4_cost["created_at"]
```
gtarpenning: the canonical way of doing this is with the `FuzzyDateTimeMatcher`

tests/trace_server/test_opentelemetry.py
```diff
@@ -1128,6 +1129,100 @@ def test_opentelemetry_usage_output_tokens_with_explicit_total(self):
...
+
+        assert len(res_with_cost.calls) == 1
+        assert len(res_no_cost.calls) == 1
+
+        call_with_cost = res_with_cost.calls[0]
+        call_no_cost = res_no_cost.calls[0]
+
+        # Verify cost information is present when requested
+        assert "weave" in call_with_cost.summary
+        assert "costs" in call_with_cost.summary["weave"]
```
gtarpenning: probably don't need these

gtarpenning: the main assert does this anyway

gtarpenning (approved):
lgtm

weave/trace_server/token_costs.py
jwlee64: this sounds like it might eventually become a bug lol, but for now non blocking,

zbirenbaum: The scenario I could see is with OTEL rollups which would have to happen at query layer. This won't be an issue or relevant until then

jwlee64: Its more like some provider that we havent accounted for might just duplicate the usage numbers between the two, but most likely wont be an issue

tests/trace_server/test_opentelemetry.py
zbirenbaum: Tried but ran into issues with it. Not worth blocking over so I'm going to keep this.

Issue #5844, opened 2025-12-12
[Bug]: Google GenAI (`from google import genai`) is not implicitly patched despite docs claiming automatic tracing


oekekezie:
### Describe the bug

Weave’s documentation states that Google GenAI is automatically traced via implicit patching, but in practice `from google import genai` is **not patched unless `weave.integrations.patch_google_genai()` is called explicitly**, even when the import occurs *after* `weave.init()`.

This differs from OpenAI, which *is* implicitly patched without an explicit patch call.

### Reproduction

```python
import weave # version 0.52.22

weave.init(project_name="google-genai")

from google import genai

client = genai.Client(api_key="...")
client.models.generate_content(
    model="gemini-2.0-flash",
    contents="What's the capital of France?",
)
```

**Observed behavior**

* No Weave traces are captured unless `weave.integrations.patch_google_genai()` is called explicitly.

**Expected behavior (per docs)**

* Google GenAI should be implicitly patched and traced without calling a patch function.

### Documentation reference

Official docs show this exact pattern and state tracing is automatic:
[https://docs.wandb.ai/weave/guides/integrations/google](https://docs.wandb.ai/weave/guides/integrations/google)

### Likely cause

From inspection of `weave/integrations/patch.py`, implicit patching appears to rely on:

* root-module matching (`root_module = fullname.split(".")[0]`)
* and/or `fullname == root_module`

This works for `openai` but not for Google GenAI, which is imported as `google.genai` under the shared `google` namespace. As a result, the import hook does not trigger patching for `google.genai`, even when imported after `weave.init()`.

### Why this matters

* The documented usage pattern does not reliably work.
* Behavior differs from OpenAI despite similar “automatic tracing” claims.
* Users must discover and apply `patch_google_genai()` manually, which contradicts the docs.

### Suggested direction

* Ensure implicit patching recognizes `google.genai` explicitly (without patching the entire `google` namespace), or
* Update documentation to reflect the need for explicit patching if this is intentional.

neutralino1:
Thank you, here's a proposed fix. https://github.com/wandb/weave/pull/5879

andrewtruong:
Hi @oekekezie this should be fixed as of the latest release, so closing as resolved.

Can you please try the latest version of weave and see if it works for you?  If not, feel free to re-open.  Thanks!

PR #5867, opened 2025-12-19
chore(weave): Update google generative api patch autopatch target

andrewtruong:
https://wandb.atlassian.net/browse/WB-29827
https://github.com/wandb/weave/issues/5844

Updates the patch target so when the user does
```py
from google import genai
import weave

weave.init(...)
```

The genai methods will be patched automatically

### DIFF
--- a/weave/integrations/patch.py
+++ b/weave/integrations/patch.py
@@ -108,7 +108,7 @@ def patch_google_genai(settings: IntegrationSettings | None = None) -> None:
     if settings is None:
         settings = IntegrationSettings()
     if get_google_genai_patcher(settings).attempt_patch():
-        _PATCHED_INTEGRATIONS.add("google.generativeai")
+        _PATCHED_INTEGRATIONS.add("google.genai")
 
 
 def patch_vertexai(settings: IntegrationSettings | None = None) -> None:


jamie-rasmussen (approved):

jamie-rasmussen:
Does the `_PATCHED_INTEGRATIONS.add("google.generativeai")` line need to be updated? Will this make logging stop working for old Google SDKs?

andrewtruong:
Yep, thanks for catching.  It will stop autopatching the old Google SDKs, but those are long deprecated by Google anyways:
https://pypi.org/project/google-generativeai/

andrewtruong:
Actually I guess we can keep patching for that path anyways.  We do that for crewai

neutralino1:
I'm not sure this is working due to the fact that we use `split(".")[0]` to identify the root module.
I have a different fix here: https://github.com/wandb/weave/pull/5879/changes

andrewtruong:
@neutralino1 my solution seems to work for me.  Are you seeing something different?

https://wandb.ai/megatruong/testing/r/call/019b9059-e994-719f-aad9-b163ef49226d
image

neutralino1:
ok then 🤔

PR #5873, opened 2025-12-22
fix(weave): Remove chars causing op creation to fail

zbirenbaum:
## Description

- Fixes WB-29975

Customer ran into issue exporting to weave through intermediate collector. Turns out it was unrelated to OTEL but was instead an issue with chars that break op creation not being removed by the safe name method.

docs are not required

## Testing

Validated full e2e flow and can ingest with this change.

### DIFF
--- a/tests/trace/test_server_object_creation_utils.py
+++ b/tests/trace/test_server_object_creation_utils.py
@@ -228,4 +228,6 @@ def test_make_safe_name_with_angle_brackets() -> None:
             return x + 1
 
         # Verify the op was successfully created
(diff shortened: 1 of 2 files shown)


zbirenbaum:
> Does this need to me deployed asap or can it wait until the end of code freeze?
> 
> Also, maybe a unit test would be good here.

Sure I'll add a test. Lets wait until after freeze

zbirenbaum:
@clboetticher Tagging you on this

tssweeney (approved):

PR #5879, opened 2025-12-30
fix(weave): Special-casing google.genai module patching

neutralino1:
## Description

- Fixes WB-30014

This PR addressed the bug reported in https://github.com/wandb/weave/issues/5844. The `google.genai` module is not being autopatched because it is a namespaced module.
This PR special-cases this module.

## Testing

How was this PR tested?

- [x] Unit testing

### DIFF
--- a/tests/integrations/google_genai/test_google_genai_autopatch.py
+++ b/tests/integrations/google_genai/test_google_genai_autopatch.py
@@ -1,13 +1,11 @@
 """Tests for Google GenAI autopatching behavior."""
 
-from __future__ import annotations
-
 from unittest.mock import MagicMock, patch
 
 from weave.integrations.patch import register_import_hook
 
 
-def test_foo():
+def test_google_genai_autopatch():
     register_import_hook()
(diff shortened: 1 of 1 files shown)


weave/integrations/patch.py
```diff
@@ -315,8 +315,11 @@ def find_spec(self, fullname, path, target=None):  # type: ignore
         integration is being imported and schedule it for patching after import.
         """
         # Check if this is a root module we support (not a submodule)
-        root_module = fullname.split(".")[0]
-
+        # Google GenAI is a special case because it's a namespace package
+        if fullname.startswith("google.genai"):
+            root_module = "google.genai"
+        else:
+            root_module = fullname.split(".")[0]
```
neutralino1, line 322: This is ugly, any better ideas?

adrnswanberg (approved):

andrewtruong:
@neutralino1 this fix works because you're using `google.genai` instead of `google.generativeai`, not because of the special casing.

PR #5888, opened 2026-01-07
chore(weave): Allows the client to enforce a minimum trace server version

tssweeney:
This PR sets up the mechanics to allow the client to require a minimum server version. by default it is set to None, so it is a no op, but once https://github.com/wandb/core/pull/37714 is deployed, we can then start requiring a minimum version for certain features.

### DIFF
--- a/weave/trace/weave_init.py
+++ b/weave/trace/weave_init.py
@@ -184,7 +184,7 @@ def init_weave(
         # In the future, we may want to throw here.
         min_required_version = "0.0.0"
         trace_server_version = None
-    trace_server_url = remote_server.trace_server_url
+    trace_server_url = env.weave_trace_server_url()
     if not init_message.check_min_weave_version(min_required_version, trace_server_url):
         return init_weave_disabled()
     if not init_message.check_min_trace_server_version(


weave/trace/init_message.py
```diff
@@ -96,6 +96,47 @@ def assert_min_weave_version(
         raise ValueError(message)
 
 
+def assert_min_trace_server_version(
+    trace_server_version: str | None,
+    min_required_version: str | None,
+    trace_server_host: str = "https://trace.wandb.ai",
```
andrewtruong: extract const to top

weave/trace/weave_init.py
```diff
@@ -173,16 +174,21 @@ def init_weave(
...
     except Exception:
         # Set to a minimum version that will always pass the check
         # In the future, we may want to throw here.
         min_required_version = "0.0.0"
+        trace_server_version = None
     init_message.assert_min_weave_version(min_required_version)
+    init_message.assert_min_trace_server_version(
+        trace_server_version,
+        MIN_TRACE_SERVER_VERSION,
+    )
```
neutralino1: Shouldn't we explicitly pass `trace_server_host`? Won't this be incorrect on dedicated and self-managed deployment?

weave/trace/init_message.py
neutralino1: Seems risky to default to prod.

andrewtruong: Might make sense for this to rely on the settings object

tssweeney: fixing

tssweeney: fixing

weave/trace/weave_init.py
tssweeney: yup - exactly how i fixed it

PR #5890, opened 2026-01-07
chore(weave): more datadog reporting on types of calls queries

gtarpenning:
## Description

Adds a couple tags to more easily correlate problematic/slow queries in datadog. Include flags for cost query, feedback query, and expand column query. 

We might want these tags on the root, but for now on the stream call feels okay.

## Testing

image

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -640,7 +640,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
 
         # We put summary_dump last so that when we compute the costs and summary its in the right place
         if req.include_costs:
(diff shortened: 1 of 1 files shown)


mscavezze-cw (approved):
Looks good.

calls_query_stream is a little bit long. Consider whether there is a clean way to break it into multiple methods. The goal would be to improve readability and make future reviews easier.

gtarpenning:
> Looks good.
> 
> calls_query_stream is a little bit long. Consider whether there is a clean way to break it into multiple methods. The goal would be to improve readability and make future reviews easier.
@mscavezze-cw 

Sounds like a great idea. I'll be the first to raise my hand and admit I have jammed way too much stuff in there. It's also our second most important endpoint, so it really ought to be clean and easy to audit.

PR #5891, opened 2026-01-07
feat(weave): Audio support in LLMAsAJudgeScorer

mscavezze-cw:
## Description

- WB-26366
- DOCS-1969

This PR allows audio to be used with LLMAsAJudgeScorer

This is part of a series of PRs to address https://wandb.atlassian.net/browse/WB-26366
- Core Deployments: https://github.com/wandb/core-deployments/pull/68
- Weave: https://github.com/wandb/weave/pull/5891
- Core: https://github.com/wandb/core/pull/37746

## Testing

- [x] _Updated unit tests_
- [x] _Added integration tests_
- [x] _Tested on local environment_
- [x] _Tested on QA environment_

### DIFF
--- a/tests/trace/data_serialization/test_cases/library_cases.py
+++ b/tests/trace/data_serialization/test_cases/library_cases.py
@@ -74,7 +74,7 @@ def evaluation_equality_check(a, b):
 
 llm_as_a_judge_scorer_digest = (
     "t7mpggmkUX0XDd9gi2K2J01r657NtnfGlnc0cb8hScs"
-    if sys.version_info.minor >= 13
+    if sys.version_info.major >= 3 and sys.version_info.minor >= 13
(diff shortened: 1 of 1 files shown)


weave/scorers/llm_as_a_judge_scorer.py
```diff
@@ -12,6 +12,8 @@
 class LLMAsAJudgeScorer(Scorer):
     model: LLMStructuredCompletionModel
     scoring_prompt: str
+    include_audio_output: bool = False
+    audio_output_json_paths: list[str] | None = None
```
zbirenbaum: Could you add a comment here about what impact these have? In the test case above audio_output_json_paths  is None and include_audio_output is false so I have no idea what they do.

mscavezze-cw: I'll add a comment. I'll also investigate whether it makes sense to add them to the test case (though those cases are not exactly self-documenting)

PR #5894, opened 2026-01-08
chore(weave): Add autopatching helper to simplify integration patching

andrewtruong:
This adds a utility function to reduce boilerplate and make it easier have multiple targets for autopatching.

For example, you can now more easily define both `google.genai` and `google.generativeai` as symbols to trigger the google autopatching function

The utility function requires named kwargs to avoid ambiguity

### DIFF
--- a/tests/integrations/test_autopatch.py
+++ b/tests/integrations/test_autopatch.py
@@ -319,6 +319,8 @@ def test_patch_integration(setup_env, success):
 
         mock_getter.assert_called_once()
         mock_patcher.attempt_patch.assert_called_once()
+
+        # True if patched, False if not
         assert ("single_symbol" in patch_module._PATCHED_INTEGRATIONS) is success
 
         # Second call behavior depends on whether first patch succeeded
@@ -373,6 +375,8 @@ def test_patch_integration_multi(setup_env, success):
 
         mock_getter.assert_called_once()
         mock_patcher.attempt_patch.assert_called_once()
+
+        # All True if patched, all False if not
         assert ("symbol1" in patch_module._PATCHED_INTEGRATIONS) is success
         assert ("symbol2" in patch_module._PATCHED_INTEGRATIONS) is success
         assert ("symbol3" in patch_module._PATCHED_INTEGRATIONS) is success


weave/integrations/patch.py
```diff
@@ -6,10 +6,12 @@
 3. Import hook for automatic patching of libraries imported after weave.init()
 """
 
+from __future__ import annotations
```
gtarpenning, line 9: why do we need this?

weave/integrations/patch.py
```diff
@@ -18,7 +20,37 @@
...
+        settings: Optional integration settings
+    """
+    # If symbols are already patched, don't patch again
+    if any(name in _PATCHED_INTEGRATIONS for name in triggering_symbols):
+        return
+
+    if settings is None:
+        settings = IntegrationSettings()
+
+    module = importlib.import_module(module_path)
```
gtarpenning, line 48: I presume we get the same deferred import latency benefits using this path?

weave/integrations/patch.py
andrewtruong: This just defers annotations where possible.  In general we should have this everywhere

weave/integrations/patch.py
andrewtruong: Yes, as written this means the import is deferred until `_patch_integration`

weave/integrations/patch.py
```diff
@@ -265,12 +290,13 @@ def patch_llamaindex() -> None:
...
-
-    if settings is None:
-        settings = IntegrationSettings()
-    if get_openai_realtime_websocket_patcher(settings).attempt_patch():
-        _PATCHED_INTEGRATIONS.add("openai_realtime")
+    """Enable Weave tracing for OpenAI Realtime."""
+    _patch_integration(
+        module_path="weave.integrations.openai_realtime",
+        patcher_func_getter_name="get_openai_realtime_websocket_patcher",
+        triggering_symbols=["openai_realtime"],
```
zbirenbaum, line 297: Does triggering symbols refer to the weave module or the external module? It looks like the external module above. Realtime patches various websocket modules

PR #5895, opened 2026-01-08
chore(weave): retry 413 with smaller batches if possible

gtarpenning:
## Description

I think this is a bug: if we ever get a 413 from the server, we should _always_ try to send a smaller batch. Even if we tried before, it was best effort, and we should attempt to split and send again when we get 413. 

Also lightly refactors for DRY
Also cleans up an import

## Testing

adds test

### DIFF
--- a/weave/trace_server_bindings/http_utils.py
+++ b/weave/trace_server_bindings/http_utils.py
@@ -173,12 +173,7 @@ def process_batch_with_retry(
         send_batch_fn(encoded_data)
     except Exception as e:
         # Handle 413 specially: server rejected as too large, split and retry
-        is_413 = (
-            isinstance(e, httpx.HTTPStatusError)
-            and e.response is not None
-            and e.response.status_code == 413
-        )
-        if is_413 and len(batch) > 1:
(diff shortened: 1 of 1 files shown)


weave/trace_server_bindings/http_utils.py
```diff
@@ -159,7 +172,27 @@ def process_batch_with_retry(
     try:
         send_batch_fn(encoded_data)
     except Exception as e:
-        from weave.utils.retry import _is_retryable_exception
+        # Handle 413 specially: server rejected as too large, split and retry
+        is_413 = (
+            isinstance(e, httpx.HTTPStatusError)
+            and e.response is not None
+            and e.response.status_code == 413
+        )
```
mscavezze-cw: Consider extracting a helper, and calling it directly from the if below

mscavezze-cw (approved):
Looks good. Thanks for the minor refactors.

weave/trace_server_bindings/http_utils.py
gtarpenning: yep I like it

PR #5900, opened 2026-01-10
feat(weave): autoconvert base64 safely in trace server

zbirenbaum:
## Description

- WB-30277

docs are not required
Requires [wandb/core #37876](https://github.com/wandb/core/pull/37876) to work

### DIFF
--- a/weave/trace_server/base64_content_conversion.py
+++ b/weave/trace_server/base64_content_conversion.py
@@ -148,7 +148,7 @@ def replace_base64_with_content_objects(
                     # More complicated false positives or failed detections will show 'application/octet-stream'
                     # The uncovered scenario is if a user has encoded a plaintext document as Base64
                     # We don't handle text content objects in a special way on the clients, so this is acceptable.
-                    content = Content.from_base64(val)
+                    content: Content[Any] = Content.from_base64(val)
                     if content.mimetype not in (
                         "text/plain",
                         "application/octet-stream",


weave/trace_server/base64_content_conversion.py
```diff
@@ -23,10 +23,28 @@
...
+    Args:
+        value: String to check
+    Returns:
+        True if the string is valid base64
+    """
+    # Check if it matches base64 pattern
+    if not BASE64_PATTERN.match(value):
+        return False
+
+    return True
```
neutralino1:
```suggestion
    return (BASE64_PATTERN.match(value) is not None)
```

neutralino1 (approved):
One nit.

weave/trace_server/base64_content_conversion.py
```diff
@@ -128,6 +146,26 @@ def _visit(val: Any) -> Any:
...
+                        "text/plain",
+                        "application/octet-stream",
+                    ):
+                        return store_content_object(
+                            content,
+                            project_id,
+                            trace_server,
+                        )
+                except Exception as e:
+                    logger.warning(
```
mscavezze-cw, line 162: `warning` might be a little noisy, given that we're attempting to decode something that we're not sure is actually base64. Maybe `info` would be better?

tests/trace_server/test_base64_content_conversion.py
```diff
@@ -222,5 +223,128 @@ def test_process_call_req_to_content_start_and_end(self):
...
+class TestStandaloneBase64Detection:
+    """Test detection and conversion of standalone base64 strings (new functionality)."""
+
+    def test_non_media_base64_strings_not_converted(self):
+        """Test that various base64 strings of different lengths (mod 4) that decode to
+        text/plain or application/octet-stream are NOT converted."""
+        # Mock trace server
+        trace_server = MagicMock()
+
+        # Test strings of various lengths mod 4 (0, 1, 2, 3)
```
mscavezze-cw: I'm not sure what `(0, 1, 2, 3)` is meant to convey here. Consider omitting or elaborating.

weave/trace_server/base64_content_conversion.py
zbirenbaum: Nice catch, the `return bool(BASE64_PATTERN.match(value))` I use below is even more concise so I swapped to that for consistency

neutralino1: I actually prefer `is not None`, doesn't use truthiness which always generates bugs.

tests/trace_server/test_base64_content_conversion.py
```diff
@@ -222,5 +223,128 @@ def test_process_call_req_to_content_start_and_end(self):
...
+        # These should all decode to text/plain or application/octet-stream
+        test_cases = [
```
mscavezze-cw, line 238: Why not parameterize the test (`@pytest.mark.parametrize`)? That would keep the test cases a little more separate, improving readability. It might also render any test errors more clearly.

weave/trace_server/base64_content_conversion.py
```diff
@@ -128,6 +146,26 @@ def _visit(val: Any) -> Any:
...
 
+            if is_maybe_base64(val):
+                try:
+                    content = Content.from_base64(val)
+                    # 'aaaa' is valid base64 and will come out as text/plain
+                    # More complicated false positives or failed detections will show 'application/octet-stream'
+                    # We don't handle either in a special way so false negatives in our detection aren't an issue
+                    if content.mimetype not in (
+                        "text/plain",
+                        "application/octet-stream",
```
mscavezze-cw, line 154: Could you add some documentation on why these mimetypes indicate that we shouldn't convert?
`text/plain` seems reasonable enough, since it's not really `content`, and is probably not exceeding the 3.5mb limit (though we should consider what to do in cases where a `text/plain` does exceed the 3.5mb limit.
`application/octet-stream` is less obvious

weave/trace_server/base64_content_conversion.py
```diff
@@ -128,6 +146,26 @@ def _visit(val: Any) -> Any:
                         f"Failed to create and store content from data URI with error {e}"
                     )
 
+            if is_maybe_base64(val):
+                try:
```
mscavezze-cw, line 145: Consider extracting a helper function that containing either all of the new code, or perhaps leaving just the `if is_maybe_base64(val):` here. This would help improve readability by keeping replace_base64_with_content_objects shorter, and by keeping the details of this new heuristic separate. I get the sense that the heuristic may grow over time, either because of additional implementation complexity or additional documentation (see below).

tests/trace_server/test_base64_content_conversion.py
```diff
@@ -222,5 +223,128 @@ def test_process_call_req_to_content_start_and_end(self):
...
+        wav_data.extend(audio_data_size.to_bytes(4, "little"))
+        # Add audio data (silence)
+        wav_data.extend(b"\x00" * audio_data_size)
+
+        # Update file size in RIFF header (total size - 8 bytes for RIFF header)
+        file_size = len(wav_data) - 8
+        wav_data[file_size_pos:file_size_pos + 4] = file_size.to_bytes(4, "little")
+
+        # Encode as base64
+        wav_base64 = base64.b64encode(bytes(wav_data)).decode("ascii")
```
mscavezze-cw, line 318: Consider extracting to helper for improved readability of the test (and maybe later reuse)

tests/trace_server/test_base64_content_conversion.py
zbirenbaum: Addressed on call

tests/trace_server/test_base64_content_conversion.py
zbirenbaum: resolved on call

weave/trace_server/base64_content_conversion.py
zbirenbaum: Discussed on call

weave/trace_server/base64_content_conversion.py
zbirenbaum: Discussed on call

weave/trace_server/base64_content_conversion.py
zbirenbaum: discussed on call

weave/trace_server/base64_content_conversion.py
zbirenbaum: Addressed

PR #5901, opened 2026-01-12
fix(weave): respect HTTP_PROXY/HTTPS_PROXY environment variables in httpx.Client

### LINKED ISSUE #5849: [Bug] HTTP_PROXY/HTTPS_PROXY environment variables are ignored in http_requests client
### Describe the bug

https://github.com/wandb/weave/blob/ed52f1a6299fcfc371bf9cf2c6fa0e40e916fc7b/weave/utils/http_requests.py#L167-L171

The `httpx.Client` in `weave/utils/http_requests.py` does not respect `HTTP_PROXY`/`HTTPS_PROXY` environment variables, making it impossible for users behind corporate proxies to use Weave.

### Root Cause

When `transport` is explicitly provided, httpx skips proxy configuration from environment variables.

https://github.com/encode/httpx/blob/ae1b9f66238f75ced3ced5e4485408435de10768/httpx/_client.py#L685C1-L687C1

gemelom:
## Description

- Fixes #5849

### DIFF
--- a/weave/utils/http_requests.py
+++ b/weave/utils/http_requests.py
@@ -164,8 +164,15 @@ def _get_http_timeout() -> float:
     return http_timeout()
 
 
+proxy_url = (
+    os.environ.get("HTTPS_PROXY")
+    or os.environ.get("https_proxy")
+    or os.environ.get("HTTP_PROXY")
+    or os.environ.get("http_proxy")
+)
+
 client = httpx.Client(
-    transport=LoggingHTTPTransport(),
+    transport=LoggingHTTPTransport(proxy=proxy_url),
     timeout=_get_http_timeout(),
     limits=httpx.Limits(max_connections=None, max_keepalive_connections=None),
 )


gemelom:

weave/utils/http_requests.py
```diff
@@ -164,8 +164,15 @@ def _get_http_timeout() -> float:
     return http_timeout()
 
 
+proxy_url = (
+    os.environ.get("HTTPS_PROXY")
+    or os.environ.get("https_proxy")
+    or os.environ.get("HTTP_PROXY")
+    or os.environ.get("http_proxy")
+)
```
gtarpenning, line 172: @andrewtruong will we need to use these anywhere else?

gtarpenning, line 172: @andrewtruong will we need to use these anywhere else?

andrewtruong:
Hey @gemelom thanks for your contrib.  We have an issue with the non-wandb testing workflow atm, but I will open an equivalent branch with small changes and get this merged!

weave/utils/http_requests.py
andrewtruong: I think you just need to do this for HTTPX clients which we don't use elsewhere atm

PR #5903, opened 2026-01-13
chore(weave): Generalize the ClickhouseMigrator to remove TraceServer-specific details

tssweeney:
This PR improves our migrator in 2 distinct ways:
1. We allow the caller to specify a `migration_dir` which contains the migration files
2. We generalize the execution flow such that the cost insertion is handled via a lifecycle hook.

For the normal trace server, there are no changes since the getter function injects the same defaults. However, this allows me to re-use the migrator mechanics in the trace server's new llm proxy extension

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_migrator.py
+++ b/weave/trace_server/clickhouse_trace_server_migrator.py
@@ -898,7 +898,8 @@ def get_clickhouse_trace_server_migrator(
     use_distributed: bool | None = None,
     management_db: str = "db_management",
     migration_dir: str | None = None,
-    post_migration_hook: PostMigrationHook | None = _default_trace_server_costs_post_migration_hook,
+    post_migration_hook: PostMigrationHook
+    | None = _default_trace_server_costs_post_migration_hook,
 ) -> BaseClickHouseTraceServerMigrator:
     """Factory function to create the appropriate migrator based on configuration.
 


weave/trace_server/clickhouse_trace_server_migrator.py
```diff
@@ -510,12 +566,22 @@ def __init__(
         replicated_path: str | None = None,
         replicated_cluster: str | None = None,
         management_db: str = "db_management",
+        *,
+        migration_dir: str,
+        post_migration_hook: PostMigrationHook | None = None,
```
gtarpenning, line 571: pretty anti this pattern, this is good for not changing other code but it's easy to just swallow args. Any chance we can refactor the other callsites?

gtarpenning: or, perhaps, why do we need the `*` explicitly?

gtarpenning: okay looking at the tests this is pretty reasonable.

weave/trace_server/clickhouse_trace_server_migrator.py
```diff
@@ -867,16 +939,39 @@ def get_clickhouse_trace_server_migrator(
             "Set replicated=True or use_distributed=False."
         )
 
+    if migration_dir is None:
+        migration_dir = os.path.abspath(
+            os.path.join(os.path.dirname(__file__), "migrations")
+        )
+
+    if post_migration_hook is None:
+        post_migration_hook = _default_trace_server_costs_post_migration_hook
```
gtarpenning: this is a bit overloaded, obviously I get why we are doing this but it might not be clear that by not providing a hook we insert costs, or by providing one we dont... if that makes sense. could be worth calling out

weave/trace_server/clickhouse_trace_server_migrator.py
tssweeney: which pattern do you not like - the * in args?

gtarpenning: thats what I was commenting on yeah, little convo by myself, it's a bit unique in our codebase

weave/trace_server/clickhouse_trace_server_migrator.py
tssweeney: i'll make this more clear

weave/trace_server/clickhouse_trace_server_migrator.py
andrewtruong: In this context I think the `*` is fine.

`*` is intended to force the user to specify the kwarg by name to avoid ambiguity.  In general I prefer specific names to avoid confusion, but it can get annoying.  For example, pydantic requires named args for everything.

PR #5911, opened 2026-01-15
chore(weave): add calls_complete v2 write path

gtarpenning:
[WB-29712](https://wandb.atlassian.net/browse/WB-29712)

## Description

Write path for `calls_complete` table. 

Adds 3 new endpoints:
```python
def calls_complete(self, req: CallsUpsertCompleteReq) -> CallsUpsertCompleteRes: ...
def call_start_v2(self, req: CallStartV2Req) -> CallStartV2Res: ...
def call_end_v2(self, req: CallEndV2Req) -> CallEndV2Res: ...
```

The main one, which should see the vast vast majority of traffic: `calls_complete` is a batch endpoint only accepting completed calls, or those with both start and end data. `call_start_v2` accepts starts and inserts them into the `calls_complete` table, while `call_end_v2` accepts call end data for calls that have already been started, and uses the lightweight `UPDATE` function to update the `calls_complete` table for the given record. 

One interesting part of the UPDATE path to note is that because `started_at` is part of the primary key, call end `UPDATES` to the `calls_complete` table are _much_ better if they have the `started_at` of the original call. This is a requirement to updating a call end. This will be automatically handled on the sdk, but could be confusing for users trying to use the endpoints directly. This constraint can always be relaxed in the future, although it does hurt performance.

## Testing

Adds unit tests mostly, more end to end tests coming once the read side is in place. 

[WB-29712]: https://wandb.atlassian.net/browse/WB-29712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/calls_query_builder/calls_query_builder.py
+++ b/weave/trace_server/calls_query_builder/calls_query_builder.py
@@ -2007,6 +2007,11 @@ def build_calls_complete_update_end_query(
         where_clauses.append(
             f"started_at = fromUnixTimestamp64Micro({{{started_at_param}:Int64}}, 'UTC')"
         )
+    else:
+        # TODO: try to optimistically parse uuidv7, grabbing timestamps from the ID
+        # then use that to narrow the granules we need to search.
+        pass
+
     where_clauses.append(f"id = {{{id_param}:String}}")
     where_clause = " AND ".join(where_clauses)
 
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -618,23 +618,24 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         Returns:
             CallsUpsertCompleteRes: Empty response on success.
         """
-        for complete_call in req.batch:
-            complete_call = process_complete_call_to_content(complete_call, self)
+        with self.call_batch():
+            for complete_call in req.batch:
+                complete_call = process_complete_call_to_content(complete_call, self)
 
-            # Determine write target based on project, this should be the same for all
-            # calls in the batch, subsequent calls just hit the in-memory cache. This
-            # is here for technical correctness, in case we relax project_id target
-            # constraints intra-batch
-            write_target = self.table_routing_resolver.resolve_v2_write_target(
-                complete_call.project_id,
-                self.ch_client,
-            )
+                # Determine write target based on project, this should be the same for all
+                # calls in the batch, subsequent calls just hit the in-memory cache. This
+                # is here for technical correctness, in case we relax project_id target
+                # constraints intra-batch
+                write_target = self.table_routing_resolver.resolve_v2_write_target(
+                    complete_call.project_id,
+                    self.ch_client,
+                )
 
-            ch_call = _complete_call_to_ch_insertable(complete_call)
-            if write_target == WriteTarget.CALLS_COMPLETE:
-                self._insert_call_complete(ch_call)
-            else:
-                self._insert_call_to_v1(ch_call)
+                ch_call = _complete_call_to_ch_insertable(complete_call)
+                if write_target == WriteTarget.CALLS_COMPLETE:
+                    self._insert_call_complete(ch_call)
+                else:
+                    self._insert_call_to_v1(ch_call)
 
         return tsi.CallsUpsertCompleteRes()
 


weave/trace_server/clickhouse_schema.py
```diff
@@ -97,6 +100,52 @@ class CallUpdateCHInsertable(BaseModel):
...
+    _op_name_v = field_validator("op_name")(validation.op_name_validator)
+    _input_refs_v = field_validator("input_refs")(validation.refs_list_validator)
+    _output_refs_v = field_validator("output_refs")(validation.refs_list_validator)
+    _display_name_v = field_validator("display_name")(validation.display_name_validator)
+    _wb_user_id_v = field_validator("wb_user_id")(validation.wb_user_id_validator)
+    _wb_run_id_v = field_validator("wb_run_id")(validation.wb_run_id_validator)
+    _wb_run_step_v = field_validator("wb_run_step")(validation.wb_run_step_validator)
+    _wb_run_step_end_v = field_validator("wb_run_step_end")(
+        validation.wb_run_step_validator
+    )
```
neutralino1: Can you explain this pattern?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -565,6 +601,162 @@ def call_end(
...
+        elif write_target == WriteTarget.CALLS_MERGED:
+            ch_end = _end_call_for_insert_to_ch_insertable_end_call(req.end)
+            self._insert_call(ch_end)
+            if self._flush_immediately:
+                self._flush_calls()
+
+        return tsi.CallEndV2Res()
+
+    @ddtrace.tracer.wrap(
+        name="clickhouse_trace_server_batched._update_call_end_in_calls_complete"
```
neutralino1, line 695: Nit: is `name=` necessary when == the func name?

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1963,3 +1963,60 @@ def _is_minimal_filter(filter: tsi.CallsFilter | None) -> bool:
...
+            exception = {{{exception_param}:Nullable(String)}},
+            output_dump = {{{output_dump_param}:String}},
+            summary_dump = {{{summary_dump_param}:String}},
+            output_refs = {{{output_refs_param}:Array(String)}},
+            wb_run_step_end = {{{wb_run_step_end_param}:Nullable(UInt64)}},
+            updated_at = now64(3)
+        WHERE project_id = {{{project_id_param}:String}}
+            AND started_at = fromUnixTimestamp64Micro({{{started_at_spec}}}, 'UTC')
+            AND id = {{{id_param}:String}}
+        """
```
neutralino1, line 2031: 👌

neutralino1 (commented):
Couple of nits, looks neat 😎

weave/trace_server/clickhouse_schema.py
gtarpenning: Follows from the other schema types, basically it forces Pydantic to use our special validators. We have one for each of our types, defined in `validation.py`.

neutralino1: And why are they not set directly on the actual fields?

gtarpenning: I _think_ the only other way is to do something like this (from the pydantic docs): 
```python
class Model(BaseModel):
    text: str

    @field_validator('text', mode='after')
    def remove_stopwords(cls, v: str, info: ValidationInfo) -> str:
        if isinstance(info.context, dict):
            stopwords = info.context.get('stopwords', set())
            v = ' '.join(w for w in v.split() if w.lower() not in stopwords)
        return v
```

which isn't DRY

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -556,8 +560,12 @@ def test_completions_create_stream_multiple_choices():
...
         # Get all the calls
         call_args = [call[0][0] for call in mock_insert_call.call_args_list]
-        start_calls = [call for call in call_args if hasattr(call, "started_at")]
-        end_calls = [call for call in call_args if hasattr(call, "ended_at")]
+        start_calls = [
+            call for call in call_args if isinstance(call, CallStartCHInsertable)
+        ]
+        end_calls = [
+            call for call in call_args if isinstance(call, CallEndCHInsertable)
+        ]
```
bcsherma, line 568: this seems more correct, was this necessary for the pr? not saying you shoudl remove it just curious

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -681,8 +689,12 @@ def test_completions_create_stream_single_choice_unified_wrapper():
...
```
bcsherma, line 697: samez

tests/trace_server/test_calls_complete.py
```diff
@@ -0,0 +1,508 @@
...
+        trace_server: External trace server fixture.
+
+    Returns:
+        ClickHouseTraceServer: The internal ClickHouse trace server.
+
+    Examples:
+        >>> internal = clickhouse_trace_server
+    """
+    internal_server = trace_server._internal_trace_server
+    if isinstance(internal_server, SqliteTraceServer):
```
bcsherma, line 33: this would be nice to make shared I love skipping tests on sqlite

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -693,3 +693,26 @@ def score_delete(self, req: tsi.ScoreDeleteReq) -> tsi.ScoreDeleteRes:
...
+        """Start a single call (v2), converting project_id."""
+        req.start.project_id = self._idc.ext_to_int_project_id(req.start.project_id)
+        if req.start.wb_user_id is not None:
+            req.start.wb_user_id = self._idc.ext_to_int_user_id(req.start.wb_user_id)
+        return self._ref_apply(self._internal_trace_server.call_start_v2, req)
+
+    def call_end_v2(self, req: tsi.CallEndV2Req) -> tsi.CallEndV2Res:
+        """End a single call (v2), converting project_id."""
+        req.end.project_id = self._idc.ext_to_int_project_id(req.end.project_id)
+        return self._ref_apply(self._internal_trace_server.call_end_v2, req)
```
bcsherma, line 718: is this just for evals? should we just have evals use the evals api

tests/trace_server/test_clickhouse_trace_server_batched.py
gtarpenning: technically now the call end can have "started_at", which is a perf thing mentioned in the pr description, so we can't use that as a check anymore

tests/trace_server/test_calls_complete.py
gtarpenning: yeah, good call....

weave/trace_server/external_to_internal_trace_server_adapter.py
gtarpenning: For MVP yes, although we will probably want it to be more flexible. 

I was under the impression the evals API is more of a work in progress?

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1963,3 +1963,60 @@ def _is_minimal_filter(filter: tsi.CallsFilter | None) -> bool:
...
+        SET
+            ended_at = fromUnixTimestamp64Micro({{{ended_at_spec}}}, 'UTC'),
+            exception = {{{exception_param}:Nullable(String)}},
+            output_dump = {{{output_dump_param}:String}},
+            summary_dump = {{{summary_dump_param}:String}},
+            output_refs = {{{output_refs_param}:Array(String)}},
+            wb_run_step_end = {{{wb_run_step_end_param}:Nullable(UInt64)}},
+            updated_at = now64(3)
+        WHERE project_id = {{{project_id_param}:String}}
+            AND started_at = fromUnixTimestamp64Micro({{{started_at_spec}}}, 'UTC')
```
tssweeney: is this `started_at` condition just a gut check to avoid potential id collisions?

gtarpenning: Its so the query planner uses the primary key.

gtarpenning: I could make it optional, but the performance should be better this way.

weave/trace_server/project_version/project_version.py
```diff
@@ -63,25 +83,11 @@ def resolve_read_table(self, project_id: str, ch_client: CHClient) -> ReadTable:
             return ReadTable.CALLS_MERGED
 
         residence = self._get_residence(project_id, ch_client)
+        set_current_span_dd_tags({"project_version.residence": residence.value})
```
tssweeney: any reason you aren't also setting the project id like you do above?

gtarpenning: oversight, nice.

weave/trace_server/project_version/types.py
```diff
@@ -6,20 +6,20 @@
...
-# │ BOTH                    │ COMPLETE / COMPLETE  │ MERGED / BOTH               │ COMPLETE / BOTH                │ MERGED / MERGED      │ MERGED / MERGED      │
-# └─────────────────────────┴──────────────────────┴─────────────────────────────┴────────────────────────────────┴──────────────────────┴──────────────────────┘
+# ┌─────────────────────────┬──────────────────────┬──────────────────────┬──────────────────────┐
+# │ Project Data Residence  │ AUTO                 │ FORCE_LEGACY         │ OFF                  │
+# │ (Physical State)        │ (Read / Write)       │ (Read / Write)       │ (Read / Write)       │
+# ├─────────────────────────┼──────────────────────┼──────────────────────┼──────────────────────┤
+# │ EMPTY                   │ COMPLETE / COMPLETE  │ MERGED / MERGED      │ MERGED / MERGED      │
+# │ MERGED_ONLY             │ MERGED / MERGED      │ MERGED / MERGED      │ MERGED / MERGED      │
+# │ COMPLETE_ONLY           │ COMPLETE / COMPLETE  │ MERGED / MERGED      │ MERGED / MERGED      │
+# │ BOTH                    │ COMPLETE / COMPLETE  │ MERGED / MERGED      │ MERGED / MERGED      │
```
tssweeney, line 21: it seems we have completely removed the dual write path - am i reading this table and the implementation correctly?

gtarpenning: yes. no dual writes. because this is opt in, I think dual writes just add massive complexity for very little gain. In addition to lots of extra logic across the insertion and read path, It also complicates the billing path, if users are dual writing we shouldn't charge them for both.

weave/trace_server/calls_query_builder/calls_query_builder.py
tssweeney: seems like a very odd constraint to build into the api. i'd think this should be optional and used by us internally for speed, but not something we make as a hard requirement

weave/trace_server/project_version/types.py
tssweeney: so then:
1. once enabled, this is a 1-way door (unless we do a migration)
2. since you are not throwing on the `BOTH` case, we have a weird scenario where an old client could write to a project written by the new client which then hides the data

gtarpenning: 1. Only for users who opt-in with a flag in the SDK. We can have warnings. We can also backfill from this table back into calls_merged easily if we revert.
2. No, we throw in that case:
```python
        # Check write target - v1 call_start cannot write to calls_complete
        write_target = self.table_routing_resolver.resolve_v1_write_target(
            ch_call.project_id,
            self.ch_client,
        )
        if write_target == WriteTarget.CALLS_COMPLETE:
            raise CallsCompleteModeRequired(ch_call.project_id)
```

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: It does feel odd I agree. Our client can easily keep track of it, but users of only the service endpoints will find that annoying. It should be possible to do something almost as good by deconstructing uuidv7s and grabbing the timestamp from that...
i'll make this optional and add a todo to implement that timestamp optimization for uuidv7s

PR #5914, opened 2026-01-15
fix(weave): Fix nesting issue when tracing generators

andrewtruong:
Internal: https://wandb.atlassian.net/browse/WB-30196

---

This PR fixes an issue where calling a generator repeatedly would incorrectly nest the 2nd iteration under the 1st, instead of having them as siblings.

See before and after for this sample code:
```py
import weave

weave.init("generator_issue")


@weave.op
async def inner():
    yield 1


@weave.op
async def outer():
    async for x in inner():
        yield x


@weave.op
async def parent():
    for name in ["first", "second"]:
        async for val in outer():
            ...


await parent()
```

## Before
image

## After
image

(there are 2 of these and they are not nested)

### DIFF
--- a/tests/trace/test_op_generators.py
+++ b/tests/trace/test_op_generators.py
@@ -3,6 +3,7 @@ from collections.abc import AsyncGenerator, Generator
 import pytest
 
 import weave
+from weave.trace.context import call_context
 
 
 @weave.op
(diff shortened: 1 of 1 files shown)


tests/trace/test_op_generators.py
jtschoonhoven, line 1: These are some high-quality tests ❤️

tests/trace/test_op_generators.py
```diff
@@ -238,3 +239,101 @@ async def test_async_generator_with_decorator_accumulator(client):
     calls = client.get_calls()
     assert len(calls) == 1
     assert calls[0].output == [0, 1, 2]
+
+
+def test_nested_generator_multiple_iterations(client):
```
jtschoonhoven, line 244: Tiny nit: add type annotations?

PR #5915, opened 2026-01-15
fix(weave): Avoid importing IPython at weave startup

shawnlewis:
[The debugging and code writing for this PR were done by codex with gpt-5.2-codex reasoning=high.]

Summary

  - Weave’s is_running_interactively() imported IPython unconditionally. That is called when serializing op code, which happens any time weave traces an op. In CLI contexts this can initialize prompt_toolkit, which mutates terminal mode and breaks TUIs (e.g. Textual apps).
  - Introduce a small helper gate to only query IPython when it is already loaded in sys.modules, and reuse that gate for get_notebook_source().
  - Add tests to cover both interactive and non‑interactive behavior.

  Manual testing

  - Non‑interactive CLI check:

    ./.venv/bin/python - <<'PY'
    import sys
    from weave.utils import ipython as wip
    print("before", "IPython" in sys.modules)
    print("interactive", wip.is_running_interactively())
    print("after", "IPython" in sys.modules)
    PY
    Expected: before False, interactive False, after False (no IPython import side effects).

  - Interactive check in IPython:

    ./.venv/bin/ipython

    import sys
    from weave.utils import ipython as wip
    ("IPython" in sys.modules, wip.is_running_interactively())

    Expected: (True, True).

### DIFF
--- a/weave/utils/ipython.py
+++ b/weave/utils/ipython.py
@@ -1,5 +1,7 @@
 """Utilities for working with IPython/Jupyter notebooks."""
 
+from __future__ import annotations
+
 import ast
 import sys
 from collections.abc import Callable


weave/utils/ipython.py
```diff
@@ -10,21 +11,26 @@ class NotInteractiveEnvironmentError(Exception): ...
...
-def is_running_interactively() -> bool:
-    """Check if the code is running in an interactive environment."""
+def _get_ipython():
+    # Avoid importing IPython in non-interactive CLI runs since it may
+    # initialize prompt_toolkit and mutate terminal mode (e.g., break Textual apps).
+    if "IPython" not in sys.modules:
+        return None
     try:
         from IPython import get_ipython
+    except Exception:
```
zbirenbaum: nit: `except (ImportError, ModuleNotFoundError):`

zbirenbaum (approved):
One nit, otherwise looks good

PR #5917, opened 2026-01-15
chore(weave): fix test flake in poison pill test

gtarpenning:
## Description

Flakes: https://github.com/wandb/weave/actions/runs/21048097194/job/60527622202?pr=5911

Now we don't restart the processor explicitly which was causing issues. 

## Testing

test


bcsherma (approved):
thanks!

PR #5924, opened 2026-01-16
fix(weave): Remove redundant response capture

zbirenbaum:
## Description

- Fixes WB-30413

Currently responses in Agents SDK are double captured. The Agents integration captures a 'Response' and the OpenAI captures the same response as 'openai.response.create'

Latency and cost data is not accessible for the agents integration but the underlying call to OpenAI does have this info

PR removes Agents SDK response capturing and lets the OpenAI integration handle it

docs are not required

## Testing

Added unit tests and ran a script to confirm behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,10 +19,6 @@ You can use Weave to:
 
 Our goal is to bring rigor, best-practices, and composability to the inherently experimental process of developing Generative AI software, without introducing cognitive overhead.
 
-<div align="center">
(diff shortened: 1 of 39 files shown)


andrewtruong:
Seems like some tests are failing, can you please fix those?

zbirenbaum:
> Seems like some tests are failing, can you please fix those?

I think they are fixed. I can't run the one that failed locally but was able to reason through it so should be passing now.

andrewtruong (approved):

PR #5956, opened 2026-01-21
fix(weave): retry Empty query with fresh insert generator

gtarpenning:
## Description

https://wandb.atlassian.net/browse/WB-27557

Attempts to fix a longstanding "Empty query" error that intermittently occurs during ClickHouse inserts. This happens when clickhouse-connect's internal serialization generator gets exhausted during an HTTP connection retry (typically after CH Cloud's 10-second keep-alive timeout causes a connection reset).

Examples from this [week](https://us5.datadoghq.com/apm/traces?query=env%3Aprod%20service%3Aweave-trace%20resource_name%3Aclickhouse_trace_server_batched._flush_calls%20status%3Aerror&agg_m=count&agg_m_source=base&agg_t=count&cols=service%2Cresource_name%2C%40duration%2C%40http.status_code%2C%40_span.count%2C%40_duration.by_service%2C%40weave_trace_server.auth_scope.email%2C%40weave_trace_server.auth_scope.project&fromUser=false&graphType=span_list&historicalData=true&messageDisplay=inline&query_translation_version=v0&refresh_mode=sliding&shouldShowLegend=true&sort=time&spanType=all&spanViewType=logs&storage=hot&traceQuery=&view=spans&start=1768424025275&end=1769028825275&paused=false) (96)

This PR:
- Adds retry logic to `_insert` for "Empty query" errors only (max 3 attempts); Retrying works because the source `data` list is intact - only the internal generator was exhausted
- Adds `INSERT_MAX_RETRIES` constant to `clickhouse_trace_server_settings.py`
- Extracts error handling into explicit helpers: `_raise_if_insert_too_large`, `_should_retry_empty_query`, `_log_and_raise_insert_error`

## Testing

Adds test + existing coverage

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -5199,8 +5199,8 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
 
             # InsertTooLarge: raise immediately, no retry
             except ValueError as e:
-                _raise_if_insert_too_large(e)
-                _log_and_raise_insert_error(e, table, data)
+                converted = _convert_to_insert_too_large(e)
+                _log_and_raise_insert_error(converted, table, data)
 
             # Empty query error: RETRY (generator was consumed during HTTP retry)
             # We should retry with a fresh generator
@@ -5949,18 +5949,25 @@ def _sanitize_name_for_object_id(name: str) -> str:
 # -----------------------------------------------------------------------------
 
 
-def _raise_if_insert_too_large(e: Exception) -> None:
-    """Raise InsertTooLarge if the error indicates data is too large."""
+def _convert_to_insert_too_large(e: Exception) -> Exception:
+    """Convert ValueError to InsertTooLarge if the error indicates data is too large."""
     if isinstance(e, ValueError) and "negative shift count" in str(e):
-        raise InsertTooLarge(
+        return InsertTooLarge(
             "Database insertion failed. Record too large. "
             "A likely cause is that a single row or cell exceeded "
             "the limit. If logging images, save them as `Image.PIL`."
-        ) from e
+        )
+    return e
 
 
 def _should_retry_empty_query(e: Exception, table: str, attempt: int) -> bool:
-    """Check if we should retry an empty query error. Logs warning if retrying."""
+    """Check if we should retry an empty query error. Logs warning if retrying.
+
+    Attempts to fix a longstanding "Empty query" error that intermittently
+    occurs during ClickHouse inserts. This happens when clickhouse-connect's
+    internal serialization generator gets exhausted during an HTTP connection
+    retry (after CH Cloud's keep-alive timeout causes a connection reset).
+    """
     is_empty_query = isinstance(e, DatabaseError) and "Empty query" in str(e)
     should_retry = is_empty_query and attempt < ch_settings.INSERT_MAX_RETRIES - 1
     if should_retry:
@@ -5989,4 +5996,4 @@ def _log_and_raise_insert_error(
             "data_bytes": data_bytes,
         },
     )
-    raise
+    raise e


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5189,38 +5190,28 @@ def _insert(
...
+
+        for attempt in range(ch_settings.INSERT_MAX_RETRIES):
+            try:
+                return self.ch_client.insert(
+                    table, data=data, column_names=column_names, settings=settings
+                )
+
+            # InsertTooLarge: raise immediately, no retry
+            except ValueError as e:
+                _raise_if_insert_too_large(e)
```
mscavezze-cw: It seems odd to me to have code after something with `raise` in the name. I wonder of this would be a better pattern
```
e = convert_clickhouse_exception
_log_and_raise_insert_error(e, table, data)
```

That would eliminate the partially hidden conditional that is inside of _raise_if_insert_too_large. It would also route all of these exceptions through a common flow, so that logging would occur. I assume that is desirable.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5951,3 +5942,51 @@ def _setup_completion_model_info(
...
+            },
+        )
+    return should_retry
+
+
+def _log_and_raise_insert_error(
+    e: Exception, table: str, data: Sequence[Sequence[Any]]
+) -> None:
+    """Log insert error with data size info and re-raise."""
+    data_bytes = sum(_num_bytes(row) for row in data)
```
mscavezze-cw, line 5989: The comment about this being an expensive operation is being removed. Was that inaccurate? If not, consider retaining the comment.

If the presence of the expensive operation prevents others from reusing this function, consider adding a parameter like `include_expensive_size_calculation` and making the calculation optional.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5951,3 +5942,51 @@ def _setup_completion_model_info(
...
+    if isinstance(e, ValueError) and "negative shift count" in str(e):
+        raise InsertTooLarge(
+            "Database insertion failed. Record too large. "
+            "A likely cause is that a single row or cell exceeded "
+            "the limit. If logging images, save them as `Image.PIL`."
+        ) from e
+
+
+def _should_retry_empty_query(e: Exception, table: str, attempt: int) -> bool:
+    """Check if we should retry an empty query error. Logs warning if retrying."""
```
mscavezze-cw: Consider adding some detail from the pr description here:
```
Attempts to fix a longstanding "Empty query" error that intermittently occurs during ClickHouse inserts. This happens when clickhouse-connect's internal serialization generator gets exhausted during an HTTP connection retry (typically after CH Cloud's 10-second keep-alive timeout causes a connection reset).
```

mscavezze-cw (approved):

weave/trace_server/clickhouse_trace_server_batched.py
mscavezze-cw: > It would also route all of these exceptions through a common flow, so that logging would occur. I assume that is desirable.

Maybe this is a bad assumption, if the `too large` error can happen in response to bad user input. That could add a bunch of noise.

gtarpenning: Yeah that makes perfect sense.

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: eh it's not that expensive in retrospect. we are doing much more expensive stuff in other places... Good callout though.

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: too large error is very rare, and we are protecting against it in other places so shouldn't be deterministically triggered.

PR #5957, opened 2026-01-21
feat(weave): Add /trace/usage endpoint to dynamically aggregate usage data

andrewtruong:
https://wandb.atlassian.net/browse/WB-30250

---

Pair with: https://github.com/wandb/core/pull/38473

This PR adds an endpoint to dynamically aggregate usage metrics, like tokens and costs, for the trace details view.  A second PR (https://github.com/wandb/weave/pull/5995) introduces an endpoint for dynamically aggregating usage metrics for the trace table view.  This is useful for cases where calls are created outside of our SDKs meaning client-side pre-aggregation is not possible.

The new endpoint `/trace/usage` returns recursively aggregated tokens and costs for all calls in the trace.

This implementation queries for all the relevant calls then does the aggregation on the server.  An alternate implementation could use a recursive CTE (see experiment: https://github.com/wandb/weave/pull/5938), but is not included in this PR.

In most cases where this endpoint is used, a `call_stream_query` is probably already issued.  This is an optimization that can be made in future, but is not considered for this MVP.

### DIFF
--- a/tests/trace_server/test_trace_usage.py
+++ b/tests/trace_server/test_trace_usage.py
@@ -1,34 +1,12 @@
 """Tests for the trace_usage endpoint."""
 
-import random
-import time
 from datetime import datetime, timezone
+import uuid
 
 import pytest
 
 from tests.trace_server.conftest import TEST_ENTITY
 from weave.trace_server import trace_server_interface as tsi
-
-
-def uuid7() -> str:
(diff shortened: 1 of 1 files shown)


weave/trace_server_bindings/stainless_remote_http_trace_server.py
andrewtruong, line 1: will fix this later this sprint with the Stainless work

PR #5961, opened 2026-01-21
chore(weave): Add some helpful instructions for agents

andrewtruong:
This PR adds some helpful instructions for agents to better work with our codebase

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,6 +6,7 @@
 - If there is something that doesn't make sense architecturally, devex-wise, or product-wise, please update the `Requests to Humans` section below.
 - Always follow the established coding patterns and conventions in the codebase.
 - Document any significant architectural decisions or changes.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -35,13 +36,44 @@ def my_function():
...
+# Install for trace server development
+uv sync --group test --group trace_server --group trace_server_tests
+```
+
+### Available Dependency Groups
+
+- `dev` - Development tools (pre-commit, linting, codegen)
+- `test` - Test framework (pytest, nox, coverage)
+- `trace_server` - Trace server runtime dependencies
+- `trace_server_tests` - Additional trace server test dependencies
```
jtschoonhoven, line 69: These sections contain overlapping info and could probably be tightened up a bit. An agent should probably _always_ have `dev` and `test`, so maybe we should explicitly tell it to install those groups and then make it aware of extra deps it might need.

Would it be worth just telling the agent to install all groups?

PR #6023, opened 2026-01-30
feat(weave): Add basic weave CLI with login that mirrors wandb login

andrewtruong:
This PR simply adds a CLI with one command called `login` that mirrors `wandb login`

Eventually the CLI will have other more useful commands!


tests/cli/test_login.py
bcsherma, line 1: are these running in ci?

PR #6073, opened 2026-02-06
feat(weave): Add wb_username resolution in calls queries

scottire:
## Summary
Adds automatic `wb_username` resolution for calls where `wb_user_id` matches the authenticated user. Username hydration happens entirely in the adapter layer - no core repo changes needed.

## Changes
- `weave/trace_server/trace_server_interface.py`: Add `wb_username` field to `CallSchema`
- `weave/trace_server/external_to_internal_trace_server_adapter.py`:
  - Add `get_auth_user()` method to `IdConverter` (returns `None` by default)
  - Resolve `wb_username` in `call_read`, `calls_query`, and `calls_query_stream` when user matches
- `weave/trace_server/sqlite_trace_server.py`: Exclude `wb_username` from SELECT columns (computed field, not stored)
- `tests/trace_server/conftest_lib/trace_server_external_adapter.py`: Add test support (`DummyAuthUser`, `get_auth_user()`)
- `tests/trace_server/test_username_resolution.py`: Add unit tests (5 tests covering all query methods)

## How it works
1. `IdConverter.get_auth_user()` returns the authenticated user (or `None`)
2. In `call_read`, `calls_query`, and `calls_query_stream`, after converting `wb_user_id`, check if it matches `auth_user.id`
3. If match, set `wb_username = auth_user.username`

## Core repo requirement
The core repo's `AuthenticatedConverter` needs to implement `get_auth_user()` to return the auth user from `authentication.maybe_get_auth_user(auth_params)`.

~~## Related PR~~
~~Core repo PR for AuthenticatedConverter implementation: [link]~~


neutralino1 (commented):
If adding the username to the response is cheap (i.e. no new queries to Gorilla), why not do it by default and not add another flag?

gtarpenning:
we get `username` for free in the `_auth_user_query` query, it should be in auth scope. Lets add this in the `trace_server.py` calls query method.

tests/trace_server/test_username_resolution.py
```diff
@@ -0,0 +1,244 @@
+"""Tests for wb_username resolution from auth scope in calls query."""
```
gtarpenning: probably don't want a whole new test file just for this one field, where do we do the other auth field assertions/checks?

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -35,6 +35,13 @@ def ext_to_int_user_id(self, user_id: str) -> str:
     def int_to_ext_user_id(self, user_id: str) -> str:
         raise NotImplementedError()
 
+    def get_auth_user(self) -> typing.Any:
+        """Override to provide auth user for username resolution.
+
+        Returns an object with `id` and `username` attributes, or None.
+        """
+        return None
```
gtarpenning, line 44: ?? what is this for

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -151,6 +158,10 @@ def call_read(self, req: tsi.CallReadReq) -> tsi.CallReadRes:
             res.call.wb_run_id = self._idc.int_to_ext_run_id(res.call.wb_run_id)
         if res.call.wb_user_id is not None:
             res.call.wb_user_id = self._idc.int_to_ext_user_id(res.call.wb_user_id)
+            # Resolve username from auth scope when user matches
+            auth_user = self._idc.get_auth_user()
+            if auth_user and res.call.wb_user_id == auth_user.id:
+                res.call.wb_username = auth_user.username
```
gtarpenning: does this mean users can only resolve their own username? that seems weird.

weave/trace_server/trace_server_interface.py
```diff
@@ -131,6 +124,7 @@ class CallSchema(BaseModel):
 
     # WB Metadata
     wb_user_id: str | None = None
+    wb_username: str | None = None  # Resolved username from wb_user_id
```
gtarpenning: remove this comment, i think it's actually wrong, the username is not directly resolved from the `wb_user_id`

PR #6086, opened 2026-02-09
fix(weave): create trace spans for argument binding failures in weave.op

adamwdraper:
This is one of a series of pr's I am producing from different Coding agents for comparison for fixing the same bug: Fixes [WB-30436](https://wandb.atlassian.net/browse/WB-30436)

## Agent
Conductor: Claude Opus 4.6 Thinking w/plan

## Comparison PRs
#6084 

## Summary

Fixes [WB-30436](https://wandb.atlassian.net/browse/WB-30436)

- When a `@weave.op` decorated function is called with wrong arguments (e.g., LLM sends wrong parameter names), `sig.bind()` fails with `TypeError` which gets wrapped in `OpCallError`. Previously this error was raised **before** a Call object was created, so no trace span appeared in the UI, making debugging agent behavior difficult.
- Added a `_handle_op_call_error` helper that catches `OpCallError` in all 4 `_call_*` functions (sync, async, sync gen, async gen), creates a Call with the raw kwargs as inputs, finishes it with the error, then re-raises. The helper is wrapped in try/except so tracing errors never mask the original error.
- Updated existing tests and added 5 new tests covering all function types and error detail recording.

## Test plan
- [x] All 33 tests in `tests/trace/test_op_decorator_behaviour.py` pass
- [x] All 208 broader op tests pass (`tests/trace/ -k "op"`)
- [x] `OpCallError` is still raised to the caller
- [x] Trace span shows raw inputs and the exception
- [ ] CI passes

[WB-30436]: https://wandb.atlassian.net/browse/WB-30436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/test_op_decorator_behaviour.py
+++ b/tests/trace/test_op_decorator_behaviour.py
@@ -142,6 +142,12 @@ def test_sync_method_call(client, weave_obj, py_obj):
     with pytest.raises(OpCallError):
         res2, call2 = py_obj.method.call(1)
 
+    # Verify a trace span was created for the failed call
+    calls = list(client.get_calls())
+    error_calls = [c for c in calls if c.exception is not None]
+    assert len(error_calls) == 1
+    assert "OpCallError" in error_calls[0].exception or "Error calling" in error_calls[0].exception
+
 
 @pytest.mark.asyncio
 async def test_async_method(client, weave_obj, py_obj):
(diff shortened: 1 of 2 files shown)


andrewtruong:
Sounds like the use case here is detecting that an agent hallucinated args to a func?  I can see the value in that.

I think we can accept something like this, and it might warrant its own error state.

adamwdraper:
@andrewtruong to clarify — when you say "it might warrant its own error state," are you thinking something like:

1. A distinguishing attribute on the call (e.g., `attributes.weave.binding_error = True`) so the UI/API can tell "binding failed before execution" apart from "function ran and threw"
2. A new first-class error state on the Call object itself (separate from `exception`)
3. Something else?

Currently the span uses the existing `exception` field, which looks the same as any runtime error. Happy to add a distinguishing marker if that's what you have in mind.

PR #6097, opened 2026-02-10
perf(weave): pass [] to sort_by for unsorted query

gtarpenning:
## Description

Handles the empty list case differently than `None`, now taking this as an explicit request for no sort order. This can be a major optimization when doing needle in the haystack searches, where you know that the result set is small, and don't care about the order. Anecdotally this can bring up to a 10x speedup in certain searches, like filtering by the expensive inputs/output fields.

## Testing

- adds unit tests
- adds end to end functional tests


gtarpenning:

PR #6104, opened 2026-02-11
feat(weave): Implement client-side mutation-tracking API

andrewtruong:
This PR completes the client-side implementation of mutations API by adding `Traceable.save` and implementing the relevant wiring

There is no user-facing impact today as it's not defined on the server yet, but it will be useful later to improve upload perf


neutralino1:
What is the purpose of this change?

PR #6131, opened 2026-02-13
chore(weave): strip large payloads to content, dont drop

gtarpenning:
## Summary
- Stop dropping oversized call payloads: offload large string leaves to Content (file) storage before insert, stripping to the `<EXCEEDS_LIMITS>` sentinel only as a last resort.
- Runs proactively before every insert (calls / calls_complete / OTEL) at a 1 MiB JSON budget; offloads progressively (large then medium leaves) so payloads made of many medium strings are preserved, not dropped.
- Keeps the `InsertTooLarge` -> row-by-row fallback so a single oversized row can't fail the whole batch.

## Performance
A cheap `chars * 4` upper bound skips the full byte-encode on the common small-row path (master did no measurement here at all). Heavy ~80 KiB trace rows: ~3.9 -> ~0.7 us/row on the offload gate.

## Testing
offload functional tests in clickhouse mode + unit coverage for progressive offload and the strip fallback.

### DIFF
--- a/weave/trace_server/base64_content_conversion.py
+++ b/weave/trace_server/base64_content_conversion.py
@@ -212,12 +212,12 @@ def replace_large_strings_with_content_objects(
     vals: T,
     project_id: str,
     trace_server: TraceServerInterface,
-    min_chars: int,
+    max_chars: int,
 ) -> T:
     """Recursively replace large string values with Content objects in file storage.
 
     Walks the value tree and converts any string leaf whose character count exceeds
-    ``min_chars`` into a Content object stored via the trace server's file storage.
+    ``max_chars`` into a Content object stored via the trace server's file storage.
     The original string is preserved as ``text/plain`` content, and a Content
     reference dict is returned in its place.
 
@@ -231,7 +231,7 @@ def replace_large_strings_with_content_objects(
             return {k: _visit(v) for k, v in val.items()}
         elif isinstance(val, list):
             return [_visit(v) for v in val]
-        elif isinstance(val, str) and len(val) > min_chars:
+        elif isinstance(val, str) and len(val) > max_chars:
             try:
                 content: Content[Any] = Content.from_text(val)
                 return store_content_object(content, project_id, trace_server)
@@ -240,6 +240,7 @@ def replace_large_strings_with_content_objects(
                     "Failed to offload large string (%d chars) to content storage: %s",
                     len(val),
                     e,
+                    exc_info=True,
                 )
                 return val
         return val
--- a/weave/trace_server/base64_content_conversion.py
+++ b/weave/trace_server/base64_content_conversion.py
@@ -233,7 +233,7 @@ def replace_large_strings_with_content_objects(
             return [_visit(v) for v in val]
         elif isinstance(val, str) and len(val) > min_chars:
             try:
-                content = Content.from_text(val)
+                content: Content[Any] = Content.from_text(val)
                 return store_content_object(content, project_id, trace_server)
             except Exception as e:
                 logger.warning(
--- a/weave/trace_server/base64_content_conversion.py
+++ b/weave/trace_server/base64_content_conversion.py
@@ -1,7 +1,10 @@
-"""Base64 content conversion utilities for trace server.
+"""Content conversion utilities for trace server.
 
-This module handles automatic detection and replacement of base64 encoded content
-with content objects stored in bucket storage.
+This module handles automatic detection and replacement of large or encoded content
+with Content objects stored in bucket storage. Supports:
+- Base64 encoded strings (detected by pattern matching)
+- Data URI strings (e.g., data:image/png;base64,...)
+- Arbitrary large strings (offloaded when payloads exceed ClickHouse row limits)
 """
 
 import json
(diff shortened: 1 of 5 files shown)


weave/trace_server/base64_content_conversion.py
```diff
@@ -174,6 +178,46 @@ def _visit(val: Any) -> Any:
     return _visit(vals)
 
 
+@ddtrace.tracer.wrap(name="replace_large_strings_with_content_objects")
+def replace_large_strings_with_content_objects(
+    vals: T,
+    project_id: str,
+    trace_server: TraceServerInterface,
+    min_chars: int,
```
jtschoonhoven: Tiny nit: should this be `max_chars`? That makes more sense to me. Feel free to ignore.

weave/trace_server/base64_content_conversion.py
```diff
@@ -174,6 +178,46 @@ def _visit(val: Any) -> Any:
...
+        elif isinstance(val, str) and len(val) > min_chars:
+            try:
+                content: Content[Any] = Content.from_text(val)
+                return store_content_object(content, project_id, trace_server)
+            except Exception as e:
+                logger.warning(
+                    "Failed to offload large string (%d chars) to content storage: %s",
+                    len(val),
+                    e,
+                )
```
jtschoonhoven, line 244: Just confirming you want to log at `warning` level and not `exception`? If using `warning` it's probably a good idea to capture the exception info in the log:
```suggestion
            except Exception as e:
                logger.warning(
                    "Failed to offload large string (%d chars) to content storage: %s",
                    len(val),
                    e,
                    exc_info=True,
                )
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6046,72 +6046,143 @@ def _flush_calls_complete(self) -> None:
...
+                if total_json_bytes <= row_limit:
                     break
 
-                # Replace this large JSON value with placeholder, update running size
-                stripped_item[col_idx] = ch_settings.ENTITY_TOO_LARGE_PAYLOAD
-                total_json_bytes -= size - entity_too_large_payload_byte_size
-                stripped_count += 1
+                current_col_size = original_size
+
+                # Try to offload large strings to Content objects
```
jtschoonhoven: Nit: not the worst, but these loops/conditionals are getting nested enough that it might be a good idea to factor out more of the inner logic into its own function for test/readability

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5964,13 +5967,10 @@ def _insert_call(self, ch_call: CallCHInsertable) -> None:
...
-        except InsertTooLarge:
-            logger.info("Retrying with large objects stripped.")
-            batch = self._strip_large_values(self._call_batch)
-            # Insert rows one at a time after stripping large values
-            for row in batch:
-                self._insert_call_batch([row])
+            batch = self._offload_large_values(
+                self._call_batch, ALL_CALL_INSERT_COLUMNS
+            )
+            self._insert_call_batch(batch)
```
jtschoonhoven: Are you sure it's safe to remove handling for `InsertTooLarge`? Is `_offload_large_values` guaranteed to prevent the `InsertTooLarge` error?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6046,72 +6046,143 @@ def _flush_calls_complete(self) -> None:
...
+) -> str | None:
+    """Try to reduce a JSON column by offloading large string leaves to Content storage.
+
+    Parses the JSON dump, replaces string values whose character count exceeds
+    ``min_chars`` with Content object references stored in file storage, and
+    re-serializes.
+    """
+    try:
+        parsed = json.loads(json_dump)
+        processed = replace_large_strings_with_content_objects(
```
jtschoonhoven, line 7548: Do we need to worry about the performance implications of parsing large json objects + the recursive tree walk? Looks like this potentially gets called for every trace on flush.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6046,72 +6046,143 @@ def _flush_calls_complete(self) -> None:
...
-                total_json_bytes -= size - entity_too_large_payload_byte_size
-                stripped_count += 1
+                current_col_size = original_size
+
+                # Try to offload large strings to Content objects
+                new_dump = _try_offload_json_column(
+                    processed_item[col_idx],
+                    processed_item[project_id_idx],
+                    self,
+                    min_chars=ch_settings.LARGE_STRING_OFFLOAD_MIN_CHARS,
```
jtschoonhoven: Nit: there's an edge case where a row contains many string leaves that are individually under the `LARGE_STRING_OFFLOAD_MIN_CHARS` threshold, but in total they still exceed `PROACTIVE_OFFLOAD_BYTES_LIMIT`. Flagging it, but I'm not sure it's actually worth addressing.

jtschoonhoven (approved):
Mostly nits and minor suggestions. Looks great!

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: i think this case will be super rare, but yes technically totally true

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: done

weave/trace_server/base64_content_conversion.py
gtarpenning: yep, done

weave/trace_server/base64_content_conversion.py
gtarpenning: yep

gtarpenning:
Closing as won't-do.

PR #6179, opened 2026-02-20
chore(weave): add TTL settings LRU cache module

gtarpenning:
## Summary

- Two-layer TTL settings cache (L1 in-process TTLCache, L2 Redis) for per-project retention lookups
- argMax-based ClickHouse query avoids FINAL, with graceful Redis failure fallback
- `compute_expire_at` helper handles zero/positive/negative retention_days
- Datadog tracing on cache hits (L1/L2/L3 tags)

NO hot code path changes, this new code should not be hit yet.

PR #6178, opened 2026-02-20
chore(weave): add ttl to CH insertable types

gtarpenning:
## Description

Adds clickhouse types for new `expire_at` ttl settings

## Testing

existing coverage

Manual testing in qa

PR #6177, opened 2026-02-20
chore(weave): add TTL migration

gtarpenning:
## Description

TTL! For calls! Add a new table: `project_ttl_settings` where you can set a TTL at a project level, currently for traces. This is not a one-way door, we can always expand this table to include information about specific Ops, or different types like Feedback or Objects. 

More information in the design spec here: [https://www.notion.so/wandbai/TTL-Spec-314e2f5c7ef38093b5e3e5bd078bf89f](https://www.notion.so/wandbai/TTL-Spec-314e2f5c7ef38093b5e3e5bd078bf89f?source=copy_link)

## Testing

nope! (not yet, later in stack)

PR #6180, opened 2026-02-20
chore(weave): wire TTL computation into insert paths

gtarpenning:
## Description

[WB-33319](https://coreweave.atlassian.net/browse/WB-33319)

Wires project-level TTL computation into the Clickhouse insert paths, adds SQLite parity, and adds end-to-end coverage.

- `expire_at` is populated at call-start / call-end / `calls_complete` from the resolved TTL policy (retention_days → days, 0 → 2100-01-01 sentinel, negative → minutes).
- Adds an `expire_at` column + `project_ttl_settings` table to SQLite so the backend behaves the same way.
- Adds 4 parametrized end-to-end tests (retention = 30 days, 0 sentinel, -5 minutes) that set a project TTL, insert via each endpoint (`call_start`+`call_end`, `call_start_batch`, `calls_complete`, `call_start_v2`+`call_end_v2`), and raw-read `expire_at` from the backend to verify. Tests run against both SQLite and ClickHouse.

## Testing
- Adds tests
- Existing coverage
- Manual qa

[WB-33319]: https://coreweave.atlassian.net/browse/WB-33319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -272,7 +272,6 @@ from weave.trace_server.trace_server_common import (
 from weave.trace_server.ttl_settings import (
     compute_expire_at,
     get_project_retention_days,
-    invalidate_ttl_cache,
 )
 from weave.trace_server.workers.evaluate_model_worker.evaluate_model_worker import (
     EvaluateModelArgs,
@@ -808,6 +807,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         if write_target == WriteTarget.CALLS_COMPLETE:
             raise CallsCompleteModeRequired(ch_call.project_id)
 
+        self._add_call_expire_at(ch_call, ch_call.started_at)
         # Inserts the call into the clickhouse database, verifying that
         # the call does not already exist
         self._insert_call(ch_call)
@@ -837,6 +837,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         if write_target == WriteTarget.CALLS_COMPLETE:
             raise CallsCompleteModeRequired(ch_call.project_id)
 
+        self._add_call_expire_at(ch_call, ch_call.ended_at)
         # Inserts the call into the clickhouse database, verifying that
         # the call does not already exist
         self._insert_call(ch_call)
@@ -892,6 +893,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
                 )
 
                 ch_call = complete_call_to_ch_insertable(processed_complete_call)
+                self._add_call_expire_at(ch_call, ch_call.started_at)
                 if write_target == WriteTarget.CALLS_COMPLETE:
                     self._insert_call_complete(ch_call)
                 else:
@@ -921,8 +923,10 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         )
         if write_target == WriteTarget.CALLS_COMPLETE:
             ch_complete_start = start_call_insertable_to_complete_start(ch_start)
+            self._add_call_expire_at(ch_complete_start, ch_complete_start.started_at)
             self._insert_call_complete(ch_complete_start)
         else:
+            self._add_call_expire_at(ch_start, ch_start.started_at)
             self._insert_call(ch_start)
 
         return tsi.CallStartV2Res(id=ch_start.id, trace_id=ch_start.trace_id)
@@ -952,6 +956,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
             self._update_call_end_in_calls_complete(req.end)
         elif write_target == WriteTarget.CALLS_MERGED:
             ch_end = end_call_for_insert_to_ch_insertable(req.end)
+            self._add_call_expire_at(ch_end, ch_end.ended_at)
             self._insert_call(ch_end)
             if self._flush_immediately:
                 self._flush_calls()
@@ -6146,6 +6151,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
                 wb_user_id=req.wb_user_id,
             )
             ch_call = complete_call_to_ch_insertable(completed)
+            self._add_call_expire_at(ch_call, ch_call.started_at)
             self._insert_call_complete(ch_call)
         else:
             # Write to call_parts/calls_merged via start/end pattern
@@ -6281,8 +6287,12 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
             # Insert immediately so that callers can see the call in progress
             if write_target == WriteTarget.CALLS_COMPLETE:
                 ch_complete_start = start_call_insertable_to_complete_start(start_call)
+                self._add_call_expire_at(
+                    ch_complete_start, ch_complete_start.started_at
+                )
                 self._insert_call_complete(ch_complete_start)
             else:
+                self._add_call_expire_at(start_call, start_call.started_at)
                 self._insert_call(start_call)
 
         # Set the combined messages (with template vars replaced) for LiteLLM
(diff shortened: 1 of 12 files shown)


gtarpenning:
* **#6180** <a href="https://app.graphite.com/github/pr/wandb/weave/6180?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6181](https://github.com/wandb/weave/pull/6181) <a href="https://app.graphite.com/github/pr/wandb/weave/6181?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6677](https://github.com/wandb/weave/pull/6677) <a href="https://app.graphite.com/github/pr/wandb/weave/6677?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6180?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6179** <a href="https://app.graphite.com/github/pr/wandb/weave/6179?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6178** <a href="https://app.graphite.com/github/pr/wandb/weave/6178?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6177** <a href="https://app.graphite.com/github/pr/wandb/weave/6177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6741,6 +6746,19 @@ def _insert(
 
     @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched._insert_call")
     def _insert_call(self, ch_call: CallCHInsertable) -> None:
+        ttl_anchor = None
+        if isinstance(ch_call, CallStartCHInsertable):
+            ttl_anchor = ch_call.started_at
+        elif isinstance(ch_call, CallEndCHInsertable):
+            ttl_anchor = ch_call.started_at
```
gtarpenning: call ends dont have a started_at, but we will still want to expire them along with the start, maybe we should use ended_at?

gtarpenning: Good catch — switched end rows to anchor on ended_at. Also (per your other ask) moved expire_at stamping out of _insert_call into each public entry point via _add_call_expire_at, and flipped get_project_retention_days to a MintClient factory so the lookup runs on its own CH session. 9384f92241

weave/trace_server/ttl_settings.py
```diff
@@ -46,14 +51,18 @@
...
 ) -> int:
     """Return retention_days for a project (0 = no TTL / infinite). Cached.
 
     Read path: L1 (in-process) -> L2 (Redis, optional) -> ClickHouse (argMax).
     A fall-through to ClickHouse is a full cache miss; if ClickHouse has no
     row for the project, returns 0 (no TTL configured).
+
+    On cache miss we mint a fresh ClickHouse client with its own session
+    rather than reusing the caller's thread-local client, to avoid
+    SESSION_IS_LOCKED when this lookup runs alongside an ongoing insert.
```
jtschoonhoven, line 65: Probably out of scope for this PR, but this seems like a weird bandaid. Is there a more permanent fix? Maybe add a TODO?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -816,6 +821,7 @@ def call_end(
         if write_target == WriteTarget.CALLS_COMPLETE:
             raise CallsCompleteModeRequired(ch_call.project_id)
 
+        self._add_call_expire_at(ch_call, ch_call.ended_at)
```
jtschoonhoven, line 824: Makes me a little nervous that the `call_start` expires before `call_end`. Can we have them expire at the same time? The sqlite implementation uses a single `expire_at`.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6749,6 +6769,17 @@ def _insert_call(self, ch_call: CallCHInsertable) -> None:
...
+    def _add_call_expire_at(
+        self,
+        ch_call: CallStartCHInsertable | CallEndCHInsertable | CallCompleteCHInsertable,
+        anchor: datetime.datetime,
+    ) -> None:
+        """Add expire_at to the insertable from the project's retention policy."""
+        retention_days = get_project_retention_days(
+            ch_call.project_id, self._mint_client
+        )
+        ch_call.expire_at = compute_expire_at(retention_days, anchor)
```
jtschoonhoven, line 6781: Forgive me if you explained this somewhere and I missed it, but why do we need to store the expiration on the call itself? Isn't this trivial to derive from the call's timestamp and the project's TTL? What happens if a customer changes the TTL for their project?

weave/trace_server/sqlite_trace_server.py
```diff
@@ -469,6 +482,26 @@ def setup_tables(self) -> None:
...
+        row = cursor.fetchone()
+        return int(row[0]) if row is not None else 0
+
+    def _compute_call_expire_at(
+        self, project_id: str, anchor: datetime.datetime
+    ) -> str:
+        """Compute the ISO-formatted expire_at for a call row."""
+        retention_days = self._get_project_retention_days(project_id)
+        if retention_days == 0:
+            return EXPIRE_AT_NEVER.isoformat(sep=" ")
```
jtschoonhoven, line 502: Nit: this check is redundant with `compute_expire_at`

jtschoonhoven (commented):
A few questions before approving

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: at the db layer when they get combined into `calls_merged` we pick `min(start.expired_at, end.expired_at)` so for 99% of calls they get expired at the same time for the user. The individual call parts are really only relevant at the db layer, so i dont think a couple seconds really matters.

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: we want to store it on the call so the database engine itself does the clean up. This way we can have exactly one TTL statement in clickhouse `TTL expire_at DELETE` which uses the field on the call to clean up when appropriate. In a mtsaas environment we can now have 100s of different projects all with different TTLs, but all that complexity gets pushed down to the db layer, all we need to do is stash the appropriate value on the call.

weave/trace_server/ttl_settings.py
gtarpenning: yeah also my fix to mint new sessions actually didn't completely resolve this issue in prod so i'll revert this .

weave/trace_server/clickhouse_trace_server_batched.py
jtschoonhoven: SGTM

weave/trace_server/clickhouse_trace_server_batched.py
jtschoonhoven: That makes sense. How do we handle the case where the user changes the TTL for their project? Would we issue a giant UPDATE or is that just not supported?

gtarpenning: we could do that, which shouldn't be that bad as its only targeting exactly one small metadata column, however its also out of scope for mvp.

weave/trace_server/ttl_settings.py
gtarpenning: completely refactored this out

weave/trace_server/clickhouse_trace_server_batched.py
jtschoonhoven: Makes sense to keep it out of scope for the MVP. It should get fixed at some point though. I'd recommend adding a TODO comment that links to a JIRA ticket so we don't forget. I don't want customers to be surprised by data loss.

jtschoonhoven (commented):
Ready to approve once we decide whether `_add_call_expire_at` belongs inside `_insert_call`.

jtschoonhoven (approved):
Sic dope

PR #6181, opened 2026-02-20
chore(weave): add TTL settings GET/POST API endpoints

gtarpenning:
## Description

Adds project-level TTL settings APIs for trace retention. Retention uses `None` at the API/app layer to mean no TTL, while ClickHouse/SQLite keep `expire_at` non-null by converting `None` to the storage sentinel (`2100-01-01`) at the DB boundary.

Includes:
- per-project TTL read/update handlers with audit user validation
- cached retention lookup and insert-path `expire_at` computation
- sentinel conversion for `expire_at` on CH/SQLite write/read paths
- focused coverage for TTL settings, insert paths, and sentinel conversion

## Testing

- lots of unit tests
- end to end tests as well


gtarpenning:
* **#6181** <a href="https://app.graphite.com/github/pr/wandb/weave/6181?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6181?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6180** <a href="https://app.graphite.com/github/pr/wandb/weave/6180?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6677](https://github.com/wandb/weave/pull/6677) <a href="https://app.graphite.com/github/pr/wandb/weave/6677?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6179** <a href="https://app.graphite.com/github/pr/wandb/weave/6179?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6178** <a href="https://app.graphite.com/github/pr/wandb/weave/6178?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6177** <a href="https://app.graphite.com/github/pr/wandb/weave/6177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/trace_server_interface.py
```diff
@@ -1555,6 +1558,28 @@ class ProjectStatsRes(BaseModel):
...
+# TTL Settings API
+# ================
+
+
+class ProjectTTLSettingsReq(BaseModelStrict):
+    project_id: str
+
+
+class ProjectTTLSettingsRes(BaseModel):
+    retention_days: int  # 0 = no TTL (infinite retention)
```
jtschoonhoven, line 1570: Why not have `null` = "infinite retention"? Seems more intuitive and less likely to accidentally cause massive data loss.

```suggestion
    retention_days: int | None  # None = no TTL (infinite retention)
```

weave/trace_server/trace_server_interface.py
```diff
@@ -1555,6 +1558,28 @@ class ProjectStatsRes(BaseModel):
...
+
+
+class SetProjectTTLSettingsReq(BaseModelStrict):
+    project_id: str
+    retention_days: int  # 0 = disable TTL; must be 0 or >= 1
+    wb_user_id: str | None = None
+
+
+class SetProjectTTLSettingsRes(BaseModel):
+    retention_days: int
```
jtschoonhoven, line 1580: Why not `retention_seconds`? Seems more flexible. We can still show users "days" in the UI.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2617,6 +2617,42 @@ def _default_true(val: bool | None) -> bool:
...
+    def project_ttl_settings_read(
+        self, req: tsi.ProjectTTLSettingsReq
+    ) -> tsi.ProjectTTLSettingsRes:
+        retention_days = get_project_retention_days(req.project_id, self.ch_client)
+        return tsi.ProjectTTLSettingsRes(retention_days=retention_days)
+
+    @ddtrace.tracer.wrap(
+        name="clickhouse_trace_server_batched.project_ttl_settings_set"
+    )
+    def project_ttl_settings_set(
```
jtschoonhoven, line 2632: Use `_update` for consistency with the rest of the module.

```suggestion
    def project_ttl_settings_update(
```

weave/trace_server/sqlite_trace_server.py
```diff
@@ -2792,6 +2792,20 @@ def project_stats(self, req: tsi.ProjectStatsReq) -> tsi.ProjectStatsRes:
             "project_stats is not implemented for SQLite trace server"
         )
 
+    def project_ttl_settings_read(
+        self, req: tsi.ProjectTTLSettingsReq
+    ) -> tsi.ProjectTTLSettingsRes:
+        raise NotImplementedError(
+            "project_ttl_settings_read is not implemented for SQLite trace server"
+        )
```
jtschoonhoven, line 2800: I thought in the other PR you added support for this?

weave/trace_server/trace_server_interface.py
```diff
@@ -155,6 +155,9 @@ class CallSchema(BaseModel):
 
     deleted_at: datetime.datetime | None = None
 
+    # Expiration timestamp for this call (sentinel 2100-01-01 = no expiry)
+    expire_at: datetime.datetime | None = None
```
jtschoonhoven, line 159: Are we using both `None` _and_ `2100-01-01` to mean "no expiry"? I prefer `None`.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1878,14 +1878,15 @@ def get_calls_table_name(read_table: ReadTable) -> str:
         join_table_name=ROLLED_UP_CALL_MERGED_STATS_TABLE_NAME,
     ),
     "otel_dump": CallsMergedAggField(field="otel_dump", agg_fn="any"),
+    "expire_at": CallsMergedAggField(field="expire_at", agg_fn="min"),
```
jtschoonhoven, line 1881: Nit: shouldn't matter in practice, but `max` seems marginally safer

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2617,6 +2617,42 @@ def _default_true(val: bool | None) -> bool:
...
+        self, req: tsi.SetProjectTTLSettingsReq
+    ) -> tsi.SetProjectTTLSettingsRes:
+        # Negative values encode minutes-based retention (e.g. -5 = 5 minutes) for testing
+        if req.retention_days < -1440:
+            raise InvalidRequest(
+                "retention_days must be 0 (no TTL), >= 1, or negative for minutes-based retention (min -1440)"
+            )
```
jtschoonhoven, line 2639: This is a little too surprising. Switching to retention_seconds would be more obvious.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2617,6 +2617,42 @@ def _default_true(val: bool | None) -> bool:
...
+
+        self.ch_client.insert(
+            "project_ttl_settings",
+            data=[
+                [
+                    req.project_id,
+                    req.retention_days,
+                    datetime.datetime.now(datetime.UTC),
+                    req.wb_user_id or "",
```
jtschoonhoven, line 2648: Why do we allow this to be empty?

jtschoonhoven (commented):
Some comments before approving

weave/trace_server/clickhouse/schema_converters.py
```diff
@@ -94,9 +95,18 @@ def ch_call_dict_to_call_schema_dict(ch_call_dict: dict) -> dict:
...
+        "expire_at": _expire_at_from_ch(ch_call_dict.get("expire_at")),
     }
 
 
+def _expire_at_from_ch(value: Any) -> Any:
+    """Map the CH far-future sentinel back to None at the API boundary."""
+    expire_at = ensure_datetimes_have_tz(value)
+    if expire_at == EXPIRE_AT_NEVER:
+        return None
+    return expire_at
```
jtschoonhoven: There's already a loop on line 48 above that does all this for `SENTINEL_DATETIME_FIELDS`. Probably you should just add `expire_at` to SENTINEL_DATETIME_FIELDS?

EDIT: Looks like the `expire_at` sentinel value is 2100-01-01 so this wouldn't work unless it matched `SENTINEL_DATETIME`'s 1970-01-01. Why does expire_at need its own special sentinel value?

If there's a good reason that `expire_at` needs its own sentinel then this is fine, but it would be nice if everything worked the same way.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2603,6 +2606,45 @@ def _default_true(val: bool | None) -> bool:
...
+        name="clickhouse_trace_server_batched.project_ttl_settings_update"
+    )
+    def project_ttl_settings_update(
+        self, req: tsi.ProjectTTLSettingsUpdateReq
+    ) -> tsi.ProjectTTLSettingsUpdateRes:
+        if req.retention_days is not None and req.retention_days < 1:
+            raise InvalidRequest(
+                "retention_days must be None (no TTL) or >= 1 (days of retention)"
+            )
+        assert req.wb_user_id, "wb_user_id is required for audit trail"
```
jtschoonhoven: Don't raise AssertionError in production code, raise `InvalidRequest`. But then why is this optional on `ProjectTTLSettingsUpdateReq`?

weave/trace_server/sqlite_trace_server.py
```diff
@@ -2835,6 +2846,39 @@ def project_stats(self, req: tsi.ProjectStatsReq) -> tsi.ProjectStatsRes:
...
+    def project_ttl_settings_update(
+        self, req: tsi.ProjectTTLSettingsUpdateReq
+    ) -> tsi.ProjectTTLSettingsUpdateRes:
+        if req.retention_days is not None and req.retention_days < 1:
+            raise InvalidRequest(
+                "retention_days must be None (no TTL) or >= 1 (days of retention)"
+            )
+        assert req.wb_user_id, "wb_user_id is required for audit trail"
+
+        stored_days = req.retention_days if req.retention_days is not None else 0
```
jtschoonhoven: Tiny nit: this could just be `stored_days = req.retention_days or 0`

weave/trace_server/trace_server_interface.py
```diff
@@ -155,6 +155,10 @@ class CallSchema(BaseModel):
 
     deleted_at: datetime.datetime | None = None
 
+    # Expiration timestamp for this call. None = no TTL configured for the
+    # project (the row will not be expired).
+    expire_at: datetime.datetime | None = None
```
jtschoonhoven: Nit: if the comment is a field description then it's discoverable by tooling
```suggestion
    expire_at: datetime.datetime | None = Field(
        default=None,
        description="Expiration timestamp for this call. None = no TTL configured for the project (the row will not be expired)."
    )
```

weave/trace_server/trace_server_interface.py
```diff
@@ -1555,6 +1559,28 @@ class ProjectStatsRes(BaseModel):
...
+# TTL Settings API
+# ================
+
+
+class ProjectTTLSettingsReadReq(BaseModelStrict):
+    project_id: str
+
+
+class ProjectTTLSettingsReadRes(BaseModel):
+    retention_days: int | None  # None = no TTL (infinite retention)
```
jtschoonhoven: Nit
```suggestion
    retention_days: int | None = Field(default=None, description="None = no TTL (infinite retention)")
```

weave/trace_server/trace_server_interface.py
```diff
@@ -1555,6 +1559,28 @@ class ProjectStatsRes(BaseModel):
...
+
+
+class ProjectTTLSettingsUpdateReq(BaseModelStrict):
+    project_id: str
+    retention_days: int | None  # None disables TTL; must be None or >= 1
```
jtschoonhoven: Nit
```suggestion
    retention_days: int | None = Field(default=None, description="None disables TTL; must be None or >= 1")
```

jtschoonhoven (commented):
Still a couple questions before approving. Mainly: why does `expire_at` need its own sentinel value that's different from `SENTINEL_DATETIME`? The other comments are non-blocking.

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: its not optional, every request gets a wb_user_id injected by the `external_to_internal_converter`. User doesn't send us anything, we grab it from the API key. Because we overload the type, in practice this is always set, but mypy doesn't know about it

jtschoonhoven: Ok that makes sense. Resolving.

weave/trace_server/sqlite_trace_server.py
```diff
@@ -2835,6 +2846,39 @@ def project_stats(self, req: tsi.ProjectStatsReq) -> tsi.ProjectStatsRes:
...
+            "VALUES (?, ?, ?, ?)",
+            (
+                req.project_id,
+                stored_days,
+                datetime.datetime.now(datetime.timezone.utc).isoformat(),
+                req.wb_user_id,
+            ),
+        )
+        conn.commit()
+        return tsi.ProjectTTLSettingsUpdateRes(retention_days=req.retention_days)
```
jtschoonhoven, line 2893: If sqlite uses a cache then you should invalidate it here too
```suggestion
        conn.commit()
        invalidate_ttl_cache(req.project_id)
        return tsi.ProjectTTLSettingsUpdateRes(retention_days=req.retention_days)
```

weave/trace_server/clickhouse/schema_converters.py
```diff
@@ -94,13 +94,19 @@ def ch_call_dict_to_call_schema_dict(ch_call_dict: dict) -> dict:
...
 
 
-def ch_call_to_row(ch_call: CallCHInsertable) -> list[Any]:
+def ch_call_to_row(ch_call: CallCHInsertable | CallCompleteCHInsertable) -> list[Any]:
     """Convert a CH insertable call to a row for batch insertion with the correct defaults."""
     call_dict = ch_call.model_dump()
-    return [call_dict.get(col) for col in ALL_CALL_INSERT_COLUMNS]
+    return [
+        ch_sentinel_values.to_ch_value(col, call_dict.get(col))
+        if col == "expire_at"
```
jtschoonhoven: This list comprehension feels a little janky. Why is `expire_at` the only sentinel that gets converted here (instead of ALL_SENTINEL_FIELDS)? Deserves a comment at least.

Also, don't you need to do the same conversion in `ch_complete_call_to_row`?

weave/trace_server/ch_sentinel_values.py
```diff
@@ -50,13 +51,14 @@
...
-        "updated_at",
-        "deleted_at",
-    }
-)
+SENTINEL_DATETIME_VALUES: dict[str, datetime.datetime] = {
+    "ended_at": SENTINEL_DATETIME,
+    "updated_at": SENTINEL_DATETIME,
+    "deleted_at": SENTINEL_DATETIME,
+    "expire_at": EXPIRE_AT_NEVER,
+}
```
jtschoonhoven, line 62: I like this approach. Please add a comment though explaining why `expire_at` gets its own sentinel value.

weave/trace_server/ch_sentinel_values.py
```diff
@@ -69,12 +71,15 @@
...
+SENTINEL_IN_CALLS_MERGED_FIELDS = frozenset({"expire_at"})
+
 # Single source of truth for ClickHouse DateTime64 precision per column.
 # Must match the column definitions in migration 024_calls_complete_v2.up.sql.
 DATETIME_PRECISION: dict[str, int] = {
     "ended_at": 6,
     "updated_at": 3,
     "deleted_at": 3,
+    "expire_at": 3,
 }
```
jtschoonhoven, line 91: Nit: the keys here have to exactly match the keys in `SENTINEL_DATETIME_VALUES`. Can you add a simple unit test somewhere that does something like `assert set(SENTINEL_DATETIME_VALUES) == set(DATETIME_PRECISION)`? Or if there's an elegant way to make mypy assert this for you, that would be even better. Not blocking.

weave/trace_server/ch_sentinel_values.py
```diff
@@ -141,7 +146,11 @@ def sentinel_ch_literal(field: str) -> str:
         return "''"
     if field in SENTINEL_DATETIME_FIELDS:
         precision = DATETIME_PRECISION[field]
-        return f"toDateTime64(0, {precision})"
+        sentinel = SENTINEL_DATETIME_VALUES[field]
+        if sentinel == SENTINEL_DATETIME:
+            return f"toDateTime64(0, {precision})"
```
jtschoonhoven: Why do we need a separate branch for this? Doesn't the below logic do the same thing? Or if this is a performance optimization, probably worth a comment.
```suggestion
```

weave/trace_server/ch_sentinel_values.py
```diff
@@ -69,12 +71,15 @@
     SENTINEL_STRING_FIELDS | SENTINEL_DATETIME_FIELDS | SENTINEL_INT_FIELDS
 )
 
+SENTINEL_IN_CALLS_MERGED_FIELDS = frozenset({"expire_at"})
```
jtschoonhoven, line 82: It's worth writing a comment to explain that all the above sentinel fields only apply to `calls_complete` because the legacy `calls_merged` just uses nullable columns for everything except `expire_at`.

I hadn't realized that until I read `_uses_sentinel_for_read_table()`

weave/trace_server/sqlite_trace_server.py
jtschoonhoven: Bump on this. You can resolve if the cache isn't relevant to sqlite.

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -1008,6 +1008,26 @@ def project_stats(self, req: tsi.ProjectStatsReq) -> tsi.ProjectStatsRes:
             "/project/stats", req, tsi.ProjectStatsReq, tsi.ProjectStatsRes
         )
 
+    def project_ttl_settings_read(
+        self, req: tsi.ProjectTTLSettingsReadReq
+    ) -> tsi.ProjectTTLSettingsReadRes:
+        return self._generic_request(
+            "/project/ttl_settings",
```
jtschoonhoven: Nit: the URL convention in this module seems to be `/project/{resource}/{verb}`. Idk why we're not doing REST, but for consistency this should probably be `/project/ttl_settings/read`.

weave/trace_server/ch_sentinel_values.py
```diff
@@ -34,6 +34,7 @@
     from weave.trace_server.orm import ParamBuilder
 
 SENTINEL_DATETIME = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)
```
jtschoonhoven: It seems dangerous that this module exports both `SENTINEL_DATETIME` and `SENTINEL_DATETIME_FIELDS`. `SENTINEL_DATETIME` is no longer the sentinel for all datetimes, so it seems safer to name it `SENTINEL_EPOCH`.

That would also flush out any call sites that are still using `SENTINEL_DATETIME` instead of `SENTINEL_DATETIME_FIELDS`.

weave/trace_server/trace_server_interface.py
```diff
@@ -1555,6 +1563,32 @@ class ProjectStatsRes(BaseModel):
...
+
+
+class ProjectTTLSettingsReadReq(BaseModelStrict):
+    project_id: str
+
+
+class ProjectTTLSettingsReadRes(BaseModel):
+    retention_days: int | None = Field(
+        default=None, description="None = no TTL (infinite retention)"
+    )
```
jtschoonhoven, line 1578: Should this include the `updated_by` field? Or are we intentionally not exposing that anywhere?

weave/trace_server/clickhouse_schema.py
```diff
@@ -3,10 +3,9 @@
 from pydantic import BaseModel, Field, field_validator
 
 from weave.shared import refs_internal as ri
-from weave.trace_server import validation
+from weave.trace_server import ch_sentinel_values, validation
 
-# Sentinel value meaning "never expire". Matches the CH column default.
-EXPIRE_AT_NEVER = datetime.datetime(2100, 1, 1, tzinfo=datetime.timezone.utc)
+EXPIRE_AT_NEVER = ch_sentinel_values.EXPIRE_AT_NEVER
```
jtschoonhoven: Nit: pointless redefinition
```suggestion
```

weave/trace_server/clickhouse_schema.py
```diff
@@ -69,7 +68,7 @@ class CallStartCHInsertable(
     attributes_dump: str
     inputs_dump: str
     otel_dump: str | None = None
-    expire_at: datetime.datetime = EXPIRE_AT_NEVER
+    expire_at: datetime.datetime | None = None
```
jtschoonhoven: Nit: instead of adding this to every child class, want to define `expire_at` on `CallBaseCHInsertable`?

weave/trace_server/clickhouse/schema_converters.py
```diff
@@ -94,13 +94,19 @@ def ch_call_dict_to_call_schema_dict(ch_call_dict: dict) -> dict:
         "display_name": display_name,
         "storage_size_bytes": ch_call_dict.get("storage_size_bytes"),
         "total_storage_size_bytes": ch_call_dict.get("total_storage_size_bytes"),
+        "expire_at": sv["expire_at"],
     }
 
 
-def ch_call_to_row(ch_call: CallCHInsertable) -> list[Any]:
+def ch_call_to_row(ch_call: CallCHInsertable | CallCompleteCHInsertable) -> list[Any]:
```
jtschoonhoven, line 101: There's already `ch_complete_call_to_row(ch_call: CallCompleteCHInsertable)` farther down, so this might be a mistake.
```suggestion
def ch_call_to_row(ch_call: CallCHInsertable) -> list[Any]:
```

weave/trace_server/trace_server_interface.py
```diff
@@ -1555,6 +1563,32 @@ class ProjectStatsRes(BaseModel):
...
+
+
+class ProjectTTLSettingsUpdateReq(BaseModelStrict):
+    project_id: str
+    retention_days: int | None = Field(
+        default=None, description="None disables TTL; must be None or >= 1"
+    )
+    wb_user_id: str | None = None
```
jtschoonhoven, line 1586: I'm still a little unclear on how this works, but we definitely shouldn't let the client set this. We should _only_ derive this from the API key. I think the `ProjectTTLSettingsUpdateReq` is the wrong place to store this.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1804,14 +1804,15 @@ def get_calls_table_name(read_table: ReadTable) -> str:
         join_table_name=ROLLED_UP_CALL_MERGED_STATS_TABLE_NAME,
     ),
     "otel_dump": CallsMergedAggField(field="otel_dump", agg_fn="any"),
+    "expire_at": CallsMergedAggField(field="expire_at", agg_fn="max"),
```
jtschoonhoven: I asked for this to be `max` instead of `min` earlier, but robotim flagged that in calls_merged this column is `SimpleAggregateFunction(min, DateTime64(3))`. Probably the logic here should agree.
```suggestion
    "expire_at": CallsMergedAggField(field="expire_at", agg_fn="min"),
```

jtschoonhoven (approved):
Approving so long as these get fixed:

Blocking:
- Rename `SENTINEL_DATETIME` to `SENTINEL_EPOCH`. 
- The list comprehension in `schema_converters.py` looks busted, see comments.
- See comment in `trace_server_interface.py`: I think we shouldn't store `wb_user_id` on `ProjectTTLSettingsUpdateReq`.

Everything else is nonblocking. LMK if you want me to take another look before merging.

weave/trace_server/trace_server_interface.py
gtarpenning: this is how literally all of our endpoints work, happy to chat about this in another context, but if you have issues with this there are 60 other places in this file that will also need to be updated.

PR #6395, opened 2026-03-19
fix(weave): Raise errors in completion create stream

jwlee64:
## Description

https://wandb.atlassian.net/browse/WB-32221

Before we would swallow errors
Return 200 
and not save the error on the call

properly return 500, save the error message on the call, and return
Screenshot 2026-03-19 at 12 44 34 PM

we also properly return weave call ids so you can see the errored trace in the playground 
Screenshot 2026-03-19 at 12 44 55 PM

## Testing

manually, added some tests on core side

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -5913,36 +5913,23 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         prompt = getattr(req.inputs, "prompt", None)
         template_vars = getattr(req.inputs, "template_vars", None)
 
-        try:
-            # Use helper to resolve prompt, combine messages, and apply template vars
-            combined_messages, initial_messages = resolve_and_apply_prompt(
-                prompt=prompt,
-                messages=getattr(req.inputs, "messages", None),
-                template_vars=template_vars,
-                project_id=req.project_id,
-                obj_read_func=self.obj_read,
-            )
-        except Exception as e:
-            logger.exception("Failed to resolve and apply prompt")
-
-            # Yield error as single chunk then stop.
-            def _single_error_iter(err: Exception) -> Iterator[dict[str, str]]:
-                yield {"error": f"Failed to resolve and apply prompt: {err!s}"}
-
-            return _single_error_iter(e)
+        # Use helper to resolve prompt, combine messages, and apply template vars.
+        # Raises on failure — callers (trace_server.py) let this propagate to
+        # FastAPI's exception handlers for a proper error response.
+        combined_messages, initial_messages = resolve_and_apply_prompt(
+            prompt=prompt,
+            messages=getattr(req.inputs, "messages", None),
+            template_vars=template_vars,
+            project_id=req.project_id,
+            obj_read_func=self.obj_read,
+        )
 
         # --- Shared setup logic (copy of completions_create up to litellm call)
         model_info = self._model_to_provider_info_map.get(req.inputs.model)
-        try:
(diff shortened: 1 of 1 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5775,6 +5753,98 @@ def actions_execute_batch(
 
         return tsi.ActionsExecuteBatchRes()
 
+    def _record_tracked_completions_create_failure(
```
jtschoonhoven, line 5756: Ultra-tiny-nit: I prefer to do the inverted pyramid thing and sort private helper functions _after_ the public function that users them. So I'd usually put this below `completions_create`. TIOLI.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5775,6 +5753,98 @@ def actions_execute_batch(
...
+        self,
+        req: tsi.CompletionsCreateReq,
+        *,
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+        error_message: str,
+        initial_messages: list,
```
jtschoonhoven, line 5763: I think it's worth typing this more explicitly. Even if it's just `list[Any]` it's good to know what we're expecting here. Probably it's something like `list[str | ChatMessage]`?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5799,8 +5872,19 @@ def completions_create(
 
             except Exception as e:
                 logger.exception("Failed to resolve prompt")
+                err = f"Failed to resolve and apply prompt: {e!s}"
```
jtschoonhoven, line 5875: I'd recommend not making our own server's exceptions visible to customers. It's _probably_ fine, but not really a security best-practice. Same comment applies in the exception handler below.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5775,6 +5753,98 @@ def actions_execute_batch(
...
+        req: tsi.CompletionsCreateReq,
+        *,
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+        error_message: str,
+        initial_messages: list,
+    ) -> str:
+        """Persist a failed completions_create to ClickHouse when ``track_llm_call`` is enabled.
+
+        Mirrors the success-path insert (calls_complete vs calls_merged) so failures
```
jtschoonhoven, line 5767: There's also some code duplication between this function and `calls_complete` — consider factoring out the shared logic to a helper function.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5799,8 +5872,19 @@ def completions_create(
...
+                        start_time=start_time_tracked,
+                        end_time=datetime.datetime.now(),
+                        error_message=err,
+                        initial_messages=initial_messages,
+                    )
                 return tsi.CompletionsCreateRes(
-                    response={"error": f"Failed to resolve prompt: {e!s}"}
+                    response={"error": err},
+                    weave_call_id=weave_call_id,
                 )
```
jtschoonhoven, line 5888: I don't see where the status code gets set to 500 here, unless it happens automagically somewhere else?

weave/trace_server/clickhouse_trace_server_batched.py
jtschoonhoven, line 5926: This probably deserves some error handling as well

jtschoonhoven (commented):
A few questions and comments. Probably this deserves a unit test as well.

PR #6399, opened 2026-03-20
chore(weave): Add job manager for tracking long-running operations via Redis

neutralino1:
## Summary
This PR adds a Redis-backed `JobManager` to keep track of long-running tasks. Those include call scores-backfilling jobs, evaluation playground jobs.

`JobManager` enables keeping track of progress (`total_items`, `successful_items`, `failed_items`) as well as cancellation requests.

Job details expire after 60 minutes to avoid keeping orphaned tasks (e.g. worker dies, never completes task). This possibility yields a sub-optimal UX, but I argue we should address it if it becomes a real issue.

## Tests

Unit tests


jtschoonhoven:
What's the use case for this?

neutralino1:
@jtschoonhoven there are two uses cases currently (more to come):
- The evaluation playground launched background jobs. They need be cancelable, and also show progress.
- the score backfill feature also launches background jobs, and there is currently no UI feedback that something is happening. Providing feedback and progress requires some task tracking.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6196,6 +6197,18 @@ def calls_score(self, req: tsi.CallsScoreReq) -> tsi.CallsScoreRes:
 
         return tsi.CallsScoreRes()
 
+    def tasks_list(self, req: tsi.TasksListReq) -> tsi.TasksListRes:
```
jtschoonhoven: Nit: "task" is overloaded, probably worth adding docstrings to all the functions/classes introduced in this PR so folks know what this is.

weave/trace_server/task_manager.py
```diff
@@ -0,0 +1,170 @@
...
+    """Manages task state in Redis for tracking long-running operations."""
+
+    def __init__(self, project_id: str, wb_user_id: str | None) -> None:
+        self._redis_client = get_redis_client()
+
+        if self._redis_client is None:
+            raise ValueError("Redis client is not initialized")
+
+        self._project_id = project_id
+        self._wb_user_id = wb_user_id or "anonymous"
```
jtschoonhoven: Nit: this could be a bit more concise as a dataclass, plus you'd get type safety and sane defaults for python boilerplate "dunder" methods:
```suggestion
@dataclass(frozen=True)
class TaskManager:
    """Manages task state in Redis for tracking long-running operations."""
    project_id: str
    wb_user_id: str = "anonymous"
    _redis_client = field(init=False, default_factory=get_redis_client)
```

EDIT: Actually I don't think this needs to be a class at all, see other comment.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6196,6 +6197,18 @@ def calls_score(self, req: tsi.CallsScoreReq) -> tsi.CallsScoreRes:
...
+        task_manager = TaskManager(req.project_id, req.wb_user_id)
+        tasks = task_manager.list_tasks()
+        return tsi.TasksListRes(tasks=[tsi.TaskDetails(**t) for t in tasks])
```
jtschoonhoven: I believe `list_tasks` already returns a list of `TaskDetails` so you shouldn't need to make new `TaskDetails`.
```suggestion
        return tsi.TasksListRes(tasks=tasks)
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6196,6 +6197,18 @@ def calls_score(self, req: tsi.CallsScoreReq) -> tsi.CallsScoreRes:
...
+
+    def task_cancel(self, req: tsi.TaskCancelReq) -> tsi.TaskCancelRes:
+        task_manager = TaskManager(req.project_id, req.wb_user_id)
+        task = task_manager.cancel_task(req.task_id)
+        if task is None:
+            raise ValueError(f"Task {req.task_id} not found")
+        return tsi.TaskCancelRes(task=tsi.TaskDetails(**task))
```
jtschoonhoven:
```suggestion
        return tsi.TaskCancelRes(task=task)
```

weave/trace_server/task_manager.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class TaskID(str):
+    """A unique identifier for a task."""
+
+    def __new__(cls) -> "TaskID":
+        return super().__new__(cls, str(uuid.uuid4()))
+
+
+class TaskManager:
```
jtschoonhoven: Nit: I don't see any benefit to this being a class. Seems like it would be more straightforward to write these as functions instead of as class methods.

weave/trace_server/task_manager.py
```diff
@@ -0,0 +1,170 @@
...
+    """Manages task state in Redis for tracking long-running operations."""
+
+    def __init__(self, project_id: str, wb_user_id: str | None) -> None:
+        self._redis_client = get_redis_client()
```
jtschoonhoven: You should probably cache and reuse a single redis client.
```python3
# Something along these lines, this is from memory
@cached(LRUCache(maxsize=1)
def _get_redis_client() -> RedisClient:
   return get_redis_client()
```

weave/trace_server/task_manager.py
```diff
@@ -0,0 +1,170 @@
...
+        """Get task details by ID."""
+        if self._redis_client is None:
+            return None
+
+        task_details = self._redis_client.get(self._make_task_key(task_id))
+
+        if task_details is None:
+            return None
+
+        return json.loads(task_details)
```
jtschoonhoven: There's no runtime guarantee that this actually returns `TaskDetails`. Since you're frequently (de)serializing to/from TaskDetails I think it would be a good idea to make it a pydantic class that has some runtime validation.

weave/trace_server/task_manager.py
```diff
@@ -0,0 +1,170 @@
...
+    def increment_successful_items(self, task_id: str) -> TaskDetails | None:
+        """Increment the successful items count for a task."""
+        task_details = self.get_task(task_id)
+
+        if task_details is None:
+            return None
+
+        task_details["successful_items"] += 1
+
+        self._set_task(task_id, task_details)
```
jtschoonhoven: Why write the counts to TaskDetails (2 network requests, json dump/load, race conditions) instead of writing them to redis directly (1 network request, atomic)?

weave/trace_server/task_manager.py
```diff
@@ -0,0 +1,170 @@
...
+            return []
+
+        tasks: list[TaskDetails] = []
+
+        cursor = 0
+
+        while True:
+            cursor, keys = self._redis_client.scan(
+                cursor=cursor, match=f"{self._task_key_prefix}*"
+            )
```
jtschoonhoven: `scan` iterates over the entire keyspace. That's probably OK if the redis instance stores _only_ these short-lived task details, but if we're reusing this redis as a general-purpose cache then there might be millions of keys and you'll get bad performance. You could instead maintain a set of all the tasks.

weave/trace_server/task_manager.py
```diff
@@ -0,0 +1,170 @@
...
+
+            for key in keys:
+                data = self._redis_client.get(key)
```
jtschoonhoven: This is pretty inefficient. You should fetch all these tasks in a single shot. Consider `MGET` or a pipeline. Or store all the keys in a set.

jtschoonhoven (commented):
I think it's worth using redis a bit more efficiently. I made some suggestions.

weave/trace_server/task_manager.py
neutralino1: Need to keep project and user IDs to form keys. It's that or they must be passed in each function call.

weave/trace_server/task_manager.py
neutralino1: It's already a singleton in `redis_client.py`.

weave/trace_server/task_manager.py
neutralino1: In the current set up, with a payload stored on a single class, I have to read the current task values, increment and persist again. What do you mean writing the to redis directly?

What I considered doing is, instead of persisting a serialized object, I could persist each individual value as a separate Redis key (e.g. `"weave:task:{self._project_id}:{self._wb_user_id}:{task_id}:total_items"`), then I could use the native redis INCR command. That's a valid option.

weave/trace_server/task_manager.py
neutralino1: Where would I maintain this set of all tasks? As a separate Redis key? This is scoped per project and user though, I would have to maintain a list for each user. Plus I could not handle the TTL in this case.

weave/trace_server/clickhouse_trace_server_batched.py
neutralino1: Renamed to job and added docstring.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6368,6 +6369,28 @@ def calls_score(self, req: tsi.CallsScoreReq) -> tsi.CallsScoreRes:
...
 
+    def jobs_list(self, req: tsi.JobsListReq) -> tsi.JobsListRes:
+        """List all jobs for a project/user.
+
+        A job tracks the progress of a long-running server-side operation
+        (e.g. batch scoring). State is stored in Redis with a 1-hour TTL.
+        """
+        job_manager = JobManager(req.project_id, req.wb_user_id)
+        jobs = job_manager.list_jobs()
+        return tsi.JobsListRes(jobs=jobs)
```
jtschoonhoven, line 6380: Why do these methods live in `clickhouse_trace_server_batched` if they only touch redis?

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+    return JobDetails.model_validate(
+        {**data, "canceled_at": data.get("canceled_at") or None}
+    )
+
+
+class JobID(str):
+    """A unique identifier for a job."""
+
+    def __new__(cls) -> Self:
+        return super().__new__(cls, str(uuid.uuid4()))
```
jtschoonhoven, line 45: This class is only referenced in one place, so it's not currently adding any value over doing `str(uuid4())` inline.

If you want some type safety you could pass job IDs around as UUIDs instead of as strings. But I don't think you need this class either way.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+@dataclass(frozen=True)
+class JobManager:
+    """Manages job state in Redis for tracking long-running operations."""
+
+    project_id: str
+    wb_user_id: str | None = "anonymous"
```
jtschoonhoven, line 53: If this is meant to default to "anonymous" then I suspect you don't actually want to allow `None` here.

jtschoonhoven: I see below the `_user_id` property returns `self.wb_user_id or "anonymous"`. No need to set the default in two different places, I'd drop the default here and keep the `or "anonymous"` below.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+    _redis_client: "Redis" = field(
+        init=False, repr=False, default_factory=_get_required_redis_client
+    )
+
+    @property
+    def _user_id(self) -> str:
+        return self.wb_user_id or "anonymous"
```
jtschoonhoven, line 60: Tiny nit: "anonymous" could be a constant. TIOLI.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+
+    @property
+    def _job_key_prefix(self) -> str:
+        return f"weave:job:{self.project_id}:{self._user_id}:"
+
+    @property
+    def _index_key(self) -> str:
+        return f"weave:job:{self.project_id}:{self._user_id}:_index"
```
jtschoonhoven, line 68: Nit: "_job_key_prefix" is pretty self-explanatory, but it's not immediately obvious to me what "_index_key" is used for, so I'd recommend a docstring here.

Also, do you want to DRY this up and return `f"{self._job_key_prefix}:_index"`?

EDIT: Ok now I understand how this is used. The word "index" is overloaded, and I think it would be more self-explanatory to call this `_job_ids_key`.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _make_job_key(self, job_id: str) -> str:
+        return f"{self._job_key_prefix}{job_id}"
+
+    def create_job(self, total_items: int) -> JobDetails:
+        """Create a new job with the given total number of items."""
```
jtschoonhoven, line 74: What are "items" in this context? Why do you have to specify `total_items` up front? Probably this is more clear in context, but I think it would be good to explain in the docstring.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+        job_id = JobID()
+        key = self._make_job_key(job_id)
+
+        created_at = datetime.now().isoformat()
+
+        self._redis_client.hset(
+            key,
+            mapping={
+                "id": str(job_id),
```
jtschoonhoven, line 83: Nit: "job_id" is already `str`, right?

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+from weave.trace_server.trace_server_interface import JobDetails
+
+if TYPE_CHECKING:
+    from redis import Redis
+
+JOB_TTL_SECONDS = 3600  # 1 hour - prevents orphaned jobs from persisting indefinitely
+
+
+def _get_required_redis_client() -> "Redis":
+    client = get_redis_client()
```
jtschoonhoven, line 24: Nit: consider updating `get_redis_client` to accept an argument for `decode_responses`. Redis returns bytes by default, and it's nice to know what to expect without having to inspect `redis_client.py`.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+        return f"{self._job_key_prefix}{job_id}"
+
+    def create_job(self, total_items: int) -> JobDetails:
+        """Create a new job with the given total number of items."""
+        job_id = JobID()
+        key = self._make_job_key(job_id)
+
+        created_at = datetime.now().isoformat()
+
+        self._redis_client.hset(
```
jtschoonhoven, line 80: You can combine several redis commands into a single network request with a pipeline:
```
with self._redis_client.pipeline() as pipe:
    pipe.hset(...)
    pipe.expire(...)
    pipe.sadd(...)
    pipe.execute()
```

weave/trace_server/trace_server_interface.py
```diff
@@ -1936,6 +1936,36 @@ class CallsScoreRes(BaseModel):
...
 
+class JobDetails(BaseModelStrict):
+    """Details about a job's progress."""
+
+    id: str
+    total_items: int
+    successful_items: int
+    failed_items: int
+    created_at: str
+    canceled_at: str | None = None
```
jtschoonhoven, line 1947: I recommend making these `datetimes` instead of strings. Pydantic can automatically parse to `datetime` from an ISO-formatted string.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+    def increment_successful_items(self, job_id: str) -> JobDetails | None:
+        """Atomically increment the successful items count for a job."""
+        key = self._make_job_key(job_id)
+
+        if not self._redis_client.exists(key):
+            return None
+
+        self._redis_client.hincrby(key, "successful_items", 1)
+
+        return self.get_job(job_id)
```
jtschoonhoven, line 120: Nit: here and elsewhere you're making two synchronous network requests (increment-then-read). It should be possible to use a pipeline to combine these into one request.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+class JobID(str):
+    """A unique identifier for a job."""
+
+    def __new__(cls) -> Self:
+        return super().__new__(cls, str(uuid.uuid4()))
+
+
+@dataclass(frozen=True)
+class JobManager:
+    """Manages job state in Redis for tracking long-running operations."""
```
jtschoonhoven, line 50: I notice all these methods are synchronous. Is there a reason we can't use async redis? I know the official redis library supports asyncio out of the box.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+
+        return self.get_job(job_id)
+
+    def complete_job(self, job_id: str) -> None:
+        """Mark a job as complete by removing it from Redis."""
+        self._redis_client.delete(self._make_job_key(job_id))
+        self._redis_client.srem(self._index_key, job_id)
+
+    def is_canceled(self, job_id: str) -> bool:
+        """Return True if the job has been canceled."""
```
jtschoonhoven, line 139: What's the use case for `is_canceled`? Flagging that it's only possible to detect canceled jobs until they auto-expire after an hour. That might be what we want, but it also might be confusing depending on the case.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+            return None
+
+        self._redis_client.hincrby(key, "failed_items", 1)
+
+        return self.get_job(job_id)
+
+    def complete_job(self, job_id: str) -> None:
+        """Mark a job as complete by removing it from Redis."""
+        self._redis_client.delete(self._make_job_key(job_id))
+        self._redis_client.srem(self._index_key, job_id)
```
jtschoonhoven, line 136: It's a little surprising to me that we have `is_canceled` but not `is_complete`. Maybe we should also track `completed_at` so it's possible to retrieve recently-completed jobs? Not sure exactly how we plan to use this, but I can imagine it being useful.

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+        return self.get_job(job_id)
+
+    def list_jobs(self) -> list[JobDetails]:
+        """List all jobs for this project and user."""
+        job_ids = self._redis_client.smembers(self._index_key)
+        jobs: list[JobDetails] = []
+        stale_ids: list[str] = []
+
+        for job_id in job_ids:
+            data = self._redis_client.hgetall(self._make_job_key(job_id))
```
jtschoonhoven, line 165: You can batch these requests:
```
with self._redis_client.pipeline(transaction=False) as pipe:
    for key in keys:
        pipe.hgetall(key)
    jobs_data = pipe.execute()
```

weave/trace_server/job_manager.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def clear_jobs(self) -> None:
+        """Delete all job keys for this project and user from Redis."""
+        job_ids = self._redis_client.smembers(self._index_key)
+
+        if job_ids:
+            keys = [self._make_job_key(job_id) for job_id in job_ids]
+            self._redis_client.delete(*keys)
+
+        self._redis_client.delete(self._index_key)
```
jtschoonhoven, line 185: Nit: you could batch these requests as well

weave/trace_server/sqlite_trace_server.py
```diff
@@ -2880,6 +2880,12 @@ def evaluation_status(
     def calls_score(self, req: tsi.CallsScoreReq) -> tsi.CallsScoreRes:
         raise NotImplementedError("Calls scoring is not supported in SQLite")
 
+    def jobs_list(self, req: tsi.JobsListReq) -> tsi.JobsListRes:
+        raise NotImplementedError("Job listing is not supported in SQLite")
+
+    def job_cancel(self, req: tsi.JobCancelReq) -> tsi.JobCancelRes:
+        raise NotImplementedError("Job cancellation is not supported in SQLite")
+
```
jtschoonhoven, line 2888: Why isn't this supported for sqlite? Seems like the only dependency is on redis.

jtschoonhoven (commented):
A few more nits and suggestions. The important suggestions are mostly related to batching network requests. Also wondering if we should be using asyncio?

PR #6408, opened 2026-03-24
fix(weave): stabilize version_index on re-publish (WB-32435)

rgao-coreweave:
## Summary

Fixes version_index instability when object content is re-published. Previously, re-publishing an existing digest caused version indices to shift because `version_index` was computed via `ROW_NUMBER() ORDER BY created_at`, and re-publishing created a new row with a newer timestamp.

**Jira**: WB-32435

**Two changes:**

1. **Query: `min(created_at)` anchor** — version_index now orders by the *first* time each digest was published, not the surviving row's timestamp. Adds `min(created_at) OVER (PARTITION BY project_id, kind, object_id, digest)` and uses `_first_created_at ASC, digest ASC` for deterministic ordering.

2. **Dedup-before-insert** — `obj_create` now checks if a non-deleted row with the same digest already exists before inserting. If so, it returns early. This prevents new rows from being created for existing content, keeping `min(created_at)` stable across ReplacingMergeTree merges.

## Correctness

| Test | Before | After |
|------|--------|-------|
| P1-T1: Republish oldest | FAIL (shifted) | **PASS** |
| P1-T2: Republish middle | FAIL (shifted) | **PASS** |
| P1-T3: Republish latest | PASS | PASS |
| P1-T4: Delete+republish | FAIL (shifted) | FAIL (known gap) |
| P1-T5: 5 versions, republish v0 | FAIL (shifted) | **PASS** |
| P1-T6: Republish two versions | FAIL (shifted) | **PASS** |

P1-T4 (delete+republish) remains a known gap — a structural limitation of ReplacingMergeTree where the original row is lost after compaction.

### Object creation (100 objects × 100 versions = 10,000)

| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| Throughput | 144/s | 94/s | -35% |
| Per version | 6.95ms | 10.66ms | +3.7ms |

The dedup SELECT adds ~3.7ms per `obj_create` (CH network round-trip). Acceptable — object creation is interactive, not high-throughput. 94/s is still 5-10x above typical user publishing rates.

### Query performance (median of 5 runs)

| Versions | Before | After | Delta |
|----------|--------|-------|-------|
| 1 | 8.4ms | 8.5ms | +1% |
| 100 | 20.3ms | 22.7ms | +12% |
| 500 | 58.8ms | 59.0ms | +0% |
| 1000 | 105.7ms | 107.2ms | +1% |

The `min(created_at)` window adds negligible overhead. Scaling slope unchanged at ~0.1ms/version.

## Files Changed

| File | Change |
|------|--------|
| `weave/trace_server/query_builder/objects_query_builder.py` | `min(created_at)` window + `_first_created_at` ordering + `digest ASC` tiebreak |
| `weave/trace_server/clickhouse_trace_server_batched.py` | `_obj_digest_exists_non_deleted()` + dedup check in `obj_create()` |

## Test plan

- [x] Version index stability tests pass on SQLite (6/6)
- [x] Version index stability tests pass on ClickHouse (5/6, P1-T4 is known gap)
- [x] Benchmark shows acceptable performance regression
- [ ] Run full `trace` shard
- [ ] Run `trace_server` shard


rgao-coreweave:
* **#6408** <a href="https://app.graphite.com/github/pr/wandb/weave/6408?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6409](https://github.com/wandb/weave/pull/6409) <a href="https://app.graphite.com/github/pr/wandb/weave/6409?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6411](https://github.com/wandb/weave/pull/6411) <a href="https://app.graphite.com/github/pr/wandb/weave/6411?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6408?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1626,6 +1625,42 @@ def _update_calls_complete(
...
+        self, project_id: str, object_id: str, digest: str
+    ) -> bool:
+        """Check if a non-deleted row with the same digest already exists.
+
+        Uses argMax(deleted_at, created_at) to handle pre-merge state where
+        multiple rows may exist for the same digest.  Best-effort: async
+        inserts mean a small race window, but the read-path dedup and
+        ReplacingMergeTree handle duplicates correctly.
+        """
+        query = """
```
gtarpenning: can we move raw SQL to the `objects_query_builder.py`

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1157,3 +1157,182 @@ def test_file_batch_clears_on_insert_failure():
             f"Memory leak: _file_batch retained {len(server._file_batch)} items "
             f"after failed insert. Batch should be cleared on any exception."
         )
+
+
+# ── obj_create dedup-before-insert ───────────────────────────────────
+
+
+def _get_ch_server(trace_server):
```
gtarpenning: why do we need this? there should already be a fixture for this if you need that

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1157,3 +1157,182 @@ def test_file_batch_clears_on_insert_failure():
...
+
+    r1 = _obj_create(server, project_id, obj_id, val)
+    r2 = _obj_create(server, project_id, obj_id, val)
+
+    assert r1.digest == r2.digest
+    assert r2.object_id == obj_id
+    assert len(_objs_query(server, project_id, obj_id)) == 1
+
+
+def test_obj_create_dedup_allows_different_content(trace_server):
```
gtarpenning: can we combine this test with the one above? generally fewer tests with more assertions are faster in CI

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1157,3 +1157,182 @@ def test_file_batch_clears_on_insert_failure():
...
+    non_deleted = [
+        o for o in _objs_query(server, project_id, obj_id) if o.deleted_at is None
+    ]
+    assert len(non_deleted) >= 1
+
+
+# ── _obj_digest_exists_non_deleted ───────────────────────────────────
+
+
+def test_obj_digest_exists_non_deleted_returns_false_for_missing(trace_server):
```
gtarpenning: we can combine this with another test

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1157,3 +1157,182 @@ def test_file_batch_clears_on_insert_failure():
...
+
+    for i in range(n):
+        _obj_create(server, project_id, obj_id, {"i": i})
+
+    objs = _objs_query(server, project_id, obj_id)
+    assert len(objs) == n
+    assert sorted(o.version_index for o in objs) == list(range(n))
+
+
+def test_delete_preserves_version_index_gaps(trace_server):
```
gtarpenning: i think there is already coverage for most of these tests, can you ask claude to find similar tests?

tests/trace/test_client_trace.py
```diff
@@ -1041,14 +1041,15 @@ def mock_now(*args, **kwargs):
...
-
+    # Create a datetime subclass that overrides now() but preserves the real
+    # class for isinstance() checks — replacing datetime.datetime with a
+    # MagicMock breaks isinstance() in any code sharing the datetime module.
+    class _MockDatetime(datetime.datetime):
+        @classmethod
+        def now(cls, *args, **kwargs):
+            return mock_now()
+
+    with mock.patch("weave.trace.weave_client.datetime.datetime", _MockDatetime):
```
gtarpenning: why do we need to do this?

gtarpenning (commented):
couple nits, @ me for quick re-review

weave/trace_server/clickhouse_trace_server_batched.py
rgao-coreweave: Moved to `make_obj_digest_exists_non_deleted_query()` in `obj_tags_query_builder.py`, following the same pattern as the existing `make_obj_version_exists_query()`.

tests/trace_server/test_clickhouse_trace_server_batched.py
rgao-coreweave: Replaced with a `@pytest.fixture` named `ch_server` in the same file. The existing `clickhouse_trace_server` fixture in `test_otel_calls_complete.py` also sets AUTO routing mode which isn't needed here, so kept it separate.

tests/trace_server/test_clickhouse_trace_server_batched.py
rgao-coreweave: Combined all three dedup tests into a single `test_obj_create_dedup` — covers same-digest skip, different-content insert, and delete+re-publish in one test.

tests/trace_server/test_clickhouse_trace_server_batched.py
rgao-coreweave: Combined into a single `test_obj_digest_exists_non_deleted` covering all three cases (missing, after insert, after delete).

tests/trace_server/test_clickhouse_trace_server_batched.py
rgao-coreweave: `test_obj_dedupe` in `test_weave_client.py` already covers sequential version_index at the client level. Removed `test_version_index_sequential` (redundant). Kept `test_version_index_stable_on_republish` (tests new `_first_created_at` behavior, no existing coverage) and `test_delete_preserves_version_index_gaps` (no existing coverage for delete+gap preservation at CH level).

tests/trace/test_client_trace.py
rgao-coreweave: Reverted — this was fixing a pre-existing fragile mock that broke `isinstance()` when the dedup code path ran in the same process. Not related to the core changes in this PR.

PR #6409, opened 2026-03-24
chore(weave): add version_index stability tests and benchmarks (WB-32435)

rgao-coreweave:
## Summary

Adds test suite and benchmark script for version_index stability and object creation performance. Stacked on #6408.

**New files:**

- `tests/trace/test_version_index_stability.py` — 6 correctness tests covering re-publish and delete+republish scenarios (P1-T1 through P1-T6)
- `tests/trace/bench_obj_create.py` — performance benchmark for object creation and query paths

## Running the benchmarks

```bash
cd services/weave-python/weave-public

# Full benchmark (ClickHouse):
nox --no-install -e "tests-3.12(shard='trace')" -- \
  tests/trace/bench_obj_create.py -v -s --trace-server=clickhouse --clickhouse-process=true -n0

# Query scaling only:
nox --no-install -e "tests-3.12(shard='trace')" -- \
  tests/trace/bench_obj_create.py::test_bench_query_versions -v -s --trace-server=clickhouse --clickhouse-process=true -n0

# Stability tests:
nox --no-install -e "tests-3.12(shard='trace')" -- \
  tests/trace/test_version_index_stability.py -v -s --trace-server=clickhouse --clickhouse-process=true -n0
```

Note: `-n0` disables xdist parallelism so benchmark output is visible.

## Benchmark coverage

| Test | What it measures |
|------|-----------------|
| `test_bench_obj_create` | 100 objects × 100 versions: create throughput, query, republish |
| `test_bench_query_versions` | Query scaling from 1 to 1000 versions per object |
| `test_bench_obj_create_many_objects_few_versions` | 500 objects × 5 versions: wide partition pattern |

## Test plan

- [x] All benchmarks pass on SQLite
- [x] All benchmarks pass on ClickHouse
- [x] Stability tests pass on SQLite (6/6)
- [x] Stability tests pass on ClickHouse (5/6, P1-T4 is documented known gap)

PR #6411, opened 2026-03-24
feat(weave): write explicit "latest" alias on every obj_create (WB-32435)

rgao-coreweave:
## Summary

Implements proposal for Problem 2: on every `obj_create` (including dedup hits), write a "latest" alias to the aliases table. This moves "latest" to the re-published digest, matching Artifacts' "promote on re-publish" behavior. Stacked on #6409.

**Jira**: WB-32435

### Write path

- `obj_create` calls `_insert_aliases(["latest"], digest)` after every insert and dedup-skip
- Both new versions and re-published content move the "latest" pointer

### Read path (3 touch points)

- `_maybe_resolve_alias`: "latest" now resolves from aliases table, with **computed** **`is_latest`** **fallback** for legacy objects or failed alias writes
- `add_aliases_condition`: checks aliases table OR `is_latest=1` (backward compat)
- `_enrich_objs_with_tags_and_aliases`: prefers explicit alias, only synthesizes from `is_latest` for legacy objects without alias row

### Consistency guarantee

`_maybe_resolve_alias("latest")` has a fallback to `_resolve_computed_latest()` which uses the `is_latest` window function. This ensures all three read paths agree:

| Path | Primary | Fallback |
| --- | --- | --- |
| `obj_read(digest="latest")` | Alias table | Computed `is_latest` |
| `objs_query(latest_only=True)` | Computed `is_latest` | — |
| `objs_query(aliases=["latest"])` | `is_latest=1 OR` alias table | — |

If the alias INSERT fails, all paths converge on the computed `is_latest` (graceful degradation to old behavior, not data corruption).

### Object creation (100 objects × 100 versions = 10,000)

| Metric | Baseline | \+ P1 | \+ P1 + P2 (this PR) | Delta (P1→P1+P2) |
| Throughput | 144/s | 94/s | 62/s | \-34% |
| Per version | 6\.95ms | 10\.66ms | 16\.01ms | \+5.4ms |

The alias INSERT adds ~5.4ms per `obj_create`. Acceptable — object creation is interactive, 16ms per version is imperceptible.

### Query scaling (median of 5 runs)

| Versions | Baseline | \+ P1 | \+ P1 + P2 |
| --- | --- | --- | --- |
| 1 | 8\.4ms | 8\.5ms | 8\.6ms |
| 100 | 20\.3ms | 22\.7ms | 20\.8ms |
| 500 | 58\.8ms | 59\.0ms | 60\.2ms |
| 1000 | 105\.7ms | 107\.2ms | 115\.2ms |

Query path unaffected by the alias write.

### Other paths

| Metric | Baseline | \+ P1 | \+ P1 + P2 |
| --- | --- | --- | --- |
| Republish rate | 21,961/s | 21,343/s | 22,264/s |
| Query all latest (100 obj) | 28\.2ms | 27\.8ms | 35\.9ms |
| Wide create (500×5) | 159/s | 99/s | 65/s |
| Query 500 latest | 78\.1ms | 70\.5ms | 70\.8ms |

## Test updates

- `test_aliases_list_returns_distinct`: expects "latest" in list (now an explicit alias)
- `test_aliases_list_excludes_removed`: expects "latest" in list

## Test plan

- [x] Version index stability: 5/6 pass on CH (P1-T4 is known gap)
- [x] Alias tests: 127 pass on CH (2 assertions updated)
- [x] SQLite: all pass
- [x] Benchmarks: all pass
- [ ] Run full `trace` shard
- [ ] Run `trace_server` shard


rgao-coreweave:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/wandb/weave/6411?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6411** <a href="https://app.graphite.com/github/pr/wandb/weave/6411?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6411?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6408** <a href="https://app.graphite.com/github/pr/wandb/weave/6408?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6409](https://github.com/wandb/weave/pull/6409) <a href="https://app.graphite.com/github/pr/wandb/weave/6409?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

gtarpenning (commented):
can we add the updated benchmarks on this pr? (and the dowstream one too?)

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1336,3 +1336,91 @@ def test_delete_preserves_version_index_gaps(trace_server):
...
+        data=[list(ch_obj.model_dump().values())],
+        column_names=list(ch_obj.model_fields.keys()),
+    )
+
+    # No alias written — _maybe_resolve_alias should fall back to computed latest
+    resolved = server._maybe_resolve_alias(project_id, obj_id, "latest")
+    assert resolved == digest_result.digest
+
+
+def test_resolve_computed_latest_returns_none_for_missing(trace_server):
```
gtarpenning, line 1421: probably can roll this into one of the above tests

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1703,6 +1713,15 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
...
+            req.obj.project_id,
+            req.obj.object_id,
+            ["latest"],
+            digest,
+            wb_user_id=req.obj.wb_user_id or "",
+        )
+
         return tsi.ObjCreateRes(
             digest=digest,
             object_id=req.obj.object_id,
```
gtarpenning, line 1727: <p class='graphite__hidden'><i>[Re: line +1731]</i></p>
we need to make changes here too right? probably should extract the insert both logic into one function call so we can call that from both places safely. 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/wandb/weave/6411?utm_source=unchanged-line-comment">Graphite</a>.</p>

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1680,6 +1680,16 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
         if self._obj_digest_exists_non_deleted(
             req.obj.project_id, req.obj.object_id, digest
         ):
+            # Even on dedup: move "latest" alias to this digest.
+            # This matches Artifacts behavior where re-publish promotes
+            # the version to latest (see Problem 2 design doc).
+            self._insert_aliases(
```
gtarpenning, line 1686: cant we just insert this after we know the digest, instead of gating behind this conditional and calling it twice? like, can we just move this up the function?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2150,12 +2187,25 @@ def _enrich_objs_with_tags_and_aliases(
         tags_map = self._get_tags_for_objects(project_id, object_ids)
         aliases_map = self._get_aliases_for_objects(project_id, object_ids)
 
+        # Build a set of object_ids that have an explicit "latest" alias
+        # so we only fall back to computed is_latest for legacy objects.
```
gtarpenning, line 2191: consider: "Legacy objects
created before this change dan't have an explicit alias row. We synthesize 'latest' from the computed `is_latest` column only when no object with this object_id has any explicit 'latest'
alias."

PR #6440, opened 2026-03-26
chore(weave): log user request payload to dd

gtarpenning:
## Summary
- When `InvalidFieldError` is raised (e.g. user queries `columns=["status"]`), we previously only saw the error message with zero request context
- `calls_query_stream` and `calls_query_stats` now include the req in the datadog tags

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -119,7 +119,6 @@ from weave.trace_server.digest_validation import validate_expected_digest
 from weave.trace_server.errors import (
     CallsCompleteModeRequired,
     InsertTooLarge,
-    InvalidFieldError,
     InvalidRequest,
     MissingLLMApiKeyError,
     NotFoundError,
@@ -255,6 +254,36 @@ _CALLS_COMPLETE_SENTINEL_COLUMNS: list[tuple[int, str]] = [
     for idx, col in enumerate(ALL_CALL_COMPLETE_INSERT_COLUMNS)
     if ch_sentinel_values.is_sentinel_field(col)
 ]
+
+
+_DD_TAG_MAX_LEN = 500
+
+
+def _calls_query_req_dd_tags(
+    req: tsi.CallsQueryReq | tsi.CallsQueryStatsReq,
+    endpoint: str,
+) -> dict[str, str]:
+    """Extract user-provided query fields as DD span tags for debuggability.
+
+    When InvalidFieldError fires, the DD span already has the request context
+    so operators can see exactly what query triggered the error.
+    """
+    tags: dict[str, str] = {
+        "calls_query.endpoint": endpoint,
+        "calls_query.project_id": req.project_id,
+    }
+    if req.query is not None:
+        tags["calls_query.query"] = req.query.model_dump_json()[:_DD_TAG_MAX_LEN]
+    if req.filter is not None:
+        tags["calls_query.filter"] = req.filter.model_dump_json()[:_DD_TAG_MAX_LEN]
+    if isinstance(req, tsi.CallsQueryReq):
+        if req.columns is not None:
+            tags["calls_query.columns"] = ",".join(req.columns)[:_DD_TAG_MAX_LEN]
+        if req.sort_by is not None:
+            tags["calls_query.sort_by"] = ",".join(
+                f"{s.field}:{s.direction}" for s in req.sort_by
+            )[:_DD_TAG_MAX_LEN]
+    return tags
 
 
 class ClickHouseTraceServer(tsi.FullTraceServerInterface):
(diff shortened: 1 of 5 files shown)


weave/trace_server/datadog.py
```diff
@@ -78,3 +81,13 @@ def set_current_span_dd_tags(tags: dict[str, str | float | int]) -> None:
     current_span = ddtrace.tracer.current_span()
     if current_span is not None:
         current_span.set_tags(tags)
+
+
+def tag_request(
+    req: BaseModel,
+    tag_prefix: str = "calls_query",
```
jtschoonhoven: "calls_query" seems specific to one use case, but this function seems general-purpose. So I'd remove the default here. Or move the function to `weave/trace_server/clickhouse_trace_server_batched.py` since it's only used there.

weave/trace_server/datadog.py
```diff
@@ -78,3 +81,13 @@ def set_current_span_dd_tags(tags: dict[str, str | float | int]) -> None:
...
+) -> dict[str, str | float | int]:
+    """Serialize a pydantic request model as a DD span tag for debuggability."""
+    return {
+        f"{tag_prefix}.req": req.model_dump_json()[:DD_TAG_MAX_LEN],
```
jtschoonhoven: Rather than just truncating the json payload, which will give you invalid json on truncation, it would only be slightly more complicated to build this payload up one key at a time and just drop keys that would exceed the threshold

gtarpenning: fair

weave/trace_server/datadog.py
gtarpenning: true, probably can just remove the default

weave/trace_server/datadog.py
```diff
@@ -78,3 +82,25 @@ def set_current_span_dd_tags(tags: dict[str, str | float | int]) -> None:
...
+    tags: dict[str, str | float | int] = {}
+    remaining = DD_TAG_MAX_LEN
+    for key, value in req.model_dump().items():
+        if isinstance(value, BaseModel):
+            serialized = value.model_dump_json()
+        else:
+            serialized = json.dumps(value)
+        if len(serialized) <= remaining:
+            tags[f"{tag_prefix}.{key}"] = serialized
+            remaining -= len(serialized)
```
jtschoonhoven, line 105: Nit: is `DD_TAG_MAX_LEN` a limit on just the tag values? The values + tag names? The overall payload? You might also want to account for the length of the tag names here. Or just make sure that `DD_TAG_MAX_LEN` has enough headroom that any of that is unlikely to matter. This nit might be ignorable but I wanted to check.

weave/trace_server/datadog.py
```diff
@@ -78,3 +82,25 @@ def set_current_span_dd_tags(tags: dict[str, str | float | int]) -> None:
...
+    """Serialize a pydantic request model as DD span tags for debuggability.
+
+    Builds tags one field at a time, dropping fields that would exceed the
+    per-tag size limit so every emitted value is valid JSON.
+    """
+    tags: dict[str, str | float | int] = {}
+    remaining = DD_TAG_MAX_LEN
+    for key, value in req.model_dump().items():
+        if isinstance(value, BaseModel):
+            serialized = value.model_dump_json()
```
jtschoonhoven, line 100: `model_dump` is recursive, so this branch is dead code

jtschoonhoven (approved):
Couple notes but LGTM

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1000,6 +1001,7 @@ def calls_query_stats(self, req: tsi.CallsQueryStatsReq) -> tsi.CallsQueryStatsR
         """Returns a stats object for the given query. This is useful for counts or other
         aggregate statistics that are not directly queryable from the calls themselves.
         """
+        set_current_span_dd_tags(tag_request(req, "calls_query_stats"))
```
jtschoonhoven, line 1004: Are there security concerns here? Are we going to log sensitive data from the request?

gtarpenning: This was really what I wanted to cover in the review process. I believe not. The request payload _should_ only include filter criteria. If users are storing sensitive information in plain text, and then trying to filter on those sensitive payloads, yeah we will see it. Technically we can already see these in the _logs_, when a stream query request is made, but this will make the request information more available for datadog viewers. I'm not 100% sure if thats kosher from a legal/compliance perspective, but from a security one it doesn't increase the cardinality of our exposure, merely the availability of the information. I think the gain in debugging is worth the cost, but i'm willing to negotiate on that point. One potential mitigation would be to only log the `req` span on error?

jtschoonhoven: I'm not up to speed on how stringent coreweave/wandb are on PII concerns. My instinct is to ship it for the reasons you mention. But we could confirm with someone on the compliance team I suppose 🤷

jtschoonhoven: A middle ground might be to specify which specific keys from the request payload you want to include in the tags. If the client is filtering on PII then we'd still see that in datadog, but the scope of the issue feels more constrained that way.

jtschoonhoven: If you wanted to get fancy (and not feel obligated to involve compliance) you could scrub _all_ scalar values from the request payload. It's probably most useful to understand the structure of the query vs the specific values they're filtering on.

This PR still seems OK as-is to me but, like I said, I'm not a subject matter expert on PII at wandb.

PR #6475, opened 2026-03-27
chore(weave): add ddtrace instrumentation to otel_export

gtarpenning:
## Summary

Datadog flame graphs for `POST /otel/v1/traces` show uninstrumented gaps inside `otel_export` before table routing. This extracts three helpers and wraps them with `@ddtrace.tracer.wrap()` to make that time visible:

- `_otel_parse_and_convert_spans` — protobuf deserialization + `to_call` conversion
- `_otel_resolve_ops` — op cache lookup, CH query, placeholder creation
- `_otel_convert_and_insert_rows` — row conversion + CH batch insert

Refactor only — no behavior change.

## Test plan
- [x] Verify flame graph shows new child spans under `otel_export` in Datadog QA

master:
Screenshot 2026-03-30 at 8 51 03 AM

You can see the span looks much more complete with the new branch:
image


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -578,9 +630,16 @@ def otel_export(self, req: tsi.OTelExportReq) -> tsi.OTelExportRes:
                             wb_run_id=wb_run_id,
                         )
                     )
+        return calls, rejected_spans, error_messages
```
zbirenbaum: lets group the error with the span. Maybe a new dataclass to keep the association.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -554,9 +608,7 @@ def otel_export(self, req: tsi.OTelExportReq) -> tsi.OTelExportRes:
                     try:
                         span = Span.from_proto(proto_span, resource)
                     except AttributePathConflictError as e:
-                        # Record and skip malformed spans so we can partially accept the batch
```
zbirenbaum, line 557: This comment actually seems pretty useful.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -533,13 +533,67 @@ def _ensure_placeholder_file_exists(self, project_id: str) -> None:
...
+    @ddtrace.tracer.wrap(
+        name="clickhouse_trace_server_batched.otel_parse_and_convert_spans"
+    )
+    def _otel_parse_and_convert_spans(
+        self, req: tsi.OTelExportReq
+    ) -> tuple[
+        list[tuple[tsi.StartedCallSchemaForInsert, tsi.EndedCallSchemaForInsert]],
+        int,
+        list[str],
+    ]:
```
zbirenbaum: This type is pretty hard to read, could we make a container or two for it? OTelSpanParseResult maybe?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -533,13 +533,67 @@ def _ensure_placeholder_file_exists(self, project_id: str) -> None:
     @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched.otel_export")
     def otel_export(self, req: tsi.OTelExportReq) -> tsi.OTelExportRes:
         assert_non_null_wb_user_id(req)
+
+        calls, rejected_spans, error_messages = self._otel_parse_and_convert_spans(req)
+
+        set_current_span_dd_tags({"weave_trace_server.insert_call_count": len(calls)})
```
zbirenbaum: We should log the rejected spans with reasons in datadog. We don't care much about successful spans except for usage tracking. Failed spans are the real insight.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -593,10 +652,8 @@ def otel_export(self, req: tsi.OTelExportReq) -> tsi.OTelExportRes:
         )
 
         for op_name, op_ref_uri in existing_ops.items():
-            # Modify each of the matched start calls in place
             for idx in obj_id_idx_map[op_name]:
                 calls[idx][0].op_name = op_ref_uri
-            # Remove this ID from the mapping so that once the for loop is done we are left with only new objects
```
zbirenbaum, line 599: Definitely keep these comments. They were specifically requested during the review for this PR.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -554,9 +608,7 @@ def otel_export(self, req: tsi.OTelExportReq) -> tsi.OTelExportRes:
                     try:
                         span = Span.from_proto(proto_span, resource)
                     except AttributePathConflictError as e:
```
zbirenbaum, line 612: Not in scope but I want to do a followup to catch more than this...

zbirenbaum (changes requested):
Few notes. Am I correct in my understanding that this mostly just logs spans for each call? I'm curious exactly what data is being recorded.

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: i'm just trying to log the duration, error logging should handle this

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: done

weave/trace_server/trace_server_interface.py
```diff
@@ -360,6 +360,23 @@ class OTelExportRes(BaseModel):
...
+    error: str
+    name: str = ""
+    trace_id: str = ""
+    span_id: str = ""
+
+
+class OTelSpanParseResult(BaseModel):
+    calls: list[CallPair] = Field(default_factory=list)
+    rejected: list[RejectedOTelSpan] = Field(default_factory=list)
+
```
zbirenbaum: Why not Iterable?

zbirenbaum (approved):
One comment, looks good!

weave/trace_server/trace_server_interface.py
gtarpenning: idk because we are doing like indexing and mutation over it? I feel like iterable is normally only used in specific cases like streaming

PR #6481, opened 2026-03-28
feat(weave): write explicit 'latest' alias on every obj_create (WB-32435)

rgao-coreweave:
## Summary

Make `obj_read("latest")`, `objs_query(latest_only=True)`, and `objs_query(aliases=["latest"])` behave correctly across **both** real-world scenarios this PR cares about:

1. **Re-publishing existing content promotes it to `"latest"`** — the headline WB-32435 / W&B Artifacts-parity requirement.
2. **Deleting the version that currently holds `"latest"` falls back to the next surviving version** — no silent disappearance.

The shipped design is **hybrid `is_latest`**: an explicit `"latest"` alias row in the `aliases` table is the override; absence of a live alias row falls through to a computed most-recent-surviving rank over `object_versions`. The SQL `is_latest` projection unions the two paths, so every caller (alias filters, `latest_only`, `obj_read("latest")`) goes through one expression.

Mental model: **`"latest"` is a pointer with a default**, the same way `HEAD` works in git. The alias is the override; natural ordering is the default. Not "two sources of truth" — one model with a well-defined default and an override mechanism for the cases that need it.

## Background

Pre-PR, `is_latest` was a window function over `object_versions` ranked by `(deleted_at IS NULL) DESC, _first_created_at DESC, digest DESC`. That can't express "this older digest was just re-published and should now be latest," because dedup-republish reuses the existing `object_versions` row — `_first_created_at` doesn't change, so any ordering-based notion of latest is stuck on the most-recently-published-by-time digest.

The product behavior we need (matching W&B Artifacts):

- publish v0 (content A) → latest = A
- publish v1 (content B) → latest = B
- re-publish content A (dedup hit, no new `object_versions` row) → **latest should move back to A**

This requires an explicit override mechanism. The aliases table provides it.

### Before
toggling a monitor off creates a new digest (active=false) which becomes "latest" by first-seen time; toggling it back on dedupes to the prior active=true digest but doesn't move "latest" off the active=false row, so the UI keeps showing the monitor as off.

### After
the explicit-latest-alias write makes `"latest"` an alias that `obj_create` re-asserts on every call (including dedup), and the hybrid `is_latest` ensures the alias wins over natural ordering.

## The regression that drove the redesign

The PR's first cut (commits up through `28dd3704a4`) made the aliases table the **sole** source of truth for `"latest"`. It broke in one scenario: when `obj_delete` removes the version currently holding `"latest"`, the cascade tombstones the alias row, and nothing re-points `"latest"` to the surviving most-recent version. The author's red-team test `test_delete_current_latest_promotes_prior_surviving_version` caught this — and was the failing test that blocked CI.

A second red-team test (`test_migration_031_alias_shadow_picks_wrong_digest_as_latest`, added by gtarpenning) caught a separate but related bug: the migration that backfills `"latest"` alias rows for legacy objects was producing the **wrong digest** for every legacy object whose newest version wasn't lex-smallest in its digest history.

## Alternatives explored and rejected

**Option A — repoint `"latest"` inside `obj_delete`.** After cascading tombstone of the alias row, look up the surviving most-recent version via `FINAL` on `object_versions` and write a fresh `"latest"` alias row pointing at it. Explored three sub-variants:

- **A1: write the repoint at `now()`.** Loses the race against a concurrent `obj_create(C)` whose alias write lands *before* `obj_delete`'s repoint — the janitorial repoint clobbers the fresh publish, latest = surviving-at-delete-time instead of = C. Real correctness bug, not just aesthetic.

- **A2: pin the repoint's `created_at` to `surviving_version.object_versions.created_at`** (intuition: deterministically lose to a fresh `now()` publish). Walks: `publish A, publish B, republish-A-via-dedup, delete A`. After step 4, the alias row for surviving B would be written at `t1` (B's `object_versions.created_at`), but the tombstone for A was written at `t_delete > t1`. Under `ReplacingMergeTree(created_at)` with `ORDER BY (project_id, object_id, alias)` (digest **not** in the sort key, migration 025), the tombstone wins the merge, the `HAVING argMax(deleted_at, created_at) = 0` filter drops the row, and there's still no latest. Breaks the very dedup-republish feature this PR exists to support.

- **A3: pin the repoint's `created_at` to `tombstone.created_at + 1µs`, emitted atomically in one INSERT with the tombstone.** Solves the dedup-republish case (repoint deterministically wins over the just-emitted tombstone) and concurrent-delete races (idempotent — both deletes converge on the same `(digest, created_at)` pair). Still has a narrow window where a concurrent `obj_create(C)` lands its alias write **before** the delete's pair-insert: `obj_delete`'s repoint at `t_del + 1µs` outranks `obj_create`'s alias at `t_create_alias`, silently demoting a fresh publish. The clean fix requires linearization (Redis / CH Keeper) — out of scope, no new infra for this.

**Option B — query-time hybrid (chosen).** When no live alias row exists for `(project_id, object_id)`, fall back to a computed window-function rank over `object_versions` (CH) or the column-based `objects.is_latest` (SQLite). The aliases table remains the override that drives dedup-promote behavior; the natural-ordering default takes over when the override is absent. No `obj_delete` repoint logic, no concurrency invariants to engineer, no `FINAL` queries, legacy data handled for free.

**Option C — hard-delete the alias row in `obj_delete` instead of soft-delete.** Doesn't actually fix the failing test, since there's no older live alias row for the surviving version (`"latest"` was only ever written for the digest being deleted). Rejected.

## Why hybrid is the right model

We initially worried this was "two sources of truth" — the same concern that motivated the alias-only design. On closer review, it isn't:

- The `aliases` table is still the **only** mechanism that can express "this digest is latest even though it isn't most-recent-by-publish-time" (the dedup-republish-promote semantic).
- The computed fallback is the **default** behavior when no override is set. That's not a competing source of truth; that's a default.

Same shape as `HEAD` in git, the default branch in any VCS, `current` in k8s. Override-with-default is **one** model, not two.

What we give up: a small performance cost on the read path (a second window function over `latest_row_per_digest`), bounded by project size and the already-filtered `rn=1` rows.

### CH read CTE — `weave/trace_server/query_builder/objects_query_builder.py`

```sql
object_versions_with_index AS (
    SELECT *,
        row_number() OVER (PARTITION BY project_id, kind, object_id
                           ORDER BY _first_created_at ASC, digest ASC) - 1 AS version_index,
        count(*) OVER (PARTITION BY project_id, kind, object_id) AS version_count,
        row_number() OVER (PARTITION BY project_id, kind, object_id
                           ORDER BY (deleted_at IS NULL) DESC,
                                    _first_created_at DESC, digest DESC) AS _computed_latest_rank,
        if(
            (project_id, object_id, digest) IN
                (SELECT project_id, object_id, digest FROM latest_alias_per_object)
            OR (
                _computed_latest_rank = 1
                AND (project_id, object_id) NOT IN
                    (SELECT project_id, object_id FROM latest_alias_per_object)
            ),
            1, 0
        ) AS is_latest
    FROM latest_row_per_digest
    WHERE rn = 1
)
```

The fallback **only** fires when no live alias row exists for the `(project_id, object_id)` — guarantees the alias path and computed path never both claim is_latest=1 for the same object.

### SQLite read expression — `_IS_LATEST_FROM_ALIASES_SQL`

```sql
(
    (objects.project_id, objects.object_id, objects.digest) IN
        (SELECT project_id, object_id, digest FROM aliases WHERE alias = 'latest')
    OR (
        (objects.project_id, objects.object_id) NOT IN
            (SELECT project_id, object_id FROM aliases WHERE alias = 'latest')
        AND objects.is_latest = 1
    )
)
```

SQLite's fallback uses the column-based `objects.is_latest` (master-style), maintained on insert by `obj_create` AND re-pointed by `obj_delete` to the most-recent surviving version (new in this PR — see next section).

### SQLite `obj_delete` re-points the column

After the cascade hard-delete of alias rows, `obj_delete` runs a single `UPDATE … CASE WHEN digest = (SELECT … LIMIT 1) THEN 1 ELSE 0 END` inside the existing transaction to set `objects.is_latest = 1` on the most-recent non-deleted version of the object_id (and 0 on the rest). Symmetric to CH's window-function fallback. Without this, the SQLite fallback would read a stale column where a deleted digest still has `is_latest = 1`, returning the wrong (or no) answer.

### Resolver, enrichment, filter routing

- **`_maybe_resolve_alias`** (both backends): unchanged contract — resolves any alias name, including `"latest"`, via the aliases table. Returns `None` when no live alias row exists (the HAVING tombstone filter excludes tombstones). When it returns `None` for `"latest"`, `obj_read` falls through to `digest = "latest"` → `is_latest = 1` via the hybrid SQL expression, which handles both alias and computed-fallback paths.
- **`_enrich_objs_with_tags_and_aliases`** (both backends): synthesizes `"latest"` into `obj.aliases` when `obj.is_latest == 1` and `"latest"` isn't already in the explicit alias map. Includes a **read-skew guard** that precomputes the set of `object_id`s with a live `"latest"` alias in the just-fetched aliases map and suppresses synthesis on any digest whose `object_id` is in that set — protects against a concurrent `obj_create` write landing between the metadata query (T1) and the aliases-map query (T2).
- **`add_aliases_condition`** (CH) and the SQLite mirror inside `objs_query`: route `aliases=["latest"]` through the hybrid `is_latest` column, not the raw alias subquery. A raw subquery alone would miss objects whose alias row was tombstoned by `obj_delete` and now rely on the computed/column fallback. Non-`"latest"` aliases use the raw alias subquery as before. Mixed filters like `aliases=["latest", "prod"]` OR the two paths.

### Write side

- ClickHouse `obj_create` / `obj_create_batch`: call `_insert_aliases("latest")` after every insert (dedup path included).
- SQLite `obj_create`: `INSERT OR REPLACE INTO aliases` for `"latest"` in both dedup and new-insert paths, inside `BEGIN IMMEDIATE TRANSACTION`.

## Migration 031 alias-shadow fix

The migration that backfills explicit `"latest"` alias rows for legacy objects was producing the wrong digest for any object whose newest version wasn't lex-smallest in its digest history. Two CH analyzer rewrites compounded:

1. **WHERE alias-shadow**: the SELECT projected `toDateTime64(0, 3) AS deleted_at`, and the analyzer resolved `deleted_at` in the WHERE clause to that constant projection alias instead of the `object_versions.deleted_at` column. The predicate `deleted_at = toDateTime64(0, 3)` became `toDateTime64(0,3) = toDateTime64(0,3)`, always TRUE — soft-deleted rows were included in the backfill.
2. **argMax alias-shadow**: the SELECT projected `toDateTime64('1970-01-01 …', 3) AS created_at`, and `argMax(digest, created_at)` resolved its second argument to that constant. All rows tied on a constant; CH broke the tie by storage order (`ORDER BY (project_id, kind, object_id, digest)`), so the lex-smallest digest won.

Net effect: migration 031 set `"latest"` to the lex-smallest digest of all rows in `object_versions` for every legacy object.

Fix: qualify all column references with the source-table alias (`ov.project_id`, `ov.digest`, `ov.created_at`, `ov.deleted_at`) and use `WHERE ov.deleted_at IS NULL` to avoid the constant-comparison shadow.

The hybrid design makes the backfill **less load-bearing** in general — the API's hybrid `is_latest` falls back to the computed window when no alias row exists — but a wrong backfill row is worse than no backfill row (it would actively mis-credit the wrong digest as latest). The fix is required.

## Concurrency

The hybrid makes `obj_create` ↔ `obj_delete` races benign:

- Concurrent `obj_create(C)` lands its alias write → alias path returns C.
- `obj_delete` tombstones first, then `obj_create(C)` lands its version row but the alias write fails or races out → alias path empty for this object → fallback to computed → `_first_created_at` ordering surfaces C anyway.
- `obj_delete` cascades a tombstone for a digest that wasn't actually `"latest"` → alias path still matches the still-live `"latest"` row pointing at a different digest. No change.

The read path has one residual narrow window (the metadata query and aliases-map query are two reads of the `aliases` table; a concurrent write can move `"latest"` between them, leaving `is_latest=1` on the stale digest while the aliases map credits a different digest). The read-skew guard in `_enrich_objs_with_tags_and_aliases` closes the user-visible inconsistency — no response ever shows two digests both claiming `"latest"`.

No new locks, no Keeper, no Redis, no `FINAL` queries.

## Test plan

- [x] **`test_delete_current_latest_promotes_prior_surviving_version`** (CH-only red-team test in `tests/trace_server/test_clickhouse_trace_server_batched.py`) — the previously-failing CI test. Passes via the computed fallback. No `obj_delete` repoint code needed.
- [x] **`test_migration_031_alias_shadow_picks_wrong_digest_as_latest`** (CH-only red-team test) — passes after the migration fix (qualified `ov.*` column references, `IS NULL`).
- [x] **`test_hybrid_latest_full_lifecycle`** (both backends, in `tests/trace/test_obj_tags_aliases.py`) — walks the full lifecycle in one test: publish A/B/C → linear-delete C → fallback to B → dedup-republish A → alias path wins → delete A → fallback to B → delete B → no latest → publish D → self-heal. At every step, asserts `obj_read('latest')`, `objs_query(latest_only=True)`, and `objs_query(aliases=['latest'])` agree, via a shared `_resolve_latest_digest` helper.
- [x] **`test_hybrid_latest_multi_object_query`** (both backends) — three objects in one query: one resolves via the alias path, one via the fallback path (post-delete), one with a non-`"latest"` alias `prod`. Covers `latest_only=True` across all three AND the mixed `aliases=["latest", "prod"]` filter.
- [x] **Existing `test_republish_promotes_to_latest` and `test_alias_resolution`** still pass — confirms the dedup-promote headline feature works under hybrid.
- [x] **Updated query-builder SQL snapshot tests** to reflect the new CTE and the special-cased `aliases=["latest"]` filter.
- [x] **CH obj_create probe tests** updated to call `obj_read("latest")` (public API) instead of `_maybe_resolve_alias` (kept its original contract).
- [x] **Full alias suite** (`tests/trace/test_obj_tags_aliases.py`, 50 tests): pass on SQLite, pass on ClickHouse.
- [x] **Full `trace_server` shard**: SQLite 1028 passed; ClickHouse 669 passed.
- [x] **Ruff check + format** clean on touched files.

## Performance

The hybrid adds one extra `row_number()` window over `latest_row_per_digest` (already filtered to `rn=1`). The dominant read cost is the existing aliases-table scan, which the alias-only design already paid. Rough magnitudes from earlier benchmarks (100k objects × 1 version per object on a synthetic dataset):

| Scale | OLD (computed-only) | NEW (alias-only) | Hybrid (estimated) |
|---|---|---|---|
| 1k objs × 1 ver | 8 ms | 8 ms | ~9 ms |
| 10k objs × 1 ver | 16 ms | 17 ms | ~19 ms |
| 100k objs × 1 ver | 90 ms | 105 ms (+17%) | ~110–130 ms |
| 10k objs × 10 ver | 82 ms | 74 ms (−10%) | ~85 ms (wash) |

The worst case is high-object-count, low-versions-per-object projects. Acceptable given correctness on the headline product semantic AND the delete-current-latest fallback, all without new infra or write-side coordination. Worth watching in prod telemetry.

## Changes

**Read path (hybrid):**
- `weave/trace_server/query_builder/objects_query_builder.py`: `make_metadata_query` adds `_computed_latest_rank` window and ORs it into the `is_latest` projection; `add_aliases_condition` special-cases `"latest"` → `is_latest = 1` with a clear docstring on the routing.
- `weave/trace_server/clickhouse_trace_server_batched.py`: `_enrich_objs_with_tags_and_aliases` synthesizes `"latest"` when `is_latest == 1` and absent, with a read-skew guard against concurrent alias writes.
- `weave/trace_server/sqlite_trace_server.py`: `_IS_LATEST_FROM_ALIASES_SQL` is now hybrid; `_select_objs_query` projection mirrors it; `objs_query` aliases filter special-cases `"latest"`; `_enrich_objs_with_tags_and_aliases` synthesizes with the same read-skew guard.

**Write side (unchanged from earlier in the PR — still needed for the override):**
- ClickHouse `obj_create` / `obj_create_batch`: `_insert_aliases("latest")` after every insert (dedup path included).
- SQLite `obj_create`: `INSERT OR REPLACE INTO aliases` for `"latest"` in both dedup and new-insert paths.

**SQLite `obj_delete` re-points `objects.is_latest`** to the most-recent surviving version after the cascade — symmetric to CH's window-function fallback.

**Migration 031 fix**: qualified `ov.*` column references and `IS NULL` to defeat CH analyzer alias-shadow.

**Docs / comments:**
- Updated TODOs in `clickhouse_schema.py` and the SQLite `objects` table DDL to describe `is_latest` as hybrid (override + default) rather than "drop the column".
- Migration 031 comment: backfill is for consistency, not correctness; documents the alias-shadow trap.

## Known limitations / future work

- The hybrid read path costs a second window function vs alias-only. Watch in prod telemetry.
- gtarpenning's two open nits on earlier commits are not addressed in this PR:
  - The untyped 4-tuple in CH `obj_create_batch` (build `AliasCHInsertable` directly inside the loop).
  - The `_CursorProxy` class in `test_sqlite_trace_server.py` (use `Mock(wraps=...)` instead).
- (Footnote per Tim's review) `latest_alias_per_object` doesn't filter or group by `kind`; the computed rank partitions by `(project_id, kind, object_id)`. Today `object_id` is unique per project so this isn't a bug, but if we ever allow the same `object_id` across kinds, this would be a footgun. Not addressed here.


rgao-coreweave:
* **#6481** <a href="https://app.graphite.com/github/pr/wandb/weave/6481?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6481?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

rgao-coreweave:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rgao-coreweave:
### Code review

Found 1 issue:

1. ~~SQLite `_maybe_resolve_alias` was not updated to resolve "latest" from the aliases table. It still has `if digest == "latest": return None` (line 4760), which bypasses the alias table entirely. The ClickHouse version was updated to resolve "latest" from aliases with a computed fallback, but the SQLite path was missed. This means the core PR semantic (re-publishing old content promotes to latest) does not work for SQLite `obj_read("latest")`. The dedup test (`test_obj_create_moves_latest_alias_on_dedup`) only runs against ClickHouse via `ch_server`.~~

https://github.com/wandb/weave/blob/054b8edaf5797004e6b4d8e2797f846a7345ee6b/weave/trace_server/sqlite_trace_server.py#L4759-L4762

jtschoonhoven:
@rgao-coreweave mind adding a comment to explain why we have to write to the aliases table instead of deriving `is_latest` from `created_at`?

weave/trace_server/query_builder/objects_query_builder.py
```diff
@@ -368,6 +364,14 @@ def make_metadata_query(self) -> str:
     ) AS fc USING (object_id, digest)
     WHERE ov.project_id = {{project_id: String}}{self.object_id_conditions_part}
 ),
+latest_alias_per_object AS (
```
gtarpenning, line 392: whats the performance implications of this? at first glance this adds an additional entire db read, not sure that was expected. When we talked about this it seemed like the perf hit was on the write side, but its also on the read side?

rgao-coreweave:
> @rgao-coreweave mind adding a comment to explain why we have to write to the aliases table instead of deriving `is_latest` from `created_at`?

Because dedup re-publishes don't change `created_at`.

  The product requirement (WB-32435, matching W&B Artifacts) is: re-publishing existing   
  content must promote it to latest. When you re-publish content whose digest already     
  exists, the write dedups to the existing object_versions row — no new row, no updated   
  created_at. 

  So any is_latest derived from created_at (or _first_created_at from the   
  object_version_first_seen MV, which only tracks the earliest created_at per digest) is
  structurally unable to express "this old digest was just re-published and should now be
  latest."

weave/trace_server/query_builder/objects_query_builder.py
rgao-coreweave: There is one extra scan against aliases per make_metadata_query call (this query is in the hot path — objs_query and any obj_read with digest="latest" go through it). 

 1. The aliases table is much smaller than object_versions.
  2. The PREWHERE+sort-key prefix make it cheap. Sort key starts with project_id, so CH 
  prunes to that project's granules before applying WHERE alias = 'latest'.
  3. Same shape as the existing add_aliases_condition subquery. Filtering by arbitrary    
  aliases (e.g. prod) already did this. The new CTE just generalizes it to latest and     
  lifts it once into the WITH so it's not re-evaluated.                                 
  4. The previous design wasn't free either. It did a per-partition row_number() sort     
  across every latest_row_per_digest row. The new path replaces sort+rank over a large CTE
   with a hash IN against a small one.

rgao-coreweave: > whats the performance implications of this? at first glance this adds an additional entire db read, not sure that was expected. When we talked about this it seemed like the perf hit was on the write side, but its also on the read side?

there is an extra read, and I should have called
that out earlier. I benchmarked it against a synthetic dataset on a local
single-node CH; numbers below.

**Setup:** seeded one synthetic project at four scales, ran each query 7×
after 2 warmups, pulled durations from `system.query_log`. Same dataset
served both variants. NEW = current CTE-based `is_latest`; OLD =
hand-rewritten window-function variant from before this change.

| Scale | OLD p50 | NEW p50 | Δ | Alias CTE in isolation |
|---|---|---|---|---|
| 1k objs × 1 ver | 8 ms | 8 ms | +0% | 1 ms |
| 10k objs × 1 ver | 16 ms | 17 ms | +6% | 5 ms |
| **100k objs × 1 ver** | **90 ms** | **105 ms** | **+17%** | 28 ms |
| **10k objs × 10 ver** | **82 ms** | **74 ms** | **−10%** | 5 ms |

**Three things worth noting:**

1. **The extra read scales with objects, not versions.** The alias CTE is
   `PREWHERE project_id = ... WHERE alias = 'latest'` against a table sorted
   by `(project_id, object_id, alias)`. It reads ~one row per object,
   bounded by the project's object count. At 100k objects it costs ~28ms
   standalone; at 10k it's ~5ms.

2. **For projects with many versions per object the change is a wash or a
   win.** In the 10k × 10 case NEW is actually 10% faster, because we
   replaced a `row_number() OVER (PARTITION BY ... ORDER BY ...)` window-sort
   over all version rows with a hash IN against the small CTE. `EXPLAIN
   PIPELINE` confirms: NEW drops one outer `WindowTransform`+sort and adds a
   `CreatingSets`/`DelayedPorts` step for the IN subquery.

3. **The worst case is high-object, low-version projects** (e.g. a project
   that publishes lots of distinct artifacts but rarely re-versions them).
   That's where the alias scan is unamortized: ~+15ms / +17% at 100k
   objects. We don't have many projects shaped like this today, but it's
   the shape to watch in prod telemetry.

**What I'm planning to do:**

- Land as-is, since the correctness win (dedup-promotes-to-latest matches
  Artifacts) is real and the regression is bounded and only hits an
  atypical shape.

jtschoonhoven (approved):
LGTM but I'd like to see @gtarpenning sign off on the performance implications before merging

rgao-coreweave:
> LGTM but I'd like to see @gtarpenning sign off on the performance implications before merging

Got signed off from Griffen: https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1778697277165449?thread_ts=1777940966.409049&cid=C03BSTEBD7F

Locally tested again, will land this PR.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1852,6 +1880,7 @@ def obj_create_batch(
                 f"obj_create_batch only supports updating a single project. Supplied projects: {unique_projects}"
             )
 
+        alias_rows: list[tuple[str, str, str, str]] = []
```
gtarpenning, line 1915: untyped 4-tuple is easy to misindex. build the `AliasCHInsertable` rows directly inside the existing `for obj in batch` loop and drop the intermediate list entirely — one pass, named fields.

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1480,3 +1480,114 @@ def run_slow_query() -> None:
...
+
+    # Establish a prior latest.
+    r1 = _obj_create(ch_server, project_id, obj_id, {"v": 1})
+    assert ch_server._maybe_resolve_alias(project_id, obj_id, "latest") == r1.digest
+
+    # Simulate failure on the alias write for the next obj_create.
+    def failing_insert_aliases(*args, **kwargs):
+        raise RuntimeError("simulated alias write failure")
+
+    ch_server._insert_aliases = failing_insert_aliases  # type: ignore[method-assign]
```
gtarpenning: `monkeypatch.setattr(ch_server, "_insert_aliases", failing_insert_aliases)` here — drops the type-ignore and the try/finally del cleanup.

tests/trace_server/test_sqlite_trace_server.py
```diff
@@ -404,3 +408,97 @@ def test_sqlite_calls_query_empty_thread_ids_against_real_db():
...
+
+def _make_obj_create_req(
+    project_id: str, object_id: str, val: dict
+) -> tsi.ObjCreateReq:
+    return tsi.ObjCreateReq(
+        obj=tsi.ObjSchemaForInsert(project_id=project_id, object_id=object_id, val=val)
+    )
+
+
+class _CursorProxy:
```
gtarpenning: nit:`Mock(wraps=real_cursor, execute=side_effect)` would do the same thing in ~3 lines and read more clearly.

gtarpenning (commented):
couple nits on style

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: `list[list[Any]]` leaks `Any` past this boundary. Build a `list[AliasCHInsertable]` and serialize at the `self._insert` call site — drops the intermediate list, and the typing dies in exactly one place (`model_dump().values()`) instead of being defined as untyped from the start.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -2272,13 +2326,31 @@ def _enrich_objs_with_tags_and_aliases(
         tags_map = self._get_tags_for_objects(project_id, object_ids)
         aliases_map = self._get_aliases_for_objects(project_id, object_ids)
```
gtarpenning: This block + its comment lands verbatim at sqlite_trace_server.py:5050. Consider extracting to a shared helper (e.g. `apply_latest_synth(objs, aliases_map)` in trace_server_common)

weave/trace_server/sqlite_trace_server.py
```diff
@@ -4966,12 +5040,31 @@ def _enrich_objs_with_tags_and_aliases(
...
         aliases_map = self._get_aliases_for_objects(project_id, object_ids)
 
+        # Read-skew guard: the projection query and the aliases-map query
+        # are two separate reads of the `aliases` table. If a concurrent
+        # write moves "latest" between them, we can see is_latest=1 on the
+        # old digest (from the first read) while the aliases map already
+        # credits "latest" to the new digest. Skip synthesizing "latest" on
+        # any digest whose object_id already has an explicit "latest" alias
+        # in the just-fetched map.
+        object_ids_with_alias_latest = {
```
gtarpenning: Verbatim copy of the CH version at clickhouse_trace_server_batched.py:2328 — consider moding to ts-common, or an object utils context

tests/trace_server/test_sqlite_trace_server.py
```diff
@@ -404,3 +408,160 @@ def test_sqlite_calls_query_empty_thread_ids_against_real_db():
...
+    conn_map[sqlite_db_path] = slts.ConnCursor(real_conn, proxy)
+    try:
+        with pytest.raises(RuntimeError, match="simulated alias write failure"):
+            server.obj_create(_make_obj_create_req(project_id, object_id, {"v": 2}))
+    finally:
+        conn_map[sqlite_db_path] = saved
+        # Defensive: clear any lingering open transaction from the failed call.
+        try:
+            real_conn.rollback()
+        except Exception:
```
gtarpenning: Narrow to `sqlite3.Error`, or drop the try/except entirely if rollback is provably safe here. Bare `except Exception: pass` in cleanup hides real regressions in the next test's setup, is this prior art? adding something new here?

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1480,3 +1480,761 @@ def run_slow_query() -> None:
...
+    _seed_alias_row(ch_server, project_id, obj_id, digest_dead, t_alias)
+
+    # _maybe_resolve_alias resolves to the digest (it only knows about the
+    # aliases table, not whether the target is alive in object_versions).
+    assert ch_server._maybe_resolve_alias(project_id, obj_id, "latest") == digest_dead
+
+    # obj_read must NOT return the tombstoned version.  Either raises, or
+    # falls through to the CTE's computed fallback — which itself filters
+    # tombstones via `(deleted_at IS NULL) DESC` and would find nothing.
+    with pytest.raises((NotFoundError, Exception)) as exc_info:
```
gtarpenning: consider Tighten to `pytest.raises((NotFoundError, ObjectDeletedError))`.

tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1480,3 +1480,761 @@ def run_slow_query() -> None:
...
+    # Up again: backfill re-added.  Real alias still wins.
+    ch_server._query(_MIGRATION_031_UP_SQL, {"p": project_id, "o": obj_id})
+    assert (
+        ch_server._maybe_resolve_alias(project_id, obj_id, "latest") == digest_real
+    ), (
+        "after second up: real alias should still outrank a freshly-inserted backfill row"
+    )
+
+
+def test_obj_create_batch_duplicate_object_id_last_entry_wins_latest(ch_server):
```
gtarpenning, line 1744: does this flake?

weave/trace_server/migrations/031_backfill_latest_alias.up.sql
```diff
@@ -0,0 +1,40 @@
...
+--      constant `toDateTime64(0, 3) AS deleted_at` — making the predicate
+--      `toDateTime64(0,3) = toDateTime64(0,3)`, always TRUE, which would
+--      include tombstoned rows in the backfill.
+--   2. `argMax(ov.digest, ov.created_at)` uses qualified column names so
+--      the second argument binds to `object_versions.created_at`, not the
+--      SELECT-projection constant `toDateTime64('1970-01-01 ...', 3) AS
+--      created_at`. Without qualification CH would tie on the constant
+--      for every row and break ties by storage order (lex-smallest
+--      digest wins), producing the wrong "latest" digest.
+INSERT INTO aliases (project_id, object_id, alias, digest, wb_user_id, created_at, deleted_at)
```
gtarpenning: Heads up for deploy: this is one large INSERT...SELECT scanning the full `object_versions` table with no per-project chunking. For large deployments worth either batching per project or running in a maintenance window. we should confirm that this runs without issue on a prod replica

weave/trace_server/clickhouse_trace_server_batched.py
rgao-coreweave: Decided to leave `obj_create_batch` as-is in eb42c2e5. It's an internal OTel-only path (never user-facing), and on that path the original `alias_rows` intermediate + second-pass `AliasCHInsertable` comprehension is fine — not worth churning a hot path that already has its own perf invariants. Resolving.

tests/trace_server/test_clickhouse_trace_server_batched.py
rgao-coreweave: Done in 8bd91640 — switched to `mock.patch.object(ch_server, "_insert_aliases", ...)`. I went with `mock.patch.object` over `monkeypatch.setattr` because the former correctly `delattr`s on exit when the original was a class method (it inspects `target.__dict__` to decide between `delattr` and setting back), so the session-scoped `ch_server.__dict__` stays clean and `test_reset_server_state_covers_all_attrs` still passes. Resolving.

tests/trace_server/test_sqlite_trace_server.py
rgao-coreweave: Done in 8bd91640 — `_CursorProxy` replaced with `Mock(wraps=real_cursor)` plus a local `_execute_or_fail` that raises on the alias INSERT and delegates everything else to the wrapped cursor. Resolving.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1920,6 +1952,28 @@ def obj_create_batch(
...
+                    .values()
+                )
+                for (project_id, object_id, digest, wb_user_id) in alias_rows
+            ]
+            self._insert(
+                "aliases",
+                data=ch_alias_rows,
+                column_names=list(AliasCHInsertable.model_fields.keys()),
+            )
+
```
gtarpenning, line 1977: can we de-dupe by digest or something, can we make sure we only insert the minimal set of aliases required? if there are duplicate objs in the batch we should not insert dupes

gtarpenning (approved):
couple more small nits, and we really should benchmark the migration on the CH replica, maybe you can ask @bcsherma for 15min to test?

tests/trace_server/test_sqlite_trace_server.py
rgao-coreweave: Done in 8c48e8f8 — narrowed to `except sqlite3.Error: pass`. The exception we actually expect is `sqlite3.OperationalError("cannot rollback - no transaction is active")` (sqlite raises when the prior BEGIN already auto-rolled back); `sqlite3.Error` catches that class without hiding unrelated cleanup regressions. Added a comment explaining the case. Resolving.

tests/trace_server/test_clickhouse_trace_server_batched.py
rgao-coreweave: Done in 8c48e8f8 — tightened to `pytest.raises((NotFoundError, ObjectDeletedError))`. Those are the two real possibilities when 'latest' resolves to a tombstoned digest: CTE filters it out → `NotFoundError`, or `obj_read` returns it and rejects on `deleted_at` → `ObjectDeletedError`. Resolving.

weave/trace_server/migrations/031_backfill_latest_alias.up.sql
rgao-coreweave: Dropped the migration entirely in eb935df8. The hybrid `is_latest` projection in the metadata CTE already has a computed-rank fallback for any (project_id, object_id) absent from the alias CTE (added in c508900e95 `fix(weave): make is_latest hybrid (alias + computed fallback)`), so legacy objects resolve `'latest'` correctly without backfill — identical to the pre-PR behavior.

Removes:
- both `031_backfill_latest_alias.{up,down}.sql` files
- the 5 `test_migration_031_*` tests + `_MIGRATION_031_*` SQL constants (~420 LoC)

Keeps `test_legacy_no_alias_row_resolves_latest_via_computed_fallback`, which pins the exact behavior that makes the migration unnecessary.

Trade-off: legacy objects use the window-function rank in the CTE until their next re-publish writes an alias row organically (~+17% worst case per the benchmarks upthread). That seems worth it to avoid the full-table `INSERT...SELECT` on prod and the deploy timing question. Resolving.

rgao-coreweave:
> couple more small nits, and we really should benchmark the migration on the CH replica, maybe you can ask @bcsherma for 15min to test?

TLDR: Migration is dropped.

Dropped the migration entirely in https://github.com/wandb/weave/commit/eb935df8ac46233a0eb17ebeaeb855ef43adf7e4. The hybrid is_latest projection in the metadata CTE already has a computed-rank fallback for any (project_id, object_id) absent from the alias CTE (added in https://github.com/wandb/weave/commit/c508900e95889068fc82c8cca510cae26a2f49f3 fix(weave): make is_latest hybrid (alias + computed fallback)), so legacy objects resolve 'latest' correctly without backfill — identical to the pre-PR behavior.

weave/trace_server/clickhouse_trace_server_batched.py
rgao-coreweave: Done in dd5a4fc8 — extracted to `apply_tags_and_synth_latest_in_place(objs, tags_map, aliases_map)` in `trace_server_common.py`. Both backends already produced identically-keyed `(object_id, digest) → list[str]` maps, so the helper drops in cleanly. Read-skew guard and synth-when-is_latest=1 rule both moved with it; each call site collapses to three lines. Pure refactor — no behavior change. Resolving.

weave/trace_server/sqlite_trace_server.py
rgao-coreweave: Done in dd5a4fc8 — see paired thread on the CH side. Both call sites now invoke `apply_tags_and_synth_latest_in_place` from `trace_server_common.py`. Resolving.

tests/trace_server/test_clickhouse_trace_server_batched.py
rgao-coreweave: Decorated with `@pytest.mark.flaky(reruns=3, reruns_delay=0.2)` in dbe19f65 — matches the repo convention for other storage-order/eventual-consistency flakes. The failure mode here is the documented `argMax` storage-order tiebreaker on tied `now64(3)` created_at, not a real regression, so reruns absorb the noise without hiding genuine breaks. Resolving.

weave/trace_server/clickhouse_trace_server_batched.py
rgao-coreweave: The test-side symptom (the duplicate-object_id case being non-deterministic) is handled by `@pytest.mark.flaky(reruns=3, reruns_delay=0.2)` in dbe19f65 — see the paired thread on the test.

Holding off on dedup'ing `alias_rows` inside `obj_create_batch` itself: it's the OTel-only ingest path where the documented contract is "lose-versioning," and duplicate `(project_id, object_id)` entries in a single batch are themselves a sign of a caller-side issue we don't want to silently paper over here. Happy to revisit in a followup if we see this pattern in prod telemetry. Resolving.

PR #6521, opened 2026-04-01
perf(weave): eliminate CH reads from OTel ingest hot path

gtarpenning:
https://coreweave.atlassian.net/browse/WB-32667

Removes all ClickHouse reads from the OTel ingest hot path.

The op ref URI is fully deterministic (`OTEL_PLACEHOLDER_OP_DIGEST`), so we blind-insert and let `ReplacingMergeTree` deduplicate — no need to query CH first.

### What changed
- Replaced `_op_ref_cache` (TTLCache) with `_inserted_ops: set` — process-local, with a 50K sanity cap that clears the set on overflow (re-inserts are CH-idempotent, so dropping the whole set is harmless). The TTL was only needed when we cached resolved digests from CH reads. Now the digest is deterministic, so staleness isn't possible. Subsequent requests for the same op hit the set and skip the insert entirely.
- Deleted `_get_existing_ops_from_spans` (the CH read), `_cache_op_ref`, and `OP_REF_CACHE_*` constants
- Capped `obj_create_batch` async insert timeout at 200ms (`OTEL_ASYNC_INSERT_TIMEOUT_MS`) to avoid blocking on the default adaptive flush timer (up to 1000ms) — these inserts are tiny, and on the objects table, max of 10/s is probably fine. If not, we can increase the flush timer but then disable waiting for the flush ack, which will be insanely fast (at the cost of marginally less robust). 

### Safety
- **Duplicate inserts:** RMT + deduped view collapse them
- **Thread safety / restart:** TOCTOU harmless, empty set on restart just re-inserts
- **SDK op + OTel same name:** OTel inserts a placeholder version alongside the real one — uncommon, and display/filtering is digest-agnostic

### Tests
- `test_otel_export_reuses_existing_real_op` → `test_otel_export_uses_placeholder_even_with_existing_real_op` (asserts both versions coexist)


zbirenbaum:
Couple concerns here:
1. Explicitly not blocking on this but just wanted to raise that OTel failures are transient (except with some opt-in and perf intensive configuration by the user) If the timeout is succeeded users will most likely never know, their data will just be missing.
2. I believe the logic removed was used to check if the op that would be created for a span already exists (Op name and placeholder are used to calculate, no code to change so version never changes). This was an optimization over creating the op every time which was much more expensive. The PR description seems to imply something else.

PR #6542, opened 2026-04-02
fix(weave): dedupe concurrent op saves

gtarpenning:
## Summary

Fix a concurrency bug in op saving that was surfacing as a flaky evaluation teardown failure.

During `Evaluation.evaluate(...)`, parallel scorer execution could race on the same shared function scorer op. Multiple workers could observe that scorer op as unsaved and all call `_save_object_basic`, leading to redundant saves of the same op and occasionally destabilizing later background ref resolution.

This PR makes `_save_op` singleflight per client: one caller performs the save, concurrent callers wait and reuse the same ref.

Original failure ([GitHub Actions job](https://github.com/wandb/weave/actions/runs/23918560565/job/69758718183?pr=6523)):
```
ERROR    weave.trace.concurrent.futures:futures.py:192 Task failed: AttributeError: 'list' object has no attribute 'result'
```

## Root Cause

The flake was exposing a check-then-act race:

- evaluation rows are processed concurrently
- function scorers are shared module-scope `@weave.op` objects
- multiple threads could see `get_ref(op) is None` at the same time
- each thread could then independently call `_save_object_basic(op, ...)`

A deterministic reproducer added in this PR showed that exact behavior on both SQLite and ClickHouse. Before the fix, each scorer op was being saved 11 times in a single evaluation run. After the fix, each scorer is saved once.

That repeated-save race is the most plausible root cause of the rarer teardown error: the evaluation itself completed, but a later background task tripped over unstable async/ref state.

## Options Considered

1. Keep the test-only mitigation
- Disable autoflush around the flaky test and rely on reruns.
- Rejected because it reduces noise without fixing the race.

2. Redesign ref ownership entirely
- Move ref state off Python op objects and into a client-scoped cache.
- Better long term, but much larger than needed for this bug.

3. Add singleflight to op saves
- Make concurrent saves of the same Python op object collapse to one underlying save.
- Chosen because it directly addresses the observed race while preserving current semantics.

## Decision And Fix

`WeaveClient._save_op` now:

- returns an existing same-project ref immediately when present
- tracks in-flight saves per client, keyed by Python object identity
- lets one thread own the save while concurrent callers wait for the same future
- raises on conflicting concurrent names instead of silently picking one
- always clears the in-flight entry in `finally`

This keeps synchronization at the ownership layer for op saving rather than leaking it into evaluation.

## Why This Fix Works

The bug was not “evaluation is flaky”. The bug was “saving the same op object is not concurrency-safe”.

This change closes that race for ops:

- one caller performs the save
- all concurrent callers reuse the same in-flight result
- the shared op object ends up with one stable ref instead of several competing save attempts

This is a correctness fix, not just a flake-rate reduction.

## Tests

Added:
- an evaluation-level reproducer that forces concurrent scorer saves and asserts each scorer is saved once
- a focused `_save_op` concurrency test that asserts one underlying save is reused by concurrent callers

### DIFF
--- a/tests/trace/test_evaluate_oldstyle.py
+++ b/tests/trace/test_evaluate_oldstyle.py
@@ -1,4 +1,5 @@
 import asyncio
+import threading
 
 import pytest
 
@@ -6,6 +7,8 @@ import weave
 from tests.conftest import LATENCY_TOL
 from weave import Dataset, Evaluation, Model
 from weave.scorers import MultiTaskBinaryClassificationF1
+from weave.trace.ref_util import get_ref, remove_ref
+from weave.trace.weave_client import WeaveClient
 
 dataset_rows = [{"input": "1 + 2", "target": 3}, {"input": "2**4", "target": 15}]
 dataset = Dataset(rows=dataset_rows)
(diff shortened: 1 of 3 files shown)


tests/trace/test_evaluate_oldstyle.py
```diff
@@ -100,12 +103,19 @@ def test_evaluate_rows_only(client):
...
 def test_evaluate_both_styles(client):
+    client.set_autoflush(False)
     evaluation = Evaluation(
         dataset=dataset_rows,
         scorers=[score_oldstyle, score_newstyle],
     )
     model = EvalModel()
-    result = asyncio.run(evaluation.evaluate(model))
+    try:
+        result = asyncio.run(evaluation.evaluate(model))
```
andrewtruong: This test should use `pytest.mark.asyncio` and this should be an `await`

tests/trace/test_evaluate_oldstyle.py
```diff
@@ -1,11 +1,14 @@
...
 
 import weave
 from tests.conftest import LATENCY_TOL
 from weave import Dataset, Evaluation, Model
 from weave.scorers import MultiTaskBinaryClassificationF1
+from weave.trace.ref_util import get_ref, remove_ref
+from weave.trace.weave_client import WeaveClient
 
 dataset_rows = [{"input": "1 + 2", "target": 3}, {"input": "2**4", "target": 15}]
 dataset = Dataset(rows=dataset_rows)
```
andrewtruong: This should be a fixture

tests/trace/test_evaluate_oldstyle.py
```diff
@@ -114,6 +124,68 @@ def test_evaluate_both_styles(client):
     }
 
 
+@pytest.mark.disable_logging_error_check
+def test_evaluate_both_styles_saves_each_function_scorer_once(client, monkeypatch):
+    expanded_rows = [dict(row) for _ in range(5) for row in dataset_rows]
```
andrewtruong: This dataset_rows appeared out of nowhere.  The dataset should be a fixture!

tests/trace/test_evaluate_oldstyle.py
```diff
@@ -114,6 +124,68 @@ def test_evaluate_both_styles(client):
...
+    evaluation = Evaluation(
+        dataset=expanded_rows,
+        scorers=[score_oldstyle, score_newstyle],
```
andrewtruong, line 180: same here

tests/trace/test_evaluate_oldstyle.py
```diff
@@ -114,6 +124,68 @@ def test_evaluate_both_styles(client):
...
+    }
+    save_counts_lock = threading.Lock()
+    concurrent_first_save_gates = {
+        id(score_oldstyle): {"waiting": 0, "release": threading.Event()},
+        id(score_newstyle): {"waiting": 0, "release": threading.Event()},
+    }
+    gate_lock = threading.Lock()
+    original_save_object_basic = WeaveClient._save_object_basic.__get__(
+        client, type(client)
+    )
```
andrewtruong: What is this?

tests/trace/test_evaluate_oldstyle.py
```diff
@@ -114,6 +124,68 @@ def test_evaluate_both_styles(client):
...
+        if val is score_oldstyle or val is score_newstyle:
+            if get_ref(val) is None:
+                gate_state = concurrent_first_save_gates[id(val)]
+                with gate_lock:
+                    gate_state["waiting"] += 1
+                    if gate_state["waiting"] == 2:
+                        gate_state["release"].set()
+                gate_state["release"].wait(timeout=0.5)
+            with save_counts_lock:
+                save_counts[val.name] += 1
```
andrewtruong: Can you annotate what you're trying to do here?

gtarpenning: yeah, also pulled it out to make it more clear

tests/trace/test_evaluate_oldstyle.py
gtarpenning: done

tests/trace/test_evaluate_oldstyle.py
gtarpenning: yeah, see those upstream changes.

PR #6569, opened 2026-04-09
feat(weave): add api_key and base_url params to weave.init() with client-scoped credential isolation

rgao-coreweave:
## Summary

Resolves [WB-33023](https://coreweave.atlassian.net/browse/WB-33023)

Adds optional `api_key`, `base_url`, and `trace_server_url` keyword arguments to `weave.init()`, enabling programmatic authentication without environment variables.

### Motivation

External partners need to set the API key and base URL at runtime to support multi-tenant scenarios where multiple auth contexts coexist in the same process. The current workaround of mutating `os.environ` before calling `weave.init()` is brittle and error-prone.

### Design

Credentials are **eagerly resolved** at `WeaveClient` construction time and stored as immutable instance fields (`_api_key`, `_base_url`). When explicit params are provided they're used directly; otherwise the current env/netrc values are snapshotted once. This means:

- **No module-level globals** — removed `_explicit_base_url` from `env.py` and `_explicit_api_key` from `context.py`
- **No cross-client leakage** — each client carries its own credentials
- **Thread-safe by construction** — immutable fields readable from any `FutureExecutor` worker thread without synchronization
- **No lazy fallback** — downstream consumers read from the client, never from env

| Parameter | Overrides | Purpose |
|---|---|---|
| `api_key` | `WANDB_API_KEY` / netrc / `wandb.login()` | Authentication for trace server and entity resolution |
| `base_url` | `WANDB_BASE_URL` | W&B platform API URL; `trace_server_url` auto-derived from this |
| `trace_server_url` | `WF_TRACE_SERVER_URL` | Trace server URL — only needed when on a different host |

### Usage

```python
# Existing env var approach (unchanged)
weave.init("my-project")

# Direct params — trace_server_url auto-derived from base_url
weave.init("my-project", api_key=key, base_url="https://api.wandb.ai")

# Multi-client isolation
client_x = weave.init("proj-x", api_key=key_x, base_url=url_x)
client_y = weave.init("proj-y", api_key=key_y, base_url=url_y)
# client_x retains x's credentials — no cross-contamination
```

### Changes

**Credential storage & propagation (core):**
- `weave/trace/weave_client.py` — `WeaveClient` stores `_api_key`/`_base_url` (eagerly resolved); threads them to `ensure_project_exists`, URL generation, `create_call`, `get_calls`/`get_call`, `Call.children()`, WAL callbacks, `Chat`/`Completions`/`InferenceModels`
- `weave/trace/weave_init.py` — `init_weave()` resolves creds per call (no globals), derives `trace_server_url` from `base_url`, passes creds to all helpers
- `weave/wandb_interface/context.py` — Removed `_explicit_api_key` global; `get_wandb_api_context()` only reads env/netrc
- `weave/trace/env.py` — Removed `_explicit_base_url` global; `wandb_base_url()` only reads env/settings

**Credential consumers (all read from client instance):**
- `weave/compat/wandb/wandb_thin/internal_api.py` — `Api`/`ApiAsync` accept `api_key`/`base_url` in constructor
- `weave/trace/urls.py` — All URL helpers accept optional `base_url`
- `weave/trace/call.py` — `Call` carries `_base_url`; `make_client_call`/`_make_calls_iterator` thread it through
- `weave/chat/completions.py`, `weave/chat/inference_models.py` — Read from `self._client._api_key`
- `weave/wandb_interface/project_creator.py` — `ensure_project_exists()` accepts and passes `api_key`/`base_url`
- `weave/trace/init_message.py` — Uses explicit `base_url` for init message URL
- `weave/trace/api.py` — `init()` forwards new params; `publish()` passes client base_url; `finish()` simplified

**Server interface:**
- `weave/trace_server/service_interface.py` — `ensure_project_exists` accepts optional `api_key`/`base_url`
- `weave/trace_server_bindings/remote_http_trace_server.py` — Passes creds to `project_creator`
- `weave/trace_server_bindings/stainless_remote_http_trace_server.py` — Same
- `weave/trace_server/fake_service.py` — Updated signature

### Unit tests (`tests/trace/test_weave_init_auth.py` — 34 tests)

**Init flow:**
- Explicit `api_key` skips env/netrc auth
- Credentials stored on client instance (eagerly resolved from env when not explicit)
- Client reuse when no credential params; forced re-init when creds change
- `trace_server_url` derived from `base_url` (custom host, default host, explicit override)
- Version check uses effective trace server URL

**Credential passthrough (verified with mock assertions):**
- `ensure_project_exists` receives correct `api_key`/`base_url`
- `get_username` and `get_entity_project_from_project_name` receive creds
- `print_init_message` uses correct `base_url`
- `Completions`/`InferenceModels` prefer client key
- `Api` constructor stores creds

**Multi-client isolation:**
- Two clients retain independent credentials
- Module-level env readers unaffected by explicit init
- `Call.ui_url` uses client's `base_url` (with env fallback)
- Finishing one client doesn't affect another
- `create_call` stamps `_base_url` on Call

**Thread safety & callbacks:**
- `FutureExecutor` closure reads correct client credentials
- WAL callback passes `self._base_url` to `redirect_call`

**Other tests updated:**
- `test_client_trace.py`, `test_weave_client.py` — Updated expected `Call` objects to include `_base_url`
- `test_client_server_caching.py` — Made env-independent for `base_url`

### Manual validation (`rgao/weave_init_api_key_demo.ipynb`)
Tested against live prod and QA: env vars, multi-tenant switching, mixed env+params with auto-derivation, and client-scoped isolation.

[WB-33023]: https://coreweave.atlassian.net/browse/WB-33023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/test_weave_init_auth.py
+++ b/tests/trace/test_weave_init_auth.py
@@ -287,8 +287,8 @@ def test_init_weave_reuse_and_reinit(mock_wandb_api):
 
 
 def test_init_weave_env_fallback_paths(mock_wandb_api, monkeypatch):
-    """When api_key is omitted it reads from env; base_url stays None
-    (env default is used at point of use by URL helpers).
+    """When api_key is omitted it reads from env; base_url eagerly
+    resolves from WANDB_BASE_URL.
     """
     mock_wandb_api.default_entity_name.return_value = "test-entity"
     mock_server = _make_mock_server()
@@ -306,9 +306,7 @@ def test_init_weave_env_fallback_paths(mock_wandb_api, monkeypatch):
         client = weave_init.init_weave("test-project")
         mock_ctx.assert_called_once()
         assert client._api_key == "env-key"
-        # base_url is None when not explicitly provided; env default is
-        # resolved at point of use by URL helpers.
-        assert client._base_url is None
+        assert client._base_url == "https://env.example.com"
         client.finish()
 
 
@@ -515,18 +513,18 @@ def test_api_eager_credential_resolution(monkeypatch):
     assert api_default._base_url == "https://env.example.com"
 
 
-def test_client_without_explicit_key_leaves_api_key_none():
-    """WeaveClient with api_key=None stores None; Completions and
-    InferenceModels fall back to env at point of use.
+def test_client_resolves_env_key_for_completions_and_models(monkeypatch):
+    """WeaveClient eagerly resolves api_key from env so Completions and
+    InferenceModels see it without their own fallback logic.
     """
+    monkeypatch.setenv("WANDB_API_KEY", "env-key")
     mock_server = _make_mock_server()
     client = WeaveClient(
         "entity", "project", mock_server, ensure_project_exists=False, api_key=None
     )
 
-    assert client._api_key is None
-    assert Completions(client)._client._api_key is None
-    assert InferenceModels(client)._client._api_key is None
+    assert Completions(client)._client._api_key == "env-key"
+    assert InferenceModels(client)._client._api_key == "env-key"
     client.finish()
 
 
--- a/tests/trace/test_weave_init_auth.py
+++ b/tests/trace/test_weave_init_auth.py
@@ -931,9 +931,7 @@ def test_completions_falls_back_to_env_when_no_client_key():
         "weave.chat.completions.get_wandb_api_context", return_value="env-key"
     ) as mock_ctx:
         api_key = (
-            comp._client._api_key
-            if comp._client._api_key is not None
-            else mock_ctx()
+            comp._client._api_key if comp._client._api_key is not None else mock_ctx()
         )
         assert api_key == "env-key"
         mock_ctx.assert_called_once()
@@ -969,17 +967,19 @@ def test_inference_models_falls_back_to_env_when_no_client_key():
 # --- Coverage: project_creator with explicit credentials ---
 
 
-def test_project_creator_threads_credentials():
-    """ensure_project_exists should pass api_key/base_url to _ensure_project_exists."""
-    with patch(
-        "weave.wandb_interface.project_creator._ensure_project_exists",
-        return_value={"project_name": "proj"},
-    ) as mock_inner:
-        from weave.wandb_interface.project_creator import ensure_project_exists
+def test_project_creator_passes_credentials_to_api():
+    """_ensure_project_exists should construct wandb.Api with api_key and base_url."""
+    from weave.wandb_interface.project_creator import _ensure_project_exists
 
-        ensure_project_exists(
-            "entity", "project", api_key="key", base_url="https://example.com"
+    mock_api = MagicMock()
+    mock_api.project.return_value = {"project": {"name": "proj"}}
+
+    with patch("weave.wandb_interface.project_creator.wandb") as mock_wandb:
+        mock_wandb.Api.return_value = mock_api
+        result = _ensure_project_exists(
+            "entity", "proj", api_key="key", base_url="https://example.com"
         )
-        mock_inner.assert_called_once_with(
-            "entity", "project", api_key="key", base_url="https://example.com"
+        mock_wandb.Api.assert_called_once_with(
+            api_key="key", base_url="https://example.com"
         )
+        assert result == {"project_name": "proj"}
--- a/tests/trace/test_weave_init_auth.py
+++ b/tests/trace/test_weave_init_auth.py
@@ -983,3 +983,41 @@ def test_project_creator_passes_credentials_to_api():
             api_key="key", base_url="https://example.com"
         )
         assert result == {"project_name": "proj"}
+
+
+# --- Coverage: Call.children() passes base_url ---
+
+
+def test_call_children_passes_base_url():
+    """Call.children() should pass client._base_url to _make_calls_iterator."""
+    from weave.trace.call import Call
+    from weave.trace.weave_client import WeaveClient
+
+    mock_server = _make_mock_server()
+    client = WeaveClient(
+        "entity",
+        "project",
+        mock_server,
+        ensure_project_exists=False,
+        base_url="https://children.example.com",
+    )
+    weave_client_context.set_weave_client_global(client)
+
+    call = Call(
+        _op_name="test-op",
+        project_id="entity/project",
+        trace_id="trace-123",
+        parent_id=None,
+        inputs={},
+        id="call-123",
+    )
+
+    with patch("weave.trace.call._make_calls_iterator") as mock_iter:
+        mock_iter.return_value = iter([])
+        list(call.children())
+        mock_iter.assert_called_once()
+        _, kwargs = mock_iter.call_args
+        assert kwargs.get("base_url") == "https://children.example.com"
+
+    client.finish()
+    weave_client_context.set_weave_client_global(None)
--- a/tests/trace/test_weave_init_auth.py
+++ b/tests/trace/test_weave_init_auth.py
@@ -287,8 +287,8 @@ def test_init_weave_reuse_and_reinit(mock_wandb_api):
 
 
 def test_init_weave_env_fallback_paths(mock_wandb_api, monkeypatch):
-    """When api_key is omitted it reads from env; when base_url is omitted it
-    eagerly resolves from WANDB_BASE_URL.
+    """When api_key is omitted it reads from env; base_url stays None
+    (env default is used at point of use by URL helpers).
     """
     mock_wandb_api.default_entity_name.return_value = "test-entity"
     mock_server = _make_mock_server()
@@ -306,7 +306,9 @@ def test_init_weave_env_fallback_paths(mock_wandb_api, monkeypatch):
         client = weave_init.init_weave("test-project")
         mock_ctx.assert_called_once()
         assert client._api_key == "env-key"
-        assert client._base_url == "https://env.example.com"
+        # base_url is None when not explicitly provided; env default is
+        # resolved at point of use by URL helpers.
+        assert client._base_url is None
         client.finish()
 
 
@@ -513,18 +515,18 @@ def test_api_eager_credential_resolution(monkeypatch):
     assert api_default._base_url == "https://env.example.com"
 
 
-def test_client_resolves_env_key_for_completions_and_models(monkeypatch):
-    """WeaveClient eagerly resolves api_key from env so Completions and
-    InferenceModels see it without their own fallback logic.
+def test_client_without_explicit_key_leaves_api_key_none():
+    """WeaveClient with api_key=None stores None; Completions and
+    InferenceModels fall back to env at point of use.
     """
-    monkeypatch.setenv("WANDB_API_KEY", "env-key")
     mock_server = _make_mock_server()
     client = WeaveClient(
         "entity", "project", mock_server, ensure_project_exists=False, api_key=None
     )
 
-    assert Completions(client)._client._api_key == "env-key"
-    assert InferenceModels(client)._client._api_key == "env-key"
+    assert client._api_key is None
+    assert Completions(client)._client._api_key is None
+    assert InferenceModels(client)._client._api_key is None
     client.finish()
 
 
--- a/tests/trace/test_weave_init_auth.py
+++ b/tests/trace/test_weave_init_auth.py
@@ -259,31 +259,52 @@ def test_init_weave_credential_storage_and_env_skipping(mock_wandb_api):
 
 
 def test_init_weave_reuse_and_reinit(mock_wandb_api):
-    """Same project without credential params reuses client; passing any
-    credential param forces re-init.
+    """Reuse client when identical explicit credentials are passed;
+    recreate when credentials differ or are env-derived (None).
     """
     mock_wandb_api.default_entity_name.return_value = "test-entity"
     mock_server = _make_mock_server()
 
-    with _init_patches(mock_server):
+    with (
+        _init_patches(mock_server),
+        patch(
+            "weave.trace.weave_init.get_wandb_api_context",
+            return_value="env-key",
+        ),
+    ):
         client_a = weave_init.init_weave("test-project", api_key="key")
-        assert weave_init.init_weave("test-project") is client_a  # reuse
+        # Same explicit credentials → reuse
+        assert weave_init.init_weave("test-project", api_key="key") is client_a
 
-        client_b = weave_init.init_weave("test-project", api_key="new-key")
+        # No explicit api_key (env-derived) → recreate to pick up env changes
+        client_b = weave_init.init_weave("test-project")
         assert client_b is not client_a
 
-        client_c = weave_init.init_weave(
-            "test-project", api_key="new-key", base_url="https://new.example.com"
-        )
+        # Different explicit key → recreate
+        client_c = weave_init.init_weave("test-project", api_key="new-key")
         assert client_c is not client_b
 
+        # Same key but added base_url → recreate
         client_d = weave_init.init_weave(
+            "test-project", api_key="new-key", base_url="https://new.example.com"
+        )
+        assert client_d is not client_c
+        # Same key + same base_url → reuse
+        assert (
+            weave_init.init_weave(
+                "test-project", api_key="new-key", base_url="https://new.example.com"
+            )
+            is client_d
+        )
+
+        # Added trace_server_url → recreate
+        client_e = weave_init.init_weave(
             "test-project",
             api_key="new-key",
             trace_server_url="https://trace.new.example.com",
         )
-        assert client_d is not client_c
-        client_d.finish()
+        assert client_e is not client_d
+        client_e.finish()
 
 
 def test_init_weave_env_fallback_paths(mock_wandb_api, monkeypatch):
--- a/tests/trace/test_weave_init_auth.py
+++ b/tests/trace/test_weave_init_auth.py
@@ -16,16 +16,19 @@ from weave.trace.weave_init import (
 @pytest.fixture(autouse=True)
 def reset_weave_client():
     """Reset the global weave client state before each test."""
-    from weave.trace.env import set_wandb_base_url
-    from weave.wandb_interface.context import set_wandb_api_context
-
     weave_client_context.set_weave_client_global(None)
-    set_wandb_api_context(None)
-    set_wandb_base_url(None)
     yield
     weave_client_context.set_weave_client_global(None)
-    set_wandb_api_context(None)
-    set_wandb_base_url(None)
+
+
+@pytest.fixture(autouse=True)
+def mock_project_creator():
+    """Mock project_creator.ensure_project_exists for all tests that create clients."""
+    with patch(
+        "weave.wandb_interface.project_creator.ensure_project_exists",
+        return_value={"project_name": "test-project"},
+    ):
+        yield
 
 
 @dataclass
@@ -239,16 +242,12 @@ def test_weave_init_passes_all_params(mock_wandb_api):
         weave_client_context.set_weave_client_global(None)
 
 
-def test_init_weave_with_explicit_params_sets_context(mock_wandb_api):
-    """When api_key and base_url are provided, they should be set in context so
-    downstream code (e.g. entity resolution, trace server URL derivation) can
-    use them without env vars.
+def test_init_weave_stores_creds_on_client(mock_wandb_api):
+    """When api_key and base_url are provided, they should be stored on the
+    client instance (not in module-level globals).
     """
     mock_wandb_api.default_entity_name.return_value = "test-entity"
     mock_server = _make_mock_server()
-
-    import weave.trace.env as env_module
-    import weave.wandb_interface.context as context_module
 
     with (
         patch("weave.trace.weave_init.init_weave_get_server", return_value=mock_server),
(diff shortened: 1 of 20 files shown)


rgao-coreweave:
* **#6569** <a href="https://app.graphite.com/github/pr/wandb/weave/6569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6569?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace/env.py
```diff
@@ -12,6 +12,12 @@
 
 logger = logging.getLogger(__name__)
 
+# Module-level override for runtime configuration via weave.init() params.
+# When set, overrides env-based lookup so the SDK can be configured without
+# touching environment variables. Uses a plain global (not contextvars) because
+# this value must be visible across threads (e.g. call link generation).
+_explicit_base_url: str | None = None
```
tssweeney: Are you sure you want to keep this in a global? It seems that we should be moving away from globals in general. for example what happens in this situation:
```
client_x = weave.init(base_url=x)
client_y = weave.init(base_url=y)

client_x.do_thing()
```

the y variant will have overloaded the global. I don't think you want to explicitly modify the environment itself, but rather modify the reader to hold private state

weave/trace/weave_init.py
```diff
@@ -105,21 +112,36 @@ def init_weave(
...
         ):
             return current_client
         else:
             # Flush any pending calls before switching to a new project
             current_client.finish()
             weave_client_context.set_weave_client_global(None)
 
-    from weave.wandb_interface.context import get_wandb_api_context
+    # Set explicit base_url in context so downstream code (e.g. entity resolution
+    # via wandb.Api(), trace server URL derivation) uses it instead of env vars.
```
tssweeney: this relates to my comment above. It seems like a leaky abstraction to have other global constants effected by this

weave/trace/weave_init.py
```diff
@@ -105,21 +112,36 @@ def init_weave(
...
+    if base_url is not None:
+        env.set_wandb_base_url(base_url)
 
-    api_key = get_wandb_api_context()
     if api_key is None:
-        url = wandb.app_url(env.wandb_base_url())
-        logger.info("Please login to Weights & Biases (%s) to continue...", url)
-        wandb.login(anonymous="never", force=True, referrer="weave")  # type: ignore
+        from weave.wandb_interface.context import get_wandb_api_context
```
tssweeney: import loop?

weave/wandb_interface/context.py
```diff
@@ -2,7 +2,24 @@
 
 from weave.trace.env import weave_wandb_api_key
 
+# When set, overrides env-based API key lookup. This allows weave.init(api_key=...)
+# to propagate the explicit key to all downstream code (e.g. entity resolution)
+# without touching environment variables. Uses a plain global (not contextvars)
+# because this value must be visible across threads.
+_explicit_api_key: str | None = None
```
tssweeney: oofff - similar to the comment at the top - this design make our system less thread safe and less capable for handling multiple clients

weave/trace/env.py
rgao-coreweave: Good point!

The real problem with "store on client": The readers like urls.redirect_call(), wandb.Api(), and Call.ui_url don't currently have a reference to the client. Call is a standalone dataclass — it doesn't hold a client reference. And wandb.Api() is a stateless utility.

What would actually work: Rather than putting state on the client, the cleaner fix is to make wandb.Api accept credentials at construction time so callers can pass them through.

rgao-coreweave: Addressed in c4f74c6. `_explicit_base_url` and `set_wandb_base_url()` removed entirely. Credentials now live as private fields on each `WeaveClient` instance (`_api_key`, `_base_url`) and are threaded through to all downstream consumers (Api, urls, Call, Chat, project_creator). 42 tests cover isolation, thread safety, and all credential read points.

weave/trace/weave_init.py
rgao-coreweave: Addressed in c4f74c6. `init_weave()` no longer calls `set_wandb_base_url()` or `set_wandb_api_context()`. Credentials are passed directly to `WeaveClient` constructor and threaded to helpers (`get_username`, `get_entity_project_from_project_name`, `ensure_project_exists`).

weave/trace/weave_init.py
rgao-coreweave: Addressed in c4f74c6. The local import was originally to avoid circular deps with `set_wandb_api_context`. Now that `get_wandb_api_context` is just a thin wrapper around `weave_wandb_api_key()` (env/netrc reader), the circular concern is gone. Will move to top-level import in a follow-up.

weave/wandb_interface/context.py
rgao-coreweave: Addressed in c4f74c6. `_explicit_api_key` and `set_wandb_api_context()` removed. `get_wandb_api_context()` now just reads env/netrc. Each `WeaveClient` holds its own `_api_key` — no globals, fully thread-safe for multiple clients.

weave/trace/weave_init.py
```diff
@@ -105,21 +112,36 @@ def init_weave(
...
             and current_client.ensure_project_exists == ensure_project_exists
+            and api_key is None
+            and base_url is None
+            and trace_server_url is None
         ):
             return current_client
         else:
             # Flush any pending calls before switching to a new project
             current_client.finish()
             weave_client_context.set_weave_client_global(None)
```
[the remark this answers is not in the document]
rgao-coreweave: Addressed in c4f74c6. Module-level overrides no longer exist — there is nothing to leak. Each `WeaveClient` carries its own credentials as instance fields, so re-initialization creates a fresh client with its own state.

weave/trace/weave_client.py
```diff
@@ -368,9 +372,15 @@ def __init__(
         self.ensure_project_exists = ensure_project_exists
 
         if ensure_project_exists:
-            resp = self.server.ensure_project_exists(entity, project)
+            from weave.wandb_interface.project_creator import (
+                ensure_project_exists as _ensure_project,
+            )
```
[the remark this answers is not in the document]
rgao-coreweave: Valid point. Will move the import to the top of the file in a follow-up — no circular dependency concern here.

weave/trace/weave_init.py
```diff
@@ -105,30 +116,39 @@ def init_weave(
...
+    entity_name, project_name = get_entity_project_from_project_name(
+        project_name, api_key=api_key, base_url=base_url
+    )
     wb_run_context = get_global_wb_run_context()
     if wb_run_context:
         wandb_run_id = f"{entity_name}/{project_name}/{wb_run_context.run_id}"
         check_wandb_run_matches(wandb_run_id, entity_name, project_name)
 
-    remote_server = init_weave_get_server(api_key)
+    remote_server = init_weave_get_server(api_key, trace_server_url=trace_server_url)
```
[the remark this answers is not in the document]
rgao-coreweave: Valid point. When `base_url` is provided but `trace_server_url` is not, we should derive the trace server URL from `base_url` (matching the documented contract). Will address in a follow-up.

weave/trace/urls.py
chance-wnb, line 1: Nit: My guess: there used to be a time that having those utility functions made sense, because the `env.wandb_base_url()` can always retrieved from a global/neutral location. But now weave_client holds the contextual information of these base_url, trace_server_url, then does it make sense to put those function to the weave_client itself?

weave/trace/call.py
```diff
@@ -80,6 +80,7 @@ class Call:
     # These are the live children during logging
     _children: list[Call] = dataclasses.field(default_factory=list)
     _feedback: RefFeedbackQuery | None = None
+    _base_url: str | None = None
```
chance-wnb, line 83: make_client_call() might not have this info passed.

When a user calls client.get_calls() or client.get_call(call_id), the returned Call objects won't have _base_url set, so call.ui_url will use the env-based default instead of the client's explicit base_url. This is inconsistent with calls created via client.create_call() (which do get _base_url at weave_client.py:835).

A fix would require threading _base_url through _make_calls_iterator and make_client_call, or having the client set _base_url on calls after they're constructed. This is also related to my comment about moving URL functions onto WeaveClient itself — if URL generation lived on the client, this gap wouldn't exist.

chance-wnb (commented):
Hi Rick, sorry I am kinda in a rush here, cause I need to take care of another emergent fix. I have to pause my code review here:

Could you help check whether your implementation might break the work of WAL (recent work from Andrew) ? I estimate that this might be related, since WAL mechanism might need to decide which server to send traffic to. Sorry I couldn't get a chance to check that myself.

weave/trace/call.py
rgao-coreweave: Good observation. I considered this but kept them as standalone functions because they're used from multiple callsites that don't hold a client reference (e.g. `Call.ui_url`, `init_message.print_init_message`). Moving them into `WeaveClient` would require either giving `Call` a back-reference to the client or duplicating logic. The optional `base_url` param approach keeps them stateless and avoids coupling. Happy to revisit if we refactor `Call` to carry a client reference in the future.

weave/trace/urls.py
rgao-coreweave: Agree this is a natural next step. For now keeping them as standalone functions with optional base_url param since they're used from callsites without a client ref. Will revisit when we refactor Call to carry a client reference.

weave/trace/call.py
rgao-coreweave: Fixed in 93cd042. `make_client_call` and `_make_calls_iterator` now accept `base_url` param. `WeaveClient.get_calls()` and `get_call()` pass `self._base_url` through, so retrieved calls generate correct URLs.

tests/trace/test_weave_init_auth.py
```diff
@@ -138,3 +150,679 @@ def test_get_entity_project_from_project_name_exceptions(case: ExceptionCase):
...
+        mock_get_server.assert_called_with(
+            "key", trace_server_url="https://explicit-trace.example.com"
+        )
+        weave_client_context.set_weave_client_global(None)
+
+
+def _make_mock_server():
+    from weave.trace_server_bindings.remote_http_trace_server import (
+        RemoteHTTPTraceServer,
+    )
```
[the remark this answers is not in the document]
rgao-coreweave: Fixed in 93cd042. All function-local imports moved to module level.

tests/trace/test_weave_init_auth.py
```diff
@@ -138,3 +150,679 @@ def test_get_entity_project_from_project_name_exceptions(case: ExceptionCase):
...
+            base_url="https://explicit.example.com",
+        )
+        weave_client_context.set_weave_client_global(None)
+
+
+def test_completions_uses_client_api_key():
+    """Completions and InferenceModels should prefer client._api_key over env."""
+    from weave.chat.completions import Completions
+    from weave.chat.inference_models import InferenceModels
+    from weave.trace.weave_client import WeaveClient
```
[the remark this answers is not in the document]
rgao-coreweave: Fixed in 93cd042. All function-local imports moved to module level.

tests/trace/test_weave_init_auth.py
```diff
@@ -138,3 +150,679 @@ def test_get_entity_project_from_project_name_exceptions(case: ExceptionCase):
...
+    assert client_b._base_url == "https://b.example.com"
+
+    client_b.finish()
+    weave_client_context.set_weave_client_global(None)
+
+
+@pytest.mark.disable_logging_error_check
+def test_client_create_call_sets_base_url(mock_wandb_api):
+    """WeaveClient.create_call should set _base_url on the Call object."""
+    from weave.trace.weave_client import WeaveClient
```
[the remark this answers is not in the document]
rgao-coreweave: Fixed in 93cd042. All function-local imports moved to module level.

tests/trace/test_weave_init_auth.py
```diff
@@ -138,3 +150,679 @@ def test_get_entity_project_from_project_name_exceptions(case: ExceptionCase):
...
+
+        client_y.finish()
+        weave_client_context.set_weave_client_global(None)
+
+
+def test_call_ui_url_uses_client_base_url(mock_wandb_api):
+    """Call.ui_url should use the base_url from the client that created it,
+    not a global or the default.
+    """
+    from weave.trace.call import Call
```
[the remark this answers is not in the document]
rgao-coreweave: Fixed in 93cd042. All function-local imports moved to module level.

weave/trace/call.py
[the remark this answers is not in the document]
rgao-coreweave: Good catch. Fixed in 15225ce — `Call.children()` now passes `client._base_url` to `_make_calls_iterator`, consistent with `get_calls()` and `get_call()`. Added `test_call_children_passes_base_url` to verify.

rgao-coreweave:
> Hi Rick, sorry I am kinda in a rush here, cause I need to take care of another emergent fix. I have to pause my code review here:
>
> Could you help check whether your implementation might break the work of WAL (recent work from Andrew) ? I estimate that this might be related, since WAL mechanism might need to decide which server to send traffic to. Sorry I couldn't get a chance to check that myself.

​Good call on the WAL.

The WAL system works with my branch because it receives the already-authenticated server object — it doesn't need to handle credentials independently.  `ensure_project_exists()` changes add optional params are backwards-compatible and the WAL sender never calls that method anyway.

rgao-coreweave:
### Code review

Found 1 issue:

1. Function-local imports in `tests/trace/test_weave_init_auth.py` violate CLAUDE.md import rule. CLAUDE.md states: "IMPORTANT: Always place imports at the top of Python files. All imports must be at the module level (top of the file), not inside functions or methods." and "Imports inside functions are not caught by linting. Agents must self-enforce this rule." There are 11 function-local imports across 7 newly-added test functions (lines 877, 879, 900, 909, 921-922, 944-945, 972, 993-994). These should be moved to the module-level imports at the top of the file.

https://github.com/wandb/weave/blob/15225ce5f17bc67d6e9e2db4c95664da3ab68917/tests/trace/test_weave_init_auth.py#L876-L994

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with :+1:. Otherwise, react with :-1:.</sub>

weave/chat/inference_models.py
```diff
@@ -26,7 +26,11 @@ def __init__(self, client: "WeaveClient"):
         self._client = client
 
     def list(self) -> ModelsResponse:
-        api_key = get_wandb_api_context()
+        api_key = (
+            self._client._api_key
+            if self._client._api_key is not None
```
tssweeney: after client is constructed, won't it always have api key?

weave/compat/wandb/wandb_thin/internal_api.py
```diff
@@ -20,14 +20,28 @@
 
 
 class Api:
+    def __init__(
+        self,
+        api_key: str | None = None,
+        base_url: str | None = None,
+    ) -> None:
+        self._api_key = api_key
```
tssweeney: seems like your api key resolution should happen here

tssweeney: effectively making the api key always non-null

weave/chat/inference_models.py
rgao-coreweave: No, it won't always have an api key. When users rely on env vars / netrc (no explicit api_key param), `_client._api_key` is None — the auth is set directly on the server via set_auth(). The fallback to get_wandb_api_context() covers that path.

weave/compat/wandb/wandb_thin/internal_api.py
rgao-coreweave: It's a design preference. Eager resolution at construction means Api becomes a snapshot of credentials at creation time. The current lazy approach means it always reads the latest env var. For multi-client isolation, eager is better.

weave/chat/inference_models.py
rgao-coreweave: Fixed. `WeaveClient.__init__` now eagerly resolves `api_key` from env/netrc when not explicitly provided. Downstream consumers (`Completions`, `InferenceModels`, `Api`) use the resolved value directly — no fallback logic.

weave/compat/wandb/wandb_thin/internal_api.py
rgao-coreweave: Done. `Api.__init__` and `ApiAsync.__init__` now eagerly resolve both `api_key` and `base_url` at construction time. `query()` uses the resolved values directly — no lazy env reads.

rgao-coreweave:
**Question:** Is the client reuse check in `init_weave()` intentionally strict about explicit credentials?

Currently, if a user calls `weave.init("proj", api_key="key1")` twice in a row (e.g., at module scope or in a loop), the second call tears down and recreates the client because the reuse condition requires all credential params to be `None`:

```python
if (
    current_client.project == project_name
    and current_client.ensure_project_exists == ensure_project_exists
    and api_key is None
    and base_url is None
    and trace_server_url is None
):
    return current_client
```

This means identical credentials still trigger a full teardown (`finish()` → flush pending calls, close connections, clear global state) and re-init. Would it make sense to compare the provided values against the current client's stored values so identical credentials allow reuse?

```python
and (api_key is None or api_key == current_client._api_key)
and (base_url is None or base_url == current_client._base_url)
and (trace_server_url is None or trace_server_url == current_client.server.trace_server_url)
```

rgao-coreweave:
> **Question:** Is the client reuse check in `init_weave()` intentionally strict about explicit credentials?
> 
> Currently, if a user calls `weave.init("proj", api_key="key1")` twice in a row (e.g., at module scope or in a loop), the second call tears down and recreates the client because the reuse condition requires all credential params to be `None`:
> 
> ```python
> if (
>     current_client.project == project_name
>     and current_client.ensure_project_exists == ensure_project_exists
>     and api_key is None
>     and base_url is None
>     and trace_server_url is None
> ):
>     return current_client
> ```
> 
> This means identical credentials still trigger a full teardown (`finish()` → flush pending calls, close connections, clear global state) and re-init. Would it make sense to compare the provided values against the current client's stored values so identical credentials allow reuse?
> 
> ```python
> and (api_key is None or api_key == current_client._api_key)
> and (base_url is None or base_url == current_client._base_url)
> and (trace_server_url is None or trace_server_url == current_client.server.trace_server_url)
> ```

 weave/trace/weave_init.py:
  - Snapshot the raw caller-supplied params (project_name, ensure_project_exists, api_key, base_url, trace_server_url) as a tuple before any
  resolution/mutation
  - Store this tuple on the client as _raw_init_params after construction
  - Reuse logic: if api_key is not None (explicit) and all raw params match → reuse the client. Otherwise tear down and recreate.

  This means:
  - weave.init("proj", api_key="key1") twice → reuses (identical explicit credentials)
  - weave.init("proj", api_key="key1") then weave.init("proj") → recreates (env may have changed)
  - weave.init("proj") then weave.init("proj") → recreates (env may have changed)
  - weave.init("proj", api_key="key1") then weave.init("proj", api_key="key2") → recreates (different credentials)

weave/trace/api.py
```diff
@@ -562,14 +582,12 @@ def finish() -> None:
...
-    # Capture client before teardown so we can still flush outstanding work.
     wc = weave_client_context.get_weave_client()
-    weave_init.finish()
-
-    # Flush any remaining calls
     if wc is not None:
         wc.finish()
 
+    weave_init.finish()
+
```
jtschoonhoven, line 590: Why this change?

weave/trace/call.py
```diff
@@ -179,6 +180,7 @@ def children(self, *, page_size: int = DEFAULT_CALLS_PAGE_SIZE) -> CallsIter:
             self.project_id,
             CallsFilter(parent_ids=[self.id]),
             page_size=page_size,
+            base_url=client._base_url,
```
jtschoonhoven, line 183: Shouldn't this be `self._base_url`?

weave/trace/call.py
jtschoonhoven, line 83: Safer to do `_base_url: str | None = field(default=None, compare=False, repr=False)` so this new field is excluded from `__eq__` and `__repr__`.

weave/trace/weave_init.py
```diff
@@ -95,40 +90,83 @@ def _weave_is_available(server: TraceServerClientInterface) -> bool:
...
         wandb_run_id = f"{entity_name}/{project_name}/{wb_run_context.run_id}"
         check_wandb_run_matches(wandb_run_id, entity_name, project_name)
 
-    remote_server = init_weave_get_server(api_key)
+    # Derive trace_server_url from base_url when not explicitly provided,
+    # honoring the documented contract: "The trace server URL is derived from
+    # this if trace_server_url is not provided."
+    resolved_trace_server_url = trace_server_url
+    if resolved_trace_server_url is None and base_url is not None:
+        normalized = base_url.rstrip("/")
```
jtschoonhoven, line 161: I'd make this `normalized = base_url.rstrip("/").lower()` and I'd also recommend checking for `https://` and adding it if it's missing.

jtschoonhoven (commented):
Some minor notes. Overall looks good to me.

chance-wnb:
> > Hi Rick, sorry I am kinda in a rush here, cause I need to take care of another emergent fix. I have to pause my code review here:
> > Could you help check whether your implementation might break the work of WAL (recent work from Andrew) ? I estimate that this might be related, since WAL mechanism might need to decide which server to send traffic to. Sorry I couldn't get a chance to check that myself.
> 
> ​Good call on the WAL.
> 
> The WAL system works with my branch because it receives the already-authenticated server object — it doesn't need to handle credentials independently. `ensure_project_exists()` changes add optional params are backwards-compatible and the WAL sender never calls that method anyway.

Hi Rick, thanks for checking it. You are correct that the in-process case, each weave-client keeps a WAL sender that pins to a server object. I am under the impression that the WAL is also supposed to work in a "process crash and recover" mode, I wonder what it will look like in that case. It might be fine for now, since it might out of the scope of this PR. May I look loop @andrewtruong to comment please?

PR #6553, opened 2026-04-07
feat(weave): resolve dataset-backed eval inputs via table_rows JOIN

nikumar1206:
## Description

This PR pushes resolving row-refs ot the existing clickhouse query, bringing it closer in functionality to what the /calls/stream_query endpoint does.  
Just separating this out so its easier to review PR2.

## Testing

- Unit tests

PR #6572, opened 2026-04-10
feat(weave): server-side sort/filter/pagination for /eval_results

nikumar1206:
## Description

  Pushes sort, filter, and pagination for /eval_results into ClickHouse via a CTE chain.

  CTE chain (eval_results_query_builder.py)

  - predict_and_score_calls → filter to PAS calls, extract row_digest
  - predict_and_score_calls_resolved → LEFT JOIN table_rows to resolve dataset-backed inputs
  - ranked_digests → GROUP BY row_digest, HAVING filters, ROW_NUMBER for sort order
  - ranked_digest_count → total matching rows for pagination metadata
  - page_digests → paginated slice of ranked_digests
  - page_rows → expand page digests to call IDs + resolved inputs

  Request/response (trace_server_interface.py)

  - EvalResultsSortBy: extends SortBy with evaluation_call_id and mode (value/difference)
  - EvalResultsFilter: evaluation_call_id + query expression
  - Added sort_by, filters to EvalResultsQueryBody

## Testing

- Unit tests


nikumar1206:
* **#6572** <a href="https://app.graphite.com/github/pr/wandb/weave/6572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6572?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6553** <a href="https://app.graphite.com/github/pr/wandb/weave/6553?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

tests/trace_server/test_trace_server_evaluation_apis.py
```diff
@@ -827,3 +833,289 @@ def test_eval_results_dataset_backed_no_resolve(client):
...
+    res = client.server.eval_results_query(
+        EvalResultsQueryReq(
+            project_id=client.project_id,
+            evaluation_call_ids=[eval_id],
+            include_raw_data_rows=True,
+            sort_by=[EvalResultsSortBy(field="outputs.answer", direction="asc")],
+        )
+    )
+    assert res.total_rows == 3
+    assert len(res.rows) == 3
```
bcsherma: Sorting by outputs doesn't work and this test should assert that the order is correct, I will leave a comment elsewhere with a summary of the bug

weave/trace_server/trace_server_interface.py
```diff
@@ -2667,6 +2667,36 @@ class ScoreDeleteRes(BaseModel):
...
+    evaluation_call_id: str | None = Field(
+        default=None,
+        description=("Scope the sort to a specific evaluation's scores."),
+    )
+    mode: Literal["value", "difference"] = Field(
+        default="value",
+        description=(
+            "When 'value', sort by the field value for the specified evaluation. "
+            "When 'difference', sort by max-min spread of the field across all "
+            "evaluations (evaluation_call_id is ignored)."
```
bcsherma, line 2682: is this useful? I don't think it's exposed anywhere in the UI - maybe for the mvp if this is null we just default to the first eval id in the list passed to the query?

weave/trace_server/eval_results_query_builder.py
```diff
@@ -102,3 +116,256 @@ def build_predict_and_score_calls_resolved_cte(
...
+
+    if field_path.startswith("inputs."):
+        remaining = field_path[len("inputs.") :]
+        parts = split_escaped_field_path(remaining)
+        path = _string_param(pb, quote_json_path_parts(parts))
+        inner = f"nullIf(JSON_VALUE(resolved_inputs, {path}), 'null')"
+        inner = _wrap_with_eval_scope(inner, evaluation_call_id, pb)
+        return f"any({inner})", {"inputs_dump"}
+
+    if field_path.startswith("outputs."):
```
bcsherma: In the weave api the names for call input/output fields are `inputs` (plural) and `output` (singular). I think we should stick with this even though it is funky

weave/trace_server/eval_results_query_builder.py
```diff
@@ -102,3 +116,256 @@ def build_predict_and_score_calls_resolved_cte(
...
+        remaining = field_path[len("scores.") :]
+        parts = ["scores"] + split_escaped_field_path(remaining)
+        path = _string_param(pb, quote_json_path_parts(parts))
+        inner = f"nullIf(JSON_VALUE(output_dump, {path}), 'null')"
+        inner = _wrap_with_eval_scope(inner, evaluation_call_id, pb)
+        return f"avg(toFloat64OrNull({inner}))", {"output_dump"}
+
+    raise InvalidRequest(
+        f"Unsupported eval results field: '{field_path}'. "
+        f"Supported prefixes: scores.*, inputs.*, outputs.*, row_digest."
```
bcsherma: how do we use row_digest. ?

weave/trace_server/eval_results_query_builder.py
```diff
@@ -102,3 +116,256 @@ def build_predict_and_score_calls_resolved_cte(
...
+        inner = _wrap_with_eval_scope(inner, evaluation_call_id, pb)
+        return f"any({inner})", {"output_dump"}
+
+    if field_path.startswith("scores."):
+        remaining = field_path[len("scores.") :]
+        parts = ["scores"] + split_escaped_field_path(remaining)
+        path = _string_param(pb, quote_json_path_parts(parts))
+        inner = f"nullIf(JSON_VALUE(output_dump, {path}), 'null')"
+        inner = _wrap_with_eval_scope(inner, evaluation_call_id, pb)
+        return f"avg(toFloat64OrNull({inner}))", {"output_dump"}
```
bcsherma: boolean score sorting isnt working because  these values come back as json serialized bool values, so you  need to do something like:
```
coerced = f"multiIf({inner} = 'true', '1', {inner} = 'false', '0', {inner})"
return f"avg(toFloat64OrNull({coerced}))", {"output_dump"}
```

bcsherma (changes requested):
I found some 🐛  but overall this is definitely the right direction, I think we are close to a majorly improved api <3

weave/trace_server/eval_results_query_builder.py
nikumar1206: row_digest is the default way we sort when no sort_by is provided. I think we can just remove that from the error message though. we dont need to expose that to the public interface.

weave/trace_server/trace_server_interface.py
nikumar1206: this is the only spot where this is used:

Screenshot 2026-04-13 at 12 56 45 PM

the compare evaluations landing page has table rows sorted by which rows have the biggest accuracy difference between the 2 evals.

weave/trace_server/eval_results_query_builder.py
nikumar1206: ✅

weave/trace_server/eval_results_query_builder.py
nikumar1206: ✅

tests/trace_server/test_trace_server_evaluation_apis.py
nikumar1206: yep, fixed the test to assert the order.
also yep outputs sorting was broken.

the sql generated was looking for output.output inside 'output_dump'. this should be resolved now.

bcsherma (approved):

PR #6573, opened 2026-04-10
fix(weave): use fresh CH session for write-target resolution to prevent SESSION_IS_LOCKED

gtarpenning:
https://coreweave.atlassian.net/browse/WB-30704

## Summary

- Seeing `SESSION_IS_LOCKED (373)` errors on the `upsert_batch` / write paths in production against CH Cloud
- Root cause: `resolve_v{1,2}_write_target` does a `ch_client.query()` (on residence cache miss) using the same thread-local session that subsequent inserts use. With CH Cloud's load balancer and the shared `_CH_POOL_MANAGER`, these can route through different HTTP connections where the session lock hasn't fully cleared
- Same fix pattern as 463e92daed (`query_stream` path): mint a fresh client with its own session for the residence query so it never conflicts with mutations
- Added `_resolve_v2_write_target` / `_resolve_v1_write_target` helpers that encapsulate the `with_new_client()` pattern, applied to all 10 call sites

### DIFF
--- a/tests/trace_server/test_llm_completion.py
+++ b/tests/trace_server/test_llm_completion.py
@@ -284,9 +284,13 @@ class TestLLMCompletionStreaming(unittest.TestCase):
 
     def setUp(self):
         """Set up test fixtures before each test."""
-        self.server = chts.ClickHouseTraceServer(host="test_host")
         mock_ch_client = MagicMock()
         mock_ch_client.query.return_value = MagicMock(result_rows=[[0, 0]])
+        self._mint_patcher = patch.object(
+            chts.ClickHouseTraceServer, "_mint_client", return_value=mock_ch_client
+        )
+        self._mint_patcher.start()
+        self.server = chts.ClickHouseTraceServer(host="test_host")
         self.server._thread_local.ch_client = mock_ch_client
         self.mock_secret_fetcher = MagicMock()
         self.mock_secret_fetcher.fetch.return_value = {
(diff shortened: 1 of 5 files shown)


neutralino1 (approved):

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -454,6 +454,20 @@ def table_routing_resolver(self) -> TableRoutingResolver:
             self._table_routing_resolver = TableRoutingResolver()
             return self._table_routing_resolver
 
+    def _resolve_v2_write_target(self, project_id: str) -> WriteTarget:
```
mscavezze-cw: Putting on my `defensive coding hat` for a moment...

Is there ever a reason why someone would legitimately want to call `self.table_routing_resolver.resolve_v2_write_target` without `with_new_client`?

If not, should the `with self.with_new_client()` go directly in `resolve_v2_write_target` in `TableRoutingResolver`?

Alternately, should we update `resolve_v2_write_target` there to make it more apparent that there's a potential foot-gun?

tests/trace_server/test_llm_completion.py
```diff
@@ -1308,11 +1313,14 @@ def mock_obj_read(req):
 @pytest.fixture
 def completions_mock_server():
     """Create a mock ClickHouseTraceServer for completions tests."""
-    server = chts.ClickHouseTraceServer(host="test_host")
     mock_ch_client = MagicMock()
     mock_ch_client.query.return_value = MagicMock(result_rows=[[0, 0]])
-    server._thread_local.ch_client = mock_ch_client
-    return server
+    with patch.object(
+        chts.ClickHouseTraceServer, "_mint_client", return_value=mock_ch_client
```
mscavezze-cw, line 1319: Consider adding an assert that `_mint_client` was actually called.

mscavezze-cw (approved):
Looks good. I added some comments about eliminating a potential foot-gun and reducing the number of wrappers we have.

bcsherma (approved):
Very low risk, rip it

PR #6577, opened 2026-04-10
fix(weave): avoid returning same call multiple times if it has multiple feedbacks

nikumar1206:
## Description

- Fixes: [Jira Ticket](https://coreweave.atlassian.net/browse/WB-33100)

Small bug where if the same call has multiple feedbacks, the call is returned multiple times.
This occurs due to a `LEFT JOIN` between the `calls_complete` and the `feedback` table.

This is not an issue on calls_merged as data is de-duped thanks to the `GROUP BY (project_id, id)` 

This PR adds a `DISTINCT` to calls_complete query when we do a join on the `feedback` table.

## Before

https://github.com/user-attachments/assets/3deff179-763b-4673-9862-69778ce5a91d

## After

Screenshot 2026-04-10 at 2 13 10 PM

## Testing

- Unit tests
- QA testing

### DIFF
--- a/tests/trace_server/query_builder/test_calls_query_builder.py
+++ b/tests/trace_server/query_builder/test_calls_query_builder.py
@@ -711,7 +711,7 @@ def test_query_with_simple_feedback_sort_and_filter() -> None:
 
 
 def test_query_with_simple_feedback_filter_calls_complete() -> None:
-    """Test feedback filter on calls_complete uses DISTINCT to avoid row duplication."""
+    """Test feedback filter on calls_complete table - should NOT use GROUP BY or aggregation."""
     cq = CallsQuery(project_id="project", read_table=ReadTable.CALLS_COMPLETE)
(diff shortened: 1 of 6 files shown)


tests/trace_server/query_builder/test_calls_query_builder.py
```diff
@@ -3456,7 +3456,7 @@ def test_calls_complete_with_feedback_filter() -> None:
     assert_sql(
         cq,
         """
-        SELECT
+        SELECT DISTINCT
             calls_complete.id AS id
```
gtarpenning, line 3460: can we add another select fieeld here, just curious what happens

gtarpenning: jk see cost query

gtarpenning (approved):
thanks for the fix, good tests

PR #6583, opened 2026-04-13
chore(weave): Make async tests actually async

andrewtruong:
Surprisingly this was not the case previously


neutralino1 (approved):
async you very much!

PR #6576, opened 2026-04-10
chore(weave): replicated ClickHouse CI test run for trace shard

gtarpenning:
## Summary

Adds CI coverage for replicated ClickHouse mode — the mode that broke in #6574 and had zero test coverage.

- New `--clickhouse-replicated` pytest flag configures the test CH server in replicated mode (Keeper-enabled, `ReplicatedMergeTree` tables, `WF_CLICKHOUSE_REPLICATED_CLUSTER` set)
- Runs the `trace` shard (end-to-end client->server->CH tests) on an 8-core runner with pytest-xdist parallel workers
- Python 3.12 only

**Non blocking, because it takes forever**

## Overhead

+1 `ubuntu-latest-8-cores` runner, ~30 min timeout.

PR #6587, opened 2026-04-13
test: verify replicated CI catches _local suffix bug

gtarpenning:
**DO NOT MERGE** — this PR exists only to verify that the replicated CI shard catches the bug fixed in #6574.

Reverts the fix from #6574 on top of the replicated CI branch (#6576). If the replicated ClickHouse tests fail, it proves the new shard would have caught the original bug.


gtarpenning:
didn't work

PR #6591, opened 2026-04-13
chore(weave): session-scoped ClickHouse fixture for faster CI

gtarpenning:
## Summary

Replaces per-test DB creation + 28 migrations (~1s each) with a single session-scoped DB that truncates tables between tests (~0.03s each).

**CI results (ClickHouse runs):**

| Shard | Baseline | This PR (3.10 / 3.11 / 3.13) | Improvement |
|-------|----------|-------------------------------|-------------|
| trace_server | 6m10s | 3m36s / 2m59s / 3m20s | ~50% faster |
| trace (-n8 xdist) | 7m56s | 3m10s / 3m06s / 3m07s | ~61% faster |

All 180 CI checks pass.

### DIFF
--- a/tests/trace_server/conftest.py
+++ b/tests/trace_server/conftest.py
@@ -17,6 +17,7 @@ from tests.trace_server.workers.evaluate_model_test_worker import (
     EvaluateModelTestDispatcher,
 )
 from weave.trace_server import clickhouse_trace_server_batched
+from weave.trace_server import clickhouse_trace_server_migrator as wf_migrator
 from weave.trace_server.clickhouse_trace_server_batched import ClickHouseTraceServer
 from weave.trace_server.project_version import project_version
 from weave.trace_server.secret_fetcher_context import secret_fetcher_context
@@ -120,11 +121,15 @@ def reset_project_version_cache():
     project_version.reset_project_residence_cache()
 
 
-def _get_worker_db_suffix(request) -> str:
-    """Get database suffix for pytest-xdist worker isolation."""
+def _get_worker_db_suffix(request, default: str = "_test") -> str:
+    """Get database suffix for pytest-xdist worker isolation.
+
+    Returns worker-specific suffix like '_w0' for xdist, or `default` for
+    single-process runs.
+    """
     worker_input = getattr(request.config, "workerinput", None)
     if worker_input is None:
-        return ""
+        return default
     worker_id = worker_input.get("workerid", "master")
     return f"_w{worker_id.replace('gw', '')}"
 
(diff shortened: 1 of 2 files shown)


gtarpenning:
@tssweeney Would appreciate your eyes on this. The change is a single file (`tests/trace_server/conftest.py`) — replaces the per-test DB create+migrate pattern with a session-scoped DB + table truncation between tests.

The main thing I'd want your take on:

1. **`_reset_server_state` completeness** — I'm resetting `_op_ref_cache`, `_placeholder_file_projects`, `_table_routing_resolver`, `_file_storage_client`, and the batch queues. If there's other mutable state on `ClickHouseTraceServer` I'm missing, tests will get order-dependent failures. Does this list look complete to you?

2. **`SEED_DATA_TABLES` approach** — `llm_token_prices` is excluded from truncation because migration 006 seeds it. If other migrations start seeding reference data, they'd need to be added here too. Is there a better pattern?

3. **`_flush_immediately = True`** — I force synchronous writes on the shared server so tests see data immediately after insert. The old per-test servers used batched writes that happened to flush because each server was short-lived. Any concerns with always-flush in tests?

tests/trace_server/conftest.py
```diff
@@ -127,102 +129,174 @@ def _get_worker_db_suffix(request) -> str:
...
+    server._placeholder_file_projects.clear()
+    # Reset table routing resolver to None so it's lazily re-created
+    # (tests may have set _mode to AUTO or accessed the resolver)
+    server._table_routing_resolver = None
+    # Reset file storage client so tests that mock env vars get a fresh client
+    server._file_storage_client = None
+    server._file_storage_client_initialized = False
+    # Clear batch queues (thread-local, clear for current thread)
+    tl = server._thread_local
+    if hasattr(tl, "call_batch"):
```
andrewtruong: Why do you need this?

andrewtruong: The hasattr checks are unfortunate

gtarpenning: the hasattr? isn't this because of the trace_server composition? will confirm

tests/trace_server/conftest.py
```diff
@@ -127,102 +129,174 @@ def _get_worker_db_suffix(request) -> str:
...
+            id_converter=id_converter
+        ),
+    )
+
+    # Drop and recreate from scratch once
+    ch_server.ch_client.command(f"DROP DATABASE IF EXISTS {management_db}")
+    ch_server.ch_client.command(f"DROP DATABASE IF EXISTS {unique_db}")
+    ch_server._database_ensured = False
+
+    import weave.trace_server.clickhouse_trace_server_migrator as wf_migrator
```
andrewtruong: do you need to defer?

gtarpenning: no, bug

tests/trace_server/conftest.py
gtarpenning: fixed

andrewtruong (approved):
nice!

PR #6592, opened 2026-04-13
chore(weave): auto-detect distributed tables from cluster shard count

gtarpenning:
## Summary
- Auto-detect whether distributed tables are needed by querying `system.clusters` for the cluster's shard count, instead of requiring `WF_CLICKHOUSE_USE_DISTRIBUTED_TABLES` to be explicitly set.
- When the env var is unset: >1 shard → distributed mode, 1 shard → replicated only, 0 shards (cluster not found) → non-distributed.
- Explicit `true`/`false` env var still overrides auto-detection.

## Changes
- **`database_engine.py`**: Added `detect_cluster_shard_count()` — queries `system.clusters` for distinct shard count.
- **`environment.py`**: `wf_clickhouse_use_distributed_tables()` now returns `bool | None` — `None` when env var is unset (signals auto-detect).
- **`clickhouse_trace_server_migrator.py`**: Factory function auto-detects when `use_distributed=None` and `replicated=True`. Falls back to `False` on query failure.
- **`clickhouse_trace_server_batched.py`**: `use_distributed_mode` property auto-detects and caches the result.

## Test plan
- [ ] Verify on a multi-shard cluster: migrator auto-selects `DistributedClickHouseTraceServerMigrator`
- [ ] Verify on a single-shard cluster: migrator auto-selects `ReplicatedClickHouseTraceServerMigrator`
- [ ] Verify explicit `WF_CLICKHOUSE_USE_DISTRIBUTED_TABLES=true` still overrides to distributed
- [ ] Verify explicit `WF_CLICKHOUSE_USE_DISTRIBUTED_TABLES=false` still overrides to non-distributed

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -132,10 +132,17 @@ from weave.trace_server.clickhouse_schema import (
     SelectableCHObjSchema,
     TagCHInsertable,
 )
+from weave.trace_server.clickhouse_trace_server_migrator import (
+    DEFAULT_REPLICATED_CLUSTER,
+)
 from weave.trace_server.common_interface import AnnotationQueueItemsFilter
 from weave.trace_server.constants import (
     COMPLETIONS_CREATE_OP_NAME,
     IMAGE_GENERATION_CREATE_OP_NAME,
+)
+from weave.trace_server.database_engine import (
+    EngineDiscoveryError,
+    detect_cluster_shard_count,
 )
 from weave.trace_server.datadog import (
     generator_trace,
@@ -340,6 +347,10 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         self._op_ref_cache_lock = threading.Lock()
         self._placeholder_file_projects: set[str] = set()
         self._database_ensured = False
+        # Cached result of distributed-mode auto-detection (None = not yet resolved).
+        # Once resolved by querying system.clusters, the bool is stored here so
+        # we don't re-query on every call to use_distributed_mode.
+        self._use_distributed_cached: bool | None = None
 
         if wf_env.wf_clickhouse_disable_lightweight_update():
             logger.warning(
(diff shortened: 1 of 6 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -458,12 +469,47 @@ def table_routing_resolver(self) -> TableRoutingResolver:
...
-            bool: True if using distributed tables, False otherwise.
+        Resolution order:
+        1. If WF_CLICKHOUSE_USE_DISTRIBUTED_TABLES is explicitly set → use that.
+        2. Otherwise, auto-detect by querying system.clusters for the shard
+           count of the configured cluster (>1 shard → distributed).
+        3. Cache the result so we only query system.clusters once.
         """
-        return wf_env.wf_clickhouse_use_distributed_tables()
+        # Fast path: already resolved and cached.
+        if self._use_distributed_cached is not None:
```
rgao-coreweave, line 479: Every other lazily-initialized property in `ClickHouseTraceServer` (`file_storage_client`, `kafka_producer`, `table_routing_resolver`) uses double-checked locking with `self._init_lock`. The new `use_distributed_mode` property reads `_use_distributed_cached`, and if `None`, performs a network query to ClickHouse and writes back — all without any lock.

In a multi-threaded server, concurrent threads hitting this property before it's cached will each independently fire the `system.clusters` query. More critically, if the query fails intermittently, one thread could cache `False` while a concurrent thread's query would have succeeded with `True`.

```python
# Suggestion: use the same double-checked locking pattern
@property
def use_distributed_mode(self) -> bool:
    if self._use_distributed_cached is not None:
        return self._use_distributed_cached
    with self._init_lock:
        if self._use_distributed_cached is not None:
            return self._use_distributed_cached
        # ... env check and auto-detect logic ...
        self._use_distributed_cached = result
        return result
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -458,12 +469,47 @@ def table_routing_resolver(self) -> TableRoutingResolver:
...
+            # Explicit override from env var — cache and return.
+            self._use_distributed_cached = env_value
+            return env_value
+
+        # Auto-detect: query system.clusters for the shard count.
+        cluster = (
+            wf_env.wf_clickhouse_replicated_cluster() or DEFAULT_REPLICATED_CLUSTER
+        )
+        try:
+            shard_count = detect_cluster_shard_count(self.ch_client, cluster)
```
rgao-coreweave, line 494: The migrator factory guards auto-detection with `if use_distributed is None and replicated:` and explicitly sets `use_distributed = False` when not replicated. But here there's no equivalent check — if the env var is unset, it always queries `system.clusters` regardless of whether replication is enabled.

Is this intentional? On a non-replicated single-node deployment with no env var set, this would make a pointless query to `system.clusters` and potentially log a confusing `EngineDiscoveryError` warning.

weave/trace_server/clickhouse_trace_server_batched.py
rgao-coreweave: Two threads (A and B) both call use_distributed_mode at the same time when the cache is None:                                                   
```
  1. Thread A passes the first is not None check → enters with self._init_lock
  2. Thread B also passes the first is not None check → blocks waiting on the lock
  3. Thread A passes the second is not None check → queries ClickHouse → sets _use_distributed_cached = True → releases lock
  4. Thread B acquires the lock → hits the second is not None check → sees True (already set by A) → returns immediately without querying
```

Without that second check, Thread B would also query ClickHouse redundantly after acquiring the lock.

PR #6593, opened 2026-04-13
test(weave): skip calls query builder unit tests in clickhouse CI

gtarpenning:
## Description

Skip the entire `tests/trace_server/query_builder` directory in the ClickHouse `trace_server` CI shard. These are SQL-generation unit tests and are already covered in the SQLite `trace_server` run.

Measured on the latest completed `Trace nox tests (3, 13, trace_server)` job for `master` vs this branch's skip-change run:
- ClickHouse step: `6m 21s` on `master` -> `5m 4s` on this branch, saving `1m 17s` (`%20`)

## Testing

test only change

---

<a href="https://app.devin.ai/review/wandb/weave/pull/6593" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tests/trace_server/query_builder/conftest.py
+++ b/tests/trace_server/query_builder/conftest.py
@@ -3,6 +3,7 @@ import pytest
 
 def pytest_collection_modifyitems(items):
     # These are SQL-generation unit tests; keep them in the SQLite trace_server run
-    # and exclude the whole directory from the ClickHouse CI shard.
+    # and exclude only this directory from the ClickHouse CI shard.
     for item in items:
-        item.add_marker(pytest.mark.skip_clickhouse_client)
+        if "query_builder" in item.path.parts:
+            item.add_marker(pytest.mark.skip_clickhouse_client)
--- a/tests/trace_server/query_builder/conftest.py
+++ b/tests/trace_server/query_builder/conftest.py
@@ -1,0 +1,6 @@
+import pytest
+
+
+def pytest_collection_modifyitems(items):
+    for item in items:
+        item.add_marker(pytest.mark.skip_clickhouse_client)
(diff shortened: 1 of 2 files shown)


tests/trace_server/query_builder/conftest.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+
+def pytest_collection_modifyitems(items):
```
jtschoonhoven: I hadn't seen `pytest_collection_modifyitems` before and I didn't know it was a pytest hook. Consider adding a docstring that mentions how this works and maybe links to https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems.

Also, from the docs it looks like `items` is the third positional arg. Are you sure this signature is right? Consider adding mypy type annotations.

tests/trace_server/query_builder/conftest.py
```diff
@@ -0,0 +1,9 @@
...
+    # These are SQL-generation unit tests; keep them in the SQLite trace_server run
+    # and exclude only this directory from the ClickHouse CI shard.
+    for item in items:
+        if "query_builder" in item.path.parts:
```
jtschoonhoven, line 14: Isn't this guaranteed to be true since we're already in the `query_builder` subdir?

EDIT: Yeah it's necessary. Probably worth documenting that this hook receives _all_ tests, not just those in this directory.

jtschoonhoven (approved):
Approved with suggestions for comments

tests/trace_server/query_builder/conftest.py
gtarpenning: you would think...

PR #6597, opened 2026-04-13
fix(weave): restore replicated table engines in replicated DBs

gtarpenning:
## Summary
This fixes a replicated ClickHouse migration regression where fresh `ENGINE = Replicated(...)` databases were creating plain `MergeTree`-family tables instead of `Replicated*MergeTree` tables.

## Actual cause
The regression came from a bad assumption introduced in `f654147b91` ([chore(weave): actually test the replicated ch tests in CI](https://github.com/wandb/weave/pull/6419)). We taught the migrator to skip `ON CLUSTER` for `Replicated` databases to avoid ClickHouse error 80, which was correct. But we also started skipping the table-engine rewrite entirely under the assumption that a `Replicated` database would auto-convert `MergeTree` tables into replicated table engines for us.

That assumption is false for the clusters we run against. The result was:
- fresh replicated-only DBs created plain `MergeTree` / `ReplacingMergeTree` / `AggregatingMergeTree` tables
- each replica stored independent local data! Not replicated!

## Fix
The migrator now treats engine conversion and `ON CLUSTER` as separate concerns.

Engine handling matrix:
- replicated-only DB: `Replicated*MergeTree()` with no `ON CLUSTER`
- distributed + Atomic DB: explicit ZooKeeper args plus `ON CLUSTER`
- distributed + Replicated DB: `Replicated*MergeTree()` with no explicit ZooKeeper args and no `ON CLUSTER`

This preserves the original reason for the March fix:
- we still do not emit `ON CLUSTER` inside `ENGINE = Replicated` databases
- we still avoid explicit ZooKeeper args in distributed local tables when ClickHouse forbids them in `Replicated` databases

## Why this is durable
This change makes the logic explicit instead of relying on implicit ClickHouse behavior:
- the replicated-only and distributed codepaths now encode the engine-handling matrix directly
- the migrator comments explain the expected behavior at the decision points
- tests now assert actual table engines, not just database engine / table existence

## Preventing regressions
The main failure before was that our tests proved DDL executed, but not that the resulting storage engines were actually replicated.

This change tightens that by:
- updating unit tests to assert the exact DDL emitted for replicated vs distributed cases
- updating functional tests to check `system.tables.engine_full` for replicated tables in replicated mode and replicated local tables in distributed mode

That means future changes that accidentally turn replicated tables back into plain `MergeTree` tables should fail in CI instead of only showing up on a fresh multi-replica cluster.

### DIFF
--- a/tests/trace_server/test_clickhouse_trace_server_migrator.py
+++ b/tests/trace_server/test_clickhouse_trace_server_migrator.py
@@ -362,14 +362,14 @@ def test_replicated_management_table_follows_database_engine(replicated_migrator
     def _normalize(sql: str) -> str:
         return " ".join(sql.split())
 
-    # Replicated DB: plain MergeTree, no ON CLUSTER (auto-replicated)
+    # Replicated DB: explicit ReplicatedMergeTree, no ON CLUSTER
     replicated_migrator._replicated_db_engine_cache[
         replicated_migrator.management_db
     ] = True
     assert _normalize(replicated_migrator._create_management_table_sql()) == _normalize(
         """CREATE TABLE IF NOT EXISTS db_management.migrations
            ( db_name String, curr_version UInt64, partially_applied_version UInt64 NULL, )
-           ENGINE = MergeTree() ORDER BY (db_name)"""
+           ENGINE = ReplicatedMergeTree() ORDER BY (db_name)"""
     )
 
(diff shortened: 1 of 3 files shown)


jtschoonhoven (approved):
LGTM.

I miss postgres 😿

PR #6612, opened 2026-04-15
chore(weave): Add tool call auto-retry, compact events support to pi-coding-agent

chance-wnb:
## Description

Implements OpenTelemetry tracing for Pi Coding Agent tool execution and session management events. Adds span creation and tracking for tool calls with proper attributes including tool name, call ID, and conversation ID. Implements tool result handling with error status tracking and content capture. Adds session compaction tracing with reason, abort status, and retry information. Implements auto-retry event tracking with attempt counts, error messages, and success status.

## Testing

TBD

PR #6611, opened 2026-04-15
chore(weave): pi-coding-agent, chat spans

chance-wnb:
## Description

Implements OpenTelemetry tracing for Pi Coding Agent chat interactions by adding span creation and event tracking.

- `onTurnStart` now creates a chat span with model information, operation name, and conversation ID attributes
- `onContext` adds system message events to the chat span when content capture is enabled
- `onMessageEnd` adds assistant message events to the chat span when content capture is enabled
- `onTurnEnd` sets response attributes including model, finish reasons, token usage, cache usage, and cost information before ending the chat span
- `endChatSpan` properly terminates and cleans up the chat span

## Testing

Implemented in https://github.com/wandb/weave/pull/6612

PR #6610, opened 2026-04-15
chore(weave): basic session tracking set up

chance-wnb:
## Description

Implements session and agent invocation span tracking for the Pi Coding Agent OpenTelemetry adapter.

- Adds conversation ID tracking that persists throughout the session
- Implements session span creation in `onSessionStart` with agent name, working directory, and conversation ID attributes
- Implements agent invocation span creation in `onBeforeAgentStart` with operation name, agent name, provider, and conversation ID attributes
- Adds user message event capture when content capture is enabled
- Implements proper span cleanup in `onAgentEnd` and `onSessionShutdown` handlers
- Adds error handling for open tool spans when agent ends unexpectedly
- Tracks current model state from model selection events

## Testing

Implemented in https://github.com/wandb/weave/pull/6612

PR #6609, opened 2026-04-15
chore(weave): initial set up for the pi-coding-agent integration

chance-wnb:
## Description

Initial setup for the integration for pi.dev coding agent.

- Add `@opentelemetry/api` dependency (^1.9.1)
- Implement `PiCodingAgentOtelAdapter` class that creates OpenTelemetry spans for the full agent lifecycle
- Add `createOtelExtension()` factory function to create pi coding agent extensions
- Export new integration functions and types from integrations index
- Include comprehensive type definitions for pi.dev coding agent extension system
- Support GenAI semantic conventions for tracing agent operations, tool calls, and session management

## Testing

Implemented in https://github.com/wandb/weave/pull/6612

---

<a href="https://app.devin.ai/review/wandb/weave/pull/6609" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #6613, opened 2026-04-15
chore(weave): allows the pi-coding-agent to automatically send OTEL spans to the weave endpoint

chance-wnb:
## Description

This PR adds OpenTelemetry trace export capabilities to the Pi Coding Agent integration. The implementation automatically configures an OTLP trace exporter when a Weave global client is available, enabling traces to be sent to the Weave trace server automatically without needing a user sets up https://docs.wandb.ai/weave/guides/integrations/pydantic_ai#configure-otel-tracing-in-weave.

Key changes:

- Added OpenTelemetry dependencies: `@opentelemetry/exporter-trace-otlp-proto`, `@opentelemetry/resources`, and `@opentelemetry/sdk-trace-base (These are must-haves, because the client appplication does not necessarily install them into the runtime)`
- Enhanced `PiCodingAgentOtelAdapter` to auto-configure a `BasicTracerProvider` with OTLP export when no custom tracer is provided
- Configured the exporter to send traces to the Weave trace server endpoint with proper authentication headers
- Added proper cleanup by shutting down the private provider during session shutdown
- Made session shutdown asynchronous to handle the provider shutdown properly

The adapter now falls back gracefully: if a custom tracer is provided, it uses that; if a Weave client exists, it auto-configures OTLP export; otherwise, it uses the global tracer.

## Testing

TBD

---

<a href="https://app.devin.ai/review/wandb/weave/pull/6613" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/sdks/node/package.json
+++ b/sdks/node/package.json
@@ -140,10 +140,8 @@
     "ext": "ts,json",
     "exec": "tsx examples/evaluate.ts"
   },
-  "peerDependencies": {
-    "@opentelemetry/api": "^1.9.1"
-  },
   "dependencies": {
+    "@opentelemetry/api": "^1.9.1",
     "@opentelemetry/exporter-trace-otlp-proto": "^0.53.0",
     "@opentelemetry/resources": "^1.26.0",
     "@opentelemetry/sdk-trace-base": "^1.26.0",
--- a/sdks/node/package.json
+++ b/sdks/node/package.json
@@ -144,6 +144,9 @@
     "@opentelemetry/api": "^1.9.1"
   },
   "dependencies": {
+    "@opentelemetry/exporter-trace-otlp-proto": "^0.53.0",
+    "@opentelemetry/resources": "^1.26.0",
+    "@opentelemetry/sdk-trace-base": "^1.26.0",
     "cli-progress": "^3.12.0",
     "cross-spawn": "^7.0.5",
     "form-data": "^4.0.4",
@@ -155,7 +158,6 @@
     "uuidv7": "^1.0.1"
   },
   "devDependencies": {
-    "@opentelemetry/sdk-trace-base": "^1.26.0",
     "@types/cli-progress": "^3.11.6",
     "@types/ini": "^1.3.3",
     "@types/jest": "^29.5.13",
--- a/sdks/node/src/integrations/piCodingAgent.ts
+++ b/sdks/node/src/integrations/piCodingAgent.ts
@@ -412,7 +412,10 @@ export class PiCodingAgentOtelAdapter {
           [ATTR.PI_USAGE_COST_USD]: usage.cost.total,
         });
 
-        if (event.message.stopReason === 'error' && event.message.errorMessage) {
+        if (
+          event.message.stopReason === 'error' &&
+          event.message.errorMessage
+        ) {
           this.chatSpan.setAttribute(ATTR.ERROR_TYPE, 'llm_error');
           this.chatSpan.setStatus({
             code: SpanStatusCode.ERROR,
--- a/sdks/node/src/integrations/piCodingAgent.ts
+++ b/sdks/node/src/integrations/piCodingAgent.ts
@@ -8,14 +8,17 @@
  * Usage:
  * ```typescript
  * import { init } from 'weave';
- * import { createAgentSession } from '@mariozechner/pi-coding-agent';
  * import { createOtelExtension } from 'weave';
+ * import { createAgentSession, DefaultResourceLoader } from '@mariozechner/pi-coding-agent';
  *
  * await init('my-entity/my-project');
  *
- * const session = await createAgentSession({
- *   extensions: [createOtelExtension()],
+ * const resourceLoader = new DefaultResourceLoader({
+ *   extensionFactories: [createOtelExtension()],
  * });
+ * await resourceLoader.reload();
+ *
+ * const { session } = await createAgentSession({ resourceLoader });
  * ```
  */
 
@@ -32,7 +35,7 @@ import {OTLPTraceExporter} from '@opentelemetry/exporter-trace-otlp-proto';
 import {Resource} from '@opentelemetry/resources';
 import {
   BasicTracerProvider,
-  BatchSpanProcessor,
+  SimpleSpanProcessor,
 } from '@opentelemetry/sdk-trace-base';
 
 import {getGlobalClient} from '../clientApi';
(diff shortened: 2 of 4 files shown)


chance-wnb:
* **#6613** <a href="https://app.graphite.com/github/pr/wandb/weave/6613?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6613?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6612** <a href="https://app.graphite.com/github/pr/wandb/weave/6612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6611** <a href="https://app.graphite.com/github/pr/wandb/weave/6611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6610** <a href="https://app.graphite.com/github/pr/wandb/weave/6610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6609** <a href="https://app.graphite.com/github/pr/wandb/weave/6609?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sdks/node/package.json
```diff
@@ -142,6 +142,9 @@
   },
   "dependencies": {
     "@opentelemetry/api": "^1.9.1",
+    "@opentelemetry/exporter-trace-otlp-proto": "^0.53.0",
+    "@opentelemetry/resources": "^1.26.0",
+    "@opentelemetry/sdk-trace-base": "^1.26.0",
```
andrewtruong: These are quite heavy, I wonder if they can be optional deps?

sdks/node/src/integrations/piCodingAgent.ts
```diff
@@ -462,13 +516,16 @@ export class PiCodingAgentOtelAdapter {
     }
   }
 
-  private endSessionSpan(): void {
+  private async endSessionSpan(): Promise<void> {
```
andrewtruong, line 523: Why is this the only one changed to be async?

sdks/node/package.json
chance-wnb: These are not dependencies customer would necessarily include in their apps. And they are indeed needed by us in the runtime. 

I made them as `optional peer dependencies`. This adds some adoption friction, since the host application will have to explicitly do 

```
npm install @opentelemetry/sdk-trace-base @opentelemetry/exporter-trace-otlp-proto @opentelemetry/resources
```

but I doubt any user would like to do it in this verbose way at all. 

In case these are missing we will purely rely on the OTEL sink the user configures to ingest the OTEL spans. In case it is not the weave endpoint, we are going to receive nothing. 

is it really better? If not, I can revert.

sdks/node/src/integrations/piCodingAgent.ts
chance-wnb: Good question,

because endSessionSpan needs to await this.privateProvider.shutdown() to flush buffered spans before the process exits. BatchSpanProcessor buffers spans and exports them in batches — without awaiting the shutdown, the process could exit with unflushed spans.

But this only matters when privateProvider is non-null (auto-configure path). The PiExtensionHandler type already supports Promise<void> | void return, so the pi agent will await it.

sdks/node/package.json
andrewtruong: Does TS have an idea of "extras"?  Like can you `npm install weave[otel]`?



If not, we can revert

sdks/node/src/integrations/piCodingAgent.ts
andrewtruong: Should the other methods be async too?

chance-wnb: I checked the other event handlers: for those who are sync, they are not invoking async procedures, so they should remain sync. 

There is no need to make sync async. async can call sync easily, but not the other way around. So we don't have to make sync async, which would cause extra hassle for the callers.

sdks/node/package.json
chance-wnb: Nope, npm doesn't have that.

I have reverted.

sdks/node/src/integrations/piCodingAgent.ts
chance-wnb: Sorry, I need to make a correction: the pi-coding-agent doesn't wait on an async callback (with pi.on('evtName', callback)).

Since the  `session_shutdown` event is not waited. I had to put a 
```
process.once('beforeExit', () => .. ) 
```
to ensure that the `provider.shutdown()` can complete its job.

sdks/node/src/integrations/piCodingAgent.ts
```diff
@@ -144,8 +156,59 @@ export class PiCodingAgentOtelAdapter {
...
+            this.endInvokeAgentSpan();
+            if (this.sessionSpan) {
+              this.sessionSpan.end();
+              this.sessionSpan = null;
+              this.sessionCtx = ROOT_CONTEXT;
+            }
+            await provider.shutdown().catch(() => {});
+          });
+        } catch {
+          this.tracer = trace.getTracer('pi-coding-agent-weave-ext');
```
andrewtruong, line 215: should we provide a warning here?

sdks/node/src/integrations/piCodingAgent.ts
```diff
@@ -144,8 +156,59 @@ export class PiCodingAgentOtelAdapter {
...
+        }
+      } else {
+        this.tracer = trace.getTracer('pi-coding-agent-weave-ext');
```
andrewtruong, line 221: and here?

sdks/node/src/integrations/piCodingAgent.ts
```diff
@@ -144,8 +156,59 @@ export class PiCodingAgentOtelAdapter {
...
+          // endSessionSpan may not complete. This hook catches the
+          // event loop draining and flushes any remaining spans.
+          process.once('beforeExit', async () => {
+            this.endInvokeAgentSpan();
+            if (this.sessionSpan) {
+              this.sessionSpan.end();
+              this.sessionSpan = null;
+              this.sessionCtx = ROOT_CONTEXT;
+            }
+            await provider.shutdown().catch(() => {});
```
andrewtruong: should provide a warning here?

sdks/node/src/integrations/piCodingAgent.ts
```diff
@@ -144,8 +156,59 @@ export class PiCodingAgentOtelAdapter {
...
+          const endpoint = `${client.traceServerApi.baseUrl}/otel/v1/traces`;
+
+          const authHeader = `Basic ${Buffer.from(`api:${apiKey}`).toString('base64')}`;
+          const provider = new BasicTracerProvider({
+            resource: new Resource({
+              'wandb.entity': entity,
+              'wandb.project': project,
+            }),
+            spanProcessors: [
+              new SimpleSpanProcessor(
```
andrewtruong, line 178: should this be BatchSpanProcessor instead?

chance-wnb: Good question. With the beforeExit hook + provider.shutdown() force-flushing, BatchSpanProcessor would now work — the buffered spans get flushed on exit.

But SimpleSpanProcessor is still the better choice here:

1. Coding agent sessions produce tens of spans over minutes of runtime, not thousands per second. Batching doesn't meaningfully reduce network overhead.
2. SimpleSpanProcessor ships each span immediately, so we can see live progress in the Weave UI as the agent works — not just after it finishes.
3. If the process crashes (not a clean exit), beforeExit doesn't fire. SimpleSpanProcessor would have already sent all ended spans; BatchSpanProcessor would lose everything in the buffer.

I suggest we Keep it.

andrewtruong (approved):

PR #6618, opened 2026-04-15
fix(weave): sanitize file paths in MemTraceFilesArtifact

gtarpenning:
## Summary

- Add `_safe_join()` helper that rejects absolute paths, `..` traversal, and symlink escapes
- Apply to `MemTraceFilesArtifact.path()` and `writeable_file_path()` to ensure all file writes stay within the intended temp directory
- Add tests covering valid relative paths, absolute path rejection, traversal rejection, and filename override

https://coreweave.atlassian.net/browse/VULNMGMT-1008
https://coreweave.atlassian.net/browse/WB-32491

### DIFF
--- a/tests/trace/test_mem_artifact.py
+++ b/tests/trace/test_mem_artifact.py
@@ -1,0 +1,40 @@
+import pytest
+
+from weave.trace.serialization.mem_artifact import MemTraceFilesArtifact
+
+
+def test_mem_artifact_path_sanitization():
+    art = MemTraceFilesArtifact(
+        path_contents={
+            "audio.wav": b"RIFF",
+            "subdir/file.txt": b"nested",
+            "/etc/passwd": b"abs",
+            "../../etc/shadow": b"traversal",
+        }
+    )
+
+    # Valid relative paths work, including nested
+    assert art.path("audio.wav").endswith("audio.wav")
+    result = art.path("subdir/file.txt")
+    assert result.endswith("subdir/file.txt")
+
+    # Absolute paths and traversals are rejected in both path() and filename override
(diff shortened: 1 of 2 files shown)


tests/trace/test_mem_artifact.py
```diff
@@ -0,0 +1,42 @@
...
+        }
+    )
+
+    # Valid relative paths work, including nested
+    assert art.path("audio.wav").endswith("audio.wav")
+    result = art.path("subdir/file.txt")
+    assert result.endswith(os.path.join("subdir", "file.txt"))
+
+    # Absolute paths and traversals are rejected in both path() and filename override
+    with pytest.raises(ValueError, match="absolute path|escapes base directory"):
```
andrewtruong: we should be more specific about the match

weave/trace/serialization/mem_artifact.py
```diff
@@ -82,7 +108,8 @@ def path(self, path: str, filename: str | None = None) -> str:
             raise FileNotFoundError(path)
 
         self.temp_read_dir = tempfile.TemporaryDirectory()
```
andrewtruong, line 114: Shouldn't this be a in contextmanager?

andrewtruong (approved):

tests/trace/test_mem_artifact.py
gtarpenning: yep, breaking up or condition, should make this possible

weave/trace/serialization/mem_artifact.py
gtarpenning: probably, but not our change

PR #6621, opened 2026-04-15
fix(weave): restrict load_op fallback to Op type only

gtarpenning:
The `_decode_custom_obj` fallback path would call `wc.get()` on an arbitrary user-supplied op URI for any unrecognized type string. This restricts the `load_op` fallback to the `Op` type only, raising `ValueError` for all other unknown types. Prevents fetching and executing untrusted ops during deserialization.

Practically, only `Op` types should be trigger `load_op`, not sure why this wasn't the case since the beginning. Probably a bug. 

https://coreweave.atlassian.net/browse/WB-33243

### DIFF
--- a/tests/trace/test_custom_objs.py
+++ b/tests/trace/test_custom_objs.py
@@ -1,9 +1,7 @@
 from __future__ import annotations
 
 from datetime import datetime, timezone
-from unittest.mock import MagicMock, patch
 
-import pytest
 import rich.markdown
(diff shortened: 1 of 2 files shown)


gtarpenning:
Closing — the load_op fallback is intentionally general. It's the mechanism for ALL user-registered custom types to survive across sessions (not just Op). The server-side protection is handled at the worker boundary in #6622 instead.

PR #6622, opened 2026-04-15
fix(weave): reject CustomWeaveType payloads in evaluate model worker

gtarpenning:
## Summary

- Add `_assert_safe_payload` pre-deserialization check to the evaluate model worker that recursively rejects any `CustomWeaveType` nodes in raw object payloads before `client.get()` is called
- Validates both `evaluation_ref` and `model_ref` payloads via `obj_read` before deserialization
- Matches the shape of the existing scoring worker's `_assert_safe_scorer_payload`

https://coreweave.atlassian.net/browse/VULNMGMT-1007

## Overhead

Adds 2 extra `obj_read` calls (one per ref) plus a shallow recursive walk of the raw JSON vals. Refs inside the payload are strings and skipped immediately — we don't resolve them. Negligible compared to running the evaluation itself.

## Note on load_op fallback

The `_decode_custom_obj` fallback in `custom_objs.py` allows any `CustomWeaveType` with a `load_op` field to load via a stored op URI. This is intentional — it's the mechanism for user-registered custom types to survive across sessions. Server-side protection is handled here at the worker boundary.

## Surface area

This blocks ALL `CustomWeaveType` deserialization in the evaluate model worker. If evaluations or models contain `CustomWeaveType` values (e.g. images, audio, custom serialized objects), those will now be rejected at the worker boundary.

### DIFF
--- a/weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
+++ b/weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
@@ -1,6 +1,5 @@
 import asyncio
 from abc import ABC, abstractmethod
-from typing import Any
 
 import ddtrace
 from pydantic import BaseModel, ConfigDict
@@ -15,12 +14,9 @@ from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.interface.builtin_object_classes.llm_structured_model import (
     LLMStructuredCompletionModel,
 )
+from weave.trace_server.validation import assert_safe_payload
 
 EVALUATE_MODEL_WORKER_MARKER = {"_weave_eval_meta": {"evaluate_model_worker": True}}
-
-
-class UnsafePayloadError(TypeError):
-    """Raised when a ref resolves to a payload that is not safe for server-side deserialization."""
 
 
 class EvaluateModelArgs(BaseModel):
@@ -39,43 +35,6 @@ class EvaluateModelDispatcher(ABC):
         pass
 
 
-def _assert_safe_payload(value: Any, path: str = "root") -> None:
-    """Recursively walk a raw serialized payload and reject CustomWeaveType nodes.
-
-    CustomWeaveType payloads can trigger code execution during deserialization
-    (e.g. Op types call __import__ on user-uploaded Python files, and unknown
-    types fall back to a load_op path that does the same). None of these should
-    be deserialized in a server-side worker process.
-
-    https://coreweave.atlassian.net/browse/VULNMGMT-1007
-    """
-    if isinstance(value, str):
-        return
-
-    if isinstance(value, list):
-        for idx, item in enumerate(value):
-            _assert_safe_payload(item, f"{path}[{idx}]")
-        return
-
-    if not isinstance(value, dict):
-        return
-
-    if value.get("_type") == "CustomWeaveType":
-        weave_type = value.get("weave_type")
-        custom_type = (
-            weave_type.get("type", "unknown")
-            if isinstance(weave_type, dict)
-            else "unknown"
-        )
-        raise UnsafePayloadError(
-            f"Evaluate model worker does not allow CustomWeaveType payloads "
-            f"({custom_type} at {path})"
-        )
-
-    for key, item in value.items():
-        _assert_safe_payload(item, f"{path}.{key}")
-
-
 def _assert_safe_ref(client: WeaveClient, ref_uri: str, label: str) -> None:
     """Read the raw object for a ref and reject it if it contains unsafe CustomWeaveType payloads."""
     ref = Ref.parse_uri(ref_uri)
@@ -90,7 +49,7 @@ def _assert_safe_ref(client: WeaveClient, ref_uri: str, label: str) -> None:
             digest=ref.digest,
         )
     )
-    _assert_safe_payload(read_res.obj.val, label)
+    assert_safe_payload(read_res.obj.val, label)
 
 
 def evaluate_model(args: EvaluateModelArgs) -> None:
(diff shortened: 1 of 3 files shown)


weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -41,6 +101,11 @@ def evaluate_model(args: EvaluateModelArgs) -> None:
 def _evaluate_model(args: EvaluateModelArgs) -> None:
     client = require_weave_client()
 
+    # Validate raw payloads before deserialization.
+    # https://coreweave.atlassian.net/browse/VULNMGMT-1007
+    _assert_safe_ref(client, args.evaluation_ref, "evaluation_ref")
```
gtarpenning, line 48: @andrewtruong this is going to walk literally every ref in an eval.... impossible. okay...

gtarpenning: wait, but then how was it working before 🤔

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -33,6 +39,60 @@ def dispatch(self, args: EvaluateModelArgs) -> None:
         pass
 
 
+def _assert_safe_payload(value: Any, path: str = "root") -> None:
```
jtschoonhoven: Nit: this function probably doesn't need to be sorted at the top of the module

andrewtruong: this should probably be a general helper in the server.  I'm sure more than the evaluate worker will use it

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -8,15 +9,20 @@
...
 from weave.trace.weave_client import WeaveClient
+from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.interface.builtin_object_classes.llm_structured_model import (
     LLMStructuredCompletionModel,
 )
 
 EVALUATE_MODEL_WORKER_MARKER = {"_weave_eval_meta": {"evaluate_model_worker": True}}
 
 
+class UnsafePayloadError(TypeError):
```
andrewtruong: This should be `ValueError`

jtschoonhoven (approved):
Ship it. But I think we should fix this in a more permanent way.

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -33,6 +39,60 @@ def dispatch(self, args: EvaluateModelArgs) -> None:
...
+
+    CustomWeaveType payloads can trigger code execution during deserialization
+    (e.g. Op types call __import__ on user-uploaded Python files, and unknown
+    types fall back to a load_op path that does the same). None of these should
+    be deserialized in a server-side worker process.
+
+    https://coreweave.atlassian.net/browse/VULNMGMT-1007
+    """
+    if isinstance(value, str):
+        return
```
andrewtruong: isn't this redundant with the `not dict` check below?

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -33,6 +39,60 @@ def dispatch(self, args: EvaluateModelArgs) -> None:
...
+            f"({custom_type} at {path})"
+        )
+
+    for key, item in value.items():
+        _assert_safe_payload(item, f"{path}.{key}")
+
+
+def _assert_safe_ref(client: WeaveClient, ref_uri: str, label: str) -> None:
+    """Read the raw object for a ref and reject it if it contains unsafe CustomWeaveType payloads."""
+    ref = Ref.parse_uri(ref_uri)
```
andrewtruong: nit: not from this PR but it seems like this should be `Ref.from_uri`

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
andrewtruong: I'm not sure I understand your comment

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
gtarpenning: fixed

gtarpenning: done

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
gtarpenning: fixed

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
gtarpenning: This pr implements a walk of refs in the eval object to ensure safety. I was concerned by the additional overhead, because for deeply nested objects this might be hundreds of additional refs. but actually evaluation objects aren't structured like this, i was thinking of an evaluation trace. the object should just have one level of nesting. either way we only do a shallow walk, adding literally two object read requests.

andrewtruong (approved):

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
jtschoonhoven: +1

PR #6624, opened 2026-04-15
chore(weave): reduce Python 3.13 deprecation warnings

gtarpenning:
## Summary
- replace repo-owned Pydantic patterns that now warn on Python 3.13 / Pydantic 2.11+
- keep behavior the same while switching instance-level `model_fields` access to class-level access
- stop introspecting deprecated Pydantic compatibility attributes in `ObjectRecord`

## Why
`weave-trace-tests` started emitting a large warnings summary in CI. A subset of that noise was coming from `weave-public` code that still used older Pydantic APIs. These changes are the mechanical fixes on our side before deciding what to do about the remaining third-party warnings.

## Notes
- this branch contains the 5 remaining cleanup hunks that were still needed on current `master`
- instance reads like `self.model_fields` / `obj.model_fields` are switched to `type(self).model_fields` / `type(obj).model_fields`
- `ObjSchemaForInsert.model_post_init` now reads the deprecated compatibility field from `__dict__` so we preserve backward compatibility without triggering the deprecation
- `ObjectRecord.getmembers()` now skips deprecated Pydantic compatibility members when the target is a `BaseModel`

---

<a href="https://app.devin.ai/review/wandb/weave/pull/6624" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/weave/trace/object_record.py
+++ b/weave/trace/object_record.py
@@ -11,6 +11,15 @@ from pydantic import BaseModel
 from weave.shared.pydantic_util import pydantic_asdict_one_level
 from weave.trace.op import is_op
 
+_PYDANTIC_DEPRECATED_MEMBER_NAMES = frozenset(
+    {
+        "__fields__",
+        "__fields_set__",
+        "model_computed_fields",
+        "model_fields",
+    }
+)
+
 
 class ObjectRecord:  # noqa: PLW1641
     _class_name: str
@@ -105,6 +114,10 @@ def getmembers(
     results = []
     processed = set()
     names = dir(object)
+    if isinstance(object, BaseModel):
+        names = [
+            name for name in names if name not in _PYDANTIC_DEPRECATED_MEMBER_NAMES
+        ]
     # :dd any DynamicClassAttributes to the list of names if object is a class;
     # this may result in duplicate entries if, for example, a virtual
     # attribute with the same name as a DynamicClassAttribute exists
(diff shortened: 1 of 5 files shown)


andrewtruong:
We don't support pydantic v1 anymore, we should just update these callsites

weave/trace/object_record.py
```diff
@@ -11,6 +11,15 @@
...
 
+_PYDANTIC_DEPRECATED_MEMBER_NAMES = frozenset(
+    {
+        "__fields__",
+        "__fields_set__",
+        "model_computed_fields",
+        "model_fields",
+    }
+)
+
```
andrewtruong, line 22: this can be removed now, right?

weave/trace_server/trace_server_interface.py
```diff
@@ -321,8 +321,9 @@ class ObjSchemaForInsert(BaseModel):
 
     def model_post_init(self, context: Any, /) -> None:
         # If set_base_object_class is provided, use it to set builtin_object_class for backwards compatibility
-        if self.set_base_object_class is not None and self.builtin_object_class is None:
-            self.builtin_object_class = self.set_base_object_class
+        set_base_object_class = self.__dict__.get("set_base_object_class")
+        if set_base_object_class is not None and self.builtin_object_class is None:
+            self.builtin_object_class = set_base_object_class
```
andrewtruong, line 326: What is this used for?

PR #6627, opened 2026-04-16
chore(weave): Make WAL compatible with multi-client setup

andrewtruong:


rgao-coreweave:
**Question on the HMAC approach vs plain SHA-256:**

The HMAC secret exists to prevent brute-forcing API keys from directory names. But the `.hmac_key` file lives in the same `~/.weave/wal/` directory as the WAL data — so an attacker with read access to the directory names also has read access to the secret, which lets them brute-force the HMAC anyway.

Given that the threat model doesn't separate the secret from the data it protects, would a plain `sha256(api_key)` be sufficient here? It would give you the same two properties you need (deterministic + non-reversible for casual observation) while avoiding the `.hmac_key` file entirely — which eliminates the race condition during first-time key creation, the orphaning risk if the key file gets corrupted, and the multi-site monkeypatching needed in tests.

Not a blocker — just wondering if the extra machinery is justified given where the secret is stored.

rgao-coreweave (approved):
LGTM, but @chance-wnb has to approve the new HMAC dependency.

tests/durability/test_wal_client_id.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def test_same_api_key_gets_same_directory(self, tmp_path):
+        """Same API key always resolves to the same WAL directory."""
+        mgr_a = WALManager("entity", "project", api_key="wk-key-aaa")
+        mgr_b = WALManager("entity", "project", api_key="wk-key-aaa")
+
+        assert mgr_a.wal_dir == mgr_b.wal_dir
+
+
+class TestCrashRecovery:
```
chance-wnb, line 63: nit: maybe this test suite should go to a separate dedicated file. The test scenarios test many things other than `compute_client_id`.

chance-wnb (approved):
Thanks for the PR! Approved with one nit.

PR #6637, opened 2026-04-17
feat(weave): Add GEPA integration

andrewtruong:
Adds a preliminary GEPA integration to automatically log traces to weave


weave/integrations/gepa/gepa_sdk.py
```diff
@@ -0,0 +1,146 @@
...
+        return False
+    params = sig.parameters
+    if "callbacks" in params:
+        return True
+    return any(p.kind is inspect.Parameter.VAR_KEYWORD for p in params.values())
+
+
+def _inject_weave_callback(fn: Callable, kwargs: dict[str, Any]) -> None:
+    """Add a WeaveGEPACallback to the `callbacks` kwarg if one is not present."""
+    if not _fn_accepts_callbacks(fn):
```
chance-wnb: nit: this heuristic checking seems an overkill. I think we are only patching 3 methods. so you can design a parameter to _wrap_optimize().

```
_wrap_optimize(optimize_settings, should_inject_callback=True),
```

isn't this easier to do among lines of 126 -142 ? (human decides, code is simplified, no surprises) 

Nitpick, not a big deal.

tests/integrations/gepa/gepa_test.py
```diff
@@ -0,0 +1,537 @@
...
+        reflection_lm=_stub_reflection_lm,
+        max_metric_calls=8,
+        display_progress_bar=False,
+        raise_on_exception=False,
+        skip_perfect_score=False,
+    )
+
+    # Query the published `gepa_candidate` object via the trace server to
+    # confirm at least two versions exist (seed + one accepted candidate).
+    from weave.trace_server.trace_server_interface import (
```
chance-wnb: Shall we put imports to the beginning of the file?

tests/integrations/gepa/gepa_test.py
```diff
@@ -0,0 +1,537 @@
...
+        if d == propose_depth + 1
+        and all(names[j][1] > propose_depth for j in range(propose_idx + 1, i + 1))
+    ]
+    assert "short_reflection_lm" in propose_children, (
+        f"reflection LM should nest under gepa.propose; "
+        f"propose children={propose_children!r}"
+    )
+
+    # At least two `gepa_candidate` versions: seed and the accepted shorter one.
+    from weave.trace_server.trace_server_interface import (
```
chance-wnb: Shall we put imports to the beginning of the file?

tests/integrations/gepa/gepa_test.py
```diff
@@ -0,0 +1,537 @@
...
+@pytest.mark.disable_logging_error_check(reason="gepa logs its own error traceback")
+def test_gepa_optimize_records_errors(client: WeaveClient) -> None:
+    """GEPA's engine dispatches `on_error` when an iteration raises.
+    Exceptions escaping the adapter's `evaluate()` propagate out of the
+    reflective-mutation proposer (after the seed eval) and reach the engine's
+    `on_error` dispatch. We should emit a `gepa.error` span with the
+    exception type/message so users can count and inspect failures.
+    """
+
+    class _ExplodingAdapter(_StubAdapter):
```
chance-wnb: 1. This test case has a lot of similarities with test_gepa_optimize_records_multiple_errors . Maybe test_gepa_optimize_records_multiple_errors could cover the test scenario of this test case. 

2. I believe in you define the _FlakyAdapter outside of the test function, so it is somewhat inconsistent.

tests/integrations/gepa/gepa_test.py
```diff
@@ -0,0 +1,537 @@
...
+    )
+
+    # Query the published `gepa_candidate` object via the trace server to
+    # confirm at least two versions exist (seed + one accepted candidate).
+    from weave.trace_server.trace_server_interface import (
+        ObjectVersionFilter,
+        ObjQueryReq,
+    )
+
+    resp = client.server.objs_query(
```
chance-wnb: does this duplicate with the logic around line 449? maybe this test can fold into test_gepa_optimize_end_to_end_with_reflection_lm. 

Your call.

chance-wnb (approved):
Great work! Thanks!

tests/integrations/gepa/gepa_test.py
rgao-coreweave: +1

rgao-coreweave (approved):
amazing work!

LGTM

weave/integrations/gepa/gepa_callback.py
```diff
@@ -0,0 +1,360 @@
...
+            "gepa.iteration",
+            inputs={"iteration": iteration},
+            display_name=f"gepa.iteration.{iteration}",
+        )
+
+    def on_iteration_end(self, event: dict[str, Any]) -> None:
+        gc = weave_client_context.require_weave_client()
+        if self._iteration_call is not None:
+            gc.finish_call(
+                self._iteration_call,
```
rgao-coreweave: **[Should Fix]** `gc.finish_call` raising leaks the lifecycle Call reference

If `gc.finish_call` raises (transient trace-server hiccup, network blip), `self._iteration_call = None` never runs. GEPA's `notify_callbacks` wraps each callback invocation in `try/except` and only logs a warning — so the exception is silently swallowed. The next `on_iteration_start` then overwrites the still-set `_iteration_call`, and the previous span is leaked permanently (stays open in the trace).

`on_error` already does this correctly (lines 296–305): the `setattr(self, attr, None)` runs unconditionally inside an `except` wrapper. The three `on_*_end` paths should mirror that.

Suggestion:
```python
if self._iteration_call is not None:
    try:
        gc.finish_call(self._iteration_call, {...})
    finally:
        self._iteration_call = None
```

Apply the same try/finally to `on_evaluation_end` (line 163) and `on_proposal_end` (line 223).

PR #6639, opened 2026-04-17
chore(weave): fix flaky test_basic_evaluation on Windows

gtarpenning:
## Summary

`test_basic_evaluation` indexed into `client.get_calls()` positionally. `get_calls()` sorts by `(started_at, id)`; on Windows `time.time()` has ~15ms resolution, so a parent op and its immediate child can share a millisecond bucket — the UUIDv7 id tiebreak then falls to the random suffix and child can sort before parent. Restructured the assertions to match calls by `op_name` and child relationships by `parent_id`, so sibling ordering within a millisecond no longer matters. Added `@pytest.mark.flaky(reruns=3)` as a safety net.

Original failure ([Windows Trace nox tests (3, 11, flow)](https://github.com/wandb/weave/actions/runs/24575214484/job/71858621172)):
```
AssertionError: assert 'Model.predict' == 'Evaluation.predict_and_score'
```

## Testing

Test-only change. Ran `tests/flow/test_evaluation_imperative.py::test_basic_evaluation` locally (passes). Ruff clean.

### DIFF
--- a/tests/flow/test_evaluation_imperative.py
+++ b/tests/flow/test_evaluation_imperative.py
@@ -1,6 +1,7 @@
 import asyncio
 import inspect
 import json
+from collections import defaultdict
 from collections.abc import Callable
 from typing import TypedDict
 
@@ -36,7 +37,6 @@ def user_model():
     return func
 
 
-@pytest.mark.flaky(reruns=3)
 def test_basic_evaluation(
     client, user_dataset: list[ExampleRow], user_model: Callable[[int, int], int]
 ):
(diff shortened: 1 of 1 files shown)


tests/flow/test_evaluation_imperative.py
```diff
@@ -60,11 +61,25 @@ def test_basic_evaluation(
...
 
-    evaluate_call = calls[0]
-    assert op_name_from_call(evaluate_call) == "Evaluation.evaluate"
+    # Group by op_name and match children by parent_id instead of positional
+    # indexing. get_calls() sorts by (started_at, id); on Windows time.time()
+    # has ~15ms resolution so a parent op and its immediate child can share a
+    # millisecond, and the UUIDv7 tiebreak is random within that bucket.
+    by_op: dict[str, list] = {}
+    for c in calls:
+        by_op.setdefault(op_name_from_call(c), []).append(c)
```
jtschoonhoven: Tiny nit: I'd use `defaultdict` for this

```suggestion
    by_op: dict[str, list] = defaultdict(list)
    for c in calls:
        by_op[op_name_from_call(c)].append(c)
```

tests/flow/test_evaluation_imperative.py
```diff
@@ -36,6 +36,7 @@ def func(a: int, b: int) -> int:
     return func
 
 
+@pytest.mark.flaky(reruns=3)
```
jtschoonhoven: Is there still a reason to think this is helpful? If the test is fixed then let's just drop it.

jtschoonhoven (approved):
Sic

tests/flow/test_evaluation_imperative.py
gtarpenning: fixed

PR #6641, opened 2026-04-17
chore(weave): mark obj tests flaky

gtarpenning:
## Summary

Race between `weave.publish()` returning and the `WeaveList` becoming queryable on ClickHouse caused `NotFoundError` in CI. Mirrors the existing `time.sleep(0.2) + @pytest.mark.flaky(reruns=3)` pattern used for similar publish→get races.

Also preemptively covers 5 sibling tests in the same file that share the exact publish-nested-container → get shape and are structurally at the same risk.

Original failure ([job](https://github.com/wandb/weave/actions/runs/24582238190/job/71882800455?pr=6639)):
```
NotFoundError: Obj WeaveList:huU6RuytfnYediLjJV0lg6BgGH8ElEngI76SKuXxyXo not found
```

## Testing

Test-only change.

### DIFF
--- a/tests/trace/test_weave_client_mutations.py
+++ b/tests/trace/test_weave_client_mutations.py
@@ -1,3 +1,5 @@
+import time
+
 import pytest
 from pydantic import Field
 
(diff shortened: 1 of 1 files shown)


gtarpenning:

PR #6642, opened 2026-04-17
perf(weave): avoid per-request trace payload copies

gtarpenning:
## Description

This PR reduces avoidable memory churn in the trace-server `call_start` / `call_end` hot path.

### Problem

Large trace payloads were being copied multiple times per request before insertion:

- ref conversion round-tripped request models through `model_dump(...)` and `model_validate(...)`
- base64 content conversion rebuilt every visited dict/list, even when nothing changed
- ClickHouse row building dumped insert models to dicts just to read the fields back out

This is not a leak, but under concurrency spikes it increases peak heap usage in proportion to request size.

### What changed

- Ref conversion now uses a copy-on-write traversal for supported payload trees.
  - unchanged `BaseModel` / dict / list / tuple / set branches are reused
  - only paths containing rewritten refs are rebuilt
  - if a model contains unsupported objects in an `Any` payload, the converter automatically falls back to the legacy `model_dump` / `model_validate` path to preserve existing behavior
- Base64 content conversion now uses the same copy-on-write behavior.
  - branches with no converted content are returned as-is
- ClickHouse insert-row construction now reads model attributes directly instead of allocating intermediate dicts.

### Why this is safe

The behavior is intentionally unchanged:

- the same values are still traversed and the same scalar ref-mapping logic is applied
- aliased query models still serialize correctly with `by_alias=True`
- models that relied on the old serialization round-trip still get it automatically when needed
- excluded fields are still skipped
- ClickHouse row ordering and sentinel handling are preserved

### DIFF
--- a/tests/trace/test_errors.py
+++ b/tests/trace/test_errors.py
@@ -1,0 +1,52 @@
+import httpx
+
+from weave.trace.errors import format_http_error, response_error_message
+
+
+def test_response_error_message_prefers_common_json_fields_and_falls_back_to_body():
+    request = httpx.Request("POST", "https://trace.wandb.ai/obj/read")
+
+    # Structured API errors use different message field names.
+    assert (
+        response_error_message(
+            httpx.Response(
+                403,
+                json={"reason": "reason", "detail": "detail", "message": "message"},
+                request=request,
+            )
+        )
+        == "reason"
+    )
+    assert (
+        response_error_message(
+            httpx.Response(403, json={"detail": "Project not found"}, request=request)
+        )
+        == "Project not found"
+    )
+    assert (
+        response_error_message(
+            httpx.Response(403, json={"message": "Forbidden"}, request=request)
+        )
+        == "Forbidden"
+    )
+
+    # Unstructured and empty responses preserve useful diagnostic context.
+    assert (
+        response_error_message(
+            httpx.Response(502, content=b"upstream unavailable", request=request)
+        )
+        == "upstream unavailable"
+    )
+    assert (
+        response_error_message(httpx.Response(500, request=request))
+        == "<empty response body>"
+    )
+
+    assert format_http_error(
+        httpx.Response(403, json={"detail": "Project not found"}, request=request),
+        "Unable to read object for ref uri: weave:///test/test-project/object/frozen-dataset:abc123",
+    ) == (
+        "Unable to read object for ref uri: "
+        "weave:///test/test-project/object/frozen-dataset:abc123 "
+        "(status 403): Project not found"
+    )
--- a/weave/trace/errors.py
+++ b/weave/trace/errors.py
@@ -1,0 +1,27 @@
+"""Shared helpers for client-side error reporting."""
+
+from typing import Any
+
+import httpx
+
+
+def response_error_message(response: httpx.Response) -> str:
+    """Extract the most actionable message from an HTTP error response."""
+    body: Any = response.text or "<empty response body>"
+    try:
+        parsed_body = response.json()
+    except (UnicodeDecodeError, ValueError):
+        return str(body)
+
+    if isinstance(parsed_body, dict):
+        for key in ("reason", "detail", "message"):
+            if message := parsed_body.get(key):
+                return str(message)
+    return str(parsed_body)
+
+
+def format_http_error(response: httpx.Response, context: str) -> str:
+    """Add request context and status to an HTTP error response message."""
+    return (
+        f"{context} (status {response.status_code}): {response_error_message(response)}"
+    )
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import copy
 import dataclasses
 import datetime
-import json
 import logging
 import os
 import re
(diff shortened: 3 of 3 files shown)


weave/trace_server/base64_content_conversion.py
```diff
@@ -125,12 +125,23 @@ def replace_base64_with_content_objects(
...
         elif isinstance(val, list):
-            return [_visit(v) for v in val]
+            updated_list: list[Any] | None = None
+            for index, child in enumerate(val):
+                visited_child = _visit(child)
+                if visited_child is not child:
+                    if updated_list is None:
+                        updated_list = list(val)
+                    updated_list[index] = visited_child
+            return val if updated_list is None else updated_list
```
neutralino1: Can you add a comment to explain why the added apparent complexity? Just say it's to avoid unnecessary copies or such...

weave/trace_server/trace_server_converter.py
```diff
@@ -1,5 +1,6 @@
+import dataclasses
```
neutralino1, line 24: Same throughout this file.

weave/trace_server/base64_content_conversion.py
gtarpenning: yeah the reason is in-place edits vs fully memory re-assignment

gtarpenning: will add comment

weave/trace_server/trace_server_converter.py
```diff
@@ -150,20 +151,115 @@ def mapper(obj: D) -> D:
...
+    new_obj = func(obj)
+    if new_obj is not obj:
+        return new_obj, True, True
+    return new_obj, False, not dataclasses.is_dataclass(obj)
+
+
+def _map_model_values(
+    obj: BaseModel,
+    func: Callable[[Any], Any],
+) -> tuple[BaseModel, bool, bool]:
```
jtschoonhoven: This is complicated enough to deserve a docstring and probably some more comments so it's easier to follow.

weave/trace_server/base64_content_conversion.py
```diff
@@ -125,12 +125,23 @@ def replace_base64_with_content_objects(
...
-                result[k] = _visit(v)
-            return result
+            updated_dict: dict[Any, Any] | None = None
+            for key, child in val.items():
+                visited_child = _visit(child)
+                if visited_child is not child:
+                    if updated_dict is None:
+                        updated_dict = dict(val)
+                    updated_dict[key] = visited_child
+            return val if updated_dict is None else updated_dict
```
jtschoonhoven: Looks good to me, but you should add a comment explaining the copy-on-write thing, otherwise it's not obvious why it needs to be so complicated

jtschoonhoven (commented):
Mostly looks good, but the changes in  `weave/trace_server/trace_server_converter.py` are hard to follow. I'd like to see that file cleaned up a little.

weave/trace_server/base64_content_conversion.py
gtarpenning: agreed, Emmanuel said the same thing.

weave/trace_server/base64_content_conversion.py
```diff
@@ -125,12 +128,26 @@ def replace_base64_with_content_objects(
 
     def _visit(val: Any) -> Any:
         if isinstance(val, dict):
-            result = {}
-            for k, v in val.items():
-                result[k] = _visit(v)
-            return result
+            # Clone on first change so no-op requests can reuse the original
+            # payload.
```
jtschoonhoven: Super nit: I hate that "payload" is on its own line, this will still pass lint checks on one line

weave/trace_server/trace_server_converter.py
```diff
@@ -150,20 +153,133 @@ def mapper(obj: D) -> D:
 
 
 def _map_values(obj: E, func: Callable[[E], E]) -> E:
+    """Recursively apply ``func`` while preserving untouched object identity."""
```
jtschoonhoven: Tiny nit: Claude says the \``double backticks`` are a Sphinx convention, which I'm pretty sure we don't use. I'd search-and-replace them back to single ticks

weave/trace_server/trace_server_converter.py
```diff
@@ -150,20 +153,133 @@ def mapper(obj: D) -> D:
...
+    new_obj, _, _ = _map_values_copy_on_write(obj, func)
+    return cast(E, new_obj)
+
+
+def _map_values_copy_on_write(
+    obj: E,
+    func: Callable[[E], E],
+) -> tuple[E, bool, bool]:
```
jtschoonhoven: Nit: consider returning a `NamedTuple` or similar so that you don't have to keep "value, changed, fully_supported" in your head

weave/trace_server/trace_server_converter.py
```diff
@@ -150,20 +153,133 @@ def mapper(obj: D) -> D:
 
 
 def _map_values(obj: E, func: Callable[[E], E]) -> E:
+    """Recursively apply ``func`` while preserving untouched object identity."""
+    new_obj, _, _ = _map_values_copy_on_write(obj, func)
+    return cast(E, new_obj)
```
jtschoonhoven: Any idea why you have to use `cast` here to make mypy happy? I would have thought that mypy could follow this on its own. Not important, I'm just curious.

jtschoonhoven (commented):
I have to head to the airport in a minute so I might not be able to finish reviewing today. Left a few nits so far.

The logic gets pretty hairy. Personally I'm willing to sacrifice some memory/CPU for clean, maintainable code. I need some more convincing that this added complexity is worth it.

PR #6644, opened 2026-04-20
feat(weave): auto opt in calls_complete on wandb-entity projects

gtarpenning:
https://coreweave.atlassian.net/browse/WB-33669

Pivots from "flip `WEAVE_USE_CALLS_COMPLETE` default to true" to an entity-based dogfood gate. Entities in `CALLS_COMPLETE_ENTITY_ALLOWLIST` (currently `{"wandb"}`) auto-opt-in to the calls_complete write path regardless of the env var; everyone else stays on the legacy default.

validate calls_complete on wandb/* traffic first. Flipping the env default remains a follow-up once we have more evidence that its fully functional.

Entity is threaded from `weave.init()` → `init_weave_get_server` → `RemoteHTTPTraceServer`. `StainlessRemoteHTTPTraceServer` does not use this path and is unaffected.

### DIFF
--- a/weave/trace/settings.py
+++ b/weave/trace/settings.py
@@ -20,6 +20,12 @@ from pydantic import BaseModel, ConfigDict, PrivateAttr
 DEFAULT_RETRY_MAX_INTERVAL_SECONDS = 60 * 5  # 5 minutes
 SETTINGS_PREFIX = "WEAVE_"
 
+# Entities that auto-opt-in to the calls_complete write path, independent of
+# the WEAVE_USE_CALLS_COMPLETE env var. This lets us dogfood calls_complete on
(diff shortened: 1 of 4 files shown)


weave/trace/settings.py
```diff
@@ -214,16 +214,16 @@ class UserSettings(BaseModel):
     Can be overridden with the environment variable `WEAVE_USE_STAINLESS_SERVER`
     """
 
-    use_calls_complete: bool = False
+    use_calls_complete: bool = True
```
andrewtruong: Wont this immediately break users logging to an existing non-calls-complete project?

gtarpenning: no, the complete processor logs to calls_merged projects too.

andrewtruong (approved):

PR #5864, opened 2025-12-18
fix(weave): actually support distributed replicated clickhouse tables

gtarpenning:
## Description

[WB-29945](https://wandb.atlassian.net/browse/WB-29945)

Enables proper distributed table support for replicated ClickHouse deployments by automatically creating distributed tables on top of replicated local tables during migrations. This allows queries to be routed correctly across the cluster rather than hitting individual replicas, no changes needed on the query side. 

Follows implementation explained in the docs here: https://clickhouse.com/docs/engines/table-engines/special/distributed

This PR:
- Adds environment variable helpers for ClickHouse replication settings (`WF_CLICKHOUSE_REPLICATED`, `WF_CLICKHOUSE_REPLICATED_PATH`, `WF_CLICKHOUSE_REPLICATED_CLUSTER`, `WF_CLICKHOUSE_USE_DISTRIBUTED_TABLES`)
- Passes replication configuration from environment to migrator (previously wasn't wired up, we use `./migrate_local` but still we should do this)
- Implements special handling for ALTER/CREATE/DROP operations (indexes only on local, regular columns on both)
- Adds materialized view transformation using DROP/CREATE pattern with `_local` suffix routing
- Skips unsupported operations (MATERIALIZE commands) in distributed mode
- Refactors the migrator, 
        - push methods that just generate sql out of class. Now, all class methods actually execute sql. 
        - all SQL regex is in its own helper class
        - split Base, Replicated, Distributed handling into their own classes with hierarchy 
- add distributed properties to the clickhouse_batched class
- Add comprehensive tests for distributed table creation and replicated mode behavior
- Fixed potential bug where passing an explicit ReplicatedMergeTree engine type still gets "Replicated" prepended. 

## Testing

updates existing tests and adds new ones

Local testing:

```bash
> WF_CLICKHOUSE_REPLICATED_CLUSTER=cluster_2S_2R WF_CLICKHOUSE_REPLICATED=true WF_CLICKHOUSE_USE_DISTRIBUTED_TABLES=true make migrate_local
./migrate_local.sh
INFO:weave.trace_server.clickhouse_trace_server_migrator:ClickHouseTraceServerMigrator initialized with: replicated=True, use_distributed=True, replicated_cluster=cluster_2S_2R, replicated_path=/clickhouse/tables/{db}, management_db=db_management
INFO:weave.trace_server.clickhouse_trace_server_migrator:`default` migration status: {'db_name': 'default', 'curr_version': 0, 'partially_applied_version': None}
[...]

PR #6648, opened 2026-04-20
chore(weave): add 1s3r and 2s2r migrator CI jobs with path filters

amwarrier:
## What

  Adds two new pre-merge CI jobs that run the existing migrator functional suite against multi-replica ClickHouse topologies:

  - **`trace-server-migrator-1s3r`** — 1 shard × 3 replicas, embedded Keeper in `ch-s1-r1`.
  - **`trace-server-migrator-2s2r`** — 2 shards × 2 replicas, embedded Keeper in `ch-s1-r1`, distributed tables available for tests that opt in.

  The existing `trace-server-migrator-tests` job is renamed to `trace-server-migrator-1s1r` for clarity. Behavior is unchanged — still runs on every push.

  Both new jobs are path-filtered: they run only when a PR touches `weave/trace_server/migrations/`, `tests/trace_server_migrator/`, or `tests/trace_server/conftest_lib/`. They also run on every push to
  `master`, matching the broader pattern in this workflow.

  The `trace-tests-matrix-check` aggregator whitelists the two new jobs via `alls-green`'s `allowed-skips` so path-gated skips don't wedge merge on unrelated PRs.

  ## Why

  Prod runs ClickHouse Cloud at 2 shards × 3 replicas. CI today tests only at 1 shard × 1 replica. Between those two points, every replication-specific and distributed-specific code path has no pre-merge coverage anywhere in the pipeline.

  That gap is not theoretical. In the last ~14 months, 22+ fix PRs from @gtarpenning have addressed bugs invisible to 1s1r CI:

  - #6574  — `_local` suffix missing in ON CLUSTER mutations (needs 2s+)
  - #6503  — replica sync failure during migration (needs 1s3r+)
  - #6597  — replicated engines in replicated DBs (needs 1s3r+)
  - #6413  — distributed sharded DB engine misconfig (needs 2s+)
  - #6082  — distributed calls sharding key (needs 2s+)
  - #6335 , #6039 , #5968 , #5969 , #5864 , #5885  — similar classes

  Every one of these shipped to production and was caught post-merge.

  ## Design notes

  **Why 2s2r and not 2s3r (prod-exact).** GH Actions `ubuntu-latest` runners have ~7 GB RAM. The 3rd replica per shard only exercises Keeper quorum behavior, which has never been in our incident history. Dropping to 2
  replicas saves ~25% runner memory without losing any bug classes.
  Revisit if quorum ever becomes an incident surface.

  **Why 1 Keeper and not 3.** Keeper replica count and CH replica count are independent. The bug classes above all live in the CH protocol surface (replication, DDL, distributed routing), not in Keeper Raft behavior.
  A single Keeper catches everything we care about at a fraction of the complexity. Same revisit criterion.

  **Why path-filter to just `migrations/` and test dirs.** Per review feedback: the existing migrator functional suite only exercises end-to-end application of migration SQL. Gating on the migrator engine, schema, or environment files would over-trigger without adding coverage the tests actually validate.

  ## Local verification

  Both topologies were stood up via docker-compose locally (OrbStack on macOS) and verified end-to-end before push:

  - 1s3r: 4-node cluster (3 CH + embedded Keeper), all 3 replicas visible in
    `system.clusters` with zero errors, `ON CLUSTER` DDL propagates,
    `ReplicatedMergeTree` inserts replicate correctly. All 8 upstream
    migrator functional tests pass in 63s.
  - 2s2r: 4-node cluster, sharded + replicated, `Distributed` engine routes
    writes across both shards and reads fan out correctly. All 8 tests pass in 76s.
  - Memory footprint on both: ~2.7 GB peak across all nodes, well under
    the runner budget.

  ## Risk / rollout

  - New jobs are **not** wired into branch protection as individual required checks. They only appear via the aggregator (`trace-tests-matrix-check`).
  - Plan to run on master for ~1 week before proposing branch-protection changes that actually gate merge.
  - If CI flakes, the first lever is the `ubuntu-latest-8-cores` runner upgrade. Second lever is moving 2s2r to a nightly job. Neither is proposed in this PR.

  ## Follow-ups (out of scope)

  - **Regression tests encoding specific past bug classes** (Griffin's `_local` suffix fix, replica-sync-during-migration, distributed sharding key). Without these, the matrix validates syntax + end-to-end flow but does not assert topology-specific correctness. High priority.
  - **Nightly all-shards run** across a broader topology matrix, per reviewer suggestion. Catches indirect bugs from non-migrator PRs that path filters intentionally skip.
  - **Branch protection update** to require `trace-tests-matrix-check` (deferred until jobs are stable for a week).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -275,7 +275,6 @@ jobs:
         env:
           CI: 1
           WF_CLICKHOUSE_KEEPER_PORT: 8130
-          WF_CLICKHOUSE_TOPOLOGY: 1s1r
           DD_TRACE_ENABLED: false
         run: |
           nox -e "tests-${{ matrix.python-version-major }}.${{ matrix.python-version-minor }}(shard='trace_server_migrator')"
@@ -327,14 +326,11 @@ jobs:
           CI: 1
           # Existing fixture reads WF_CLICKHOUSE_KEEPER_PORT and connects
           # to localhost at that port — we map the 1s3r entrypoint to 8130.
-          WF_CLICKHOUSE_KEEPER_PORT: 8130
-          # Topology marker read by tests/trace_server_migrator/conftest.py
-          # hook that skips @pytest.mark.topology-gated tests. Other
-          # WF_CLICKHOUSE_* vars are intentionally NOT set here: the
+          # Other WF_CLICKHOUSE_* vars are intentionally NOT set: the
           # existing migrator tests pass replicated/distributed flags
           # explicitly to the migrator factory, so env vars would be
-          # misleading (unread) documentation.
-          WF_CLICKHOUSE_TOPOLOGY: 1s3r
+          # unread documentation.
+          WF_CLICKHOUSE_KEEPER_PORT: 8130
           DD_TRACE_ENABLED: false
         run: |
           nox -e "tests-${{ matrix.python-version-major }}.${{ matrix.python-version-minor }}(shard='trace_server_migrator')"
@@ -397,10 +393,8 @@ jobs:
         env:
           CI: 1
           # 2s2r entrypoint is on host port 8131 (1s3r uses 8130).
+          # See 1s3r job for why only KEEPER_PORT is set.
           WF_CLICKHOUSE_KEEPER_PORT: 8131
-          # See 1s3r job for why only TOPOLOGY is set; other
-          # WF_CLICKHOUSE_* vars would be unread today.
-          WF_CLICKHOUSE_TOPOLOGY: 2s2r
           DD_TRACE_ENABLED: false
         run: |
           nox -e "tests-${{ matrix.python-version-major }}.${{ matrix.python-version-minor }}(shard='trace_server_migrator')"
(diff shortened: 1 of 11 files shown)


amwarrier (commented):

tests/trace_server/conftest_lib/keeper_1s3r.xml
```diff
@@ -0,0 +1,67 @@
+<clickhouse>
+    <!--
+      Cluster config for 1-shard, 3-replica topology.
```
gtarpenning, line 3: i think technically we might only need 1s2r but this is fine

.github/workflows/test.yaml
```diff
@@ -275,15 +275,154 @@ jobs:
         env:
           CI: 1
           WF_CLICKHOUSE_KEEPER_PORT: 8130
+          WF_CLICKHOUSE_TOPOLOGY: 1s1r
```
gtarpenning: where does this get used?

gtarpenning (approved):
update pr description when you have a chance, and maybe consider only running these tests if there are any changes to the migrations, as that's all we test. 

also this would be great as a nightly, but on all shards (obviously out of scope)

.github/workflows/test.yaml
amwarrier: I added this for being used in the future for regression tests. Will remove it. YAGNI.

tests/trace_server/conftest_lib/keeper_1s3r.xml
amwarrier: 👍

amwarrier:
recheck

amwarrier (commented):

amwarrier:

PR #6578, opened 2026-04-10
chore(weave): consolidate test_obj_tags_aliases.py from 127 to 29 tests

gtarpenning:
## Summary
- Consolidates 127 test functions (142 cases with parametrize) into 29 functions (42 cases)
- Groups tests by functionality with denser assertions per test
- No coverage lost — all original behaviors preserved
- File reduced from 2780 to 1744 lines

Was taking a ton of time in CI. 

## Grouping
- **Validation** (3): tag format rules, alias format rules, digest_is_content_hash
- **Server CRUD** (2): tag lifecycle, alias lifecycle
- **Server errors** (4): nonexistent/deleted objects (separate functions due to SQLite transaction state)
- **Server enrichment/read** (2): enrichment toggle, obj_read digest types
- **Server filtering** (3): by tags, by aliases, combined
- **Server isolation** (3): version/object isolation, batch enrichment, list endpoints
- **SDK** (5): tag lifecycle, alias lifecycle, URI strings, error cases
- **Top-level weave.\*** (2): tag functions, alias functions
- **Integration** (5): alias resolution, full lifecycle, publish-time tags/aliases, cross-project isolation, deletion cascades

PR #6649, opened 2026-04-20
fix(weave): retry on ch read paths for individual objs

gtarpenning:
## Summary

- Inlines the existing `_read_with_retry` (3 attempts, 50ms→1s backoff) into the public `obj_read`, `file_content_read`, `obj_delete`, and `op_delete` methods. Previously only a handful of internal create-then-read call sites knew to go through the `_obj_read_with_retry` / `_file_content_read_with_retry` wrappers — direct callers got nothing. Wrappers removed.
- `obj_delete` / `op_delete` do a read-select before inserting the soft-delete tombstones, so they hit the same replica-lag race as reads and now retry the same way.
- Adds `@ddtrace.tracer.wrap` on the four public methods (replaces spans lost when the wrappers were deleted) and tags the active span with `clickhouse.read_retry.attempt` via a tenacity `before_sleep` callback, so retries are queryable in DD.

## Flake context

Flake-finder on master picked up 10+ tests hitting the same read-after-write race against ClickHouse replica lag. Previous per-test patches (#6431, #6423, #6469, #6514, #6599, #6578, #6625) chased symptoms. Examples this targets:

- `test_content_in_dataset`, `test_images_in_dataset_without_client`, `test_large_files`, `test_large_file_migration` — `File with digest ... not found`
- `test_should_capture_code_env`, `test_gotten_object_method_is_callable_with_call_func`, `test_list_mutation_saving_nested_dicts`, `test_op_no_repeats`, `test_refs_read_batch_with_extra` — `Obj ... not found`
- `test_read_deleted_op` — `Object ... not found when deleting`

Scope is narrow on purpose: only identity-fetch methods that raise `NotFoundError` on miss. Query/list methods are untouched — empty is a valid result there.

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -1875,9 +1875,13 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         object_query_builder.set_include_deleted(include_deleted=True)
         metadata_only = req.metadata_only or False
 
-        objs = self._select_objs_query(object_query_builder, metadata_only)
-        if len(objs) == 0:
-            raise NotFoundError(f"Obj {req.object_id}:{req.digest} not found")
+        def _query_objs() -> list[SelectableCHObjSchema]:
+            results = self._select_objs_query(object_query_builder, metadata_only)
+            if len(results) == 0:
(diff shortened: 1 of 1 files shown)


gtarpenning:
do this client side: https://github.com/wandb/weave/pull/6653

PR #6651, opened 2026-04-20
chore(weave): inline CHValidationError into errors.py

gtarpenning:
## Summary

- Moves `CHValidationError` from `weave/trace_server/validation_util.py` into `weave/trace_server/errors.py`
- `validation_util.py` re-exports it, so every existing import path keeps working
- `errors.py` no longer imports from `validation_util`, so it is now a self-contained public contract module (no cross-file deps inside `weave/trace_server/` except the public sibling set and `weave.shared`)

## Why

Prep work for a planned refactor that splits `weave/trace_server/` into a small frozen public contract (schemas, errors, ids, query AST, builtin object classes) and a closed server implementation. `errors.py` is part of that public contract and was the only file in the set still reaching into a would-be-closed module. Inlining the one-line class here removes that last edge so the contract can be enforced with a lint guard in a follow-up.

Zero behavior change: identical exception class (same name, same bases), `validation_util.CHValidationError is errors.CHValidationError`, and the error registry still maps it to HTTP 400.

## Test plan

- [x] `uvx ruff check` passes on both edited files
- [x] `pytest --collect-only tests/trace_server/test_clickhouse_batching.py tests/trace/test_client_trace.py` collects cleanly (139 tests)
- [x] Smoke-tested `require_otel_trace_id`, `require_otel_span_id`, `require_uuid`, `require_base64`, `require_max_str_len` each raise `CHValidationError` on bad input
- [x] Smoke-tested `handle_server_exception(CHValidationError(...))` still returns HTTP 400
- [x] `validation_util.CHValidationError is errors.CHValidationError` holds
- [ ] CI green

### DIFF
--- a/weave/trace_server/errors.py
+++ b/weave/trace_server/errors.py
@@ -7,8 +7,6 @@ from typing import Any, Optional
 import httpx
 from gql.transport.exceptions import TransportQueryError, TransportServerError
 
-from weave.trace_server.validation_util import CHValidationError
-
 # =============================================================================
 # Error Codes - Machine-readable codes for client-side error detection
 # =============================================================================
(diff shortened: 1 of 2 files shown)


gtarpenning:
punt

PR #6653, opened 2026-04-20
fix(weave): client-side retry on replica-lag 404s for reads

gtarpenning:
## Summary

Adds a narrow, client-side retry for 404s on `obj_read` and `file_content_read` so that eventual-consistency races between a write and a subsequent read don't surface as errors. Replaces the server-side approach in #6649 (closed).

- Retry 1 time (2 attempts total), 1s wait, only on HTTP 404.
- Skip retry when the response body carries `deleted_at` -> authoritative `ObjectDeletedError`, not a race.
- `not_found_retry_disabled()` context manager for callers that want a fast definitive 404 (e.g., existence probes).
- Applied to `RemoteHTTPTraceServer.obj_read` and `RemoteHTTPTraceServer.file_content_read`. Stainless binding untouched (separate migration).

## Why not #6649

That PR retried every call through `obj_read` / `file_content_read` / `obj_delete` / `op_delete` server-side, which meant legit 404s (genuinely missing object, existence probes, delete-time lookups) paid up to 1s of extra latency and three extra round-trips. Pushing the retry to the client lets us scope it to the flows that actually suffer without punishing every consumer of those endpoints.

## Test plan

- [x] `uv run --group test python -m pytest tests/trace_server_bindings/test_http_utils.py tests/trace_server_bindings/test_http_behavior_remote.py` -> 16/16 pass
- [x] `uvx ruff check weave/trace_server_bindings/ tests/trace_server_bindings/` -> clean
- [ ] CI green

## Flake check: `Trace nox tests (3, 11, trace)` x 10 reruns

Re-ran the shard 10 times in a row on top of the original attempt. All 11 runs passed identically (`831 passed, 105 skipped` SQLite + `917 passed, 19 skipped` ClickHouse, per run). No flakes observed.

| Attempt | Job | SQLite | ClickHouse | Failed | Errored |
|--------:|-----|-------:|-----------:|-------:|--------:|
| 1 (original) | [72226919719](https://github.com/wandb/weave/actions/runs/24695406734/job/72226919719) | 93.27s | 192.43s | 0 | 0 |
| 2 | [72228230873](https://github.com/wandb/weave/actions/runs/24695406734/attempts/2) | 88.80s | 191.81s | 0 | 0 |
| 3 | [72229053701](https://github.com/wandb/weave/actions/runs/24695406734/attempts/3) | 89.88s | 193.74s | 0 | 0 |
| 4 | [72229751349](https://github.com/wandb/weave/actions/runs/24695406734/attempts/4) | 92.77s | 184.15s | 0 | 0 |
| 5 | [72230440864](https://github.com/wandb/weave/actions/runs/24695406734/attempts/5) | 94.21s | 204.91s | 0 | 0 |
| 6 | [72231216841](https://github.com/wandb/weave/actions/runs/24695406734/attempts/6) | 90.38s | 184.82s | 0 | 0 |
| 7 | [72231881322](https://github.com/wandb/weave/actions/runs/24695406734/attempts/7) | 95.73s | 190.97s | 0 | 0 |
| 8 | [72232569557](https://github.com/wandb/weave/actions/runs/24695406734/attempts/8) | 92.45s | 185.71s | 0 | 0 |
| 9 | [72233256344](https://github.com/wandb/weave/actions/runs/24695406734/attempts/9) | 91.80s | 196.75s | 0 | 0 |
| 10 | [72233924465](https://github.com/wandb/weave/actions/runs/24695406734/attempts/10) | 90.96s | 190.18s | 0 | 0 |
| 11 | [72234591817](https://github.com/wandb/weave/actions/runs/24695406734/attempts/11) | 91.35s | 196.23s | 0 | 0 |

### DIFF
--- a/tests/trace_server_bindings/test_http_utils.py
+++ b/tests/trace_server_bindings/test_http_utils.py
@@ -5,10 +5,13 @@ from unittest.mock import Mock
 import httpx
 import pytest
 
+from weave.trace_server_bindings import http_utils
 from weave.trace_server_bindings.http_utils import (
     CallsCompleteModeRequired,
     handle_response_error,
+    not_found_retry_disabled,
     process_batch_with_retry,
+    retry_on_not_found,
 )
 
 
@@ -51,3 +54,62 @@ def test_calls_complete_mode_required_raises():
 
     with pytest.raises(CallsCompleteModeRequired, match="calls_complete mode required"):
         handle_response_error(response, "/call/upsert_batch")
+
+
+def _raise_404(body: dict) -> httpx.HTTPStatusError:
+    response = httpx.Response(
+        404,
+        json=body,
+        request=httpx.Request("POST", "http://example.com"),
+    )
+    return httpx.HTTPStatusError("404", request=response.request, response=response)
+
+
+def test_retry_on_not_found_behavior(monkeypatch):
+    """Retries a replica-lag 404 once, passes on ObjectDeletedError, ignores non-404."""
+    monkeypatch.setattr(http_utils, "NOT_FOUND_RETRY_WAIT_SECONDS", 0.0)
+    calls = {"missing": 0, "deleted": 0, "server": 0, "disabled": 0}
+
+    @retry_on_not_found
+    def flaky_missing():
+        calls["missing"] += 1
+        if calls["missing"] == 1:
+            raise _raise_404({"reason": "Obj foo:bar not found"})
+        return "ok"
+
+    @retry_on_not_found
+    def deleted():
+        calls["deleted"] += 1
+        raise _raise_404({"reason": "deleted", "deleted_at": "2024-01-01T00:00:00Z"})
+
+    @retry_on_not_found
+    def server_error():
+        calls["server"] += 1
+        response = httpx.Response(
+            500, request=httpx.Request("POST", "http://example.com")
+        )
+        raise httpx.HTTPStatusError("500", request=response.request, response=response)
+
+    @retry_on_not_found
+    def missing_but_disabled():
+        calls["disabled"] += 1
+        raise _raise_404({"reason": "not found"})
+
+    # Retries once, second attempt succeeds.
+    assert flaky_missing() == "ok"
+    assert calls["missing"] == 2
+
+    # ObjectDeletedError (body has `deleted_at`) is not retried.
+    with pytest.raises(httpx.HTTPStatusError):
+        deleted()
+    assert calls["deleted"] == 1
+
+    # Non-404s are not in scope for this decorator.
+    with pytest.raises(httpx.HTTPStatusError):
+        server_error()
+    assert calls["server"] == 1
+
+    # Context-scoped opt-out short-circuits retry.
+    with not_found_retry_disabled(), pytest.raises(httpx.HTTPStatusError):
+        missing_but_disabled()
+    assert calls["disabled"] == 1
--- a/tests/trace_server_bindings/test_http_utils.py
+++ b/tests/trace_server_bindings/test_http_utils.py
@@ -5,11 +5,9 @@ from unittest.mock import Mock
 import httpx
 import pytest
 
-from weave.trace_server_bindings import http_utils
 from weave.trace_server_bindings.http_utils import (
     CallsCompleteModeRequired,
     handle_response_error,
-    not_found_retry_disabled,
     process_batch_with_retry,
     retry_on_not_found,
 )
@@ -56,7 +54,7 @@ def test_calls_complete_mode_required_raises():
         handle_response_error(response, "/call/upsert_batch")
 
 
-def _raise_404(body: dict) -> httpx.HTTPStatusError:
+def _make_404(body: dict) -> httpx.HTTPStatusError:
     response = httpx.Response(
         404,
         json=body,
@@ -66,50 +64,26 @@ def _raise_404(body: dict) -> httpx.HTTPStatusError:
 
 
 def test_retry_on_not_found_behavior(monkeypatch):
-    """Retries a replica-lag 404 once, passes on ObjectDeletedError, ignores non-404."""
-    monkeypatch.setattr(http_utils, "NOT_FOUND_RETRY_WAIT_SECONDS", 0.0)
-    calls = {"missing": 0, "deleted": 0, "server": 0, "disabled": 0}
+    """Retries on a non-deleted 404; skips deleted 404s (same error class)."""
+    monkeypatch.setenv("WEAVE_RETRY_MAX_ATTEMPTS", "2")
+    monkeypatch.setenv("WEAVE_RETRY_MAX_INTERVAL", "0.01")
+    calls = {"missing": 0, "deleted": 0}
 
     @retry_on_not_found
     def flaky_missing():
         calls["missing"] += 1
         if calls["missing"] == 1:
-            raise _raise_404({"reason": "Obj foo:bar not found"})
+            raise _make_404({"reason": "Obj foo:bar not found"})
         return "ok"
 
     @retry_on_not_found
     def deleted():
         calls["deleted"] += 1
-        raise _raise_404({"reason": "deleted", "deleted_at": "2024-01-01T00:00:00Z"})
+        raise _make_404({"reason": "deleted", "deleted_at": "2024-01-01T00:00:00Z"})
 
-    @retry_on_not_found
-    def server_error():
-        calls["server"] += 1
-        response = httpx.Response(
-            500, request=httpx.Request("POST", "http://example.com")
-        )
-        raise httpx.HTTPStatusError("500", request=response.request, response=response)
-
-    @retry_on_not_found
-    def missing_but_disabled():
-        calls["disabled"] += 1
-        raise _raise_404({"reason": "not found"})
-
-    # Retries once, second attempt succeeds.
     assert flaky_missing() == "ok"
     assert calls["missing"] == 2
 
-    # ObjectDeletedError (body has `deleted_at`) is not retried.
     with pytest.raises(httpx.HTTPStatusError):
         deleted()
     assert calls["deleted"] == 1
-
-    # Non-404s are not in scope for this decorator.
-    with pytest.raises(httpx.HTTPStatusError):
-        server_error()
-    assert calls["server"] == 1
-
-    # Context-scoped opt-out short-circuits retry.
-    with not_found_retry_disabled(), pytest.raises(httpx.HTTPStatusError):
-        missing_but_disabled()
-    assert calls["disabled"] == 1
--- a/tests/trace_server_bindings/test_http_utils.py
+++ b/tests/trace_server_bindings/test_http_utils.py
@@ -5,11 +5,13 @@ from unittest.mock import Mock
 import httpx
 import pytest
 
+from weave.trace_server_bindings import http_utils
 from weave.trace_server_bindings.http_utils import (
     CallsCompleteModeRequired,
     handle_response_error,
     process_batch_with_retry,
     retry_on_not_found,
+    write_then_read_scope,
 )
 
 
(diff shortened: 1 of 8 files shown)


weave/trace_server_bindings/http_utils.py
```diff
@@ -21,8 +26,15 @@
 REMOTE_REQUEST_BYTES_LIMIT = (32 - 1) * 1024 * 1024
 ROW_COUNT_CHUNKING_THRESHOLD = 1000
 
+# Retry once on a 404 to smooth over eventual consistency. Short fixed wait
+# keeps added latency bounded for genuinely missing objects.
+NOT_FOUND_RETRY_ATTEMPTS = 2
+NOT_FOUND_RETRY_WAIT_SECONDS = 1.0
```
jtschoonhoven: How'd you pick 1s as the retry timeout?

tests/trace_server_bindings/test_http_utils.py
```diff
@@ -51,3 +54,62 @@ def test_calls_complete_mode_required_raises():
 
     with pytest.raises(CallsCompleteModeRequired, match="calls_complete mode required"):
         handle_response_error(response, "/call/upsert_batch")
+
+
+def _raise_404(body: dict) -> httpx.HTTPStatusError:
```
nikumar1206: super nit: tech this isn't raising anything

nikumar1206 (approved):

tests/trace_server_bindings/test_http_utils.py
gtarpenning: true

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -862,6 +864,7 @@ def file_create(self, req: tsi.FileCreateReq) -> tsi.FileCreateRes:
         return tsi.FileCreateRes.model_validate(r.json())
 
     @with_retry
+    @retry_on_not_found
```
jtschoonhoven: `@with_retry` is applied on top of `@retry_on_not_found` which is probably not what you want?

weave/trace_server_bindings/http_utils.py
```diff
@@ -340,6 +352,52 @@ def _is_413_error(e: Exception) -> bool:
...
+    """True for a 404 worth retrying; False if the body signals a deleted object."""
+    if not isinstance(exc, httpx.HTTPStatusError) or exc.response is None:
+        return False
+    if exc.response.status_code != 404:
+        return False
+    try:
+        body = exc.response.json()
+    except (json.JSONDecodeError, ValueError):
+        return True
+    return not (isinstance(body, dict) and "deleted_at" in body)
```
gtarpenning: okay this is not great

weave/trace_server_bindings/http_utils.py
```diff
@@ -340,6 +352,52 @@ def _is_413_error(e: Exception) -> bool:
...
+
+
+def retry_on_not_found(func: Callable[P, R]) -> Callable[P, R]:
```
jtschoonhoven, line 381: There's already a `@with_retry` decorator that pulls its config from the environment, can you build your own retry on top of that?

jtschoonhoven (commented):
Looks pretty good, a few questions/suggestions. Thanks for taking the trouble to do this on the client.

weave/trace_server_bindings/http_utils.py
gtarpenning: we can clean this up with a change to the server, but let's try this first and punt on that if it works

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -685,6 +686,7 @@ def obj_create(self, req: tsi.ObjCreateReq) -> tsi.ObjCreateRes:
         )
 
     @validate_call
+    @retry_on_not_found
     def obj_read(self, req: tsi.ObjReadReq) -> tsi.ObjReadRes:
```
jtschoonhoven, line 688: I still prefer doing this in the client, but it would be even better if we could be even more targeted than this. Are there specific write-then-read paths we could patch instead of doing this globally for all `obj_read`?

weave/trace_server_bindings/http_utils.py
gtarpenning: total guess, but if the eventual consistency in CH is > 1s we have major problems (and way way more of our tests would be busted)

weave/trace_server_bindings/http_utils.py
gtarpenning: yeah

weave/trace_server_bindings/remote_http_trace_server.py
gtarpenning: yep doing that

weave/trace_server_bindings/http_utils.py
```diff
@@ -340,6 +352,52 @@ def _is_413_error(e: Exception) -> bool:
     )
 
 
+_not_found_retry_disabled: ContextVar[bool] = ContextVar(
+    "weave_not_found_retry_disabled", default=False
+)
+
+
+@contextmanager
+def not_found_retry_disabled() -> Iterator[None]:
```
jtschoonhoven: Are you planning to use this somewhere? It's the reverse of what I'd expect to decorate using `not_found_retry_disabled` instead of decorating with `retry_on_not_found` (opt-out VS opt-in)

weave/trace_server_bindings/http_utils.py
jtschoonhoven: I don't have any intuition for how quickly CH writes propagate, but it would be neat if this could be 100ms instead of 1s.

jtschoonhoven:
From `/talk-with-tim`:
>⏺ Here are the actual read-after-write race sites, verified against the code:

>  SDK client (where the read follows an own-write in the same flow):

>  - weave/trace/weave_client.py:464 in WeaveClient.get(): Races when invoked via save() at line 458, which calls _save_object() then self.get(ref) immediately. Not a race when called directly on an existing ref.
>  - weave/trace/serialization/custom_objs.py:140 in _load_custom_obj_files(): Hydrates CustomWeaveType files — races when hydrating an object whose file_create just completed (e.g. publish() → get() of the same
  custom obj).
>  - weave/trace/serialization/serialize.py:298 in _load_custom_obj_files(): Same pattern as above, parallel implementation.
>  - weave/trace/vals.py:804 in make_trace_obj(): Dereferences nested ObjectRefs during deserialization — races when the parent was just created and contains refs to siblings written in the same publish.

>  Server-side internal create-then-read (weave/trace_server/clickhouse_trace_server_batched.py):

>  - clickhouse_trace_server_batched.py:3341 in op_create(): Reads back the obj immediately after insert to populate version_index in the response.
>  - clickhouse_trace_server_batched.py:3570 in dataset_create(): Same pattern — post-insert read for version_index.
>  - clickhouse_trace_server_batched.py:3719 in scorer_create(): Same.
>  - clickhouse_trace_server_batched.py:3841 in evaluation_create(): Same.
>  - clickhouse_trace_server_batched.py:3989 in model_create(): Same.

>  Not races (do NOT wrap):

>  - weave_client.py:1399, 1424, 1490 (get_tags, get_tags_and_aliases, get_aliases) — pure introspection, no preceding write.
>  - clickhouse_trace_server_batched.py:3590, 3743, 3865, 4022 (dataset_read, scorer_read, evaluation_read, model_read entry points) — user-initiated reads on arbitrary refs.
>  - clickhouse_trace_server_batched.py:3400, 3467, 4039, 4088 (file_content_read inside op_read/op_list/model_read/model_list) — reading source code for an already-existing object; not a read-after-own-write.

weave/trace_server_bindings/http_utils.py
gtarpenning: starting with `250`

weave/trace_server_bindings/http_utils.py
gtarpenning: i was, but reworked and removed

weave/trace_server_bindings/http_utils.py
```diff
@@ -340,6 +353,75 @@ def _is_413_error(e: Exception) -> bool:
...
+    try:
+        body = exc.response.json()
+    except (json.JSONDecodeError, ValueError):
+        return True
+    return not (isinstance(body, dict) and "deleted_at" in body)
+
+
+_in_write_then_read: ContextVar[bool] = ContextVar(
+    "weave_in_write_then_read", default=False
+)
```
jtschoonhoven: I think the ContextVar pattern adds needless complexity. It's hard to know if you're inside a write-then-read scope without tracing up through the call stack

weave/trace_server_bindings/http_utils.py
```diff
@@ -340,6 +353,75 @@ def _is_413_error(e: Exception) -> bool:
...
+    # Remote HTTP 404 -- retry unless the body indicates a delete.
+    if not isinstance(exc, httpx.HTTPStatusError) or exc.response is None:
+        return False
+    if exc.response.status_code != 404:
+        return False
+    try:
+        body = exc.response.json()
+    except (json.JSONDecodeError, ValueError):
+        return True
+    return not (isinstance(body, dict) and "deleted_at" in body)
```
jtschoonhoven: `"deleted_at" in body` is `True` even if `deleted_at` is `None`, is that what you want?

weave/trace_server_bindings/http_utils.py
```diff
@@ -340,6 +353,75 @@ def _is_413_error(e: Exception) -> bool:
...
+        return True
+    # Remote HTTP 404 -- retry unless the body indicates a delete.
+    if not isinstance(exc, httpx.HTTPStatusError) or exc.response is None:
+        return False
+    if exc.response.status_code != 404:
+        return False
+    try:
+        body = exc.response.json()
+    except (json.JSONDecodeError, ValueError):
+        return True
```
jtschoonhoven: Why does this return `True` on JSONDecodeError and ValueError?

weave/trace/weave_client.py
```diff
@@ -453,13 +455,17 @@ def save(self, val: Any, name: str, branch: str = "latest") -> Any:
         ref = self._save_object(val, name, branch)
         if not isinstance(ref, ObjectRef):
             raise TypeError(f"Expected ObjectRef, got {ref}")
-        return self.get(ref)
+        # Write-then-read: absorb eventual-consistency 404s on the read-back
+        # and on any nested reads (custom-type files, nested refs) during
+        # deserialization within `get()`.
+        with write_then_read_scope():
+            return self.get(ref)
```
jtschoonhoven: Is this the only place we need to retry? Can we just retry the `set.get` directly instead of using `write_then_read_scope`?

jtschoonhoven (commented):
I think the current implementation is more complicated than it needs to be, see comments.

weave/trace_server_bindings/http_utils.py
gtarpenning: I think there are valid reasons to use this, namely we will have to prop drill through a couple unrelated layers and cross-thread there are race conditions (in theory), but i'm sympathetic... lemme see

weave/trace_server_bindings/http_utils.py
gtarpenning: well... okay fine. the real fix is to add a proper error code (which is in a follow up: https://github.com/wandb/weave/pull/6658)

gtarpenning: but yes you are correct

gtarpenning: (server never emits that though)

weave/trace_server_bindings/http_utils.py
gtarpenning: good point, depends if you think it's more conservative to retry or not. One retry is pretty cheap, and does cover a case where the backend fails mid request for some reason, but thats exceedingly rare and who am I kidding this should be false.

jtschoonhoven: Yeah I think `False` makes the most sense in this context

weave/trace_server_bindings/http_utils.py
jtschoonhoven: I mean, do you want to check that `deleted_at` is some truthy value? Currently it just checks whether the key is present. But maybe I'm misunderstanding the intent here.

weave/trace_server_bindings/http_utils.py
```diff
@@ -340,6 +350,48 @@ def _is_413_error(e: Exception) -> bool:
...
+        return False
+    if exc.response.status_code != 404:
+        return False
+    try:
+        body = exc.response.json()
+    except (json.JSONDecodeError, ValueError):
+        return False
+    if not isinstance(body, dict):
+        return False
+    return body.get("deleted_at") is None
```
gtarpenning, line 378: TODO make this better

gtarpenning: (https://github.com/wandb/weave/pull/6658)

jtschoonhoven (approved):
🚢

PR #6667, opened 2026-04-21
chore(weave): stress tests + probe to diagnose obj_read NotFound flake

gtarpenning:
## Summary

Opt-in stress suite (`--run-stress`) that hammers `publish → ref.get()` in four shapes to bisect the ClickHouse read-after-write race:

1. sequential hammer (main thread)
2. concurrent per-thread (8 × 50)
3. cross-thread handoff (publish on A, get on B via barrier)
4. explicit `client.flush()` between publish and get

`ch_probe` fixture monkey-patches `obj_create` (post-insert visibility count) and `obj_read` (on-NotFound state dump + 1s-delayed re-query). Test-only, no product-code change.

`stress-tests.yaml` workflow runs on-demand (`workflow_dispatch` + `**/stress-*` branches) with the same CH 25.11.2.24 service the normal trace shard uses. `repeats` input controls pass count per job.

Normal CI unchanged — tests skip unless `--run-stress` is passed.

### DIFF
--- a/.github/workflows/stress-tests.yaml
+++ b/.github/workflows/stress-tests.yaml
@@ -1,0 +1,84 @@
+name: Stress tests (obj_read flake)
+
+# Opt-in stress tests for the publish→ref.get() NotFoundError flake.
+# Triggered manually via workflow_dispatch; optionally runs on push to
+# branches matching "*/stress-*" for iterative investigation.
+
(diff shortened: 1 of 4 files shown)


gtarpenning:
wow 100% pass rate!

PR #6668, opened 2026-04-21
fix(weave): retry NotFound at CachingMiddleware layer

gtarpenning:
## Summary

Moves `retry_on_not_found` one layer down from scattered client-side call sites into `CachingMiddlewareTraceServer.obj_read` and `.file_content_read`. Every path — SDK APIs, test helpers that hit `client.server.obj_read` directly, internal workers — now shares the same retry semantics in one place.

`fluke --all --days 7` on master shows **14 of 21** flake occurrences are `obj_read` / `file_content_read` NotFound. Most come from test helpers like `get_saved_code` in `test_op_versioning.py` that call `client.server.obj_read(...)` directly, bypassing the SDK-level retry added in #6653 / #6663. The middleware layer catches them uniformly.

Leaves the existing `weave/trace/weave_client.py` wraps in place — harmless (bounded nested-retry) and out of scope for this fix.

### DIFF
--- a/weave/trace/serialization/custom_objs.py
+++ b/weave/trace/serialization/custom_objs.py
@@ -23,6 +23,7 @@ from weave.trace_server.trace_server_interface import (
     FileContentReadReq,
     TraceServerInterface,
 )
+from weave.trace_server_bindings.http_utils import retry_on_not_found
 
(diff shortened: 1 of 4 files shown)


gtarpenning:
no, already doing this in the http layer

PR #6673, opened 2026-04-22
fix(weave): Fixes issue client.responses.create(...).withResponse is not a function

chance-wnb:
# Preserve OpenAI SDK `APIPromise` helpers through weave's wrapper

## Problem

Calling `withResponse` helpers on a weave-wrapped OpenAI client throws:

```
Exception: client.responses.create(...).withResponse is not a function
```

`weave.op()`'s async wrapper `await`s the inner callback and returns a fresh `Promise<T>` — the SDK's original `APIPromise` (and its `.withResponse()` / `.asResponse()` / `.parse()` / `._thenUnwrap()` helpers) never reaches the caller.

Affects `responses.create`, `chat.completions.create`, and `beta.chat.completions.parse`.

## Fix

- New helper `wrapAsAPIPromiseLike` — a `Proxy` that keeps `then`/`catch`/`finally` on the traced promise (so tracing still runs on `await`) and forwards every other property access to the original `APIPromise`.
- Both `makeOpenAIResponsesCreateOp` (renamed from `…Proxy`) and `makeOpenAIChatCompletionsOp` now capture the `APIPromise` from inside the op callback and hand it to `wrapAsAPIPromiseLike`.
- `makeOpenAIResponsesCreateOp` also converted from a `new Proxy(originalCreate, {apply})` form to a plain returned function matching `makeOpenAIChatCompletionsOp`; `originalCreate` is pre-bound at the `wrapOpenAI` call site. `shouldAdoptThis` is preserved so `self` still appears in traces.
- Silent `tracedPromise.catch(() => {})` on both paths: if a caller awaits only a forwarded helper, the outer traced promise has no handler — this avoids an unhandled-rejection warning in error cases (weave still records the exception inside the op).

## Test plan

- [x] Existing OpenAI integration tests still pass.
- [x] New regression tests added in `openai2.test.ts` covering `.withResponse()` on both `responses.create` and `chat.completions.create`.
- [x] Manual smoke test against a OpenAI + pi-coding-agent scenario.

### DIFF
--- a/sdks/node/src/__tests__/integrations/openai2.test.ts
+++ b/sdks/node/src/__tests__/integrations/openai2.test.ts
@@ -207,6 +207,57 @@ describe('OpenAI Integration', () => {
       // Verify wrapped matches unwrapped
       expect(wrappedResult).toEqual(unwrappedResult);
     });
+
+    it('should forward APIPromise.withResponse() through the chat completions wrapper', async () => {
+      const mockResponse = {
+        headers: new Map([['x-request-id', 'req_chat_id']]),
+      };
+      const parsedData = {
+        id: 'chat_test_id',
+        choices: [{message: {role: 'assistant', content: 'Hello'}}],
+        model: 'gpt-4',
+        usage: {prompt_tokens: 1, completion_tokens: 1, total_tokens: 2},
+      };
+
+      class MockAPIPromise {
+        then(onFulfilled: any, onRejected: any) {
+          return Promise.resolve(parsedData).then(onFulfilled, onRejected);
+        }
+        catch(onRejected: any) {
+          return Promise.resolve(parsedData).catch(onRejected);
+        }
+        finally(onFinally: any) {
+          return Promise.resolve(parsedData).finally(onFinally);
+        }
+        asResponse() {
+          return Promise.resolve(mockResponse);
+        }
+        async withResponse() {
+          return {
+            data: parsedData,
+            response: mockResponse,
+            request_id: mockResponse.headers.get('x-request-id'),
+          };
+        }
+      }
+
+      mockOpenAI.chat.completions.create = jest
+        .fn()
+        .mockImplementation(() => new MockAPIPromise());
+      wrappedOpenAI = wrapOpenAI(mockOpenAI);
+
+      const pending = wrappedOpenAI.chat.completions.create({
+        model: 'gpt-4',
+        messages: [{role: 'user', content: 'Hi'}],
+      });
+
+      expect(typeof pending.withResponse).toBe('function');
+      const {data, response, request_id} = await pending.withResponse();
+
+      expect(data).toEqual(parsedData);
+      expect(response).toBe(mockResponse);
+      expect(request_id).toBe('req_chat_id');
+    });
   });
 });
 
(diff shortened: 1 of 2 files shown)


chance-wnb:
* **#6673** <a href="https://app.graphite.com/github/pr/wandb/weave/6673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6673?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/__tests__/integrations/openai2.test.ts
```diff
@@ -207,6 +207,57 @@ describe('OpenAI Integration', () => {
...
+        headers: new Map([['x-request-id', 'req_chat_id']]),
+      };
+      const parsedData = {
+        id: 'chat_test_id',
+        choices: [{message: {role: 'assistant', content: 'Hello'}}],
+        model: 'gpt-4',
+        usage: {prompt_tokens: 1, completion_tokens: 1, total_tokens: 2},
+      };
+
+      class MockAPIPromise {
```
rgao-coreweave, line 222: The MockAPIPromise class is near-identical at lines 222-242 (chat completions test) and lines 495-515 (responses test). Consider extracting to a shared test helper to reduce duplication and make it easier to update if the OpenAI SDK adds new helpers.

sdks/node/src/integrations/openai.ts
```diff
@@ -444,29 +460,68 @@ export function summarizer(result: any) {
...
+// Why this is needed: weave's op() awaits the inner callback and
+// returns a fresh Promise of the parsed value — the original
+// APIPromise (and its helpers) is gone by the time the caller sees
+// the result. The factory captures the APIPromise in a closure and
+// hands it here; this proxy keeps then/catch/finally on the traced
+// promise so tracing still runs on await, and forwards every other
+// property access to the original APIPromise.
+function wrapAsAPIPromiseLike(tracedPromise: Promise<any>, apiPromise: any) {
+  return new Proxy(tracedPromise, {
+    get(target, p) {
```
rgao-coreweave, line 516: The Proxy only traps `get`. The `has` trap (`'withResponse' in result`) and `ownKeys`/`getOwnPropertyDescriptor` are not trapped, so they fall through to `tracedPromise` (a plain `Promise`). This means `'withResponse' in result` returns `false` even though `result.withResponse` works.

Consider implement `has` but the other traps are not commonly used.

```
                                                           
  return new Proxy(tracedPromise, {                                                                                                                                           
    get(target, p) {                                                                                                                                                          
      if (p === 'then' || p === 'catch' || p === 'finally') {                                                                                                                 
        const val = Reflect.get(target, p);                                                                                                                                   
        return val.bind(target);                                                                                                                                              
      }
      const val = Reflect.get(apiPromise, p);                                                                                                                                 
      return typeof val === 'function' ? val.bind(apiPromise) : val;          
    },                                                                                                                                                                        
    has(target, p) {
      if (p === 'then' || p === 'catch' || p === 'finally') {                                                                                                                 
        return Reflect.has(target, p);      // check the Promise              
      }                                                                                                                                                                       
      return Reflect.has(apiPromise, p);    // check the APIPromise
    },                                                                                                                                                                        
  });  
```

rgao-coreweave (approved):
@chance-wnb Amazing work!

The fix introduces a Proxy-based shim that splits property access: then/catch/finally route to the traced promise (preserving weave tracing on await), while every other property access forwards to the original APIPromise (preserving SDK helpers). 

The Proxy works nicely so that callers don't need to know whether they're interacting with a real APIPromise or the shim. Learned a lot!

sdks/node/src/__tests__/integrations/openai2.test.ts
```diff
@@ -413,6 +464,74 @@ describe('OpenAI responses.create Integration', () => {
...
+        .mockImplementation(() => new MockAPIPromise());
+      wrappedOpenAI = wrapOpenAI(mockOpenAI);
+
+      const pending = wrappedOpenAI.responses.create({
+        model: 'gpt-4o-2024-08-06',
+        messages: [{role: 'user', content: 'Hello!'}],
+      });
+
+      expect(typeof pending.withResponse).toBe('function');
+      const {data, response, request_id} = await pending.withResponse();
```
andrewtruong, line 528: What about the case where `stream: true`?  Can you please add a test for that as well?

chance-wnb:
closing in favor of https://github.com/wandb/weave/pull/6691

PR #6675, opened 2026-04-22
chore(weave): Add EvaluationLoggerV2

andrewtruong:
This PR adds a variant of `EvaluationLogger` that uses the server-side APIs instead of our client-side hack.

### DIFF
--- a/tests/flow/test_evaluation_imperative_contract.py
+++ b/tests/flow/test_evaluation_imperative_contract.py
@@ -1,0 +1,218 @@
+"""Contract tests parametrized across EvaluationLogger (V1) and EvaluationLoggerV2.
+
+These tests cover observable, user-facing behavior that both implementations
+must support identically. They intentionally avoid peeking at implementation
+details (specific op names, call-tree shapes, internal trace structure) so
+the same test body can run against V1 and V2.
+
+Anything that checks V1-specific trace structure (op names like
+"Evaluation.evaluate", specific parent_id relationships, V1 summary shape)
+belongs in ``test_evaluation_imperative.py``. Anything that checks V2-only
+behavior (prediction_list, evaluation_run status, deferred output update)
+belongs in ``test_evaluation_imperative_v2.py``.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from weave.evaluation.eval_imperative import EvaluationLogger
+from weave.evaluation.eval_imperative_v2 import EvaluationLoggerV2
+
+
+@pytest.fixture(params=[EvaluationLogger, EvaluationLoggerV2], ids=["v1", "v2"])
+def logger_cls(request):
+    return request.param
+
+
+def test_log_example_roundtrip(client, logger_cls):
+    ev = logger_cls(model="m", dataset=[{"q": "x"}])
+    ev.log_example(
+        inputs={"q": "x"},
+        output="y",
+        scores={"correct": 1.0, "fluent": 0.8},
+    )
+    ev.log_summary()
+
+    pred = ev._accumulated_predictions[0]
+    assert pred._captured_scores == {"correct": 1.0, "fluent": 0.8}
+    assert pred._has_finished
+
+
+def test_log_example_multiple_examples(client, logger_cls):
+    ev = logger_cls(model="m", dataset=[{"q": "a"}, {"q": "b"}])
+    ev.log_example(inputs={"q": "a"}, output="A", scores={"s": 1.0})
+    ev.log_example(inputs={"q": "b"}, output="B", scores={"s": 0.5})
+    ev.log_summary()
+
+    assert len(ev._accumulated_predictions) == 2
+    assert ev._accumulated_predictions[0]._captured_scores == {"s": 1.0}
+    assert ev._accumulated_predictions[1]._captured_scores == {"s": 0.5}
+
+
+def test_log_example_after_finalization_raises(client, logger_cls):
+    ev = logger_cls(model="m", dataset=[{"q": "x"}])
+    ev.log_example(inputs={"q": "x"}, output="y", scores={"s": 1.0})
+    ev.log_summary()
+    with pytest.raises(ValueError, match="finalized"):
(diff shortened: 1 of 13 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -4711,18 +4786,67 @@ def prediction_finish(
...
+        # We mitigate the cost by:
+        #   (a) only calling this when the new output actually differs from
+        #       what's already stored (diff check below), and
+        #   (b) the V2 logger's common flow (output provided upfront OR
+        #       assigned before any log_score) never triggers this because
+        #       create-time output already matches the final output.
+        # In practice the mutation only fires when a caller mutates
+        # `pred.output` AFTER logging a score, or overrides via
+        # `pred.finish(output=...)`. For high-frequency eval workloads that
+        # rely on that pattern, the right long-term fix is to change
```
chance-wnb: > the right long-term fix is to

I am afraid that's the correct fix and should happen before this ships to production traffic. But I think directly attacking the issue by refactoring the AggregatingMergeTree is better. or otherwise sooner or later we will pay. 

Your mentioning of  `the right long-term fix` is the bottom line :)

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -4711,18 +4786,67 @@ def prediction_finish(
...
             )
         )
         self.call_end(call_end_req)
 
+        # If a new `output` was supplied on finish, overwrite the aggregated
+        # output for this call. `prediction_create` already emitted a
+        # `call_end` row with the create-time output, and `calls_merged` uses
+        # `SimpleAggregateFunction(any, ...)` for `output_dump` — so a second
+        # `call_end` row alone cannot win against the first one. We issue an
+        # `ALTER TABLE ... UPDATE` mutation on both `call_parts` (source) and
```
chance-wnb: IMHO, `ALTER TABLE ... UPDATE ` is very costly. You know our calls_merged table contains data not related to eval business as well, so it might rewrite a multi-GB part per row, on every replica, while blocking merges. (An ALTER TABLE ... UPDATE mutation rewrites parts, not granules.)

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -4711,18 +4786,67 @@ def prediction_finish(
...
                 summary={},
             )
         )
         self.call_end(call_end_req)
 
+        # If a new `output` was supplied on finish, overwrite the aggregated
+        # output for this call. `prediction_create` already emitted a
+        # `call_end` row with the create-time output, and `calls_merged` uses
+        # `SimpleAggregateFunction(any, ...)` for `output_dump` — so a second
+        # `call_end` row alone cannot win against the first one. We issue an
```
chance-wnb: > so a second `call_end` row alone cannot win

I think this could work with the current AggregatingMergeTree

- Add `updated_at DateTime64(6)` to `call_parts`, stamped `now64(6)` per insert.

- `calls_merged`: `MODIFY COLUMN output_dump AggregateFunction(argMax, String, DateTime64(6))` 

- `calls_merged_view`: change from `anySimpleState(output_dump)` to `argMaxState(output_dump, updated_at)`.

- Query time: `SELECT output_dump` → `argMaxMerge(output_dump)`.

This will need to rebuild a column of output_dump. I am not sure how realistic it is. Let's ask expert's opinion: @gtarpenning

chance-wnb (changes requested):
Let's zoom into the `ALTER TABLE ... UPDATE` issue.

tests/flow/test_evaluation_imperative_contract.py
```diff
@@ -0,0 +1,217 @@
...
+    ev.log_example(inputs={"q": "a"}, output="A", scores={"s": 1.0})
+    ev.log_example(inputs={"q": "b"}, output="B", scores={"s": 0.5})
+    ev.log_summary()
+
+    assert len(ev._accumulated_predictions) == 2
+    assert ev._accumulated_predictions[0]._captured_scores == {"s": 1.0}
+    assert ev._accumulated_predictions[1]._captured_scores == {"s": 0.5}
+
+
+def test_log_example_after_finalization_raises(client, logger_cls):
```
gtarpenning: can we collapse these tests into like 4 testing real user flows with multiple assertions instead of like a zillion ones doing very simple things?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1030,6 +1030,78 @@ def _update_call_end_in_calls_complete(
...
+        # than lightweight updates but are the only correct way to override
+        # an `any()` aggregation after the fact.
+        settings: dict[str, int | str] = {"mutations_sync": 1}
+
+        pb_parts = ParamBuilder()
+        project_id_param_parts = pb_parts.add_param(project_id)
+        id_param_parts = pb_parts.add_param(call_id)
+        output_dump_param_parts = pb_parts.add_param(output_dump)
+        output_refs_param_parts = pb_parts.add_param(output_refs)
+        call_parts_query = f"""
```
gtarpenning: oh man we are going to do two updates? yikes

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1030,6 +1030,78 @@ def _update_call_end_in_calls_complete(
             settings=ch_settings.CLICKHOUSE_LIGHTWEIGHT_UPDATE_SETTINGS,
         )
 
+    @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched._update_call_output")
+    def _update_call_output(
```
gtarpenning: most of this belongs in a `/query_builder/` file

gtarpenning (changes requested):
This seems very dangerous

neutralino1 (commented):
Concerns from @gtarpenning and @chance-wnb make me ask: should we migrate the data model first? As in, create new tables dedicated to evals and decouple from the calls table, before we do this switch over?

PR #6567, opened 2026-04-09
feat(weave): add GenAI observability schema, extraction, and query layer

bcsherma:
## Summary

Adds a GenAI / agent observability layer to the Weave trace server. Extraction promotes GenAI semantic-convention attributes into typed ClickHouse columns; everything else lands in three typed `custom_attrs*` Maps plus a lossless raw span dump. Reads go through dedicated REST endpoints backed by the `spans` table and three materialized rollups.

Any OTel span can be ingested, queried via the spans API, and browsed in a (future) span explorer UI. GenAI-semconv attributes unlock the agent-specific surfaces: the agents list, per-agent invocation counts / token totals, conversation view with per-turn chat trajectory, and full-text message search.

## Data model

Three first-class concepts derived from span metadata, not stored as separate rows:

- **Session** (a.k.a. Conversation) — spans sharing `conversation_id`.
- **Turn** — one `trace_id` within a session. Matches how every mainstream agent SDK (OpenAI Agents SDK, Google ADK, LangChain, Anthropic) scopes traces. Verified against the actual instrumentation source for each. Subagents nested in the same trace count as invocations on their own agent row, not as new turns.
- **Invocation** — one `invoke_agent` span, counted per `agent_name`.

## What's included

**Migration** (`migrations/030_add_agent_tables.up.sql`)

- `spans` — `ReplacingMergeTree`, one row per OTel span. Wide GenAI schema + three typed `custom_attrs*` Maps + raw OTel dumps. Partitioned by month, TTL on `expire_at`.
- `agents` / `agent_versions` — `AggregatingMergeTree`s fed by MVs off `spans`, keyed by `(project, agent_name)` and `(project, agent_name, agent_version)`. Track invocation / span / token / duration / error totals. Duration arithmetic guarded against UInt64 underflow when `ended_at` defaults to epoch.
- `messages` — `MergeTree`, one row per (span, message occurrence). Fed by **five independent MVs** — output_messages, input_messages, system_instructions, tool_call_arguments, tool_call_result — each targeting the same table. `content_digest = murmurHash3_128` stored raw as `FixedString(16)`; hex-encoded on the read path. Skip indexes: tokenbf on content, bloom on digest / span_id / trace_id / conversation_id.

**Agent module** (`weave/trace_server/agents/`)

- `schema.py` — Pydantic model mirroring the ClickHouse row; source of truth for insert column order.
[...]

PR #6678, opened 2026-04-22
feat(weave): add typed attribute maps for fast filtering

gtarpenning:
## Summary

Adds four typed Map columns to `call_parts`, `calls_merged`, and `calls_complete` alongside `attributes_dump`. At ingest, each attribute value is routed into `attributes_map_str` / `_int` / `_float` / `_bool` by Python type. Filters wrapped in `$convert` hit the typed column directly instead of `JSON_VALUE(attributes_dump, ...)`.

- ORDER BY and unconverted filters keep the JSON_VALUE path — no behavior change for those.
- Modeled on the `custom_attrs*` pattern in #6567.

## Back-compat

Rows inserted before migration 030 (or before backfill completes) have empty typed maps. The query builder emits a row-level `if(mapContains(...), ...)` guard so the fast path fires only when the map was populated for that row; legacy rows fall back to the existing JSON_VALUE path. Mixed-backfill tables stay correct without a cutoff timestamp or table-wide flag.

Minimal example — `$convert(attributes.retries, int) == 3` on `calls_complete`:

```sql
if(attributes_map_int != NULL,  -- fix me (can break out to top)
   attributes_map_int['retries'],
   toInt64OrNull(coalesce(nullIf(JSON_VALUE(attributes_dump, '$."retries"'), 'null'), '')))
= 3
```

## Test plan

- [x] 10 new unit tests in `test_attributes_map_dispatch.py` (dispatch per cast, nested paths, fallback, extractor edge cases)
- [x] 366 existing `query_builder/` tests pass
- [x] `clickhouse_trace_server_migrator` tests pass

Local benchmarking:

### DIFF
--- a/tests/trace_server/query_builder/test_attributes_map_dispatch.py
+++ b/tests/trace_server/query_builder/test_attributes_map_dispatch.py
@@ -5,12 +5,13 @@ cast, the query builder emits a hybrid
 ``if(mapContains(map, key), map[key], clickhouse_cast(JSON_VALUE(...)))``
 expression: the typed Map column wins on rows where migration 030 populated
 it at ingest, and the old JSON_VALUE path takes over on legacy rows whose
-maps are still empty. This file exercises the dispatch SQL shape end-to-end
-on both read tables plus the fall-through cases (missing cast, ``exists``
-cast, empty extra_path).
-"""
+maps are still empty. This file asserts the full SQL shape end-to-end on
+both read tables plus the fall-through cases (no cast, ``exists`` cast).
 
-import pytest
+Each test is intentionally un-parametrized and pins the complete query
+shape — the SQL layer is load-bearing enough that a shared template or
+fuzzy match would miss regressions at the punctuation level.
+"""
 
 from tests.trace_server.query_builder.utils import assert_sql
 from weave.trace_server.calls_query_builder.calls_query_builder import CallsQuery
@@ -20,35 +21,18 @@ from weave.trace_server.clickhouse.utilities import (
     extract_typed_attrs,
 )
 from weave.trace_server.interface import query as tsi_query
-from weave.trace_server.orm import ParamBuilder
 from weave.trace_server.project_version.types import ReadTable
 
 # ---------------------------------------------------------------------------
-# Query-builder dispatch
+# calls_merged dispatch (one test per supported cast)
 # ---------------------------------------------------------------------------
 
 
-@pytest.mark.parametrize(
-    ("cast", "expected_column", "expected_param_type", "fallback_cast", "literal"),
-    [
-        ("int", "attributes_map_int", "Int64", "toInt64OrNull", 1),
-        ("double", "attributes_map_float", "Float64", "toFloat64OrNull", 1.5),
-        ("bool", "attributes_map_bool", "Bool", "toUInt8OrNull", True),
-        ("string", "attributes_map_str", "String", "toString", "prod"),
-    ],
-)
-def test_typed_map_dispatch_on_calls_merged(
-    cast: str,
-    expected_column: str,
-    expected_param_type: str,
-    fallback_cast: str,
-    literal: object,
-) -> None:
-    """Each supported cast emits an if(mapContains, map-read, cast(JSON_VALUE)) gate.
+def test_typed_map_dispatch_on_calls_merged_int() -> None:
+    """``$convert(attributes.env, int)`` on calls_merged.
 
-    The fast branch reads the agg-wrapped typed column, and the else branch is
-    the existing JSON_VALUE-over-attributes_dump path preserved for legacy rows
-    whose maps haven't been backfilled.
+    Fast branch: ``any(attributes_map_int)[key]``.
+    Fallback: ``toInt64OrNull(coalesce(nullIf(JSON_VALUE(...), 'null'), ''))``.
     """
     cq = CallsQuery(project_id="p", read_table=ReadTable.CALLS_MERGED)
     cq.add_field("id")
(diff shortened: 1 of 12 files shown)


weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -2074,6 +2085,48 @@ def _get_multi_value_feedback_field(
     return None
 
 
+def _maybe_attributes_map_filter_sql(
```
mscavezze-cw, line 2221: Does this make an assumption that the table in question has the attributes_map column, and further that it is populated for all rows? Should there be handling for partially back-filled tables?

weave/trace_server/clickhouse/utilities.py
```diff
@@ -23,6 +24,11 @@
 
 logger = logging.getLogger(__name__)
 
+# Per-map cap for the typed attribute maps populated by extract_typed_attrs.
+# Applied independently to each of the four maps, so a single call can
+# contribute up to 4 * MAX_TYPED_ATTR_ENTRIES_PER_MAP entries across all maps.
```
mscavezze-cw: What happens if we exceed that?
- do we fail the insert?
- do we insert a subset of the values?
- do we insert a sentinel value so that we can be sure not to use the fast sql for this row or table?

mscavezze-cw (commented):
Cool - I skimmed the PR and asked some high-level questions.

gtarpenning:
lets consider back compat path with a version number instead, at least be able to break out the check for map existence to the top, all cols should be present if any are.

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: [:robot: automated] Considered a few back-compat gate options; kept the per-filter `mapContains` for now and wanted to leave the trade-offs in the thread instead of the code.

1. **Per-filter `mapContains(map, key)` (current).** Correct for mixed-backfill rows and preserves cross-type cast semantics — `$convert(attributes.enabled, string) = 'true'` on a bool attribute falls back to `JSON_VALUE` and matches the pre-maps behavior exactly. Downside: duplicates the check across every `$convert` filter in a query.
2. **Row-level "any map has content" gate.** `length(map_str) + length(map_int) + length(map_float) + length(map_bool) > 0` would hoist the check to a single per-row expression (all four maps populate together at ingest). Cheaper, but changes semantics on wrong-type casts: the fast branch returns the Map default (`''`/`0`/`false`/`0.0`) for a key that lives in a sibling typed map, where the JSON fallback would cast it successfully. Also misclassifies empty-attrs rows (all maps empty even post-extraction).
3. **Version column (e.g. `attrs_extracted UInt8`).** Same per-row shape as (2), same semantic trade-off, plus a new migration to add + backfill the marker. Cleanest long-term answer once a backfill job exists — probably worth revisiting then.

Picked (1) because correctness > perf during backfill, and CH inlines the common-subexpression `mapContains` + `map[key]` on populated rows so the duplicated work is smaller than it looks in the SQL.

Separately, removed `MAX_TYPED_ATTR_ENTRIES_PER_MAP` in this push — `_strip_large_values` now treats `attributes_dump` + the four `attributes_map_*` columns as one slot, so oversized rows get the existing `ENTITY_TOO_LARGE_PAYLOAD` treatment instead of silent truncation at extract time.

weave/trace_server/clickhouse/utilities.py
gtarpenning: [🤖 automated] Went with option (b) effectively — removed the cap entirely. Oversized rows now fall through to the existing `_strip_large_values` path (which this push also teaches about the `attributes_map_*` columns, so the dump + typed maps get the `ENTITY_TOO_LARGE_PAYLOAD` replacement together). No silent truncation at extract time.

tests/trace_server/query_builder/test_attributes_map_dispatch.py
```diff
@@ -0,0 +1,534 @@
...
+    cq.add_field("id")
+    cq.add_condition(
+        tsi_query.EqOperation.model_validate(
+            {
+                "$eq": [
+                    {
+                        "$convert": {
+                            "input": {"$getField": "attributes.env"},
+                            "to": "int",
+                        }
```
bcsherma, line 44: Why do we need to do this `$convert`? Since we are comparing to a literal and constrained in general to literal comparisons can't we just infer from the literal type?

tests/trace_server/query_builder/test_attributes_map_dispatch.py
```diff
@@ -0,0 +1,534 @@
...
+        cq,
+        """
+        SELECT calls_merged.id AS id
+        FROM calls_merged
+        PREWHERE calls_merged.project_id = {pb_3:String}
+        GROUP BY (calls_merged.project_id, calls_merged.id)
+        HAVING (
+            ((if(mapContains(any(calls_merged.attributes_map_int), {pb_0:String}),
+                 any(calls_merged.attributes_map_int)[{pb_0:String}],
+                 toInt64OrNull(coalesce(nullIf(JSON_VALUE(any(calls_merged.attributes_dump), {pb_1:String}), 'null'), '')))
```
bcsherma, line 62: So I make sure I understand, we still load the attributes for all rows but we avoid the json parse cost if the typed maps are populated and the key is found

bcsherma (commented):
The columns look good, I guess the tricky bit here is backcompat and right now the query is eating that complexity by coalescing new -> old data sources. I am wondering how much this is really going to help bc the attribute dumps are typically pretty small. For inputs/outputs this is much higher value right? How much impact does this have?

tests/trace_server/query_builder/test_attributes_map_dispatch.py
gtarpenning: Currently all fields are stored as strings in the backend, this is what the frontend does. I do agree that it's possibly not required. I'll investigate.

PR #6681, opened 2026-04-22
chore(weave): Make serialization more clear

andrewtruong:
This PR splits out the serializers into their own functions which makes it easier to test each case and understand how payloads get serialized

### DIFF
--- a/tests/trace/test_serialize_encoders.py
+++ b/tests/trace/test_serialize_encoders.py
@@ -1,0 +1,511 @@
+"""Tests for the encoder ladder in ``weave.trace.serialization.serialize``.
+
+Each ``_encode_*`` helper is tested in isolation for two behaviors:
+  1. Match: returns the correctly-shaped JSON payload for an input it owns.
+  2. Miss: returns the ``_MISS`` sentinel for an input it does not own.
+
+Separately, ``to_json`` is tested end-to-end to verify the dispatch order —
+higher-priority encoders win when multiple could match — and that the terminal
+``stringify`` fallback catches anything that falls through.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from typing import Any, NamedTuple
+
+import pytest
+from pydantic import BaseModel
+
+from weave.flow.scorer import WeaveScorerResult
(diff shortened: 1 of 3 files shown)


tests/trace/test_serialize_encoders.py
```diff
@@ -0,0 +1,511 @@
...
+    to_json,
+)
+
+PROJECT_ID = "entity/project"
+
+
+# ---------- _encode_ref ----------
+
+
+def test_encode_ref_table_ref_returns_uri() -> None:
```
chance-wnb: **AI's Suggestion: consolidate a few test groups via `@pytest.mark.parametrize`**
Nice coverage on the encoder ladder. A few spots where parametrize could cut ~40 lines without losing any assertions — feel free to take or leave:
1. **`_encode_ref` match** (`tests/trace/test_serialize_encoders.py:45-52`) — the two tests differ only by the `Ref` subclass. Could parametrize on `ref` directly since both assert `== ref.uri`.
2. **`_encode_container` match** (`tests/trace/test_serialize_encoders.py:126-176`) — 7 tests collapse into one parametrize over `(obj, expected)` pairs (list/tuple/dict/empty/nested/namedtuple). The namedtuple-priority case (`:150-158`) is already implied by `test_encode_container_namedtuple_returns_dict` (`:142-147`) returning a dict rather than a list, so one of those could go.
3. **`_encode_pydantic_schema` misses** (`tests/trace/test_serialize_encoders.py:200-217`) — `test_encode_pydantic_instance_returns_miss` and `test_encode_pydantic_base_model_itself_returns_miss` could fold into the existing parametrized miss test by adding a module-level `_PydFoo(x=1)` instance and `BaseModel` to the value list.
4. **`test_encode_primitive_long_string_passes_through`** (`tests/trace/test_serialize_encoders.py:231-234`) — probably redundant with the parametrized identity test above it; no length-sensitive logic lives in `_encode_primitive`. The truncation comment belongs on `stringify` if anywhere.
5. **`_encode_try_to_dict` match** (`tests/trace/test_serialize_encoders.py:364-379`) — two tests differ only in `to_dict()`'s return; parametrize on `payload`.

chance-wnb (approved):
Approved with a comment to consolidate tests.

PR #6684, opened 2026-04-22
feat(weave): add agent_turn/agent_conversation/agent_span ref kinds

nikumar1206:
## Description

- Fixes [WB-33571](https://coreweave.atlassian.net/browse/WB-33571)

First of three stacked PRs for Phase 1 of human feedback on agent turns.

Adds three new ref kinds (external + internal): \`agent_turn/{trace_id}\`, \`agent_conversation/{conversation_id}\`, \`agent_span/{span_id}\`. No consumers yet — just the grammar. \`conversation_id\` is URL-encoded since it's app-chosen.

Stack: this → #6685 → #6686

## Testing

- unit tests

[WB-33571]: https://coreweave.atlassian.net/browse/WB-33571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6685, opened 2026-04-22
feat(weave): add helper for querying feedback by agent target refs

nikumar1206:
## Description

- Fixes [WB-33571](https://coreweave.atlassian.net/browse/WB-33571)

Second of three stacked PRs for Phase 1 of human feedback on agent turns.

Adds two helpers: \`make_agent_feedback_query_req\` builds one feedback query for a mix of turn/conversation/span targets; \`group_agent_feedback_by_target\` buckets the result by identifier. Tiny refactor: shared feedback field list extracted to a constant so the call and agent helpers don't duplicate it.

Stack: #6684 → this → #6686

## Testing

- unit tests

[WB-33571]: https://coreweave.atlassian.net/browse/WB-33571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/trace_server_common.py
+++ b/weave/trace_server/trace_server_common.py
@@ -3,12 +3,27 @@ import datetime
 import json
 from collections import OrderedDict, defaultdict
 from collections.abc import Iterator
+from dataclasses import dataclass, field
 from typing import Any, Literal, cast
 
 from weave.shared import refs_internal as ri
 from weave.trace_server import trace_server_interface as tsi
 
 CallStatus = Literal["running", "completed", "failed"]
+
+FEEDBACK_QUERY_FIELDS = [
+    "id",
+    "feedback_type",
+    "weave_ref",
+    "payload",
+    "creator",
+    "created_at",
+    "wb_user_id",
+    "runnable_ref",
+    "call_ref",
+    "trigger_ref",
+    "annotation_ref",
+]
 
 
 def make_feedback_query_req(
@@ -34,22 +49,78 @@ def make_feedback_query_req(
     )
     feedback_query_req = tsi.FeedbackQueryReq(
         project_id=project_id,
-        fields=[
-            "id",
-            "feedback_type",
-            "weave_ref",
-            "payload",
-            "creator",
-            "created_at",
-            "wb_user_id",
-            "runnable_ref",
-            "call_ref",
-            "trigger_ref",
-            "annotation_ref",
-        ],
+        fields=FEEDBACK_QUERY_FIELDS,
         query=query,
     )
     return feedback_query_req
+
+
+@dataclass(frozen=True)
+class AgentFeedbackByTarget:
+    """Feedback rows grouped by agent target kind + raw identifier.
+
+    Keys are raw trace_id / conversation_id / span_id strings (not ref URIs)
+    so callers can look up by whatever identifier they already have from the
+    spans row.
+    """
+
+    by_trace_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
+    by_conversation_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
+    by_span_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
+
+
+def make_agent_feedback_query_req(
+    project_id: str,
+    refs: list[str],
+) -> tsi.FeedbackQueryReq:
+    """Build a FeedbackQueryReq that returns feedback for any of `refs`.
+
+    The refs are unioned (OR'd) into a single `$in` over `weave_ref` so
+    the chat-view handler can fetch feedback for a turn plus all of its
+    steps (or a conversation plus all its turns) in one round-trip.
+    """
+    query = tsi.Query(
+        **{
+            "$expr": {
+                "$in": [
+                    {"$getField": "weave_ref"},
+                    [{"$literal": ref} for ref in refs],
+                ]
+            }
+        }
+    )
+    return tsi.FeedbackQueryReq(
+        project_id=project_id,
+        fields=FEEDBACK_QUERY_FIELDS,
+        query=query,
+    )
+
+
+def group_agent_feedback_by_target(
+    feedback: tsi.FeedbackQueryRes,
+) -> AgentFeedbackByTarget:
+    """Group feedback rows by agent target kind + raw identifier."""
+    by_trace_id: dict[str, list[dict[str, Any]]] = defaultdict(list)
+    by_conversation_id: dict[str, list[dict[str, Any]]] = defaultdict(list)
+    by_span_id: dict[str, list[dict[str, Any]]] = defaultdict(list)
+
+    for raw in feedback.result:
+        # `FeedbackQueryRes.result` is intentionally `dict[str, Any]` because
+        # callers can request arbitrary fields; this helper always passes
+        # FEEDBACK_QUERY_FIELDS, so the rows match FeedbackDict shape.
+        ref = ri.parse_internal_uri(raw.get("weave_ref", ""))
+        if isinstance(ref, ri.InternalAgentTurnRef):
+            by_trace_id[ref.trace_id].append(raw)
+        elif isinstance(ref, ri.InternalAgentConversationRef):
+            by_conversation_id[ref.conversation_id].append(raw)
+        elif isinstance(ref, ri.InternalAgentSpanRef):
+            by_span_id[ref.span_id].append(raw)
+
+    return AgentFeedbackByTarget(
+        by_trace_id=by_trace_id,
+        by_conversation_id=by_conversation_id,
+        by_span_id=by_span_id,
+    )
 
 
 def hydrate_calls_with_feedback(
--- a/tests/trace_server/test_agent_feedback_helpers.py
+++ b/tests/trace_server/test_agent_feedback_helpers.py
@@ -1,0 +1,62 @@
+from weave.shared import refs_internal as ri
+from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.trace_server_common import (
+    group_agent_feedback_by_target,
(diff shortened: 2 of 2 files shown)


nikumar1206:
* **#6686** <a href="https://app.graphite.com/github/pr/wandb/weave/6686?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6685** <a href="https://app.graphite.com/github/pr/wandb/weave/6685?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6685?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6684** <a href="https://app.graphite.com/github/pr/wandb/weave/6684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6567** <a href="https://app.graphite.com/github/pr/wandb/weave/6567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/trace_server_common.py
```diff
@@ -34,24 +49,93 @@ def make_feedback_query_req(
...
-            "trigger_ref",
-            "annotation_ref",
-        ],
+        fields=FEEDBACK_QUERY_FIELDS,
         query=query,
     )
     return feedback_query_req
 
 
+@dataclass
```
jtschoonhoven:
```suggestion
@dataclass(frozen=True)
```

weave/trace_server/trace_server_common.py
```diff
@@ -34,24 +49,93 @@ def make_feedback_query_req(
...
+    project_id: str,
+    trace_ids: list[str] | None = None,
+    conversation_ids: list[str] | None = None,
+    span_ids: list[str] | None = None,
+) -> tsi.FeedbackQueryReq:
+    """Build a FeedbackQueryReq for a batch of agent targets.
+
+    Combines agent_turn / agent_conversation / agent_span refs in a single
+    query so the chat-view handler can fetch feedback for a turn plus all of
+    its steps (or a conversation plus all its turns) in one round-trip.
```
jtschoonhoven: This means the filters are OR and not AND, right? Meaning the query matches on trace_id OR conversation_id OR span_id? Probably worth saying that explicitly in the docstring, usually filters are ANDed.

weave/trace_server/trace_server_common.py
```diff
@@ -34,24 +49,93 @@ def make_feedback_query_req(
...
+    """Feedback rows grouped by agent target kind + raw identifier.
+
+    Keys are raw trace_id / conversation_id / span_id strings (not ref URIs)
+    so callers can look up by whatever identifier they already have from the
+    spans row.
+    """
+
+    by_trace_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
+    by_conversation_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
+    by_span_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
```
jtschoonhoven, line 69: Why can't we type these fields more specifically than `dict[str, list[dict[str, Any]]]`? Whenever possible we should avoid passing around opaque dicts. Can we deserialize the rows before we load them into the class?

weave/trace_server/trace_server_common.py
```diff
@@ -34,24 +49,93 @@ def make_feedback_query_req(
...
+            by_trace_id[ref.trace_id].append(item)
+        elif isinstance(ref, ri.InternalAgentConversationRef):
+            by_conversation_id[ref.conversation_id].append(item)
+        elif isinstance(ref, ri.InternalAgentSpanRef):
+            by_span_id[ref.span_id].append(item)
+
+    return AgentFeedbackByTarget(
+        by_trace_id=dict(by_trace_id),
+        by_conversation_id=dict(by_conversation_id),
+        by_span_id=dict(by_span_id),
```
jtschoonhoven: These are already dicts, there shouldn't be any need to coerce them again

weave/trace_server/trace_server_common.py
```diff
@@ -34,24 +49,93 @@ def make_feedback_query_req(
...
+    by_trace_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
+    by_conversation_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
+    by_span_id: dict[str, list[dict[str, Any]]] = field(default_factory=dict)
+
+
+def make_agent_feedback_query_req(
+    project_id: str,
+    trace_ids: list[str] | None = None,
+    conversation_ids: list[str] | None = None,
+    span_ids: list[str] | None = None,
```
jtschoonhoven: I'm not necessarily opposed, but are you sure this is the right shape for this API? Constructing refs is trivial, so I don't know that it adds that much benefit to transform ID -> Ref in this function. It might make sense to push this responsibility onto the client and just expose an endpoint that accepts a list of refs. That would be more flexible.

weave/trace_server/trace_server_common.py
nikumar1206: agreed added docstring

weave/trace_server/trace_server_common.py
```diff
@@ -34,24 +49,81 @@ def make_feedback_query_req(
...
+    by_span_id: dict[str, list[tsi.FeedbackDict]] = field(default_factory=dict)
+
+
+def make_agent_feedback_query_req(
+    project_id: str,
+    refs: list[str],
+) -> tsi.FeedbackQueryReq:
+    """Build a FeedbackQueryReq that returns feedback for any of ``refs``.
+
+    The refs are unioned (OR'd) into a single ``$in`` over ``weave_ref`` so
```
jtschoonhoven: Nit: I don't know why Claude has been using \``double-ticks`` but it's unnecessary to use more than one tick.
```suggestion
    """Build a FeedbackQueryReq that returns feedback for any of `refs`.

    The refs are unioned (OR'd) into a single `$in` over `weave_ref` so
```

weave/trace_server/trace_server_common.py
```diff
@@ -34,24 +49,81 @@ def make_feedback_query_req(
...
+    by_trace_id: dict[str, list[tsi.FeedbackDict]] = defaultdict(list)
+    by_conversation_id: dict[str, list[tsi.FeedbackDict]] = defaultdict(list)
+    by_span_id: dict[str, list[tsi.FeedbackDict]] = defaultdict(list)
+
+    for raw in feedback.result:
+        # `FeedbackQueryRes.result` is intentionally `dict[str, Any]` because
+        # callers can request arbitrary fields; this helper always passes
+        # FEEDBACK_QUERY_FIELDS, so the rows match FeedbackDict shape.
+        item = cast(tsi.FeedbackDict, raw)
+        ref = ri.parse_internal_uri(item.get("weave_ref", ""))
```
jtschoonhoven: If the item doesn't have a valid "weave_ref" then this will raise `InvalidInternalRef`. It might be better to log that exception rather than raise it so one bad row doesn't fail the entire request.

weave/trace_server/trace_server_interface.py
```diff
@@ -78,6 +78,11 @@ class FeedbackDict(TypedDict, total=False):
     creator: str | None
     created_at: datetime.datetime | None
     wb_user_id: str | None
+    # below were added in migration 007 (`007_add_refs_to_feedback`)
+    annotation_ref: str | None
+    runnable_ref: str | None
+    call_ref: str | None
+    trigger_ref: str | None
```
jtschoonhoven: There are other schemas that currently use `FeedbackDict`. Have those been updated to include these new fields? If not then you might want to define a new dict rather than modify one that's currently in use.

weave/trace_server/trace_server_common.py
```diff
@@ -34,24 +49,81 @@ def make_feedback_query_req(
...
+    """Group feedback rows by agent target kind + raw identifier."""
+    by_trace_id: dict[str, list[tsi.FeedbackDict]] = defaultdict(list)
+    by_conversation_id: dict[str, list[tsi.FeedbackDict]] = defaultdict(list)
+    by_span_id: dict[str, list[tsi.FeedbackDict]] = defaultdict(list)
+
+    for raw in feedback.result:
+        # `FeedbackQueryRes.result` is intentionally `dict[str, Any]` because
+        # callers can request arbitrary fields; this helper always passes
+        # FEEDBACK_QUERY_FIELDS, so the rows match FeedbackDict shape.
+        item = cast(tsi.FeedbackDict, raw)
```
jtschoonhoven: You should pretty much never use `cast`. Since you're passing `FEEDBACK_QUERY_FIELDS`, you actually _do_ know the shape of the response and you should be able to marshal it into a typed object without casting.

jtschoonhoven (commented):
Mostly looks good, but I think we can get proper type safety here if we don't use `cast`

weave/trace_server/trace_server_interface.py
bcsherma: is this needed for this pr? or is this just something else that happened to be missing

bcsherma (approved):
I have a few qs but not blocking

weave/trace_server/trace_server_interface.py
nikumar1206: im gonna revert this change.
i was previously using tsi.FeedbackDict as the types for feedback, but realized that in most spots we actually mark feedback with type Any.
this is something i should have reverted as well.

nikumar1206:
### Merge activity

* **May 12, 6:40 PM UTC**: @nikumar1206 merged this pull request with [Graphite](https://app.graphite.com/github/pr/wandb/weave/6685).

PR #6686, opened 2026-04-22
feat(weave): fold feedback into agent chat-view responses via include_feedback

nikumar1206:
## Description

- Fixes [WB-33571](https://coreweave.atlassian.net/browse/WB-33571)

Third of three stacked PRs for Phase 1 of human feedback on agent turns.

Adds \`include_feedback: bool\` to \`agent_traces_chat\` and \`agent_conversation_chat\`. When true, the handler adds feedback into the response: turn feedback onto the turn, step feedback onto each message by span_id, conversation feedback onto the conversation.

Stack: #6684 → #6685 → this

## Testing

- unit tests

[WB-33571]: https://coreweave.atlassian.net/browse/WB-33571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace_server/helpers.py
+++ b/tests/trace_server/helpers.py
@@ -1,3 +1,5 @@
+"""Shared helpers for trace_server tests."""
+
 import base64
 import uuid
 
--- a/weave/trace_server/agents/types.py
+++ b/weave/trace_server/agents/types.py
@@ -30,9 +30,13 @@ from weave.trace_server.agents.schema import (
 from weave.trace_server.interface.query import Query
 
 if TYPE_CHECKING:
-    from weave.trace_server.trace_server_interface import ProcessedResourceSpans
+    from weave.trace_server.trace_server_interface import (
+        FeedbackDict,
+        ProcessedResourceSpans,
+    )
 else:
     ProcessedResourceSpans = Any
+    FeedbackDict = dict
 
 SearchMessageRole = Literal[
     "",
@@ -539,7 +543,7 @@ class AgentChatMessage(BaseModel):
             )
         return self
 
-    feedback: list[dict[str, Any]] | None = None
+    feedback: list[FeedbackDict] | None = None
 
 
 class AgentTraceChatReq(BaseModel):
@@ -566,7 +570,7 @@ class AgentTraceChatRes(BaseModel):
         ),
     )
     messages: list[AgentChatMessage] = Field(default_factory=list)
-    turn_feedback: list[dict[str, Any]] | None = None
+    feedback: list[FeedbackDict] | None = None
 
 
 class AgentConversationChatReq(BaseModel):
@@ -607,7 +611,7 @@ class AgentConversationChatRes(BaseModel):
     has_more: bool = False
     limit: int = MAX_CONVERSATION_CHAT_TURNS
     offset: int = 0
-    conversation_feedback: list[dict[str, Any]] | None = None
+    feedback: list[FeedbackDict] | None = None
 
 
 class AgentSchema(BaseModel):
--- a/weave/trace_server/agents/types.py
+++ b/weave/trace_server/agents/types.py
@@ -539,12 +539,15 @@ class AgentChatMessage(BaseModel):
             )
         return self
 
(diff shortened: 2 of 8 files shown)


nikumar1206:
* **#6686** <a href="https://app.graphite.com/github/pr/wandb/weave/6686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6685** <a href="https://app.graphite.com/github/pr/wandb/weave/6685?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6684** <a href="https://app.graphite.com/github/pr/wandb/weave/6684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6567** <a href="https://app.graphite.com/github/pr/wandb/weave/6567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

tests/trace_server/helpers.py
jtschoonhoven, line 1: Nit: I don't think we want a module that just exports one 2-line function. If there's no other sensible place to put this function then maybe rename this file fixtures.py or unit_test_helpers.py or something like that so we can add to it later.

weave/trace_server/agents/types.py
```diff
@@ -388,6 +391,7 @@ class AgentTraceChatRes(BaseModel):
         ),
     )
     messages: list[AgentChatMessage] = Field(default_factory=list)
+    turn_feedback: list[dict[str, Any]] | None = None
```
jtschoonhoven: This needs a more specific type

weave/trace_server/agents/types.py
```diff
@@ -427,6 +432,7 @@ class AgentConversationChatRes(BaseModel):
     has_more: bool = False
     limit: int = MAX_CONVERSATION_CHAT_TURNS
     offset: int = 0
+    conversation_feedback: list[dict[str, Any]] | None = None
```
jtschoonhoven: Nit: "conversation" is already specified by the class so I think this should just be called "feedback".

Same for "turn_feedback" above.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6582,24 +6582,28 @@ def calls_score(self, req: tsi.CallsScoreReq) -> tsi.CallsScoreRes:
     # observability API is considered stable.
 
     def agent_spans_query(self, req: AgentSpansQueryReq) -> AgentSpansQueryRes:
-        return AgentQueryHandler(self._query).spans_query(req)
+        return AgentQueryHandler(self._query, self.feedback_query).spans_query(req)
```
jtschoonhoven, line 6600: I think it's a code smell that you're passing the instance method `self.feedback_query` as an argument and binding it to `AgentQueryHandler`. Is there another approach that could work?

bcsherma: The only reason we have this `AgentQueryHandler` pattern is because we want to move stuff out of `clickhouse_trace_server_batched.py` but given that our queries use each other maybe this isn't quite the right pattern. Maybe `AgentQueryHandler`should just take a `ClickhouseTraceServerBatched` or maybe its parent class, ie the whole trace server rather than just some methods?

nikumar1206: i think the current iteration is ok. 
AgentQueryHandler takes 2 server methods, instead of taking a whole server.
i think we can re-visit when/if we need to add more methods to AgentQueryHandler.

if thats ok, i think we can just add a docstring for this decision.

bcsherma: fair enough let's keep it moving

bcsherma (approved):

PR #6688, opened 2026-04-22
chore(weave): Remote scorer class

mscavezze-cw:
## Description

[WB-33547](https://coreweave.atlassian.net/browse/WB-33547)

Add Remote Scorer class

## Testing

unit tests

[WB-33547]: https://coreweave.atlassian.net/browse/WB-33547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/scorers/remote_scorer.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@register_object
+class RemoteScorer(Scorer):
+    """Scorer backed by a customer-managed HTTPS endpoint.
+
+    The Python SDK stores configuration for publication with monitors. Remote
+    scoring is **invoked from the Weave scoring worker**, which performs the
+    outbound ``POST`` and feedback writes; it does not run by calling
+    :meth:`score` in user code.
```
neutralino1, line 20: I wonder if this class should also be the one users use to encapsulate the actual scoring logic that will run inside their server. It could bring two advantages:

- If we eventually do offer a `.serve()` API, it could be defined on this class
- The scorer object would be persisted as an object in our backend and users could select it from a drop down instead of configuring it by hand in the UI form

Not a blocker here, just food for thought.

weave/trace_server/environment.py
```diff
@@ -134,6 +134,28 @@ def wf_scoring_worker_debounced_scoring_max_call_history() -> int:
...
+    When set, the scoring worker's remote scorer path may prefer this value
+    over those other sources; exact precedence is defined by that
+    implementation. When unset, the worker resolves credentials from other
+    configuration only.
+
+    Returns:
+        The token string, or None if unset. Whitespace-only values are treated as
+        None.
+    """
+    raw = os.environ.get("WF_SCORING_WORKER_REMOTE_SCORER_BEARER_TOKEN")
```
neutralino1: These `WF_` prefixes really bug me. They don't mean anything anymore. Should we avoid them going forward?

weave/scorers/remote_scorer.py
mscavezze-cw: That seems like a reasonable idea worth exploring. It could be elegant if it works out. I imagine that .serve could live in this class, and then  customers could make subclasses with their own custom logic.

I probably wouldn't put the implementation that makes the http call here (some reasons in [this comment](https://github.com/wandb/core/pull/42925/changes#r3138768387), and also this spike pr gives a little more context). Some of that reasoning might apply to the .serve() API, but to a lesser extent. I'd like to think about that a little more.

I think it's compatible with my current plan. A couple things might change, but not in a major way.

I do have a drop-down scorer selector in the works, just for switching between LLM and Remote. Adding additional scorers to it could be a natural evolution.

Package dependencies could be an issue, but as we discussed before, those can probably be solved.

weave/trace_server/environment.py
mscavezze-cw: They don't carry their original meaning any more. However, they may serve to identify these variables as a group, since they all share the `WF_` prefix. There can be value in consistency over multiple evolving forms of correctness.

My opinions here are not too strong. I'm curious if there's a particular reason why these bug you so much.

mscavezze-cw (commented):
Thanks for the comments!

weave/trace_server/environment.py
mscavezze-cw: I'm moving these env vars to a separate PR.

weave/scorers/remote_scorer.py
```diff
@@ -0,0 +1,88 @@
...
+from urllib.parse import urlparse
+
+from pydantic import Field, field_validator
+
+from weave.flow.scorer import Scorer
+from weave.trace.objectify import register_object
+from weave.trace.op import op
+
+
+def _validate_remote_scorer_endpoint_url(v: str) -> str:
```
gtarpenning: probably better to have helpers at the bottom of the file

weave/scorers/remote_scorer.py
```diff
@@ -0,0 +1,88 @@
...
+            "local development, e.g. http://127.0.0.1:8000/score)."
+        ),
+    )
+    config: dict[str, Any] | None = Field(
+        default=None,
+        description="Optional customer-defined JSON-serializable configuration.",
+    )
+
+    @op
+    def score(self, *, output: Any, **kwargs: Any) -> Any:
```
gtarpenning, line 67: not really sure what this is for based on the docstring

gtarpenning (approved):

weave/scorers/remote_scorer.py
mscavezze-cw: I've moved _validate_remote_scorer_endpoint_url down (but github doesn't seem to have updated comment placement)

weave/scorers/remote_scorer.py
mscavezze-cw: I've updated the docstring in RemoteScorer.score (now line 67)

PR #6690, opened 2026-04-23
chore(weave): Reduce usage of globals in the client

andrewtruong:
Changes post-processing and attributes to be stored on the client instead of globally


chance-wnb (approved):
LGTM.

PR #6691, opened 2026-04-23
fix(weave_ts): Attemp 2 Fixes issue client.responses.create(...).withResponse is not a function

chance-wnb:
# Preserve OpenAI SDK APIPromise semantics in the weave wrapper

Replaces #6673 (to be closed).

## Problem — two bugs

1. **`.withResponse is not a function`** on the weave-wrapped OpenAI client. Callers chaining any `APIPromise` helper (`.withResponse()`, `.asResponse()`, `.parse()`, `._thenUnwrap()`) threw, because weave's wrapper returned a plain `Promise<T>` in place of the SDK's native `APIPromise`.

2. **Streaming `.withResponse()` leaves traces pending forever.** Libraries like `@mariozechner/pi-ai` (used by `pi-coding-agent`) consume every streaming response via

   ```ts
   const { data, response } = await client.chat.completions.create(params).withResponse();
   ```

   When `.withResponse()` resolved to the SDK's raw `Stream`, iteration bypassed weave's chunk-tapping generator — the `finally` block that calls `finishCall` never fired, so every streaming call stayed in an "in-progress" state indefinitely.

   **Why this is load-bearing:** an async iterator has no "done" callback, so weave can only observe drain by sitting *inside* the iteration — intercepting every chunk, forwarding it to the caller, and firing `finishCall` in a generator `finally` when the underlying iterator completes (or throws). The tap also feeds the reducer that assembles the traced output. If anything hands the caller a stream weave hasn't wrapped, both signals — drain event and assembled content — are lost, and the trace never closes. This is the constraint every streaming fix has to respect.

## Why a new PR

#6673 addressed both bugs by layering a `Proxy` over `op()`'s returned `Promise<T>` (`wrapAsAPIPromiseLike`), capturing the `APIPromise` via a closure side-channel, and writing a special-case `.withResponse()` handler to splice weave's tapped stream back into the SDK-shaped result. Reviewers flagged the stack of hacks; looking harder at the root cause, the problem wasn't any one forwarding rule — it was the abstraction:

- `op()` assumes the wrapped function returns a "value-producing" Promise. It awaits the return, takes the awaited value for tracing, and hands the caller a **fresh** `Promise<T>`.
- OpenAI SDK (and other provider SDKs) return an `APIPromise` — a rich thenable whose callers rely on attached helpers, not just the awaited value.

Routing the second pattern through an abstraction built for the first forced every fix to be a compensation for something the outer layer had already thrown away. This PR removes that mismatch rather than patching around it.

## What's different

Skip `op()` for the two APIPromise-returning factories and drive `WeaveClient` primitives directly:

| | PR #6673 | This PR |
|---|---|---|
| Tracing lifecycle | Through `op()` | `client.pushNewCall` / `createCall` / `runWithCallStack` / `finishCall` / `finishCallWithException` called directly |
| Returned object | `Proxy(tracedPromise, { get, has })` wrapping a plain `Promise<T>` | The SDK's native `APIPromise`, built via `apiPromise._thenUnwrap(transform)` |
| APIPromise helpers (`withResponse`, `asResponse`, `parse`, `_thenUnwrap`, future additions) | Manually forwarded in the Proxy; `withResponse` specially spliced with `{...sdkResult, data}` | Inherited from `APIPromise` — no per-helper handling needed |
| Stream tracing | `op()`'s `WeaveIterator` logic + Proxy-level `withResponse` substitution | One helper `wrapStreamForTracing` invoked from the `_thenUnwrap` transform |
| Error path | `tracedPromise.catch(() => {})` to silence unhandled rejections | Side `.catch` on the traced APIPromise calls `finishCallWithException` and implicitly marks the rejection as handled |
| Closure side-channel for the APIPromise | Required (`pendingApiPromise`) | Removed — `originalCreate(...)` is called directly where we need it |
| `then/catch/finally` binding gymnastics | Needed (V8 rejects non-Promise receivers) | Not needed — we return a real Promise subclass |

## Changes worth reviewing

- **`traceOpenAICall` + `wrapStreamForTracing`** ([openai.ts]) — the new core. `traceOpenAICall` drives `createCall` / `_thenUnwrap` / `finishCall` / `finishCallWithException`; `wrapStreamForTracing` is where the Proxy-over-`Symbol.asyncIterator` tap lives. These two are where to spend review time.
- **Factory renames**: `makeOpenAIChatCompletionsOp` → `wrapOpenAIChatCompletionsCreate`, `makeOpenAIResponsesCreateProxy`/`Op` → `wrapOpenAIResponsesCreate`. Reflects that they no longer produce weave `Op`s. `makeOpenAIImagesGenerateOp` intentionally keeps its name — it still goes through `op()`, which is the right fit for a non-APIPromise return.
- **Test mocks upgraded** to look like `APIPromise`: `makeAPIPromiseShim<T>` exported from `openaiMock.ts`, used to wrap both branches of `makeMockOpenAIChat`. Without this, the integration's `_thenUnwrap` call hits `undefined` in tests.
- **New regression test** for streaming `.withResponse()` against `InMemoryTraceServer` — fails without the fix, passes with it.

## SDK coupling

Relies on `APIPromise._thenUnwrap` (present in `openai` v4.x and v6.x). The underscore prefix makes it a loose contract rather than a documented-stable API; if the SDK ever removes it, the fallback is the Proxy approach from #6673. Noted so a future reader knows where the escape hatch is.

## Test plan

- [x] All existing openai integration tests pass (4 suites, 24 tests).
- [x] New regression test for streaming + `.withResponse()` — verified to fail without the `_thenUnwrap` + `wrapStreamForTracing` substitution, and pass with it.
- [x] Manual verification against `@mariozechner/pi-coding-agent` — traces should now finish (not hang in "in-progress").

---

<a href="https://app.devin.ai/review/wandb/weave/pull/6691" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/sdks/node/src/__tests__/integrations/openai2.test.ts
+++ b/sdks/node/src/__tests__/integrations/openai2.test.ts
@@ -207,6 +207,57 @@ describe('OpenAI Integration', () => {
       // Verify wrapped matches unwrapped
       expect(wrappedResult).toEqual(unwrappedResult);
     });
+
+    it('should forward APIPromise.withResponse() through the chat completions wrapper', async () => {
+      const mockResponse = {
+        headers: new Map([['x-request-id', 'req_chat_id']]),
+      };
+      const parsedData = {
+        id: 'chat_test_id',
+        choices: [{message: {role: 'assistant', content: 'Hello'}}],
+        model: 'gpt-4',
+        usage: {prompt_tokens: 1, completion_tokens: 1, total_tokens: 2},
+      };
+
+      class MockAPIPromise {
+        then(onFulfilled: any, onRejected: any) {
+          return Promise.resolve(parsedData).then(onFulfilled, onRejected);
+        }
+        catch(onRejected: any) {
+          return Promise.resolve(parsedData).catch(onRejected);
+        }
+        finally(onFinally: any) {
+          return Promise.resolve(parsedData).finally(onFinally);
+        }
+        asResponse() {
+          return Promise.resolve(mockResponse);
+        }
+        async withResponse() {
+          return {
+            data: parsedData,
+            response: mockResponse,
+            request_id: mockResponse.headers.get('x-request-id'),
+          };
+        }
+      }
+
+      mockOpenAI.chat.completions.create = jest
+        .fn()
+        .mockImplementation(() => new MockAPIPromise());
+      wrappedOpenAI = wrapOpenAI(mockOpenAI);
+
+      const pending = wrappedOpenAI.chat.completions.create({
+        model: 'gpt-4',
+        messages: [{role: 'user', content: 'Hi'}],
+      });
+
+      expect(typeof pending.withResponse).toBe('function');
+      const {data, response, request_id} = await pending.withResponse();
+
+      expect(data).toEqual(parsedData);
+      expect(response).toBe(mockResponse);
+      expect(request_id).toBe('req_chat_id');
+    });
   });
 });
 
@@ -411,6 +462,74 @@ describe('OpenAI responses.create Integration', () => {
       });
 
       expect(mockOpenAI.responses.create).toHaveBeenCalledWith(options);
+    });
+
+    it('should forward APIPromise.withResponse() through the weave wrapper', async () => {
+      const mockResponse = {
+        headers: new Map([['x-request-id', 'req_test_id']]),
+      };
+      const parsedData = {
+        id: 'resp_test_id',
+        object: 'response',
+        status: 'completed',
+        model: 'gpt-4o-2024-08-06',
+        output: [
+          {
+            id: 'msg_test_id',
+            type: 'message',
+            status: 'completed',
+            content: [
+              {
+                type: 'output_text',
+                annotations: [],
+                text: 'Hello! How can I help you today?',
+              },
+            ],
+            role: 'assistant',
+          },
+        ],
+        usage: {input_tokens: 1, output_tokens: 1, total_tokens: 2},
+      };
+
+      // Simulate APIPromise: thenable with .withResponse()/.asResponse().
+      class MockAPIPromise {
+        then(onFulfilled: any, onRejected: any) {
+          return Promise.resolve(parsedData).then(onFulfilled, onRejected);
+        }
+        catch(onRejected: any) {
+          return Promise.resolve(parsedData).catch(onRejected);
+        }
+        finally(onFinally: any) {
+          return Promise.resolve(parsedData).finally(onFinally);
+        }
+        asResponse() {
+          return Promise.resolve(mockResponse);
+        }
+        async withResponse() {
+          return {
+            data: parsedData,
+            response: mockResponse,
+            request_id: mockResponse.headers.get('x-request-id'),
+          };
+        }
+      }
+
+      mockOpenAI.responses.create = jest
+        .fn()
+        .mockImplementation(() => new MockAPIPromise());
+      wrappedOpenAI = wrapOpenAI(mockOpenAI);
+
+      const pending = wrappedOpenAI.responses.create({
+        model: 'gpt-4o-2024-08-06',
+        messages: [{role: 'user', content: 'Hello!'}],
+      });
+
+      expect(typeof pending.withResponse).toBe('function');
+      const {data, response, request_id} = await pending.withResponse();
+
+      expect(data).toEqual(parsedData);
+      expect(response).toBe(mockResponse);
+      expect(request_id).toBe('req_test_id');
     });
 
     it('should handle streaming responses and skip deltas', async () => {
--- a/sdks/node/src/__tests__/integrations/openai2.test.ts
+++ b/sdks/node/src/__tests__/integrations/openai2.test.ts
@@ -12,6 +12,35 @@ import {WeaveClient} from '../../weaveClient';
 jest.mock('../../generated/traceServerApi');
 jest.mock('../../wandb/wandbServerApi');
 
+// Stands in for the OpenAI SDK's APIPromise: thenable with the helpers
+// (.withResponse(), .asResponse()) the wrapper must preserve.
+function createMockAPIPromise<T>(
+  parsedData: T,
+  mockResponse: {headers: {get(key: string): string | null | undefined}}
+) {
+  return {
+    then(onFulfilled: any, onRejected: any) {
+      return Promise.resolve(parsedData).then(onFulfilled, onRejected);
+    },
+    catch(onRejected: any) {
+      return Promise.resolve(parsedData).catch(onRejected);
+    },
+    finally(onFinally: any) {
+      return Promise.resolve(parsedData).finally(onFinally);
+    },
+    asResponse() {
+      return Promise.resolve(mockResponse);
+    },
+    async withResponse() {
+      return {
+        data: parsedData,
+        response: mockResponse,
+        request_id: mockResponse.headers.get('x-request-id'),
+      };
+    },
+  };
+}
+
 describe('OpenAI Integration', () => {
   let mockOpenAI: any;
   let wrappedOpenAI: any;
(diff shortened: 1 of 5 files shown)


chance-wnb:
* **#6691** <a href="https://app.graphite.com/github/pr/wandb/weave/6691?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6691?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

andrewtruong (approved):
Do you need to handle `asResponse` as well?

chance-wnb:
> Do you need to handle `asResponse` as well?

Thanks Andrew, I think what you mentioned is indeed worth considering. 

What `asResponse()` does: get the raw `Response` instance without parsing it. 

Without parsing it, the response body is not consumed. So no LLM message is returned back, it has nothing to do with our tracing business. 

Users might have a reason to call `asResponse()`, but I feel that's too exceptional. If they don't drain, the trace is left open — we started it before the SDK call returned. Nothing fails, but the trace stays in-progress indefinitely. For callers explicitly opting out of parsing, that's consistent: they told the SDK "I don't want the parsed body," and we have nothing meaningful to record.

however if they do, then they not only hacked us around, but they also broke OpenAI SDK's mechanism. asResponse() is an explicit SDK escape hatch — callers who use it are opting out of the OpenAI SDK's parsing layer. Weave not observing what they do after that is consistent with that opt-out. At the same time, I feel it is pretty hard to patch against this case. So maybe we should accept it as a limitation and leave this edge case alone until an explicit ask to address this situation.

sdks/node/src/integrations/openai.ts
```diff
@@ -444,29 +462,176 @@ export function summarizer(result: any) {
...
+    startTime,
+    undefined,
+    {kind: 'llm'}
+  );
+
+  const apiPromise = client.runWithCallStack(newStack, () =>
+    originalCreate(...callArgs)
+  );
+
+  const traced = apiPromise._thenUnwrap((value: any) => {
```
rgao-coreweave, line 557: **[Should Fix]** No runtime guard for missing `_thenUnwrap`

`apiPromise._thenUnwrap(...)` is called unconditionally. `_thenUnwrap` is an underscore-prefixed (private) method on the OpenAI SDK's `APIPromise`. If a future SDK version removes or renames it, users get an opaque `TypeError` at runtime with no diagnostic pointing to the SDK version incompatibility.

A simple guard would make the failure mode graceful — just detect the missing method and fall back to returning the plain promise (tracing degrades but the call doesn't crash).

rgao-coreweave (commented):
Code review report: https://github.com/wandb/weave/pull/6691

sdks/node/src/__tests__/integrations/openai2.test.ts
```diff
@@ -207,6 +236,37 @@ describe('OpenAI Integration', () => {
       // Verify wrapped matches unwrapped
       expect(wrappedResult).toEqual(unwrappedResult);
     });
+
+    it('should forward APIPromise.withResponse() through the chat completions wrapper', async () => {
```
rgao-coreweave: **[Should Fix]** Test names are misleading — these only test the no-client passthrough

No global client is initialized in this describe block, so `getGlobalClient()` returns null and the wrapper falls through to `originalCreate(...)`. The test name "should forward APIPromise.withResponse() through the chat completions wrapper" implies it tests the tracing path, but it only tests the passthrough case.

Additionally, `createMockAPIPromise` (defined above) lacks `_thenUnwrap`, so if someone adds a client to these tests, they'd break.

Consider either renaming the tests to clarify (e.g., "should preserve APIPromise when no client is initialized") or adding `_thenUnwrap` to `createMockAPIPromise` and initializing a client to actually test the traced path.

sdks/node/src/integrations/openai.ts
rgao-coreweave, line 557: **[Question]** Is there a plan for the `_thenUnwrap` escape hatch?

The PR description mentions "the fallback is the Proxy approach from #6673" if `_thenUnwrap` is ever removed. Is there an intent to keep #6673's code documented somewhere so a future developer can quickly implement the fallback? Or is the plan to fix it when/if it breaks?

andrewtruong:
OK, I think that's fine.  If we get reports about this we can fix it then.  Thanks @chance-wnb !

sdks/node/src/integrations/openai.ts
chance-wnb: Great suggestion, yes, a guard is needed. I added a early check above.

chance-wnb: Great question. Here’s what I think we should do if that happens:

In this file:
https://github.com/wandb/weave/blob/a648a679d53fe7bf5b543f55bddb0c55e5e162bf/sdks/node/src/integrations/openai.ts#L671

we already have a branching mechanism for version-based patching. So we can say: for OpenAI SDK versions >= 4.0.0, <= 6.0.0, we apply cjsPatchOpenAI. For versions >= 7.0.0, we apply cjsPatchOpenAIGen2. That means cjsPatchOpenAIGen2 doesn’t need to worry about the Gen1 implementation at all, and is free to choose whatever patching strategy makes sense.

Even though we don’t currently need version branching, the system is already set up to support it.

As for how we’d refactor the code structure to support multiple versions, and how to handle the case where _thenUnwrap gets removed — let’s cross that bridge when we come to it :) It’s possible that #6673 won’t be the right approach when that time comes.

sdks/node/src/__tests__/integrations/openai2.test.ts
chance-wnb: Great catch. On a second thought, I don't think these tests test anything meaningful -- the case that we patch the OpenAI SDK without a globalClient is almost illusionary. The cases are better verified in `integrationOpenAI.test.ts` where we have the in-memory based client to check the trace state. 

I am pulling out these test cases now.

sdks/node/src/integrations/openai.ts
```diff
@@ -444,29 +463,189 @@ export function summarizer(result: any) {
...
+      for await (const chunk of stream) {
+        state = reduceFn(state, chunk);
+        yield chunk;
+      }
+    } finally {
+      finalizeFn(state);
+      await client.finishCall(
+        call,
+        state,
+        currentCall,
```
rgao-coreweave: **[Should Fix]** Stream iteration errors are recorded as successful completions

In `wrapStreamForTracing`, the `finally` block always calls `client.finishCall(...)` (success path), even if the stream threw an error during iteration. Compare with `traceOpenAICall`'s `traced.catch(...)` which correctly calls `finishCallWithException` for SDK-level errors.

If the underlying stream throws mid-iteration (network error, server error, etc.), the trace will show a "completed" status with partial output instead of recording the exception.

This is pre-existing behavior from the old `op()`-based `WeaveIterator`, but since this PR is rewriting stream handling, it's a good opportunity to fix it:

```typescript
async function* WeaveIterator() {
  let iterationError: any = undefined;
  try {
    for await (const chunk of stream) {
      state = reduceFn(state, chunk);
      yield chunk;
    }
  } catch (e) {
    iterationError = e;
    throw e;
  } finally {
    if (iterationError) {
      await client.finishCallWithException(
        call, iterationError, currentCall, parentCall,
        new Date(), startCallPromise
      );
    } else {
      finalizeFn(state);
      await client.finishCall(
        call, state, currentCall, parentCall,
        summarize, new Date(), startCallPromise
      );
    }
  }
}
```

chance-wnb: Great catch. Thanks! Adopted your solution.

rgao-coreweave (approved):
LGTM, thanks for the detailed description!

PR #6696, opened 2026-04-23
feat(weave): stub out Session SDK API surface

rgao-coreweave:
## Summary

[WB-33608](https://coreweave.atlassian.net/browse/WB-33608)

- Introduces the Session SDK API surface as functional stubs (no OTel span emission yet)
- Adds 5 core classes: `Session`, `Turn`, `LLM`, `Tool`, `SubAgent`
- Adds 4 data types: `Message`, `Usage`, `Reasoning`, `LogResult`
- Adds 11 top-level functions: `start_session`, `start_turn`, `start_llm`, `end_*`, `get_current_*`, `log_turn`, `log_session`
- Adds OTel attribute builder stubs in `session_otel.py`
- 52 tests covering the full API surface

## Key design decisions

- **`turn.llm()`, `turn.tool()`, `turn.subagent()`** — LLM calls, tool executions, and subagents are all direct children of a turn
- **`start_turn(user_message=...)`** for eager user message visibility
- **`include_content` privacy flag** on Session
- **Contextvar cross-module tracing**: `start_*` / `get_current_*` / `end_*` functions
- **One type per concept**: real-time and batch APIs share `LLM`/`Tool` — no separate batch-only types
- **No-op when context missing**: `start_turn()` without a session returns a disconnected object (no contextvar set, no crash). Warning to be added when OTel integration lands.
- **SubAgent has OOP API only** (no `start_subagent()` / `get_current_subagent()`) — always created via `turn.subagent()`, no cross-module use case

## What this does NOT include

OTel span creation, LiveSpanProcessor, BatchSpanProcessor, attribute population, `weave.init()` integration — all come in follow-up PRs.

## Test plan

- [x] 52 unit tests pass (`uv run --group test python -m pytest tests/trace/test_session.py -v`)
- [x] Ruff lint clean on all changed files
- [x] All exports importable from `weave` top-level

[WB-33608]: https://coreweave.atlassian.net/browse/WB-33608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace/session.py
```diff
@@ -0,0 +1,534 @@
...
+        if self._ended:
+            return
+        self._ended = True
+        self.ended_at = datetime.now(timezone.utc)
+        if self._token is not None:
+            _current_turn.reset(self._token)
+            self._token = None
+
+    def __enter__(self) -> Self:
+        self.started_at = datetime.now(timezone.utc)
```
rgao-coreweave: **[Should Fix]** `started_at` overwritten when combining `start_*()` with context managers

`start_turn()` (line 447) and `start_chat()` (line 472) both set `started_at` on the returned object. But `Turn.__enter__` (line 310) and `Chat.__enter__` (line 186) unconditionally overwrite `started_at`. So `with start_turn() as t:` sets the timestamp twice.

The `_token` field already has the right guard (`if self._token is None`), but `started_at` does not. While negligible for stubs, this will matter when OTel spans are emitted -- the span start time should reflect when the user logically started the turn/chat, not when the `with` block was entered.

**Suggestion:**
```python
def __enter__(self) -> Self:
    if self.started_at is None:
        self.started_at = datetime.now(timezone.utc)
    if self._token is None:
        self._token = _current_turn.set(self)
    return self
```

Same fix needed for `Chat.__enter__` (line 186) and `SubAgent.__enter__` (line 241).

weave/trace/session.py
```diff
@@ -0,0 +1,534 @@
...
+    model: str = "",
+    agent_name: str = "",
+) -> Turn:
+    """Create and activate a turn. Uses the current session if available."""
+    session = get_current_session()
+    if session is not None:
+        turn = session.start_turn(
+            user_message=user_message, model=model, agent_name=agent_name
+        )
+    else:
```
rgao-coreweave: **[Should Fix]** ContextVar leak when calling `start_turn()` consecutively without a session

Without an active session, calling `start_turn()` twice without ending the first creates a ContextVar ordering problem:

1. `t1 = start_turn()` -- contextvar = t1, token saves "previous=None"
2. `t2 = start_turn()` -- contextvar = t2, token saves "previous=t1"
3. `t2.end()` -- resets to t1 (stale, never ended)
4. `t1.end()` -- resets to None, but now t2's token is invalid

With a session, `session.start_turn()` auto-ends the previous turn (line 350-351). But the `else` branch doesn't.

**Suggestion:** Auto-end the current turn in the session-less path:
```python
else:
    current = get_current_turn()
    if current is not None:
        current.end()
    turn = Turn(agent_name=agent_name, model=model)
```

Same pattern needed for `start_chat()` (line 465). Also add a test for consecutive `start_turn()` without a session.

rgao-coreweave (commented):
Code review posted. See full report for detailed analysis.

weave/trace/session.py
```diff
@@ -0,0 +1,534 @@
...
+
+
+def start_session(
+    *,
+    agent_name: str = "",
+    model: str = "",
+    session_id: str = "",
+    session_name: str = "",
+    include_content: bool = True,
+    base_url: str = "",
```
rgao-coreweave: **[Should Fix]** `start_session()` silently discards connection parameters

`base_url`, `api_key`, `entity`, and `project` are accepted but never passed to the Session constructor (and Session has no corresponding fields). Users may write code passing these expecting them to work.

**Suggestion:** Either remove these params with a TODO comment referencing the follow-up PR, or add them as Session fields even if unused by stubs. Removing is safer -- avoids users depending on a signature that might change when OTel integration lands.

weave/trace/session.py
```diff
@@ -0,0 +1,534 @@
...
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: types.TracebackType | None,
+    ) -> bool:
+        self.end()
+        return False
+
+
+class SubAgent(BaseModel):
```
rgao-coreweave: **[Question]** SubAgent has no functional API -- intentional?

Session, Turn, and Chat all have both an OOP API (class methods + context managers) and a functional API (top-level functions + contextvars). SubAgent only has the OOP API -- no `_current_subagent` contextvar, no `start_subagent()`, etc. Is this intentional because sub-agents are always nested within a turn and accessed via `turn.subagent()`? Or will a functional API be added in a follow-up?

tests/trace/test_session.py
```diff
@@ -0,0 +1,444 @@
...
+        assert sa.name == "research-bot"
+        assert sa.model == ""
+
+    def test_llm_returns_llm(self) -> None:
+        sa = SubAgent(name="research-bot", model="gpt-4o-mini")
+        c = sa.llm(model="gpt-4o-mini")
+        assert isinstance(c, LLM)
+        assert c.model == "gpt-4o-mini"
+
+    def test_llm_inherits_model(self) -> None:
```
chance-wnb: This test can be merged with above. 

Same for 

- TestTurn.test_subagent_returns_subagent  + test_subagent_inherits_model 
- TestSubAgent.test_llm_returns_llm  + test_llm_inherits_model 
- TestSession.test_start_turn_inherits_agent_name  + test_start_turn_overrides_agent_name

weave/session/session.py
```diff
@@ -0,0 +1,495 @@
...
+    finish_reasons: list[str] = Field(default_factory=list)
+    input_messages: list[Message] = Field(default_factory=list)
+    output_messages: list[Message] = Field(default_factory=list)
+    started_at: datetime | None = None
+    ended_at: datetime | None = None
+
+    _ended: bool = PrivateAttr(default=False)
+    _token: Token[LLM | None] | None = PrivateAttr(default=None)
+
+    def output(self, content: str) -> LLM:
```
andrewtruong, line 126: should these be `output` or `log_output`?

tests/trace/test_session.py
```diff
@@ -0,0 +1,444 @@
...
+
+    def test_output_appends_assistant_message(self) -> None:
+        c = LLM(model="gpt-4o")
+        result = c.output("Hello!")
+        assert result is c  # chainable
+        assert len(c.output_messages) == 1
+        assert c.output_messages[0].role == "assistant"
+        assert c.output_messages[0].content == "Hello!"
+
+    def test_think_sets_reasoning(self) -> None:
```
chance-wnb: nit: could combine with above.

weave/trace/session.py
```diff
@@ -0,0 +1,495 @@
...
+        """Append an assistant message to output_messages."""
+        self.output_messages.append(Message(role="assistant", content=content))
+        return self
+
+    def think(self, content: str) -> LLM:
+        """Set reasoning/chain-of-thought content."""
+        self.reasoning = Reasoning(content=content)
+        return self
+
+    def attach_file(self, file_id: str, *, modality: str = "document") -> LLM:
```
andrewtruong: what does `modality = document` mean?

weave/session/session.py
```diff
@@ -0,0 +1,495 @@
...
+        """Attach a file reference. Stub — stores nothing yet."""
+        return self
+
+    def attach_image(
+        self, content: bytes | str, *, mime_type: str = "image/png"
+    ) -> LLM:
+        """Attach an image. Stub — stores nothing yet."""
+        return self
+
+    def attach_uri(self, uri: str, *, modality: str = "image") -> LLM:
```
andrewtruong, line 146: same here

weave/session/session.py
```diff
@@ -0,0 +1,495 @@
...
+    usage: Usage = Field(default_factory=Usage)
+    reasoning: Reasoning = Field(default_factory=Reasoning)
+    finish_reasons: list[str] = Field(default_factory=list)
+    input_messages: list[Message] = Field(default_factory=list)
+    output_messages: list[Message] = Field(default_factory=list)
+    started_at: datetime | None = None
+    ended_at: datetime | None = None
+
+    _ended: bool = PrivateAttr(default=False)
+    _token: Token[LLM | None] | None = PrivateAttr(default=None)
```
andrewtruong, line 120: what are these fields and why are they private?

chance-wnb (approved):
Beautiful code. Great work Rick.

I made some suggestions to consolidate some test cases. But if you feel they should be separate, feel free to leave as is.

weave/trace/session.py
```diff
@@ -0,0 +1,495 @@
...
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: types.TracebackType | None,
+    ) -> Literal[False]:
+        self.end()
+        return False
+
+
+class Session(BaseModel):
```
andrewtruong: Session is a group of Turn?

andrewtruong (approved):

weave/session/session.py
rgao-coreweave: Keeping `output` — it's the fluent builder pattern (`llm.output("Hello").think("...")`) rather than a logging side-effect. `log_output` implies it's writing somewhere, but this just appends to the in-memory `output_messages` list.

weave/trace/session.py
rgao-coreweave: Good call — removed it. The method name `attach_file` already implies document. Kept `modality` on `attach_uri` since a URI could point to an image, document, audio, etc.

weave/session/session.py
rgao-coreweave: Keeping `modality` on `attach_uri` since unlike `attach_file`/`attach_image`, a URI doesn't imply a specific content type.

weave/session/session.py
rgao-coreweave: `_ended` prevents double-end bugs (idempotent `end()` — calling it twice is a no-op). `_token` stores the ContextVar reset token so `end()` can restore the previous contextvar value. They're `PrivateAttr` so Pydantic excludes them from serialization/validation — they're internal bookkeeping, not part of the data model.

weave/trace/session.py
rgao-coreweave: Yes — a Session groups Turns by `conversation_id`. It has no OTel span of its own; it provides a shared `session_id` that each Turn's root span carries as an attribute. Think of it as a conversation thread.

tests/trace/test_session.py
rgao-coreweave: Done — merged the test pairs you flagged: `test_fluent_builders`, `test_llm_returns_llm_and_inherits_model`, `test_subagent_returns_subagent_and_inherits_model`, `test_start_turn_inherits_and_overrides_agent_name`, and the Turn version too.

tests/trace/test_session.py
rgao-coreweave: Done — merged into `test_fluent_builders`.

PR #6697, opened 2026-04-24
chore(weave): Retrofit `.js` extensions on relative imports

chance-wnb:
# Weave Node SDK — dual-build PR 2: retrofit `.js` extensions on relative imports

Mechanical change. Every relative import in `src/**/*.ts` now ends in `.js`. No logic changes, no build-tool changes, no published-output changes. Background and full plan: [Tech-Report-Should-the-Weave-Node-SDK-ship-a-dual-CJS-ESM-build](https://www.notion.so/wandbai/Tech-Report-Should-the-Weave-Node-SDK-ship-a-dual-CJS-ESM-build-348e2f5c7ef380149c80fdeb5a546f3c?source=copy_link) (§5 "Output shape", §7 PR 2).

## Why

PR 3 (next) emits **file-per-module** ESM (`bundle: false`) — the shape `openai`, `@openai/agents`, `@anthropic-ai/sdk` all ship. That requires explicit file extensions on every relative import in the emitted `.mjs` files; without them Node's ESM resolver throws `ERR_MODULE_NOT_FOUND`. This PR is the precondition.

The current CJS build is unaffected — TypeScript's resolver maps `'./foo.js'` back to `'./foo.ts'` at typecheck time under both `node` and `nodenext` resolution.

## What changed

- **228 imports rewritten** across 50 files (18 SDK source + 32 tests) via a one-shot codemod.

- **One manual fix:** [`src/index.ts`](sdks/node/src/index.ts) had `from './integrations'` (directory import); the codemod naively added `.js` → corrected to `./integrations/index.js`.

- **Jest config:** added `"^(\\.{1,2}/.*)\\.js$": "$1"` to `moduleNameMapper` in all three jest projects. Strips the `.js` suffix at jest's resolve time so its source-level resolver finds the underlying `.ts` files. Without this, every test would fail with `Cannot find module`. This simply because Jest loads source file (ts), not built files(js), so we have to retrofit for the Jest use case here

## What's not in this PR

- No `tsup`, no `package.json` exports changes — that's PR 3.
- No ESLint rule preventing future extensionless imports. ESLint isn't configured in this SDK; setting it up is its own work, deferred.

## Test plan

- [x] `pnpm run build` — clean.
- [x] `WANDB_API_KEY=dummy pnpm run test` — **183 passed, 7 skipped**. Same counts as `master`.

### DIFF
--- a/sdks/node/package.json
+++ b/sdks/node/package.json
@@ -56,7 +56,8 @@
         },
         "moduleNameMapper": {
           "^weave$": "<rootDir>/src/index.ts",
-          "^weave/(.+)$": "<rootDir>/src/$1"
+          "^weave/(.+)$": "<rootDir>/src/$1",
+          "^(\\.{1,2}/.*)\\.js$": "$1"
         }
       },
       {
@@ -74,7 +75,8 @@
         },
         "moduleNameMapper": {
           "^weave$": "<rootDir>/src/index.ts",
-          "^weave/(.+)$": "<rootDir>/src/$1"
+          "^weave/(.+)$": "<rootDir>/src/$1",
+          "^(\\.{1,2}/.*)\\.js$": "$1"
         }
       },
       {
@@ -97,7 +99,8 @@
         },
         "moduleNameMapper": {
           "^weave$": "<rootDir>/src/index.ts",
-          "^weave/(.*)$": "<rootDir>/src/$1"
+          "^weave/(.*)$": "<rootDir>/src/$1",
+          "^(\\.{1,2}/.*)\\.js$": "$1"
         }
       }
     ],
--- a/sdks/node/src/__tests__/attributes.test.ts
+++ b/sdks/node/src/__tests__/attributes.test.ts
@@ -1,8 +1,8 @@
 import {withAttributes, op} from 'weave';
 
-import {initWithCustomTraceServer} from './clientMock';
-import {InMemoryTraceServer} from '../inMemoryTraceServer';
-import {Settings} from '../settings';
+import {initWithCustomTraceServer} from './clientMock.js';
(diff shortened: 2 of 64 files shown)


chance-wnb:
* **#6697** <a href="https://app.graphite.com/github/pr/wandb/weave/6697?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6697?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6682** <a href="https://app.graphite.com/github/pr/wandb/weave/6682?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6737](https://github.com/wandb/weave/pull/6737) <a href="https://app.graphite.com/github/pr/wandb/weave/6737?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

andrewtruong (approved):

andrewtruong:
Oh actually 1 question: This is a one time change.  How do we have this behaviour going forward?

chance-wnb:
> Oh actually 1 question: This is a one time change. How do we have this behaviour going forward?

Yes, this question makes a lot sense.

1. Going forward, we as developers need to be aware of ESM relative import requirement and always add `.js` extension. 

2. The need for `.js` extension is just a fact for ESM based node.js development. AI knows about it. So when it generates new code. It will reference the current code base and should work correctly. 

3. In case it is missed, there will be build/lint errors. (eg. TS2835 or TS2307)

andrewtruong:
Great, thank you!

chance-wnb:
This PR is abandoned, since the approach in https://github.com/wandb/weave/pull/6737 doesn't need such retrofit.

PR #6699, opened 2026-04-24
feat(weave): dogfood calls_complete on wandb-entity projects

gtarpenning:
Adds an entity-based gate for the `calls_complete` write path so wandb-internal projects dogfood it first. Entities in `CALLS_COMPLETE_ENTITY_ALLOWLIST` (currently `{"wandb"}`) auto-opt-in regardless of `WEAVE_USE_CALLS_COMPLETE`.

Related to #6644 — de-risks flipping the default for all users by letting us validate on wandb/* traffic first. Expand the allowlist (or flip the env default) once we have confidence.

Entity is threaded from `weave.init()` → `init_weave_get_server` → `RemoteHTTPTraceServer`. `StainlessRemoteHTTPTraceServer` is unaffected (does not use this path).

### DIFF
--- a/tests/trace_server_bindings/test_http_behavior_remote.py
+++ b/tests/trace_server_bindings/test_http_behavior_remote.py
@@ -23,6 +23,7 @@ from tests.trace_server_bindings.conftest import (
     generate_start,
 )
 from weave.trace.display.term import configure_logger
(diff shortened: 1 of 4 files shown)


gtarpenning:
Superseded by #6644 — allowlist approach moved onto that branch.

PR #6705, opened 2026-04-27
chore(weave): fix flaky call link capture test

gtarpenning:
## Summary

The call link test was racing future callback logging against teardown of the test's temporary root logger handler. The CI failure shows the donut link did print, but pytest captured it after the test's StringIO capture had already missed it.

Original failure (https://github.com/wandb/weave/actions/runs/25004069199/job/73221851118):
```
FAILED tests/trace/test_call_behaviours.py::test_call_prints_link - AssertionError: assert 0 == 1
```

## Testing

Test-only change.

### DIFF
--- a/tests/trace/test_call_behaviours.py
+++ b/tests/trace/test_call_behaviours.py
@@ -1,7 +1,11 @@
 import pytest
 
 import weave
-from tests.trace.util import capture_output, flushing_callback
+from tests.trace.util import (
+    capture_output,
+    flush_and_wait_for_output,
+    flushing_callback,
+)
 from weave.trace.constants import TRACE_CALL_EMOJI
 from weave.trace.display.term import configure_logger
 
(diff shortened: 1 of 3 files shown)


nikumar1206 (approved):

tests/trace/test_call_behaviours.py
```diff
@@ -19,9 +23,9 @@ async def afunc():
 
 
 def test_call_prints_link(client):
-    callbacks = [flushing_callback(client)]
-    with capture_output(callbacks) as captured:
+    with capture_output([]) as captured:
```
chance-wnb: I wonder if every occasion of `capture_output()` should use your new pattern. May I ask why some don't need? (if all of them are converted, maybe we can eliminate the callback parameter? )

gtarpenning: hmm yes let me think about this, i didn't consider the broader scope...

chance-wnb (approved):
Awesome. Thanks Griffin!

PR #6708, opened 2026-04-27
fix(weave): lazy httpx.Client to honor env vars set after import

rgao-coreweave:
Fixes [WB-33539](https://wandb.atlassian.net/browse/WB-33539).

## Problem
The shared `httpx.Client` in `weave/utils/http_requests.py` was constructed at module import time, baking in `ssl_verify()` / `http_timeout()` before the caller had a chance to set env vars. Setting `WEAVE_INSECURE_DISABLE_SSL=true` *after* `import weave` was a no-op:

```python
import weave                                          # client created, verify=True
os.environ["WEAVE_INSECURE_DISABLE_SSL"] = "true"     # too late
weave.init(...)                                       # SSL fails, surfaces as
                                                      # "Weave is not available on the server"
```

## Fix
Build the `httpx.Client` lazily on first use instead of at module import time. Env-driven config is read when the client is constructed.

```python
def get_client() -> httpx.Client:
    """Return the shared httpx.Client, building it on first use."""
    global _client
    with _client_lock:
        if _client is None:
            _client = httpx.Client(
                event_hooks={"request": [_log_request], "response": [_log_response]},
                timeout=http_timeout(),
                limits=CLIENT_LIMITS,
                verify=ssl_verify(),
            )
        return _client
```

## Contract: frozen after first use
**Once the client is built, its config is immutable.** Changing `WEAVE_INSECURE_DISABLE_SSL` or `WF_HTTP_TIMEOUT` after the first HTTP call has no effect.

This is the same contract `httpx` itself enforces: SSL config is owned by the connection pool, and `httpx.Client.get/post/send` do not accept a per-request `verify=` ([encode/httpx#554](https://github.com/encode/httpx/issues/554)). It also matches the openai and anthropic Python SDKs, which build their `httpx.Client` eagerly per SDK instance and never reconfigure it.

Callers who need to change SSL/timeout config must set the env var **before** the first HTTP call (typically before `weave.init`).

Both behaviors are pinned by tests:
- `test_env_var_takes_effect_when_set_after_import` — the WB-33539 regression
- `test_config_is_frozen_after_first_use` — documents the intentional freeze

## Out of scope (follow-up tickets)
- Six other `httpx.Client` instances in the codebase ignore `ssl_verify()` entirely (`weave/chat/completions.py`, `weave/chat/inference_models.py`, `weave/utils/pypi_version_check.py`, `weave/trace_server/costs/update_costs.py`, `weave/trace_server/model_providers/{model_providers.py,catalog_augment_hf.py}`). `WEAVE_INSECURE_DISABLE_SSL=true` will fix the trace ingestion path but not these. Worth centralizing through `get_client()` in a follow-up.

## Test plan
- [x] `uvx ruff check` clean
- [x] `tests/utils/test_http_requests.py` + `tests/utils/test_ssl_verification.py` — 16/16 pass
- [x] `tests/trace_server_bindings/test_http_behavior_remote.py` — 13/13 pass
- [ ] CI green

<details>
<summary>Performance (irrelevant to correctness, included for completeness)</summary>

`import weave`, n=20 cold subprocesses each:

| Op                          | Before (eager)                                     | After (lazy)                                       | Δ                          |
|-----------------------------|----------------------------------------------------|----------------------------------------------------|----------------------------|
| `import weave` (cold)       | 1428.7 ms median (mean 1439.5, stdev 42.7)         | 1421.3 ms median (mean 1422.7, stdev 67.3)         | within noise               |
| `httpx.Client()` ctor       | 24.6 ms median                                     | 24.8 ms median                                     | unchanged (sanity check)   |

The ~25 ms `httpx.Client()` construction simply moves from import to the first HTTP request, where it's negligible next to network latency.
</details>

[WB-33539]: https://coreweave.atlassian.net/browse/WB-33539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/utils/test_http_requests.py
+++ b/tests/utils/test_http_requests.py
@@ -25,7 +25,7 @@ def clear_http_env(monkeypatch):
 
 
 def test_client_uses_default_httpx_transport():
-    assert isinstance(http_requests.client._transport, httpx.HTTPTransport)
+    assert isinstance(http_requests.get_client()._transport, httpx.HTTPTransport)
 
 
 def test_request_hook_logs_when_enabled(monkeypatch):
--- a/tests/utils/test_ssl_verification.py
+++ b/tests/utils/test_ssl_verification.py
@@ -39,48 +39,69 @@ class TestSslVerifyEnv:
 class TestHttpxClientSslVerify:
     """The shared httpx.Client in http_requests respects ssl_verify().
 
-    The client is now created lazily on first use, so we use ``reset_client()``
-    to drop the cached client and let the next access re-read the env var.
+    The client is built lazily on first call to ``get_client()`` and then
+    frozen — to assert behavior tied to env vars read at construction time
+    we drop the cached client first via ``_reset_client_for_tests``.
     """
 
     def test_client_verify_disabled(self, monkeypatch):
         monkeypatch.setenv(WEAVE_INSECURE_DISABLE_SSL, "true")
-        http_requests.reset_client()
+        http_requests._reset_client_for_tests()
         try:
             # httpx bakes the verify flag into an ssl_context on the transport pool;
             # CERT_NONE means SSL certificate verification is disabled.
-            transport = http_requests.client._transport
-            assert isinstance(transport, httpx.HTTPTransport)
-            pool = transport._pool
-            assert pool._ssl_context.verify_mode.name == "CERT_NONE"
+            client = http_requests.get_client()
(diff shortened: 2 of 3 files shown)


rgao-coreweave:
Closing this due to avoiding global var.


Done in https://github.com/wandb/weave/pull/7061

PR #6709, opened 2026-04-27
chore(weave): add ch query failure prediction guard

gtarpenning:
Addresses: https://coreweave.atlassian.net/browse/WB-33069

## Summary
- Adds `max_estimated_execution_time`, `timeout_before_checking_execution_speed=5`, `timeout_overflow_mode=throw` to read-query settings so SELECTs with bad plans fail fast.
- Splits command settings out of query settings so mutations skip the guard.
- Single env var `WF_CLICKHOUSE_DISABLE_QUERY_FAILURE_PREDICTION` is the kill-switch.

## Testing
- unit test


gtarpenning:
* **#6710** <a href="https://app.graphite.com/github/pr/wandb/weave/6710?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6709** <a href="https://app.graphite.com/github/pr/wandb/weave/6709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6709?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/clickhouse_trace_server_settings.py
```diff
@@ -39,19 +39,41 @@
 # https://clickhouse.com/docs/operations/settings/settings#max_execution_time
 DEFAULT_MAX_EXECUTION_TIME = 60 * 1  # 1 minute
 
+# https://clickhouse.com/docs/operations/settings/settings#timeout_before_checking_execution_speed
+DEFAULT_TIMEOUT_BEFORE_CHECKING_EXECUTION_SPEED = 5
```
mscavezze-cw, line 55: Why not add this to environment.py?

weave/trace_server/clickhouse_trace_server_settings.py
```diff
@@ -39,19 +39,41 @@
 # https://clickhouse.com/docs/operations/settings/settings#max_execution_time
 DEFAULT_MAX_EXECUTION_TIME = 60 * 1  # 1 minute
 
+# https://clickhouse.com/docs/operations/settings/settings#timeout_before_checking_execution_speed
```
mscavezze-cw, line 54: Consider adding a one-line description here, in addition to the docs link:
- it would lower the cost of understanding what this constant means
- maybe it would make LLMs work better, since there would be fewer http calls and less context bloat
- the docs don't seem especially informative - something tailored to our usage might be more easily understandable.

mscavezze-cw: For example,

"We don't bother calculating estimated execution time for queries that are reasonably quick, because that would be unnecessary overhead."

weave/trace_server/clickhouse_trace_server_settings.py
mscavezze-cw, line 55: There's some interaction between the timeout and the max estimated execution time, right? Do we want some kind of log letting us know if the settings are incompatible, and documenting the consequence of that incompatibility?

tests/trace_server/test_clickhouse_trace_server_settings.py
```diff
@@ -0,0 +1,57 @@
+import importlib
+
+from weave.trace_server import clickhouse_trace_server_settings as ch_settings
+
+
+def test_clickhouse_query_failure_prediction(monkeypatch):
```
mscavezze-cw: Consider splitting into multiple tests. The benefits include
- improved readability, since the purpose of each would be more obvious
- improved reporting when failures occur

mscavezze-cw (commented):
Looks pretty good. With the tests all in a single method, it's hard to understand whether the test coverage is complete or has gaps.

weave/trace_server/clickhouse_trace_server_settings.py
gtarpenning: do we think we need that? I had it initially, but it seems unlikely that we will want to tune this. could totally be wrong though

tests/trace_server/test_clickhouse_trace_server_settings.py
gtarpenning: Noted, I will also mention some cons just for posterity:

Cons include: 
- harder to maintain 
- longer test time

weave/trace_server/clickhouse_trace_server_settings.py
gtarpenning: made this impossible by aligning hard max timeouts with estimated values

weave/trace_server/clickhouse_trace_server_settings.py
gtarpenning: yeah good idea, done

tests/trace_server/test_clickhouse_trace_server_settings.py
gtarpenning: broke them out into more functional oriented tests

weave/trace_server/clickhouse_trace_server_settings.py
mscavezze-cw: It's pretty easy to add now. If we deploy this and then later want to tune it, we'll have to make a PR, get it reviewed, deal with any flaky tests, and then deploy.

The cost of having it tunable and never tuned is pretty low - just a little bit more code in a file that is rarely viewed or edited.

mscavezze-cw: I see the alignment between max time and max estimated time. (`DEFAULT_MAX_ESTIMATED_EXECUTION_TIME = DEFAULT_MAX_EXECUTION_TIME`)

I was thinking of the dependency between `DEFAULT_TIMEOUT_BEFORE_CHECKING_EXECUTION_SPEED` and those. A timeout that was longer than those would effectively defeat the feature, and seems like a configuration issue. Reading the new comment about how `ClickHouse extrapolates from rows already scanned`, it seems like there might be a minimum value below which the quality of the extrapolations would suffer and we might erroneously cancel queries. 

That might also be a reason to make `DEFAULT_TIMEOUT_BEFORE_CHECKING_EXECUTION_SPEED` configurable, so that we could have a quick response to outages. It might be worth adding some logging or alerting on the number of canceled queries, with query text, so that we could understand if this is happening.

weave/trace_server/clickhouse_trace_server_settings.py
```diff
@@ -36,22 +36,62 @@
...
+# longer than this.
 # https://clickhouse.com/docs/operations/settings/settings#max_execution_time
 DEFAULT_MAX_EXECUTION_TIME = 60 * 1  # 1 minute
 
+# Projection-based ceiling — ClickHouse extrapolates from rows already scanned
+# and aborts early if the projected total runtime would exceed this. Lets us
+# fail fast on doomed queries without waiting for max_execution_time.
+# Available in ClickHouse 24.1+.
+# https://clickhouse.com/docs/operations/settings/settings#max_estimated_execution_time
+DEFAULT_MAX_ESTIMATED_EXECUTION_TIME = DEFAULT_MAX_EXECUTION_TIME
```
mscavezze-cw, line 49: I think we're making some assumptions here:
* the accuracy of clickhouse's estimate of execution time
* the stability of the relationship between estimated and actual estimated time, with respect to clickhouse versions and query shapes

Given that those assumptions might not be true for all clickhouse versions or customers, I wonder if we want to have these configurable with independent environment variables.

weave/trace_server/clickhouse_trace_server_settings.py
mscavezze-cw: Thinking more [about it](https://github.com/wandb/weave/pull/6709/changes#r3157497536), enforcing relationships between these values might not be foolproof.

mscavezze-cw (approved):
I left some subjective comments. It's probably fine to merge as-is, but there's some potential that we'll want to revisit some things in the future once we have a better idea how the system works with real queries. Additional logging and tunables could make that easier.

gtarpenning:
[Here](https://us5.datadoghq.com/dashboard/bji-qgv-fru?fromUser=false&graphType=flamegraph&historicalData=true&index=%2A&refresh_mode=sliding&shouldShowLegend=true&traceQuery=&from_ts=1776874553171&to_ts=1777479353171&live=true) is the dashboard that tracks errors, with a special section for this error case. @mscavezze-cw

PR #6711, opened 2026-04-27
chore(weave): fix flaky call query filters

gtarpenning:
## Summary

_sigh_ just retry these ones...

The eventual consistency issue is especially hard when you are querying for _some_ number of calls. we don't want to retry in the client NOR server, because zero is totally legitimate...

Original failure ([GitHub Actions job](https://github.com/wandb/weave/actions/runs/25014685065/job/73259283402?pr=6695)):
```
Task failed: AttributeError: 'str' object has no attribute 'result'\n
```
## Testing
Test-only change

### DIFF
--- a/tests/trace/test_client_trace.py
+++ b/tests/trace/test_client_trace.py
@@ -495,6 +495,12 @@ def get_all_calls_asserting_finished(
     return res
 
 
+def flush_and_settle(client: ClientType) -> None:
+    client.flush()
+    time.sleep(0.2)
+    client.flush()
+
+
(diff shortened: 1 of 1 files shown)


nikumar1206 (approved):

nikumar1206:
consider adding a delay if needed, but lgtm

gtarpenning:
> consider adding a delay if needed, but lgtm

i considered (check git history) but really this should be passing 90% of the time anyway, it only fails when we get unlucky

PR #6719, opened 2026-04-28
fix(weave): make future executor flush wait for callbacks

gtarpenning:
## Summary

Strict red/green TDD split-out for the `FutureExecutor.flush()` callback barrier.

`FutureExecutor.flush()` used to wait on a snapshot of active futures. With `then(...)`, a completed future can run a done callback that submits/runs chained work after that snapshot is already drained, so `flush()` can return before the chained callback work is complete.

This PR adds exactly one regression test and then fixes `flush()` by tracking logical futures, including the public future returned by `then()`.

## Red Phase

Red commit: `d9b7fdec237554d0c9eadce884a99bd912caf112` (`test(weave): expose future executor flush callback race`)

The commit adds only `tests/trace/concurrent/test_futures.py::test_flush_waits_for_then_callback_work`. It blocks inside a `then(...)` callback, calls `flush()` from another thread, and asserts that `flush()` does not return until the callback is released.

Red result on that commit:

```text
FAILED tests/trace/concurrent/test_futures.py::test_flush_waits_for_then_callback_work
E       assert not True
```

CI red result on that commit:

- Run: https://github.com/wandb/weave/actions/runs/25063213064
- Failed job: https://github.com/wandb/weave/actions/runs/25063213064/job/73423002037
- Job: `Trace non-trace server tests (3, 13, trace_no_server)`

## Green Phase

Green commit: `1d60251505c5f04cf0122f37dbd87dc9902e3a84` (`fix(weave): wait for future callbacks during flush`)

The patch:

- adds `_track_future(...)` as the single owner of `_active_futures` membership
- tracks the `then()` `result_future` directly, because that is the logical work unit callers receive and `flush()` should wait on
- guards dependency callbacks with a lock/boolean so multiple completed dependencies cannot submit the chained callback twice
- keeps failure logging inside the tracked-future completion barrier, so `flush()` cannot return before executor-owned error logs are emitted
- makes `flush()` loop until tracked logical futures are drained, recomputing timeout across the full drain

### DIFF
--- a/weave/trace/concurrent/futures.py
+++ b/weave/trace/concurrent/futures.py
@@ -32,6 +32,7 @@ from __future__ import annotations
 import atexit
 import concurrent.futures
 import logging
+import time
 from collections.abc import Callable
 from concurrent.futures import Future, wait
 from contextvars import ContextVar
@@ -116,7 +117,9 @@ class FutureExecutor:
         Returns:
             Future[U]: A new Future object representing the result of applying g to the results of the futures.
         """
-        result_future: Future[U] = Future()
+        result_future: Future[U] = self._track_future(Future(), log_exception=False)
+        callback_submit_lock = Lock()
+        callback_submitted = False
 
         def callback() -> None:
             try:
@@ -139,12 +142,20 @@ class FutureExecutor:
 
             self._safe_add_done_callback(g_future, on_g_done)
 
+        def submit_callback_once() -> None:
+            nonlocal callback_submitted
+            with callback_submit_lock:
+                if callback_submitted:
+                    return
+                callback_submitted = True
+            self._safe_submit(callback)
+
         def on_done_callback(_: Future[T]) -> None:
             if all(fut.done() for fut in futures):
-                self._safe_submit(callback)
+                submit_callback_once()
 
         if not futures:
-            self._safe_submit(callback)
+            submit_callback_once()
         else:
             for f in futures:
                 self._safe_add_done_callback(f, on_done_callback)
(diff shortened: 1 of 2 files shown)


weave/trace/concurrent/futures.py
```diff
@@ -167,28 +178,34 @@ def flush(self, timeout: float | None = None) -> bool:
...
             raise RuntimeError("Cannot flush from within a thread")
 
-        for future in concurrent.futures.as_completed(futures_to_wait, timeout=timeout):
-            try:
-                future.result()
-            except Exception as e:
-                if get_raise_on_captured_errors():
-                    raise
+        start = time.monotonic()
+        while True:
```
chance-wnb, line 198: How about add some comments here? I think it is the key step of your work. (Spectacular, I enjoyed the show a lot!) 

You might want to explain why it would loop and what each loop does. ( It took me quite some effort to connect the dots 👍 )

weave/trace/concurrent/futures.py
```diff
@@ -116,7 +117,9 @@ def then(self, futures: list[Future[T]], g: Callable[[list[T]], U]) -> Future[U]
         Returns:
             Future[U]: A new Future object representing the result of applying g to the results of the futures.
         """
-        result_future: Future[U] = Future()
+        result_future: Future[U] = self._track_future(Future(), log_exception=False)
+        callback_submit_lock = Lock()
+        callback_submitted = False
```
chance-wnb, line 127: I think it's also worth adding some comments here to explain under which scenario the callback would be added multiple times so its necessity is justified.

chance-wnb (approved):
Fantastic work! Added some suggestions for comments

weave/trace/concurrent/futures.py
gtarpenning: Done in fc94d086 — added a comment block above the loop explaining the snapshot → drain → re-check cycle and why chained work shows up mid-flight. Thanks for the kind words!

weave/trace/concurrent/futures.py
gtarpenning: Done in fc94d086 — comment now spells out the concurrent-completion race that lets multiple `on_done_callback` threads pass the `all(fut.done())` gate, which is why we need the lock + flag.

PR #6721, opened 2026-04-28
fix(weave): handle error in OpenAI raw-response unwrap

gtarpenning:
## Summary

Tightens `maybe_unwrap_api_response` so the OpenAI integration handles `httpx.ResponseNotRead` from `LegacyAPIResponse.parse()` / `APIResponse.parse()` by reading the underlying `http_response` and re-parsing, instead of silently falling back to the raw response object via a broad `try/except`.

## Original Flake

`LegacyAPIResponse.parse() can fail in streaming contexts due to async race conditions (e.g. httpx.ResponseNotRead on Windows CI).` That broad fallback was first introduced for the flaky crewai async test (#6433); this PR replaces it with a targeted handler that actually unwraps the response so tracing sees the parsed `ChatCompletion` instead of the raw legacy response.

### Testing
- adds 3 tests, codecov reports clean

### DIFF
--- a/tests/integrations/openai/openai_test.py
+++ b/tests/integrations/openai/openai_test.py
@@ -18,6 +18,12 @@ from weave.integrations.openai.openai_sdk import (
 from weave.trace.weave_client import WeaveClient
 
 model = "gpt-4o"
+RawChatCompletionResponse = (
+    LegacyAPIResponse[ChatCompletion] | APIResponse[ChatCompletion]
+)
+RawChatCompletionResponseCls = (
+    type[LegacyAPIResponse[ChatCompletion]] | type[APIResponse[ChatCompletion]]
+)
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


tests/integrations/openai/openai_test.py
```diff
@@ -21,6 +29,71 @@ def patch_openai() -> Generator[None, None, None]:
...
+            method="post",
+            url="/chat/completions",
+        ),
+        retries_taken=0,
+    )
+
+
+@pytest.mark.parametrize("response_cls", [LegacyAPIResponse, APIResponse])
+def test_maybe_unwrap_api_response_reads_unread_raw_response(
+    response_cls: type,
```
andrewtruong: This should be `LegacyAPIResponse | APIResponse` right

andrewtruong (approved):

tests/integrations/openai/openai_test.py
gtarpenning: yup done

PR #6723, opened 2026-04-28
test(weave): add test client flush barriers tests

gtarpenning:
## Summary

The trace test suite has been flaky on ClickHouse (e.g. `test_dataset_rows_ref`, `test_object_mismatch_project_ref_nested`, `test_dirty_model_op_retrieval`, `test_object_mutation_saving`) with `NotFoundError: Obj … not found` immediately after a publish. Investigation in #6717 showed two test-harness gaps that let one test's queued background writes leak into the next test's setup, racing the per-test ClickHouse reset:

1. `TestOnlyFlushingWeaveClient.__getattribute__` only flushed on the success path. If a wrapped client method raised, the autoflush was skipped — but the failing call could still have enqueued background writes via `future_executor`. Those writes then landed *after* the test failed, during the next test's truncate.
2. Tests routinely bypass the client and call `client.server.*` directly. Those direct calls had no flush barrier at all, so any background work from preceding client calls could be unobserved when the test read state through the raw server.
3. The `client` and `client_creator` fixture teardowns closed the server (or its cache) without first flushing, so pending background writes from the test body could outlive the fixture.

## Why we need `TestOnlyFlushingTraceServer` (the new proxy)

The race isn't `server.write() → server.read()` — raw `TraceServerInterface` methods are synchronous. The race is **`client.X() → client.server.Y()`**, because the *client* queues work onto `future_executor` before it ever reaches the server:

```
client.create_call(...)
  → future_executor.submit(server.call_start, ...)   # queued, not yet run
  → returns to test
client.server.calls_query_stats(...)                 # may miss the queued call
```

`TestOnlyFlushingWeaveClient`'s existing wrapper already flushes after every `client.*` method, so `client.X() → client.read()` is safe. The gap is when the read goes directly through `client.server.*` after a queueing client call — there's no barrier today. The proxy adds one before AND after each direct server call (the post-call flush covers cases where `Y()` itself triggers callbacks that enqueue more work).

The executor-aware guard (`_flush_client_if_safe`) is non-optional: if a server method is invoked from inside a `future_executor` worker (e.g. an `@op` running on the executor that calls back into `client.server.*`), calling `client._flush()` would wait for that very worker to finish → self-deadlock. The check on `executor._in_thread_context` lets the proxy skip the flush in that case.

Note: this is a mostly a test-harness issue, not a client correctness issue — `client.server` is private and real users generally only call `WeaveClient` public methods, which are responsible for their own read-after-write semantics; tests bypass that boundary to assert raw server state, so the harness has to recreate the flush guarantee manually.

## What's in this PR

- `TestOnlyFlushingWeaveClient.__getattribute__` now flushes in a `try/finally` so failed calls still drain queued writes.
- New `TestOnlyFlushingTraceServer` proxy wraps `client.server` and applies the same before/after flush barrier to direct `client.server.*` calls. It's executor-aware: it skips flushing when called from inside a `future_executor` thread to avoid deadlocks.
- New `_flush_for_teardown` / `_flush_test_client_for_teardown` helpers flush without re-entering the autoflush wrapper.
- The `client` and `client_creator` fixtures now flush before clearing the global client and closing the server/cache.

## Testing

test only change


tests/conftest.py
```diff
@@ -290,15 +294,83 @@ def __getattribute__(self, name):
...
+
+
+def _flush_test_client_for_teardown(client: weave_client.WeaveClient) -> None:
+    if isinstance(client, TestOnlyFlushingWeaveClient):
+        TestOnlyFlushingWeaveClient._flush_for_teardown(client)
+    else:
+        client._flush()
+
+
+class TestOnlyFlushingTraceServer:
```
mscavezze-cw, line 322: Consider extracting some of these classes to stand-alone files (`TestOnlyFlushingTraceServer`, `TestOnlyFlushingWeaveClient`, maybe others). There would be a couple of benefits:
- shorter files
- easier to quickly understand and review PRs
  - the file list would give more of an indication of what changed
  - it would be easier to navigate between changes
  - in the truncated view of the diff, it would be easier to tell which class is being modified
- maybe less context bloat for LLMs

tests/conftest.py
```diff
@@ -290,15 +294,83 @@ def __getattribute__(self, name):
...
+    """Flush a test client around direct client.server method calls.
+
+    Tests often bypass WeaveClient methods and call client.server.* directly.
```
mscavezze-cw: > Tests often bypass WeaveClient methods and call client.server.* directly.

Is this a reasonable thing to do, or is it weakening our tests? Would it make sense to move away from this pattern? Or at least document some guidance on when that's safe and when it's a problem? It seems like this might not be the only case where bypassing WeaveClient methods could lead to a problem.

tests/conftest.py
```diff
@@ -290,15 +294,83 @@ def __getattribute__(self, name):
...
+            object.__delattr__(self, name)
+        else:
+            delattr(self._server, name)
+
+    def __getattr__(self, name):
+        attr = getattr(self._server, name)
+        if name.startswith("_") or not callable(attr):
+            return attr
+
+        def wrapper(*args, **kwargs):
```
mscavezze-cw, line 384: Consider adding some of the PR description into the source code as comments or docstrings

tests/conftest.py
```diff
@@ -393,11 +465,14 @@ def client(zero_stack, request, trace_server, caching_client_isolation):
...
     finally:
-        weave_client_context.set_weave_client_global(None)
         try:
-            client.server.close()
-        except Exception:
-            pass
+            _flush_test_client_for_teardown(client)
+        finally:
+            weave_client_context.set_weave_client_global(None)
+            try:
```
mscavezze-cw: This is 3 layers of `try` statements. Consider whether some helper function(s) would improve readability.

tests/conftest.py
```diff
@@ -393,11 +465,14 @@ def client(zero_stack, request, trace_server, caching_client_isolation):
...
+                client.server.close()
+            except Exception:
+                pass
```
mscavezze-cw: Could you document why we just swallow exceptions?

tests/conftest.py
```diff
@@ -431,17 +506,20 @@ def client(
...
-            # The trace_server is shared across multiple clients within this
-            # fixture and will be cleaned up by its own fixture teardown.
             try:
-                client.server._cache.close()
-            except Exception:
-                pass
+                _flush_test_client_for_teardown(client)
+            finally:
+                weave_client_context.set_weave_client_global(None)
+                weave.trace.settings.parse_and_apply_settings(
```
mscavezze-cw: Why does this fixture call parse_and_apply_settings, but the client fixture does not?

If this is intentional, could you document why?

If this is not, this could be a signal that this cleanup code should be extracted to a helper that both fixtures use.

Either way, moving the cleanup code to a helper would probably improve readability, reduce the likelihood of them getting unintentionally out of sync, make the similarities and differences more obvious, and add an opportunity to make the differences self-documenting through the use of well named parameters.

tests/conftest.py
```diff
@@ -290,15 +294,83 @@ def __getattribute__(self, name):
...
+
+    def __init__(
+        self,
+        client: TestOnlyFlushingWeaveClient,
+        server: tsi.TraceServerInterface,
+    ) -> None:
+        object.__setattr__(self, "_client", client)
+        object.__setattr__(self, "_server", server)
+
+    def __setattr__(self, name, value):
```
mscavezze-cw, line 367: I'm curious why we need all these `__*attr__` functions, but too lazy to load this branch locally and ask AI.

mscavezze-cw (commented):
Thanks for addressing test flakes! I left a couple of comments.

tests/conftest.py
gtarpenning: agree, i think this is a bit out of scope, pr is already a bit bloated. Happy to do this as a follow up though

tests/conftest.py
gtarpenning: this happens in a zillion places in our tests, and is a reasonable thing to do because is users CURL or whatever these are the endpoints they hit. the real issue is when we _mix_ these, which is very unlikely to be done in user land.

tests/conftest.py
gtarpenning: yep done

tests/conftest.py
gtarpenning: extracted `_teardown_test_client` totally agree. 

in general i'm less strict with stuff like this in the test suite, but then again thats how we got in this mess in the first place.

tests/conftest.py
gtarpenning: done

tests/conftest.py
gtarpenning: The client fixture never mutates settings, so it has nothing to reset.

agree though, refactored and commented

tests/conftest.py
gtarpenning: added comment:

```python
Dispatch rules:
    - `__init__` uses `object.__setattr__` to bootstrap `_client`/`_server`,
      because our own `__setattr__` would forward to `_server` — which doesn't
      exist yet.
    - `__getattr__` intercepts method calls and adds the flush barrier.
    - `__setattr__` forwards writes to the underlying server. Tests *do* write
      through the proxy (e.g. `client.server.attribute_access_log = []` to
      reset the recorder); without forwarding, those writes would silently
      land on the proxy and the recorder underneath would never see them.
    - `__delattr__` mirrors `__setattr__`.
```

tests/conftest.py
```diff
@@ -380,6 +486,51 @@ def zero_stack():
...
+       context manager multiple times. The autouse `reset_digest_settings`
+       fires once per test, which is too coarse — we need a reset per
+       CM-exit so settings don't leak across nested usages.
+    4. Close the server. `client_creator` shares one trace_server across
+       multiple clients in a single test, so `close_cache_only=True`
+       closes only the caching middleware; the outer `trace_server`
+       fixture handles full teardown.
+
+    The server may already be partially torn down by the time we get here
+    (sqlite file removed, HTTP connection closed, etc). Teardown shouldn't
```
mscavezze-cw, line 513: It makes sense to me that we need to be resilient to this teardown order. I'm not sure I'd generalize this as far as saying that teardown shouldn't allow a green test to turn red. If that's happening for unexpected reasons, it might indicated an error with the test or the code being tested. I totally buy that there could be different schools of thought here.

mscavezze-cw (approved):
Thanks for making those updates!

tests/conftest.py
gtarpenning: i understand the concern... thinking about this in a followup.

PR #6724, opened 2026-04-28
fix(weave): handle newer OpenAI Agents span types

gtarpenning:
## Summary
Weave is out of sync with newer `openai-agents` tracing spans. Newer SDKs emit real `TaskSpanData` and `TurnSpanData` objects; the current processor only recognizes the older span classes, so these newer spans fall through to generic handling. That is why deployed traces can show turn work as `Unknown` and lose the task/turn metadata and usage fields.

This PR keeps runtime compatibility broad while covering the updated shape:

- optionally imports `TaskSpanData` / `TurnSpanData` from the user's installed SDK
- routes with `isinstance` when those classes exist, with a `span_data.type` fallback for older SDKs
- records task/turn display names, metadata, and usage metrics
- keeps the public `openai-agents>=0.0.4` extra unchanged
- bumps only the `openai_agents` nox shard to `openai-agents>=0.14.7` so CI exercises the real newer classes

SDK shape reference:
https://openai.github.io/openai-agents-python/ref/tracing/span_data/

### Testing
- adds regression tests

The regression test now uses real `TaskSpanData` / `TurnSpanData` from the upgraded nox shard and also monkeypatches the optional imports to `None` to prove the older-SDK fallback path still works.

Local older-version import check:

```text
openai-agents==0.2.0
TaskSpanData=None
TurnSpanData=None
```

### Reported by @zapulam

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -129,6 +129,11 @@ def tests(session: nox.Session, shard: str):
 
     session.run(*sync_args)
 
+    if shard == "openai_agents":
+        # Keep the public extra broad for runtime compatibility, but exercise the
+        # newer SDK span classes in CI.
+        session.run("uv", "pip", "install", "--upgrade", "openai-agents>=0.14.7")
+
     env = {
         k: session.env.get(k) or os.getenv(k)
         for k in [
--- a/tests/integrations/openai_agents/openai_agents_test.py
+++ b/tests/integrations/openai_agents/openai_agents_test.py
@@ -526,7 +526,7 @@ def test_newer_agent_task_and_turn_fields_prevent_unknown_blocks(
         ),
     ],
 )
-def test_usage_to_metrics_branches(usage, expected) -> None:
+def test_usage_to_metrics_branches(client: WeaveClient, usage, expected) -> None:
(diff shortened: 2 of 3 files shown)


noxfile.py
```diff
@@ -130,6 +129,11 @@ def tests(session: nox.Session, shard: str):
 
     session.run(*sync_args)
 
+    if shard == "openai_agents":
+        # Keep the public extra broad for runtime compatibility, but exercise the
+        # newer SDK span classes in CI.
+        session.run("uv", "pip", "install", "--upgrade", "openai-agents>=0.14.7")
```
andrewtruong, line 135: not sure how I feel about this, but approving as it's just tests

andrewtruong (approved):

noxfile.py
gtarpenning: yeah like i mentioned in the [thread](https://weightsandbiases.slack.com/archives/C03BSTEBD7F/p1777413041199989) the testing for thsi is not a great solution. We should consider @chance-wnb's feedback. Firmly in SDK domain though, don't want to step on your toes.

PR #6725, opened 2026-04-28
fix(weave): treat partial obj value-row miss as not-found

gtarpenning:
## Summary

`_select_objs_query` runs two SELECTs: one for object metadata, one for value rows. If the metadata row is visible but the value row hasn't replicated yet, the old code silently filled `val_dump` with `"{}"` — `obj_read` returned success and the client decoded a corrupted empty object.

Now: if any metadata row is missing its value row, return `[]`. `obj_read` raises `NotFoundError` and the existing `retry_on_not_found` wrapper reruns the read once replication catches up.

## Test plan
- [x] Added `test_select_objs_query_partial_value_miss_returns_empty` covering the partial-miss and happy-path cases
- [x] Full `test_clickhouse_trace_server_batched.py` suite passes (28/28)

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -6629,9 +6629,10 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         # All-or-nothing: if any metadata row is missing its value row (e.g. due
         # to ClickHouse replication lag), return empty so callers raise
         # NotFoundError and retry, instead of silently filling with "{}".
-        if any(
-            (obj.object_id, obj.digest) not in object_values for obj in metadata_result
-        ):
+        all_value_rows_found = all(
+            (obj.object_id, obj.digest) in object_values for obj in metadata_result
+        )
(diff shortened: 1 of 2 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6626,9 +6626,15 @@ def _select_objs_query(
             (object_id, digest, val_dump) = row
             object_values[object_id, digest] = val_dump
 
-        # update the val_dump for each object
+        # All-or-nothing: if any metadata row is missing its value row (e.g. due
+        # to ClickHouse replication lag), return empty so callers raise
+        # NotFoundError and retry, instead of silently filling with "{}".
+        if any(
```
andrewtruong: I feel like this would read better as `not all`

andrewtruong (approved):

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: agree, done and assigned var for extra clarity

PR #6727, opened 2026-04-28
fix(weave): use buffered _query for object reads (test flake repro)

gtarpenning:
## Summary

Isolates one change from #6717 to test in CI: switch `_select_objs_query` in `clickhouse_trace_server_batched.py` from `_query_stream` to `_query`.

Theory: `_query_stream` sets `server_wait=False` in clickhouse-connect, so the HTTP response can be acked before the query finishes server-side. For exact object reads issued immediately after a write, that opens a read-after-write gap and produces `NotFoundError` flakes (e.g. `test_dataset_rows_ref`). Buffered `_query` keeps `server_wait=True`.

This PR is solely to run CI 5x and see if the flake rate drops.


gtarpenning:
object streaming does not cause obj not found flake

PR #6728, opened 2026-04-28
feat(weave): infer typed attribute filters from literals

gtarpenning:
## Summary

Follow-up to #6678 and the review question about whether `$convert` is necessary when the comparison literal already carries the intended type. This proves the answer is "yes, we can infer it" for scalar attribute filters.

- Adds typed `attributes_map_str` / `_int` / `_float` / `_bool` columns via migration `031_add_attributes_map`.
- Populates those maps at ingest from `attributes_dump`, using the same dot-flattened keys the query builder reads.
- Routes `attributes.<path>` filters through the typed map when the type is known from either explicit `$convert` or from a scalar `$literal` peer.
- Keeps the row-level `mapContains(..., key)` guard so legacy or partially backfilled rows fall back to `JSON_VALUE(attributes_dump, ...)` instead of leaking ClickHouse map defaults.

## Why Users Might Want This

Filter builders and hand-written queries often already know the type from the literal they are comparing against: `3` means int, `0.9` means float, `true` means bool, and `"prod"` means string. Requiring callers to wrap the field in `$convert` exposes an internal query-shaping detail and is easy to miss. When missed, numeric/bool dynamic-field comparisons either stay on the slower JSON path or can hit ClickHouse common-type errors. Literal inference lets users write the natural query shape while still getting the typed-map fast path.

## Query Shape Changes

The explicit `$convert` form still works:

```json
{
  "$expr": {
    "$eq": [
      {"$convert": {"input": {"$getField": "attributes.retries"}, "to": "int"}},
      {"$literal": 3}
    ]
  }
}
```

The new inferred form is also accepted and emits the same typed-map hybrid:

```json
{
  "$expr": {
    "$eq": [
      {"$getField": "attributes.retries"},
      {"$literal": 3}
    ]
  }
}
```

For `calls_complete`, that becomes the equivalent of:

```sql
if(
  mapContains(calls_complete.attributes_map_int, {key:String}),
  calls_complete.attributes_map_int[{key:String}],
  toInt64OrNull(coalesce(nullIf(JSON_VALUE(calls_complete.attributes_dump, {json_path:String}), 'null'), ''))
) = {literal:Int64}
```

Literal dispatch rules:

- `str` literal -> `attributes_map_str`, JSON fallback stays string-shaped.
- `int` literal -> `attributes_map_int`, JSON fallback uses `toInt64OrNull(...)`.
- `float` literal -> `attributes_map_float`, JSON fallback uses `toFloat64OrNull(...)`.
- `bool` literal -> `attributes_map_bool`, JSON fallback uses `JSON_VALUE(...) = 'true'` because ClickHouse does not cast JSON bool strings through `toUInt8OrNull` correctly.
- `$in` uses the typed map when all list literals share one scalar type.
- `$convert(..., "exists")`, `null`, structured literals, mixed-type `$in` lists, and non-attribute fields keep the existing JSON_VALUE path.

## Test Plan

- [x] `uv run --group test python -m pytest tests/trace_server/query_builder/test_attributes_map_dispatch.py -v`
- [x] `uv run --group test python -m pytest tests/trace_server/query_builder/ -v`
- [x] `uv run --group test python -m pytest tests/trace_server/test_attributes_map_fast_filter.py -v --trace-server=clickhouse --clickhouse-process=true`
- [x] `uv run --group test python -m pytest tests/trace_server/test_clickhouse_trace_server_migrator.py -v`
- [x] `uvx ruff check weave/trace_server/calls_query_builder/calls_query_builder.py tests/trace_server/query_builder/test_attributes_map_dispatch.py tests/trace_server/query_builder/test_calls_query_builder.py tests/trace_server/test_attributes_map_fast_filter.py`

### DIFF
--- a/tests/trace_server/conftest.py
+++ b/tests/trace_server/conftest.py
@@ -20,6 +20,7 @@ from weave.trace_server import clickhouse_trace_server_batched
 from weave.trace_server import clickhouse_trace_server_migrator as wf_migrator
 from weave.trace_server.clickhouse_trace_server_batched import ClickHouseTraceServer
 from weave.trace_server.project_version import project_version
+from weave.trace_server.project_version.types import CallsStorageServerMode
 from weave.trace_server.secret_fetcher_context import secret_fetcher_context
 from weave.trace_server.sqlite_trace_server import SqliteTraceServer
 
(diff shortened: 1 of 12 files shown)


gtarpenning:
Closing this draft: it proved the wrong attrs-map-specific direction. Reworking from origin/master for general dynamic JSON filter cast inference.

PR #6729, opened 2026-04-28
feat(weave): infer dynamic json filter casts

gtarpenning:
https://coreweave.atlassian.net/browse/WB-33798

## Summary

- Infers ClickHouse casts for dynamic JSON call fields from typed comparison literals, so callers can omit `$convert` when the literal already communicates the expected type.
- Applies to raw dynamic fields under `inputs`, `output`, `attributes`, and `summary`, including dotted paths and scalar root fields.
- Covers `$eq`, `$gt`, `$lt`, `$gte`, `$lte`, and homogeneous literal `$in` lists; string, `null`, structured literals, and mixed `$in` lists keep the existing uncast path.
- Keeps this independent of the attrs-map branch. This PR is based directly on `origin/master`.

## Why

Users often build filters from typed UI state or typed Python values where the field path is dynamic JSON but the comparison literal is already the type they intend to compare against. Requiring `$convert` in every numeric or boolean raw-field comparison makes those query shapes more verbose and easy to get wrong.

If the user sends a well-typed literal, the query builder can infer the field-side conversion while preserving the existing query API.

## Compatibility note

This is a semantic change for existing queries that compared a dynamic JSON field against a numeric or boolean `$literal` *without* an explicit `$convert`. Previously those queries fell through to string comparison on the JSON_VALUE output (e.g. `'10' > '3'` lexicographic); they now emit a typed cast and compare numerically. Queries that used `$convert` explicitly, or compared against string literals, are unchanged.

## Query shape change

Before, typed raw JSON filtering required an explicit conversion:

```json
{
  "$expr": {
    "$eq": [
      {"$convert": {"input": {"$getField": "inputs.x"}, "to": "int"}},
      {"$literal": 1}
    ]
  }
}
```

Now the same comparison can omit `$convert`:

```json
{
  "$expr": {
    "$eq": [
      {"$getField": "inputs.x"},
      {"$literal": 1}
    ]
  }
}
```

For ClickHouse this emits a typed field-side extraction like:

```sql
toInt64OrNull(coalesce(nullIf(JSON_VALUE(any(calls_merged.inputs_dump), {path:String}), 'null'), '')) = {literal:Int64}
```

Inference mapping:

- `int` literals -> `toInt64OrNull(...)`
- `float` literals -> `toFloat64OrNull(...)`
- `bool` literals -> `multiIf(... = 'true', 1, ... = 'false', 0, toUInt8OrNull(...))`
- homogeneous typed `$in` literal lists -> one cast on the field side


bcsherma (approved):
Approving, but one interaction I’d keep an eye on: the inferred int/float casts now make direct JSON-field comparisons typed, while `_create_like_optimized_eq_condition` can still add a raw JSON-text `LIKE` prefilter for the same direct field/literal shape.

For example, a float literal like `1.0` could produce a final typed comparison that would treat JSON `1` and `1.0` equivalently, but the LIKE prefilter may only search for the text form of the literal. I think this is probably fine in practice because this optimization is already approximate and most callers are likely comparing values serialized in the same shape they queried with, but it may be worth a follow-up test or a small guard if we see false negatives around mixed int/float JSON encodings.

gtarpenning:
> Approving, but one interaction I’d keep an eye on: the inferred int/float casts now make direct JSON-field comparisons typed, while `_create_like_optimized_eq_condition` can still add a raw JSON-text `LIKE` prefilter for the same direct field/literal shape.
> 
> For example, a float literal like `1.0` could produce a final typed comparison that would treat JSON `1` and `1.0` equivalently, but the LIKE prefilter may only search for the text form of the literal. I think this is probably fine in practice because this optimization is already approximate and most callers are likely comparing values serialized in the same shape they queried with, but it may be worth a follow-up test or a small guard if we see false negatives around mixed int/float JSON encodings.

Great callout, adding an explicit test.

PR #6731, opened 2026-04-29
fix(weave): Sanitize invalid UTF8 surrogates

andrewtruong:
Fixes an issue where we could pass invalid characters to clickhouse that are not insertable.


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6908,12 +6909,24 @@ def _insert(
...
 
         start = time.monotonic()
+        sanitized_invalid_utf8 = False
         for attempt in range(ch_settings.INSERT_MAX_RETRIES):
             try:
                 result = self.ch_client.insert(
                     table, data=data, column_names=column_names, settings=settings
                 )
 
+            # Invalid client Unicode: sanitize the batch and retry once.
```
rgao-coreweave, line 6919: Should we apply the same fix to SQLite path?
`weave/trace_server/sqlite_trace_server.py`

SQLite's Python driver raises the same UnicodeEncodeError for the same input. Confirmed via repro:
```
import sqlite3
conn = sqlite3.connect(":memory:")
cur = conn.cursor()
cur.execute("CREATE TABLE t (x TEXT)")
cur.execute("INSERT INTO t VALUES (?)", ("hello \ud83d",))
# UnicodeEncodeError: 'utf-8' codec can't encode character '\ud83d' in position 6: surrogates not allowed
```

rgao-coreweave (approved):
Thx for the fix, learned about lone surrogate.

LGTM!

PR #6735, opened 2026-04-29
feat(weave): infer json filter casts in feedback path

gtarpenning:
[WB-33832](https://coreweave.atlassian.net/browse/WB-33832)

## Summary

- Original problem: `POST /feedback/query` with a Python `bool` literal against a JSON feedback payload field (`feedback.[...].payload.is_positive == False`) 502'd in prod with ClickHouse `Code: 386 (NO_COMMON_TYPE)`. `JSON_VALUE` returns `String`, `python_value_to_ch_type(False)` binds the literal as `Bool`, and with no cast on either side CH refused the compare. #6729 fixed the analogous shape on `CallsQuery` but gated inference to `CallsMergedDynamicField`, so the feedback / threads / objects paths kept hitting the same failure.
- Lifts the literal-cast inference into `orm.py`'s `_process_query_to_conditions` so any caller of `Select.where(...)` (feedback, threads, objects, etc.) gets typed JSON comparisons, not just the `CallsQuery` path.
- Adds `clickhouse_cast_json_value` for the bool quirk (`JSON_VALUE` on a JSON bool returns the strings `'true'` / `'false'`; `toUInt8OrNull` would NULL them out, so wrap in `multiIf('true' -> 1, 'false' -> 0, toUInt8OrNull(...))`).
- Collapses #6729's `interface/query.py` inference helpers and the inline JSON-bool `multiIf` in `calls_query_builder.json_dump_field_as_sql` so there's one source of truth across `CallsQuery`, `sqlite_trace_server`, and `orm.py`. Net: -50 lines after the dedup commit.

[Datadog trace](https://us5.datadoghq.com/apm/trace/69f2605b000000004396a487a93635d0) for the original failure:

```
Code: 386. DB::Exception: There is no supertype for types String, Bool ...
while executing function equals on arguments
  toString(JSON_VALUE(__table1.payload_dump, '$."is_positive"'_String)) String,
  _CAST(0_Bool, 'Bool'_String) Bool
```

## Compatibility note

Same as #6729: pre-existing queries that compared a JSON-extracted field against an `int` / `float` / `bool` `$literal` *without* `$convert` previously fell through to lexicographic string compare (`'10' > '3'`); they now emit a typed cast and compare numerically. Queries with explicit `$convert` or string literals are unchanged.

## Stacking

Currently stacked on #6729. When #6729 lands, rebase drops those commits and this PR is just the orm.py expansion + dedup (3 commits, ~ +90 / -10 net).

## Testing

`tests/trace_server/test_orm.py` covers bool / int / float / string / `$convert` / literal-on-LHS / homogeneous + mixed `$in` shapes against `Table.select().where(...).prepare("clickhouse")`. `tests/trace/test_feedback.py` exercises the feedback path against sqlite. Query-builder snapshot tests (multiIf SQL) keep the inline-removal honest.

[WB-33832]: https://coreweave.atlassian.net/browse/WB-33832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/test_feedback.py
+++ b/tests/trace/test_feedback.py
@@ -981,10 +981,10 @@ def test_feedback_query_contains_numeric_literal(client) -> None:
 def test_feedback_query_typed_payload_filters(client: WeaveClient) -> None:
     """Regression for WB-33832: /feedback/query 500s on typed payload literals.
 
-    Without inferred field-side casts, ClickHouse refuses ``JSON_VALUE(...)``
+    Without inferred field-side casts, ClickHouse refuses `JSON_VALUE(...)`
     (String) compared against a typed param (Bool / Int64 / Float64) with
-    ``Code: 386 NO_COMMON_TYPE``. JSON_VALUE on a JSON boolean emits the
-    literal string ``'true'`` / ``'false'``, so the bool path must coerce
+    `Code: 386 NO_COMMON_TYPE`. JSON_VALUE on a JSON boolean emits the
+    literal string `'true'` / `'false'`, so the bool path must coerce
     those before any numeric fallback.
 
     Asserts on both backends so the same query shape works through sqlite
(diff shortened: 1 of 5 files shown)


bcsherma (approved):
🔥

PR #6737, opened 2026-04-30
chore(weave): dual CJS/ESM emission via tsc-multi

chance-wnb:
# Weave Node SDK — dual-build PR 2: dual CJS/ESM emission via tsc-multi

Migrates the Weave Node SDK to a true file-per-module dual CJS/ESM build. ESM consumers loading `weave` (or `--import=weave/instrument`) now get real `.mjs` output that imports other `.mjs` siblings; CJS consumers get the same shape they shipped before. Background: [`Tech Report — Should the Weave Node SDK ship a dual CJS/ESM build`](https://www.notion.so/wandbai/Tech-Report-Should-the-Weave-Node-SDK-ship-a-dual-CJS-ESM-build-348e2f5c7ef380149c80fdeb5a546f3c?source=copy_link), particularly §5 ("Output shape") and §7 PR 2.

## TL;DR

- **Build pipeline:** [`scripts/build.mjs`](sdks/node/scripts/build.mjs) calls `tsc-multi`'s public `build()` function **twice** — once per target — with a different tsconfig per call.
- **Per-target file selection:** [`tsconfig.cjs.json`](sdks/node/tsconfig.cjs.json) excludes `src/esm/**`; [`tsconfig.esm.json`](sdks/node/tsconfig.esm.json) excludes `src/utils/commonJSLoader.ts`. Each format only sees the files it should.
- **One file rename:** `src/esm/instrument.ts` → `src/esm/instrument.mts` so TypeScript emits it as `.mjs` natively. Source uses `import.meta.url` directly — no eval workaround, no `@ts-ignore`.
- **One source-side change in** **`src/index.ts`:** runtime-conditional `require('./utils/commonJSLoader')` so the ESM build never tries to import the (now-not-emitted) CJS-only file.

## Why this is non-trivial

A naive dual-emit setup would be: "run tsc twice with different `module` settings, ship both outputs." That falls over for two reasons specific to this SDK.

### The two format-specific files

The SDK has exactly two source files that legitimately have no business in the _other_ format:

- [**`src/utils/commonJSLoader.ts`**](sdks/node/src/utils/commonJSLoader.ts) **(CJS-only).** Patches Node's `Module.prototype.require` so CJS hosts auto-instrument supported packages. There is no `require` chain to patch under ESM — the equivalent role is filled by an `import-in-the-middle` loader hook (registered from [`src/esm/instrument.mts`](sdks/node/src/esm/instrument.mts)). Compiling this file as ESM produces dead, broken code: a `.mjs` that calls `require()` (ReferenceError under ESM) and would crash if loaded.
- [**`src/esm/instrument.mts`**](sdks/node/src/esm/instrument.mts) **(ESM-only).** The `--import=weave/instrument` preload entry. Calls `register()` with `import.meta.url` to install the loader hook. `import.meta` is **a syntax error** under `module: commonjs`, so the file simply cannot be compiled as CJS without contortions (eval indirection, `@ts-ignore`, or a separate file split).

Each file is _intentionally_ format-specific and trying to compile each in the other format produces nonsense at best and broken output at worst.

### `tsc-multi` doesn't expose per-target file selection

`tsc-multi` is the right core: its AST transformer rewrites relative-import specifiers per target (so `.mjs` output imports `.mjs` siblings, `.js` output imports `.js` siblings — no dual-package hazard). But its config file (`tsc-multi.json`) only supports per-target **compiler-option overrides** (`module`, `moduleResolution`, `extname`, etc.). It does not support per-target `include` / `exclude`. From its [Target schema](https://github.com/tommy351/tsc-multi/blob/main/src/config.ts):

```ts
const targetSchema = type({
  extname: optional(string()),
  transpileOnly: optional(boolean()),
});
// plus arbitrary compiler options merged into compilerOptions
```

So out of the box, every project source file is compiled under every target. The two format-specific files above can't be excluded per target through `tsc-multi.json` alone.

## Approaches we explored, and why they were not ideal

The PR is the third design we tried. The earlier attempts taught us what doesn't work; capturing them here so future readers don't re-tread.

### 1\. `tsup` with `bundle: false`

The migration plan in §5 originally recommended `tsup`. Empirically, `tsup` with `bundle: false` does **not** rewrite import-specifier extensions — `.mjs` files end up importing `.js` siblings, which Node's ESM resolver loads as the **CJS** sibling next to it. The dual build degrades into "ESM-shaped CJS." The community workaround ([`esbuild-plugin-file-path-extensions`](https://github.com/favware/esbuild-plugin-file-path-extensions)) explicitly requires `bundle: true`, which loses the file-per-module shape we committed to in §5 ("Output shape"). Long-standing tsup issue: [#953](https://github.com/egoist/tsup/issues/953).

### 2\. `tsc-multi` + dedicated `tsc` invocation for `instrument`

Run `tsc-multi` against the bulk of the source, then run plain `tsc -p tsconfig.instrument.esm.json` for `instrument.mts`, then a `move-esm-instrument.mjs` script to copy just the wanted files into `dist/`.

This worked but had real drawbacks:

- The dedicated `tsc` invocation transitively compiles every imported source file (TypeScript's design — `include` doesn't gate emission), so we needed a sandbox `outDir` and a manual cherry-pick to avoid clobbering `tsc-multi`'s output.
- Asymmetric: `instrument.mts` got special-case build infrastructure; `commonJSLoader.ts` did not. The build pipeline had to grow more carve-outs as new format-specific files arrived.
- ~55 lines of build infrastructure (extra tsconfig + move script) for a one-file problem.

### 3\. `process.argv[1]` as `parentURL`, no `import.meta.url`

If `instrument`'s source could avoid `import.meta.url`, it would compile cleanly under both `module: commonjs` and `module: esnext`, and the whole dedicated-tsc workaround disappears. We tried passing `pathToFileURL(process.argv[1])` (the user's entry script) to `register()` as `parentURL`.

This worked for typical Node invocations but is fragile for global CLIs, npx, workspace-hoisted-`node_modules` setups, and any case where the entry script lives outside the project's `node_modules` ancestry. A tracing SDK can't tolerate "doesn't auto-instrument when launched in pattern X." Rejected.

### 4\. Rename source files to `.cts` / `.mts` to encode format at the filename level

Considered for both files; rejected for `commonJSLoader.cts` because `tsc-multi`'s extension-rewriting transformer is naive about `.cts`/`.mts` siblings. It rewrites `'./commonJSLoader'` → `'./commonJSLoader.js'` (CJS target), but the actual emitted file is `commonJSLoader.cjs` (TypeScript's per-extension rule). Mismatch → runtime resolution failure. We saw the same issue earlier in this stack with `.cts` source.

We **did** rename `instrument.ts` → `instrument.mts` because nothing imports that file from source (it's only invoked as a preload), so the transformer's mismatch behavior never gets a chance to fire. The rename makes the file's ESM-only nature self-documenting at the file-tree level.

### 5\. Roll our own dual-emit driver (vendor `tsc-multi`'s transformer)

Two tsconfigs (CJS, ESM) extending a shared base, plus our own AST transformer (lifted from `tsc-multi`'s public source) plus a custom build script that drives `ts.createProgram` + `program.emit` × 2.

Architecturally clean — full per-target file selection, no dependency on `tsc-multi` internals. ~240 lines of build infrastructure to own forever, including a TypeScript custom transformer to maintain across compiler-API drift. Not worth it for two format-specific files.

## What this PR actually does — and why it's the right shape

**Use** **`tsc-multi`'s public** **`build()`** **API directly, twice, with different tsconfigs per call.**

```js
// scripts/build.mjs
import {build} from 'tsc-multi';

await build({
  cwd,
  projects: ['tsconfig.cjs.json'],
  targets: [{extname: '.js', module: 'commonjs', moduleResolution: 'node'}],
});

await build({
  cwd,
  projects: ['tsconfig.esm.json'],
  targets: [{extname: '.mjs', module: 'esnext'}],
});
```

Each `build()` call runs `tsc-multi` for **one project against one target**. The per-target tsconfig declares its own `exclude` list (the file selection that `tsc-multi.json` cannot express), and the per-target settings live where they belong (the `module` field is in each per-format tsconfig, not in the root). The AST transformer runs as part of `tsc-multi`'s normal pipeline, so import-extension rewriting Just Works.

### Why this is a great leverage of `tsc-multi`'s design

The CLI surface (`tsc-multi` + `tsc-multi.json`) was designed for the typical "one source tree, two targets, no per-target file selection" case. We have a less typical case (per-target file selection), and the CLI doesn't cover it. But the **library API** — the exported `build()` function — accepts everything we need: the project paths, the target options, the cwd. It happily runs once per call with a one-target array. Nothing about its design objects to being called multiple times.

By calling `build()` ourselves we:

1. Get full control over which tsconfig pairs with which target.
2. Inherit `tsc-multi`'s well-tested AST transformer for free, with no internal-API coupling.
3. Avoid maintaining a `tsc-multi` clone (~240 lines of compiler-API code).
4. Stay on the upgrade path: a future `tsc-multi` version that fixes bugs or adds features ships to us via a normal `pnpm update`.

This is a relatively common pattern for build tools that ship a CLI thin wrapper around a programmatic API: when the CLI's config doesn't cover your case, drop down one layer and drive the API directly.

### [`src/index.ts`](sdks/node/src/index.ts) — runtime-conditional `commonJSLoader` import

```ts
// CJS-only side-effect: install the `require()` patcher so CJS hosts
// auto-instrument supported modules. ESM hosts use the loader hook in
// `./esm/instrument.mjs` instead, registered via `--import=weave/instrument`.
if (typeof require === 'function' && typeof module === 'object') {
  require('./utils/commonJSLoader');
}
```

The `commonJSLoader.ts` source file is **excluded from the ESM tsconfig**, so no `commonJSLoader.mjs` ships in `dist/`. The runtime guard prevents the `require()` from being called under ESM (where the file wouldn't exist anyway). CJS consumers are unchanged: the guard passes, and `commonJSLoader.js` patches `Module.prototype.require` like before.

### [`src/esm/instrument.mts`](sdks/node/src/esm/instrument.mts) — `.mts` rename

Renamed from `.ts` so TypeScript emits `.mjs` natively (and `.d.mts` for the declaration). Source uses `import.meta.url` directly with no contortions — works because the file is excluded from the CJS tsconfig and only ever compiled under `module: esnext`.

### [`src/__tests__/integrations/openaiIITM.test.ts`](sdks/node/src/__tests__/integrations/openaiIITM.test.ts)

Two occurrences of `dist/esm/instrument.js` updated to `dist/esm/instrument.mjs` since the runtime artifact is now `.mjs`.

## Tsconfig structure

- [`tsconfig.json`](sdks/node/tsconfig.json) — shared base. `target: es2022` (matches our Node-18-minimum commitment, drops unnecessary downleveling), `moduleResolution: node`, `strict`, `paths`, etc. **No** **`module`** **setting** — that's a per-target concern.
- [`tsconfig.cjs.json`](sdks/node/tsconfig.cjs.json) — extends base, pins `module: commonjs` explicitly, excludes `src/esm/**`. Passed to the first `build()` call with `extname: '.js'`.
- [`tsconfig.esm.json`](sdks/node/tsconfig.esm.json) — extends base, pins `module: esnext` explicitly, excludes `src/utils/commonJSLoader.ts`. Passed to the second `build()` call with `extname: '.mjs'`.

Each per-format tsconfig is self-explanatory at a glance: you don't have to chase the `extends` chain to know what format it produces.

## What the published package looks like now

```
dist/
├── index.js / index.mjs / index.d.ts          # main entry, both formats
├── clientApi.js / clientApi.mjs / clientApi.d.ts
├── …                                           # every src/ file mirrored
├── integrations/
│   ├── openai.js / openai.mjs / openai.d.ts
│   └── …
├── utils/
│   ├── commonJSLoader.js / commonJSLoader.d.ts # CJS-only — no .mjs sibling
│   └── …
└── esm/
    └── instrument.mjs / instrument.d.mts       # ESM-only — no .js sibling
```

`exports` map:

```jsonc
{
  ".": {
    "types": "./dist/index.d.ts",
    "import": "./dist/index.mjs",
    "require": "./dist/index.js"
  },
  "./instrument": {
    "types": "./dist/esm/instrument.d.mts",
    "import": "./dist/esm/instrument.mjs"
  }
}
```

`./instrument` only routes the `import` condition — CJS consumers don't use this entry; they use the `commonJSLoader` patch baked into `require('weave')`.

## Tested scenarios

- [x] `pnpm run build` — clean (`[js]: Found 0 errors. [mjs]: Found 0 errors.`).
- [x] Smoke test 1: `node -e "require('weave')"` — succeeds, exposes expected keys.
- [x] Smoke test 2: `node --input-type=module -e "import * as w from 'weave'"` — succeeds, exposes expected keys.
- [x] Smoke test 3: `node --input-type=module --import='weave/instrument' -e "console.log('OK')"` — succeeds (preload completes without errors).
- [x] `WANDB_API_KEY=dummy pnpm run test` — **185 passed, 7 skipped** (live-network tests, pre-existing skip), 0 failed. Same counts as `master` plus the 2 retest fixes.
- [x] Manual validation against several host applications (compiled with both CJS and ESM).
- [x] `dist/esm/instrument.mjs` import audit: imports from `.mjs` siblings (`./integrations/instrumentations.mjs`, etc.), not via ESM→CJS interop. True file-per-module dual build.

---

<a href="https://app.devin.ai/review/wandb/weave/pull/6737" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/sdks/node/package.json
+++ b/sdks/node/package.json
@@ -8,16 +8,16 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "require": "./dist/index.js",
-      "import": "./dist/index.js"
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js"
     },
     "./instrument": {
-      "types": "./dist/esm/instrument.d.ts",
-      "import": "./dist/esm/instrument.js"
+      "types": "./dist/esm/instrument.d.mts",
+      "import": "./dist/esm/instrument.mjs"
     }
   },
   "scripts": {
-    "build": "tsc --outDir dist",
+    "build": "node scripts/build.mjs",
     "prepare": "npm run build",
     "test": "jest --silent",
     "test:coverage": "jest --coverage",
@@ -167,6 +167,7 @@
     "swagger-typescript-api": "^13.0.22",
     "ts-jest": "^29.2.5",
     "ts-node": "^10.9.2",
+    "tsc-multi": "^1.1.0",
     "tsconfig-paths": "^4.2.0",
     "tsx": "^4.19.1",
     "typedoc": "^0.28",
--- a/sdks/node/pnpm-lock.yaml
+++ b/sdks/node/pnpm-lock.yaml
@@ -81,6 +81,9 @@ importers:
       ts-node:
         specifier: ^10.9.2
         version: 10.9.2(@types/node@22.14.1)(typescript@5.5.4)
+      tsc-multi:
+        specifier: ^1.1.0
+        version: 1.1.0(typescript@5.5.4)
       tsconfig-paths:
         specifier: ^4.2.0
         version: 4.2.0
@@ -518,6 +521,18 @@ packages:
 
   '@jridgewell/trace-mapping@0.3.9':
     resolution: {integrity: sha512-3Belt6tdc8bPgAtbcmdtNJlirVoTmEb5e2gC94PnkwEW9jI6CAHUeoG85tjWP5WquqfavoMtMwiG4P926ZKKuQ==}
+
+  '@nodelib/fs.scandir@2.1.5':
+    resolution: {integrity: sha512-vq24Bq3ym5HEQm2NKCr3yXDwjc7vTsEThRDnkp2DK9p1uqLR+DHurm/NOTo0KG7HYHU7eppKZj3MyqYuMBf62g==}
+    engines: {node: '>= 8'}
+
+  '@nodelib/fs.stat@2.0.5':
+    resolution: {integrity: sha512-RkhPPp2zrqDAQA/2jNhnztcPAlv64XdhIp7a7454A5ovI7Bukxgt7MX7udwAu3zg1DcpPU0rz3VV1SeaqvY4+A==}
+    engines: {node: '>= 8'}
+
+  '@nodelib/fs.walk@1.2.8':
+    resolution: {integrity: sha512-oGB+UxlgWcgQkgwo8GcEGwemoTFt3FIO9ababBmaGwXIoBKZ+GTy0pP185beGg7Llih/NSHSV2XAs1lnznocSg==}
+    engines: {node: '>= 8'}
 
   '@opentelemetry/api-logs@0.53.0':
     resolution: {integrity: sha512-8HArjKx+RaAI8uEIgcORbZIPklyh1YLjPSBus8hjRmvLi6DeFzgOcdZ7KwPabKj8mXF8dX0hyfAyGfycz0DbFw==}
@@ -1097,11 +1112,18 @@ packages:
     resolution: {integrity: sha512-2Zks0hf1VLFYI1kbh0I5jP3KHHyCHpkfyHBzsSXRFgl/Bg9mWYfMW8oD+PdMPlEwy5HNsR9JutYy6pMeOh61nw==}
     engines: {node: ^14.15.0 || ^16.10.0 || >=18.0.0}
 
+  fast-glob@3.3.3:
+    resolution: {integrity: sha512-7MptL8U0cqcFdzIzwOTHoilX9x5BrNqye7Z/LuC7kCMRio1EMSyqRK3BEAUD7sXRq4iT4AzTVuZdhgQ2TCvYLg==}
+    engines: {node: '>=8.6.0'}
+
   fast-json-stable-stringify@2.1.0:
     resolution: {integrity: sha512-lhd/wF+Lk98HZoTCtlVraHtfh5XYijIjalXck7saUtuanSDyLMxnHhSXEDJqHxD7msR8D0uCmqlkwjCV8xvwHw==}
 
   fast-safe-stringify@2.1.1:
     resolution: {integrity: sha512-W+KJc2dmILlPplD/H4K9l9LcAHAfPtP6BY84uVLXQ6Evcz9Lcg33Y2z1IVblT6xdY54PXYVHEv+0Wpq8Io6zkA==}
+
+  fastq@1.20.1:
+    resolution: {integrity: sha512-GGToxJ/w1x32s/D2EKND7kTil4n8OVk/9mycTc4VDza13lOvpUZTGX3mFSCtV9ksdGBVzvsyAVLM6mHFThxXxw==}
 
   fb-watchman@2.0.2:
     resolution: {integrity: sha512-p5161BqbuCaSnB8jIbzQHOlpgsPmK5rJVDfDKO91Axs5NC1uu3HRQm6wt9cd9/+GtQQIO53JdGXXoyDpTAsgYA==}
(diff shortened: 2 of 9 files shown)


chance-wnb:
* **#6737** <a href="https://app.graphite.com/github/pr/wandb/weave/6737?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6737?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6682** <a href="https://app.graphite.com/github/pr/wandb/weave/6682?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6697](https://github.com/wandb/weave/pull/6697) <a href="https://app.graphite.com/github/pr/wandb/weave/6697?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

andrewtruong (approved):

PR #6738, opened 2026-04-30
perf(weave): reduce per-op overhead in async tracing

gtarpenning:
## Summary

Reduces per-op overhead in the weave Python client. **Strictly internal/non-user-visible changes.** The user contract is identical to master.

Linked: [WB-33844](https://coreweave.atlassian.net/browse/WB-33844)

## Scope

Three fixes only. The deferred output-side `_save_nested_objects` walk that an earlier revision included has been moved to a follow-up PR (TBD) because it exposed a latent worker-pool starvation hazard. See "Why the deferral isn't here" below.

1. **Drop the redundant per-op `model_dump_json` size check** in both `create_call` and `finish_call`. Each was running a full Pydantic encode just to compute `len()` for a warning log.
2. **Cache the op URI per Op**. After first save the digest is a resolved string, so subsequent calls read `.uri` directly without wrapping in a future.
3. **Cache `inspect.signature` + `parse_from_signature`** at op decoration time. Both were re-running per-call.

## Bench (paired, simulated H200 GIL pressure ~33% contention, 1460 ops/step)

| Branch                                | Step avg | Slowdown | p50 stall | Max stall |
|---------------------------------------|---------:|---------:|----------:|----------:|
| Master                                | 1.15s    | 9.1x     | 176ms     | 1319ms    |
| **This PR (internal-only, 3 fixes)**  | 1.02s    | 8.0x     | 170ms     | **649ms** |
| Delta vs master                       | -11%     | -12%     | -3%       | **-51%**  |

The headline win is in **tail latency** (max stall 1319ms -> 649ms). p50 step time barely moved because the dominant per-step cost is the synchronous `_save_nested_objects` + `to_json` walk on the calling thread, which is unchanged here. That's what the follow-up PR addresses.

## Why the deferral isn't here

An earlier draft also deferred `_save_nested_objects` + `map_to_refs` from the calling thread into the `send_end_call` worker. That cut step time to 0.57s (~50% improvement) but exposed a latent starvation in `_save_object_basic`:

- Each `send_obj_create` calls `to_json(parent_with_refs)` from a worker, which synchronously waits on child `digest_future.result()`.
- Under highly concurrent flushes, all main-pool workers can land in `to_json` simultaneously, each blocked on a child whose `obj_create` is queued *behind* them in the same executor.
- Master happens to escape this hazard by ~1 worker of timing headroom; the deferral consumed it.

The fix is structural (pre-walk to collect child digest futures, chain `to_json + send` via `then(...)` so workers don't block on `result()`), and it's the right home for the deferral. Splitting it lets this PR ship safely now and the structural fix land independently.

## What is **not** changed

User contract identical to master:
- Errors in user-supplied `postprocess_output` and `_on_finish_handler` still propagate on the calling thread.
- `call.output` is set to the postprocessed value when `finish_call` returns.
- `call.summary` is computed and set when `finish_call` returns.
- Existing tests pass without modification.
- Production worker-pool defaults unchanged.

## Risks

- **Op decoration is slightly heavier.** `inspect.signature` and `parse_from_signature` now run once per `@weave.op()` decoration. For modules with many op definitions, import time goes up by a tiny fraction. (`parse_from_signature` did the same regex parsing per call before.)
- **Size-warning telemetry lost.** A user with a single-call payload exceeding `MAX_TRACE_PAYLOAD_SIZE` (3.5 MiB) will no longer see a client-side warning. They'll learn from a server rejection instead. Removing the check was the largest single win in this PR.

## Test plan

- [x] Local: full `trace` shard with `--trace-server=sqlite -n4`: passes.
- [x] Local: full `trace_calls_complete_only` shard (`WEAVE_USE_CALLS_COMPLETE=true`): passes.
- [x] Local: targeted resilience/postprocessing/op-decorator suites: pass.
- [x] CI: previously-hung trace shards (Linux 3.10/3.11/3.13, Windows 3.10/3.11/3.13) all green.
- [x] Local bench across no-GIL, light-GIL, heavy-GIL pressure (recorded in `bench_history.jsonl`).
- [ ] Hand to MRE author David Corbitt for prod repro.

## Related

- Follow-up structural fix (defer + pre-walk digest collection): TBD draft PR off master.
- Bench tool and history: `~/Desktop/local_testing/weavetest/qa-scripts/bench_async_overhead.py` + `bench_history.jsonl`.

[WB-33844]: https://coreweave.atlassian.net/browse/WB-33844


weave/trace/op.py
```diff
@@ -1355,6 +1367,19 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
             wrapper.kind = kind  # type: ignore
             wrapper.color = color  # type: ignore
 
+            # Cache signature + parsed annotations at decoration time. These
+            # are otherwise recomputed in `_default_on_input_handler` on every
+            # call, which is the hot path for high-throughput async ops.
+            from weave.trace.annotation_parser import parse_from_signature  # circular
```
chance-wnb: Do we really have a circular dependency by importing `parse_from_signature` ? how so?

weave/trace/op.py
```diff
@@ -1355,6 +1367,19 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+
+            try:
+                cached_sig = inspect.signature(func)
+                wrapper._cached_signature = cached_sig  # type: ignore
```
chance-wnb: I think you can probably make `_cached_signature` an implementation detail of inspect.signature itself. So all callers of inspect.signature() transparently get the cached result. `inspect.signature()` just returns func._cached_signature if it detects it. As I see, there are a few other callers of inspect.signature() than the spots you changed.

chance-wnb (commented):
Thanks for the fix!

I left a couple suggestions.

weave/trace/op.py
gtarpenning: yeah true, lets think about this...

chance-wnb: Ok, I just realized that `inspect.signature` is a system library function, which is out of our control. So my comment above doesn't quite make sense. But I now see that you found a workaround. Great job and thanks for addressing my comment.

chance-wnb (approved):
LGTM. Thanks!

weave/trace/op.py
```diff
@@ -1355,6 +1369,31 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
             wrapper.kind = kind  # type: ignore
             wrapper.color = color  # type: ignore
 
+            # `inspect.signature` returns `__signature__` directly, so this
+            # makes all callers hit the cached signature for the wrapper.
```
andrewtruong: I dont understand this comment

weave/trace/op.py
```diff
@@ -313,8 +321,14 @@ def _default_on_input_handler(func: Op, args: tuple, kwargs: dict) -> ProcessedI
 
     # Annotated return type flow
     # If user defines postprocess_output manually, trust it instead of running this
-    if not func.postprocess_output and sig.return_annotation:
-        parsed = parse_content_annotation(str(sig.return_annotation))
+    if not func.postprocess_output:
+        parsed = getattr(func, "_weave_cached_parsed_return_annotation", None)
```
andrewtruong: why are we doing getattr/hasattr checks?

weave/trace/op.py
```diff
@@ -276,16 +281,15 @@ def _is_unbound_method(func: Callable) -> bool:
...
-        ContentAnnotation,
-        parse_content_annotation,
-        parse_from_signature,
-    )
+    # Avoid circular import: Content -> serialization.op_type -> op.
     from weave.type_wrappers import Content
 
     try:
         sig = inspect.signature(func)
+    except (TypeError, ValueError) as e:
```
andrewtruong: When would you hit this?

weave/trace/weave_client.py
```diff
@@ -812,7 +812,14 @@ def create_call(
...
 
-        op_name_future = self.future_executor.defer(lambda: op_def_ref.uri)
+        # Skip the future allocation once the op's digest is resolved (any
+        # call after the first). Reading `.uri` directly is a no-op string
+        # build for an already-resolved ref.
+        op_name_future: str | Future[str]
+        if isinstance(op_def_ref._digest, Future):
+            op_name_future = self.future_executor.defer(lambda: op_def_ref.uri)
+        else:
+            op_name_future = op_def_ref.uri
```
andrewtruong, line 819: is this the main perf improvement?

weave/trace/op.py
gtarpenning: :robot: yeah it was vague. rewrote it to say what it actually does: setting `wrapper.__signature__` makes later `inspect.signature(wrapper)` calls short-circuit to the cached object instead of re-walking `func`.

weave/trace/op.py
gtarpenning: :robot: good call, dead defense. the decorator sets `_weave_cached_parsed_input_annotations` and `_weave_cached_parsed_return_annotation` on the wrapper in both the success and except branches, so they always exist on any `Op`. dropped the getattr/hasattr ladder in favor of direct attr access. the `if parsed is None and not hasattr(...)` arm was unreachable too, deleted.

weave/trace/op.py
gtarpenning: :robot: practically never. `inspect.signature(wrapper)` is just an attr fetch once we set `__signature__`, so this only fires for unintrospectable callables (built-ins / C extensions). the split was pointless, collapsed back to the single combined try/except w/ TypeError that master had.

weave/trace/weave_client.py
gtarpenning: :robot: nope, this is one of three. the size-check removal in `create_call`/`finish_call` (skipping the per-call `model_dump_json` just to `len()` for a warning) was the biggest single win. signature/annotation caching + this URI defer round it out. together they cut max stall ~51% (1319ms -> 649ms), p50 step time barely moves b/c the dominant cost is the synchronous `_save_nested_objects` walk that lives in the follow-up PR.

weave/trace/op.py
```diff
@@ -296,7 +302,9 @@ def _default_on_input_handler(func: Op, args: tuple, kwargs: dict) -> ProcessedI
     # If user defines postprocess_inputs manually, trust it instead of running this
     to_weave_inputs = {}
     if not func.postprocess_inputs:
-        parsed_annotations = parse_from_signature(sig)
+        parsed_annotations = func._weave_cached_parsed_input_annotations  # type: ignore[attr-defined]
```
andrewtruong, line 305: why are you doing a type ignore here?

weave/trace/op.py
```diff
@@ -313,8 +321,14 @@ def _default_on_input_handler(func: Op, args: tuple, kwargs: dict) -> ProcessedI
 
     # Annotated return type flow
     # If user defines postprocess_output manually, trust it instead of running this
-    if not func.postprocess_output and sig.return_annotation:
-        parsed = parse_content_annotation(str(sig.return_annotation))
+    if not func.postprocess_output:
+        parsed = func._weave_cached_parsed_return_annotation  # type: ignore[attr-defined]
```
andrewtruong, line 325: and here

andrewtruong (approved):
please fix the type ignores

weave/trace/op.py
gtarpenning: i think this is the best we can do. the type here is the public Op signature, to properly type it would be to add `_weave_cached_parsed_input_annotations` to the actual type, users would see it etc. we could do that if you  think its okay.

andrewtruong (approved):

PR #6739, opened 2026-04-30
chore(weave): EvaluationLogger: add internal support for applying eval metadata to all calls

kelu-wandb:
## Description

- Step of https://coreweave.atlassian.net/browse/WB-32723

Internal-only: Add `_create_with_meta` factory to `EvaluationLogger` that allows passing attributes that will then get propagated to all child calls.  (This is unlike `eval_attributes`, which only applies to the `Evaluation.evaluate` call.)

See followup prototype `wandb/wandb` usage: https://github.com/wandb/wandb/pull/11809

## Testing

- [x] Added unit tests
- [x] End-to-end tested with corresponding `wandb/wandb` change.


weave/evaluation/eval_imperative.py
```diff
@@ -665,10 +673,13 @@ def __init__(
         dataset: Dataset | list[dict] | str | None = None,
         eval_attributes: dict[str, Any] | None = None,
         scorers: list[str] | None = None,
+        _meta_attributes: dict[str, Any] | None = None,
```
andrewtruong: is there a nicer way to expose this?  This is not intended to be used by normal `EvaluationLogger` users.  Maybe you can expose a classmethod just for you instead?

andrewtruong: I can see `EvaluationLogger._with_meta`; or better a private function that instantiates and modifies `EvaluationLogger` but doesn't otherwise change the `EvaluationLogger` classdef

kelu-wandb: Replaced with `_with_meta`. I can't just construct first because there's stuff that happens in `__init__` that needs those attributes already, so I have to get in there by calling `__new__` and setting them before calling `__init__`.

weave/evaluation/eval_imperative.py
```diff
@@ -767,13 +780,31 @@ def summarize(self: Evaluation) -> dict:
...
+        eval_meta: dict[str, Any],
+        **init_kwargs: Any,
+    ) -> EvaluationLogger:
+        """Internal factory: build an EvaluationLogger with extra
+        `_weave_eval_meta` keys propagated to every emitted call.
+
+        Used by wandb.EvalTable to tag calls with `wandb_eval_table: True`
+        without exposing the side channel on the public constructor.
+        """
+        instance = cls.__new__(cls)
```
andrewtruong, line 806: I think this is more simply written as:

```python
instance = cls(**init_kwargs)
instance._eval_meta = ...
```

weave/evaluation/eval_imperative.py
```diff
@@ -354,11 +359,14 @@ def __init__(
         evaluate_call: Call,
         predict_call: Call,
         predefined_scorers: list[str] | None = None,
+        _eval_meta: dict[str, Any] | None = None,
```
andrewtruong: this should be removed

weave/evaluation/eval_imperative.py
```diff
@@ -354,11 +359,14 @@ def __init__(
...
     ) -> None:
         self.predict_and_score_call = predict_and_score_call
         self.evaluate_call = evaluate_call
         self.predict_call = predict_call
         self.predefined_scorers = predefined_scorers
+        _eval_meta = _eval_meta if _eval_meta is not None else {}
+        self._score_call_meta = {**_eval_meta, **SCORE_META_MARKER}
```
andrewtruong: Does this need to be here?  I'm guessing probably not?

weave/evaluation/eval_imperative.py
```diff
@@ -669,6 +677,9 @@ def __init__(
         self.name = name
         self.scorers = scorers
         self.eval_attributes = eval_attributes if eval_attributes is not None else {}
+        # Init with just the marker if _eval_meta hasn't been pre-seeded by _create_with_meta.
+        if not hasattr(self, "_eval_meta"):
```
andrewtruong: Please dont use hasattr checks

weave/evaluation/eval_imperative.py
andrewtruong: You don't need to use our init.  You can just write your own custom alternate constructor!

weave/evaluation/eval_imperative.py
kelu-wandb: Obsolete.

weave/evaluation/eval_imperative.py
kelu-wandb: So that doesn't work because I need to set the eval meta BEFORE the body of the constructor runs.

weave/evaluation/eval_imperative.py
kelu-wandb: Changed this to a setter.

weave/evaluation/eval_imperative.py
kelu-wandb: See above

weave/evaluation/eval_imperative.py
kelu-wandb: So the key factors are:
- I want to run everything in the main constructor.
- The main constructor uses the eval meta both when declaring the predict_and_score inner function and when creating the initial evaluate call, so I have to get in there and set it before that happens, i.e. I can't just run the constructor first and then set the eval meta.

My new approach is to factor out the body of init into an `_init_common` helper function, and the default constructor and my alternate factory each set eval meta as they want and then run that helper function.

For ScoreLogger, the constructor doesn't do any reads, so I can set the eval meta via a setter (though honestly the constructor is only used internally right? The class can be used externally but we don't really expect anyone to construct them?)

weave/evaluation/eval_imperative.py
```diff
@@ -771,13 +801,42 @@ def summarize(self: Evaluation) -> dict:
...
+        """Internal factory: build an EvaluationLogger with extra
+        `_weave_eval_meta` keys propagated to every emitted call.
+        """
+        # Use __new__ instead of cls(**init_kwargs) because we need to set the
+        # _eval_meta before running the body of the constructor, which uses it when
+        # declaring the predict_and_score inner class and creating the actual
+        # evaluate call.
+        instance = cls.__new__(cls)
+        instance._eval_meta = {**eval_meta, **IMPERATIVE_EVAL_META_MARKER}
+        instance._init_common(
```
andrewtruong: You can use `__new__` but then just call the regular init instead of making a helper func for init

andrewtruong (approved):

weave/evaluation/eval_imperative.py
kelu-wandb: Done.

PR #6740, opened 2026-04-30
perf(weave): defer output _save_nested_objects in client

gtarpenning:
## TL;DR

Cuts per-op calling-thread CPU ~5× in async tracing (p50 stall 176ms → 38ms in heavy-GIL bench) by deferring output-side `_save_nested_objects` + `map_to_refs` + `to_json` from `finish_call` into the worker pool. A structural fix to `_save_object_basic` / `_save_table` keeps a naive defer from deadlocking on thread-pool starvation.

Builds on #6738 (the conservative subset). Linked: [WB-33844].

## How the work moves off the calling thread

```
master:
   ┌─ calling thread ──────────────────────────┐    ┌─ worker ──────────┐
   │ finish_call:                              │    │ send_end_call:    │
   │   _save_nested_objects(output)   <-- walk │    │   redact          │
   │   map_to_refs(output)            <-- walk │    │   to_json         │
   │   defer(send_end_call)  ──────────────────┼──► │   server.call_end │
   │   return                                  │    └───────────────────┘
   └───────────────────────────────────────────┘

this PR:
   ┌─ calling thread ──────────┐    ┌─ worker: schedule_send ─────────────┐
   │ finish_call:              │    │   _save_nested_objects(output)      │
   │   defer(schedule_send) ───┼──► │   map_to_refs(output)               │
   │   return                  │    │   pending = collect_pending_digests │
   └───────────────────────────┘    │   if pending:                       │
                                    │     then(pending, finalize_send) ───┼─► continuation
                                    │     # worker returns to pool        │   runs after
                                    │   else:                             │   digests resolve
                                    │     finalize_send(...)              │
                                    └─────────────────────────────────────┘
                                          │
                                          ▼ finalize_send: redact + to_json + server.call_end
```

## What changes

1. **Output walks deferred** in `finish_call`. `_save_nested_objects`, `map_to_refs`, redact, and `to_json` move into the worker pool. `call.output` is still set sync, so user code reading it after `finish_call` returns is unchanged.
2. **Save sites gate on pending digests.** `_save_object_basic`, `_save_table`, and the chunked-table paths now go through one `_defer_after_pending_digests` helper that pre-walks the value and uses `then(pending, send)` instead of `defer(send)`. The worker registering the callback returns to the pool instead of blocking inside `to_json -> digest.result()`.
3. **`send_end_call` split** into Phase 1 (`schedule_send` — walks + collect_pending) and Phase 2 (`finalize_send` — redact + `to_json` + `call_end`). A barrier `end_complete_future` is tracked in the executor so `_flush()` waits for the whole pipeline; `log_exception=True` keeps the same "Task failed" log line existing resilience tests assert on.
4. **Output snapshot before defer** ([review fix](#output-mutation-race-review-fix-from-andrewtruong)). `_snapshot_mutable_containers` rebuilds `dict`/`list`/`set`/`tuple` on the calling thread so a caller mutating the returned value can't reach into the deferred walk.

## Calling-thread contract (unchanged)

- User `postprocess_output` and `_on_finish_handler` still raise on the calling thread.
- `call.output` is set to the postprocessed value before `finish_call` returns.
- `call.summary` is computed and set before `finish_call` returns.
- Production worker-pool defaults unchanged.

## Risks

- **Refs attach asynchronously after `finish_call` returns.** Today (master) refs attach before `finish_call` returns. User code that synchronously inspects `output.ref` immediately after `finish_call` may briefly see no ref attached. Behavior change; not expected to affect any current pattern but called out.
- **Larger surface than #6738.** Touches `_save_object_basic`, `_save_table`, both chunked-table paths, and `finish_call`.
- **`_track_future` is private.** Used deliberately to register the barrier future; will need an update if the futures module evolves.

## Output mutation race (review fix from @andrewtruong)

Andrew flagged that with the output walks deferred, a caller mutating the returned value after `finish_call` returns has the mutation recorded:

```py
@weave.op
def func(): return {"result": [1]}

res = func()
res["result"].append(2)   # without the fix, this leaks into the recorded output
```

Fix: `_snapshot_mutable_containers` rebuilds `dict`/`list`/`set`/`tuple` at every level on the calling thread before queuing `schedule_send`. Leaf types (Refs, weave Objects, pydantic models) are returned by identity so async ref attachment via `_save_nested_objects` still flows through to the caller's alias.

Regression test: `test_output_mutation_after_finish_is_not_recorded` gates the worker at the start of `schedule_send` so the mutation deterministically lands first, then asserts the recorded `call_end` request carries the pre-mutation value. Fails on the unfixed branch with `{"result": [1, 2]}`, passes with the snapshot.

### Cost (sqlite micro-bench, calling-thread p50, n=200, no GIL pressure)

| Payload                       | Without fix | With snapshot | Delta    |
|-------------------------------|------------:|--------------:|---------:|
| completions=200               |     0.20ms  |       0.56ms  |  +0.36ms |
| completions=1000              |     0.35ms  |       2.18ms  |  +1.83ms |
| flat dict width=10000         |     1.68ms  |       4.06ms  |  +2.38ms |
| huge 1MB string leaf          |     0.16ms  |       0.17ms  |     ~0   |

Linear in container element count. Leaf strings/bytes/numerics are O(1) because they are returned by identity. Bench script: `~/Desktop/local_testing/weavetest/qa-scripts/local_bench_payloads.py`.

Compared the `copy.deepcopy` alternative on the same payloads: ~15-30% slower at every size AND breaks weave Object ref attachment for the caller's original alias (deepcopy clones the Object, ref attaches to the clone, original never gets one). Snapshot wins on both axes.

### Residual scope

Snapshot catches container mutation (Andrew's exact example). Does NOT catch:
- pydantic / dataclass field mutation: `res.x = 5`
- in-place leaf mutation: `numpy_arr += 1`
- weave Object field mutation between `finish_call` and the deferred save

The deferred `_save_nested_objects` still sees post-mutation state for those. Full master-style correctness would require moving `_save_nested_objects + map_to_refs` back to sync.

## Bench

Heavy GIL pressure (1460 ops/step, 30 steps). Median per-step is the stable comparison; avg is dominated by occasional GC stalls under heavy GIL.

| Branch                            | Median step | p50 stall | Max stall |
|-----------------------------------|------------:|----------:|----------:|
| Master                            |       ~1.0s |     176ms |    1319ms |
| Internal-only (#6738)             |       ~1.0s |     170ms |     649ms |
| **This PR**                       |   **~0.4s** |  **38ms** |     ~3s\* |

\* Max under heavy GIL is dominated by occasional consecutive bad steps (a 9.54s outlier appears once in the 30-step run, then the system recovers). Median/p50 are the meaningful numbers — calling-thread time per op is cut roughly 5×.

The 38ms p50 above is from before the snapshot fix. Need to re-run the heavy-GIL bench with the snapshot on top; the local micro-bench above suggests +0.5-2ms per op on typical payloads, so the regression should be small but should be confirmed before merge.

## Tests

`tests/trace/test_save_nested_defer.py` (17 tests):
- 12 unit tests for `_collect_pending_digests` covering primitives, nested containers, ObjectRef/OpRef/TableRef (resolved + pending), TableRef with both digest *and* row_digests pending, ObjectRecord, cycles, and resolved-future short-circuit.
- 4 integration tests against a paused-server fixture: single-op finish doesn't deadlock; 20 concurrent async finishes drain via the then-chain; a real `call_end` failure surfaces as `"Task failed"` via `_track_future` (regression guard for `log_exception=True`); `finish_call` returns before the deferred work runs.
- 1 regression test for the output-mutation race (`test_output_mutation_after_finish_is_not_recorded`): captures `call_end` via a server proxy, gates the worker at the start of `schedule_send`, mutates the returned dict between `finish_call` returning and the gate releasing, asserts the recorded output is the pre-mutation value.

Existing suites green:
- Local: `trace` shard (839 passed), `trace_calls_complete_only` shard (839 passed).
- Local: `test_evaluation_performance` 5× consecutive at default parallelism (the deadlock fixture; would hang without the structural gate).
- Local: `test_tracing_resilience` (29/29).

## Test plan

- [x] `trace` and `trace_calls_complete_only` shards green
- [x] Deadlock fixture (`test_evaluation_performance`) green at default parallelism, 5× consecutive
- [x] Bench under heavy GIL, 30 steps. p50 stall 38ms vs master 176ms.
- [x] Regression test for output mutation race (Andrew's review)
- [ ] Re-run heavy-GIL bench with the snapshot fix to confirm overhead is small
- [ ] CI sweep across trace, trace_calls_complete_only, integrations, Windows
- [ ] Hand to MRE author David Corbitt for prod repro

## Related

- #6738 — internal-only fixes. Safe to ship without this. Modest p50 win (~3%) but halves max stall (1319ms → 649ms).
- Bench tool and history: `~/Desktop/local_testing/weavetest/qa-scripts/bench_async_overhead.py` + `bench_history.jsonl`.

<details>
<summary><b>Deep dive: why naive defer deadlocks (and how the gate fixes it)</b></summary>

**Part 1 alone** — just deferring the output walks, no gate — is a one-line change but reproduces a real deadlock against `test_evaluation_performance::paused_client`:

`_save_object_basic`'s `send_obj_create` calls `to_json(parent_with_refs)`, which synchronously waits on each nested ref's `digest_future.result()`. Under high concurrency every main-pool worker can land in `to_json` at once, each blocked on a child whose `obj_create` is queued *behind* it in the same executor — classic thread-pool starvation. Master happens to escape by ~1 worker of timing headroom; the deferral consumes it.

**Part 2** — this PR — restructures the dispatch so workers don't synchronously block on digest futures. At every save site that calls `to_json(value_with_refs)`, we pre-walk `value` for unresolved digest futures via `_collect_pending_digests`. If any are pending, we register the save via `future_executor.then(pending, send)` instead of `defer(send)`. The worker that wanted to schedule the save attaches a callback and returns to the pool; the continuation runs once digests resolve, by which point `to_json` is non-blocking. No worker holds a slot waiting for another worker.

`_collect_pending_digests` walks any value (primitives, dicts, lists, tuples, `ObjectRecord`, Refs) and collects unresolved digest / row-digest futures. Cycle-safe via `seen` set, skips done futures, treats Refs as leaves (does not recurse into their fields).

</details>

[WB-33844]: https://coreweave.atlassian.net/browse/WB-33844

### DIFF
--- a/tests/trace/test_save_nested_defer.py
+++ b/tests/trace/test_save_nested_defer.py
@@ -9,6 +9,8 @@ Covers:
 
 from __future__ import annotations
 
+import asyncio
+import logging
 import threading
 import time
 from concurrent.futures import Future
@@ -220,7 +222,6 @@ async def test_concurrent_finishes_drain(client: WeaveClient) -> None:
     Object whose digest is pending while the server is paused; releasing the
     pause must let every pending finalize_send fire and `_flush()` return.
     """
-    import asyncio
 
     class Item(weave.Object):  # type: ignore[name-defined]
         i: int = 0
@@ -235,24 +236,55 @@ async def test_concurrent_finishes_drain(client: WeaveClient) -> None:
     assert results == [k * 2 for k in range(20)]
 
 
-def test_call_end_error_is_logged_once(client: WeaveClient) -> None:
-    """Errors in the deferred Phase 2 (server.call_end) propagate via
-    `end_complete_future` and surface to the user the same way master does.
+class _CallEndRaisingServer:
+    """Server proxy that raises `RuntimeError` from `call_end`.
 
-    Regression guard: in earlier drafts we set `log_exception=False` on the
-    barrier future, which silently swallowed call_end errors. The fix is to
-    log on `end_complete_future`; this test fails if that flips back.
+    Plain class (no `TraceServerInterface` subclass) so `__getattr__`
+    forwards every other method to the inner server — subclassing the
+    interface would resolve abstract methods via MRO and never reach
+    `__getattr__`.
+    """
+
+    def __init__(self, inner: tsi.TraceServerInterface) -> None:
+        self._inner = inner
+
+    def __getattr__(self, item: str) -> Any:
+        return getattr(self._inner, item)
+
+    def call_end(self, *args: Any, **kwargs: Any) -> Any:
+        raise RuntimeError("simulated call_end failure")
+
+
+@pytest.mark.disable_logging_error_check
+def test_call_end_error_is_logged(
+    client: WeaveClient,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """Errors raised by `server.call_end` in the deferred Phase 2 must surface
+    as a "Task failed" log line via `_track_future`'s logger.
+
+    Regression guard: an earlier draft tracked the barrier future with
+    `log_exception=False`, which silently swallowed `call_end` errors. If
+    that flips back, the assertion below fails.
     """
 
     @weave.op
     def f(x: int) -> int:
         return x + 1
 
-    f(1)
-    f(2)
-    client._flush()
-    # Smoke: just asserting flush returns. The failure mode was a hang/silence,
-    # not a wrong return value.
+    original = client.server
+    client.server = _CallEndRaisingServer(original)
+    try:
+        with caplog.at_level(logging.ERROR):
+            f(1)
+            client._flush()
+    finally:
+        client.server = original
+
+    messages = [r.getMessage() for r in caplog.records]
+    assert any(
+        "Task failed" in m and "simulated call_end failure" in m for m in messages
+    ), f"Expected 'Task failed' log with our error; got: {messages}"
 
 
 def test_finish_call_returns_before_send(client: WeaveClient) -> None:
--- a/tests/trace/test_save_nested_defer.py
+++ b/tests/trace/test_save_nested_defer.py
@@ -1,0 +1,289 @@
+"""Tests for the deferred-output structural fix.
+
+Covers:
+1. `_collect_pending_digests` — pre-walk that finds unresolved digest futures.
+2. The `then(...)`-chained `_save_object_basic` path (no worker block on digest).
+3. The deferred `send_end_call` (Phase 1 + Phase 2) doesn't deadlock under
+   high concurrency or paused-server scenarios.
+"""
+
+from __future__ import annotations
+
+import threading
+import time
+from concurrent.futures import Future
+from contextlib import contextmanager
+from typing import Any
+
+import pytest
+
+import weave
+from weave.trace.object_record import ObjectRecord
+from weave.trace.refs import ObjectRef, OpRef, TableRef
+from weave.trace.weave_client import WeaveClient, _collect_pending_digests
+from weave.trace_server import trace_server_interface as tsi
+
+
+def _make_pending_future() -> Future[str]:
+    f: Future[str] = Future()
+    return f
+
+
+def _make_resolved_future(value: str) -> Future[str]:
+    f: Future[str] = Future()
+    f.set_result(value)
+    return f
+
+
+# ---------------------------------------------------------------------------
+# _collect_pending_digests: unit tests over arbitrary value shapes.
+# These don't touch the client; they verify the pre-walk that gates `to_json`
+# from blocking inside a worker thread.
+# ---------------------------------------------------------------------------
+
+
+class TestCollectPendingDigests:
+    def test_primitives_return_empty(self) -> None:
+        assert _collect_pending_digests(None) == []
+        assert _collect_pending_digests("hello") == []
+        assert _collect_pending_digests(42) == []
+        assert _collect_pending_digests(3.14) == []
+        assert _collect_pending_digests(True) == []
+
+    def test_containers_of_primitives_return_empty(self) -> None:
+        assert _collect_pending_digests([]) == []
+        assert _collect_pending_digests([1, 2, "a"]) == []
+        assert _collect_pending_digests((1, 2)) == []
+        assert _collect_pending_digests({}) == []
+        assert _collect_pending_digests({"a": 1, "b": [2, 3]}) == []
+        assert _collect_pending_digests({"nested": {"deep": {"value": 1}}}) == []
+
+    def test_object_ref_with_resolved_digest_returns_empty(self) -> None:
+        ref = ObjectRef(
+            entity="e", project="p", name="obj", _digest="resolved-digest"
+        )
+        assert _collect_pending_digests(ref) == []
+
+    def test_object_ref_with_pending_digest_is_collected(self) -> None:
+        pending = _make_pending_future()
+        ref = ObjectRef(entity="e", project="p", name="obj", _digest=pending)
+        assert _collect_pending_digests(ref) == [pending]
+
+    def test_object_ref_with_resolved_future_treated_as_pending_caller_handles(
+        self,
+    ) -> None:
+        # Done futures are skipped — the caller will read `.result()` without
+        # blocking, so we don't need the `then(...)` gate.
+        f = _make_resolved_future("done")
+        ref = ObjectRef(entity="e", project="p", name="obj", _digest=f)
+        assert _collect_pending_digests(ref) == []
+
+    def test_table_ref_collects_both_digest_and_row_digests(self) -> None:
+        digest = _make_pending_future()
+        row_digests: Future[list[str]] = Future()
+        ref = TableRef(
+            entity="e", project="p", _digest=digest, _row_digests=row_digests
+        )
+        result = _collect_pending_digests(ref)
+        assert digest in result
+        assert row_digests in result
+        assert len(result) == 2
+
+    def test_op_ref_collected(self) -> None:
+        pending = _make_pending_future()
+        ref = OpRef(entity="e", project="p", name="my_op", _digest=pending)
+        assert _collect_pending_digests(ref) == [pending]
+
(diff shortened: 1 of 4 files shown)


andrewtruong:
I think there may be a correctness (or at least expectation) issue here.

If this work is now pushed to a background thread, I think it's possible that changes to an op's output that happen after the function is executed could be recorded as if they happened during the function's execution.

For example:

```py
@weave.op
def func(): return {"result": [1]}

res = func()
res["result"].append(2)
```

In the previous case, I think the res call's output is always `[1]`, but in this case it could be either `[1]` or `[1, 2]`

Can you add a test for this?

gtarpenning:
> I think there may be a correctness (or at least expectation) issue here.
> 
> If this work is now pushed to a background thread, I think it's possible that changes to an op's output that happen after the function is executed could be recorded as if they happened during the function's execution.
> 
> For example:
> 
> ```python
> @weave.op
> def func(): return {"result": [1]}
> 
> res = func()
> res["result"].append(2)
> ```
> 
> In the previous case, I think the res call's output is always `[1]`, but in this case it could be either `[1]` or `[1, 2]`
> 
> Can you add a test for this?

added test and fixed most cases

tests/trace/test_save_nested_defer.py
```diff
@@ -0,0 +1,544 @@
...
+
+# ---------------------------------------------------------------------------
+# Integration: under paused-server (the deadlock fixture from
+# test_evaluation_performance), a basic op chain must complete cleanly at
+# default executor parallelism. Master without the structural fix deadlocks
+# here when the deferred `_save_nested_objects` is present.
+# ---------------------------------------------------------------------------
+
+
+class _BlockingTraceServer(tsi.TraceServerInterface):
```
andrewtruong: This could be generally useful in other tests. Should we move this to conftest?

tests/trace/test_save_nested_defer.py
andrewtruong, line 1: I may have missed it.  Is there a test checking that an object create failure will not hang flushing?

tests/trace/test_save_nested_defer.py
gtarpenning: Addressed in b31b56908c.

Moved `BlockingTraceServer` and the `paused` context manager into `tests/trace/conftest.py` (dropping the leading underscores since they're now shared helpers).

---
<sub>Resolved with AI assistance</sub>

tests/trace/test_save_nested_defer.py
gtarpenning: Addressed in 45b4de7705.

Added `test_obj_create_error_does_not_hang_flush`: an op whose output contains a `weave.Object` triggers `obj_create` via the deferred `schedule_send` path; the test wraps the server with an `_ObjCreateRaisingServer`, asserts `_flush()` returns within 10s (no hang on the `end_complete_future` barrier), and confirms the failure surfaces as a `Task failed` log line.

---
<sub>Resolved with AI assistance</sub>

gtarpenning: added

tests/trace/test_save_nested_defer.py
gtarpenning: moved

weave/trace/weave_client.py
```diff
@@ -993,8 +1121,12 @@ def finish_call(
         if self.postprocess_output:
             postprocessed_output = self.postprocess_output(postprocessed_output)
 
-        self._save_nested_objects(postprocessed_output)
-        output_as_refs = map_to_refs(postprocessed_output)
+        # `_save_nested_objects` and `map_to_refs` walk the output tree once
```
chance-wnb: This comment makes sene for the scope of the PR, but does it make sense to stay here to guide future developers? isn't it confusing that it talks about something(_save_nested_objects, map_to_refs) no longer happens at this spot?

chance-wnb:
> User code that synchronously inspects output.ref immediately after finish_call may briefly see no ref attached. Behavior change; 

I am slightly concerned on this: User will have no way to assert when the refs will be ready. When user writes code to read the refs after finish_call(), whether the refs fill up quickly or slowly makes no difference. And `_flush()` is a private method which user cannot call and depend on, this essentially leaves user in an awkward situation that they have to write sleep + loop waiting. 

Do we have something to say for this behavior change? does it make sense to expose _flush() or some await-able future? 

We could resolve this in a separate PR.

weave/trace/weave_client.py
```diff
@@ -1065,40 +1193,86 @@ def _deep_update(dst: dict[str, Any], src: dict[str, Any]) -> None:
...
-            maybe_redacted_output_as_refs = output_as_refs
-            if should_redact_pii():
-                from weave.utils.pii_redaction import redact_pii
+        # Phase 2: serialize `output_as_refs`, build the request, and send.
+        # Runs after any output-side digest futures resolve (see Phase 1).
+        # Errors are routed to `end_complete_future` so the on_end_complete
+        # callback fires once exactly when the whole pipeline finishes.
+        # Tracked in the executor's active set so `_flush()` waits for it,
+        # and `log_exception=True` so a server.call_end failure logs the
+        # same "Task failed" line master does (matches resilience tests).
```
chance-wnb: `master` is mentioned here. This code will become `master` :) might want to reword a bit.

weave/trace/weave_client.py
```diff
@@ -323,6 +323,134 @@ def map_to_refs(obj: Any) -> Any:
     return obj
 
 
+def _snapshot_mutable_containers(obj: Any, _memo: dict[int, Any] | None = None) -> Any:
```
chance-wnb: nit: does it make sense to push it to a utility file? (I think weave_client.py is pretty large already)

chance-wnb (approved):
LGTM!

andrewtruong (approved):

PR #6746, opened 2026-04-30
fix(weave): convert content to base 64 before completions requests

jwlee64:
## Description

https://coreweave.atlassian.net/browse/WB-32672

Before we would get an error abotu how openai doesnt know what to do with weave.content
Converts content objects back to base64, before sending completions create calls
allows responding to images in the playground

https://github.com/user-attachments/assets/332336b6-7784-4bf2-988c-146dd657904d

## Testing

manually
unit tests


weave/trace_server/base64_content_conversion.py
```diff
@@ -227,3 +230,84 @@ def process_complete_call_to_content(
...
+            metadata = json.loads(metadata_res.content)
+            mimetype = metadata.get("mimetype", "application/octet-stream")
+
+            content_res = trace_server.file_content_read(
+                FileContentReadReq(project_id=project_id, digest=content_digest)
+            )
+            b64_data = base64.b64encode(content_res.content).decode("ascii")
+        except Exception as e:
+            logger.warning("Failed to restore Content object to data URL: %s", e)
+            return val
```
gtarpenning, line 291: This `except Exception` silently re-introduces the bug this PR is fixing. If `file_content_read` errors or the metadata JSON fails to parse, we return the Content dict unchanged, which then hits litellm and 4xx's with the original `openai doesn't know what to do with weave.content` error, just buried under a WARN log instead of an exception.

Catch the specific failure modes (storage 404, `json.JSONDecodeError`) and let anything unexpected raise so the call fails loudly. Right now the fallback path is the failure mode, dressed up as resilience.

weave/trace_server/base64_content_conversion.py
```diff
@@ -227,3 +230,84 @@ def process_complete_call_to_content(
...
+        if isinstance(v, dict):
+            # Handle Weave's custom ContentMessage format: {type: "image", image: <content_object>}
+            # The frontend stores images as {type: K, [K]: CustomWeaveTypePayload} where K is
+            # the content type. Convert to the standard OpenAI image_url format so litellm
+            # can forward it to any LLM provider.
+            content_type = v.get("type")
+            if content_type == "image" and _is_content_object(v.get("image")):
+                data_url = _restore_one(v["image"])
+                if isinstance(data_url, str):
+                    return {"type": "image_url", "image_url": {"url": data_url}}
```
gtarpenning, line 307: This `image -> image_url` reshape is OpenAI/litellm message-protocol logic, but the module's docstring is `automatic detection and replacement of base64 encoded content with content objects stored in bucket storage`. Storage conversion shouldn't know what an OpenAI vision message looks like.

If a future caller of `restore_content_objects_to_base64` (agent path, storage-side restore, etc.) doesn't want the reshape, they get it anyway. Suggest splitting: keep this function pure (Content -> data URL), and put the message-format adaptation next to the completions handler that depends on it. That also gives `image`/`image_url`/`url`/`type` a natural home as constants instead of inline strings.

tests/trace_server/test_base64_content_conversion.py
```diff
@@ -353,5 +359,214 @@ def test_wav_file_base64_converted(self):
...
+    trace_server.file_content_read = MagicMock(
+        side_effect=[
+            FileContentReadRes(content=metadata),
+            FileContentReadRes(content=raw_bytes),
+        ]
+    )
+    return trace_server
+
+
+class TestRestoreContentObjectsToBase64:
```
gtarpenning, line 387: None of these tests exercise `completions_create` / `completions_create_stream`, which are the paths this PR actually changes. Every test calls `restore_content_objects_to_base64` directly with a MagicMock server.

The roundtrip test below is the one good high-fidelity case, but it still bypasses the call handlers. Worth adding one integration test that drives `completions_create` against a SQLite trace server with a faked litellm, asserting (1) messages handed to litellm contain `image_url` data URLs, not Content dicts, and (2) stored call inputs contain Content dicts, not raw base64. Otherwise this regresses silently the next time someone refactors the completions path.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6205,7 +6213,19 @@ def completions_create(
...
         req.inputs.messages = initial_messages
+        stored_inputs = {
+            **req.inputs.model_dump(exclude_none=True, exclude={"vertex_credentials"})
+        }
+        stored_inputs = replace_base64_with_content_objects(
+            stored_inputs, req.project_id, self
+        )
+        stored_output = replace_base64_with_content_objects(
+            res.response, req.project_id, self
+        )
```
gtarpenning, line 6227: This `stored_output = replace_base64_with_content_objects(...)` exists in the non-streaming path, but the streaming path's `_create_tracked_stream_wrapper` aggregates output without the equivalent replace before storing. So an image-returning streamed completion would store raw base64, while the non-streaming twin stores a Content object.

Today's user-facing bug is image inputs (symmetric across both paths), but the storage half is now asymmetric. Either apply the replace to `aggregated_output` inside the wrapper before building the end call, or document why streaming output is exempt.

PR #6748, opened 2026-04-30
chore(weave): Dont call server_info_res twice, speeding up init

andrewtruong:
Calling it twice is unnecessary.  This saves about ~150ms of init time


chance-wnb (approved):
Good optimization. LGTM!

rgao-coreweave (approved):
LGTM, nice work!

-150ms is very nice

PR #6749, opened 2026-04-30
chore(weave): agent span stats API

bcsherma:
## Summary

Adds a chart-ready stats API for GenAI agent spans.

This introduces `/agents/spans/stats`, which lets callers request time-bucketed aggregations over agent span data using the existing agent span filter/query model. The endpoint supports grouping by span columns or typed custom attributes, and lets clients request explicit metric specs such as counts, sums, averages, min/max, distinct counts, boolean counts, and numeric percentiles.

## API Shape

The request is centered around:

- `project_id`
- `query` for existing span filters
- `start` / `end` time bounds
- optional `granularity`
- optional `timezone`
- optional `group_by`
- required `metrics`

The response returns resolved time bounds, granularity, column metadata, and tabular rows suitable for frontend charts and grouped timeline views.

## Implementation

- Adds `AgentSpanStatsReq` / `AgentSpanStatsRes` and related metric/column types.
- Adds a ClickHouse query builder for bucketed span stats.
- Supports derived metrics like `duration_ms`, `total_tokens`, `is_error`, and `is_invocation`.
- Supports aggregating raw span fields and typed custom attribute map values.
- Adds result-size safeguards for grouped time-series queries.
- Wires the API through the trace server, external adapter, and agent query handler.

## Testing

Adds coverage for:

- stats request validation
- SQL generation
- grouped and ungrouped stats queries
- custom attribute metrics and group-bys
- endpoint auth and empty-result behavior


weave/trace_server/query_builder/agent_stats_query_builder.py
```diff
@@ -0,0 +1,785 @@
...
+    ]
+
+    start_epoch = start.replace(tzinfo=datetime.timezone.utc).timestamp()
+    end_epoch = end.replace(tzinfo=datetime.timezone.utc).timestamp()
+    start_param = pb.add_param(start_epoch)
+    end_param = pb.add_param(end_epoch)
+    tz_param = pb.add_param(tz)
+    bucket_interval_param = pb.add_param(granularity_seconds)
+    group_limit = _effective_group_limit(req, start, end, granularity_seconds)
+    group_limit_slot = pb.add(group_limit, param_type="UInt64")
```
nikumar1206: does group_limit matter when we are doing ungrouped queries?

maybe this should only be part of the _build_ungrouped_stats_query()

weave/trace_server/query_builder/agent_stats_query_builder.py
```diff
@@ -0,0 +1,785 @@
...
+
+_COLUMN_VALUE_TYPES: dict[str, AgentSpanStatsValueType] = {
+    **{
+        column: _SEMCONV_TYPE_TO_STATS_TYPE[semconv.ATTRIBUTES[key].type]
+        for key, column in semconv.CANONICAL_KEY_TO_COLUMN.items()
+    },
+    **_CORE_SPAN_VALUE_TYPES,
+}
+
+_DERIVED_VALUE_TYPES: dict[AgentSpanStatsDerivedMetric, AgentSpanStatsValueType] = {
```
nikumar1206: this is defined twice right?

weave/trace_server/agents/types.py
```diff
@@ -41,6 +44,174 @@
...
+]
+AgentSpanStatsDerivedMetric = Literal[
+    "duration_ms",
+    "total_tokens",
+    "is_error",
+    "is_invocation",
+]
+
+_IDENT_RE = r"^[a-zA-Z_][a-zA-Z0-9_]*$"
+_DERIVED_VALUE_TYPES: dict[AgentSpanStatsDerivedMetric, AgentSpanStatsValueType] = {
```
nikumar1206: defined here as well?

nikumar1206 (approved):
lgtm, left 2 comments

weave/trace_server/agents/types.py
gtarpenning: doesn't this also live in `weave/trace_server/query_builder/agent_stats_query_builder.py`?

gtarpenning: probably should keep this

weave/trace_server/agents/types.py
```diff
@@ -41,6 +44,174 @@
...
+
+class AgentSpanStatsRes(BaseModel):
+    """Response containing chart-ready agent span stats rows."""
+
+    start: datetime.datetime
+    end: datetime.datetime
+    granularity: int
+    timezone: str
+    columns: list[AgentSpanStatsColumn] = Field(default_factory=list)
+    rows: list[dict[str, Any]] = Field(default_factory=list)
```
gtarpenning: maybe: 
```
AgentSpanStatsCell = datetime.datetime | str | int | float | bool | None 

list[dict[str, AgentSpanStatsCell]]
```

tests/trace_server/query_builder/test_agent_stats_query_builder.py
```diff
@@ -0,0 +1,442 @@
...
+
+
+def test_basic_stats_query_uses_query_filter_and_bucket() -> None:
+    pb = ParamBuilder("genai")
+    req = _req(
+        query=Query.model_validate(
+            {
+                "$expr": {
+                    "$eq": [
+                        {"$getField": "agent.name"},
```
gtarpenning, line 291: where does this come from? we use `agent_name` in the `AgentGroupByRef` seems weird to have both `agent.name` and `agent_name`

bcsherma: good call, bc we should just use the semconv names for the group by keys, eg `agent.name` not a special `agent_name` key, I made that change

PR #6751, opened 2026-04-30
feat(weave): emit `weave.score_agent_spans` event for genai `turn_ended` spans

jtschoonhoven:
## Description

Wires up the producer side of the new agent scoring worker for genai spans. [This PR](https://github.com/wandb/core/pull/43212) wires up the consumer side.

- Adds a new `weave.score_agent_spans` Kafka event that gets produced for genai `turn_ended` spans
- Kafka event is gated on `WEAVE_ENABLE_AGENT_SCORING` env var
- Hardcodes prompt headers and footer for `ClassifierMonitors` (previously these were hardcoded in the frontend which has been buggy) and adds a new header for classifying agent spans

## JIRA

[WB-33383](https://coreweave.atlassian.net/browse/WB-33383?atlOrigin=eyJpIjoiMjdkMGIxZWEyNTAzNGVmOGE3MTM0MmI1NTBiMDY4ZTkiLCJwIjoiaiJ9)

## Testing

- [x] Unit tests
- [x] Tested locally

[WB-33383]: https://coreweave.atlassian.net/browse/WB-33383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/kafka.py
```diff
@@ -184,6 +189,51 @@ def produce_score_calls(
         if flush_immediately:
             self.flush(0)
 
+    @ddtrace.tracer.wrap(name="kafka_producer.produce_score_agent_spans")
+    def produce_score_agent_spans(self, event: ScoreAgentSpansEvent) -> None:
```
gtarpenning, line 153: DRY with `produce_score_calls`?

weave/trace_server/kafka.py
```diff
@@ -184,6 +189,51 @@ def produce_score_calls(
...
+                    "event_type": event.event_type,
+                    "project_id": event.project_id,
+                    "trace_id": event.trace_id,
+                },
+            )
+            set_root_span_dd_tags({"kafka.producer.buffer_size": buffer_size})
+            return
+
+        if buffer_size >= self.max_buffer_size * 0.5:
+            buffer_percentage = (buffer_size / self.max_buffer_size) * 100
```
gtarpenning: we should also pull this out as const while we are here

weave/trace_server/agents/clickhouse.py
```diff
@@ -377,11 +383,27 @@ def insert_otel_spans(self, req: GenAIOTelExportReq) -> GenAIOTelExportRes:
...
+
         if span_rows:
             self._ch_client.insert(
                 "spans", data=span_rows, column_names=ALL_SPAN_INSERT_COLUMNS
             )
 
+        if flush_kafka_events and self._kafka_producer is not None:
+            try:
+                self._kafka_producer.flush(0)
+            except Exception as e:
```
gtarpenning: `e` unused?

weave/trace_server/agents/kafka_events.py
```diff
@@ -0,0 +1,68 @@
...
+    conversation_id: str | None = None
+    agent_name: str | None = None
+    operation_name: str | None = None
+    request_model: str | None = None
+
+    def emit(self, producer: KafkaProducer) -> None:
+        """Produce this event to Kafka. Logs failures without raising."""
+        try:
+            producer.produce_score_agent_spans(self)
+        except Exception as e:
```
gtarpenning: `e` unused here too

weave/flow/monitor.py
```diff
@@ -110,6 +114,81 @@ def from_obj(cls, obj: WeaveObject) -> Self:
...
+        "<agent>",
+        "<name>{agent_name}</name>",
+        "<version>{agent_version}</version>",
+        "<description>{agent_description}</description>",
+        "<status>",
+        "  <code>{status_code}</code>",
+        "  <message>{status_message}</message>",
+        "</status>",
+        "<messages>",
+        "  <conversation-name>{conversation_name}</conversation_name>",
```
gtarpenning: technically this is a typo

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -6602,7 +6602,10 @@ def agent_conversation_chat(
         return AgentQueryHandler(self._query).conversation_chat(req)
 
     def genai_otel_export(self, req: GenAIOTelExportReq) -> GenAIOTelExportRes:
-        return AgentWriteHandler(self.ch_client).insert_otel_spans(req)
+        # Only emit score_agent_spans events when the consumer side is enabled.
+        # There are 2 gates: `wf_enable_agent_scoring` and `wf_enable_online_eval`.
+        producer = self.kafka_producer if wf_env.wf_enable_agent_scoring() else None
```
gtarpenning: could add `wf_enable_online_eval()` here too just to be safe

weave/trace_server/agents/clickhouse.py
```diff
@@ -377,11 +383,27 @@ def insert_otel_spans(self, req: GenAIOTelExportReq) -> GenAIOTelExportRes:
                     span_rows.append(genai_span_to_row(genai_row))
                     accepted += 1
 
+                    # Emit kafka events when the producer is wired in (gated
+                    # off in environments where the consumer hasn't shipped).
+                    if self._kafka_producer is not None and (
+                        event := ScoreAgentSpansEvent.from_row(genai_row)
+                    ):
+                        flush_kafka_events = True
+                        event.emit(self._kafka_producer)
```
gtarpenning: what's our contract here, i think we should consider publishing to bufstream after we get an ack from clickhouse? could be wrong though depending on what you are using this for.

weave/trace_server/kafka.py
```diff
@@ -13,6 +14,10 @@
 )
 
 from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.agents.kafka_events import (
+    SCORE_AGENT_SPANS_TOPIC,
+    ScoreAgentSpansEvent,
```
gtarpenning, line 19: i was sorta thinking this file would stay only Kafka related, more like a common infra file, seeing this import has me thinking about where `ScoreAgentSpansEvent` should really be used 🤔

weave/trace_server/agents/clickhouse.py
gtarpenning: flushing _could_ happen at any time, so emitting after insert is probably best i think

weave/trace_server/agents/clickhouse.py
```diff
@@ -311,10 +314,12 @@ def _run_trace_detail_query(
 class AgentWriteHandler:
     """Write-side operations for agent data.
 
-    Takes a `ch_client` for `insert` calls, which have no query wrapper.
+    Takes a `ch_client` for `insert` calls, which have no query wrapper and
+    a `kafka_producer` so the ingest path can publish a "weave.score_agent_spans" event.
     """
 
     _ch_client: CHClient
+    _kafka_producer: KafkaProducer | None = None
```
gtarpenning: maybe we can do the kafka emitting in `genai_otel_export` to mirror the other insert flows, and centralizes our main bufstream contract to just the clickhouse file (I realize how that sounds). Then `genai_otel_export` could be: 

```python
def genai_otel_export(self, req: GenAIOTelExportReq) -> GenAIOTelExportRes:
      res, score_event_rows = AgentWriteHandler(self.ch_client).insert_otel_spans(req)

      if not wf_env.wf_enable_agent_scoring() or not score_event_rows:
          return res

      producer = self.kafka_producer
      for row in score_event_rows:
          if event := ScoreAgentSpansEvent.from_row(row):
              event.emit(producer)
      try:
          producer.flush(0)
      except Exception:
          logger.exception("Failed to flush Kafka producer during OTel span ingest")

      return res
```

weave/trace_server/agents/kafka_events.py
```diff
@@ -0,0 +1,68 @@
...
+    ended_at: datetime.datetime
+    conversation_id: str | None = None
+    agent_name: str | None = None
+    operation_name: str | None = None
+    request_model: str | None = None
+
+    def emit(self, producer: KafkaProducer) -> None:
+        """Produce this event to Kafka. Logs failures without raising."""
+        try:
+            producer.produce_score_agent_spans(self)
```
gtarpenning, line 40: this is all small metadata right?

gtarpenning (commented):
couple comments about the structure, happy to fast review or resolve if you have conflicting thoughts.

weave/trace_server/agents/kafka_events.py
jtschoonhoven: Yeah just short strings, probably a few dozen bytes. Since you mention it I can drop a couple of these fields. We don't need request_model in the event message

weave/trace_server/kafka.py
jtschoonhoven: I agree, but this appears to be the current pattern and I'm not certain it's worth the refactor

weave/trace_server/agents/clickhouse.py
```diff
@@ -320,13 +323,15 @@ class AgentWriteHandler:
...
+    def insert_otel_spans(
+        self, req: GenAIOTelExportReq
+    ) -> tuple[GenAIOTelExportRes, list[AgentSpanCHInsertable]]:
         """Ingest OTel spans into the spans table.
 
         The `messages` search table is populated by a ClickHouse
         materialized view off the spans table (migration 030).
         """
-        span_rows: list[list[object]] = []
+        span_rows: list[AgentSpanCHInsertable] = []
```
gtarpenning, line 334: nice

gtarpenning (approved):
lgtm!

PR #6756, opened 2026-05-01
fix(weave): fix tz-aware/naive compare bug in agent-scoring-worker

jtschoonhoven:
## Description

Bugfix: was previously comparing tz-aware with tz-naive datetimes in agents/kafka_events.py.

This code is gated behind an env var that is currently disabled in all environments, so no services were affected.


bcsherma (approved):
were you testing me?

jtschoonhoven:
> were you testing me?

CONSTANT VIGILANCE

PR #6757, opened 2026-05-02
feat(weave): Session SDK ergonomics for manually-instrumented agents

rgao-coreweave:
## Summary

A bundle of ergonomic additions for apps that hand-build chat / tool / agent spans (instead of using one of the autopatched integrations). Every manual-instrumentation caller hits the same friction: re-encoding tool I/O, re-parsing data URLs, hand-converting OpenAI message shapes, and assigning ten fields per LLM call. This PR pulls those into the Session SDK so the call site stays compact.

**Consumer:** [wandb/core#43000](https://github.com/wandb/core/pull/43000) — wb_agent dropped ~150 LOC of weave-shape transforms and now uses the new classmethod surface throughout.

### Structured payloads on `Tool` and the tool message parts

`Tool.arguments` / `Tool.result` and `ToolCallPart.arguments` / `ToolCallResponsePart.response` accept `str | dict | list | int | float | bool | None` (alias `JSONStringInput`). Strings pass through unchanged; non-strings are JSON-encoded — at construction for the message parts (with `validate_assignment=True` so post-hoc edits also coerce), at OTel emission for the `Tool` span. **The wire format remains a string per GenAI semconv.**

```python
# Before
t.result = json.dumps({"hits": 3})
# After
t.result = {"hits": 3}
```

### `LLM.attach_media_url(url, modality=...)`

Unifies `data:` URLs (parsed into mime + payload, emitted as `kind=blob`) and plain URIs (emitted as `kind=uri`). Empty URL is a no-op. Returns `self` for chaining.

### High-level `Message` constructors

`Message.user(text)`, `Message.system(text)`, `Message.assistant(text=..., tool_calls=[...])`, `Message.tool_result(call_id, output)`. The `assistant` constructor promotes to the explicit-parts API only when `tool_calls` are present.

### `LLM.record(...)` one-shot recorder

Collapses the eight-attribute end-of-call recording into one keyword call. Only fields explicitly passed (non-`None`) are applied — partial calls preserve existing values, which matters for code paths that record input early (so mid-stream errors still have it on the span) and output later. `reasoning` accepts either a `Reasoning` instance or a plain string.

### Provider extractors as classmethods on the Session SDK types

The discoverable surface for manually-instrumented agents — IDE autocompletion on `Message.<dot>` / `Usage.<dot>` / `Reasoning.<dot>` shows the per-provider converters:

- `Message.from_openai_responses_input(items)` → `(messages, media_attachments)` ready to assign to `LLM.input_messages` / `LLM.media_attachments`
- `Reasoning.from_openai_responses(part)` → `Reasoning | None` for the OpenAI Responses reasoning shape
- `Usage.from_openai_responses(response)` → `Usage` from an OpenAI Responses `Response`. Empty when `response.usage is None`.
- `Usage.from_anthropic(message)` → `Usage` from an Anthropic `Message`. Handles `Optional[int]` cache fields.

Implementations live in `weave.integrations.openai.responses` (lazy-imported into the classmethods), keeping the core types module free of provider-shape knowledge. Module-level `input_to_weave` / `reasoning_to_weave` are kept as back-compat aliases for older imports.

### `start_llm` provider auto-detection

When `provider_name=` is omitted, the SDK derives it from the model identifier:

| Prefix | Provider |
|---|---|
| `gpt-*`, `o1-*`, `o3-*`, `o4-*`, `o5-*`, `text-*` | `openai` |
| `claude-*` | `anthropic` |
| `gemini-*`, `models/gemini-*` | `google` |
| `mistral-*`, `open-mistral-*` | `mistral` |
| `command-*` | `cohere` |
| `llama-*`, `meta-llama/*` | `meta` |
| `deepseek-*` | `deepseek` |

Conservative — unknown names resolve to `""` (unknown), and explicit `provider_name=` always wins. Removes a foot-gun where forgetting `provider_name=` produced silently bad OTel telemetry (`gen_ai.provider.name=""`).

## Design notes

- **Why classmethods on the SDK types instead of loose functions?** Discoverability. `Message.<dot>` in an IDE surfaces the converter; an opaque `from weave.integrations.openai.responses import input_to_weave` doesn't. Implementation still lives in the integrations package — the classmethods are thin lazy-import dispatchers — so the core types module stays provider-agnostic.
- **Why a union type instead of `Any` for the JSON-string fields?** The `JSONStringInput` alias enumerates the JSON-value types so mypy / pyright can verify call sites without `# type: ignore`. After validation the stored value is always `str`.
- **Why coerce at construction for parts but at emission for `Tool`?** Parts flow through `_message_to_parts` → `model_dump(exclude_defaults=True)`, which would emit a dict-as-dict if storage weren't a string. `Tool` skips `model_dump` — emission goes through a hand-built attrs dict — so we have a natural conversion point in `Tool.end()`.
- **Why `validate_assignment=True` on the parts?** Without it, `part.arguments = some_dict` would silently store a dict (bypassing the validator), and the wire serializer would emit it as a nested object.
- **Why is provider auto-detection prefix-based instead of regex / external table?** Conservative on purpose. The mapping is short, longer prefixes come first, and unknown names fall through to `""` rather than guessing.

## Test plan

- 8 new test classes covering structured-payload coercion, message builders, `attach_media_url`, `LLM.record`, and the OpenAI Responses adapters
- 31 test functions / 47 parametrized cases — variants that differ only in input/output (string vs dict tool args, data URL vs plain URL images, four `returns-None` reasoning shapes, etc.) live under `@pytest.mark.parametrize`; structurally distinct cases stay as separate methods
- `uv run --group test python -m pytest tests/session/` — 201 / 201 pass
- `uvx ruff check` / `format --check` clean
- `uvx pyright` clean

## Files

- `weave/session/types.py` — `_to_json_string` helper, `JSONStringInput` alias, validators on `ToolCallPart` / `ToolCallResponsePart`, `Message` builders + `Message.from_openai_responses_input`, `Usage.from_openai_responses` + `Usage.from_anthropic`, `Reasoning.from_openai_responses`
- `weave/session/session.py` — `Tool` field types widened with emission-time serialization, `LLM.attach_media_url`, `LLM.record`, `_parse_data_url`, `_detect_provider` + auto-detect in `start_llm`
- `weave/integrations/openai/responses.py` — implementation backing the classmethods (`_input_to_weave`, `_reasoning_to_weave`) plus back-compat aliases
- `tests/session/test_session_otel.py` — 8 new test classes (parametrized where input/output is the only variant)


weave/session/types.py
```diff
@@ -134,12 +241,60 @@ class Usage(BaseModel):
...
+        ``Usage`` in that case so they can still pass the result through to
+        ``LLM.record(usage=...)`` unconditionally.
+        """
+        usage = response.usage
+        if usage is None:
+            return cls()
+        return cls(
+            input_tokens=usage.input_tokens,
+            output_tokens=usage.output_tokens,
+            reasoning_tokens=usage.output_tokens_details.reasoning_tokens,
```
chance-wnb: usage.output_tokens_details and usage.input_tokens_details could be None.

weave/integrations/openai/responses.py
```diff
@@ -0,0 +1,205 @@
...
+
+# ---------------------------------------------------------------------------
+# Back-compat module-level aliases
+# ---------------------------------------------------------------------------
+#
+# The preferred public surface is now ``Message.from_openai_responses_input``
+# and ``Reasoning.from_openai_responses``. These wrappers exist so older
+# call sites that import directly from this module keep working.
+
+input_to_weave = _input_to_weave
```
chance-wnb: This is a new file, I bet there are no consumers of it, and `so older
call sites that import directly from this module keep working` probably doesn't stand.

weave/integrations/openai/responses.py
```diff
@@ -0,0 +1,205 @@
...
+        if item_type == "function_call_output":
+            messages.append(_function_call_output_message(item))
+            continue
+        if item_type == "reasoning":
+            continue
+
+        if role in _USER_LIKE_ROLES:
+            content = item.get("content")
+            text = _extract_text_content(content)
+            if text:
```
chance-wnb: It seems this requires text to be present. what if the response has an image only? would it break `_serialize_input_messages()` ?

weave/integrations/openai/responses.py
```diff
@@ -0,0 +1,205 @@
...
+    messages: list[Message] = []
+    attachments: list[MediaAttachment] = []
+    seen_urls: set[str] = set()
+
+    for item in items:
+        item_type = item.get("type")
+        role = item.get("role")
+
+        if item_type == "function_call":
+            messages.append(_function_call_message(item))
```
chance-wnb: It seems this is producing `1 message - 1 tool call` message list. 

does this match the

```
def assistant(
        cls,
        text: str = "",
        *,
        tool_calls: list[ToolCallPart] | None = None,
    ) -> Message
```
shape? which seems to suggest 1 message, multiple tool calls.

weave/session/types.py
rgao-coreweave: Good catch — fixed in f0f6d46. Defaulting both to 0 when the details object is missing via `getattr(..., 0) or 0`, matching the defensive pattern in `openai_sdk.py:_normalize_openai_cache_tokens` for partial / streaming responses.

weave/integrations/openai/responses.py
rgao-coreweave: You're right — the file was new in this PR with no real older callers, only the tests in this PR imported the aliases. Removed the shim entirely and migrated the tests to call `Message.from_openai_responses_input` / `Reasoning.from_openai_responses` directly. f0f6d46.

weave/integrations/openai/responses.py
rgao-coreweave: You're right — image-only user input would silently drop the attachment because `_serialize_input_messages` short-circuits on `messages == []`, and even with mixed inputs the image would bind to the wrong (last text-bearing) user message. Fixed by always emitting a user `Message` so the attachment has a slot to bind to. Added a regression test (`test_image_only_user_message_keeps_message_slot`). f0f6d46.

weave/integrations/openai/responses.py
rgao-coreweave: Right — separate Messages don't match the `assistant(tool_calls=[...])` shape and split a single parallel-call turn into N turns. Now buffering consecutive `function_call` items and flushing them as one `Message.assistant(tool_calls=...)` whenever a non-function-call item arrives (or at end-of-list). Added `test_parallel_function_calls_coalesce_to_one_assistant_message`. f0f6d46.

rgao-coreweave:
Thank you @chance-wnb for the feedback!

weave/session/adapters/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Provider adapters for the Weave Session SDK.
+
+These modules convert provider-specific wire formats (OpenAI Responses,
+Anthropic Messages, etc.) into the provider-agnostic types defined in
+``weave.session.types``. Each adapter is a leaf module — it imports
+types but is not imported by them, so ``types.py`` stays free of
+provider knowledge.
+
+Adapters are optional: importing ``weave.session.adapters.openai``
+requires ``openai`` to be installed; ``weave.session.adapters.anthropic``
```
chance-wnb: nit: 

> importing ``weave.session.adapters.openai``
requires ``openai`` to be installed;

is this really true? You have

```
if TYPE_CHECKING:
    from openai.types.responses import Response
```

in adapters/openai.py. so it only matters at type-checking(mypy) time. Client side runtime import should be fine.

Same for `anthropic`.

weave/session/adapters/openai.py
```diff
@@ -0,0 +1,240 @@
...
+
+# ---------------------------------------------------------------------------
+# Internal helpers
+# ---------------------------------------------------------------------------
+
+
+def _to_tool_call_part(item: dict[str, Any]) -> ToolCallPart:
+    arguments = item.get("arguments", "")
+    if not isinstance(arguments, str):
+        arguments = json.dumps(arguments)
```
chance-wnb: In your definition of ToolCallPart, you have

```
``JSONString`` annotation lets callers pass a dict / list / scalar
    at construction or assignment and the SDK JSON-encodes it
```

so here probably you don't even need to care about manually encoding of the arguments field here. Non-str type is acceptable and will be encoded automatically downstream.

weave/session/adapters/openai.py
```diff
@@ -0,0 +1,240 @@
...
+    return ToolCallPart(
+        id=str(item.get("call_id", "")),
+        name=str(item.get("name", "")),
+        arguments=arguments,
+    )
+
+
+def _function_call_output_message(item: dict[str, Any]) -> Message:
+    output = item.get("output")
+    response = output if isinstance(output, str) else json.dumps(output)
```
chance-wnb: Same issue as above. Manual encoding of the `response` field is probably not needed.

chance-wnb (approved):
Several minor nits. Overall looks great. Approving... Thanks!

weave/session/adapters/__init__.py
rgao-coreweave: Good catch — that line was wrong. Both provider imports are `TYPE_CHECKING`-only and the call sites duck-type `response.usage` / `message.usage`, so the adapter modules import and run without `openai` / `anthropic` installed. Docstring updated in d8883ed.

weave/session/adapters/openai.py
rgao-coreweave: Right — `JSONString` already coerces on assignment, so the manual `json.dumps` was dead. Dropped it in d8883ed (side benefit: the validator uses `ensure_ascii=False, default=str`, friendlier than the local `json.dumps` for non-ASCII tool args).

weave/session/adapters/openai.py
rgao-coreweave: Same fix — letting `JSONString` handle it (d8883ed). One edge note: a missing `output` field now serializes to `""` instead of the literal string `"null"` (since `_to_json_string(None) → ""`). OpenAI never actually omits `output` here, and `""` is a friendlier sentinel, so this seems fine. Also dropped the now-unused `import json`.

rgao-coreweave:
> Several minor nits. Overall looks great. Approving... Thanks!

Thanks for the feedback, I appreciate it

PR #6762, opened 2026-05-04
chore(weave): partition agent scoring worker topic on trace_id

jtschoonhoven:
## Description

Changes the partition key for the agent scoring worker from `project_id` to `trace_id`. See [this slack thread](https://coreweave.slack.com/archives/C03BSTEBD7F/p1777915325401149) for details and discussion. The agent scoring worker isn't deployed yet so this won't affect any live systems.


gtarpenning (approved):
lgtm

PR #6764, opened 2026-05-04
perf(weave): add trace_id bloom filter index on calls_merged

gtarpenning:
https://coreweave.atlassian.net/browse/WB-34437

## Summary

Adds a `bloom_filter(0.01)` skip-index on `ifNull(trace_id, '')` for `calls_merged` so trace-scoped lookups can prune granules instead of full-scanning the multi-TB table.

**Why**: trace-scoped reads (`WHERE trace_id = 'x'` or `trace_id IN (...)`) are the hot path behind every "open this trace" navigation in the UI, plus stats/cost rollups grouped by trace. `calls_merged` is ordered by `(project_id, id)`, and trace IDs are random (UUIDs / OTel span IDs), so the primary key gives zero locality on `trace_id` and today these queries fan out across every granule in the project's partition. A bloom filter on `trace_id` lets ClickHouse skip granules that provably don't contain a given trace, turning a full project scan into a sparse read.

**Why the `ifNull` wrap**: `trace_id` on `calls_merged` is `SimpleAggregateFunction(any, Nullable(String))` because pre-merge rows can transiently be NULL until the AggregatingMergeTree merges parts. A bloom filter on a Nullable column does NOT prune for `WHERE trace_id = 'x'` (Nullable equality is three-valued). Wrapping both the index expression and the query predicate as `ifNull(trace_id, '')` makes both sides non-nullable so the index actually fires. `calls_complete` keeps raw `trace_id` because the column is non-nullable there (and 028 already added a bloom on the raw expression).

**Example shape after the change** (trace detail load, the common case):

```sql
WITH filter_candidate_ids AS (
    SELECT id FROM calls_merged
    PREWHERE project_id = {pb_1:String}
    WHERE ifNull(trace_id, '') = {pb_0:String}       -- bloom-pruned
)
SELECT ... FROM calls_merged
WHERE id IN filter_candidate_ids
  AND (ifNull(trace_id, '') = {pb_0:String} OR trace_id IS NULL)
GROUP BY (project_id, id)
```

The CTE prunes the granule list using the index; the outer query only re-reads matched ids to materialize heavy columns. `OR trace_id IS NULL` is preserved on the outer arm so rows whose `trace_id` hasn't merged yet still come back correctly.

**Changes**:
- Migration 031: `ADD INDEX IF NOT EXISTS idx_trace_id_bloom ifNull(trace_id, '') TYPE bloom_filter(0.01) GRANULARITY 1`, then `MATERIALIZE INDEX idx_trace_id_bloom` to backfill existing parts (`mutations_sync` defaults to 0, so the migration returns immediately and ClickHouse backfills asynchronously). Manual run on a multi-TB replica completed materialization within a couple seconds.
- Query layer: `process_trace_id_filter_to_sql` and `build_calls_filter_sql` emit `ifNull(trace_id, '')` for the equality/IN arm only on `calls_merged`. `calls_complete` keeps raw `trace_id`.

## Testing

- Query-builder SQL tests updated for both `calls_merged` (ifNull + candidate-id CTE) and `calls_complete` (raw, no CTE). New `test_trace_id_filter_calls_complete_no_candidate_cte` pins the calls_complete-side behavior so we don't regress to wrapping it.
- Manual: ran on a replica with a couple TB; index materialization completed within a couple seconds.


weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1169,8 +1171,34 @@ def as_sql(self, pb: ParamBuilder, table_alias: str | None = None) -> str:
...
+        # call-part correctness and restricts to `id IN filter_candidate_ids`.
+        candidate_cte_name: str | None = None
+        candidate_cte_sql = self._build_filter_candidate_ids_cte_sql(
+            pb, table_alias_resolved
+        )
+        if candidate_cte_sql is not None:
+            ctes.add_cte(CTE_FILTER_CANDIDATE_IDS, candidate_cte_sql)
+            candidate_cte_name = CTE_FILTER_CANDIDATE_IDS
+
         if not should_optimize and not self.include_costs and not object_ref_conditions:
```
mscavezze-cw: The containing function is a monster. Having an early return in the middle of it doesn't feel great, especially as the contents of this if expands. It seems like a little bit of a trap for the next person who wants to add something that must be done before the function completes.

I wonder if it would make sense to do a little bit of refactoring in a follow-up PR:
- assign these to a well named local variable: `not should_optimize and not self.include_costs and not object_ref_conditions`.
  - I'm curious why those additional checks are there. If we want to include costs, then that implies optimization?
- move all (or the majority) of the code below this if statement into a separate function. Would it be called `optimize_sql`? Maybe not, given the number of checks in the if statement.

mscavezze-cw: If you were going to do a fast-follow refactor, consider extracting helper functions for each path of the `if use_filter_cte` block below

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1169,8 +1171,34 @@ def as_sql(self, pb: ParamBuilder, table_alias: str | None = None) -> str:
...
+        # Pre-narrow rows via the `idx_trace_id_bloom` skip-index when the
+        # query has a hardcoded trace_ids filter on calls_merged. The CTE
+        # uses the strict (no OR-IS-NULL) form so the bloom filter can prune
+        # granules; the outer query keeps the OR-IS-NULL form for unmerged-
+        # call-part correctness and restricts to `id IN filter_candidate_ids`.
+        candidate_cte_name: str | None = None
+        candidate_cte_sql = self._build_filter_candidate_ids_cte_sql(
+            pb, table_alias_resolved
+        )
+        if candidate_cte_sql is not None:
```
mscavezze-cw, line 1209: I am curious why this happens here and not inside the `if use_filter_cte` block below? It seems like we have a `should_optimize` flag that disables optimizations, but we perform this optimization regardless of that flag. Is there more nuance about what `should_optimize` means?

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -2367,35 +2425,85 @@ def process_op_name_filter_to_sql(
...
 
 
+def _trace_id_match_sql(
+    trace_ids: list[str],
+    field_expr: str,
+    param_builder: ParamBuilder,
+) -> str:
+    """Build a `field_expr = ?` or `field_expr IN ?` clause for `trace_ids`.
+
+    Returns "" when `trace_ids` is empty. Single-element lists use equality
```
mscavezze-cw, line 2460: I wonder whether the cut-over point for optimizing lists to equality compares happens when list length is 1, or somewhat higher.

I'm not sure if is worthwhile to tune that right now, but I thought I'd call it out since some consideration for optimization has already happened.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -2367,35 +2425,85 @@ def process_op_name_filter_to_sql(
...
+        hardcoded_filter.filter.trace_ids, field_expr, param_builder
+    )
+    if not trace_cond:
+        return ""
+
+    # `calls_complete.trace_id` is non-nullable `String`, so the OR-IS-NULL
+    # arm only applies to `calls_merged`, where unmerged call parts can have
+    # a NULL aggregated trace_id.
+    if read_table != ReadTable.CALLS_MERGED:
+        return f" AND {trace_cond}"
```
mscavezze-cw: Consider adding some parenthesis for obvious safety (rather than clever safety, or maybe subtly buggy)

`return f" AND ({trace_cond})"`

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -2367,35 +2425,85 @@ def process_op_name_filter_to_sql(
...
+    hardcoded_filter: HardCodedFilter | None,
+    param_builder: ParamBuilder,
+    table_alias: str,
+) -> str:
+    """Strict (no `OR IS NULL`) trace_id condition for the candidate-id CTE.
+
+    The OR-IS-NULL clause that the outer query keeps for unmerged-call-part
+    correctness defeats the bloom filter, so this strict form runs alone inside
+    `filter_candidate_ids` to maximize granule pruning. Returns "" when no
+    trace_ids are set. Only meaningful for calls_merged.
```
mscavezze-cw, line 2517: While not being a benefit, it is still safe for `calls_complete`, right? Or would there be issues because `calls_complete` uses different column types?

mscavezze-cw: Maybe it's a moot point anyway, since we raise below.

weave/trace_server/migrations/032_add_bloom_filter_trace_id_index.up.sql
```diff
@@ -0,0 +1,25 @@
...
+-- `ifNull(trace_id, '')` (and matching the predicate in the query layer
+-- character-for-character) makes the comparison non-nullable on both
+-- sides so the index actually fires.
+ALTER TABLE calls_merged
+    ADD INDEX IF NOT EXISTS idx_trace_id_bloom
+        ifNull(trace_id, '') TYPE bloom_filter(0.01) GRANULARITY 1;
+
+-- Kick off background materialization for existing parts. mutations_sync
+-- defaults to 0, so this returns immediately and ClickHouse runs the
+-- mutation asynchronously instead of blocking the migration on a full-
```
mscavezze-cw, line 22: What happens to queries that happen before the index has been fully created? Do they succeed at the current speed? Or do they fail (or return incorrect results) because they try to reference a missing or incomplete index?

mscavezze-cw (commented):
Partial review, will take a look at tests next

mscavezze-cw (commented):
Reviewed tests. No additional comments.

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: i'm pretty sure the bloom-filter prune will be universally beneficial for trace-scoped reads — even the trivial single-pass query benefits from inheriting a candidate-id list, because it shrinks the rows scanned by the outer GROUP BY/aggregation. `should_optimize` is indeed poorly named, i'll update that

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: its mostly for cleanliness of the final query, I think we can actually remove this.

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: yeah

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: its safe, and unhelpful. There is already an index on `calls_complete` and we dont need to do the `ifNull` handling because its already non-null

weave/trace_server/migrations/032_add_bloom_filter_trace_id_index.up.sql
gtarpenning: In theory, ClickHouse skip indexes are purely an optimization layer, never a correctness layer. That being said there have been issues in the past. I strongly suspect that its a total non-issue, but worth a once-over of the docs. IIRC skip index marks only appear for the query planner if they are complete for a given granule; as the planner goes through the plan if there is ever an incomplete skip index it just doesn't use it.

weave/trace_server/calls_query_builder/calls_query_builder.py
mscavezze-cw: Consider keeping it as-is for this PR. There's a non-trivial chance of a perf regression, and it'll suck if there are multiple things to look at. If you do want to make the change, maybe make a ticket and revisit a week or two after the bloom filter has some time to soak in the wild.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1182,6 +1186,13 @@ def as_sql(self, pb: ParamBuilder, table_alias: str | None = None) -> str:
         # uses the strict (no OR-IS-NULL) form so the bloom filter can prune
         # granules; the outer query keeps the OR-IS-NULL form for unmerged-
         # call-part correctness and restricts to `id IN filter_candidate_ids`.
+        #
+        # Always-on: independent of `should_use_filter_cte`, because the bloom
+        # prune pays for itself even on the trivial single-pass path and the
+        # CTE itself is cheap (one extra SELECT, no replicated ordering).
+        # TODO: as_sql is getting long. Follow-up to hoist the early-return
```
mscavezze-cw: Consider adding a jira ticket, unless you're strongly confident that you'll get to this work before something higher priority comes up. (For example, if you already have a child branch and just need a review)

Even then, maybe it makes sense to add one so that it shows up on the sprint board.

mscavezze-cw (approved):
Strong recommendation to back out the removal of the `if len(trace_ids) == 1` branch for now, to limit the number of changes that could have an impact on perf. 

https://github.com/wandb/weave/pull/6764#discussion_r3267934822

weave/trace_server/calls_query_builder/calls_query_builder.py
gtarpenning: yep, fair

neutralino1 (approved):

PR #6765, opened 2026-05-05
fix(weave): set obj `is_latest` based on `created_at`

jtschoonhoven:
## JIRA

[WB-34036](https://coreweave.atlassian.net/browse/WB-34036?atlOrigin=eyJpIjoiNWZiMDcxMmZjMTc5NGU1YmEzYWQ4YzMxMjMyZGY1OGYiLCJwIjoiaiJ9
)

## Description

When querying on the object versions table we append an `is_latest` column that is based on when each object version was "first seen". The version is based on a hash of the object, so if a change reverts back to a previously-seen version, we won't treat that as new. Currently, we also don't update `is_latest`, which is a bug:

```
Error-classifiers(active=true)  -> version 1  # new monitor (latest)
Error-classifiers(active=false) -> version 2  # updated monitor (latest)
Error-classifiers(active=true)  -> version 1  # toggled back to true (NOT latest)
```

To fix this we instead need to set `is_latest` based on `created_at` instead of `first_seen`.

## Testing

Tested locally

## Before

https://github.com/user-attachments/assets/df065b89-6c22-401c-b9b0-ea49c8742df4

## After

https://github.com/user-attachments/assets/2d6c5651-34c1-4ca0-9113-d89c83de27ac

[WB-34036]: https://coreweave.atlassian.net/browse/WB-34036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/query_builder/objects_query_builder.py
+++ b/weave/trace_server/query_builder/objects_query_builder.py
@@ -336,10 +336,10 @@ class ObjectMetadataQueryBuilder:
         #
         # object_versions_with_index: For each object, number its versions
         # 0, 1, 2, ... ordered by when each digest was first published
-        # (_first_created_at).  Also marks is_latest (1 for the most recently
-        # published non-deleted version; callers always filter with
-        # deleted_at IS NULL, so if all versions are deleted the query returns
-        # no results), and version_count.
+        # (_first_created_at).  is_latest tracks the most recent publish
+        # (by ov.created_at) so re-publishing a prior digest correctly
+        # promotes that digest back to latest; this can yield is_latest=1
+        # on a row whose version_index is not the maximum.
         query = f"""
 WITH latest_row_per_digest AS (
     SELECT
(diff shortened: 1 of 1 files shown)


rgao-coreweave (commented):
Q: If version 2 is deleted, version 1 will become the latest. What does reverted monitor here mean?

```
Error-classifiers(active=true)  -> version 1  # new monitor (latest)
Error-classifiers(active=false) -> version 2  # updated monitor (latest)
Error-classifiers(active=true)  -> version 1  # reverted monitor (NOT latest)
```

This PR breaks stable version index and is unlikely to achieve what you want to do. `set obj is_latest based on created_at` requires an insert a record into db updating created_at, so for it to behave like ^ the request needs to get pass caching (because the records need to land on the tracer server), and the request needs to survive deduping (because clickhouse merges records).

I think https://github.com/wandb/weave/pull/6481 is a better approach as this is a valid use case: re-publishing old content promotes it to "latest" — matching W&B Artifacts behavior.

PR #6766, opened 2026-05-05
chore(weave): Dont use client fixture when not necessary

andrewtruong:
Trivial change that will speed up testing.  We don't always need the client fixture for all tests, which saves setup and teardown time.

### DIFF
--- a/tests/trace/test_weave_client_mutations.py
+++ b/tests/trace/test_weave_client_mutations.py
@@ -3,10 +3,8 @@ from pydantic import Field
 
 import weave
 
-pytestmark = pytest.mark.usefixtures("client")
 
-
-def test_object_mutation_saving():
+def test_object_mutation_saving(client):
     class Thing(weave.Object):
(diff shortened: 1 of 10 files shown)


tests/trace/test_weave_client_mutations.py
```diff
@@ -3,8 +3,10 @@
 
 import weave
 
+pytestmark = pytest.mark.usefixtures("client")
```
chance-wnb: Could you help explain why it is needed? (add a comment). I am curious, because the client fixtures are removed from the test cases, then I wonder why it is still required here. Might not be that obvious for the readers.

chance-wnb (approved):
Thanks!

tests/trace/test_weave_client_mutations.py
andrewtruong: Thanks for catching, this is a mistake!  I added the fixture back

PR #6770, opened 2026-05-05
chore(weave): Fix nightly tests

andrewtruong:
Fixes an issue with the verifiers shard (hopefully)


bcsherma (approved):
approving to unblock

PR #6785, opened 2026-05-08
chore(weave): remove usage from agent classifier prompt

jtschoonhoven:
## Description

Drop token usage from the prompt we use for agent signals.

Token usage isn't typically available on the root span for a turn, and would require fetching and aggregating over all child spans in the turn. We might want to add that in the future, but no one is yet asking for it.

### DIFF
--- a/weave/flow/monitor.py
+++ b/weave/flow/monitor.py
@@ -14,7 +14,7 @@ DebounceAggregationField: TypeAlias = Literal["trace_id", "thread_id"]
 DebounceAggregationMethod: TypeAlias = Literal["last_message", "all_messages"]
 
 # Monitors can be configured to score agent spans by including an AgentSpanOpName in their `op_names` list.
-AgentSpanOpName: TypeAlias = Literal["weave.genai.turn"]
+AgentSpanOpName: TypeAlias = Literal["weave.genai.turn_ended"]
 AGENT_SPAN_OP_NAMES: frozenset[AgentSpanOpName] = frozenset(get_args(AgentSpanOpName))
 
 # Runtime-valid sets derived from Literals above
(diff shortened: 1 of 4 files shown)


weave/flow/monitor.py
```diff
@@ -14,7 +14,7 @@
 DebounceAggregationMethod: TypeAlias = Literal["last_message", "all_messages"]
 
 # Monitors can be configured to score agent spans by including an AgentSpanOpName in their `op_names` list.
-AgentSpanOpName: TypeAlias = Literal["weave.genai.turn"]
+AgentSpanOpName: TypeAlias = Literal["weave.genai.turn_ended"]
```
neutralino1, line 17: Is this a placeholder "fake" op to signal to the worker that this monitor targets an agent turn?
I'm guessing we'll also have to store on the monitor another settings to know what agent to target. We should use this value from the semconv: https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-agent-name

jtschoonhoven: Correct. There's no OTel span that represents a "turn" so this is an internal name that represents a turn op.

And yes, users will want to be able to filter their monitors to specific agents. Monitors currently have two types of filters, `op` and `query`. "Turn ended", "Conversation ended" will be op filters, and agent name will be a query filter.

weave/trace_server/agents/kafka_events.py
```diff
@@ -49,10 +48,10 @@ def from_row(row: AgentSpanCHInsertable) -> ScoreAgentSpansEvent | None:
 
         Currently only "turn_ended" is supported, but the interface is designed to support multiple types.
         """
-        event_type: ScoreAgentSpansEventType | None = None
+        event_type: AgentSpanOpName | None = None
         # A span with no parent is assumed to represent the end of a turn
         if not row.parent_span_id:
-            event_type = "turn_ended"
+            event_type = "weave.genai.turn_ended"
```
mscavezze-cw, line 54: Consider making a constant for this string

mscavezze-cw (approved):

PR #6786, opened 2026-05-08
chore(weave): Add agent trace references to eval results

seanmor5:
This is a draft which adds agent trace references to evaluation calls so they can be linked in the UI. This is strictly an API-surface area change, which introduces a new `agent_trace_ref` to the eval call attributes. I wanted to go as simple and small surface area as possible to avoid committing to anything too extreme during the transition between data models


weave/trace_server/trace_server_interface.py
```diff
@@ -2836,6 +2848,7 @@ class EvalResultsTrial(BaseModel):
     model_latency_seconds: float | None = None
     total_tokens: int | None = None
     scorer_call_ids: dict[str, str] = Field(default_factory=dict)
+    agent_trace_ref: AgentTraceRef | None = None
```
neutralino1: I wonder if we should call this `genai_span_ref` as it is more specific than `agent_trace_ref`. It directly points to a row in the `spans` table. `genai` because these spans follow the OTel GenAI Semantic Conventions.

weave/trace_server/trace_server_interface.py
```diff
@@ -2529,6 +2529,14 @@ class EvaluationRunFinishRes(BaseModel):
     )
 
 
+class AgentTraceRef(BaseModel):
+    trace_id: str
+    span_id: str | None = Field(default=None, exclude_if=_exclude_if_none)
+    conversation_id: str | None = Field(default=None, exclude_if=_exclude_if_none)
+    agent_name: str | None = Field(default=None, exclude_if=_exclude_if_none)
+    agent_version: str | None = Field(default=None, exclude_if=_exclude_if_none)
```
neutralino1: Do we need all this info, not just the `span_id` of the `invoke_agent` span?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -4529,6 +4532,14 @@ def prediction_create(
                 version=predict_and_score_op_res.digest,
             )
 
+            predict_and_score_weave_attrs = {
```
neutralino1, line 4537: Should it be attached to the `predict_and_score` call or the `predict` call? Since that's the call invoking the agent.

neutralino1 (commented):
A few notes.

seanmor5:

weave/trace_server/trace_server_interface.py
seanmor5: I kept because I thought it may be useful for display purposes, but it's not strictly necessary and everything here can be retrieved with just the trace and span ID. I've simplified to only include those

weave/trace_server/trace_server_interface.py
seanmor5: [4404407](https://github.com/wandb/weave/pull/6786/commits/44044073938268560a2dd9cea241bc9789f1eb78)

weave/trace_server/clickhouse_trace_server_batched.py
seanmor5: I attached to `predict_and_score` because the eval results query builders are all based off of `predict_and_score` calls and some of the eval UI paths seem to depend on `predict_and_score` and not just `predict`

neutralino1 (approved):
LGTM but a ✅ from @andrewtruong or @bcsherma would do good.

PR #6790, opened 2026-05-11
fix(weave): Fix potential mem leak in streaming accumulator

andrewtruong:
Previously IteratorWrapper could leak a tiny amount that would build up over time in long running procs


rgao-coreweave (approved):
LGTM

PR #6795, opened 2026-05-11
perf(weave): parallelize bucket uploads in upsert_batch hot path

ro31337:
## JIRA Issue(s)

[WB-34172](https://coreweave.atlassian.net/browse/WB-34172)

## Description

When `/call/upsert_batch` receives a payload with multimodal/inline-base64 content, the trace server fans the content out into one bucket PUT per file. Today's `_file_create_bucket` issues those PUTs one at a time on the request thread; with ninety PUTs in a single batch (a real production trace, captured below), the bucket phase alone takes around 6.5 seconds and pushes whole `upsert_batch` walls past ten seconds.

This change keeps the same code path for one-shot `file_create` (no parallelism gain there) but, when we are inside a `call_batch()`, defers each PUT into a per-thread queue. A new phase 0 of `_flush_all_batches_in_order` drains that queue across a process-wide `ThreadPoolExecutor` before the file-chunks insert runs. The pool has eight workers and a `thread_name_prefix="bucket-put"`, sized to fit comfortably under the default `requests` HTTPAdapter pool of ten.

The key for each PUT is deterministic from `(project_id, digest)`, so we can pre-compute the final `file_storage_uri` before the PUT starts. The chunk record carrying that URI is only committed to `_file_batch` after the PUT actually succeeds. If a single PUT fails we route only that one digest through `_file_create_clickhouse`, mirroring the previous per-call fallback (`try/except FileStorageWriteError`). ddtrace context flows into each worker through a per-submit `contextvars.copy_context()`, so child PUT spans stay attached to the originating trace.

## Why this approach

The team's recent perf fixes in this area (e.g. PRs #6783, #6789, and the prior `core` ports) ship without env flags or canary rollouts — the safety net is strong pre-merge testing rather than a runtime toggle. We keep that shape here. The pool is class-level (not per instance) so many concurrent `call_batch()` handlers on the same pod share one thread cap instead of multiplying threads with each in-flight request.

### This is important — production data, not a thought experiment

DDtrace APM trace `6a02107200000000d228fc15a7d6b7ee` (env:prod, 24h before this change) shows a 9.92s `upsert_batch` whose breakdown is:

| Span | Count | Sum | % of trace |
|---|---|---|---|
| `requests.request` (HTTP PUT to bucket) | 90 | 6.49s | **65%** |
| `_file_create_bucket` | 90 | 6.57s | 66% |
| `process_call_req_to_content` | 5 | 6.78s | 68% |
| `flush_calls` + ClickHouse insert | 2 | 0.77s | 8% |
| Per-PUT latency | — | 94–139ms (avg ~72ms) | — |

Two-thirds of a representative slow `upsert_batch` is sequential bucket PUTs. The HTTPS LB also recorded 7,538 `client_disconnected_before_any_response` events on `/call/upsert_batch` over the same 24h window — the same shape with more PUTs, where the client gave up before the bucket phase finished.

### Real improvement — measured, not projected

Local micro-benchmark using a stub storage client with `time.sleep(72ms)` per PUT (the average per-PUT latency from the trace above). Each row is one call to `file_create` per N, sequential = outside `call_batch()` (old path), parallel = inside `call_batch()` (this PR). Numbers from one run on a laptop, single Python process, eight worker threads:

| N PUTs | Sequential wall | Parallel wall | Speedup |
|--------|----------------|---------------|---------|
| 1 | 77 ms | 78 ms | 1.00x |
| 10 | 774 ms | 155 ms | 5.0x |
| 50 | 3835 ms | 537 ms | 7.1x |
| 100 | 7670 ms | 992 ms | 7.7x |
| 500 | 38343 ms | 4780 ms | 8.0x |

The N=1 row is the regression guardrail: a one-shot upload takes the same wall as before (the synchronous path is preserved for `file_create` outside a batch). For the production-shaped N=90 case, sequential 6.5s drops to roughly 0.8s on the bucket phase, collapsing the 9.9s trace to about 4.3s. The 500-PUT tail (which the LB-disconnect traces look like) drops from ~36s to about 5s.

### Safety — what was checked and why this is hard to regress on

The chunk record's `file_storage_uri` is computed via the same `key_for_project_digest` helper that `store_in_bucket` uses internally, so the pre-computed and post-upload URIs are byte-for-byte identical; there is a unit test pinning that. A second test forces one PUT in a batch to fail and asserts that exactly that digest's chunks land in ClickHouse with `file_storage_uri=None`, while the other digest keeps its bucket URI — the prior per-call fallback semantics are preserved.

The pool is process-wide and lazily initialized behind a lock, so handler concurrency cannot multiply thread counts; the eight-worker cap also sits under the default `requests` connection pool of ten, so we never starve connections. A third test asserts that child spans created inside a worker stay on the same `trace_id` as the originating `upsert_batch` trace, so DDtrace observability does not degrade. ddtrace v4 holds the active span in a `ContextVar`, so `contextvars.copy_context()` is enough to propagate it into the worker; the test pins that wiring.

A fourth test catches a subtle dedup bug: the existing `_file_batch` dedup did not cover the new pending queue, so without an extra check a duplicate `file_create` inside one batch would have re-enqueued. The new code dedups against the queue too and the test asserts only one PUT per digest. A final test asserts the pool defaults (`max_workers=8`, `thread_name_prefix="bucket-put"`) so a future bump cannot quietly invalidate the projected speedup.

## Testing

New test file `tests/trace_server/test_parallel_bucket_flush.py` covers eight scenarios: pool defaults; one-shot synchronous path; parallel flush wall-clock vs sequential; deterministic URI matching the PUT target; per-digest ClickHouse fallback on failure; in-batch dedup; ddtrace trace-id continuity across the worker boundary; and immutability of the queue items. All eight pass locally on SQLite. Existing trace_server tests still pass against SQLite (131 passed, 28 skipped — the skips are ClickHouse-only tests). ClickHouse-backend CI runs via `tests-3.12(shard='trace_server')` will exercise the full insert path.

## Backward compatibility

No API or schema changes. Outside a `call_batch()` the synchronous path is preserved exactly. Inside a batch, behavior is identical except for latency and the order in which PUTs hit the bucket — every successful PUT still lands the same chunk row, and every failing PUT still falls back to ClickHouse for that digest only. Rollback is one or two revert commits; no flags, no migrations.

[WB-34172]: https://coreweave.atlassian.net/browse/WB-34172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace_server/conftest.py
+++ b/tests/trace_server/conftest.py
@@ -176,6 +176,7 @@ def _reset_server_state(server: ClickHouseTraceServer) -> None:
     server._call_batch = []
     server._file_batch = []
     server._calls_complete_batch = []
+    server._pending_bucket_uploads = []
     server._flush_immediately = False
 
 
--- a/tests/trace_server/test_parallel_bucket_flush.py
+++ b/tests/trace_server/test_parallel_bucket_flush.py
@@ -1,0 +1,389 @@
+"""Tests for the phase-0 parallel bucket-upload flush inside ``call_batch()``.
+
+These tests exercise the deferred-PUT path added to ``ClickHouseTraceServer``:
+
+  * ``_file_create_bucket`` queues an upload instead of running ``store_in_bucket``
+    synchronously when we are inside a ``call_batch()`` (``_flush_immediately`` is
+    ``False``).
+  * ``_flush_pending_bucket_uploads`` fans the queued PUTs out across the
+    shared ``ThreadPoolExecutor``, then appends a chunk record per success.
+  * On per-digest failure the fallback re-routes that one digest through the
+    ClickHouse-chunked write path, preserving the prior single-shot semantics.
+
+Tests intentionally avoid the heavyweight ClickHouse fixtures — they patch
+the low-level ``_mint_client`` and ``file_storage_client`` so they run as
+plain unit tests without Docker.
+"""
+
+from __future__ import annotations
+
+import base64
+import threading
+import time
+from unittest.mock import MagicMock, patch
+
+import ddtrace
+import pytest
+
+from weave.shared.digest import compute_file_digest
+from weave.trace_server import clickhouse_trace_server_batched as cts_module
+from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.clickhouse_trace_server_batched import (
+    _BUCKET_UPLOAD_POOL_MAX_WORKERS,
+    ClickHouseTraceServer,
+    _PendingBucketUpload,
+)
+from weave.trace_server.file_storage_uris import GCSFileStorageURI
+
+
+def _make_project_id() -> str:
+    return base64.b64encode(b"test_entity/test_project").decode("utf-8")
+
+
+class _FakeStorageClient:
+    """Minimal FileStorageClient stand-in: records every store call.
+
+    Optionally sleeps to model GCS PUT latency (so parallel-vs-sequential
+    tests are observable). The ``fail_paths`` set forces ``FileStorageWriteError``
+    for any path in it, mirroring the wrapped exception ``store_in_bucket``
+    raises in production.
+    """
+
+    def __init__(
+        self,
+        *,
+        bucket: str = "wandb-weave-trace-prod",
+        per_put_sleep: float = 0.0,
+        fail_paths: set[str] | None = None,
+    ):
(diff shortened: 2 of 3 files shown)


ro31337:
Will get back to it later.

PR #6797, opened 2026-05-11
perf(weave): cut allocation churn in upsert_batch base64 detection

ro31337:
## JIRA Issue(s)

[WB-34182](https://coreweave.atlassian.net/browse/WB-34182)

## Description

`replace_base64_with_content_objects` is on the hot path of every call ingested through `/call/upsert_batch`: the trace server walks the entire inputs / outputs tree of each call to find inline base64 content that should be lifted out into bucket storage. Today that walk does two pieces of work that are unnecessary on the dominant traffic shape, where the tree is large but contains no binary content.

The first piece of unnecessary work is regex-matching long-but-not-binary strings. Every string longer than the auto-conversion threshold goes through `is_data_uri` and `is_base64`, and sometimes through a full `Content.from_base64` decode, even though the entire 1–8 KiB band is essentially never real binary — long LLM outputs, serialized tool-call arguments, and big system prompts all land there. Genuine multimodal blobs (PNG, JPEG, WebP, audio) sit comfortably above 8 KiB once base64-encoded, so raising the threshold from 1 KiB to 8 KiB removes the regex work for the false-positive band while keeping every legitimate binary on the same bucket-storage path it took before.

The second piece is allocation. The inner `_visit` recursion currently allocates a fresh `dict` or `list` on every level of the tree, even when none of the children changed. For a payload that has no binary content at all — which is most production traffic — that is a complete structural copy of the input for no reason. The change keeps `_visit`'s public behaviour identical but only materialises a copy on subtrees that actually had a replacement; everywhere else it returns the original object. The single caller pattern downstream is `req.start.inputs = replace_base64_with_content_objects(...)`, which rebinds the field, so identity-as-aliasing is a no-op for the trace server.

## Evidence

I measured both effects on a synthetic 18.7 MB payload (~8000 nested chat-message dicts, no inline binaries) that matches the shape of the slow tail of `/call/upsert_batch` in production. DDtrace shows env:prod requests with `content_length` between 14 MB and 22 MB on the >2 s tail at roughly 26 req/s overall, so this is a realistic worst-case rather than a microbenchmark fiction. Wall time on this payload drops from ~5.1 ms to ~3.5 ms per `replace_base64_with_content_objects` call (1.44×), and the p95 of 34 ms — almost certainly a major GC collection triggered by the megabytes of throwaway dicts the old code was producing — flattens to ~4 ms.

The memory side is the bigger story. On the same prod-shape payload, peak per-call allocation drops from ~2.7 MB to ~12 KiB (the latter is `tracemalloc` overhead and a MagicMock call recording; the function itself allocates zero new Python objects when nothing changed). Multiplied across `/call/upsert_batch` traffic with multiple `process_call_req_to_content` invocations per batch, that removes hundreds of MB/s of dict-copy churn from one code path per pod. Smaller payloads see proportionally smaller gains, and multimodal payloads with real data URIs are unchanged within noise — the 8 KiB threshold sits well below any real encoded image or audio, so the bucket-storage path still triggers identically.

For reference, the bench script and the before/after tables for typical, medium, multimodal, wide-100, wide-500 and prod-scale shapes are kept locally; the headline numbers are the only ones worth quoting in a PR body and they are above. The numbers above were generated with `tracemalloc.reset_peak` between iterations so each value is the peak of a single call, not a cumulative sum.

## Safety

Code change is small (~12 lines of behaviour, plus comments explaining the two knobs). Five new unit tests pin the new contract:

- the threshold is 8192 (regression guard against silently reverting it),
- a 1–8 KiB string does not invoke `is_data_uri` / `is_base64` / `file_create`,
- a no-replacement input returns the same dict object (identity preserved at every level),
- a partial replacement only copies the affected subtrees,
- the caller pattern `req.start.inputs = replace_base64_with_content_objects(...)` remains safe.

All existing tests in `test_base64_content_conversion.py`, `test_clickhouse_batching.py`, `test_calls_complete.py`, and the broader trace_server subset (54 tests) pass locally on SQLite. ClickHouse-backed CI exercises the full insert path. There is no API or schema change; rollback is a single revert commit.

The only behavioural difference users could in principle notice is that base64-encoded blobs smaller than 8 KiB no longer get auto-lifted into bucket storage — they stay inline in ClickHouse. This is the pre-feature behaviour, handled correctly by the storage path, and the original intent of the auto-conversion was specifically to peel off large media. Small icons inlined in chat tool results are the most plausible affected case; ClickHouse column-store compression handles them fine, just without bucket-level deduplication.

## Testing

Locally on this branch: `uv run --group test python -m pytest tests/trace_server/test_base64_content_conversion.py tests/trace_server/test_clickhouse_batching.py tests/trace_server/test_calls_complete.py tests/trace_server/test_clickhouse_trace_server_batched.py --trace-server=sqlite` — 54 passed, 53 skipped (ClickHouse-only). Ruff format and ruff check are clean. CI will run the full trace_server shard against ClickHouse.

## Backward compatibility

No breaking changes. Behaviour for large data URIs and standalone base64 ≥ 8 KiB is identical. Behaviour for the 1–8 KiB band reverts to pre-auto-conversion (inline storage). No new dependencies, no new flags, no migration.

[WB-34182]: https://coreweave.atlassian.net/browse/WB-34182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## QA tier A/B

To check that the local-bench numbers held up under a real ClickHouse insert path, I ran the same synthetic load against the QA tier in two back-to-back runs with eleven minutes between them. The only thing that changed between runs was which build of the trace server was deployed: the master commit immediately before this PR, then this PR's HEAD. The loader sends `/call/upsert_batch` at one request per second for three minutes, and each request is a five-call batch totalling roughly 19 MB of nested chat-shaped JSON with no inline binary content — the same shape DDtrace shows on the slow tail of this endpoint in env:prod (content_length between 14 and 22 MB for requests over two seconds).

The server-side `process_call_req_to_content` distribution narrowed at every percentile under load. Median dropped from 4 ms to 3 ms, p99 dropped from 8 ms to 4 ms, and the per-window maximum collapsed from 387 ms to 4 ms. That last number is the one worth pausing on. The 387 ms outlier on the baseline build is a major garbage-collection cycle on the request thread, triggered by the megabytes of throwaway dicts the old recursion produced for every large payload; that is exactly the failure mode the allocation cut in this PR is meant to remove, and across a three-minute window of the same load against the patched build it does not appear at all. The local bench predicted a p95 spike in the 30 – 40 ms range from this; in QA with a real ClickHouse insert running alongside, the spike turned out to be roughly an order of magnitude larger before the fix and entirely gone after.

There was also a customer-visible payoff that I had not predicted. The QA HTTPS load balancer returns `502 Temporary backend error` when the upstream trace-server pod does not respond inside the LB's read-timeout — the same signature that shows up in our production LB logs as `client_disconnected_before_any_response`. On the baseline build 37.6 % of the 181 requests timed out at the LB that way; on the patched build the same loader against the same project saw 19.3 %. So on this payload shape, this PR roughly halves the rate of real upstream-timeout failures, not just trimming CPU baseline. That is on top of the percentile narrowing and the GC-spike disappearing. This is one synthetic shape on a non-prod environment, so I would not overweight the absolute milliseconds, but the *shape* of the win (median tightens, max collapses, LB-timeout rate halves) lines up exactly with what the allocation reduction was designed to do.

| metric | master baseline | this PR | delta |
|---|---:|---:|---|
| LB `502 Temporary backend error` rate | 37.6 % | 19.3 % | roughly half as many upstream timeouts |
| `process_call_req_to_content` p50 | 4 ms | 3 ms | -33 % |
| `process_call_req_to_content` p99 | 8 ms | 4 ms | -50 % |
| `process_call_req_to_content` max | **387 ms** | 4 ms | GC-pause outlier removed |

Raw per-request samples and the DDtrace span dumps from both runs are saved locally; happy to share specifics in DM if it is useful.


weave/trace_server/base64_content_conversion.py
```diff
@@ -124,13 +133,29 @@ def replace_base64_with_content_objects(
...
         elif isinstance(val, list):
-            return [_visit(v) for v in val]
+            new_items: list[Any] | None = None
+            for i, v in enumerate(val):
+                new_v = _visit(v)
+                if new_v is not v:
+                    if new_items is None:
+                        new_items = list(val)
+                    new_items[i] = new_v
+            return val if new_items is None else new_items
```
neutralino1: We do this no-copy logic in a couple of places already, i wonder if we could DRY it into some sort of `no_copy_walk` tool. Not a blocker.

ro31337: fixed!

tests/trace_server/test_base64_content_conversion.py
```diff
@@ -353,5 +353,167 @@ def test_wav_file_base64_converted(self):
         assert metadata["mimetype"] in {"audio/wav", "audio/x-wav", "audio/wave"}
 
 
+class TestThresholdAndStructuralIdentity:
```
gtarpenning, line 356: not blocking but _in general_ we don't really use classes in our tests. I personally prefer longer, denser tests with many assertions that demonstrate a functional area rather than lots of disparate little one-offs.

weave/trace_server/base64_content_conversion.py
```diff
@@ -30,8 +30,17 @@
...
+#
+# Lifted from 1 KiB to 8 KiB: under the old threshold every long LLM output,
+# serialized tool-call argument, and large system prompt in the 1-8 KiB range
+# was put through ``is_data_uri`` + ``is_base64`` + (occasionally)
+# ``Content.from_base64`` even though those strings are essentially never
+# real binary payloads. Genuine multimodal blobs (PNG/JPEG/WebP/audio) start
+# comfortably above 8 KiB once base64-encoded; small icons that happen to
+# fall below 8 KiB now stay inline in ClickHouse, which is the pre-feature
+# behaviour and is handled correctly by the existing storage path.
+AUTO_CONVERSION_MIN_SIZE = 8192  # 8 KiB
```
gtarpenning, line 43: there is a security / data governance concern here, where users who rely on bring your own bucket storage want their B64 data stored on their own bucket storage rather than CH. not blocking, but this might not be good enough for some ppl

tests/trace_server/test_base64_content_conversion.py
```diff
@@ -353,5 +353,167 @@ def test_wav_file_base64_converted(self):
...
+            start=StartedCallSchemaForInsert(
+                project_id="proj",
+                op_name="op",
+                started_at=datetime.now(timezone.utc),
+                attributes={},
+                inputs=inputs_before,
+            )
+        )
+        processed = process_call_req_to_content(start_req, trace_server)
+        # Pydantic copies inputs into the model on assignment, but the
```
gtarpenning: technically pydantic copies on construction but for our purposes thats the same

tests/trace_server/test_base64_content_conversion.py
ro31337: Fair point — matched the existing class-namespacing in this file for local consistency, but happy to do a flat-style sweep separately if the team wants to standardise.

weave/trace_server/base64_content_conversion.py
ro31337: Noted — anything under 8 KiB now stays in CH (the pre-feature behaviour for inline content), so BYO-bucket tenants miss those; happy to revisit with a per-project knob if anyone hits it.

tests/trace_server/test_base64_content_conversion.py
ro31337: Good catch — just pushed a one-line fix to that comment so it reads 'copies on construction'.

gtarpenning (approved):

PR #6799, opened 2026-05-11
feat(weave): add eval/genai span linking primitive

seanmor5:
This is meant to build on #6786 with SDK changes that make it easy to link evaluation call results to Gen AI spans. I'm leaving as draft because I've gone back and forth a bunch of times on what the best API surface is for this and I still don't like this implementation.

The current state is such that if you are using the normal evaluation path, you need to track the current Gen AI span and plumb it through to your prediction method, so for example:

```python
    @weave.op
    def predict(question: str) -> str:
        weave.link_genai_span(
            # somebody needs to explicitly track these come from
            ref_trace_id,
            ref_span_id,
         )
        return invoke_agent(question)
```

Notice that this requires the caller to somehow know the current `ref_trace_id` and `ref_span_id` inside of `predict` before they call `invoke_agent`. I imagine that most implementations here for agents essentially have their `predict` as a wrapper for a broader `invoke_agent` implementation, and `invoke_agent` would already have OTel-style instrumentation. You could assume something like:

```python
    @weave.op
    def predict(question: str) -> str:
        result = invoke_agent(question)

        weave.link_genai_span(
            result.trace_id,
            result.span_id,
         )

         return result
```

But that now requires callers to propagate instrumentation metadata back, just for the purposes of connecting an eval call to a full agent span.

I think a more ideal path is probably something like:

```python
    @weave.op
    def predict(question: str) -> str:
        with weave.capture_genai_span():
            return invoke_agent(question)
```

Where `weave.capture_genai_span()` uses whatever OTel-style helpers we want to expose via the SDK and helps link the eval call to the span. But, that's still a bit handwave-y until that's further fleshed out


weave/evaluation/eval.py
```diff
@@ -179,9 +250,13 @@ def model_post_init(self, context: Any, /) -> None:
 
     @op
     async def predict_and_score(self, model: Op | Model, example: dict) -> dict:
-        apply_model_result = await apply_model_async(
-            model, example, self.preprocess_model_input
-        )
+        token = set_current_eval_predict_and_score_call(call_context.get_current_call())
```
andrewtruong: prob should be a context manager

weave/trace_server/eval_results_helpers.py
```diff
@@ -289,16 +289,17 @@ def best_effort_scorer_call_ids(
     return result
 
 
-def extract_genai_span_ref(call: tsi.CallSchema | None) -> tsi.GenAISpanRef | None:
-    """Extract an optional GenAI span ref from call attributes."""
-    if call is None or not isinstance(call.attributes, dict):
+def extract_genai_span_ref_from_weave_namespace(
+    data: Any,
```
andrewtruong: why is this `Any`?

seanmor5: I guess more correctly it's `dict[str, Any] | SummaryMap` since it's always called on `call.attributes` or `call.summary`

seanmor5:
```suggestion
    data: dict[str, Any] | tsi.SummaryMap,
```

weave/evaluation/eval.py
seanmor5: [48307a4](https://github.com/wandb/weave/pull/6799/commits/48307a434059d04f846e85ce794b28265bbd924d)

neutralino1:
If the user is tracing their agents with the SDK, we could probably access some global context and do this linking automatically.
If they are using an OTel collector, it is more tricky, no shared context. Either we do as you propose here, or we do something more complex with an on-disk context (e.g. temp file with trace ID).

PR #6809, opened 2026-05-12
chore(weave): Centralize should_use_accumulator check

andrewtruong:
Small change that shares `should_use_accumulator` between integrations


rgao-coreweave (approved):
LGTM

PR #6814, opened 2026-05-12
chore(weave): Condense settings into one contextvar

andrewtruong:
Previously we had multiple contextvars for settings, but since they are all settings, it's easier to just have one

### DIFF
--- a/weave/trace/settings.py
+++ b/weave/trace/settings.py
@@ -1,21 +1,36 @@
 """Settings for Weave.
 
-## `disabled`
+The settings module exposes a process-level configuration value, scoped per
+async-context via a single :class:`ContextVar`.  Read settings via
+:func:`current` (returns a :class:`SettingsView` with environment-variable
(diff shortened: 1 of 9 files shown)


weave/trace/settings.py
```diff
@@ -276,138 +314,288 @@ class UserSettings(BaseModel):
...
-        if self._is_first_apply:
-            self._is_first_apply = False
-        else:
-            self._reset()
 
-        for name in self.model_fields:
-            context_var = _context_vars[name]
-            context_var.set(getattr(self, name))
+@contextmanager
+def override(**fields: Unpack[_SettingsOverrides]) -> Iterator[UserSettings]:
```
neutralino1: How about `override_settings`? Matches `replace_settings`, makes it more explicit at call sites, and also makes it clear what the args should be.

andrewtruong: Done

neutralino1 (approved):

PR #6820, opened 2026-05-13
fix(weave): use normal WHERE filters for agent project reads

bcsherma:
## Summary
- Revert agent project filters back to normal column-left equality.
- Move agent query builders from explicit `PREWHERE` clauses to regular `WHERE` clauses and let ClickHouse choose any safe automatic PREWHERE rewrite.
- Disable `use_query_condition_cache` for ClickHouse read queries.

## Why
The earlier operand-order workaround was chasing the wrong issue. The local bad results line up with ClickHouse condition-cache corruption:

- `WHERE project_id = p` returned 0 with the cache enabled.
- `WHERE project_id = p SETTINGS use_query_condition_cache = 0` returned the expected rows.
- The failure class is tracked upstream in https://github.com/ClickHouse/ClickHouse/issues/104781.

Explicitly splitting filters into `PREWHERE project_id ... WHERE non_pk IN ...` is one of the known trigger shapes, so the agent builders should not force that split. The global read-query setting is the correctness guard for affected ClickHouse 26.x builds.

## Testing
- `/Users/ben/repos/core/services/weave-python/weave-public/.venv/bin/python -m pytest tests/trace_server/query_builder/test_agent_query_builder.py tests/trace_server/query_builder/test_agent_stats_query_builder.py tests/trace_server/test_clickhouse_trace_server_settings.py`
- `nox -e lint`

### DIFF
--- a/tests/trace_server/query_builder/test_agent_query_builder.py
+++ b/tests/trace_server/query_builder/test_agent_query_builder.py
@@ -72,7 +72,7 @@ class TestMakeSpansCountQuery:
         query = make_spans_count_query(pb, AgentSpansQueryReq(project_id="p1"))
 
         expected = (
-            "SELECT count() FROM spans s PREWHERE {genai_0:String} = s.project_id"
+            "SELECT count() FROM spans s WHERE s.project_id = {genai_0:String}"
         )
         assert_sql(expected, {"genai_0": "p1"}, query, pb.get_params())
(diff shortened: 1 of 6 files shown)


gtarpenning (approved):
lgtm, might want to add a couple TODOs to remind us to revert the change when CH patches the issue

PR #6822, opened 2026-05-13
feat(weave): Link Gen AI spans with Evals Using OTel span processor

seanmor5:
I think this is a better alternative to #6799 because it doesn't require any explicit linking.

For users of the Python SDK, we attach an `EvalLinkSpanProcessor` which is just a span processor that will detect if a span has been emitted from inside a `Evaluation.predict_and_score` call and automatically grab the span and trace ID, so we can attach them to the call.

We also add `weave.eval` attributes to the spans automatically, which should enable us to filter/deep-link to spans which are specific to a given eval.


tests/trace/test_eval_otel_linking.py
```diff
@@ -0,0 +1,172 @@
...
+async def test_genai_span_ref_attached_to_eval_call(client, otel_setup):
+    """A GenAI OTel span emitted during a prediction should produce a
+    GenAISpanRef on the predict_and_score call summary.
+    """
+    exporter = otel_setup
+
+    @weave.op
+    async def model_predict(input) -> str:
+        _emit_genai_span()
+        return str(eval(input))
```
andrewtruong: Why are you calling eval here?

tests/trace/test_eval_otel_linking.py
```diff
@@ -0,0 +1,172 @@
...
+            project_id=client.project_id,
+            evaluation_call_ids=[evaluate_call.id],
+            include_predict_and_score_children=False,
+        )
+    )
+
+    trial = res.rows[0].evaluations[0].trials[0]
+    assert trial.genai_span_ref is not None
+    assert trial.genai_span_ref.trace_id == format(
+        genai_spans[0].context.trace_id, "032x"
```
andrewtruong, line 85: What does `032x` mean?

tests/trace/test_eval_otel_linking.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+    assert trial.genai_span_ref.span_id == format(
+        genai_spans[0].context.span_id, "016x"
```
andrewtruong, line 88: and this

weave/evaluation/eval.py
```diff
@@ -48,6 +53,71 @@
...
+
+    Our EvalLinkSpanProcessor calls this to attach GenAISpanRef to an Eval
+    call. The trace server can than extract span refs from call summaries
+    by looking for the presence of the GenAISpanRef attribute key.
+    """
+    if call.summary is None:
+        call.summary = {}
+
+    weave_summary = call.summary.get(constants.WEAVE_ATTRIBUTES_NAMESPACE)
+    if not isinstance(weave_summary, dict):
```
andrewtruong: can this ever be true?

weave/evaluation/eval.py
```diff
@@ -48,6 +53,71 @@
...
+        weave_summary = {}
+        call.summary[constants.WEAVE_ATTRIBUTES_NAMESPACE] = weave_summary
+
+    weave_summary[constants.GENAI_SPAN_REF_ATTR_KEY] = genai_span_ref.model_dump()
+
+
+def _find_current_predict_and_score_call() -> Call | None:
```
andrewtruong: The presence of this is unfortunate, but I see why you're doing.  I wonder if there's a better way?

weave/evaluation/otel_eval_linker.py
```diff
@@ -0,0 +1,125 @@
...
+if TYPE_CHECKING:
+    from opentelemetry.context import Context
+    from opentelemetry.sdk.trace import ReadableSpan, Span
+
+    from weave.trace.call import Call
+
+logger = logging.getLogger(__name__)
+
+# GenAI semantic convention attribute that identifies a span as a GenAI operation.
+_GENAI_OPERATION_NAME_ATTR = "gen_ai.operation.name"
```
andrewtruong, line 43: Can you just get this directly from the OTEL library?

weave/evaluation/otel_eval_linker.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _get_evaluation_name(predict_and_score_call: Call) -> str | None:
```
andrewtruong: I think this is repeated from the code before. You may want to reuse implementations.

weave/evaluation/otel_eval_linker.py
```diff
@@ -0,0 +1,125 @@
...
+        during a predict_and_score context, not just GenAI spans. This is because
+        gen_ai.operation.name is typically set after span creation and is therefore
+        not available at start time.  The query side can intersect with gen_ai.operation.name
+        to narrow to GenAI spans.
+        """
+        call = self._find_predict_and_score_call()
+        if call is None:
+            return
+
+        span.set_attribute("weave.eval.predict_and_score_call_id", call.id)
```
andrewtruong: Please pull these out into consts.

weave/trace/weave_init.py
```diff
@@ -168,6 +170,10 @@ def _setup_session_tracing(entity: str, project: str, api_key: str | None) -> No
         exporter._certificate_file = False
     provider = TracerProvider(resource=resource)
     provider.add_span_processor(BatchSpanProcessor(exporter))
+    # Auto-link GenAI OTel spans to eval predictions and inject eval
+    # metadata (call ID, project, evaluation name) onto spans for
+    # deep-linking in the agent traces UI.
+    provider.add_span_processor(EvalLinkSpanProcessor())  # type: ignore[arg-type]
```
andrewtruong: Why is there a type ignore here?

andrewtruong (approved):
Just a few nits

tests/trace/test_eval_otel_linking.py
```diff
@@ -0,0 +1,172 @@
...
+
+    OTel only allows set_tracer_provider once per process, so we add our
+    processors to whatever provider is already active and clear the exporter
+    between tests.
+    """
+    exporter = InMemorySpanExporter()
+    span_processor = SimpleSpanProcessor(exporter)
+    eval_processor = EvalLinkSpanProcessor()
+
+    provider = otel_trace.get_tracer_provider()
```
andrewtruong: Do you need a teardown for this?  Wondering if it will leak state between tests

weave/evaluation/eval.py
```diff
@@ -48,6 +53,71 @@
...
+    """
+    if call.summary is None:
+        call.summary = {}
+
+    weave_summary = call.summary.get(constants.WEAVE_ATTRIBUTES_NAMESPACE)
+    if not isinstance(weave_summary, dict):
+        weave_summary = {}
+        call.summary[constants.WEAVE_ATTRIBUTES_NAMESPACE] = weave_summary
+
+    weave_summary[constants.GENAI_SPAN_REF_ATTR_KEY] = genai_span_ref.model_dump()
```
andrewtruong, line 103: This implies you can only link one span, right?  Is there ever a case where they may be multiple spans?

seanmor5: This is actually problematic, I made a more complex eval use case and this tries to attach to every span inside the full trace, instead of just the root span which it should be linked to, which causes issues on the rendering side. I will take a closer look at how to best address this

tests/trace/test_eval_otel_linking.py
seanmor5: Added some comments, but basically this formats the trace IDs (which are integers) to padded hexadecimal strings as recommended here: https://www.w3.org/TR/trace-context/#trace-id

tests/trace/test_eval_otel_linking.py
seanmor5: Same as above

weave/evaluation/eval.py
seanmor5: [150cc3c](https://github.com/wandb/weave/pull/6822/commits/150cc3cf60c476d77ba3b17e4e1d3d9c69205646)

tests/trace/test_eval_otel_linking.py
seanmor5: I removed, it's unnecessary for the tests, the cases were all "1 + 1" so the model predict was just getting the output by evaling that but it doesn't matter for these tests. Claude-ism

weave/evaluation/otel_eval_linker.py
seanmor5: [de95b81](https://github.com/wandb/weave/pull/6822/commits/de95b817a97a5f0b580eb0fa140f50932732ecbb)

weave/trace/weave_init.py
seanmor5: [6d074e7](https://github.com/wandb/weave/pull/6822/commits/6d074e751071fbae18906f3878f78cff29679fd6)

weave/evaluation/otel_eval_linker.py
seanmor5: [cccf133](https://github.com/wandb/weave/pull/6822/commits/cccf133179ad3acd4ecb2bcf50ecc078d8f1e4e8)

weave/evaluation/otel_eval_linker.py
seanmor5: [7135183](https://github.com/wandb/weave/pull/6822/commits/71351838b034fb320bb62ffb66b401a21dbc059b)

tests/trace/test_eval_otel_linking.py
seanmor5: So this causes the global provider to accumulate span processors, and will add these span processors to any test that also tries to use OTel. I thought maybe a session-based fixture would work, but that still doesn't solve the teardown problem for tests outside of this module.

Apparently, the `set_tracer_provider` has a `do_once` guard which prevents us from setting/unsetting the tracer provider "normally." I saw some hacks in GitHub issues which use `importlib` to reload opentelemetry: https://github.com/open-telemetry/opentelemetry-specification/issues/3302

The suggestion I went with here instead is to use `pytest.monkeypatch` which monkey patches the internal tracer provider attribute and will automatically unset it after the test runs. So this should not impact tests running in this module, or other tests outside of this module

weave/evaluation/eval.py
seanmor5: I'm going to address this in a follow-on PR as it touches some trace-server logic

PR #6836, opened 2026-05-13
chore(weave): Update settings call sites

andrewtruong:
Updates remaining settings call sites to match new pattern


rgao-coreweave (approved):
LGTM

PR #6838, opened 2026-05-13
chore(weave): Mark weave tests that depend on weave side effects and not the client

andrewtruong:
Some tests don't need the weave client specifically, even though they use the `client` fixture.  They instead depend on some side effect of client initialization.  This PR makes it clear what those tests are so we can we better segment these tests in future


rgao-coreweave (approved):
Very good change, thanks @andrewtruong.

PR #6841, opened 2026-05-14
feat(weave): Preliminary Google ADK integration

andrewtruong:
Initial OTEL-based integration for Google ADK that supports auto-tracking agent spans.
image

### DIFF
--- a/weave/integrations/google_adk/extractors.py
+++ b/weave/integrations/google_adk/extractors.py
@@ -1,0 +1,296 @@
+"""ADK → GenAI semconv attribute extractors.
+
+Given a parsed ADK object (``LlmRequest`` / ``LlmResponse`` / agent / tool),
+emit the GenAI semantic-convention attributes Weave extracts into dedicated
+columns.
+
+These functions take an OTel ``Span`` and write attributes directly to it.
+They never raise into ADK's flow — if a value isn't present, the attribute
+just isn't set, which is the correct behaviour for optional GenAI fields.
+
+The per-content-part wire shape (``{"type": "text", "content": ...}``,
+``{"type": "tool_call", ...}``, etc.) comes from ADK's own ``_to_part``
+helper — same TypedDicts ADK uses on its native spans — with one local
+override that strips the ``Blob`` binary payload because bytes aren't
+span-attribute-safe and Weave already stores blobs out of band.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+from typing import Any
+
+from google.adk.models.llm_request import LlmRequest
+from google.adk.models.llm_response import LlmResponse
+
+# ADK's experimental semconv module exports the tool-definition
+# discriminator, the per-part ``_to_part`` builder, and
+# ``_to_system_instructions``, which normalises ``config.system_instruction``
+# — a polymorphic field that ADK accepts as a bare ``str``, a
+# ``google.genai.types.Content``, a list of either, a dict that quacks like
+# ``Content`` / ``Part``, or even a PIL image — into the parts-model wire
+# shape Weave's column extractor reads. Reusing ADK's helpers means we
+# stay in lock-step with google-genai's content-normalisation logic
+# (``_transformers.t_contents``) rather than duplicating its many
+# branches here.
+from google.adk.telemetry._experimental_semconv import (
+    FUNCTION_TOOL_DEFINITION_TYPE,
+    Part,
+    _to_part,
+    _to_system_instructions,
+)
+from google.genai.types import Content, GenerateContentConfig
+from opentelemetry.semconv._incubating.attributes.gen_ai_attributes import (
+    GEN_AI_INPUT_MESSAGES,
+    GEN_AI_OUTPUT_MESSAGES,
+    GEN_AI_OUTPUT_TYPE,
+    GEN_AI_REQUEST_CHOICE_COUNT,
+    GEN_AI_REQUEST_FREQUENCY_PENALTY,
+    GEN_AI_REQUEST_MAX_TOKENS,
+    GEN_AI_REQUEST_PRESENCE_PENALTY,
+    GEN_AI_REQUEST_SEED,
+    GEN_AI_REQUEST_STOP_SEQUENCES,
+    GEN_AI_REQUEST_TEMPERATURE,
+    GEN_AI_REQUEST_TOP_P,
+    GEN_AI_RESPONSE_ID,
+    GEN_AI_RESPONSE_MODEL,
+    GEN_AI_SYSTEM_INSTRUCTIONS,
+    GEN_AI_TOOL_DEFINITIONS,
+    GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS,
+    GEN_AI_USAGE_REASONING_OUTPUT_TOKENS,
+    GenAiOutputTypeValues,
+    GenAiSystemValues,
+)
+from opentelemetry.trace import Span
+
+# Runtime config — env vars the integration reads on every span. Both the
+# extractors below and the patcher wrappers in ``google_adk_sdk`` consult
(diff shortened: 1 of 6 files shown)


weave/integrations/patch.py
```diff
@@ -443,18 +477,20 @@ def __getattr__(self, name):  # type: ignore
 
 def _patch_if_needed(module_name: str) -> None:
     """Apply patching for a module if it hasn't been patched yet."""
-    if (
-        module_name not in _PATCHED_INTEGRATIONS
-        and module_name in INTEGRATION_MODULE_MAPPING
-    ):
+    if module_name not in INTEGRATION_MODULE_MAPPING:
+        return
+    with _PATCH_LOCK:
```
francesca-wandb: This looks like the source of the import-time issue we saw with ADK auto-patching.

`_patch_if_needed("google.adk")` acquires `_PATCH_LOCK`, then calls `patch_google_adk()`. That function goes through `_patch_integration(...)`, which also tries to acquire `_PATCH_LOCK`.

Because `_PATCH_LOCK` is a regular `threading.Lock`, the same thread ends up waiting on a lock it already holds, so the ADK import never finishes.

In the demo we worked around this by avoiding the implicit auto-patching path.

Possible fixes:
- use `threading.RLock`, or
- release `_PATCH_LOCK` before calling `patch_func()`, or
- have `_patch_if_needed` call the lower-level patcher directly without re-entering `_patch_integration`.

weave/integrations/google_adk/extractors.py
```diff
@@ -0,0 +1,239 @@
...
+    if usage.thoughts_token_count is not None:
+        span.set_attribute(
+            GEN_AI_USAGE_REASONING_OUTPUT_TOKENS, int(usage.thoughts_token_count)
+        )
+    if usage.cached_content_token_count is not None:
+        span.set_attribute(
+            GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS,
+            int(usage.cached_content_token_count),
+        )
+    if usage.cache_creation_token_count is not None:
```
francesca-wandb: This raised in the live ADK demo with current `google-genai`:

`AttributeError: 'GenerateContentResponseUsageMetadata' object has no attribute 'cache_creation_token_count'`

Workaround we used while testing: treat this as optional and default to `None`.

Suggested change:

```python
cache_creation_token_count = getattr(usage, "cache_creation_token_count", None)
if cache_creation_token_count is not None:
    span.set_attribute(
        GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS,
        int(cache_creation_token_count),
    )
```

Might be worth using getattr(..., None) for any usage metadata fields that are not guaranteed across google-genai versions.

weave/integrations/google_adk/google_adk_sdk.py
```diff
@@ -0,0 +1,295 @@
...
+    ``.span``.
+    """
+    if _AdkGenerateContentSpan is not None and isinstance(
+        span, _AdkGenerateContentSpan
+    ):
+        return span.span
+    return span
+
+
+def _wrap_trace_agent_invocation(original: Any) -> Any:
```
francesca-wandb: This integration gives us the ADK-native `invocation` trace, but Weave Threads do not render usable turns/messages from the pure ADK trace yet.

Local demo workaround:

- mark the ADK `invocation` span as the Weave Thread turn
- attach message fields to that same span:
  - `input.value`
  - `input.mime_type`
  - `output.value`
  - `output.mime_type`

Without this, Threads either do not group correctly or show empty turns.

Ideal behaviour for this PR: `weave.init(...)` + `runner.run_async(...)` should produce both:
1. an ADK trace with internals, and
2. a usable Weave Thread row with the user/assistant messages.

I think the integration should map ADK’s top-level `invocation` span into the Weave Thread turn and ensure the final user/assistant message content is available on that span.

weave/integrations/patch.py
```diff
@@ -334,6 +373,7 @@ def patch_openai_realtime(settings: IntegrationSettings | None = None) -> None:
     "google.generativeai": patch_google_genai,
     "google.genai": patch_google_genai,
     "vertexai": patch_vertexai,
+    "google.adk": patch_google_adk,
```
francesca-wandb: Tested this latest branch in our `weave_adk_trace_demo`. The two non-thread issues I flagged earlier look fixed:

1. ADK import-time hang: fixed by using `threading.RLock` for `_PATCH_LOCK`. I also verified `_patch_if_needed("google.adk")` completes successfully.

2. `google-genai` usage metadata: fixed because the ADK integration no longer reads `usage.cache_creation_token_count`, which is absent in our current google-genai version.

One remaining setup question for the ADK integration path:

In our demo, we still explicitly call:

```
from weave.integrations.patch import patch_google_adk
patch_google_adk()
```

after:

`weave.init(..., settings={"implicitly_patch_integrations": False})`

This is intentional. We want the ADK integration enabled, but we do not want broad implicit patching to also patch lower-level `google.genai` calls. When `google.genai` is patched underneath ADK, we see duplicate standalone `google.genai.models.AsyncModels.generate_content` root traces next to the ADK OTEL trace tree, which makes the trace view noisy and can make agent granularity look lost.

Should the next iteration document this as the recommended ADK setup, or should implicit patching detect the ADK case and avoid creating duplicate lower-level `google.genai` root traces automatically?

This applies to Vertex AI ADK usage too: the goal is still to keep ADK as the source of truth for the agent trace tree, and avoid lower-level Google/Vertex model SDK auto-patching creating separate root traces outside the ADK hierarchy.

weave/integrations/google_adk/google_adk_sdk.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+def get_google_adk_patcher(
+    settings: IntegrationSettings | None = None,
+) -> MultiPatcher | NoOpPatcher:
+    """Return the patcher that enriches ADK OTel spans with the OTLv2 superset.
+
+    The patcher hooks ADK's tracing functions in-place; the spans created
+    by ADK still go through the global OTel TracerProvider so they ride
+    whichever BatchSpanProcessor was configured by ``weave.init()`` (or
```
francesca-wandb, line 233: Question from testing the ADK demo: since ADK spans ride whichever `BatchSpanProcessor` is configured by `weave.init()` or the user, should the ADK integration support classic Weave Traces as a fallback too?

In our demo, the Agents/Conversations path alone was not visible/reliable for us yet. Classic Traces visibility only worked after adding a second exporter to `/otel/v1/traces` with `project_id` and `wandb-api-key` headers.

Our local `tracing.py` is now just a compatibility shim for that fallback, not a replacement for this ADK patcher. But if classic Traces are expected to work for ADK, it feels like that exporter/header behavior should live in the integration rather than in customer code.

weave/integrations/patch.py
chance-wnb: Hey Andrew, I believe the current implementation still implicitly auto-patches. Do you want to take it out?

chance-wnb (commented):
LGTM. I would like to see how you will respond to removing the auto-patching. Happy to re-review anytime soon after seeing your decision.

andrewtruong:
@chance-wnb what do you mean by removing autopatching?  The intent is to autopatch unless the user opts out

weave/integrations/patch.py
andrewtruong: @chance-wnb oh I see, I think this is what you were referring to.
1. The RLock issue is fixed in another PR: https://github.com/wandb/weave/pull/6923;
2. I can add a check to not patch genai if the ADK OTEL patching is enabled.  This is a bit tricky because the `genai` patching is actually the old tracing patch

chance-wnb (approved):
Looks like you are aware of the patching bug and have plans on avoiding double patching this and Genai. 

Approving. Thanks for the change!

PR #6846, opened 2026-05-14
feat(weave): New SDK calls respect global settings

rgao-coreweave:
## Summary

- Wire four Weave global settings into the Session SDK: `WEAVE_DISABLED`, `WEAVE_REDACT_PII`, `WEAVE_CAPTURE_CLIENT_INFO`, `WEAVE_CAPTURE_SYSTEM_INFO`. Brings the new Session SDK (`weave.start_session` / `start_turn` / `start_tool` / `start_llm` / `start_subagent` + batch `log_turn` / `log_session`) to feature parity with the `@op` tracer.
- Add a new `weave/session/_redaction.py` helper module that routes through the existing `weave.utils.pii_redaction.redact_pii` so redaction output is byte-identical to `@op` for the same input. Helpers: `redact_string`, `redact_messages`, `redact_system_instructions`.
- Fix a pre-existing leak: `LLM.reasoning` was passed to `llm_attributes` regardless of `include_content`, leaking reasoning text into `gen_ai.output.messages` even when sessions opted out of content capture. Now gated correctly in both streaming (`LLM.end`) and batch (`_attrs_for_span`) paths.

## Behavior changes

1. **`WEAVE_DISABLED=true` now also silences Session SDK telemetry.** Short-circuits at `_setup_session_tracing` (init), `_SpanBase._start_otel_span` (streaming), `_emit_span_now` (batch), and the top of `log_turn` / `log_session`. Customers who had `WEAVE_DISABLED=true` but were still seeing Session spans will see them stop.

2. **`WEAVE_REDACT_PII=true` now also redacts Session SDK content.** Tool args/result, LLM input/output messages, system instructions, Turn messages, and LLM reasoning content are all redacted via the existing Presidio-backed `redact_pii`. Skips Presidio entirely when `include_content=False` (no NLP cost on already-dropped content). Out of v1 redaction scope: agent metadata, conversation IDs, model names, tool descriptions, request_*, and binary media — these are configuration identifiers or request metadata per the OTel GenAI semconv, not user content.

3. **Session spans now include `weave.*` metadata by default.** Adds `weave.client_version`, `weave.source`, `weave.sys_version`, `weave.os_name`, `weave.os_version`, `weave.os_release` to every Session span. Matches `@op` defaults; opt out via `WEAVE_CAPTURE_CLIENT_INFO=false` / `WEAVE_CAPTURE_SYSTEM_INFO=false`.

4. **Bug fix: reasoning content no longer leaks when `include_content=False`.** Previously `LLM.reasoning.content` would still appear in `gen_ai.output.messages` because `reasoning=self.reasoning` was passed ungated to `llm_attributes`. Now gated like every other content field. Regression tests cover both streaming and batch paths.

## Design

Single chokepoint inside `weave/session/session.py`. The pure attribute builders in `weave/session/session_otel.py` stay decoupled from settings — they remain importable and testable without weave init / settings setup.

Each span class (`Tool`, `LLM`, `Turn`) exposes a `_content_fields(*, include_content, redact)` method that returns its content-bearing fields as a dict shaped for spreading into the matching OTel attribute builder kwargs. The streaming (`.end()`) and batch (`_attrs_for_span` / `log_turn`) paths share that single definition, so future content fields are added in one place instead of six.

`_capture_info_attrs()` is a per-span helper (not OTel `Resource`) so env-var toggles take effect on every emit, matching `@op` semantics in `weave_client.py:805-812`.

## Settings classification (per OTel GenAI semconv)

When `include_content=False`:
- **Stripped (content):** `gen_ai.input.messages`, `gen_ai.output.messages`, `gen_ai.system_instructions`, `gen_ai.tool.call.arguments`, `gen_ai.tool.call.result`, reasoning content.
- **Kept (request/response metadata):** `gen_ai.request.*` (temperature, max_tokens, top_p, stop_sequences, etc.), `gen_ai.response.*` (id, model, finish_reasons), `gen_ai.usage.*`, agent/tool/conversation names and IDs.

Matches the [OTel GenAI Spans spec](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/) which gates only `input.messages` / `output.messages` / `system_instructions` as sensitive content; request params are recommended metadata with no content-gating restrictions.

## Test plan

- [x] 31 tests in `tests/session/test_session_settings.py` (26 named, 3 parametrized cases covering invariants across span types) — pass
- [x] Full session suite (234 tests) passes locally
- [x] `uvx ruff check` and `uvx ruff format --check` clean on all modified files
- [x] `nox --no-install -e "tests-3.12(shard='trace')" -- tests/session/ --trace-server=sqlite` passes (1 pre-existing unrelated flake in `test_flask_server`)
- [x] Verify in staging: `WEAVE_DISABLED=true` produces no spans for a Session SDK exercise
- [x] Verify in staging: `WEAVE_REDACT_PII=true` redacts a test email in tool args and LLM messages
- [x] Verify in staging: `WEAVE_CAPTURE_CLIENT_INFO=false` removes `weave.*` attrs from Session spans


rgao-coreweave:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/wandb/weave/6846?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6846** <a href="https://app.graphite.com/github/pr/wandb/weave/6846?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6846?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6858** <a href="https://app.graphite.com/github/pr/wandb/weave/6858?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/integrations/openai_realtime/state_exporter.py
```diff
@@ -843,11 +843,8 @@ def _schedule_fifo_check(self) -> None:
                     pass
 
             def _cb() -> None:
-                try:
-                    self._advance_fifo()
-                finally:
-                    # If more remain and head not ready, a new timer will be scheduled
-                    pass
+                # If more remain and head not ready, a new timer will be scheduled
+                self._advance_fifo()
```
andrewtruong: Why do you even need this?  Can't you just defer `self._advance_fifo`?

weave/session/_redaction.py
```diff
@@ -0,0 +1,86 @@
...
+# import error will surface clearly.
+try:
+    from weave.utils.pii_redaction import redact_pii
+except ImportError:  # pragma: no cover — exercised in environments without presidio
+
+    def redact_pii(data: dict[str, Any] | str) -> dict[str, Any] | str:  # type: ignore[no-redef]
+        raise ImportError(
+            "presidio is required for PII redaction. "
+            "Install with `pip install 'weave[presidio]'`."
+        )
```
andrewtruong: Why is this defined here?  It should be defined in `pii_redaction`

andrewtruong: I feel like this whole file probably should be moved into there actually

weave/session/session.py
```diff
@@ -107,6 +117,53 @@
...
+    reasoning: Reasoning | None
+
+
+class _TurnContentFields(TypedDict):
+    """Turn content fields, shaped for ``invoke_agent_attributes(**...)``."""
+
+    input_messages: list[Message] | None
+
+
+def _capture_info_attrs() -> dict[str, Any]:
```
andrewtruong, line 116: Don't we already have this somewhere?  Why is this redefined?  Probably should share the code here

weave/trace/weave_init.py
```diff
@@ -117,6 +118,8 @@ def _setup_session_tracing(entity: str, project: str, api_key: str | None) -> No
     returns early if opentelemetry is unavailable. Other errors propagate
     so misconfiguration is visible to the user.
     """
+    if should_disable_weave():
```
andrewtruong: why is this up here?

weave/integrations/openai_realtime/state_exporter.py
rgao-coreweave: This change was carved into downstack PR #6858 which was subsequently closed — `state_exporter.py` on this branch now matches master.

weave/session/session.py
rgao-coreweave: Grep-checked — these aren't redefined anywhere; they were new in this PR (commit `cdea1249cc`), added to typecheck the `**self._content_fields(...)` spread against the builder kwargs. You're right that they subset-mirror the builder signatures and that's a drift surface. The cdea1249cc commit itself noted "Same dict shape at runtime; behavior unchanged" — they caught no concrete bugs, just provided a typecheck at the spread site.

Reverted to `dict[str, Any]` in 88e54c526a; the type-safety win wasn't worth the duplication.

weave/session/_redaction.py
rgao-coreweave: Moved to `pii_redaction`

weave/trace/weave_init.py
rgao-coreweave: You're right — it was dead code. `weave.init()` in `api.py:167` already short-circuits to `init_weave_disabled()` when `should_disable_weave()` is true, so `init_weave` (and `_setup_session_tracing` along with it) is never reached when disabled. Dropped the guard in 3386c55363 and rewrote the regression test to assert the real invariant (`init_weave_disabled` doesn't call `_setup_session_tracing`) instead of the redundant inner check.

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,584 @@
+"""Tests that the Session SDK respects weave global settings.
+
+See spec: rgao/superpowers/specs/2026-05-13-session-sdk-respects-global-settings-design.md
```
andrewtruong: what is this?

weave/session/session.py
andrewtruong: Are you sure?  These must be captured in weave.op somewhere

weave/utils/pii_redaction.py
```diff
@@ -71,6 +103,9 @@ def redact_recursive(value: Any) -> Any:
 
 
 def redact_pii_string(data: str) -> str:
+    if not _HAS_PRESIDIO:
```
andrewtruong: What is the point of the guard if you are going to just ImportError here anyways?

tests/session/test_session_settings.py
rgao-coreweave: Stale local artifact reference — that path was in my private specs directory, not shared. Removed in cd3ae54b68.

weave/session/session.py
rgao-coreweave: You're right — I misread your original comment. The duplication was the gating + values in `weave_client.py:805-812`. Extracted to `weave/utils/capture_info.py::get_capture_info_items()` which returns `[(key, value)]` pairs with the setting gating in one place. Session SDK prefixes them with `weave.`; `weave_client.py` feeds them through `_set_weave_item()`. cd3ae54b68

weave/utils/pii_redaction.py
rgao-coreweave: The point is deferring the failure from import time to call time, so (1) tests can `mock.patch("weave.utils.pii_redaction.redact_pii")` without installing presidio + spacy + transformers, and (2) users who never set `WEAVE_REDACT_PII=true` never have to install the NLP stack. Expanded the inline comment near the guard in cd3ae54b68 so the reasoning isn't only in the module docstring.

rgao-coreweave: You were right — I checked `weave/scorers/presidio_guardrail.py`, which uses lazy imports inside the function plus `TYPE_CHECKING` for hints, no flag. That gives the same "import the module without presidio installed" property with less code. Refactored in [02f57e71c9](https://github.com/wandb/weave/commit/02f57e71c9): dropped `_HAS_PRESIDIO`, moved presidio imports into a `_get_engines()` helper that raises with the install hint if presidio is missing. My earlier defense of the flag was wrong.

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,581 @@
...
+    log_turn,
+    start_session,
+)
+from weave.trace.settings import override_settings
+from weave.trace.weave_init import init_weave_disabled
+from weave.utils import pii_redaction
+
+
+@pytest.fixture(autouse=True)
+def _reset_contextvars():
```
andrewtruong: I think there is a fixture for this already.  Shouldn't that fixture be autoused?

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,581 @@
...
+    if (llm := get_current_llm()) is not None:
+        llm.end()
+    if (turn := get_current_turn()) is not None:
+        turn.end()
+    if (session := get_current_session()) is not None:
+        session.end()
+
+
+@pytest.fixture
+def otel_spans(monkeypatch: pytest.MonkeyPatch):
```
andrewtruong: Can we try to reuse fixtures?  Surely this already exists in the other OTEL code?

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,581 @@
...
+                    pass
+                with t.tool(name="search"):
+                    pass
+    assert otel_spans.get_finished_spans() == ()
+
+
+@pytest.mark.parametrize(
+    "batch_call",
+    [
+        pytest.param(
```
andrewtruong: Why use funcs in one place but lambda here?

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,581 @@
...
+                turns=[
+                    Turn(agent_name="a", messages=[Message(role="user", content="hi")])
+                ],
+                session_id="sess",
+            ),
+            id="log_session",
+        ),
+    ],
+)
+def test_disabled_batch_emits_no_spans(
```
andrewtruong, line 139: what is a batch call supposed to be?

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,581 @@
...
+    batch_call: Callable[[], Any], otel_spans: InMemorySpanExporter
+):
+    with override_settings(disabled=True):
+        result = batch_call()
+    assert result.span_count == 0
+    assert otel_spans.get_finished_spans() == ()
+
+
+def test_disabled_init_skips_session_tracing():
```
andrewtruong: what does this have to do with these tests?

weave/utils/pii_redaction.py
```diff
@@ -71,14 +105,67 @@ def redact_recursive(value: Any) -> Any:
...
-    analyzer = AnalyzerEngine()
-    anonymizer = AnonymizerEngine()
+    analyzer, anonymizer = _get_engines()
     entities = _get_redaction_entities()
     results = analyzer.analyze(text=data, language="en", entities=entities)
     redacted = anonymizer.anonymize(text=data, analyzer_results=results)
     return redacted.text
 
 
+def redact_string(s: str) -> str:
```
andrewtruong: What is the point of this?  Why not just use `redact_pii`?

tests/session/test_session_settings.py
rgao-coreweave: You're right — same fixture lived in `test_session.py:30`, `test_session_otel.py:56`, and here. Moved both `_reset_contextvars` (autouse) and `otel_spans` into the new `tests/session/conftest.py`; dropped from all three test files. `59c83f1caa`

tests/session/test_session_settings.py
rgao-coreweave: Same fix as the `_reset_contextvars` thread above — moved into the new `tests/session/conftest.py` so `test_session_otel.py` and `test_session_settings.py` share one definition. `59c83f1caa`

tests/session/test_session_settings.py
rgao-coreweave: Fair — everything else in this file uses named `_exercise_*` helpers. Extracted `_exercise_log_turn_minimal` / `_exercise_log_session_minimal` so the parametrize entries match the convention. `431ff25b94`

tests/session/test_session_settings.py
rgao-coreweave: Renamed `batch_call` → `exercise` to match the variable used in `test_redact_pii_applied` and the other parametrize blocks. The two cases exercise the batch entry points (`log_turn`, `log_session`) under `disabled=True`. `431ff25b94`

tests/session/test_session_settings.py
rgao-coreweave: It's the init-time half of the same disabled-tracing invariant — the parametrized tests above verify that runtime span emission is suppressed under `disabled=True`; this one verifies that `init_weave_disabled()` (the path `weave.init()` takes under `WEAVE_DISABLED=true`) doesn't install a session TracerProvider at all. Without it, a regression that re-introduced the inner guard removed in `116289e1cd` would pass the runtime tests but still install a provider against the disabled stub server.

Renamed to `test_disabled_init_does_not_install_session_tracer_provider` and rewrote the docstring to spell that out. `88a816fefc`

weave/utils/pii_redaction.py
rgao-coreweave: You're right — the wrapper was doing two things, an empty-skip plus a `cast(str, redact_pii(s))` to escape the `dict|str` return type. The empty-skip belongs on `redact_pii_string` itself (Presidio doesn't need to analyze `""`), and `redact_pii_string` already has the clean `(str) -> str` signature callers want — no cast needed.

Moved the empty-skip into `redact_pii_string`, deleted `redact_string`, and updated the four call sites (three in `session.py`: `Tool.arguments`, `Tool.result`, `LLM.reasoning.content`; one internal: `redact_system_instructions`). `0057057e79`

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,592 @@
...
+    if expects_engines_loaded:
+        assert result == "REDACTED"
+        assert mock_get_engines.called
+    else:
+        assert result == ""
+        assert not mock_get_engines.called
+
+
+@pytest.mark.parametrize(
+    ("fn", "empty"),
```
andrewtruong, line 208: what is the point of the empty param?

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,592 @@
...
+        ),
+    ],
+)
+def test_redact_collection_helpers_passthrough_none_and_empty(fn, empty):
+    """Collection helpers passthrough None → None and empty → empty."""
+    assert fn(None) is None
+    assert fn(empty) == empty
+
+
+def test_redact_messages_dump_redact_revalidate():
```
andrewtruong, line 222: I dont understand this

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,592 @@
...
+    assert out is not None
+    assert len(out) == 2
+    assert out[0].content == "EMAIL ME AT ALICE@EXAMPLE.COM"
+    assert out[0].role == "user"  # role is a Literal — preserved by validation
+    assert out[1].parts[0].content == "OK"
+    # ToolCallPart.arguments goes through JSONString coercion
+    assert "ALICE" in out[1].parts[1].arguments
+
+
+def test_redact_system_instructions():
```
andrewtruong, line 257: wasnt this already tested above?

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,592 @@
...
+        out = pii_redaction.redact_system_instructions(["secret instruction"])
+    assert out == ["<REDACTED>"]
+
+
+# ---------------------------------------------------------------------------
+# redact_pii — applied across streaming + batch paths
+# ---------------------------------------------------------------------------
+
+
+def _redact_streaming_tool() -> None:
```
andrewtruong, line 271: there is no redacting in any of these methods, these just look like setup...

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,592 @@
...
+        pytest.param(
+            _redact_batch_child_tool,
+            "execute_tool",
+            ("gen_ai.tool.call.arguments",),
+            id="batch_child_tool",
+        ),
+    ],
+)
+def test_redact_pii_applied(
+    exercise: Callable[[], None],
```
andrewtruong, line 384: I feel like this should be called something else, maybe `user_func` or something that describes that this is what the user is expected to be doing...

tests/session/test_session_settings.py
```diff
@@ -0,0 +1,592 @@
...
+    otel_spans: InMemorySpanExporter,
+):
+    """redact_pii=True applies redaction across streaming + batch paths."""
+    substitutor = _email_substitutor()
+    with override_settings(redact_pii=True):
+        with (
+            patch("weave.utils.pii_redaction.redact_pii", side_effect=substitutor),
+            patch(
+                "weave.utils.pii_redaction.redact_pii_string", side_effect=substitutor
+            ),
```
andrewtruong, line 396: why do you need to patch these?  Isn't the point that PII redaction does this for you?

weave/utils/capture_info.py
```diff
@@ -0,0 +1,44 @@
...
+import sys
+
+from weave.trace.settings import (
+    should_capture_client_info,
+    should_capture_system_info,
+)
+from weave.version import VERSION
+
+
+def get_capture_info_items() -> list[tuple[str, str]]:
```
andrewtruong, line 21: why return a tuple?  Why not return typedict?  These are very specific keys

weave/utils/pii_redaction.py
```diff
@@ -71,14 +105,63 @@ def redact_recursive(value: Any) -> Any:
...
+    out: list[Message] = []
+    for m in msgs:
+        dumped = m.model_dump()
+        redacted = cast(dict[str, Any], redact_pii(dumped))
+        _restore_literals(redacted, dumped)
+        out.append(Message.model_validate(redacted))
+    return out
+
+
+def redact_system_instructions(insts: list[str] | None) -> list[str] | None:
```
andrewtruong, line 158: use the whole word

weave/utils/pii_redaction.py
```diff
@@ -71,14 +105,63 @@ def redact_recursive(value: Any) -> Any:
...
+    if not data:
+        return data
+    analyzer, anonymizer = _get_engines()
     entities = _get_redaction_entities()
     results = analyzer.analyze(text=data, language="en", entities=entities)
     redacted = anonymizer.anonymize(text=data, analyzer_results=results)
     return redacted.text
 
 
+def _restore_literals(redacted: dict[str, Any], original: dict[str, Any]) -> None:
```
andrewtruong, line 118: I dont understand why you need this

rgao-coreweave:
This PR is getting too hard to review. 

Closing in flavor of https://github.com/wandb/weave/pull/6943

PR #6847, opened 2026-05-14
chore(weave_ts): Relocate semconv, add types and new semconv keys

chance-wnb:
### Summary

Foundation PR for the new GenAI session SDK under `src/genai/`. No runtime behavior yet — this lands data types, relocates the OTel GenAI semconv constants out of the integrations folder so a future shared OTel module can consume them, and adds the new attribute keys the SDK will need.

### Changes

- **Relocate `src/integrations/common/genai.ts` → `src/otel/semconv.ts`.**
  The constants in that file aren't integration-specific; they're the OTel GenAI semconv key namespace. Both the future `src/genai/` module and the   existing `piCodingAgent` integration should pull from one place.
- **New semconv keys** added to the relocated file:
  - `gen_ai.input.messages`, `gen_ai.output.messages` — Python SDK encodes message lists as JSON-string attributes; the TS SDK will mirror that for cross-SDK parity.
  - `gen_ai.usage.reasoning_tokens`.
  - New `WEAVE_RESOURCE_ATTR` bundle: `wandb.entity`, `wandb.project`,  `weave.sdk.version`, `weave.sdk.language` — resource attributes the  server uses to fingerprint emitter origin.
- **New data types** in `src/genai/types.ts`: `Message`, `MessagePart`,  `Modality`, `Reasoning`, `Role`, `Usage`. Plain TS interfaces, no Zod.
- **Module entry** at `src/genai/index.ts` — type-only re-exports for now;  runtime surface lands in later PRs.

### Tests

All compile and tests still pass. No new tests yet.

### DIFF
--- a/sdks/node/src/otel/semconv.ts
+++ b/sdks/node/src/otel/semconv.ts
@@ -1,6 +1,6 @@
 /**
- * Shared constants for OpenTelemetry integrations that emit spans conforming
- * to the GenAI semantic conventions:
+ * Constants for OpenTelemetry emitters in the Weave SDK that produce spans
+ * conforming to the GenAI semantic conventions:
  * https://opentelemetry.io/docs/specs/semconv/gen-ai/
  */
 
@@ -21,6 +21,7 @@ export const GEN_AI_ATTR = {
   GEN_AI_USAGE_INPUT_TOKENS: 'gen_ai.usage.input_tokens',
   GEN_AI_USAGE_OUTPUT_TOKENS: 'gen_ai.usage.output_tokens',
   GEN_AI_USAGE_TOTAL_TOKENS: 'gen_ai.usage.total_tokens',
+  GEN_AI_USAGE_REASONING_TOKENS: 'gen_ai.usage.reasoning_tokens',
   GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS: 'gen_ai.usage.cache_read.input_tokens',
   GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS:
     'gen_ai.usage.cache_creation.input_tokens',
@@ -28,6 +29,8 @@ export const GEN_AI_ATTR = {
   GEN_AI_TOOL_NAME: 'gen_ai.tool.name',
   GEN_AI_TOOL_CALL_ID: 'gen_ai.tool.call.id',
   GEN_AI_OUTPUT_TYPE: 'gen_ai.output.type',
+  GEN_AI_INPUT_MESSAGES: 'gen_ai.input.messages',
+  GEN_AI_OUTPUT_MESSAGES: 'gen_ai.output.messages',
 } as const;
 
 // ---------------------------------------------------------------------------
@@ -50,4 +53,16 @@ export const GEN_AI_EVENT = {
 /** General OpenTelemetry attribute keys used alongside GenAI spans. */
 export const OTEL_ATTR = {
   ERROR_TYPE: 'error.type',
+} as const;
+
+// ---------------------------------------------------------------------------
+// Resource attribute keys
+// ---------------------------------------------------------------------------
+
+/** Resource-level attributes attached to every span emitted by this SDK. */
+export const WEAVE_RESOURCE_ATTR = {
+  WANDB_ENTITY: 'wandb.entity',
+  WANDB_PROJECT: 'wandb.project',
+  WEAVE_SDK_VERSION: 'weave.sdk.version',
+  WEAVE_SDK_LANGUAGE: 'weave.sdk.language',
 } as const;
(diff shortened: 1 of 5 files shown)


chance-wnb:
* **#6847** <a href="https://app.graphite.com/github/pr/wandb/weave/6847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6847?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

rgao-coreweave (approved):
LGTM

sdks/node/src/otel/semconv.ts
andrewtruong, line 1: Do we need this file?  If we depend on the OTEL library directly, can we just import these definitions from there?

sdks/node/src/otel/semconv.ts
```diff
@@ -51,3 +54,15 @@ export const GEN_AI_EVENT = {
 export const OTEL_ATTR = {
   ERROR_TYPE: 'error.type',
 } as const;
+
+// ---------------------------------------------------------------------------
+// Resource attribute keys
+// ---------------------------------------------------------------------------
+
+/** Resource-level attributes attached to every span emitted by this SDK. */
+export const WEAVE_RESOURCE_ATTR = {
```
andrewtruong: I don't think this is part of semconv

sdks/node/src/otel/semconv.ts
chance-wnb: Hey Andrew, based on your suggestion I attempted to do this today, and it turned out to be pretty difficult:

1. None of our existing series of `@opentelemetry/*` dependencies has those values defined. 
2. To see those values, we need an additional dependency `@opentelemetry/semantic-conventions`.
3. however this package doesn't expose the values directly from the root exports, but from a relative `@opentelemetry/semantic-conventions/incubating` path, meaning these value are experimental and tentative to change.
4. Even worse, in order to import from the subpath of `@opentelemetry/semantic-conventions/incubating`, we need some surgery to the dual-build setup. which a: ruins the cjs build extension names, b. enforces a self-declaration of the types which is almost like rewriting the values ourselves, also risking consistency.

So we should better just leave it as is for now. When `@opentelemetry/semantic-conventions` ever thinks of moving the stuff out of `incubating` , we then consider adoption again.

sdks/node/src/otel/semconv.ts
chance-wnb: Indeed, separated them as `weaveResource.ts‎`

sdks/node/src/otel/semconv.ts
andrewtruong: I'm not sure putting it here is better, we're just shifting the dependency from `incubating` to here.  On the python side we do import `semantic-conventions` (which I think is already a transitive dep anyways).  
  
Can you expand on the issue with dual-build?

andrewtruong (approved):
Approving, but have some concern about manually keeping our own version of `semconv.ts`

andrewtruong (approved):
Approving, but have some concern about maintaining `semconv.ts`

sdks/node/src/otel/semconv.ts
chance-wnb: > Can you expand on the issue with dual-build?

The package.json of @opentelemetry/semantic-conventions has

```
"exports": {
  ".": {
    "esnext": "./build/esnext/index.js",
    "module": "./build/esm/index.js",
    "types":  "./build/src/index.d.ts",
    "default": "./build/src/index.js"
  },
  "./incubating": {
    "esnext": "./build/esnext/index-incubating.js",
    "module": "./build/esm/index-incubating.js",
    "types":  "./build/src/index-incubating.d.ts",
    "default": "./build/src/index-incubating.js"
  }
}
```
so `./incubating` is a different entry point other than the root entry point.

Currently we have `moduleResolution: node` in SDK project's tsconfig.json . this config is not capable of following "exports" field and import from an alternative entry point.

`moduleResolution: node` is the only known config that works with both `module: commonjs` and `module: esnext` from my prior dual-build work. 

This combination might work: 
`moduleResolution: node` <-> `module: commonjs`
`moduleResolution: bundler` <-> `module: esnext`

but we should experiment and try out in a separate PR. We can come back to this issue after the import issue is fixed.

andrewtruong: Ok thanks for investigating, I'm ok with the code as written but agree we should try to see if we can just depend on the upstream `semantic-conventions` in a separate PR

drtangible: If we are open to adding the dependency, thoughts on pointing to the documentation in `@opentelemetry/semantic-conventions` that describes their recommendation for using a combination of stable + unstable conventions (https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv)?



Being able to distinguish between stable and unstable conventions quickly would be nice -- it appears some of these are already stable?


Would be nice hygiene to then periodically update as things move from unstable -> stable:

> Occasionally, one should review changes to @opentelemetry/semantic-conventions to see if any used unstable conventions have changed or been stabilized.




-----

Using things directly from their library might let us not have to create our own naming / organization conventions + boilerplate on top of the ones they already follow, which might be nice.

Image


But I'm less familiar with the dual-build issue -- thanks for writing up notes, agree that a separate PR makes sense for introducing that dependency. (Most of the above thoughts were from prior to the last couple comments here)

drtangible (approved):
👍

sdks/node/src/otel/semconv.ts
chance-wnb: Great find Dan! Yeah, I agree that literally copying those constants makes more sense. I will address it on the top of this PR stack.

PR #6848, opened 2026-05-14
chore(weave_ts): Shared GenAI OTel infrastructure

chance-wnb:
### Summary

Lands the process-wide singleton OTel `TracerProvider` that every GenAI emitter in the SDK will use to ship spans to the trace server's `/agents/otel/v1/traces` endpoint. Also collapses the temporary `src/otel/` folder introduced in PR 1 back into `src/genai/` 

### Changes

- **Consolidation refactor.** `src/otel/semconv.ts` → `src/genai/semconv.ts`.
  All new files in this PR land under `src/genai/`. `piCodingAgent`'s semconv import path updated to use the new path
- **`src/genai/provider.ts`** — singleton `BasicTracerProvider` with:
  - OTLP exporter targeting `${client.traceServerApi.baseUrl}/agents/otel/v1/traces` (the new GenAI ingest endpoint).
  - Resource attrs: `wandb.entity`, `wandb.project`, `weave.sdk.version`, `weave.sdk.language='node'`.
  - `getWeaveTracer(name)` returns a no-op tracer (via standalone `BasicTracerProvider` with `AlwaysOffSampler`) if `weave.init()` was not called — downstream API calls remain safe to invoke.
  - One-time `beforeExit` hook (registered the first time a real provider is built) that calls `provider.shutdown()` so in-flight spans flush before the event loop drains.
- **`src/genai/flush.ts`** — public `flushWeaveOTel()` helper that  delegates to `provider.forceFlush()`. Re-exported from `src/index.ts`.
- **Settings extension** in `src/settings.ts`: new `GenAISettingsInit`
  interface with `spanProcessor: 'batch' | 'simple' | SpanProcessor`  (default `'batch'`) and `batchOptions`. The exporter itself is non-replaceable; only the processor is user-selectable.
- **New Tests** under `src/__tests__/genai/provider.test.ts`.

### Design notes

- piCodingAgent's `/otel/v1/traces` endpoint is still wrong, it is out of scope of this project. will address in separate PRs.
- In the future, we could consider exposing `getWeaveOTLPExporter()` so callers can wrap-not-replace the exporter.

### Tested

- [x] Prettier clean.
- [x] All unit tests pass.


chance-wnb:
* **#6848** <a href="https://app.graphite.com/github/pr/wandb/weave/6848?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6848?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

rgao-coreweave (approved):
LGTM

rgao-coreweave (approved):
LGTM

sdks/node/src/__tests__/genai/provider.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  );
+  setGlobalClient(client);
+  return client;
+}
+
+function clearGlobalClient(): void {
+  // setGlobalClient is typed for a real client, but the underlying holder
+  // accepts null. Tests need this to exercise the "weave.init() not called"
+  // branch.
+  setGlobalClient(null as unknown as WeaveClient);
```
andrewtruong: shouldn't we change `setGlobalClient`'s signature to `WeaveClient | null` then?

sdks/node/src/genai/provider.ts
```diff
@@ -0,0 +1,147 @@
...
+
+const SDK_LANGUAGE = 'node';
+const GENAI_OTLP_PATH = '/agents/otel/v1/traces';
+
+// Dual-package-hazard-safe: routed through globalThis so CJS and ESM copies
+// of this module share one provider instance per process.
+const _providerHolder = globalSingleton<{
+  provider: BasicTracerProvider | null;
+  beforeExitRegistered: boolean;
+}>('_weave_genai_provider', () => ({
```
andrewtruong: This seems like a magic const used in a few places.  Where is this defined, and can we pull out into a const?

andrewtruong (approved):

sdks/node/src/__tests__/genai/provider.test.ts
chance-wnb: yes, fixed

sdks/node/src/genai/provider.ts
chance-wnb: Introduced a constant.

sdks/node/src/index.ts
```diff
@@ -23,6 +23,7 @@ export {
   patchRealtimeSession,
   createOtelExtension,
 } from './integrations';
+export {flushWeaveOTel} from './genai/flush';
```
drtangible: Couple small thoughts I'm mostly just curious about:


* If this is part of the public API from `weave`, is including the name `Weave` here redundant?

* Similarly, do we expect consumers to need to know that OTel is used under the hood? Not sure if there's a more generic description of the consumer behavior we're affording for by including it in the public API, eg...

```tsx
  import {init, startSession, flushSpans} from 'weave'; 
  ...

  await init(...);
  const session = startSession({agentName: 'research-bot'});
  ...

  flushSpans();
```

drtangible (approved):
👍👍👍

sdks/node/src/index.ts
chance-wnb: Yep, `weave` is redundant, remove it now. 

On the OTEL thing, I am afraid I don't have a mature answer on that yet. I might need to rely on your all's judgement. :) 

On the buildOtlpExporter's header comment, I left this note:

```
// Private today. A user-supplied SpanProcessor (via
// settings.genai.spanProcessor) fully owns its own exporter and bypasses
// this one — fine for tests and for users who don't care about reaching
// the Weave Agents tab, but it leaves no clean path for "custom
// queue/retry logic AND spans land in Weave."
//
// If that use case shows up, expose this as `getWeaveOTLPExporter()` so
// callers can wrap-not-replace:
//
//   const exporter = weave.getWeaveOTLPExporter();
//   const processor = new MyCustomBatchProcessor(exporter, { ... });
//   await weave.init('entity/project', { genai: { spanProcessor: processor } });
//
// Not done now: nothing needs it, and widening the public API has support
// cost.
function buildOtlpExporter(client: WeaveClient): OTLPTraceExporter {
```

So I planned a future feature that would let users wrap the OTEL exporter with their own logic and then pass the wrapped exporter back to us. That way, OTEL spans could be sent both to our destination and to a user-specified one. If this is the direction we want to go in—then exposing the OTEL concept is unavoidable

drtangible: 👍 👍 👍

PR #6849, opened 2026-05-14
chore(weave_ts): Core classes: Session, Turn, LLM, Tool, SubAgent (manual start/end)

chance-wnb:
### Summary

Lands the five GenAI session classes and their chained child-creation methods. Manual `start()` / `.end({ error? })` is the primary idiom (no ALS yet — that's PR 4). The Session→Turn→LLM→Tool/SubAgent hierarchy is fully traceable end-to-end via the chained form. Classes are exposed as **types only** at the package boundary so consumers can name them in their own signatures but cannot construct them directly — construction is private to the SDK's top-level entry points (which land in PR 4).

### Changes

- **New class files:**
  - `src/genai/session.ts` — no span, propagates `sessionId` (UUID v7 if not supplied) as `gen_ai.conversation.id` to every child.
  - `src/genai/turn.ts` — emits `invoke_agent` root span (always rooted at `ROOT_CONTEXT`, so external OTel context can never accidentally  parent us). Owns chained `llm()` / `tool()` / `subagent()` factories.
  - `src/genai/llm.ts` — emits `chat` span as child of Turn. Mutable  `inputMessages` / `outputMessages` / `usage` / `reasoning` fields  populated by the user between `create()` and `end()`; serialized to span attributes (GenAI semconv) on close. Owns chained  `startTool()` / `startSubagent()` for the nested form.
  - `src/genai/tool.ts` — emits `execute_tool` span. `result?: string`  field for free-form output capture.
  - `src/genai/subagent.ts` — emits nested `invoke_agent` span.
- **`src/genai/common.ts`** — `ChildSpanContext { parentContext: Context;  conversationId?: string }` threaded from parent to child. Each non-leaf class stores its own derived `Context` and passes it down  (children build on parents' context rather than reconstructing from `ROOT_CONTEXT` + parent span).
- **Tool parent resolution** (design §6.3): both forms supported.
  - `turn.tool(...)` — Tool is a sibling of LLM under Turn (flat).
  - `llm.startTool(...)` — Tool is a child of the LLM span (nested).
- **Public type contract** in `src/index.ts`: `export type { ... }` for the five class types plus init types and data types.
- **`WEAVE_GENAI_TRACER_NAME = 'weave-genai'`** added to `src/genai/semconv.ts`.
- **New Tests** — per-class test files under `src/__tests__/genai/`

### Design notes

- Static `create()` factories sit on the class but are unreachable from consumer code because the class is exported as `export type` only. This lets the chained methods inside the package construct via `LLM.create()` while keeping `LLM.create(...)` invisible to consumers — they go through the top-level entry points (PR 4) or chained calls.
- `tool.result` is a public mutable field (set before `end()`), matching the Python SDK's `step.usage = Usage(...)` field-assignment pattern.

### Tested

- [x] Prettier clean.
- [x] All unit tests pass.

### DIFF
--- a/sdks/node/src/__tests__/genai/common.ts
+++ b/sdks/node/src/__tests__/genai/common.ts
@@ -1,0 +1,105 @@
+import {
+  InMemorySpanExporter,
+  type ReadableSpan,
+  SimpleSpanProcessor,
+} from '@opentelemetry/sdk-trace-base';
+
+import {setGlobalClient} from '../../clientApi';
+import {Api as TraceServerApi} from '../../generated/traceServerApi';
+import {PROVIDER_HOLDER_SYMBOL_NAME} from '../../genai/provider';
+import {Settings, type SettingsInit} from '../../settings';
+import {WandbServerApi} from '../../wandb/wandbServerApi';
+import {WeaveClient} from '../../weaveClient';
+
+export const TEST_BASE_URL = 'http://localhost:8080';
+export const TEST_PROJECT = 'test-entity/test-project';
+
+export function installFakeClient(settings: SettingsInit = {}): WeaveClient {
+  const traceServerApi = {baseUrl: TEST_BASE_URL} as TraceServerApi<any>;
+  const client = new WeaveClient(
+    traceServerApi,
+    {} as WandbServerApi,
+    TEST_PROJECT,
+    new Settings(
+      settings.printCallLink ?? true,
+      settings.globalAttributes ?? {},
+      settings.genai ?? {}
+    )
+  );
+  setGlobalClient(client);
+  return client;
+}
+
+export function clearGlobalClient(): void {
+  // setGlobalClient is typed for a real client, but the underlying holder
+  // accepts null. Tests need this to exercise the "weave.init() not called"
+  // branch.
+  setGlobalClient(null as unknown as WeaveClient);
+}
+
+// Reach into the global singleton's holder directly to reset its state.
+export function resetProviderSingleton(): void {
+  const holder = (globalThis as Record<symbol, unknown>)[
+    Symbol.for(PROVIDER_HOLDER_SYMBOL_NAME)
+  ] as {provider: unknown; beforeExitRegistered: boolean} | undefined;
+  if (holder) {
+    holder.provider = null;
+    holder.beforeExitRegistered = false;
+  }
+}
+
+/**
+ * Install the standard before/after-each pattern for GenAI tests: stubs
+ * `WANDB_API_KEY` (so the provider's OTLP exporter can be built), resets
+ * the provider singleton, and clears the global Weave client. Call inside
+ * the consumer's `describe` block.
+ */
+export function setupGenAITestEnvironment(): void {
+  const originalApiKey = process.env.WANDB_API_KEY;
+
+  beforeEach(() => {
+    process.env.WANDB_API_KEY = 'test-api-key';
+    resetProviderSingleton();
+    clearGlobalClient();
+  });
+
+  afterEach(() => {
+    resetProviderSingleton();
+    clearGlobalClient();
+    if (originalApiKey === undefined) {
+      delete process.env.WANDB_API_KEY;
+    } else {
+      process.env.WANDB_API_KEY = originalApiKey;
+    }
+  });
+}
+
+/** Find a finished span by name; throw with a helpful message if missing. */
+export function findSpan(spans: ReadableSpan[], name: string): ReadableSpan {
+  const span = spans.find(s => s.name === name);
+  if (!span) {
+    throw new Error(
+      `no span named '${name}' (saw: ${spans.map(s => s.name).join(', ')})`
+    );
+  }
+  return span;
+}
+
+/**
+ * Per-test setup that installs a fresh `InMemorySpanExporter` + fake client.
+ * Returns a getter so tests can read `getExporter().getFinishedSpans()` after
+ * spans are ended.
+ *
+ * Call inside a describe block, after `setupGenAITestEnvironment()` so the
+ * API key and singleton resets happen first.
+ */
+export function setupExporterPerTest(): () => InMemorySpanExporter {
+  let current: InMemorySpanExporter;
+  beforeEach(() => {
+    current = new InMemorySpanExporter();
+    installFakeClient({
+      genai: {spanProcessor: new SimpleSpanProcessor(current)},
+    });
+  });
+  return () => current;
+}
--- a/sdks/node/src/__tests__/genai/provider.test.ts
+++ b/sdks/node/src/__tests__/genai/provider.test.ts
@@ -4,75 +4,15 @@ import {
   SimpleSpanProcessor,
 } from '@opentelemetry/sdk-trace-base';
 
-import {setGlobalClient} from '../../clientApi';
-import {Api as TraceServerApi} from '../../generated/traceServerApi';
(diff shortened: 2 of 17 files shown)


chance-wnb:
* **#6849** <a href="https://app.graphite.com/github/pr/wandb/weave/6849?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6849?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/__tests__/genai/common.ts
andrewtruong, line 1: Was this re-created?  I thought I saw this in the last PR

sdks/node/src/__tests__/genai/provider.test.ts
andrewtruong, line 1: oh I see it was deleted from here

sdks/node/src/__tests__/genai/tool.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    expect(toolSpan.attributes[GEN_AI_ATTR.GEN_AI_CONVERSATION_ID]).toBe(
+      'conv-1'
+    );
+    expect(toolSpan.parentSpanId).toBe(turnSpan.spanContext().spanId);
+  });
+
+  it('attaches to the LLM span when started via llm.startTool() (nested)', async () => {
+    const turn = await Turn.create({});
+    const llm = await turn.llm({model: 'gpt-4o'});
+    const tool = await llm.startTool({name: 'get_weather'});
```
andrewtruong: Is this concept new? I don't think `startTool` exists in Python.

sdks/node/src/genai/llm.ts
```diff
@@ -0,0 +1,141 @@
...
+    });
+  }
+
+  async startSubagent(opts: SubAgentInit): Promise<SubAgent> {
+    return SubAgent.create({
+      ...opts,
+      parentContext: this.context,
+      conversationId: this.conversationId,
+    });
+  }
```
andrewtruong: Are these concepts new?  I don't think they exist in python

sdks/node/src/__tests__/genai/tool.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  const getExporter = setupExporterPerTest();
+
+  it('attaches to the turn span when started via turn.tool() (flat)', async () => {
+    const turn = await Turn.create({conversationId: 'conv-1'});
+    const tool = await turn.tool({
+      name: 'get_weather',
+      args: '{"city":"Tokyo"}',
+      toolCallId: 'tc-1',
+    });
+    tool.result = '75F';
```
andrewtruong, line 23: You set this, but I don't think it's checked anywhere.

sdks/node/src/__tests__/genai/tool.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+describe('Tool', () => {
+  setupGenAITestEnvironment();
+  const getExporter = setupExporterPerTest();
+
+  it('attaches to the turn span when started via turn.tool() (flat)', async () => {
+    const turn = await Turn.create({conversationId: 'conv-1'});
+    const tool = await turn.tool({
+      name: 'get_weather',
+      args: '{"city":"Tokyo"}',
```
andrewtruong, line 20: This too.  Maybe worth doing another check to see all of the things you set are checked

sdks/node/src/genai/tool.ts
```diff
@@ -0,0 +1,58 @@
...
+  private _ended = false;
+
+  private constructor(
+    private readonly span: Span,
+    public readonly name: string,
+    public readonly args: string,
+    public readonly toolCallId: string
+  ) {}
+
+  static async create(opts: ToolInit & ChildSpanContext): Promise<Tool> {
```
andrewtruong: Per the other class definitions, it doesn't seem like there's any reason why create should be async. What's the benefit of making it async here? Is it just consistency?

sdks/node/src/__tests__/genai/common.ts
chance-wnb: Yes, this is a migration.

sdks/node/src/__tests__/genai/tool.test.ts
chance-wnb: they are there: https://github.com/wandb/weave/blob/669c98c7968b1e2ccf27d1b42a6bf255c86bb66d/weave/session/__init__.py#L41

sdks/node/src/genai/llm.ts
chance-wnb: They do exist in python: https://github.com/wandb/weave/blob/669c98c7968b1e2ccf27d1b42a6bf255c86bb66d/weave/session/__init__.py#L41

sdks/node/src/__tests__/genai/tool.test.ts
chance-wnb: True, added some checks now.

sdks/node/src/__tests__/genai/tool.test.ts
chance-wnb: This should be validated now.

sdks/node/src/genai/tool.ts
chance-wnb: yeah, these are issued caused by repetitive refactoring. Now removed most of the async/promise usage.

sdks/node/src/genai/turn.ts
```diff
@@ -0,0 +1,102 @@
...
+import {LLM, type LLMInit} from './llm';
+import {getWeaveTracer} from './provider';
+import {GEN_AI_ATTR, WEAVE_GENAI_TRACER_NAME} from './semconv';
+import {SubAgent, type SubAgentInit} from './subagent';
+import {Tool, type ToolInit} from './tool';
+
+export interface TurnInit {
+  agentName?: string;
+  model?: string;
+  userMessage?: string;
```
andrewtruong: it doesn't seem like this used in `create` below.  What is this for?

andrewtruong: oh maybe this is consumed in a future PR?

sdks/node/src/genai/llm.ts
```diff
@@ -0,0 +1,143 @@
...
+import {getWeaveTracer} from './provider';
+import {GEN_AI_ATTR, WEAVE_GENAI_TRACER_NAME} from './semconv';
+import {SubAgent, type SubAgentInit} from './subagent';
+import {Tool, type ToolInit} from './tool';
+import type {Message, Reasoning, Usage} from './types';
+
+export interface LLMInit {
+  model: string;
+  providerName?: string;
+  systemInstructions?: string[];
```
andrewtruong: same here

andrewtruong (approved):
lgtm overall; I think some of my comments can be separately addressed in a PR6

drtangible (approved):
👍 👍 👍

sdks/node/src/genai/turn.ts
chance-wnb: Great catch. This extra field is pulled out now. no one consumes it. The actual message is embodied from the LLM usages.

sdks/node/src/genai/llm.ts
chance-wnb: Pulled this field out as well (no use). Thanks!

PR #6850, opened 2026-05-14
chore(weave_ts): Top-level entry points + AsyncLocalStorage context

chance-wnb:
### Summary

Wires the public top-level API: `weave.startSession`, `startTurn`, `startLLM`, `startTool`, `startSubagent`, `endSession`, `endTurn`, `endLLM`, and `getCurrentSession` / `getCurrentTurn` / `getCurrentLLM`.
Parent resolution between calls happens through three `AsyncLocalStorage` stores — Python `ContextVar` parity, install on `create()` and restore on `end()`. With this PR, three of the four design-doc patterns work end-to-end (context manager equivalent via `try/finally`, manual start/end, and cross-module via ALS — batch logging stays for a later PR).

### Changes

- **`src/genai/context.ts`** — Three `AsyncLocalStorage` stores
  (`_currentSession`, `_currentTurn`, `_currentLLM`). Public accessors  `getCurrentSession()`, `getCurrentTurn()`, `getCurrentLLM()`.
- **ALS install / restore in Session, Turn, LLM constructors and  `end()`.** Each constructor captures the previous value via  `_currentX.getStore()` and calls `enterWith(this)`. Each `end()`  restores via `_currentX.enterWith(this._previousX)`. Idempotent (guarded by `_ended`). Tool and SubAgent are leaves — no ALS slot.
- **`src/genai/api.ts`** — Top-level functions:
  - `startSession` — always works, installs Session in ALS.
  - `startTurn` — inherits `conversationId` from current Session if present; otherwise starts a Turn with no conversation id.
  - `startLLM`, `startTool`, `startSubagent` — read parent via ALS;  throw with a clear error message if the required parent is missing. `startTool` / `startSubagent` prefer current LLM, fall back to current Turn.
  - `endSession` / `endTurn` / `endLLM` — no-ops when nothing is active;  safe in `finally` blocks.
- **Public exports** from `src/index.ts`: value exports for the new top-level fns and `getCurrent*` accessors, alongside the existing  `export type` block for the class types.
- **New Tests** — `src/__tests__/genai/api.test.ts` (10 tests).

### Design notes

`startLLM` / `startTool` / `startSubagent` are strict (throw on missing parent) rather than permissively creating a root span. Forces the user to structure their calls; clearer error model.

### Tested

- [x] Prettier clean.
- [x] All unit tests pass.
----
# Addendum — context model refactor

The original PR 4 wired Session / Turn / LLM tracking through three separate `AsyncLocalStorage` instances, each installed on entry with `als.enterWith(instance)` and restored on `end()` by re-running `enterWith(previous)`. That model allowed arbitrary stacking — nested Sessions, nested Turns, nested LLMs — because each `end()` could pop back to the previously stashed instance.

We've since tightened the contract: **at most one active Session, one active Turn, and one active LLM per async chain**. There is no semantic use case for nesting (a "nested Turn" within a Turn isn't a thing in our agent model), so the implementation no longer needs to support it. This addendum is the refactor that drops the stacking machinery and replaces it with something simpler and safer for parallel work.

## What changed

- **One ALS, one container.** A single `AsyncLocalStorage<GenAIState>` holds a mutable `{session, turn, llm}` container. `start*` paths assign into the container's slots; `end*` paths null them back out. No `enterWith` anywhere, no per-instance stack.
- **Nesting guards.** `Session.create` / `Turn.create` / `LLM.create` throw if their slot in the active container is already populated. Makes the "one active at a time per chain" rule explicit instead of silently overwriting.
- **`runIsolated(fn)`.** New public API. Wraps `fn` in `als.run(freshState(), fn)` so any Session / Turn / LLM started inside `fn` lives in that frame's container and cannot clash with sibling `runIsolated` frames. This is the supported way to run parallel GenAI work.
- **Process-wide default container.** When no `runIsolated` frame is on the stack, `_getGenaiState()` falls back to a single shared `_defaultState`. Sequential single-flight callers (the common case) don't need any wrapper.
- **Test-only reset lives in the test helper.** `resetGenaiDefaultState` in `__tests__/genai/common.ts` reaches the holder via `Symbol.for(DEFAULT_STATE_SYMBOL_NAME)`, mirroring how `resetProviderSingleton` reaches the provider holder. Production `context.ts` no longer exports a test hatch.

## Why dropping `enterWith` matters

`enterWith` mutates the *current* async frame's store and that mutation is inherited by every async continuation forked from that frame, including continuations that live above any `als.run(...)` boundary a user might wrap around their code. With the old three-ALS-plus-`enterWith` design, two `startSession(...)` calls running in parallel `Promise.all` branches both wrote into the same ALS and clashed — there was no clean way to scope concurrent independent sessions. The new design routes isolation through `als.run` (which *does* respect frame boundaries), and uses plain object mutation on the container the run-frame owns.

## Trade-off and rationale

The default-container fallback is **not safe for concurrent independent sessions** — that's the whole reason `runIsolated` exists. The split is intentional: keep the casual `weave.startSession(...)` → `weave.startTurn(...)` ergonomic with zero boilerplate, and route users who need parallelism through an explicit `runIsolated` wrapper. This matches the Python SDK's contextvar story (contextvars copy on async fork; an explicit `copy_context().run(...)` is the parallel-isolation primitive).

## Tests added

- `runIsolated isolates concurrent Sessions in Promise.all` — sibling frames each see their own session, neither leaks to the outer chain.
- `runIsolated supports a full Session → Turn → LLM stack per frame` — the full chain works per-frame.
- `parallel sessions WITHOUT runIsolated clash on the default state and throw` — documents and locks in the nesting-guard behavior on the unwrapped path.
- `starting a Session / Turn / LLM while one is already active throws` — covers the one-at-a-time contract directly.


chance-wnb:
* **#6850** <a href="https://app.graphite.com/github/pr/wandb/weave/6850?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6850?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/__tests__/genai/common.ts
```diff
@@ -63,11 +64,19 @@ export function setupGenAITestEnvironment(): void {
     process.env.WANDB_API_KEY = 'test-api-key';
     resetProviderSingleton();
     clearGlobalClient();
+    // Clear ALS stores so test bodies start without inherited Session/Turn/LLM
+    // state from any prior test in the same async chain.
+    _currentSession.enterWith(undefined);
+    _currentTurn.enterWith(undefined);
+    _currentLLM.enterWith(undefined);
```
andrewtruong: Maybe you can make this a helper.

sdks/node/src/__tests__/genai/api.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    expect(getCurrentSession()).toBeUndefined();
+  });
+
+  // -------------------------------------------------------------------------
+  // ALS restore-on-end with nested same-class instances
+  // -------------------------------------------------------------------------
+
+  it('ending the inner Turn restores the outer as current', async () => {
+    const outer = await startTurn({agentName: 'outer'});
+    const inner = await startTurn({agentName: 'inner'});
```
andrewtruong: The practical use case for this would be a nested subagent with its own set of turns, right?

andrewtruong: Maybe it's worth explicitly testing that case too.

andrewtruong: Otherwise, I don't think it makes sense to nest turns. Turns, by definition, should be siblings.

sdks/node/src/__tests__/genai/api.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    await turn.end();
+
+    const spans = getExporter().getFinishedSpans();
+    const toolSpan = findSpan(spans, 'execute_tool');
+    const turnSpan = findSpan(spans, 'invoke_agent');
+    expect(toolSpan.parentSpanId).toBe(turnSpan.spanContext().spanId);
+  });
+
+  // -------------------------------------------------------------------------
+  // Error paths
```
andrewtruong, line 116: I was expecting to see an `endSession` test somewhere to see how it interacts with the methods added in this PR

sdks/node/src/__tests__/genai/common.ts
chance-wnb: Thanks. This is not applicable after the refactoring.

sdks/node/src/__tests__/genai/api.test.ts
chance-wnb: Right, nested turns don't make sense. I have added some protection to prevent nested objects. This lead to some opportunity to simplify the ALS usage. I have done some significant refactoring in this PR. Please kindly help take another look. I included an addendum section in the PR description to explain.

sdks/node/src/__tests__/genai/api.test.ts
chance-wnb: Fixed. now `endLLM / endTurn / endSession clear their current instances` test case has it.

sdks/node/src/__tests__/genai/api.test.ts
```diff
@@ -0,0 +1,259 @@
...
+    ]);
+    expect(results[0].status).toBe('fulfilled');
+    expect(results[1].status).toBe('rejected');
+    if (results[1].status === 'rejected') {
+      expect(String(results[1].reason)).toMatch(/Session is already active/);
+    }
+  });
+
+  // -------------------------------------------------------------------------
+  // end* are idempotent no-ops when no instance is active
```
andrewtruong, line 231: I think we also need an `endSession` test that checks for cascading.  If the session ends, then its children should have also ended first

andrewtruong (approved):
Overall lgtm, please see last comment

sdks/node/src/index.ts
```diff
@@ -24,6 +24,20 @@ export {
...
+  getCurrentLLM,
+  getCurrentSession,
+  getCurrentTurn,
+  runIsolated,
+  startLLM,
+  startSession,
+  startSubagent,
+  startTool,
+  startTurn,
+} from './genai';
```
drtangible, line 40: super tiny organization nit: Did we want flushWeaveOTel to be exported / imported from ./genai similar to all these, now that they're here?

sdks/node/src/index.ts
```diff
@@ -24,6 +24,20 @@ export {
   createOtelExtension,
 } from './integrations';
 export {flushWeaveOTel} from './genai/flush';
+export {
+  endLLM,
+  endSession,
+  endTurn,
+  getCurrentLLM,
+  getCurrentSession,
+  getCurrentTurn,
```
drtangible, line 33: Just adding a note for my future reference: discussed offline with @chance-wnb about whether these `getFoo` functions afford for add'l behavior not possible via the top-level API. If not, could consider removing to keep the public API smaller.

drtangible (approved):
👍 👍 👍 


Adding note just for my future reference -- discussed this bit offline with @chance-wnb:

> With this PR, three of the four design-doc patterns work end-to-end (context manager equivalent via try/finally

We don't have an explicit "context manager" equivalent (but don't think we need one, as the explicit APIs are in place).

sdks/node/src/__tests__/genai/api.test.ts
chance-wnb: Great reminder. the feature was actually missing. added the cascading close logic and a test to validate as well.

sdks/node/src/index.ts
chance-wnb: Yeah, refactored and it is cleaner now.

sdks/node/src/index.ts
chance-wnb: Just add an additional note:

Lets consider this example:

```typescript
//main.ts

import {Engine} from "FabulousGame";

import { create, update, draw} from "./myGame";

const session = weave.startSession();

const myGame = new Engine({
  create, update, draw
})


//myGame.ts
export function update() {
  // const session = weave.getCurrentSession();
}
```

When such a framework doesn't provide a way to pass the `session` object over. Then user has to use the `weave.getCurrentSession()` helper to get it. `myGame.ts` cannot re-import `main.ts` without causing circular import. There are work arounds. Say user can manually build up the AsyncLocalStorage themselves to store the `session` object or pushing the `const session = weave.startSession();` into a separate file to avoid circular import. 

So the `getFoo` functions are good to have utilities. they are not something to fill up any function gaps. They are purely convenience.

Just my two cents. You and the team own the final decision :)

drtangible: Cool cool -- makes sense!

PR #6851, opened 2026-05-14
chore(weave): preserve unpaired calls on flush via eager start

gtarpenning:
## Summary
- `CallBatchProcessor` parks starts in `_pending_starts` until a matching end arrives, then queues the pair as a single `CompleteBatchItem`. On `stop_accepting_new_work_and_flush_queue`, anything still unpaired after `FLUSH_TIMEOUT_SECONDS` was written to the disk-fallback log and silently dropped, which loses long-running ops on process exit even when the network was healthy.
- Send orphan starts/ends via the eager v2 endpoints (`/call/start`, `/call/end`) instead by routing them through `_queue_item` -> `_process_mixed_batch` -> `eager_processor_fn`. `_queue_item` keeps the disk fallback as a final safety net if the queue itself is full.
- Bump `FLUSH_TIMEOUT_SECONDS` from 60s to 5min.
- Orphans are queued **before** `stop_accepting_work_event.set()` so the processing thread actually drains them before exiting.

## Testing
renamed `test_flush_drops_unpaired_items_and_clears_state` to `test_flush_eager_sends_unpaired_items_and_clears_state`; asserts the orphan start and orphan end are passed to `eager_processor_fn`.


weave/trace_server_bindings/call_batch_processor.py
```diff
@@ -190,40 +187,39 @@ def stop_accepting_new_work_and_flush_queue(self) -> None:
...
+                    "endpoint (did not pair within %ss).",
                     len(self._pending_starts),
                     FLUSH_TIMEOUT_SECONDS,
                 )
-                for call_id, start_item in self._pending_starts.items():
-                    self._write_item_to_disk(
-                        start_item,
-                        f"Unpaired start dropped after flush timeout: {call_id}",
-                    )
+                for start_item in self._pending_starts.values():
```
andrewtruong: nit: Don't need to call `values()`

weave/trace_server_bindings/call_batch_processor.py
```diff
@@ -190,40 +187,39 @@ def stop_accepting_new_work_and_flush_queue(self) -> None:
...
+                    "Flush timeout: sending %s unpaired ends via the eager v2 "
+                    "endpoint (did not pair within %ss).",
                     len(self._pending_ends),
                     FLUSH_TIMEOUT_SECONDS,
                 )
-                for call_id, end_item in self._pending_ends.items():
-                    self._write_item_to_disk(
-                        end_item, f"Unpaired end dropped after flush timeout: {call_id}"
-                    )
+                for end_item in self._pending_ends.values():
```
andrewtruong: same here

andrewtruong (commented):
It doesn't seem like you implemented the fallback to eager endpoints.  Does that already exist?  If yes, why didn't it do that before?

gtarpenning:
> It doesn't seem like you implemented the fallback to eager endpoints. Does that already exist? If yes, why didn't it do that before?

@andrewtruong 

The reason was due to performance, kicking the tires on very long `UPDATE` statements. but prod clickhouse seems to be able to handle this, will need to track this though. That's why i conservatively bumped the timeout.

weave/trace_server_bindings/call_batch_processor.py
gtarpenning: done

weave/trace_server_bindings/call_batch_processor.py
gtarpenning: done

andrewtruong (approved):

PR #6856, opened 2026-05-14
fix(weave): set `opentelemetry>=1.38.0` to fix `import weave` if you `pip install wandb` before `weave`

kelu-wandb:
## Description

This PR sets a minimum version of `opentelemetry` so we don't end up with disjoint versions that are mutually incompatible if a too-new version of `protobuf` is already installed, e.g. by `wandb`.

### Steps to reproduce:

```
uv venv
uv pip install wandb==0.27.0 weave==0.52.40
uv run python -c 'import weave'
```
This throws an error:

```
    _descriptor.FieldDescriptor(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
...
TypeError: Descriptors cannot be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
```

(Full error below.)

Workaround is to reverse the order of installing

```
uv venv
uv pip install weave==0.52.40 wandb==0.27.0 
uv run python -c 'import weave'
```
This passes.

The cause is that wandb just requires `protobuf < 8`, so when you install it first, you end up with:

```
protobuf                               7.34.1
opentelemetry-api                      1.41.1
opentelemetry-exporter-otlp-proto-http 1.11.1
opentelemetry-proto                    1.11.1
opentelemetry-sdk                      1.41.1
```

Newer versions of `opentelemetry-proto` have a minimum `protobuf` version requirement, whereas version `1.11.1` doesn't, so we end up with that very old version.

The mismatch between different `opentelemetry` components then causes the error.

If you install `weave` before `wandb`, you get:

```
protobuf                                 6.33.6
opentelemetry-api                        1.41.1
opentelemetry-exporter-otlp-proto-http   1.41.1
opentelemetry-proto                      1.41.1
opentelemetry-sdk                        1.41.1
```

### Full Error

```log
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/weave/__init__.py", line 15, in <module>
    from weave.trace.api import (
    ...<27 lines>...
    )
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/weave/trace/api.py", line 13, in <module>
    from weave import type_handlers  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/weave/type_handlers/__init__.py", line 3, in <module>
    from weave.type_handlers.Content import content
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/weave/type_handlers/Content/content.py", line 12, in <module>
    from weave.trace.serialization.custom_objs import MemTraceFilesArtifact
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/weave/trace/serialization/custom_objs.py", line 22, in <module>
    from weave.trace_server.trace_server_interface import (
    ...<2 lines>...
    )
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/weave/trace_server/trace_server_interface.py", line 13, in <module>
    from opentelemetry.proto.trace.v1.trace_pb2 import ResourceSpans
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/opentelemetry/proto/trace/v1/trace_pb2.py", line 14, in <module>
    from opentelemetry.proto.common.v1 import common_pb2 as opentelemetry_dot_proto_dot_common_dot_v1_dot_common__pb2
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/opentelemetry/proto/common/v1/common_pb2.py", line 36, in <module>
    _descriptor.FieldDescriptor(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
      name='string_value', full_name='opentelemetry.proto.common.v1.AnyValue.string_value', index=0,
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
      is_extension=False, extension_scope=None,
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      serialized_options=None, file=DESCRIPTOR,  create_key=_descriptor._internal_create_key),
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kelu/wandb/examples-dev/evals-for-models/.venv/lib/python3.13/site-packages/google/protobuf/descriptor.py", line 675, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
TypeError: Descriptors cannot be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```

## Testing

How was this PR tested?

```
uv venv
uv pip install wandb==0.27.0 -e ~/wandb/weave
uv run python -c 'import weave'
```

Now no longer throws an error, and this is what gets installed:

```
protobuf                                 6.33.6
opentelemetry-api                        1.41.1
opentelemetry-exporter-otlp-proto-http   1.41.1
opentelemetry-proto                      1.41.1
opentelemetry-sdk                        1.41.1
```


pyproject.toml
```diff
@@ -51,9 +51,9 @@ dependencies = [
   "polyfile-weave>=0.5.9",
 
   # Session SDK — uses OTel GenAI semantic conventions for span emission
-  "opentelemetry-api",
-  "opentelemetry-sdk",
-  "opentelemetry-exporter-otlp-proto-http",
+  "opentelemetry-api>=1.38.0",
+  "opentelemetry-sdk>=1.38.0",
+  "opentelemetry-exporter-otlp-proto-http>=1.38.0",
```
andrewtruong, line 56: Do you need to pin these as well or just the proto below?

kelu-wandb: The `opentelemetry-proto` below is in an optional dep. If anything, that's the one that we maybe don't even need to upgrade? I just did it for consistency.

The real problem is that `opentelemetry-exporter-otlp-proto-http` here indirectly depends on `opentelementary-proto`. And again `opentelemetry-api` and `opentelemetry-sdk` here also don't technically need the pin, but I'm doing it for consistency.

The minimum change needed to fix the problem is:

`opentelemetry-exporter-otlp-proto-http>=1.15.0`

If you just run `uv pip install wandb==0.27.0 -e ~/wandb/weave` with that, you get:

```
protobuf                                 6.33.6
opentelemetry-api                        1.41.1
opentelemetry-exporter-otlp-proto-http   1.41.1
opentelemetry-proto                      1.41.1
opentelemetry-sdk                        1.41.1
```

I felt that making them all the same seemed more consistent, but I'm fine with that minimal version as well.

And fwiw I verified that if I were to force `==1.15.0` to simulate something forcing it to be that low, it still works, with:

```
protobuf                               4.25.9
opentelemetry-api                      1.41.1
opentelemetry-exporter-otlp-proto-http 1.15.0
opentelemetry-proto                    1.15.0
opentelemetry-sdk                      1.41.1
```

but that feels janky to me?

pyproject.toml
rgao-coreweave: opentelemetry-exporter-otlp is still unpinned. Do we need to do the same thing here?

rgao-coreweave (approved):
Thanks for fixing, LGTM.

As a follow-up, it might be worth adding a CI job that runs the resolver in minimum-version mode, so we catch this whole class of bug automatically instead of reactively.

The idea: a declaration like opentelemetry-api>=1.38.0 is a promise that the code works with any version ≥1.38.0. In normal CI we only ever exercise the newest version (1.41.x today), so the floor goes untested. 

The bug in this PR is exactly that — opentelemetry-api was effectively unbounded, and nothing in CI ever tried to resolve it to an old version. It only broke when wandb's protobuf<8 constraint nudged the resolver into picking ancient opentelemetry-proto==1.11.1.

With uv:
```
uv venv
uv pip install --resolution=lowest-direct .
python -c "import weave"
```

# plus a couple of smoke tests

--resolution=lowest-direct picks the oldest version of each direct dependency that still satisfies the declared bounds (transitive deps resolve normally, which avoids spurious failures from under-constrained third-party packages). So post-merge it would install opentelemetry-api==1.38.0 exactly, and verify that import weave actually works at the floor we promised. Pre-merge, on main today, the same job would have resolved to whatever the earliest opentelemetry-api is and failed loudly, forcing us to add a real lower bound before a user hit it.

It's a cheap job to add (one venv, one import, a handful of smokes) and it keeps the dependency floors in pyproject.toml honest going forward.

pyproject.toml
kelu-wandb: done for consistency.

kelu-wandb:
@rgao-coreweave yeah it'll take a bit of work getting that working:

Trying that with just installing weave:

```shell
% rm -rf /tmp/v1 && uv venv /tmp/v1 && VIRTUAL_ENV=/tmp/v1 uv pip install --resolution=lowest-direct weave
% VIRTUAL_ENV=/tmp/v1 uv run python -c 'import weave'
  File "<string>", line 1, in <module>
    import weave
  File "/tmp/v1/lib/python3.13/site-packages/weave/__init__.py", line 6, in <module>
    from . import weave_types as types
  File "/tmp/v1/lib/python3.13/site-packages/weave/weave_types.py", line 10, in <module>
    from . import box
  File "/tmp/v1/lib/python3.13/site-packages/weave/box.py", line 5, in <module>
    import pytz
ModuleNotFoundError: No module named 'pytz'
...
```

from `wandb/weave` clone:

```shell
~/wandb/weave% rm -rf /tmp/v1 && uv venv /tmp/v1 && VIRTUAL_ENV=/tmp/v1 uv pip install --resolution=lowest-direct .
...
      Built weave @ file:///Users/kelu/wandb/weave
  × Failed to build `pydantic-core==2.0.1`
...
```

So yeah there's a bunch of stuff to fix before we can set up that CI :P

kelu-wandb:
Btw it looks like with this fix, installing both wandb and weave basically forces protobuf to downgrade from v7 to v6. I'm not sure if that will have any negative impact on the off chance that there are users trying to use protobuf v7?

opentelemetry seems to pin to <7 though.


Looks like someone filed a ticket against opentelemetry about this just yesterday :P 
https://github.com/open-telemetry/opentelemetry-python/issues/5205

kelu-wandb:
Reverted:
- #6860

PR #6859, opened 2026-05-14
chore(weave): trace dead gaps inside otel_export

gtarpenning:
## Summary
- A typical `POST /otel/v1/traces` request shows ~840ms total with `clickhouse_trace_server_batched.otel_export` consuming 96% of it, but only 4 traced inner children. The bulk is uninstrumented Python work (proto->call conversion, row materialization) showing up as dead space in the flamegraph.
- Wraps `objs_query` (called by `_get_existing_ops_from_spans` from `otel_export`) so the Python builder cost around the CH query is visible.
- Adds two `ddtrace.tracer.trace` blocks inside `otel_export`: `proto_to_calls` (around the `Span.from_proto` / `span.to_call` loop) and `build_rows` (around the `*_to_ch_insertable` + `ch_call_to_row` list-build before insert). Both tagged with row/call counts so flamegraph clicks land on actionable counts.
- Example trace: https://us5.datadoghq.com/apm/trace/6a064604000000009517be91c0f6ec5c

Pairs with wandb/core PR adding `@ddtrace.tracer.wrap` to `otel_export_deserialize` for the gzip/proto-parse gap before `otel_export` opens.

## Testing
non-functional change. Verified locally with `uvx ruff check` and `uvx ruff format --check`.

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -602,52 +602,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
     @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched.otel_export")
     def otel_export(self, req: tsi.OTelExportReq) -> tsi.OTelExportRes:
         assert_non_null_wb_user_id(req)
-        calls: list[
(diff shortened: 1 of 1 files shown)


rgao-coreweave (approved):
Nice, thx for fixing.


LGTM

PR #6863, opened 2026-05-15
feat(weave): add grouped span custom attr distributions

bcsherma:
## Description

Extends grouped agent span queries so each returned group can include requested value distributions alongside the normal grouped result data. This lets callers fetch a page of grouped spans and, for each visible group, also get compact distributions for selected fields.

The immediate use case is the agents UI: when grouping by conversation, turn, or similar span dimensions, we need to render custom attribute summaries in tables and filters without adding a separate conversation-specific API path.

Supported distribution shapes:

- numeric values return histogram bins
- string and bool values return top value counts plus other/missing counts
- distributions are keyed by caller-provided aliases on each grouped span row

## Testing

- `uv run pytest tests/trace_server/test_genai_agent_queries.py -k conversation_custom_attr_distributions`
- `uv run prek run --hook-stage=pre-push ruff-check --files weave/trace_server/agents/clickhouse.py tests/trace_server/test_genai_agent_queries.py`
- `uv run prek run --hook-stage=pre-push mypy --files weave/trace_server/agents/clickhouse.py tests/trace_server/test_genai_agent_queries.py`
- `uv run prek run --hook-stage=pre-push ty --files weave/trace_server/agents/clickhouse.py tests/trace_server/test_genai_agent_queries.py`
- `nox -e lint`

### DIFF
--- a/weave/trace_server/agents/clickhouse.py
+++ b/weave/trace_server/agents/clickhouse.py
@@ -29,9 +29,11 @@ from weave.trace_server.agents.schema import (
     AgentSpanCHInsertable,
 )
 from weave.trace_server.agents.types import (
+    AGENT_CUSTOM_ATTR_SOURCE_VALUE_TYPES,
     AgentConversationChatReq,
     AgentConversationChatRes,
     AgentCustomAttrSchemaItem,
+    AgentCustomAttrSource,
     AgentCustomAttrsSchemaReq,
     AgentCustomAttrsSchemaRes,
     AgentSchema,
@@ -39,6 +41,9 @@ from weave.trace_server.agents.types import (
     AgentSearchMatchedMessage,
     AgentSearchReq,
     AgentSearchRes,
+    AgentSpanGroupDistributionBin,
+    AgentSpanGroupDistributionItem,
+    AgentSpanGroupDistributionValue,
     AgentSpanGroupRow,
     AgentSpanSchema,
     AgentSpansQueryReq,
@@ -71,11 +76,16 @@ from weave.trace_server.query_builder.agent_query_builder import (
     make_conversation_chat_turns_count_query,
     make_custom_attrs_schema_query,
     make_message_search_query,
+    make_span_group_categorical_distributions_query,
+    make_span_group_distribution_counts_query,
+    make_span_group_numeric_distributions_query,
     make_spans_count_query,
     make_spans_list_query,
     make_trace_detail_spans_query,
+    safe_float,
     safe_int,
     safe_str,
+    span_group_distribution_key,
 )
 from weave.trace_server.query_builder.agent_stats_query_builder import (
     build_agent_span_stats_query,
@@ -151,6 +161,8 @@ class AgentQueryHandler:
         aliases = [group_by_ref_alias(ref) for ref in req.group_by]
         measure_aliases = [measure.alias for measure in req.measures]
         groups = [_hydrate_group_row(r, aliases, measure_aliases) for r in rows]
+        if req.group_distributions:
+            self._hydrate_group_distributions(req, groups, aliases[0])
         return AgentSpansQueryRes(groups=groups, total_count=total)
 
     def spans_stats(self, req: AgentSpanStatsReq) -> AgentSpanStatsRes:
@@ -190,6 +202,108 @@ class AgentQueryHandler:
             offset=req.offset,
             has_more=len(rows) > req.limit,
         )
+
+    def _hydrate_group_distributions(
+        self,
+        req: AgentSpansQueryReq,
+        groups: list[AgentSpanGroupRow],
+        group_alias: str,
+    ) -> None:
+        """Attach requested custom attribute distributions to returned groups."""
+        # TODO: Split item setup and row hydration into smaller testable helpers.
+        if not groups:
+            return
+
+        groups_by_key = {
+            span_group_distribution_key(group.group_keys.get(group_alias)): group
+            for group in groups
+        }
+        group_values = [group.group_keys.get(group_alias) for group in groups]
+        pb = ParamBuilder(PARAM_NAMESPACE)
+        counts_sql = make_span_group_distribution_counts_query(pb, req, group_values)
+        total_counts = {
+            safe_str(row.get("group_key")): safe_int(row.get("total_count"))
+            for row in _rows_as_dicts(self._query(counts_sql, pb.get_params()))
+        }
+
+        items: dict[tuple[str, str], AgentSpanGroupDistributionItem] = {}
+        for group_key, group in groups_by_key.items():
+            for spec in req.group_distributions:
+                total_count = total_counts.get(group_key, 0)
+                source = cast("AgentCustomAttrSource", spec.value.source)
+                item = AgentSpanGroupDistributionItem(
+                    alias=spec.alias,
+                    source=source,
+                    key=spec.value.key,
+                    value_type=AGENT_CUSTOM_ATTR_SOURCE_VALUE_TYPES[source],
+                    total_count=total_count,
+                    # Start at total_count; distribution rows below subtract
+                    # present values for groups where the custom attr exists.
+                    missing_count=total_count,
+                )
+                items[group_key, spec.alias] = item
+                group.distributions[spec.alias] = item
+
+        numeric_specs = [
+            spec
+            for spec in req.group_distributions
+            if spec.value.source in {"custom_attrs_int", "custom_attrs_float"}
+        ]
+        if numeric_specs:
+            pb = ParamBuilder(PARAM_NAMESPACE)
+            sql = make_span_group_numeric_distributions_query(
+                pb, req, group_values, numeric_specs
+            )
+            for row in _rows_as_dicts(self._query(sql, pb.get_params())):
+                key = (safe_str(row.get("group_key")), safe_str(row.get("alias")))
+                distribution_item = items.get(key)
+                if distribution_item is None:
+                    continue
+                distribution_item.present_count = safe_int(row.get("present_count"))
+                distribution_item.missing_count = max(
+                    0,
+                    distribution_item.total_count - distribution_item.present_count,
+                )
+                distribution_item.bins.append(
+                    AgentSpanGroupDistributionBin(
+                        index=safe_int(row.get("bucket_index")),
+                        min=safe_float(row.get("bucket_min")),
+                        max=safe_float(row.get("bucket_max")),
+                        count=safe_int(row.get("count")),
+                    )
+                )
+
+        categorical_specs = [
+            spec
+            for spec in req.group_distributions
+            if spec.value.source in {"custom_attrs_string", "custom_attrs_bool"}
+        ]
+        if categorical_specs:
+            pb = ParamBuilder(PARAM_NAMESPACE)
+            sql = make_span_group_categorical_distributions_query(
+                pb, req, group_values, categorical_specs
+            )
+            for row in _rows_as_dicts(self._query(sql, pb.get_params())):
+                key = (safe_str(row.get("group_key")), safe_str(row.get("alias")))
+                distribution_item = items.get(key)
+                if distribution_item is None:
+                    continue
+                distribution_item.present_count = safe_int(row.get("present_count"))
+                distribution_item.missing_count = max(
+                    0,
+                    distribution_item.total_count - distribution_item.present_count,
+                )
+                distribution_item.values.append(
+                    AgentSpanGroupDistributionValue(
+                        value=safe_str(row.get("value")),
+                        count=safe_int(row.get("count")),
+                    )
+                )
+
+        for item in items.values():
+            if item.values:
+                top_count = sum(value.count for value in item.values)
+                item.other_count = max(0, item.present_count - top_count)
 
     # ------------------------------------------------------------------
     # AMT-backed agents queries
--- a/weave/trace_server/agents/clickhouse.py
+++ b/weave/trace_server/agents/clickhouse.py
@@ -11,7 +11,7 @@ import datetime
 import logging
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, TypeAlias, TypeVar, cast
+from typing import TYPE_CHECKING, Any, NamedTuple, TypeAlias, TypeVar, cast
 
 from weave.shared import refs_internal as ri
 from weave.trace_server.agents.chat_view import build_trace_chat
@@ -33,7 +33,6 @@ from weave.trace_server.agents.types import (
     AgentConversationChatReq,
     AgentConversationChatRes,
     AgentCustomAttrSchemaItem,
-    AgentCustomAttrSource,
     AgentCustomAttrsSchemaReq,
     AgentCustomAttrsSchemaRes,
     AgentSchema,
@@ -60,6 +59,7 @@ from weave.trace_server.agents.types import (
     AgentVersionsQueryRes,
     GenAIOTelExportReq,
     GenAIOTelExportRes,
+    group_by_ref_alias,
 )
 from weave.trace_server.datadog import record_db_insert
 from weave.trace_server.opentelemetry.genai_extraction import extract_genai_span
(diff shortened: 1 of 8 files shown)


bcsherma:
* **#6863** <a href="https://app.graphite.com/github/pr/wandb/weave/6863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6863?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -710,6 +722,20 @@ def ensure_group_filters_match(
             )
 
 
+def ensure_group_measure_aliases_do_not_collide(
```
jtschoonhoven: Tiny nit: could be `_` prefixed to mark as private

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -710,6 +722,20 @@ def ensure_group_filters_match(
...
+) -> None:
+    """Reject dynamic measure aliases that would overwrite grouped row fields."""
+    reserved = SPAN_GROUP_RESULT_COLS.union(frozenset(group_aliases))
+    collisions = sorted(
+        {measure.alias for measure in measures if measure.alias in reserved}
+    )
+    if collisions:
+        raise ValueError(
+            f"measure aliases collide with grouped row fields: {collisions!r}"
+        )
```
jtschoonhoven, line 724: Does this just cause the request to 500? Could we fail more gracefully than that?

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -991,6 +1020,184 @@ def make_custom_attrs_schema_query(
     """
 
 
+def span_group_distribution_key(value: Any) -> str:
+    """Normalize a returned group key for page-scoped distribution lookups."""
+    return "" if value is None else str(value)
```
jtschoonhoven: Nit: this is a lot of high-falootin' language for a wrapper around `str()`

```suggestion
def to_str(value: Any) -> str:
    """Call `str(value)` or return an empty string for None."""
    return "" if value is None else str(value)
```

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -991,6 +1020,184 @@ def make_custom_attrs_schema_query(
...
+
+
+def _span_group_distribution_context(
+    pb: ParamBuilder,
+    req: AgentSpansQueryReq,
+    group_values: Sequence[Any],
+) -> tuple[str, str]:
```
jtschoonhoven: Nit: `tuple[str, str]` is easy to mess up, how about returning a NamedTuple or dataclass?

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -991,6 +1020,184 @@ def make_custom_attrs_schema_query(
...
 
+def span_group_distribution_key(value: Any) -> str:
+    """Normalize a returned group key for page-scoped distribution lookups."""
+    return "" if value is None else str(value)
+
+
+def _span_group_distribution_context(
+    pb: ParamBuilder,
+    req: AgentSpansQueryReq,
+    group_values: Sequence[Any],
```
jtschoonhoven: I _think_ you can (and should) type this more specifically than `Any` here. AFAICT these values come from `AgentSpanGroupDistributionItem` which is typed, so I think you can narrow the type here.

weave/trace_server/agents/clickhouse.py
```diff
@@ -190,6 +202,95 @@ def custom_attrs_schema(
             has_more=len(rows) > req.limit,
         )
 
+    def _hydrate_group_distributions(
+        self,
+        req: AgentSpansQueryReq,
+        groups: list[AgentSpanGroupRow],
+        group_alias: str,
+    ) -> None:
+        """Attach requested custom attribute distributions to returned groups."""
```
jtschoonhoven, line 218: This function is too complicated (several nested `for` loops with conditional logic). I think it's worth splitting these inner loops out into their own functions. Easier to test and review.

weave/trace_server/agents/clickhouse.py
```diff
@@ -190,6 +202,95 @@ def custom_attrs_schema(
...
+                items[group_key, spec.alias] = item
+                group.distributions[spec.alias] = item
+
+        for spec in req.group_distributions:
+            pb = ParamBuilder(PARAM_NAMESPACE)
+            if spec.value.source in {"custom_attrs_int", "custom_attrs_float"}:
+                sql = make_span_group_numeric_distribution_query(
+                    pb, req, group_values, spec
+                )
+                for row in _rows_as_dicts(self._query(sql, pb.get_params())):
```
jtschoonhoven: Probably not a good idea query ClickHouse in a synchronous `for` loop. At a minimum these should run in parallel. Better would be to find a way to merge the queries.

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -991,6 +1020,184 @@ def make_custom_attrs_schema_query(
...
+                   value_count,
+                   sum(value_count) OVER (PARTITION BY group_key) AS present_count,
+                   row_number() OVER (
+                     PARTITION BY group_key
+                     ORDER BY value_count DESC, raw_value ASC
+                   ) AS value_rank
+            FROM value_counts
+          )
+        SELECT group_key,
+               substring(raw_value, 1, 256) AS value,
```
jtschoonhoven, line 1292: Nit: I think in general it's better to truncate _after_ grouping. But 256 chars is pretty generous so I doubt it will matter in practice.

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -991,6 +1020,184 @@ def make_custom_attrs_schema_query(
...
+    """
+
+
+def make_span_group_numeric_distribution_query(
+    pb: ParamBuilder,
+    req: AgentSpansQueryReq,
+    group_values: Sequence[Any],
+    spec: AgentSpanGroupDistributionSpec,
+) -> str:
+    """Build per-group histograms for one numeric custom attribute."""
```
jtschoonhoven: This is kinda complicated and AFAICT there's not much test coverage. Probably worth a unit test at least.

weave/trace_server/agents/clickhouse.py
```diff
@@ -190,6 +202,95 @@ def custom_attrs_schema(
...
+            for spec in req.group_distributions:
+                total_count = total_counts.get(group_key, 0)
+                source = cast("AgentCustomAttrSource", spec.value.source)
+                item = AgentSpanGroupDistributionItem(
+                    alias=spec.alias,
+                    source=source,
+                    key=spec.value.key,
+                    value_type=AGENT_CUSTOM_ATTR_SOURCE_VALUE_TYPES[source],
+                    total_count=total_count,
+                    missing_count=total_count,
```
jtschoonhoven, line 248: Nit: worth adding a comment to clarify that this value gets modified later. Otherwise it looks weird.

jtschoonhoven (commented):
This looks good, but the synchronous N+1 queries in `_hydrate_group_distributions` are a blocker. Other comments are non-blocking (but recommended)

weave/trace_server/agents/clickhouse.py
bcsherma: oooph yeah I will batch this loop into one query

weave/trace_server/query_builder/agent_query_builder.py
bcsherma: added many more tests

weave/trace_server/agents/clickhouse.py
bcsherma: I added a todo to refactor here

weave/trace_server/agents/clickhouse.py
```diff
@@ -191,6 +203,108 @@ def custom_attrs_schema(
...
+        }
+        group_values = [group.group_keys.get(group_alias) for group in groups]
+        pb = ParamBuilder(PARAM_NAMESPACE)
+        counts_sql = make_span_group_distribution_counts_query(pb, req, group_values)
+        total_counts = {
+            safe_str(row.get("group_key")): safe_int(row.get("total_count"))
+            for row in _rows_as_dicts(self._query(counts_sql, pb.get_params()))
+        }
+
+        items: dict[tuple[str, str], AgentSpanGroupDistributionItem] = {}
```
jtschoonhoven: Nit: a namedtuple would make this a lot more readable. I can't guess what the key represents by looking at the type.

weave/trace_server/agents/clickhouse.py
```diff
@@ -191,6 +203,108 @@ def custom_attrs_schema(
...
+        for group_key, group in groups_by_key.items():
+            for spec in req.group_distributions:
+                total_count = total_counts.get(group_key, 0)
+                source = cast("AgentCustomAttrSource", spec.value.source)
```
jtschoonhoven: Nit: Cast is a code smell. You shouldn't need this.

weave/trace_server/agents/clickhouse.py
```diff
@@ -191,6 +203,108 @@ def custom_attrs_schema(
...
+                    distribution_item.total_count - distribution_item.present_count,
+                )
+                distribution_item.bins.append(
+                    AgentSpanGroupDistributionBin(
+                        index=safe_int(row.get("bucket_index")),
+                        min=safe_float(row.get("bucket_min")),
+                        max=safe_float(row.get("bucket_max")),
+                        count=safe_int(row.get("count")),
+                    )
+                )
```
jtschoonhoven, line 284: Nit: I know you left this as a TODO and it's still not a blocker, but it _does_ seem like it would be pretty easy to factor out these `if` blocks into their own functions. That would also make it easier to run the queries in parallel.

weave/trace_server/agents/clickhouse.py
```diff
@@ -191,6 +203,108 @@ def custom_attrs_schema(
...
+            spec
+            for spec in req.group_distributions
+            if spec.value.source in {"custom_attrs_string", "custom_attrs_bool"}
+        ]
+        if categorical_specs:
+            pb = ParamBuilder(PARAM_NAMESPACE)
+            sql = make_span_group_categorical_distributions_query(
+                pb, req, group_values, categorical_specs
+            )
+            for row in _rows_as_dicts(self._query(sql, pb.get_params())):
```
jtschoonhoven, line 296: Nit: consider adding a TODO to run these two queries in parallel.

weave/trace_server/agents/clickhouse.py
```diff
@@ -191,6 +203,108 @@ def custom_attrs_schema(
...
+                    AgentSpanGroupDistributionValue(
+                        value=safe_str(row.get("value")),
+                        count=safe_int(row.get("count")),
+                    )
+                )
+
+        for item in items.values():
+            if item.values:
+                top_count = sum(value.count for value in item.values)
+                item.other_count = max(0, item.present_count - top_count)
```
jtschoonhoven, line 320: Nit: a comment here would be helpful

weave/trace_server/agents/types.py
```diff
@@ -93,6 +98,7 @@
 )
 
 _IDENT_RE = r"^[a-zA-Z_][a-zA-Z0-9_]*$"
+_FIELD_GROUP_BY_SOURCES = {"field", "column"}
```
jtschoonhoven, line 103: Nit: probably worth adding a comment to explain what these are. Especially since it's easy to misread the constant as "field group, by sources".

weave/trace_server/agents/types.py
```diff
@@ -500,6 +506,60 @@ class AgentGroupByRef(BaseModel):
     alias: str | None = None  # output key in AgentSpanGroupRow.group_keys
 
 
+def _group_by_ref_alias(ref: AgentGroupByRef) -> str:
```
jtschoonhoven: Nit: this function could use a docstring (and maybe a better name). From the name I assumed it was actually grouping some data, not just returning a key. _get_group_by_ref_alias would be more obvious.

EDIT: there's already a nearly-identical function in `agent_query_builder.py` named `group_by_ref_alias`. You should delete one or the other.

weave/trace_server/agents/types.py
```diff
@@ -500,6 +506,60 @@ class AgentGroupByRef(BaseModel):
...
+    def validate_distribution_spec(self) -> AgentSpanGroupDistributionSpec:
+        if self.value.source not in AGENT_CUSTOM_ATTR_SOURCES:
+            raise ValueError("distribution specs must reference custom attr sources")
+        return self
+
+
+class AgentSpanGroupDistributionBin(BaseModel):
+    """One numeric histogram bin for a custom attribute in a span group."""
+
+    index: int
```
jtschoonhoven, line 561: Nit: what's `index`? Maybe it's obvious in the context where this type is used, but I can't guess from just looking at the type. Worth a comment.

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -192,6 +185,14 @@
     AGENT_CUSTOM_ATTR_SOURCE_VALUE_TYPES.items()
 )
 
+SpanGroupKeyValue = str | int | float | bool | None
```
jtschoonhoven: Nit
```suggestion
type SpanGroupKeyValue = str | int | float | bool | None
```

Alternatively you could maybe call this type `Scalars`

weave/trace_server/query_builder/agent_query_builder.py
jtschoonhoven: Bump

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -997,6 +1015,319 @@ def make_custom_attrs_schema_query(
...
+def make_span_group_categorical_distribution_query(
+    pb: ParamBuilder,
+    req: AgentSpansQueryReq,
+    group_values: Sequence[SpanGroupKeyValue],
+    spec: AgentSpanGroupDistributionSpec,
+) -> str:
+    """Build per-group top value counts for one categorical custom attr."""
+    return make_span_group_categorical_distributions_query(
+        pb, req, group_values, [spec]
+    )
```
jtschoonhoven: Pointless wrapper function
```suggestion
```

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -997,6 +1015,319 @@ def make_custom_attrs_schema_query(
...
+
+
+def make_span_group_numeric_distribution_query(
+    pb: ParamBuilder,
+    req: AgentSpansQueryReq,
+    group_values: Sequence[SpanGroupKeyValue],
+    spec: AgentSpanGroupDistributionSpec,
+) -> str:
+    """Build per-group histograms for one numeric custom attribute."""
+    return make_span_group_numeric_distributions_query(pb, req, group_values, [spec])
```
jtschoonhoven: Pointless wrapper function
```suggestion
```

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -997,6 +1015,319 @@ def make_custom_attrs_schema_query(
...
+    pb: ParamBuilder,
+    specs: Sequence[AgentSpanGroupDistributionSpec],
+    *,
+    limit_attr: str,
+) -> str:
+    spec_tuples = [
+        _span_group_distribution_spec_tuple_sql(
+            pb,
+            spec,
+            limit_value=cast(int, getattr(spec, limit_attr)),
```
jtschoonhoven: Nit: you should almost never need to use cast. Here there's no reason you can't verify the type returned by getattr using isinstance.

Personally I'd add a method on AgentSpanGroupDistributionSpec like `def get_limit_value(self) -> int:`

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -997,6 +1015,319 @@ def make_custom_attrs_schema_query(
...
+    key_slot = pb.add(spec.value.key, param_type="String")
+    limit_slot = pb.add(limit_value, param_type="UInt64")
+    return f"tuple({alias_slot}, {source_slot}, {key_slot}, {limit_slot})"
+
+
+def _span_group_distribution_specs_array_sql(
+    pb: ParamBuilder,
+    specs: Sequence[AgentSpanGroupDistributionSpec],
+    *,
+    limit_attr: str,
```
jtschoonhoven:
```suggestion
    limit_attr: Literal["bins", "top_n"],
```

...though really it's a code smell that you're passing in the name of an attribute to use with `getattr`. Better to avoid that in the first place. See the suggestion in the next comment for example.

jtschoonhoven (approved):
Dope. Nits only.

PR #6864, opened 2026-05-15
chore(weave): trim spans on upsert_batch hot path

gtarpenning:
## Summary

**Motivation**: prod is downsampling apm spans due to volume. let's clean up unnecessary stuff so we can trace more important things. 

Reduces Datadog span-byte volume on the hottest `/call/upsert_batch` path by dropping spurious per-item spans.

- `ttl_settings.get_project_retention_days`: stop spanning L1 cache hits (~99% of calls, 16 µs p50). Restores `ttl_settings.get_project_retention_days` wrapping the miss path only. Parent span still gets the `ttl.cache_hit:L1` tag.
- `process_call_req_to_content` / `process_complete_call_to_content`: drop decorators. Real work (`store_content_object`) is already separately spanned.
- `_insert_file_chunks`: drop decorator. It's a list-append in batch mode and a passthrough to the already-spanned `_insert` in flush-immediately mode.

Observed per-trace span counts on a typical 10-call `/call/upsert_batch`: 11x `ttl_settings.get_project_retention_days` (all L1 hits, 0.02 ms each) + 11x `process_call_req_to_content` (10 trivial + 1 real). After: 0 + 0 (plus the existing `store_content_object` when content is actually uploaded).

## Testing

no functional changes

### DIFF
--- a/weave/trace_server/ttl_settings.py
+++ b/weave/trace_server/ttl_settings.py
@@ -57,12 +57,8 @@ def get_project_retention_days(
     (argMax). A fall-through to ClickHouse is a full cache miss; if ClickHouse
     has no row for the project, returns RETENTION_DAYS_NO_TTL.
 
-    Only the cache-miss path is wrapped in a tracer span. L1 hits (the vast
-    majority of calls on the hot path) tag the parent span and return, to keep
-    Datadog span volume down.
-
     Redis client is resolved lazily via get_redis_client() (lru_cached
-    process singleton). ch_client must come from the calling thread, it is
+    process singleton). ch_client must come from the calling thread — it is
     a thread-local resource in ClickHouseTraceServer.
     """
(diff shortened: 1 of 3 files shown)


weave/trace_server/ttl_settings.py
```diff
@@ -67,24 +66,24 @@ def get_project_retention_days(
         set_current_span_dd_tags({"ttl.cache_hit": "L1"})
```
nikumar1206, line 66: dont think it matters, but this line will become a no-op right since there is no current trace to tag anymore

nikumar1206 (approved):

weave/trace_server/ttl_settings.py
gtarpenning: hmm i think it will target the root span maybe

PR #6844, opened 2026-05-14
perf(weave): copy-on-write ref conversion in trace_server_converter

gtarpenning:
## Summary
- The trace_server hot path runs every request payload through `universal_ext_to_int_ref_converter`, which cloned every dict / list / tuple / set / pydantic model on every request. Switch to copy-on-write: subtrees with no rewrites return by identity; only paths containing a rewritten ref allocate.
- For ref-free payloads zero new objects are allocated. For payloads with refs, only the path from the root to each rewrite is rebuilt.
- Reworked from the trace_server_converter portion of #6642 with inline review feedback addressed. The base64 piece of that PR landed via #6797; the clickhouse-row piece is up as #6843.

## Review feedback addressed
- Module docstring + per-function docstrings explain the COW strategy (neutralino1, jtschoonhoven)
- `_MapResult` NamedTuple replaces the unnamed `(value, changed, fully_supported)` triple (jtschoonhoven)
- Four near-identical container arms consolidated into three small helpers (`_walk_mapping`, `_walk_sequence` for list+tuple, `_walk_set`); the COW invariant lives in one place per container kind (jtschoonhoven: "logic gets pretty hairy")
- Inner helpers are typed `Any`, so `cast` is needed only once at the TypeVar boundary in `_map_values` (jtschoonhoven question on cast usage)

## Testing
new `tests/trace_server/test_trace_server_converter.py` covers (a) identity preservation on no-op + sibling-subtree-untouched walks, (b) aliased query models with `$`-prefixed fields, (c) the round-trip fallback when an `Any` field contains a nested dataclass. Existing `test_adapter_does_not_mutate_req_when_inner_raises` continues to enforce that the adapter boundary stays mutation-safe (it `model_copy(deep=True)`s before invoking the converter, so the in-place fast path is contained to the deep copy).

PR #6865, opened 2026-05-15
perf(weave): reuse int-to-ext ref mapper across stream items

gtarpenning:
## Summary

Follow-up to #6844. `_stream_ref_apply` was rebuilding the entire `universal_int_to_ext_ref_converter` closure (cache dict + `replace_ref` + `mapper`) on every yielded item, and was wrapping that with its own stream-scoped `cached_int_to_ext_project_id` block, so each item paid for a redundant pair of caches and three closure allocations.

- New `make_int_to_ext_ref_mapper(resolver) -> Callable` builds the closures and project_id cache once and returns a reusable applier.
- `universal_int_to_ext_ref_converter` is now a 3-line wrapper around it, so non-streaming callers (`_ref_apply` and every read/write method) are unchanged.
- `_stream_ref_apply` builds the mapper once per stream, yields `mapper(item)` per item, and drops the redundant cache block.

## Performance

Local micro-bench, stream-discard mode (median of 7 runs):

| payload | items | cpu (old -> new) | peak mem (old -> new) |
|---|---|---|---|
| ref-heavy, 64 projects x 8 refs | 1000 | 9.88 -> 8.77ms (-11.2%) | 11.4 -> 10.3KB (-9.9%) |
| ref-heavy, 1 project x 4 refs   | 5000 | 28.48 -> 26.03ms (-8.6%) | 22.2 -> 21.9KB (-1.5%) |
| ref-free                        | 5000 | 37.71 -> 36.24ms (-3.9%) | 2.8 -> 2.5KB (-11.9%) |
| nested depth=6                  | 1000 | 12.28 -> 11.43ms (-6.9%) | 36.0 -> 35.5KB (-1.2%) |
| mixed call-end-shaped           | 10000 | 142.74 -> 136.73ms (-4.2%) | 31.2 -> 30.9KB (-1.1%) |

## Testing

unit test on cache reuse across N items (resolver hit exactly once per distinct project_id, ref-free items preserve identity, cached `None` survives across items, malformed ref raises without poisoning the cache). adapter mutation contract tests still pass.

### DIFF
--- a/tests/trace_server/test_trace_server_converter.py
+++ b/tests/trace_server/test_trace_server_converter.py
@@ -257,3 +257,27 @@ def test_make_int_to_ext_ref_mapper_shares_cache_across_items():
     out_a3 = mapper(item_a3)
     assert out_a3 == {"ref": f"weave:///{ext_a}/op/qux:v4"}
     assert resolver_calls == [int_a, int_b, int_private]
+
+    # Item 7: BaseModel input. _map_values walks model fields via COW;
+    # ref-bearing model rewrites the ref and returns a (re-validated) new
+    # instance, while a ref-free model returns its input by identity.
+    # Resolver remains untouched on the cached hit.
+
+    class Item(BaseModel):
+        op_name: str
+        attributes: dict[str, str]
+
+    model_with_ref = Item(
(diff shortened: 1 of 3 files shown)


gtarpenning:
closing - perf gain is too marginal to justify the diff

PR #6866, opened 2026-05-15
fix(weave): `set opentelemetry*>=1.28.0` to fix `import weave` if you `pip install wandb weave`

kelu-wandb:
## Description

- Fixes https://coreweave.atlassian.net/browse/WB-34375

(Re-submitting PR #6856 which was reverted in PR #6856, except being more lenient about opentelemetry version.)

This PR sets a minimum version of `opentelemetry` so we don't end up with disjoint versions that are mutually incompatible if a too-new version of `protobuf` is already installed, e.g. by `wandb`.

Also preps `weave` for being depended on by `wandb`.

### Version Selection

1.28.0 is the minimum version that is compatible with wandb, as tested by forcing:

`opentelemetry-proto==1.28.0`

in `wandb/wandb` repo `requirements/requirements_dev.txt` and running `requirements/generate.sh`.

1.27.0 threw an error.

### Steps to reproduce:

```
uv venv
uv pip install wandb==0.27.0 weave==0.52.40
uv run python -c 'import weave'
```
This throws an error:

```
    _descriptor.FieldDescriptor(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
...
TypeError: Descriptors cannot be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
```

Workaround is to require a minimum version of `opentelemetry`:

```
uv venv
uv pip install weave==0.52.40 wandb==0.27.0 "opentelemetry-proto>=1.28.0"
uv run python -c 'import weave'
```
This passes.  (So does ordering `weave` before `wandb`.)

The cause is that wandb just requires `protobuf < 8`, so when you install it first, you end up with:

```
opentelemetry-api                      1.41.1
opentelemetry-exporter-otlp-proto-http 1.11.1
opentelemetry-proto                    1.11.1
opentelemetry-sdk                      1.41.1
protobuf                               7.34.1
```

Newer versions of `opentelemetry-proto` have a minimum `protobuf` version requirement, whereas version `1.11.1` doesn't, so we end up with that very old version.

The mismatch between different `opentelemetry` components then causes the error.

If you pin a minimum `opentelemetry` version or install `weave` before `wandb`, you get:

```
opentelemetry-api                        1.41.1
opentelemetry-exporter-otlp-proto-common 1.41.1
opentelemetry-exporter-otlp-proto-http   1.41.1
opentelemetry-proto                      1.41.1
opentelemetry-sdk                        1.41.1
protobuf                                 6.33.6
```

### With wandb release 0.27.0

```
uv venv
uv pip install wandb==0.27.0 -e ~/wandb/weave
uv run python -c 'import weave'
```

Now no longer throws an error, and this is what gets installed:

```
opentelemetry-api                        1.41.1
opentelemetry-exporter-otlp-proto-common 1.41.1
opentelemetry-exporter-otlp-proto-http   1.41.1
opentelemetry-proto                      1.41.1
opentelemetry-sdk                        1.41.1
protobuf                                 6.33.6
```

### With local wandb that depends on weave

I temporarily modified my local wandb with:

```
dependencies = [
...
    "weave>=0.52.40",
]
...
[tool.uv.sources]
weave = { path = "../weave", editable = true }
```

and then:

```
uv pip install -e ~/wandb/wandb
```

works and gives the same list of packages as above. with 1.41.1 of opentelemetry packages.
Note that protobuf is being dowgraded from 7.34.1 though. Confirming with SDK team on whether that's acceptable.

Also, in `wandb/wandb`, I ran `requirements/generate.sh` and then temporarily modified `requirements/requirements_dev.3.10.darwin.txt` to replace `.` and `weave==0.52.40` with `-e ${HOME}/wandb/wandb` and `-e ${HOME}/wandb/weave` respectively, and confirmed that

```
uv pip install -r ~/wandb/wandb/requirements/requirements_dev.3.10.darwin.txt
```

passed. (That is part of `wandb/wandb` CI.)

To verify that 1.28.0 would really work if some other dependency needed it:

```
rm -rf /tmp/v1 && uv venv /tmp/v1 && VIRTUAL_ENV=/tmp/v1 uv pip install -e ~/wandb/wandb "opentelemetry-proto==1.28.0"
```

`import weave` still works with that, and the versions we get are:

```
opentelemetry-api                        1.28.2
opentelemetry-exporter-otlp-proto-common 1.28.0
opentelemetry-exporter-otlp-proto-http   1.28.0
opentelemetry-proto                      1.28.0
opentelemetry-sdk                        1.28.2
protobuf                                 5.29.6
```

So I feel confident that this will work, and allow `wandb` to add a dependency on `weave` as planned as well.


rgao-coreweave (approved):
LGTM

PR #6871, opened 2026-05-18
fix(weave): resolve recursive LiteralOperation schema to fix Mintlify warnings

### LINKED ISSUE #5619: OpenAPI spec contains recursive schemas that cause documentation build warnings in Mintlify
## Description

The Weave service API OpenAPI specification contains recursive schema definitions that cause warnings when processed by documentation generators like Mintlify. While the API documentation is still generated successfully, the warnings create noise in the build output.

## Problem Details

The OpenAPI spec at `docs/docs/reference/service-api/openapi.json` contains MongoDB-style query operators with recursive schema definitions that are challenging for documentation parsers.

### Specific Issue

The `LiteralOperation` schema has a recursive `additionalProperties` that references itself:

```json
{
  "properties": {
    "$literal": {
      "anyOf": [
        {
          "type": "string"
        },
        {
          "type": "integer"
        },
        {
          "type": "number"
        },
        {
          "type": "boolean"
        },
        {
          "additionalProperties": {
            "$ref": "#/components/schemas/LiteralOperation"  // Recursive self-reference
          },
          "type": "object"
        },
        {
          "items": {
            "$ref": "#/components/schemas/LiteralOperation"  // Recursive self-reference
          },
          "type": "array"
        },
        {
          "type": "null"
        }
      ],
      "title": "$Literal"
    }
  },
  "type": "object",
  "required": ["$literal"],
  "title": "LiteralOperation"
}
```

## Build Warnings

When Mintlify processes this OpenAPI spec in the W&B documentation, it produces these warnings:

```
File weave/reference/service-api/openapi.json not found in openApiDataFiles
#/paths/\/cost/purge/post/requestBody/content/application\/json/schema/properties/query/properties/$expr/properties/$and/items/properties/$literal/additionalProperties
error converting additionalProperties schema Error [ConversionError]: missing schema definition
```

Similar warnings occur for other recursive operations like `$convert`.

## Impact

- **Functionality**: API documentation is still generated and works correctly ✅
- **Developer Experience**: Build output is noisy with repeated warnings ⚠️
- **No User Impact**: End users can access all API documentation normally ✅

## Suggested Solutions

1. **Simplify Recursive Schemas**: Modify the OpenAPI generation to avoid infinite recursion
   - Use a maximum depth limit for recursive types
   - Replace `additionalProperties: {"$ref": "self"}` with `additionalProperties: true` or a simpler type
   
2. **Generate Documentation-Specific OpenAPI**: Create a separate, simplified OpenAPI spec specifically for documentation that avoids these complex patterns while maintaining the full spec for actual API usage

3. **Use OpenAPI 3.1 Features**: Consider if newer OpenAPI features could represent these schemas without recursion

## Technical Context

- The OpenAPI spec is valid according to OpenAPI 3.1.0 standards
- The issue is with documentation generators' ability to handle recursive schemas
- This pattern is common in query languages (MongoDB, GraphQL, etc.)
- FastAPI is generating these schemas from Python type definitions

## Related Files

- Source of OpenAPI generation: Likely in the Weave trace server FastAPI routes
- Affected schemas: `LiteralOperation`, `ConvertOperation`, and other query operators
- Documentation repository: `wandb/docs` (where the warnings appear)

## Environment

- OpenAPI Version: 3.1.0
- Documentation Tool: Mintlify
- API Framework: FastAPI

Would it be possible to adjust the schema generation to avoid these recursive patterns while maintaining the same API functionality?

rajkumar-prog:
## Problem
Fixes #5619

The `LiteralOperation` schema in `sdks/node/weave.openapi.json` contains recursive self-references (`$ref: "#/components/schemas/LiteralOperation"`) that cause Mintlify documentation build warnings:

```
error converting additionalProperties schema Error [ConversionError]: missing schema definition
```

## Fix
Replaced the recursive `$ref` self-references inside `LiteralOperation` with:
- `additionalProperties: true` for the object case (allows any value)
- `items: {}` for the array case (allows any item)

This removes the infinite recursion while keeping the schema valid and the API behaviour unchanged.

## Changes
- `sdks/node/weave.openapi.json` — 2 lines changed inside `LiteralOperation` schema definition

### DIFF
--- a/tests/trace/test_ref_trace.py
+++ b/tests/trace/test_ref_trace.py
@@ -3,6 +3,88 @@ from dataclasses import FrozenInstanceError
 import pytest
 
 import weave
+from weave.trace.refs import (
+    AgentConversationRef,
+    AgentSpanRef,
+    AgentTurnRef,
+    CallRef,
+    ObjectRef,
+    OpRef,
+    TableRef,
+)
+
+
+@pytest.mark.parametrize(
+    ("ref", "expected_uri"),
+    [
+        (
+            ObjectRef(entity="e", project="p", name="obj", _digest="dig", _extra=()),
+            "weave:///e/p/object/obj:dig",
+        ),
+        (
+            OpRef(entity="e", project="p", name="op", _digest="dig", _extra=()),
+            "weave:///e/p/op/op:dig",
+        ),
+        (
+            CallRef(entity="e", project="p", id="call-id", _extra=()),
+            "weave:///e/p/call/call-id",
+        ),
+        (
+            TableRef(entity="e", project="p", _digest="dig"),
+            "weave:///e/p/table/dig",
+        ),
+        (
+            AgentTurnRef(entity="e", project="p", trace_id="t"),
+            "weave:///e/p/agent_turn/t",
+        ),
+        (
+            AgentConversationRef(entity="e", project="p", conversation_id="c"),
+            "weave:///e/p/agent_conversation/c",
+        ),
+        (
+            AgentSpanRef(entity="e", project="p", span_id="s"),
+            "weave:///e/p/agent_span/s",
+        ),
+    ],
+)
+def test_str_returns_uri(ref, expected_uri):
+    assert str(ref) == expected_uri
+    assert f"{ref}" == expected_uri
+    assert f"{ref!s}" == expected_uri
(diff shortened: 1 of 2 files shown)


rajkumar-prog:

PR #6873, opened 2026-05-18
fix(weave): release failed obj_create payloads from user objects

gtarpenning:
## Summary
- WB-31070: a final failure on `obj/create` or `table/create` (e.g. 502 after retries) left the failed `digest_future` attached to the user's object via `obj.ref._digest`. The future stores its exception with `__traceback__`, which pins the serialized payload (`json_val` / table rows) via frame locals for the lifetime of the user's object.
- Adds `clear_refs_on_failure(digest_future, cleanup)` in `ref_util` and registers it at three sites:
  - `_save_object_basic`: clears `orig_val.ref` on failure (covers dict, raw `ObjectRecord`, op saves).
  - `_save_nested_objects` pydantic Object case: clears the user-visible pydantic `obj.ref` (the inner `_save_object_basic` call only saw `obj_rec`, so the outer obj kept pinning).
  - `_save_table`: clears `table.ref` and (for `WeaveTable`) `table.table_ref`.
- `set_ref` refactored alongside: drops the duplicate middle branch, replaces the bare `except:` with `except Exception:`, swallow path now emits a debug log.

Jira: [WB-31070](https://coreweave.atlassian.net/browse/WB-31070)

## Testing
parametrized regression test in `test_tracing_resilience.py` exercising `client.save(pydantic_obj)` and `client.save(Table)` against the throwing-server fixture. asserts the ref is cleared on failure and that `weakref(obj)` dies after `del obj; gc.collect()`. sanity-checked by reverting each cleanup site and watching the corresponding test parameter fail.

[WB-31070]: https://coreweave.atlassian.net/browse/WB-31070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/trace/test_tracing_resilience.py
```diff
@@ -81,6 +84,48 @@ def simple_op():
...
+    obj = make_obj()
+    obj_weak = weakref.ref(obj)
+
+    try:
+        client_with_throwing_server.save(obj, name="payload")
+    except Exception:
+        pass
+    client_with_throwing_server.future_executor.flush()
+
+    assert getattr(obj, "ref", None) is None
```
andrewtruong: Why are you doing a getattr check?

weave/trace/ref_util.py
```diff
@@ -26,13 +31,36 @@ def set_ref(obj: Any, ref: Ref | None) -> None:
...
+        try:
+            obj.__dict__["ref"] = ref
+        except Exception:
+            logger.debug(
+                "set_ref: could not attach ref to object of type %s",
+                type(obj).__name__,
+            )
+            raise ValueError(
+                f"Failed to set ref on object of type {type(obj)}"
+            ) from None
```
andrewtruong: Pull this out into a func.  Nesting like this is dangerous and harder to reason about

gtarpenning: sure, although tbf i'm not adding nesting

andrewtruong (approved):

PR #6881, opened 2026-05-19
fix(weave): allow linking all spans to a single eval call

seanmor5:
The previous implementation assumes single span to single call, but we should actually capture all spans. This accumulates spans under a call and links them all. The frontend can then handle the returned list of spans appropriately.

A few notes:

* The assumption here is that `EvalLinkSpanProcessor` is attached _once_ to avoid duplicate spans getting attached. We don't do anything to enforce ref uniqueness so we can potentially store duplicate refs on an eval call. I think this okay though because the frontend can be smart about this and we don't have to worry about making the span processor more complex than it needs to be.
* This keeps the existing `genai_span_ref` attribute name, but we can just go with plural for the breaking change. I think I'm the only person with linked calls and it would basically just break the link in the frontend (e.g. it wouldn't show up). As-is this still has some backwards compatibility things which allow for singular (non-list) refs.

### DIFF
--- a/tests/trace_server/test_eval_results_genai_span_ref.py
+++ b/tests/trace_server/test_eval_results_genai_span_ref.py
@@ -59,10 +59,52 @@ def test_build_eval_rows_returns_genai_span_ref_without_children() -> None:
     )
 
     trial = rows[0].evaluations[0].trials[0]
-    assert trial.genai_span_ref == tsi.GenAISpanRef.model_validate(_genai_span_ref())
+    assert trial.genai_span_ref == [tsi.GenAISpanRef.model_validate(_genai_span_ref())]
 
 
-def test_build_trial_prefers_prediction_genai_span_ref() -> None:
+def test_eval_results_trial_wraps_legacy_single_genai_span_ref() -> None:
+    trial = tsi.EvalResultsTrial.model_validate(
+        {
+            "predict_and_score_call_id": "pas-1",
+            "genai_span_ref": _genai_span_ref(),
+        }
+    )
+
+    assert trial.genai_span_ref == [tsi.GenAISpanRef.model_validate(_genai_span_ref())]
+
+
+def test_build_eval_rows_returns_genai_span_ref_list_without_children() -> None:
+    predict_and_score_call = _call(
+        call_id="pas-1",
+        attributes={
+            constants.WEAVE_ATTRIBUTES_NAMESPACE: {
+                constants.GENAI_SPAN_REF_ATTR_KEY: [
+                    _genai_span_ref("agent-trace-1"),
+                    _genai_span_ref("agent-trace-2"),
+                ],
+            }
+        },
+        inputs={"example": {"x": 1}, "model": "model://agent"},
+        output={"output": "result", "scores": {}},
+    )
+
+    rows = eval_helpers.build_eval_rows(
+        [predict_and_score_call],
+        ["eval-1"],
+        {"pas-1": "row-1"},
+        include_raw_data_rows=False,
+        include_predict_and_score_children=False,
+    )
+
+    trial = rows[0].evaluations[0].trials[0]
+    assert trial.genai_span_ref is not None
+    assert [ref.trace_id for ref in trial.genai_span_ref] == [
+        "agent-trace-1",
+        "agent-trace-2",
+    ]
+
+
+def test_build_trial_combines_prediction_and_parent_genai_span_refs() -> None:
     predict_and_score_call = _call(
         call_id="pas-1",
         attributes={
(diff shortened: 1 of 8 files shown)


tests/trace_server/test_eval_results_genai_span_ref.py
```diff
@@ -59,10 +59,52 @@ def test_build_eval_rows_returns_genai_span_ref_without_children() -> None:
     )
 
     trial = rows[0].evaluations[0].trials[0]
-    assert trial.genai_span_ref == tsi.GenAISpanRef.model_validate(_genai_span_ref())
+    assert trial.genai_span_ref == [tsi.GenAISpanRef.model_validate(_genai_span_ref())]
 
 
-def test_build_trial_prefers_prediction_genai_span_ref() -> None:
+def test_eval_results_trial_wraps_legacy_single_genai_span_ref() -> None:
```
seanmor5: Note: can be removed if it's okay to just break the existing "singular" links (which are probably just on my project)

weave/evaluation/eval.py
```diff
@@ -100,7 +100,14 @@ def _attach_genai_span_ref_to_call_summary(
 
     weave_summary = call.summary[constants.WEAVE_ATTRIBUTES_NAMESPACE]
 
-    weave_summary[constants.GENAI_SPAN_REF_ATTR_KEY] = genai_span_ref.model_dump()
+    new_ref = genai_span_ref.model_dump()
+    existing = weave_summary.get(constants.GENAI_SPAN_REF_ATTR_KEY)
+    if isinstance(existing, list):
```
seanmor5, line 105: Note: can be simplified if okay to go all in on plural (e.g. list) links

weave/trace_server/trace_server_interface.py
```diff
@@ -2845,7 +2855,14 @@ class EvalResultsTrial(BaseModel):
     model_latency_seconds: float | None = None
     total_tokens: int | None = None
     scorer_call_ids: dict[str, str] = Field(default_factory=dict)
-    genai_span_ref: GenAISpanRef | None = None
+    genai_span_ref: list[GenAISpanRef] | None = None
+
+    @field_validator("genai_span_ref", mode="before")
```
seanmor5: Note: can be removed if okay to go all in on plural (e.g. list) links

weave/evaluation/eval.py
neutralino1: This has not shipped yet, ok to make it backward incompatible.

tests/trace_server/test_trace_server_evaluation_apis.py
```diff
@@ -509,7 +509,7 @@ def test_eval_results_query_returns_genai_span_ref_without_children(client):
     assert res.total_rows == 1
     trial = res.rows[0].evaluations[0].trials[0]
     assert trial.predict_call_id is None
-    assert trial.genai_span_ref == genai_span_ref
+    assert trial.genai_span_ref == [genai_span_ref]
```
neutralino1, line 512: Begs the question, should the relationship be persisted the other way around: genai spans having a foreign key to the `predict_and_score` call. Not saying this is the way to go as it would require a new column on the genai_spans table but worth discussing.

Another option is to keep the single reference and do 2 queries: first get the root span (the one ref'd), then get all the children spans of that span.

weave/trace_server/trace_server_interface.py
seanmor5: [8647cd5](https://github.com/wandb/weave/pull/6881/commits/8647cd547c52cb7bc53ba3b0dcb9a18e604641ae)

weave/evaluation/eval.py
seanmor5: [8647cd5](https://github.com/wandb/weave/pull/6881/commits/8647cd547c52cb7bc53ba3b0dcb9a18e604641ae)

tests/trace_server/test_eval_results_genai_span_ref.py
seanmor5: [8647cd5](https://github.com/wandb/weave/pull/6881/commits/8647cd547c52cb7bc53ba3b0dcb9a18e604641ae)

tests/trace_server/test_trace_server_evaluation_apis.py
seanmor5: > Begs the question, should the relationship be persisted the other way around: genai spans having a foreign key to the predict_and_score call. Not saying this is the way to go as it would require a new column on the genai_spans table but worth discussing.

We actually already invert this relationship in the span data: during eval execution we attach the predict-and-score call ID as a span attribute, so spans can be queried directly by eval call without relying only on the call summary ref. We can switch at any point

> Another option is to keep the single reference and do 2 queries: first get the root span (the one ref'd), then get all the children spans of that span.

I’m hesitant to go with the root span route because I think it assumes there is always a single obvious root span. That may not hold if instrumentation creates multiple top-level GenAI spans, if an eval intentionally invokes multiple agents/models (although maybe they'd end up under a single span), or if there is some instrumentation issue which does not connect parent/child spans correctly, so we end up having to guess about grouping/hierarchy. If we collect all we at least have everything present and can decide about hierarchy / grouping later, or at least display to the user that the hiearchy / grouping is broken

neutralino1 (approved):

PR #6884, opened 2026-05-19
perf(weave): fast-path base64 content detection on long low-entropy strings

gtarpenning:
## Summary
- Adds an O(1) entropy guard to `is_base64`. DD trace [`6a0c7d99...975ce4fb10f7c246`](https://us5.datadoghq.com/apm/traces?trace=6a0c7d9900000000975ce4fb10f7c246) shows a 454ms leaf span on `process_complete_call_to_content`: a 50000-char run of `"x"` passes the base64 alphabet regex, then pays for a `Content.from_base64` decode + libmagic mimetype probe whose result is discarded. The single new `if` clause short-circuits before that work.
- Real base64 of binary content has uniformly high entropy in any short window. Low-entropy fillers (`"x" * N`, padding-token runs, repeated placeholders in dataset rows) are the only inputs the guard rejects.

## Performance

Local microbench, 1000 iterations, mocked `file_create`:

| case | master median ns | branch median ns | delta |
|---|---:|---:|---:|
| pathological (`"x" * 50000`) | 499,959 | 1,084 | **-99.8%** |
| chat_history (50KB natural text) | 5,500 | 6,958 | +26% (noise, regex still bails on space) |
| data_uri (real 1x1 PNG) | 159,583 | 162,084 | +1.6% (real path, no change) |

`tracemalloc` peak for the pathological case: 188,754 -> 92,422 bytes.

## Testing
one regression test (`test_low_entropy_long_string_does_not_invoke_storage`) pinning the `"x" * 50000` short-circuit. Existing `test_non_media_base64_strings_not_converted` drops its `is_base64(...)` assertion since the guard now legitimately rejects some short low-entropy strings; the observable round-trip assertion still holds.


gtarpenning:
this is dumb

PR #6885, opened 2026-05-19
perf(weave): honor sort_by=[] as no-sort, stream-aggregate setting

gtarpenning:
## Summary
- Treat `sort_by=None` as default order, `sort_by=[]` as an explicit no-sort signal on the streaming endpoint.
- When `calls_merged` + LIMIT + no order fields, inject `optimize_aggregation_in_order=1` so ClickHouse streams the GROUP BY in `(project_id, id)` order. ORDER BY defeats this, so we only inject when `order_fields` is empty.
- The matching FE auto-engages this when the user filters on heavy fields (`inputs.*`/`output.*`/`summary.*`/`attributes.*`).

## Performance
Bench on local CH (10M rows / 5M ids, 2 events/call + 1% deletes, best of 3, wall / read):

| variant | stream pagination (LIMIT 100) | stream + heavy filter (LIMIT 100) |
|---|---|---|
| PROD baseline | 548ms / 320MB | 694ms / 524MB |
| PROD + setting alone (no no-sort) | 1088ms / 320MB (regresses 2x) | -- |
| no-sort alone | 263ms / 320MB | 345ms / 524MB |
| shipped (no-sort + setting) | 53ms / 37MB | 260ms / 139MB |
| gain vs prod | **10x / 9x** | **2.7x / 3.8x** |

The setting alone (with ORDER BY) regresses because ClickHouse inserts a SortingTransform after the in-order aggregate. The no-sort signal is what unlocks the streaming win.

## Testing
- Parametrized unit tests for the four `sort_by` shapes (None / [] / user / no-limit) against expected order fields and setting injection (`test_clickhouse_calls_query_stream_sort_modes`).
- Matching FE PR: https://github.com/wandb/core/pull/44245

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -227,7 +227,7 @@ from weave.trace_server.orm import ParamBuilder, Row
 from weave.trace_server.project_version.project_version import (
     TableRoutingResolver,
 )
-from weave.trace_server.project_version.types import WriteTarget
+from weave.trace_server.project_version.types import ReadTable, WriteTarget
 from weave.trace_server.query_builder import eval_results_query_builder
 from weave.trace_server.query_builder.annotation_queues_query_builder import (
     make_annotator_progress_insert_query,
@@ -1522,8 +1522,12 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
         if req.query is not None:
             cq.add_condition(req.query.expr_)
 
-        # Sort with empty list results in no sorting
-        if req.sort_by:
+        # sort_by=None -> default order; sort_by=[] -> explicit no sort
+        # (falls through with no order_fields, enabling stream-aggregate-in-order).
+        if req.sort_by is None:
+            cq.add_order("started_at", "asc")
+            cq.add_order("id", "asc")
+        elif len(req.sort_by) > 0:
             for sort_by in req.sort_by:
                 cq.add_order(sort_by.field, sort_by.direction)
             # If user isn't already sorting by id, add id as secondary sort for consistency.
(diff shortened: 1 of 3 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1522,8 +1522,11 @@ def calls_query_stream(self, req: tsi.CallsQueryReq) -> Iterator[tsi.CallSchema]
         if req.query is not None:
             cq.add_condition(req.query.expr_)
 
-        # Sort with empty list results in no sorting
-        if req.sort_by:
+        # sort_by=None -> default order; sort_by=[] -> explicit no sort.
+        if req.sort_by is None:
+            cq.add_order("started_at", "asc")
+            cq.add_order("id", "asc")
+        elif req.sort_by:
```
neutralino1: Something stricter than just truthiness? Like `isinstance(list,` and `len == 0`?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1522,8 +1522,12 @@ def calls_query_stream(self, req: tsi.CallsQueryReq) -> Iterator[tsi.CallSchema]
...
             cq.add_condition(req.query.expr_)
 
-        # Sort with empty list results in no sorting
-        if req.sort_by:
+        # sort_by=None -> default order; sort_by=[] -> explicit no sort
+        # (falls through with no order_fields, enabling stream-aggregate-in-order).
+        if req.sort_by is None:
+            cq.add_order("started_at", "asc")
+            cq.add_order("id", "asc")
+        elif len(req.sort_by) > 0:
```
neutralino1, line 1530: Actually, technically, a simple `else:` would suffice but 🤷

neutralino1 (approved):

PR #6889, opened 2026-05-19
chore(weave): add manual cost overrides and update costs 5/19

jwlee64:
## Description

Introduces a `manual_costs.json` file and a corresponding `fetch_manual_costs()` function to allow manually curated cost overrides for models that litellm and `modelsBegin.json` do not cover or cover incorrectly. Manual costs take precedence over both litellm and modelsBegin sources when merging in `main()`.

New manual costs 

- `gemini-3.1-flash-lite` (via Gemini, Vertex AI, and OpenRouter)

The `fetch_manual_costs()` function gracefully handles missing or malformed files by returning an empty dict and printing a warning, so the pipeline continues running from litellm and modelsBegin sources.

Ran make update_costs and make update_model_providers after adding the manual costs file  
  
Testing

Unit tests were added for `fetch_manual_costs()` covering:

- Missing file returns empty dict without crashing
- Malformed JSON is tolerated and returns empty dict
- Well-formed entries load correctly while malformed entries are skipped
- When a model has multiple historical entries, the most recent one is used
- Manual costs correctly override litellm costs for the same model ID in the `main()` pipeline

Existing `test_main` tests were updated to mock `fetch_manual_costs` to maintain test isolation.


gtarpenning (approved):
what happens if both litellm reports and we manually enter? who wins? probably should be litellm?

jwlee64:
It was the manual costs, but makes sense to me to default to litellm

PR #6899, opened 2026-05-20
[codex] Fix agent GenAI columns and chat refs

bcsherma:
## Description

Fixes the Agent spans/query and chat surfaces for GenAI spans.

- Adds cache token fields to the default and chat/detail projections so cache read/write token values can be fetched back, not just used in stats histograms.
- Extends the agent span schema/query allowlist to cover persisted GenAI typed columns plus `raw_span_dump` when explicitly requested.
- Normalizes ref-like strings inside GenAI OTLP protobuf attributes at the external ingest boundary, so `weave:///...` refs are stored internally as `weave-trace-internal:///...` before ClickHouse insert.
- Adds focused query-builder and converter regression tests.

Root cause:

- Cache token columns and several typed GenAI fields existed in persisted spans/stats but were missing from the span list/detail projection path.
- GenAI OTLP ingest converted the request `project_id` and `wb_user_id`, but did not walk refs embedded inside protobuf `AnyValue` attributes. Those external refs were stored directly in span fields such as `content_refs`. Later, the public response adapter correctly treated external refs in an internal response as invalid and raised `InvalidInternalRef`, surfacing as a 500.

## Testing

- `uv run pytest tests/trace_server/test_trace_server_interface_util.py tests/trace_server/test_trace_server_converter.py tests/trace_server/query_builder/test_agent_query_builder.py`
- Verified new GenAI OTLP ingest stores seeded `content_refs` / `object_refs` as `weave-trace-internal:///...` in local ClickHouse.
- Verified locally against `tilt prod_weavetrace`: `/agents/traces/chat` and `/agents/conversations/chat` returned 200 for newly seeded `bcanfieldsherman/agents-lorem-ipsum` rows.

### DIFF
--- a/tests/trace_server/query_builder/test_agent_query_builder.py
+++ b/tests/trace_server/query_builder/test_agent_query_builder.py
@@ -12,6 +12,7 @@ import pytest
 import sqlparse
 from pydantic import ValidationError
 
+from weave.trace_server.agents.schema import AgentSpanCHInsertable
 from weave.trace_server.agents.types import (
     AgentConversationChatReq,
     AgentCustomAttrsSchemaReq,
@@ -20,6 +21,7 @@ from weave.trace_server.agents.types import (
     AgentSortBy,
     AgentSpanGroupFilter,
     AgentSpanMeasureSpec,
+    AgentSpanSchema,
     AgentSpansQueryReq,
     AgentSpanValueRef,
     AgentsQueryFilters,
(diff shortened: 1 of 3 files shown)


bcsherma:
* **#6899** <a href="https://app.graphite.com/github/pr/wandb/weave/6899?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6899?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #6906, opened 2026-05-20
chore(weave): add scorer_* columns to feedback table (031)

nikumar1206:
## Description

Adds four typed columns to the \`feedback\` table:

- \`scorer_tags\` \`Array(String)\`
- \`scorer_ratings\` \`Map(String), Float64)\`
- \`scorer_reasons\` \`Map(String, String)\`
- \`scorer_confidences\` \`Map(String, Float64)\`

Non-null with empty defaults so existing rows need no backfill. data layer only — no readers or writers touch the columns yet.

This is in effort to enable stronger-typed agent scoring.

## Testing

Applied locally with \`make migrate_local\`; up + down both verified.

PR #6908, opened 2026-05-20
feat(weave): typed scorer feedback columns, agent_monitor type, and query filters

nikumar1206:
## Description

Backend support for the typed scorer feedback schema ([proposal](https://www.notion.so/wandbai/Scorer-Feedback-Schema-Proposal-361e2f5c7ef38013a18bc254a40b43b5)). Builds on the migration in #6906 (already merged). Collapses the former stack of #6908 / #6909 / #6918 into one PR.

**Typed columns wiring**
- ORM gains `array_string`, `map_string_string`, `map_string_float` column types. ClickHouse round-trips them as native list/dict; SQLite stores JSON text and decodes on read.
- `TABLE_FEEDBACK`, `FeedbackCreateReq`, `format_feedback_to_row`, and `FEEDBACK_QUERY_FIELDS` expose the six `scorer_*` columns (`scorer_tags`, `scorer_tag_reasons`, `scorer_tag_confidences`, `scorer_ratings`, `scorer_rating_reasons`, `scorer_rating_confidences`). Fields are non-nullable with empty-container defaults.

**`wandb.agent_monitor` feedback type**
- New feedback type; scorer identity comes from `runnable_ref` (no `.<name>` suffix).
- Validation requires `runnable_ref`, `call_ref`, and `trigger_ref`. The `scorer_*` columns are rejected (when non-empty) on any other feedback type.
- `feedback_replace` validates the replacement payload before purging so a rejected replace can't destroy the existing row.

**Query filters**
- `$contains` on an `Array(String)` column compiles to `has(col, value)` (CH) / `json_each` subquery (SQLite), with case-insensitive support. Falls back to substring search for non-array operands.
- Dotted access on a `Map(String, *)` column resolves to a typed map lookup (`col['key']` on CH guarded by `mapContains`, `json_extract` on SQLite), so missing keys read as NULL instead of CH's `0`/`""` defaults.

## Testing

Unit tests, against both SQLite and ClickHouse backends.

### DIFF
--- a/weave/trace_server/orm.py
+++ b/weave/trace_server/orm.py
@@ -99,9 +99,9 @@ ColumnType = Literal[
     "float",
     # Array(String) in ClickHouse; JSON text in SQLite.
     "array_string",
-    # Map(*, String) in ClickHouse; JSON text in SQLite.
+    # Map(String, String) in ClickHouse; JSON text in SQLite.
     "map_string_string",
-    # Map(*, Float64) in ClickHouse; JSON text in SQLite.
+    # Map(String, Float64) in ClickHouse; JSON text in SQLite.
     "map_string_float",
 ]
 
@@ -193,7 +193,12 @@ class Table:
         if database_type == "sqlite":
             return f"DELETE FROM {self.name}"
 
-    def tuple_to_row(self, tup: tuple, fields: list[str]) -> Row:
+    def tuple_to_row(
+        self,
+        tup: tuple,
+        fields: list[str],
+        database_type: DatabaseType,
+    ) -> Row:
         d = {}
         for i, field in enumerate(fields):
             normalized_field = field[:-5] if field.endswith("_dump") else field
@@ -201,18 +206,28 @@ class Table:
             col_type = self.col_types.get(normalized_field)
             if col_type == "json":
                 d[normalized_field] = json.loads(value)
-            elif col_type in _JSON_TEXT_BACKED_COL_TYPES and isinstance(value, str):
-                # SQLite stores Array/Map columns as JSON text; ClickHouse returns
-                # them as native list/dict already, so only decode on string values.
+            elif col_type in _JSON_TEXT_BACKED_COL_TYPES and database_type == "sqlite":
+                # SQLite stores Array/Map columns as JSON text; ClickHouse
+                # returns them as native list/dict already.
+                if not isinstance(value, str):
+                    raise ValueError(
+                        f"Unexpected value for {normalized_field}: "
+                        f"expected str, got {type(value).__name__}"
+                    )
                 d[normalized_field] = json.loads(value)
             else:
                 d[normalized_field] = value
         return d
 
-    def tuples_to_rows(self, tuples: list[tuple], fields: list[str]) -> Rows:
+    def tuples_to_rows(
+        self,
+        tuples: list[tuple],
+        fields: list[str],
+        database_type: DatabaseType,
+    ) -> Rows:
         rows = []
         for t in tuples:
-            rows.append(self.tuple_to_row(t, fields))
+            rows.append(self.tuple_to_row(t, fields, database_type))
         return rows
 
 
--- a/weave/trace_server/orm.py
+++ b/weave/trace_server/orm.py
@@ -47,7 +47,7 @@ class ParamBuilder:
         param_builder_count += 1
         self._params: dict[str, Any] = {}
         self._prefix = (prefix or f"pb_{param_builder_count}") + "_"
-        self._database_type = database_type
+        self.database_type = database_type
         self._param_to_name: dict[Any, str] = {}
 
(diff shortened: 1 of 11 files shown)


nikumar1206:
* **#6918** <a href="https://app.graphite.com/github/pr/wandb/weave/6918?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6909** <a href="https://app.graphite.com/github/pr/wandb/weave/6909?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6908** <a href="https://app.graphite.com/github/pr/wandb/weave/6908?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6908?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6906** <a href="https://app.graphite.com/github/pr/wandb/weave/6906?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/orm.py
```diff
@@ -97,8 +97,23 @@ def get_params(self) -> dict[str, Any]:
     "datetime",
     "json",  # Represented as string in ClickHouse
     "float",
+    # Array(String) in ClickHouse; JSON text in SQLite.
+    "array_string",
+    # Map(*, String) in ClickHouse; JSON text in SQLite.
+    "map_string_string",
+    # Map(*, Float64) in ClickHouse; JSON text in SQLite.
+    "map_string_float",
```
jtschoonhoven: Really tiny nit, I'd just be explicit instead of using `*`
```suggestion
    # Map(String, String) in ClickHouse; JSON text in SQLite.
    "map_string_string",
    # Map(String, Float64) in ClickHouse; JSON text in SQLite.
    "map_string_float",
```

weave/trace_server/orm.py
```diff
@@ -183,10 +198,12 @@ def tuple_to_row(self, tup: tuple, fields: list[str]) -> Row:
...
             normalized_field = field[:-5] if field.endswith("_dump") else field
             value = tup[i]
-            if (
-                normalized_field in self.col_types
-                and self.col_types[normalized_field] == "json"
-            ):
+            col_type = self.col_types.get(normalized_field)
+            if col_type == "json":
+                d[normalized_field] = json.loads(value)
+            elif col_type in _JSON_TEXT_BACKED_COL_TYPES and isinstance(value, str):
```
jtschoonhoven: The comment explains that this check is for sqlite and I think it's better to express that explicitly. This way the logic matches to write path below.

```suggestion
            elif col_type in _JSON_TEXT_BACKED_COL_TYPES and database_type == "sqlite":
```

jtschoonhoven: If you want the extra safety you could add an additional check:
```suggestion
            elif col_type in _JSON_TEXT_BACKED_COL_TYPES and database_type == "sqlite":
                if not isinstance(value, str):
                    raise ValueError(f"Unexpected value {normalized_field}: expected str, got {type(value)}")
```

weave/trace_server/trace_server_interface.py
```diff
@@ -1185,6 +1185,28 @@ class FeedbackCreateReq(BaseModelStrict):
         examples=["018f1f2a-9c2b-7d3e-b5a1-8c9d2e4f6a7b"],
     )
 
+    # typed scorer outputs; populated by agent-monitor scorers
+    scorer_tags: list[str] | None = Field(
+        default=None,
+        description="Tags applied to the ref by a scorer",
+        examples=[["nsfw", "high-quality"]],
+    )
```
jtschoonhoven: The columns are non-nullable in the database, so there shouldn't be any need to support `None` here. Same applies to the attributes below.

```suggestion
    scorer_tags: list[str] = Field(
        default_factory=list,
        description="Tags applied to the ref by a scorer",
        examples=[["nsfw", "high-quality"]],
    )
```

jtschoonhoven (approved):
Approving to unblock, but see comments. Lmk if you want to discuss more.

weave/trace_server/orm.py
```diff
@@ -178,24 +193,41 @@ def truncate_sql(self, database_type: DatabaseType) -> str:
         if database_type == "sqlite":
             return f"DELETE FROM {self.name}"
 
-    def tuple_to_row(self, tup: tuple, fields: list[str]) -> Row:
+    def tuple_to_row(
+        self,
+        tup: tuple,
+        fields: list[str],
+        database_type: DatabaseType = "clickhouse",
```
jtschoonhoven: I don't think this should have a default. Seems dangerous.

jtschoonhoven (approved):
One comment to address, otherwise LGMT. 

Let's not merge this until after the migration has been deployed just in case.

PR #6914, opened 2026-05-20
fix(weave): handle mixed BaseModel/WeaveDict in scorer auto_summarize

### LINKED ISSUE #5483: AttributeError: 'WeaveDict' object has no attribute 'model_dump'
The following line throw AttributeError at the end of an eval run:
https://github.com/wandb/weave/blob/7f85a83ab8d714d0b15ed0427ba4d7b9f212ac9d/weave/flow/scorer.py#L185

The code assumes that if val is a BaseModel, then data will support model_dump, but it's encountering a WeaveDict object instead.

Stack trace:
```
  File ".../weave/trace/op.py", line 1232, in wrapper
    res, _ = await _call_async_func(
             ^^^^^^^^^^^^^^^^^^^^^^^
        cast(Op[P, R], wrapper), *args, __should_raise=True, **kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File ".../weave/trace/op.py", line 639, in _call_async_func
    res = await func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../weave/flow/eval.py", line 263, in evaluate
    summary = await self.summarize(eval_results)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../weave/trace/op.py", line 1232, in wrapper
    res, _ = await _call_async_func(
             ^^^^^^^^^^^^^^^^^^^^^^^
        cast(Op[P, R], wrapper), *args, __should_raise=True, **kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File ".../weave/trace/op.py", line 639, in _call_async_func
    res = await func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../weave/flow/eval.py", line 210, in summarize
    scored = summarize_fn(score_table)
  File ".../weave/flow/scorer.py", line 185, in auto_summarize
    return auto_summarize([x.model_dump() for x in data])
                           ^^^^^^^^^^^^
AttributeError: 'WeaveDict' object has no attribute 'model_dump'
```

Possible fix:
```
    elif isinstance(val, BaseModel):
        return auto_summarize([x.model_dump() if hasattr(x, 'model_dump') else dict(x) for x in data])
```

ro31337:
## Description
`auto_summarize` picks its aggregation branch from the type of the first row. When that row is a pydantic `BaseModel` it calls `.model_dump()` on every item — but scorer outputs can round-trip through the trace server and come back as `WeaveDict` (a `dict` subclass with no `model_dump`), so a list like `[BaseModel, WeaveDict, ...]` crashed with `AttributeError`.

The fix guards the conversion per element: only `BaseModel` items get `model_dump()`, everything else passes through to the recursive `dict` branch, which already handles `WeaveDict`. This mirrors that branch and is the smallest change that dispatches correctly, since the branch only inspects `data[0]`.

## Testing
Added one regression test that summarizes `[BaseModel, WeaveDict, WeaveDict]` and checks it aggregates cleanly; it fails with the original `AttributeError` when the fix is reverted. Also regenerated the one affected non-legacy serialization fixture — the captured `auto_summarize` source shifts the `Scorer.summarize` op digest and its dependents — and the serialization suite passes on Python 3.12 and 3.13.

Closes #5483.

### DIFF
--- a/tests/flow/test_scorer.py
+++ b/tests/flow/test_scorer.py
@@ -1,8 +1,9 @@
 from unittest.mock import MagicMock
 
+import pytest
 from pydantic import BaseModel
 
-from weave.flow.scorer import Scorer
+from weave.flow.scorer import Scorer, auto_summarize
 from weave.trace.object_record import ObjectRecord
 from weave.trace.vals import WeaveDict, WeaveObject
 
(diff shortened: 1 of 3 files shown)


ro31337:
The Codecov patch-coverage report above is from the first push, when the new tests were silently deselected by the `flow` shard's `-m "trace_server"` filter. The follow-up commit `cca150942b` adds `@pytest.mark.trace_server` to all five new tests so they now run on CI, and the second run is green across every shard.

Locally, `pytest tests/flow/test_scorer.py --cov=weave.flow.scorer --cov-report=xml` shows `weave/flow/scorer.py` lines 192–194 with `hits="1"`, i.e. the new BaseModel branch is exercised by `test_auto_summarize_mixed_basemodel_and_weave_dict` (and indeed by every other new test that touches the branch). The regression test was also confirmed to fail with the original `AttributeError` when the fix is reverted, so the coverage signal here looks like a stale-report artefact rather than a real gap. Happy to dig further if anyone disagrees.

gtarpenning (approved):

PR #6916, opened 2026-05-20
feat(weave): support multi-alias semantic convention keys

andrewtruong:
Add support for multi-alias semconv keys.

The main reason for this PR is to allow multiple semconv keys to map to a single concept on the server. This may happen for a number of reasons, including:
1. Promoting a key from incubating to main;
2. Two keys just happen to represent the same concept

One specific case of this is with reasoning tokens, which have two forms:
1. `gen_ai.usage.reasoning_tokens`; and
2. `gen_ai.usage.experimental.reasoning_tokens`

### DIFF
--- a/weave/trace_server/agents/semconv.py
+++ b/weave/trace_server/agents/semconv.py
@@ -28,12 +28,32 @@ AttributeType = Literal["string", "int", "float", "string[]", "json"]
 
 @dataclass(frozen=True, slots=True)
 class Attribute:
-    """A single semantic convention attribute."""
+    """A single semantic convention attribute.
+
+    ``gen_ai_alias`` is the *primary* OTel ``gen_ai.*`` wire key for this
+    attribute — usually whatever the upstream OpenTelemetry semconv has
+    standardised on. ``additional_aliases`` lets the server recognise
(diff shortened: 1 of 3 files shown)


weave/trace_server/agents/semconv.py
```diff
@@ -28,12 +28,32 @@
...
+
+    The catalog is a recognizer registry; the OTel spec is the source of
+    truth, not this file.
+    """
 
     key: str  # canonical weave.* key
     type: AttributeType
     description: str
-    gen_ai_alias: str = ""  # OTel gen_ai.* equivalent, if any
+    gen_ai_alias: str = ""  # primary OTel gen_ai.* wire key, if any
```
zbirenbaum: Could gen_ai_alias just be changed to accept a string or a tuple? Then there's only one field to deal with
Not sure I understand the comment `# primary OTel gen_ai.* wire key`. The old comment seemed a bit clearer

zbirenbaum (commented):
Nice, just one comment, I think this could be simplified a little bit

bcsherma (approved):
nice, I would recommend we add a test case to ensure that the reasoning token aliases get recognized

PR #6917, opened 2026-05-20
feat(weave): Add OpenAI Agents integration v2

andrewtruong:
Initial integration for OpenAI Agents using OTEL.

### DIFF
--- a/weave/integrations/openai_agents/openai_agents.py
+++ b/weave/integrations/openai_agents/openai_agents.py
@@ -31,13 +31,9 @@ except ImportError:  # pragma: no cover - older openai-agents SDKs lack these
     TaskSpanData = None
     TurnSpanData = None
 
-from weave.integrations.patcher import NoOpPatcher, Patcher
-from weave.trace.autopatch import IntegrationSettings
 from weave.trace.call import Call
 from weave.trace.context import call_context
 from weave.trace.context.weave_client_context import get_weave_client
-
-_openai_agents_patcher: OpenAIAgentsPatcher | None = None
 
 
 def _call_type(span: Span) -> str:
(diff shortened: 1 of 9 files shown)


weave/integrations/openai_agents/openai_agents.py
andrewtruong, line 1: This content moved to `patcher.py`

weave/integrations/openai_agents/otel_processor.py
andrewtruong, line 1: Most of the new impl lives here and it's corresponding test file

weave/integrations/openai_agents/otel_processor.py
```diff
@@ -0,0 +1,617 @@
...
+        return []
+    messages: list[Message] = []
+    for entry in raw:
+        if not isinstance(entry, dict):
+            continue
+        role = str(entry.get("role") or default_role)
+        content = entry.get("content")
+        if isinstance(content, str):
+            messages.append(Message(role=role, content=content))
+        elif content is None:
```
drtangible: Is `content` going to be `None` in the `Non-string content (block lists)...` case referenced above?

weave/integrations/openai_agents/otel_processor.py
```diff
@@ -0,0 +1,617 @@
...
+        return _transcription_attrs(span)
+    if isinstance(sd, SpeechSpanData):
+        return _speech_attrs(span)
+    if isinstance(sd, SpeechGroupSpanData):
+        return _speech_group_attrs(span)
+    if isinstance(sd, MCPListToolsSpanData):
+        return _mcp_list_tools_attrs(span)
+    if isinstance(sd, CustomSpanData):
+        return _custom_attrs(span)
+    return {}
```
drtangible, line 457: Out of curiosity -- if a new type of span was introduced by a future version of the OpenAI Agents SDK, would we detect it prior to looking at the documentation? Not sure if logging / warning on unhandled span types is worthwhile.

drtangible (approved):
👍 👍 👍

weave/integrations/openai_agents/otel_processor.py
andrewtruong: That's a good point, I'll add that

PR #6909, opened 2026-05-20
feat(weave): add wandb.agent_monitor feedback type

nikumar1206:
## Description

Adds the \`wandb.agent_monitor\` feedback_type. Scorer identity comes from \`runnable_ref\` (no \`.<name>\` suffix like \`wandb.runnable.*\`).

Validation:
- Requires \`runnable_ref\`.
- Requires \`call_ref\` when \`runnable_ref\` is an op-ref.
- Rejects \`scorer_tags\` / \`scorer_ratings\` / \`scorer_reasons\` / \`scorer_confidences\` on any other feedback_type, including explicit empty containers.

## Testing

Unit tests.

PR #6918, opened 2026-05-20
feat(weave): query filters for scorer_* Array/Map columns

nikumar1206:
## Description

Filter support for the Array/Map scorer columns.

- \`Table\` exposes \`array_string_cols\` and \`map_cols\`.
- \`_transform_external_field_to_internal_field\` resolves \`scorer_ratings._rating_\` to \`scorer_ratings['_rating_']\` on ClickHouse and \`json_extract(scorer_ratings, '\$.\"_rating_\"')\` on SQLite. Backslash-escaped dots in keys are preserved (\`scorer_reasons.tag\\.nsfw\` → key \`tag.nsfw\`).
- \`\$contains\` on a bare Array(String) column compiles to \`has(col, value)\` on ClickHouse and \`EXISTS (SELECT 1 FROM json_each(col) WHERE value = ?)\` on SQLite — was \`position()\` / \`INSTR()\` before, which errors on arrays. Other operand shapes still hit the substring path.

## Testing

Unit tests.

### DIFF
--- a/weave/trace_server/orm.py
+++ b/weave/trace_server/orm.py
@@ -47,7 +47,7 @@ class ParamBuilder:
         param_builder_count += 1
         self._params: dict[str, Any] = {}
         self._prefix = (prefix or f"pb_{param_builder_count}") + "_"
-        self._database_type = database_type
+        self.database_type = database_type
         self._param_to_name: dict[Any, str] = {}
 
     def add_param(self, param_value: Any) -> str:
@@ -76,7 +76,7 @@ class ParamBuilder:
         """
         param_name = param_name or self._prefix + str(len(self._params))
         self._params[param_name] = param_value
-        if self._database_type == "clickhouse":
+        if self.database_type == "clickhouse":
             ptype = param_type or python_value_to_ch_type(param_value)
             return f"{{{param_name}:{ptype}}}"
         return ":" + param_name
@@ -159,12 +159,22 @@ class Table:
     # Fields derived from cols
     col_types: dict[str, ColumnType]
     json_cols: list[str]
+    array_string_cols: list[str]
+    map_string_cols: list[str]
+    map_float_cols: list[str]
 
     def __init__(self, name: str, cols: Columns | None = None):
         self.name = name
         self.cols = cols or []
         self.col_types = {c.name: c.type for c in self.cols}
         self.json_cols = [c.name for c in self.cols if c.type == "json"]
+        self.array_string_cols = [c.name for c in self.cols if c.type == "array_string"]
+        self.map_string_cols = [
+            c.name for c in self.cols if c.type == "map_string_string"
+        ]
+        self.map_float_cols = [
+            c.name for c in self.cols if c.type == "map_string_float"
+        ]
 
     def create_sql(self) -> str:
         sql = f"CREATE TABLE IF NOT EXISTS {self.name} (\n"
@@ -347,7 +357,7 @@ class Select:
         param_builder: ParamBuilder | None = None,
     ) -> PreparedSelect:
         param_builder = param_builder or ParamBuilder(None, database_type)
-        assert database_type == param_builder._database_type
+        assert database_type == param_builder.database_type
 
         sql = ""
         if self.action == "SELECT":
@@ -355,8 +365,10 @@ class Select:
             internal_fields = [
                 _transform_external_field_to_internal_field(
                     f,
-                    self.all_columns,
-                    self.table.json_cols,
+                    all_columns=self.all_columns,
+                    json_columns=self.table.json_cols,
+                    map_string_columns=self.table.map_string_cols,
+                    map_float_columns=self.table.map_float_cols,
                     param_builder=param_builder,
                 )[0]
                 for f in fieldnames
@@ -376,7 +388,13 @@ class Select:
         # Returns {join type} JOIN {table name} ON {join condition}
         for j in self.joins:
             query_conds, fields_used = _process_query_to_conditions(
-                j.query, self.all_columns, self.table.json_cols, param_builder
+                j.query,
+                all_columns=self.all_columns,
+                json_columns=self.table.json_cols,
+                array_string_columns=self.table.array_string_cols,
+                map_string_columns=self.table.map_string_cols,
+                map_float_columns=self.table.map_float_cols,
+                param_builder=param_builder,
             )
             joined = combine_conditions(query_conds, "AND")
             sql += f"\n{j.join_type + ' ' if j.join_type else ''}JOIN {j.table.name} ON {joined}"
@@ -389,7 +407,13 @@ class Select:
             conditions = [f"project_id = {param_project_id}"]
         if self._query:
             query_conds, fields_used = _process_query_to_conditions(
-                self._query, self.all_columns, self.table.json_cols, param_builder
+                self._query,
+                all_columns=self.all_columns,
+                json_columns=self.table.json_cols,
+                array_string_columns=self.table.array_string_cols,
+                map_string_columns=self.table.map_string_cols,
+                map_float_columns=self.table.map_float_cols,
+                param_builder=param_builder,
             )
             conditions.extend(query_conds)
 
@@ -401,8 +425,10 @@ class Select:
             internal_fields = [
                 _transform_external_field_to_internal_field(
                     f,
-                    self.all_columns,
-                    self.table.json_cols,
+                    all_columns=self.all_columns,
+                    json_columns=self.table.json_cols,
+                    map_string_columns=self.table.map_string_cols,
+                    map_float_columns=self.table.map_float_cols,
                     param_builder=param_builder,
                 )[0]
                 for f in self._group_by
@@ -418,6 +444,7 @@ class Select:
                 # For each order by field, if it is a dynamic field, we generate
                 # 3 order by terms: one for existence, one for float casting, and one for string casting.
                 # The effect of this is that we will have stable sorting for nullable, mixed-type fields.
+                options: list[tuple[tsi_query.CastTo | None, str]]
                 if _is_dynamic_field(field, self.table.json_cols):
                     # Prioritize existence, then cast to double, then str
                     options = [
@@ -426,7 +453,9 @@ class Select:
                         ("string", direction),
                     ]
                 else:
-                    options = [(field, direction)]
+                    # Static columns don't need a cast; the function ignores
+                    # the param when no JSON/Map extraction is happening.
+                    options = [(None, direction)]
 
                 # For each option, build the order by term
                 for cast, direct in options:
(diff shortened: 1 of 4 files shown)


nikumar1206:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/wandb/weave/6918?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6986** <a href="https://app.graphite.com/github/pr/wandb/weave/6986?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6982** <a href="https://app.graphite.com/github/pr/wandb/weave/6982?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6918** <a href="https://app.graphite.com/github/pr/wandb/weave/6918?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6918?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6909** <a href="https://app.graphite.com/github/pr/wandb/weave/6909?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6908** <a href="https://app.graphite.com/github/pr/wandb/weave/6908?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6906** <a href="https://app.graphite.com/github/pr/wandb/weave/6906?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/orm.py
```diff
@@ -159,12 +159,21 @@ class Table:
...
     def __init__(self, name: str, cols: Columns | None = None):
         self.name = name
         self.cols = cols or []
         self.col_types = {c.name: c.type for c in self.cols}
         self.json_cols = [c.name for c in self.cols if c.type == "json"]
+        self.array_string_cols = [c.name for c in self.cols if c.type == "array_string"]
+        self.map_cols = {
+            c.name: c.type
+            for c in self.cols
+            if c.type in {"map_string_string", "map_string_float"}
```
jtschoonhoven: Nit: this should probably be a constant declared along with `_JSON_TEXT_BACKED_COL_TYPES`

weave/trace_server/orm.py
```diff
@@ -159,12 +159,21 @@ class Table:
     # Fields derived from cols
     col_types: dict[str, ColumnType]
     json_cols: list[str]
+    array_string_cols: list[str]
+    # Map column name → value column-type (`map_string_string` / `map_string_float`).
+    map_cols: dict[str, ColumnType]
```
jtschoonhoven: There are two different patterns being used here, I think we should pick one or the other.

Either:
- Change `array_string_cols: list[str]` to `array_cols: dict[str, ColumnType]`
- Split `map_cols` into `map_string_cols: list[str]` and `map_float_cols: list[str]`

I think the latter is more readable but either is fine.

weave/trace_server/orm.py
```diff
@@ -624,13 +649,36 @@ def _transform_external_field_to_internal_field(
     field: str,
     all_columns: Sequence[str],
     json_columns: Sequence[str],
-    cast: str | None = None,
```
jtschoonhoven, line 642: Tiny nit: I'd move `map_columns` up so it follows `json_columns` so all the `*_columns` args are grouped

weave/trace_server/orm.py
```diff
@@ -624,13 +649,36 @@ def _transform_external_field_to_internal_field(
...
     raw_fields_used = set()
+    map_columns = map_columns or {}
     json_path = None
+    map_access: tuple[list[str], ColumnType] | None = None
+    for prefix, value_type in map_columns.items():
+        if field == prefix:
+            # Bare map column — return as-is and let the caller compare it.
+            break
+        if field.startswith(prefix + "."):
+            raw_key = field[len(prefix + ".") :]
```
jtschoonhoven:
```suggestion
            raw_key = field[len(prefix) + 1 :]
```

weave/trace_server/orm.py
```diff
@@ -624,13 +649,36 @@ def _transform_external_field_to_internal_field(
...
+            # Honor backslash-escaped dots in the key so reasons/confidences
+            # can reference `tag.<name>` / `rating.<name>` keys without
+            # ambiguity (`scorer_reasons.tag\.nsfw` -> key="tag.nsfw").
+            parts = split_escaped_field_path(raw_key)
+            map_access = (parts, value_type)
+            raw_fields_used.add(field)
+            field = prefix
+            break
```
jtschoonhoven: Now that the plan is to have separate `scorer_tag_reasons ` and `scorer_rating_confidences` (etc) this will need to be updated.

jtschoonhoven (commented):
Partial review. Will look again after you've updated to handle the new `scorer_tag_reasons` columns, etc. Looks pretty good so far.

weave/trace_server/orm.py
```diff
@@ -639,18 +670,40 @@ def _transform_external_field_to_internal_field(
...
+        if field == prefix:
+            # Bare map column — return as-is and let the caller compare it.
+            break
+        if field.startswith(prefix + "."):
+            raw_key = field[len(prefix) + 1 :]
+            map_access_parts = split_escaped_field_path(raw_key)
+            map_access_is_string = prefix in map_string_columns
+            raw_fields_used.add(field)
+            field = prefix
+            break
```
jtschoonhoven, line 698: I don't think `split_escaped_field_path` is doing anything useful here.

```suggestion
        if field.startswith(prefix + "."):
            map_key = field[len(prefix) + 1 :]
            raw_fields_used.add(field)
            field = prefix
            break
```

weave/trace_server/orm.py
```diff
@@ -677,7 +730,28 @@ def _transform_external_field_to_internal_field(
...
+        else:
+            key_param = param_builder.add(key_literal, None, "String")
+            # Missing CH map keys default to 0 / "" — guard with mapContains
+            # so absent keys read as NULL (matches SQLite's json_extract).
+            field = f"if(mapContains({field}, {key_param}), {field}[{key_param}], NULL)"
+            # Map(String, Float64) values are already typed; only Map(String,
+            # String) values need explicit conversion when a non-string cast
+            # is requested.
+            if map_access_is_string and cast not in {None, "string"}:
+                field = clickhouse_cast(field, cast)
```
jtschoonhoven: This looks like a bug, or at least an unnecessary constraint. If the client wants to cast a Float64 value then I don't see any reason to stop them. E.g. the client might want to check `cast="exists"` on a Base64 map.

```suggestion
            field = clickhouse_cast(field, cast)
```

weave/trace_server/orm.py
```diff
@@ -677,7 +730,28 @@ def _transform_external_field_to_internal_field(
         raise ValueError(f"Unknown field: {field}")
 
     raw_fields_used.add(unprefixed_field)
-    if json_path is not None:
+    if map_access_parts is not None:
+        # The key is whatever sits after the map-column prefix; for CH the
+        # `col[key]` literal expects the full (possibly dotted) string.
+        key_literal = ".".join(map_access_parts)
+        is_sqlite = param_builder._database_type == "sqlite"
```
jtschoonhoven: `_database_type` is private, so you shouldn't be accessing it here. If you don't think `_database_type` should be private then you can remove the `_` prefix.

weave/trace_server/orm.py
```diff
@@ -755,23 +848,41 @@ def process_operation(operation: tsi_query.Operation) -> str:
...
+                rhs_part = process_operand(operation.contains_.substr)
+                raw_fields_used.add(array_col)
+                if pb._database_type == "sqlite":
+                    cond = (
+                        f"EXISTS (SELECT 1 FROM json_each({array_col}) "
+                        f"WHERE value = {rhs_part})"
+                    )
                 else:
-                    cond = f"INSTR({lhs_part}, {rhs_part}) > 0"
+                    cond = f"has({array_col}, {rhs_part})"
```
jtschoonhoven, line 873: Here you've removed support for case-insensitive matching. Worth adding back I think. Claude suggests `arrayExists(x -> lower(x) = lower(?), col)`

weave/trace_server/orm.py
```diff
@@ -694,10 +768,29 @@ def _transform_external_field_to_internal_field(
...
+    return None
+
+
 def _process_query_to_conditions(
     query: tsi.Query,
     all_columns: Sequence[str],
     json_columns: Sequence[str],
+    array_string_columns: Sequence[str] = (),
+    map_string_columns: Sequence[str] = (),
+    map_float_columns: Sequence[str] = (),
```
jtschoonhoven: It will make the diff slightly bigger, but I think it's worth forcing the calling function to use named arguments. The types are all the same, so passing in arguments positionally is brittle. And while you're here, might as well set a default value for `all_columns` and `json_columns` as well
```suggestion
    *
    all_columns: Sequence[str] = (),
    json_columns: Sequence[str] = (),
    array_string_columns: Sequence[str] = (),
    map_string_columns: Sequence[str] = (),
    map_float_columns: Sequence[str] = (),
```

weave/trace_server/orm.py
```diff
@@ -639,18 +670,40 @@ def _transform_external_field_to_internal_field(
     field: str,
     all_columns: Sequence[str],
     json_columns: Sequence[str],
```
jtschoonhoven: Same comment here as for `_process_query_to_conditions`. These types are all the same so it would be safer to force callers to use named arguments. Also, if `map_string_columns` has a default then I think so should `json_columns` (or neither should).
```suggestion
    *,
    all_columns: Sequence[str] = (),
    json_columns: Sequence[str] = (),
```

jtschoonhoven (commented):
A few more notes before approving

weave/trace_server/orm.py
```diff
@@ -159,12 +159,22 @@ class Table:
...
         self.name = name
         self.cols = cols or []
         self.col_types = {c.name: c.type for c in self.cols}
         self.json_cols = [c.name for c in self.cols if c.type == "json"]
+        self.array_string_cols = [c.name for c in self.cols if c.type == "array_string"]
+        self.map_string_cols = [
+            c.name for c in self.cols if c.type == "map_string_string"
+        ]
+        self.map_float_cols = [
+            c.name for c in self.cols if c.type == "map_string_float"
```
jtschoonhoven, line 176: Nit: `"array_string"`, `"map_string_string"`, and `"map_string_float"` should all be constants

weave/trace_server/orm.py
```diff
@@ -637,20 +668,39 @@ def quote_json_path_parts(parts: list[str]) -> str:
...
+    cast: tsi_query.CastTo | None = None,
     param_builder: ParamBuilder | None = None,
 ) -> tuple[str, ParamBuilder, set[str]]:
-    """Transforms a request for a dot-notation field to a clickhouse field."""
+    """Transforms a request for a dot-notation field to a clickhouse field.
+
+    For Map(String, *) columns, a dotted path like `scorer_ratings._rating_`
+    resolves to a typed map access (`col['key']` in ClickHouse,
+    `json_extract(col, '$."key"')` in SQLite). Bare references to a map
+    column pass through to the column itself.
```
jtschoonhoven, line 684: NIt: the wording of this example could be more clear
```suggestion
    For Map columns, a dotted path like `scorer_ratings._rating_` resolves
    to a typed map access (`scorer_ratings['_rating_']` in ClickHouse,
    `json_extract(scorer_ratings, '$."_rating_"')` in SQLite). Bare references
    to a map column pass through to the column itself.
```

jtschoonhoven (approved):

PR #6920, opened 2026-05-20
perf(weave): force vertical merge on wide calls tables

gtarpenning:
## Summary

Adds migration 031 that sets the following on `call_parts`, `calls_merged`, and `calls_complete`:

- `vertical_merge_algorithm_min_columns_to_activate = 1` (default 11)
- `vertical_merge_algorithm_min_rows_to_activate = 1` (default 131072)
- `vertical_merge_algorithm_min_bytes_to_activate = 0`
- `merge_max_block_size = 65536` (default 8192)

Forces ClickHouse to pick the Vertical merge algorithm for these wide-row tables. The default column threshold of 11 was causing the engine to fall back to Horizontal even on schemas where each row is ~17 KB and column-by-column merging is much more efficient.

Context: a production cluster was observed running a single Horizontal merge on `calls_merged` for ~32 min at 28% progress (1.57M rows, 26.79 GB uncompressed, ~17 KB/row). A neighboring narrower table (`call_parts` at the time of investigation, post-mutation) was already merging Vertical and finishing in seconds. Lowering the activation thresholds is the safest single knob to flip.

Only applies to future merges, no data rewrite. Reversible via the `.down.sql` (`RESET SETTING ...`).

## Testing
todo!

### DIFF
--- a/weave/trace_server/migrations/031_vertical_merge_settings.down.sql
+++ b/weave/trace_server/migrations/031_vertical_merge_settings.down.sql
@@ -1,0 +1,22 @@
+-- Rollback Migration 031: reset vertical-merge tuning to cluster defaults.
+
+ALTER TABLE call_parts
+    RESET SETTING
+        vertical_merge_algorithm_min_columns_to_activate,
+        vertical_merge_algorithm_min_rows_to_activate,
+        vertical_merge_algorithm_min_bytes_to_activate,
+        merge_max_block_size;
+
+ALTER TABLE calls_merged
+    RESET SETTING
+        vertical_merge_algorithm_min_columns_to_activate,
+        vertical_merge_algorithm_min_rows_to_activate,
+        vertical_merge_algorithm_min_bytes_to_activate,
+        merge_max_block_size;
+
+ALTER TABLE calls_complete
+    RESET SETTING
+        vertical_merge_algorithm_min_columns_to_activate,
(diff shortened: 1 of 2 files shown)


gtarpenning:
Closing — local load test ([bench harness](https://github.com/wandb/weave/issues)) showed the vertical-merge settings are a no-op on `calls_merged` in CH 26.3.

`AggregatingMergeTree` (and `SummingMergeTree`) are excluded from the vertical-merge algorithm regardless of threshold settings. Verified empirically: setting `min_columns=1, min_rows=1, min_bytes=0` on a 240k-row / 7.5GB AMT merge still picks Horizontal. `MergeTree` / `ReplacingMergeTree` / `CollapsingMergeTree` pick Vertical with the same settings.

So the prod 32-min Horizontal merge on `calls_merged` is just how AMT works on wide rows — this migration can't flip it.

Per-table impact of this PR:
- `call_parts` (MergeTree): would force V on small (<131k row) merges; defaults already pick V at larger scales. Peak mem drops ~4.6× at small scale.
- `calls_merged` (AggregatingMergeTree): no algorithm change. Only `merge_max_block_size=65536` had any effect (~20-28% faster Horizontal at 300k-500k rows in local bench).
- `calls_complete` (ReplacingMergeTree): same story as `call_parts`.

May reopen as a `merge_max_block_size`-only change for `calls_merged` if the local ~20% holds up in prod.

PR #6922, opened 2026-05-21
fix(weave): distinguish 429 rate limit types using Retry-After header in Node SDK

rajkumar-prog:
## Problem

Fixes #6543

The Node SDK retry logic in \`sdks/node/src/utils/retry.ts\` treats all 429 responses identically — it always retries with exponential backoff regardless of the \`Retry-After\` header. This means:

- **Quota/billing exhaustion** (long \`Retry-After\`, e.g. 600s) → retried anyway, amplifying pressure
- **Transient pressure** (short \`Retry-After\`, e.g. 30s) → retried with wrong delay (exponential backoff instead of the server-specified wait)

When Weave traces LLM calls at volume and hits a quota exhaustion 429, the SDK retries instead of surfacing the failure immediately — making things worse.

## Fix

Added \`Retry-After\` header inspection to \`retry.ts\`:

1. **Long \`Retry-After\` (> 60s)** → quota/billing exhaustion, return immediately without retrying
2. **Short \`Retry-After\` (≤ 60s)** → use the server-specified delay instead of exponential backoff
3. **No \`Retry-After\`** → existing exponential backoff behaviour (unchanged)

Also added a \`parseRetryAfterMs()\` helper that handles both numeric seconds and HTTP-date formats per the spec.

## Bug fix included

During testing, a negative-value edge case was found: a \`Retry-After: -1\` value passed the numeric \`seconds >= 0\` guard (false), fell through to the HTTP-date parser, was interpreted as year 1 BCE by \`new Date('-1')\`, and returned \`0\` instead of \`null\`. Fixed by returning \`null\` explicitly for any numeric value below zero before reaching the date parser.

## Changes

- \`sdks/node/src/utils/retry.ts\` — \`Retry-After\` classification logic + \`parseRetryAfterMs()\` helper (exported); negative-value edge case fixed
- \`sdks/node/src/utils/__tests__/retry.test.ts\` — 11 tests covering: numeric seconds, HTTP date format, negative value, no header, over-threshold (>60s) stops retries, short value uses server delay (not backoff), and \`retryOnStatus\` disabling 429 retries

### DIFF
--- a/sdks/node/src/utils/__tests__/retry.test.ts
+++ b/sdks/node/src/utils/__tests__/retry.test.ts
@@ -1,0 +1,162 @@
+import {createFetchWithRetry, parseRetryAfterMs} from '../retry';
+
+// ---------------------------------------------------------------------------
+// parseRetryAfterMs
+// ---------------------------------------------------------------------------
+
+function headersFrom(record: Record<string, string>): Headers {
+  return new Headers(record);
+}
+
+describe('parseRetryAfterMs', () => {
+  test('returns null when Retry-After header is absent', () => {
+    expect(parseRetryAfterMs(headersFrom({}))).toBeNull();
+  });
+
+  test('returns milliseconds for a positive numeric value', () => {
+    expect(parseRetryAfterMs(headersFrom({'retry-after': '30'}))).toBe(30_000);
+    expect(parseRetryAfterMs(headersFrom({'retry-after': '0'}))).toBe(0);
+  });
+
+  test('returns null for a negative numeric value', () => {
+    // Negative values are invalid per RFC 7231 and must not be used as a delay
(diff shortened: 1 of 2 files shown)


rajkumar-prog:

rajkumar-prog:
Addressed both Codex review comments:

1. **Respect `retryOnStatus` for 429**: Added `&& retryOnStatus(response.status)` check before entering the 429-specific branch — callers who disable retries for 429 are now respected.

2. **Reject negative Retry-After values**: Added `&& seconds >= 0` guard in `parseRetryAfterMs` — negative values like `-1` now fall through to the HTTP-date parser and return `null` if unparseable, avoiding tight retry loops.

rajkumar-prog:
Bumping to prevent auto-close — this PR is still relevant and ready for review. Happy to address any feedback.

PR #6923, opened 2026-05-21
chore(weave): Add RLock for patching to prevent deadlock

andrewtruong:
Prevents a deadlocking case that can happen when using litellm


drtangible (approved):
👍 👍 👍

PR #6935, opened 2026-05-21
fix(weave): handle coexisting JSON string and dotted OTel attribute keys

zbirenbaum:
- When an OTel exporter sends both `gen_ai.completion` as a JSON string and `gen_ai.completion.0.role` as a dotted subkey, the span was silently rejected due to `AttributePathConflictError`
- Adds `_list_to_numeric_dict` to convert lists to numeric-keyed dicts so `_validate_structure` can descend into them
- Adds `_deep_merge` and `_merge_structure` so structured values (JSON string → parsed list, dotted subkeys → nested dict) merge instead of conflicting
- Unit tests for all merge orderings + integration test confirming the span is no longer dropped

- Fixes WB-34637

### DIFF
--- a/tests/trace_server/test_opentelemetry.py
+++ b/tests/trace_server/test_opentelemetry.py
@@ -925,6 +925,28 @@ class TestAttributes:
             {"role": "assistant", "content": "hello", "extra": True}
         ]
 
+    def test_expand_attributes_nested_list_merge_preserves_dotted_only_elements(self):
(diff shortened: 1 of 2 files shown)


gtarpenning (approved):
pretty hairy problem, can you just confirm that this can't be broken with some clever input params? maybe you can use my red-team-pr tool, [here](https://gist.github.com/gtarpenning/0b08fccf0245fbe90654cfb1a6168aab)

PR #6940, opened 2026-05-21
fix(weave): agent spans query include_details + ext-to-int ref conversion

bcsherma:
## Summary

Two paired changes to the agents pipeline. Shipping together because (2) is a prerequisite for (1) to not 500.

1. **`include_details` flag on `AgentSpansQueryReq`.** Lets the span detail panel pull heavy per-row fields (messages, tool args/result, `system_instructions`, `compaction_summary`, the typed `*_refs` columns, `raw_span_dump`) without bloating the spans-table list query. Also expands the always-on scalar projection to include cheap fields the UI table already reads but the API was dropping on the floor (cache tokens, every `request_*` parameter, `output_type`, server addr/port, `wb_run_step*`, compaction counters).

2. **OTel ingest now converts the typed ref attributes ext→int.** In `external_to_internal_trace_server_adapter.genai_otel_export`, walk each span's protobuf `attributes` and rewrite the values of `weave.content_refs` / `weave.artifact_refs` / `weave.object_refs` from `weave:///entity/project/...` to `weave-trace-internal:///<project_id>/...` before forwarding the request to the inner trace server. Handles both flat-dotted (`weave.object_refs`) and nested-kvlist (`weave` → `object_refs`) OTel attribute encodings.

## Why the protobuf walk lives in the adapter

Every other resolver gets ext→int conversion for free by wrapping `_ref_apply`, whose `universal_ext_to_int_ref_converter` walks the request and rewrites any `weave:///` strings it finds. The walker handles `BaseModel` / `dict` / `list` / `tuple` / `set` and treats everything else as a scalar leaf.

`GenAIOTelExportReq.processed_spans[*].resource_spans` is a raw protobuf `ResourceSpans` (`arbitrary_types_allowed=True`), so the walker bounces off it — the refs inside the OTel attribute values escape conversion entirely. That's fine for fields the inner server stores as JSON blobs (the strict reader sees one string starting with `{`, no `weave://` prefix match), but it 500s every detail-mode query the moment the typed `Array(String)` ref columns get selected, because the read-path `int→ext` converter walks each list element as its own bare string and explicitly raises on `weave://`-prefixed values.

Pre-walking the protobuf in the adapter restores the invariant every other resolver relies on: the inner trace server receives a request that's already in internal-ref form. The inner trace server stays oblivious to ext/int — no signature changes anywhere downstream of the adapter.

## What's intentionally _not_ converted

- `raw_span_dump`, `attributes_dump`, `events_dump`, `resource_dump` — preserved byte-for-byte from the OTel sender. They're JSON strings, the strict reader doesn't trip on them, and they're meant to be a verbatim record of what the client emitted.
- Refs embedded inside message content / tool arguments / event payloads — same reasoning. The walker only touches values of the three known ref-typed attribute keys.
- Other schemes (e.g. `wandb-artifact://`) are left alone — `_convert_external_weave_ref` only rewrites strings that match `weave:///`.

## Why not just relax the read-path strict check?

That check is what catches genuine data corruption (refs stored in the wrong scheme on disk). The right fix is to make the on-disk invariant hold, not to weaken the invariant check.

### DIFF
--- a/weave/trace_server/external_to_internal_trace_server_adapter.py
+++ b/weave/trace_server/external_to_internal_trace_server_adapter.py
@@ -1,12 +1,78 @@
 import abc
-from collections.abc import Callable, Iterator
+from collections.abc import Callable, Iterable, Iterator
 from typing import Any, TypeVar
 
+from opentelemetry.proto.common.v1.common_pb2 import KeyValue
+
+from weave.shared import refs_internal as ri
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.trace_server_converter import (
     universal_ext_to_int_ref_converter,
     universal_int_to_ext_ref_converter,
 )
+
+# OTel attribute keys whose values are typed `Array(String)` ref columns
+# in the agents `spans` table. Refs in these columns are surfaced as bare
+# strings by the read-path response, so they must be in internal form on
+# disk for the int→ext converter to round-trip them.
+_OTEL_REF_ATTR_KEYS = frozenset(
+    {
+        "weave.content_refs",
+        "weave.artifact_refs",
+        "weave.object_refs",
+    }
+)
+_WEAVE_EXT_PREFIX = ri.WEAVE_SCHEME + ":///"
+
+
+def _convert_external_weave_ref(
+    ref: str, ext_to_int_project_id: Callable[[str], str]
+) -> str:
+    """Rewrite a single `weave:///entity/project/tail` ref to internal form.
+
+    Anything else (other schemes, malformed refs, already-internal refs)
+    is returned unchanged.
+    """
+    if not ref.startswith(_WEAVE_EXT_PREFIX):
+        return ref
+    rest = ref[len(_WEAVE_EXT_PREFIX) :]
+    parts = rest.split("/", 2)
+    if len(parts) != 3:
+        return ref
+    entity, project, tail = parts
+    internal_project_id = ext_to_int_project_id(f"{entity}/{project}")
+    return f"{ri.WEAVE_INTERNAL_SCHEME}:///{internal_project_id}/{tail}"
+
+
+def _rewrite_otel_ref_attrs_inplace(
+    attrs: Iterable[KeyValue],
+    ext_to_int_project_id: Callable[[str], str],
+    prefix: str = "",
+) -> None:
+    """Recursively rewrite typed ref attributes from external to internal form.
+
+    OTel encodes attributes either flat (`weave.object_refs`) or nested as
+    a `kvlist_value` (top-level key `weave` with a child `object_refs`),
+    so we descend through `kvlist_value`s and accumulate dotted prefixes.
+    Only `Array(String)` values under the known ref keys are touched;
+    everything else (including refs embedded in event payloads, message
+    content, `raw_span_dump`, etc.) is left exactly as the client sent it.
+    """
+    for kv in attrs:
+        full_key = f"{prefix}{kv.key}" if prefix else kv.key
+        value = kv.value
+        if full_key in _OTEL_REF_ATTR_KEYS and value.HasField("array_value"):
+            for item in value.array_value.values:
+                if item.HasField("string_value"):
+                    item.string_value = _convert_external_weave_ref(
+                        item.string_value, ext_to_int_project_id
+                    )
+        elif value.HasField("kvlist_value"):
+            _rewrite_otel_ref_attrs_inplace(
+                value.kvlist_value.values,
+                ext_to_int_project_id,
+                prefix=f"{full_key}.",
+            )
 
 
 class IdConverter:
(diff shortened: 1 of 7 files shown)


weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -1164,6 +1230,18 @@ def genai_otel_export(
...
         if req.wb_user_id is not None:
             req.wb_user_id = self._idc.ext_to_int_user_id(req.wb_user_id)
+        # `_ref_apply`'s universal walker can't descend into the raw
+        # protobuf `ResourceSpans` payload, so the typed-ref OTel
+        # attribute values (`weave.content_refs`, `weave.artifact_refs`,
+        # `weave.object_refs`) escape the standard ext→int conversion.
+        # Rewrite them in-place here so the inner trace server sees a
+        # request that's already in internal-ref form, matching the
+        # invariant every other resolver relies on.
+        ext_to_int = self._idc.ext_to_int_project_id
```
neutralino1, line 1229: Do we cache ext-to-int mappings?

gtarpenning (commented):
no tests?

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -1,13 +1,79 @@
...
+    {
+        "weave.content_refs",
+        "weave.artifact_refs",
+        "weave.object_refs",
+    }
+)
+_WEAVE_EXT_PREFIX = ri.WEAVE_SCHEME + ":///"
+
+
+def _convert_external_weave_ref(
```
gtarpenning: why do we need to use this instead of `universal_ext_to_int_ref_converter::replace_ref`?

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -923,8 +966,9 @@ def make_spans_list_query(pb: ParamBuilder, req: AgentSpansQueryReq) -> str:
         custom_attr_projection = _custom_attr_map_projection(
             pb, req.custom_attr_columns
         )
+        details_projection = f", {SPANS_DETAILS_COLS}" if req.include_details else ""
```
gtarpenning, line 969: what about the `group_by` or `validator` paths?

gtarpenning (commented):
super nitty but this is def a `fix` not a `feat` also whats the perf impact?

weave/trace_server/external_to_internal_trace_server_adapter.py
bcsherma: Not before this PR — the inner `_convert_external_weave_ref` was calling `ext_to_int_project_id` per ref. Pushed [2d27ef7](https://github.com/wandb/weave/commit/2d27ef76ee) to thread a single per-request `dict[str, str]` cache through the OTel walker so the converter runs at most once per distinct `entity/project` pair across every span in the batch (verified by `test_genai_otel_export_caches_project_id_lookup_across_batch`). Symmetric with the existing closure-local cache inside `universal_ext_to_int_ref_converter`.

weave/trace_server/external_to_internal_trace_server_adapter.py
bcsherma: Good catch — `replace_ref` was a closure inside `universal_ext_to_int_ref_converter`, so it wasn't reusable. Refactored in [2d27ef7](https://github.com/wandb/weave/commit/2d27ef76ee): extracted it as a module-level `replace_external_weave_ref(ref_str, convert_ext_to_int_project_id, cache=None)` in `trace_server_converter`. Both `universal_ext_to_int_ref_converter` (closure-local cache) and the adapter's OTel walker (per-request cache) now share the same parse/replace/cache logic. Deleted the duplicated `_convert_external_weave_ref` from the adapter.

weave/trace_server/query_builder/agent_query_builder.py
bcsherma: The grouped branch returns aggregates, not per-span rows, so projecting `SPANS_DETAILS_COLS` there is meaningless — silently dropping it is the foot-gun. Added a validator in [2d27ef7](https://github.com/wandb/weave/commit/2d27ef76ee) that rejects `include_details=True` combined with `group_by`, symmetric with the existing `custom_attr_columns` + `group_by` rule. Covered by `test_include_details_rejected_with_group_by`.

bcsherma:
Thanks for the review — addressed all three thread comments plus the top-level ones in [2d27ef7](https://github.com/wandb/weave/commit/2d27ef76ee):

**`feat` → `fix`**: PR title updated.

**Tests**: added in this push (1,024 trace_server tests pass locally):
- `tests/trace_server/test_trace_server_converter.py` — `replace_external_weave_ref` caching + error semantics.
- `tests/trace_server/test_trace_server_interface_util.py` — `genai_otel_export` ref rewriting end-to-end: flat-dotted attrs (`weave.object_refs`), the nested-kvlist form, non-ref attrs (`weave.raw_span_dump`) left byte-identical, mixed arrays with non-`weave://` elements passed through, and the per-request `ext_to_int_project_id` cache amortizing lookups across spans.
- `tests/trace_server/query_builder/test_agent_query_builder.py` — `include_details=True` projects `SPANS_DETAILS_COLS`; rejected when combined with `group_by`.

**Perf impact** (genai_otel_export only — everything else is unchanged):
- **Without refs in the batch**: ~one `KeyValue.HasField` call per attribute per span. Comparable to the existing attribute iteration the inner server already does.
- **With refs in the batch**: one `replace_external_weave_ref` per matching string (constant-time parse + `dict` get/set on a per-request cache). The cache amortizes the `ext_to_int_project_id` call to **once per distinct entity/project pair**, regardless of ref count (verified by `test_genai_otel_export_caches_project_id_lookup_across_batch`).
- No new allocations on the no-ref-attrs path beyond the empty cache dict; refs that don't start with `weave:///` short-circuit before the parse.

`include_details` is a pure additive projection — no perf change for the default `False` case (it's how the spans-list table already reads).

gtarpenning (approved):

PR #6942, opened 2026-05-21
fix(weave): keep calls_complete update payload out of the URL

gtarpenning:
https://coreweave.atlassian.net/browse/WB-34650

## Summary
- `build_calls_complete_update_end_query` used CH's `{name:Type}` syntax. clickhouse_connect serializes those params into the URL, so a 2 MB `output_dump` becomes a 2 MB URL -> LB drops the connection -> broken-pipe 502. Switch to `%(name)s` so values inline into the SQL body via `escape_str`.
- Bump `max_query_size` to 32 MiB in `CLICKHOUSE_LIGHTWEIGHT_UPDATE_SETTINGS`.
- Map CH `TOO_LARGE_QUERY` -> 413 instead of leaking 502.
- DD: https://us5.datadoghq.com/apm/trace/6a0eee88000000005dd68ce876650999

## Testing
- unit tests for placeholders + injection safety + 413 mapping.
- verified locally against CH 26.3: 2 MB `output_dump` ships in a 332-byte URL, row updates, `EXPLAIN indexes=1` confirms full PK use.

### DIFF
--- a/tests/trace_server/query_builder/test_calls_query_builder.py
+++ b/tests/trace_server/query_builder/test_calls_query_builder.py
@@ -2132,16 +2132,16 @@ def test_build_calls_complete_update_end_query() -> None:
     expected = """
         UPDATE calls_complete
         SET
-            ended_at = fromUnixTimestamp64Micro({ended_at:Int64}, 'UTC'),
-            exception = {exception:String},
-            output_dump = {output_dump:String},
-            summary_dump = {summary_dump:String},
-            output_refs = {output_refs:Array(String)},
-            wb_run_step_end = {wb_run_step_end:UInt64},
+            ended_at = fromUnixTimestamp64Micro(%(ended_at)s, 'UTC'),
+            exception = %(exception)s,
+            output_dump = %(output_dump)s,
+            summary_dump = %(summary_dump)s,
+            output_refs = %(output_refs)s,
+            wb_run_step_end = %(wb_run_step_end)s,
             updated_at = now64(3)
-        WHERE project_id = {project_id:String}
-            AND started_at = fromUnixTimestamp64Micro({started_at:Int64}, 'UTC')
-            AND id = {id:String}
+        WHERE project_id = %(project_id)s
+            AND started_at = fromUnixTimestamp64Micro(%(started_at)s, 'UTC')
+            AND id = %(id)s
     """
 
     exp_formatted = sqlparse.format(expected, reindent=True)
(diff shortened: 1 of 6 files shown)


tests/trace_server/query_builder/test_calls_query_builder.py
```diff
@@ -2152,6 +2152,57 @@ def test_build_calls_complete_update_end_query() -> None:
     )
 
 
+def test_build_calls_complete_update_end_query_client_side_binding() -> None:
+    """The update-end query MUST use Python `%(name)s` (client-side) placeholders.
+
+    `clickhouse_connect`'s `{name:Type}` server-side substitution puts params in
+    the URL, which makes multi-MB outputs blow past LB URI limits and surface as
```
mscavezze-cw: Is there a perf difference? Do we want to make the client/server resolution conditional on the size of the things being substituted?

tests/trace_server/query_builder/test_calls_query_builder.py
```diff
@@ -2152,6 +2152,57 @@ def test_build_calls_complete_update_end_query() -> None:
     )
 
 
+def test_build_calls_complete_update_end_query_client_side_binding() -> None:
+    """The update-end query MUST use Python `%(name)s` (client-side) placeholders.
+
+    `clickhouse_connect`'s `{name:Type}` server-side substitution puts params in
```
mscavezze-cw: This is a useful comment. Would it make sense to put it in the implementation instead of the test?

tests/trace_server/query_builder/test_calls_query_builder.py
```diff
@@ -2152,6 +2152,57 @@ def test_build_calls_complete_update_end_query() -> None:
...
+    the URL, which makes multi-MB outputs blow past LB URI limits and surface as
+    `BrokenPipeError`. We deliberately use client-side substitution so that the
+    `output_dump`/`summary_dump` values land in the SQL body instead.
+
+    Also verifies that running the query through `clickhouse_connect.driver.binding.bind_query`
+    with a malicious string literal does not let the value escape its quotes.
+    """
+    from clickhouse_connect.driver.binding import bind_query
```
gtarpenning: move...

weave/trace_server/errors.py
```diff
@@ -429,6 +429,12 @@ def handle_clickhouse_query_error(e: Exception) -> None:
             "Lightweight update queries are not supported by this ClickHouse version or configuration. "
             "This is a ClickHouse version issue that needs to be resolved."
         ) from e
+    if "Max query size exceeded" in error_str:
```
gtarpenning, line 432: better way here?

tests/trace_server/query_builder/test_calls_query_builder.py
```diff
@@ -2152,6 +2152,57 @@ def test_build_calls_complete_update_end_query() -> None:
...
 
+def test_build_calls_complete_update_end_query_client_side_binding() -> None:
+    """The update-end query MUST use Python `%(name)s` (client-side) placeholders.
+
+    `clickhouse_connect`'s `{name:Type}` server-side substitution puts params in
+    the URL, which makes multi-MB outputs blow past LB URI limits and surface as
+    `BrokenPipeError`. We deliberately use client-side substitution so that the
+    `output_dump`/`summary_dump` values land in the SQL body instead.
+
+    Also verifies that running the query through `clickhouse_connect.driver.binding.bind_query`
```
mscavezze-cw: Consider moving the injection test into a separate test for clarity of code and test-failure reporting

tests/trace_server/query_builder/test_calls_query_builder.py
mscavezze-cw: It's been a while since I noticed an import in a function. Are there some anti-patterns nearby that are being copied? Maybe a quick refactor of nearby code would be useful.

weave/trace_server/errors.py
mscavezze-cw: I'm not aware of one, but I don't have a ton of context on clickhouse.

If you produce the error in the test ([described here](https://coreweave.slack.com/archives/C03BSTEBD7F/p1779402602042099?thread_ts=1779401717.392039&cid=C03BSTEBD7F)), at least you'll know if an upgrade breaks the code. Assuming that an upgrade would go through code review on `core` and not be done externally.

Update, I see this is in the `weave` repo. I'm not sure where clickhouse versions are set anyway.

mscavezze-cw (approved):
The injection test has been removed ([link](https://github.com/wandb/weave/pull/6942/changes/b5c2c5c832ea58c8e696879d9331a096c8013171..b85d823ae1cb740ddbdf3d159f5c70d68e8e8a55#diff-700698ef9ae2130e76b9de908a1f2dd7269af005fa96a03568728a87a924436cL2183)). Was that intentional, because the test was weak for some reason, or accidental?

PR #6943, opened 2026-05-21
feat(weave): Session SDK respects global settings

rgao-coreweave:
Replaces #6846 — fresh single-commit version off master, with the unresolved review feedback from #6846 pre-applied.

Wires four Weave global settings into the Session SDK so it matches `@op`, and fixes a reasoning leak.

## Changes

- **`WEAVE_DISABLED`** — silences Session SDK spans (streaming + batch).
- **`WEAVE_REDACT_PII`** — redacts Tool args/result, LLM input/output messages, system instructions, LLM reasoning, and Turn messages. Routes through the existing `weave.utils.pii_redaction` primitives. Skips Presidio entirely when `include_content=False`.
- **`WEAVE_CAPTURE_CLIENT_INFO` / `WEAVE_CAPTURE_SYSTEM_INFO`** — stamps `weave.*` metadata (version, source, OS info) on every Session span.
- **Bug fix** — `LLM.reasoning` was leaking into `gen_ai.output.messages` even when `include_content=False`. Now gated correctly.

## Shape

Each span class has a single `_build_attrs()` shared between the streaming (`.end()`) and batch (`log_turn`) paths, so the two can't drift.

`weave/utils/capture_info.py` is new — shared by Session SDK + `weave_client.py` for the `weave.*` attr gating.

## Refactor commit applies #6846 review feedback up front

To avoid another round of review on the same points, the second commit pre-applies the unresolved comments from #6846:

- `capture_info`: return `dict[str, str]` instead of `list[tuple[str, str]]` — keys are specific, dict gives accurate typing without TypedDict boilerplate.
- `pii_redaction`: drop abbreviations — `msgs`→`messages`, `insts`→`instructions`, `m`→`message`, `s`→`instruction`.
- Tests: patch `_get_engines` (the single Presidio dependency) instead of the high-level `redact_pii` / `redact_pii_string`. Real redaction code paths now run end-to-end against deterministic stub engines via a new `fake_presidio` fixture. `include_content=False` tests now assert `_get_engines` isn't called — matches the actual cost (no NLP engine loading on already-dropped content).
- Test helpers: `_redact_streaming_*` → `_streaming_*_with_pii`; helpers set up SDK usage, the SUT does redaction. `_ic_off_*` → `_content_off_*` for clarity. `exercise` → `user_func`.

## Tests

22 settings tests in `tests/session/test_session_settings.py`, all passing. Full session suite (225 tests) passes. `tests/trace/test_trace_settings.py` (40 tests) also passes — validates `@op`'s use of `capture_info` still works after the rename.

### DIFF
--- a/weave/utils/capture_info.py
+++ b/weave/utils/capture_info.py
@@ -18,27 +18,20 @@ from weave.trace.settings import (
 from weave.version import VERSION
 
 
-def get_capture_info_items() -> list[tuple[str, str]]:
-    """Return weave.* (subkey, value) pairs gated by capture-info settings.
+def get_capture_info() -> dict[str, str]:
+    """Return weave.* metadata keys (unprefixed), gated by capture-info settings.
 
-    Subkeys are unprefixed (``client_version``, not ``weave.client_version``)
+    Keys are unprefixed (``client_version``, not ``weave.client_version``)
     so call sites can prefix or nest as their attribute shape requires.
+    Both settings can independently include or exclude their key group.
     """
-    items: list[tuple[str, str]] = []
+    info: dict[str, str] = {}
     if should_capture_client_info():
-        items.extend(
-            [
-                ("client_version", VERSION),
-                ("source", "python-sdk"),
-                ("sys_version", sys.version),
-            ]
-        )
+        info["client_version"] = VERSION
+        info["source"] = "python-sdk"
+        info["sys_version"] = sys.version
     if should_capture_system_info():
-        items.extend(
-            [
-                ("os_name", platform.system()),
-                ("os_version", platform.version()),
-                ("os_release", platform.release()),
-            ]
-        )
-    return items
+        info["os_name"] = platform.system()
+        info["os_version"] = platform.version()
+        info["os_release"] = platform.release()
+    return info
--- a/weave/utils/capture_info.py
+++ b/weave/utils/capture_info.py
@@ -1,0 +1,44 @@
+"""Shared client/system info capture — used by both @op tracer and Session SDK.
+
+Returns ``weave.*`` metadata key/value pairs gated by
+``should_capture_client_info()`` / ``should_capture_system_info()``. Call
+sites apply their own key prefix (``"weave."`` dotted vs nested
+``_WeaveKeyDict``), but the gating + value sources stay in one place.
+"""
+
+from __future__ import annotations
+
+import platform
+import sys
+
+from weave.trace.settings import (
+    should_capture_client_info,
+    should_capture_system_info,
+)
+from weave.version import VERSION
+
+
+def get_capture_info_items() -> list[tuple[str, str]]:
+    """Return weave.* (subkey, value) pairs gated by capture-info settings.
+
+    Subkeys are unprefixed (``client_version``, not ``weave.client_version``)
+    so call sites can prefix or nest as their attribute shape requires.
+    """
+    items: list[tuple[str, str]] = []
+    if should_capture_client_info():
+        items.extend(
+            [
+                ("client_version", VERSION),
+                ("source", "python-sdk"),
+                ("sys_version", sys.version),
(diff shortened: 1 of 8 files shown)


weave/utils/capture_info.py
```diff
@@ -0,0 +1,37 @@
...
+import sys
+
+from weave.trace.settings import (
+    should_capture_client_info,
+    should_capture_system_info,
+)
+from weave.version import VERSION
+
+
+def get_capture_info() -> dict[str, str]:
```
andrewtruong, line 21: would rename: `get_system_info`

weave/utils/pii_redaction.py
```diff
@@ -71,14 +107,43 @@ def redact_recursive(value: Any) -> Any:
 
 
 def redact_pii_string(data: str) -> str:
-    analyzer = AnalyzerEngine()
-    anonymizer = AnonymizerEngine()
+    """Redact PII in a single string."""
+    if not data:
+        return data  # _get_engines() is uncached; skip the engine-load cost on empty defaults
```
andrewtruong: I dont understand this comment

weave/utils/pii_redaction.py
```diff
@@ -71,14 +107,43 @@ def redact_recursive(value: Any) -> Any:
...
+    analyzer, anonymizer = _get_engines()
     entities = _get_redaction_entities()
     results = analyzer.analyze(text=data, language="en", entities=entities)
     redacted = anonymizer.anonymize(text=data, analyzer_results=results)
     return redacted.text
 
 
+def redact_messages(messages: list[Message] | None) -> list[Message] | None:
```
andrewtruong: Is it ever valid for `messages` to be `None`?  Or is that a thing we're introducing?

weave/utils/pii_redaction.py
rgao-coreweave: basically empty string in, empty out.

It is an optimization that skips `_get_engines`. It matters because `_get_engines` is not cached, so if an empty string is passed in, it will occur cost of redaction even if it is empty.

Traces can have a lot of empty strings passed in, so we want this code in place.

weave/utils/capture_info.py
rgao-coreweave: The function returns both client info (`client_version`, `source`, `sys_version`) and system info (`os_name`, `os_version`, `os_release`), gated by two independent settings (`should_capture_client_info`() / `should_capture_system_info`()). 

Renaming to get_system_info would conflate it with one of the two gates, therefore keeping `get_capture_info`.

weave/utils/pii_redaction.py
rgao-coreweave: Tighten the signature to be List of messages. The empty-list short-circuit (if not messages) still handles [] cleanly.

rgao-coreweave: Good callout! thanks!

weave/utils/pii_redaction.py
```diff
@@ -109,15 +109,17 @@ def redact_recursive(value: Any) -> Any:
 def redact_pii_string(data: str) -> str:
     """Redact PII in a single string."""
     if not data:
-        return data  # _get_engines() is uncached; skip the engine-load cost on empty defaults
+        # Short-circuit empty input — _get_engines() is uncached and traces
+        # commonly contain empty strings, so loading Presidio every call is costly.
```
andrewtruong, line 113: well it seems silly that we have to reinstantiate the object every call.  Why not cache it?

andrewtruong: That can be a separate PR

andrewtruong (approved):

weave/utils/pii_redaction.py
rgao-coreweave: agreed, this is an pre-existing issue. filed a ticket

PR #6947, opened 2026-05-22
feat(weave): hold eager call_start to pair into complete

gtarpenning:
## Summary
- Eager-flagged starts (`@op(eager_call_start=True)`, used by `Evaluation.evaluate` + `EvaluationLogger`) are held in `_pending_starts` for up to `eager_hold_window_seconds` (default 1.0s) waiting for their end. If the end arrives in the window, the pair flows through the existing `calls/complete` path as one `CompleteBatchItem` (one DB-friendly insert) instead of two separate v2/call/start + v2/call/end records.
- `_promote_expired_eager_starts` runs on every processor tick; starts whose hold deadline has passed are promoted to the eager v2 endpoint for UI visibility (legacy behavior). Their ends follow via the eager path as before.
- Adds `dev_docs/client_ingest_architecture.md` to document the ingest path this change touches.
- `eager_hold_window_seconds=0` preserves legacy immediate-eager dispatch.

## Testing
- `test_call_batch_processor.py`: new tests cover quick-end pairing, zero-hold-window legacy path, and promotion after window expiry. 13/13 in the file pass.
- `trace_server_bindings/` shard: 57/57 pass.

### DIFF
--- a/dev_docs/client_ingest_architecture.md
+++ b/dev_docs/client_ingest_architecture.md
@@ -21,7 +21,7 @@ How a `@weave.op` call gets from user code to the trace server. Written as groun
 | Feedback batching | `weave/trace_server_bindings/async_batch_processor.py` | AsyncBatchProcessor (used for feedback only) |
 | HTTP transport | `weave/trace_server_bindings/remote_http_trace_server.py` | dispatches single vs batch endpoints |
 | Oversize handling | `weave/trace_server_bindings/http_utils.py` | binary-tree split, 413 retry, repackage |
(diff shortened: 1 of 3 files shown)


gtarpenning:
Closing; this drifted from the original prediction.finish() perf investigation. The hold-eager-pair-into-complete change is a real structural improvement but doesn't move the bench that motivated this work (imperative eval predictions are normal calls, not eager). Restarting against the actual bottleneck. Architecture doc content preserved at `~/Desktop/specs/weave_client_ingest_architecture.md` for the next attempt.

PR #6954, opened 2026-05-24
feat(weave_ts): add Turn.setAttribute and Turn.addEvent

rgao-coreweave:
## Summary

Adds two narrow methods to the genai `Turn` class:

- `setAttribute(key: string, value: AttributeValue): this`
- `addEvent(name: string, attributes?: Attributes, startTime?: TimeInput): this`

Both methods are 1:1 mirrors of the underlying OTel `Span` API, are post-`end()` no-ops, and return `this` for chaining. No exposure of the underlying span itself.

## Motivation

Plugins that need to stamp running attributes (e.g. cumulative cost, usage totals) or mark non-span events (e.g. context compaction, tool loop detection) on an `invoke_agent` Turn span had no canonical surface to do so without piercing encapsulation. The two new methods provide a minimal API surface for this. Required by the openclaw-plugin-weave.

## Test Plan

- [x] `Turn.setAttribute` writes to the underlying span (verified via in-memory exporter)
- [x] `Turn.setAttribute` returns `this`
- [x] `Turn.setAttribute` is a no-op after `end()`
- [x] `Turn.addEvent` writes a named event with attributes
- [x] `Turn.addEvent` returns `this`
- [x] `Turn.addEvent` is a no-op after `end()`
- [x] Full genai test suite (Turn/Session/LLM/Tool/SubAgent): 245/245 passing

### DIFF
--- a/sdks/node/src/genai/turn.ts
+++ b/sdks/node/src/genai/turn.ts
@@ -1,9 +1,12 @@
 import {
+  type AttributeValue,
+  type Attributes,
   type Context,
   ROOT_CONTEXT,
   type Span,
   SpanKind,
   SpanStatusCode,
+  type TimeInput,
   trace,
 } from '@opentelemetry/api';
 
@@ -122,6 +125,34 @@ export class Turn {
       parentContext: this.context,
       conversationId: this.conversationId,
     });
+  }
+
+  /**
+   * Set a single attribute on the Turn span. Useful for stamping running
+   * totals (e.g. cumulative cost, token usage) or other metadata that becomes
+   * known mid-turn. No-op after `end()`. Mirrors OTel `Span.setAttribute`.
+   *
+   * @example
+   * turn.setAttribute('gen_ai.usage.input_tokens', totalInputTokens);
+   */
+  setAttribute(key: string, value: AttributeValue): this {
+    if (this._ended) return this;
+    this.span.setAttribute(key, value);
+    return this;
+  }
+
+  /**
+   * Add a named event to the Turn span. Useful for marking non-span moments
+   * such as context compaction, tool-loop detection, or guardrail trips.
+   * No-op after `end()`. Mirrors OTel `Span.addEvent`.
+   *
+   * @example
+   * turn.addEvent('context_compacted', {removedMessages: 12});
+   */
+  addEvent(name: string, attributes?: Attributes, startTime?: TimeInput): this {
+    if (this._ended) return this;
+    this.span.addEvent(name, attributes, startTime);
+    return this;
   }
 
   /** Close the Turn span. Idempotent. Pass `error` to mark it as failed. */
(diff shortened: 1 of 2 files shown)


sdks/node/src/index.ts
```diff
@@ -38,6 +38,12 @@ export {
   startTool,
   startTurn,
 } from './genai';
+// Low-level: get the Weave-configured OTel tracer so plugins /
+// integrations can emit spans directly into the Weave pipeline
+// (resource attrs, OTLP exporter, auth). The tracer is a no-op
+// until `init()` has been called. `init()` itself requires
+// WANDB_API_KEY or a ~/.netrc entry to be present.
+export {getWeaveTracer} from './genai/provider';
```
drtangible: Think this is a separate change w/another PR (#6944), shall we remove from this one?

rgao-coreweave: I am not sure what does `shall we remove from this one?` mean?

I checked that PR it doesn't look like it is covered this.

drtangible: Aren't the changes here https://github.com/wandb/weave/pull/6944/changes the same as this change? I wasn't sure if it was intentional here, or if we only want to introduce `setAttribute` and `addEvent` in this PR.

rgao-coreweave: oh my bad, I clicked the wrong PR number. I clicked the link and yes you were right.

Let me prepare a PR and close out the other.

drtangible: Thanks! Can we update the PR description w/details about why we're exposing `getWeaveTracer` now? Still currently just describing the other two fns added.


I'm a bit curious what that's being used for. Seems like if we expose that, we couple our public API with the `Tracer` interface provided by `@opentelemetry/api`. If the reason we need it exposed is well-defined, I could see benefit in adding a function on `weave` that does whatever we need it to, so we have more control over breaking changes.

sdks/node/src/genai/turn.ts
```diff
@@ -99,6 +102,22 @@ export class Turn {
     });
   }
 
+  setAttribute(key: string, value: AttributeValue): this {
+    if (this._ended) return this;
+    this.span.setAttribute(key, value);
+    return this;
+  }
+
+  addEvent(
```
drtangible: After pulling latest, the methods in here should all have some basic jsdocs (that'll eventually be extracted to docs.wandb.ai) -- mind adding to these two new fns so we get them covered in the SDK reference?

sdks/node/src/index.ts
rgao-coreweave: Good pushback on `getWeaveTracer`, we shouldn't expose it to SDK users.

Let's remove it for now, and I will do a workaround in openclaw plugin.

sdks/node/src/genai/turn.ts
rgao-coreweave: Good feedback, done!

Ty

drtangible (approved):
🔥 🔥 🔥

sdks/node/src/index.ts
drtangible: Ok, cool! This LGTM :shipit:

andrewtruong:
merging as all checks pass and it does not affect core

PR #6962, opened 2026-05-26
fix(weave): bump vcrpy 7 -> 8, fixes crewai async flake

gtarpenning:
## Summary

Bumps `vcrpy` 7.0.0 → 8.x. `vcrpy` 8.0.0 (release [PR #943](https://github.com/kevin1024/vcrpy/pull/943)) rewrote httpx support to patch at the `httpcore` layer instead of `httpx`, explicitly fixing the [`ResponseNotRead`](https://github.com/kevin1024/vcrpy/issues/832) issue ([also #834](https://github.com/kevin1024/vcrpy/issues/834)) we were hitting on Windows.

The old 7.0.0 stub (`vcr/stubs/httpx_stubs.py:_from_serialized_response`) was decorated with `@patch("httpx.Response.read", MagicMock())`, which replaces `Response.read` at the **class level** for the duration of the call -- visible to every thread. crewai's `kickoff_for_each_async` runs each kickoff via `asyncio.to_thread(self.kickoff, ...)`, so two sync OpenAI completions execute on separate threads. If thread A is mid-replay (class-level mock active) while thread B is inside `httpx.Client.send` running its post-replay `response.read()` (`httpx/_client.py:921`, the call that normally populates `_content` from the ByteStream), B's read resolves to A's MagicMock and is a no-op. B's response has no `_content`, and litellm's later `raw_response.parse()` -> `LegacyAPIResponse._parse` -> `response.json()` -> `httpx.Response.content` raises `ResponseNotRead`. Windows GIL timing widens the overlap window enough for the test to fail.

`vcrpy` 8.x's `httpcore_stubs.py` constructs an `httpcore.Response` directly with the bytes; httpx then runs its normal response-construction path against the httpcore response, so `_content` is populated correctly without any class-level mocking.

**Repro (`openai.with_raw_response.create` + `.parse()` under VCR replay, 8 concurrent threads, 50 trials = 400 calls):**

| | Failures |
|---|---|
| vcrpy 7.0.0 | 399 / 400 |
| vcrpy 8.1.1 | 0 / 400 |

8.0.0 breaking changes are non-issues here: drops Python 3.9 (weave needs >=3.10) and urllib3<2 (lockfile already on 2.6.3).

Original failure ([CI job](https://github.com/wandb/weave/actions/runs/26458318344/job/77898646520?pr=6868)):
```
FAILED tests/integrations/crewai/crewai_test.py::test_crewai_simple_crew_kickoff_for_each_async - litellm.exceptions.APIError: APIError: OpenAIException - error - Attempted to access streaming response content, without having called `read()`.
```

## Testing

local 8-thread stress repro (above). no test-side band-aids; the buggy test is unchanged.


bcsherma (approved):
🦸🏻

PR #6965, opened 2026-05-26
perf(weave): parallel bucket uploads inside upsert_batch

gtarpenning:
## Summary
- Jira: [WB-34755](https://coreweave.atlassian.net/browse/WB-34755)
- Fan out bucket uploads collected during a `call_batch` so an N-attachment upsert pays one parallel round-trip instead of N serial ones.
- New `BucketUploadBatch` module owns staging + parallel flush; `clickhouse_trace_server_batched.py` becomes a thin caller (`stage` during `file_create`, `flush` inside `_flush_all_batches_in_order` before the file_chunks insert).
- Preserves per-file `FileStorageWriteError` fallback to inline ClickHouse chunks. Non-batch path (`_flush_immediately=True`) is unchanged.
- Falls out of a Datadog investigation of `/call/upsert_batch` tail latency: today `_file_create_bucket` p50 = 61ms / p95 = 163ms and runs serially inside one worker, so heavy batches pay 11-20 sequential round-trips per request.

### Local
| | before | after (workers=8) |
|---|---|---|
| 20-attachment batch, GCS p50 = 60ms | ~1.2s | ~180ms |
| 50-attachment batch | ~3.0s | ~420ms |
| 1-attachment batch | unchanged | unchanged |
### QA
20 batches x 25 calls x 2 attachments x 32KB, 4 client threads (50 GCS uploads/batch), same endpoint before vs after the deploy:

|  | pre (master) | post (this PR) | speedup |
|---|---|---|---|
| p50 batch | 13.15s | 1.89s | 6.9x |
| p95 batch | 13.93s | 2.66s | 5.2x |
| p99 batch | 14.82s | 2.67s | 5.6x |
| attachments/sec | 14.7 | 94.1 | 6.4x |
| errors | 0/500 | 0/500 | - |

Variance compressed too: p99 lands ~tied with p95 post-deploy (wall time is now a few parallel waves instead of a long serial chain). Script: `qa-scripts/parallel_bucket_uploads_load_test.py`.

image

## Testing
two end-to-end tests in `tests/trace/test_server_file_storage.py`: parallel fan-out + within-batch dedup + project-prefix correctness (asserts `concurrent_peak >= 4` with the default 8-worker pool), and per-file `FileStorageWriteError` -> inline CH fallback (parametrized single-chunk + multi-chunk payloads, verified via `file_content_read` round-trip).

[WB-34755]: https://coreweave.atlassian.net/browse/WB-34755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/parallel_bucket_uploads.py
```diff
@@ -0,0 +1,169 @@
...
+# Hides GCS RTT (~60ms p50, ~150ms p95) without exhausting the per-pod
+# google-cloud-storage HTTP session pool.
+DEFAULT_BUCKET_UPLOAD_CONCURRENCY = 8
+
+
+@dataclass(slots=True)
+class _Pending:
+    req: tsi.FileCreateReq
+    digest: str
+    client: FileStorageClient
```
amwarrier: I don't like this. Why all _Pending have to carry the pointer to the same client? Its not part of the request. It should be passed only once to the one who actually does the upload, right?

weave/trace_server/parallel_bucket_uploads.py
```diff
@@ -0,0 +1,169 @@
...
+        """Defer a bucket upload until `flush()`.
+
+        Caller computes the digest, validates `expected_digest`, and dedups
+        against the in-flight ClickHouse chunk buffer. Within-batch dedup
+        across bucket uploads is tracked here via `has()`.
+        """
+        self._pending.append(_Pending(req=req, digest=digest, client=client))
+        self._seen.add((req.project_id, digest))
+
+    def has(self, project_id: str, digest: str) -> bool:
```
amwarrier, line 125: i think we are silently creating a contract here. Are you expecting the Bucket to always have its content de-duped? If yes, I would `raise` from the `stage` above if the digest is present on the seen. This way - you can enforce the caller to do:
```
if not bucket.has:
    bucket.stage()
```

weave/trace_server/parallel_bucket_uploads.py
gtarpenning: nice - `client` is `self.file_storage_client` (a per-instance property), so every `_Pending` in a batch was carrying the same pointer. fixed in 34f76ade10: `flush(client)` takes it once and `_upload_one` receives it via `pool.submit(..., p, client)`. dropped the field from `_Pending`.

weave/trace_server/parallel_bucket_uploads.py
gtarpenning: fair, raising in `stage()` on duplicate (project_id, digest) in a2d3236398 so the `if not bucket.has(): bucket.stage()` pattern is enforced... the existing caller already pre-checks via `has()` so no caller change.

weave/trace_server/parallel_bucket_uploads.py
```diff
@@ -0,0 +1,191 @@
...
+        race on `if_generation_match=0`, so we raise instead of silently
+        accepting it.
+        """
+        key = (req.project_id, digest)
+        if key in self._seen:
+            raise ValueError(
+                f"BucketUploadBatch.stage() called twice for {key}; callers "
+                "must pre-check via `has()` and skip duplicates."
+            )
+        self._pending.append(_Pending(req=req, digest=digest))
```
amwarrier, line 121: A client can potentially cause OOMs here. Should we have a guard rail here which tracks the size before we upload instead of blindly appending?

weave/trace_server/parallel_bucket_uploads.py
amwarrier: also add a unit test to make sure that the raise happens.

weave/trace_server/parallel_bucket_uploads.py
gtarpenning: sure, although we already have this vuln. i'll add it

amwarrier (approved):

PR #6969, opened 2026-05-26
chore(weave_ts): remove unnecessary type assertions in `openai.agents.ts`

drtangible:
## Description

* We're going to be reworking our OpenAI Agents SDK integration. While reviewing, noticed a few type assertions that appear unnecessary, so removing.

PR #6970, opened 2026-05-26
chore(weave_ts): Use types provided by @openai/agents

drtangible:
## Description

* We manually maintain some duplicated types for 3rd party libraries. This can mask type issues / bugs that can be caught by pointing to the library-provided types via `devDependencies`.

* This change removes some of the duplicated `@openai/agents` types.

### DIFF
--- a/sdks/node/src/integrations/openai.agent.types.ts
+++ b/sdks/node/src/integrations/openai.agent.types.ts
@@ -1,96 +1,26 @@
 /**
  * Type definitions for OpenAI Agents integration
  *
- * These types are duck-typed to match @openai/agents without importing from that package.
- * This allows users to use the integration without forcing a dependency on @openai/agents.
  */
 
 // ============================================================================
 // Span Data Types
 // ============================================================================
 
-/**
- * Base span data structure
- */
-type SpanDataBase = {
-  type: string;
-};
+import type {
+  AgentSpanData,
+  CustomSpanData,
+  FunctionSpanData,
+  GenerationSpanData,
+  GuardrailSpanData,
+  HandoffSpanData,
+  ResponseSpanData,
+  Span as OpenAIAgentsSpan,
+} from '@openai/agents';
 
-/**
- * Agent execution span data
- */
-type AgentSpanData = SpanDataBase & {
-  type: 'agent';
-  name: string;
-  handoffs?: string[];
-  tools?: string[];
-  output_type?: string;
-};
+export type {CustomSpanData, Trace, TracingProcessor} from '@openai/agents';
 
-/**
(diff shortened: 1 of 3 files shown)


drtangible:
* **#6983** <a href="https://app.graphite.com/github/pr/wandb/weave/6983?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6984](https://github.com/wandb/weave/pull/6984) <a href="https://app.graphite.com/github/pr/wandb/weave/6984?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6985](https://github.com/wandb/weave/pull/6985) <a href="https://app.graphite.com/github/pr/wandb/weave/6985?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6970** <a href="https://app.graphite.com/github/pr/wandb/weave/6970?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6970?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6969** <a href="https://app.graphite.com/github/pr/wandb/weave/6969?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sdks/node/src/integrations/openai.agent.types.ts
```diff
@@ -1,96 +1,26 @@
...
-
-/**
- * Agent handoff span data
- */
-type HandoffSpanData = SpanDataBase & {
-  type: 'handoff';
-  from_agent?: string;
-  to_agent?: string;
-};
+export type {CustomSpanData, Trace, TracingProcessor} from '@openai/agents';
```
drtangible, line 21: Note: eventually we can just use types directly from the library, just keeping change minimal w/this PR.

andrewtruong: Just want to make sure: will this incur cost for anyone not using openai?

andrewtruong (approved):

sdks/node/src/integrations/openai.agent.types.ts
drtangible: It should not impact consumers of the `weave` package who are not using OpenAI. `type` imports don't make their way to the built `dist/` customers pull down when installing the node module.

(Note: we should be able to add some safeguards to ensure types are imported correctly to prevent adding a runtime dependency on 3rd-party libs we don't wish to. (eg, via `import {CustomSpanData} from '@openai/agents'` without `type`) Will follow up with that.)

PR #6972, opened 2026-05-27
fix(weave): validate image_completion URLs before fetch

gtarpenning:
## Summary

Fixes [WB-34796](https://coreweave.atlassian.net/browse/WB-34796). `_process_image_data_item` passed the upstream litellm response URL straight into `Content.from_url`, which performs an unrestricted server-side HTTP GET. A malformed or attacker-influenced upstream URL could reach cloud metadata (`169.254.169.254`), loopback, RFC1918 ranges, or `file://`.

Rather than hand-roll another URL check (we already have one in `provider.py`, plus a third lighter check in `remote_scorer.py`), extracts a shared `weave/trace_server/helpers/url_safety.py` with `is_publicly_routable_url(url) -> bool` and routes both `image_completion` and `provider` through it.

Tightens two gaps the previously-shipped `provider.py` check had:
- IPv4/IPv6 multicast (`224.0.0.1`, `ff00::/8`), `is_global` returned True
- `localhost`, `*.localhost`, `localhost.localdomain`, accepted as hostnames

Defense in depth, primary control is pod egress policy.

## Testing

57 parametrized cases in new `tests/trace_server/test_url_safety.py` (schemes, malformed URLs, localhost variants, cloud-metadata hostnames, non-global IPv4/IPv6, alternative IPv4 encodings); 4 wiring cases in `test_image_completion.py` asserting `Content.from_url` is not called for non-publicly-routable URLs; existing `test_provider_base_url_validation` unchanged.

[WB-34796]: https://coreweave.atlassian.net/browse/WB-34796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/helpers/url_safety.py
jtschoonhoven, line 1: Nit: I think this should go under some kind of `./helpers` directory, rather than directly under `weave/trace_server` just to keep things organized around here

weave/trace_server/image_completion.py
```diff
@@ -46,6 +47,11 @@ def _process_image_data_item(
         # Handle URL-based images
         if "url" in data_item and data_item["url"] and trace_server and project_id:
             url = data_item["url"]
+            if not is_publicly_routable_url(url):
+                processed_item["error"] = (
+                    f"refusing to fetch image from disallowed URL at index {index}"
+                )
+                return processed_item
```
jtschoonhoven, line 58: You should emit a log here so we can audit this
```suggestion
            if not is_publicly_routable_url(url):
                processed_item["error"] = (
                    f"refusing to fetch image from disallowed URL at index {index}"
                )
                logger.warning("Blocked non-publicly routable url %s", url)
                return processed_item
```

jtschoonhoven (approved):

weave/trace_server/helpers/url_safety.py
gtarpenning: Moved to `weave/trace_server/helpers/url_safety.py` with a stub `helpers/__init__.py`. Updated the two import sites and the test module docstring.

weave/trace_server/image_completion.py
gtarpenning: Added. `logger.warning("Blocked non-publicly routable url %s", url)` with `logger = logging.getLogger(__name__)` at module top.

PR #6975, opened 2026-05-27
feat(weave): bucket project_id partition keys for HPA-friendly scaling

gtarpenning:
## Summary
- Adds `WF_KAFKA_PROJECT_ID_BUCKET_COUNT` env var (default `1`, no-op). When `N > 1`, the produce key on `weave.call_ended` and `weave.score_calls` becomes `project_id:crc32(salt) % N`, spreading one hot project across up to N partitions/workers so we can scale on consumer-group lag.
- `produce_call_end` salts on `call_end.id`; `produce_score_calls` on the chunk's first call_id. `produce_score_agent_spans` untouched (already keyed by conversation/trace).

## Prereqs before setting `N > 1` in prod
- alert_worker `recent_alerts_cache` is a per-pod 60s TTLCache that assumes project locality; splitting a project across pods can fire duplicate alerts (move to redis).
- Confirm `weave.call_ended` / `weave.score_calls` topics have `> 1` partitions, else bucketing is a silent no-op.

## Testing
unit tests for the env helper and the bucketed-key formula

### DIFF
--- a/tests/trace_server/test_agent_scorer_event_producer.py
+++ b/tests/trace_server/test_agent_scorer_event_producer.py
@@ -10,7 +10,7 @@ from weave.trace_server.agents.kafka_events import (
     SCORE_AGENT_SPANS_TOPIC,
     ScoreAgentSpansEvent,
 )
-from weave.trace_server.kafka import KafkaProducer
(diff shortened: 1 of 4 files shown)


gtarpenning:
bad idea

PR #6976, opened 2026-05-27
feat(weave): use content for attached media

zbirenbaum:
## Description

Replaces raw media storage (blobs, URIs, file IDs) in `MediaAttachment` with published `Content` object refs. When `LLM.attach_media` or `attach_media_url` is called, the media is immediately published as a `Content` object and only the `weave://` ref URI is stored on the attachment.

Key changes:
- `MediaAttachment` now holds a single `ref: str` field instead of `kind`/`content`/`uri`/`file_id`
- New `_publish_media_content()` helper converts raw bytes, base64, data URLs, and plain URLs into published `Content` objects
- OTel serialization emits all attachments as `uri` parts pointing to the content ref
- New `weave.content_refs` span attribute lists all content refs for downstream indexing
- `NormalizedMessage` and ClickHouse schema gain a `parts` field for lossless rendering of multimodal messages
- Registered `audio/wav` MIME type in mimetypes module

## Testing

- Updated unit tests in `test_session.py` and `test_session_otel.py` to mock `_publish_media_content` and validate ref-based attachments
- All existing session and OTel serialization tests pass with the new model

### DIFF
--- a/tests/session/test_session_otel.py
+++ b/tests/session/test_session_otel.py
@@ -1766,6 +1766,7 @@ class TestAttachMediaUrl:
     ) -> None:
         llm = LLM()
         llm.attach_media_url(url, modality="image")
+        llm._await_uploads()
         (att,) = llm.media_attachments
         assert att.ref.startswith("weave:///")
         assert att.mime_type == expected_mime
@@ -1778,6 +1779,7 @@ class TestAttachMediaUrl:
         att = llm.media_attachments[0]
         assert att.modality == "image"
         assert att.mime_type == "image/jpeg"
+        llm._await_uploads()
 
     def test_empty_url_ignored(self) -> None:
         llm = LLM()
@@ -1789,6 +1791,7 @@ class TestAttachMediaUrl:
         llm = LLM()
         result = llm.attach_media_url("https://e.com/a.png", modality="image")
         assert result is llm
+        llm._await_uploads()
 
 
 class TestLLMRecord:
--- a/tests/session/test_session_otel.py
+++ b/tests/session/test_session_otel.py
@@ -2,11 +2,11 @@
 
 from __future__ import annotations
 
-import base64
 import json
 import time
 from datetime import datetime, timedelta, timezone
 from typing import Any
+from unittest.mock import patch
 
 import pytest
 from opentelemetry import trace as otel_trace
@@ -390,17 +390,13 @@ class TestLLMAttributes:
         assert "finish_reason" not in raw[0]
         assert raw[1]["finish_reason"] == "length"
 
-    def test_attach_media_blob_serialized_as_blob_part(self) -> None:
+    def test_attach_media_serialized_as_uri_part(self) -> None:
+        ref = "weave:///test/project/object/img:v1"
         attrs = llm_attributes(
             model="gpt-4o",
             input_messages=[Message(role="user", content="what is this?")],
             media_attachments=[
-                MediaAttachment(
-                    kind="blob",
-                    modality="image",
-                    mime_type="image/png",
-                    content=b"\x89PNG",
-                )
+                MediaAttachment(ref=ref, modality="image", mime_type="image/png")
             ],
         )
         raw = json.loads(attrs["gen_ai.input.messages"])
@@ -410,59 +406,29 @@ class TestLLMAttributes:
                 "parts": [
                     {"type": "text", "content": "what is this?"},
                     {
-                        "type": "blob",
+                        "type": "uri",
                         "mime_type": "image/png",
                         "modality": "image",
-                        "content": base64.b64encode(b"\x89PNG").decode("ascii"),
+                        "uri": ref,
                     },
                 ],
             }
         ]
 
-    def test_attach_media_uri_serialized_as_uri_part(self) -> None:
+    def test_content_refs_emitted(self) -> None:
+        ref = "weave:///test/project/object/img:v1"
         attrs = llm_attributes(
             model="gpt-4o",
             input_messages=[Message(role="user", content="describe")],
             media_attachments=[
-                MediaAttachment(
-                    kind="uri",
-                    modality="image",
-                    mime_type="image/jpeg",
-                    uri="https://example.com/photo.jpg",
-                )
+                MediaAttachment(ref=ref, modality="image", mime_type="image/jpeg")
             ],
         )
-        raw = json.loads(attrs["gen_ai.input.messages"])
-        assert raw[0]["parts"][1] == {
-            "type": "uri",
-            "mime_type": "image/jpeg",
-            "modality": "image",
-            "uri": "https://example.com/photo.jpg",
-        }
-
-    def test_attach_media_file_serialized_as_file_part(self) -> None:
-        attrs = llm_attributes(
-            model="gpt-4o",
-            input_messages=[Message(role="user", content="transcribe")],
-            media_attachments=[
-                MediaAttachment(
-                    kind="file",
-                    modality="audio",
-                    mime_type="audio/wav",
-                    file_id="file-abc",
-                )
-            ],
-        )
-        raw = json.loads(attrs["gen_ai.input.messages"])
-        assert raw[0]["parts"][1] == {
-            "type": "file",
-            "mime_type": "audio/wav",
-            "modality": "audio",
-            "file_id": "file-abc",
-        }
+        assert attrs["weave.content_refs"] == [ref]
 
     def test_media_attached_to_last_user_message(self) -> None:
         """When multiple user messages exist, media goes on the most recent one."""
+        ref = "weave:///test/project/object/img:v1"
         attrs = llm_attributes(
             model="gpt-4o",
             input_messages=[
@@ -471,9 +437,7 @@ class TestLLMAttributes:
                 Message(role="user", content="last"),
             ],
             media_attachments=[
-                MediaAttachment(
-                    kind="uri", modality="image", mime_type="image/png", uri="u"
-                )
+                MediaAttachment(ref=ref, modality="image", mime_type="image/png")
             ],
         )
         raw = json.loads(attrs["gen_ai.input.messages"])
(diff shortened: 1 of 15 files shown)


weave/session/adapters/openai.py
```diff
@@ -220,15 +221,9 @@ def _collect_image_attachments(
 def _url_to_attachment(url: str) -> MediaAttachment:
     if url.startswith("data:"):
         mime_type, payload = _parse_data_url(url)
-        return MediaAttachment(
-            kind="blob",
-            modality="image",
-            mime_type=mime_type,
-            content=payload,
+        ref_uri = _publish_media_content(
```
rgao-coreweave, line 224: The industry standard for tracing SDKs is "instrumentation must never block the application"

Every image attachment now goes through `_publish_media_content` → `weave.publish()` with no skip path. 

this introduces synchronous requests.get with 30s timeout + synchronous upload. Is it possible to use async or this will be a known gap?

weave/session/session.py
```diff
@@ -184,6 +184,38 @@ def _record_otel_error(self, exc_val: BaseException) -> None:
...
+            content_obj = Content.from_base64(content, mimetype=mime_type or None)
+        else:
+            content_obj = Content.from_bytes(content, mimetype=mime_type or None)
+    elif uri:
+        if uri.startswith("data:"):
+            content_obj = Content.from_data_url(uri)
+        else:
+            content_obj = Content.from_url(uri)
+    else:
+        content_obj = Content.from_text(file_id, mimetype=mime_type or None)
```
rgao-coreweave: this publishes the literal `"file_id"` string as Content bytes

https://github.com/wandb/weave/blob/4939275f3ec90b04acf565ba3eb0c5f76d1e91e9/weave/type_wrappers/Content/content.py#L220-L223

weave/trace_server/agents/helpers.py
```diff
@@ -23,6 +23,7 @@ class MessageTuple(NamedTuple):
     role: str
     content: str
     finish_reason: str
+    parts: str
```
rgao-coreweave: do we need a db migration for `parts`??

rgao-coreweave (commented):
Thanks for working on fixing this, left some questions.

tests/session/test_session_otel.py
```diff
@@ -410,59 +406,29 @@ def test_attach_media_blob_serialized_as_blob_part(self) -> None:
                 "parts": [
                     {"type": "text", "content": "what is this?"},
                     {
-                        "type": "blob",
+                        "type": "uri",
```
andrewtruong, line 409: is everything URI now?

tests/trace/data_serialization/test_cases/media_cases.py
```diff
@@ -300,9 +300,7 @@ def markdown_equality_check(a, b):
             "weave_type": {"type": "weave.type_wrappers.Content.content.Content"},
             "files": {
                 "content": "xfOhnNfgQxRzgWZ6DC1QEGt9vrJWcathymKPPZQmmIw",
-                "metadata.json": "v5gYkptifYVGd0JYbW3U0jYQPSKkLzEI3eamc0cnCG8"
-                if sys.platform == "win32"
-                else "0tY4LYkQE9BXzCQDItzaUoFLd3lesQ0RkuHNMuXQJIk",
+                "metadata.json": "v5gYkptifYVGd0JYbW3U0jYQPSKkLzEI3eamc0cnCG8",
```
andrewtruong, line 303: Why did this change?

tests/trace/data_serialization/test_cases/media_cases.py
```diff
@@ -319,12 +317,8 @@ def markdown_equality_check(a, b):
         ],
         exp_files=[
             {
-                "digest": "v5gYkptifYVGd0JYbW3U0jYQPSKkLzEI3eamc0cnCG8"
-                if sys.platform == "win32"
-                else "0tY4LYkQE9BXzCQDItzaUoFLd3lesQ0RkuHNMuXQJIk",
-                "exp_content": b'{"size": 88244, "mimetype": "audio/wav", "digest": "c5f3a19cd7e043147381667a0c2d50106b7dbeb25671ab61ca628f3d9426988c", "filename": "audio.wav", "content_type": "file", "input_type": "str", "encoding": "utf-8", "metadata": null, "extension": ".wav"}'
-                if sys.platform == "win32"
-                else b'{"size": 88244, "mimetype": "audio/x-wav", "digest": "c5f3a19cd7e043147381667a0c2d50106b7dbeb25671ab61ca628f3d9426988c", "filename": "audio.wav", "content_type": "file", "input_type": "str", "encoding": "utf-8", "metadata": null, "extension": ".wav"}',
+                "digest": "v5gYkptifYVGd0JYbW3U0jYQPSKkLzEI3eamc0cnCG8",
```
andrewtruong, line 320: same here and rest of the file

tests/trace/type_handlers/Content/test_content.py
```diff
@@ -53,10 +52,9 @@ def audio_file(tmp_path_factory) -> Path:
 
 
 # New parameterization list using fixture names
-# Note: Windows uses "audio/wav" while Unix uses "audio/x-wav"
 MEDIA_TEST_PARAMS = [
     ("image_file", ".png", "image/png"),
-    ("audio_file", ".wav", "audio/wav" if sys.platform == "win32" else "audio/x-wav"),
+    ("audio_file", ".wav", "audio/wav"),
```
andrewtruong, line 57: Same here.  I guess this implies everything is happening server side now?

weave/session/session.py
```diff
@@ -184,6 +184,38 @@ def _record_otel_error(self, exc_val: BaseException) -> None:
...
+        if isinstance(content, str):
+            content_obj = Content.from_base64(content, mimetype=mime_type or None)
+        else:
+            content_obj = Content.from_bytes(content, mimetype=mime_type or None)
+    elif uri:
+        if uri.startswith("data:"):
+            content_obj = Content.from_data_url(uri)
+        else:
+            content_obj = Content.from_url(uri)
+    else:
```
andrewtruong, line 221: prefer making this case explicit and having `else` raise for unexpected case handling

weave/trace_server/agents/helpers.py
```diff
@@ -69,13 +72,14 @@ def normalize_span_row(d: dict[str, Any]) -> dict[str, Any]:
...
-                        f"{key} tuple must have 3 values (role, content, finish_reason), got {len(m)}"
+                        f"{key} tuple must have 3 or 4 values, got {len(m)}"
                     )
-                normalized.append(
-                    {"role": m[0], "content": m[1], "finish_reason": m[2]}
-                )
+                row = {"role": m[0], "content": m[1], "finish_reason": m[2]}
+                if len(m) == 4:
+                    row["parts"] = m[3]
+                normalized.append(row)
```
andrewtruong: This seems cryptic and would benefit from a detailed comment

weave/session/adapters/openai.py
andrewtruong: Agreed, that will be an issue here.

Can we calc the digest client side for ref construction and do a publish async?  I suspect there will be issues if the client is on windows, but I'm not sure what a better way would be

tests/trace/data_serialization/test_cases/media_cases.py
zbirenbaum: Windows had a different mimetype for WAV but now its the same across the board

tests/trace/type_handlers/Content/test_content.py
zbirenbaum: No, I just added 'wav' to the default python mimetypes library to fix a separate bug, so now all platforms detect it as audio/wav instead of just window.

weave/session/adapters/openai.py
zbirenbaum: Ideally we would have an RPC or daemon running on the user system which handles the upload and sends back a ref to export immediately. This is an intermediate step which is better than sending the full data inside the OTel span but obviously has some drawbacks...

weave/trace_server/agents/helpers.py
zbirenbaum: Nice catch, parts should be in content, not in a separate field

tests/session/test_session_otel.py
zbirenbaum: Yeah, since everything is a weave ref it should be a URI.

weave/session/session.py
zbirenbaum: This whole method feels a little off. It's doing what "_from_guess" does but with much less functionality and fewer cases handled.  I'll probably end up changing this but I want to figure out what to do for the file ID case. Updated to error when file_id is passed with instructions.

zbirenbaum:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/wandb/weave/6976?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7017** <a href="https://app.graphite.com/github/pr/wandb/weave/7017?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6976** <a href="https://app.graphite.com/github/pr/wandb/weave/6976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6976?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7018** <a href="https://app.graphite.com/github/pr/wandb/weave/7018?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/session/session.py
zbirenbaum: Fixed and added a not-supported message

weave/session/adapters/openai.py
zbirenbaum: > Agreed, that will be an issue here.
> 
> Can we calc the digest client side for ref construction and do a publish async? I suspect there will be issues if the client is on windows, but I'm not sure what a better way would be

We can do async uploads the way the SDK does. Remote files over HTTP breaks for pre-computed refs. Using background threads works for local files, raw data, and HTTP resolution. I added it in https://github.com/wandb/weave/pull/7017

We would also need to handle the scenario for a span being uploaded with a ref to an object that doesn't exist. It may be fine, but thats a large surface area which deserves more thorough testing. Lets leave it as a followup for now to #7017

andrewtruong (approved):

PR #6982, opened 2026-05-27
feat(weave): Support scorer_* fields for wandb.runnable scorers

benjreinhart:
## Description

Piggy backing off of @nikumar1206's work, starting to support the new structured feedback columns (scorer_*) for wandb.runnable scorers.

## Testing

Unit tests.

PR #6986, opened 2026-05-27
fix(weave): populate structured scorer feedback for runnable scores

benjreinhart:
## Description

For older runnable call scorers using llm-as-judge, if the llm output matches what we expect for the typed outputs (scorer_* columns), populate them in the to-be-inserted feedback row.

This should allow the runnable scorers to start adopting the new typed columns while remaining backwards compatible for the ones that exist.

I ported the code from core (pydantic models / validations) over here so we can reuse it. Will need to follow up after merge with an update to core to 🔪  the duplication and import from weave.

## Testing

Unit tests

### DIFF
--- a/weave/trace_server/feedback.py
+++ b/weave/trace_server/feedback.py
@@ -28,6 +28,7 @@ from weave.trace_server.interface.feedback_types import (
 )
 from weave.trace_server.orm import Column, Row, Table
 from weave.trace_server.refs_internal_server_util import ensure_ref_is_valid
+from weave.trace_server.scorer_feedback import ScorerColumns, ScorerFeedbackFields
 from weave.trace_server.validation import (
     validate_purge_req_multiple,
     validate_purge_req_one,
@@ -96,97 +97,31 @@ def process_feedback_payload(
     return processed_payload
 
 
-def _optional_float(value: Any) -> float | None:
-    if isinstance(value, bool):
-        return None
-    if isinstance(value, (int, float)):
-        return float(value)
-    if isinstance(value, str):
-        try:
-            return float(value)
-        except ValueError:
-            return None
-    return None
-
-
-def _optional_reason(value: Any) -> str | None:
-    if isinstance(value, str) and value:
-        return value
-    return None
-
-
-def _output_to_score_items(output: Any) -> list[dict[str, Any]]:
-    if isinstance(output, list):
-        return [item for item in output if isinstance(item, dict)]
-
-    if not isinstance(output, dict):
-        return []
-
-    scores = output.get("scores")
-    if isinstance(scores, list):
-        return [item for item in scores if isinstance(item, dict)]
-
-    if "value" in output:
-        return [output]
-
-    return []
+_MISSING_OUTPUT = object()
 
 
 def _derive_scorer_fields_from_payload(
     feedback_req: tsi.FeedbackCreateReq,
     processed_payload: dict[str, Any],
-) -> dict[str, Any]:
+) -> ScorerFeedbackFields:
     if not (
         feedback_type_is_runnable(feedback_req.feedback_type)
         or feedback_type_is_agent_monitor(feedback_req.feedback_type)
     ):
         return {}
 
-    output = processed_payload.get("output")
-    scorer_tags = []
-    scorer_tag_reasons = {}
-    scorer_tag_confidences = {}
-    scorer_ratings = {}
-    scorer_rating_reasons = {}
-    scorer_rating_confidences = {}
-    has_score_item = False
-
-    for score_item in _output_to_score_items(output):
-        value = score_item.get("value")
-        reason = _optional_reason(score_item.get("reason"))
-        confidence = _optional_float(score_item.get("confidence"))
-
-        if isinstance(value, str) and value:
-            has_score_item = True
-            scorer_tags.append(value)
-            if reason is not None:
-                scorer_tag_reasons[value] = reason
-            if confidence is not None:
-                scorer_tag_confidences[value] = confidence
-            continue
-
-        rating = _optional_float(value)
-        if rating is None:
-            continue
-
-        has_score_item = True
-        rating_name = "_rating_"
-        scorer_ratings[rating_name] = rating
-        if reason is not None:
-            scorer_rating_reasons[rating_name] = reason
-        if confidence is not None:
-            scorer_rating_confidences[rating_name] = confidence
-
-    if not has_score_item:
+    output = processed_payload.get("output", _MISSING_OUTPUT)
+    if output is _MISSING_OUTPUT:
         return {}
-    return {
-        "scorer_tags": scorer_tags,
-        "scorer_tag_reasons": scorer_tag_reasons,
-        "scorer_tag_confidences": scorer_tag_confidences,
-        "scorer_ratings": scorer_ratings,
-        "scorer_rating_reasons": scorer_rating_reasons,
-        "scorer_rating_confidences": scorer_rating_confidences,
-    }
+
+    try:
+        return ScorerColumns.from_raw_output(output).to_feedback_fields()
+    except (TypeError, ValueError, ValidationError) as e:
+        if feedback_type_is_agent_monitor(feedback_req.feedback_type):
+            raise InvalidRequest(
+                f"Invalid scorer output for feedback_type {feedback_req.feedback_type}: {e}"
+            ) from e
+        return {}
 
 
 def validate_feedback_create_req(
@@ -374,7 +309,7 @@ def format_feedback_to_row(
     feedback_id = feedback_req.id or generate_id()
     created_at = datetime.datetime.now(ZoneInfo("UTC"))
 
-    request_scorer_fields = {
+    request_scorer_fields: ScorerFeedbackFields = {
         "scorer_tags": feedback_req.scorer_tags,
         "scorer_tag_reasons": feedback_req.scorer_tag_reasons,
         "scorer_tag_confidences": feedback_req.scorer_tag_confidences,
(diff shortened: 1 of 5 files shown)


benjreinhart:
* **#6986** <a href="https://app.graphite.com/github/pr/wandb/weave/6986?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/6986?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6982** <a href="https://app.graphite.com/github/pr/wandb/weave/6982?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

weave/trace_server/feedback.py
```diff
@@ -96,6 +96,99 @@ def process_feedback_payload(
...
+    if "value" in output:
+        return [output]
+
+    return []
+
+
+def _derive_scorer_fields_from_payload(
+    feedback_req: tsi.FeedbackCreateReq,
+    processed_payload: dict[str, Any],
+) -> dict[str, Any]:
```
jtschoonhoven: You should be able to replace most of this by reusing code from https://github.com/wandb/core/pull/44328/changes#diff-daee47bf3fb4333f5265212ed054429a7a15ad728e17225fa7a09993400d04f0

E.g. this is similar to `ScorerLlmOutputGroup.from_agent_scorer_output()`

It probably makes sense to pull some of those classes out of the agent scoring worker and into a shared module.

weave/trace_server/feedback.py
```diff
@@ -281,6 +374,27 @@ def format_feedback_to_row(
...
     created_at = datetime.datetime.now(ZoneInfo("UTC"))
 
+    request_scorer_fields = {
+        "scorer_tags": feedback_req.scorer_tags,
+        "scorer_tag_reasons": feedback_req.scorer_tag_reasons,
+        "scorer_tag_confidences": feedback_req.scorer_tag_confidences,
+        "scorer_ratings": feedback_req.scorer_ratings,
+        "scorer_rating_reasons": feedback_req.scorer_rating_reasons,
+        "scorer_rating_confidences": feedback_req.scorer_rating_confidences,
+    }
```
jtschoonhoven: In https://github.com/wandb/core/pull/44328/changes#diff-daee47bf3fb4333f5265212ed054429a7a15ad728e17225fa7a09993400d04f0 there's a pydantic model `ScorerColumns` you could use to parse and validate these.

jtschoonhoven (commented):
We should DRY this up with the code in [this branch](https://github.com/wandb/core/pull/44328) of [agent_scoring_types.py](https://github.com/wandb/core/pull/44328/changes#diff-daee47bf3fb4333f5265212ed054429a7a15ad728e17225fa7a09993400d04f0)

Apologies if you weren't aware of that branch, hate to duplicate work.

benjreinhart:
@jtschoonhoven i was not aware but sounds good to me, excited to simplify+reuse. I'll update this when that gets merged.

weave/trace_server/scorer_feedback.py
```diff
@@ -0,0 +1,166 @@
...
+    value: _TagType | _RatingType
+    reason: str | None = Field(default=None, max_length=_MAX_REASON_LENGTH)
+    confidence: float | None = Field(default=None, ge=0.0, le=1.0)
+
+    @field_validator("value", "confidence", mode="before")
+    @classmethod
+    def reject_bool_as_number(cls, v: Any) -> Any:
+        if isinstance(v, bool):
+            raise ValueError("boolean values are not valid scorer numbers")  # noqa: TRY004
+        return v
```
jtschoonhoven, line 44: Tiny nit: not sure it's worth bothering to check this. On the off-chance the LLM returns `{"value": "myTag", "confidence": true}` it seems fine to interpret that as `1.0`. But nbd either way.

weave/trace_server/scorer_feedback.py
```diff
@@ -0,0 +1,166 @@
...
+        if not isinstance(obj, dict):
+            raise TypeError(f"Invalid scorer output type: {type(obj).__name__}")
+        if isinstance(obj.get("scores"), list):
+            return ScorerLlmOutputGroup.model_validate(obj)
+        return ScorerLlmOutputGroup(scores=[ScorerLlmOutputSchema.model_validate(obj)])
+
+    @staticmethod
+    def from_agent_scorer_output(obj: Any) -> ScorerLlmOutputGroup:
+        """Backward-compatible alias for the agent scoring worker."""
+        return ScorerLlmOutputGroup.from_raw_output(obj)
```
jtschoonhoven, line 70: I don't see the need to maintain an alias for `from_raw_output` I think we can drop this method.

weave/trace_server/scorer_feedback.py
```diff
@@ -0,0 +1,166 @@
...
+        return v
+
+
+class ScorerLlmOutputGroup(BaseModel):
+    """Schema for a list of scorer tag or rating outputs."""
+
+    scores: list[ScorerLlmOutputSchema]
+
+    @staticmethod
+    def from_raw_output(obj: Any) -> ScorerLlmOutputGroup:
```
jtschoonhoven, line 53: Nit: this was previously typed `obj: AgentScorerOutput` why did it change? Better to use more specific types when possible.

weave/trace_server/scorer_feedback.py
```diff
@@ -0,0 +1,166 @@
...
+    tag_reasons: dict[str, str] = Field(default_factory=dict)
+    tag_confidences: dict[str, float] = Field(default_factory=dict)
+    rating_reasons: dict[str, str] = Field(default_factory=dict)
+    rating_confidences: dict[str, float] = Field(default_factory=dict)
+
+    @classmethod
+    def from_raw_output(cls, output: Any) -> ScorerColumns:
+        """Return normalized `feedback.score_*` columns from raw scorer output."""
+        group = ScorerLlmOutputGroup.from_raw_output(output)
+        return cls.from_scorer_llm_output_group(group)
```
jtschoonhoven, line 89: Nit: again, I think it's better to type methods explicitly when possible and avoid using `Any`.

weave/trace_server/scorer_feedback.py
```diff
@@ -0,0 +1,166 @@
...
+    def to_feedback_fields(self) -> ScorerFeedbackFields:
+        """Return the column names/values expected by FeedbackCreateReq and feedback rows."""
+        return {
+            "scorer_tags": self.tags,
+            "scorer_tag_reasons": self.tag_reasons,
+            "scorer_tag_confidences": self.tag_confidences,
+            "scorer_ratings": self.ratings,
+            "scorer_rating_reasons": self.rating_reasons,
+            "scorer_rating_confidences": self.rating_confidences,
+        }
```
jtschoonhoven, line 116: What is this for? `ScorerColumns` exists to validate and type the LLM output, but this method drops all that type info and coerces it back to `dict[str, Any]`. Seems like this should at least return a TypedDict.

weave/trace_server/scorer_feedback.py
```diff
@@ -0,0 +1,166 @@
...
+import logging
+import re
+from typing import Annotated, Any, TypeAlias
+
+from pydantic import BaseModel, Field, field_validator
+
+logger = logging.getLogger("weave.trace_server.scorer_feedback")
+
+ScorerFeedbackFieldValue: TypeAlias = list[str] | dict[str, Any]
+ScorerFeedbackFields: TypeAlias = dict[str, ScorerFeedbackFieldValue]
```
jtschoonhoven, line 12: This dict has a fixed set of keys, so it should be a `TypedDict`.

weave/trace_server/scorer_feedback.py
```diff
@@ -0,0 +1,166 @@
...
+                _MAX_TAGS_PER_ROW,
+                tag,
+            )
+            return
+        if tag := self._sanitize_tag(tag):
+            if tag in self.tags:
+                logger.warning("Skipping duplicate tag '%s'", tag)
+                return
+            self.tags.append(tag)
+            self.tags.sort()
```
jtschoonhoven, line 132: Tiny nit: this drops a comment I added:

```suggestion
            self.tags.sort()  # Inefficient, but good enough for very short lists
```

weave/trace_server/feedback.py
```diff
@@ -96,6 +97,33 @@ def process_feedback_payload(
...
+) -> ScorerFeedbackFields:
+    if not (
+        feedback_type_is_runnable(feedback_req.feedback_type)
+        or feedback_type_is_agent_monitor(feedback_req.feedback_type)
+    ):
+        return {}
+
+    output = processed_payload.get("output", _MISSING_OUTPUT)
+    if output is _MISSING_OUTPUT:
+        return {}
```
jtschoonhoven, line 115: Nit: this works exactly the same with `None` instead of `_MISSING_OUTPUT` so there's no need to define a special sentinel value just for this one check. I'd drop `_MISSING_OUTPUT`

weave/trace_server/feedback.py
```diff
@@ -96,6 +97,33 @@ def process_feedback_payload(
...
+
+    try:
+        return ScorerColumns.from_raw_output(output).to_feedback_fields()
```
jtschoonhoven, line 118: Older Signal monitors use a different schema, so this won't work. Check out `defaultClassifiersConfig.ts`in the core repo.

weave/trace_server/feedback.py
```diff
@@ -281,6 +309,27 @@ def format_feedback_to_row(
...
+
+    # If typed fields (scorer_*) exist on the feedback req object,
+    # use those values. Otherwise, try to derive them from the output
+    if any(request_scorer_fields.values()):
+        scorer_fields = request_scorer_fields
+    else:
+        scorer_fields = _derive_scorer_fields_from_payload(
+            feedback_req,
+            processed_payload,
+        )
```
jtschoonhoven, line 331: I think this is the wrong place to add backwards-compatibility. Why not just update the scoring worker itself?

IIRC Griffin updated the scoring worker to write to ClickHouse directly without passing through this endpoint (worth double checking)

jtschoonhoven (commented):
Thanks for working on this, I left some nitpicks and some structural questions. Happy to take another look when it's ready!

PR #6988, opened 2026-05-28
ci(weave): add replicated CH backends to nightly tests

gtarpenning:
## Summary
- Adds `nightly-tests-replicated-1s3r` and `nightly-tests-replicated-2s2r` to `nightly-tests.yaml`, reusing the existing topology compose files in `tests/trace_server/conftest_lib/topologies/`.
- Two production bugs surfaced by the replicated topologies are fixed here:
  - **Double-fire MV**: distributed-mode `CREATE MATERIALIZED VIEW` was leaving the MV pointed at Distributed-engine source/target; later `ALTER ... MODIFY QUERY` added a `_local` twin, so every insert fired both MVs. Migrator now creates only the `_local` variant. `DROP VIEW` also drops the `_local` twin.
  - **call_parts cross-shard split**: call_parts was sharded by `rand()`, so call_start (parent_id=root_id) and call_end (parent_id=NULL) for the same call could land on different shards; `parent_id IS NULL` then matched the orphaned call_end row for every child. Sharded by `id` now (id is non-Nullable on every part).
- Test-only fixture changes for replication-lag determinism:
  - forward `WF_CLICKHOUSE_REPLICATED*` env into the conftest migrator
  - `ON CLUSTER` truncate + drop in the session fixture so peer replicas don't keep stale state
  - distributed-aware `force_optimize_calls_merged` helper targeting `calls_merged_local ON CLUSTER` in distributed mode
  - `users.d/test_overrides.xml` mounted only into the test CH nodes:
    - `distributed_foreground_insert=1` so write-then-read is deterministic
    - `load_balancing=in_order` so Distributed reads hit the replica that received the write (no replication-lag race). Production semantics unchanged.

## Testing
- Local 2s2r: `test_call_stats` + `test_call_lifecycle` 14/14, `test_calls_complete` 26/26, `test_clickhouse_trace_server_batched` 41/41, migrator unit tests 100/100.
- CI smoke: https://github.com/wandb/weave/actions/runs/26554846068

### DIFF
--- a/.github/workflows/nightly-tests.yaml
+++ b/.github/workflows/nightly-tests.yaml
@@ -192,10 +192,368 @@ jobs:
         run: |
           .github/scripts/retry.sh nox -e "tests-${{ matrix.python-version-major }}.${{ matrix.python-version-minor }}(shard='${{ matrix.shard }}')"
 
+  nightly-tests-replicated-1s3r:
+    name: Nightly Tests (replicated 1s3r)
+    timeout-minutes: 90
+    runs-on: ubuntu-latest
+    env:
+      MAX_ATTEMPTS: 3
+    strategy:
+      matrix:
+        python-version-major: ["3"]
+        python-version-minor: [
+            "10",
+            "11",
+            "12",
+            "13",
+            #
+          ]
+        shard:
+          [
+            "trace",
+            "flow",
+            "trace_server",
+            "trace_server_bindings",
+            "anthropic",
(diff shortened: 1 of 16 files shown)


gtarpenning:
Superseded by the stacked breakdown #6996 → #7001 (6 draft PRs, one per logical unit + Jira ticket). Cumulative diff is identical; per-file delta verified to be 0 lines on all 8 touched files.

PR #6991, opened 2026-05-28
fix(weave): normalize datetime literals in ORM query path

gtarpenning:
## Summary
- `feedback_query` 500'd on `created_at` filters: ISO-8601 strings carrying a `T` separator / `Z` suffix (e.g. `2026-05-27T17:49:15.491230Z`) can't be compared against the `DateTime64(3)` column, so ClickHouse raises TYPE_MISMATCH (code 53). The generic ORM `where` path now normalizes datetime-column literals to ClickHouse's canonical `YYYY-MM-DD HH:MM:SS.ffffff` form, mirroring the existing calls-query-builder fix.
- To avoid duplicating that logic, the two datetime helpers (`parse_string_to_utc_timestamp`, `timestamp_to_datetime_str`) and the operand converter move down into `orm.py` (the base layer both paths import; `calls_query_builder` can't be imported back into `orm`). The calls path now delegates to the shared `maybe_convert_datetime_operands` instead of its own copy.
- Scope: only the binary comparison ops (`$eq`/`$gt`/`$lt`/`$gte`/`$lte`) normalize datetime literals, matching the calls path. `$in` against a datetime column is unchanged (different operand shape, and exact-timestamp `IN` lists aren't a real query pattern).
- [WB-34897](https://coreweave.atlassian.net/browse/WB-34897)

## Testing
unit coverage in test_orm.py for the conversion (incl. a non-datetime-column control) + a functional feedback_query test that filters on `created_at` and asserts on the returned row; calls path delegates to the shared helper so existing calls coverage applies.

[WB-34897]: https://coreweave.atlassian.net/browse/WB-34897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/orm.py
```diff
@@ -580,6 +587,59 @@ def python_value_to_ch_type(value: Any) -> str:
...
 
+def timestamp_to_datetime_str(timestamp: float) -> str:
+    """Convert a unix timestamp to a ClickHouse-compatible datetime string.
+
+    Args:
+        timestamp (int | float): Unix timestamp in seconds.
+
+    Returns:
+        str: Datetime string in the format `YYYY-MM-DD HH:MM:SS.ffffff`,
+            matching the precision of ClickHouse `DateTime64(6)` columns.
```
nikumar1206, line 598: i know this is just a lift and shift, but the calls and the feedback table seem to use `DateTime64(3)` not `DateTime64(6)`.
e.g
`services/weave-python/weave-public/weave/trace_server/migrations/003_feedback.up.sql:36`

i assume this just works because CH auto-handles the conversion for us?

nikumar1206: actually makes sense to me since we use both `DateTime64(3)` and `DateTime64(6)`. better to leave the extra floats there and have ch drop when not needed, rather than the other way around.

nikumar1206 (approved):

PR #7002, opened 2026-05-28
feat(weave): Add Initial Claude Agents SDK OTEL integration

andrewtruong:
Initial Claude Agents SDK integration in OTEL


tests/integrations/claude_agent_sdk/claude_agent_sdk_otel_test.py
```diff
@@ -0,0 +1,294 @@
...
+    assert "The answer is 4." in get_all_text(
+        get_messages(agent_span, "gen_ai.output.messages")
+    )
+
+    chat_span = chat_spans[0]
+    chat_attrs = get_attrs(chat_span)
+    assert chat_attrs["gen_ai.request.model"] == "claude-sonnet-4-6"
+    assert chat_attrs["gen_ai.conversation.id"] == "s-abc123"
+    assert chat_attrs["gen_ai.usage.input_tokens"] == 25
+    assert chat_attrs["gen_ai.usage.output_tokens"] == 10
```
drtangible: Looks like the `pi-coding-agent` integration sets `gen_ai.usage.total_tokens` -- though I don't see that attribute in our `semconv.py` mapping. Should I remove it from there?

Assume some of the other additional attributes set there (https://github.com/wandb/weave/blob/master/sdks/node/src/__tests__/integrations/piCodingAgent.test.ts#L270-L287) aren't relevant or possible to set in this integration's case?

weave/integrations/claude_agent_sdk/otel_integration.py
```diff
@@ -0,0 +1,433 @@
...
+
+    ``conversation_id_holder`` carries the SDK ``session_id`` across turns of a
+    single ``ClaudeSDKClient``: the ``system/init`` message (which holds the
+    session_id) is only sent on the first turn, so later turns must inherit it.
+    """
+    tracer = _tracer()
+    root = tracer.start_span("invoke_agent " + _AGENT_NAME)
+    token = otel_context.attach(otel_trace.set_span_in_context(root))
+    state = _TurnState(user_prompt=user_prompt)
+    if conversation_id_holder and conversation_id_holder[0]:
```
drtangible: The check on `converation_id_holder` below (https://github.com/wandb/weave/pull/7002/changes#diff-dde7d901a02de47152fdf7285a6e65dd6f81a91099d7236aebccadad784eea60R314-R315) uses `is not None`. Should we be consistent, or that intentional?

weave/integrations/claude_agent_sdk/otel_integration.py
```diff
@@ -0,0 +1,433 @@
...
+    conversation_id_holder: list[str] | None = None,
+) -> AsyncIterator[Any]:
+    """Wrap a message stream, emitting the span tree for one turn.
+
+    ``conversation_id_holder`` carries the SDK ``session_id`` across turns of a
+    single ``ClaudeSDKClient``: the ``system/init`` message (which holds the
+    session_id) is only sent on the first turn, so later turns must inherit it.
+    """
+    tracer = _tracer()
+    root = tracer.start_span("invoke_agent " + _AGENT_NAME)
```
drtangible: super small nit, just curious. Below we do `tracer.start_span(f"execute_tool {block.name}")`. In general do we have a preference between f-strings and appending?

drtangible (approved):
👍 👍 👍

tests/integrations/claude_agent_sdk/claude_agent_sdk_otel_test.py
drtangible: Separate comment: this might not be very Pythonic, but on the TS side, I'm hoping to lean into more exhaustive assertions as we maintain more integrations. YMMV, but when confirming to a spec, kinda like being able to see the full data picture rather than parsing individual assertions and trying to understand if they cover everything. eg...

```py
    assert chat_attrs == {
      'gen_ai.conversation.id': 's-abc123',
      'gen_ai.operation.name': 'chat',
      'gen_ai.provider.name': 'anthropic',
      'gen_ai.request.model': 'claude-sonnet-4-6',
      'gen_ai.output.messages': '[{"role": "assistant", "parts": [{"type": "text", ''"content": "The answer is 4."}]}]',
      'gen_ai.usage.input_tokens': 25,
      'gen_ai.usage.output_tokens': 10
    }
```

weave/integrations/claude_agent_sdk/otel_integration.py
andrewtruong: We should use fstrings -- fixed, thanks!

weave/integrations/claude_agent_sdk/otel_integration.py
andrewtruong: Yeah that's fair, changed for consistency

tests/integrations/claude_agent_sdk/claude_agent_sdk_otel_test.py
andrewtruong: Agreed, we can do it this way.  It can get a little noisy for nested dicts, but for plain dicts this is ok

PR #7004, opened 2026-05-28
fix(weave): refuse code-bearing custom objects on server-side decode

gtarpenning:
## Summary
- Server-side workers (eval) reconstruct user payloads; an `Op`, or any custom type that hits the packaged `load_op` fallback, runs user-uploaded code. Adds `WeaveClient.allow_unsafe_custom_obj_decode` (default `True`); the worker flips it off.
- With it off, `_decode_custom_obj` still decodes data-only types (images, audio) via their in-process serializer, but refuses `Op`/unknown types and the `load_op` fallback, so multimodal eval datasets keep working while code never loads. No active client fails closed.
- Also rejects op refs passed as the eval/model ref, which `client.get` would run before decode applies. Tracking: [WB-34909](https://coreweave.atlassian.net/browse/WB-34909)

## Testing
unit tests for `is_safe_to_decode` + the ref guard + `KNOWN_TYPES` sync; decode tests for data-only decode under the worker policy, `Op` refusal, and the forged `load_op` fallback; e2e eval rejects an `Op` dataset row at decode time.

[WB-34909]: https://coreweave.atlassian.net/browse/WB-34909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -105,18 +109,10 @@ def _run_evaluation(
...
 
-def _assert_safe_ref(client: WeaveClient, ref_uri: str, label: str) -> None:
-    """Read the raw object for a ref and reject it if it contains unsafe CustomWeaveType payloads."""
+def _assert_object_ref(ref_uri: str, label: str) -> None:
+    """Reject anything but a plain object ref. `OpRef` subclasses `ObjectRef`, so
+    exclude it explicitly: an op ref here would be loaded and executed.
+    """
     ref = Ref.parse_uri(ref_uri)
-    if not isinstance(ref, ObjectRef):
+    if not isinstance(ref, ObjectRef) or isinstance(ref, OpRef):
```
andrewtruong: Why do you need the OpRef check?  Now the func name is a bit confusing

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -43,10 +41,16 @@ def evaluate_model(args: EvaluateModelArgs) -> None:
...
 
-    # Validate raw payloads before deserialization.
-    # https://coreweave.atlassian.net/browse/VULNMGMT-1007
-    _assert_safe_ref(client, args.evaluation_ref, "evaluation_ref")
-    _assert_safe_ref(client, args.model_ref, "model_ref")
+    # This worker reconstructs user-supplied objects; it must never deserialize
+    # code-bearing custom objects (Op / load_op). The decode guard in custom_objs.py
+    # enforces this for every payload, including dataset rows fetched lazily during
+    # evaluation. https://coreweave.atlassian.net/browse/WB-34909
+    client._allow_unsafe_custom_obj_decode = False
```
andrewtruong: this check seems important enough to warrant something like a `require_secure_weave_client` helper

weave/trace/weave_client.py
```diff
@@ -398,6 +398,10 @@ def __init__(
         self.future_executor_fastlane = FutureExecutor(max_workers=parallelism_upload)
         self.ensure_project_exists = ensure_project_exists
 
+        # Internal: server-side workers (eval, scoring) flip this off so deserialization
+        # never reconstructs code-bearing custom objects. See custom_objs.py.
+        self._allow_unsafe_custom_obj_decode = True
```
andrewtruong, line 407: see previous comment

jtschoonhoven:
I attempted something similar here (now closed): https://github.com/wandb/weave/pull/6623

tests/trace_server/workers/test_evaluate_model_worker.py
```diff
@@ -1,78 +1,47 @@
...
+        ("Op", True, True),
+        ("TotallyMadeUp", True, True),
+        # Worker client (allow_unsafe=False): data-only types decode, code-loading
+        # ("Op") and unknown types are refused. The load_op fallback is gated separately
+        # (see test_unsafe_decode_disabled_refuses_code_bearing in test_custom_objs.py).
+        ("Op", False, False),
+        ("TotallyMadeUp", False, False),
+        ("PIL.Image.Image", False, True),
+        ("weave.type_wrappers.Content.content.Content", False, True),
+    ],
```
andrewtruong, line 25: please add `ids=` to make it easier to see the test cases

tests/trace/test_custom_objs.py
```diff
@@ -95,6 +97,46 @@ def make_datetime():
...
+    assert encoded["weave_type"]["type"] == "PIL.Image.Image"
+
+    client._allow_unsafe_custom_obj_decode = False
+    decoded = decode_custom_obj(encoded)
+    assert isinstance(decoded, Image.Image)
+    assert decoded.tobytes() == img.tobytes()
+
+
+@pytest.mark.parametrize(
+    "encoded",
```
andrewtruong, line 116: same here

weave/trace/weave_client.py
jtschoonhoven: This seems like something I should be able to set from an env var. Customers might also want to disable remote code execution, even if it's supposedly their own code.

gtarpenning: hmm, but this is the weave client, being run in the weave worker 🤔  not sure how we expose it to our backend as an env var but not to users. It would be totally useless if a user set this in their own client, or worse, it would break things that are fine for users on their own compute.

tests/trace_server/workers/test_evaluate_model_worker.py
gtarpenning: yeah done

weave/trace/weave_client.py
gtarpenning: done

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
gtarpenning: done

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
gtarpenning: the decode guard in `custom_objs.py` fires on `CustomWeaveType` payloads so technically if you had an Op inside a dataset row, we would then blindly call the load_op, so we just check to make sure you don't have an op here.
I think this also protects the initial worker call: `client.get(Ref.parse_uri(evaluation_ref))`. If `evaluation_ref` is an op ref, `client.get` loads and runs that op directly...

weave/trace/weave_client.py
jtschoonhoven: >It would be totally useless if a user set this in their own client, or worse, it would break things that are fine for users on their own compute.

As a user of weave I want the ability to disable local execution of code that's hosted remotely. Are many users relying on deserializing code-bearing objects? I know some users publish and share ops, but I don't have the sense that it's especially common.

Also, if there was an env var then we could set `ALLOW_UNSAFE_CUSTOM_OBJ_DECODE=false` in the helm charts and disable the behavior globally, rather than doing it in code service-by-service.

weave/trace/weave_client.py
```diff
@@ -398,6 +398,11 @@ def __init__(
         self.future_executor_fastlane = FutureExecutor(max_workers=parallelism_upload)
         self.ensure_project_exists = ensure_project_exists
 
+        # Internal: server-side workers (eval, scoring) flip this off via
+        # `require_secure_weave_client` so deserialization never reconstructs
+        # code-bearing custom objects. See custom_objs.py.
+        self._allow_unsafe_custom_obj_decode = True
```
jtschoonhoven: If I was a user and I read this block I would feel concerned 🤨 

Probably worth writing the comment with customers in mind, to explain why this _should_ be safe for them.

Also as a user I might want the same mechanism to disable code-bearing custom objects.

weave/trace_server/workers/evaluate_model_worker/evaluate_model_worker.py
```diff
@@ -38,12 +36,17 @@ def evaluate_model(args: EvaluateModelArgs) -> None:
...
+    # code-bearing custom objects (Op / load_op). The secure client locks the decode
+    # guard in custom_objs.py off, which enforces this for every payload, including
+    # dataset rows fetched lazily during evaluation.
+    # https://coreweave.atlassian.net/browse/WB-34909
+    client = require_secure_weave_client()
+
+    # An op ref passed as the evaluation/model ref would be loaded and run by
+    # `client.get` directly, before the decode guard applies, so reject it here.
+    _assert_non_op_ref(args.evaluation_ref, "evaluation_ref")
+    _assert_non_op_ref(args.model_ref, "model_ref")
```
jtschoonhoven: I think this comment is wrong and `client.get` now would raise an `UnsafeDeserializationError`. So the `_assert_non_op_ref` isn't necessary. But if I'm wrong that seems like a problem, since `client.get` should definitely check `_allow_unsafe_custom_obj_decode`

weave/trace/serialization/custom_objs.py
```diff
@@ -193,6 +233,25 @@ def _decode_custom_obj(
     load_instance_op_uri: str | None = None,
 ) -> Any:
     type_ = weave_type["type"]
+
+    # Decode policy lives on the process-global client (`get_weave_client`): workers
+    # flip it off there and this guard reads it back, even from worker decode threads.
+    # If client lookup becomes context-local, re-plumb this or it fails open there.
```
jtschoonhoven: Seems like another argument in favor of using an env var

weave/trace/serialization/custom_objs.py
```diff
@@ -65,6 +68,43 @@ class EncodedCustomObjDict(
...
+SAFE_CUSTOM_WEAVE_TYPES = frozenset(
+    {
+        "PIL.Image.Image",
+        "wave.Wave_read",
+        "weave.type_handlers.Audio.audio.Audio",
+        "datetime.datetime",
+        "rich.markdown.Markdown",
+        "moviepy.video.VideoClip.VideoClip",
+        "weave.type_wrappers.Content.content.Content",
+    }
```
jtschoonhoven, line 91: Claude recommends adding `File` and claims it's safe and "just reads metadata.json and a file out of the artifact — pure data, no __import__, no user code. It's exactly as safe as Image/Audio."

```suggestion
        "weave.type_wrappers.Content.content.Content",
        "weave.type_handlers.File.file.File",
    }
```

Also to `KNOWN_TYPES`

jtschoonhoven (approved):
A few comments. I still recommend reading `_allow_unsafe_custom_obj_decode` from an env var, but otherwise this looks like a straightforward improvement.

weave/trace/serialization/custom_objs.py
gtarpenning: confirmed

weave/trace/weave_client.py
gtarpenning: done

PR #7017, opened 2026-06-01
feat(weave): upload attach_media content on background threads

zbirenbaum:
Prevents `LLM.attach_media` from blocking user code.

## What changed

- `attach_media` now appends a placeholder and dispatches a daemon that runs `_publish_media_content` and writes the resolved `weave://` ref back onto the attachment. Threads run concurrently — the call returns immediately.
- `_await_uploads()` joins all in-flight upload threads refs are guaranteed populated before the span is emitted
- Failed uploads are logged (not raised, so a failed upload can't crash the caller) and their empty-ref attachments are dropped so the span never emits a broken URI part.

## Testing
- Added new tests and ensured existing pass

### DIFF
--- a/tests/session/test_session.py
+++ b/tests/session/test_session.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import threading
 from unittest.mock import patch
 
 import pytest
@@ -161,6 +162,7 @@ class TestLLM:
         ):
             c = LLM(model="gpt-4o")
             assert c.attach_media(content=b"png_bytes", mime_type="image/png") is c
+            c._await_uploads()
 
     def test_context_manager_sets_timestamps(self) -> None:
         with LLM(model="gpt-4o") as c:
@@ -187,6 +189,7 @@ class TestAttachMedia:
         result = llm.attach_media(content=b"png_bytes", mime_type="image/png")
         assert result is llm
         assert len(llm.media_attachments) == 1
+        llm._await_uploads()
         att = llm.media_attachments[0]
         assert att.ref.startswith("weave:///")
         assert att.modality == "image"
@@ -200,6 +203,7 @@ class TestAttachMedia:
             mime_type="image/jpeg",
         )
         assert len(llm.media_attachments) == 1
+        llm._await_uploads()
         att = llm.media_attachments[0]
         assert att.ref.startswith("weave:///")
         assert att.modality == "image"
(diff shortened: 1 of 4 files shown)


zbirenbaum:
* **#7017** <a href="https://app.graphite.com/github/pr/wandb/weave/7017?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7017?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6976** <a href="https://app.graphite.com/github/pr/wandb/weave/6976?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7018** <a href="https://app.graphite.com/github/pr/wandb/weave/7018?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

bcsherma (approved):

andrewtruong:
Do you need this PR?  By calling `publish` as you had previously, the work is already offloaded to a thread pool.  Is there actually a perf gain here?  If so can you bench and show it?

zbirenbaum:
> Do you need this PR? By calling `publish` as you had previously, the work is already offloaded to a thread pool. Is there actually a perf gain here? If so can you bench and show it?

I'm a little confused with this in the context of our discussion here https://github.com/wandb/weave/pull/6976/changes#r3314065716

This PR lets us:
 1. Not block user code while waiting for upload (raised in discussion above)
 2. Force spans to wait until the refs contained in them can actually be resolved until they are sent

If we just published inline we get the ref back when the upload completes (or a future that resolves to one)
If we compute the ref and fire off later we could get a broken ref (and breaks HTTP uploads)

I'll benchmark it but I think my mental model is a little off here

zbirenbaum:
> Do you need this PR? By calling `publish` as you had previously, the work is already offloaded to a thread pool. Is there actually a perf gain here? If so can you bench and show it?

| Metric | base (sync) | this branch (async) |
| --- | --- | --- |
| **Scenario A — no overlapping work** | | |
| `attach_media()` blocking | **415.8 ms** | **0.3 ms** |
| `_build_attrs()` drain | 0.0 ms | 105.0 ms |
| caller blocked on upload | **415.8 ms** | **105.2 ms** |
| **Scenario B — caller does 200 ms of work** | | |
| `attach_media()` blocking | 412.6 ms | 0.3 ms |
| caller work | 202.8 ms | 202.6 ms |
| `_build_attrs()` drain | 0.1 ms | 0.2 ms |
| caller blocked on upload | **412.7 ms** | **0.4 ms** |
| end-to-end | 615.5 ms | 203.1 ms |

rgao-coreweave (approved):
Very nice, thanks for the benchmark

LGTM

PR #7020, opened 2026-06-01
fix(weave): harden clickhouse migration lock for rolling deploys

gtarpenning:
## Summary
- Fixes [WB-35030](https://coreweave.atlassian.net/browse/WB-35030): ~32 weave-trace pods contend on the ClickHouse `migration_lock` on rolling deploys, stalling rollout 5-10 min.
- Lock-free pre-check skips the lock entirely when no schema migration is pending, so warm deploys never contend.
- Lease + heartbeat (120s lease / 30s heartbeat) replaces the fixed 600s TTL: a crashed holder frees in <=120s, a live slow migration is not stolen. Heartbeat connection defaults from `WF_CLICKHOUSE_*` env, so no caller change is needed.
- Release via lightweight `DELETE FROM` instead of an async `ALTER ... DELETE` mutation. `heartbeat_at` column added to existing lock tables via idempotent `ALTER ... ADD COLUMN IF NOT EXISTS`.

## Testing
unit tests; plus a 24-process race harness against real clickhouse: warm deploys take the lock 0 times, cold race applies migrations exactly once (no partial), live holder not stolen, dead holder recovers in ~lease.

[WB-35030]: https://coreweave.atlassian.net/browse/WB-35030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


neutralino1:
Is there a way to make sure the script only runs on a single pod? Either at the helm chart level, or at the script level?

amwarrier (approved):

amwarrier:
Curious why @neutralino1 would like to run this on a single pod.

gtarpenning:
> Is there a way to make sure the script only runs on a single pod? Either at the helm chart level, or at the script level?

this is pretty hard because i think we do k8s rollout by statefulset so many can come up at a time. we did briefly have a k8s JOB that managed this, but if the job fails then all the pods still roll and you get code running on out-of-date databases. 

the initcontainers should be idempotent, the only bad cases are races between the initcontainers.

neutralino1 (approved):

gtarpenning:
tested manually again in qa, seems like at least not a regression.

PR #7026, opened 2026-06-02
chore(weave_ts): Un-skip `live/` test suites

drtangible:
## Description

* Update: rather than removing these tests that haven't been running for ~16months, introducing some mocking that should allow us to un-`skip` them.

* The tests in `live/` act as integration tests, but are currently skipped. If we record their interaction with the network and replay it in tests, they should be less flaky.

* Similarly, a lot of our tests around our integrations with 3rd-party APIs would benefit from less manual mocking -- if we can record against an actual contract, we'll be less likely to introduce errors.

* Python side already uses https://vcrpy.readthedocs.io/en/latest/, this change introduces `vcr-test` for similar network recording / playback on the javascript side.

* Cassettes live in `/__tests__/cassettes/`, and are automatically recorded when not present.

* This library is a small wrapper on top of https://mswjs.io/, which has some other cool stuff we could potentially leverage, such as auto-mocking based on an OpenAPI spec.

* This change unskips the tests in `/live/`, and introduces a couple test helpers:
  * `withCassette()` for specifying that a fn should mock network requests w/a recorded cassette
  * `vcrTest` for automatically wrapping a test case `withCassette`.

### DIFF
--- a/sdks/node/package.json
+++ b/sdks/node/package.json
@@ -200,6 +200,7 @@
     "typedoc-plugin-markdown": "^4.2.9",
     "typescript": "^5.9.3",
     "typescript-eslint": "^8.61.0",
+    "vcr-test": "^1.4.0",
     "zod": "^4.4.3"
   },
   "packageManager": "pnpm@10.8.1+sha512.c50088ba998c67b8ca8c99df8a5e02fd2ae2e2b29aaf238feaa9e124248d3f48f9fb6db2424949ff901cffbb5e0f0cc1ad6aedb602cd29450751d11c35023677"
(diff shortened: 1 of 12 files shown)


drtangible:
* **#7026** <a href="https://app.graphite.com/github/pr/wandb/weave/7026?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7026?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

andrewtruong:
These seem like important behaviours.  I'm not sure why they were skipped, but shouldn't we fix them?

drtangible:
> These seem like important behaviours. I'm not sure why they were skipped, but shouldn't we fix them?

Happy to write some new tests around this behavior -- these ones have been been skipped for 15 months, so can start from scratch.

andrewtruong (approved):
nice 😎

PR #7027, opened 2026-06-02
chore(weave): retry flaky genai agent query tests

gtarpenning:
WB-35048

## Summary.
- Spans inserted via `ch_client.insert("spans", ...)` are occasionally not visible to the immediately-following read under CI ClickHouse. No production retry gap: production ingests via OTel (same insert call) then queries from the UI later, never reading its own write within ms. Same class as #6852.

TODO:
- build into the test suite a read after write handler to fix all of these.

Original failures ([3,10](https://github.com/wandb/weave/actions/runs/26788728287/job/78970293659), [3,11](https://github.com/wandb/weave/actions/runs/26788728287/job/78970293767)):
```
test_group_by_conversation_id - AssertionError: assert "conv-29be9732" in {}
test_agent_span_stats_ungrouped_metrics - assert 0.0 == 300
```

## Testing
test only change.

### DIFF
--- a/tests/trace_server/test_genai_agent_queries.py
+++ b/tests/trace_server/test_genai_agent_queries.py
@@ -185,6 +185,7 @@ def test_group_by_trace_id(ch_server):
 # ---------------------------------------------------------------------------
 
 
+# flaky: CI ClickHouse occasionally doesn't surface the just-inserted spans to the immediate read.
 @pytest.mark.flaky(reruns=3)
 def test_group_by_conversation_id(ch_server):
     """Grouping spans by conversation_id returns per-conversation aggregates."""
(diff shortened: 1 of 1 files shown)


tests/trace_server/test_genai_agent_queries.py
```diff
@@ -185,6 +185,7 @@ def test_group_by_trace_id(ch_server):
 # ---------------------------------------------------------------------------
 
 
+@pytest.mark.flaky(reruns=3)
```
andrewtruong, line 189: Please leave a comment explaining why it's flaky (and how to fix it in future if you have any thoughts)

gtarpenning: im working on a comprehensive fix, its just read after write consistency, which accounts for 95% of the flake in the repo.

andrewtruong (approved):

PR #7037, opened 2026-06-02
perf(weave): bounded conversation message previews for grouped spans query

bcsherma:
## Problem

The Agents conversations table shows a first-message and last-message preview per row. Each row was hydrated separately via `useConversationPagePreviews` (frontend), firing up to **two `agentConversationChat` requests per conversation** through a 6-worker pool — ~100 round trips per page, re-run on every sort/page change. This is a 1-2 order or magnitude speedup for loading the messages in the conversations table if the table is showing many rows with many and long messages.

## Why not just fold argMin/argMax into the grouped query

The obvious fix — adding `argMinIf(input_messages, …)` / `argMaxIf(output_messages, …)` to the grouped list query — has a nasty cost: ClickHouse aggregates **all** groups before applying `LIMIT`, and those aggregates need their argument columns for every qualifying row. So it would read the wide `input_messages` / `output_messages` array columns for **every span matching the list filters** (potentially the whole project / time window), then throw away all but the displayed page. `spans` is `ORDER BY (project_id, started_at, span_id)` with only a bloom-filter skip index on `conversation_id`, so that's an expensive scan on busy projects.

## Change — bounded second query

Compute previews server-side, but in a **second query scoped to the page's already-computed conversation_ids**:

- `make_conversation_previews_query` (`agent_query_builder.py`): `argMinIf`/`argMaxIf` over `spans`, `WHERE project_id = ? AND conversation_id IN (<page ids>)` plus the optional time range. The wide message columns are read **only for the conversations actually shown**; the bloom-filter index + time bounds prune the scan further.
- `AgentQueryHandler.spans_query` runs it **only when grouping by `conversation_id`** (other group-bys pay nothing), then attaches `first_message` / `last_message` to each row.
- Previews use the chat view's exact role resolution (`first_user_preview_text` / `last_assistant_preview_text`) so they match the opened conversation, truncated to `CONVERSATION_PREVIEW_CHARS = 280`.
- New `AgentConversationMessagePreview` + `first_message`/`last_message` on `AgentSpanGroupRow` (`types.py`).

One extra round trip to ClickHouse, vs ~N requests to the browser. The matching frontend change is a separate PR in `wandb/core` (#44907).

## Validation

- ClickHouse-backed integration test `test_group_by_conversation_id_message_previews` inserts a multi-turn conversation (opening user prompt, an empty tool span, final assistant turn) and runs the **real** two-query path — confirms `first_message → "hello first"`, `last_message → "final reply"`, empty tool span skipped by the `-If` guard.
- Handler unit test asserts 3 queries (count, grouped list, bounded preview), that the grouped list query does **not** reference message columns, and that the preview query is scoped with `conversation_id IN`.
- Query-builder tests for the new SQL (scoping + time range).
- Full suites green: 25 CH integration + 83 query/handler/helper/chat-view tests.

## Notes / follow-ups

- The page currently requests up to 1000 conversations; the preview query hydrates exactly that returned set. If a project's displayed set approaches its total conversation count, the bounded query's savings shrink (still never worse than folding in). A natural follow-up is to hydrate only the visible page.
- Previews derive from the earliest/latest *span* rather than the oldest/newest *turn*; in practice these coincide and the role resolution is identical.

### DIFF
--- a/weave/trace_server/agents/clickhouse.py
+++ b/weave/trace_server/agents/clickhouse.py
@@ -14,27 +14,36 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, NamedTuple, TypeAlias, TypeVar, cast
 
 from weave.shared import refs_internal as ri
-from weave.trace_server.agents.chat_view import build_trace_chat
+from weave.trace_server.agents.chat_view import (
+    build_trace_chat,
+    first_user_preview_text,
+    last_assistant_preview_text,
+)
 from weave.trace_server.agents.constants import (
+    CONVERSATION_PREVIEW_CHARS,
     MAX_INGEST_ERRORS_REPORTED,
     NO_CONVERSATION_LABEL,
 )
 from weave.trace_server.agents.helpers import (
     genai_span_to_row,
+    messages_from_ch_value,
     normalize_span_row,
     unpack_string_array,
 )
 from weave.trace_server.agents.schema import (
     ALL_SPAN_INSERT_COLUMNS,
     AgentSpanCHInsertable,
+    NormalizedMessage,
 )
 from weave.trace_server.agents.types import (
     AGENT_CUSTOM_ATTR_SOURCE_VALUE_TYPES,
     AgentConversationChatReq,
     AgentConversationChatRes,
+    AgentConversationMessagePreview,
     AgentCustomAttrSchemaItem,
     AgentCustomAttrsSchemaReq,
     AgentCustomAttrsSchemaRes,
+    AgentGroupByRef,
     AgentSchema,
     AgentSearchConversationResult,
     AgentSearchMatchedMessage,
@@ -73,6 +82,7 @@ from weave.trace_server.query_builder.agent_query_builder import (
     make_agents_list_query,
     make_conversation_chat_spans_query,
     make_conversation_chat_turns_count_query,
+    make_conversation_previews_query,
     make_custom_attrs_schema_query,
     make_message_search_query,
     make_span_group_categorical_distributions_query,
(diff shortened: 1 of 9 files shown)


bcsherma:
* **#7037** <a href="https://app.graphite.com/github/pr/wandb/weave/7037?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7037?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/agents/clickhouse.py
```diff
@@ -169,8 +179,54 @@ def spans_query(self, req: AgentSpansQueryReq) -> AgentSpansQueryRes:
...
+        this page so the wide message columns are read only for the displayed
+        conversations — not for every span matching the list filters. Failures
+        are swallowed: previews are best-effort display data and must not fail
+        the list query.
+        """
+        conversation_ids = [
+            cid
+            for g in groups
+            if isinstance((cid := g.group_keys.get(group_alias)), str) and cid
+        ]
```
jtschoonhoven: Nit: the conditional assignment inside an `isinstance` inside a list comprehension is a little intense, can you make this a `for` loop?

weave/trace_server/agents/clickhouse.py
```diff
@@ -169,8 +179,54 @@ def spans_query(self, req: AgentSpansQueryReq) -> AgentSpansQueryRes:
...
+    ) -> None:
+        """Attach first/last message previews to conversation rows.
+
+        Runs a second, bounded query scoped to exactly the conversation_ids on
+        this page so the wide message columns are read only for the displayed
+        conversations — not for every span matching the list filters. Failures
+        are swallowed: previews are best-effort display data and must not fail
+        the list query.
+        """
+        conversation_ids = [
```
jtschoonhoven: Q: should this be a `set`?

weave/trace_server/agents/clickhouse.py
```diff
@@ -169,8 +179,54 @@ def spans_query(self, req: AgentSpansQueryReq) -> AgentSpansQueryRes:
...
+            req.project_id,
+            conversation_ids,
+            started_after=req.started_after,
+            started_before=req.started_before,
+        )
+        try:
+            rows = _rows_as_dicts(self._query(sql, pb.get_params()))
+        except Exception:
+            logger.exception("failed to hydrate conversation message previews")
+            return
```
jtschoonhoven, line 219: Nit: should this catch a more specific exception?

weave/trace_server/agents/clickhouse.py
```diff
@@ -169,8 +179,54 @@ def spans_query(self, req: AgentSpansQueryReq) -> AgentSpansQueryRes:
         groups = [_hydrate_group_row(r, aliases, measure_aliases) for r in rows]
         if req.group_distributions:
             self._hydrate_group_distributions(req, groups, aliases[0])
+        if _is_conversation_grouping(req.group_by):
+            self._hydrate_conversation_previews(req, groups, aliases[0])
```
jtschoonhoven, line 183: Below you write

>previews are best-effort display data and must not fail the list query.

...do you want to wrap this in `try/except`?

weave/trace_server/agents/clickhouse.py
```diff
@@ -782,10 +838,63 @@ def _hydrate_group_row(
         conversation_names=unpack_string_array(row.get("conversation_names")),
         first_seen=_datetime_or_none(row.get("first_seen")),
         last_seen=_datetime_or_none(row.get("last_seen")),
+        # first_message / last_message are hydrated separately for conversation
+        # groupings via _hydrate_conversation_previews; left None here.
```
jtschoonhoven, line 842: I think this ship has probably sailed, but really we should avoid polymorphic API response types. If the same endpoint can return multiple types of responses, then it should be split into multiple endpoints.

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -1017,6 +1017,50 @@ def make_spans_list_query(pb: ParamBuilder, req: AgentSpansQueryReq) -> str:
...
+
+    Deliberately scoped to ``conversation_id IN (...)`` — the page's already
+    computed conversation_ids — so the wide ``input_messages`` / ``output_messages``
+    columns are only read for the conversations actually shown, not for every
+    span matching the list filters. (A grouped ``argMin``/``argMax`` folded into
+    the main list query would read those columns for the entire filter match
+    before LIMIT.) The bloom-filter skip index on ``conversation_id`` plus the
+    optional time range bound the scan further.
+
+    ``argMinIf``/``argMaxIf`` pick the earliest input and latest output span that
```
jtschoonhoven, line 1038: \``🤮 ``

jtschoonhoven (approved):
Nits only

weave/trace_server/agents/clickhouse.py
bcsherma: I think maybe in the future but for now I just want to ensure that we don't fail the query at this stage, we can observe errors and see if there are more specific things to except in future

weave/trace_server/agents/clickhouse.py
bcsherma: 🫠

PR #7039, opened 2026-06-02
fix(weave): treat an empty parent_id as null in call-ingest validation

ro31337:
## JIRA Issue(s)
https://coreweave.atlassian.net/browse/WB-35079

## Description
On the ingest path (`call_start` / `call_batch`), a root call's `parent_id` sometimes arrives as an empty string `""` instead of `null`. The UUID validator rejects `""`, and because validation runs inside the batch loop before the flush, that one bad item fails the whole batch with a 400, dropping the valid calls batched with it too.

`parent_id` is already optional (`None` means a root call), so this normalizes `""` to `None`. A non-empty but malformed `parent_id` is still rejected, and `""` was never really stored anyway, because the storage layer already uses it as the "no parent" sentinel and reads it back as `None`. This fires continuously in prodd, roughly 430 ingest batches were dropped this way in the last 7 days. See it in [Datadog (us5, prod)](https://us5.datadoghq.com/apm/traces?query=env%3Aprod%20service%3Aweave-trace%20%40error.type%3Aweave.trace_server.validation_util.CHValidationError%20%28resource_name%3Acall_start%20OR%20resource_name%3Acall_batch%29&agg_m=count&agg_m_source=base&agg_t=count&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code%2C%40_span.count%2C%40_duration.by_service&fromUser=false&historicalData=true&messageDisplay=inline&query_translation_version=v0&sort=desc&spanType=all&storage=hot&traceQuery=&view=spans&start=1779826008025&end=1780430808025&paused=false).

## Testing
Added a validator unit test and a batch test mirroring the existing invalid-`trace_id` one. Both fail without the fix (with the exact production error) and pass with it.


ro31337:
The bug is quiet, see Slack thread (link in Jira).

Checked 2026-06-22: the empty-`parent_id` signature (`Invalid UUID: . Unable to parse` on `call_start` / `call_batch`) shows 0 occurrences across the full searchable APM window in DataDog (~last 2 weeks). That matches the drop-off around June 2 noted in the thread — it's stayed quiet since, and nothing in the code looks like it actually fixed it, so whatever client was sending `parent_id=""` just stopped on its own.

Closing it for now with a "no fix" resolution. The change itself is correct and harmless, so we can revive it if the signature ever comes back.

PR #7041, opened 2026-06-02
fix(weave): mirror clickhouse CI image via ghcr 

gtarpenning:
## Summary
- The `weave_clickhouse` service container pulled from Docker Hub, which intermittently fails CI at image-pull time (`net/http: request canceled ... Client.Timeout`) due to the unauthenticated per-IP rate limit.
- Convert it to an in-step `docker run` that pulls from `ghcr.io/wandb/clickhouse-server`, mirroring from Docker Hub on first use (`.github/scripts/ch-image.sh`).
- Now CH version is single-sourced as the script's default tag (was pinned in 3 spots, incl. nightly's stale `25.10`).

## Performance
Per trace-tests job. The `Initialize containers` phase is gone (no service containers), so even the one-time seeding run beats the old Hub pull.

| stage | master (Hub service) | run #1 (seed) | run #2 (ghcr-hit) |
|---|---|---|---|
| Initialize containers | 21s | 0s | 0s |
| Start ClickHouse step | n/a | ~16s | ~12-13s |

## Testing
first run seeds ghcr (one hub pull); subsequent runs pull from ghcr, no hub access. 161/161 jobs green. make the ghcr package public so fork/dependabot PRs skip hub too.

### DIFF
--- a/.github/scripts/ch-image.sh
+++ b/.github/scripts/ch-image.sh
@@ -10,10 +10,11 @@ ghcr="ghcr.io/wandb/clickhouse-server:${tag}"
 hub="clickhouse/clickhouse-server:${tag}"
 
 if [ -n "${GITHUB_TOKEN:-}" ]; then
+  # best-effort: fork/dependabot PRs lack packages:write and fall back to the Hub path below.
   echo "${GITHUB_TOKEN}" | docker login ghcr.io -u "${GITHUB_ACTOR:-wandb}" --password-stdin 1>&2 || true
 fi
 
-if docker pull "$ghcr" 1>&2 2>/dev/null; then
+if docker pull "$ghcr" 1>&2; then
   echo "$ghcr"
   exit 0
 fi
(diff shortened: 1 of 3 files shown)


.github/scripts/ch-image.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo "${GITHUB_TOKEN}" | docker login ghcr.io -u "${GITHUB_ACTOR:-wandb}" --password-stdin 1>&2 || true
+fi
+
+if docker pull "$ghcr" 1>&2 2>/dev/null; then
+  echo "$ghcr"
+  exit 0
+fi
+
+echo "Mirroring ${hub} -> ${ghcr}" 1>&2
+docker pull "$hub" 1>&2
```
ro31337: nit: if this command fails, the script will continue execution... can be a bit unpredictable. Maybe

```
if ! docker pull "$hub" 1>&2; then
  echo "Failed to pull ${hub}" 1>&2
  exit 1
fi
```

?

ro31337 (approved):
Looks good, left one comment up to you

.github/scripts/ch-image.sh
gtarpenning: nice, done

PR #7042, opened 2026-06-02
feat(weave): migration to add agent columns to feedback table

jtschoonhoven:
First of several stacked PRs adding a pre-aggregated feedback API for the Agent Signals charts.

## Changes

Migration to add 3 new columns to the `feedback` table that are duplicated from the agent `spans` table: `span_agent_name`, `span_agent_version`, `span_status_code`. 

Denormalizing the data this ways allows for efficient queries on agent data. Includes only the data that we frequently want to filter/group on for scores. This will used only for visualizations and `spans` remains the source of truth.

## Stack

1. **#7042 (this PR)** — migration
2. #7043 — ORM plumbing + aggregate query builder + models
3. #7044 — feedback_aggregate endpoint impl + bindings

## Testing

- [x] Deploy to QA

### DIFF
--- a/weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
+++ b/weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
@@ -1,0 +1,23 @@
+/*
+Duplicate select genai metadata from the `spans` table into `feedback`.
+Denormalizing the data this ways allows for efficient queries on agent data.
+Includes only the data that we frequently want to filter/group on for scores.
+This is used only for visualizations and `spans` remains the source of truth.
+
+- `span_agent_name`: display name of the scored agent (from `spans.agent_name`).
+- `span_agent_version`: agent version string (from `spans.agent_version`).
+- `span_status_code`: status of the scored turn (from `spans.status_code`).
+*/
+ALTER TABLE feedback
+    ADD COLUMN IF NOT EXISTS span_agent_name    String DEFAULT '',
+    ADD COLUMN IF NOT EXISTS span_agent_version String DEFAULT '',
+    ADD COLUMN IF NOT EXISTS span_status_code   Enum8('UNSET'=0, 'OK'=1, 'ERROR'=2) DEFAULT 'UNSET';
+
+/*
+Add data-skipping indexes on the columns we want to filter on.
+- `idx_span_agent_name`: skip granules with agents not in the query.
+- `idx_span_agent_version`: skip granules with versions not in the query.
+*/
+ALTER TABLE feedback
+    ADD INDEX IF NOT EXISTS idx_span_agent_name   span_agent_name    TYPE set(64)  GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_span_agent_version span_agent_version TYPE set(128) GRANULARITY 1;
--- a/weave/trace_server/migrations/033_add_agent_columns_to_feedback.down.sql
+++ b/weave/trace_server/migrations/033_add_agent_columns_to_feedback.down.sql
@@ -1,0 +1,8 @@
+ALTER TABLE feedback
+    DROP INDEX IF EXISTS idx_span_agent_name,
+    DROP INDEX IF EXISTS idx_span_agent_version;
+
+ALTER TABLE feedback
+    DROP COLUMN IF EXISTS span_agent_name,
+    DROP COLUMN IF EXISTS span_agent_version,
+    DROP COLUMN IF EXISTS span_status_code;


weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
```diff
@@ -0,0 +1,14 @@
...
+This is used only for visualizations and `spans` remains the source of truth.
+
+- `span_agent_name`: display name of the scored agent (from `spans.agent_name`).
+- `span_agent_version`: agent version string (from `spans.agent_version`).
+- `span_status_code`: status of the scored turn (from `spans.status_code`).
+*/
+ALTER TABLE feedback
+    ADD COLUMN IF NOT EXISTS span_agent_name    String DEFAULT '',
+    ADD COLUMN IF NOT EXISTS span_agent_version String DEFAULT '',
+    ADD COLUMN IF NOT EXISTS span_status_code   String DEFAULT '';
```
gtarpenning: why not use the actual source type? `Enum8('UNSET'/'OK'/'ERROR')`

weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
```diff
@@ -0,0 +1,14 @@
+/*
+Duplicate select genai metadata from the `spans` table into `feedback`.
```
gtarpenning, line 2: did we consider adding skip indices on these fields here? they are not in the primary key and if they are going to be explicitly used for filtering...

bcsherma (approved):
the migration looks good and low risk to me. I do have a feeling we could be a ways from aggregations on feedback values being a big perf win, just given the volume of feedback we are likely to see in the short-medium term. But tracking the span agent metadata seems like a good thing to do regardless of the materialized views

weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
jtschoonhoven: Good catch: updated

weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
jtschoonhoven: Updated!

weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
```diff
@@ -0,0 +1,25 @@
...
+/*
+Add data-skipping indexes on the columns we want to filter on.
+- `idx_created_at`: skip granules outside the time window of the query.
+- `idx_span_agent_name`: skip granules with agents not in the query.
+- `idx_span_agent_version`: skip granules with versions not in the query.
+*/
+ALTER TABLE feedback
+    ADD INDEX IF NOT EXISTS idx_created_at        created_at         TYPE minmax     GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_span_agent_name   span_agent_name    TYPE set(512)  GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_span_agent_version span_agent_version TYPE set(1024) GRANULARITY 1;
```
gtarpenning: this suggest we think there will be 1000 agent versions in an 8000 row block, seems like an insanely high number no?

gtarpenning: at 1000 rows its not even really that much better to hit the index instead of the table, if we think that there will actually be that many. in that case a bloom filter would be way better

jtschoonhoven: Good point. I can probably drop these by ~10x. `set(64)` and `set(128)`

weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
```diff
@@ -0,0 +1,25 @@
...
+    ADD COLUMN IF NOT EXISTS span_status_code   Enum8('UNSET'=0, 'OK'=1, 'ERROR'=2) DEFAULT 'UNSET';
+
+/*
+Add data-skipping indexes on the columns we want to filter on.
+- `idx_created_at`: skip granules outside the time window of the query.
+- `idx_span_agent_name`: skip granules with agents not in the query.
+- `idx_span_agent_version`: skip granules with versions not in the query.
+*/
+ALTER TABLE feedback
+    ADD INDEX IF NOT EXISTS idx_created_at        created_at         TYPE minmax     GRANULARITY 1,
```
gtarpenning: hmm interesting, i don't think this is really that useful? id be surprised if we were able to prune much using this because of the current primary key co-locating data based first on weave_ref (with otel ids). i would suspect this index to essentiall just include a very small min and a very high max for every granule, maybe we can test this with a dev instance and some normal usage that fills up 100 granules?

weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
jtschoonhoven: I'd consider going even lower if you think that's better

weave/trace_server/migrations/033_add_agent_columns_to_feedback.up.sql
jtschoonhoven: ~The data is already sorted by uuid7 timestamps, but I don't think that will help the query planner when filtering on `created_at`. I can do some EXPLAINing to confirm.~

Crap, otel IDs aren't uuid7. Nevermind I'll remove this index.

gtarpenning: what does a weave_ref to a call with an OTel uuidv4 look like?

jtschoonhoven: `weave-trace-internal:///UHJvamVjdEludGVybmFsSWQ6NDM0/agent_turn/38da85678e584070bfeb02d5c4bfcf2b`

jtschoonhoven: `weave-trace-internal:///<projectId>/<spanType>/<randomHex>`

gtarpenning (approved):

PR #7045, opened 2026-06-03
chore(weave): Allow creation of monitors using op names

mscavezze-cw:
## Description

- Fixes [WB-33908](https://coreweave.atlassian.net/browse/WB-33908)

Allow creation of monitors using op names. Now the sample code in the docs works.

## Testing

Unit tests, manual testing using both scripts and UI

[WB-33908]: https://coreweave.atlassian.net/browse/WB-33908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weave/flow/monitor.py
```diff
@@ -107,8 +152,47 @@ def deactivate(self) -> ObjectRef:
...
+                continue
+            short = name.strip().strip("/")
+            if not short:
+                raise ValueError(f"op_names entries must be non-empty; got {name!r}")
+            if "/" in short:
+                raise ValueError(
+                    "op_names entries must be a weave URI starting with 'weave://', "
+                    f"or a single op short name (no '/'); got {name!r}"
+                )
+            normalized.append(f"weave:///{entity}/{project}/op/{short}:*")
```
gtarpenning: there must be helpers for these kind of constructions

weave/flow/monitor.py
```diff
@@ -107,8 +152,47 @@ def deactivate(self) -> ObjectRef:
...
+        To satisfy both use-cases, we perform a conversion here using the
+        active client's entity/project.
+
+        Entries that are already ``weave://`` refs, or agent-span literals
+        like ``"weave.genai.turn_ended"``, are returned unchanged.
+        """
+
+        def needs_expansion(name: str) -> bool:
+            return name not in AGENT_SPAN_OP_NAMES and not name.strip().startswith(
+                "weave://"
```
gtarpenning: we use `weave:///` no?

tests/flow/test_monitor.py
```diff
@@ -85,3 +89,49 @@ def test_activate(weave_active):
 
     deactivated_ref = monitor.deactivate()
     assert deactivated_ref.get().active == False
+
+
+def test_normalizes_short_op_names_at_construction(weave_active):
```
gtarpenning: seems redundant with above?

gtarpenning (approved):
cool, this (finally) makes sense. not 100% sure we need the `model_post_init`, most people will probably know to `import weave`, but up to you. 

the earlier comments still apply, and one more meta comment:  technically here we are assuming that there can be no cross-project objects. That's not actually the case. I have multiple long running scripts that init a client in one project and pull objects from others (mostly datasets). Not sure if that's at all relevant for ops/monitors, but might be worth printing a warning that we made a best guess attempt at figuring out what project the op belonged to.

mscavezze-cw:
> cool, this (finally) makes sense. not 100% sure we need the `model_post_init`, most people will probably know to `import weave`, but up to you.

I introduced it to handle the case where you create a monitor but don't activate it. Without it, the monitor could show up in the UI with `Incorrect op ref:`. Maybe that's not a common case and we're being too defensive. That said, I don't see much harm in having it here.

 
> the earlier comments still apply, and one more meta comment: technically here we are assuming that there can be no cross-project objects. That's not actually the case. I have multiple long running scripts that init a client in one project and pull objects from others (mostly datasets). Not sure if that's at all relevant for ops/monitors, but might be worth printing a warning that we made a best guess attempt at figuring out what project the op belonged to.

That's an interesting use-case. It's not obvious how I could detect this case and warn about it without either 1) doing a read to see if the ref is valid or 2) issuing the warning every time, which adds noise and could confuse regular users.

I'll add a comment to the code. If someone runs into this scenario, they'll be able to work around it by supplying full refs instead of short names.

tests/flow/test_monitor.py
mscavezze-cw: Removed

weave/flow/monitor.py
mscavezze-cw: Updated

weave/flow/monitor.py
mscavezze-cw: Updated - thanks for catching that.

PR #7047, opened 2026-06-03
feat(weave): push traces from playgrounds to spans table

seanmor5:
This introduces the backend changes necessary to push traces from playgrounds to our new agent spans table. We just emit a full `AgentSpanCHInsertable` with information sent to us from the playground chat completion request. The frontend sends us all of the information we need.

There are some considerations here with the `chat_view` rendering logic that I think need to be followed up on in my next couple of PRs:

- Tools / functions render, but there may be some edge cases to look out for when rendering through the playground. They also obviously don't emit as tool call spans right now, but no reason they couldn't
- I noticed the chat view renders these conversations strangely. Each span carries `input_messages` which has the system prompt, prior assistant messages, etc. but the conversations view only likes to render the last user input message and then the output assistant message.
- In the case of "retries" we get repeated rendering of the same input message and new response. I think it's fine to treat this as the same conversation, but maybe we should only render the input message once.

Spans are sent with `agent_name="playground"`, so they are rendered as a separate agent. I think this is probably fine. One consideration though is that this is used to reconstruct playground history, so things get weird if a customer has their agent named `playground` as well


jwlee64 (approved):
This looks good
Theres a couple more tests that need to be updated within
tests/trace_server/test_llm_completion.py
tests/trace_server/test_clickhouse_trace_server_batched.py

weave/trace_server/agents/playground.py
```diff
@@ -0,0 +1,246 @@
...
+    AgentSpanCHInsertable,
+    NormalizedMessage,
+)
+from weave.trace_server.ch_sentinel_values import EXPIRE_AT_NEVER, SENTINEL_EPOCH
+from weave.trace_server.trace_server_interface import CompletionsCreateRequestInputs
+from weave.trace_server.ttl_settings import compute_expire_at
+
+logger = logging.getLogger(__name__)
+
+PLAYGROUND_AGENT_NAME = "playground"
```
neutralino1: Maybe we should make it more specific, like at least "Weave Chat Playground"?

PR #7048, opened 2026-06-03
chore(weave): Flip OTEL on by default

andrewtruong:
Enable v2 mode by default

### DIFF
--- a/weave/trace/settings.py
+++ b/weave/trace/settings.py
@@ -309,20 +309,14 @@ class UserSettings:
     """
     Routes OTel-capable integrations through their OTel variant.
 
-    When True (the default), integrations that ship a sibling OTel patcher
-    (e.g. ``openai_agents`` → ``openai_agents_otel``, ``claude_agent_sdk`` → its
-    OTel variant) dispatch to the OTel variant on implicit import-hook patching.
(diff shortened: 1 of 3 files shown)


zbirenbaum (approved):
LGTM

weave/trace/settings.py
```diff
@@ -305,7 +305,7 @@ class UserSettings:
     Can be overridden with the environment variable `WEAVE_DISABLE_WAL_SENDER`
     """
 
-    use_otel_v2: bool = False
+    use_otel_v2: bool = True
```
ro31337, line 308: As I understand it would switch off patching for openai.* traces? (for those who didn't do `WEAVE_USE_OTEL_V2=false`). Just wanted to double check that part

Image

ro31337: Asking because says "chore" and confused me for a bit

ro31337 (approved):

drtangible (approved):

weave/trace/settings.py
andrewtruong: Yep.  I guess it is technically a behaviour change but hopefully the description is sufficient.  I think most users will be on the v2 case anyways

PR #7044, opened 2026-06-02
feat(weave): implement feedback_aggregate endpoint (ClickHouse) + bindings

jtschoonhoven:
Stacked on #7043 (review/merge that first).

## Purpose

Third PR of the pre-aggregated feedback API for Agent Signals charts. Wires the query builder from #7043 into the server.

## Changes

- ClickHouse impl (`methods/feedback_aggregate.py`): runs the built query and maps each `(bucket, group)` row to a `FeedbackAggregateBucket`. `sumMap` comes back as a `Tuple(Array, Array)`, coerced to a dict.
- SQLite raises `NotImplementedError` (ClickHouse-only: `sumMap` / `toStartOfInterval`).
- Remote HTTP client binding (`POST /feedback/aggregate`) and the external→internal adapter forwarding method (project-id translation + ref resolution).

The FastAPI route and the `tracing_trace_server` wrapper live in the **core repo** and land separately (they depend on the submodule SHA bump).

## Testing

- `tests/trace/test_feedback.py::test_feedback_aggregate` — ClickHouse integration test: writes agent-monitor rows across two scorers with varied tags/ratings, calls `feedback_aggregate` grouped by `runnable_ref`, folds the sparse buckets, and asserts per-scorer totals / tag counts / rating sum+count. Asserts SQLite raises `NotImplementedError`.
- Full feedback + query-builder suites: 35 passed on ClickHouse. `ruff` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7052, opened 2026-06-03
feat(weave): `feedback_agg_query_builder.py`

jtschoonhoven:
Stacked on #7043.

## Summary

Adds `feedback_agg_query_builder.py` which will be the basis for a new endpoint to return aggregate scores from the feedback table (added in #7044). This isn't wired up to anything yet.

## Testing

- [x] Unit tests
- [x] Tested locally

### DIFF
--- a/tests/trace/test_feedback.py
+++ b/tests/trace/test_feedback.py
@@ -5,13 +5,21 @@ from clickhouse_connect.driver.exceptions import DatabaseError
 
 import weave
 from tests.trace.util import client_is_sqlite
+from tests.trace_server.conftest_lib.trace_server_external_adapter import (
+    DummyIdConverter,
+)
 from weave import AnnotationSpec
 from weave.trace.weave_client import WeaveClient, get_ref
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.common_interface import SortBy
 from weave.trace_server.errors import InvalidRequest
+from weave.trace_server.feedback_agg_query_builder import (
+    build_feedback_aggregate_query,
+)
 from weave.trace_server.interface.query import Query
+from weave.trace_server.orm import ParamBuilder
 from weave.trace_server.trace_server_interface import (
+    FeedbackAggregateReq,
     FeedbackCreateReq,
     FeedbackQueryReq,
     FeedbackReplaceReq,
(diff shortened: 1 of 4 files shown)


weave/trace_server/feedback_agg_query_builder.py
```diff
@@ -0,0 +1,169 @@
...
+    if req.scorer_ids:
+        # Scorers are always stored in the `runnable_ref` column
+        clauses.append(
+            _any_prefix_clause(_ref_object_id_sql("runnable_ref"), req.scorer_ids, pb)
+        )
+    if req.span_agent_names:
+        names_param = pb.add_param(req.span_agent_names)
+        clauses.append(f"span_agent_name IN {param_slot(names_param, 'Array(String)')}")
+    if req.span_types:
+        # weave_ref looks like `<weave-scheme>:///<project>/<span_type>/<id>`
```
gtarpenning: ultranit: `<project_id>`

tests/trace_server/query_builder/test_feedback_aggregate.py
```diff
@@ -0,0 +1,245 @@
...
+            tags=["nsfw"],
+            rating_min=0.5,
+            rating_max=0.9,
+        ),
+        project,
+        after,
+        before,
+        pb,
+    )
+    assert "project_id = {pb_0:String}" in where
```
gtarpenning: can we assert against the actual full sql? catches small query generation bugs in formatting

gtarpenning: this goes for most of the tests in this file actually

gtarpenning (commented):
partial

weave/trace_server/trace_server_interface.py
```diff
@@ -1460,6 +1460,154 @@ class FeedbackStatsRes(BaseModel):
...
 
+# --- Feedback aggregate schema (for scores grouped by time bucket) ---
+
+FeedbackAggregateGroupByColumn: TypeAlias = Literal[
+    "runnable_ref", "span_agent_name", "span_agent_version", "span_status_code"
+]
+
+# Valid GROUP BY columns for aggregate feedback requests.
+FEEDBACK_AGGREGATE_GROUP_BY_COLUMNS: frozenset[FeedbackAggregateGroupByColumn] = (
+    frozenset(
```
nikumar1206: instead of defining twice u can use get_args for this:

```
FEEDBACK_AGGREGATE_GROUP_BY_COLUMNS = frozenset(
    get_args(FeedbackAggregateGroupByColumn)
)
```

nikumar1206: https://docs.python.org/3/library/typing.html#typing.get_args

tests/trace_server/query_builder/test_feedback_aggregate.py
```diff
@@ -0,0 +1,325 @@
...
+    FEEDBACK_AGGREGATE_GROUP_BY_COLUMNS,
+    FeedbackAggregateReq,
+)
+
+# 2024-01-01 and 2024-01-02 UTC, in unix epoch ms (a 1-day window).
+_AFTER_MS = 1704067200000
+_BEFORE_MS = 1704153600000
+_BUCKET_SECONDS = 21600
+
+_DAY_MS = 86_400_000
```
nikumar1206: super nit, but i think this is defined twice.

is this not the same as  `DAY_IN_MS`

nikumar1206: we can prolly just import the const from the tsi file

gtarpenning (commented):
hey can we also do functional tests over these? claude found at least 2 different functional issues: 
1. filtering by `monitor_ids=["mon"]` compiles to `startsWith(splitByChar('/', ifNull(trigger_ref,''))[-1], 'mon')` which will match `monday` or `monkeyman1000` etc. even without the `*` . also on this path like a bare `*` will literally match every single monitor, which feels like a perf issue? 
2. `position(weave_ref, '/agent_turn/') > 0` matches anywhere in the ref url, weirdly named stuff might match no? why cant we just do it like: `splitByChar('/', weave_ref)[6] = {param} ` or something?

weave/trace_server/feedback_agg_query_builder.py
```diff
@@ -0,0 +1,169 @@
...
+_RATING_KEY = "_rating_"
+
+# SQL to add up tags and ratings by name. Each returns Map(<key-name>, <count-or-sum>).
+# Sum and count are separate for ratings so the client can caculalate averages across buckets.
+_TAG_COUNTS_SQL = "sumMap(scorer_tags, arrayMap(x -> toUInt64(1), scorer_tags))"
+_RATING_SUMS_SQL = "sumMap(mapKeys(scorer_ratings), mapValues(scorer_ratings))"
+_RATING_COUNTS_SQL = "sumMap(mapKeys(scorer_ratings), arrayMap(x -> toUInt64(1), mapValues(scorer_ratings)))"
+
+# Count a row if it produced at least one tag or rating. Used to show the total volume of scores.
+_SCORED_COUNT_SQL = "countIf(notEmpty(scorer_tags) OR length(scorer_ratings) > 0)"
```
gtarpenning: aren't these the same:`notEmpty(scorer_tags) OR length(scorer_ratings) > 0)`

gtarpenning: like why not: `notEmpty` for both?

weave/trace_server/feedback_agg_query_builder.py
```diff
@@ -0,0 +1,169 @@
...
+    sql: str
+    columns: list[str]
+    parameters: dict[str, QueryParamValue] = field(default_factory=dict)
+
+
+# Agent-monitor scores land under this scorer_ratings key; the rating filter targets it.
+_RATING_KEY = "_rating_"
+
+# SQL to add up tags and ratings by name. Each returns Map(<key-name>, <count-or-sum>).
+# Sum and count are separate for ratings so the client can caculalate averages across buckets.
```
gtarpenning: nit typo

weave/trace_server/trace_server_interface.py
```diff
@@ -1460,6 +1460,154 @@ class FeedbackStatsRes(BaseModel):
...
+        {"runnable_ref", "span_agent_name", "span_agent_version", "span_status_code"}
+    )
+)
+
+# Span types, matched on the feedback's weave_ref path segment
+SpanType: TypeAlias = Literal["agent_turn", "agent_conversation"]
```
gtarpenning: nit: `FeedbackSpanType`

tests/trace/test_feedback.py
```diff
@@ -993,6 +1001,79 @@ async def test_filter_by_feedback(client: WeaveClient, no_autoflush) -> None:
...
+def test_feedback_aggregate_filter_matching_functional(client: WeaveClient) -> None:
+    """Functional checks (ClickHouse-only) that the WHERE filters match precisely.
+
+    The aggregate endpoint lands separately, so this executes the built query
+    directly. Guards against over-broad matching that string assertions miss:
+    object-id filters match the id exactly (a trailing '*' opts into prefix), and
+    span_types matches the ref's span-type segment, not an arbitrary substring.
+    """
+    if client_is_sqlite(client):
+        pytest.skip("feedback_aggregate query is ClickHouse-only (sumMap/splitByChar)")
```
gtarpenning, line 1013: nit for followup: why is the ch only? its a claude 1-shot for sqlite too

gtarpenning (approved):

tests/trace/test_feedback.py
jtschoonhoven: I'm inclined to skip support for sqlite. I'd either need to special-case behavior just for sqlite, or I'd need an entirely separate query builder. Seems like a lot of extra code/complexity for not a lot of added value. I have the opinion that we should avoid adding special cases just for tests into production code.

It would be really nice to have an in-memory clickhouse stub so we didn't have to maintain a separate interface just for sqlite...

PR #7059, opened 2026-06-04
fix(weave): defer jsonschema import to speed up `import weave`

andrewtruong:
## Summary

This PR defers the `jsonschema` import inside trace server which speeds up client import time by ~450ms (~50%).

That import is not intended to be used by the client at all, and is an unfortunate side effect of trace server coupling.

## Benchmarks
| Metric | Before | After | Improvement |
|---|--:|--:|--:|
| `import weave` — mean | **1.064s** | **0.515s** | **−52% (~2.1×)** |
| `import weave` — median | 1.027s | 0.510s | −50% |
| `import weave` — min | 0.989s | 0.488s | −51% |
| `-X importtime` total self-time | 1.061s | 0.543s | −0.52s |
| `weave.init()` (offline/disabled code path) | ~0.000s | ~0.000s | n/a (network-bound) |

### DIFF
--- a/scripts/benchmarks/.gitignore
+++ b/scripts/benchmarks/.gitignore
@@ -1,1 +1,2 @@
 *.csv
+.artifacts/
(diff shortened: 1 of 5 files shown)


rgao-coreweave (approved):
LGTM

thanks for fixing!

drtangible (commented):
🥇

PR #7061, opened 2026-06-04
fix(weave): lazy-init httpx client so WEAVE_INSECURE_DISABLE_SSL works post-import

rgao-coreweave:
## Summary

Fixes WB-33539. Setting `WEAVE_INSECURE_DISABLE_SSL=true` *after* `import weave` but before `weave.init()` had no effect, producing a misleading `RuntimeError: Weave is not available on the server` even though the server was fine.

Root cause: `weave/utils/http_requests.py` built a module-level `httpx.Client(verify=ssl_verify())` at import time, freezing the SSL setting before users got a chance to set the env var.

## Changes

- **`weave/utils/http_requests.py`** — replaced the module-level `client = httpx.Client(...)` with `_get_client()`, which reads `ssl_verify()` / `http_timeout()` per call and looks up the matching client from `_build_client(verify, timeout)` (cached with `@functools.cache`). Flipping `WEAVE_INSECURE_DISABLE_SSL` mid-process picks up the right client on the next request; unchanged settings reuse the same connection pool.
- **`weave/trace/weave_init.py`** — `_get_server_info` now logs with `exc_info=True`, so the real underlying exception (SSL failure, network error, etc.) is visible instead of just the generic "not available" warning.
- **tests** — added a regression test for the env-flip behavior; updated existing tests to call `_get_client()` directly instead of relying on a module-level `client` attribute.

## Repro (before this PR)

```python
import os
import weave  # httpx client created here with verify=True
os.environ["WEAVE_INSECURE_DISABLE_SSL"] = "true"  # too late
weave.init("entity/project")  # → "Weave is not available on the server"
```

After this PR the env var is honored because the client is built on first request and re-resolved per request via the settings-keyed cache.

## Benchmark

Ran `scripts/benchmarks/import_init_profile.py` on master vs. this branch (15 iterations, cold subprocess, 1 warmup discarded):

| Phase | master | branch | Δ |
|---|---|---|---|
| `import weave` (mean) | 0.569s | 0.547s | **−22 ms (−4%)** |
| `import weave` (median) | 0.563s | 0.533s | −30 ms |
| `init()` [disabled, offline] | <1 ms | <1 ms | noise |

Per-module breakdown explains the win: on master `weave.utils.http_requests` was the #2 most expensive module at **25 ms self-time** (4.4% of total) — that's the cost of constructing the `httpx.Client` and its SSL context at import. On this branch the module is no longer in the top-30 because client construction is deferred to first request. The cost doesn't disappear, it just moves to a place where you were going to pay it anyway (you can't make an HTTP request without building the client).

## Test plan

- [x] `tests/utils/test_ssl_verification.py` — 10 pass, including new `test_env_flip_swaps_cached_client` covering verify → no-verify → verify
- [x] `tests/utils/test_http_requests.py` — all 5 pass
- [x] `uvx ruff format` / `uvx ruff check` clean
- [ ] Manual repro of WB-33539 scenario against a self-signed endpoint


drtangible (approved):
👍

PR #7071, opened 2026-06-05
fix(weave): qualify bare wb_run_ids in calls-query filters

gtarpenning:
Jira: [WB-35246](https://coreweave.atlassian.net/browse/WB-35246)

## Summary
- Querying calls with a bare wandb run id (e.g. `nf18qjc6`) in `wb_run_ids` 400s with `InvalidIdFormat` (the converter expects `entity/project/run`), spiking prod errors on `/calls/stream_query` + `/calls/query_stats`.
- Fix in the external->internal adapter, where every read consumer converges (SDK `get_calls`, stainless, raw REST): the query is already project-scoped, so bare ids are qualified to the request's `entity/project` before conversion. Full `entity/project/run` ids pass through; malformed ids still raise.
- Documents the accepted formats on `CallsFilter.wb_run_ids` for SDK users.

## Testing
extends `test_trace_call_query_filter_wb_run_ids` (bare resolves same as qualified, mixed ids, both stream + query_stats paths); passes sqlite + clickhouse.

[WB-35246]: https://coreweave.atlassian.net/browse/WB-35246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/external_to_internal_trace_server_adapter.py
+++ b/weave/trace_server/external_to_internal_trace_server_adapter.py
@@ -265,14 +265,13 @@ class ExternalTraceServer(tsi.FullTraceServerInterface):
         return res
 
     def _ext_to_int_run_ids(self, run_ids: list[str], ext_project_id: str) -> list[str]:
-        # Queries are already scoped to a project, so qualify bare run ids
-        # (no "/") to the "entity/project/run" the converter requires.
-        return [
-            self._idc.ext_to_int_run_id(
-                run_id if "/" in run_id else f"{ext_project_id}/{run_id}"
-            )
-            for run_id in run_ids
-        ]
+        # Bare run ids (no "/") are qualified to the queried project, since the
+        # converter requires the full "entity/project/run" form.
+        int_run_ids = []
+        for run_id in run_ids:
+            qualified = run_id if "/" in run_id else f"{ext_project_id}/{run_id}"
+            int_run_ids.append(self._idc.ext_to_int_run_id(qualified))
+        return int_run_ids
 
     def calls_query(self, req: tsi.CallsQueryReq) -> tsi.CallsQueryRes:
         req = req.model_copy(deep=True)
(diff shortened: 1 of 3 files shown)


weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -264,16 +264,24 @@ def call_read(self, req: tsi.CallReadReq) -> tsi.CallReadRes:
             res.call.wb_user_id = self._idc.int_to_ext_user_id(res.call.wb_user_id)
         return res
 
+    def _ext_to_int_run_ids(self, run_ids: list[str], ext_project_id: str) -> list[str]:
+        # Bare run ids (no "/") are qualified to the queried project, since the
+        # converter requires the full "entity/project/run" form.
```
mscavezze-cw: I find this comment and the naming of the parameter confusing.

`entity/project/run` suggests to me that we're mashing together 3 ids: entity, project, and run. The naming of the parameters suggest that we have 2 types of ids, project, and runs (and not in that order). Does `ext_project_id` contain both entity and project (my assumption based on the comment and diff), or just project (my assumption based on naming)?

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -355,13 +362,13 @@ def calls_delete(self, req: tsi.CallsDeleteReq) -> tsi.CallsDeleteRes:
...
+        original_project_id = req.project_id
+        req.project_id = self._idc.ext_to_int_project_id(original_project_id)
         if req.filter is not None:
             if req.filter.wb_run_ids is not None:
-                req.filter.wb_run_ids = [
-                    self._idc.ext_to_int_run_id(run_id)
-                    for run_id in req.filter.wb_run_ids
-                ]
+                req.filter.wb_run_ids = self._ext_to_int_run_ids(
+                    req.filter.wb_run_ids, original_project_id
```
mscavezze-cw, line 372: I feel like I've reviewed the same change 3 times. Consider a follow-up PR to extract a bunch of duplicated code to a helper function.

mscavezze-cw (approved):
Looks good, aside from a couple of comments about readability. Thanks for the ergonomic improvement for callers!

PR #7076, opened 2026-06-08
feat(weave_ts): setAttributes/addEvent on Tool; plural setAttributes on Turn

rgao-coreweave:
## Description

Brings the GenAI `Tool` and `Turn` emitters in line with the landed Python SDK (#7131), which exposes `set_attributes(dict)` / `add_event(...)` on the shared span mixin.

### Tool
- **`setAttributes(attributes)`** — stamp arbitrary OTel attributes on the tool span (e.g. `weave.display_name`, enterprise tags). Pass an object whether you have one key or many.
- **`addEvent(name, attributes?, startTime?)`** — record an OTel span event (e.g. `weave.permission_request`).
- Both are chainable and, after `end()`, warn-and-no-op via the shared `_warnIfEnded` helper.

### Turn
- **`setAttributes(attributes)`** — plural form added, mirroring Python + OTel JS `Span.setAttributes`.
- **`setAttribute(key, value)` kept as a `@deprecated` alias** delegating to `setAttributes`. `Turn.setAttribute` already shipped on `master` and has live consumers (e.g. `weave-openclaw`), so this is intentionally **non-breaking** — existing callers keep working, with the doc-comment steering new code to the plural form.

### Naming
`setAttributes` (plural, object arg) matches the Python SDK's `set_attributes(dict)` and OTel JS's `Span.setAttributes`. This is part of a coordinated plural rename across the emitter family — `LLM` (#7078) and `SubAgent` (#7077) get the same treatment. Only `Turn` carries the back-compat alias, since it's the only emitter whose singular form had already released; adding a "deprecated-from-birth" alias to the others would be dead surface.

## Testing

`cd sdks/node && npx jest src/__tests__/genai/tool.test.ts src/__tests__/genai/turn.test.ts` — 14 passing. Coverage:
- Tool: `setAttributes` (multi-key) + `addEvent` (with attributes and a backdated timestamp) land on the span; after `end()` both warn once per method and nothing reaches the span.
- Turn: `setAttributes` writes multiple keys / returns `this` / no-ops after `end()`; the deprecated `setAttribute` alias still lands its attribute and chains.


rgao-coreweave:
* **#7076** <a href="https://app.graphite.com/github/pr/wandb/weave/7076?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7076?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/genai/tool.ts
```diff
@@ -76,6 +83,35 @@ export class Tool {
     return new Tool(span, opts.name, opts.args ?? '', opts.toolCallId ?? '');
   }
 
+  /** Set a single attribute on the Tool span. Warns and no-ops after `end()`.
+   *  Mirrors OTel `Span.setAttribute`. */
+  setAttribute(key: string, value: AttributeValue): this {
+    if (this._warnIfEnded('setAttribute')) return this;
+    this.span.setAttribute(key, value);
+    return this;
+  }
```
drtangible: Looks like we went with `set_attributes` on the Python side. Should we follow suit? We could mark `Turn.setAttribute` as deprecated and add there as well?

sdks/node/src/__tests__/genai/tool.test.ts
```diff
@@ -62,4 +62,61 @@ describe('Tool', () => {
...
+    tool.addEvent('weave.permission_request');
+    turn.end();
+
+    // One warning per method, naming the method that was called too late.
+    expect(warnSpy).toHaveBeenCalledTimes(2);
+    for (const method of ['setAttribute', 'addEvent']) {
+      expect(warnSpy).toHaveBeenCalledWith(
+        expect.stringContaining(`Tool.${method}() called after end()`)
+      );
+    }
```
drtangible, line 116: Nit: wonder if the test is more concise + illustrative of the behavior w/o parameterizing here.

```tsx
expect(warnSpy).toHaveBeenCalledTimes(2);
expect(warnSpy).toHaveBeenCalledWith('weave.Tool.setAttribute() called after end() — data will not be recorded on the span.');
expect(warnSpy).toHaveBeenCalledWith('weave.Tool.addEvent() called after end() — data will not be recorded on the span.');
```

drtangible (approved):
👍 👍 👍

rgao-coreweave:
Superseded by #7195, which adds setAttributes/addEvent to all four emitters (Tool/LLM/SubAgent/Turn) via a shared SpanBase, mirroring Python #7131. All of this PR's Tool + Turn attribute work lives there now.

PR #7077, opened 2026-06-08
feat(weave_ts): nest LLM/Tool/SubAgent under a SubAgent

rgao-coreweave:
## Summary

Add child-span factories to `SubAgent` — `startLLM`, `startTool`, `startSubagent` — mirroring `Turn`, so a sub-agent's own model calls, tools, and nested sub-agents nest under its `invoke_agent` span instead of flattening onto the parent Turn. `SubAgent` now holds its OTel context + `conversationId` (captured at `create()`) so it can parent children.

The `setAttributes` / `addEvent` mutators and post-hoc start/end timing already landed on master via the shared `SpanBase` (#7195, #7078); this is the remaining child-factory piece.

## Design note — no out-of-block bug (unlike the Python SDK)

The Python counterpart (#7133) had to add explicit OTel-context threading because it creates a child span **lazily** at `with child:` from whatever ambient context is current — so a child entered after its parent exits could detach. The TS SDK instead binds each child's parent context **eagerly** inside `start*()` (via `ChildSpanContext`) and never reads ambient context, so the parent link holds regardless of teardown order or when the child is created. A regression test locks this in.

## Test plan

- [x] `pnpm build` (sdks/node) — 0 errors (js + mjs), `typecheck:cjs` clean, eslint clean
- [x] `pnpm test` genai suite green; `subagent.test.ts` (11 tests) covers:
  - child LLM + Tool nest under a SubAgent
  - `startSubagent` full chain: `turn → outer → inner → llm`, each parenting to the level above
  - regression guard: a child created **after** the SubAgent *and* Turn `end()` still nests under the sub (proves eager binding, not ambient-context parenting)

## Notes

Rebased onto current `master` (the genai `Conversation` / `SubAgent` refactor). `subagent.ts` was resolved to keep master's private-field / `record()` / agent-identity structure and graft in the child factories + stored OTel context, mirroring how `Turn` does it.


rgao-coreweave:
* **#7077** <a href="https://app.graphite.com/github/pr/wandb/weave/7077?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7077?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdks/node/src/genai/subagent.ts
```diff
@@ -52,17 +63,21 @@ export interface SubAgentInit extends SpanInitBase {
  * });
  *
  * try {
+ *   const llm = sub.startLLM({model: 'gpt-4o', providerName: 'openai'});
  *   // ... orchestrate the sub-agent's LLM/Tool calls ...
+ *   llm.end();
```
drtangible: Do we want these in the first example in this typedoc for consistency?

sdks/node/src/genai/subagent.ts
```diff
@@ -32,7 +40,10 @@ export interface SubAgentInit extends SpanInitBase {
  * its model.
  *
  * Created by `weave.startSubagent()` (or `turn.startSubagent()`, or
- * `llm.startSubagent()`) and terminated with `end()`.
+ * `llm.startSubagent()`) and terminated with `end()`. Children (LLM, Tool,
+ * SubAgent) attach via the `startLLM`, `startTool`, `startSubagent` methods,
+ * so a sub-agent's own model calls and tools nest under its `invoke_agent`
```
drtangible: This change doesn't introduce this, but this line shows we refer to subagent as a single word, a hyphenated word, and two separate words. Would be nice if we eventually make that consistent across the user-facing pieces of our SDK, at least.

drtangible (approved):
🔥 🔥 🔥

sdks/node/src/genai/subagent.ts
rgao-coreweave: true, sticking to one word `subagent`

rgao-coreweave: checked 

- http://code.claude.com/docs/en/agent-sdk/subagents
- https://developers.openai.com/codex/subagents

sdks/node/src/genai/subagent.ts
rgao-coreweave: true, `SubAgent` looks good for now in types.

PR #7078, opened 2026-06-08
feat(weave_ts): post-hoc start/end times on GenAI spans

rgao-coreweave:
## Description

Lets every GenAI emitter reconstruct a span from **post-hoc data** — replaying a transcript or persisting a completed call after the fact, where the real wall-clock window is only known once the call has finished.

The shared `SpanBase` carries the capability so all four emitters (`LLM`, `Tool`, `SubAgent`, `Turn`) get post-hoc times uniformly, mirroring how `setAttributes`/`addEvent` are defined once in `SpanBase` (#7195 / Python #7131):

- **`SpanInitBase.startTime?: TimeInput`** — backdate the span's start to when the call actually began (vs. when `start*()` was invoked). Each emitter's `Init` extends `SpanInitBase`, and each `create()` forwards `opts.startTime` into its `tracer.startSpan({ startTime })`.
- **`SpanEndOptions.endTime?: TimeInput`** — backdate the span's close, so a replayed span carries an accurate duration. Each emitter's `end()` takes `SpanEndOptions` and delegates to `SpanBase._closeSpan(opts)`, which applies `endTime` to `span.end(endTime)`.

`startTime`/`endTime` are forwarded verbatim to OTel, which treats `undefined` as "use the current time" — so the params are purely additive and the default path is unchanged.

`_closeSpan` also consolidates the error-status + `span.end()` tail that was previously copy-pasted across all four emitters (the TS counterpart of Python `_SpanBase._end_otel_span`). The start path has no shared logic to extract beyond the `startTime` field, so each `create()` keeps its own inline `tracer.startSpan` call.

## Testing

`cd sdks/node && npx jest src/__tests__/genai/` — **62 passing**, including a `startTime`/`endTime` backdate test for each emitter (LLM/Tool/SubAgent/Turn) asserting the exported span's start/end seconds via the shared `expectSpanTimesToMatch` helper. `prettier --check` + `tsc -p tsconfig.cjs.json` clean.


rgao-coreweave:
* **#7078** <a href="https://app.graphite.com/github/pr/wandb/weave/7078?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7078?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

drtangible (approved):
👍 👍 👍

PR #7087, opened 2026-06-08
feat(weave_ts): Emit `chat` spans for OpenAI Agents SDK.

drtangible:
## Description

* Emit `chat` spans via `WeaveOtelTracingProcessor`.
* Note: Much of the data present on these will be introduced further up the stack, after getting plumbing for the other supported types in place.
* See #6917 for reference on the Python SDK side.

PR #7085, opened 2026-06-08
feat(weave_ts): Emit `invoke_agent` spans for OpenAI Agents SDK.

drtangible:
## Description

* Emit `invoke_agent` spans via `WeaveOtelTracingProcessor`.
* See #6917 for reference on the Python SDK side.

PR #7083, opened 2026-06-08
chore(weave_ts): Add some more tests around @openai/agents integration

drtangible:
## Description

* As we introduce the new tracing processor, covering the existing one with a one more illustrative test case, so we can compare the previously-emitted data w/the new stuff.

PR #7089, opened 2026-06-08
feat(weave_ts): Emit message data on 'chat' spans for OpenAI Agents SDK.

drtangible:
## Description

* Followup from #7087 to include add'l data.
* See #6917 for reference on the Python SDK side.
* Updates example, getting data flowing through to the `Agents` tab:

```bash
➜  node git:(06-04-emit-more-data-on-chat-spans) WEAVE_USE_OTEL_V2=true tsx examples/openaiAgents.ts
Initializing project: benmocha-/example

Query: What's the temperature difference between Tokyo and San Francisco in celsius? Use the calculator to subtract.
Answer: Tokyo is 28°C and San Francisco is 18°C, so the difference is **10°C**.

Query: If the temperature in London tripled, what would it be? Get the weather first, then use the calculator.
Answer: London is 12°C, so if it tripled, it would be 36°C.
```

Screenshot 2026-06-08 at 12 40 35 PM

* Usages looks off here, taking a look at that.

### DIFF
--- a/sdks/node/src/integrations/openai-agents/messages.ts
+++ b/sdks/node/src/integrations/openai-agents/messages.ts
@@ -1,0 +1,375 @@
+/**
+ * The OpenAI Agents SDK populates `ResponseSpanData._input`, `ResponseSpanData._response`,
+ * `GenerationSpanData.input`, and `GenerationSpanData.output`.
+ *
+ * These functions help deal with that data, when present.
+ */
+
+import type OpenAI from 'openai';
+import type {GenerationSpanData, ResponseSpanData} from '@openai/agents';
+import type {Message, MessagePart} from '../../genai';
+import {MediaPart, urlToMediaPart} from '../../utils/urlToMediaPart';
+import {withIndex} from '../../utils/withIndex';
+import {findLastIndex} from '../../utils/findLastIndex';
+
+/**
+ * `ResponseSpanData` tightened to the shape the OpenAI Responses-API model
+ * (and any adapter that normalizes through it) populates. Upstream types
+ * `_input` loosely.
+ */
+type ResponseSpanDataWithInput = Omit<ResponseSpanData, '_input'> & {
+  _input: string | OpenAI.Responses.ResponseInputItem[];
+};
+
+/**
+ * `ResponseSpanData` tightened to the shape the OpenAI Responses-API model
+ * (and any adapter that normalizes through it) populates. Upstream types
+ * `_response` loosely.
+ */
+type ResponseSpanDataWithOutput = Omit<ResponseSpanData, '_response'> & {
+  _response: OpenAI.Responses.Response;
+};
+
+/**
+ * `GenerationSpanData` tightened to the shape the OpenAI Chat Completions-API model
+ * (and any adapter that normalizes through it) populates. Upstream types
+ * `input` loosely.
+ */
+type GenerationSpanDataWithInput = GenerationSpanData & {
+  input: OpenAI.Chat.Completions.ChatCompletionMessageParam[];
+};
+
+/**
+ * `GenerationSpanData` tightened to the shape the OpenAI Chat Completions-API model
+ * (and any adapter that normalizes through it) populates. Upstream types
+ * `output` loosely.
+ */
+type GenerationSpanDataWithOutput = GenerationSpanData & {
+  output: OpenAI.Chat.Completions.ChatCompletionMessage[];
+};
+
+export function hasResponsesInput(
+  spanData: ResponseSpanData
+): spanData is ResponseSpanDataWithInput {
+  return !!spanData._input;
+}
+
+export function hasResponsesOutput(
+  spanData: ResponseSpanData
+): spanData is ResponseSpanDataWithOutput {
+  return !!(spanData._response && spanData._response.object === 'response');
+}
+
+export function hasChatCompletionInput(
+  spanData: GenerationSpanData
+): spanData is GenerationSpanDataWithInput {
+  return !!(spanData.input && hasRole(spanData.input));
+}
+
+export function hasChatCompletionOutput(
+  spanData: GenerationSpanData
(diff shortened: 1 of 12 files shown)


drtangible:
* **#7125** <a href="https://app.graphite.com/github/pr/wandb/weave/7125?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7107** <a href="https://app.graphite.com/github/pr/wandb/weave/7107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7089** <a href="https://app.graphite.com/github/pr/wandb/weave/7089?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7089?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7088** <a href="https://app.graphite.com/github/pr/wandb/weave/7088?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7087** <a href="https://app.graphite.com/github/pr/wandb/weave/7087?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7086** <a href="https://app.graphite.com/github/pr/wandb/weave/7086?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7085** <a href="https://app.graphite.com/github/pr/wandb/weave/7085?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7084** <a href="https://app.graphite.com/github/pr/wandb/weave/7084?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7083** <a href="https://app.graphite.com/github/pr/wandb/weave/7083?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7082** <a href="https://app.graphite.com/github/pr/wandb/weave/7082?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7091** <a href="https://app.graphite.com/github/pr/wandb/weave/7091?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sdks/node/examples/openaiAgents.ts
```diff
@@ -80,18 +86,17 @@ async function main() {
     tools: [getWeatherTool, calculateTool],
   });
 
+  const runner = new Runner({groupId: generateGroupId()});
+
   const queries = [
-    'What is the weather like in Tokyo and San Francisco?',
-    'What is (17 * 4) + 93?',
+    "What's the temperature difference between Tokyo and San Francisco in celsius? Use the calculator to subtract.",
+    'If the temperature in London tripled, what would it be? Get the weather first, then use the calculator.',
```
andrewtruong: Seems odd to change it here and not in the test.  Is there any reason for that?

sdks/node/src/integrations/openai-agents/messages.ts
```diff
@@ -0,0 +1,428 @@
...
+
+/**
+ * Split a `data:` URL into (mimeType, payload). Payload is the raw string
+ * after the comma — base64-encoded content is NOT decoded; it's passed
+ * through as-is so the wire format matches what the producer embedded.
+ * Returns `undefined` for non-data URLs.
+ */
+function parseDataUrl(
+  url: string
+): {mimeType: string; payload: string} | undefined {
```
andrewtruong: nit: things like this seem generic and prob belong in a util file.  I'm sure other integrations/we internally will use it too.

drtangible: Good call -- there are several functions in here that we'll likely pull out for reuse for constructing attrs for other integrations. (Think Python has similar fns eg, `invoke_agent_attributes`,  `execute_tool_attributes`). I can take a quick pass on that yank out things that aren't specific to OpenAI here.

sdks/node/examples/openaiAgents.ts
drtangible: Was using this locally to get more robust data all the way into Weave. These `examples/` are sort of acting similar to a manual version of the `live/` e2e tests (which aren't actually running, as they're all `.skip`ped).

I could update the unit test to also cover a more robust case, though I'm hoping to update our test suite to act on recorded+mocked responses rather than setting up + mocking implementation details (similar to what we have going on w/VCR recordings on the python side, I believe). Will back out this change and work on that, so we can start to simplify things into just documentation + tests.

sdks/node/src/integrations/openai-agents/messages.ts
drtangible: Took a quick pass to hoist out pure utility fns.

sdks/node/src/utils/findLastIndex.ts
andrewtruong, line 1: is it idiomatic to have a file with just 1 function?  Why not just a single util file?

andrewtruong (approved):

sdks/node/src/utils/findLastIndex.ts
drtangible: Yeah, pretty common! In some style guides / lint setups I've worked with in the past, a preference for a single export is sometimes enforced. (Or at least a preference for file names to match the name of their single named export, when applicable).

I typically default to a 1:1 mapping between file name and a named export, and then consolidate as organization seems natural. For example, I could imagine an `arrayPolyfills.ts` module that exported  `findLastIndex.ts` and `withIndex`), if we found that helpful in the future. (But a generic context-less `utils.ts` file would typically be an antipattern IMO).

Many of our existing functions in this directory follow this convention (`warnOnce`, `userAgent`, `packageVersion`, `globalSingleton`, etc.). An example of one that doesn't is `npmModuleUtils.ts`, which only exports `requirePackageJson`. I'd argue that we don't even need to introduce something called `npmModuleUtils` into our codebase until organization around that concept emerged, and that it'd be preferable to rename that file `requirePackageJson.ts` for discoverability.

PR #7095, opened 2026-06-08
fix(weave): honor skip_clickhouse_client marker so VCR tests use sqlite

gtarpenning:
## Summary
- `verifiers_test` and other VCR integration tests fail nightly with `CannotOverwriteExistingCassetteException`: they run on the default clickhouse backend, so VCR intercepts the clickhouse HTTP flush, which the cassette has no match for. The `skip_clickhouse_client` marker was meant to prevent exactly this but was dead, `get_trace_server_flag` never honored it (and `test_skip_clickhouse_client` already asserts it should).
- Honor the marker at the single backend chokepoint, routing marked tests to sqlite.

## Testing
exercised by the nightly verifiers/integration shards.

### DIFF
--- a/tests/trace_server/conftest.py
+++ b/tests/trace_server/conftest.py
@@ -93,6 +93,11 @@ def pytest_collection_modifyitems(config, items):
 
 
 def get_trace_server_flag(request):
+    # Tests marked `skip_clickhouse_client` run on sqlite: they patch HTTP via VCR,
+    # and a real clickhouse client would emit unmatched localhost traffic.
+    node = getattr(request, "node", None)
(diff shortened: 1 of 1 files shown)


gtarpenning:
Closing: this approach is too broad. Honoring `skip_clickhouse_client` in `get_trace_server_flag` would flip all ~29 integration suites (anthropic, openai, langchain, ...) from the clickhouse backend to sqlite, not just verifiers. The marker is already enforced for trace_server-marked shards via the workflow `-m "... and not skip_clickhouse_client"` filter; integration-only shards intentionally run on CH. A correct fix must be scoped to the verifiers VCR config (CH/localhost passthrough), not the global backend selector.

PR #7105, opened 2026-06-09
fix(weave): route agent chat media to the user or assistant message by direction

zbirenbaum:
- Media attached to an agent's LLM call was rendering on the assistant bubble instead of the user's
- chat view now routes each content ref to the user or assistant message based on whether it appears in the input or output message parts matched by object digest

Before:
image

After:
image

### DIFF
--- a/weave/trace_server/agents/chat_view.py
+++ b/weave/trace_server/agents/chat_view.py
@@ -500,6 +500,94 @@ def _content_refs(span: AgentSpanSchema) -> list[str]:
     return [str(r) for r in (span.content_refs or []) if r]
 
 
+# Content part types that reference uploaded media by ref rather than inlining
+# text. The session SDK emits attached media as ``uri`` parts (see
+# weave/session/session_otel.py::_media_to_part); ``blob``/``file`` are accepted
+# defensively in case other producers use a ref-bearing variant.
+_MEDIA_PART_TYPES = {"uri", "blob", "file"}
+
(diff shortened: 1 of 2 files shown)


weave/trace_server/agents/chat_view.py
```diff
@@ -500,6 +500,94 @@ def _content_refs(span: AgentSpanSchema) -> list[str]:
...
+def _ref_digest(ref: str) -> str:
+    """Return the trailing object digest of a content ref.
+
+    The same object is referenced in two forms sharing one digest: the inline
+    message ``uri`` part uses the external ``weave:///entity/project/...`` form,
+    while the span-level ``content_refs`` uses the internal
+    ``weave-trace-internal:///...`` form. Both end in ``...:<digest>``, so the
+    digest is the stable key for matching one to the other.
+    """
+    return ref.rsplit(":", 1)[-1] if ":" in ref else ref
```
rgao-coreweave: `if ":" in ref else ref` is redundant

rgao-coreweave: rsplit already handles the no-colon case

rgao-coreweave (approved):
LGTM

one nit

weave/trace_server/agents/chat_view.py
zbirenbaum: Thanks, fixed!

PR #7108, opened 2026-06-09
refactor(weave): remove bare type:ignore / cast(Any) in trace server

gtarpenning:
## Summary
- `trace_server_interface.py`: replace `__pydantic_config__ = ...  # type: ignore` with the `@with_config(ConfigDict(extra="allow"))` decorator on `ExtraKeysTypedDict`
- `annotation_spec.py`: drop the stale ignore on `model_json_schema()`
- `schema_converters.py`: use `None` directly instead of `cast(Any, None)` for `storage_size_bytes`; drop the now-unused `cast` import
- `agents/types.py`: replace 3x `# type: ignore[index]` with a typed `_derived_value_type()` lookup that narrows the str key
- `calls_query_builder.py`: drop the stale bare ignore on `add_param(operand.literal_)`

## Testing
type-only change; pyright + ty + ruff clean on touched files, agent/calls/table query-builder unit tests pass

### DIFF
--- a/weave/trace_server/agents/types.py
+++ b/weave/trace_server/agents/types.py
@@ -117,6 +117,13 @@ _ALLOWED_AGGS_BY_TYPE: dict[AgentSpanStatsValueType, set[AgentSpanStatsAggregati
 }
 
 
+def _derived_value_type(key: str) -> AgentSpanStatsValueType:
+    for metric, value_type in AGENT_SPAN_STATS_DERIVED_VALUE_TYPES.items():
+        if metric == key:
+            return value_type
+    raise ValueError(f"unknown derived span value: {key!r}")
+
+
 def _as_utc(dt: datetime.datetime) -> datetime.datetime:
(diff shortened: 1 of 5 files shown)


weave/trace_server/agents/types.py
```diff
@@ -117,6 +117,13 @@
 }
 
 
+def _derived_value_type(key: str) -> AgentSpanStatsValueType:
+    for metric, value_type in AGENT_SPAN_STATS_DERIVED_VALUE_TYPES.items():
+        if metric == key:
+            return value_type
+    raise ValueError(f"unknown derived span value: {key!r}")
```
jtschoonhoven, line 124: Are you sure this function is necessary? Could you instead just type the constant like this?
```
AgentSpanStatsValueType: dict[str, AgentSpanStatsValueType] = {...}`
```

jtschoonhoven (approved):

PR #7111, opened 2026-06-09
fix(weave): route feedback/cost purge DELETE to _local in distributed mode

gtarpenning:
https://coreweave.atlassian.net/browse/WB-35450

## Summary
- `feedback` and `llm_token_prices` purge issued DELETE against the Distributed engine table, failing `Code 36 BAD_ARGUMENTS: DELETE ... not supported for Distributed` on the distributed 2s2r nightly ([CI run](https://github.com/wandb/weave/actions/runs/27110446368), WB-35378).
- Distributed mode: route both purge DELETEs to `{table}_local` + `ON CLUSTER` (via `_mutation_table_name` / `_mutation_cluster_name`) so the mutation fans across shards.
- Replicated and cloud modes keep the plain `DELETE FROM {table}`: replicated tables self-replicate the mutation via Keeper, so `ON CLUSTER` is unnecessary there.
- Execution moves to `_command` (+ lightweight-mutation settings) in all modes.

## Testing
full-SQL builder tests parametrized over cloud / replicated / distributed in test_on_cluster_query_builder.py.

verified on live clusters: `feedback_purge`/`cost_purge` functional tests pass on both 1S3R (replicated, plain `DELETE FROM feedback`) and 2s2r (distributed, `DELETE FROM feedback_local ON CLUSTER`); on 1S3R the plain delete propagated to all 3 replicas with zero pending/failed mutations and an empty replication queue.


weave/trace_server/orm.py
```diff
@@ -497,6 +500,17 @@ def prepare(
...
 
 
+def _format_table_name_with_cluster(table_name: str, cluster_name: str | None) -> str:
+    """Append `ON CLUSTER {cluster_name}` to a table reference when clustered.
+
+    Callers pass the already-resolved table name (e.g. `<table>_local` in
+    distributed mode); this only appends the cluster clause.
+    """
+    if cluster_name:
+        return f"{table_name} ON CLUSTER {cluster_name}"
```
amwarrier, line 448: is it safe to add ON CLUSTER blindly when cluster_name is passed not matter what the action is?

amwarrier (commented):
replicated mode now gets ON CLUSTER where it didn't before. Is that expected? From the PR description, we want to use ON CLUSTER only for S>1 and not R>1
Please ignore if you tested this on 1S3R

weave/trace_server/orm.py
gtarpenning: good flag. tested on 1S3R and it was safe (replicated tables self-replicate the lightweight DELETE via Keeper, so the plain delete hit all 3 replicas with zero pending/failed mutations and an empty replication queue). ON CLUSTER is only actually needed for S>1, to fan the mutation across shards.

still, adding it whenever a cluster is set is sloppy, so I tightened it to be explicit about intent: `_mutation_cluster_name()` only returns the cluster in distributed mode (mirrors `_mutation_table_name`), so replicated + cloud go back to the plain `DELETE FROM {table}`. reverified the distributed path on 2s2r too (`DELETE FROM feedback_local ON CLUSTER`).

amwarrier (approved):

PR #7112, opened 2026-06-09
fix(weave): GLOBAL JOIN the cost-merge query on distributed clusters

gtarpenning:
## Summary
- WB-35378: the cost-merge query LEFT JOINs `llm_token_prices` against per-call usage; on distributed clusters both are Distributed tables so the JOIN is double-distributed and fails with `Code 288 Double-distributed IN/JOIN subqueries denied`.
- That failed JOIN cascades to `KeyError: 'costs'` in `test_summary_tokens_cost` / `test_opentelemetry_cost_calculation` (the costs object is omitted when the query returns nothing). CI run: https://github.com/wandb/weave/actions/runs/27110446368
- CH 24.3+ analyzer silently ignores `distributed_product_mode='global'`, so the JOIN must be written as explicit `GLOBAL LEFT JOIN`. Adds a typed `global_` flag to the ORM `join()` and passes it from `get_ranked_prices`.

## Testing
full-SQL-string tests for the GLOBAL cost CTE + ORM join; functional cost path validated by the 2s2r nightly.


jwlee64 (approved):
lgtm

PR #7113, opened 2026-06-09
fix(weave): emit GLOBAL IN for object-ref UNION subqueries on distributed clusters

gtarpenning:
## Summary
- The object-ref condition CTEs read Distributed `object_versions` + `table_rows` via `UNION ALL`. The CH analyzer won't globalize an `IN (SELECT ref FROM cte)` against that shape via `distributed_product_mode` (CH #99370), so the 2s2r nightly hit `Code 36` "Invalid local IN function name globalIn". Emit `GLOBAL IN` explicitly at those sites only.
- Scoped down from the original blanket rewrite: `distributed_product_mode='global'` still covers the other (non-UNION) IN/JOIN shapes, so calls-query / eval-results builders are untouched. The annotation Code 288 fix is split into #7118.

## Performance
Local benchmark, object-ref filter as `count()` over the full filtered scan, `IN` vs `GLOBAL IN`:

| cluster | plain `IN` | `GLOBAL IN` |
|---|---|---|
| single-node, 500k calls | ~195 ms | ~199 ms |
| 2s2r distributed, 50k calls | Code 36 (fails) | ~110 ms |

Single-node is neutral (delta flips sign run-to-run, within ~±5% noise, identical row counts). On distributed, `GLOBAL IN` is what lets the query run at all.

## Testing
object_ref query-builder unit tests updated to the GLOBAL IN SQL; full `query_builder` unit suite green. Functional + benchmark paths validated on a local single-node and 2s2r cluster.

### DIFF
--- a/tests/trace_server/query_builder/test_object_ref_calls_query_builder.py
+++ b/tests/trace_server/query_builder/test_object_ref_calls_query_builder.py
@@ -53,10 +53,10 @@ def test_object_ref_filter_simple() -> None:
                   OR calls_merged.ended_at IS NULL)
            GROUP BY (calls_merged.project_id,
                      calls_merged.id)
-           HAVING (((coalesce(nullIf(JSON_VALUE(any(calls_merged.output_dump), {pb_3:String}), 'null'), '') IN
+           HAVING (((coalesce(nullIf(JSON_VALUE(any(calls_merged.output_dump), {pb_3:String}), 'null'), '') GLOBAL IN
                       (SELECT ref
                        FROM obj_filter_0)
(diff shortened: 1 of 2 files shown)


jtschoonhoven (approved):
I see in the PR description you decided against using `distributed_product_mode='global'` but I'm curious why?

gtarpenning:
> * Scoped down from the original blanket rewrite: `distributed_product_mode='global'` still covers the other

@jtschoonhoven 
its actually: "Scoped down from the original blanket rewrite, distributed_product_mode='global' still covers the other..." as in: "`distributed_product_mode='global'` is a better solution, no need to do the insane thing claude tried to do first"

PR #7119, opened 2026-06-09
fix(weave): run INSERT on distributed tables during migration

gtarpenning:
## Summary
- On distributed (2s2r) clusters the migrator skipped every `INSERT INTO`, so migration 006's `llm_token_prices` cost seeds never landed. The cost query then ranked over only the checkpoint-seeded rows and returned a different price than single-node (e.g. gpt-4o), failing `test_summary_tokens_cost`.
- Run `INSERT ... VALUES` seeds synchronously against the Distributed table (the engine fans rows to shards); keep skipping `INSERT ... SELECT` backfills (per-shard). Also exclude the `_local` twins of seed tables from the per-test TRUNCATE so seeded prices survive between tests.

## Testing
`test_summary_tokens_cost` passes on a local 2s2r cluster and on single-node; migrator (104) + costs (10) unit suites green.

### DIFF
--- a/tests/trace_server/conftest.py
+++ b/tests/trace_server/conftest.py
@@ -159,7 +159,10 @@ def _discover_truncatable_tables(ch_client, database: str) -> list[str]:
         "AND engine NOT IN ('View', 'MaterializedView', 'Distributed') "
         "ORDER BY name"
     )
-    return [row[0] for row in result.result_rows if row[0] not in SEED_DATA_TABLES]
+    seed_tables = SEED_DATA_TABLES | {
+        t + ch_settings.LOCAL_TABLE_SUFFIX for t in SEED_DATA_TABLES
+    }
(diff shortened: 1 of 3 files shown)


amwarrier (commented):
I understand that the INSERT INTO ... VALUES now work for any new clusters. What about the existing clusters? Will we fix them?

gtarpenning:
> I understand that the INSERT INTO ... VALUES now work for any new clusters. What about the existing clusters? Will we fix them?

@amwarrier 

this is only relevant for costs, which get cleaned up at query time. this bug fix is really just to get the tests to pass cleanly, i dont think its an issue. also, very few existing multi s deployments, happy to write one off patches if this does come up. finally, there is also remedy through the app, where you can insert your own custom costs, which should always work correctly.

amwarrier (approved):

PR #7131, opened 2026-06-10
feat(weave): add set_attributes and add_event to Tool/LLM/SubAgent/Turn

rgao-coreweave:
## Summary

Two escape hatches on the `_SpanBase` mixin — inherited by `Tool`, `LLM`, `SubAgent`, `Turn`:

- `set_attributes(attributes: dict)` — stamp arbitrary OTel attributes (e.g. `weave.display_name` for Claude Code, enterprise tags). Mirrors OTel's `Span.set_attributes`. Callers pass a dict whether they have one key or many.
- `add_event(name, attributes=None, timestamp=None)` — record OTel span events (e.g. `weave.permission_request`)

Both warn (via `logger.warning`) when called on a span that hasn't been started or has already ended — silent only when OTel isn't installed or Weave is disabled. Naming/fix hint is included in the warning, e.g. `set_attributes({'weave.tag'}) ignored: span not started. Use \`with\` for live tracing, or log_turn() for batch ingest.`

Mirrors the TypeScript SDK's `setAttribute`/`addEvent` surface from wandb/weave#7076, wandb/weave#7077, wandb/weave#7078, lifted onto the shared mixin so the implementation covers all four classes uniformly.

## Test plan

- [x] 23 tests in `tests/session/test_span_attributes.py` — parametrized across `(Tool, LLM, SubAgent, Turn)` for: lands-on-span, no-op-after-end, returns-self, plus mixin-level checks for sequence-value support and warning behavior (unstarted + already-ended)
- [x] Existing session tests still pass
- [x] `uvx ruff check` + `uvx ruff format --check` clean
- [x] `mypy weave/session/session.py` clean (no errors introduced)

### DIFF
--- a/tests/session/test_span_attributes.py
+++ b/tests/session/test_span_attributes.py
@@ -93,11 +93,16 @@ def test_add_event_no_op_after_end(
     assert len(finished_span.events) == 0
 
 
-def test_returns_self_for_chaining() -> None:
-    """Both methods return ``self`` — checked on Tool (mixin, same on all)."""
-    tool = Tool(name="test-tool")
-    assert tool.set_attribute("key", "value") is tool
-    assert tool.add_event("event-name") is tool
+@pytest.mark.parametrize(
+    ("_class_label", "factory", "_span_name"), CASES, ids=CLASS_LABELS
+)
+def test_returns_self_for_chaining(
+    otel_spans: InMemorySpanExporter, _class_label, factory, _span_name
+) -> None:
+    """Both methods return ``self`` for fluent chaining on a live span."""
+    with Session(session_id="test-session"), factory() as span_obj:
+        assert span_obj.set_attribute("key", "value") is span_obj
+        assert span_obj.add_event("event-name") is span_obj
 
 
 def test_set_attribute_accepts_sequence_value(
@@ -110,11 +115,12 @@ def test_set_attribute_accepts_sequence_value(
     assert tuple(chat_span.attributes["gen_ai.response.finish_reasons"]) == ("stop",)
 
 
-def test_no_op_on_unentered_span() -> None:
-    """No OTel span ever created (caller never entered the context manager)."""
+def test_no_op_on_unentered_span(otel_spans: InMemorySpanExporter) -> None:
+    """Unentered span doesn't crash and emits no OTel output."""
     tool = Tool(name="test-tool")
-    assert tool.set_attribute("key", "value") is tool
-    assert tool.add_event("event-name") is tool
+    tool.set_attribute("key", "value")
+    tool.add_event("event-name")
+    assert len(otel_spans.get_finished_spans()) == 0
 
 
 def test_warns_when_span_not_started(caplog: pytest.LogCaptureFixture) -> None:
--- a/tests/session/test_span_attributes.py
+++ b/tests/session/test_span_attributes.py
@@ -1,0 +1,224 @@
+"""Tests for ``set_attribute`` and ``add_event`` on session span classes.
+
+These two methods are escape hatches that delegate to the underlying OTel
+span: ``set_attribute`` lands arbitrary keys not covered by the declared
+GenAI semconv fields, and ``add_event`` records OTel span events.
+
+Both live on ``_SpanBase`` so all four span classes (Tool, LLM, SubAgent,
+Turn) get the same behavior. This file exercises each class independently
+to lock in that uniformity.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timezone
+
+import pytest
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+
+from weave.session.session import (
+    LLM,
+    Session,
+    SubAgent,
+    Tool,
+    Turn,
+)
+
+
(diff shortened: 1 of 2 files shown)


weave/session/session.py
```diff
@@ -187,6 +187,28 @@ def _record_otel_error(self, exc_val: BaseException) -> None:
         self._otel_span.set_status(StatusCode.ERROR, str(exc_val))
         self._otel_span.record_exception(exc_val)
 
+    def set_attribute(self, key: str, value: Any) -> Self:
+        """Stamp an arbitrary OTel attribute. No-op after ``end()``."""
+        if self._otel_span is not None and self._otel_span.is_recording():
+            self._otel_span.set_attribute(key, value)
```
andrewtruong: Might be good to warn the user if the attribute is not set so they know if something is broken

weave/session/session.py
```diff
@@ -187,6 +187,28 @@ def _record_otel_error(self, exc_val: BaseException) -> None:
...
+        return self
+
+    def add_event(
```
andrewtruong, line 239: same here.

Also what would this be used for?

tests/session/test_span_attributes.py
```diff
@@ -0,0 +1,116 @@
...
+        llm.set_attribute("gen_ai.response.finish_reasons", ["stop"])
+    chat_span = _only_span(otel_spans.get_finished_spans(), "chat gpt-4o")
+    assert tuple(chat_span.attributes["gen_ai.response.finish_reasons"]) == ("stop",)
+
+
+def test_no_op_on_unentered_span() -> None:
+    """No OTel span ever created (caller never entered the context manager)."""
+    tool = Tool(name="test-tool")
+    assert tool.set_attribute("key", "value") is tool
+    assert tool.add_event("event-name") is tool
```
drtangible: Is this the same as `test_returns_self_for_chaining`?

tests/session/test_span_attributes.py
```diff
@@ -0,0 +1,116 @@
...
+        span_obj.add_event("weave.late")
+    finished_span = _only_span(otel_spans.get_finished_spans(), span_name)
+    assert len(finished_span.events) == 0
+
+
+def test_returns_self_for_chaining() -> None:
+    """Both methods return ``self`` — checked on Tool (mixin, same on all)."""
+    tool = Tool(name="test-tool")
+    assert tool.set_attribute("key", "value") is tool
+    assert tool.add_event("event-name") is tool
```
drtangible: Any reason this can't follow the above pattern and test all of the classes we expect to exhibit this behavior?

weave/session/session.py
```diff
@@ -187,6 +187,28 @@ def _record_otel_error(self, exc_val: BaseException) -> None:
         self._otel_span.set_status(StatusCode.ERROR, str(exc_val))
         self._otel_span.record_exception(exc_val)
 
+    def set_attribute(self, key: str, value: Any) -> Self:
+        """Stamp an arbitrary OTel attribute. No-op after ``end()``."""
+        if self._otel_span is not None and self._otel_span.is_recording():
+            self._otel_span.set_attribute(key, value)
+        return self
```
drtangible: What're our thoughts on a convenience `set_attributes` function as well?

drtangible (approved):

weave/session/session.py
rgao-coreweave: add_event appends to an ordered list, an event is a specific instant within the span.

set_attribute writes to a key→value map on the span, an attribute describes the span as a whole.

weave/session/session.py
rgao-coreweave: GREAT FEEDBACK! Users might not have started the span correctly, so added a warn here.

tests/session/test_span_attributes.py
rgao-coreweave: good call, differentiate the tests for distinct intent.

tests/session/test_span_attributes.py
rgao-coreweave: Good call!  Parametrized across all four span classes and moved inside with so it exercises chaining on a live recording span.

weave/session/session.py
rgao-coreweave: Makes sense, we can proactively add a convenience function that mirrors OTel's bulk-setter.

weave/session/session.py
andrewtruong: What is an "event"?

weave/session/session.py
```diff
@@ -187,6 +187,85 @@ def _record_otel_error(self, exc_val: BaseException) -> None:
...
+            logger.warning(
+                "%s(%s) ignored: span already ended. Set attributes before "
+                "exiting `with` or calling .end().",
+                operation,
+                key_repr,
+            )
+            return None
+        return self._otel_span
+
+    def set_attribute(self, key: str, value: Any) -> Self:
```
andrewtruong: instead of having two methods, let's just have one `set_attributes`

andrewtruong (approved):

weave/session/session.py
rgao-coreweave: sounds good, just keeping `set_attributes` for now.

PR #7132, opened 2026-06-10
fix(weave): preserve caller-set started_at/ended_at in streaming end()

rgao-coreweave:
## Summary

Streaming `end()` on `LLM` / `SubAgent` / `Turn` overwrote caller-supplied `ended_at` with `now()`, and all four span classes' `_end_otel_span` calls dropped the value before reaching OTel. `SubAgent.__enter__` did the same for `started_at`. Net: caller-set timestamps on the SDK object never reached the emitted OTel span.

Batch path (`log_turn` / `log_session`) already worked. After this PR, streaming and batch produce identical spans for the same inputs.

## Changes

- `LLM.end()` / `SubAgent.end()` / `Turn.end()`: preserve existing `ended_at` (matching `Tool.end()`'s pre-existing pattern)
- All four `end()`s: pass `end_time_ns=_to_ns(self.ended_at)` to `_end_otel_span`
- `SubAgent.__enter__`: preserve existing `started_at` and pass `start_time_ns` (the only `__enter__` that ignored it)

## Why

Reconstructing chat spans from upstream transcripts (replay, queue worker, callback) needs to backdate the OTel span_time to when the LLM call actually happened, not when the SDK called `end()`.

## Test plan

- [x] 13 tests in `tests/session/test_post_hoc_times.py` — 4×3 parametrized across `(Tool, LLM, SubAgent, Turn)` covering `ended_at` round-trip, OTel span times match caller values, default `now()` fallback; plus a streaming-vs-batch regression on `chat` spans
- [x] 241 existing session tests still pass
- [x] `uvx ruff check` + `uvx ruff format --check` clean

### DIFF
--- a/tests/session/test_post_hoc_times.py
+++ b/tests/session/test_post_hoc_times.py
@@ -1,238 +1,146 @@
-"""Tests that streaming end() honors caller-supplied ``started_at`` / ``ended_at``.
+"""Tests that streaming end() honors caller-supplied started_at / ended_at.
 
-Before this PR:
-- ``LLM.end()`` / ``SubAgent.end()`` / ``Turn.end()`` unconditionally clobbered
-  ``self.ended_at = now()``, so a caller-set value was lost.
-- All four span classes' ``_end_otel_span()`` calls omitted ``end_time_ns``,
-  so the emitted OTel span carried ``now()`` as its end time regardless of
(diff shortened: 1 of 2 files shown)


drtangible (approved):
👍 👍 👍

PR #7135, opened 2026-06-10
feat(weave_ts): add Google ADK integration core (WeaveAdkPlugin)

andrewtruong:
## Description

Adds tracing support for the Google Agent Development Kit for TypeScript ([`@google/adk`](https://github.com/google/adk-js)) to the weave TS SDK. **Stack PR 1/4** — core integration.

The integration is an ADK *plugin*: ADK dispatches plugin callbacks at its runner/model/tool lifecycle boundaries, and `WeaveAdkPlugin` mirrors those boundaries into weave calls via `saveCallStart`/`saveCallEnd` (the same transport as the OpenAI Agents integration):

- `beforeRun`/`afterRun` → `google.adk.invocation` root call (user message, app/session metadata, final response output)
- model callbacks → `google.adk.call_llm` calls with sanitized `LlmRequest` inputs, `LlmResponse` output, and per-model Gemini token usage; streaming partials are skipped so only the final aggregated response closes the call
- tool callbacks → `google.adk.execute_tool` calls keyed by `functionCallId`, with tool args/results and error capture
- per-agent `google.adk.invoke_agent` calls are synthesized lazily from `agentName` + the agent tree captured at run start, because `@google/adk` 1.2.0 never dispatches plugin agent callbacks (the `PluginManager` methods exist but have no call sites); the agent callbacks are still implemented so nesting tightens automatically once ADK wires them up

**Why a plugin rather than riding ADK's native OTel spans:** capturing those would require weave to register the process-global OTel TracerProvider and ContextManager, silently breaking any user observability stack registered later. Plugin callbacks fire regardless of OTel setup and hand us live request/response objects instead of JSON span attributes.

**Auto-instrumentation:** the existing CJS/ESM module hooks patch `Runner.prototype.runAsync`/`runEphemeral` to self-register a shared plugin instance idempotently (covers subclasses like `InMemoryRunner`). `WeaveAdkPlugin` is also exported for explicit use where module hooks cannot run (e.g. bundlers). ADK types are duck-typed locally (`googleAdk.types.ts`) so the SDK takes no dependency on `@google/adk`; every callback swallows its own errors because ADK's `PluginManager` rethrows plugin exceptions into the user's run.

## Testing

- cjs/esm typechecks pass at this level; integration tests, the `@google/genai` double-count exclusion, and a runnable example land in the rest of this stack (#7136, #7137, #7138).
- Verified end to end against prod Weave in both CJS (`require` hook) and ESM (`--import=weave/instrument`) hosts: trace structure, parenting, usage, and error capture read back from the server ([example trace project](https://wandb.ai/megatruong/google-adk-ts-test/weave)).

## Breaking changes

None — new integration surface only.

### DIFF
--- a/sdks/node/src/index.ts
+++ b/sdks/node/src/index.ts
@@ -18,6 +18,7 @@ export {GetCallsOptions} from './weaveClient';
 export {
   wrapOpenAI,
   wrapGoogleGenAI,
+  WeaveAdkPlugin,
   createOpenAIAgentsTracingProcessor,
   instrumentOpenAIAgents,
   patchRealtimeSession,
--- a/sdks/node/src/integrations/googleAdk.ts
+++ b/sdks/node/src/integrations/googleAdk.ts
@@ -1,0 +1,975 @@
+/**
+ * Weave integration for the Google Agent Development Kit for TypeScript
+ * (`@google/adk`).
+ *
+ * The integration is an ADK *plugin*: ADK dispatches plugin callbacks at its
+ * runner / model / tool lifecycle boundaries, and `WeaveAdkPlugin` mirrors
+ * those boundaries into Weave calls (`saveCallStart` / `saveCallEnd`), the
+ * same transport the OpenAI Agents integration uses.
+ *
+ * Why a plugin rather than ADK's OpenTelemetry spans: riding ADK's native
+ * OTel tracing would require Weave to register the process-global OTel
+ * TracerProvider and ContextManager, which would silently break any user
+ * observability stack registered later. Plugin callbacks fire regardless of
+ * OTel setup and hand us live `LlmRequest` / `LlmResponse` / tool objects
+ * instead of JSON-serialized span attributes.
+ *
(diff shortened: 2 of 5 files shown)


andrewtruong:
* **#7138** <a href="https://app.graphite.com/github/pr/wandb/weave/7138?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7137** <a href="https://app.graphite.com/github/pr/wandb/weave/7137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7136** <a href="https://app.graphite.com/github/pr/wandb/weave/7136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7135** <a href="https://app.graphite.com/github/pr/wandb/weave/7135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7135?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #7141, opened 2026-06-10
chore(weave): add claude-fable-5 to model providers and costs

nichochar:
## Description

Enables Anthropic's new Claude Fable 5 model (`claude-fable-5`, [announcement](https://www.anthropic.com/news/claude-fable-5-mythos-5)) in the trace server so it can be used from the model playground:

- `model_providers.json`: maps `claude-fable-5` → `anthropic` / `ANTHROPIC_API_KEY`
- `cost_checkpoint.json`: adds costs ($10/$50 per MTok input/output, cache read 10% of input, cache write 1.25x input, matching the pattern of other Anthropic entries)

Mirrors the prior model-add PR #7009 (opus 4.8). Frontend counterpart in wandb/core adds the model to the playground dropdown.

## Testing

Validated both JSON files parse; entries follow the same structure as existing Anthropic models.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,6 +73,11 @@ _Important:_ For OpenAI Codex agents (most likely you!), your environment does n
   - `weave/` - Python package implementation
   - `weave/trace_server` - Backend server implementation
 
+## Generated Files — Do Not Hand-Edit
+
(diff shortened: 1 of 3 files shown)


nichochar:

jwlee64 (changes requested):
This should be auto generated with make update_costs and make update_model_providers something like this
https://github.com/wandb/weave/compare/master...josiah/fable-5

jwlee64 (approved):

PR #7159, opened 2026-06-11
feat(weave_ts): add Google ADK plugin core (run/model/tool span tracing)

andrewtruong:
## Description

Adds tracing support for the Google Agent Development Kit for TypeScript
(@google/adk). The integration is an ADK plugin: ADK dispatches plugin
callbacks at its runner/model/tool lifecycle boundaries, and WeaveAdkPlugin
mirrors those boundaries into OpenTelemetry spans carrying the GenAI
semantic conventions, exported to Weave's agents endpoint
(/agents/otel/v1/traces) through the shared Weave tracer provider — the
same pipeline as the Pi coding-agent integration and the openai-agents OTel
processor, and the same conventions as the Python google_adk integration.
Runs land in the Weave Agents view.

- beforeRun/afterRun -> root `invoke_agent {agent}` span (gen_ai.agent.*,
  conversation id = ADK session id, user message in / final answer out as
  gen_ai.input/output.messages)
- model callbacks -> `chat {model}` spans with request parameters, system
  instructions, tool definitions, input/output messages in the GenAI parts
  model, token usage (input/output/total/reasoning/cache-read), response
  id/model/finish reasons; streaming partials are skipped so only the final
  aggregated response closes the span; model errors set span status ERROR +
  error.type
- tool callbacks -> `execute_tool {tool}` spans keyed by functionCallId
  (with synthetic FIFO keys when ADK omits one), recording
  gen_ai.tool.call.arguments/result and errors
- message content honours ADK's own opt-out
  (ADK_CAPTURE_MESSAGE_CONTENT_IN_SPANS=false) so regulated users get the
  same gate ADK gives them natively; provider name mirrors
  GOOGLE_GENAI_USE_VERTEXAI (gemini vs vertex_ai)
- dangling spans are force-ended (marked weave.google_adk.interrupted) at
  afterRun and on process beforeExit, registered before the tracer
  provider's own flush hook so late ends still export

In this slice every chat/tool span parents directly to the root span.
Per-agent invoke_agent nesting and module-hook auto-instrumentation land
next in this stack; until then the plugin is registered explicitly
(plugins: [new WeaveAdkPlugin()]).

Why a plugin rather than riding ADK's native OTel spans: capturing those
would require Weave to register the process-global OTel TracerProvider and
ContextManager, silently breaking any user observability stack registered
later. Plugin callbacks fire regardless of OTel setup and hand us live
request/response objects instead of JSON span attributes.

ADK types are duck-typed locally (googleAdk.types.ts) so the SDK takes no
dependency on @google/adk; every callback swallows its own errors because
ADK's PluginManager rethrows plugin exceptions into the user's run. Adds
the gen_ai.agent.description/id and gen_ai.tool.definitions semconv
constants.

## Testing

Jest integration tests run the real ADK runner (InMemoryRunner + a
scripted BaseLlm, no network) for the tool flow and model-error capture,
and drive the plugin callbacks directly for streaming partials, tool errors
plus the late afterToolCallback ADK always sends, dangling-span cleanup,
the message-content opt-out, and no-client safety — spans captured via an
InMemorySpanExporter injected through settings.genai.spanProcessor. Adds
@google/adk + lodash devDependencies and the lodash-es -> lodash jest
mapping (@google/adk's CJS build requires the ESM-only lodash-es; legal
under Node >= 22 require(esm), fatal under jest). cjs/esm typechecks pass.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -220,6 +220,22 @@ npm i
 npm run test
 ```
 
+### TypeScript SDK integration patterns (sdks/node)
+
+- Integrations live in `sdks/node/src/integrations/`.
+- Agent-framework integrations either write calls directly with
+  `client.saveCallStart` / `saveCallEnd` (the openai-agents default,
+  `openai-agents/weave-tracing-processor.ts`) or emit GenAI-semconv OTel
+  spans to the Weave agents endpoint (`/agents/otel/v1/traces`) via
+  `getWeaveTracer` (`googleAdk.ts`, `piCodingAgent.ts`, openai-agents under
+  `WEAVE_USE_OTEL_V2`) — never by wrapping framework functions in `op()`.
+- Framework types are duck-typed locally (e.g. `googleAdk.types.ts`) so the
+  SDK has no runtime dependency on the framework package. Keep these types
+  free of index signatures or the real framework types stop being assignable.
+- `@google/adk`'s CJS dist `require()`s the ESM-only `lodash-es` (legal under
+  Node >= 22 `require(esm)`, fatal under jest); the default jest project maps
+  `^lodash-es$` → `lodash`.
+
 ## Code Review & PR Guidelines
 
 ### PR Requirements
--- a/sdks/node/package.json
+++ b/sdks/node/package.json
@@ -119,7 +119,8 @@
         },
         "moduleNameMapper": {
           "^weave$": "<rootDir>/src/index.ts",
-          "^weave/(.*)$": "<rootDir>/src/$1"
+          "^weave/(.*)$": "<rootDir>/src/$1",
+          "^lodash-es$": "lodash"
         }
       }
     ],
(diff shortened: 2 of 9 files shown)


andrewtruong:
* **#7162** <a href="https://app.graphite.com/github/pr/wandb/weave/7162?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7161** <a href="https://app.graphite.com/github/pr/wandb/weave/7161?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7160** <a href="https://app.graphite.com/github/pr/wandb/weave/7160?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7159** <a href="https://app.graphite.com/github/pr/wandb/weave/7159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7159?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #7180, opened 2026-06-11
chore(weave_ts): Pull up module-level genai state

drtangible:
## Description

* We reference the "dual package hazard" (see: https://github.com/wandb/weave/pull/6682) a bunch of places in the codebase.

* Think our integrations would be a bit more readable if we consolidated some of this boilerplate and hoisted up module-level state. This'll also give us a clearer picture about _what_ global state we're relying on, rather than needing to jump through different call sites.

* This first change is just a POC of that -- starting with the state in the `genai/` stuff.

### DIFF
--- a/sdks/node/src/__tests__/genai/common.ts
+++ b/sdks/node/src/__tests__/genai/common.ts
@@ -5,12 +5,11 @@ import {
 } from '@opentelemetry/sdk-trace-base';
 
 import {setGlobalClient} from '../../clientApi';
-import {DEFAULT_STATE_SYMBOL_NAME} from '../../genai/context';
 import {Api as TraceServerApi} from '../../generated/traceServerApi';
-import {PROVIDER_HOLDER_SYMBOL_NAME} from '../../genai/provider';
 import {Settings, type SettingsInit} from '../../settings';
 import {WandbServerApi} from '../../wandb/wandbServerApi';
 import {WeaveClient} from '../../weaveClient';
+import state from 'weave/state';
 
 export const TEST_BASE_URL = 'http://localhost:8080';
 export const TEST_PROJECT = 'test-entity/test-project';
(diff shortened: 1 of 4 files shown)


drtangible:
* **#7182** <a href="https://app.graphite.com/github/pr/wandb/weave/7182?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7181** <a href="https://app.graphite.com/github/pr/wandb/weave/7181?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7180** <a href="https://app.graphite.com/github/pr/wandb/weave/7180?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7180?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

andrewtruong:
Does it have to be module state?  Is this better served with context?

drtangible:
> Does it have to be module state? Is this better served with context?

@andrewtruong -- I should note that a goal of this stack is to refactor existing behavior w/o changing it, but that's a good question.

Looking at the things we currently store as module-level state, it seems lilke that there might be a mix of things that are truly module-level state, and things that might be reasonably scoped to a AsyncLocalStorage context. For example, keeping track of the currently instrumented integrations, vs. genai state (already provided by `AsyncLocalStorage`) or the current provider / domain (not yet provided by `AsyncLocalStorage`). 

For the latter, I could see moving that, if we wanted to support calling `weave.init(...)` with different projects in the same application. Do we have customers doing that? In any case, I feel like centralizing the current module-level state sort of make it easier to have those conversations, so pulling it up feels like a win IMO.

Disclaimer: will follow up w/you offline to make sure I'm understanding the suggestion :)

andrewtruong:
Yeah I think the no-behaviour-change refactor is good, so I'm going to approve this stack.  The question is more generally whether we should have global state at all.

RE: ALS, I think that's specific to Node and wouldn't work if the user was using Bun or the browser.  Are there any alternatives?

andrewtruong (approved):

drtangible:
> RE: ALS, I think that's specific to Node and wouldn't work if the user was using Bun or the browser. Are there any alternatives?

I believe Bun supports it (https://bun.com/reference/node/async_hooks/AsyncLocalStorage), but no, wouldn't work in the browser.

Would have to better understand the use cases we'd be aiming to support, but in the browser, could store state wherever we see fit (in-memory, `sessionStorage`, `localStorage`, etc.), and access it by way of a key of whatever resource we're aiming to scope it to.

PR #7183, opened 2026-06-11
fix(weave): exclude op methods from LLMAsAJudgeScorer publish payload

ro31337:
## JIRA Issue(s)
[WB-35184](https://coreweave.atlassian.net/browse/WB-35184)

## Description
Creating an `LLMAsAJudgeScorer` in the SDK and publishing it to attach to an online-scoring monitor produced a payload the scoring worker rejects, so the monitor silently never scored — even though building the same scorer through the UI worked. On publish, the client serializes each `@op` method as a ref that deserializes to a `CustomWeaveType(Op)`, and the worker's safety check fails closed on those. A published judge scorer carried three of them: its own `score`, the inherited `summarize`, and — because the nested `model` is published as its own object — `LLMStructuredCompletionModel.predict`.

This sets `_weave_exclude_ops_from_record = True` on both `LLMAsAJudgeScorer` and `LLMStructuredCompletionModel`, the same opt-out `RemoteScorer` already uses (#7036). The op methods still run and trace at runtime; only their stored ref — which nothing reads — is dropped, so the published shape now matches what the UI already persists. It is SDK-only: a `ClassVar` is not a pydantic field, so there is no schema/zod change and no core-side dependency.

## Testing
Added a regression test that `pydantic_object_record` drops `score`/`summarize`/`predict` and that a published scorer round-trips with no op refs in the stored payload. Regenerated the one serialization fixture this affects and ran the serialization suite on Python 3.13 (the version that produces it) and 3.12, plus the judge-scorer, structured-model, evaluation-ref-get, and remote-scorer suites — all green.

[WB-35184]: https://coreweave.atlassian.net/browse/WB-35184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace/data_serialization/test_cases/library_cases.py
+++ b/tests/trace/data_serialization/test_cases/library_cases.py
@@ -277,6 +277,219 @@ library_cases = [
         python_version_code_capture=(3, 13),
     ),
     SerializationTestCase(
+        id="Library Objects - Scorer, Evaluation, Dataset, LLMAsAJudgeScorer, LLMStructuredCompletionModel (legacy v6)",
+        runtime_object_factory=make_evaluation,
+        inline_call_param=False,
+        is_legacy=True,
+        exp_json={
+            "_type": "Evaluation",
+            "name": None,
+            "description": None,
+            "dataset": "weave:///shawn/test-project/object/Dataset:YLYVrBqCtlMOa770T1oPssqYnf9rgqdnY5hVCwRcrm8",
+            "scorers": [
+                "weave:///shawn/test-project/object/MyScorer:erh2OhYuvmiYF5MAHd2iNbtJkiAfvkHYn5l78CV6XrU",
+                "weave:///shawn/test-project/object/LLMAsAJudgeScorer:usU9eU7is5YeNlwmYcSOHYfjJB8xHGCXXUVpm6dBbfc",
+            ],
+            "preprocess_model_input": None,
+            "trials": 1,
+            "metadata": None,
+            "evaluation_name": None,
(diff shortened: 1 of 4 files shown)


tests/trace/data_serialization/test_cases/library_cases.py
mscavezze-cw, line 78: It looks like this PR updates the test case instead of creating a new one. I don't think that matches the original design of these tests.

From test_serialization_correctness.py:22:
```
"""
IMPORTANT RULES: Once a SerializationTestCase is created, it should never be modified.
As the code base evolves, it is expected that some of these test cases will break (since
the serialization format changes, op code changes, etc...). In such cases:
1. Copy the failing test case to a new test case.
2. Set the is_legacy flag to True on the new test case.
3. Rerun the test: this should PASS. If it does not, then it means you have made a
backwards incompatible change and data written by older clients will not be able to
be deserialized by newer clients.
4. Now you can modify the original test case to pass.

This methodology allows us to lock in the legacy serialization formats as a contact,
independent of the actual code that is used to serialize the data.
"""
```

mscavezze-cw: Also, the change to this file in this PR but not the one for RemoteScorer demonstrates that we have a hole in our test coverage. I've filed a new ticket. https://coreweave.atlassian.net/browse/WB-35583

mscavezze-cw (approved):

PR #7184, opened 2026-06-11
feat(weave): add wandb.agent_user_feedback feedback type

nikumar1206:
## What

Backend support for human reactions on agents.

- New `wandb.agent_user_feedback` feedback type for human-applied tags (vs scorer-applied `wandb.agent_monitor`). The value is stored as a tag in `scorer_tags`.
- The chat view now exposes each turn/message's agent name, version, and status, so the UI can attach them to a reaction (used for dashboard filtering).

Frontend: wandb/core#45619.

## Testing
- [x] Manual
- [x] Unit
- [x] QA

### DIFF
--- a/weave/trace_server/agents/chat_view.py
+++ b/weave/trace_server/agents/chat_view.py
@@ -151,12 +151,18 @@ def build_trace_chat(
     messages = build_chat_messages(spans)
 
     root_span_name: str | None = None
+    root_agent_name: str | None = None
+    root_agent_version: str | None = None
+    root_status_code: str | None = None
     provider: str | None = None
     total_duration_ms: int | None = None
 
     if spans:
         root = _select_root_span(spans)
         root_span_name = root.agent_name or root.span_name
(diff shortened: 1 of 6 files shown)


ro31337 (approved):

weave/trace_server/agents/chat_view.py
```diff
@@ -151,12 +151,18 @@ def build_trace_chat(
     messages = build_chat_messages(spans)
 
     root_span_name: str | None = None
+    root_agent_name: str | None = None
+    root_agent_version: str | None = None
+    root_status_code: str | None = None
```
jtschoonhoven, line 156: The `status_code` is declared an Enum8 in ClickHouse, with possible values "OK", "ERROR", or "UNSET". You should update the types throughout this PR to match.

jtschoonhoven (approved):
Make sure you type `status_code` correctly (see comment). Otherwise LGTM.

PR #7186, opened 2026-06-11
feat(weave): allow unbounded range for ungrouped agent spans/stats

nikumar1206:
## Summary
`MAX_AGENT_STATS_RANGE_DAYS` (31d) caps the `/agents/spans/stats` time range. That cap exists to guard expensive grouped/bucketed/filtered aggregations — but it also blocks a plain ungrouped rollup (e.g. an all-time span count).

This lifts the cap **only** for a single-series request — no `group_by`, no `group_filters`, no numeric bucket. Grouped/bucketed/filtered requests are still capped.

No request/response shape or query-builder changes (per review): clients get an all-time scalar by requesting the whole range as one bucket — `start=epoch`, `end=now`, `granularity=now-epoch` — which the existing time-bucket path already collapses to a single row.

## Testing
- [x] Manual tests
- [x] Unit tests
- [x] QA tests

### DIFF
--- a/weave/trace_server/agents/types.py
+++ b/weave/trace_server/agents/types.py
@@ -373,8 +373,15 @@ class AgentSpanStatsReq(BaseModel):
         end = self.end or datetime.datetime.now(datetime.timezone.utc)
         if end < self.start:
             raise ValueError("AgentSpanStatsReq end must be after start")
+        # Unfiltered, ungrouped requests are not bound by
+        # MAX_AGENT_STATS_RANGE_DAYS.
+        apply_max_range_days = (
+            bool(self.group_by)
+            or bool(self.group_filters)
+            or numeric_bucket is not None
+        )
         max_range = datetime.timedelta(days=MAX_AGENT_STATS_RANGE_DAYS)
-        if end - self.start > max_range:
+        if apply_max_range_days and end - self.start > max_range:
             raise ValueError(
                 "AgentSpanStatsReq date range cannot exceed "
                 f"{MAX_AGENT_STATS_RANGE_DAYS} days"
--- a/tests/trace_server/query_builder/test_agent_stats_query_builder.py
+++ b/tests/trace_server/query_builder/test_agent_stats_query_builder.py
@@ -822,8 +822,18 @@ def test_request_validation_rejects_numeric_bucket_group_by() -> None:
         )
 
 
-def test_request_validation_rejects_large_range() -> None:
+def test_request_validation_rejects_large_range_when_grouped() -> None:
     with pytest.raises(ValidationError):
         _req(
             end=datetime.datetime(2026, 2, 15, tzinfo=datetime.timezone.utc),
+            group_by=[
+                AgentGroupByRef(source="column", key="agent_name", alias="agent")
+            ],
         )
+
+
+def test_request_validation_allows_large_range_when_ungrouped() -> None:
+    # No group_by / group_filters / numeric bucket: a single-series rollup may
+    # span any range (used for all-time totals), so the cap does not apply.
+    req = _req(end=datetime.datetime(2026, 2, 15, tzinfo=datetime.timezone.utc))
+    assert req.end == datetime.datetime(2026, 2, 15, tzinfo=datetime.timezone.utc)
--- a/tests/trace_server/test_genai_agent_queries.py
+++ b/tests/trace_server/test_genai_agent_queries.py
@@ -899,6 +899,49 @@ def test_agent_span_stats_ungrouped_metrics(ch_server):
     assert row["count_true_invocations"] == 1
 
 
+def test_agent_span_stats_ungrouped_all_time(ch_server):
+    """Ungrouped >31-day request with the whole range as one bucket returns a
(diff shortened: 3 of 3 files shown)


nikumar1206:
* **#7186** <a href="https://app.graphite.com/github/pr/wandb/weave/7186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7186?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/query_builder/agent_stats_query_builder.py
```diff
@@ -317,9 +321,108 @@ def build_agent_span_stats_query(
...
+    SELECT
+      {outer_metric_sql}
+    FROM {_CTE_AGGREGATED_DATA}
+    """
+
+
 def _resolve_time_bounds(
     req: AgentSpanStatsReq,
 ) -> tuple[datetime.datetime, datetime.datetime]:
+    assert req.start is not None
```
jtschoonhoven: Never `assert` in production code, raise ValueError. Even better: move this to a custom validator in the `AgentSpanStatsReq`.

weave/trace_server/query_builder/agent_stats_query_builder.py
```diff
@@ -317,9 +321,108 @@ def build_agent_span_stats_query(
...
+
+
+def _build_scalar_stats_query(
+    *,
+    source_filter: _SpanSourceFilterSQL,
+    metric_exprs: list[str],
+    agg_selects: list[str],
+    outer_metric_selects: list[str],
+) -> str:
+    """Render a single aggregate row over the filtered span set (no buckets)."""
```
jtschoonhoven: Tiny nit: i dunno if `_build_scalar_stats_query` is the right name... what makes this "scalar"? That it returns one row instead of several? 🤔

weave/trace_server/agents/types.py
```diff
@@ -400,11 +435,11 @@ def validate_stats_request(self) -> AgentSpanStatsReq:
...
     """Response containing chart-ready agent span stats rows."""
 
-    start: datetime.datetime
-    end: datetime.datetime
+    start: datetime.datetime | None = None
+    end: datetime.datetime | None = None
     granularity: int | None = None
     timezone: str
-    bucket_type: Literal["time", "number"] = "time"
+    bucket_type: Literal["time", "number", "none"] = "time"
```
jtschoonhoven: Just using `None` (null) instead of `"none"` seems more idiomatic.

```suggestion
    bucket_type: Literal["time", "number"] | None = "time"
```

weave/trace_server/agents/types.py
```diff
@@ -325,7 +336,9 @@ def validate_numeric_bucket_spec(self) -> AgentSpanStatsNumericBucketSpec:
 
 
 AgentSpanStatsBucketSpec = Annotated[
-    AgentSpanStatsTimeBucketSpec | AgentSpanStatsNumericBucketSpec,
+    AgentSpanStatsTimeBucketSpec
+    | AgentSpanStatsNumericBucketSpec
+    | AgentSpanStatsNoneBucketSpec,
     Field(discriminator="type"),
 ]
```
jtschoonhoven, line 330: Nit: I think we should avoid polymorphic response types. It makes our APIs way harder to use when you can't rely on it to return a fixed shape.

That ship has already sailed here, but can we avoid adding _another_ type here? What if you just reused `AgentSpanStatsTimeBucketSpec` for a very wide time range?

If we really need to return a different response type _and_ write custom sql, I think that's a sign that we really ought to just make a separate endpoint.

jtschoonhoven (commented):
If you disagree I could be persuaded otherwise, but I think it would be better and much more simple if you didn't modify the request, response, or query builder _at all_.

To get all-time stats, the client should use the endpoint as-is and just set `start=EPOCH` and `end=NOW` and `granularity=NOW-EPOCH`. Then the endpoint returns a single-row response.

Then the only change required in this PR is to bypass the MAX_DAYS check if the user isn't adding any extra groups or filters.

WDYT?

nikumar1206:
updated to reflect that. my only concern is the query will be slower, but if we find that its blocking loading up pages on the UI, we can always update it in the future to use a different query.

weave/trace_server/agents/types.py
```diff
@@ -373,8 +373,15 @@ def validate_stats_request(self) -> AgentSpanStatsReq:
         end = self.end or datetime.datetime.now(datetime.timezone.utc)
         if end < self.start:
             raise ValueError("AgentSpanStatsReq end must be after start")
+        # Unfiltered, ungrouped requests are not bound by
+        # MAX_AGENT_STATS_RANGE_DAYS.
+        capped = (
```
jtschoonhoven: Tiny nit: I don't find "capped" to be very descriptive. Consider renaming this "apply_max_range_days" or something

jtschoonhoven (approved):

PR #7191, opened 2026-06-11
chore(weave): validate monitor query fields on save

gtarpenning:
## Summary
- Validate `Monitor.query` field refs against `ALLOWED_CALL_FIELDS` at object-create time; reject unknown fields (e.g. `operation_name`) with the allowed list in the error.
- Prevents the silent per-cycle InvalidFieldError storm in the scoring worker (a bad monitor query fails sampling forever today with no user signal).
- Validation lives in its own `monitor_query_validation.py` (keeps `calls_query_builder.py` untouched).

## Testing
functional obj_create tests: bad field rejected with allowed list + not stored; malformed query rejected; static/dynamic/absent queries accepted.

### DIFF
--- a/weave/trace_server/calls_query_builder/calls_query_builder.py
+++ b/weave/trace_server/calls_query_builder/calls_query_builder.py
@@ -31,7 +31,7 @@ from collections.abc import Callable, KeysView, Sequence
 from dataclasses import dataclass
 from typing import Any, Literal, cast
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, ValidationError
 from typing_extensions import Self
 
 from weave.shared.trace_server_interface_util import (
@@ -64,7 +64,7 @@ from weave.trace_server.calls_query_builder.utils import (
     trace_id_index_expr,
 )
 from weave.trace_server.common_interface import SortBy
-from weave.trace_server.errors import InvalidFieldError
+from weave.trace_server.errors import InvalidFieldError, InvalidRequest
 from weave.trace_server.interface import query as tsi_query
 from weave.trace_server.interface.feedback_types import MULTI_VALUE_FEEDBACK_TYPES
 from weave.trace_server.interface.query import (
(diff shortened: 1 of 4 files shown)


weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1943,6 +1943,94 @@ def get_field_by_name(name: str) -> CallsMergedField:
     return ALLOWED_CALL_FIELDS[name]
 
 
+# Field references get_field_by_name accepts beyond exact ALLOWED_CALL_FIELDS
```
mscavezze-cw: It's hard to know what these constants mean until reading the rest of the code and using claude to analyze the PR. I think these would be more self-documenting if a couple of changes were made:

- It looks like they're mainly for creating the error message, but they're up above the validation code. Consider moving them to be closer to the error message and/or giving a little high-level context before launching into the details.
- Consider moving the error messaging and maybe the validation into a separate file. This file is huge, and adding more new code to it isn't going to help readability or agent context limits.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1943,6 +1943,94 @@ def get_field_by_name(name: str) -> CallsMergedField:
...
+# purpose: the trace server must not import from weave/flow.
+MONITOR_OBJECT_CLASSES = frozenset({"Monitor", "ClassifierMonitor"})
+
+
+def validate_monitor_query_fields(
+    base_object_class: str | None,
+    leaf_object_class: str | None,
+    val: object,
+) -> None:
+    """Reject a Monitor whose `query` references a field outside the allowed set."""
```
mscavezze-cw: The comment and implementation don't quite match. While it is true that this will detect a query with unacceptable fields, I think it will also detect queries that don't compile. At least that's how the code reads. That also seems reasonable, but the mismatch between the comment and code is confusing.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1943,6 +1943,94 @@ def get_field_by_name(name: str) -> CallsMergedField:
...
+    if (
```
mscavezze-cw: Half of this function is answering the question "is this a monitor that we should validate". That's not immediately obvious. Consider ways to make the code more self-documenting.

weave/trace_server/calls_query_builder/calls_query_builder.py
```diff
@@ -1943,6 +1943,94 @@ def get_field_by_name(name: str) -> CallsMergedField:
...
+        query = tsi_query.Query.model_validate(cleaned)
+    except ValidationError:
+        # Not a recognizable query (e.g. a user object that happens to share
+        # the Monitor class name); leave the write untouched.
+        return
+    validate_query_compiles(query)
+
+
+def validate_query_compiles(query: tsi_query.Query) -> None:
+    """Validate that `query` references only allowed call fields and is well-formed."""
```
mscavezze-cw: Similar comment about mismatch between comment and what the code appears to be doing. Consider connecting the dots here.

mscavezze-cw (approved):
This looks functional, but was difficult to read without claude. I offered some suggestions for this specific PR. There are probably things that could be done to the rest of this file, but that's out of scope for this PR and review.

PR #7193, opened 2026-06-11
chore(weave): Env var for Remote Scoring result schema

mscavezze-cw:
## Description

Add env var to allow Remote Scoring to enforce a schema for results

## Testing

Unit testing


weave/trace_server/environment.py
```diff
@@ -223,6 +226,21 @@ def wf_scoring_worker_remote_scorer_allow_insecure_http() -> bool:
     )
 
 
+def wf_scoring_worker_remote_scorer_require_structured_result_schema() -> bool:
+    """Whether remote scorer results must match the structured scorer schema.
+
+    Defaults to true to nudge new RemoteScorer users onto the typed scorer
+    result shape. Set to ``false`` only for controlled compatibility testing or
```
jtschoonhoven: Tiny nit: there's no reason to use \``double-ticks`` and they look weird to me

jtschoonhoven (approved):
Code looks good. I'll defer to you on whether this is actually a necessary precaution.

weave/trace_server/environment.py
mscavezze-cw: Yeah, they annoy me too but I've gotten tired of telling agents to take them out. I'll try again - the memory feature seems better lately.

PR #7194, opened 2026-06-11
feat(weave_ts): add Turn.setAttributes (plural), deprecate setAttribute

rgao-coreweave:
## Description

Aligns `Turn`'s attribute API with the landed Python SDK (#7131), which exposes `set_attributes(dict)` on the shared span mixin.

- **Adds `Turn.setAttributes(attributes)`** (plural, object arg) — "pass an object whether you have one key or many." Mirrors the Python `set_attributes` and OTel JS's `Span.setAttributes`.
- **Keeps `Turn.setAttribute(key, value)` as a `@deprecated` alias** that delegates to `setAttributes({[key]: value})`. `Turn.setAttribute` already shipped on `master` and has live consumers (e.g. `weave-openclaw`, ~8 call sites), so this is intentionally **non-breaking** — existing callers keep working, with the doc-comment steering new code to the plural form.

This completes a coordinated plural rename across the GenAI emitter family:

| Emitter | PR | Surface |
|---|---|---|
| Tool | #7076 | `setAttributes` only (never released singular) |
| LLM | #7078 | `setAttributes` only |
| SubAgent | #7077 | `setAttributes` only |
| Turn | this PR | `setAttributes` + `@deprecated setAttribute` alias |

Only `Turn` carries the back-compat alias because it's the only one whose singular form had shipped; adding a "deprecated-from-birth" alias to the others would be dead surface.

## Testing

`cd sdks/node && npx jest src/__tests__/genai/turn.test.ts` — 10 passing. Coverage:
- `setAttributes` writes multiple keys, returns `this`, no-ops after `end()`.
- The deprecated `setAttribute` alias still lands its attribute and chains.

### DIFF
--- a/sdks/node/src/__tests__/genai/turn.test.ts
+++ b/sdks/node/src/__tests__/genai/turn.test.ts
@@ -52,34 +52,44 @@ describe('Turn', () => {
   });
 });
 
-describe('Turn.setAttribute', () => {
+describe('Turn.setAttributes', () => {
   setupGenAITestEnvironment();
   const getExporter = setupExporterPerTest();
 
-  it('writes an attribute to the underlying span', () => {
+  it('writes attributes to the underlying span', () => {
     const turn = Turn.create({});
-    turn.setAttribute('weave.cost.usd', 0.42);
+    turn.setAttributes({'weave.cost.usd': 0.42, 'weave.tag': 'enterprise'});
     turn.end();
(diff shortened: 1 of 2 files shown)


rgao-coreweave:
Folded into #7076 (cherry-picked the Turn `setAttributes` + deprecated `setAttribute` alias commit). Closing this standalone PR.

PR #7195, opened 2026-06-11
feat(weave_ts): add setAttributes/addEvent to Tool/LLM/SubAgent/Turn via shared base

rgao-coreweave:
## Description

Mirrors the landed Python SDK (#7131), which added `set_attributes` and `add_event` to `Tool`/`LLM`/`SubAgent`/`Turn` via a shared `_SpanBase` mixin. This is the TypeScript equivalent: a single `SpanBase` that all four GenAI span wrappers extend, so the escape-hatch surface is defined once and applied uniformly. (`Session` is excluded — same as Python, where `Session` does not extend `_SpanBase`.)

### `SpanBase` (`sdks/node/src/genai/spanBase.ts`)
Holds the underlying OTel span + the `_ended` guard and exposes:

- **`setAttributes(attributes)`** — stamp multiple attributes at once. Pass an object whether you have one key or many. Mirrors OTel `Span.setAttributes` and the Python SDK's `set_attributes(dict)`.
- **`addEvent(name, attributes?, startTime?)`** — record a span event (e.g. `weave.permission_request`, context compaction).

Both are chainable (`return this`). Mutating after `end()` **warns** (`weave.<Emitter>.<method>() called after end() — …`) and no-ops, matching Python #7131 (the warning names the concrete emitter via the runtime constructor name).

`Tool`, `LLM`, `SubAgent`, and `Turn` now `extend SpanBase`; LLM's private `_warnIfEnded` is dropped in favor of the inherited one (identical message).

### Deprecated singular alias (Turn only)
`Turn.setAttribute(key, value)` shipped previously and has a live consumer (`weave-openclaw`), so it's retained as a thin **`@deprecated`** alias delegating to `setAttributes`. The other three emitters never shipped a singular form, so they're plural-only — no dead surface.

## Testing

`cd sdks/node && npx jest src/__tests__/genai` — **58 passing** across 8 suites. Behavior is tested **per-emitter** (in each emitter's own suite, oblivious to the base class so it can be refactored freely): `setAttributes`/`addEvent` land on the exported span, return `this`, and warn + no-op after `end()`. Turn additionally covers the deprecated `setAttribute` alias.

## Supersedes
Consolidates the per-emitter attribute/event work that was split across #7076 (closed) / #7077 / #7078 into one PR, matching #7131's single-PR shape. #7077 and #7078 are reduced to their non-attribute parts (SubAgent nesting, LLM post-hoc times).


sdks/node/src/genai/spanBase.ts
```diff
@@ -0,0 +1,66 @@
...
+   * usage). No-op after `end()`. Mirrors OTel `Span.setAttribute`.
+   *
+   * @example
+   * span.setAttribute('weave.cost.usd', 0.42);
+   */
+  setAttribute(key: string, value: AttributeValue): this {
+    if (this._ended) return this;
+    this.span.setAttribute(key, value);
+    return this;
+  }
```
drtangible: Should we leave this deprecated and only on `Turn`?

drtangible (approved):
👍 👍 👍

sdks/node/src/__tests__/genai/spanBase.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  {
+    label: 'Turn',
+    start: () => {
+      const turn = Turn.create({});
+      return {
+        target: turn,
+        endAll: () => turn.end(),
+        locate: spans => findSpan(spans, 'invoke_agent'),
+      };
+    },
```
drtangible: nit: I think it'd still make sense to test behavior for all of these in their own test suites. The fact that this behavior is provided by way of a base class is an implementation detail that the test should remain oblivious to. This'd allow us to refactor implementation details as we see fit.

rgao-coreweave: added 4 tests. I think that makes sense as well!

rgao-coreweave: Nice, future-proofing!

sdks/node/src/genai/spanBase.ts
rgao-coreweave: Moved to turn only and deprecated setAttribute.

sdks/node/src/__tests__/genai/spanBase.test.ts
drtangible: 🔥 🔥 🔥

PR #7196, opened 2026-06-11
feat(weave): add OTel tracing primitives to weave.trace_server (PR-7a of ddtrace migration)

amwarrier:
## Summary

First PR of the `weave/trace_server/` ddtrace → OpenTelemetry migration. Adds the abstraction layer (`@traced`, `@traced_generator`) that subsequent PRs will use to replace `@ddtrace.tracer.wrap` decorators across 12 files (66 wraps total).

**This PR is intentionally scope-limited to the decorator surface + tests. No call sites are migrated.** Call-site migration follows:

- PR-7b1 — `clickhouse_trace_server_batched.py` (47 of 66 wraps) — draft opened immediately after this PR.
- PR-7b2 — remaining 11 files (`kafka.py`, `usage_utils.py`, `ttl_settings.py`, etc).
- PR-7c — `datadog.py` rewrite + remove `ddtrace>=2.7.0` from `pyproject.toml`.

## Why a new module

`weave.trace_server.datadog` keeps its existing public symbols (`record_db_insert`, `set_root_span_dd_tags`, `db_insert_path`, etc.). Only the *tracing* surface — `@traced` and the generator variant — moves to the new module. This keeps the PR-7a diff scoped to additions, and the module name is honest: the SDK is OpenTelemetry, DD is just the backend.

## Public surface

```python
from weave.trace_server.tracing import traced, traced_generator

@traced(name="my_op")
def my_op(x: int) -> int:
    ...

@traced(name="my_async_op")
async def my_async_op(x: int) -> int:
    ...

@traced_generator(name="stream_calls")
def stream_calls(limit: int) -> Iterator[dict]:
    yield from ...
```

## Why no try/except in the decorator body

OTel's `start_as_current_span` context manager already:
- Records the exception as a span event on `__exit__`
- Sets `Status(ERROR, "{ExcType}: {msg}")` for `Exception` subclasses
- Lets `BaseException` (GeneratorExit, CancelledError, KeyboardInterrupt, SystemExit) pass through without marking the span errored

Reference: opentelemetry-python's `use_span` helper catches `except Exception`, not `BaseException`, per [open-telemetry/opentelemetry-python#4484](https://github.com/open-telemetry/opentelemetry-python/issues/4484).

This also means the historical `generator_trace` wrapper in `datadog.py` (which existed to suppress ddtrace's auto-mark-error-on-GeneratorExit) is removable in PR-7c.

## Why `_tracer = trace.get_tracer(...)` is hoisted to module scope

Measured: 16.5μs/call with `get_tracer(...)` inside the wrapper vs 9.95μs/call with it hoisted (40% per-call overhead saved). These decorators run on hot CH-query paths with dozens of spans per request, so the hoist matters.

`get_tracer` returns a `ProxyTracer` that caches its real-tracer reference on first span open and never re-resolves. To preserve test isolation when the global `TracerProvider` is swapped between cases, a `reset_tracer_cache()` helper is exported. Production code never needs to call it — there is exactly one `TracerProvider` per process, set during application startup before any span opens.

## Test plan

- [ ] All 12 unit tests pass (validated locally against opentelemetry-sdk 1.28.0+):
  - Sync function: span name, error marking, `functools.wraps` metadata preservation
  - Async function: span name, error marking, `CancelledError` pass-through
  - Generator/async-generator refusal at decoration time (clear TypeError)
  - `@traced_generator`: span covers full iteration, `GeneratorExit` pass-through, exception marking
  - Nested decorator calls produce correct parent/child span linkage
- [ ] No regressions in the existing `tests/trace_server/` suite (no call sites migrated yet)
- [ ] Tested in the wandb/core image via submodule bump on wandbench-small before merging

## Out of scope

- No `traced_with_service` (per-span service override) — none of upstream's existing wraps use `@ddtrace.tracer.wrap(service="X")`, so adding it would be speculative. Will add in a future PR if a call site needs it.
- Phase A in wandb/core (`services/weave-trace/src/tracing/decorator.py`) has its own equivalent decorator. Once this lands and the upstream snapshot advances in wandb/core, that file becomes deletable — tracked separately.

### DIFF
--- a/tests/trace_server/test_tracing.py
+++ b/tests/trace_server/test_tracing.py
@@ -1,0 +1,275 @@
+"""Unit tests for `weave.trace_server.tracing`.
+
+These tests use an isolated `TracerProvider` with an in-memory exporter so
+they don't depend on (or interfere with) any process-global tracer state.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Generator, Iterator
+from typing import Any
+
+import pytest
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
+    InMemorySpanExporter,
+)
+from opentelemetry.trace.status import StatusCode
+
+from weave.trace_server.tracing import reset_tracer_cache, traced, traced_generator
+
+
+@pytest.fixture
+def exporter() -> Generator[InMemorySpanExporter, None, None]:
+    """Install an isolated TracerProvider + in-memory exporter for the test.
+
+    Swaps the global TracerProvider for the duration of the test, then drops
+    the module-level cached ProxyTracer reference via `reset_tracer_cache()`
+    so the next span open re-resolves against our new provider.
+    """
+    previous = trace._TRACER_PROVIDER  # noqa: SLF001
+    provider = TracerProvider()
+    exp = InMemorySpanExporter()
+    provider.add_span_processor(SimpleSpanProcessor(exp))
+    trace._TRACER_PROVIDER = provider  # noqa: SLF001
+    reset_tracer_cache()
+    try:
+        yield exp
+    finally:
+        trace._TRACER_PROVIDER = previous  # noqa: SLF001
+        reset_tracer_cache()
+
+
+def _single_span(exporter: InMemorySpanExporter) -> Any:
+    spans = exporter.get_finished_spans()
+    assert len(spans) == 1, f"expected one span, got {len(spans)}: {spans!r}"
+    return spans[0]
+
+
+# ---------------------------------------------------------------------------
+# @traced — sync function shape
+# ---------------------------------------------------------------------------
+
+
+def test_sync_function_span_name(exporter: InMemorySpanExporter) -> None:
+    @traced(name="my_sync_op")
+    def my_sync_op(x: int) -> int:
+        return x * 2
+
+    assert my_sync_op(3) == 6
+    span = _single_span(exporter)
+    assert span.name == "my_sync_op"
+    assert span.status.status_code == StatusCode.UNSET
+
+
+def test_sync_function_marks_error_and_reraises(
+    exporter: InMemorySpanExporter,
+) -> None:
+    @traced(name="raises_op")
+    def raises_op() -> None:
+        raise RuntimeError("boom")
+
+    with pytest.raises(RuntimeError, match="boom"):
+        raises_op()
+
+    span = _single_span(exporter)
+    assert span.name == "raises_op"
+    assert span.status.status_code == StatusCode.ERROR
+    # OTel's start_as_current_span sets description to "{ExcType}: {msg}".
+    assert span.status.description == "RuntimeError: boom"
+    # An exception event should be recorded on the span.
+    event_names = [e.name for e in span.events]
+    assert "exception" in event_names
+
+
+def test_sync_function_preserves_metadata(
+    exporter: InMemorySpanExporter,  # noqa: ARG001
+) -> None:
+    """`functools.wraps` should preserve __name__, __doc__, __module__."""
+
+    @traced(name="span_name_unrelated_to_fn_name")
+    def some_fn(x: int) -> int:
+        """Docstring."""
+        return x
+
+    assert some_fn.__name__ == "some_fn"
+    assert some_fn.__doc__ == "Docstring."
+    assert some_fn.__wrapped__(5) == 5  # type: ignore[attr-defined]
+
+
+# ---------------------------------------------------------------------------
+# @traced — async function shape
+# ---------------------------------------------------------------------------
+
+
+def test_async_function_span_name(exporter: InMemorySpanExporter) -> None:
+    @traced(name="my_async_op")
+    async def my_async_op(x: int) -> int:
+        return x * 2
+
+    assert asyncio.run(my_async_op(3)) == 6
+    span = _single_span(exporter)
+    assert span.name == "my_async_op"
+    assert span.status.status_code == StatusCode.UNSET
+
+
+def test_async_function_marks_error_and_reraises(
+    exporter: InMemorySpanExporter,
+) -> None:
+    @traced(name="async_raises_op")
+    async def async_raises_op() -> None:
+        raise ValueError("async-boom")
+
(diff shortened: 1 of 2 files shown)


weave/trace_server/tracing.py
```diff
@@ -0,0 +1,178 @@
...
+        _tracer._real_tracer = None  # type: ignore[attr-defined]  # noqa: SLF001
+    except AttributeError:
+        pass
+
+
+def traced(name: str) -> Callable[[Callable[..., T]], Callable[..., T]]:
+    """Wrap a function in an OTel span named `name`.
+
+    Args:
+        name: The span name. This becomes the DD APM `resource_name` via DD's
```
gtarpenning, line 90: claude thought this was a red flag: L19 says dashboards key on this name 1:1; here it's `resource_name`. ddtrace `name` = *operation* name, but OTel->DD maps span name -> `resource_name` and synthesizes operation name from the scope/kind. So monitors filtered on operation_name go dark on migration. Is that true?

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,178 @@
...
+    Only intended for tests that swap the global TracerProvider mid-process.
+    Production code should never need to call this — there is exactly one
+    TracerProvider per process and it is set during application startup
+    before any span is opened.
+    """
+    # `_real_tracer` is a private attr on `opentelemetry.trace.ProxyTracer`.
+    # If a future OTel release renames it, we'd rather tests silently lose
+    # isolation than crash with AttributeError mid-suite.
+    try:
+        _tracer._real_tracer = None  # type: ignore[attr-defined]  # noqa: SLF001
```
gtarpenning, line 81: drop the `# type: ignore` (hard rule). this whole private-attr poke is test-only -- it belongs in a conftest fixture, not the prod module's public surface.

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,178 @@
...
+    except AttributeError:
+        pass
```
gtarpenning, line 83: maybe at least an error log here? silent isolation-loss means the suite stays green for the wrong reason if OTel ever renames `_real_tracer`.

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,178 @@
...
+            )
+
+        if asyncio.iscoroutinefunction(fn):
+
+            @wraps(fn)
+            async def awrap(*args: Any, **kwargs: Any) -> Any:
+                with _tracer.start_as_current_span(name):
+                    return await fn(*args, **kwargs)
+
+            return awrap  # type: ignore[return-value]
```
gtarpenning, line 133: kill both `# type: ignore[return-value]` (here + L140) with `@overload` on `traced` -- one for async, one for sync.

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,178 @@
...
+        ... def stream_calls(limit: int) -> Iterator[dict]:
+        ...     for i in range(limit):
+        ...         yield {"id": i}
+    """
+
+    def decorator(
+        fn: Callable[..., Iterator[Any]],
+    ) -> Callable[..., Generator[Any, None, None]]:
+        @wraps(fn)
+        def wrapper(*args: Any, **kwargs: Any) -> Generator[Any, None, None]:
```
gtarpenning, line 172: no guard: decorating a non-generator silently `yield from`s its return (dict -> keys; scalar -> opaque TypeError at call time). we should refuse non-generators loudly at decoration time I think

tests/trace_server/test_tracing.py
```diff
@@ -0,0 +1,275 @@
...
+
+    Swaps the global TracerProvider for the duration of the test, then drops
+    the module-level cached ProxyTracer reference via `reset_tracer_cache()`
+    so the next span open re-resolves against our new provider.
+    """
+    previous = trace._TRACER_PROVIDER  # noqa: SLF001
+    provider = TracerProvider()
+    exp = InMemorySpanExporter()
+    provider.add_span_processor(SimpleSpanProcessor(exp))
+    trace._TRACER_PROVIDER = provider  # noqa: SLF001
```
gtarpenning, line 37: i think this is a bit of a smell, forces `reset_tracer_cache` to exist. can we scope this to a conftest fixture?

tests/trace_server/test_tracing.py
```diff
@@ -0,0 +1,275 @@
...
+    reset_tracer_cache()
+    try:
+        yield exp
+    finally:
+        trace._TRACER_PROVIDER = previous  # noqa: SLF001
+        reset_tracer_cache()
+
+
+def _single_span(exporter: InMemorySpanExporter) -> Any:
```
gtarpenning, line 46: `-> ReadableSpan` (already importing from the sdk). no `Any`.

tests/trace_server/test_tracing.py
```diff
@@ -0,0 +1,275 @@
...
+    """`functools.wraps` should preserve __name__, __doc__, __module__."""
+
+    @traced(name="span_name_unrelated_to_fn_name")
+    def some_fn(x: int) -> int:
+        """Docstring."""
+        return x
+
+    assert some_fn.__name__ == "some_fn"
+    assert some_fn.__doc__ == "Docstring."
+    assert some_fn.__wrapped__(5) == 5  # type: ignore[attr-defined]
```
gtarpenning, line 100: `getattr(some_fn, "__wrapped__")(5)` drops the `# type: ignore`.

gtarpenning: type ignores should be avoided at all costs even in tests

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,178 @@
...
+
+from opentelemetry import trace
+
+T = TypeVar("T")
+
+# Hoisted to module scope — re-resolving `get_tracer(...)` inside every wrapper
+# call costs ~7us per span open (measured), and these decorators run on hot
+# CH-query paths with dozens of spans per request. `get_tracer` returns a
+# `ProxyTracer` that lazy-resolves the real TracerProvider on first
+# `start_span`/`start_as_current_span` call and caches it. Tests that need to
```
gtarpenning, line 63: we can tighten up these comments

gtarpenning (commented):
some thoughts, also you should definitely tell your claude to stop `type: ignore`ing and `noqa:...` that's quite nasty.

amwarrier:
Superseded by the clean review-ready version at #7258. This PoC PR stays open as historical reference (validated end-to-end on a wandbench cluster). The new PR is the merge candidate.

PR #7197, opened 2026-06-11
chore(weave): Update samples to use the new feedback schema

mscavezze-cw:
## Description

Update samples to use the new feedback schema

## Testing

How was this PR tested?


mscavezze-cw:
Replacing with https://github.com/wandb/weave/pull/7207, which removes the unnecessary dependency on https://github.com/wandb/weave/pull/7193

PR #7198, opened 2026-06-11
chore(weave): add dataset_sources table migration

jwlee64:
## Description

Goal: see a dataset's **sources** — the calls, agent spans, and conversations that produced its rows — and the reverse: which datasets a given call, span, or conversation is a part of.

Migration `035` creates the `dataset_sources` ClickHouse table — storage for dataset-row provenance: links each dataset row to the call, agent span, or conversation that produced it, queryable both directions. **Schema only**; endpoints land in follow-up PRs, so nothing reads or writes it yet.

## Design

**Engine** — `ReplacingMergeTree(updated_at)`, read via `GROUP BY <logical key> + argMax(col, (updated_at, id))`, never `FINAL`. Writes are insert-only: link, relink, and soft-delete are all inserts of a new version row (delete = tombstone with `deleted_at` set and a newer `updated_at`); reads pick the latest version via argMax. RMT (over plain MergeTree) is what keeps storage and scan size bounded — it collapses superseded versions at merge time, whereas plain MergeTree would accumulate relink/concurrent duplicates forever with no cleanup.

**Deterministic link id** — `id = UUIDv5(project_id, dataset_object_id, row_digest, source_kind, source_id, source_trace_id)`. `source_trace_id` is part of the key because a span's identity is `(trace_id, span_id)` — two spans sharing a `span_id` across different traces are distinct edges and must not collapse (redundant for `call`, empty `''` for `conversation`, kept uniform across kinds). Relinking the same logical key always yields the same id (idempotent, no read-before-write), and merge-collapse can never dangle an id a client holds.

**`updated_at`** is `DateTime64(6)` (microsecond). It's both the RMT version column and the read-side tiebreaker. Because ids are deterministic, `id` can't break ties between versions of the same link — `updated_at` must, so it needs enough precision that sequential ops (delete → relink) are always distinct.

**Bloom filters** on `source_id` and `source_trace_id` power the reverse lookup (source → datasets), letting ClickHouse skip data parts that can't contain the queried ids.

### Perf

**Reads are flat from 5M to 100M rows** — the ORDER BY prefix bounds the forward  
scan and the bloom filter bounds the reverse scan, so latency is independent of  
total table size:

| **table size** | **on-disk** | **forward p50 / p95** | **reverse p50 / p95** |
| --- | --- | --- | --- |
| 5,000,000 | ~520 MB | 8\.2 / 13.5 ms | 12\.4 / 18.6 ms |
| 50,000,000 | ~5.2 GB | 9\.2 / 15.7 ms | 13\.8 / 29.9 ms |
| 100,000,000 | ~10.4 GB | 8\.8 / 13.5 ms | 11\.7 / 16.1 ms |

**Writes are fast and independent of table size** — 80k links insert in a single  
statement in **~290ms**, same at 1M and 100M baseline; throughput plateaus  
around ~290k rows/s:

| **batch size** | **median latency** | **throughput** |
| --- | --- | --- |
| 1000 | 71 ms | 14k rows/s |
| 10,000 | 93 ms | 108k rows/s |
| **80,000** | **296 ms** | **270k rows/s** |
| 250,000 | 853 ms | 293k rows/s |


jwlee64:
* **#7198** <a href="https://app.graphite.com/github/pr/wandb/weave/7198?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7199](https://github.com/wandb/weave/pull/7199) <a href="https://app.graphite.com/github/pr/wandb/weave/7199?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7238](https://github.com/wandb/weave/pull/7238) <a href="https://app.graphite.com/github/pr/wandb/weave/7238?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/migrations/035_add_dataset_sources.up.sql
```diff
@@ -0,0 +1,49 @@
...
+    source_trace_id String,
+    -- source_started_at / source_display_name are denormalized SNAPSHOTS of the
+    -- source captured at link time, NOT live joins. This table owns no SQL
+    -- against the calls/spans tables, and a conversation has no source row to
+    -- join at all; provenance is also a frozen claim about write-time. Contract:
+    -- these reflect the source as of when it was linked — a later rename of the
+    -- call/span is intentionally NOT reflected here.
+    source_started_at DateTime64(6),
+    source_display_name String,
+    link_metadata String DEFAULT '',
```
gtarpenning, line 21: i realize it's a bit of a pain but we should already have macros to deal with nullable columns, can we just use `''` as the empty value? it's more performant and we can add skip indexes on it in the future. 

same with all the nullables like `deleted_at`, we should already have some existing handling for dealing with datetime sentinels. `Nullable` is just not great in clickhouse

weave/trace_server/migrations/035_add_dataset_sources.up.sql
```diff
@@ -0,0 +1,49 @@
...
+    INDEX idx_source_id source_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_source_trace_id source_trace_id TYPE bloom_filter(0.01) GRANULARITY 1
+) ENGINE = ReplacingMergeTree(updated_at)
+-- A span's identity is (trace_id, span_id), so source_trace_id is part of the
+-- logical key (last, since reads filter by the source_id bloom, never trace_id).
+-- It is redundant for calls (call_id is globally unique) and empty ('') for
+-- conversations (a conversation spans traces, so conversation_id is sufficient
+-- identity); kept uniform across all three source kinds. The write layer
+-- enforces these invariants (conversation -> '', call/span -> non-empty trace).
+ORDER BY (project_id, dataset_object_id, row_digest, source_kind, source_id, source_trace_id)
```
gtarpenning, line 49: i think we should consider adding a `PartitionByYYYYMM` unless we think it will be common to have sources cross-month, in which case we should instead set the MAX size of the file part to like 50GB (default I think it's 150GB)

jwlee64: Added a PartitionByYYYYMM

weave/trace_server/migrations/035_add_dataset_sources.up.sql
jwlee64: added defaults instead of nullables

gtarpenning (approved):

weave/trace_server/migrations/035_add_dataset_sources.up.sql
```diff
@@ -0,0 +1,57 @@
...
+-- let retention drop whole months. A dataset's links can span months, which is
+-- fine. Matches the agent spans table's monthly partitioning.
+PARTITION BY toYYYYMM(created_at)
+-- A span's identity is (trace_id, span_id), so source_trace_id is part of the
+-- logical key (last, since reads filter by the source_id bloom, never trace_id).
+-- It is redundant for calls (call_id is globally unique) and empty ('') for
+-- conversations (a conversation spans traces, so conversation_id is sufficient
+-- identity). Kept uniform across all three source kinds. The write layer
+-- enforces these invariants (conversation -> '', call/span -> non-empty trace).
+ORDER BY (project_id, dataset_object_id, row_digest, source_kind, source_id, source_trace_id)
```
gtarpenning, line 49: @jwlee64 wait dude one qq, are we planning on using this entire key every time we read? like i think to use the primary key we would need to always filter down in order, so if you wanted to filter by `source_kind` you must always know `row_digest` etc (all the things above it in the chain). is that true? if so, awesome!

jwlee64: For the main query, ie `what are the sources of the dataset` we are intending to use project_id, dataset_object_id always, and not necessarily filter based on source kind as much, since you tend to want all sources of the dataset. But yes the order by should significantly speed this up

PR #7200, opened 2026-06-12
chore(weave): add structured fields when ref objects are not found

domphan-wandb:
## Description

- [https://coreweave.atlassian.net/browse/](https://coreweave.atlassian.net/browse/WB-35394)

This PR surfaces the missing refs in a structured field, `missing_digests.` This PR allows for better error handling on the application when a batched read is made. 

See accompanying FE PR for how the error gets surfaced on an eval table: https://github.com/wandb/core/pull/45716

When a run dies before the Weave SDK flushes all of its objects, refs_read_batch references objects the server doesn't have and fails the whole batch with a generic 404:
`{ "reason": "Ref read contains 4 digests, but found 3 objects. Diff digests: {'aq4j…'}" }`

The model evals client, reads every compared run's eval + model objects in one batch) can't tell which object is missing without scraping that human-readable string. This adds the missing digests as a structured field so clients can attribute the failure programmatically (we want to map a digest back to the run that referenced it) instead of parsing the message.

**What changed**
* New `RefObjectsNotFoundError(NotFoundError)` carrying missing_object_digests: list[str]. 
* Subclasses NotFoundError, so existing except NotFoundError callers and 404 handling are unaffected.

**Response shape**
Status code (404) and reason are unchanged; the only addition is the new field:
```json
{
  "reason": "Ref read contains 4 digests, but found 3 objects. Diff digests: {'aq4j…'}",
  "missing_object_digests": ["aq4j…"]
}
```

**Compatibility**
Additive:
* Existing clients that only read reason/status see no change.
* RefObjectsNotFoundError is a NotFoundError, so type-based handling continues to work.

## Testing

How was this PR tested?
* Added a unit test.

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -5813,7 +5813,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
                 if len(ref_digests) != len(found_digests):
                     missing_digests = sorted(ref_digests - found_digests)
                     raise RefObjectsNotFoundError(
-                        f"Ref read contains {len(ref_digests)} digests, but found {len(found_digests)} objects. Diff digests: {ref_digests - found_digests}",
+                        f"Ref read contains {len(ref_digests)} digests, but found {len(found_digests)} objects. Diff digests: {missing_digests}",
                         missing_digests,
                     )
                 # filter out deleted objects
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -179,6 +179,7 @@ from weave.trace_server.errors import (
     NotFoundError,
     ObjectDeletedError,
     ObjectNameTypeCollision,
+    RefObjectsNotFoundError,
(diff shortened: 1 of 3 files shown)


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5810,8 +5802,10 @@ def get_object_refs_root_val(
                 objs = self._select_objs_query(object_query_builder)
                 found_digests = {obj.digest for obj in objs}
                 if len(ref_digests) != len(found_digests):
-                    raise NotFoundError(
-                        f"Ref read contains {len(ref_digests)} digests, but found {len(found_digests)} objects. Diff digests: {ref_digests - found_digests}"
+                    missing_digests = sorted(ref_digests - found_digests)
+                    raise RefObjectsNotFoundError(
+                        f"Ref read contains {len(ref_digests)} digests, but found {len(found_digests)} objects. Diff digests: {ref_digests - found_digests}",
+                        missing_digests,
```
andrewtruong: why duplicate `ref_digests - found_digests`?

andrewtruong (approved):

PR #7201, opened 2026-06-12
feat(weave): migrate clickhouse_trace_server_batched.py to @traced (PR-7b1)

amwarrier:
## Summary

**Stacked on #7196 (PR-7a).** This is sub-PR 2 of 4 in the ddtrace → OpenTelemetry migration of `weave/trace_server/`.

Replaces all 47 `@ddtrace.tracer.wrap(name=X)` decorators in `clickhouse_trace_server_batched.py` with `@traced(name=X)` from `weave.trace_server.tracing` (added in PR-7a). 

## Notable: two generators move to `@traced_generator`

`_table_query_stream` and `annotation_queues_query_stream` are generator functions. With `@ddtrace.tracer.wrap`, these were **silently producing wrong span durations** — ddtrace's wrap closes the span when the wrapper returns the generator object, not when the generator is exhausted. The new `@traced_generator` uses `yield from` inside the span body so the span correctly covers the full iteration lifetime.

This is a behavior change: span durations on these two methods will be larger (and more accurate) after this PR.

## Test plan

- [ ] `python -m py_compile weave/trace_server/clickhouse_trace_server_batched.py` ✅ (validated locally)
- [ ] AST audit: every `@traced` lands on a non-generator, every `@traced_generator` lands on a generator ✅ (validated locally)
- [ ] No remaining `ddtrace` references in the file ✅
- [ ] Existing `tests/trace_server/` suite passes (CI)
- [ ] Tested end-to-end in wandb/core image on wandbench-small (post-merge of full stack)

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -15,7 +15,6 @@ from typing import Any, NamedTuple, TypeVar, cast
 from zoneinfo import ZoneInfo
 
 import clickhouse_connect
-import ddtrace
 from cachetools import TTLCache
 from clickhouse_connect import common as ch_common
 from clickhouse_connect.driver.client import Client as CHClient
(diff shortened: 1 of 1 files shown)


amwarrier:
Superseded by the clean review-ready version at #7263 (stacked on #7258). This PoC PR stays open as historical reference. The clean PR is the merge candidate.

PR #7203, opened 2026-06-12
feat(weave): drop ddtrace dep, inline DogStatsD in datadog.py (PR-7c)

amwarrier:
## Summary

**Stacked on PR-7b2.** Final sub-PR (4 of 4) of the ddtrace → OpenTelemetry migration of `weave/trace_server/`.

After this PR, `pip install . && pip list | grep ddtrace` returns empty.

### `weave/trace_server/datadog.py` rewritten

The module historical name (`datadog.py`) is preserved for backwards compatibility, but internally ddtrace is gone. Public symbols preserved:
- `DB_INSERT_METRIC`, `DB_INSERT_PATH_UNKNOWN` (constants)
- `db_insert_path(path)` (context manager)
- `tag_db_insert_path(path)` (decorator)
- `record_db_insert(*, table, count, path=None)` (function)
- `set_root_span_dd_tags(tags)`, `set_current_span_dd_tags(tags)` (functions)

Removed:
- `generator_trace(span_name)` — it existed only to suppress ddtrace's auto-mark-error-on-GeneratorExit. OTel's `start_as_current_span` doesn't have that behavior (BaseException subclasses propagate through `use_span` without marking the span errored), so this decorator becomes dead code. Callers should use `@traced_generator` from `weave.trace_server.tracing` instead. **Empirically verified** with the OTel SDK before deletion.

### Inline DogStatsD

The DogStatsD client is now a ~30-line `_StatsDClient` class:
- Opens a UDP socket to `${DD_DOGSTATSD_URL}` (or `${DD_AGENT_HOST}:8125` fallback)
- Encodes counters in DogStatsD wire format `name:value|c|#tags`
- Swallows `OSError` so stats hiccups never crash the request path
- Lazily creates the socket (import is side-effect free)

The metric name (`weave_trace_server.db_inserts`) and tag format are preserved, so existing DD dashboards keep resolving.

### `set_root_span_dd_tags` semantics change

The original used `ddtrace.tracer.current_root_span()`. OTel has no built-in current-root-span accessor, so this now uses `trace.get_current_span()` — the currently-active span. Per analysis in `services/weave-trace/docs/phase-b-weave-public-plan.md` §3.4, none of the three historical call sites actually need root semantics; they just need "the span currently active when this helper is called", which `get_current_span()` provides. **NOT a semantic regression** for known callers.

### `pyproject.toml`

Removed `ddtrace>=2.7.0` from both occurrences (line 80 in `trace_server` optional deps, line 425 in the `trace_server` dependency group).

## Test plan

- [ ] `python -m py_compile weave/trace_server/datadog.py` ✅
- [ ] `pip install . && pip list | grep ddtrace` returns empty (post-merge validation)
- [ ] Existing `tests/trace_server/` suite passes (CI)
- [ ] End-to-end: deploy weave-trace image with all of 7a/7b1/7b2/7c vendored into wandb/core, observe `weave_trace_server.db_inserts` counter still ingesting in DD

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,6 @@ rich = ["rich"]           # Optional dependency for enhanced console output
 # this to a separate package. Note, when that happens, we will need to pull along some of the
 # default dependencies as well.
 trace_server = [
-  "ddtrace>=2.7.0",
   # BYOB - S3
   "boto3>=1.34.0",
   # BYOB - Azure
@@ -422,7 +421,6 @@ dev = [
 ]
 trace_server = [
   # Same dependencies as the trace_server optional dependency, but as a dependency group for tests
-  "ddtrace>=2.7.0",
   # BYOB - S3
   "boto3>=1.34.0",
   # BYOB - Azure
--- a/weave/trace_server/datadog.py
+++ b/weave/trace_server/datadog.py
@@ -1,21 +1,42 @@
-"""Datadog integration utilities for trace server."""
+"""Datadog-integration utilities for trace server.
+
+Historical name kept for backward compatibility — internally this module
+no longer depends on `ddtrace`. The two things it provides are:
+
+  1. `db_insert_path` / `tag_db_insert_path` / `record_db_insert` —
+     emit DogStatsD counters for trace-server DB inserts. The wire
+     format and metric name (`weave_trace_server.db_inserts`) are
+     preserved so existing DD dashboards keep working.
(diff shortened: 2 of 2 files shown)


amwarrier:
Superseded by the clean review-ready version at #7268 (stacked on #7264 / #7263 / #7258). This PoC PR stays open as historical reference. The clean PR is the merge candidate.

PR #7208, opened 2026-06-12
feat(weave_ts)!: Remove export of `WeaveClient` value, only export as type.

drtangible:
## Description

* `WeaveClient` is currently exported as a value, likely accidentally since it's by way of a wildcard export here:

```js
// src/index.ts
export * from './types';
```

* This class is very likely unused directly by consumers -- clients are typically instantiated by way of `weave.init()`. (Functions for constructing the required argument types are also not exposed via the public API (`TraceServerApi`, `WandbServerApi`, `Settings`).

* ~This change marks this export (and the `constructor` itself) as deprecated, so we may remove it from the public API in a future version of the SDK.~

* This change remove the export as a value, while still exporting as a `type`. Breaking change, so we'll consider when bumping version number for next release.

### DIFF
--- a/sdks/node/scripts/loadTest.ts
+++ b/sdks/node/scripts/loadTest.ts
@@ -1,4 +1,5 @@
 import * as weave from 'weave';
+import type {WeaveClient} from 'weave';
 
 const func = weave.op(async () => 1);
 const myFunction = async (a: number = 1, b: string = 'hello', c: boolean = true) => {
@@ -13,7 +14,7 @@ const myFunction2 = async ({ a, b = 'wuh' }: { a?: number; b?: string }) => {
 };
 const func4 = weave.op(myFunction2, { parameterNames: 'useParam0Object' });
 
-async function bench(func: weave.Op<any>, calls: number, client: weave.WeaveClient) {
+async function bench(func: weave.Op<any>, calls: number, client: WeaveClient) {
   console.log(`Benchmarking with ${calls} calls...`);
   const startTime = Date.now();
   const promises = Array(calls)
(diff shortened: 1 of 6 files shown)


drtangible:
* **#7218** <a href="https://app.graphite.com/github/pr/wandb/weave/7218?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7217** <a href="https://app.graphite.com/github/pr/wandb/weave/7217?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7209** <a href="https://app.graphite.com/github/pr/wandb/weave/7209?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7208** <a href="https://app.graphite.com/github/pr/wandb/weave/7208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7208?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

andrewtruong:
I think we still want this export, otherwise how will users get data back?

drtangible:
> I think we still want this export, otherwise how will users get data back?

The functions on `WeaveClient` are still accessible on the client instance returned by `init()`.

```typescript
const client = init('example');
client.getWhatever()
```

This deprecates direct access to the constructor -- `new Weave(...)`. I wouldn't expect customers to be using that directly, but let's chat about that.

drtangible:
Moving the back to draft temporarily -- we discussed offline and are comfortable w/a more aggressive step here, dropping this export entirely.

drtangible:
Updated code and description for removing the export rather than deprecating.

andrewtruong (approved):

PR #7211, opened 2026-06-13
chore(weave): drop the wandb local-testcontainer from CI test jobs

andrewtruong:
Removes the wandb `local-testcontainer`, which is not required for CI.


drtangible (approved):
👍 👍 👍

PR #7216, opened 2026-06-15
fix(weave): don't stall flush() on unpaired eager calls under calls_complete

gtarpenning:
## Summary
- A mid-eval `client.flush()` cleared `_eager_call_ids`, so the later eager call-end no longer matched and parked in `_pending_ends` as a never-pairing orphan; the next flush then waited the full 5-min `FLUSH_TIMEOUT_SECONDS` on it.
- Keep eager tracking across a flush, exclude eager-unpaired ends from the pairing wait, and send them via the eager v2 endpoint right away. Non-eager in-flight work keeps the full pairing budget.
- WB-34557: https://coreweave.atlassian.net/browse/WB-34557

## Testing
new test asserts eager-unpaired flush returns fast while non-eager pending still consumes the full timeout.

### DIFF
--- a/tests/trace_server_bindings/test_call_batch_processor.py
+++ b/tests/trace_server_bindings/test_call_batch_processor.py
@@ -217,8 +217,8 @@ def test_missing_trace_id_raises_value_error() -> None:
         processor.stop_accepting_new_work_and_flush_queue()
 
 
-def test_flush_eager_sends_unpaired_items_and_clears_state() -> None:
-    """Flush sends unpaired items via the eager v2 endpoints and clears caches."""
+def test_flush_sends_unpaired_items_via_eager_endpoints() -> None:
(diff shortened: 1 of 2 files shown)


tests/trace_server_bindings/test_call_batch_processor.py
jtschoonhoven: Claude nit:

>Lingering assertion in the renamed test is now misleading (239): you dropped "and clears state"/"clears caches" from the name and docstring but kept assert len(processor._eager_call_ids) == 0. That assertion still passes only because both items in that test are non-eager (enqueue(...) with eager_call_start=False), so no eager ids are ever created, not because anything clears them. It now reads like it's verifying clear-on-flush, which is exactly the behavior you removed. Either drop the line or add a comment noting it's just confirming no eager IDs leaked from non-eager calls.

jtschoonhoven (approved):

PR #7219, opened 2026-06-15
feat(weave): add generic agent_name_override for OTel invoke_agent spans

andrewtruong:
By default, our auto-instrumentation for OTEL integrations create their own `invoke_agent` span and set `gen_ai.agent.name`.  This is not configurable today, so this PR introduces a way to configure it.


drtangible (approved):
👍 👍

PR #7221, opened 2026-06-15
fix(weave): tcp keep-alive on gcs pool for kill files/create

gtarpenning:
## Summary
- `files/create` p99 ~30s / max ~38s is a stalled GCS upload hitting a hard 30s read timeout, then a retry on a fresh connection succeeds in ~40ms (a 577-byte file took 31.4s at <10% pod CPU, so it is network I/O wait, not CPU). Root cause: GCP's VPC connection tracking silently reaps idle sockets (no RST) in the long-lived `storage.Client` pool, and urllib3's connection-reuse check can't detect a blackholed socket.
- Fix: TCP keep-alive on the GCS pool (custom `AuthorizedSession`) so idle sockets stay warm and are never reaped; split the single `timeout=30` into `(connect=10, read=30)` so connect fast-fails; size the pool to the concurrent-upload budget. Read timeout is unchanged, so legitimate slow-but-progressing transfers are unaffected.
- Jira: [WB-35637](https://coreweave.atlassian.net/browse/WB-35637); follow-up for s3 + azure: [WB-35655](https://coreweave.atlassian.net/browse/WB-35655)
- Impact (Datadog, 30d, sampled): ~337 GCS read-timeout stalls; 624 `files/create` >10s across 46 tenants, 180 >25s.

## Testing
existing file-storage suite green on clickhouse. The keep-alive effect (defeating GCP's idle reaper) isn't unit-testable - it needs a real socket + a 10-min idle window - so validation is the prod tail post-deploy.

[WB-35637]: https://coreweave.atlassian.net/browse/WB-35637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WB-35655]: https://coreweave.atlassian.net/browse/WB-35655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/file_storage.py
+++ b/weave/trace_server/file_storage.py
@@ -41,6 +41,7 @@ There are two ways to authenticate with Azure Blob Storage:
 """
 
 import logging
+import socket
 from abc import abstractmethod
 from collections.abc import Callable
 from typing import TypeVar, cast
@@ -51,8 +52,11 @@ from azure.storage.blob import BlobServiceClient
 from botocore.config import Config
 from botocore.exceptions import ClientError
 from google.api_core import exceptions as gcp_exceptions
+from google.auth import default as google_auth_default
+from google.auth.transport.requests import AuthorizedSession
 from google.cloud import storage
 from google.oauth2.credentials import Credentials as GCPCredentials
+from requests.adapters import HTTPAdapter
 from tenacity import (
     RetryCallState,
     before_sleep_log,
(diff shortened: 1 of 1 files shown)


amwarrier (approved):
Approving. But we might need a similar fix for s3 and azure.

PR #7223, opened 2026-06-15
feat(weave): per-conversation span sequences for the spans minimap

nikumar1206:
## JIRA Issue(s)

- [Turns / Events Minimap in Sessions list (Notion)](https://www.notion.so/wandbai/Turns-Minimap-in-Sessions-list-375e2f5c7ef380ec8e5fd994696125b2)

## Description

Feature half of the Sessions spans minimap backend (the trace-server interface/adapter/delegate wiring is in the stacked follow-up wandb/weave#7323).

Adds the query + transforms that produce, for a set of conversations:

- `spans` — an ordered per-span sequence (`kind`, `trace_id`, `span_id`, `status`, `duration_ms`), classified from span scalar columns (`operation_name` / `status_code`); no message bodies read. Capped at `MAX_CONVERSATION_SPANS`, keeping the most recent spans.
- `spans_feedback` — turn-anchored feedback markers (`trace_id`, `feedback_type`, detoned emoji `reactions`), fetched in one batched feedback query.

Exposed as `AgentQueryHandler.conversation_spans`. Replaces the `include_spans` flag on the grouped `agentSpansQuery` (removed here, along with `spans`/`spans_feedback` on `AgentSpanGroupRow`).

Also renames `ConversationEventKind` to `ConversationSpanKind` and documents why it is distinct from `AgentChatMessageType` (a coarse span-level classification, not the message taxonomy).

Stacked: wandb/weave#7323 (wiring). Frontend: wandb/core#45791 (+ wandb/core#45792).

## Testing

- [x] Manual
- [x] Unit
- [x] QA

### DIFF
--- a/weave/trace_server/agents/clickhouse.py
+++ b/weave/trace_server/agents/clickhouse.py
@@ -27,6 +27,10 @@ from weave.trace_server.agents.constants import (
     MAX_INGEST_ERRORS_REPORTED,
     NO_CONVERSATION_LABEL,
 )
+from weave.trace_server.agents.conversation_spans import (
+    parse_conversation_spans,
+    span_feedback_marker,
+)
 from weave.trace_server.agents.helpers import (
     genai_span_to_row,
     messages_from_ch_value,
@@ -43,6 +47,11 @@ from weave.trace_server.agents.types import (
     AgentConversationChatReq,
     AgentConversationChatRes,
     AgentConversationMessagePreview,
+    AgentConversationSpan,
+    AgentConversationSpanFeedback,
+    AgentConversationSpans,
+    AgentConversationSpansReq,
+    AgentConversationSpansRes,
     AgentCustomAttrSchemaItem,
     AgentCustomAttrsSchemaReq,
     AgentCustomAttrsSchemaRes,
@@ -88,6 +97,7 @@ from weave.trace_server.query_builder.agent_query_builder import (
     make_conversation_chat_spans_query,
     make_conversation_chat_turns_count_query,
     make_conversation_previews_query,
+    make_conversation_spans_query,
     make_custom_attrs_schema_query,
     make_message_search_query,
     make_span_group_categorical_distributions_query,
@@ -152,7 +162,7 @@ class AgentQueryHandler:
     Takes a `query_fn` (typically the server's `_query` method) so queries
     participate in the same logging / ddtrace / error-handling wrapper as the
     rest of the trace server. Also takes a `feedback_query_fn` (the server's
-    `feedback_query` method), invoked only when ``include_feedback=True`` to
+    `feedback_query` method), invoked only when `include_feedback=True` to
     fold agent-target feedback into the chat response.
     """
 
@@ -231,6 +241,111 @@ class AgentQueryHandler:
             preview = previews_by_conv.get(cid) if isinstance(cid, str) else None
             if preview is not None:
                 g.first_message, g.last_message = preview
+
+    def conversation_spans(
+        self, req: AgentConversationSpansReq
+    ) -> AgentConversationSpansRes:
+        """Return per-conversation span sequences + feedback markers.
+
+        Runs one bounded query scoped to the requested conversation_ids that
+        reads only scalar columns (no message bodies), then a single batched
+        feedback query for those conversations and their turns. Best-effort:
+        failures yield empty sequences so the spans minimap never hard-fails.
+        """
+        if not req.conversation_ids:
+            return AgentConversationSpansRes()
+        pb = ParamBuilder(PARAM_NAMESPACE)
+        sql = make_conversation_spans_query(
+            pb,
+            req.project_id,
+            req.conversation_ids,
+            started_after=req.started_after,
+            started_before=req.started_before,
+        )
+        try:
+            rows = _rows_as_dicts(self._query(sql, pb.get_params()))
+        except Exception:
+            logger.exception(
+                "failed to query conversation spans (project=%s, conversations=%d)",
+                req.project_id,
+                len(req.conversation_ids),
+            )
+            rows = []
+
+        spans_by_conv: dict[str, list[AgentConversationSpan]] = {}
+        # Map each turn (trace_id) back to its conversation so turn-level
+        # feedback can be bucketed without a second lookup.
+        trace_to_conv: dict[str, str] = {}
+        for row in rows:
+            cid = safe_str(row.get("conversation_id"))
+            if not cid:
+                continue
+            spans = parse_conversation_spans(row.get("spans"))
+            spans_by_conv[cid] = spans
+            for span in spans:
+                if span.trace_id:
+                    trace_to_conv.setdefault(span.trace_id, cid)
+
+        feedback_by_conv = self._fetch_conversation_spans_feedback(
+            req.project_id, req.conversation_ids, trace_to_conv
+        )
+
+        return AgentConversationSpansRes(
+            conversations=[
+                AgentConversationSpans(
+                    conversation_id=cid,
+                    spans=spans_by_conv.get(cid, []),
+                    spans_feedback=feedback_by_conv.get(cid, []),
+                )
+                for cid in req.conversation_ids
+            ]
+        )
+
+    def _fetch_conversation_spans_feedback(
+        self,
+        project_id: str,
+        conversation_ids: list[str],
+        trace_to_conv: dict[str, str],
+    ) -> dict[str, list[AgentConversationSpanFeedback]]:
+        """Batch-fetch conversation- and turn-level feedback for one list page.
+
+        One feedback query covering every conversation + turn target on the
+        page; results are bucketed per conversation and shaped into positioned
+        markers (turn markers carry their trace_id; conversation-level markers
+        carry none). Step-level (span) feedback is intentionally not fetched —
+        markers are anchored to turns for the list view. Best-effort.
+        """
+        by_conv: dict[str, list[AgentConversationSpanFeedback]] = {
+            cid: [] for cid in conversation_ids
+        }
+        try:
+            grouped = self._fetch_agent_feedback(
+                project_id=project_id,
+                conversation_ids=list(conversation_ids),
+                trace_ids=list(trace_to_conv),
+            )
+        except Exception:
+            logger.exception(
+                "failed to fetch conversation spans feedback "
+                "(project=%s, conversations=%d, turns=%d)",
+                project_id,
+                len(conversation_ids),
+                len(trace_to_conv),
+            )
+            return by_conv
+
+        for cid, items in grouped.by_conversation_id.items():
+            if cid in by_conv:
+                by_conv[cid].extend(
+                    span_feedback_marker(raw, trace_id="") for raw in items
+                )
+        for tid, items in grouped.by_trace_id.items():
+            conv_id = trace_to_conv.get(tid)
+            if conv_id is not None and conv_id in by_conv:
+                by_conv[conv_id].extend(
+                    span_feedback_marker(raw, trace_id=tid) for raw in items
+                )
+        return by_conv
 
     def spans_stats(self, req: AgentSpanStatsReq) -> AgentSpanStatsRes:
         """Return chart-ready aggregations over spans."""
@@ -779,9 +894,9 @@ class AgentWriteHandler:
     def insert_span(self, span: AgentSpanCHInsertable) -> None:
         """Insert a single pre-built span into the spans table.
 
-        Unlike ``insert_otel_spans`` this skips OTel protobuf parsing and
+        Unlike `insert_otel_spans` this skips OTel protobuf parsing and
         GenAI extraction — the caller is responsible for constructing a
-        fully populated ``AgentSpanCHInsertable``.
+        fully populated `AgentSpanCHInsertable`.
         """
         self.insert_spans([span])
 
--- a/weave/trace_server/agents/clickhouse.py
+++ b/weave/trace_server/agents/clickhouse.py
@@ -28,6 +28,7 @@ from weave.trace_server.agents.constants import (
     NO_CONVERSATION_LABEL,
 )
 from weave.trace_server.agents.conversation_spans import (
+    is_supported_feedback,
     parse_conversation_spans,
     span_feedback_marker,
 )
@@ -337,13 +338,17 @@ class AgentQueryHandler:
         for cid, items in grouped.by_conversation_id.items():
             if cid in by_conv:
                 by_conv[cid].extend(
-                    span_feedback_marker(raw, trace_id=None) for raw in items
+                    span_feedback_marker(raw, trace_id=None)
+                    for raw in items
+                    if is_supported_feedback(safe_str(raw.get("feedback_type")))
                 )
         for tid, items in grouped.by_trace_id.items():
             conv_id = trace_to_conv.get(tid)
             if conv_id is not None and conv_id in by_conv:
                 by_conv[conv_id].extend(
-                    span_feedback_marker(raw, trace_id=tid) for raw in items
+                    span_feedback_marker(raw, trace_id=tid)
+                    for raw in items
+                    if is_supported_feedback(safe_str(raw.get("feedback_type")))
                 )
         return by_conv
 
--- a/weave/trace_server/agents/clickhouse.py
+++ b/weave/trace_server/agents/clickhouse.py
@@ -337,7 +337,7 @@ class AgentQueryHandler:
         for cid, items in grouped.by_conversation_id.items():
             if cid in by_conv:
                 by_conv[cid].extend(
-                    span_feedback_marker(raw, trace_id="") for raw in items
+                    span_feedback_marker(raw, trace_id=None) for raw in items
                 )
         for tid, items in grouped.by_trace_id.items():
             conv_id = trace_to_conv.get(tid)
--- a/weave/trace_server/agents/conversation_spans.py
+++ b/weave/trace_server/agents/conversation_spans.py
@@ -85,9 +85,13 @@ def _feedback_tags(feedback_type: str, raw: dict[str, Any]) -> list[str]:
 
 
 def span_feedback_marker(
-    raw: dict[str, Any], *, trace_id: str
+    raw: dict[str, Any], *, trace_id: str | None
 ) -> AgentConversationSpanFeedback:
-    """Shape a raw feedback row into a positioned spans feedback marker."""
+    """Shape a raw feedback row into a positioned feedback marker.
+
+    `trace_id` is the turn the feedback is anchored to, or None for
+    conversation-level feedback.
+    """
     feedback_type = safe_str(raw.get("feedback_type"))
     return AgentConversationSpanFeedback(
         trace_id=trace_id,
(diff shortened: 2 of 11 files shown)


weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,54 @@ class AgentConversationMessagePreview(BaseModel):
...
+    "assistant",
+    "tool",
+    "agent",
+    "handoff",
+    "compaction",
+    "unknown",
+]
+"""Kind of a single conversation spans event, mirroring the chat-view
+message taxonomy (AgentChatMessageType). Classified from span scalar columns (operation_name +
+status_code) on the cheap spans path, not from message bodies."""
```
jtschoonhoven: This should go above the thing its describing.

Also tiny nit, I think the convention is to use google-style docstrings, that leave the trailing `"""` on a newline

```
"""Like
this
"""
```

Though if it's not enforced with a required lint check then it's not really a rule.

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,54 @@ class AgentConversationMessagePreview(BaseModel):
...
 
+ConversationEventKind = Literal[
+    "user",
+    "assistant",
+    "tool",
+    "agent",
+    "handoff",
+    "compaction",
+    "unknown",
+]
```
jtschoonhoven: Why do we need this _and_ `AgentChatMessageType`? They look the same to me.

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,54 @@ class AgentConversationMessagePreview(BaseModel):
...
+"""Kind of a single conversation spans event, mirroring the chat-view
+message taxonomy (AgentChatMessageType). Classified from span scalar columns (operation_name +
+status_code) on the cheap spans path, not from message bodies."""
+
+
+class AgentConversationSpan(BaseModel):
+    """One lightweight event in a conversation's spans sequence.
```
jtschoonhoven: Let's not say "event" when we mean "span". And I think "sequence" here means "trace"?
```suggestion
    """One span in a conversation's trace.
```

weave/trace_server/agents/types.py
```diff
@@ -640,6 +686,11 @@ class AgentSpanGroupRow(BaseModel):
     # latest span carried no renderable message text.
     first_message: AgentConversationMessagePreview | None = None
     last_message: AgentConversationMessagePreview | None = None
+    # Per-conversation spans sequence + feedback markers, populated only when
+    # AgentSpansQueryReq.include_spans is set (conversation grouping only) via
+    # _hydrate_conversation_spans. Empty otherwise.
+    spans: list[AgentConversationSpan] = Field(default_factory=list)
+    spans_feedback: list[AgentConversationSpanFeedback] = Field(default_factory=list)
```
jtschoonhoven: I don't love the pattern of having one giant endpoint that gives different-shaped responses depending on which flags are included. Why does this new feature belong on `AgentSpanGroupRow` and not in a new endpoint?

jtschoonhoven (commented):
Partial review

nikumar1206:
* **#7323** <a href="https://app.graphite.com/github/pr/wandb/weave/7323?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7223** <a href="https://app.graphite.com/github/pr/wandb/weave/7223?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7223?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,77 @@ class AgentConversationMessagePreview(BaseModel):
     text: str = ""
 
 
+# Coarse span kind from operation_name / status_code. Distinct from
+# AgentChatMessageType: the classifier only emits agent / tool / assistant
+# (+ unknown), not the full message taxonomy.
+ConversationSpanKind = Literal[
```
nikumar1206: so this looks somewhat similar to `AgentChatMessageType`, but these are just the event types we can glean from the `operation_name`, while AgentChatMessageType requires looking for message bodies.

nikumar1206: i think we can force usign AgentChatMessageType and that would make it one less constant to keep updated, but semantically they are used for different things.

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+    only (no message bodies). Spans are ordered by `started_at`, which
+    approximates — but does not exactly match — the detail chat view's
+    parent/child tree-walk order.
+    """
+
+    kind: ConversationSpanKind = "unknown"
+    trace_id: str = ""
+    span_id: str = ""
+    status: StatusCodeLiteral = "UNSET"
+    duration_ms: int = 0
```
jtschoonhoven: Small nit: I don't see any advantage to setting default values here. Seems like it just increases the risk that we'll forget to set one of these.

Other small nit: it might be worth translating the empty sentinel values back into explicit `None`.

```suggestion
    kind: ConversationSpanKind
    trace_id: str | None
    span_id: str | None
    status: StatusCodeLiteral
    duration_ms: int | None
```

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+    trace_id: str = ""
+    feedback_type: str = ""
+    reactions: list[str] = Field(
+        default_factory=list,
+        description=(
+            "Detoned emoji `:shortcode:` aliases on this feedback; "
+            "empty when it carries none."
+        ),
+        examples=[[":thumbs_up:"]],
+    )
```
jtschoonhoven, line 673: Blocker: this should be updated to use tags.

Up to you if you want to make this backwards-compatible with reactions, but we should design for tags going forwards.

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -1065,6 +1067,96 @@ def make_conversation_previews_query(
...
+    latest activity. `started_at` is carried only as the sort key; the handler
+    ignores it when building `AgentConversationSpan` rows.
+
+    `max_spans` bounds the returned (and wire) payload, not the intermediate
+    aggregation: `groupArray` still materializes the full per-conversation
+    tuple array before the sort + slice, so server-side memory is bounded by
+    page-size x spans-per-conversation. A `groupArraySorted(max_spans)` could
+    bound the intermediate too, but it sorts ascending (keeping the head), so
+    keeping the recent tail would need a descending sort key.
+    """
```
jtschoonhoven: This docstring is way too long to be useful. I usually tell claude to limit itself to 3 sentences max. If it really takes 3 paragraphs to explain the code then something is wrong.

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+
+class AgentConversationSpansReq(BaseModel):
+    """Request the span sequences for an explicit set of conversations.
+
+    Reads span scalar columns only (no message bodies) for the given
+    `conversation_ids`. Powers the conversations-list spans minimap.
+    """
+
+    project_id: str
+    conversation_ids: list[str] = Field(default_factory=list)
```
jtschoonhoven: There should be an upper bound on the number of conversation IDs here so the client can't abuse this endpoint.

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -1065,6 +1067,96 @@ def make_conversation_previews_query(
     """
 
 
+def _span_kind_sql(op_col: str) -> str:
```
jtschoonhoven: I like that this is factoring out re-usable logic, but as a best-practice, we should avoid writing functions that build SQL by injecting arbitrary strings. Plus it's not obvious that `op_col` needs an aliased table name prefix like `s.operation_name`.

An easy fix might be to just put this logic back in `make_conversation_spans_query`.

weave/trace_server/agents/clickhouse.py
```diff
@@ -228,6 +238,111 @@ def _hydrate_conversation_previews(
             if preview is not None:
                 g.first_message, g.last_message = preview
 
+    def conversation_spans(
+        self, req: AgentConversationSpansReq
+    ) -> AgentConversationSpansRes:
```
jtschoonhoven, line 248: Probably worth testing this too. Currently only `make_conversation_spans_query` is tested.

EDIT: Looks like you add more testing in the next PR so this comment is optional

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+# AgentChatMessageType: the classifier only emits agent / tool / assistant
+# (+ unknown), not the full message taxonomy.
+ConversationSpanKind = Literal[
+    "user",
+    "assistant",
+    "tool",
+    "agent",
+    "handoff",
+    "compaction",
+    "unknown",
```
jtschoonhoven: Claude flagged:

>The classifier (_span_kind_sql) only ever emits agent/tool/assistant, and parse_conversation_spans coerces anything else to unknown

So you should probably update either `parse_conversation_spans` or `ConversationSpanKind` so they match

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -1065,6 +1067,96 @@ def make_conversation_previews_query(
...
+    )
+    kind = _span_kind_sql("s.operation_name")
+    return f"""
+        SELECT s.conversation_id AS conversation_id,
+               arraySlice(
+                   arraySort(
+                       x -> x.1,
+                       groupArray(
+                           tuple(
+                               s.started_at,
```
jtschoonhoven, line 1203: Nit: consider adding a tie-breaker field to this sort in case two spans share the same `started_at`.

jtschoonhoven (commented):
A few comments before approving. Main blocker is that this API still uses reactions instead of tags.

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+        examples=[["\U0001f44d"]],
+    )
+
+
+class AgentConversationSpans(BaseModel):
+    """One conversation's span sequence and its feedback markers."""
+
+    conversation_id: str
+    spans: list[AgentConversationSpan] = Field(default_factory=list)
+    spans_feedback: list[AgentConversationSpanFeedback] = Field(default_factory=list)
```
jtschoonhoven, line 685: Should the response also include conversation-level feedback?

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+
+class AgentConversationSpanFeedback(BaseModel):
+    """A feedback marker for a conversation's trace.
+
+    Positioned client-side by matching `trace_id` (turn) against the spans.
+    `tags` holds the human tags on this feedback (emoji reactions are tags too).
+    """
+
+    trace_id: str = ""
+    feedback_type: str = ""
```
jtschoonhoven: This should have a more specific type

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+
+
+class AgentConversationSpanFeedback(BaseModel):
+    """A feedback marker for a conversation's trace.
+
+    Positioned client-side by matching `trace_id` (turn) against the spans.
+    `tags` holds the human tags on this feedback (emoji reactions are tags too).
+    """
+
+    trace_id: str = ""
```
jtschoonhoven: Why is this optional?

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+    feedback_type: str = ""
+    tags: list[str] = Field(
+        default_factory=list,
+        description=(
+            "Human tags on this feedback, skin-tone-detoned. Emoji tags are the "
+            "glyph itself (e.g. a thumbs-up); empty when it carries none."
```
jtschoonhoven: I'm pretty sure you want to include _all_ tags, not just manually-applied tags. This comment is confusing, it should probably just say "Arbitrary descriptive tags applied to this span".

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+    """
+
+    kind: ConversationSpanKind = "unknown"
+    trace_id: str = ""
+    span_id: str = ""
+    status: StatusCodeLiteral = "UNSET"
+    duration_ms: int = 0
+
+
+class AgentConversationSpanFeedback(BaseModel):
```
jtschoonhoven, line 658: Nit: I'd call this `AgentConversationSpanTags` (if you're just including tags) and `AgentConversationSpanScores` (if you're including ratings).

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,82 @@ class AgentConversationMessagePreview(BaseModel):
...
+    """A feedback marker for a conversation's trace.
+
+    Positioned client-side by matching `trace_id` (turn) against the spans.
+    `tags` holds the human tags on this feedback (emoji reactions are tags too).
+    """
+
+    trace_id: str = ""
+    feedback_type: str = ""
+    tags: list[str] = Field(
```
jtschoonhoven, line 669: Why aren't you including ratings here? Seems like this should include ratings.

weave/trace_server/agents/conversation_spans.py
```diff
@@ -0,0 +1,139 @@
...
+)
+
+
+def is_supported_feedback(feedback_type: str) -> bool:
+    """Whether a feedback row's type is surfaced by agent_conversation_spans."""
+    return feedback_type in _FEEDBACK_TYPES
+
+
+# Valid span kinds; unexpected SQL values coerce to "unknown".
+_SPAN_KINDS: frozenset[ConversationSpanKind] = frozenset(get_args(ConversationSpanKind))
```
jtschoonhoven: I recommend declaring this immediately below ConversationSpanKind in types.py so that it's easy to see what it contains

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,93 @@ class AgentConversationMessagePreview(BaseModel):
...
+    kind: ConversationSpanKind
+    trace_id: str
+    span_id: str
+    status: StatusCodeLiteral
+    duration_ms: int
+
+
+# Feedback types this endpoint surfaces: human tags (agent_user_feedback) and
+# scorer tags + ratings (agent_monitor). Other feedback types are ignored.
+AgentSpanFeedbackType = Literal["wandb.agent_user_feedback", "wandb.agent_monitor"]
```
jtschoonhoven: This should probably be moved to `feedback_types.py` instead of `types.py`. 

Also, the comment "feedback types this endpoint surfaces..." should be updated, since it's not clear what endpoint you're referring to.

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,93 @@ class AgentConversationMessagePreview(BaseModel):
...
+class AgentConversationSpan(BaseModel):
+    """One span in a conversation's trace.
+
+    Returned by `agent_conversation_spans`, which reads span scalar columns
+    only (no message bodies). Spans are ordered by `started_at`, which
+    approximates — but does not exactly match — the detail chat view's
+    parent/child tree-walk order.
+    """
+
+    kind: ConversationSpanKind
```
jtschoonhoven: OTel already defines "span kind" a different way, so this should be renamed. Just return the `operation_name`

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -1065,6 +1067,74 @@ def make_conversation_previews_query(
...
+        "0)"
+    )
+    # Coarse span kind from operation_name (see ConversationSpanKind): agent
+    # invocations -> agent, tool executions -> tool, everything else -> assistant.
+    kind = (
+        "multiIf("
+        f"s.operation_name = '{OP_INVOKE_AGENT}', 'agent', "
+        f"s.operation_name = '{OP_EXECUTE_TOOL}', 'tool', "
+        "'assistant')"
+    )
```
jtschoonhoven: Are you sure it's safe for this to default to "assistant"? Why not "unknown"?

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,93 @@ class AgentConversationMessagePreview(BaseModel):
     text: str = ""
 
 
+# Coarse span kind classified from operation_name / status_code. Distinct from
+# AgentChatMessageType: the classifier only emits these values, not the full
+# message taxonomy (which needs message bodies).
+ConversationSpanKind = Literal["agent", "tool", "assistant", "unknown"]
```
jtschoonhoven: The OTel spec already defines `span_kind` and `operation_name`. Here you're introducing a new classification that exists only for this one endpoint. That's too confusing. I think this endpoint should return the original `operation_name` and you should transform it client-side if needed

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,93 @@ class AgentConversationMessagePreview(BaseModel):
...
+    feedback_type: AgentSpanFeedbackType
+    tags: list[str] = Field(
+        default_factory=list,
+        description="Arbitrary descriptive tags applied to this feedback.",
+        examples=[["\U0001f44d"]],
+    )
+    scores: list[AgentConversationSpanScore] = Field(
+        default_factory=list,
+        description="Numeric scores (ratings) applied to this feedback.",
+    )
```
jtschoonhoven: "Scores" also include tags. Call these "ratings"
```suggestion
    ratings: list[AgentConversationSpanScore] = Field(
        default_factory=list,
        description="Numeric scores (ratings) applied to this feedback.",
    )
```

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,93 @@ class AgentConversationMessagePreview(BaseModel):
...
+    status: StatusCodeLiteral
+    duration_ms: int
+
+
+# Feedback types this endpoint surfaces: human tags (agent_user_feedback) and
+# scorer tags + ratings (agent_monitor). Other feedback types are ignored.
+AgentSpanFeedbackType = Literal["wandb.agent_user_feedback", "wandb.agent_monitor"]
+
+
+class AgentConversationSpanScore(BaseModel):
```
jtschoonhoven: "Rating"
```suggestion
class AgentConversationSpanRating(BaseModel):
```

weave/trace_server/agents/types.py
```diff
@@ -610,6 +610,93 @@ class AgentConversationMessagePreview(BaseModel):
...
+    """
+
+    trace_id: str | None = Field(
+        description="The turn this feedback is anchored to; None for conversation-level."
+    )
+    feedback_type: AgentSpanFeedbackType
+    tags: list[str] = Field(
+        default_factory=list,
+        description="Arbitrary descriptive tags applied to this feedback.",
+        examples=[["\U0001f44d"]],
```
jtschoonhoven: Can you add a readable example here? I'm not even sure we'll parse unicode escape sequences correctly.

jtschoonhoven (commented):
Partial review

weave/trace_server/query_builder/agent_query_builder.py
jtschoonhoven: EDIT: this should just return the raw operation name

bcsherma (approved):

PR #7237, opened 2026-06-15
feat(weave): send trace_id on calls-ingest requests that lack it

ro31337:
## JIRA Issue(s)
https://coreweave.atlassian.net/browse/WB-35682

## Description
Today a call's `trace_id` reaches the server on only some calls-ingest messages. Call-start carries it, but call-end and the end-parts of the batched upsert path do not. A trace arrives as many separate messages (start, end, children, batches), so without the trace key on every message the server cannot reason about a whole trace at ingest time — it has to merge first.

This change carries `trace_id` on every calls-ingest message that currently lacks it, in the cheapest place to read it later. Single-call requests (`/call/start`, `/call/end`, their v2 variants, and the eager v2 path that evaluations use) now send an `X-Weave-Trace-Id` header, so a server-side middleware can read it without parsing the body. Batched requests can't use a header because one request spans many traces, so the end schema (`EndedCallSchemaForInsert`) gains an optional `trace_id` field that a body-parsing reader can use; this single field covers `/call/end`, the end-parts of `/call/upsert_batch`, and v2 `call/end`. The `calls/complete` path already carries `trace_id` and is untouched.

This is groundwork with no server behavior change yet. `/call/update`, `/calls/delete`, and the feedback endpoints are a separate follow-up.

## Why this approach
The header keeps the trace key where a cheap pre-body check can see it, the same pattern `RequestSizeLimitMiddleware` and the existing `X-Weave-Retry-Id` header already use. The schema field reuses the exact path that recently added `started_at` to the same end message, so one optional field fans out to the three end endpoints.

## Backward compatibility
The header and the field are both optional with safe defaults, so it is fail-open in both directions: a new client against an old server has its extra header/field ignored, and an old client against a new server simply omits them. The server-side call-end path doesn't read the field, so nothing is written or changed.

## Testing
Added transport tests that decode the posted request and assert the header on single-call and eager-v2 requests, the field in the end body, that a batch post (which spans traces) carries no per-trace header, and the no-trace_id backward-compat case. Verified locally: the `trace_server_bindings` shard (61 passed), the full client-trace suite (127 passed), and the evaluation suite (20 passed).

### DIFF
--- a/tests/trace_server_bindings/test_http_behavior_remote.py
+++ b/tests/trace_server_bindings/test_http_behavior_remote.py
@@ -28,6 +28,7 @@ from weave.trace_server_bindings.async_batch_processor import AsyncBatchProcesso
 from weave.trace_server_bindings.call_batch_processor import CallBatchProcessor
 from weave.trace_server_bindings.http_utils import (
     ERROR_CODE_CALLS_COMPLETE_MODE_REQUIRED,
+    TRACE_ID_HEADER,
 )
 from weave.trace_server_bindings.models import (
     CompleteBatchItem,
@@ -49,6 +50,11 @@ def make_calls_complete_required_response() -> httpx.Response:
         },
         request=httpx.Request("POST", "http://example.com/call/upsert_batch"),
     )
+
+
+def _request_headers(mock_call) -> dict:
+    """Pull the headers kwarg from a mocked http_requests.* call."""
+    return mock_call.kwargs.get("headers", {})
 
 
 @pytest.fixture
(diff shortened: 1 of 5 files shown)


tests/trace_server_bindings/test_http_behavior_remote.py
```diff
@@ -459,3 +480,160 @@ def test_eager_calls_complete_required_is_reraised(mock_post, monkeypatch):
...
+                project_id="entity/project",
+                id=call_id,
+                trace_id="trace-batch",
+                ended_at=datetime.datetime.now(tz=datetime.timezone.utc),
+                summary={"result": "ok"},
+            )
+        )
+    )
+
+    try:
```
andrewtruong, line 617: Why does this need to be wrapped in a try?

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -127,14 +137,16 @@ def get(self, url: str, *args: Any, **kwargs: Any) -> httpx.Response:
...
+        self, url: str, *args: Any, trace_id: str | None = None, **kwargs: Any
+    ) -> httpx.Response:
+        headers = self._build_dynamic_request_headers(trace_id=trace_id)
 
         return http_requests.post(
             self.trace_server_url + url,
             *args,
             auth=self._auth,
-            headers=headers,
+            headers={**headers, **kwargs.pop("headers", {})},
```
andrewtruong: why modify this?  Shouldn't your logic be inside the helper above?

andrewtruong (approved):
lgtm, just some nits

tests/trace_server_bindings/test_http_behavior_remote.py
ro31337: Mirrors the same pattern in the this file, for example  `test_calls_complete_batch_endpoint_and_payload`. As I understand, `should_batch=True` spins up threads and we need to stop them

weave/trace_server_bindings/remote_http_trace_server.py
ro31337: Hold on, fixing

PR #7199, opened 2026-06-11
feat(weave): dataset_sources provenance endpoints

jwlee64:
## Description

- Fixes WB-NNNNN
- Fixes #NNNN

What does the PR do? Include a concise description of the PR contents.

## Testing

How was this PR tested?

PR #7238, opened 2026-06-15
feat(weave): dataset_sources query builders + unit tests

jwlee64:
## Description

https://coreweave.atlassian.net/browse/WB-35605

Adds the `dataset_sources` query builder module and a corresponding batch span existence query to the agent query builder.  
Follows a similar pattern to `annotation_queues_query_builder.py` as the table structure is similar

Pure SQL/query-builder layer for `dataset_sources` — no DB calls, fully unit-tested.

- `deterministic_link_id(...)` — UUIDv5 over the logical key
`(project_id, dataset_object_id, row_digest, source_kind, source_id, source_trace_id)`,
so the same link always maps to the same id (relink → same row; delete/restore are version rows).
- `make_dataset_sources_select(...)` — forward query (dataset → sources): `GROUP BY` the logical
key + `argMax(col, (updated_at, id))`, never `FINAL`; excludes soft-deleted by default.
- `make_source_datasets_select(...)` — reverse query (source → datasets), with row-digest capping.
- `make_link_state_query` / `make_link_state_by_ids_query` — read current state for relink/delete.
- `make_spans_existence_query(...)` — validate span sources against the `spans` table.

`make_spans_existence_query(...)` lives in `agent_query_builder.py` rather than the
`dataset_sources` module because the `spans` table is owned by the agents module. Linking a
span source has to confirm the span exists, and we want that SQL written/maintained alongside
the rest of the spans-table queries — provenance code calls into the agents query builder
instead of hand-rolling its own query against a table it doesn't own.

## Testing

Unit tests are added in `tests/trace_server/query_builder/test_dataset_sources_query_builder.py`, covering:

- 8 unit tests — id determinism (incl. distinct-trace D1), GROUP BY/argMax dedup without  
FINAL, soft-delete exclusion, param-builder usage.


jwlee64:
* **#7239** <a href="https://app.graphite.com/github/pr/wandb/weave/7239?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7240](https://github.com/wandb/weave/pull/7240) <a href="https://app.graphite.com/github/pr/wandb/weave/7240?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7241](https://github.com/wandb/weave/pull/7241) <a href="https://app.graphite.com/github/pr/wandb/weave/7241?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7238** <a href="https://app.graphite.com/github/pr/wandb/weave/7238?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7238?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7198** <a href="https://app.graphite.com/github/pr/wandb/weave/7198?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7199](https://github.com/wandb/weave/pull/7199) <a href="https://app.graphite.com/github/pr/wandb/weave/7199?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
+"""Query builder for the dataset_sources membership table.
+
+`dataset_sources` records a membership between dataset rows and their provenance
+sources. Shared invariants: deterministic ids, soft-delete tombstones,
+idempotent relink, and read-side dedup via GROUP BY + argMax (never FINAL).
+
+ARCHITECTURE NOTE: this module deliberately owns NO SQL against tables it does
```
gtarpenning: probably can clean up this comment

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+from __future__ import annotations
+
+import uuid
+
+from weave.trace_server.calls_query_builder.calls_query_builder import (
+    _format_table_name_with_cluster,
+)
+from weave.trace_server.orm import ParamBuilder
+
+# Stable namespace for deterministic UUIDv5 link ids. Generated once and frozen;
```
gtarpenning: uuidv5? not 7? not 4?

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+    dataset_object_id: str,
+    row_digest: str,
+    source_kind: str,
+    source_id: str,
+    source_trace_id: str,
+) -> str:
+    """Deterministic UUIDv5 over the logical key.
+
+    The logical key
+    ``(project_id, dataset_object_id, row_digest, source_kind, source_id,
```
gtarpenning, line 66: why double ``

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+    _format_table_name_with_cluster,
+)
+from weave.trace_server.orm import ParamBuilder
+
+# Stable namespace for deterministic UUIDv5 link ids. Generated once and frozen;
+# changing this would re-key every existing link and break idempotent relink.
+# The KEY COMPOSITION is frozen too: the set and order of fields hashed into the
+# id (see deterministic_link_id) — including source_trace_id — must not change,
+# or existing links would re-key.
+DATASET_SOURCE_LINK_NAMESPACE = uuid.UUID("9f2c1d3e-7b4a-5c6d-8e9f-0a1b2c3d4e5f")
```
gtarpenning, line 32: is `uuid.UUID` the same as `uuid.uuid5` (below)

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+    if offset is not None:
+        offset_param = pb.add(offset, param_type="Int64")
+        query += f"\n        OFFSET {offset_param}"
+
+    return query
+
+
+def make_source_datasets_select(
+    project_id: str,
+    sources: list[tuple[str, str, str]],
```
gtarpenning, line 306: i think we need to guard against sources being empty in this fn

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,205 @@
...
+    query = make_dataset_sources_select(
+        project_id=project_val,
+        dataset_object_id=dataset_val,
+        pb=pb,
+    )
+    params = pb.get_params()
+    normalized = " ".join(query.split()).lower()
+
+    # Read-side dedup: GROUP BY + argMax, NEVER FINAL.
+    assert "group by" in normalized
```
gtarpenning: every one of these tests needs to assert against the whole sql shape not just x in y, it should be x == ....

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -1405,6 +1405,42 @@ def make_trace_detail_spans_query(
     """
 
 
+def make_spans_existence_query(
```
gtarpenning, line 1508: do we need a unit test against this?

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+    "project_id",
+    "dataset_object_id",
+    "row_digest",
+    "source_kind",
+    "source_id",
+    "source_trace_id",
+]
+
+
+def make_link_state_query(
```
gtarpenning, line 151: same here, probably need a unit test against this one?

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+        FROM dataset_sources AS ds
+        WHERE ds.project_id = {project_param}
+            AND ds.dataset_object_id = {dataset_param}
+            AND ds.row_digest IN {row_digests_param}
+            AND ds.source_id IN {source_ids_param}
+        GROUP BY {group_by}
+    """
+
+
+def make_link_state_by_ids_query(
```
gtarpenning, line 197: and this one

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,205 @@
...
+    # is a sentinel-epoch column (not Nullable), so "live" is `= toDateTime64(0)`.
+    assert "having deleted_at = todatetime64(0, 3)" in normalized
+
+
+# ---------------------------------------------------------------------------
+# Reverse query (sources -> datasets)
+# ---------------------------------------------------------------------------
+
+
+def test_make_source_datasets_select_dedups_via_group_by_argmax_no_final() -> None:
```
gtarpenning: does this test the two level aggregation? 

also why is this test called `_no_final` like I get it you aren't using `FINAL` great

gtarpenning: (other tests probably need that treatment)

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+(benchmarked ~100x slower). All user-supplied values go through ``ParamBuilder``;
+no f-string interpolation of user data.
+"""
+
+from __future__ import annotations
+
+import uuid
+
+from weave.trace_server.calls_query_builder.calls_query_builder import (
+    _format_table_name_with_cluster,
```
gtarpenning: do we use this? `_format_table_name_with_cluster`

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+    """
+
+
+# Re-export so callers can format the local mutation table name if ever needed
+# (dataset_sources never mutates — insert-only — but keep symmetry with the
+# annotation_queues builder which exposes the same helper).
+__all__ = [
+    "DATASET_SOURCES_INSERT_COLUMNS",
+    "DATASET_SOURCE_LINK_NAMESPACE",
+    "_format_table_name_with_cluster",
```
gtarpenning: why export something we are importing?

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+    "updated_at",
+    "deleted_at",
+]
+
+
+def _argmax_select(columns: list[str]) -> str:
+    """Build ``argMax(ds.col, (ds.updated_at, ds.id)) AS col`` per payload column.
+
+    Note: ``id`` itself is resolved with argMax over ``updated_at`` only (it is
+    part of the version-tiebreaker tuple and constant within a logical key, so
```
gtarpenning: do we need the id tie-breaker if the id will always be identical for two versions tied on updated_at? not sure we need the ID

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,205 @@
...
+    make_dataset_sources_select,
+    make_source_datasets_select,
+)
+
+# ---------------------------------------------------------------------------
+# deterministic_link_id
+# ---------------------------------------------------------------------------
+
+
+def test_deterministic_link_id_is_stable_for_same_logical_key() -> None:
```
gtarpenning: can we combine some of these `deterministic_link_id` tests they all do the same thing

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,398 @@
...
+                ds.source_id,
+                ds.source_trace_id,
+                argMax(ds.created_at, (ds.updated_at, ds.id)) AS created_at,
+                argMax(ds.deleted_at, (ds.updated_at, ds.id)) AS deleted_at
+            FROM dataset_sources AS ds
+            WHERE ds.project_id = {project_param}
+                AND ds.source_id IN {source_ids_param}
+                AND ({exact_match_sql})
+            GROUP BY {group_by_inner}
+            {inner_having}
```
gtarpenning, line 390: is there no order by here intentionally? i think you are truncating right, so this needs an order?

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: Trimmed the architecture note down to two sentences. Fixed in commit 4e7d906288.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: UUIDv5 is name-based — a deterministic hash of (namespace, key) — so the same logical key always yields the same id, which is exactly what lets relink be idempotent without a read-before-write. v4 is random and v7 is time-ordered, so neither is reproducible from the key. Added a comment. Fixed in commit 4e7d906288.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: Those are reStructuredText inline-code markers (the docstring convention used across this module) — they render as monospace in generated docs and have no functional effect.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: Nope different calls: `uuid.UUID("9f2c...")` parses the fixed string into the frozen *namespace* constant, while `uuid.uuid5(namespace, name)` derives a new v5 id by hashing that namespace plus the logical key. The former defines the namespace; the latter consumes it.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: Added a `ValueError` guard for empty `sources` (the CH consumer already short-circuits empty requests, but the builder itself shouldn't be able to emit `AND ()`). Fixed in commit 4e7d906288.

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
jwlee64: Reworked every builder test to assert the full whitespace-normalized SQL with `==`, not substring `in` checks. Fixed in commit 4e7d906288.

weave/trace_server/query_builder/agent_query_builder.py
jwlee64: Added `test_make_spans_existence_query_sql`. Fixed in commit 4e7d906288.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: Added `test_make_link_state_query_sql`. Fixed in commit 4e7d906288.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: Added `test_make_link_state_by_ids_query_sql`. Fixed in commit 4e7d906288.

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
jwlee64: Yes — the reverse-query test now asserts the full SQL including the nested inner SELECT, so the two-level aggregation is covered explicitly. Renamed it to `test_make_source_datasets_select_sql_is_two_level_aggregation` and dropped the `_no_final` name (FINAL-absence is now its own assertion). Fixed in commit 4e7d906288.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: No — it was imported and re-exported but never actually used. Removed the import. Fixed in commit 4e7d906288.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: Agreed — it was a leftover re-export, removed from `__all__`. Fixed in commit 4e7d906288.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: You're right. `id` is a pure function of the logical key (the GROUP BY key), so it's identical for every version in a group and can't break a tie. Dropped it — argMax now tiebreaks on `updated_at` alone — and fixed the docstring that wrongly claimed the id added determinism. Fixed in commit 4e7d906288.

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
jwlee64: Combined the four into one `test_deterministic_link_id` covering stability, per-field distinctness (incl. the D1 same-span/distinct-trace case), and the UUIDv5 check. Fixed in commit 4e7d906288.

weave/trace_server/query_builder/dataset_sources_query_builder.py
jwlee64: Good catch — it was truncating a bare `groupArray(N)`, so the kept subset was arbitrary. Now uses `groupArraySorted(cap)(row_digest)`, which is both deterministic (the N smallest) and memory-bounded during aggregation. Verified the semantics against CH 26.2. Fixed in commit 4e7d906288.

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,411 @@
...
+        GROUP BY {_GROUP_BY}
+        HAVING deleted_at = toDateTime64(0, 3)
+        ORDER BY ds.row_digest ASC, ds.source_kind ASC, ds.source_id ASC,
+            ds.source_trace_id ASC
+        """
+    )
+    assert _norm(query) == expected
+
+    # Dedup via GROUP BY + argMax, never FINAL.
+    assert "final" not in _norm(query).lower()
```
gtarpenning: remove

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,411 @@
...
+
+    # Values are bound via ParamBuilder, not string-interpolated.
+    params = pb.get_params()
+    assert project_val not in query
+    assert dataset_val not in query
```
gtarpenning: remove

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,411 @@
...
+        FROM dataset_sources AS ds
+        WHERE ds.project_id = {{pb_0:String}}
+            AND ds.dataset_object_id = {{pb_1:String}}
+            AND ds.row_digest IN {{pb_2:Array(String)}}
+            AND ds.source_id IN {{pb_3:Array(String)}}
+        GROUP BY {_GROUP_BY}
+        """
+    )
+    assert _norm(query) == expected
+    assert "final" not in _norm(query).lower()
```
gtarpenning: remove

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,411 @@
...
+            {_KEY_SELECT},
+            {_PAYLOAD}
+        FROM dataset_sources AS ds
+        WHERE ds.project_id = {{pb_0:String}}
+            AND ds.id IN {{pb_1:Array(String)}}
+        GROUP BY {_GROUP_BY}
+        """
+    )
+    assert _norm(query) == expected
+    assert "final" not in _norm(query).lower()
```
gtarpenning: remove

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,411 @@
...
+                ds.source_kind, ds.source_id, ds.source_trace_id
+            HAVING deleted_at = toDateTime64(0, 3)
+        )
+        GROUP BY source_kind, source_id, source_trace_id, dataset_object_id
+        ORDER BY source_kind ASC, source_id ASC, source_trace_id ASC,
+            dataset_object_id ASC
+        """
+    )
+    assert _norm(query) == expected
+    assert "final" not in _norm(query).lower()
```
gtarpenning: remove

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,411 @@
...
+
+    # Source / trace / project values are bound, not interpolated.
+    params = pb.get_params()
+    assert "call_123" not in query
+    assert "trace_abc" not in query
+    assert project_val not in query
```
gtarpenning: we should just assert on the actual payload

gtarpenning: in general we shouldn't use `not in` in the sql tests at all

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,411 @@
...
+def test_make_source_datasets_select_include_deleted_drops_inner_having() -> None:
+    pb = ParamBuilder("pb")
+    query = make_source_datasets_select(
+        project_id="proj",
+        sources=[("call", "call_123", "trace_abc")],
+        pb=pb,
+        include_deleted=True,
+        row_digests_cap=100,
+    )
+    assert "having" not in _norm(query).lower()
```
gtarpenning: remove this entire test

tests/trace_server/query_builder/test_dataset_sources_query_builder.py
```diff
@@ -0,0 +1,411 @@
...
+            argMax(spans.started_at, spans.created_at) AS started_at
+        FROM spans
+        WHERE project_id = {pb_0:String}
+          AND trace_id IN {pb_1:Array(String)}
+          AND span_id IN {pb_2:Array(String)}
+        GROUP BY project_id, trace_id, span_id
+        """
+    )
+    assert _norm(query) == expected
+    assert "final" not in _norm(query).lower()
```
gtarpenning: remove

weave/trace_server/query_builder/dataset_sources_query_builder.py
```diff
@@ -0,0 +1,406 @@
...
+# break idempotent relink.
+# The KEY COMPOSITION is frozen too: the set and order of fields hashed into the
+# id (see deterministic_link_id) — including source_trace_id — must not change,
+# or existing links would re-key.
+DATASET_SOURCE_LINK_NAMESPACE = uuid.UUID("9f2c1d3e-7b4a-5c6d-8e9f-0a1b2c3d4e5f")
+
+# Physical column order for INSERTs into dataset_sources. Both backends build
+# rows in exactly this order (the ClickHouse insert and the SQLite insert), so
+# keep this list authoritative.
+DATASET_SOURCES_INSERT_COLUMNS = [
```
gtarpenning, line 37: do we use this?

jwlee64: Yes — the consumer uses it as the column order for both the link and tombstone inserts. Cleaned up the stale comment. Fixed in commit 44d3857fba. Though it is used later up the stack

gtarpenning (approved):
🚢

PR #7239, opened 2026-06-15
chore(weave): dataset_sources trace server interface types + method signatures

jwlee64:
## Description

https://coreweave.atlassian.net/browse/WB-35605

Adds the Dataset Sources API to the trace server interface, which links dataset rows to their provenance sources (calls, agent spans, or conversations).

This introduces four new endpoint definitions following the existing membership pattern (see annotation_queue_items):

- `dataset_sources_link` — links one or more dataset rows to their provenance sources, with optional created-status reporting
- `dataset_sources_link_delete` — soft-deletes dataset source links by ID
- `dataset_sources_query` — forward query from a dataset to its associated sources, filterable by row digests and source kinds
- `source_datasets_query` — reverse query from a set of sources back to the datasets they belong to, returning membership info including row digests (capped at `MAX_ROW_DIGESTS_PER_RESULT`) and first-seen timestamps

Supporting models include `SourceRef`, `SourceKind`, `DatasetSourceLinkPayload`, `DatasetSourceLinkSchema`, and `SourceDatasetMembership`.

PR #7241, opened 2026-06-15
feat(weave): dataset_sources ClickHouse implementation, wiring, and tests

jwlee64:
## Description

- Implements the four `dataset_sources` trace-server methods against ClickHouse + `(+1079/-0)` of tests

- `dataset_sources_link` — link dataset rows → sources (write)
Insert-only: link, relink, and restore are all new versions with a fresh
`updated_at`. Flattens `(row_digest, source)` pairs preserving order and:
  - Enforces the **D4 invariants** — `conversation` links must carry
  `source_trace_id=''` (a conversation spans traces); `call`/`span` links
  require a non-empty trace (their identity is `(trace_id, id)`).
  - **Validates existence** of call/span sources and snapshots their display
  name + started_at; any missing source fails fast with **no partial writes**.
  Conversation links are *asserted* (no backing table → no existence check).
  - With `include_created_status=True`, returns per-link `created` and carries
  `created_at` forward (a relink of a tombstoned link is a **restore**, not a
  create). Caps links per request.

- `dataset_sources_link_delete` — soft-delete by id (write)
 Soft-deletes via insert-only **tombstone** versions (same logical key + id,
fields carried forward, `deleted_at=now`). Fails fast on ids not in the
project; **idempotent** — an already-deleted link returns `deleted=False` and
writes nothing.

- `dataset_sources_query` — dataset → sources (forward read)
Prefix scan on `(project_id, dataset_object_id)` with optional `row_digests` /
`source_kinds` filters and limit/offset. Tombstones are filtered **after**
the version collapse via `HAVING` (so a later tombstone hides an earlier live
version); `include_deleted=True` bypasses it.

- `source_datasets_query` — sources → datasets (reverse read)
Bloom-assisted lookup by `source_id`, then a two-level aggregation: inner
collapses versions per logical key, outer groups by `(source, dataset)` and
returns the row_digests **truncated deterministically** to the N
lexicographically-smallest (`groupArraySorted(N)`) alongside the true total
count + a `truncated` flag. Empty `sources` short-circuits to an empty result.

- `external_to_internal_trace_server_adapter.py`: external→internal `project_id`/`wb_user_id`
translation passthrough for the four methods.

## Testing

Tests: 23 ClickHouse integration tests covering link/relink/dedup, created-status, atomicity,
soft-delete/restore, forward filters (row_digests, source_kinds, limit/offset, include_deleted),
reverse query (membership, truncation, mixed kinds), span sources, and design invariants D1–D4.
`tests/trace/test_client_dataset_sources.py` — integration tests through the
real ClickHouse client covering all four methods: link/relink/restore,
soft-delete + idempotency, missing-source and batch-cap errors, forward
filters + pagination + `include_deleted`, deterministic reverse-query
truncation, and call / span / conversation source kinds (incl. the
same-span-id-distinct-traces and conversation-subset cases).


jwlee64:
* **#7281** <a href="https://app.graphite.com/github/pr/wandb/weave/7281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7241** <a href="https://app.graphite.com/github/pr/wandb/weave/7241?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7241?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7239** <a href="https://app.graphite.com/github/pr/wandb/weave/7239?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7240](https://github.com/wandb/weave/pull/7240) <a href="https://app.graphite.com/github/pr/wandb/weave/7240?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7238** <a href="https://app.graphite.com/github/pr/wandb/weave/7238?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7198** <a href="https://app.graphite.com/github/pr/wandb/weave/7198?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7199](https://github.com/wandb/weave/pull/7199) <a href="https://app.graphite.com/github/pr/wandb/weave/7199?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

tests/trace/test_client_dataset_sources.py
```diff
@@ -0,0 +1,1079 @@
...
+    fq = forward_query(client, dataset_object_id=ds_obj, limit=cap + 100)
+    assert len(fq.links) == 0
+
+
+# ---------------------------------------------------------------------------
+# Soft delete
+# ---------------------------------------------------------------------------
+
+
+def test_dataset_sources_soft_delete_removes_from_forward_and_reverse(client):
```
gtarpenning, line 448: can we consolidate some of the tests? lots of overlap

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -3672,6 +3683,386 @@ def annotator_queue_items_progress_update(
...
+
+        self._insert(
+            "dataset_sources",
+            rows,
+            column_names=DATASET_SOURCES_INSERT_COLUMNS,
+        )
+
+        return tsi.DatasetSourcesLinkRes(entries=entries)
+
+    def _fetch_call_source_fields(
```
gtarpenning: can we move this and refracotr above into its own file, so that most of this business logic stuff is outside of `clickhouse_trace_server_batched.py` its already massive

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -3672,6 +3683,386 @@ def annotator_queue_items_progress_update(
...
+                # Conversation: asserted link, no cached source fields.
+                display_name, started_at = "", now
+
+            link_id = deterministic_link_id(
+                req.project_id,
+                req.dataset_object_id,
+                row_digest,
+                kind,
+                source.source_id,
+                source.source_trace_id,
```
gtarpenning: can we use the trace_id we grabbed from `_fetch_call_source_fields`? bit worried clients might not supply the right thing, or maybe we can just make it simpler for clients

gtarpenning: claude says: 
```python
if source.source_kind == tsi.SourceKind.CALL:
    display_name, started_at, source_trace_id = call_fields[source.source_id]
elif source.source_kind == tsi.SourceKind.SPAN:
    key = (source.source_trace_id, source.source_id)
    display_name, started_at = span_fields[key]
    source_trace_id = source.source_trace_id
else:
    display_name, started_at, source_trace_id = "", now, ""


Then use source_trace_id (not source.source_trace_id) at L3786, L3789 and in the row at L3815.
```

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -3672,6 +3683,386 @@ def annotator_queue_items_progress_update(
...
+            created_at = now
+            if req.include_created_status:
+                prior = prior_state.get(logical_key)
+                # created=False means an existing row (live or tombstoned) was
+                # matched; a soft-deleted link being relinked is a RESTORE, not
+                # a create. created_at carries forward to preserve first_seen.
+                created = prior is None
+                if prior is not None and prior.get("created_at") is not None:
+                    created_at = prior["created_at"]
+            else:
```
gtarpenning: when `include_created_status=False`, `created_at = now` which messes up `first_seen_at` i think... we could do `min(created_at)` in the query

weave/trace_server/external_to_internal_trace_server_adapter.py
```diff
@@ -862,6 +862,51 @@ def annotator_queue_items_progress_update(
...
+        )
+
+    def dataset_sources_query(
+        self, req: tsi.DatasetSourcesQueryReq
+    ) -> tsi.DatasetSourcesQueryRes:
+        req = req.model_copy(deep=True)
+        req.project_id = self._idc.ext_to_int_project_id(req.project_id)
+        if req.wb_user_id is not None:
+            req.wb_user_id = self._idc.ext_to_int_user_id(req.wb_user_id)
+        return self._ref_apply(
```
gtarpenning: need `added_by` translation too! (functional tests don't cover this btw)

gtarpenning: in the res

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -3672,6 +3683,386 @@ def annotator_queue_items_progress_update(
             req.project_id, req.queue_id, req.item_id
         )
 
+    # Dataset Sources API
+    @ddtrace.tracer.wrap(name="clickhouse_trace_server_batched.dataset_sources_link")
+    def dataset_sources_link(
+        self, req: tsi.DatasetSourcesLinkReq
```
gtarpenning, line 3681: what do we do with `dataset_digest`? says audit log only but like, do we audit it?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -3672,6 +3683,386 @@ def annotator_queue_items_progress_update(
...
+                # a create. created_at carries forward to preserve first_seen.
+                created = prior is None
+                if prior is not None and prior.get("created_at") is not None:
+                    created_at = prior["created_at"]
+            else:
+                # Blind write: no lookup ran; created is "not requested" (None)
+                # and created_at is refreshed to now.
+                created = None
+
+            link_metadata_str = json.dumps(meta) if meta else ""
```
gtarpenning:
```suggestion
            link_metadata_str = json.dumps(meta) if meta is not None else ""
```

gtarpenning (approved):

weave/trace_server/clickhouse_trace_server_batched.py
jwlee64: Added the audit log

weave/trace_server/clickhouse_trace_server_batched.py
jwlee64: 👍  CALL links now derive source_trace_id authoritatively from _fetch_call_source_fields (the call record) and ignore the client value; SPAN keeps the client trace (it's the validated lookup key) and CONVERSATION stays ''. Also relaxed the write-invariant so clients no longer need to send source_trace_id for calls. The same derived value flows into deterministic_link_id, the logical key, and the row, so relinks are idempotent regardless of what the client sends. Covered by test_dataset_sources_call_link_uses_authoritative_trace_id (wrong trace, empty trace, relink idempotency).

weave/trace_server/external_to_internal_trace_server_adapter.py
jwlee64: Added test_dataset_sources_forward_query_added_by_is_external, which exercises the non-identity translation via the fixture's base64 id-converter — it asserts the external id, not the stored internal one.

weave/trace_server/clickhouse_trace_server_batched.py
jwlee64: The reverse query's inner aggregation is now min(ds.created_at), so a blind relink (include_created_status=False) no longer inflates first_seen_at. New functional test test_source_datasets_first_seen_at_survives_blind_relink.

jwlee64: Callout from claude for @jwlee64 in the future
Caveat: this is robust pre-merge; after a ReplacingMergeTree merge collapses versions only the max-updated_at row survives, so a blind relink could still skew it post-merge. A fully durable fix would carry created_at forward on blind writes (needs the prior-state lookup the blind path deliberately skips) — happy to do that as a follow-up if we expect blind relinks here.

weave/trace_server/clickhouse_trace_server_batched.py
jwlee64: pulled helpers into dataset_sources.py

PR #4220, opened 2025-04-22
fix(weave): Parse Bedrock inference profile ARN to properly get the Model ID for tracing

michaelromagne:
## Description

Currently, if someone uses an inference profile on AWS Bedrock, cost is not properly tracked by Weave as the model ID is not recognised (see screenshot below).

In order to fix cost tracking, I propose adding a function to get inference profile detail from Bedrock or directly processing the ARN. This way, tracing is fixed.

- Fixes #4218 

## Testing

This PR was tested running `nox -e "tests-3.12(shard='bedrock')"` + I adapted two bedrock converse tests to run with an inference profile ARN

PR #7242, opened 2026-06-16
fix(weave): parse Bedrock inference profile ARN to model ID for invoke api

joshross99:
## Description

Extends #4220 (which fixed the Converse API) to the `invoke_model` path.

When an inference profile ARN is passed as `modelId` to `invoke_model`, Weave logs usage
keyed by the raw ARN. Cost is computed from the usage summary, which is keyed off the
input `modelId` — so the trace-server cost lookup never matches and no cost is shown in
the UI. (The resolved model appears in the response body, but that value is not used for
cost.)

`postprocess_inputs_invoke` now resolves an inference-profile ARN to the underlying model
id before the usage summary is built, reusing `extract_model_name_from_inference_profile_arn`
— the same helper the Converse path already uses. Behavior is unchanged for plain model-id
inputs.

## Testing

- `uv run --group test python -m pytest tests/integrations/bedrock/bedrock_test.py` — 9 passed.
- Parametrized `test_bedrock_invoke` over a plain model id and an inference-profile ARN
  (mirrors the existing Converse tests). Also fixed a shared mock-body-stream leak the
  parametrization exposed — each invoke call now gets a fresh response-body stream.
- Verified end-to-end against wandb.ai SaaS: invoked a real application inference-profile
  ARN through the patched SDK and confirmed cost now renders in the Weave UI (was blank
  before the fix).

🤖 AI-assisted (Claude Code); reviewed and validated by the author.

### DIFF
--- a/tests/integrations/bedrock/bedrock_test.py
+++ b/tests/integrations/bedrock/bedrock_test.py
@@ -101,29 +101,24 @@ MOCK_STREAM_EVENTS = [
     },
 ]
 
-MOCK_INVOKE_RESPONSE = {
-    "body": io.BytesIO(
-        json.dumps(
+MOCK_INVOKE_BODY = json.dumps(
+    {
+        "id": "msg_bdrk_01WpFc3918C93ZG9ZMKVqzCd",
+        "type": "message",
+        "role": "assistant",
+        "model": "claude-3-5-sonnet-20240620",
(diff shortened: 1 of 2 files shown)


joshross99:

jwlee64 (approved):
nice lgtm

PR #7245, opened 2026-06-16
feat(weave): add Claude Agent SDK integration for the TypeScript SDK

rgao-coreweave:
## Summary

Adds a Weave integration for the **Claude Agent SDK** (`@anthropic-ai/claude-agent-sdk`) to the TypeScript SDK, at parity with the Python `claude_agent_sdk` integration.

The SDK has no trace-processor hook and runs the model in a spawned Claude Code subprocess, so the `@anthropic-ai/sdk` patch never sees those calls. Instead we wrap the SDK's exported `query()` and emit **GenAI agent spans** from the streamed `SDKMessage`s through the shared Weave GenAI tracer (targeting `/agents/otel/v1/traces`), surfacing in the **Agents tab** — the same substrate as the pi-coding-agent and OpenAI-Agents OTel integrations.

## Trace shape

```
invoke_agent claude_agent_sdk     (root)        inputs: prompt
├─ chat <model>                   (per assistant message)  output.messages = text / reasoning / tool_call parts, finish reasons, model
└─ execute_tool <name>            (per tool_use)           closed by the matching tool_result; is_error → ERROR
```

- Every span carries `gen_ai.conversation.id` (the SDK `session_id`) so a session's turns group together.
- Per-model usage, `total_cost_usd`, `num_turns`, and the final result are lifted onto the root `invoke_agent` span.
- Error paths set ERROR status (thrown stream error, non-`success` terminal subtype, tool result flagged `is_error`); open tool spans are swept on finalize. Tracing failures are swallowed so instrumentation never breaks the caller's stream.

Auto-instrumented via the existing CJS/ESM module-loader hooks (pinned to `>= 0.3.178`); `wrapClaudeAgentSdk` is also exported for manual instrumentation.

## Changes

- `src/integrations/claude-agent-sdk/{otelTracer,messages}.ts` — maps the message stream to agent spans; local SDK message types + camelCase→snake_case usage mapping.
- `src/integrations/claudeAgentSdk.ts` — `query()` wrapper, `patchClaudeAgentSdk`, auto-instrumentation registration.
- `src/integrations/hooks.ts`, `src/index.ts` — register and export.
- `examples/claudeAgents.ts` — a multi-prompt transcript demo.
- `package.json` — adds `@anthropic-ai/claude-agent-sdk` as a devDependency (the integration source imports types from it; the example + host-app e2e also consume it).

## Testing

- Unit tests assert the span tree, attributes, parent linkage, conversation id, usage aggregation, error paths, and the `query()`-wrapping behaviors against an in-memory OTel exporter.
- The real-package **hostApps end-to-end** test — runs the actual `query()` under `node --import=weave/instrument` against a fake Claude Code executable, fully offline — is added in the stacked follow-up #7399.

## Out of scope (follow-up)

- Multi-turn / streaming-input sessions (`query({prompt: <AsyncIterable>})` / `Query.streamInput`) get a single root turn rather than one per turn.
- Per-`chat` usage and input messages: the SDK only reports aggregate usage in the terminal result (lifted onto the root) and the stream doesn't expose per-turn input context.

## Reviewer note

`@anthropic-ai/claude-agent-sdk` is a devDependency because the integration source imports types from it (`claudeAgentSdk.ts`, `otelTracer.ts`, `messages.ts`); it bundles the Claude Code CLI (platform binaries), so it adds notable install weight — flagging in case you'd prefer to vendor the handful of types instead.


AGENTS.md
```diff
@@ -214,10 +214,28 @@ nox --no-install -e "tests-3.12(shard='langchain')" -- tests/integrations/langch
 
 ## Typescript Testing Guidelines
 
+The Node SDK (`sdks/node`) is a **pnpm** project — it ships a `pnpm-lock.yaml`
+and pins `"packageManager": "pnpm@10.8.1"` in `package.json`. Do **not** run
+`npm i`: npm's resolver crashes trying to dedupe pnpm's symlink `node_modules`
+(`TypeError: Cannot read properties of null (reading 'matches')`).
+
+Use `corepack pnpm`, which reads the pinned version and works regardless of any
```
andrewtruong: why do we need this?  AFAIK we have never used this

AGENTS.md
```diff
@@ -214,10 +214,28 @@ nox --no-install -e "tests-3.12(shard='langchain')" -- tests/integrations/langch
...
+corepack pnpm install
+corepack pnpm run test
+```
+
+To run an example (e.g. the Claude Agent SDK demo), `dist/` must be built first
+(`import 'weave'` self-resolves via the package `exports` to `dist/index.mjs`);
+`pnpm install` builds it via the `prepare` script. Then:
+
+```
+corepack pnpm exec tsx examples/claudeAgents.ts
```
drtangible: We might want to just update the imports in our examples to...

```ts
import * as weave from '../src';
```

And remove the documentation around building.

drtangible: Might be a good idea to add corepack to our dev env setup / instructions, but think we should remove it from here until we intentionally add it.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,429 @@
...
+    return String(content);
+  }
+}
+
+/**
+ * Sum the per-model `modelUsage` (or fall back to the aggregate `usage`) into
+ * Weave's snake_case token keys. Reuses {@link toWeaveUsage} so the
+ * camelCase→snake_case handling stays in one place.
+ */
+function aggregateUsage(result: SDKResultMessage): Record<string, number> {
```
andrewtruong: Do we need this?  The server should be doing aggregations, not the client

sdks/node/src/integrations/claudeAgentSdk.ts
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Register automatic instrumentation for `@anthropic-ai/claude-agent-sdk`.
+ * Called once from `integrations/hooks.ts`.
+ */
+export function instrumentClaudeAgentSdk(): void {
+  // Floor pinned to the version whose message/usage shape this integration was
+  // validated against (camelCase `modelUsage`, result `subtype` enum, structured
+  // `system` messages) — the same version carried as a devDependency. Older
+  // 0.3.x builds pass through untraced rather than risk a silently-wrong shape.
```
drtangible: This comment is a bit coupled to technical details, but describes an important version requirement. I wonder if we can omit this and enumerate the version requirement in a comment that would be visible to an end-user? (For example, if we hoist up to the JSDoc for `instrumentClaudeAgentSdk`, it'd show up in the reference docs and in the IDE if the user hovered over the fn).

sdks/node/src/integrations/claudeAgentSdk.ts
```diff
@@ -0,0 +1,191 @@
...
+      return Reflect.get(target, prop, receiver);
+    },
+  });
+}
+
+/**
+ * Register automatic instrumentation for `@anthropic-ai/claude-agent-sdk`.
+ * Called once from `integrations/hooks.ts`.
+ */
+export function instrumentClaudeAgentSdk(): void {
```
drtangible, line 223: I believe our other integrations add these functions to our top-level API, so that users can manually instrument libraries should they run into issues w/our automatic instrumentation.

sdks/node/src/integrations/claudeAgentSdk.ts
```diff
@@ -0,0 +1,191 @@
+/**
+ * Weave integration for the Claude Agent SDK (`@anthropic-ai/claude-agent-sdk`).
+ *
+ * The SDK has no trace-processor hook (unlike `@openai/agents`) and runs the
+ * model inside a spawned Claude Code subprocess, so the `@anthropic-ai/sdk`
+ * patch never sees those calls. We therefore wrap the SDK's exported `query()`
+ * and emit GenAI agent spans from the streamed messages through the shared
+ * Weave GenAI tracer — `invoke_agent`/`chat`/`execute_tool` spans to the
+ * `/agents/otel` endpoints (the Agents tab). See {@link ClaudeAgentOtelTracer}.
```
drtangible: It looks like the Agent SDK has a concept of [hooks](https://code.claude.com/docs/en/agent-sdk/hooks). Did we explore leveraging those rather than monkey-patching internals?

AGENTS.md
rgao-coreweave: Fixed the local/ci env so we don't need to use corepack

AGENTS.md
rgao-coreweave: dropped corepack by fixing the dev/ci env.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Good callout that the server already does the aggregation, we should drop the summing here.

thanks!

sdks/node/src/integrations/claude-agent-sdk/messages.ts
```diff
@@ -0,0 +1,143 @@
+/**
+ * Local types and the usage-mapping helper for the Claude Agent SDK
+ * integration: a minimal mirror of the `@anthropic-ai/claude-agent-sdk` message
+ * shapes the tracer consumes.
+ *
+ * We deliberately avoid depending on `@anthropic-ai/claude-agent-sdk` for these
+ * types — the integration patches the SDK's exported `query()` at runtime, the
+ * same way `anthropic.ts` patches `@anthropic-ai/sdk` without importing it — so
+ * only the subset of fields used by the tracer is modelled here.
+ */
```
drtangible, line 8: We've been updating integration to use actual types -- assume we can pull these from `@anthropic-ai/claude-agent-sdk`?

AGENTS.md
rgao-coreweave: dropped the `corepack`, so this is not needed. thanks.

sdks/node/src/integrations/claudeAgentSdk.ts
```diff
@@ -0,0 +1,191 @@
...
+ * single root turn rather than one per turn.
+ */
+import {getGlobalClient} from '../clientApi';
+import {addCJSInstrumentation, addESMInstrumentation} from './instrumentations';
+import {ClaudeAgentOtelTracer} from './claude-agent-sdk/otelTracer';
+import type {SDKMessage, SDKResultMessage} from './claude-agent-sdk/messages';
+
+// Marks an exports object whose `query` we've already wrapped, so repeated hook
+// invocations (e.g. CJS + ESM both firing) don't double-wrap.
+const PATCHED = Symbol.for('weave.claudeAgentSdk.patched');
```
drtangible: We have similar module-level state abstracted away in `state.ts`. Is it possible to follow suit with this?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,473 @@
...
+  const output = usage.output_tokens;
+  const cacheRead = usage.cache_read_input_tokens;
+  const cacheCreation = usage.cache_creation_input_tokens;
+  if (typeof input === 'number') {
+    attrs[ATTR_GEN_AI_USAGE_INPUT_TOKENS] = input;
+  }
+  if (typeof output === 'number') {
+    attrs[ATTR_GEN_AI_USAGE_OUTPUT_TOKENS] = output;
+  }
+  if (typeof cacheRead === 'number') {
```
drtangible: Could `toWeaveUsage` produce a map with known types? It looks like it's aware of the possible keys it produces.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,473 @@
...
+  if (typeof cacheCreation === 'number') {
+    attrs[ATTR_GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS] = cacheCreation;
+  }
+  if (typeof input === 'number' && typeof output === 'number') {
+    attrs[ATTR_GEN_AI_USAGE_TOTAL_TOKENS] = input + output;
+  }
+  return attrs;
+}
+
+export interface ClaudeAgentOtelTracerOptions {
```
drtangible: Is this used externally?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,473 @@
...
+        ATTR_GEN_AI_INPUT_MESSAGES,
+        JSON.stringify([{role: 'user', content: opts.prompt}])
+      );
+    }
+  }
+
+  processMessage(msg: SDKMessage): void {
+    // The SDK stamps session_id on every message; capture it once as the
+    // conversation id so the root and its children all carry it.
+    const sessionId = (msg as {session_id?: string}).session_id;
```
drtangible: If this is on every message, is there a reason its not present on `SDKMessage`? Is the coercion necessary?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,473 @@
...
+    if (this.conversationId == null && sessionId) {
+      this.conversationId = sessionId;
+      this.invokeAgentSpan.setAttribute(ATTR_GEN_AI_CONVERSATION_ID, sessionId);
+    }
+
+    switch (msg.type) {
+      case 'assistant':
+        this.processAssistant(msg as SDKAssistantMessage);
```
drtangible: Is the check on `type === 'assistant'` not already narrowing this to `SDKAssistantMessage`?

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: Good point! Added `wrapClaudeAgentSdk`

Thanks for the feedback.

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: Good question, I took a look at the hooks, but hooks are partial as they can't produce LLM chat:  a hook callback only sees `tool_name/tool_input/session_id`, but never the `assistant message content, model, token usage, or total_cost_usd`.

So therefore we are wrapping the public exported `query()`.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Not used externally, dropped the `export`

sdks/node/src/integrations/claude-agent-sdk/messages.ts
rgao-coreweave: Good call, swapped hand rolled to pulling exported types from claude agent.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Good call, dropped the type cast.

thank you!

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: done. thank you!

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: `state.ts`'s flags (openaiAgents.instrumented, openaiAgentsRealtime.patched) are process-wide one-time booleans. PATCHED is a per-object idempotency marker, and patchClaudeAgentSdk runs on distinct exports objects.

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: Good call, I put it on `wrapClaudeAgentSdk` rather than `instrumentClaudeAgentSdk`, since that's the public entry point users actually call.

sdks/node/src/integrations/claudeAgentSdk.ts
drtangible: That's a bummer! I see `MessageDisplay - An assistant message with text completes, once per message with the full message text` in https://code.claude.com/docs/en/agent-sdk/hooks#available-hooks. But maybe usage is not yet exposed through any of the available hooks?

In any case, since this is that library's blessed way of hooking into agent events, seems worth us adding a link to it in a comment in our implementation? That way we can explain why we're not using it, and keep tabs on whether functionality we'd want is eventually added to the hooks. (On that note -- worth adding an Issue to their repo requesting what we're looking for?)

sdks/node/src/integrations/claudeAgentSdk.ts
drtangible: Ah -- ok, so we need to keep track of multiple distinct `exports` as patched? Looks like this pattern is consistent with another integration or two in the codebase, but at some point I might lift that up so we're keeping track of state separately. (eg, we could have a `state.integrations.claudeAgents.patchedExports` that's a `WeakMap` keyed by the the objects, rather than mutating them directly).

sdks/node/src/integrations/claudeAgentSdk.ts
```diff
@@ -0,0 +1,236 @@
...
+    },
+  });
+}
+
+/**
+ * Manually instrument the `@anthropic-ai/claude-agent-sdk` module.
+ *
+ * Requires `@anthropic-ai/claude-agent-sdk` >= 0.3.178 — the message/usage
+ * shape this integration was validated against (camelCase `modelUsage`, the
+ * result `subtype` enum, structured `system` messages). Automatic
```
drtangible: Since this is part of the public API, I believe it'll show up on reference docs in docs.wandb.ai.

For someone reading that just trying to understand how to instrument `@anthropic-ai/claude-agent-sdk`, referring to some of these implementation details might be confusing:

> (camelCase `modelUsage`, the result `subtype` enum, structured `system` messages)

sdks/node/src/integrations/claudeAgentSdk.ts
```diff
@@ -0,0 +1,236 @@
...
+ * instrumentation passes older builds through untraced; this manual path does
+ * not gate on the version, so wrapping an older build risks a wrong span shape.
+ *
+ * Reach for this when automatic instrumentation doesn't apply — e.g. a bundler
+ * whose module loading the CJS/ESM hooks can't observe, or an import path they
+ * don't cover. Returns a view of the module whose `query` export is traced; use the
```
drtangible: Similarly, this might be a bit technical for a reader of the docs -- not a blocker, but think we might eventually benefit from linking folks to other documentation describing auto vs. manual instrumentation.

sdks/node/examples/claudeAgents.ts
```diff
@@ -0,0 +1,161 @@
...
+  return oneLine.length > max ? `${oneLine.slice(0, max)}…` : oneLine;
+}
+
+/** Render a tool_result's content (string or content-block array) as a string. */
+function stringifyToolResult(content: unknown): string {
+  if (content == null) return '';
+  return typeof content === 'string' ? content : JSON.stringify(content);
+}
+
+interface TurnResult {
```
drtangible: Assuming this could be a simple `type`.

sdks/node/examples/claudeAgents.ts
```diff
@@ -0,0 +1,161 @@
...
+const MODEL = process.env.ANTHROPIC_MODEL || 'claude-sonnet-4-5';
+
+/** Collapse whitespace and clip long strings so the transcript stays readable. */
+function clip(text: string, max = 120): string {
+  const oneLine = text.replace(/\s+/g, ' ').trim();
+  return oneLine.length > max ? `${oneLine.slice(0, max)}…` : oneLine;
+}
+
+/** Render a tool_result's content (string or content-block array) as a string. */
+function stringifyToolResult(content: unknown): string {
```
drtangible: Might be able to type this with `function stringifyToolResult(content: string | object | undefined): string {`

sdks/node/src/integrations/claudeAgentSdk.ts
```diff
@@ -0,0 +1,236 @@
...
+// import — each needs its `query` wrapped exactly once, so the marker must
+// travel with the object rather than gate all of them on one boolean. Using
+// Symbol.for keeps the marker identical across duplicate CJS/ESM copies of this
+// module (the dual-package hazard state.ts otherwise handles via globalSingleton).
+const PATCHED = Symbol.for('weave.claudeAgentSdk.patched');
+
+/** Minimum `@anthropic-ai/claude-agent-sdk` version this integration supports. */
+const SUPPORTED_VERSION_RANGE = '>= 0.3.178';
+
+type QueryFn = (args: {prompt?: unknown; [k: string]: unknown}) => unknown;
```
drtangible: Might be able to more precisely type this as...

```ts
import type * as ClaudeAgentsSDK from "@anthropic-ai/claude-agent-sdk";

type QueryFn = typeof ClaudeAgentsSDK["query"];
```

sdks/node/src/integrations/claudeAgentSdk.ts
```diff
@@ -0,0 +1,236 @@
...
+      let streamError: unknown;
+      try {
+        for await (const msg of realQuery as AsyncIterable<SDKMessage>) {
+          if (msg && (msg as SDKMessage).type === 'result') {
+            result = msg as SDKResultMessage;
+          } else {
+            // Tracing must never break the caller's stream: swallow any
+            // mapping error (the message is still yielded below).
+            try {
+              tracer.processMessage(msg as SDKMessage);
```
drtangible: When properly typing `QueryFn`, I think a lot of the `as Foo` are not necessary:

Image

sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  name: string;
+  spanId: string;
+  parentSpanId?: string;
+  traceId: string;
+  attributes: Record<string, unknown>;
+  statusCode: number;
+}
+
+describe('hostApps — claude-agent-sdk', () => {
+  test('auto-instruments the real @anthropic-ai/claude-agent-sdk query() and emits a GenAI agent span tree', async () => {
```
drtangible: Cool -- we don't yet have other of these `hostApps` tests for specific integrations. As we add more, I think we could eventually document how much we want to mock at this layer, and how. (ie, patching vs recording and playing back network activity, etc). Just a comment for myself for later.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+): Record<string, number> {
+  const usage = toWeaveUsage(rawUsage);
+  const attrs: Record<string, number> = {};
+  const input = usage.input_tokens;
+  const output = usage.output_tokens;
+  const cacheRead = usage.cache_read_input_tokens;
+  const cacheCreation = usage.cache_creation_input_tokens;
+  if (typeof input === 'number') {
+    attrs[ATTR_GEN_AI_USAGE_INPUT_TOKENS] = input;
+  }
```
drtangible: Now that these are typed, a nullish check might be a bit more concise, since it's a `number` if it passes?

```ts
  if (input) {
    attrs[ATTR_GEN_AI_USAGE_INPUT_TOKENS] = input;
  }
```

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+  if (typeof cacheCreation === 'number') {
+    attrs[ATTR_GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS] = cacheCreation;
+  }
+  if (typeof input === 'number' && typeof output === 'number') {
+    attrs[ATTR_GEN_AI_USAGE_TOTAL_TOKENS] = input + output;
+  }
+  return attrs;
+}
+
+interface ClaudeAgentOtelTracerOptions {
```
drtangible: Assuming this could be a simple `type`

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+  /** The user prompt, when invoked as a string (recorded as input on the root). */
+  prompt?: string;
+  /** Tracer override (tests inject one backed by an in-memory exporter). */
+  tracer?: Tracer;
```
drtangible: Where is this option being used?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+    // prompts in one session are linked via gen_ai.conversation.id, not a
+    // shared parent.
+    this.invokeAgentSpan = this.tracer.startSpan(
+      `invoke_agent ${AGENT_NAME}`,
+      {
+        kind: SpanKind.INTERNAL,
+        attributes: {
+          ...CLAUDE_AGENT_SDK_OTEL_ATTRS,
+          [ATTR_GEN_AI_OPERATION_NAME]: 'invoke_agent',
+          [ATTR_GEN_AI_AGENT_NAME]: AGENT_NAME,
```
drtangible: I believe we currently have a requirement that this attribute be present on child spans, otherwise some data will not show up in the Weave UI. (eg, input / output token counts). Are we seeing that behavior with this integration?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+      );
+    }
+
+    if (result) {
+      // Emit per-model usage on child spans and let the trace server cost and
+      // roll it up, rather than summing on the client. See
+      // {@link emitModelUsageSpans}.
+      this.emitModelUsageSpans(result);
+
+      if (result.total_cost_usd != null) {
```
drtangible: It appears from the type that `total_cost_usd` is always a `number`?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+        this.invokeAgentSpan.setAttribute(ATTR_COST_USD, result.total_cost_usd);
+      }
+      if (result.num_turns != null) {
```
drtangible: Same on this one.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+        this.invokeAgentSpan.setAttribute(ATTR_NUM_TURNS, result.num_turns);
+      }
+      // `result` is only on the success member of the SDKResultMessage union.
+      if ('result' in result && result.result != null) {
```
drtangible: By the time the null check is happening, it looks like `result.result` is already narrowed to `string`.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+        break;
+      default:
+        break;
+    }
+  }
+  return parts;
+}
+
+/** Stringify tool-result content (string, content-block array, or other). */
+function toolResultText(content: unknown): string {
```
drtangible: Similar comment to https://github.com/wandb/weave/pull/7245#discussion_r3461023457. We could technically probably get at the type by doing something like...


```ts
type ToolResultBlock = Extract<
  SDKUserMessage['message']['content'][number],
  {type: 'tool_result'}
>;

function toolResultText(content: ToolResultBlock["content"]): string {
```

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    // A thrown stream error, or a non-success terminal subtype, marks the root
+    // as failed — broadening the Python integration's `is_error`-only check so
+    // failed runs (error_max_turns, error_during_execution, …) surface as errors.
+    const errored =
+      error != null ||
+      (result != null &&
+        (result.is_error ||
+          (result.subtype != null && result.subtype !== 'success')));
```
drtangible: Do we need this null check?

Image

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+        );
+      }
+    }
+
+    // A thrown stream error, or a non-success terminal subtype, marks the root
+    // as failed — broadening the Python integration's `is_error`-only check so
+    // failed runs (error_max_turns, error_during_execution, …) surface as errors.
+    const errored =
+      error != null ||
+      (result != null &&
```
drtangible: nit: can probably drop the `!= null` here.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+        (result.is_error ||
+          (result.subtype != null && result.subtype !== 'success')));
+    if (errored) {
+      // `result`/`errors` live on opposite members of the SDKResultMessage
+      // union — narrow with `in` before reading either.
+      const resultText =
+        result && 'result' in result ? result.result : undefined;
```
drtangible: A more idiomatic way of narrowing this type might be to check against the `subtype`:

```ts
      const resultText =
        result && result.subtype === "success" ? result.result : undefined;
```

Then we can drop the comment above.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+      const errorsText =
+        result && 'errors' in result ? result.errors.join('; ') : undefined;
```
drtangible: Similarly,

```ts
      const errorsText =
        result && result.subtype !== "success" ? result.errors.join('; ') : undefined;
```

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+
+  // The conversation/session id, attached to child spans only once it's known.
+  private conversationAttrs(): Attributes {
+    return this.conversationId
+      ? {[ATTR_GEN_AI_CONVERSATION_ID]: this.conversationId}
+      : {};
+  }
+
+  // A `chat` span (CLIENT, child of the root) keyed by model. `extraAttrs`
+  // carries the per-message response (processAssistant) or the per-model usage
```
drtangible: It looks like `processAssistant` doesn't pass anything for `extraAttrs` -- is this comment accurate?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,483 @@
...
+    const chatSpan = this.startChatSpan(model);
+    const parts = assistantParts(content);
+    if (parts.length > 0) {
+      const outputMessage: Message = {role: 'assistant', parts};
+      chatSpan.setAttribute(
+        ATTR_GEN_AI_OUTPUT_MESSAGES,
+        JSON.stringify([outputMessage])
+      );
+    }
+    if (msg.message?.stop_reason) {
```
drtangible: Do we need this guard?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Good call!

Dropped the redundant check

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: dropped the redundant check here.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Good call, dropped the null check.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: message is required, dropped `?.`.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Used in tests, but dead code. Removed both.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: true,  subtype always present, dropped the null check.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Switched to type. thanks!

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: input is a number, if so input = 0, `if (input)` gets dropped. 

Updated with null check `if (input) != null`. thanks!

sdks/node/examples/claudeAgents.ts
rgao-coreweave: Sounds good. thank you!

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: You're right, it was redundant. It was never null after typing it.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: thank you! Narrowed errorsText on subtype.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: checking `subtype`, thanks.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: interesting, never seen this trick to get typed param. thanks!

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: yes, `extraAttrs is used only by emitModelUsageSpans`. processAssistant doesn't have extraAttrs because it doesn't carry token usage.

sdks/node/examples/claudeAgents.ts
rgao-coreweave: good call. thank you!

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: true, thanks for the styling guide!

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: good trick, thank you!

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: good call, will stamp gen_ai.agent.name onto the child spans.

rgao-coreweave: image

AGENT_NAME is properly showing up though

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: Good point, the impl-detail is moved out of JSDoc and moved to a plain // comment (won't render on docs.wandb.ai).

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: Good point, the impl-detail is moved out of JSDoc and moved to a plain // comment (won't render on docs.wandb.ai).

sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
rgao-coreweave: true, host apps with fake clis provides a good test coverage.

rgao-coreweave:
Spans are populated correctly.

image


https://wandb.ai/wandb/example/weave/agents/conversations/6b9af871-f396-4e30-a78b-2fad4a2ad665

sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
andrewtruong, line 1: what is `hostApps`?

sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
```diff
@@ -0,0 +1,94 @@
+import {randomUUID} from 'crypto';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+import {fixturePath, genProjectId, launchAppFrom} from './utils';
+
+interface CapturedSpan {
```
drtangible: believe this could be a simple `type`

sdks/node/src/integrations/claude-agent-sdk/messages.ts
```diff
@@ -0,0 +1,73 @@
...
+  SDKMessage,
+  SDKResultMessage,
+  SDKUserMessage,
+  SDKUserMessageReplay,
+} from '@anthropic-ai/claude-agent-sdk';
+
+// ── Usage mapping ───────────────────────────────────────────────────────
+
+/** Weave's snake_case usage shape — the only keys {@link toWeaveUsage} emits. */
+interface WeaveUsage {
```
drtangible: Could be a simple `type`

sdks/node/src/integrations/claude-agent-sdk/messages.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * The message shapes the tracer consumes — `SDKMessage` and the members it
+ * reads (`SDKAssistantMessage`, `SDKUserMessage`, `SDKResultMessage`) — are
+ * re-exported straight from `@anthropic-ai/claude-agent-sdk` rather than
+ * hand-rolled, so the integration tracks the SDK's own definitions instead of a
+ * mirror that can silently drift (this matches the OpenAI-Agents integration,
+ * which pulls its span/usage types from `@openai/agents`). These are
+ * `import type` re-exports — erased at compile time, so they add no runtime
+ * import or dependency and don't change how the integration patches the SDK's
+ * exported `query()` at runtime.
```
drtangible: nbd, but comments referencing implementation details and past decisions make things a bit harder to read and review. For things like this that are cross-integration concepts, wonder if documenting somewhere higher level makes sense.

sdks/node/src/integrations/claude-agent-sdk/messages.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+import type {ModelUsage} from '@anthropic-ai/claude-agent-sdk';
+
+export type {
+  SDKAssistantMessage,
+  SDKMessage,
+  SDKResultMessage,
+  SDKUserMessage,
+  SDKUserMessageReplay,
+} from '@anthropic-ai/claude-agent-sdk';
```
drtangible: IMO, we can import directly from the library wherever we need types, rather than re-exporting. Would make the comment a few lines above unnecessary, too.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,507 @@
...
+  SDKUserMessage['message']['content'][number],
+  {type: 'tool_result'}
+>;
+
+/** Stringify tool-result content (string, content-block array, or other). */
+function toolResultText(content: ToolResultBlock['content']): string {
+  if (typeof content === 'string') {
+    return content;
+  }
+  if (Array.isArray(content)) {
```
drtangible:
```ts
function toolResultText(content: ToolResultBlock['content']): string {
  if (!content) {
    return ''
  };

  if (typeof content === 'string') {
    return content;
  }

  // ...
```

Nit -- if we had an early return at the top to guard against the nullish case, this would be an array already.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,507 @@
...
+    const text = content
+      .map(block => ('text' in block ? String(block.text) : ''))
```
drtangible: Would checking `block.type === "text"` narrow to the only type we care about here?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,507 @@
...
+   * that descendant rollup; the SDK's authoritative `total_cost_usd` still rides
+   * on the root as a reference attribute. Sourcing usage from the result's
+   * `modelUsage` (rather than per-assistant-message usage) also captures models
+   * the SDK used without emitting an assistant message — e.g. a fast model for
+   * an internal step.
+   */
+  private emitModelUsageSpans(result: SDKResultMessage): void {
+    // Prefer the per-model breakdown; fall back to the flat aggregate keyed by
+    // the session's primary model when the SDK reports no `modelUsage`.
+    const perModel: Array<[string | undefined, Record<string, unknown>]> =
```
drtangible: Could we type this as...

```ts
    const perModel: Array<[string | undefined, ModelUsage | NonNullableUsage]> =
```

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,507 @@
...
+/**
+ * Map one model's SDK usage object to the scalar `gen_ai.usage.*` span
+ * attributes. {@link toWeaveUsage} normalizes both the camelCase per-model
+ * `modelUsage` values and the snake_case aggregate `usage`.
+ *
+ * `input_tokens` is emitted as the FULL prompt — fresh + cache-read +
+ * cache-creation — see the folding note below.
+ */
+function usageAttributes(
+  rawUsage: Record<string, unknown>
```
drtangible: If we do the aforementioned typing, we can probably make this (and `toWeaveUsage` more precise).

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,507 @@
...
+  }
+  if (input != null && output != null) {
+    attrs[ATTR_GEN_AI_USAGE_TOTAL_TOKENS] = input + output;
+  }
+  return attrs;
+}
+
+type ClaudeAgentOtelTracerOptions = {
+  /** The user prompt, when invoked as a string (recorded as input on the root). */
+  prompt?: string;
```
drtangible, line 209: Is this ever constructed without a `prompt`?

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,507 @@
...
+      const errorsText =
+        result && result.subtype !== 'success'
+          ? result.errors.join('; ')
+          : undefined;
+      const message =
+        error != null
+          ? error instanceof Error
+            ? error.message
+            : String(error)
+          : resultText || errorsText || 'Conversation ended with error';
```
drtangible: I'm not totally sure this is equivalent, but could we possibly combine the check for an error state w/the construction of the message to make this more readable? somthing like..

```ts
function getError({
  error,
  result,
}: {
  error: unknown;
  result: SDKResultMessage | undefined;
}): {message: string} | undefined {
  if (error instanceof Error) {
    return {message: error.message};
  }

  if (error) {
    return {message: String(error)};
  }

  if (result && result.subtype !== 'success') {
    return {message: result.errors.join('; ')};
  }

  if (result && result.is_error) {
    return {message: result.result};
  }
}

  // ...
// ...

// A thrown stream error, or a non-success terminal subtype, marks the root
// as failed — broadening the Python integration's `is_error`-only check so
// failed runs (error_max_turns, error_during_execution, …) surface as errors.
const err = getError();
if (err) {
  this.invokeAgentSpan.setAttribute(ATTR_ERROR_TYPE, 'agent_error');
  this.invokeAgentSpan.setStatus({
    code: SpanStatusCode.ERROR,
    message: err.message || 'Conversation ended with error',
  });
}

File: sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts [hunk 44]
```diff
@@ -0,0 +1,507 @@
...
+  toWeaveUsage,
+  type SDKAssistantMessage,
+  type SDKMessage,
+  type SDKResultMessage,
+  type SDKUserMessage,
+  type SDKUserMessageReplay,
+} from './messages';
+
+const TRACER_NAME = 'claude-agent-sdk';
+const AGENT_NAME = 'claude_agent_sdk';
```
drtangible, line 72: Do we want agent name to always be this?

It looks like the SDK takes an `agent`, and `agents` option:

```ts
    /**
     * Agent name for the main thread. When specified, the agent's system prompt,
     * tool restrictions, and model will be applied to the main conversation.
     * The agent must be defined either in the `agents` option or in settings.
     *
     * This is equivalent to the `--agent` CLI flag.
     *
     * @example
     * ```typescript
     * agent: 'code-reviewer',
     * agents: {
     *   'code-reviewer': {
     *     description: 'Reviews code for best practices',
     *     prompt: 'You are a code reviewer...'
     *   }
     * }
     * ```
     */
    agent?: string;

    /**
     * Programmatically define custom subagents that can be invoked via the Agent tool.
     * Keys are agent names, values are agent definitions.
     *
     * @example
     * ```typescript
     * agents: {
     *   'test-runner': {
     *     description: 'Runs tests and reports results',
     *     prompt: 'You are a test runner...',
     *     tools: ['Read', 'Grep', 'Glob', 'Bash']
     *   }
     * }
     * ```
     */
```

sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
rgao-coreweave: https://github.com/wandb/weave/blob/master/sdks/node/src/__tests__/hostApps/README.md

"Host app" is the application that hosts the library as an installed dependency

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    switch (msg.type) {
+      case 'assistant':
+        // `type: 'assistant'` already narrows msg to SDKAssistantMessage.
+        this.processAssistant(msg);
+        break;
+      case 'user':
+        // `type: 'user'` narrows to SDKUserMessage | SDKUserMessageReplay; both
+        // carry `message: MessageParam`, so processUser reads either.
```
drtangible: nit -- these comments are kinda describing stuff that the types themselves communicate.

sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
rgao-coreweave: done, changed to a simple type.

sdks/node/src/integrations/claude-agent-sdk/messages.ts
rgao-coreweave: changed to a simple type

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: done, thx!

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: done, thx!

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: yes, changed to `block.type === 'text' `

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: nice!

sdks/node/src/integrations/claude-agent-sdk/messages.ts
rgao-coreweave: Trimmed the docstring down to just what the file does

rgao-coreweave: dropped the past-decision and cross-integration notes. thx

sdks/node/src/integrations/claude-agent-sdk/messages.ts
rgao-coreweave: good call, dropped the re-export. Importing the SDK message types directly at each use sites

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Good call, `claude_agent_sdk` should only be a fallback. set `options.agent` should be wired up.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: removed these verbose comments.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Yep, with streaming input (query({prompt: <AsyncIterable>})), there's no string prompt since it is optional.

sdks/node/src/integrations/claudeAgentSdk.ts
rgao-coreweave: it is weird that they don't have usage on hooks. even for claude code plugin, we had to get usage from transcript because they were not emitted on hooks.

Good point on adding a header comment linking the hooks docs.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: thanks, much cleaner!

sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
rgao-coreweave: moved it out to https://github.com/wandb/weave/pull/7399

sdks/node/src/integrations/claude-agent-sdk/messages.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Translate a Claude Agent SDK usage object into Weave's snake_case usage shape.
+ *
+ * The two SDK shapes are disjoint by casing — the per-model `result.modelUsage`
+ * values ({@link ModelUsage}, camelCase) and the aggregate `result.usage`
+ * ({@link NonNullableUsage}, snake_case) — so an `in` check on a camelCase key
+ * proves which one we hold and narrows the union. Both shapes type every token
+ * field as a required number (`NonNullableUsage` strips the `| null` off
+ * `BetaUsage`), so the result is fully populated; no field can be absent.
```
drtangible: Not important for this change, but a thought: this kinda describes what the code is showing. Wonder if there's a way we can tone done LLMs echoing implementation details on internal fns to make it quicker to review PRs.

sdks/node/src/__tests__/integrations/claude-agent-sdk/otelTracer.test.ts
```diff
@@ -0,0 +1,356 @@
...
+import {ClaudeAgentOtelTracer} from '../../integrations/claude-agent-sdk/otelTracer';
+import {
+  findSpan,
+  setupExporterPerTest,
+  setupGenAITestEnvironment,
+} from '../genai/common';
+
+const INVOKE = 'invoke_agent claude_agent_sdk';
+
+describe('Claude Agent SDK — OTel tracer', () => {
```
drtangible, line 227: nit: If this is intended to be a pure unit test of `ClaudeAgentOtelTracer`, convention would be to have it mimic the src folder structure and have this be `otelTracer.test.ts`.


(However, it might be better to test this behavior at a higher level anyways, by either recording or mocking at network boundaries and letting `ClaudeAgentOtelTracer` be an implementation detail that tests are oblivious to. I think that's what your next PR in the stack is aiming for 👍 )

sdks/node/src/__tests__/integrations/claudeAgentSdkOtel.test.ts
```diff
@@ -0,0 +1,356 @@
...
+  test('emits invoke_agent + chat + execute_tool GenAI spans for one query', () => {
+    const tracer = new ClaudeAgentOtelTracer({
+      prompt: 'What is the weather in Tokyo?',
+    });
+    tracer.processMessage({
+      type: 'system',
+      subtype: 'init',
+      session_id: 'sess-1',
+      model: 'claude-x',
+    } as any);
```
drtangible: Can these be `as SDKMessage`?

drtangible (approved):
🙌 🙌 🙌 

Just a couple non-blocking minor comments! Ran the example and seeing this, looks cool!

Image

sdks/node/src/integrations/claude-agent-sdk/messages.ts
rgao-coreweave: done, will make a memory from the pr reviews. thx!

sdks/node/src/__tests__/integrations/claude-agent-sdk/otelTracer.test.ts
rgao-coreweave: changed to `otelTracer.test.ts`

sdks/node/src/__tests__/integrations/claudeAgentSdkOtel.test.ts
rgao-coreweave:
```
  assistantMessage({
        sessionId: 'sess-1',
        model: 'claude-x',
        stopReason: 'tool_use',
        content: [
          thinkingBlock('I should check the weather.'),
          textBlock('Let me check.'),
          toolUseBlock('t1', 'Bash', {command: 'ls'}),
        ],
      })
```
made a function to return `SDKAssistantMessage` since SDKMessage wouldn't work on the partial.

PR #7247, opened 2026-06-16
chore(weave): fix flaky test_call_batch_uploads_files_to_bucket_in_parallel

gtarpenning:
## Summary
- `test_call_batch_uploads_files_to_bucket_in_parallel` asserted `concurrent_peak >= 4`, but the gcs mock recorded the peak then slept \`delay=0.1s\`. Under \`-n8\` xdist CI contention the 4th upload thread could be scheduled only after the 1st finished its sleep and decremented \`inflight\`, so the peak capped at 3.
- Production parallelism is correct (pool dispatches all 4 unique uploads); this is purely a test-side measurement race. Replaced the timing window with a \`threading.Barrier\` so all 4 uploads are provably in-flight at once.

Original failure ([CI job](https://github.com/wandb/weave/actions/runs/27635430571/job/81721250225?pr=7246)):
```
E  AssertionError: expected 4 concurrent uploads, peak=3
E  assert 3 >= 4
```

## Testing
test only change.

### DIFF
--- a/tests/trace/test_server_file_storage.py
+++ b/tests/trace/test_server_file_storage.py
@@ -556,8 +556,10 @@ def test_call_batch_uploads_files_to_bucket_in_parallel(client: WeaveClient, gcs
     collapses to one upload, and every stored object lands under the
     expected project prefix.
     """
-    gcs.state.delay = 0.1
     # 4 unique blobs + 2 duplicates of the first => 4 GCS uploads after dedup.
+    # Barrier makes the peak deterministic so it can't flake under CI scheduler
+    # jitter (each upload blocks until all 4 are simultaneously in-flight).
+    gcs.state.expected_concurrency = 4
     payload_size = 50_000
     payloads = [
         _unique_payload("alpha", payload_size),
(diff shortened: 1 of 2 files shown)


tests/trace/test_server_file_storage.py
```diff
@@ -556,8 +556,10 @@ def test_call_batch_uploads_files_to_bucket_in_parallel(client: WeaveClient, gcs
     collapses to one upload, and every stored object lands under the
     expected project prefix.
     """
-    gcs.state.delay = 0.1
```
ro31337, line 559: There is also

```
                if state.delay:
                    time.sleep(state.delay)
```

In conftest.py.

Can be removed now? not a blocker.

Image

tests/trace/test_server_file_storage.py
```diff
@@ -577,10 +579,9 @@ def test_call_batch_uploads_files_to_bucket_in_parallel(client: WeaveClient, gcs
                 )
             )
 
-    # Pool defaults to 8 workers, so all 4 unique uploads should run
-    # concurrently. concurrent_peak is the load-bearing parallelism signal;
-    # wall-time assertions on top would flake on contended CI runners.
-    assert gcs.state.concurrent_peak >= 4, (
+    # Pool defaults to 8 workers, so all 4 unique uploads run concurrently;
+    # the upload barrier guarantees the peak reaches 4.
+    assert gcs.state.concurrent_peak == 4, (
```
ro31337, line 584: Love `==` part instead of `>=`

ro31337 (approved):

PR #7254, opened 2026-06-16
feat(weave): attribute agent / conversation identity at read time

bcsherma:
## Problem

Agent identity (`agent_name` / `agent_version` / `agent_id`) and conversation identity (`conversation_id`) are typically reported by instrumentors **only on the `invoke_agent` span**, not on the child llm/tool spans it encloses. Because `/agents` filters, groups, sorts, and aggregates are span-exact (`s.agent_name = 'X'`), they see blank identity on every child span — so span lists, counts, charts, tokens, latency, costs, and conversation counts all **undercount** an agent's real work.

Computing an "effective agent" and storing it on each span was the original plan, but two constraints rule it out: (1) OTel spans export on *end*, so child spans usually arrive **before** the `invoke_agent` parent that defines their context; (2) `spans` is a `ReplacingMergeTree` whose `agents` / `agent_versions` / `messages` MVs fire on the INSERT trigger, so re-inserting enriched rows would double-count aggregates.

## Approach: read-time trace attribution

Attribute identity at **read time** by wrapping the `spans` table in an *attributed-spans* subquery (`agent_trace_attribution.attributed_spans_source`). The rule is one line:

> if a span declares its own agent (`agent_name != ''`), use its own identity (span-exact); otherwise inherit the identity of its trace.

Builders swap `FROM spans s` for `FROM {attributed_spans_source(...)} s`, so filter, `GROUP BY`, projection, sort and aggregates all transparently see the attributed value — **what you filter on equals what you display**. A span that set its own identity always keeps it, so sub-agent spans are never re-attributed.

This **replaces** the `$anyInTrace` existential operator that earlier revisions of this branch added (and which this PR originally described). Read-time attribution is strictly better here: `$anyInTrace` made a span match *every* agent in its trace, so per-agent aggregate sums double-counted shared spans; assigning each span exactly one identity keeps aggregates exact.

### Coherent identity

`agent_name` / `agent_version` / `agent_id` identify *one* agent, so they are inherited **together as a single tuple** taken from one span — the earliest span in the trace that declares an `agent_name` (`argMinIf(..., started_at, ...)`). Picking each column independently (e.g. four separate `anyIf`s) can synthesize a `(name, version)` pair that never co-occurred on any real span. The own-vs-inherited choice is likewise made once for the whole triple, gated on a raw-column marker (`(s0.agent_name != '') AS has_own_agent_identity`) so it isn't fooled by the rewritten `agent_name` output alias. `conversation_id` is trace-scoped and inherited on its own.

```sql
-- per-trace fallback (one row per trace)
argMinIf((agent_name, agent_version, agent_id), started_at, agent_name != '') AS fb_agent_identity,
anyIf(conversation_id, conversation_id != '')                                 AS fb_conversation_id
-- coalesce on each span
if(has_own_agent_identity, agent_name,    fb_agent_identity.1) AS agent_name,
if(has_own_agent_identity, agent_version, fb_agent_identity.2) AS agent_version,
if(has_own_agent_identity, agent_id,      fb_agent_identity.3) AS agent_id,
if(conversation_id != '', conversation_id, fb_conversation_id) AS conversation_id
```

## Changes

- **`query_builder/agent_trace_attribution.py`** (new) — `attributed_spans_source(...)` builds the wrapping subquery; `query_references_identity` / `fields_reference_identity` gate when attribution is needed. The fallback scan is bounded to the outer time window widened by one max-trace-duration of slack, so `(project_id, started_at)` primary-key pruning is preserved and the in-memory `trace_id` set stays bounded (no-window queries stay all-time). `base_relation` lets the source compose over the cost-augmented source.
- **`agent_query_builder.py`** — `_spans_source(...)` returns the attributed source (optionally over the cost source). The spans **list** always attributes (it exists to display identity); **count / distribution** attribute only when an identity column is filtered/grouped, so non-identity queries skip the extra trace scan.
- **`agent_stats_query_builder.py`** — `_spans_source_filter_sql` composes attribution over the (optionally cost-augmented) base and routes every CTE through that single source.
- **`agent_query_compiler.py`** — removes the `$anyInTrace` operator branch (net minimal).

Chat-view queries (which already inherit hierarchically in `chat_view`) and the `agents` / `agent_versions` rollup MVs are intentionally left untouched.

## Testing

- **Unit** (`test_agent_trace_attribution.py`) — pins the attributed-source SQL shape (coherent tuple via `argMinIf`, slack-widened fallback) and the `references_identity` gating helpers.
- **End-to-end ClickHouse** (`test_genai_agent_queries.py`) —
  - an unset child span inherits its turn's identity and filters/displays as that agent;
  - a self-identified sub-agent span keeps its own identity and is not pulled into the parent's filter;
  - **an inherited identity is a single coherent agent** — a trace that splits name onto one span and version onto another never produces the phantom `(name, version)` pair.
- Composes with query-time **costs** (ungrouped/grouped/stats `include_costs` e2e still green).
- 157 tests green across `query_builder/` + genai agent / costs / chat-view suites; `nox -e lint` clean (ty, pyright, mypy, ruff).

## Notes / follow-ups

- Frontend wiring (wraps the `agent_name` / `agent_version` / `conversation_id` filters) is a separate **wandb/core** PR; backend deploys first.
- Flat attribution: in a rare multi-agent trace an unset span inherits the trace's earliest-declared agent (a single coherent agent, never a cross-agent mix). Exact for single-agent traces (the norm). The hierarchical "nearest enclosing agent" projection remains separate in `chat_view`.
- v1 optimization, if the per-trace fallback aggregation is too slow at scale: a per-trace `trace_agents` set-table MV; only this module's SQL would change.

### DIFF
--- a/weave/trace_server/query_builder/agent_trace_attribution.py
+++ b/weave/trace_server/query_builder/agent_trace_attribution.py
@@ -1,0 +1,209 @@
+"""Read-time trace attribution for agent-identity columns.
+
+Agent identity (`agent_name` / `agent_version` / `agent_id`) and conversation
+identity (`conversation_id`) are typically reported only on the `invoke_agent`
+span, not on the child llm / tool spans it encloses. A span-exact read of those
+columns therefore shows blank identity on every child span and undercounts the
+turn's real work.
+
+This module attributes that identity at *read time* with a single rule:
+
+    if a span declares its own agent (`agent_name != ''`), use its own
+    identity (singleton / span-exact); otherwise inherit the identity of the
+    span's trace.
+
+It is expressed by wrapping the `spans` table in an *attributed-spans*
+subquery (see :func:`attributed_spans_source`). Callers swap their
+`FROM spans s` for `FROM {attributed_spans_source(...)} s` and every
+downstream reference to `s.agent_name` (filter, GROUP BY, projection,
+aggregate) transparently sees the attributed value — what you filter on equals
+what you display.
+
+`agent_name` / `agent_version` / `agent_id` identify one agent, so they
+are inherited *together* as a single tuple taken from one span — the earliest
+span in the trace that declares an `agent_name` (`argMinIf` by
+`started_at`). Picking each column independently could synthesize a
+`(name, version)` pair that never co-occurred on any real span. Likewise, a
+span that declares its own `agent_name` keeps its *entire* own triple, so its
+version / id are never crossed with an inherited agent's. `conversation_id` is
+trace-scoped and inherited on its own.
+
+This is a deliberately *flat* attribution: in a multi-agent trace an unset span
+inherits the trace's earliest-declared agent. A span that set its own identity
+always keeps it, so sub-agent spans are never wrongly attributed. The
+hierarchical "nearest enclosing agent" projection used by the chat view lives in
+`weave.trace_server.agents.chat_view` and is intentionally separate — queries
+that feed the chat view must NOT use this source or they would double-attribute.
+"""
+
+from __future__ import annotations
+
+import datetime
+
+from weave.trace_server.agents import semconv
+from weave.trace_server.interface import query as tsi_query
+from weave.trace_server.orm import ParamBuilder
+
+# The identity columns that inherit from their trace when unset. Order is fixed
+# so the generated SQL is stable (and snapshot-testable).
+IDENTITY_COLUMNS: tuple[str, ...] = (
+    "agent_name",
+    "agent_version",
+    "agent_id",
+    "conversation_id",
+)
+
+_IDENTITY_COLUMN_SET = frozenset(IDENTITY_COLUMNS)
+
+# `agent_name` / `agent_version` / `agent_id` identify one agent and are
+# inherited together as a single tuple (see module docstring); `agent_name` is
+# the marker for "this span declares its own agent". `conversation_id` is
+# trace-scoped and inherited on its own.
+_AGENT_IDENTITY_COLUMNS: tuple[str, ...] = ("agent_name", "agent_version", "agent_id")
+_AGENT_IDENTITY_MARKER = "agent_name"
+
+# All spans of a trace fall within one turn, so a span needing attribution is at
+# most one trace-duration away from the identity-bearing span it inherits from.
+# We widen the trace-fallback scan's window by this slack so an edge span still
+# finds its identity-bearing span (which may have started just outside the outer
+# window) while ClickHouse can still prune by the (project_id, started_at)
+# primary key instead of scanning the project's whole history.
+_TRACE_FALLBACK_WINDOW_SLACK = datetime.timedelta(hours=1)
+
+
+def _collect_get_fields(node: object) -> set[str]:
+    """Recursively collect every `$getField` name in a dumped Query expr."""
+    found: set[str] = set()
+    if isinstance(node, dict):
+        for key, value in node.items():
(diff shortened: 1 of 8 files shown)


bcsherma:
Added the time-bound subquery mitigation (commit `3a8b742`).

Previously the `$anyInTrace` semi-join was scoped only by `project_id`, so the subquery scanned the project's **entire history** to gather matching trace_ids even when the outer query only read a small time window. Now `compile_agent_query` also takes the outer window (`started_after` / `started_before`) and bounds the subquery to a copy of it, widened by `_ANY_IN_TRACE_WINDOW_SLACK` (1h, one max-trace-duration) on each side:

```sql
s.trace_id IN (
  SELECT s__t.trace_id FROM spans s__t
  WHERE s__t.project_id = {p}
    AND s__t.started_at >= {after - slack}
    AND s__t.started_at <  {before + slack}
    AND (<inner>)
)
```

The slack keeps it correct at window edges (a span's identity-bearing span is in the same trace, so at most one trace-duration away), while restoring `(project_id, started_at)` primary-key pruning on the subquery and bounding the in-memory trace_id set. Queries with no window stay all-time. Slack is a module constant — bump it if real traces routinely run >1h.

Tests: unit (`test_time_window_bounds_subquery`, `test_no_time_window_leaves_subquery_unbounded`) + the end-to-end ClickHouse test now also runs a time-bounded query and confirms the metadata-less child is still attributed.

No frontend (wandb/core#45982) change needed — the window is already part of the request the page sends.

bcsherma:
* **#7254** <a href="https://app.graphite.com/github/pr/wandb/weave/7254?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7254?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

nikumar1206 (approved):

weave/trace_server/query_builder/agent_trace_attribution.py
```diff
@@ -0,0 +1,209 @@
...
+
+Agent identity (`agent_name` / `agent_version` / `agent_id`) and conversation
+identity (`conversation_id`) are typically reported only on the `invoke_agent`
+span, not on the child llm / tool spans it encloses. A span-exact read of those
+columns therefore shows blank identity on every child span and undercounts the
+turn's real work.
+
+This module attributes that identity at *read time* with a single rule:
+
+    if a span declares its own agent (``agent_name != ''``), use its own
```
nikumar1206: not a big fan of double backticks

idk why claude does this.
just `` is good.

weave/trace_server/query_builder/agent_trace_attribution.py
```diff
@@ -0,0 +1,209 @@
...
+# most one trace-duration away from the identity-bearing span it inherits from.
+# We widen the trace-fallback scan's window by this slack so an edge span still
+# finds its identity-bearing span (which may have started just outside the outer
+# window) while ClickHouse can still prune by the (project_id, started_at)
+# primary key instead of scanning the project's whole history.
+_TRACE_FALLBACK_WINDOW_SLACK = datetime.timedelta(hours=1)
+
+
+def _collect_get_fields(node: object) -> set[str]:
+    """Recursively collect every ``$getField`` name in a dumped Query expr."""
```
nikumar1206: same thing double backticks.

PR #7257, opened 2026-06-16
feat(weave_ts): send trace_id on call-end ingest messages

ro31337:
## JIRA Issue(s)
https://coreweave.atlassian.net/browse/WB-35711

## Description
A call's `trace_id` reaches the server on call-start but not on call-end. A trace arrives as many separate ingest messages, and a future server-side ingest sampler needs to make one keep/drop decision for the whole trace, so every message has to carry the trace key. This change sets `trace_id` on every call-end the Node SDK sends, so the end carries the same key the start already does.

It is the TypeScript mirror of the Python change in #7237, with one difference: the Node SDK sends calls only in batches via `/call/upsert_batch` and never makes single-call requests, so it needs only the body field — no `X-Weave-Trace-Id` header. The field is optional and nullable, so the change is purely additive; older clients omit it and the server derives it from the matching start. `/call/update` and deletes are left for a follow-up, as in Python.

## Why this approach
Python had a single funnel for call-ends; the Node SDK builds the end body by hand in 13 places (the core client plus two integrations), so `trace_id` is set at each one. The two realtime sites that send from an async closure snapshot the id into a local before the `await`, because session teardown clears it synchronously and the closure would otherwise read a cleared value. The end schema gains the optional field; the generated API types and the OpenAPI snapshot are hand-edited to match (both are already hand-maintained here), so `generate-api` is not re-run.

## Testing
`pnpm test` is green with one new test in `opFlow.test.ts`: it runs an op and asserts the call-end batch item carries the same `trace_id` as the call-start, reading the raw upsert batch so the in-memory server's start/end merge cannot produce a false pass (verified it fails without the change). Typecheck and prettier-check are clean.

### DIFF
--- a/sdks/node/src/__tests__/opFlow.test.ts
+++ b/sdks/node/src/__tests__/opFlow.test.ts
@@ -351,4 +351,30 @@ describe('Op Flow', () => {
       expect(call.trace_id).toBe(traceId);
     });
   });
+
+  test('call-end carries the same trace_id as the call-start', async () => {
+    // Inspect the raw upsert batch instead of reading calls back: the in-memory
+    // server merges end fields onto the start record, and the start already
+    // carries trace_id, so a merged read would pass even without the fix.
+    const batchSpy = jest.spyOn(
+      inMemoryTraceServer.call,
+      'callStartBatchCallUpsertBatchPost'
+    );
+
+    const myOp = op((x: number) => x * 2, {name: 'myOp'});
+    await myOp(21);
+
+    await inMemoryTraceServer.waitForPendingOperations();
+
+    const items = batchSpy.mock.calls.flatMap(([batchReq]) => batchReq.batch);
+    const startItem = items.find(item => item.mode === 'start');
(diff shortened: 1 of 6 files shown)


sdks/node/src/__tests__/opFlow.test.ts
```diff
@@ -351,4 +351,30 @@ describe('Op Flow', () => {
...
   });
+
+  test('call-end carries the same trace_id as the call-start', async () => {
+    // Inspect the raw upsert batch instead of reading calls back: the in-memory
+    // server merges end fields onto the start record, and the start already
+    // carries trace_id, so a merged read would pass even without the fix.
+    const batchSpy = jest.spyOn(
+      inMemoryTraceServer.call,
+      'callStartBatchCallUpsertBatchPost'
+    );
```
drtangible, line 362: If we find ourselves doing this more often, could maybe benefit from the fake trace server also storing end calls separately so assertions can be made on them directly.

drtangible (approved):
👍 👍 👍

PR #7258, opened 2026-06-16
feat(weave): add OTel tracing primitives to weave.trace_server

amwarrier:
## Summary

First of four PRs migrating `weave/trace_server/*` from `ddtrace` to OpenTelemetry. This PR is intentionally scope-limited: adds the new tracing decorator surface only — **no call sites are migrated**.

Two new public symbols in `weave/trace_server/tracing.py`:

```python
from weave.trace_server.tracing import traced, traced_generator

@traced(name="my_op")
def my_op(): ...

@traced_generator(name="stream_op")
def stream_op() -> Iterator[...]:
    yield ...
```

Plus `reset_tracer_cache()` — a test helper to drop the module-scope `ProxyTracer` cache when swapping the global `TracerProvider` between cases.

## Why a new module

`weave/trace_server/datadog.py` will keep its existing public symbols (`record_db_insert`, `set_root_span_dd_tags`, etc.) — only the *tracing* surface moves here. The name `datadog.py` is misleading once we're on OTel anyway; DD is just the backend, the SDK is OpenTelemetry.

Keeping this PR pure additions also means the decorator contract reads in isolation. PR 2 will be the first caller.

## Key design decisions

| Decision | Why |
|---|---|
| Two decorators, not one with a flag | `@traced` refuses generators at decoration time; `@traced_generator` `yield from`s inside the span body. Type signatures genuinely differ (`Callable[..., T]` vs `Callable[..., Generator]`). |
| Module-scope `_tracer` | Re-resolving `trace.get_tracer()` per call costs ~7µs. On hot CH-query paths with many decorated methods per request, hoisting saves real CPU. |
| No manual try/except | OTel's `start_as_current_span` already records the exception and sets `Status(ERROR, "{ExcType}: {msg}")` on `__exit__`. Manual try/except produces duplicate exception events and gets overwritten by the SDK. |
| `BaseException` pass-through | OTel's `use_span` catches `except Exception`, not `BaseException`. `CancelledError` / `GeneratorExit` / `KeyboardInterrupt` / `SystemExit` propagate without marking the span errored — correct behavior, free from the SDK. See [open-telemetry/opentelemetry-python#4484](https://github.com/open-telemetry/opentelemetry-python/issues/4484). |

## Tests

12 unit tests in `tests/trace_server/test_tracing.py` covering:

- Sync function: span name, error marking, `functools.wraps` metadata preservation
- Async function: span name, error marking, `CancelledError` pass-through
- `@traced` on generator/async-generator: `TypeError` at decoration time (clear error message)
- `@traced_generator`: span covers full iteration, `GeneratorExit` pass-through, raised exception marking
- Nested `@traced` calls produce correct parent/child span linkage

All 12 pass against `opentelemetry-sdk >= 1.28`. OTel SDK is already a declared dependency in `pyproject.toml` (lines 54-60) for the customer-facing `/otel/v1/traces` ingest endpoint, so no new deps.

## Test plan

- [x] All 12 unit tests pass (validated locally against opentelemetry-sdk 1.28+)
- [ ] No regressions in the existing `tests/trace_server/` suite (no call sites migrated yet)
- [ ] CI green

## Follow-up PRs (will open as drafts once this is reviewed)

1. ✅ **This PR** — decorator surface
2. ⏳ `clickhouse_trace_server_batched.py` migration (47 `@ddtrace.tracer.wrap` → `@traced`, plus 2 generators → `@traced_generator`)
3. ⏳ 10 other files migration (17 wraps + 2 `tracer.trace()` blocks + 3 `set_tag` → `set_attribute` + 1 `current_root_span` → `get_current_span`)
4. ⏳ `datadog.py` rewrite — ~30-line inline DogStatsD client (preserves wire format + metric name), plus remove `ddtrace>=2.7.0` from `pyproject.toml`

## Historical PoC

A proof-of-concept of this entire migration was deployed end-to-end on a wandbench cluster under real SDK load (50 `@weave.op` calls + 10 publishes + 275 streamed `get_calls` per validation run). Decorators fire correctly with realistic durations; `_query_stream` now shows 12-50ms span durations (vs near-zero under `@ddtrace.tracer.wrap` on a generator — which is a longstanding latent bug this migration also fixes).

PoC PR: #7196 — staying open as historical reference, not the merge candidate.

### DIFF
--- a/tests/trace_server/test_tracing.py
+++ b/tests/trace_server/test_tracing.py
@@ -7,43 +7,37 @@ they don't depend on (or interfere with) any process-global tracer state.
 from __future__ import annotations
 
 import asyncio
-from collections.abc import Generator, Iterator
+from collections.abc import Iterator
 from typing import Any
 
 import pytest
-from opentelemetry import trace
-from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace import ReadableSpan, TracerProvider
 from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
     InMemorySpanExporter,
 )
 from opentelemetry.trace.status import StatusCode
 
-from weave.trace_server.tracing import reset_tracer_cache, traced, traced_generator
+from weave.trace_server import tracing
+from weave.trace_server.tracing import traced, traced_generator
 
 
 @pytest.fixture
-def exporter() -> Generator[InMemorySpanExporter, None, None]:
+def exporter(monkeypatch: pytest.MonkeyPatch) -> InMemorySpanExporter:
     """Install an isolated TracerProvider + in-memory exporter for the test.
 
-    Swaps the global TracerProvider for the duration of the test, then drops
-    the module-level cached ProxyTracer reference via `reset_tracer_cache()`
-    so the next span open re-resolves against our new provider.
+    Swaps the module-level `_tracer` binding directly so the decorators
+    under test emit through our isolated provider. No process-global state
+    is mutated.
     """
-    previous = trace._TRACER_PROVIDER  # noqa: SLF001
     provider = TracerProvider()
     exp = InMemorySpanExporter()
     provider.add_span_processor(SimpleSpanProcessor(exp))
-    trace._TRACER_PROVIDER = provider  # noqa: SLF001
-    reset_tracer_cache()
-    try:
-        yield exp
-    finally:
-        trace._TRACER_PROVIDER = previous  # noqa: SLF001
-        reset_tracer_cache()
+    monkeypatch.setattr(tracing, "_tracer", provider.get_tracer("test"))
+    return exp
 
 
-def _single_span(exporter: InMemorySpanExporter) -> Any:
+def _single_span(exporter: InMemorySpanExporter) -> ReadableSpan:
     spans = exporter.get_finished_spans()
     assert len(spans) == 1, f"expected one span, got {len(spans)}: {spans!r}"
     return spans[0]
@@ -85,9 +79,7 @@ def test_sync_function_marks_error_and_reraises(
     assert "exception" in event_names
 
 
-def test_sync_function_preserves_metadata(
-    exporter: InMemorySpanExporter,  # noqa: ARG001
-) -> None:
+def test_sync_function_preserves_metadata() -> None:
     """`functools.wraps` should preserve __name__, __doc__, __module__."""
 
     @traced(name="span_name_unrelated_to_fn_name")
@@ -207,12 +199,6 @@ def test_traced_generator_client_disconnect_not_marked_error(
     """A consumer abandoning the generator (e.g. HTTP client disconnect)
     raises `GeneratorExit` inside the generator. The wrapper must NOT mark
     the span as errored.
-
-    This locks the behavior depended on by upstream's `generator_trace`
-    callers (which were originally added to suppress ddtrace's
-    auto-error-on-GeneratorExit). With OTel, the auto-mark doesn't happen,
-    but we still want this test to pin the contract so a future "let's
-    catch GeneratorExit and set status ERROR" regression is caught.
     """
 
     @traced_generator(name="abandoned_op")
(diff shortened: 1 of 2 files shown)


weave/trace_server/tracing.py
```diff
@@ -0,0 +1,173 @@
...
+    Only intended for tests that swap the global TracerProvider mid-process.
+    Production code should never need to call this — there is exactly one
+    TracerProvider per process and it is set during application startup
+    before any span is opened.
+    """
+    # `_real_tracer` is a private attr on `opentelemetry.trace.ProxyTracer`.
+    # If a future OTel release renames it, we'd rather tests silently lose
+    # isolation than crash with AttributeError mid-suite.
+    try:
+        _tracer._real_tracer = None  # type: ignore[attr-defined]  # noqa: SLF001
```
gtarpenning: why do we need these ignores? really dont like this on production code

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+
+        if asyncio.iscoroutinefunction(fn):
+
+            @wraps(fn)
+            async def awrap(*args: Any, **kwargs: Any) -> Any:
+                with _tracer.start_as_current_span(name):
+                    return await fn(*args, **kwargs)
+
+            return awrap  # type: ignore[return-value]
```
gtarpenning: we need to figure out how to type this

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,173 @@
...
+
+        @wraps(fn)
+        def swrap(*args: Any, **kwargs: Any) -> Any:
+            with _tracer.start_as_current_span(name):
+                return fn(*args, **kwargs)
+
+        return swrap  # type: ignore[return-value]
```
gtarpenning: type

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,173 @@
...
+    Production code should never need to call this — there is exactly one
+    TracerProvider per process and it is set during application startup
+    before any span is opened.
+    """
+    # `_real_tracer` is a private attr on `opentelemetry.trace.ProxyTracer`.
+    # If a future OTel release renames it, we'd rather tests silently lose
+    # isolation than crash with AttributeError mid-suite.
+    try:
+        _tracer._real_tracer = None  # type: ignore[attr-defined]  # noqa: SLF001
+    except AttributeError:
```
gtarpenning: also why do we need the try/catch? shouldn't we log at the very least an error? we aren't going to get any tracing...

gtarpenning: oh i see this is for tests, then why even try/catch at all? dont we want a loud error?

weave/trace_server/tracing.py
```diff
@@ -0,0 +1,173 @@
...
+        def swrap(*args: Any, **kwargs: Any) -> Any:
+            with _tracer.start_as_current_span(name):
+                return fn(*args, **kwargs)
+
+        return swrap  # type: ignore[return-value]
+
+    return deco
+
+
+def traced_generator(
```
gtarpenning, line 107: we dont do the same async guard here (unlike above)

tests/trace_server/test_tracing.py
```diff
@@ -0,0 +1,275 @@
...
+    trace._TRACER_PROVIDER = provider  # noqa: SLF001
+    reset_tracer_cache()
+    try:
+        yield exp
+    finally:
+        trace._TRACER_PROVIDER = previous  # noqa: SLF001
+        reset_tracer_cache()
+
+
+def _single_span(exporter: InMemorySpanExporter) -> Any:
```
gtarpenning: shouldn't use `Any` even in tests if we can help it. i think we just need: `tuple[ReadableSpan, ...]`

amwarrier: Thanks. I have changed it to ReadableSpan. _single_span returns just a single span

weave/trace_server/tracing.py
amwarrier: Deleted reset_tracer_cache() entirely, so both the # type: ignore[attr-defined] and # noqa: SLF001 are gone with it.

weave/trace_server/tracing.py
amwarrier: added two @overload stubs (async-first, sync fallback) so T doesn't get conflated between int and Coroutine[Any, Any, int]

weave/trace_server/tracing.py
amwarrier: taken care

weave/trace_server/tracing.py
amwarrier: fixed

gtarpenning (approved):

PR #7260, opened 2026-06-16
feat(weave): tag declarative evaluation child calls with eval metadata

ro31337:
## JIRA Issue(s)

[WB-35748](https://coreweave.atlassian.net/browse/WB-35748)

## Description

The imperative evaluation path (`EvaluationLogger`) already tags each of its calls with the `_weave_eval_meta` attribute, but the declarative path (`Evaluation.evaluate`) tagged nothing on its children. Those calls could be recognized only indirectly, by the op name on the root `Evaluation.evaluate` call. This change wraps the body of `evaluate` in a single `weave.attributes` context, so every call it produces — `predict_and_score`, the model call, the scorers, and `summarize` — now carries the marker. The two paths behave the same.

The reason the marker has to sit on the children is server-side ingest sampling, a planned feature that drops a portion of incoming traces to reduce storage. Evaluation traces must be kept whole, so the sampler needs to recognize eval calls. It decides one message at a time, and a child call is ingested before its root arrives, so the only reliable signal is an attribute on the child itself. The value is merged into any `_weave_eval_meta` an outer wrapper may have already set, so existing markers are preserved rather than overwritten.

## Backward compatibility

The marker is `{"declarative": true}`. The current readers of `_weave_eval_meta` key off `imperative` or `score`, neither of which this sets, so they treat these calls exactly as before. A no-`score` marker is already written on this same path in production today (by the evaluate-model worker), so this is a proven shape, not a new one. The sampler that reads the marker is a separate, later change.

## Testing

Added a declarative-path test that runs an evaluation with `trials=2` and asserts the marker lands on `predict_and_score`, the model call, all scorer calls, and `summarize` for every example. `trials=2` is what proves the marker reaches each example's subtree and not just the first. It also checks that the root stays unmarked, since it is recognized by op name.

Ran locally against ClickHouse with no regressions: the new test, the full declarative eval suites (`test_evaluations.py` and `test_evaluate.py`, 38 passed), the imperative meta-propagation tests, and the server-side eval-results tests (35 passed).

[WB-35748]: https://coreweave.atlassian.net/browse/WB-35748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/evaluation/eval.py
+++ b/weave/evaluation/eval.py
@@ -379,12 +379,8 @@ class Evaluation(Object):
 
     @op(call_display_name=default_evaluation_display_name, eager_call_start=True)
     async def evaluate(self, model: Op | Model) -> dict:
-        # Mark every child call (predict_and_score, model, scorers, summarize)
-        # so server-side ingest sampling can identify eval calls from their own
-        # attributes; children are ingested before the root, so its op_name is
-        # not yet known. Merge into any existing _weave_eval_meta (e.g. set by
-        # evaluate_model_worker) rather than overwriting it. The literal mirrors
-        # eval_imperative.EVAL_META_KEY (importing it here would be circular).
+        # Tag every eval child call (mirrors the imperative path); merge into any
+        # existing _weave_eval_meta (e.g. evaluate_model_worker), don't overwrite.
         prev_eval_meta = call_context.call_attributes.get().get("_weave_eval_meta", {})
         with attributes({"_weave_eval_meta": {**prev_eval_meta, "declarative": True}}):
             eval_results = await self.get_eval_results(model)
--- a/weave/evaluation/eval.py
+++ b/weave/evaluation/eval.py
@@ -29,6 +29,7 @@ from weave.flow.scorer import (
 )
 from weave.flow.util import make_memorable_name, transpose
 from weave.object.obj import Object
+from weave.trace.api import attributes
 from weave.trace.call import Call, CallsIter
 from weave.trace.context import call_context
 from weave.trace.context.weave_client_context import require_weave_client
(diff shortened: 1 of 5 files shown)


weave/evaluation/eval.py
```diff
@@ -378,8 +379,12 @@ async def eval_example(example: dict) -> dict:
...
     @op(call_display_name=default_evaluation_display_name, eager_call_start=True)
     async def evaluate(self, model: Op | Model) -> dict:
-        eval_results = await self.get_eval_results(model)
-        summary = await self.summarize(eval_results)
+        # Tag every eval child call (mirrors the imperative path); merge into any
+        # existing _weave_eval_meta (e.g. evaluate_model_worker), don't overwrite.
+        prev_eval_meta = call_context.call_attributes.get().get("_weave_eval_meta", {})
+        with attributes({"_weave_eval_meta": {**prev_eval_meta, "declarative": True}}):
+            eval_results = await self.get_eval_results(model)
+            summary = await self.summarize(eval_results)
```
ro31337, line 390: ^^^ here we mark every child call (predict_and_score, model, scorers, summarize).

Children are ingested before the root, so its `op_name` is not yet known. Merge into any existing `_weave_eval_meta` (e.g. set by `evaluate_model_worker`) rather than overwriting it.

The literal mirrors `eval_imperative.EVAL_META_KEY`, because importing it here would be circular. I double checked that, and this is what the codebase already does.

weave/evaluation/eval.py
```diff
@@ -378,8 +379,12 @@ async def eval_example(example: dict) -> dict:
 
     @op(call_display_name=default_evaluation_display_name, eager_call_start=True)
     async def evaluate(self, model: Op | Model) -> dict:
-        eval_results = await self.get_eval_results(model)
-        summary = await self.summarize(eval_results)
+        # Tag every eval child call (mirrors the imperative path); merge into any
+        # existing _weave_eval_meta (e.g. evaluate_model_worker), don't overwrite.
+        prev_eval_meta = call_context.call_attributes.get().get("_weave_eval_meta", {})
+        with attributes({"_weave_eval_meta": {**prev_eval_meta, "declarative": True}}):
```
andrewtruong: Can we document which fields we expect to show up in weave_eval_meta?  Something like a TypedDict would be helpful

andrewtruong (approved):
Please see comment about formalizing a type around weave_eval_meta

weave/evaluation/eval.py
ro31337: @andrewtruong done! I'll merge it shortly unless you'd want another look

PR #7267, opened 2026-06-16
feat(weave): add query-time costs to agent spans and stats APIs

bcsherma:
## Purpose

Integrate token costs into the GenAI agent spans system, mirroring how costs work for calls today. Costs are computed at read time (opt-in) and surfaced across the spans, stats, chat-view, and agents APIs so the frontend (companion PR wandb/core#46041) can show cost everywhere it shows token data.

## Approach

Spans differ from calls in a way that makes costs simpler: token usage and the model name are first-class columns, and each span has exactly one model. So per-span cost is a clean LEFT JOIN against the best-matching row in `llm_token_prices` — same pricing source of truth as calls (same table, `project > default` precedence, latest-`effective_date` tie-break). Cost is never stored; it's computed only when a caller opts in. Unpriced models yield `NULL` (not `0`) so "unpriced" stays distinguishable from "free".

## Changes

**Spans + stats**
- New `agents/span_costs.py`: ranked-model-prices subquery, the cost JOIN (keyed on `response_model`→`request_model`), per-span cost projections, and a `cost_augmented_source_sql` spans-shaped relation.
- `AgentSpansQueryReq.include_costs`: per-span cost columns (ungrouped) / summed `total_cost`/`total_input_cost`/`total_output_cost` (grouped).
- Stats: `total_cost`/`input_cost`/`output_cost` derived metrics; the stats query only reads the cost-augmented source when a cost metric is referenced.

**Chat view (trace + conversation detail)**
- The chat-view span fetch is cost-augmented, so `build_trace_chat` sums per-span cost into `AgentChatAssistantMessage.{input,output,total}_cost` (aggregated across the subtree for `invoke_agent` turns), `AgentTraceChatRes.total_cost`, and `AgentConversationChatRes.total_cost`.

**Agents / versions**
- `AgentsQueryReq` / `AgentVersionsQueryReq` gain `include_costs`; `AgentSchema` gains `total_cost`. The materialized views don't store cost, so the handler fills it from a supplementary cost-augmented grouped spans query scoped to the page's agents/versions.

## Testing

`tests/trace_server/test_genai_agent_costs.py` + chat-view unit tests, all against ClickHouse: per-span cost, cache-token subtraction, unpriced → `None`, grouped sums, stats `total_cost`, trace-chat per-message + total cost, and agents/versions cost. Full agent query/stats/chat suites pass (139 tests locally).

## Notes

- Pricing follows calls exactly (most-recent price, not time-of-span).
- Chat-view and conversations/agents queries request costs unconditionally (the price JOIN is cheap and these are low-volume detail/aggregate reads); the spans table opts in only when a cost column is shown.

### DIFF
--- a/tests/trace_server/query_builder/test_agent_query_builder.py
+++ b/tests/trace_server/query_builder/test_agent_query_builder.py
@@ -12,6 +12,7 @@ import pytest
 import sqlparse
 from pydantic import ValidationError
 
+from weave.trace_server.agents.span_costs import cost_augmented_source_sql
 from weave.trace_server.agents.types import (
     AgentConversationChatReq,
     AgentCustomAttrsSchemaReq,
@@ -32,7 +33,10 @@ from weave.trace_server.interface.query import Query
 from weave.trace_server.orm import ParamBuilder
 from weave.trace_server.query_builder.agent_query_builder import (
     CHAT_VIEW_COLS,
+    QUALIFIED_CHAT_VIEW_COLS,
+    QUALIFIED_SPANS_COST_COLS,
     SPAN_SORTABLE_COLS,
+    SPANS_COST_COLS,
     SPANS_LIST_COLS,
     build_order_by,
     make_agent_versions_count_query,
@@ -1036,18 +1040,20 @@ class TestMakeTraceDetailSpansQuery:
         pb = ParamBuilder("genai")
         query = make_trace_detail_spans_query(pb, "p1", "t1")
 
+        # Always cost-augmented (the chat/detail views render per-message and
+        # per-trace cost). Compose the expected cost source from the same
+        # helper so the snapshot stays exact without copying the price-JOIN SQL.
+        expected_pb = ParamBuilder("genai")
+        expected_pb.add("p1", param_type="String")  # genai_0: project filter
+        expected_pb.add("t1", param_type="String")  # genai_1: trace filter
+        source = cost_augmented_source_sql(expected_pb, "p1")
         expected = f"""
-            SELECT {CHAT_VIEW_COLS} FROM spans s
+            SELECT {CHAT_VIEW_COLS}, {SPANS_COST_COLS} FROM {source} s
             WHERE s.project_id = {{genai_0:String}}
             AND s.trace_id = {{genai_1:String}}
             ORDER BY s.started_at ASC
         """
-        assert_sql(
-            expected,
-            {"genai_0": "p1", "genai_1": "t1"},
-            query,
-            pb.get_params(),
-        )
+        assert_sql(expected, expected_pb.get_params(), query, pb.get_params())
 
 
 # ============================================================================
@@ -1456,9 +1462,17 @@ class TestMakeConversationChatSpansQuery:
             AgentConversationChatReq(project_id="p1", conversation_id="c1"),
         )
 
+        # Always cost-augmented (per-message + per-turn cost in the multi-turn
+        # chat view). Compose the expected cost source from the same helper.
+        expected_pb = ParamBuilder("genai")
+        expected_pb.add("p1", param_type="String")  # genai_0: project
+        expected_pb.add("c1", param_type="String")  # genai_1: conversation
+        expected_pb.add(50, param_type="UInt64")  # genai_2: limit (default)
+        expected_pb.add(0, param_type="UInt64")  # genai_3: offset
+        source = cost_augmented_source_sql(expected_pb, "p1")
         expected = f"""
-            SELECT {", ".join(f"s.{c} AS {c}" for c in CHAT_VIEW_COLS.split(", "))}
-            FROM spans s
+            SELECT {QUALIFIED_CHAT_VIEW_COLS}, {QUALIFIED_SPANS_COST_COLS}
+            FROM {source} s
             INNER JOIN (
                 SELECT trace_id, min(started_at) AS turn_started_at
                 FROM spans
@@ -1471,12 +1485,7 @@ class TestMakeConversationChatSpansQuery:
             WHERE s.project_id = {{genai_0:String}}
             ORDER BY t.turn_started_at ASC, t.trace_id ASC, s.started_at ASC
         """
-        assert_sql(
-            expected,
-            {"genai_0": "p1", "genai_1": "c1", "genai_2": 50, "genai_3": 0},
-            query,
-            pb.get_params(),
-        )
+        assert_sql(expected, expected_pb.get_params(), query, pb.get_params())
 
     def test_with_pagination(self) -> None:
         pb = ParamBuilder("genai")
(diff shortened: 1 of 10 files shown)


weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -990,9 +1020,18 @@ def make_spans_list_query(pb: ParamBuilder, req: AgentSpansQueryReq) -> str:
         aggregate_selects = (
             f"{aggregate_selects},\n               {dynamic_aggregate_selects}"
         )
+    if req.include_costs:
+        aggregate_selects = (
+            f"{aggregate_selects},\n"
+            "               sum(s.total_cost) AS total_cost,\n"
```
gtarpenning: does this return 0 or None?

tests/trace_server/query_builder/test_agent_query_builder.py
```diff
@@ -1033,21 +1036,25 @@ def test_custom_alias_used_when_key_non_identifier(self) -> None:
...
-            AND s.trace_id = {{genai_1:String}}
-            ORDER BY s.started_at ASC
-        """
-        assert_sql(
-            expected,
-            {"genai_0": "p1", "genai_1": "t1"},
-            query,
-            pb.get_params(),
-        )
+        normalized = " ".join(query.split())
```
gtarpenning: why this change? it was so nice before?

weave/trace_server/agents/types.py
```diff
@@ -69,6 +69,13 @@
     "total_tokens",
     "is_error",
     "is_invocation",
+    # Query-time costs (USD). Computed by joining the span's model against
+    # llm_token_prices; only resolvable when the query's span source is
+    # cost-augmented. Mirror of span_costs.COST_DERIVED_METRIC_NAMES — keep the
+    # two in sync. See weave/trace_server/agents/span_costs.py.
+    "total_cost",
+    "input_cost",
+    "output_cost",
```
jtschoonhoven: In my experience it's always better to transmit money values in minor units (cents) and to convert to major units (dollars) as late as possible. But it looks like these values are already stored in dollars. Still probably worth being explicit about the units (I wouldn't assume "cost" implied USD).

```suggestion
    # Query-time costs (USD). Computed by joining the span's model against
    # llm_token_prices; only resolvable when the query's span source is
    # cost-augmented. Mirror of span_costs.COST_DERIVED_METRIC_NAMES — keep the
    # two in sync. See weave/trace_server/agents/span_costs.py.
    "total_cost_usd",
    "input_cost_usd",
    "output_cost_usd",
```

I'd recommend making this change throughout if possible.

weave/trace_server/agents/types.py
```diff
@@ -670,6 +694,11 @@ class AgentSpansQueryReq(BaseModel):
     # etc.) in each row. Intended for single-trace detail fetches; do not set
     # for broad list queries.
     include_details: bool = False
+    # When true, compute per-span costs (USD) by joining each span's model
+    # against llm_token_prices. Adds input_cost/output_cost/.../total_cost to
+    # ungrouped span rows, or summed total_cost/total_input_cost/
+    # total_output_cost to grouped rows. See agents/span_costs.py.
+    include_costs: bool = False
```
jtschoonhoven, line 701: Tiny nit: it seems to me like an antipattern to have flags like this on a request payload that modify the response. It's fine when the client is our own frontend, but for external developers and agents it makes the API harder to use.

This isn't _that_ bad (it at least doesn't change the _shape_ of the response) but I was just giving Nikhil a hard time about this and thought I should flag it here too. Probably not worth changing in this case.

weave/trace_server/agents/span_costs.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def ranked_model_prices_sql(pb: ParamBuilder, project_id: str) -> str:
+    """Build a subquery returning the single best price per ``llm_id``.
+
+    Mirrors ``token_costs.build_model_prices_query``: project-level pricing
+    wins over default, ties broken by the most recent ``effective_date``. Only
+    the rank-1 row per model survives. Parameterized (no f-string project_id),
+    so it is safe to embed directly in a JOIN.
+    """
```
jtschoonhoven, line 95: Claude flagged that this function duplicates `token_costs.build_model_prices_query`

weave/trace_server/agents/clickhouse.py
```diff
@@ -409,8 +417,63 @@ def agent_versions_query(self, req: AgentVersionsQueryReq) -> AgentVersionsQuery
...
+                version.total_cost = cost_map.get(version.agent_version)
         return AgentVersionsQueryRes(versions=versions, total_count=total)
 
+    def _grouped_cost_map(
+        self,
+        project_id: str,
+        field: str,
+        values: list[str],
+        *,
+        extra_exprs: list[dict[str, Any]] | None = None,
```
jtschoonhoven, line 446: Nit: rather than `list[dict[str, Any]]]` can you make this `list[tsi_query.Operand]` to get some type safety?

weave/trace_server/agents/span_costs.py
```diff
@@ -0,0 +1,217 @@
...
+    )
+    output_cost = f"{s}.output_tokens * {p}.completion_token_cost"
+    cache_read_cost = f"{s}.cache_read_input_tokens * {p}.cache_read_input_token_cost"
+    cache_creation_cost = (
+        f"{s}.cache_creation_input_tokens * {p}.cache_creation_input_token_cost"
+    )
+    total_cost = (
+        f"({input_cost}) + ({output_cost}) "
+        f"+ ({cache_read_cost}) + ({cache_creation_cost})"
+    )
```
jtschoonhoven: Claude flagged that all this math is duplicated in `token_costs._build_cost_summary_dump_snippet`

weave/trace_server/agents/span_costs.py
```diff
@@ -0,0 +1,217 @@
...
+``effective_date`` wins" tie-break used by ``token_costs.build_model_prices_query``.
+
+Cost is never stored on the span; it is computed at read time only when a
+caller opts in (``include_costs`` on the spans query, or a cost derived metric
+on the stats query). Callers stitch three pieces into their own SQL:
+
+  * ``cost_join_sql`` — the ``LEFT JOIN`` against ranked prices, keyed on the
+    span's model.
+  * ``cost_select_exprs`` — the ``<expr> AS <name>`` cost projections.
+  * ``COST_COLUMN_NAMES`` — the resulting column names, for projections and
```
jtschoonhoven: Nit: idk why claude insists on \``double-ticks`` but they're not necessary

jtschoonhoven (approved):
Instead of passing around values like `total_cost` can we do a big search-and-replace to make that `total_cost_usd`? I'm used to seeing money values passed around in minor units, so I wouldn't assume that "cost" means "dollars", and I think it's worth being really explicit when dealing with money. Otherwise LGTM.

tests/trace_server/query_builder/test_agent_query_builder.py
bcsherma: yeah sry I reverted this

weave/trace_server/agents/types.py
bcsherma: done

PR #7264, opened 2026-06-16
feat(weave): migrate 10 remaining weave/trace_server files to @traced

amwarrier:
## Summary

PR 3 of 4 in the `weave/trace_server/*` ddtrace → OpenTelemetry migration. Migrates the remaining files — everything except `datadog.py`, which is PR-4.

**Counts (against current master):**

| Pattern | Count |
|---|---:|
| `@ddtrace.tracer.wrap` → `@traced` | 18 |
| `with ddtrace.tracer.trace(...)` → `with _tracer.start_as_current_span(...)` | 2 |
| `span.set_tag` / `set_tags` → `span.set_attribute` / `set_attributes` | 3 |
| `ddtrace.tracer.current_root_span()` → `set_root_span_dd_tags()` | 1 |

## Files touched

| File | Changes |
|---|---|
| `agents/clickhouse.py` | 1 wrap (added on master while the PR was open; rebased in) |
| `base64_content_conversion.py` | 1 wrap |
| `clickhouse/utilities.py` | 1 wrap |
| `eval_results_helpers.py` | 6 wraps via local `_trace_wrap` helper; helper deleted along with its conditional ddtrace import block (no longer needed once OTel is a direct dep) |
| `kafka.py` | 1 wrap (+ 2 callers of `set_root_span_dd_tags`) |
| `parallel_bucket_uploads.py` | 1 wrap |
| `project_version/clickhouse_project_version.py` | 1 wrap + `current_root_span()` → `set_root_span_dd_tags()` |
| `project_version/project_version.py` | 1 trace block → `_tracer.start_as_current_span` (+ 4 callers of `set_root_span_dd_tags`) |
| `ttl_settings.py` | 1 wrap + 1 trace block + 3 `set_attribute` calls |
| `usage_utils.py` | 1 wrap |
| `workers/evaluate_model_worker/evaluate_model_worker.py` | 4 wraps |

## Notable: `current_root_span` semantics

[...]

PR #7263, opened 2026-06-16
feat(weave): migrate clickhouse_trace_server_batched.py to @traced

amwarrier:
> **Stacked PR.** Base: #7258 (PR-1, decorator surface). Will auto-retarget to `master` once PR-1 merges.

## Summary

PR 2 of 4 in the `weave/trace_server/*` ddtrace → OpenTelemetry migration. Migrates the largest file: `clickhouse_trace_server_batched.py`.

**Changes:**

- 47 `@ddtrace.tracer.wrap(name=X)` → `@traced(name=X)` (span names preserved 1:1 so DD APM dashboards keep working).
- 4 generator functions → `@traced_generator(name=X)`:
  - `_table_query_stream`, `annotation_queues_query_stream` — were `@ddtrace.tracer.wrap` (see "bug fix" note below)
  - `_query_stream`, `calls_query_stream` — were `@generator_trace` (custom helper from `datadog.py`)
- `import ddtrace` removed from the file.
- `generator_trace` removed from the `datadog.py` import block (the symbol itself stays in `datadog.py` for one more PR).
- Adds `from weave.trace_server.tracing import traced, traced_generator`.

**Net diff:** +50 / −73 lines. Mostly the same shape as before, just with the decorator name swapped and (in 4 cases) the file's import line updated.

## Bug fix bundled in: generator span durations

`@ddtrace.tracer.wrap` on a generator function closes the span when the wrapper *returns the generator object* — that is, before iteration begins. So spans for `_table_query_stream` and `annotation_queues_query_stream` were reporting span durations of ~0ms regardless of how long the actual ClickHouse iteration took. We discovered this when investigating mismatches between our slow-query alerts and DD APM percentiles.

`@traced_generator` `yield from`s inside the span body, so the span correctly covers the full iteration. Post-migration we measure these at 12-50ms, matching actual CH query times.

This is a bug fix, not a regression — but call it out for reviewers who might wonder why span durations on these endpoints suddenly look longer.

The other 2 (`_query_stream`, `calls_query_stream`) were already on the custom `@generator_trace` helper, which `yield from`s correctly. They're swapped to `@traced_generator` for consistency; behavior unchanged.

## Validation

- AST audit: every `@traced` lands on a non-generator function, every `@traced_generator` on a generator. Mismatches: 0.
- `python -m py_compile weave/trace_server/clickhouse_trace_server_batched.py` — passes.
[...]

PR #7268, opened 2026-06-16
feat(weave): drop ddtrace dep, inline DogStatsD in datadog.py

amwarrier:
## Summary

PR 4 of 4 in the `weave/trace_server/*` ddtrace → OpenTelemetry migration. After this merges (plus the wandb/core submodule advance — separate PR), weave-trace stops importing ddtrace at runtime.

### `weave/trace_server/datadog.py` rewritten

Internally no longer depends on `ddtrace`. Public surface preserved:

- `DB_INSERT_METRIC`, `DB_INSERT_PATH_UNKNOWN` (constants)
- `db_insert_path(path)` (context manager)
- `tag_db_insert_path(path)` (decorator)
- `record_db_insert(*, table, count, path=None)` (function)
- `set_root_span_dd_tags(tags)`, `set_current_span_dd_tags(tags)`

**Removed:** `generator_trace(span_name)`. It existed to suppress ddtrace's auto-mark-error-on-`GeneratorExit` behavior. OTel's `use_span` only catches `except Exception` (see [opentelemetry-python#4484](https://github.com/open-telemetry/opentelemetry-python/issues/4484)), so `BaseException` subclasses pass through unmarked by default. The two callers were already migrated to `@traced_generator` from `weave.trace_server.tracing` in #7263.

### `weave/trace_server/local_root.py` added (new)

Contextvar-based "local root span" tracker. Mirrors the implementation in wandb/core's `services/weave-trace/src/tracing/local_root.py`. Provides:

- `local_root_scope(span)` — context manager; record `span` as the entry-point span for the current contextvar-visible scope
- `get_local_root()` — read it back; returns `None` if no scope is active or the recorded span is non-recording

Contextvar propagation: visible across `await`, `asyncio.create_task`, `run_in_threadpool`; NOT across `loop.run_in_executor(None, ...)` without an explicit `copy_context().run` (documented in module docstring).

### `set_root_span_dd_tags` re-implemented

ddtrace had `tracer.current_root_span()` — auto-tracked the entry-point span. OTel has no equivalent. Old behavior (pre-rebase): used `trace.get_current_span()`, which returns the *currently-active* span, NOT the request root. That silently shifted tag emission from the trace root to inner `@traced` spans for any caller inside a `@traced` boundary (caught by @gtarpenning's review on #7264).

New behavior: `set_root_span_dd_tags` calls `get_local_root().set_attributes(tags)`. Real root-span semantics restored — but only when a `local_root_scope(...)` has been entered. The wandb/core FastAPI middleware will be updated in a follow-up PR (task: bump submodule pointer + switch middleware to import `local_root_scope` from `weave.trace_server.local_root` instead of its own local copy). Until then, `set_root_span_dd_tags` is a no-op — matches the docstring's "no-op if no `local_root_scope` is active" contract.

`set_current_span_dd_tags` is unchanged: uses `trace.get_current_span()` to tag whatever span is currently active. The two helpers are now genuinely different functions — the name no longer lies.

### Inline DogStatsD

The DogStatsD client (used by `record_db_insert`) is now a ~30-line `_StatsDClient`:

- Opens a UDP socket lazily to `${DD_DOGSTATSD_URL}` (or `${DD_AGENT_HOST}:8125` fallback)
- Encodes counters in DogStatsD wire format `name:value|c|#tag1:val1,tag2:val2`
- Swallows `OSError` so stats hiccups never crash the request path

**Wire format and metric name (`weave_trace_server.db_inserts`) are byte-identical to the previous ddtrace-bundled emitter** — existing Datadog dashboards keyed on this metric continue to render unchanged.

### `pyproject.toml`

Removed `ddtrace>=2.7.0` from both occurrences (`[project.optional-dependencies].trace_server` and `[dependency-groups].trace_server`).

### `uv.lock` regenerated

ddtrace 4.0.0 + transitive deps (`bytecode`, `envier`) removed. 50 refs → 0.

## Validation

- [x] All 19 unit tests pass locally (`test_tracing.py` + `test_local_root.py`)
- [x] `grep -rn 'import ddtrace\|from ddtrace' weave/trace_server/` — empty (no remaining runtime imports)
- [x] `grep -c 'ddtrace' uv.lock` — 0
- [x] DogStatsD wire format preserved verbatim
- [x] CI green

## Stack

1. ✅ **MERGED** [#7258](https://github.com/wandb/weave/pull/7258) — decorator surface
2. ✅ **MERGED** [#7263](https://github.com/wandb/weave/pull/7263) — `clickhouse_trace_server_batched.py`
3. ⏳ [#7264](https://github.com/wandb/weave/pull/7264) — remaining files
4. ⏳ **This PR** — `datadog.py` rewrite + `local_root.py` + drop `ddtrace>=2.7.0`

After this merges + the wandb/core companion PR (submodule advance + middleware import switch), weave-trace's full migration to OTel is done.


weave/trace_server/datadog.py
```diff
@@ -27,14 +42,60 @@
...
+            parsed = urlparse(url if "://" in url else f"udp://{url}")
+            if parsed.hostname:
+                return parsed.hostname, parsed.port or 8125
+        except ValueError:
+            logger.warning(
+                "Could not parse DD_DOGSTATSD_URL=%r; falling back to localhost:8125",
+                url,
+            )
+    host = os.environ.get("DD_AGENT_HOST", "localhost")
+    port = int(os.environ.get("DD_DOGSTATSD_PORT", "8125"))
```
gtarpenning: not guarded, but the docstring right above promises bad config 'falls back ... rather than crashing the process at import time'. a non-numeric `DD_DOGSTATSD_PORT` raises `ValueError` here at import, and ~11 modules import this one, so a single bad env var takes down the whole trace server. wrap it in try/except like the URL branch does?

weave/trace_server/datadog.py
```diff
@@ -27,14 +42,60 @@
...
+            _SOCK = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            _SOCK.setblocking(False)
+        except OSError:
+            logger.warning("DogStatsD socket creation failed; metric emission disabled")
+            _SOCK_FAILED = True
+            return
+    tag_suffix = f"|#{','.join(tags)}" if tags else ""
+    packet = f"{metric}:{value}|c{tag_suffix}".encode()
+    try:
+        _SOCK.sendto(packet, _ADDR)
```
gtarpenning: `_ADDR` is a hostname in the documented prod config (`udp://datadog.datadog:8125`), so this `sendto` runs a `getaddrinfo` on every call, and `record_db_insert` fires per DB insert. `setblocking(False)` doesn't make name resolution non-blocking. resolve to an IP once in `_resolve_dogstatsd_addr`, or `connect()` the socket once and `send()`?

weave/trace_server/datadog.py
```diff
@@ -27,14 +42,60 @@
...
+    if _SOCK is None:
+        try:
+            _SOCK = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            _SOCK.setblocking(False)
+        except OSError:
+            logger.warning("DogStatsD socket creation failed; metric emission disabled")
+            _SOCK_FAILED = True
+            return
+    tag_suffix = f"|#{','.join(tags)}" if tags else ""
+    packet = f"{metric}:{value}|c{tag_suffix}".encode()
```
gtarpenning: description says this is byte-identical to the old emitter, but ddtrace's dogstatsd client usually attaches `constant_tags` (env/service/version from `DD_*`) and this only sends `table`/`path`. if those were present before, any dashboard grouping by `service`/`env` shifts. did you capture actual packets both ways, or confirm the old client had no constant tags?

weave/trace_server/datadog.py
```diff
@@ -27,14 +42,60 @@
 )
 
 
-@lru_cache(maxsize=1)
-def _dogstatsd_client() -> DogStatsd:
-    """Process-wide DogStatsd client; lazy so import is side-effect free.
+def _resolve_dogstatsd_addr() -> tuple[str, int]:
```
gtarpenning: the byte-identical wire format is the load-bearing claim of the whole PR and nothing pins it. this helper (url / host:port / malformed branches) + the packet encoding are pure and trivial to test. a small parameterized test so a future edit can't silently drift the format or the fallback chain?

weave/trace_server/datadog.py
```diff
@@ -1,21 +1,36 @@
-"""Datadog integration utilities for trace server."""
+"""Trace-server observability helpers (DogStatsD counters + OTel span tags).
+
+Historical name `datadog.py` kept because 11 internal modules import from
+it; renaming is a mechanical follow-up. The module no longer depends on
```
gtarpenning: this docstring narrates the migration (`no longer depends on ddtrace`, plus `Replaces the historical ...` on L173). reads as PR-description material, and files outlive the PR. trim to the load-bearing bit (what it does + the fixed wire format)?

weave/trace_server/datadog.py
```diff
@@ -27,14 +42,60 @@
...
+
+def _emit_statsd(metric: str, value: int, tags: list[str]) -> None:
+    """Send a single DogStatsD counter packet.
+
+    Best-effort: a socket-creation or send failure never interrupts the
+    caller. We lazy-create the UDP socket on first emission so importing
+    this module has zero side effects.
     """
-    return get_dogstatsd_client(str(_agent_config.dogstatsd_url))
+    global _SOCK, _SOCK_FAILED  # noqa: PLW0603 — lazy singleton init
```
gtarpenning: description calls this a `_StatsDClient`, but it's module globals + `global`/`# noqa`. old code used `@lru_cache` and avoided both. a tiny class instance (or a `connect()`-once socket, which also kills the per-call DNS on L95) drops the noqa. minor: two threads can both init `_SOCK` here lock-free, last write wins and leaks one socket. benign for fire-and-forget udp, but a one-line 'race is fine' note would help.

gtarpenning (commented):
ddtrace finally gone, nice. comments inline. the two i'd want fixed before merge: the import-time crash on a bad `DD_DOGSTATSD_PORT` (one env typo takes down the trace server), and a test pinning the 'byte-identical' wire format since that claim is the whole point of the PR. also worth confirming the per-call DNS on the insert hot path + whether ddtrace's constant tags got dropped.

two description nits, not code: it lists `set_root_span_dd_tags` under 'public surface preserved' but this PR deletes it (callers already moved upstream so runtime is fine, just fix the bullet), and it describes a `_StatsDClient` class that doesn't actually exist here.

weave/trace_server/tracing_root.py
```diff
@@ -0,0 +1,94 @@
+"""Contextvar-based "local root span" tracker for the OpenTelemetry tracing
```
gtarpenning, line 1: can we name this file to something about observability?

weave/trace_server/datadog.py
amwarrier: fixed in 741df8df. extracted `_parse_port(raw, *, default)` that try/excepts the int() and falls back to the default + warns. pinned by `test_parse_port_non_numeric_does_not_raise_at_import` in the new test file.

weave/trace_server/datadog.py
amwarrier: good catch — fixed in 741df8df. `_StatsDClient` now `connect()`s once in `_ensure_socket()`, so DNS resolution happens at startup and each `emit()` is a single non-blocking `send()`. no more getaddrinfo per insert.

weave/trace_server/datadog.py
amwarrier: audited it. ddtrace 4.0.0's `get_dogstatsd_client(url, namespace=None, tags=None)` — master called it with no `tags=` argument, so `constant_tags=None` on the old emitter too. the previous client wasn't attaching env/service/version either. no regression.

weave/trace_server/datadog.py
amwarrier: added in 741df8df — `tests/trace_server/test_datadog.py`, 14 tests. wire format pinned via a public `format_packet(metric, value, tags)` helper (parameterized: no tags / single / multi / zero value). address resolution pinned across all 6 fallback branches.

weave/trace_server/datadog.py
amwarrier: trimmed in 741df8df. down to 4 lines — what it does + the wire format. no more migration narration.

weave/trace_server/datadog.py
amwarrier: fixed in 741df8df. now an actual class with `threading.Lock` for safe double-init. `# noqa: PLW0603` and the module globals are gone. the `connect()`-once approach also kills the per-call DNS you flagged elsewhere.

weave/trace_server/tracing_root.py
amwarrier: done in 741df8df — `local_root.py` → `tracing_root.py`. module is now `weave.trace_server.tracing_root`. function names (`local_root_scope`, `get_local_root`) unchanged — they describe the concept (entry-point span treated as the trace's local root), only the filename changed. the wandb/core sibling will get renamed in the follow-up consolidation PR.

gtarpenning (approved):

PR #7269, opened 2026-06-17
feat(weave): extend `/agent_search` endpoint

jtschoonhoven:
## Summary

Generalizes the `messages`-table search endpoint (`agent_search`) for structured retrieval, so callers can pull a slice of messages by filters rather than only full-text search:

- `trace_id` is now one filter among many (alongside `conversation_id`, `roles`, `agent_name`, etc.).
- `query` is optional — an empty query drops the `content LIKE` clause and matches all messages.
- `truncate_content` returns truncated message content (default True).

This is to support the agent scoring worker, so it can efficiently fetch message content for scoring (in cases where it's not already present on the root span).

## Testing

- Query-builder unit test for the `trace_id` + `full_content` + empty-query path.
- ClickHouse integration test: inserts messages on child spans, retrieves by `trace_id` with `full_content`, asserts content is returned untruncated (600-char message) and tool roles are filtered out by the `roles` filter.
- Existing `TestMakeMessageSearchQuery` and search integration tests still pass (param numbering unchanged when `query` is set and `trace_id` is absent).

### DIFF
--- a/tests/trace_server/test_genai_agent_queries.py
+++ b/tests/trace_server/test_genai_agent_queries.py
@@ -1655,7 +1655,7 @@ def test_message_search_trace_id_full_content(ch_server):
             query="",
             trace_id=trace_id,
             roles=["user", "assistant", "system"],
-            full_content=True,
+            truncate_content=False,
         )
     )
     by_role: dict[str, list[str]] = {}
--- a/tests/trace_server/test_genai_agent_queries.py
+++ b/tests/trace_server/test_genai_agent_queries.py
@@ -1616,6 +1616,59 @@ def test_message_search_shared_digest_across_spans(ch_server):
     assert len(digests) == 1
 
 
+def test_message_search_trace_id_full_content(ch_server):
+    """Structured retrieval: empty query + trace_id + full_content returns the
+    trace's user/assistant/system messages untruncated, excluding tool roles.
+    This is the path the agent scoring fallback uses.
+    """
+    project_id = _make_project_id("search_trace_full")
+    trace_id = uuid.uuid4().hex
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+    long_text = "x" * 600  # exceeds the 500-char preview cap
+
+    spans = [
+        _make_span(
+            project_id,
+            trace_id=trace_id,
+            span_id=uuid.uuid4().hex,
+            operation_name="chat",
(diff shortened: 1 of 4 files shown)


weave/trace_server/agents/types.py
```diff
@@ -790,21 +790,29 @@ class AgentCustomAttrsSchemaRes(BaseModel):
...
 
+    trace_id: str | None = None
     roles: list[SearchMessageRole] | None = None
     conversation_id: str | None = None
     agent_name: str | None = None
     provider_name: str | None = None
     request_model: str | None = None
+    # Truncate message content to a preview (default) to keep search-UI payloads
+    # small; set False to return full content (e.g. for structured retrieval).
+    truncate_content: bool = True
```
nichochar, line 815: Should this be an integer for the size of the truncation rather than a boolean? 0 means no truncation but then you can control it via the filter

tests/trace_server/test_genai_agent_queries.py
```diff
@@ -1616,6 +1616,59 @@ def test_message_search_shared_digest_across_spans(ch_server):
...
+            span_id=uuid.uuid4().hex,
+            operation_name="execute_tool",
+            started_at=now + datetime.timedelta(seconds=1),
+            tool_call_result="tool output",
+        ),
+    ]
+    _insert_spans(ch_server.ch_client, spans)
+
+    res = ch_server.agent_search(
+        AgentSearchReq(
```
nikumar1206, line 1653: it looks like this model has a limit applied:
```

    limit: int = Field(default=DEFAULT_SEARCH_LIMIT, ge=0, le=MAX_SEARCH_LIMIT)
```
and the default is 20. so is this really a full search. also looks like its started at desc (so newest first). is that what we want?

nikumar1206 (approved):

weave/trace_server/agents/types.py
jtschoonhoven: I considered that, but I don't think we actually need that level of control on the client side.

It would be a backwards-compatible change if we want to change it in the future. Keeping it simple for now.

tests/trace_server/test_genai_agent_queries.py
jtschoonhoven: Yes these constraints are fine for my use case, thanks for flagging

PR #7275, opened 2026-06-17
fix(weave): sync migration status writes so a finished migration can't be left partial

gtarpenning:
## Summary
- The migrator writes migration state with async `ALTER ... UPDATE` mutations (`_update_migration_status`) and treats *submitting* them as success. If the "clear partial" mutation stalls on the replicated `db_management.migrations` table (or the process dies in the async window), the row stays `partially_applied_version=N` even though the migration's SQL already succeeded. The next boot's guard then hard-fails forever (crashlooped a managed install for ~2 weeks).
- Fix: run the status `ALTER ... UPDATE`s with `mutations_sync = 2` so they're durably applied before the migrator returns (and fail loudly if they can't).
- Jira: [WB-35755](https://coreweave.atlassian.net/browse/WB-35755)

## Testing
full migrator functional suite (cloud / replicated / distributed / legacy-Replicated-DB) passes with the now-obsolete `_sync_mutations()` test workaround removed; its removal staying green confirms the status writes are synchronous. Root cause confirmed from prod logs: no OOM/timeout/CH error in 30d, only the guard; the 026 backfill is trivially fast, so the SQL succeeded and only the async status-clear was lost.

[WB-35755]: https://coreweave.atlassian.net/browse/WB-35755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_migrator.py
+++ b/weave/trace_server/clickhouse_trace_server_migrator.py
@@ -577,12 +577,15 @@ class BaseClickHouseTraceServerMigrator(ABC):
         self, target_db: str, target_version: int, is_start: bool = True
     ) -> None:
         """Update the migration status in management database migrations table."""
+        # mutations_sync=2 makes the status write durable on every replica before we
+        # return; the async default can leave a finished migration marked partial.
+        sync = {"mutations_sync": 2}
         if is_start:
             command = f"ALTER TABLE {self.management_db}.migrations UPDATE partially_applied_version = {target_version} WHERE db_name = '{target_db}'"
-            self._run_ddl_with_retry(command)
+            self._run_ddl_with_retry(command, settings=sync)
         else:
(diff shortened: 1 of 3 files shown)


weave/trace_server/clickhouse_trace_server_migrator.py
```diff
@@ -577,12 +577,15 @@ def _update_migration_status(
         self, target_db: str, target_version: int, is_start: bool = True
     ) -> None:
         """Update the migration status in management database migrations table."""
+        # mutations_sync=2 makes the status write durable on every replica before we
```
amwarrier, line 580: would be good to refer the ticket WB-35755 for a future reader.

amwarrier (approved):

PR #7286, opened 2026-06-17
feat(weave_ts): tag declarative evaluation child calls with eval metadata

ro31337:
## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-35811 — TypeScript sibling of #7260 (the Python change, WB-35748).

## Description

The imperative evaluation path (`EvaluationLogger`) already tags every call it emits with the `_weave_eval_meta` attribute, but the declarative path (`Evaluation.evaluate`) tagged nothing on its children — they were identifiable only by the op name on the root call. This change brings the two paths to parity by wrapping the `evaluate` body in a single `withAttributes` context, so `predictAndScore`, the model call, and every scorer inherit `{declarative: true}`.

This is groundwork for a server-side ingest sampler that needs to keep evaluation traces intact. The sampler decides per message, and a child call reaches the server before its root, so the marker has to ride on every child rather than only the root. The value is merged into any existing `_weave_eval_meta` rather than replacing it, so an outer marker is preserved. The attribute key now lives in one place (`constants.ts`) and is reused by both paths instead of being hardcoded twice. This mirrors the merged Python change in #7260; the sampler that reads the marker is a separate, later change.

## Testing

Two new unit tests in `evaluation.test.ts`, using the same in-memory trace-server harness as the existing imperative-marker tests. The first runs a declarative eval with `nTrials: 2` and `maxConcurrency: 2` and asserts the marker lands on every `predictAndScore`, model, and scorer call (with exact counts), while the root stays unmarked. The second wraps the eval in an outer `_weave_eval_meta` and asserts both keys survive the merge. Full Node suite is green (284 tests); cjs+esm typecheck and prettier-check are clean.

## Backward compatibility

Purely additive: declarative children now carry one extra key inside the free-form `attributes` dict. No schema or endpoint change. Existing readers of `_weave_eval_meta` only act on `.imperative` / `.score`, so `{declarative: true}` is inert to them.

### DIFF
--- a/sdks/node/src/__tests__/evaluation.test.ts
+++ b/sdks/node/src/__tests__/evaluation.test.ts
@@ -229,12 +229,13 @@ describe('Evaluation - declarative eval metadata', () => {
       expect(call.attributes?._weave_eval_meta?.declarative).toBe(true);
     }
 
-    // The root evaluate call is recognized by op_name, so it is not tagged.
-    const rootCall = calls.find(c =>
+    // Exactly one root evaluate call; recognized by op_name, so it is
+    // intentionally not tagged (its attributes are read before the wrapper runs).
+    const rootCalls = calls.filter(c =>
       c.op_name?.includes(EVALUATION_RUN_OP_NAME)
     );
-    expect(rootCall).toBeDefined();
-    expect(rootCall?.attributes?._weave_eval_meta).toBeUndefined();
+    expect(rootCalls).toHaveLength(1);
+    expect(rootCalls[0].attributes?._weave_eval_meta).toBeUndefined();
   });
 
   // A flat overwrite would drop the outer marker; deep-merge keeps both keys.
--- a/sdks/node/src/__tests__/evaluation.test.ts
+++ b/sdks/node/src/__tests__/evaluation.test.ts
@@ -226,7 +226,7 @@ describe('Evaluation - declarative eval metadata', () => {
       ...modelCalls,
       ...scorerCalls,
     ]) {
-      expect(call.attributes?._weave_eval_meta?.declarative).toBe(true);
+      expect(call.attributes?._weave_eval_meta).toEqual({declarative: true});
     }
 
     // Exactly one root evaluate call; recognized by op_name, so it is
@@ -248,16 +248,27 @@ describe('Evaluation - declarative eval metadata', () => {
     });
 
     const calls = await getCalls(traceServer, projectId);
-    const childCalls = calls.filter(
+    const predictAndScoreCalls = calls.filter(c =>
+      c.op_name?.includes(EVALUATION_RUN_PREDICTION_AND_SCORE_OP_NAME_TS)
+    );
+    const modelCalls = calls.filter(c => c.op_name?.includes('mockPrediction'));
+    const scorerCalls = calls.filter(
       c =>
-        c.op_name?.includes(EVALUATION_RUN_PREDICTION_AND_SCORE_OP_NAME_TS) ||
-        c.op_name?.includes('mockPrediction') ||
         c.op_name?.includes('lengthScorer') ||
         c.op_name?.includes('inclusionScorer')
     );
-    expect(childCalls.length).toBeGreaterThan(0);
 
-    for (const call of childCalls) {
+    // dataset rows (5), and scorers also * scorer count (2).
+    expect(predictAndScoreCalls).toHaveLength(5);
+    expect(modelCalls).toHaveLength(5);
+    expect(scorerCalls).toHaveLength(10);
+
+    // Both the outer marker and ours survive (flat overwrite would drop foo).
+    for (const call of [
+      ...predictAndScoreCalls,
+      ...modelCalls,
+      ...scorerCalls,
+    ]) {
       expect(call.attributes?._weave_eval_meta).toEqual({
         foo: true,
         declarative: true,
--- a/sdks/node/src/__tests__/evaluation.test.ts
+++ b/sdks/node/src/__tests__/evaluation.test.ts
@@ -1,7 +1,23 @@
+import {withAttributes} from '../clientApi';
+import {
+  EVALUATION_RUN_OP_NAME,
+  EVALUATION_RUN_PREDICTION_AND_SCORE_OP_NAME_TS,
+} from '../constants';
 import {Dataset} from '../dataset';
 import {Evaluation} from '../evaluation';
 import {type ColumnMapping} from '../fn';
 import {op} from '../op';
+import {initWithCustomTraceServer} from './clientMock';
+import {InMemoryTraceServer} from './helpers/inMemoryTraceServer';
+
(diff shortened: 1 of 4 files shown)


sdks/node/src/__tests__/evaluation.test.ts
```diff
@@ -1,7 +1,23 @@
...
 import {Dataset} from '../dataset';
 import {Evaluation} from '../evaluation';
 import {type ColumnMapping} from '../fn';
 import {op} from '../op';
+import {initWithCustomTraceServer} from './clientMock';
+import {InMemoryTraceServer} from './helpers/inMemoryTraceServer';
+
+// Read all recorded calls back from the in-memory trace server (mirrors the
+// helper in evaluationLogger.test.ts).
+async function getCalls(traceServer: InMemoryTraceServer, projectId: string) {
```
drtangible, line 15: Are we able to use `const calls = await traceServer.getCalls(projectId);` instead of duplicating that function?

If so, I can take a pass through DRYing up other tests, don't worry about it for this change.

sdks/node/src/__tests__/evaluation.test.ts
```diff
@@ -169,3 +185,83 @@ describe('Evaluation', () => {
...
+
+    const calls = await getCalls(traceServer, projectId);
+    const predictAndScoreCalls = calls.filter(c =>
+      c.op_name?.includes(EVALUATION_RUN_PREDICTION_AND_SCORE_OP_NAME_TS)
+    );
+    const modelCalls = calls.filter(c => c.op_name?.includes('mockPrediction'));
+    const scorerCalls = calls.filter(
+      c =>
+        c.op_name?.includes('lengthScorer') ||
+        c.op_name?.includes('inclusionScorer')
```
drtangible, line 216: Do we have to guard against a nullish `op_name`? AFAICT the type is `string`.

sdks/node/src/__tests__/evaluation.test.ts
```diff
@@ -169,3 +185,83 @@ describe('Evaluation', () => {
...
+      expect(call.attributes?._weave_eval_meta?.declarative).toBe(true);
+    }
+
+    // Exactly one root evaluate call; recognized by op_name, so it is
+    // intentionally not tagged (its attributes are read before the wrapper runs).
+    const rootCalls = calls.filter(c =>
+      c.op_name?.includes(EVALUATION_RUN_OP_NAME)
+    );
+    expect(rootCalls).toHaveLength(1);
+    expect(rootCalls[0].attributes?._weave_eval_meta).toBeUndefined();
```
drtangible: Do we have to guard against a nullish `attributes`?

sdks/node/src/__tests__/evaluation.test.ts
```diff
@@ -169,3 +185,83 @@ describe('Evaluation', () => {
...
+    expect(predictAndScoreCalls).toHaveLength(10);
+    expect(modelCalls).toHaveLength(10);
+    expect(scorerCalls).toHaveLength(20);
+
+    for (const call of [
+      ...predictAndScoreCalls,
+      ...modelCalls,
+      ...scorerCalls,
+    ]) {
+      expect(call.attributes?._weave_eval_meta?.declarative).toBe(true);
```
drtangible: Thoughts on making these assertions look at the entire shape of `_weave_eval_meta` so they illustrate the full picture? eg:

```ts
      expect(call.attributes._weave_eval_meta).toEqual({declarative: true});
```

sdks/node/src/__tests__/evaluation.test.ts
```diff
@@ -169,3 +185,83 @@ describe('Evaluation', () => {
...
+      await evaluation.evaluate({model, maxConcurrency: 2});
+    });
+
+    const calls = await getCalls(traceServer, projectId);
+    const childCalls = calls.filter(
+      c =>
+        c.op_name?.includes(EVALUATION_RUN_PREDICTION_AND_SCORE_OP_NAME_TS) ||
+        c.op_name?.includes('mockPrediction') ||
+        c.op_name?.includes('lengthScorer') ||
+        c.op_name?.includes('inclusionScorer')
```
drtangible: Could some of these match exactly on `op_name ===` rather than using `.includes?`?

sdks/node/src/__tests__/evaluation.test.ts
```diff
@@ -169,3 +185,83 @@ describe('Evaluation', () => {
...
+    );
+    expect(childCalls.length).toBeGreaterThan(0);
```
drtangible: Nit: the above test asserts directly on the length here. Feels like either doing that, or adding separate assertions on the different types of child calls we're testing against would be a bit more robust.

sdks/node/src/__tests__/evaluation.test.ts
ro31337: yes, `traceServer.getCalls` works, just mirrored the existing helper in the `evaluationLogger.test.ts`, so I'd let your DRY pass fold both onto it (if you don't mind)

sdks/node/src/__tests__/evaluation.test.ts
ro31337: kept the `?.` to match the existing `evaluationLogger.test.ts`, which uses `op_name?.includes(...)`, LMK if you're fine with that

sdks/node/src/__tests__/evaluation.test.ts
ro31337: same as the op_name one, left the `?.` for consistency with the existing tests

sdks/node/src/__tests__/evaluation.test.ts
ro31337: these `op_name`s are full ref URIs, something like `weave:///blabla/op/Name:digest`, so a literal `===` won't match IMO. `.includes` matches the name segment (same as the existing tests)

sdks/node/src/__tests__/evaluation.test.ts
ro31337: fixed

sdks/node/src/__tests__/evaluation.test.ts
ro31337: fixed

drtangible: Ah, okay cool.

sdks/node/src/__tests__/evaluation.test.ts
drtangible: Sure, I can clean up this one as part of that.

sdks/node/src/__tests__/evaluation.test.ts
drtangible: Sure thing -- I'll take a look at that as part of the aforementioned cleanup as well.

drtangible (approved):
👍 👍 👍

PR #7291, opened 2026-06-17
perf(weave): batch traced completion-call span writes

gtarpenning:
## Summary
- Batch the per-judge traced completion-call span write: add bulk `AgentWriteHandler.insert_spans` + a new `acompletions_create_deferred` that returns the built span (instead of inserting it) so a batch caller (the scoring worker) writes all spans in one ClickHouse insert.
- Per-call `acompletions_create` is unchanged.

## Testing
unit test asserts the deferred path returns a real span and skips the per-call insert; end-to-end test bulk-inserts and reads the spans back. scoring-worker consumer + QA load test land in the paired core PR.

### DIFF
--- a/weave/trace_server/agents/clickhouse.py
+++ b/weave/trace_server/agents/clickhouse.py
@@ -686,7 +686,7 @@ class AgentWriteHandler:
                     accepted += 1
 
         if span_rows:
-            self._insert_spans(span_rows)
+            self.insert_spans(span_rows)
 
         if failure_counts:
             logger.warning(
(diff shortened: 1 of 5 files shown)


weave/trace_server/agents/clickhouse.py
```diff
@@ -718,14 +718,25 @@ def insert_span(self, span: AgentSpanCHInsertable) -> None:
...
         insert_with_empty_query_retry(
             self._ch_client,
             "spans",
-            data=[genai_span_to_row(span)],
+            data=[genai_span_to_row(s) for s in spans],
             column_names=ALL_SPAN_INSERT_COLUMNS,
             settings=self._insert_settings,
         )
-        record_db_insert(table="spans", count=1)
+        record_db_insert(table="spans", count=len(spans))
```
jtschoonhoven: Looks like you need to rebase off `master`, there's already an `AgentWriteHandler._insert_spans` function that's similar to this one. It just doesn't exist on this branch. You should be able to re-use the existing function after a rebase.

jtschoonhoven (approved):
LGTM but make sure you rebase and avoid duplicating the existing `_insert_spans`

weave/trace_server/agents/clickhouse.py
gtarpenning: done

Issue #7066, opened 2026-06-05
[Bug]: OpenTelemetry ingestion results in unusable threads.


a0js:
### Describe the bug

I'm following the [Send OpenTelemetry Traces to Weave](https://docs.wandb.ai/weave/guides/tracking/otel#endpoint-details) instructions to send gen_ai Otel spans to weave but have run into a few snags. I'm doing my best to follow [spec](https://opentelemetry.io/docs/specs/semconv/gen-ai/), granted the spec is still in development so I totally understand how things are in flux right now.

The traces are showing up fine in the traces view but threads have a few issues:

---

### `invoke_agent` span requires input messages, even when spec expects child to contain it
When following Otel spec and using both `invoke_agent` span with one or more child `chat` spans, I am purposefully excluding inputs from the `invoke_agent` span, the inputs and outputs are on the children `chat` spans. As a result, there is a `Something went wrong.` Error on the preview pane when trying to preview the thread.  

<details>
<summary>error message from the `Copy Error` button</summary>

```
  "uuid": "63b8013a-cf2b-486c-81dc-61b71e6691e4",
  "url": "https://wandb.ai/elementum-engineering/hydra-test/weave/threads?peekPath=%2Felementum-engineering%2Fhydra-test%2Fthreads%2F944ee621-9212-4c36-b09a-ee41a44d3a3d",
  "error": {
    "message": "Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7B%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.",
    "stack": [
      "Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7B%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.",
      "    at $g$1 (https://cdn.wandb.ai/production/6894dc570/assets/index-hZc5sfud.js:68:8538)",
      "    at C$28 (https://cdn.wandb.ai/production/6894dc570/assets/index-hZc5sfud.js:68:16184)",
      "    at Aj (https://cdn.wandb.ai/production/6894dc570/assets/index-hZc5sfud.js:70:1968)",
      "    at Il$2 (https://cdn.wandb.ai/production/6894dc570/assets/index-hZc5sfud.js:70:68254)",
[...]

PR #7294, opened 2026-06-18
fix(weave): respect explicit non-turn spans

jstar0:
## Summary

Part of #7066.

When OpenTelemetry spans include `gen_ai.conversation.id`, Weave currently uses that conversation id as both the thread id and an implicit turn signal. That is useful as a default, but it also means child spans that correctly belong to the conversation become extra turns even when they explicitly set `wandb.is_turn=false`.

This change lets explicit Weave metadata opt out of that turn inference while keeping the conversation id as the thread id. This is intentionally scoped to the backend ingestion rule; it does not address the separate thread preview or child-span navigation UI crashes also described in the issue.

## Changes

- Preserve explicit boolean `False` values during OTel attribute extraction.
- Resolve `wandb.is_turn` before semconv-derived turn inference.
- Keep `gen_ai.conversation.id` as `thread_id` even when `wandb.is_turn=false`.
- Add regressions for extraction and `Span.to_call()` behavior.

## Verification

```bash
pytest tests/trace_server/test_opentelemetry.py::TestPythonSpans::test_span_to_call_wandb_is_turn_false_overrides_conversation_id tests/trace_server/test_opentelemetry.py::TestSemanticConventionParsing::test_wandb_turn_and_thread_attributes tests/trace_server/test_opentelemetry.py::TestSemanticConventionParsing::test_genai_semconv_conversation_id_turn_can_be_disabled -q
pytest tests/trace_server/test_opentelemetry.py --trace-server=fake -q
ruff check weave/trace_server/opentelemetry/attributes.py weave/trace_server/opentelemetry/constants.py tests/trace_server/test_opentelemetry.py
ruff format --check weave/trace_server/opentelemetry/attributes.py weave/trace_server/opentelemetry/constants.py tests/trace_server/test_opentelemetry.py
git diff --check
```

### DIFF
--- a/tests/trace_server/test_opentelemetry.py
+++ b/tests/trace_server/test_opentelemetry.py
@@ -20,6 +20,7 @@ from opentelemetry.proto.trace.v1.trace_pb2 import (
 )
 from opentelemetry.semconv_ai import SpanAttributes as OTSpanAttr
 
+from tests.trace.util import FAKE_NOT_IMPLEMENTED
 from weave.trace import weave_client
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.constants import MAX_OP_NAME_LENGTH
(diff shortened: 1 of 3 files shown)


jstar0:

PR #7298, opened 2026-06-18
fix(weave): use async insert for agent spans path

gtarpenning:
## Summary
- Agent/genai-otel spans inserts went through `AgentWriteHandler` with no `settings`, so they always wrote synchronously even with `WEAVE_TRACE_CLICKHOUSE_USE_ASYNC_INSERT` on.
- Thread the server's resolved async-insert settings into the handler so the spans path honors the same toggle as the main `_insert` path.

## Testing
non-functional plumbing; respects existing `_use_async_insert` toggle (no change when off).


ro31337 (approved):

andrewtruong:
force-merging to deploy a fix

PR #7300, opened 2026-06-18
fix(weave): make weave installable alongside google-adk

andrewtruong:
Fixes a version compatibility issue between `google-adk` and `opentelemetry-semantic-conventions`.  The solution is to basically vendor the semconv, which is not optimal, but practical to make installation work for users!

It also adds a missing test shard that would have caught this issue initially


drtangible (approved):
👍 👍 👍

PR #7307, opened 2026-06-18
chore(weave): drop libsqlite3-dev install step from test CI workflows

andrewtruong:
Drops unnecessary libsqlite3 install


drtangible (approved):
👍 👍 👍

PR #7342, opened 2026-06-22
feat(weave): send client-capability header on ingest (Python + Node)

ro31337:
## JIRA Issue(s)

## Description

Both the Python and Node SDKs now send an `X-Weave-Client-Capabilities` header on every request they make to the Weave trace server. The value is a small comma-separated list of the sampling-relevant capabilities this SDK build guarantees — today `trace_id_on_end,eval_child_meta`, matching the two recent changes that put `trace_id` on every call message and tag evaluation child calls with eval metadata.

The header is groundwork for a future server-side ingest sampler. To sample the calls model safely the server has to drop whole traces consistently and never shred evaluations, which is only true once a client sends `trace_id` everywhere and marks its eval calls. The server cannot infer that from the data alone — an unmarked call is ambiguous (a normal call, or an eval child from an older client) — so the client states it directly, and the sampler leaves traffic without the header unsampled. Nothing reads the header yet; this mirrors the `X-Weave-Trace-Id` header added recently for the same future sampler.

## Why this approach

A capability list rather than an SDK version keeps the server out of mapping versions to behavior: the Python and Node SDKs version independently, and the relevant behavior landed across several releases. It is advisory, not a gate — a missing or unknown capability just means "don't sample," never a rejected request. It is set once in the shared request-header builder, so it rides every request (it describes the client, not one trace); a future sampler simply reads it on the ingest paths.

## Testing

Python: new tests assert the header is present on single-call and on batch upsert requests; the full `trace_server_bindings` suite passes (64). Node: a new `clientApi` test asserts `init()` wires the header onto the trace-server client; the full suite passes (313). The VCR cassette matcher now ignores this header the same way it already ignores the volatile `client_version` body field, so the recorded `live/` cassettes still match. Lint, format and typecheck (cjs + esm) are clean.

## Backward compatibility

Additive and safe both ways: an older server ignores the unknown header, and an older client simply doesn't send it. The Stainless transport and the one-shot login health check are out of scope, matching the `X-Weave-Trace-Id` change. One follow-up to track: if the Stainless transport later becomes the default, this header and `X-Weave-Trace-Id` must be ported to it together, otherwise the sampler would see header-less traffic from the new default client.

### DIFF
--- a/tests/trace_server_bindings/test_http_behavior_remote.py
+++ b/tests/trace_server_bindings/test_http_behavior_remote.py
@@ -27,6 +27,8 @@ from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server_bindings.async_batch_processor import AsyncBatchProcessor
 from weave.trace_server_bindings.call_batch_processor import CallBatchProcessor
 from weave.trace_server_bindings.http_utils import (
+    CLIENT_CAPABILITIES,
+    CLIENT_CAPABILITIES_HEADER,
     ERROR_CODE_CALLS_COMPLETE_MODE_REQUIRED,
     TRACE_ID_HEADER,
 )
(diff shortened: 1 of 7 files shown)


tests/trace_server_bindings/test_http_behavior_remote.py
```diff
@@ -637,3 +639,63 @@ def test_legacy_upsert_batch_end_carries_trace_id_in_body(mock_post):
...
+# =============================================================================
+# X-Weave-Client-Capabilities on every ingest request
+#
+# Unlike the per-trace X-Weave-Trace-Id header, the capability header describes
+# the client build itself, so it rides every request -- single calls AND
+# batches -- letting a future server-side ingest sampler recognize a
+# sampling-safe client even on the batch path, which cannot carry a per-trace
+# trace_id header.
+# =============================================================================
+
```
ro31337, line 653: ^ is it too wordy?

ro31337:
Splitting the TS and Python changes into separate PRs per review feedback (keeps reverts simple): Python #7350, Node #7351. Both are drafts with green CI. Closing this combined PR in favor of those.

PR #7343, opened 2026-06-22
fix(weave): add numeric then string fallback to eval input output sort

domphan-wandb:
## Description

- Fixes [WB-35949](https://coreweave.atlassian.net/browse/WB-35949)

This PR fixes input and output sorting on eval results by adding numeric aware ordering to `inputs.*` and `output.*` columns in the eval results query, so columns are sorted by value instead of lexicographically.

**Problem**

We noticed that input and output columns were being sorted lexicographically. Looking at the trace calls, the values were numbers, so we diagnosed that it was an issue with the backend and not our SDK write path.

The backend creates a query, via `build_sort_expression`:
1. Scores are numerically coerced for sorting
2. inputs.* and output.* were ordered as the raw JSON_VALUE String via `any(...)`, so ch sorted them lexicographically. 

**Solution**

Make the input and output column sorting numeric aware with a string fallback so these columns sort as expected. All changes are in `eval_results_query_builder.py`:
* `_build_sort_aggregate` now gives `inputs.*/output.*` a three term existence → numeric → string ordering:
1. `(toFloat64OrNull(any(x)) IS NOT NULL) DESC` — group rows that have a numeric value ahead of non-numeric/missing rows. It's pinned DESC (regardless of the requested direction) so that grouping holds for both ascending and descending sorts.
2. `toFloat64OrNull(any(x)) {dir}` — sort the numeric values by their actual number, in the requested direction.
3. `any(x) {dir}` — fall back to alphabetical order for the remaining non-numeric (text) values.
* `_build_difference_sort` now uses a new `_sort_numeric_scalar` helper, so it subtracts numeric scalars instead of raw strings.

## Testing
**Before**
Sorting on inputs:
image

Sorting on outputs:
image

**After**
Sorting on inputs:
image

Sorting on outputs:
image

- [x] Added unit tests
- [x] Added clickhouse tests 
- [x] Tested on dev_mtsaas 

Follow-up PR for the in-memory test backend: https://github.com/wandb/weave/pull/7354

[WB-35949]: https://coreweave.atlassian.net/browse/WB-35949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace_server/query_builder/test_eval_results_query_builder.py
+++ b/tests/trace_server/query_builder/test_eval_results_query_builder.py
@@ -10,6 +10,7 @@ from weave.trace_server.orm import ParamBuilder
 from weave.trace_server.query_builder.eval_results_query_builder import (
     build_eval_results_cte_chain,
     build_eval_results_query,
+    build_sort_expression,
 )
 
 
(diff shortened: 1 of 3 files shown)


tests/trace_server/query_builder/test_eval_results_query_builder.py
```diff
@@ -733,3 +734,190 @@ def test_full_query_calls_complete() -> None:
             "pb_5": "proj-1",
         },
     )
+
+
+def test_sort_output_numeric_then_string_fallback_asc() -> None:
```
gtarpenning: can we combine these tests a bit? why do we need a different whole test for asc/desc?

gtarpenning (commented):
can we add a functional test? the query unit tests are nice but ideally we would actually execute the new code in ch to confirm its safe (and prevent regressions)

tests/trace_server/query_builder/test_eval_results_query_builder.py
domphan-wandb: thanks, parameterized it.

gtarpenning (approved):

PR #7346, opened 2026-06-23
fix(weave): validate agent signal filters against the agent-spans schema

jtschoonhoven:
## Problem

In the new **Agents → Signals** tab, changing the "Only score turns matching" filter on an agent signal (e.g. `Operation name = "chat"`) failed to save with a **422 Unprocessable Content** from `POST /obj/create`:

```
Field operation_name is not allowed. Allowed fields: attributes, display_name, ended_at, ...
op_name, ... Allowed dynamic prefixes: attributes.*, feedback.*, inputs.*, output.*, summary.*, ...
```

(Same error for `agent_name`, `provider_name`, etc.)

## Root cause

An agent signal persists a `Monitor` whose `query` is a Mongo-style `$expr` built from the filter editor. The filter editor uses **logical agent-span field names** — `operation_name`, `agent_name`, `provider_name`, `request_model`, `conversation_id`, `tool_name`, … — which is correct: those are exactly the fields the agent **Spans** tab and the agent scoring worker's `agent_spans_query` understand (resolved by `compile_agent_query` against the denormalized agent-spans schema).

But at `obj_create` time, `validate_monitor_query_fields` validated **every** monitor query against the raw **`calls_merged`** schema via `process_query_to_conditions`. That schema only allows `op_name`, `attributes.*`, `inputs.*`, `output.*`, `summary.*`, etc. — so the agent-span logical fields were rejected and the save 422'd.

The two consumers of a monitor `query` want **different schemas**:

| Monitor kind | Scored over | Valid fields |
| --- | --- | --- |
| Regular monitor | raw calls (`calls_merged`) | `op_name`, `attributes.*`, `inputs.*`, … |
| Agent monitor (`op_names` ⊇ `weave.genai.turn_ended`) | agent spans (`agent_spans_query`) | `operation_name`, `agent_name`, … |

Validating an agent query against `calls_merged` was simply the wrong schema. (Conversely, a frontend-side rewrite to `attributes.weave.*` raw paths would have cleared the 422 but broken the scoring worker, which passes the stored query verbatim into `agent_spans_query` — so the fix belongs here, on the validation side.)

## Change

- `validate_monitor_query_fields` now branches on whether the monitor is an **agent monitor** (its `op_names` include an agent-span op literal). Agent-monitor queries are validated with `compile_agent_query` (the agent-spans schema); regular-monitor queries keep using `process_query_to_conditions` against `calls_merged`. Unknown fields still fail loudly on both paths.
- Added `AGENT_SPAN_OP_NAMES` to `weave/trace_server/agents/constants.py`, mirroring `weave/flow/monitor.AGENT_SPAN_OP_NAMES`, so the trace server can detect agent monitors **without importing `weave.flow`** (the validation module deliberately avoids that dependency). A test asserts the two stay in sync.

## Tests

New `tests/trace_server/calls_query_builder/test_monitor_query_validation.py` (23 cases):
- agent monitors accept `operation_name` and all the other logical agent-span fields, single and AND-ed;
- regular monitors still accept calls fields (`op_name`, `attributes.*`) and **still reject** agent logical names;
- unresolvable agent fields raise `InvalidRequest`; null query / non-Monitor objects are no-ops;
- the mirrored constant equals `weave.flow`'s.

```
tests/trace_server/calls_query_builder/test_monitor_query_validation.py .... 23 passed
tests/trace_server/query_builder/test_agent_query_compiler.py + test_calls_query_builder.py .... 162 passed
```

`ruff check` / `ruff format --check` clean on the changed files.

## Scope / notes

- Backend-only fix; no frontend (`wandb/core`) change is needed — the filter editor already emits the correct logical field names.
- The second issue from the bug report (empty `{output_messages}` in the scorer prompt) is **not** addressed here, per the request to focus on the 422.
- Could not run the full trace-server suite locally (it needs ClickHouse/Kafka infra); CI covers that. The unit tests above run standalone.

---
Re-opens #7345, which was authored by the Fixit bot and could not pass the CLA check under a bot account.

### DIFF
--- a/tests/trace_server/calls_query_builder/test_monitor_query_validation.py
+++ b/tests/trace_server/calls_query_builder/test_monitor_query_validation.py
@@ -1,0 +1,146 @@
+"""Unit tests for saved-Monitor query validation.
+
+Agent monitors (op_names include an agent-span op) validate against the
+agent-spans schema, so logical fields like `operation_name` are accepted there
+but still rejected on a regular monitor (validated against the calls schema).
+"""
+
+import pytest
+
+from weave.flow.monitor import AGENT_SPAN_OP_NAMES as FLOW_AGENT_SPAN_OP_NAMES
+from weave.trace_server.agents.constants import AGENT_SPAN_OP_NAMES
+from weave.trace_server.calls_query_builder.monitor_query_validation import (
+    validate_monitor_query_fields,
+)
+from weave.trace_server.errors import InvalidFieldError, InvalidRequest
+
+AGENT_OP = next(iter(AGENT_SPAN_OP_NAMES))
+# A normalized (non-agent) op-name ref, as a regular monitor stores them.
+CALL_OP_REF = "weave:///entity/project/op/my_op:abc123"
+
+
+def _eq(field: str, value: object) -> dict:
+    return {"$expr": {"$eq": [{"$getField": field}, {"$literal": value}]}}
+
+
+def _monitor_val(*, op_names: list[str], query: object) -> dict:
+    return {
+        "_type": "Monitor",
+        "_class_name": "Monitor",
+        "_bases": ["Monitor", "Object", "BaseModel"],
+        "op_names": op_names,
+        "query": query,
+    }
+
+
+def _validate(val: dict) -> None:
+    validate_monitor_query_fields("Monitor", "Monitor", val)
+
+
+class TestConstantInSync:
+    def test_agent_span_op_names_mirror_flow(self) -> None:
+        # The trace-server copy is mirrored from weave/flow/monitor.py so the
+        # trace server can detect agent monitors without importing weave.flow.
+        assert AGENT_SPAN_OP_NAMES == FLOW_AGENT_SPAN_OP_NAMES
+
+
+class TestAgentMonitorQuery:
+    def test_operation_name_filter_accepted(self) -> None:
+        # The reported bug: an agent signal filtering on Operation name = "chat"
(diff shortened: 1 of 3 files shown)


tests/trace_server/calls_query_builder/test_monitor_query_validation.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class TestConstantInSync:
+    def test_agent_span_op_names_mirror_flow(self) -> None:
+        # The trace-server copy is mirrored from weave/flow/monitor.py so the
+        # trace server can detect agent monitors without importing weave.flow.
+        assert AGENT_SPAN_OP_NAMES == FLOW_AGENT_SPAN_OP_NAMES
+
+
+class TestAgentMonitorQuery:
```
gtarpenning, line 47: nitty but i dont think we need a whole file for this and generally we dont do class based pytest tests

tests/trace_server/calls_query_builder/test_monitor_query_validation.py
```diff
@@ -0,0 +1,146 @@
...
+        # error carries the allowed-calls-field list (InvalidFieldError).
+        with pytest.raises(InvalidFieldError):
+            _validate(
+                _monitor_val(
+                    op_names=[CALL_OP_REF], query=_eq("operation_name", "chat")
+                )
+            )
+
+
+class TestNoQuery:
```
gtarpenning, line 133: this feels superfluous

gtarpenning (approved):
test suite went kinda wild but functional code lgtm

PR #7347, opened 2026-06-23
fix(weave): surface interleaved reasoning between tool calls in agent chat

bcsherma:
## Problem

In the Agents UI, an assistant turn's reasoning ("thinking") showed up **only when it immediately preceded the final text response** — reasoning that happened **between tool calls** was silently dropped.

Root cause is in the chat-view projection, not in instrumentation, ingest, or storage:

- `reasoning_content` is only ever attached to an `assistant_message`.
- An `assistant_message` is only emitted when a span produced output **text** (`_emit_assistant_message` returned `None` when `_extract_non_user_output_text` was empty).
- An LLM step whose output is a tool call carries `reasoning + tool_call` parts but **no text**, so it produced no `assistant_message` — and its reasoning was discarded. The actual tool call is rendered from a separate `execute_tool` span, which has no reasoning field.

Net effect: reasoning survived only on the final, text-producing step of a turn.

## Fix

Emit a **reasoning-only** assistant message for content spans that carry reasoning but no text, ordered right before the resulting tool call:

- `_emit_assistant_message` gains an `include_reasoning_only` flag. When set, a span with `reasoning_content` but no output text still yields a message (empty `text`, populated `reasoning_content`).
- `_walk_content_span` passes `include_reasoning_only=True`. Crucially, a reasoning-only step is **not** counted as the turn's final assistant text, so it does not suppress the mirrored-final-message logic on the enclosing `invoke_agent` span.

**No frontend change needed:** the chat view already renders the reasoning block independently of body text, and an empty text body renders nothing (`CollapsibleSmartText` returns `null` for empty input).

## Validation

- Unit tests: full `test_genai_chat_view.py` suite passes (22), plus two new tests:
  - interleaved reasoning surfaces as a reasoning-only step, in order before its tool call;
  - a tool-calling step with neither text nor reasoning emits no (empty) assistant bubble.
- **Real data:** replayed a production `wb-agent-prod` conversation through the patched projection. Reasoning went from surfacing on **2 of 9** assistant turns to **all 19** reasoning-bearing LLM steps (17 previously-dropped interleaved steps + 2 final), matching the reasoning content captured at ingest exactly.

## Scope / risk

- Pure projection change; no schema, ingest, or migration impact. The reasoning data was already stored in `reasoning_content` — this only stops the read path from dropping it.
- The mirror/aggregation path for `invoke_agent` is unchanged (reasoning-only steps deliberately don't flip the "emitted final text" signal).

### DIFF
--- a/tests/trace_server/test_genai_chat_view.py
+++ b/tests/trace_server/test_genai_chat_view.py
@@ -532,6 +532,148 @@ def test_reasoning_part_not_duplicated_in_assistant_text() -> None:
     assert payload.reasoning_content == reasoning
 
 
+def _tool_call_part(name: str, arguments: str) -> dict:
+    return {"type": "tool_call", "id": "tc1", "name": name, "arguments": arguments}
+
+
+def test_interleaved_reasoning_before_tool_call_is_surfaced() -> None:
+    """Reasoning that precedes a tool call (an LLM step whose only output is a
+    tool call, no assistant text) must still surface as a reasoning-only
+    message, in order before the tool call — not be dropped because the step
+    produced no final text.
+    """
+
+    def at(seconds: int) -> datetime.datetime:
+        return datetime.datetime(
+            2026, 1, 1, 0, 0, seconds, tzinfo=datetime.timezone.utc
+        )
+
+    step_reasoning = "Deciding to inspect the workspace"
+    final_reasoning = "Summarizing what I found"
+    answer = "The eval suite has 17 scenarios."
+
+    spans = [
+        _span(
+            span_id="agent",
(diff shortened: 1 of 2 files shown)


bcsherma:
* **#7347** <a href="https://app.graphite.com/github/pr/wandb/weave/7347?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7347?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

gtarpenning (approved):

PR #7348, opened 2026-06-23
feat(weave): system_instructions on turn & subagent spans

rgao-coreweave:
Carries `system_instructions` (agent system prompt) on the `invoke_agent` spans the Session SDK emits.

- `Turn` + `SubAgent` gain `system_instructions` → `gen_ai.system_instructions` (gated by `include_content`, PII-redacted).
- Settable via `start_turn` / `Session.start_turn` / `log_turn` / `start_subagent` / `turn.subagent`, or attribute assignment.

The Agents-tab agent-start marker reads `system_instructions` only from the `invoke_agent` span; before, only LLM/chat spans carried it.

First of two stacked PRs — #7429 adds batch-path fidelity + `Turn.record()`.

Tests: `pytest tests/session/` — 285 pass.


tests/session/test_session_otel.py
```diff
@@ -892,6 +910,36 @@ def test_turn_creates_invoke_agent_span(
         assert attrs["gen_ai.conversation.id"] == "sess-1"
         assert attrs["gen_ai.conversation.name"] == "Weather Chat"
 
+    def test_turn_system_instructions_emitted_on_span(
+        self, otel_spans: InMemorySpanExporter
+    ) -> None:
+        with Session(agent_name="weather-bot", session_id="sess-si") as s:
+            with s.start_turn(user_message="hi") as turn:
+                turn.system_instructions = ["You are a weather bot"]
```
drtangible, line 919: How do we decide which fields are exposed via a param to `Session.start_turn()` (or `weave.start_turn()`?

For example, why is `user_message` (and `model` and `agent_name`) settable via `start_turn` here, but not this new `system_instructions` field?

---

I ask mainly thinking about how we'd add similar functionality to the TS SDK. APIs relying on setting individual fields are less common there, so I think we'll probably end up supporting this functionality through several paths:

* `weave.startTurn({systemInstructions: ...});`
* `session.startTurn({systemInstructions: ...});`
* `turn.setSystemInstructions(...);` or `turn.update({systemInstructions: ...});`, or something.

rgao-coreweave: that's an actual gap you're pointing at!

made `system_instructions` a keyword param on `start_turn`, `Session.start_turn`, and `log_turn`, matching the LLM.

rgao-coreweave: also opened the PR: https://github.com/wandb/weave/pull/7429

weave/session/session.py
```diff
@@ -1135,15 +1182,19 @@ def start_tool(*, name: str, arguments: str = "", tool_call_id: str = "") -> Too
     return Tool(name=name, arguments=arguments, tool_call_id=tool_call_id)
 
 
-def start_subagent(*, name: str, model: str = "") -> SubAgent:
+def start_subagent(
+    *, name: str, model: str = "", system_instructions: list[str] | None = None
+) -> SubAgent:
```
drtangible, line 1187: Is this API running into a similar problem? Is it intentional that some of the subagent's fields (eg, `agent_id`, `agent_description`, `agent_version` can only be updated individually via the field setters?

drtangible (approved):
👍

weave/session/session.py
rgao-coreweave: true, will do this in https://github.com/wandb/weave/pull/7429/changes

rgao-coreweave: thank you!

PR #7350, opened 2026-06-23
feat(weave): send client-capability header on ingest (Python)

ro31337:
**1/2 of #7342 (closed)**

## JIRA Issue(s)

## Description

The Python SDK now sends an `X-Weave-Client-Capabilities` header on every request it makes to the trace server. The value is a short comma-separated list of the sampling-relevant capabilities this build guarantees — today `trace_id_on_end,eval_child_meta`, matching the two recent changes that put `trace_id` on every call message and tag evaluation child calls with eval metadata.

The header is groundwork for a future server-side ingest sampler. To sample the calls model safely the server must drop whole traces consistently and never shred evaluations, which only holds once a client sends `trace_id` everywhere and marks its eval calls. The server cannot infer that from the data alone — an unmarked call is ambiguous (a normal call, or an eval child from an older client) — so the client states it directly, and the sampler leaves traffic without the header unsampled. Nothing reads the header yet; this mirrors the `X-Weave-Trace-Id` header added recently for the same future sampler.

## Why this approach

A capability list rather than an SDK version keeps the server out of mapping versions to behavior, and the relevant behavior landed across several releases. It is advisory, not a gate — a missing or unknown capability just means "don't sample," never a rejected request. It is set once in the shared request-header builder, so it rides every request: single calls and batches alike, which matters because the batch path cannot carry a per-trace header.

## Testing

New tests assert the header is present on single-call and on batch upsert requests; the full `trace_server_bindings` suite passes (64 locally).

## Backward compatibility

Additive and safe both ways: an older server ignores the unknown header, and an older client simply doesn't send it. The Stainless transport and the one-shot login health check are out of scope, matching the `X-Weave-Trace-Id` change; if Stainless later becomes the default, both headers must be ported to it together.


weave/trace_server_bindings/http_utils.py
```diff
@@ -30,6 +30,14 @@
 # parsed, the same cheap-header pattern RequestSizeLimitMiddleware already uses.
 TRACE_ID_HEADER = "X-Weave-Trace-Id"
 
+# Request header advertising the sampling-relevant capabilities this SDK build
+# guarantees, so a future server-side ingest sampler can tell a sampling-safe
+# client from an older one and leave unsupported traffic unsampled. Sent on
+# every request (it describes the client, not one trace); a comma-separated,
+# forward-compatible token list.
+CLIENT_CAPABILITIES_HEADER = "X-Weave-Client-Capabilities"
+CLIENT_CAPABILITIES = "trace_id_on_end,eval_child_meta"
```
andrewtruong, line 39: How do we keep this in sync with the server?

tests/trace_server_bindings/test_http_behavior_remote.py
```diff
@@ -637,3 +639,63 @@ def test_legacy_upsert_batch_end_carries_trace_id_in_body(mock_post):
...
+    mock_post.return_value = httpx.Response(
+        200, json={}, request=httpx.Request("POST", "http://example.com")
+    )
+
+    call_id = "call-id"
+    start = StartBatchItem(
+        req=tsi.CallStartReq(start=generate_start(call_id, "entity/project"))
+    )
+
+    try:
```
andrewtruong, line 688: Why is this in a try block?

ro31337: I'd say it's the same answer as here: https://github.com/wandb/weave/pull/7237#discussion_r3421235247

> Mirrors the same pattern in the this file, for example `test_calls_complete_batch_endpoint_and_payload`. As I understand, `should_batch=True` spins up threads and we need to stop them

andrewtruong: hmm, it seems like a smell.  Does it work without that?

weave/trace_server_bindings/http_utils.py
ro31337: Fair question. While we're aiming for [forward compatibility](https://en.wikipedia.org/wiki/Forward_compatibility), nothing is reading these values on the server side yet.

So nothing enforces it,  and it is just a list of tokens the SDK and server agree on, similar to `X-Weave-Trace-Id`. The server only acts on tokens it knows, anything else means "don't sample". So if they ever drift, worst case we sample a bit less - nothing breaks.

Probably needs to be reflected in docs? I'm trying to put meaningful descriptions and comments, and keep docs in Notion so it's easier to see my intent.

tests/trace_server_bindings/test_http_behavior_remote.py
ro31337: IMO, if we have `should_batch=True`, we should _finally_ free up resources. So the code that was always there seems to be working correctly, and it's all about `finally` block. Navigating away from that existing pattern would require a bit of research. I would agree that there is a little bit of copy-pasta every time we add a test that can be optimized.

andrewtruong (approved):

PR #7351, opened 2026-06-23
feat(weave): send client-capability header on ingest (Node)

ro31337:
**2/2 of #7342 (closed)**

## JIRA Issue(s)

## Description

The Node SDK now sends an `X-Weave-Client-Capabilities` header on every request it makes to the trace server, carrying the same `trace_id_on_end,eval_child_meta` capability list as the Python SDK (#7350).

The header is groundwork for a future server-side ingest sampler: it lets the server recognize a client whose traces are safe to sample — one that sends `trace_id` everywhere and marks its eval calls — and leave traffic without the header unsampled. Nothing reads it yet. See #7350 for the full rationale.

## Why this approach

Advisory, not a gate — a missing or unknown capability just means "don't sample," never a rejected request. It is set once where the trace-server client is built, so it rides every request.

## Testing

The full jest suite passes (321). The VCR cassette matcher now ignores this header the same way it already ignores the volatile `client_version` body field, so the recorded `live/` cassettes still match. Prettier-check and typecheck (cjs + esm) are clean.

## Backward compatibility

Additive and safe both ways: an older server ignores the unknown header, and an older client simply doesn't send it.


sdks/node/src/__tests__/helpers/withCassette.ts
andrewtruong, line 1: I think it's ok to regenerate the cassettes.  This will make them more correct going forward

drtangible: It might be a bit painful at the moment to re-record for a sweeping change w/some of the dynamic data we have. We can improve that in the future, but for this change, manually updating the cassettes to reflect this new header might be simplest: https://github.com/epignosisx/vcr-test#how-do-i-update-an-existing-cassette-because-of-a-change

drtangible (approved):
👍 I'm happy to own updating the cassettes as a followup if helpful!

ro31337:
@drtangible thanks! Would appreciate it if you could find time to look at thse tests!

PR #7361, opened 2026-06-23
test(weave): assert complete SQL in query-builder tests

gtarpenning:
## Summary
- converts the SQL-string substring/membership assertions in `tests/trace_server/` (82 across 11 files) to assert on the complete query + params, reusing the existing `assert_sql`/`assert_raw_sql`/`assert_stats_sql`/`assert_clickhouse_sql` helpers where they apply.
- enforces the new "assert on the complete payload" test rule (#7355) for the SQL category. exception-message / ref-URI substrings are intentionally left.

## Testing
test only change


jwlee64 (approved):
lgtm
I do like having the full sql in the tests even though it makes them more brittle
Helps ai and code reviewers figure out what the intended sql output is

PR #7364, opened 2026-06-23
feat(weave): add predict-only token total to eval results summary

ro31337:
See https://github.com/wandb/core/pull/45031 for the context/screenshots to understand why we need this change. Short version - the API should expose the per-evaluation predict-only token total as one number, so the frontend can read it instead of pulling every row and summing them in the browser.

The bug: in Compare Evaluations a model's Total Tokens is too high when the eval has an LLM judge - the judge's tokens get counted as the model's. Same model and dataset, but the number jumps from 97 to 243 once you add two LLM scorers. It should only count the model's own predict() tokens.

The per-trial number is already predict-only; the API just never summed it per evaluation. That's why #45031 had to fetch all the rows and add them up on the client, which is slow on an already heavy page. So here I do the sum on the server and return it as `predict_total_tokens` on each evaluation in the summary. The summary code already loops over the trials, so I just add them up in the same place - no extra query. It's null when there's no token data, so the caller falls back to the old value and nothing breaks.

Backend only. The frontend PR comes next - it reads this field and drops the row fetch.

Jira: https://coreweave.atlassian.net/browse/WB-35134

Testing: three tests - judge tokens excluded (60, not 260), null when there are no tokens, and a real 0 stays 0 (not null).

### DIFF
--- a/weave/trace_server/trace_server_interface.py
+++ b/weave/trace_server/trace_server_interface.py
@@ -3227,6 +3227,7 @@ class EvalResultsScorerStats(BaseModel):
 class EvalResultsEvaluationSummary(BaseModel):
     evaluation_call_id: str
     trial_count: int = 0
+    predict_total_tokens: int | None = None
     scorer_stats: list[EvalResultsScorerStats] = Field(default_factory=list)
     evaluation_ref: str | None = None
     model_ref: str | None = None
--- a/weave/trace_server/trace_server_interface.py
+++ b/weave/trace_server/trace_server_interface.py
@@ -3227,8 +3227,15 @@ class EvalResultsScorerStats(BaseModel):
 class EvalResultsEvaluationSummary(BaseModel):
     evaluation_call_id: str
     trial_count: int = 0
-    predict_total_tokens: int | None = None
     scorer_stats: list[EvalResultsScorerStats] = Field(default_factory=list)
+    predict_total_tokens: int | None = Field(
+        default=None,
+        description=(
+            "Sum of per-trial predict-only token usage for this evaluation "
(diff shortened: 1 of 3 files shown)


weave/trace_server/trace_server_interface.py
```diff
@@ -3227,6 +3227,7 @@ class EvalResultsScorerStats(BaseModel):
 class EvalResultsEvaluationSummary(BaseModel):
     evaluation_call_id: str
     trial_count: int = 0
+    predict_total_tokens: int | None = None
```
nikumar1206: tiny nit:
this pydantic model will be present on our openapi schema.
i think when we add new attributes, we should include `Field(description=)` for both AI and humans to better understand what the attribute does.

nikumar1206 (approved):

nikumar1206:
lgtm, just check if @bcsherma  has any thoughts. he wrote most of /eval_results.

weave/trace_server/trace_server_interface.py
bcsherma: another tiny nit (so small it's existence can be debated) but if we could have the required attributes in a continuous block of lines and then the optional ones in a continuous block it would be nice. in other words, move `predicate_total_tokens` declaration below `scorer_stats`

bcsherma (approved):

weave/trace_server/trace_server_interface.py
ro31337: Thanks, will fix them. I decided not to add it Field(description=...) to match the sibling field -- EvalResultsTrial.total_tokens, will add it.

I'll follow the convention and will update field ordering

PR #7374, opened 2026-06-24
perf(weave): prune partitions on calls/delete with a started_at

gtarpenning:
## Summary
- `calls/delete` filtered only by `trace_id`/`id`, which prune neither the `calls_merged` sort key nor the `calls_complete` partition. On large projects (found via Datadog RCA) the descendant read OOMed (32 GiB `AggregatingTransform`) and the `calls_complete` DELETE scanned every partition (`TOO_SLOW` / `QUERY_WAS_CANCELLED`).
- Descendants start at/after their ancestors, so bound the trace read by `min(requested started_at)` and bracket the `calls_complete` DELETE by the `[min, max] started_at` of the rows being deleted (padded for clock skew).

## Performance
Same delete predicate on a real 31,446-call subtree (prod, 3-month project), `SELECT count()` before/after the `started_at` bracket:

| metric | before | after | gain |
|---|---|---|---|
| partitions scanned | 3 (all) | 1 | pruned |
| granules read | 3,024 | 36 | 84× |
| rows read | 3,992,687 | 48,385 | 82× |
| bytes read | 77.8 MB | 3.2 MB | 24× |

## Testing
query-builder unit test for the windowed DELETE SQL, plus functional cascade-delete tests on both calls_complete and calls_merged residence (descendants starting minutes after the root still fully delete).


nikumar1206:
would be cool to have some sort of before/after of the perf gains from this!

nikumar1206 (approved):

gtarpenning:
> would be cool to have some sort of before/after of the perf gains from this!

@nikumar1206  added my testing on a real prod failure to the pr description

PR #7375, opened 2026-06-24
chore(weave): expand docstrings for Weave Agents Python surface (v4)

dbrian57:
## Description

Fourth pass over the same surface as [#7362](https://github.com/wandb/weave/pull/7362), [#7363](https://github.com/wandb/weave/pull/7363), and [#7366](https://github.com/wandb/weave/pull/7366). This one is produced by `/sdk-docs` **v4** — the skill version that integrates the four authoring rules from `/docstring-writer` (workflow mapping, pinned `Note:` placement, "what not to document", `Yields:` for generators) shipped upstream in [coreweave/agent-plugins#40](https://github.com/coreweave/agent-plugins/pull/40).

Opened as **draft** for comparison only. Merge at most one of #7362, #7363, #7366, or this PR.

## What's different from #7366 (v3)

**Effectively nothing on this surface.** The v4 → v3 diff is +7 lines net (1795 vs 1788) and the prose is near-identical. The four new rules barely fire here:

| v4 rule | Impact on this PR |
|---|---|
| **Step 1.5: Map the workflow** | Formalizes work I was already doing during v3. The visible difference is that the workflow map appears in the user-facing turn before the edits, not in the final docstrings. |
| **Pinned `Note:` placement** | v3 already placed `Note:` after the extended description and before `Args:` consistently. |
| **What not to document** | Didn't apply. No trivial property accessors or redundant `__init__` docstrings in scope. |
| **`Yields:` for generators** | Didn't apply. No generator functions in scope. |

The v4 skill is the better tool for future work (especially codebases with generators or redundant `__init__`s), but on this specific surface — Pydantic data models + Protocol interface methods — v3 and v4 are interchangeable.

## Scope

Same three files as the prior PRs:

- `weave/trace_server/agents/types.py` — module docstring + ~30 public Pydantic request/response/payload types.
- `weave/trace/refs.py` — `AgentTurnRef`, `AgentConversationRef`, `AgentSpanRef` and their `parse_uri` methods.
- `weave/trace_server/trace_server_interface.py` — the nine `agent_*` / `genai_otel_export` Protocol methods.

## Testing

- `uvx ruff check` — clean.
- `uv run --group test python -m pytest tests/trace_server/test_genai_agent_queries.py tests/trace_server/test_genai_chat_view.py --trace-server=fake` — 24 passed, 27 skipped.
- Smoke import: every newly documented symbol has a populated `__doc__`.

## Reviewer note

Commit title accidentally reads "(v3)" because of a typo I caught after committing — the PR title and this body are correct as "(v4)". Following the repo's commit policy of preferring new commits over amending, I left the typo rather than rewriting history; happy to land a follow-up "chore: fix v4 commit message" if it matters before merge.

### DIFF
--- a/weave/trace_server/trace_server_interface.py
+++ b/weave/trace_server/trace_server_interface.py
@@ -3246,30 +3246,30 @@ class TraceServerInterface(Protocol):
 
     # GenAI / Agent Observability API
     #
-    # Contract every backend (ClickHouse, in-memory fake, remote HTTP)
-    # implements for the Agents view. Workflow:
+    # The API behind the Agents view. A typical workflow:
     #
-    # 1. Ingest spans via `genai_otel_export`.
-    # 2. List agents with `agent_agents_query`, drill into one agent's
-    #    versions with `agent_versions_query`, and into its conversations
-    #    by calling `agent_spans_query` in group mode with
-    #    `group_by=[conversation_id]`.
-    # 3. Open a conversation with `agent_conversation_chat`; open a single
-    #    turn with `agent_traces_chat`.
-    # 4. Side queries: `agent_spans_stats` for charts,
-    #    `agent_custom_attrs_schema` for filter builders, `agent_search`
-    #    for content or structured retrieval over messages.
+    # 1. Ingest spans with `genai_otel_export`.
+    # 2. List agents with `agent_agents_query`. Drill into one agent's
+    #    versions with `agent_versions_query`, or into its conversations
+    #    with `agent_spans_query` in group mode (group by
+    #    `conversation_id`).
+    # 3. Open a conversation with `agent_conversation_chat`, or a single
+    #    turn within it with `agent_traces_chat`.
+    # 4. Supporting queries: `agent_spans_stats` (charts),
+    #    `agent_custom_attrs_schema` (filter builders), and
+    #    `agent_search` (search messages by content and/or filters).
     #
     # User-facing concepts: https://docs.wandb.ai/weave/guides/tracking/trace-agents
     def genai_otel_export(
         self, req: agent_types.GenAIOTelExportReq
     ) -> agent_types.GenAIOTelExportRes:
-        """Ingest GenAI OTel spans into the agent observability store.
+        """Receives OpenTelemetry GenAI/agent spans exported to Weave and
+        ingests them.
 
-        The single write-side entry point. Spans land in the agent
-        spans table and refresh the downstream materialized views
-        (agents, agent_versions, messages) that the read-side
-        endpoints query.
+        Spans are written to the spans table; ClickHouse materialized
+        views defined on that table then populate the downstream summary
+        tables (agents, agent_versions, and messages) which back the
+        corresponding read-side query endpoints.
 
         Note:
             Rejected spans don't block the rest of the batch. Inspect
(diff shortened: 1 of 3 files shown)


weave/trace_server/trace_server_interface.py
```diff
@@ -3245,33 +3245,298 @@ class TraceServerInterface(Protocol):
...
-    ) -> agent_types.AgentVersionsQueryRes: ...
+    ) -> agent_types.AgentVersionsQueryRes:
+        """List the versions seen for an agent.
+
+        Run after `agent_agents_query` to diff behavior across
+        releases of one agent. Returns one `AgentVersionSchema` per
+        `(agent_name, agent_version)` pair.
+
+        Args:
+            req: Project, target `agent_name`, sort, pagination, and
```
anastasiaguspan, line 3445: SME REVIEW: This documents `sort` as an input, but `req.sort_by` is silently ignored for `agent_versions_query` — so the docstring (and the `AgentVersionsQueryReq.sort_by` field, whose own doc says "Sort steps. Allowed fields validated server-side") advertise a capability the implementation doesn't honor.

Why I believe it's a bug rather than intended behavior:

- `make_agent_versions_list_query` hardcodes `ORDER BY last_seen DESC` and never references `req.sort_by` / calls `build_order_by(...)` (`query_builder/agent_query_builder.py`, the agent-versions list query).
- The sibling `make_agents_list_query` *does* wire it through: `build_order_by(req.sort_by, AGENT_SORTABLE_COLS, "last_seen DESC, agent_name")`. So the agents endpoint honors `sort_by` and the versions endpoint doesn't — looks like a wiring gap that was missed for versions, not a deliberate omission.
- `_run_paginated` doesn't apply any ordering either, so there's no other code path that could be applying `sort_by`.

Not blocking this docs PR (the bug is in the query builder, not in this file). Flagging so the `sort_by` plumbing can be fixed, or the docs/field softened to match current behavior. Separately, minor: "one `AgentVersionSchema` per `(agent_name, agent_version)` pair" — since the query is filtered to the single required `agent_name` and groups only by `agent_version`, it's effectively one row per `agent_version` of the requested agent.

weave/trace_server/trace_server_interface.py
```diff
@@ -3245,33 +3245,298 @@ class TraceServerInterface(Protocol):
...
+        ...
     def agent_search(
         self, req: agent_types.AgentSearchReq
-    ) -> agent_types.AgentSearchRes: ...
+    ) -> agent_types.AgentSearchRes:
+        """Search agent messages by content or structured filters.
+
+        Backed by the messages MV. Two modes:
+
+        1. **Full-text search** — set `req.query` to a substring.
```
anastasiaguspan: SME REVIEW: Heads up that this "full-text search" is actually a **case-sensitive substring match**, which may be a bug (or at least a surprising default worth confirming).

How I determined it:

- `req.query` is compiled in `_search_filter_sql` (`query_builder/agent_query_builder.py`) to `content LIKE {slot}`, where the bound parameter is `f"%{_escape_like_pattern(req.query)}%"`. So it's a literal `%query%` substring scan — no tokenization/ranking, hence not really "full-text."
- It uses ClickHouse `LIKE`, not `ILIKE`. `LIKE` is case-sensitive in ClickHouse, so the matching is case-sensitive.

Why I suspect it's a bug: the `AgentSearchReq` example searches `query="rate limit"`, which would **not** match content like "Rate limit" or "RATE LIMIT". For a user-facing message search box that's almost certainly not the expected behavior — case-insensitive (`ILIKE`, or `lower(content) LIKE lower(...)`) is the usual expectation. 

If case-sensitivity is intentional, I captured it in the current docstring; if not, switching `LIKE`→`ILIKE` in `_search_filter_sql` would fix it, and then we can update the docstring.

PR #7384, opened 2026-06-24
feat(weave): add agent read methods to WeaveClient

rgao-coreweave:
## Summary

Adds eight agent-observability **read** methods to the Python `WeaveClient`, wrapping the agent read endpoints and mirroring the TS SDK. Each infers `project_id`. The list methods (`get_agents` / `get_agent_versions` / `get_agent_spans`) return a `PaginatedIterator` (like `get_calls`); the rest return the read-side `*Res` shape.

## Methods

| `WeaveClient` method | Wraps | Returns |
|---|---|---|
| `get_agents` | `agent_agents_query` | `PaginatedIterator[AgentSchema]` |
| `get_agent_versions` | `agent_versions_query` | `PaginatedIterator[AgentVersionSchema]` |
| `get_agent_spans` | `agent_spans_query` | `PaginatedIterator[AgentSpanSchema]` |
| `get_agent_turn` | `agent_traces_chat` | `AgentTraceChatRes` |
| `get_agent_turns` | `agent_conversation_chat` | `AgentConversationChatRes` |
| `get_agent_span_stats` | `agent_spans_stats` | `AgentSpanStatsRes` |
| `get_agent_custom_attributes` | `agent_custom_attrs_schema` | `AgentCustomAttrsSchemaRes` |
| `search_agents` | `agent_search` | `AgentSearchRes` |

- **Pagination:** `get_agents` / `get_agent_versions` / `get_agent_spans` return a `PaginatedIterator`, matching the `get_calls` standard — iterate to stream pages, `len()` for the total count, indexing/slicing and `.to_pandas()` supported. `limit` / `offset` / `page_size` follow `get_calls` semantics (`limit` caps the total, `None` = all). `search_agents` stays page-based because its `offset`/`limit` paginate matched *messages*, not the returned conversations.
- A *turn* is one trace: `get_agent_turn(trace_id)` returns one trace's chat; `get_agent_turns(conversation_id)` returns all turns in a conversation (mirrors TS `getAgentTurn`/`getAgentTurns`).
- Curated kwargs per method (common path); advanced fields stay on `client.server.agent_*`.
- `search_agents` is verb-first — rename to `get_agent_search` if you prefer the `get_` prefix.

## Testing

Fake-server unit tests (project_id inference, pass-through, defaults, pagination / `len` / `limit` / `offset`, `agent_name` → query translation) + end-to-end ClickHouse coverage for all eight (including the iterator round-trip via `list()` / `len()`). `ruff` + `ty` clean locally; `pyright` / `mypy` run in CI.


scripts/validate_agent_read_sdk.py
andrewtruong, line 1: Why is this file here?

tests/trace/test_client_agent_read_sdk.py
```diff
@@ -0,0 +1,258 @@
...
+"""
+
+from __future__ import annotations
+
+import datetime
+
+import pytest
+
+from weave.trace.weave_client import WeaveClient
+from weave.trace_server.agents.types import (
```
andrewtruong: Are these not exposed through trace_server_interface?  I think they should be

weave/trace/weave_client.py
```diff
@@ -260,6 +261,22 @@ def _add_scored_by_to_calls_query(
     return Query.model_validate({"$expr": {"$and": exprs}})
 
 
+def _agent_spans_query_filter(
```
andrewtruong: why is this called `query_filter`?

weave/trace/weave_client.py
andrewtruong, line 1: Should we make the methods here more helpful?  For example, could they benefit from pagination, etc?  See `PaginatedIterator`

scripts/README.md
```diff
@@ -12,3 +13,14 @@ To run a benchmark, invoke `uv run benchmark.py`. This will bring up a CLI for s
 ## 2. Stripping EXIF data from images
 
 To run the prek hook, invoke `nox -e lint`.
+
+## 3. Validating the agent read SDK
```
andrewtruong: dont think these belong here

weave/trace/weave_client.py
```diff
@@ -755,6 +772,355 @@ def get_call(
...
+                query=query,
+                group_by=group_by or [],
+                granularity=granularity,
+                timezone=timezone,
+            )
+        )
+
+    @trace_sentry.global_trace_sentry.watch()
+    @pydantic.validate_call
+    def get_agent_custom_attrs_schema(
```
drtangible: Nit: when possible, think it's good to not use abbreviations in customer-facing interfaces. Assume in the UI we spell out "attributes".

drtangible: Also is `schema` necessary for a user here? Would `get_agent_custom_attributes` be friendlier?

scripts/validate_agent_read_sdk.py
rgao-coreweave: I wanted to add the test script. but removed.

scripts/README.md
rgao-coreweave: removed.

tests/trace/test_client_agent_read_sdk.py
rgao-coreweave: gotcha, importing from `trace_server_interface`

weave/trace/weave_client.py
rgao-coreweave: Renamed to _build_agent_spans_query

weave/trace/weave_client.py
rgao-coreweave: I like `get_agent_custom_attributes`, makes sense to rename.

Please also update the typescript.

andrewtruong (approved):

weave/trace/weave_client.py
rgao-coreweave: Done, great suggestion!

PR #7382, opened 2026-06-24
chore(weave_ts): Update jsdocs on top-level `startFoo`/`endFoo` functions.

drtangible:
## Description

* Adds add'l documentation to our top-level API functions, so more can be extracted into our reference material on docs.wandb.ai.

PR #7381, opened 2026-06-24
feat(weave_ts): Support passing `endTime` to `weave.endTurn` and `weave.endLLM`

drtangible:
## Description

* Our top-level `startFoo` functions currently accept `startTime`.

* This makes our existing `endFoo` functions similar, by accepting an `endTime`.

* Also adds a test covering backfilling data with this form of our API.

PR #7378, opened 2026-06-24
fix(weave_ts): Respect `startTime` given to `Session.startTurn`.

drtangible:
## Description

* We accept `startTime` and `endTime` params to relevant functions for instrumenting agents. This allows customers to backfill data into weave with appropriate timestamps.

* There's currently a bug where `startTime` given to `Session.startTurn` is silently ignored. This make backfilled data appear incorrectly in the UI. In this example, a backfilled `chat` span from May 29th corresponds to an `invoke_agent` span from today:

Screenshot 2026-06-24 at 12 03 29 PM

Screenshot 2026-06-24 at 12 03 34 PM

* This change fixes that bug.

Screenshot 2026-06-24 at 12 13 11 PM

Screenshot 2026-06-24 at 12 13 14 PM

PR #7385, opened 2026-06-24
feat(weave_ts): Accept `endTime` on `weave.endSession` and `Session.end`

drtangible:
## Description

* While our "session" abstraction is not an OTel span itself, it does automatically `end` child `llm`s and `turn`s, when not already explicitly `end`-ed.

* This change update `weave.endSession` and `Session.end` to accept options to pass down to those instances when ending them.

### DIFF
--- a/sdks/node/src/genai/session.ts
+++ b/sdks/node/src/genai/session.ts
@@ -2,6 +2,7 @@ import {uuidv7} from 'uuidv7';
 
 import {_getGenaiState} from './context';
 import {Turn, type TurnInit} from './turn';
+import type {SpanEndOptions} from './spanBase';
 
 export interface SessionInit {
   agentName?: string;
@@ -49,7 +50,7 @@ export class Session {
     });
   }
 
-  end(): void {
+  end(opts?: SpanEndOptions): void {
     if (this._ended) {
       return;
     }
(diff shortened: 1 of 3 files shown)


drtangible:
* **#7385** <a href="https://app.graphite.com/github/pr/wandb/weave/7385?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7385?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7382** <a href="https://app.graphite.com/github/pr/wandb/weave/7382?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7381** <a href="https://app.graphite.com/github/pr/wandb/weave/7381?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7378** <a href="https://app.graphite.com/github/pr/wandb/weave/7378?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sdks/node/src/genai/session.ts
```diff
@@ -49,7 +50,7 @@ export class Session {
     });
   }
 
-  end(): void {
+  end(opts?: SpanEndOptions): void {
```
rgao-coreweave, line 53: since we are adding a new param, should we also include a jsdoc for it now?

example: https://github.com/wandb/weave/blob/5e291b179f35a2c01f2dec30c799d483cdbcbba7/sdks/node/src/genai/tool.ts#L80-L84

rgao-coreweave (approved):
LGTM 👍👍👍👍

sdks/node/src/genai/session.ts
drtangible: Good idea -- the other functions in here don't have jsdocs yet, so will follow up with a pass through them all.

PR #7386, opened 2026-06-24
feat(weave_ts): default JS client to calls_complete ingest path

gtarpenning:
## Summary
- Default the JS client to the `calls/complete` ingest path: each call start is paired with its end client-side and written as one row to `/v2/{entity}/{project}/calls/complete` (was `call/upsert_batch` / calls_merged). Mirrors the Python client.
- `eagerCallStart` is **SDK-internal** (`markOpEager`), not a public op option, and is set only on eval roots (mirrors Python's `eager_call_start`, which is set only on `eval.py` / `eval_imperative.py`). It sends a start immediately via the v2 single endpoints so a long-running eval is visible before it finishes. Streaming adapters (realtime, openai-agents) are no longer eager: their open spans pair into `calls/complete` and land on close (or via flush on disconnect/shutdown).
- Auto-upgrades to the complete path on a `CALLS_COMPLETE_MODE_REQUIRED` server error; unpaired starts/ends flush via the v2 single endpoints on shutdown so interrupted calls are not dropped.
- **Error handling** (both ingest paths, not gated by any flag): the SDK no longer calls `process.exit()`. After exhausting retries it disables tracing and lets the host process continue (previously a sustained outage killed the user's program and skipped the flush). Non-retryable 4xx are dropped instead of retried forever.
- The `calls/complete` ingest path itself is gated by `useCallsComplete` / `WEAVE_USE_CALLS_COMPLETE` (default on, kept behind the flag for old servers, same as Python).

## Testing
unit: `callsComplete.test.ts` (pairing, eager, legacy, upgrade, out-of-order, flush, non-retryable drop, graceful give-up); realtime/agent integration suites; hostApps (Python mock gained the v2 routes). all green.

prod e2e against `trace.wandb.ai`, read-back verified (Datadog samples, so not source of truth): new project, eager op (v2 start+end), legacy+complete coexisting in one project, real server-triggered auto-upgrade, interrupted-call flush, 250 concurrent, malformed/unicode.

**identical UI, different table:** logged the same trace tree to two fresh projects that differ only by `WEAVE_USE_CALLS_COMPLETE`. Renders identically; lands in different physical tables (prod CH, `count(DISTINCT id)`):

| project | `calls_merged` | `calls_complete` |
|---|---|---|
| `js-cc-demo-legacy` (flag off) | 20 | 0 |
| `js-cc-demo-complete` (flag on) | 0 | 20 |

smoketest project: https://wandb.ai/griffin_wb/js-calls-complete-smoke/weave/traces?view=traces_default


sdks/node/src/weaveClient.ts
```diff
@@ -576,40 +637,197 @@ export class WeaveClient {
...
+  private sendCallStartV2(start: CallStartParams) {
+    return this.postV2('call/start', {start});
+  }
+
+  private sendCallEndV2(end: CallEndParams) {
+    // The v2 end schema has no display_name (post-start renames go via
+    // updateCall, matching the Python client); strip it before sending.
+    const {display_name: _displayName, ...endReq} = end;
+    return this.postV2('call/end', {end: endReq});
+  }
```
drtangible, line 1112: Are these endpoints not in the server OpenAPI spec? It would be nice to use the generated type-safe functions.

drtangible (commented):
> Error handling: the SDK no longer calls process.exit(). After exhausting retries it disables tracing and lets the host process continue (previously a sustained outage killed the user's program and skipped the flush). Non-retryable 4xx are dropped instead of retried forever; gated by useCallsComplete / WEAVE_USE_CALLS_COMPLETE (default on, kept behind the flag for old servers, same as Python).

Is this suggesting that the error handling is intended to be gated behind the setting / env var? Or is the last sentence in here a separate unrelated note describing this feature as a whole?

sdks/node/src/weaveClient.ts
gtarpenning: the `v2/call/start` and `v2/call/end` are intentionally hidden until we can rate limit them. If at all possible we want consumers of the raw endpoints to just use the `calls/complete` endpoint, which _should_ be public...

gtarpenning: This commit: https://github.com/wandb/weave/pull/7386/commits/b124f48e95278d33465e797e1ebe5469d8c0fab4 should address most of your concern here I think

sdks/node/src/clientApi.ts
```diff
@@ -160,6 +161,36 @@ export async function init(
   }
 }
 
+/** Deliver all buffered calls to the server. Await before `process.exit()`. */
+export async function flush(): Promise<void> {
+  await getGlobalClient()?.flush();
+}
```
drtangible, line 167: Is this function used beyond including it in the top-level public API? Trying to understand if we need / want to make that addition.

sdks/node/src/op.ts
```diff
@@ -5,6 +5,16 @@ import {type Op, type OpOptions, type OpRef, type CallMethod} from './opType';
 import {getGlobalDomain} from './urls';
 import {warnOnce} from './utils/warnOnce';
 
+// Internal registry of ops whose call start should be sent eagerly (immediately
+// via the v2 single endpoints) instead of paired into a calls_complete write.
+// Reserved for SDK integrations with long-running spans that must be visible
+// before they finish (evaluations, streaming tracers). Not a public op option.
+const eagerOps = new WeakSet<object>();
```
drtangible, line 12: Assuming `object` could be more precisely typed as `Op<any>` here?

sdks/node/src/evaluationLogger.ts
```diff
@@ -633,7 +633,8 @@ export class EvaluationLogger {
       undefined, // No parent (root call)
       startTime,
       undefined, // displayName
-      {...IMPERATIVE_EVAL_MARKER, ...this.evalAttributes} // Merge eval attributes
+      {...IMPERATIVE_EVAL_MARKER, ...this.evalAttributes}, // Merge eval attributes
+      true // eagerCallStart: eval root must be visible before it finishes
```
drtangible, line 637: Just note for myself for later: these comments are a smell that this should be moved to named options rather than 11 positional args.

sdks/node/src/clientApi.ts
drtangible: Ah, I see it referenced in the warning message below. Maybe worth adding a test, so something at least yells at us if it's removed? Or maybe just let the `beforeExit` callback call this fn directly?

drtangible (approved):
👍 👍 👍

PR #7390, opened 2026-06-24
fix(weave): stop integrations churning object versions via repr() memory addresses

gtarpenning:
## Summary
- `dspy`, `gepa`, and `langchain` logged raw `repr()` of functions/objects (e.g. a dspy metric) into published content. `repr()` embeds `at 0x...` memory addresses that change every process, so content-addressed objects churned a new version every run (one prod project logged 20k+ `LLMJudge` versions in 24h, defeating dedup).
- Add `stable_repr()` in the central serializer (strips ` at 0x...`) and route every integration `repr()` site through it.
- Jira: [WB-36408](https://coreweave.atlassian.net/browse/WB-36408)

## Testing
unit test for `stable_repr` + gepa logging-helper coverage; dspy/langchain covered via the shared helper.

[WB-36408]: https://coreweave.atlassian.net/browse/WB-36408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace/serialization/serialize.py
+++ b/weave/trace/serialization/serialize.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import logging
-import re
 from collections.abc import Mapping, Sequence
 from types import CoroutineType
 from typing import TYPE_CHECKING, Any, Literal, Protocol, TypeAlias, TypedDict
@@ -17,7 +16,11 @@ from weave.trace_server.trace_server_interface import (
     FileCreateReq,
     TraceServerInterface,
 )
-from weave.utils.sanitize import REDACTED_VALUE, should_redact
+from weave.utils.sanitize import (
+    REDACTED_VALUE,
+    should_redact,
+    strip_memory_addresses,
+)
 
 if TYPE_CHECKING:
     from weave.trace.weave_client import WeaveClient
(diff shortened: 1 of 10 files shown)


weave/trace/serialization/serialize.py
```diff
@@ -294,6 +295,14 @@ def stringify(obj: Any, limit: int = MAX_STR_LEN) -> str:
     return rep
 
 
+_MEMORY_ADDRESS_RE = re.compile(r" at 0x[0-9a-fA-F]+")
+
+
+def stable_repr(obj: Any) -> str:
```
andrewtruong, line 301: I thought we already had this somewhere?  Can you use that one instead?

andrewtruong: Or if this impl is better then replace the other one

gtarpenning: damn, yeah i forgot about that, this impl is slightly nicer I think, will DRY it out

gtarpenning: fixed

andrewtruong (approved):

PR #7391, opened 2026-06-25
feat(weave): send is_eval on call-end ingest messages

ro31337:
## Description

A future server-side ingest sampler will decide, per call, whether to keep or drop it. For most messages it can tell an evaluation call from a normal one by looking at the call's attributes and op name. The problem is the call-END: it often arrives "bare" — carrying only ids, no attributes and no op name — so the server has nothing to recompute from. The eval root's end is always bare, so without a marker the sampler could drop the very calls that make up an evaluation.

This adds an optional `is_eval` boolean to the call-end message. The SDK already has the attributes and op name at finish time, so it computes the same predicate the server would and stamps the answer on the end. The field is optional and the server does not read it yet, so this change is inert until the sampler ships — there is no behavior change today.

The predicate is `is_eval = (eval metadata present in attributes) or (op name is the evaluation root)`. Eval children carry the metadata in their attributes; the declarative root does not (its own attributes are read before the wrapper that tags children runs), so the root is recognized by its op name. Because `op_name` is a ref URI like `weave:///entity/project/op/Evaluation.evaluate:digest`, the op-name check is a substring match — the same check the server already uses in `evaluation_status`.

## Why this approach

The shape mirrors #7237 (which put `trace_id` on the call-end for the same sampler): one optional field, computed at the single SDK point that builds the end (`send_end_call`).

Note the asymmetry with `trace_id`: the server cannot derive `trace_id`, so #7237 carries it on every message including the merged complete-record; `is_eval` it can derive from `op_name`+`attributes`, so the field is needed only on the bare call-end. Call-starts and Python's merged complete-records already carry both, and the server recomputes there.

## Testing

Integration tests run a real declarative evaluation (trials = 2) and an imperative `EvaluationLogger`, capture the outgoing call-ends, and assert the root and children come out `True` while a plain op comes out `False`; a binding test asserts the flag is serialized into the call-end body. Verified locally on both the in-memory and ClickHouse backends (43 passed across `test_evaluations.py` and `test_http_behavior_remote.py`); ruff and import-linter clean.

## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-36097


weave/trace/weave_client.py
```diff
@@ -1254,11 +1255,23 @@ def send_end_call() -> None:
...
             )
 
+            # Mark eval calls (root by op_name substring, children by attribute).
+            # op_name can raise if the op never saved, so fall back to attrs only.
+            try:
+                op_name = call.op_name
+            except Exception:
+                op_name = ""
+            is_eval = (
+                EVAL_META_KEY in (call.attributes or {})
```
andrewtruong: this should be a helper func

andrewtruong (approved):

PR #7392, opened 2026-06-25
feat(weave_ts): send is_eval on call-end ingest messages

ro31337:
## Description

A future server-side ingest sampler will decide, per call, whether to keep or drop it. It can tell an evaluation call from a normal one by looking at the call's attributes and op name — except on the call-END, which arrives carrying only ids, with no attributes or op name to recompute from. In the Node SDK every end is sent separately from its start, so this gap applies to every eval call, not just the root.

This adds an optional `is_eval` boolean to the call-end. At finish time the SDK still has the call's attributes and op name, so it computes the same predicate the server would and stamps the answer on the end. The field is optional and the server does not read it yet, so this change is inert until the sampler ships — no behavior change today.

The predicate is `is_eval = (eval metadata present in attributes) or (op name is the evaluation root)`. Eval children carry the metadata in their attributes; the declarative root does not, so it is recognized by its op name. Because `op_name` is a ref URI, the op-name check is a substring match. It is computed in `finishCall` and `finishCallWithException` from the call's own stored schema (not the live AsyncLocalStorage context, which under concurrency could return a sibling's attributes).

This is the Node companion to the Python SDK change in #7391.

## Why this approach

The shape mirrors #7257 (which put `trace_id` on the call-end for the same sampler): an optional field on `EndedCallSchemaForInsert`, filled where the call object is in scope. As in #7257, the generated `traceServerApi.ts` and `weave.openapi.json` are hand-edited rather than regenerated, so unrelated generator drift stays out of the diff; the edit matches what regeneration produces. The shared predicate is a single private method, so the two finish paths cannot drift apart.

## Testing

A new test runs a real declarative evaluation and inspects the outgoing upsert batch (the same spy approach #7257 uses), asserting the root end is `is_eval = true` (matched by op name) and every eval end is marked; a separate op-flow test asserts a plain non-eval call comes out `false`. Verified locally: the two suites pass (14 tests), and typecheck, prettier, and eslint are clean.

## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-36097

### DIFF
--- a/sdks/node/src/weaveClient.ts
+++ b/sdks/node/src/weaveClient.ts
@@ -2,7 +2,11 @@ import {AsyncLocalStorage} from 'async_hooks';
 import * as fs from 'fs';
 import {uuidv7} from 'uuidv7';
 
-import {MAX_OBJECT_NAME_LENGTH} from './constants';
+import {
+  EVAL_META_KEY,
+  EVALUATION_RUN_OP_NAME,
+  MAX_OBJECT_NAME_LENGTH,
+} from './constants';
 import {computeDigest} from './digest';
 import {
   type AgentChatMessage as AgentChatMessageSchema,
@@ -1354,6 +1358,16 @@ export class WeaveClient {
     return this.saveCallStart(startReq);
   }
 
+  // The eval root is matched by op-name substring (op_name is a ref URI); eval
+  // children carry the marker in their attributes.
+  private isEvalCall(call: InternalCall): boolean {
+    const {attributes, op_name} = call.callSchema;
+    return (
+      (attributes != null && EVAL_META_KEY in attributes) ||
+      (op_name?.includes(EVALUATION_RUN_OP_NAME) ?? false)
+    );
+  }
+
   public async finishCall(
     call: InternalCall,
     result: any,
@@ -1384,6 +1398,7 @@ export class WeaveClient {
       project_id: this.projectId,
       id: currentCall.callId,
       trace_id: currentCall.traceId,
+      is_eval: this.isEvalCall(call),
       ...callSchemaExchangeData,
       // User might change the display name of the call after the call has started.
       // take this into account when logging the end call.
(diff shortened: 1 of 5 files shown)


sdks/node/weave.openapi.json
```diff
@@ -11249,6 +11249,17 @@
...
+          "is_eval": {
+            "anyOf": [
+              {
+                "type": "boolean"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Is Eval"
```
drtangible, line 11261: Am I missing this in https://trace.wandb.ai/openapi.json? Ideally we can copy over the spec entirely. If we are worried about that making it more difficult for reviewers, a separate PR just dedicated to bumping with the latest and regenerating `traceServerApi.ts`can be helpful.

drtangible: Looks like that change would eventually make its way to the spec via #7391. Since non-breaking, probably okay -- we'll re-add this manually if we bump the spec again prior to that actually landing.

sdks/node/src/weaveClient.ts
```diff
@@ -1354,6 +1358,16 @@ export class WeaveClient {
...
 
+  // The eval root is matched by op-name substring (op_name is a ref URI); eval
+  // children carry the marker in their attributes.
+  private isEvalCall(call: InternalCall): boolean {
+    const {attributes, op_name} = call.callSchema;
+    return (
+      (attributes != null && EVAL_META_KEY in attributes) ||
+      (op_name?.includes(EVALUATION_RUN_OP_NAME) ?? false)
+    );
+  }
```
drtangible: Looks like we could pull this out of the class and make it a normal helper function in this module.

sdks/node/src/weaveClient.ts
```diff
@@ -1354,6 +1358,16 @@ export class WeaveClient {
     return this.saveCallStart(startReq);
   }
 
+  // The eval root is matched by op-name substring (op_name is a ref URI); eval
+  // children carry the marker in their attributes.
+  private isEvalCall(call: InternalCall): boolean {
+    const {attributes, op_name} = call.callSchema;
+    return (
+      (attributes != null && EVAL_META_KEY in attributes) ||
```
drtangible: Might be a bit more common to check via `attributes[EVAL_META_KEY]` rather than `in`.

drtangible (commented):
👍

drtangible (approved):
👍

PR #7399, opened 2026-06-25
chore(weave): add Claude Agent SDK host-app e2e test

rgao-coreweave:
## Summary

Follow-up to #7245 (the Claude Agent SDK integration for the TypeScript SDK). That PR was kept focused on the integration itself; this PR adds the real-package **hostApps end-to-end test** that was split out of it.

## What this adds

A real-package host-app e2e test that runs the actual `@anthropic-ai/claude-agent-sdk` `query()` under `node --import=weave/instrument` — end-to-end through the auto-instrumentation module-loader hooks, not the unit-test stand-ins:

- `src/__tests__/hostApps/claude-agent-sdk.test.ts` — drives the host app and asserts the captured GenAI span tree (`invoke_agent` root → per-message `chat` spans → `execute_tool` spans), `gen_ai.conversation.id`, and usage.
- `src/__tests__/hostApps/fixtures/claude-agent-sdk/main.mjs` — ESM host app that calls `query()`.
- `src/__tests__/hostApps/fixtures/claude-agent-sdk/fake-claude-cli.mjs` — a fake Claude Code CLI (wired via `pathToClaudeCodeExecutable`), so the run is **fully offline — no API key, no real model call**.
- `src/__tests__/hostApps/fixtures/claude-agent-sdk/package.json` — fixture manifest; `@anthropic-ai/claude-agent-sdk` resolves to the copy already vendored in the SDK's `node_modules`, and the `weave` tarball is installed at test time. The harness auto-discovers any `fixtures/*` dir with a `package.json`, so no jest config change is needed.

The `@anthropic-ai/claude-agent-sdk` devDependency itself stays in #7245 — the integration source, example, and unit tests all import types from it, so it isn't host-app-specific.

## Stacking

Based on `feat/ts-claude-agent-sdk` (#7245). The e2e exercises the integration, so it can't compile against `master` until #7245 lands; GitHub will auto-retarget this to `master` once #7245 merges.

## Testing

```
cd sdks/node
pnpm install
pnpm run test:hostApps
```

Fully offline — the harness builds + packs the SDK, installs the tarball into the fixture, spawns the in-memory mock trace server, and runs `query()` against the fake CLI. No API key required.

### DIFF
--- a/sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
+++ b/sdks/node/src/__tests__/hostApps/claude-agent-sdk.test.ts
@@ -7,18 +7,22 @@ import {fixturePath, genProjectId, launchAppFrom} from './utils';
 
 type CapturedSpan = {
   name: string;
-  spanId: string;
-  parentSpanId?: string;
-  traceId: string;
-  attributes: Record<string, unknown>;
-  statusCode: number;
 };
 
+// Smoke test for the Claude Agent SDK integration's packaging + module-loading.
+// Like the other hostApps tests, its job is narrow: confirm that launching a
+// real ESM host app under `node --import=weave/instrument` actually patches
+// `@anthropic-ai/claude-agent-sdk`'s `query()` and emits the integration's root
+// agent span through the real OTel pipeline. The detailed span-tree, usage, and
+// tool-call shape are asserted one layer down in the tracer unit test
+// (src/__tests__/integrations/claude-agent-sdk/otelTracer.test.ts), which drives
+// the tracer directly against mocked SDK messages — cheaper and more exhaustive
+// than re-asserting shape inside a spawned host app.
 describe('hostApps — claude-agent-sdk', () => {
-  test('auto-instruments the real @anthropic-ai/claude-agent-sdk query() and emits a GenAI agent span tree', async () => {
+  test('auto-instruments the real @anthropic-ai/claude-agent-sdk query() and emits the agent root span', async () => {
     const projectId = genProjectId();
     // The fixture writes its captured OTel spans here (it installs a capturing
-    // span processor via weave.init), so we assert the real emission path
+    // span processor via weave.init), so we observe the real emission path
     // without standing up an OTLP receiver.
     const spanFile = path.join(os.tmpdir(), `cas-otel-${randomUUID()}.json`);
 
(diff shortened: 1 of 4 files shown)


drtangible (approved):
I'm not opposed to landing this, but it seems like we should write up the different types of tests we have, so we can clarify what we think the responsibilities of each are.

At the moment, our 2 "host apps" tests seem to be smoke tests trying to catch packaging / module loading issues in CJS vs. ESM. I could see introducing smoke tests at this layer for each integration, though I wonder if we should reduce the scope of what we're testing for.

Similar to the current smoke test behavior:

```ts
describe('hostApps — esm-basic', () => {
  test('import("weave"), run an op, and emit a captured trace', async () => {
    const projectId = genProjectId();
    const result = await launchAppFrom({
      path: fixturePath('esm-basic'),
      projectId,
    });
    if (result.exitCode !== 0) {
      throw new Error(
        `esm-basic exited ${result.exitCode}\n` +
          `stdout:\n${result.stdout}\nstderr:\n${result.stderr}`
      );
    }
    // Confirm the fixture's `myOp` actually executed and emitted a trace
    // the mock captured under our project_id.
    const calls = await getCalls(projectId);
    expect(calls.length).toBeGreaterThanOrEqual(1);
    expect(
      calls.some(
        c => typeof c.op_name === 'string' && c.op_name.includes('myOp')
      )
    ).toBe(true);
  }, 60_000);
});
```

... I wonder if we'd want something more like...


```ts
import {fixturePath, genProjectId, getAgentSpans, launchAppFrom} from './utils';

describe('hostApps — esm-claude-agent-sdk', () => {
  test('import("weave"), run an `@anthropic-ai/claude-agent-sdk` agent, emit an OTel span', async () => {
    const projectId = genProjectId();
    const result = await launchAppFrom({
      path: fixturePath('esm-claude-agent-sdk'),
      projectId,
    });
    if (result.exitCode !== 0) {
      throw new Error(
        `esm-basic exited ${result.exitCode}\n` +
          `stdout:\n${result.stdout}\nstderr:\n${result.stderr}`
      );
    }

    // Confirm the fixture emitted some spans.
    const spans = await getAgentSpans(...);
    expect(spans.length).toBeGreaterThanOrEqual(1);
  }, 60_000);
});
```
... with one for `cjs` as well.


Feels like the bulk of the assertions done on the logic / shape of the emitted spans can be tested at a different layer, since we have mocks/recordings at our disposal that make that a bit easier than within one of these fake apps.

rgao-coreweave:
> I'm not opposed to landing this, but it seems like we should write up the different types of tests we have, so we can clarify what we think the responsibilities of each are.
> 
> At the moment, our 2 "host apps" tests seem to be smoke tests trying to catch packaging / module loading issues in CJS vs. ESM. I could see introducing smoke tests at this layer for each integration, though I wonder if we should reduce the scope of what we're testing for.
> 
> Similar to the current smoke test behavior:
> 
> ```ts
> describe('hostApps — esm-basic', () => {
>   test('import("weave"), run an op, and emit a captured trace', async () => {
>     const projectId = genProjectId();
>     const result = await launchAppFrom({
>       path: fixturePath('esm-basic'),
>       projectId,
>     });
>     if (result.exitCode !== 0) {
>       throw new Error(
>         `esm-basic exited ${result.exitCode}\n` +
>           `stdout:\n${result.stdout}\nstderr:\n${result.stderr}`
>       );
>     }
>     // Confirm the fixture's `myOp` actually executed and emitted a trace
>     // the mock captured under our project_id.
>     const calls = await getCalls(projectId);
>     expect(calls.length).toBeGreaterThanOrEqual(1);
>     expect(
>       calls.some(
>         c => typeof c.op_name === 'string' && c.op_name.includes('myOp')
>       )
>     ).toBe(true);
>   }, 60_000);
> });
> ```
> 
> ... I wonder if we'd want something more like...
> 
> ```ts
> import {fixturePath, genProjectId, getAgentSpans, launchAppFrom} from './utils';
> 
> describe('hostApps — esm-claude-agent-sdk', () => {
>   test('import("weave"), run an `@anthropic-ai/claude-agent-sdk` agent, emit an OTel span', async () => {
>     const projectId = genProjectId();
>     const result = await launchAppFrom({
>       path: fixturePath('esm-claude-agent-sdk'),
>       projectId,
>     });
>     if (result.exitCode !== 0) {
>       throw new Error(
>         `esm-basic exited ${result.exitCode}\n` +
>           `stdout:\n${result.stdout}\nstderr:\n${result.stderr}`
>       );
>     }
> 
>     // Confirm the fixture emitted some spans.
>     const spans = await getAgentSpans(...);
>     expect(spans.length).toBeGreaterThanOrEqual(1);
>   }, 60_000);
> });
> ```
> 
> ... with one for `cjs` as well.
> 
> Feels like the bulk of the assertions done on the logic / shape of the emitted spans can be tested at a different layer, since we have mocks/recordings at our disposal that make that a bit easier than within one of these fake apps.

Good call, slimmed the test down.

On a CJS variant: I left it out because `@anthropic-ai/claude-agent-sdk` ships ESM-only (no CommonJS export)

drtangible:
> On a CJS variant: I left it out because @anthropic-ai/claude-agent-sdk ships ESM-only (no CommonJS export)

Ok, thanks for checking! I believe the source code in `claudeAgentSdk.ts` references having CJS instrumentation. Should we remove that for simplicity / clarity?

PR #7407, opened 2026-06-25
test(weave_ts): fix flaky genai test by dropping real login() from beforeEach

gtarpenning:
## Summary
- `genai.test.ts` `beforeEach` called real `login()` whose only purpose here is setting `process.env.WANDB_API_KEY`. `login()` first does an unretried `GET /health` against the live backend, so a transient blip fails the whole suite.
- The test traces through an `InMemoryTraceServer`, so no backend is needed. Set the env key directly, matching the sibling genai tests (`common.ts` `setupGenAITestEnvironment`).

Original failure ([CI job](https://github.com/wandb/weave/actions/runs/28198557928/job/83531755452?pr=7401)):
```
Unable to verify connection to the weave trace server with given API Key  (clientApi.ts:48)
```

## Testing
test only change. genai.test.ts passes with `WANDB_API_KEY` unset (no network).

### DIFF
--- a/sdks/node/src/__tests__/genai/genai.test.ts
+++ b/sdks/node/src/__tests__/genai/genai.test.ts
@@ -7,9 +7,8 @@ import type {
   ChatCompletionMessageFunctionToolCall,
 } from 'openai/resources/chat/completions';
 import * as weave from '../../../src';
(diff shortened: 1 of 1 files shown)


drtangible (approved):
🙏

PR #7421, opened 2026-06-26
perf(weave): chunk the calls_complete delete by started_at window

gtarpenning:
## Summary
- Follow-up to #7374. The started_at window fixed the descendant *read*, but a single lightweight `DELETE` over the whole subtree still hit the 60s query ceiling on large `calls_complete` projects, because the delete mutation cost scales with parts-touched, not rows-matched.
- Sort descendants by `started_at` and issue one windowed `DELETE` per chunk (`DELETE_CALLS_COMPLETE_CHUNK_SIZE = 1000`), so each command prunes to a thin partition/primary-key slice and stays under the ceiling.
- Observed in prod: `POST /calls/delete` 502ing with `QUERY_WAS_CANCELLED` (CH code 394) on `calls_complete` residence after ~61s. [DD spans](https://us5.datadoghq.com/apm/traces?query=service%3Aweave-trace%20resource_name%3A%22POST%20%2Fcalls%2Fdelete%22%20%40http.status_code%3A502), [example trace](https://us5.datadoghq.com/apm/trace/13586227533205932886) (read fast, the `_command` DELETE ran 61s then cancelled).

## Testing
unit test for the chunking/windowing + CH cascade test that forces a small chunk size so a subtree spans multiple windowed deletes.


gtarpenning:
Closing - local benchmark shows this is the wrong fix. The delete cost is **per-part (full-part rewrite), not per-granule**: a tight `started_at` window doesn't make a touched part cheaper, and chunking just re-rewrites the same parts. Net result: chunking is **4-5x slower** with no per-command savings.

Setup: local CH 26.6, real `calls_complete` schema (ReplacingMergeTree, monthly partitions, skip indexes, `enable_block_number_column=1`), ~620k wide rows / ~200MB parts. Deleted a 5000-call trace as one windowed delete vs five chunked windowed deletes, for a dense (5s) and a spread (6mo) trace.

Heavyweight `ALTER DELETE` (proxy for prod's 60s full-part rewrite):

```
scenario            cmds  total_ms  max_cmd_ms
dense_single           1      2343        2343
dense_chunked(5x)      5      9078        2624
spread_single          1      2071        2071
spread_chunked(5x)     5     11188        2941
```

Lightweight `DELETE FROM` (the actual code path, on 26.6): everything ~40-60ms, but chunked still 3.5x the single (209 vs 60 dense) with no per-command win.

Takeaways:
- 5000 rows in a 5s window out of a 217MB part still took 2.3s = the whole part got rewritten. Window width does not shrink a touched part's cost.
- chunking re-rewrites the same parts (dense) and serializes parallel part-mutations (spread) -> strictly worse.
- max chunk >= single, so it doesn't even buy "stay under the 60s ceiling."

The real lever is the `started_at` window from #7374 (prunes which parts are touched). Given per-part rewrite cost, the fix for the timeout is to avoid the mutation: soft-delete tombstone (INSERT; table is already ReplacingMergeTree + `deleted_at`) or async the mutation. Pursuing those instead.

Caveat: couldn't reproduce the exact 60s on 26.6 (lightweight delete is ~40ms there); prod is 25.12/26.2, replicated, ON CLUSTER. The `ALTER DELETE` run is the faithful heavy-regime proxy, and both paths point the same direction.

PR #7427, opened 2026-06-26
fix(weave_ts): print Weave data URL on Node init

andrewtruong:
Print `View Weave data at https://<domain>/<entity>/<project>/weave` on Node `init()`, matching the Python SDK. Node previously logged only `Initializing project: <entity>/<project>` with no link.


drtangible (approved):
👍

PR #7428, opened 2026-06-27
feat(weave_ts): carry systemInstructions on Turn (invoke_agent) spans

rgao-coreweave:
## What

TypeScript Node SDK parity for #7348 (Python), which added `system_instructions` to the Session SDK's `invoke_agent` (`Turn`) span.

- `Turn` gains a mutable `systemInstructions: string[]` field, flushed to `gen_ai.system_instructions` on `end()` as a **TextPart array** (`[{type:"text", content:"…"}]`) — the same shape the chat span uses, per semconv.
- Empty → the attribute is omitted.

## Why

The Agents chat projection builds its agent-start lifecycle marker from the `invoke_agent` span's `system_instructions`. Previously only the `chat` (LLM) span could carry `gen_ai.system_instructions` on the TS side, so an agent that records its prompt on the turn had no way to surface it on the agent-start marker. This lets the turn span carry the prompt directly, matching the Python SDK.

## Implementation notes

- The field mirrors how `LLM` carries `inputMessages` (a mutable field flushed at `end()`) and the Python SDK's `Turn.system_instructions` (also a mutable field). Identity/config stays in `TurnInit`; content lives as a flushed field — consistent with the existing `LLM` surface.
- Serialized to the TextPart array per semconv, matching Python's `_serialize_system_instructions`.

## Scope notes

- The TS genai SDK has **no `include_content` / PII-redaction layer**, so those aspects of the Python PR have no counterpart here.
- `SubAgent` (also an `invoke_agent` span) can adopt the same field in a follow-up — matching the Python PR's note.
- The existing `piCodingAgent` integration already emits `gen_ai.system_instructions`, but as a plain string array rather than the TextPart shape. That's a pre-existing divergence, left out of scope for this parity PR.

## Testing

- `pnpm exec jest src/__tests__/genai/turn.test.ts` — 15 pass (13 existing + 2 new: TextPart serialization, empty-omission).
- `eslint`, `prettier --check`, and `tsc --noEmit` clean on the changed files.

Companion to #7348.

### DIFF
--- a/sdks/node/src/__tests__/genai/turn.test.ts
+++ b/sdks/node/src/__tests__/genai/turn.test.ts
@@ -5,6 +5,7 @@ import {
   ATTR_GEN_AI_CONVERSATION_ID,
   ATTR_GEN_AI_OPERATION_NAME,
   ATTR_GEN_AI_REQUEST_MODEL,
+  ATTR_GEN_AI_SYSTEM_INSTRUCTIONS,
 } from '../../genai/semconv';
 import {Turn} from '../../genai/turn';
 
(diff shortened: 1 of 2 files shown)


drtangible:
@rgao-coreweave we can close this one, correct?

sdks/node/src/__tests__/genai/turn.test.ts
```diff
@@ -53,6 +54,53 @@ describe('Turn', () => {
...
+// Mirrors the Python SDK's `system_instructions` support on the invoke_agent
+// (Turn) span (wandb/weave#7348). The TS SDK has no include_content / PII layer,
+// so only the serialization + empty-omission behavior has a counterpart here.
+describe('Turn.systemInstructions', () => {
+  setupGenAITestEnvironment();
+  const getExporter = setupExporterPerTest();
+
+  it('serializes systemInstructions as a TextPart array on gen_ai.system_instructions', () => {
+    const turn = Turn.create({agentName: 'weather-bot'});
+    turn.systemInstructions = ['Be helpful', 'Be concise'];
```
drtangible, line 66: This sort of API requiring setting fields one at a time are a bit less common on the TS side. Would love to encapsulate setting and updating via functions if possible. Mind if I pick up this thread of work with this PR (https://github.com/wandb/weave/pull/7468)? I had accidentally kicked that one off before noticing this PR.

rgao-coreweave: I will close my PR. thanks for working on this!

rgao-coreweave:
Closing in favor of @drtangible's PR stack. Thanks!

PR #7429, opened 2026-06-28
feat(weave): full-fidelity batch turn logging + Turn/SubAgent.record()

rgao-coreweave:
Stacked on #7348.

- `log_session` emits each `Turn` directly (shared `_emit_turn`) — stops dropping `system_instructions` / `agent_id` / `agent_description` / `agent_version`.
- `log_turn` gains `agent_id` / `agent_description` / `agent_version` params.
- `Turn.record(...)` and `SubAgent.record(...)` bulk setters (mirror `LLM.record`).

`log_session` rebuilt each `Turn` through `log_turn`'s scalar kwargs and silently dropped non-param fields; emitting the `Turn` directly fixes the class.

`SubAgent.record(...)` closes the same parity gap for sub-agents that `Turn.record()` closes for turns (raised in review on #7348): the identity fields (`agent_id` / `agent_description` / `agent_version`) were reachable only via individual attribute assignment. Factories stay lean (`name` / `model` / `system_instructions`); `.record()` is the bulk path for the full field set, consistent with `Turn`/`LLM`.

Tests: `pytest tests/session/` — 295 pass.

### DIFF
--- a/weave/session/session.py
+++ b/weave/session/session.py
@@ -891,6 +891,47 @@ class Turn(_SpanBase):
             system_instructions=system_instructions or [],
         )
 
+    def record(
+        self,
+        *,
+        messages: list[Message] | None = None,
+        system_instructions: list[str] | None = None,
+        agent_name: str | None = None,
+        model: str | None = None,
+        agent_id: str | None = None,
+        agent_description: str | None = None,
+        agent_version: str | None = None,
+    ) -> Turn:
+        """Set multiple turn fields in one call.
+
+        Collapses the per-field assignments a manually-instrumented agent
+        otherwise makes on a turn (``system_instructions``, ``agent_id``,
+        ...) into a single keyword call. Only fields explicitly passed
+        (non-``None``) are applied — existing values are preserved. Returns
+        ``self`` for chaining. Mirrors ``LLM.record``.
+
+        Note: on the streaming (``with``) path the turn span is named from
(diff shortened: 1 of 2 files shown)


weave/session/session.py
```diff
@@ -891,6 +891,47 @@ def subagent(
...
+        otherwise makes on a turn (``system_instructions``, ``agent_id``,
+        ...) into a single keyword call. Only fields explicitly passed
+        (non-``None``) are applied — existing values are preserved. Returns
+        ``self`` for chaining. Mirrors ``LLM.record``.
+
+        Note: on the streaming (``with``) path the turn span is named from
+        ``agent_name`` at ``__enter__``, so set ``agent_name`` via
+        ``start_turn`` rather than ``record`` if you need the span name to
+        reflect it; ``record`` still updates the ``gen_ai.agent.name``
+        attribute.
```
drtangible, line 957: Is it worth calling out that `messages` is replaced when passed here? There's at least one other method in here (`Turn.user(...)`) that appends.

drtangible (approved):
👍

weave/session/session.py
rgao-coreweave: Good call, I thought about removing the `messages` in record. but to keep `record()` maximally complete, adding to docs to call out that messages is replaced when passed here.

PR #7434, opened 2026-06-29
perf(weave): engage op_name ngram index in eval-results filter

gtarpenning:
## Summary
- `_or_any_prefix_matches` now emits `multiSearchAny(op_name, [prefixes])` instead of OR-ing `position(op_name, p) > 0`. Identical result set; engages the `idx_op_name` ngrambf index, which `position` cannot use.

## Performance
Local benchmark (CH 25.12, 20M rows, real `ngrambf_v1(8,10000,3,0)` index), the realistic N=3 prefix case:

| metric | `position` OR x3 | `multiSearchAny` x3 |
| --- | --- | --- |
| pure scan (use_skip_indexes=0) | 195 ms | 120 ms (~1.6x) |
| end-to-end | 203 ms | 126 ms (~1.6x) |

Win is one Multi-Volnitsky pass vs N substring scans. Index pruning is data-layout-dependent (0% on uniformly interleaved op_names, ~24% on a hot prod project), a variable bonus rather than the main effect. N=1 is a tie with `position`.

## Testing
SQL-assertion tests updated to the `multiSearchAny` form; functional eval-results tests pass on real ClickHouse with the result set unchanged.

### DIFF
--- a/weave/trace_server/query_builder/eval_results_query_builder.py
+++ b/weave/trace_server/query_builder/eval_results_query_builder.py
@@ -41,8 +41,8 @@ END"""
 
 
 def _or_any_prefix_matches(op_name_expr: str, op_prefix_params: list[str]) -> str:
-    """`position(op_name, p) > 0` OR'd across every prefix param."""
-    return " OR ".join(f"position({op_name_expr}, {p}) > 0" for p in op_prefix_params)
+    """`multiSearchAny(op_name, [prefixes])`: matches if any prefix is a substring (engages idx_op_name ngrambf index)."""
+    return f"multiSearchAny({op_name_expr}, [{', '.join(op_prefix_params)}])"
 
 
 def _sort_filter_uses_inputs(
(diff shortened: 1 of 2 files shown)


nikumar1206 (approved):

weave/trace_server/query_builder/eval_results_query_builder.py
```diff
@@ -41,8 +41,8 @@
 
 
 def _or_any_prefix_matches(op_name_expr: str, op_prefix_params: list[str]) -> str:
-    """`position(op_name, p) > 0` OR'd across every prefix param."""
-    return " OR ".join(f"position({op_name_expr}, {p}) > 0" for p in op_prefix_params)
+    """`multiSearchAny(op_name, [prefixes])`: matches if any prefix is a substring (engages idx_op_name ngrambf index)."""
+    return f"multiSearchAny({op_name_expr}, [{', '.join(op_prefix_params)}])"
```
amwarrier, line 45: Question: Is this supported from a certain CH version? How will this play out for on-prem customers?

gtarpenning: its been around since `2019`

PR #7435, opened 2026-06-29
perf(weave): extract-before-aggregate heavy json filters to avoid calls_merged OOM

gtarpenning:
## Summary
- Heavy sub-path filters/orders on `calls_merged` rendered `JSON_VALUE(any(inputs_dump), path)`, which holds the whole JSON dump per group in GROUP BY state and OOMs on large projects.
- Now render `anyIf(JSON_VALUE(inputs_dump, path), inputs_dump IS NOT NULL)` so only the extracted scalar enters aggregation state. NULL-guarded -> identical results.
- Scope: `CallsMergedDynamicField` sub-paths (value casts). Full-column selects, `exists`, and object-ref CTEs unchanged. Read cost is inherent to `calls_merged`; the read-side win lives in `calls_complete` (tokenbf skip index), tracked separately.

## Example
Filtering on a nested field (`inputs.options.enabled == true`):

```sql
-- before: aggregate the whole dump per group, then extract (one ~KB blob/group in GROUP BY state)
HAVING JSON_VALUE(any(inputs_dump), '$."options"."enabled"') = 'true'

-- after: extract per row, aggregate only the scalar (one tiny value/group)
HAVING anyIf(JSON_VALUE(inputs_dump, '$."options"."enabled"'), inputs_dump IS NOT NULL) = 'true'
```

(real output wraps the extraction in the existing `coalesce(nullIf(...))` + `multiIf` bool cast; only the `any(dump)` -> `anyIf(scalar)` part changes.)

## Performance
Local benchmark, ClickHouse 26.6, `AggregatingMergeTree` mirroring `calls_merged`: 400k calls, ~10 KiB `inputs_dump` each, 25% match. `mem` = peak `query_log.memory_usage`; read bytes and result count identical across shapes.

| scenario | rows/grp | paths | shape | peak mem | read | time |
|---|---|---|---|---|---|---|
| single path | 1 | 1 | old | 7.00 GB | 4.06 GB | 1203 ms |
| single path | 1 | 1 | **new** | **0.32 GB** | 4.06 GB | **215 ms** |
| prod shape | 1 | 10 | old | 7.03 GB | 4.06 GB | 4418 ms |
| prod shape | 1 | 10 | **new** | **0.68 GB** | 4.06 GB | **1323 ms** |
| start+end split | 2 | 10 | old | 6.95 GB | 4.07 GB | 4358 ms |
| start+end split | 2 | 10 | **new** | **1.22 GB** | 4.07 GB | **1467 ms** |
| 2 GB mem cap | 1 | 1 | old | **OOM (code 241)** | - | - |
| 2 GB mem cap | 1 | 1 | **new** | **0.31 GB** | 4.06 GB | 206 ms |

~6-22x lower peak memory and 3-5x faster (the dump no longer gets hashed into GROUP BY state); gain scales with groups x dump size. Read unchanged. Under a fixed memory budget the old shape OOMs where the new completes - the reported prod failure (1.6M-row project: old OOMs at the 12 GB cap, new finishes in ~17s).

Example on live data:
image

**Regression check:** the new shape evaluates `JSON_VALUE` per-row vs once-per-group, but in `calls_merged` only the start row carries inputs (end rows are NULL -> skipped by `anyIf`), so the split-row case above still wins on both axes. `JSON_VALUE` returns only scalar leaves, so aggregation state stays small regardless of which path is filtered. No read or correctness change (counts identical here; prod windowed old-vs-new diff was 0).

## Testing
updates query-builder SQL snapshots + adds a stats-count assertion pinning the new shape.

### DIFF
--- a/weave/trace_server/calls_query_builder/utils.py
+++ b/weave/trace_server/calls_query_builder/utils.py
@@ -89,6 +89,7 @@ def json_dump_field_as_sql(
     extra_path: list[str] | None = None,
     cast: tsi_query.CastTo | None = None,
     use_agg_fn: bool = True,
+    agg_fn: str | None = None,
 ) -> str:
     """Build SQL for JSON field access with optional type conversion.
 
@@ -99,6 +100,13 @@ def json_dump_field_as_sql(
         extra_path: Optional list of JSON path components
         cast: Optional type to cast the result to
         use_agg_fn: Whether to use aggregate functions
+        agg_fn: When set, `root_field_sanitized` is the raw (un-aggregated)
+            column and the aggregate wraps the extracted scalar, not the dump:
+            `{agg_fn}If(JSON_VALUE(col, path), col IS NOT NULL)`. This keeps
+            GROUP BY state at one scalar per group instead of the whole JSON
+            dump, avoiding the memory blow-up of `{agg_fn}(dump)`-then-extract.
+            NULL-guarded so only rows carrying the dump contribute, matching
+            the aggregate-then-extract result.
 
     Returns:
         str: SQL expression for accessing the JSON field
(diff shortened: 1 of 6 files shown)


nikumar1206 (approved):

weave/trace_server/calls_query_builder/utils.py
```diff
@@ -89,6 +89,7 @@ def json_dump_field_as_sql(
     extra_path: list[str] | None = None,
     cast: tsi_query.CastTo | None = None,
     use_agg_fn: bool = True,
```
amwarrier: do we need use_agg_fn and agg_fn?
can we not collapse them into a single param? if agg_fn is passed, you assume that it is to be used.

gtarpenning: technically  `use_agg_fn` is actually unused here, let me try to make this easier to understand

gtarpenning: fixed in [cf6955a](https://github.com/wandb/weave/pull/7435/commits/cf6955a96ca2478f9304a4d95ba6d42c6dc4012d)

PR #7439, opened 2026-06-29
feat(weave): propagate agent identity defaults from Conversation to turns

rgao-coreweave:
Stacked on #7429.

`Conversation` gains `agent_id` / `agent_description` / `agent_version`, mirroring its existing `agent_name` / `model` defaults. Each turn inherits them — via `start_turn(...)` on the streaming path and `log_conversation(...)` on the batch path — and a turn's own non-empty value wins per-field (the conversation default fills the rest). Closes the asymmetry that forced single-agent conversations to repeat the identity fields on every turn.

`start_turn` now also accepts `agent_id` / `agent_description` / `agent_version` kwargs, matching `Turn.record` / `log_turn` / `log_conversation` — previously it was the only turn entry point that couldn't set them at creation, forcing a follow-up `turn.record(...)`. Override a turn's identity per-field via either `start_turn(agent_id=...)` or `turn.record(...)`.

`system_instructions` is deliberately excluded: it's per-turn content (gated by `include_content`), so it stays a per-turn / per-LLM field.

Tests: `pytest tests/conversation/` — 306 pass.


weave/conversation/conversation.py
```diff
@@ -1075,6 +1075,12 @@ class Conversation(BaseModel):
     conversation_name: str = ""
     agent_name: str = ""
     model: str = ""
+    # Agent-identity defaults; each turn inherits these (its own value wins).
+    # system_instructions is excluded — gated/redacted content that varies
+    # per turn, so it stays a per-turn / per-LLM field.
+    agent_id: str = ""
+    agent_description: str = ""
+    agent_version: str = ""
```
drtangible: Nit: I wonder if inventing another term ("agent identity fields") is worth it. For example, makes me wonder why `agent_name` isn't part of those set of fields.

drtangible (approved):
👍 👍 👍

weave/conversation/conversation.py
rgao-coreweave: Good point! agent_name is already added, we need to make it symmetric.

drtangible (approved):
👍 👍 👍

PR #7441, opened 2026-06-29
perf(weave): scope table_rows_stats storage rollup to requested digests

gtarpenning:
## Summary
- `make_table_stats_query_with_storage_size` LEFT JOINed the whole project's `table_rows_stats`. This lifts `tables` into a `requested_tables` CTE and scopes the stats subquery to `digest IN (those tables' row_digests)` (`digest` = 2nd sort-key col, prunes).

## Performance
Full generated query on prod-ro CH. The isolated-subquery numbers in the first draft overstated this: the win is real only for SMALL tables. The CTE is referenced twice (scanned twice), so when requested tables span most of the project the `digest IN` set is non-selective, nothing prunes, and it regresses:

| case | master | this PR |
|---|---|---|
| small table (10 rows, 1.24M-stat project) | 1.24M rows / 36ms | 141K rows / 26ms (win) |
| giant table (8.84M rows ≈ whole project) | 8.85M rows / 3.4s | 8.90M rows / 10.1s (3x REGRESSION) |

## Production profile (Datadog, 7d)
- 274 table-stats reads (~39/day), avg 74.6K rows. Max read across the full week = 205K rows (0 over 500K): every observed request is the small-table win case; the regression case did not occur in real traffic.
- Latent risk: a large-dataset project viewing its table stats hits the 3x regression, and the builder has no table-size signal to gate on. Recommend a no-regression guard (or accept this as a small-table-only win) before merge.

## Testing
unit test asserting the full generated SQL string.

### DIFF
--- a/tests/trace_server/query_builder/test_table_query_builder.py
+++ b/tests/trace_server/query_builder/test_table_query_builder.py
@@ -4,6 +4,7 @@ from weave.trace_server.orm import ParamBuilder
 from weave.trace_server.query_builder.table_query_builder import (
     make_table_row_digests_query,
     make_table_stats_basic_query,
+    make_table_stats_query_with_storage_size,
 )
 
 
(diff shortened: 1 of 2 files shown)


zbirenbaum (approved):
nice perf boost, love it. I am a bit worried that some things might be dropped here since it is a lot more specific than the old query but the logic seems to check out.

gtarpenning:
tabling this for now, could regress in edge cases

PR #7442, opened 2026-06-29
perf(weave): scope object_versions_stats storage rollup to matched object_ids

gtarpenning:
## Summary
- The `include_storage_size` LEFT JOIN aggregated the whole project's `object_versions_stats` even for a single-object read. Master already scopes the `object_versions` scan by object_id; this closes the same gap on the stats subquery.
- Scopes the inner stats subquery by `object_id` (3rd sort-key col, prunes; `digest`, the 4th, would not) when an object_id predicate is present. Join key stays `digest`; results unchanged. List-all reads stay unscoped.

## Performance
Full generated query on prod-ro CH (the isolated-subquery numbers in the first draft overstated this):

| case | master | this PR |
|---|---|---|
| selective object (1 version), 539K-stat project | 565,248 rows / 106ms | 32,768 rows / 32ms |
| pathological object (2.06M versions) | 6.49M rows / 2.1s | 6.22M rows / 2.1s (wash) |

Never regresses: it only adds a predicate to an existing scan, no second pass. Worst case (object with many versions, or tiny stats table) is a wash.

## Production profile (Datadog, 7d)
- 2,421 object-storage reads (~346/day); 100% carry an object_id predicate, so all are scopeable.
- They land almost entirely on projects with a tiny `object_versions_stats` table (top read-volume projects <2K rows); the large-stats projects where this wins did ~0 such reads in 7d. So it's safe and future-proof but low realized yield today.

## Testing
unit tests asserting full SQL for object_id / object_ids scoping and the unscoped list-all path.

### DIFF
--- a/tests/trace_server/query_builder/test_objects_query_builder.py
+++ b/tests/trace_server/query_builder/test_objects_query_builder.py
@@ -317,6 +317,154 @@ ORDER BY created_at ASC""",
     }
 
 
+def _expected_storage_metadata_query(
+    where_clause: str, stats_scope: str, outer_clauses: str = ""
+) -> str:
+    """Build expected SQL for the include_storage_size metadata query.
+
+    Args:
+        where_clause: The WHERE condition for latest_row_per_digest.
+        stats_scope: Extra predicate appended to the object_versions_stats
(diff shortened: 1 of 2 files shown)


nikumar1206 (approved):
lgtm

PR #7443, opened 2026-06-29
fix(weave): tolerate stored external refs on agent reads instead of 500

gtarpenning:
## Summary
- A stored `weave:///` ref where an internal ref belongs makes the egress converter raise `InvalidInternalRef("Encountered unexpected ref format.")`. That class is unregistered in `errors.py`, so it hits the 500 default and every read of the row 500s forever ([13 such 500s on `POST /agents/spans/query` in the last 30d](https://us5.datadoghq.com/apm/traces?end=1782763403789&historicalData=true&paused=true&query=%40error.message%3A%2Aunexpected%5C+ref%5C+format%2A+OR+%40error.type%3AInvalidInternalRef&start=1780171403789)).
- Cause is an ingest/read asymmetry specific to agents: agent ingest only converts the typed `weave.{content,artifact,object}_refs` array keys, so a `weave:///` ref in message content / `raw_span_dump` is stored raw, then the egress walker (which validates every string) chokes on it.
- Add `tolerate_external_refs` to the int->ext converter. The 9 agent read methods pass it (log + pass through, since the string is already external-shaped); all other read paths still raise loudly to surface genuine corruption.

## Testing
unit test covering int->ext conversion, private-ref fallback, strict-default raise, and the tolerant agent pass-through.

### DIFF
--- a/tests/trace_server/test_trace_server_converter.py
+++ b/tests/trace_server/test_trace_server_converter.py
@@ -1,5 +1,6 @@
 import datetime
 import json
+import logging
 
 import pytest
 from pydantic import BaseModel, ConfigDict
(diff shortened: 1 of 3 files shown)


bcsherma (approved):
🔥 tytyty

PR #7444, opened 2026-06-29
perf(weave): scope agent trace-attribution fallback to filtered traces

gtarpenning:
## Summary
- The agent trace-attribution `tf` fallback rolls up every trace in the project even when the consuming query reads one trace. Scope it to `trace_id IN (...)` when the query restricts to concrete trace_ids.
- Speculative + aggressively gated: scope applies ONLY when the query references no attributed identity column (agent_name/agent_version/agent_id/conversation_id) AND carries literal trace_id predicates. Identity-filtered and plain-list reads are byte-for-byte unchanged.
- Narrow-but-correct: extracts literal `trace_id = X` / `trace_id IN [...]` from a top-level AND-chain rather than re-compiling an arbitrary filter against raw spans.
- Review scrutiny: the attribution-correctness gating in `extract_scopable_trace_ids` (does any path let an identity predicate or a non-literal reach the raw-spans scope scan?). `tf` still GROUPs BY trace_id over all spans of the scoped traces, so per-trace attribution is lossless.

## Performance
53M -> 798K rows, 187ms -> 50ms on a single-trace read (prod read-only CH).

## Testing
dense unit tests for the scoped single-trace SQL and the unchanged identity-filtered SQL; all 89 existing attribution tests unchanged.

### DIFF
--- a/tests/trace_server/query_builder/test_agent_query_builder.py
+++ b/tests/trace_server/query_builder/test_agent_query_builder.py
@@ -59,6 +59,7 @@ from weave.trace_server.query_builder.agent_query_builder import (
 )
 from weave.trace_server.query_builder.agent_trace_attribution import (
     attributed_spans_source,
+    trace_id_scope_subquery,
 )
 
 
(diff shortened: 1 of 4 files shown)


gtarpenning:
Closing: prod validation on the full generated query shows this regresses (953K rows/99ms scoped vs 317K rows/58ms unscoped). ClickHouse already pushes the outer trace_id predicate into the tf fallback via the join + GROUP BY key, so the explicit scope just adds a redundant spans scan. Isolated-subquery benchmark was misleading.

PR #7450, opened 2026-06-29
fix(weave): scope conversation chat view to its own conversation's spans

gtarpenning:
## Summary
- The multi-turn conversation chat view joins spans on the page's trace_ids but filtered the outer scan only by `project_id`. `trace_id` is reused across conversations in agent-eval workloads, so the view bled foreign conversations' spans and scanned the whole project (129-turn whale conversation: 805 spans returned, only 357 were the conversation's; 348 foreign across 114 other conversations).
- Scope the outer scan to the page's trace_ids (`idx_trace_id` prune) and keep only this conversation's spans plus untagged children (`conversation_id IN (cid, '')`), matching the inner page subquery + the turn-count query.

## Performance
prod whale (5.3M spans), warm interleaved, scoped vs master:

| shape | read_rows | p50 |
|---|---|---|
| point | 5.47M -> 0.35M (15.8x) | 423 -> 59ms |
| list50 | 5.47M -> 2.40M (2.3x) | 446 -> 139ms |
| deep_offset | 5.47M -> 2.37M (2.3x) | 433 -> 139ms |

2nd whale confirms 2.2-21x fewer rows, no regression on any shape.

## Testing
full-SQL unit test + functional CH test proving foreign-conversation spans no longer bleed while untagged children stay included.

### DIFF
--- a/weave/trace_server/query_builder/agent_query_builder.py
+++ b/weave/trace_server/query_builder/agent_query_builder.py
@@ -1741,10 +1741,12 @@ def make_conversation_chat_spans_query(
     # Always cost-augmented so the multi-turn chat view can render per-message
     # and per-turn cost; bounded to one conversation's turns, so cheap.
     source = cost_augmented_source_sql(pb, req.project_id)
+    # trace_id is reused across conversations in agent-eval workloads, so the
+    # page's trace_ids alone would pull in foreign conversations' spans. Scope
+    # the outer scan to the page's trace_ids (idx_trace_id prune) and keep only
+    # this conversation's spans plus untagged children (conversation_id = '').
     return f"""
-        SELECT {QUALIFIED_CHAT_VIEW_COLS}, {QUALIFIED_SPANS_COST_COLS}
-        FROM {source} s
-        INNER JOIN (
+        WITH turn_page AS (
             SELECT trace_id, min(started_at) AS turn_started_at
             FROM spans
             WHERE {_project_filter_sql("project_id", pid)}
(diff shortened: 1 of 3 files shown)


weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -1741,19 +1741,26 @@ def make_conversation_chat_spans_query(
...
             ORDER BY turn_started_at DESC, trace_id DESC
             LIMIT {limit_slot} OFFSET {offset_slot}
-        ) t ON s.trace_id = t.trace_id
+        )
+        SELECT {QUALIFIED_CHAT_VIEW_COLS}, {QUALIFIED_SPANS_COST_COLS}
+        FROM {source} s
+        INNER JOIN turn_page t ON s.trace_id = t.trace_id
         WHERE {_project_filter_sql("s.project_id", pid)}
+          AND s.trace_id IN (SELECT trace_id FROM turn_page)
+          AND s.conversation_id IN ({cid}, '')
```
gtarpenning, line 1762: 🤖 red flag (non-blocking, flagging for team discussion): this `''` clause re-opens the same bleed for *untagged children*.

Producers tag `conversation_id` only on the root/invoke span (see `test_conversation_chat_includes_child_spans_without_conversation_id`), so the children that carry the assistant text + tool bodies are `''`. On a shared `trace_id` that also hosts conv_b's turn, `conversation_id IN (cid, '')` pulls conv_b's untagged children into conv_a's view — they're un-attributable without a parent-chain walk. So this is the exact bug class the PR fixes, shrunk from "foreign tagged spans" to "foreign untagged children."

Big improvement over master and I don't think it blocks. Two paths:
1. principled: propagate `conversation_id` down to child spans at ingest, then this `''` disappears and the leak closes completely.
2. pragmatic: keep it, but add a comment that we knowingly accept untagged-child misattribution on shared traces.

Followup test pushed to this branch exercises the shape and currently asserts the leak (so we see it move when we fix it). Thoughts on which path?

bcsherma (approved):
🤤

weave/trace_server/query_builder/agent_query_builder.py
gtarpenning: @bcsherma just to confirm, you saw this callout? i'm not 100% sure i understand the agent hierarchy

PR #7449, opened 2026-06-29
perf(weave): page-prefetch two-pass for agents list-view attribution

gtarpenning:
## Summary
- The ungrouped agents spans list attributed the whole project (base scan + `tf` rollup) before its `LIMIT` (~106M rows to return 50). When the sort/filter reference no attributed-identity column and no `trace_id`, limit the page first, then attribute only that page with the fallback rollup scoped to the page's traces.
- Identity-sorted / identity-filtered / `trace_id`-filtered / grouped reads keep the existing path (`_spans_list_two_pass_applies`). Lossless: the rollup is consumed only via the `trace_id` join, so scoping drops only rows the join discards.
- Ordering: a `span_id` tiebreaker is appended to every ungrouped list read so the two-pass page is deterministic across the CTE's two references. This also lands on the unchanged full-attribution path, making its pagination deterministic where `started_at` ties were previously arbitrary (benign ordering change).

## Performance
Interleaved warm runs, two-pass vs master, 3 whale spans projects (53M / 45M / 5M). Every non-`include_costs` shape reads <= master rows; no regression.

`include_costs` reads the whole project either way (the per-span price JOIN sits below the page LIMIT, so the page can't prune it): two-pass reads ~the same rows (~1.0x) and is ~1.2-1.3x faster on prod (45M: 1.59s vs 2.06s; 5M: 0.34s vs 0.42s) from the trace-scoped attribution rollup. No regression. Stacked #7453 moves the cost JOIN above the page so costs prune like the rest (28x fewer rows on the 45M project).

**Prod reach (last 7d, `system.query_log`):** 188 of 291 agents spans-list queries (65%) take this path; the rest fall back (101 agent-identity filters, 2 `trace_id`, 0 identity sorts). Those 188 currently read **~589M rows** total at p50 261ms / p95 580ms, which the two-pass cuts 2-51x depending on shape.

| shape | rows read ↓ | p50 ↓ |
|---|---|---|
| list50 (default) | 8-42x | 4-18x |
| list1000 | 2-4x | 4-12x |
| deep offset (10k) | 5-22x | 4-18x |
| filter op_name | 5-51x | 3-7x |
| time window | 1.6-14x | ~1-1.6x |
| controls (trace_id, agent sort, agent filter) | 1x (byte-identical fallback) | 1x |

[...]

PR #7453, opened 2026-06-29
perf(weave): prune the agents cost list-read via the two-pass

gtarpenning:
## Summary
- Stacked on #7449. There, the `include_costs` list read still scanned the whole project: the per-span price JOIN sat *below* the page LIMIT, and a LIMIT can't prune below a JOIN. This builds the page on bare `spans` (prunes), then joins prices to just that page, then attributes it, so costs prune like every other shape. Cost-column sorts gate out of the two-pass (the page would then depend on the JOIN).

## Performance
Prod, two-pass vs master, whale projects (read_rows / p50, interleaved warm):

| project | rows read ↓ | p50 ↓ |
|---|---|---|
| 45M | 28x (93.7M -> 3.2M) | 16x (2.3s -> 0.14s) |
| 5M | 8x (11.9M -> 1.3M) | 3.4x (0.43s -> 0.13s) |

## Testing
full-SQL unit tests (cost JOIN over page + cost-sort fallback) + functional CH test: two-pass per-span costs == full attribution, cost is span-exact (child's own tokens, not the inherited agent's), unpriced model stays null.


bcsherma (approved):
<3 <3 <3

PR #7457, opened 2026-06-29
fix(weave): surface system instructions from chat spans in agent chat view

bcsherma:
## Problem

System instructions are logged on spans but don't appear in the Agents-tab conversation (chat) view for some traces.

The chat-view projection (`chat_view.py`) only read `system_instructions` off `invoke_agent` spans, where they're attached to the `agent_start` card the UI renders. But `system_instructions` is extracted onto **every** GenAI span (`genai_extraction.py`), so traces instrumented as plain LLM/chat spans had their system prompt silently dropped from the chat view even though it was present on the spans. This is the common shape in prod — e.g. the **WB Agent**: the `invoke_agent` span carries the agent name but *no* prompt, and the prompt is recorded on each child `chat` span. The frontend renderer was never the issue.

## Fix

Surface system instructions from content (LLM/chat) spans too, reusing the same `agent_start` card the UI already renders (`AgentChatView` reads `agent_start.system_instructions`).

- **Fold, don't duplicate:** when the enclosing `invoke_agent` already emitted a bare `agent_start` for the turn (agent identity, no prompt), the chat span's instructions fold into that card instead of emitting a redundant second one.
- **Dedup:** tracked against the last-surfaced instructions (set by both the `invoke_agent` and content-span paths) so a prompt replayed on every turn's span shows once, not once per call; a genuinely changed prompt emits a new card.

## Verification against a real prod conversation

Pulled the raw spans for a 23-turn `wb-agent-team/wandb-agent-prod` conversation and ran them through the projection:

- Instructions are stored on the `chat` spans via the `gen_ai.system_instructions` attribute; `invoke_agent` spans carry empty instructions.
- **Before:** 0 system-instruction cards (dropped).
- **After:** exactly 1 per turn, folded into the turn's existing `agent_start` — no duplicate cards, no double section headers. Each turn renders `[user_message, agent_start(+system), ...]`.

## Tests

Added coverage in `test_genai_chat_view.py`:
- content-span system instructions surface as an `agent_start` card
- identical instructions across turns are deduped to a single card
- a changed prompt mid-conversation emits a new card
- an `invoke_agent`'s instructions suppress a descendant chat span's duplicate
- a bare `invoke_agent` agent_start absorbs a descendant chat span's instructions (the prod WB Agent shape)

All 32 tests in the file pass; `nox -e lint` passes.

## Not covered (follow-up)

System instructions logged as a `system`-role entry *inside* `input_messages` (rather than via the `gen_ai.system_instructions` attribute) are still filtered out by `_NON_USER_PROMPT_ROLES`. The verified prod conversation does not use that shape, so it's out of scope here.

### DIFF
--- a/weave/trace_server/agents/chat_view.py
+++ b/weave/trace_server/agents/chat_view.py
@@ -260,6 +260,17 @@ class ChatTraversal:
     # (text, media-digests) of the most recently emitted user message, so the
     # same message replayed by a following LLM span is not emitted twice.
     _last_user_sig: tuple[str, tuple[str, ...]] | None = None
+    # The system-instructions text most recently surfaced (from either an
+    # invoke_agent or a content span), so the same instructions replayed on
+    # every turn's LLM span render once, not once per call. None until the
+    # first instructions are emitted.
+    _last_system_instructions: str | None = None
+    # A just-emitted agent_start that carries an agent identity but no system
+    # prompt (some SDKs record the prompt on the child LLM span, not the
+    # invoke_agent span). A descendant content span's instructions fold into it
+    # rather than emitting a redundant second card. Reset per invoke_agent and
+    # cleared at the turn boundary (user message).
+    _pending_agent_start: AgentChatMessage | None = None
 
     def walk_roots(self, roots: list[SpanNode]) -> None:
         for root in roots:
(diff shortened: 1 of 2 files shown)


bcsherma:
* **#7457** <a href="https://app.graphite.com/github/pr/wandb/weave/7457?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7457?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/agents/chat_view.py
```diff
@@ -260,6 +260,17 @@ class ChatTraversal:
     # (text, media-digests) of the most recently emitted user message, so the
     # same message replayed by a following LLM span is not emitted twice.
     _last_user_sig: tuple[str, tuple[str, ...]] | None = None
+    # The system-instructions text most recently surfaced (from either an
+    # invoke_agent or a content span), so the same instructions replayed on
+    # every turn's LLM span render once, not once per call. None until the
+    # first instructions are emitted.
+    _last_system_instructions: str | None = None
```
rgao-coreweave, line 267: LOVE THIS!

rgao-coreweave (approved):
LGTM! 👍👍👍

Have you tested this with WB_Agent? How does it will look like now? (we shouldn't be drowning in system prompts right)

> the same instructions replayed on every turn's LLM span render once, not once per call


https://github.com/wandb/core/pull/46423/changes#diff-91ab3dba41fe8138a995c1a071c086789d6bc3935329033f0b4944a924277bbaR58-R60 PR is probably not the right shape, but we can see that system_instructions is passed in LLM. I wonder if this is something we need to fix in wb agent's instrumentation? like setting system instructions at turns so invoke_agent span has systemInstructions? 

```
    with weave.start_llm(
        model=model, provider_name="openai", system_instructions=sys_instr
    ) as llm:
```

PR #7458, opened 2026-06-29
chore(weave): deprecate add_event / addEvent

rgao-coreweave:
Marks the public span-event methods deprecated in both SDKs; use `set_attributes` / `setAttributes` instead.

- **Python** — `_SpanBase.add_event` (inherited by `Tool` / `LLM` / `SubAgent` / `Turn`) via `typing_extensions.deprecated`; emits a `DeprecationWarning` and is flagged by type checkers / IDEs.
- **TypeScript** — `SpanBase.addEvent` via JSDoc `@deprecated` (static/IDE only; no runtime warning, matching existing TS deprecations).

**Why:** OpenTelemetry is [phasing out the Span Event API](https://opentelemetry.io/blog/2026/deprecating-span-events/) (`Span.addEvent`). We steer callers to span attributes, which Weave already records and surfaces in the Agents tab.

Both methods still work and existing span-event data stays valid. Opened as **draft** because upstream OTel SDKs haven't removed `Span.addEvent` yet — this is forward-looking.

**Testing:** 28 Python conversation tests pass (incl. a new `DeprecationWarning` assertion); 39 TS genai tests pass; ruff + prettier clean.

### DIFF
--- a/sdks/node/src/genai/spanBase.ts
+++ b/sdks/node/src/genai/spanBase.ts
@@ -72,8 +72,8 @@ export abstract class SpanBase {
    * no-ops after `end()`. Mirrors OTel `Span.addEvent`.
    *
    * @deprecated Record this data via {@link setAttributes} instead.
-   * OpenTelemetry is phasing out the Span Event API (`Span.addEvent`; OTEP
-   * 4430). This method still works and existing span-event data stays valid.
+   * OpenTelemetry is phasing out the Span Event API (`Span.addEvent`). This
+   * method still works and existing span-event data stays valid.
    * See https://opentelemetry.io/blog/2026/deprecating-span-events/
    *
    * @example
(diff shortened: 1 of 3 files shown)


sdks/node/src/genai/spanBase.ts
```diff
@@ -71,6 +71,10 @@ export abstract class SpanBase {
    * context compaction, tool-loop detection, or guardrail trips. Warns and
    * no-ops after `end()`. Mirrors OTel `Span.addEvent`.
    *
+   * @deprecated Record this data via {@link setAttributes} instead.
+   * OpenTelemetry is phasing out the Span Event API (`Span.addEvent`; OTEP
+   * 4430). This method still works and existing span-event data stays valid.
```
drtangible: Worth linking to the post? I'm not sure `OTEP 4430` would mean anything to a reader.

drtangible (approved):
👍 👍 👍

sdks/node/src/genai/spanBase.ts
rgao-coreweave: Good call! fixed.

PR #7469, opened 2026-06-30
feat(weave): add agent eval span fields

bcsherma:
## Description

First PR in the agent evals bridge stack.

This adds first-class eval metadata to Weave agent spans so spans emitted during an evaluation can be queried as agent data by eval run, prediction/trial, row, and example. The goal is to keep evals as the source of truth for runs/scores while letting the Agents product find the corresponding OTel spans without depending on op-tracing call trees or scanning arbitrary custom attrs.

What this PR changes:

- Adds `weave.eval.*` semantic convention fields:
  - `weave.eval.run_id`
  - `weave.eval.predict_and_score_call_id`
  - `weave.eval.kind`
  - `weave.eval.row_digest`
  - `weave.eval.example_id`
  - `weave.eval.trial_index`
  - `weave.eval.evaluation_name`
- Promotes those fields into the agent span schema and OTel extraction path.
- Adds ClickHouse columns and bloom filter indexes for the eval fields.
- Allows agent span queries to filter/group/sort on the promoted eval fields.
- Updates agent query builder/compiler tests and OTel extraction tests for the new fields.

What this is for:

- Enables an "agent eval run" to be represented as normal eval calls plus normal OTel agent spans.
- Lets the Agents UI query eval runs from the agent span table by `eval_run_id`, `eval_kind`, row digest, example id, and trial index.
- Keeps the agent tracing system independent of the legacy op-tracing call-tree model.
- Makes `genai_span_ref` optional compatibility data rather than the primary lookup path for new agent eval data.

Reviewer notes:

- This does not backfill old spans. Existing rows get default empty values.
- This is the base/backend-storage PR for the stack. The follow-up SDK PR stamps these attrs during imperative eval logging.
- Migration `036_add_agent_eval_span_columns` is the main schema change to review.

## Testing

Added/updated coverage in:

- `tests/trace_server/test_genai_extraction.py`
- `tests/trace_server/test_genai_agent_queries.py`
- `tests/trace_server/query_builder/test_agent_query_builder.py`
- `tests/trace_server/query_builder/test_agent_query_compiler.py`

Full-stack smoke after applying the stack:

- Ran the local trace-server migration; migration `036_add_agent_eval_span_columns` applied.
- Seeded an imperative agent eval run through the SDK bridge.
- Verified `agent_spans_query` grouped the run by `eval_run_id` / `eval_kind` and returned 42 spans, 3 eval rows, and 6 trials.


bcsherma:
* **#7476** <a href="https://app.graphite.com/github/pr/wandb/weave/7476?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7469** <a href="https://app.graphite.com/github/pr/wandb/weave/7469?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7469?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/migrations/037_add_agent_eval_span_columns.down.sql
jtschoonhoven, line 1: I believe another PR merged and claimed migration 036 so you'll need to rename these files

weave/trace_server/migrations/030_add_agent_tables.up.sql
```diff
@@ -26,6 +26,14 @@ CREATE TABLE IF NOT EXISTS spans (
     agent_description   String DEFAULT '',
     agent_version       String DEFAULT '',
 
+    eval_run_id                       String DEFAULT '',
+    eval_predict_and_score_call_id    String DEFAULT '',
+    eval_kind                         String DEFAULT '',
+    eval_row_digest                   String DEFAULT '',
+    eval_example_id                   String DEFAULT '',
+    eval_trial_index                  Int64 DEFAULT -1,
+    eval_evaluation_name              String DEFAULT '',
```
jtschoonhoven: What's the point of adding this here AND in the later migration? I guess it's fine, it just seems unnecessary.

weave/conversation/conversation.py
```diff
@@ -243,6 +254,7 @@ def set_attributes(self, attributes: dict[str, Any]) -> Self:
             span.set_attributes(attributes)
         return self
 
+    @deprecated(_ADD_EVENT_DEPRECATION_MESSAGE)
```
jtschoonhoven, line 280: This seems unrelated to the rest of the PR? Could this be its own PR?

weave/conversation/conversation.py
```diff
@@ -251,6 +263,12 @@ def add_event(
     ) -> Self:
         """Record an OTel span event at a point in time within this span.
 
+        .. deprecated::
+            Record this data with ``set_attributes`` instead. OpenTelemetry is
+            phasing out the Span Event API (``Span.add_event``).
+            ``add_event`` still works and existing span-event data stays valid.
+            See https://opentelemetry.io/blog/2026/deprecating-span-events/.
```
jtschoonhoven, line 293: Nit: Do we actually build docs from this (sphinx?) markup? `.. deprecated::` looks weird and let's not do that if we don't need it. Ditto the \``double ticks``

weave/trace_server/migrations/036_add_agent_eval_span_columns.up.sql
jtschoonhoven, line 1: It's safer to ship migrations separately from the features they depend on, can we pull this out into its own thing?

weave/trace_server/migrations/036_add_agent_eval_span_columns.up.sql
```diff
@@ -0,0 +1,17 @@
...
+    ADD COLUMN IF NOT EXISTS eval_evaluation_name           String DEFAULT '';
+
+ALTER TABLE spans
+    ADD INDEX IF NOT EXISTS idx_eval_run_id eval_run_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_eval_predict_and_score_call_id eval_predict_and_score_call_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_eval_row_digest eval_row_digest TYPE bloom_filter(0.01) GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_eval_example_id eval_example_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_eval_trial_index eval_trial_index TYPE set(128) GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_eval_kind eval_kind TYPE set(16) GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_eval_evaluation_name eval_evaluation_name TYPE ngrambf_v1(8, 10000, 3, 0) GRANULARITY 1;
```
jtschoonhoven: Do we need _all_ these skip indices? E.g. are we actually filtering on `row_digest` and `example_id`?

jtschoonhoven (commented):
Can we ship the migration and deprecation into their own PRs? IMO migrations should always ship independently of the code that depends on them.

weave/trace_server/migrations/030_add_agent_tables.up.sql
```diff
@@ -26,6 +26,14 @@ CREATE TABLE IF NOT EXISTS spans (
     agent_description   String DEFAULT '',
     agent_version       String DEFAULT '',
 
+    eval_run_id                       String DEFAULT '',
```
gtarpenning: `run` feels like nearly but not quite the right noun

weave/trace_server/migrations/030_add_agent_tables.up.sql
gtarpenning: why are we adding this? because of ordering or something?

tests/trace_server/test_genai_agent_queries.py
```diff
@@ -127,6 +127,94 @@ def test_spans_insert_and_query(ch_server):
     assert all(s.agent_name == "agent-A" for s in res_filtered.spans)
 
 
+def test_eval_span_fields_are_queryable(ch_server):
```
gtarpenning, line 139: can we assert on an unset `eval_trial_index` that its not -1?

tests/trace_server/test_genai_extraction.py
```diff
@@ -175,12 +190,24 @@ def test_extract_genai_span_comprehensive() -> None:
...
 
 
+def test_extract_genai_span_defaults_missing_eval_trial_index() -> None:
+    result = extract_genai_span(
+        _make_span(attrs={"weave.eval.run_id": "eval-run-001"}),
+        project_id="p1",
+    )
+
+    assert result.eval_run_id == "eval-run-001"
+    assert result.eval_trial_index == -1
```
gtarpenning, line 217: oh... hmm i feel liek we probably want this to be `None` to a user at least, this is an internal state?

tests/conversation/test_span_attributes.py
```diff
@@ -118,6 +118,24 @@ def test_add_event_no_op_after_end(
     assert len(finished_span.events) == 0
 
 
+def test_add_event_emits_deprecation_warning(
```
gtarpenning, line 121: break this out into a separate pr? super easy to land that

tests/trace_server/test_genai_agent_queries.py
gtarpenning, line 139: this is our only functional test? probably want more than this right?

weave/trace_server/migrations/030_add_agent_tables.up.sql
```diff
@@ -26,6 +26,14 @@ CREATE TABLE IF NOT EXISTS spans (
...
+    eval_predict_and_score_call_id    String DEFAULT '',
+    eval_kind                         String DEFAULT '',
```
gtarpenning: I think this should at least be a LOW_CARDINALITY type

weave/trace_server/migrations/030_add_agent_tables.up.sql
```diff
@@ -110,6 +118,13 @@ CREATE TABLE IF NOT EXISTS spans (
     INDEX idx_response_model response_model TYPE ngrambf_v1(8, 10000, 3, 0) GRANULARITY 1,
     INDEX idx_agent_name agent_name TYPE ngrambf_v1(8, 10000, 3, 0) GRANULARITY 1,
     INDEX idx_agent_version agent_version TYPE ngrambf_v1(8, 10000, 3, 0) GRANULARITY 1,
+    INDEX idx_eval_run_id eval_run_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_eval_predict_and_score_call_id eval_predict_and_score_call_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_eval_row_digest eval_row_digest TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_eval_example_id eval_example_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_eval_trial_index eval_trial_index TYPE set(128) GRANULARITY 1,
+    INDEX idx_eval_kind eval_kind TYPE set(16) GRANULARITY 1,
```
gtarpenning: not sure this is worth it, in projects with evals id expect at least one in 8000 rows to be part of an eval, meaning this skip index won't really be used (except it costs insert _and_ read latency)

weave/trace_server/migrations/030_add_agent_tables.up.sql
```diff
@@ -110,6 +118,13 @@ CREATE TABLE IF NOT EXISTS spans (
     INDEX idx_response_model response_model TYPE ngrambf_v1(8, 10000, 3, 0) GRANULARITY 1,
     INDEX idx_agent_name agent_name TYPE ngrambf_v1(8, 10000, 3, 0) GRANULARITY 1,
     INDEX idx_agent_version agent_version TYPE ngrambf_v1(8, 10000, 3, 0) GRANULARITY 1,
+    INDEX idx_eval_run_id eval_run_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_eval_predict_and_score_call_id eval_predict_and_score_call_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_eval_row_digest eval_row_digest TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_eval_example_id eval_example_id TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_eval_trial_index eval_trial_index TYPE set(128) GRANULARITY 1,
```
gtarpenning: would we ever filter on `eval_trial_index` without `eval_run_id`? i'm pretty sure if you use both you aren't really going to be gaining much, unless evals are like 100k rows and temporally run one at a time. probably can drop

gtarpenning: oh trial_idx not like example idx. hmm. i still think this is not worth it. are we expecting 128 trials?

weave/trace_server/migrations/030_add_agent_tables.up.sql
```diff
@@ -26,6 +26,14 @@ CREATE TABLE IF NOT EXISTS spans (
     agent_description   String DEFAULT '',
     agent_version       String DEFAULT '',
 
+    eval_run_id                       String DEFAULT '',
+    eval_predict_and_score_call_id    String DEFAULT '',
+    eval_kind                         String DEFAULT '',
+    eval_row_digest                   String DEFAULT '',
+    eval_example_id                   String DEFAULT '',
+    eval_trial_index                  Int64 DEFAULT -1,
```
gtarpenning: is this the # of rows in an eval or the number of times an eval has been run? if the latter maybe consider a smaller data type

weave/trace_server/migrations/030_add_agent_tables.up.sql
gtarpenning: or like `Enum8`

tests/conversation/test_conversation_otel.py
andrewtruong, line 983: nit: May as well call this `span` or `conversation` to be more clear.  I would do this throughout

rgao-coreweave (approved):
approve to unblock.

weave/trace_server/migrations/037_add_agent_eval_span_columns.down.sql
bcsherma: Done — renamed to `037`; the stack is now rebased on master, which owns `036`.

weave/trace_server/migrations/030_add_agent_tables.up.sql
bcsherma: Reverted `030` back to master — the eval columns now live only in `037`, whose `ADD COLUMN IF NOT EXISTS` covers fresh installs too.

weave/conversation/conversation.py
bcsherma: Dropped the deprecation change entirely.

weave/conversation/conversation.py
bcsherma: Removed.

weave/trace_server/migrations/030_add_agent_tables.up.sql
bcsherma: Kept `eval_run_id` — it maps to the `weave.eval.run_id` semconv key and matches "eval run" terminology. Happy to rename if you feel strongly.

weave/trace_server/migrations/030_add_agent_tables.up.sql
bcsherma: Reverted — `030` matches master again; the columns live only in migration `037`.

tests/trace_server/test_genai_agent_queries.py
bcsherma: Done — an unset `eval_trial_index` now reads back as `None` (see `test_eval_trial_index_unset_reads_as_none`); `-1` stays internal to the ClickHouse column.

tests/trace_server/test_genai_extraction.py
bcsherma: Agreed — `AgentSpanSchema` now maps the `-1` storage sentinel to `None`, so `-1` never leaves ClickHouse.

tests/conversation/test_span_attributes.py
bcsherma: Dropped the deprecation change, so this test is gone.

tests/trace_server/test_genai_agent_queries.py
bcsherma: Added `test_eval_trial_index_unset_reads_as_none` and a filter-by-kind / predict_and_score_call_id test, on top of the existing filter+sort+group coverage.

neutralino1 (approved):

PR #7475, opened 2026-06-30
fix(weave): restore unix-domain-socket support for dogstatsd db_inserts

gtarpenning:
## Summary
- #7268's inlined DogStatsD client is UDP-only; it dropped the unix-socket support ddtrace had. SaaS prod runs the Agent over a Unix socket (`DD_DOGSTATSD_URL=unix://...`), so `weave_trace_server.db_inserts` silently went to zero on deploy. Single-tenant (`udp://`) was unaffected.
- Resolve the URL scheme to `AF_UNIX` vs `AF_INET` and connect the matching datagram socket.

## Testing
unit tests: UDS/UDP/host-port resolution cases + a real AF_UNIX server receiving the emitted packet.

### DIFF
--- a/tests/trace_server/test_datadog.py
+++ b/tests/trace_server/test_datadog.py
@@ -8,6 +8,9 @@ can't silently drift them.
 
 from __future__ import annotations
 
+import socket
+from pathlib import Path
+
 import pytest
 
 from weave.trace_server import datadog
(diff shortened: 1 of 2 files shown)


amwarrier (approved):

gtarpenning:
Closing as won't-do.

PR #7477, opened 2026-06-30
feat(weave): filter agent conversations by signals (tags + ratings) - backend

jtschoonhoven:
## What

Server-side filtering of the agent conversations view by Signals (tags and ratings). Conversations are `agent_spans_query` results grouped by `conversation_id`; signals live in the separate `feedback` table. A conversation matches if a signal is on the conversation itself or on any turn/span within it (unioned). Designed to hold up at billions of conversations and signals.

## How

To filter cheaply at scale without a cross-table join over billions of rows:

- Migration 036 denormalizes `conversation_id` onto `feedback` (alongside the existing `span_*` columns) and adds a `bloom_filter` skip-index on `scorer_tags`.
- Feedback create resolves and stores `conversation_id` for agent feedback going forward (conversation refs self-resolve; turn/span refs resolve via a `spans` point-lookup).
- The grouped spans query gains a typed `signal_filters` field compiled to a semi-join: `... AND s.conversation_id IN (SELECT conversation_id FROM feedback WHERE <tag/rating predicate>)`. Tags use `hasAny`; ratings use a `mapContains`-guarded `Map` comparison. When no filter is set the query is byte-for-byte unchanged.

The feedback subquery is not time-constrained (a signal's timestamp is not the conversation's) and runs against the project-partitioned, tag-indexed `feedback` table, so it stays selective.

## Testing

Query-builder SQL snapshots (tag-only, rating-only, combined, empty), the write-path resolver unit tests, and an end-to-end ClickHouse test proving union semantics and mutation-correctness: tag a turn and its conversation appears; purge the feedback and it drops out; a conversation-level tag also matches.

## Notes

The frontend lands separately in wandb/core (linked from that PR). Go-forward only: pre-existing feedback rows have an empty `conversation_id` and are invisible to the filter until a backfill. Deferred follow-ups: historical backfill, negative filters (untagged / no-signal), a low-selectivity `IN`-to-`JOIN` guard, and a pre-aggregated per-conversation rollup if the feedback subquery ever becomes a read bottleneck.

https://github.com/user-attachments/assets/20eb0f9d-3b0d-4c52-a933-6606b4ed9683


weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -862,6 +863,56 @@ def _spans_filter_sql(
     return _FilterSQL(where=" AND ".join(where_conditions))
 
 
+_OP_TO_SQL: dict[str, str] = {
```
gtarpenning: seems strange this isn't done elsewhere (and maybe we can put this in a helper? this file is getting up there...

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -862,6 +863,56 @@ def _spans_filter_sql(
...
+        op = _OP_TO_SQL[cond.op]
+        # mapContains guard: Map access returns 0.0 for an absent key, which
+        # would spuriously pass a `>= 0` test.
+        conditions.append(
+            f"(mapContains(scorer_ratings, {key_slot}) "
+            f"AND scorer_ratings[{key_slot}] {op} {val_slot})"
+        )
+
+    conditions.append("conversation_id != ''")
+    where = " AND ".join(conditions)
```
gtarpenning: this AND is intentional right? can we add a functional test for this and see if this is what we expect: 

writes:
-  user puts a conversation-level tag "reviewed" (row 1),
- a monitor rates a turn 0.9 (row 2). 
read: 
- Filter `tags=["reviewed"] AND rating>=0.8`

do we expect any rows back? because the AND sub clause gets evaluated as a group against one feedback row (right?), so i'm pretty sure we will get zero rows, even though the conversation satisfies the filter.

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -1099,13 +1150,19 @@ def make_spans_list_query(pb: ParamBuilder, req: AgentSpansQueryReq) -> str:
     ensure_group_filters_match(group_filters, req.group_by, context="spans list")
     having = group_filters_having_sql(pb, group_filters)
     having_sql = f"HAVING {having}" if having else ""
+    signal_subquery = build_signal_filter_subquery(pb, req.project_id, req.signal_filters)
```
gtarpenning: is there a stats query we should do this for?

weave/trace_server/query_builder/agent_query_builder.py
```diff
@@ -862,6 +863,56 @@ def _spans_filter_sql(
...
+    project-partitioned and tag-indexed, so this scan is selective; it is NOT
+    time-constrained because a signal's timestamp is not the conversation's.
+    """
+    if signal_filters is None or signal_filters.is_empty():
+        return None
+
+    pid_slot = pb.add(project_id, param_type="String")
+    conditions: list[str] = [
+        f"project_id = {pid_slot}",
+        "feedback_type IN ('wandb.agent_user_feedback', 'wandb.agent_monitor')",
```
gtarpenning: maybe use `AGENT_SPAN_FEEDBACK_TYPES`?

weave/trace_server/agents/types.py
```diff
@@ -784,6 +798,7 @@ class AgentSpansQueryReq(BaseModel):
     offset: int = Field(default=0, ge=0)
     started_after: datetime.datetime | None = None  # filter started_at >= start
     started_before: datetime.datetime | None = None  # filter started_at < end
+    signal_filters: AgentSignalFilter | None = None
```
gtarpenning, line 801: does this only work if we set `group_by conversation_id`? should we validate that? or add some other handling? or call it `conversation_signal_filters` or something?

weave/trace_server/feedback.py
```diff
@@ -320,9 +322,54 @@ def format_feedback_to_row(
...
+) -> ResolvedAgentTargets:
+    """Resolve the conversation and turn a feedback row belongs to, for denormalization.
+
+    Producers that hold the scored span (the agent monitor worker, the UI feedback
+    controls) supply both IDs directly, and those win. As a zero-cost convenience for
+    callers that don't, we also derive what the target ref alone reveals: a
+    conversation ref gives the conversation; a turn ref gives the trace. A span ref
+    reveals neither (its id is not a trace/conversation id) and unknown/invalid refs
+    reveal nothing. Missing fields stay '' (invisible to that grain's signal filter
+    until a backfill runs). There is deliberately no spans lookup here — resolution
```
gtarpenning: `There is deliberately no spans lookup here.` that seems false given the def of `conversation_id` in the interface?

weave/trace_server/feedback.py
```diff
@@ -320,9 +322,54 @@ def format_feedback_to_row(
...
+    `trace_id` is '' for conversation-targeted feedback (no single turn) and for
+    anything unresolved; consumers treat '' as "invisible to that grain's
+    signal filter until a backfill runs".
+    """
+
+    conversation_id: str
+    trace_id: str
+
+
+def resolve_feedback_agent_targets(
```
gtarpenning: can we add a unit test on this and confirm its what we want? a bit of a contrived example, but could easily happen I think: 

```python
def test_signal_attribution_honors_conversation_ref_over_supplied_id(ch_server):
    """Feedback on conv-A's own ref must not surface conv-B in signal results."""
    conv_a_ref = ri.InternalAgentConversationRef(
        project_id=project_id, conversation_id=conv_a
    ).uri()
    ch_server.feedback_create(
        tsi.FeedbackCreateReq(
            project_id=project_id,
            weave_ref=conv_a_ref,               # ref says conv-A
            feedback_type=AGENT_USER_FEEDBACK_TYPE,
            payload={"scorer_tags": ["reviewed"]},
            conversation_id=conv_b,             # supplied says conv-B (mismatch)
        )
    )
    # The "reviewed" tag lives on conv-A; filtering for it must return conv-A, never conv-B.
    assert filtered_conv_ids(tags=["reviewed"]) == [conv_a]   # today returns [conv_b]
```

gtarpenning (approved):

weave/trace_server/agents/types.py
jtschoonhoven: Yeah good call. Adding validation and a comment to explain.

weave/trace_server/feedback.py
jtschoonhoven: Yeah, and also I'll add validation server-side in the wandb/core PR

PR #7478, opened 2026-06-30
chore(weave): support minor and major version bumps in the bump tool

andrewtruong:
Adds major and minor version bumping to the util


drtangible (approved):
👍

PR #7484, opened 2026-06-30
perf(weave): opt-in limit+1 has_more for calls_complete pagination

gtarpenning:
## Summary
- Add opt-in `include_exact_count` (default true) to calls stats. When false, calls_complete `has_more` is computed via a `limit+1` probe instead of an exact filtered `count()`, so a heavy-field-filtered page stops scanning after limit+1 matches.
- Default preserves today's exact-count behavior; frontend list views opt in separately.

## Performance
Prod read-only, heavy `inputs_dump` filter, 33,638 matches, limit=100:

| | bytes_read | elapsed |
|---|--:|--:|
| exact count | 7.28 GiB | 1.71s |
| limit+1 probe | 0.02 GiB | 0.03s |

## Testing
adds calls_query_builder snapshot coverage for both the exact and limit+1 stats SQL.

### DIFF
--- a/tests/trace_server/query_builder/test_calls_query_builder.py
+++ b/tests/trace_server/query_builder/test_calls_query_builder.py
@@ -4640,6 +4640,70 @@ def test_stats_query_calls_complete_with_feedback_filter_uses_count_distinct() -
     )
 
 
+def test_stats_query_calls_complete_include_exact_count_toggle() -> None:
+    """include_exact_count picks exact filtered count() vs a limit+1 has_more probe.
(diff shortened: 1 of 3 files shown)


gtarpenning:
Superseded: rather than an opt-in exact-count flag, we're deriving `has_more` from the row fetch itself (frontend requests pageSize+1, has_more = received > pageSize), which removes the expensive count from the calls-list pagination path entirely instead of just cheapening it. Closing in favor of the frontend change.

PR #7486, opened 2026-06-30
perf(weave): two-pass filter CTE for calls_complete heavy reads

gtarpenning:
## Summary
- Enable the two-pass filter CTE for calls_complete heavy-column reads (was calls_merged-only): pass 1 resolves the page's ids + started_at on light columns, pass 2 loads `*_dump` payloads for only those ids, bounded on the page's `[min,max]` started_at so it prunes on the `(project_id, started_at)` PK prefix. Same rows returned.
- The started_at bound keeps pass 2 fast for non-time-ordered ids (e.g. OTel span ids); without it `id IN (page ids)` can't use `idx_id` and rescans the project (one 7.8M-row project read 27 GiB / 847 ms vs 24 MiB / 57 ms).

## Performance
Prod read-only, full-row list, `ORDER BY started_at DESC LIMIT 100`, interleaved warm, master -> branch:

| project (rows, avg row) | p50 ↓ | bytes read ↓ |
|---|---|---|
| 9.9k, 4 MB/row | 81% (2.18s -> 0.42s) | 94% (30.9 GB -> 1.84 GB) |
| 1.4M, 182 KB | 35% (95 -> 62 ms) | 98% (892 MB -> 18 MB) |
| 7.8M, 5 KB, OTel-style ids | -54% (37 -> 57 ms) | 93% (369 MB -> 24 MB) |

p50 win tracks payload size: heavy reads get much faster, thin-payload projects add about 20 ms of CTE-rescan overhead (pass-1 re-inlined 3x, fixable) while still cutting bytes ~15x.

## Testing
prod benchmark above; updated calls_query_builder snapshots.

### DIFF
--- a/tests/trace_server/query_builder/test_calls_query_builder.py
+++ b/tests/trace_server/query_builder/test_calls_query_builder.py
@@ -2399,11 +2399,9 @@ def test_total_storage_size(with_filter: bool):
 
 @pytest.mark.parametrize("with_filter", [False, True])
 def test_total_storage_size_calls_complete(with_filter: bool):
-    """calls_complete scopes the total-storage rollup by id (the stats PK).
-
-    Without scoping, the rollup aggregates the entire project's
-    calls_complete_stats. The storage_scope_ids CTE narrows it to the matched
-    calls' traces so the (project_id, id) primary key prunes the scan.
+    """calls_complete scopes the total-storage rollup to the matched calls' traces
+    so the (project_id, id) primary key prunes calls_complete_stats. A filtered/paged
+    read narrows via the two-pass filtered_calls CTE; a bare read uses storage_scope_ids.
     """
     cq = CallsQuery(
         project_id="test/project",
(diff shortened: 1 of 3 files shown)


amwarrier:
**Storage-size read path now materializes the same filter twice.**

Enabling `use_filter_cte` for `include_total_storage_size=True` causes both `storage_scope_ids` and the new `filtered_calls` CTE to be emitted with identical filter bodies. Visible in the updated snapshot at `test_total_storage_size_calls_complete` (`test_calls_query_builder.py:2428-2472`) — both CTEs have the same `PREWHERE project_id = ? WHERE op_name IN [...] AND deleted_at = ? LIMIT 50`.

Since `page_started_at_bound` is gated off for storage-size reads (`calls_query_builder.py:1279-1285`), the two-pass CTE here pays the extra scan cost without getting the PK-prune benefit that justifies it. Net effect on this code path: filter scan is doubled, no perf win.

Two ways to close this:
1. **Keep single-pass for storage-size**: exclude `include_total_storage_size` (and likely `include_storage_size`) from the `use_filter_cte` gate at `calls_query_builder.py:1258-1261`. The pre-PR `storage_scope_ids`-only shape is restored.
2. **Consolidate CTEs**: drop `storage_scope_ids`, use `filtered_calls` in both places it's referenced. Bigger refactor, has implications for the storage-scope join logic.

Option 1 is smaller and preserves the pre-PR perf on this path. Also, the current snapshot test locks in the doubled-materialization shape, so whichever direction you pick, the snapshot needs to reflect the intended shape rather than the incidental one.

Not a blocker for the primary heavy-read win, but worth resolving before merge since the storage-size read is a real UI code path (aggregated storage per row).

gtarpenning:
🤖 @amwarrier benchmarked both options on prod before acting (a fat-payload `calls_complete` project, ~209 KiB avg row; `include_total_storage_size` + heavy fields, `ORDER BY started_at DESC LIMIT 50`; warm, interleaved, all return the same 50 rows):

| variant | p50 read | p50 wall |
|---|---|---|
| single-pass (Option 1 / pre-PR) | **98.4 GB** | 4671 ms |
| two-pass (this PR) | 0.11 GB | 988 ms |
| two-pass + fix `6adae04` | **0.11 GB** | 867 ms |

Option 1 is a ~900x read blowup: the stats-rollup `LEFT JOIN` blocks `LIMIT` pushdown, so single-pass materializes `*_dump` for the whole project before taking 50. Two-pass scopes the heavy read to the page's ids, so it stays.

The dead-CTE call is right though. `storage_scope_ids` was emitted but unreferenced on the two-pass path (CH elides unreferenced CTEs → 0 runtime cost, confirmed above, but it's dead SQL and locked the wrong snapshot). Pushed `6adae04`: emit it only on the single-pass path that references it; two-pass scopes via `filtered_calls`. Snapshots now assert both shapes.

amwarrier (approved):
Thanks. Approved.

PR #7487, opened 2026-06-30
feat(weave): add predict-only cost total to eval results summary

ro31337:
## JIRA Issue(s)
https://coreweave.atlassian.net/browse/WB-33374

## Description
The eval-results summary already returns a predict-only token total (#7364). This adds the matching cost.

The summary now also returns `predict_total_cost` — the model's `predict()` cost summed across trials, without the LLM judge's cost. It follows the same path as tokens: a small `extract_total_cost` helper next to `extract_total_tokens`, a per-trial `total_cost`, and the total on the summary.

It's opt-in. The request has a new `include_costs` flag, and only when it's set do we fetch the cost on the predict call (on both the ClickHouse and in-memory backends), so callers that don't ask for cost pay nothing extra. If no trial has a price (for example the model isn't in the cost table), `predict_total_cost` is `None`, so the client can tell "no cost" apart from a real `$0`.

## Testing
Server tests on both backends: two priced predict trials plus a priced judge check that the total is predict-only and leaves the judge out; a trial with no usage is skipped instead of counted as zero; and the value is `None` when nothing has a price or when `include_costs` isn't set.

## Merge order
1. wandb/weave#7487 (server, this PR)
2. wandb/core#47134 (frontend)

### DIFF
--- a/tests/trace_server/test_trace_server_evaluation_apis.py
+++ b/tests/trace_server/test_trace_server_evaluation_apis.py
@@ -20,6 +20,7 @@ from weave.trace_server.trace_server_interface import (
     CallsDeleteReq,
     CallsQueryReq,
     CallStartReq,
+    CostCreateReq,
     EndedCallSchemaForInsert,
     EvalResultsFilter,
     EvalResultsQueryReq,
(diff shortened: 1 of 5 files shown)


tests/trace_server/test_trace_server_evaluation_apis.py
```diff
@@ -1055,6 +1056,273 @@ def test_eval_results_summary_predict_total_tokens_counts_zero(client):
...
+        )
+    )
+    assert res.summary is not None
+    # 2 priced predict trials; the judge cost and the usage-less trial are excluded.
+    assert res.summary.evaluations[0].predict_total_cost == pytest.approx(
+        2 * _PREDICT_COST_PER_TRIAL
+    )
+
+
+def test_eval_results_summary_predict_total_cost_requires_include_costs(client):
```
gtarpenning: you might be able to parameterize these tests on `prices, trials, include_costs, expected` 🤷🏻

gtarpenning (commented):
@jwlee64 mind taking a look at this too?

weave/trace_server/eval_results_helpers.py
```diff
@@ -238,6 +238,47 @@ def extract_total_tokens(
     return total_tokens if found_any else None
 
 
+# The four components of a call's spent cost, summed to the canonical total_cost
+# (see the total_cost definition in trace_server_interface.py). Populated in
+# summary.weave.costs when calls are read with include_costs=True.
+_COST_COMPONENT_KEYS = (
```
jwlee64, line 244: nit - I think we have a couple places now where we are defining the components of what cost is, eval_results_helper might not be the best place for this, but we can extract later. I think costs just might need a more comprehensive overhaul

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -5662,6 +5662,7 @@ def _eval_results_via_cte(
                     filter=tsi.CallsFilter(parent_ids=batch),
                     columns=child_columns,
                     sort_by=[tsi.SortBy(field="started_at", direction="asc")],
+                    include_costs=req.include_costs,
```
jwlee64, line 5665: non-blocking, but i believe we over-fetch here: this enriches cost for all children (predict + scorer), but we only read predict cost. The scorer price joins are computed and discarded.

jwlee64 (approved):
lgtm, couple non blocking comments, down to merge as is

tests/trace_server/test_trace_server_evaluation_apis.py
ro31337: Fixed

weave/trace_server/clickhouse_trace_server_batched.py
ro31337: thanks! good catch on the over-fetch, filed https://coreweave.atlassian.net/browse/WB-36670 as a follow-up.

weave/trace_server/eval_results_helpers.py
ro31337: agree, I put it next to `extract_total_tokens` to match the existing pattern. It's not ideal, since cost is defined in a few places. Сonsolidating those is a good separate cleanup.

PR #7489, opened 2026-07-01
feat(weave): Base64 and Data URL content ref conversion

zbirenbaum:
## Description

New agents OTel export endpoint and old one were not converting base64 and data urls to content server side like they would be when contained in SDK traces. This fixes it by making chat_view.py recursively check all fields for refs and adding the object save + ref replacement to both paths.

## Testing

Tested with a basic script where images are submitted and model is asked what is in it. 
image

### DIFF
--- a/tests/trace_server/test_genai_chat_view.py
+++ b/tests/trace_server/test_genai_chat_view.py
@@ -1043,6 +1043,58 @@ def test_inline_internal_ref_surfaces_without_span_content_refs() -> None:
     assert _assistant_payload(assistant).content_refs == []
 
 
+@pytest.mark.xfail(
+    reason=(
+        "Known limitation (PR #7489 discussion): the inline-ref sweep "
+        "(_iter_internal_refs) surfaces ANY internal weave ref, not only "
+        "Content/media refs. A message that legitimately carries a ref to a "
+        "non-media object (prompt/model/dataset) in a structured field leaks "
+        "into content_refs and renders as broken media. A media ref is "
+        "indistinguishable from a prompt ref at the string level and the part "
+        "shapes are not normalized, so scoping the sweep is deferred to a "
+        "follow-up (e.g. record converted media refs authoritatively at "
+        "ingest, or scope the sweep to media part positions)."
+    ),
+    strict=False,
+)
+def test_inline_sweep_excludes_non_media_internal_refs() -> None:
+    """The inline-ref sweep must surface only Content/media refs. A message that
+    legitimately carries a ref to a non-media object (a prompt, model, dataset)
+    in a structured field must NOT land in ``content_refs``, or the UI renders
+    it as broken media. Content objects are stored under a sanitized-filename
+    object_id, so a media ref is indistinguishable at the string level from a
+    prompt ref — the walker needs to scope the match.
+    """
+    image_ref = "weave-trace-internal:///PID/object/gift_basket_png:IMGDIGEST"
+    prompt_ref = "weave-trace-internal:///PID/object/describe_image_prompt:PROMPTDIGEST"
+    spans = [
+        _span(
+            span_id="chat",
+            operation_name="chat",
+            input_messages=[
+                {
+                    "role": "user",
+                    "content": _parts(
+                        _text_part("What is in this image?"),
+                        {"type": "image_url", "image_url": {"url": image_ref}},
+                        # Agent quoting the prompt object it ran under — not media.
+                        {"type": "tool_use", "input": {"prompt_ref": prompt_ref}},
+                    ),
+                }
+            ],
+            output_messages=[
+                {"role": "assistant", "content": _parts(_text_part("A gift basket."))}
+            ],
+        ),
+    ]
+
+    messages = build_chat_messages(spans)
+    user = next(m for m in messages if m.type == "user_message")
+
+    # Only the media ref should render; the prompt ref is a false positive today.
+    assert _user_payload(user).content_refs == [image_ref]
+
+
 def test_output_media_attaches_to_assistant_message() -> None:
     """Model-generated media is a part on the output message and renders on
     the assistant bubble (mirror image of the input case).
--- a/tests/trace_server/test_genai_chat_view.py
+++ b/tests/trace_server/test_genai_chat_view.py
@@ -993,6 +993,53 @@ def test_input_media_attaches_to_user_message_not_assistant() -> None:
     assert _assistant_payload(assistant).content_refs == []
 
 
+def test_inline_internal_ref_surfaces_without_span_content_refs() -> None:
+    """Server-side content conversion embeds an internal weave ref directly in
+    the message part (e.g. an OpenAI ``image_url``) with no span-level
+    ``content_refs``. The recursive inline-ref sweep surfaces it so the image
+    still renders on the user bubble.
+    """
+    image_internal = "weave-trace-internal:///PID/object/image-abcd.png:IMGDIGEST"
+    spans = [
+        _span(
+            span_id="agent",
+            operation_name="invoke_agent",
+            agent_name="image-describer",
+            input_messages=[
+                {
+                    "role": "user",
+                    "content": _parts(_text_part("What is in this image?")),
+                }
+            ],
+        ),
+        _span(
+            span_id="chat",
+            parent_span_id="agent",
+            operation_name="chat",
+            input_messages=[
+                {
+                    "role": "user",
+                    "content": _parts(
+                        _text_part("What is in this image?"),
+                        {"type": "image_url", "image_url": {"url": image_internal}},
+                    ),
+                }
+            ],
+            output_messages=[
+                {"role": "assistant", "content": _parts(_text_part("A gift basket."))}
+            ],
+            # No span-level content_refs — the ref lives only inline in the part.
+        ),
+    ]
+
+    messages = build_chat_messages(spans)
+    user = next(m for m in messages if m.type == "user_message")
+    assistant = next(m for m in messages if m.type == "assistant_message")
+
+    assert _user_payload(user).content_refs == [image_internal]
+    assert _assistant_payload(assistant).content_refs == []
+
+
 def test_output_media_attaches_to_assistant_message() -> None:
     """Model-generated media is a part on the output message and renders on
     the assistant bubble (mirror image of the input case).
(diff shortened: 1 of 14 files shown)


tests/trace_server/test_genai_extraction.py
```diff
@@ -415,3 +421,156 @@ def test_extract_custom_attrs_skips_non_finite_floats() -> None:
...
+
+
+def test_ingest_flow_strips_base64_from_span_attribute_dumps() -> None:
+    """Replicates AgentWriteHandler.insert_otel_spans: converting span.attributes
+    before extraction strips base64 from the lossless attributes_dump /
+    raw_span_dump columns (the OpenLLMetry ``gen_ai.prompt.{i}`` shape unflattens
+    to a numeric-keyed dict with a structured content array).
+    """
+    b64 = base64.b64encode(b"a" * 15000).decode("ascii")
+    attrs = {
```
gtarpenning, line 546: what about this shape?
```python
attrs = {"gen_ai.input.messages": json.dumps([{"role": "user", "parts": [{"type": "image", "url": data_uri}]}])}
```
is that legitimate? can we test if we handle that path?

weave/trace_server/agents/clickhouse.py
```diff
@@ -847,12 +854,19 @@ def insert_otel_spans(
...
+                            span.attributes, req.project_id, self._trace_server
+                        )
+
                     try:
                         genai_row = extract_genai_span(
                             span,
                             project_id=req.project_id,
                             wb_user_id=req.wb_user_id or "",
                             wb_run_id=processed_span.run_id or "",
+                            trace_server=self._trace_server,
```
gtarpenning: this is... not ideal. maybe we can break out something like `strip_inline_blobs_from_span` above, so we don't need to pass a ref to the entire trace server? that feels strange

weave/trace_server/agents/clickhouse.py
```diff
@@ -847,12 +854,19 @@ def insert_otel_spans(
                         errors.append(str(e))
                         continue
 
+                    # strip inline base64/data-URIs into weave refs.
+                    if self._trace_server is not None:
+                        span.attributes = replace_base64_with_content_objects(
+                            span.attributes, req.project_id, self._trace_server
+                        )
```
gtarpenning, line 865: this also seems strange at a top level, why would we ever not have a _trace_server?

weave/trace_server/agents/chat_view.py
```diff
@@ -759,6 +768,71 @@ def _directional_content_refs(
     return [r for r in _content_refs(span) if _ref_digest(r) in part_digests]
 
 
+_INTERNAL_REF_PREFIX = f"{WEAVE_INTERNAL_SCHEME}:///"
+
+
+def _iter_internal_refs(value: Any) -> Iterator[str]:
```
gtarpenning: maybe add a max depth?

weave/trace_server/base64_content_conversion.py
```diff
@@ -109,6 +113,36 @@ def store_content_object(
...
+    embed a compact ``weave-trace-internal:///…`` ref in the payload instead of
+    the inline object. The version is the server-computed object digest, so
+    identical content dedupes to the same ref.
+    """
+    obj_val = store_content_object(content_obj, project_id, trace_server)
+    object_id = make_object_id(content_obj.filename, "content")
+    res = trace_server.obj_create(
+        ObjCreateReq(
+            obj=ObjSchemaForInsert(
+                project_id=project_id, object_id=object_id, val=obj_val
```
gtarpenning: don't we need the user id here?

gtarpenning: like `req.wb_user_id`

weave/trace_server/opentelemetry/genai_extraction.py
```diff
@@ -331,31 +329,40 @@ def _normalize_system_instructions(raw: Any) -> list[str]:
...
+    if val is None:
+        for event in events:
+            if event.get("name") == "gen_ai.content.completion":
+                event_attrs = event.get("attributes", {})
+                val = get_attribute(event_attrs, "gen_ai.completion")
+                if val is not None:
+                    break
+    return try_convert_numeric_keys_to_list(val)
 
     return None
```
gtarpenning: 🔪

tests/trace_server/test_genai_chat_view.py
```diff
@@ -993,6 +993,53 @@ def test_input_media_attaches_to_user_message_not_assistant() -> None:
...
+
+    messages = build_chat_messages(spans)
+    user = next(m for m in messages if m.type == "user_message")
+    assistant = next(m for m in messages if m.type == "assistant_message")
+
+    assert _user_payload(user).content_refs == [image_internal]
+    assert _assistant_payload(assistant).content_refs == []
+
+
 def test_output_media_attaches_to_assistant_message() -> None:
```
gtarpenning, line 1098: do we want an end to end test on the agents path? i see one for calls_complete but not the new agents path

gtarpenning: hmm and what about this? 
```python
def test_inline_sweep_excludes_non_media_internal_refs() -> None:
    """The inline-ref sweep must surface only Content/media refs. A message that
    legitimately carries a ref to a non-media object (a prompt, model, dataset)
    in a structured field must NOT land in `content_refs`, or the UI renders it
    as broken media. Content objects are stored under a sanitized-filename
    object_id, so a media ref is indistinguishable at the string level from a
    prompt ref — the walker needs to scope the match.
    """
    image_ref = "weave-trace-internal:///PID/object/gift_basket_png:IMGDIGEST"
    prompt_ref = "weave-trace-internal:///PID/object/describe_image_prompt:PROMPTDIGEST"
    spans = [
        _span(
            span_id="chat",
            operation_name="chat",
            input_messages=[
                {
                    "role": "user",
                    "content": _parts(
                        _text_part("What is in this image?"),
                        {"type": "image_url", "image_url": {"url": image_ref}},
                        # Agent quoting the prompt object it ran under — not media.
                        {"type": "tool_use", "input": {"prompt_ref": prompt_ref}},
                    ),
                }
            ],
            output_messages=[
                {"role": "assistant", "content": _parts(_text_part("A gift basket."))}
            ],
        ),
    ]

    messages = build_chat_messages(spans)
    user = next(m for m in messages if m.type == "user_message")

    # Only the media ref should render; the prompt ref is a false positive today.
    assert _user_payload(user).content_refs == [image_ref]
```

gtarpenning (commented):
very close to approval but passing around the `trace_server` is an anti-pattern I think, any way to avoid it?

weave/trace_server/base64_content_conversion.py
zbirenbaum: I think this is a pre-existing issue. I'll address it separately.

weave/trace_server/agents/chat_view.py
zbirenbaum: Added

weave/trace_server/agents/clickhouse.py
zbirenbaum: Addressed in latest commit

weave/trace_server/agents/clickhouse.py
zbirenbaum: Well we definitely should always have one, but the typing validation isn't set up properly. It's cosmetic

tests/trace_server/test_genai_extraction.py
zbirenbaum: Added

tests/trace_server/test_genai_chat_view.py
zbirenbaum: Known limitation. added test and documented

weave/trace_server/opentelemetry/genai_extraction.py
gtarpenning: this is still broken

weave/trace_server/base64_content_conversion.py
```diff
@@ -109,13 +113,48 @@ def store_content_object(
     }
 
 
+@traced(name="store_content_object_ref")
+def store_content_object_ref(
```
gtarpenning, line 117: just to be sure sure here, we are okay flipping this on for all the existing calls table stuff too? would be great to validate this all works in the UI etc  on the old paths as well (in the pr description i see the new path which looks great, but old path maybe could use a quick once over too)

weave/trace_server/trace_server_converter.py
```diff
@@ -34,11 +35,52 @@
...
+# payloads store refs inside a JSON-serialized string (e.g. an agent message's
+# ``content`` parts array); the mapper re-parses those and re-runs itself over
+# the decoded structure. A JSON blob can itself hold another JSON-string leaf,
+# so an adversarial or pathological payload could nest arbitrarily and exceed
+# Python's recursion limit. Realistic nesting is a handful of levels deep, so 8
+# leaves ample headroom while staying far below ``sys.getrecursionlimit()``
+# (1000 by default). This bounds only the JSON-in-JSON descent — the underlying
+# ``_walk`` container traversal is unchanged. Mirrors
+# ``chat_view._MAX_REF_SEARCH_DEPTH``.
+_MAX_REF_SEARCH_DEPTH = 8
```
gtarpenning, line 48: we have this huge comment and constant and json resurse logic in two places, consider DRYing it out, especially with the huge comment etc.

gtarpenning (approved):

weave/trace_server/base64_content_conversion.py
zbirenbaum: I tested it out and confirmed the old path does a proper ref conversion, but does not properly render an image. 

I wrote a followup in the FE, tiny change and fixes the issue

weave/trace_server/trace_server_converter.py
zbirenbaum: Streamlined logic and cleaned up comments.

weave/trace_server/opentelemetry/genai_extraction.py
zbirenbaum: Removed the return None (I believe this is what you meant by broken)

Technically since it was changing in place there wasn't a 'break' but it is sketchy looking.

PR #7496, opened 2026-07-01
fix(weave): idempotent migrations, additive 029, and partial-migration auto-recovery

gtarpenning:
## Summary

- **029 is additive, not a rename.** `ADD COLUMN IF NOT EXISTS expire_at` on `calls_complete`, keeping `ttl_at`. A `RENAME COLUMN` (even `IF EXISTS`) still spawns a mutation that collides with the patch parts from concurrent `call_end_v2` lightweight `UPDATE`s under load on CH 25.10.x ([ClickHouse#99164](https://github.com/ClickHouse/ClickHouse/pull/99164)), and it drops `ttl_at` (a rolled-back 0.80.x app then writes a column that no longer exists, silently losing writes). Additive avoids both; no application change (0.82 already writes `expire_at`).
- **Auto-recovery of partial migrations.** A migration interrupted mid-DDL leaves `partially_applied_version` set; the migrator now re-runs that idempotent migration to converge, instead of hard-raising and crash-looping on every restart. Recovery is gated: only the migration immediately after the recorded version, and never a one-shot (006 seed re-insert, 024 table-swap RENAME — re-running them is not a safe no-op), is re-run; anything else, or a re-run failure, raises for manual repair. **No retry cap by design** — a cap would re-latch into the manual-only crash-loop this PR exists to kill when ClickHouse has a transient failure streak; the crash-loop cadence is already bounded by the lock + k8s backoff.
- **Idempotent everywhere.** `IF [NOT] EXISTS` / `IF EXISTS` guards on every up + down statement, so a recovery re-run is a safe no-op through the cloud/replicated/distributed rewriters. 006's cost seed and 024's table swap remain one-shot by design (documented in their `up.sql`).
- **Divergence guard.** The migrator refuses to run when a data DB already holds migration-managed tables but has no row in the migrations table (a diverged `management_db`, e.g. a rename or a one-sided restore), rather than re-applying DDL from version 0 against tables of unknown schema.

## Testing

- `test_partial_migration_auto_recovers` (cloud/replicated/distributed, real CH): forge a crashed-mid-migration row, then re-run and assert self-heal to latest with the flag cleared and schema byte-for-byte unchanged; a one-shot swap (024) is refused and the flag survives for manual repair.
- `test_production_migrations_are_idempotent` (cloud/replicated/distributed): the apply-twice guarantee. Apply all migrations, rewind the recorded version, re-apply through each rewriter, assert table/column/engine schema is byte-for-byte unchanged. Swap set derived from SQL (`_table_swap_versions`), pinned `[24]` so a new one-shot can't land unnoticed.
- `test_migrations_refuse_populated_db_without_history` (cloud/replicated): a fresh management DB pointed at an already-migrated data DB raises.
- Unit: partial-migration auto-recovery gate (recovers `curr+1`, raises on any other partial or a denylisted one-shot without executing).
- Validated on a real 1 shard x 3 replica ClickHouse (25.12): full migrate + 3-replica schema convergence (`uniqExact(cityHash64(create_table_query)) = 1`) + auto-recovery self-heal. Leg-C A/B on CH 25.10.7.6 under concurrent call_start `INSERT` + `call_end_v2` lightweight `UPDATE`: old `RENAME` fails (`MutatePart` error 49, 2/2 rounds); additive 029 clean (0/2). Shipped for the 0.82.x line as `wandb/weave-trace:0.82.2-migration-hardening.1` (wandb/core#47401).

## Follow-ups (not in this PR)

- **Alert on migrator CrashLoopBackOff**: the real fix for "44h to notice" is alerting, not schema. Recovery self-heals the transient case; a genuinely stuck migration still crash-loops (by design, unbounded count but bounded cadence: one attempt per pod start, serialized by the migration lock, throttled by k8s backoff plateauing at ~5m) and should page.

### DIFF
--- a/tests/trace_server/test_clickhouse_trace_server_migrator.py
+++ b/tests/trace_server/test_clickhouse_trace_server_migrator.py
@@ -26,6 +26,26 @@ DEFAULT_MIGRATION_DIR = os.path.abspath(
 )
 
 
+def _count_result(n: int) -> Mock:
+    """Mock a `SELECT count()` query result with a single row/column of `n`."""
+    res = Mock()
+    res.result_rows = [(n,)]
+    return res
+
+
+def _stub_divergence_precheck(sql: str) -> Mock | None:
+    """Stub the divergence pre-check queries in `_apply_migrations_locked`.
+
+    A row in the migrations table (count 1) makes the guard a no-op, so tests
+    that mock the inner migration methods reach the path they exercise.
+    """
+    if "system.tables" in sql:
+        return _count_result(0)
+    if "count()" in sql and ".migrations" in sql:
+        return _count_result(1)
+    return None
+
+
 def _make_ch_client(database_engine: str = "Atomic") -> Mock:
     """Create a mock CH client that returns *database_engine* for system.databases queries.
 
@@ -42,6 +62,9 @@ def _make_ch_client(database_engine: str = "Atomic") -> Mock:
     def _query_side_effect(sql, *args, **kwargs):
         if "system.databases" in sql:
             return engine_result
+        precheck = _stub_divergence_precheck(sql)
+        if precheck is not None:
+            return precheck
         return Mock()
 
     ch_client.query.side_effect = _query_side_effect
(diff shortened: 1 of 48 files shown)


neutralino1 (commented):
Any chance this can hide real issues?

gtarpenning:
> Any chance this can hide real issues?

we already do this for most migrations, so i dont think this makes it meaninfully worse. to the meat of your question though, i can't really think of any case where we are in a broken db state and continuing is worse than failing hard. either way the db doesn't match the schema, its in a partial state, so neither the pre or post migration code is guaranteed to be safe... but I will think on this

amwarrier (approved):

neutralino1 (approved):

PR #7504, opened 2026-07-02
perf(weave): compute numeric-histogram bounds once (3 map scans -> 2)

gtarpenning:
## Summary
- The numeric-histogram `bounds` CTE was referenced twice, and ClickHouse re-evaluates table CTEs, so the `custom_attrs_int` map was decoded 3x (bounds x2 + aggregated_data x1).
- Computing `bounds` once as a scalar `WITH ... AS bounds_tuple` cuts it to 2 decodes. Read-time query-builder change only; 24-bucket output is byte-identical.

## Performance
local synthetic (prod read-only temporarily unreachable), 10M spans, key on 85% of rows, 24 buckets:

| metric | before | after | delta |
| --- | --- | --- | --- |
| value_rows scans (read_rows) | 3x (30.0M) | 2x (20.0M) | 3 -> 2 |
| read_bytes | 1.25 GiB | 854 MiB | -33% |
| elapsed (default threads) | 0.334s | 0.231s | -31% |
| peak_mem (max_threads=1) | 11.2 MiB | 10.7 MiB | bounded |

## Testing
updated the full-SQL unit tests; result set md5-identical before/after in local ClickHouse.

### DIFF
--- a/weave/trace_server/query_builder/agent_stats_query_builder.py
+++ b/weave/trace_server/query_builder/agent_stats_query_builder.py
@@ -801,10 +801,12 @@ def _build_numeric_bucket_stats_query(
         if bucket.max is not None
         else "max(bucket_value)"
     )
-    # bounds is a scalar tuple evaluated once: (min, max, count).
-    bounds_min = f"{_BOUNDS_TUPLE}.1"
-    bounds_max = f"{_BOUNDS_TUPLE}.2"
-    bounds_count = f"{_BOUNDS_TUPLE}.3"
+    # bounds computed once as a scalar tuple (min, max, count).
+    # ClickHouse does NOT materialize table CTEs - each reference re-scans
+    # value_rows (re-decoding the attrs map). Do NOT convert back to a CTE.
+    bounds_min = f"{_BOUNDS_TUPLE}.min_bound"
+    bounds_max = f"{_BOUNDS_TUPLE}.max_bound"
+    bounds_count = f"{_BOUNDS_TUPLE}.value_count"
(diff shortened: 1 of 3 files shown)


weave/trace_server/query_builder/agent_stats_query_builder.py
```diff
@@ -873,13 +881,14 @@ def _build_numeric_bucket_stats_query(
...
-        SELECT
-          toFloat64({min_bound_sql}) AS bucket_min_bound,
-          toFloat64({max_bound_sql}) AS bucket_max_bound,
-          count() AS value_count
+      (
+        SELECT CAST(tuple(
+          toFloat64({min_bound_sql}),
+          toFloat64({max_bound_sql}),
+          count()
+        ) AS Tuple(min_bound Nullable(Float64), max_bound Nullable(Float64), value_count UInt64))
```
jtschoonhoven, line 889: Nit: I think these don't actually need to be nullable

jtschoonhoven (approved):

weave/trace_server/query_builder/agent_stats_query_builder.py
gtarpenning: `min(bucket_value) / max(bucket_value)` over an empty `value_rows` return `NULL` so i think we need these

gtarpenning: i should have a test to cover this actually

gtarpenning: oh i do

PR #7505, opened 2026-07-02
fix(weave): preserve provider HTTP status code in completion error payload

gtarpenning:
## Summary
- litellm exceptions carry a `status_code` (they subclass the openai error hierarchy); `_litellm_error_response` currently drops it and only keeps the stringified message.
- Preserve it as `ERROR_STATUS_CODE_KEY` on the `CompletionsCreateRes` error payload so consumers can classify a failure by status class (4xx account/config vs 5xx availability) instead of matching the error string.
- Enables the agent-scoring worker to stop paging on customer quota errors routed through our inference gateway (wandb/core follow-up).

## Testing
unit test

### DIFF
--- a/weave/trace_server/llm_completion.py
+++ b/weave/trace_server/llm_completion.py
@@ -383,13 +383,19 @@ def _build_litellm_kwargs(
     return completion_kwargs
 
 
+# Key in the `CompletionsCreateRes.response` error payload holding the provider
+# HTTP status code, so consumers can classify a failure by status class (4xx
+# account/config vs 5xx availability) instead of matching the error string.
+ERROR_STATUS_CODE_KEY = "error_status_code"
+
+
 def _litellm_error_response(e: Exception) -> tsi.CompletionsCreateRes:
     response: dict[str, Any] = {"error": str(e).replace("litellm.", "")}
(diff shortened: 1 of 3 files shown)


weave/trace_server/llm_completion.py
```diff
@@ -383,9 +383,20 @@ def _build_litellm_kwargs(
...
+# Key in the `CompletionsCreateRes.response` error payload holding the provider
+# HTTP status code, so consumers can classify a failure by status class (4xx
+# account/config vs 5xx availability) instead of matching the error string.
+ERROR_STATUS_CODE_KEY = "error_status_code"
+
+
 def _litellm_error_response(e: Exception) -> tsi.CompletionsCreateRes:
-    error_message = str(e).replace("litellm.", "")
-    return tsi.CompletionsCreateRes(response={"error": error_message})
+    response: dict[str, Any] = {"error": str(e).replace("litellm.", "")}
```
jtschoonhoven: tiny nit: consider replacing just the first appearance of "litellm." instead of all occurences
```suggestion
    response: dict[str, Any] = {"error": str(e).replace("litellm.", "", 1)}
```

jtschoonhoven (approved):

PR #7507, opened 2026-07-02
fix(weave): route agent traces to active project on re-init

gtarpenning:
## Summary
- Agent/conversation OTel spans routed by the `TracerProvider` **Resource**, which is baked in on the first `weave.init()` and never rebuilt (OTel's global provider is set-once, its Resource immutable). A second `weave.init(project2)` left agent traces exporting under `project1` while classic traces correctly followed `project2`.
- Fix: carry the project in the mutable `project_id` export header instead of the Resource, and reroute the live exporter (flush-then-reheader) when init switches projects. `weave/trace/weave_init.py` `_setup_conversation_tracing`.
- Server precedence is Resource-attrs > `project_id` header, so the project is dropped from the Resource to let the header control routing (header-only routing confirmed on prod; it is the older/legacy ingest path so it is broadly supported).
- Scope: fixes sequential re-init. Routing is per-export batch, so one process interleaving concurrent work across projects is best-effort (was fully broken before, so this only improves it).

## Proof (prod, api.wandb.ai)
Procedure: `weave.init(A)` -> emit one agent turn (`log_turn`); `weave.init(B)` -> emit one agent turn; query `POST /agents/spans/query` for each project.

| project | before fix | after fix |
|---|---|---|
| A (1st init) | **2** spans (its own + B'"'"'s) | 1 span |
| B (2nd init) | **0** spans (bled to A) | 1 span |

Agents -> Spans UI for each project (note `agent-for-bleed-b` bleeding into project A, and B empty):

| | Before fix (bled) | After fix |
|---|---|---|
| 2nd-init project (B) |  |  |
| 1st-init project (A) |  |  |

## Testing
adds `tests/trace/test_conversation_tracing_setup.py` (reroute-on-reinit + leaves a foreign provider untouched); plus the prod before/after above.

### DIFF
--- a/weave/trace/weave_init.py
+++ b/weave/trace/weave_init.py
@@ -33,10 +33,22 @@ from weave.trace_server_version import MIN_TRACE_SERVER_VERSION
 from weave.wandb_interface.context import get_wandb_api_context
 
 if TYPE_CHECKING:
+    from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+        OTLPSpanExporter,
+    )
+    from opentelemetry.sdk.trace import TracerProvider
+
     from weave.trace.op import PostprocessInputsFunc, PostprocessOutputFunc
     from weave.trace_server.service_interface import ServerInfoRes
 
 logger = logging.getLogger(__name__)
+
+# The conversation/agent OTel provider is set once (OTel's global provider is
+# set-once and its Resource is immutable). A later weave.init() to a different
+# project reroutes by rewriting the exporter's project_id header instead of
+# rebuilding, so agent spans follow the active project.
+_conversation_tracer_provider: TracerProvider | None = None
+_conversation_span_exporter: OTLPSpanExporter | None = None
 
 
 class WeaveWandbAuthenticationException(Exception): ...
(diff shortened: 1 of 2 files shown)


weave/trace/weave_init.py
```diff
@@ -135,33 +150,42 @@ def _setup_conversation_tracing(entity: str, project: str, api_key: str | None)
...
-        return
-
     trace_server_url = env.weave_trace_server_url()
     if not trace_server_url:
         return
 
+    project_id = f"{entity}/{project}"
+
+    # Re-init to a new project: our provider is already the global one and can't
+    # be swapped (OTel's set-once), so reroute the live exporter. Flush first so
```
zbirenbaum: A timeout here could be good so we don't block re-init indefinitely but then you would drop the traces or route them to the wrong place so I guess there's not a better solution. Might be good to document though.

zbirenbaum (approved):
One comment looks great

weave/trace/weave_init.py
gtarpenning: added in most recent commit (+ a couple other hardening bits)

PR #7511, opened 2026-07-02
fix(weave): canonicalize self-refs in generated builtin object schemas

gtarpenning:
## Summary
- The generated builtin object class schema JSON is non-reproducible: pydantic emits a degenerate `{"title": X}` placeholder instead of `{"$ref": "#/$defs/X"}` for some recursive self-references, and which members come out degenerate depends on the installed pydantic version. Regenerating in a different env produced spurious diffs (reported on latest master).
- `build_schema()` now rewrites those placeholders to a stable `$ref`, so output is deterministic across pydantic versions. Committed file is unchanged (already canonical on master).

## Testing
adds a test asserting the committed schema equals canonical generator output and that self-refs canonicalize to `$ref`.

### DIFF
--- a/scripts/generate_base_object_schemas.py
+++ b/scripts/generate_base_object_schemas.py
@@ -17,30 +17,47 @@ OUTPUT_DIR = (
 )
 OUTPUT_PATH = OUTPUT_DIR / "generated_builtin_object_class_schemas.json"
 
+JSONValue = dict[str, "JSONValue"] | list["JSONValue"] | str | int | float | bool | None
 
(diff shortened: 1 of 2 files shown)


gtarpenning:
Closing: wrong direction. The `{title}` form is the *intended* cycle-break from the frontend's resolve_schema_circular_refs.py (quicktype can't handle recursive $ref); canonicalizing the generator to $ref fights that. Real root cause is the frontend build rewriting the committed JSON in place. Reworking as a unified generation process.

PR #7516, opened 2026-07-05
feat(weave): add Ollama integration

NishchayMahor:
## Description

Adds a Weave integration for **[Ollama](https://github.com/ollama/ollama-python)**, so chat calls made with the `ollama` Python client are automatically traced as Weave ops — no code changes required beyond `weave.init()`. Ollama is the most widely used way to run local models, so this lets Weave users observe local-model apps the same way they already can with the hosted providers.

Patches all three entry points, with metadata stamped on every call:
- `ollama.chat(...)` — the module-level helper (bound to a shared client at import, so it's patched separately)
- `ollama.Client(...).chat(...)`
- `ollama.AsyncClient(...).chat(...)`

The three symbols don't overlap on a single call, so nothing is double-counted (verified in the tests).

## Changes
- `weave/integrations/ollama/ollama_sdk.py` — `get_ollama_patcher` (sync + async + module helper).
- Registered in `weave/integrations/patch.py`, `weave/integrations/__init__.py`, and `AutopatchSettings` (`weave/trace/autopatch.py`).
- Added the `ollama` optional-dependency extra (`pyproject.toml`) and the CI integration entry (`noxfile.py`).
- Mock-based unit tests in `tests/integrations/ollama/ollama_test.py`.

## Testing
`pytest tests/integrations/ollama/ollama_test.py --trace-server=fake` → **3 passed** (module helper, explicit client, async), each producing exactly one Weave call with `integration.name == "ollama"`. The tests mock the client's HTTP layer, so no running Ollama server is needed. `ruff` and `mypy` clean on the new module.


NishchayMahor:

NishchayMahor:
still keen on landing this one, it's green and conflict-free on my end. glad to rebase or adjust anything whenever a maintainer has a moment. thanks!

NishchayMahor:
This got auto-labelled `Stale`, but it's still current and I'd like to keep it open — it's `MERGEABLE` against `main` with checks green (the earlier red check has since cleared). Happy to rebase or split it up if that helps review; just let me know what would make it easier to land.

NishchayMahor:
Rebased onto latest `master` — the branch was 156 commits behind, so I wanted it current before anyone spends time reviewing rather than after. `behind_by=0` now, still `MERGEABLE`.

A plain rebase wasn't possible (`git merge-base` against `master` came back empty, so history has been rewritten upstream since this was cut). I recreated the branch from current `master` and cherry-picked the single commit, so the diff is unchanged — 8 files, the Ollama integration plus its test.

Re-checked against current `master` that the integration contract hasn't moved: `_patch_integration` still takes the same `module_path`/`patcher_func_getter_name`/`triggering_symbols`/`settings` arguments, `INTEGRATION_MODULE_MAPPING` is still the registry, and `AutopatchSettings` still declares one `IntegrationSettings | None` field per integration. The cherry-pick was clean with no conflicts.

Happy to split or adjust if that helps review.

NishchayMahor:
Rebased onto current `master` — was 52 behind, now 0, still `MERGEABLE` with checks green. Clean rebase this time; the merge base was healthy, and nothing upstream had touched `weave/integrations/ollama/`.

While rebasing I noticed something about this PR that I think explains why it's been easy to overlook, and it's worth flagging even if the PR itself goes nowhere:

**The Ollama tests have never actually run in CI.** The PR registers the `ollama` shard in `noxfile.py` and adds the `ollama` extra to `pyproject.toml`, but `.github/workflows/test.yaml`'s `nox-shard` matrix was never updated — every other integration (`anthropic`, `cerebras`, `cohere`, `groq`, `mistral`…) has an entry there and this one doesn't. So the green checkmarks on this PR are green because `tests/integrations/ollama/` is never executed, not because it passed.

That's my omission, and I'd rather say so than let it get discovered in review.

I stopped short of fixing it here, for a reason worth knowing: adding the shard to the matrix alone would make the job fail immediately. CI installs with `uv sync --frozen`, and `uv.lock` was never regenerated for the new extra, so it errors with ``Extra `ollama` is not defined``. Regenerating the lock locally added `ollama v0.6.2` correctly but also produced ~200 lines of unrelated churn in `spacy`/`scikit-learn`/`presidio` markers, which I assume is a uv version difference from whoever generated the committed lock. That didn't feel like mine to push into your lockfile unreviewed.

So: happy to wire up the matrix entry and regenerate the lock if you tell me which uv version you'd like it done with, or to leave it to you as a one-liner. Either way you'd then get a real signal on this PR instead of a vacuous one.

I couldn't run the integration tests locally to confirm they'd pass once wired — they need a ClickHouse container via Docker, which I don't have here. What I did verify on the rebased branch is that the patcher builds and registers in `autopatch`. I'm not claiming more than that.

If Ollama isn't something you want in weave, closing this is a completely fine outcome and I won't take it personally — I'd just rather that be a decision than a 14-day timeout.

PR #7517, opened 2026-07-05
feat(weave): add Together AI integration

NishchayMahor:
## Description

Adds a Weave integration for **[Together AI](https://github.com/togethercomputer/together-python)**, so chat completion calls made with the `together` Python client are automatically traced as Weave ops after `weave.init()`. Together is a widely used inference platform (Llama, DeepSeek, Qwen, and other open models); this lets Weave users observe those calls with no code changes.

Patches both entry points, with integration metadata stamped on every call:
- `together.Together(...).chat.completions.create(...)`
- `together.AsyncTogether(...).chat.completions.create(...)`

## Changes
- `weave/integrations/together/together_sdk.py` — `get_together_patcher` (sync + async), mirroring the existing `cerebras` integration.
- Registered in `weave/integrations/patch.py`, `weave/integrations/__init__.py`, and `AutopatchSettings` (`weave/trace/autopatch.py`).
- Added the `together` optional-dependency extra (`pyproject.toml`) and the CI integration shard (`noxfile.py`).
- Mock-based unit tests in `tests/integrations/together/together_test.py`.

## Testing
`pytest tests/integrations/together/together_test.py --trace-server=fake` → **2 passed** (sync + async), each producing exactly one Weave call with `integration.name == "together"`. The tests mock the client's HTTP layer, so no Together API key or network is needed. `ruff` and `mypy` clean on the new module.


NishchayMahor:
same as the ollama pr, this is still current and passing for me. happy to iterate on anything, just flagging it's ready whenever there's review bandwidth. thanks!

NishchayMahor:
This got auto-labelled `Stale`, but it's still current and I'd like to keep it open — it's `MERGEABLE` against `main` with checks green (the earlier red check has since cleared). Happy to rebase or split it up if that helps review; just let me know what would make it easier to land.

NishchayMahor:
Rebased onto latest `master` — the branch was 156 commits behind, so I wanted it current before anyone spends time reviewing rather than after. `behind_by=0` now, still `MERGEABLE`.

A plain rebase wasn't possible (`git merge-base` against `master` came back empty, so history has been rewritten upstream since this was cut). I recreated the branch from current `master` and cherry-picked the single commit, so the diff is unchanged — 8 files, the Together AI integration plus its test.

Re-checked against current `master` that the integration contract hasn't moved: `_patch_integration` still takes the same `module_path`/`patcher_func_getter_name`/`triggering_symbols`/`settings` arguments, `INTEGRATION_MODULE_MAPPING` is still the registry, and `AutopatchSettings` still declares one `IntegrationSettings | None` field per integration. The cherry-pick was clean with no conflicts.

Happy to split or adjust if that helps review.

NishchayMahor:
Rebased onto current `master` — was 52 behind, now 0, still `MERGEABLE` with checks green. Clean rebase, healthy merge base, nothing upstream had touched `weave/integrations/together/`.

Same caveat as the Ollama PR (#7516), and it applies here identically: **the Together tests have never run in CI.** The `together` shard is registered in `noxfile.py` and the extra is in `pyproject.toml`, but there's no entry in `.github/workflows/test.yaml`'s `nox-shard` matrix, so `tests/integrations/together/` is simply never executed. The green checks on this PR aren't telling you the integration works — they're telling you nothing about it at all. My omission.

I didn't fix it in this PR because the matrix entry alone would fail: CI runs `uv sync --frozen` and `uv.lock` was never regenerated for the new extra, so it errors with ``Extra `together` is not defined``. Regenerating locally worked but dragged in ~200 lines of unrelated marker churn in other packages, which looked like a uv version mismatch with whoever generated the committed lock — not something I wanted to push into your lockfile uninvited.

Glad to do both properly if you'd like, or to leave it as a one-liner for whoever picks this up.

What I could not verify: the tests themselves. They need a ClickHouse container via Docker, which isn't available to me here, so I've confirmed only that the patcher builds and registers in `autopatch` on the rebased branch.

As with the other one — if Together AI isn't a direction you want, closing this is a fine answer. I'd just rather it be a call someone makes than a timeout.

PR #7519, opened 2026-07-06
chore(weave): add column_visibility to ComparisonViewDefinition

gtarpenning:
## Summary
- Saved eval comparison views stored only `evaluation_call_ids` + `selected_metrics`, so results-table column choices were never persisted. Adds an optional `column_visibility` (field -> visible) + regenerated schema; absent/empty = default columns.
- Frontend consumer (capture on save, restore on load): wandb/core PR for https://github.com/wandb/core/pull/47403

## Testing
additive optional field; existing base-object round-trip tests unaffected.


nikumar1206 (approved):

nikumar1206:
lgtm

PR #7524, opened 2026-07-07
feat(weave): first-class sdk feedback for agent spans + turns

gtarpenning:
## Summary
- `client.get_agent_span_feedback(span_id)` / `get_agent_turn_feedback(trace_id)` / `get_agent_conversation_feedback(conversation_id)` return a `RefFeedbackQuery` for agent refs, no more hand-building `weave:///.../agent_span/...` uris
- `add_reaction` on an agent ref now writes `wandb.agent_user_feedback` with scorer_tags (the shape the Agent UI's thumbs write and render) instead of `wandb.reaction.1`, which the agent view ignores; call/object refs unchanged

## Usage
```python
fb = client.get_agent_span_feedback(span_id)
fb.add_reaction("👍")   # renders as the Agent UI thumbs
fb.add_note("solid answer")
# same interface for turns and conversations
client.get_agent_turn_feedback(trace_id).add_reaction("👍")
```

## Testing
functional test creating span/turn/conversation/call feedback then asserting the full stored rows, both backends

### UI

image

### DIFF
--- a/tests/trace/test_client_feedback.py
+++ b/tests/trace/test_client_feedback.py
@@ -1,5 +1,6 @@
 import pytest
 
+from weave.trace.feedback import RefFeedbackQuery
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.errors import InvalidRequest
 from weave.trace_server.interface.query import Query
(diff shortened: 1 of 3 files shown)


rgao-coreweave (approved):
👍👍👍 great work!

PR #7133, opened 2026-06-10
feat(weave): nest LLM/Tool/SubAgent under Turn & SubAgent, add SubAgent.start_subagent()

rgao-coreweave:
## Summary

Two changes to the Conversation SDK's span factories. **Stacked on #7526**, which renames the factory methods to `start_*`.

1. **Add `SubAgent.start_subagent()`** — nest a sub-agent under a sub-agent (previously impossible).
2. **Fix out-of-block child nesting.** A child built inside `with parent:` but entered *after* the parent exits (queue worker, callback, async dispatch) used to detach — nesting under the drifted ambient context, or becoming a root span under a `Turn` (a `Conversation` has no span). Every span now captures its own OTel context in `_start_otel_span` and threads it to children via `_thread_otel_context` on `_SpanBase`, so `Turn`/`SubAgent` children pin to the correct parent regardless of ambient drift — **including across threads** (the parent context is threaded explicitly on the object, not via thread-local ambient context; the `_current_*` contextvar reset is now best-effort so cross-thread cleanup doesn't raise). Cross-**process** is unsupported (in-memory context; that needs OTel propagation).

```python
with turn:
    sa = turn.start_subagent(name="researcher")
    with sa:
        llm = sa.start_llm(model="gpt-4o")  # built here, not entered
    with llm:  # pre-fix: nested under Turn; post-fix: under SubAgent
        ...
```

## Test plan

- [x] `tests/conversation/test_subagent_nesting.py` (16 tests): factory shape, in-block + 3-level nesting, out-of-block nesting (Turn & SubAgent), cross-thread nesting, bare-`LLM`/`Tool` backwards-compat.
- [x] Full `tests/conversation/` suite green (325); `ruff check` / `format` clean.

PR #7526, opened 2026-07-07
chore(weave): rename Turn/SubAgent factory methods to start_*()

rgao-coreweave:
## Summary

Rename the instance factory methods on `Turn` and `SubAgent` — `llm`/`tool`/`subagent` → `start_llm`/`start_tool`/`start_subagent` — to match the module-level `start_*` functions. The old names stay as `@deprecated` aliases that forward to the new ones.

These methods have shipped since **v0.52.38**, so a hard rename would break users; the deprecated aliases keep them working (with a `DeprecationWarning`).

Split out of #7133 per review — the rename is orthogonal to the span-nesting fix there, so #7133 stacks on top of this.

## Test plan
- [x] Full `tests/conversation/` suite green (309); deprecated aliases warn + forward (`test_deprecation.py`).
- [x] `ruff check` / `ruff format` clean.

### DIFF
--- a/tests/conversation/test_conversation.py
+++ b/tests/conversation/test_conversation.py
@@ -326,16 +326,16 @@ class TestSubAgent:
 
     def test_llm_returns_llm_and_inherits_model(self) -> None:
         sa = SubAgent(name="research-bot", model="gpt-4o-mini")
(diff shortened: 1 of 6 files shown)


drtangible (approved):
🔥 🔥 🔥

PR #7528, opened 2026-07-07
perf(weave): batch content object inserts in calls_complete

gtarpenning:
## Summary
- calls_complete stored each embedded content object with its own obj_create (collision query + object_versions insert + aliases insert, 2 serial CH round trips per object); object-heavy batches spent minutes in this loop
- collect content objects during the batch walk and write them with one obj_create_batch per project (one object_versions insert + one aliases insert), deduped by digest
- collision guard runs before the insert (one query per kind, full set, incl. intra-batch conflicts); a colliding content object keeps its original inline value in the payload instead of a dangling ref (OTel path unchanged)
- content-ref cache publishes only after the whole batch commits (objects, files, calls), so a cached ref always resolves; obj_create_batch now validates expected_digest

## Performance
QA A/B, identical driver both builds (one calls_complete batch: 8 calls, 15 distinct ~50-90KB content objects, ~2.2MB body), server-side Datadog spans:

| master 62c7ff3: 9.20s cold, 15+15 object_versions/aliases inserts per request | PR 8c4d091: 2.22s cold, 1+1 inserts per request |
|---|---|
| master | pr |

one create_obj_batch span carried all 15 objects in 0.3s; no new error classes on the PR build. insert reduction scales with object count (prod exemplar with 228 objects: 456 inserts -> 2).

## Testing
functional test: batch with base64 payloads asserts refs resolve, dupes collapse, and exactly one insert per table; collision tolerance now asserts the inline value is restored verbatim and the skipped ref never poisons the cache

### DIFF
--- a/weave/trace_server/clickhouse_trace_server_batched.py
+++ b/weave/trace_server/clickhouse_trace_server_batched.py
@@ -434,6 +434,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
             self._call_batch
             or self._calls_complete_batch
             or self._file_batch
+            or self._content_obj_batch
             or self._bucket_uploads
         ):
             try:
@@ -444,6 +445,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
                 self._file_batch = []
                 self._call_batch = []
                 self._calls_complete_batch = []
+                self._content_obj_batch = []
                 self._bucket_uploads = BucketUploadBatch()
                 self._flush_immediately = True
 
@@ -503,6 +505,16 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
     @_calls_complete_batch.setter
     def _calls_complete_batch(self, value: list[list[Any]]) -> None:
         self._thread_local.calls_complete_batch = value
+
+    @property
+    def _content_obj_batch(self) -> list[tsi.ObjSchemaForInsert]:
+        if not hasattr(self._thread_local, "content_obj_batch"):
+            self._thread_local.content_obj_batch = []
+        return self._thread_local.content_obj_batch
+
+    @_content_obj_batch.setter
+    def _content_obj_batch(self, value: list[tsi.ObjSchemaForInsert]) -> None:
+        self._thread_local.content_obj_batch = value
 
     @classmethod
     def from_env(cls, use_async_insert: bool = True, **kwargs: Any) -> Self:
@@ -894,6 +906,7 @@ class ClickHouseTraceServer(tsi.FullTraceServerInterface):
             self._file_batch = []
             self._call_batch = []
             self._calls_complete_batch = []
+            self._content_obj_batch = []
             self._bucket_uploads = BucketUploadBatch()
             self._flush_immediately = True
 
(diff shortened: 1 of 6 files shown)


gtarpenning:
## 🔴 Red-team report

> [!NOTE]
> **Verdict: no red test warranted.** 5 attack vectors exercised against real ClickHouse in an isolated worktree (+ independent second pass); all held. `tests/trace_server/test_calls_complete.py` (33) green on branch. Logging for coverage.

### Attack matrix

| # | Attack vector | Hypothesis it would break | Result | Evidence |
|---|---|---|:--:|---|
| 1 | 4-char `object_id` collision | different contents → same `object_id` → wrong content resolves | ✅ holds | refs carry explicit digests → both resolve; `object_versions` keeps both rows; matches master |
| 2 | `restore_map` collapse | data-URI + bare-base64 → same ref → one call's value dropped | ✅ holds | the two forms produce **different** refs → distinct keys (diagram ↓) |
| 3 | collision-check refactor | `_find_obj_name_type_collisions` drifts from master `_reject_` | ✅ holds | identical raise cases (empty / same / single / multi mismatch); only cosmetic reorder of error payload |
| 4 | `expected_digest` mismatch | batch-wide fail on digest recompute in `obj_create_batch` | ✅ holds | both sides compute with `builtin_object_class=None` → can't mismatch |
| 5 | multi-project / auth | cross-project attribution surface | ✅ holds | per-item `ext_to_int_project_id` + per-project grouping, pre-existing |

### Why #2 (the sharp one) is safe

The same bytes embedded two ways carry different metadata (`content_type` / `encoding` / `input_type`), so the object digest — and therefore the ref — diverges. `restore_map` (keyed by ref) never collapses them.

```mermaid
flowchart LR
    P["same PNG bytes"] --> A["data:image/png;base64,…<br>content_type=image/png"]
    P --> B["bare base64<br>content_type=base64"]
    A --> DA["object digest<br>VXbF…"]
    B --> DB["object digest<br>32Cb…"]
    DA --> RA["ref …png:VXbF"]
    DB --> RB["ref …png:32Cb"]
    RA --> M{"restore_map<br>2 distinct keys"}
    RB --> M
    M --> OK["each call keeps<br>its OWN value ✅"]
```

<details>
<summary>Decisive repro — same image, two forms, under a real WB-30574 collision on the shared <code>object_id</code></summary>

```python
# object_id image-37c9.png pre-bound to a different base_object_class,
# then one calls_complete batch embeds the same PNG two ways:
batch = [call_A(inputs={"image": data_uri}),   # data:image/png;base64,iVBORw0KGgo…
         call_B(inputs={"image": bare_b64})]    # iVBORw0KGgo…
trace_server.calls_complete(CallsUpsertCompleteReq(batch=batch))

# observed:
A["image"] == data_uri     # True   ← restored to its OWN form
B["image"] == bare_b64     # True   ← restored to its OWN form
A["image"] == B["image"]   # False  ← no collapse
```

The only path that collapses (two different strings → one ref) needs whitespace-variant base64, which `b64decode(validate=True)` rejects before queuing; identical strings already dedupe at `_visit`. And same ref ⇒ same digest ⇒ same decoded bytes, so a collapse could never yield genuinely-wrong content anyway.
</details>

> [!WARNING]
> **Residual (low severity, no test):** a hard DB error inside `obj_create_batch` fails the whole `calls_complete` batch, whereas master's per-object `try/except` in `_visit` would keep the call and degrade that one object to inline. Only triggerable by an injected insert failure (not a real input) → out of reachability scope. Flagged for awareness, not action.

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -1075,11 +1077,20 @@ def calls_complete(
...
-                    complete_call, self
-                )
+            processed_calls = [
+                process_complete_call_to_content(complete_call, self, pending_objs)
+                for complete_call in req.batch
+            ]
+            # Content objects are written before the calls flush (on with-exit)
+            # so a committed call never references an unwritten object; refs of
+            # collision-skipped objects are swapped back to their raw values.
+            raw_by_ref = self._flush_pending_content_objs(pending_objs)
```
amwarrier: Ordering nit: we write object rows to CH (object_versions + aliases via obj_create_batch inside _flush_pending_content_objs) before files upload to GCS (_flush_file_batches runs on with call_batch() exit). Can we flip? If GCS fails after the object rows commit, the refs point at bytes that never landed — recoverable on retry but a live 404 window until then.

gtarpenning: nice find, yeah we should flip

gtarpenning: fixed in [38f67de](https://github.com/wandb/weave/pull/7528/commits/38f67de3439f9c7fa23519f4254e3b69057c0626)

amwarrier (approved):

PR #7531, opened 2026-07-07
feat(weave): server-side ingest sampling for the spans model

ro31337:
## Explainer to save your time

https://github.com/user-attachments/assets/79476387-3c8d-4a26-8294-8fcb684c08b9

## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-36877

## Description

Weave stores traces in two data models. The calls model already has server-side ingest sampling (#46786): set `WEAVE_INGEST_SAMPLE_RATE=0.1` and the server keeps about 10% of non-eval traces, which cuts storage cost. The *spans* model — the agents / OTel ingest path — had nothing: every span sent to it was stored. This PR closes that gap.

Same knob, and pretty much works by same rules. The two existing env vars now govern both models. The verdict is the same hash of the trace id, so a trace is kept or dropped as a whole. If any span of a trace carries a `weave.eval.*` attribute, the whole trace is kept - as it was before in *calls* model - evaluations are too valuable to sample away, and a raw-OTel client can set that attribute to opt into the same protection. A span whose trace id is broken (wrong length, or all zeros) is kept too: when in doubt, keep. One heads-up for operators: a deployment that already runs with a rate below 1.0 starts sampling the spans model when it picks up this change.

The placement is the point. The check runs inside `insert_otel_spans`, right after protobuf parsing and before anything expensive. A dropped trace therefore never writes attachment files, skips extraction entirely, and never reaches the Kafka topic that feeds online scoring — dropped means not stored and not scored. With sampling off (the default) the old single-pass code runs as before; both paths go through the same small parse/extract helpers, so their bookkeeping cannot drift apart. Dropped spans still count as accepted in the response, so clients see an ordinary success. Two fine points: a dropped span that would have failed extraction now counts as accepted (extraction never runs for it), and in sampled requests parse errors come before extraction errors in `error_message`.

Kept traces remember their rate. Every hash-kept span gets `custom_attrs_float["weave.ingest_sample_rate"] = 0.1`, so later anything counting over sampled data can weight each kept span as ten originals. No stamp means the span was kept whole and counts as itself. Dry-run keeps everything and only counts what it would have dropped — and stamps nothing, so the math stays honest. Counters go to DogStatsD (`weave_trace_server.ingest_sampling.spans.*`). No SDK, schema, or API changes.

## Testing

29 tests in `tests/trace_server/test_spans_ingest_sampling.py`. The hash and the env parsing are pinned to the calls sampler's exact behavior. End-to-end against ClickHouse: with the default config nothing changes, down to the exact error messages and their order; at rate 0 a plain trace disappears whole while an eval-marked one survives; the rate stamp lands and a client-supplied fake can't spoof it; a dropped trace leaves no files behind and triggers no scoring event; a span that fails extraction is rejected with the same error as before. A full local `tests/trace_server/` run matches master exactly (the same 5 pre-existing order-dependent failures on both, unrelated); ruff, format, and the repo-pinned mypy are clean.

### DIFF
--- a/tests/trace_server/test_spans_ingest_sampling.py
+++ b/tests/trace_server/test_spans_ingest_sampling.py
@@ -232,10 +232,12 @@ def test_unusable_trace_ids_fail_open_in_decisions(
         assert sampler_metrics["parse_failures"] == [1]
 
 
-# --- decision ladder (parsed spans, no backend) -----------------------------
+# --- per-span verdicts (parsed spans, no backend) ----------------------------
 
 
-def test_decide_spans_ladder_at_rate_zero(sampler_metrics: dict[str, list]) -> None:
+def test_decide_spans_all_verdicts_at_rate_zero(
+    sampler_metrics: dict[str, list],
+) -> None:
     config = ingest_sampling.SamplingConfig(rate=0.0, dry_run=False)
     eval_tid = (1).to_bytes(16, "big")
     plain_tid = (2).to_bytes(16, "big")
--- a/tests/trace_server/test_spans_ingest_sampling.py
+++ b/tests/trace_server/test_spans_ingest_sampling.py
@@ -29,6 +29,10 @@ from weave.trace_server.agents.types import (
     GenAIOTelExportReq,
 )
 from weave.trace_server.base64_content_conversion import AUTO_CONVERSION_MIN_SIZE
+from weave.trace_server.opentelemetry.helpers import (
+    AttributePathConflictError,
+    unflatten_key_values,
+)
 from weave.trace_server.opentelemetry.python_spans import Resource as PyResource
 from weave.trace_server.opentelemetry.python_spans import Span as PySpan
 from weave.trace_server.opentelemetry.python_spans import StatusCode
@@ -116,6 +120,13 @@ def _parse(span: PbSpan) -> PySpan:
     return PySpan.from_proto(span, PyResource.from_proto(PbResource()))
 
 
+def _conflict_error(span: PbSpan) -> str:
+    """The exact parse-failure message the server records for this span."""
+    with pytest.raises(AttributePathConflictError) as excinfo:
+        unflatten_key_values(span.attributes)
+    return str(excinfo.value)
+
+
 def _tid_with_verdict(rate: float, *, kept: bool, start: int = 1) -> bytes:
     """First 16-byte trace id (from ``start`` upward) whose verdict matches."""
     for i in range(start, start + 10_000):
@@ -154,7 +165,8 @@ def sampler_metrics(monkeypatch: pytest.MonkeyPatch) -> dict[str, list]:
 
 def test_keep_by_hash_matches_inline_sha256() -> None:
     # Pin the exact formula: the calls-model sampler uses the same one, and a
-    # trace living in both data models must get one verdict.
+    # trace living in both data models must get one verdict. Purity also gives
+    # cross-request determinism: the verdict depends on nothing but the string.
     for i in range(1, 50):
         tid = i.to_bytes(16, "big").hex()
         for rate in (0.0, 0.1, 0.5, 0.9, 1.0):
@@ -193,17 +205,29 @@ def test_sample_rate_env_falls_back_to_off_when_invalid(
 
 
 @pytest.mark.parametrize(
-    ("trace_id", "usable"),
+    ("trace_id_bytes", "usable"),
     [
-        ("a" * 32, True),
-        ("", False),  # empty proto bytes
-        ("a" * 30, False),  # 15-byte id
-        ("a" * 34, False),  # 17-byte id
-        ("0" * 32, False),  # OTel-invalid zero id
+        (b"\xaa" * 16, True),
+        (b"", False),  # empty proto bytes
+        (b"\xab" * 15, False),  # 15-byte id
+        (b"\xab" * 17, False),  # 17-byte id
+        (b"\x00" * 16, False),  # OTel-invalid zero id
     ],
 )
-def test_usable_trace_id(trace_id: str, usable: bool) -> None:
-    assert ingest_sampling._usable_trace_id(trace_id) is usable
+def test_unusable_trace_ids_fail_open_in_decisions(
+    sampler_metrics: dict[str, list], trace_id_bytes: bytes, usable: bool
+) -> None:
+    # At rate 0.0 a usable id is hash-dropped; an unusable one is kept
+    # fail-open and counted as a parse failure.
+    config = ingest_sampling.SamplingConfig(rate=0.0, dry_run=False)
+    span = _parse(_proto_span(trace_id_bytes, b"\x0a" * 8))
+    decisions = ingest_sampling.decide_spans(config, [span], [5])
+    if usable:
+        assert decisions == [ingest_sampling.SpanDecision(drop=True)]
+        assert sampler_metrics["parse_failures"] == []
+    else:
+        assert decisions == [ingest_sampling.SpanDecision(drop=False)]
+        assert sampler_metrics["parse_failures"] == [1]
 
 
 # --- decision ladder (parsed spans, no backend) -----------------------------
@@ -227,9 +251,10 @@ def test_decide_spans_ladder_at_rate_zero(sampler_metrics: dict[str, list]) -> N
         ingest_sampling.SpanDecision(drop=False),
         ingest_sampling.SpanDecision(drop=True),
     ]
-    assert sampler_metrics["seen"] == [1, 1, 1, 1]
+    # Counters are per span but emitted once per request (aggregated).
+    assert sampler_metrics["seen"] == [4]
     assert sampler_metrics["parse_failures"] == [1]
-    assert sampler_metrics["evals_kept"] == [1, 1]
+    assert sampler_metrics["evals_kept"] == [2]
     assert sampler_metrics["dropped"] == [(1, 40, False)]
 
 
@@ -281,29 +306,42 @@ def test_sampler_off_is_inert(
     monkeypatch: pytest.MonkeyPatch,
     rate_env: str | None,
 ) -> None:
-    """The default config must not change ingest behavior at all."""
+    """The default config must not change ingest behavior at all.
+
+    Pins the exact error_message (content and per-span order) on a request
+    mixing good spans with two distinct parse failures. Extraction failures
+    share the same extract_row helper in both branches, but no real OTel
+    payload can make the defensively-guarded extractor raise, so that branch
+    is covered by construction rather than by a fixture.
+    """
     if rate_env is None:
         monkeypatch.delenv("WEAVE_INGEST_SAMPLE_RATE", raising=False)
     else:
         monkeypatch.setenv("WEAVE_INGEST_SAMPLE_RATE", rate_env)
     project_id = make_project_id("spans_sampling_off")
     tid = (10).to_bytes(16, "big")
-    conflict = _proto_span(
+    conflict_a = _proto_span(
         (11).to_bytes(16, "big"),
         b"\x13" * 8,
         # "x" as a leaf and "x.y" under it can't both exist -> parse failure.
         attrs={"x": "leaf", "x.y": "nested"},
     )
+    conflict_b = _proto_span(
+        (12).to_bytes(16, "big"), b"\x14" * 8, attrs={"y": "leaf", "y.z": "nested"}
+    )
     res = _export(
         ch_server,
         project_id,
         _proto_span(tid, b"\x11" * 8),
+        conflict_a,
         _proto_span(tid, b"\x12" * 8, parent_span_id=b"\x11" * 8),
-        conflict,
+        conflict_b,
     )
     assert res.accepted_spans == 2
-    assert res.rejected_spans == 1
-    assert res.error_message != ""
+    assert res.rejected_spans == 2
+    assert res.error_message == "; ".join(
+        _conflict_error(span) for span in (conflict_a, conflict_b)
+    )
     assert len(_stored_spans(ch_server, project_id)) == 2
     # Disabled sampler emits nothing at all.
     assert sampler_metrics == {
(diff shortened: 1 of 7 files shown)


weave/trace_server/environment.py
```diff
@@ -254,6 +254,33 @@ def wf_scoring_worker_remote_scorer_require_structured_result_schema() -> bool:
     )
 
 
+# Ingest sampling (spans model) -- shared knobs with the calls-model sampler
```
mscavezze-cw: This comment doesn't quite line up with the function name. The comment is a little confusing: I'm not sure if it's telling me that this is the sampling for the spans model (since it leads with that), or that it's a shared knob (the latter part of the comment suggests that maybe the calls model is piggy-backing off of a spans tunable). The function name suggests that it is global and thus spans/calls agnostic.

weave/trace_server/environment.py
```diff
@@ -254,6 +254,33 @@ def wf_scoring_worker_remote_scorer_require_structured_result_schema() -> bool:
...
+
+
+def wf_ingest_sample_rate() -> float:
+    """Fraction of non-eval traces to keep at OTel-spans ingest (0.0-1.0).
+
+    The same environment variable also governs the calls-model sampler in the
+    trace-server service, so one operator knob covers both data models.
+    Defaults to 1.0 (sampling off); an unparseable or out-of-range value falls
+    back to 1.0, so a bad config can never start dropping data. Change the
+    rate in a quiet traffic window: an in-flight trace hashed at two different
```
mscavezze-cw: Is changing the rate sufficient, or does the worker need to be restarted?

weave/trace_server/datadog.py
```diff
@@ -225,6 +225,17 @@ def record_db_insert(*, table: str, count: int, path: str | None = None) -> None
     )
 
 
+def emit_counter(metric: str, value: int, tags: list[str]) -> None:
+    """Emit a best-effort counter with the unified service tags appended.
+
+    For counter families owned by other modules (e.g. the spans ingest
+    sampler); `record_db_insert` stays the dedicated DB-insert emitter.
```
mscavezze-cw: I'm not totally sure what this second part of the comment means, why it's here, or what action I should take in response. Is this basically saying `use emit_counter for calls metrics, but use record_db_insert for agent spans`?

Why can't record_db_insert be shared for both? Perhaps if one was be implemented by calling the other, the differences would be more obvious.

weave/trace_server/agents/ingest_sampling_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+two families deliberately stay separate -- these count *spans* while those
+count ingest messages, so summing them would be meaningless. Every emit is
+best-effort: a metrics failure never breaks ingest.
+"""
+
+from __future__ import annotations
+
+from weave.trace_server.datadog import emit_counter
+
+_METRIC_SEEN = "weave_trace_server.ingest_sampling.spans.seen"
```
mscavezze-cw: Consider extracting the prefix `weave_trace_server.ingest_sampling.spans` to avoid typos and potentially improve readability. Low pri.

weave/trace_server/agents/ingest_sampling_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+
+def evals_kept(count: int) -> None:
+    """Count spans kept by the eval carve-out."""
+    emit_counter(_METRIC_EVALS_KEPT, count, [_ROUTE_TAG])
+
+
+def dropped(count: int, byte_size: int, dry_run: bool) -> None:
+    """Count dropped spans and their serialized protobuf size.
+
+    In dry-run the tag marks a would-drop, not a real one.
```
mscavezze-cw: It seems like we've taken two different stances for how we record metrics during dry-run. In this case, we write the hypothetical sample rate to the same metric that we normally do. When recording the sample rate in the call or span, we record the actual sample rate.

Consider making the behavior match, or writing to a different metric name to avoid confusion when reviewing metrics.

weave/trace_server/agents/ingest_sampling.py
```diff
@@ -0,0 +1,174 @@
...
+    from weave.trace_server.opentelemetry.python_spans import Span
+
+# Attribute key the applied sample rate is stamped under on kept spans. The
+# flat dotted form is how custom attributes are stored on span rows; the calls
+# model records the same fact as ``attributes["weave"]["ingest_sample_rate"]``.
+INGEST_SAMPLE_RATE_ATTR = "weave.ingest_sample_rate"
+
+# sha256(trace_id) is reduced modulo this many buckets; a trace is kept when
+# its bucket falls below rate * buckets. Must stay byte-identical to the
+# calls-model sampler so a mixed trace gets one verdict across both models.
```
mscavezze-cw: If this must match, could you move the constant to a shared location and have both samplers read from the same source?

weave/trace_server/agents/ingest_sampling.py
```diff
@@ -0,0 +1,174 @@
...
+_HASH_BUCKETS = 1_000_000
+
+_ZERO_TRACE_ID = "0" * 32
```
mscavezze-cw: Consider adding a comment about the purpose of this.

weave/trace_server/agents/ingest_sampling.py
```diff
@@ -0,0 +1,174 @@
...
+
+    @property
+    def enabled(self) -> bool:
+        # rate >= 1.0 keeps everything; treat as fully off so the ingest path
+        # runs its unmodified single-pass loop and emits no metrics.
+        return self.rate < 1.0
+
+
+def request_config() -> SamplingConfig:
+    """Read the sampler config from the environment, once per request."""
```
mscavezze-cw, line 71: Instead of reading once per request, would it make sense to cache this for a short time, maybe 60 seconds?

weave/trace_server/agents/ingest_sampling.py
```diff
@@ -0,0 +1,174 @@
...
+    return SamplingConfig(
+        rate=wf_env.wf_ingest_sample_rate(),
+        dry_run=wf_env.wf_ingest_sample_dry_run(),
+    )
+
+
+def keep_by_hash(trace_id: str, rate: float) -> bool:
+    """Deterministic per-trace keep decision; True means keep.
+
+    Pure function of (trace_id, rate), byte-identical to the calls-model
```
mscavezze-cw, line 81: Instead of duplicating code, would it be better to share the code? There would be less risk of drift, and fixes/updates would only have to be made in one place.

weave/trace_server/agents/ingest_sampling.py
```diff
@@ -0,0 +1,174 @@
...
+    """Whether a trace_id can key the hash.
+
+    OTel requires a 16-byte, not-all-zero trace id. ``Span.from_proto``
+    hexlifies whatever bytes arrive, so a malformed id shows up here as a
+    wrong-length string and the spec-invalid zero id as 32 zeros. Those are
+    sentinels shared across broken clients -- hashing them would hand every
+    such client one collective verdict -- so they fail open instead (kept,
+    counted as parse failures).
+    """
+    return len(trace_id) == 32 and trace_id != _ZERO_TRACE_ID
```
mscavezze-cw: Consider naming `_ZERO_TRACE_ID` based on its purpose (sentinel) instead of what it is (zeros)

weave/trace_server/agents/ingest_sampling.py
mscavezze-cw: If there are things that are specific to calls or spans, maybe those could go in a derived class.

weave/trace_server/agents/ingest_sampling.py
```diff
@@ -0,0 +1,174 @@
...
+
+    The verdict is per trace: a trace is kept whole when any of its spans in
+    this request carries a ``weave.eval.*`` marker, otherwise the trace hash
+    decides. Spans whose trace_id is unusable are kept individually
+    (fail-open). ``byte_sizes`` -- serialized protobuf sizes aligned with
+    ``spans`` -- feed the dropped-bytes counter. Metrics count spans, not
+    traces, and are emitted once per call.
+    """
+    eval_traces: set[str] = set()
+    for span in spans:
```
mscavezze-cw, line 136: Why have two loops through spans? Why not just move this check directly into the loop below and set a bool instead of constructing a set?

tests/trace_server/test_spans_ingest_sampling.py
```diff
@@ -0,0 +1,597 @@
...
+        lambda c, b, d: calls["dropped"].append((c, b, d)),
+    )
+    return calls
+
+
+# --- pure functions ---------------------------------------------------------
+
+
+def test_keep_by_hash_matches_inline_sha256() -> None:
+    # Pin the exact formula: the calls-model sampler uses the same one, and a
```
mscavezze-cw, line 167: Consider adding a test in core that runs both formulas and verifies that they produce the same results.

tests/trace_server/test_spans_ingest_sampling.py
```diff
@@ -0,0 +1,597 @@
...
+        assert sampler_metrics["dropped"] == [(1, 5, False)]
+    else:
+        assert decisions == [ingest_sampling.SpanDecision(drop=False)]
+        assert sampler_metrics["parse_failures"] == [1]
+
+
+# --- decision ladder (parsed spans, no backend) -----------------------------
+
+
+def test_decide_spans_ladder_at_rate_zero(sampler_metrics: dict[str, list]) -> None:
```
mscavezze-cw: This name could be more self-documentating. What does `ladder` mean here?

tests/trace_server/test_spans_ingest_sampling.py
```diff
@@ -0,0 +1,597 @@
...
+        _proto_span(tid, sid, attrs={"gen_ai.output.messages": '[{"parts": 5}]'}),
+    )
+    assert res.accepted_spans == 0
+    assert res.rejected_spans == 1
+    assert res.error_message == f"Extraction failed for span {sid.hex()}: TypeError"
+    assert _stored_spans(ch_server, project_id) == []
+
+
+@pytest.mark.flaky(reruns=3)
+def test_unusable_trace_ids_fail_open(
```
mscavezze-cw, line 507: Are these `unusable` trace ids unusable in every part of the system? Or could someone who wanted to evade sampling simply set some `unusable` trace ids and still keep using the rest of Weave?

mscavezze-cw (approved):

weave/trace_server/environment.py
ro31337: FIXED: reworded the section comment so it no longer leads with "spans model"

weave/trace_server/environment.py
ro31337: FIXED: answered this in the docstring. The rate comes from the process environment and isn't mutated in-process, so changing it means a redeploy/restart with the new value, it's not like a live knob.

weave/trace_server/datadog.py
ro31337: FIXEd: `record_db_insert` now calls `emit_counter`, so the relationship is explicit (your suggestion). I also dropped the confusing second sentence from `emit_counter` docstring (it's just the low-level counter emitter now)

weave/trace_server/agents/ingest_sampling_metrics.py
ro31337: also FIXED: extracted the prefix into a `_PREFIX` constant

weave/trace_server/agents/ingest_sampling_metrics.py
ro31337: WON'T FIX (this is deliberate): in dry-run the metric must show the would-drops - that's the whole point of shadow mode - while the stamp must stay silent. Because nothing was actually sampled to weight. The same metric name mirrors the calls sampler. I added a docstring note that dashboards should split it by the dry_run tag

weave/trace_server/agents/ingest_sampling.py
ro31337: leaving for the follow-up, but would agree here. The single source has to be cross-repo (core already imports weave), so it can't land in this weave-only PR probably. It goes in the sampler-unification follow-up. For now the comment points at the core copy it must match

weave/trace_server/agents/ingest_sampling.py
ro31337: added a comment explaining the constant

weave/trace_server/agents/ingest_sampling.py
ro31337: won't fix: it's a dict lookup done once per request, so there's nothing to cache. And since the rate only changes on redeploy (see the rate comment), a TTL wouldn't buy any responsiveness

weave/trace_server/agents/ingest_sampling.py
ro31337: leaving it for follow-up. Agree on the drift risk. The duplicate lives in core, so sharing means a shared helper in weave that core imports, a cross-repo change I can't make in this weave-only PR. Tracking it as a follow-up

weave/trace_server/agents/ingest_sampling.py
ro31337: won't fix and would keep the two passes. The reason for that is a request carries many trace_ids and the eval marker can land on a later span of the same trace, so I have to scan all spans before deciding any, a bool cannot hold per-trace state or keep already-seen spans. I added a comment explaining exactly this

tests/trace_server/test_spans_ingest_sampling.py
ro31337: good idea for the follow-up. A test that runs both formulas has to live in core (that\s the other copy)

tests/trace_server/test_spans_ingest_sampling.py
ro31337: FIXED: renamed the test (as well as the section header) to drop the jargon we don't need

tests/trace_server/test_spans_ingest_sampling.py
ro31337: won't fix for v1: sampling here is cost control, not a security boundary. Evading it means bypassing the SDK. A determined client sending distinct malformed trace ids could keep its data - that's the price of fail-open, since we never drop traffic we can't classify anyway. Adn the parse_failures metric makes a spike in unusable ids visible.

Note for the follow-up: probably hash wrong-length ids and only fail-open the true shared sentinels (empty / all-zero)

ro31337:
@mscavezze-cw thanks for the comments, will leave it open for a bit (until tomorrow) in case there is something else that comes to your mind before we merge

PR #7536, opened 2026-07-08
perf(weave): narrow eval cost pricing to predict calls

ro31337:
## JIRA Issue(s)
https://coreweave.atlassian.net/browse/WB-36670

## Description

This PR does three things:

* Stops computing the LLM judge/scorer costs that eval results throws away — this was the original ask.
* Stops dragging the scorers' big rows (inputs and outputs) through the cost SQL for nothing.
* Fixes a predict call's `genai_span_ref` getting dropped on ClickHouse when costs were requested.

Below is how this was found and why we fix it.

The eval results API shows the cost of the model's `predict()` calls only, not the LLM judge. That number is already correct.

While reviewing #7487, a reviewer pointed out we do extra work to get it. To read the predict cost, the server priced *every* child of a trial — the predict call and all the scorers — and then used only the predict part. The judge prices were computed and discarded (https://github.com/wandb/weave/pull/7487#discussion_r3515543189).

I looked closer, and it is worse than just the judge math. The cost step also carries each child's whole row — inputs and outputs, which can be kilobytes — through its heavy SQL. So even plain code scorers, which have nothing to price, had their big rows hauled through for nothing. On large evals that adds up.

The fix keeps scorers out of the cost step entirely. First we fetch the trial's children with no costs. Then, only when costs are asked for, a small second query prices just the predict calls and fills in each trial's cost.

While doing this I hit a second bug. On ClickHouse, asking for costs dropped the predict call's `genai_span_ref`. The cost SQL rebuilds the summary and ends up writing the `weave` key twice, so the last copy wins and the ref is lost. Now that the predict call is read outside that step, the ref survives — which is what the in-memory backend already did.

One note on speed: this trades one query for two, so the saving grows with eval size. Large evals stop hauling all those scorer rows; small evals come out about the same.

## Testing

The five existing cost tests still pass on both backends — they lock in the rules that must not change (predict-only, opt-in, judge excluded, None when unpriced). Three new tests cover the new paths: per-trial cost on the rows path, cost on the sorted path (which had no cost coverage before), and a regression test for the `genai_span_ref` fix. Full `tests/trace_server` run locally on ClickHouse; ty, pyright, mypy, and ruff are clean.

### DIFF
--- a/tests/trace_server/test_trace_server_evaluation_apis.py
+++ b/tests/trace_server/test_trace_server_evaluation_apis.py
@@ -1273,6 +1273,200 @@ def test_eval_results_summary_predict_total_cost_none_when_absent(client):
     assert res.summary.evaluations[0].predict_total_cost is None
 
 
+def test_eval_results_rows_trial_total_cost_predict_only(client):
+    """Per-trial total_cost in rows: predict-only, filled without a summary.
+
+    include_summary=False pins that the cost fill is not tied to the summary
+    path. The judge's (differently priced) usage must not leak into the trial,
+    and a trial without a predict call reports None, not the judge's cost.
+    """
+    project_id = client.project_id
+    client.server.cost_create(
+        CostCreateReq(
+            project_id=project_id,
+            costs={"predict-llm": _PREDICT_PRICE, "judge-llm": _JUDGE_PRICE},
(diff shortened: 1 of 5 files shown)


jwlee64 (approved):
Awesome, this cost stuff is gnarly, but this is awesome, thanks for this

PR #7537, opened 2026-07-08
perf(weave): skip latest alias write for content objects in calls_complete

gtarpenning:
## Summary
- Content refs pin the version digest (`base64_content_conversion._content_ref`), so content objects are never resolved via the `latest` alias.
- `_flush_content_objs` now calls `obj_create_batch(..., write_latest_alias=False)`, dropping one CH insert per `calls_complete` batch on the ingest hot path.
- Read-transparent: the objects query builder's `is_latest` has a computed most-recent-version fallback, so `obj_read("latest")` and `latest_only` still resolve the sole version without an alias row.

## Testing
Updated `test_calls_complete_batches_content_object_inserts` to assert zero alias inserts/rows and that `latest` still resolves to the version via the computed fallback.

## QA validation (zoo-qa)
Deployed on multi-qa (zoo-qa) via core#47636, verified live via `/version`; A/B vs master.

| master `8400594` (2.78s, 20 SQL queries) | PR `4840b6e` (1.78s, 14 SQL queries) |
|---|---|
| master | pr |

Server-side Datadog spans, per `calls/complete` batch embedding content objects, grouped by build version:

| build | `aliases` | `object_versions` | `files` | full-batch traces |
|---|---|---|---|---|
| master `8400594` | 1 | 1 | 1 | 3/4 |
| PR `4840b6e` | 0 | 1 | 1 | 7/8 |

PR drops the `aliases` INSERT to 0 while `object_versions`/`files` still fire (object write path intact, only the latest-alias write skipped; the 1 bare trace on each build is a content-ref cache hit). Endpoint `@duration` P50 1.87s -> 1.76s; no new error classes on the PR build. Traces: [master](https://us5.datadoghq.com/apm/trace/257aa411722c62d09f222bfc60e10f07) / [PR](https://us5.datadoghq.com/apm/trace/51143537976376d0c903fca880328a04).

**Frontend render check (QA weave UI):** content objects written without an alias row still render and resolve `latest` via the computed fallback.

| object renders in the call view | Versions tab: empty `Aliases`, still resolves |
|---|---|
| object render | versions |

### DIFF
--- a/tests/trace_server/test_calls_complete.py
+++ b/tests/trace_server/test_calls_complete.py
@@ -661,7 +661,8 @@ def test_calls_complete_batches_content_object_inserts(
 
     Duplicate content across calls in the batch collapses to the same ref,
     and the whole batch issues exactly one insert into object_versions and
-    one into aliases instead of one pair of inserts per object.
+    none into aliases: content refs pin the version digest, so the "latest"
+    alias is never written or read for these objects.
     """
     project_id = f"{TEST_ENTITY}/calls_complete_batched_objects"
     internal_project_id = b64(project_id)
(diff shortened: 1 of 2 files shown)


rgao-coreweave (approved):
👍Great optimization, thakns!

PR #7539, opened 2026-07-08
fix(weave): stabilize client-side digests across non-string dict keys

gtarpenning:
## Summary
- `/obj/create` returned 409 `DigestMismatchError` when an object (e.g. a scorer with an int-keyed label map) carried non-string dict keys.
- Client hashed the in-memory val; the server hashed it after a JSON round-trip, which stringifies keys before `sort_keys`. Keys whose numeric vs lexicographic order differs (`{1,2,10}`) diverged.
- Fix: hash over `canonical_json` (a JSON round-trip) so both sides agree. Plain string-keyed digests are byte-for-byte unchanged, so no stored digest changes.
- Jira: [WB-37011](https://coreweave.atlassian.net/browse/WB-37011)

## Testing
adds red-then-green digest round-trip invariance tests + a server-level obj_create repro (int keys).

[WB-37011]: https://coreweave.atlassian.net/browse/WB-37011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/shared/digest.py
+++ b/weave/shared/digest.py
@@ -36,6 +36,17 @@ def str_digest(json_val: str) -> str:
     return bytes_digest(json_val.encode())
 
 
+def canonical_json(val: Any) -> str:
+    """Serialize val the way the server sees it after transport.
+
+    The client hashes an in-memory value while the server hashes the same
+    value after it round-trips through JSON over the wire, which coerces
+    non-string dict keys to strings before sort_keys orders them. Round-tripping
+    here keeps both sides byte-identical regardless of key types.
+    """
+    return json.dumps(json.loads(json.dumps(val)), sort_keys=True)
+
+
 @dataclass(slots=True, frozen=True)
 class ObjectDigestResult:
     """Deterministic digest data for object-create style payloads."""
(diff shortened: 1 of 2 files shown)


weave/shared/digest.py
```diff
@@ -36,6 +36,17 @@ def str_digest(json_val: str) -> str:
     return bytes_digest(json_val.encode())
 
 
+def canonical_json(val: Any) -> str:
```
andrewtruong, line 39: hmm this seems expensive.  Why do we need to do this?  Is there a better way?

gtarpenning: i asked claude the same thing and apparently no

gtarpenning: 🤖 good catch. this helper is shared, and the server hits it per-row on `table_create` / per-object on `obj_create`, where keys are already strings post-parse so the round-trip is pure waste there (~3x the JSON CPU on ingest). gated it on an actual non-string key in b02977861c: all-string payloads (every server call, ~all client payloads) take a single `json.dumps(sort_keys=True)`, byte-identical to the stored digests; the round-trip only fires for the rare non-string-key case that caused the 409.

andrewtruong (approved):

PR #7328, opened 2026-06-19
fix(weave): capture Bedrock Converse cache tokens on the streaming path

kimnamu:
## Summary

Thanks again for the Bedrock integration! This makes prompt-caching token counts on the Converse API actually get recorded.

Closes #7326.

Bedrock's Converse / ConverseStream `usage` object reports prompt-caching tokens as `cacheReadInputTokens` / `cacheWriteInputTokens` (confirmed against the `bedrock-runtime` botocore service model and the AWS [prompt-caching docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html)). Two things prevented them from being recorded:

1. `bedrock_on_finish_converse` read `cacheReadInputTokenCount` / `cacheWriteInputTokenCount` (a `*Count` suffix the API does not return), so `.get(...)` always returned `None`.
2. The streaming accumulator (`bedrock_stream_accumulator`) copied only `inputTokens` / `outputTokens` / `totalTokens` from the `metadata` event, so cache tokens never reached the shared finish handler on the `converse_stream` path.

The streaming accumulator already carries the other usage fields through to the shared finish handler — this just adds the two cache fields it was missing, matching the non-streaming path.

## Before / After

| Path | Cache tokens in call summary (Before) | After |
|------|----------------------------------------|-------|
| `converse` (non-streaming) | ❌ never (reads non-existent `*Count` key) | ✅ `cache_read_input_tokens` / `cache_creation_input_tokens` |
| `converse_stream` (streaming) | ❌ never (accumulator drops them) | ✅ recorded |
| `inputTokens` / `outputTokens` / `totalTokens` | ✅ | unchanged ✅ |
| Non-cached calls (no cache fields) | no cache keys in summary | unchanged ✅ (keys omitted, not zeroed) |

## Changes

- `weave/integrations/bedrock/bedrock_sdk.py`:
  - Finish handler reads `cacheReadInputTokens` / `cacheWriteInputTokens` (the service-model field names).
  - `bedrock_stream_accumulator` carries those two fields from the `metadata` event into the accumulated usage, only when present (so non-cached streams are unaffected).
- `tests/integrations/bedrock/bedrock_test.py`: streaming regression test that feeds a `metadata` event with cache tokens and asserts they reach the call summary.

## Test (before / after the fix)

Reverting only the source fix (keeping the new test) makes it fail — the streamed cache tokens never reach the summary:

```
[...]

PR #6516, opened 2026-03-31
feat(weave): Update integrations to support cached token

andrewtruong:
https://coreweave.atlassian.net/browse/WB-32634

Updates integrations to start tracking cached tokens

---

<a href="https://app.devin.ai/review/wandb/weave/pull/6516" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #7540, opened 2026-07-08
fix(weave): use real Bedrock cache token field names

### LINKED ISSUE #7326: Bedrock Converse prompt-caching tokens are never recorded (wrong key spelling + streaming accumulator drops them)
## Description

Thanks again for the Bedrock integration. This is about prompt-caching token counts on the Converse API.

Bedrock's Converse / ConverseStream `usage` object reports prompt-caching tokens. The authoritative field names from the `bedrock-runtime` botocore service model are `cacheReadInputTokens` and `cacheWriteInputTokens`:

```
$ python -c "import botocore.session as s; m=s.get_session().get_service_model('bedrock-runtime'); print(list(m.shape_for('TokenUsage').members.keys()))"
['inputTokens', 'outputTokens', 'totalTokens', 'cacheReadInputTokens', 'cacheWriteInputTokens']
```

This matches the AWS user guide ([Prompt caching](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html)):

> The `cacheReadInputTokens` and `cacheWriteInputTokens` values tell you how many tokens were read from the cache and how many tokens were written to the cache...

Two things currently prevent these tokens from being recorded:

**1. Wrong key spelling in the finish handler.** `bedrock_on_finish_converse` reads `cacheReadInputTokenCount` / `cacheWriteInputTokenCount` (a `*Count` suffix that the API does not return), so the `.get(...)` always returns `None`:

https://github.com/wandb/weave/blob/master/weave/integrations/bedrock/bedrock_sdk.py#L36-L42

**2. Streaming accumulator drops them.** For `converse_stream`, the shared finish handler reads from the accumulated dict, but `bedrock_stream_accumulator` copies only `inputTokens` / `outputTokens` / `totalTokens` from the `metadata` event and never the cache token fields:

https://github.com/wandb/weave/blob/master/weave/integrations/bedrock/bedrock_sdk.py#L413-L419

## Before / After

| Path | Cache tokens in call summary (Before) | After |
|------|----------------------------------------|-------|
| `converse` (non-streaming) | ❌ never (reads non-existent `*Count` key) | ✅ `cache_read_input_tokens` / `cache_creation_input_tokens` |
| `converse_stream` (streaming) | ❌ never (accumulator drops them) | ✅ recorded |
| `inputTokens` / `outputTokens` / `totalTokens` | ✅ | unchanged ✅ |
| Non-cached calls (no cache fields) | summary has no cache keys | unchanged ✅ (keys omitted, not zero) |

## Proposed fix

Use the field names from the botocore service model (`cacheReadInputTokens` / `cacheWriteInputTokens`) in the finish handler, and carry them through `bedrock_stream_accumulator` so both `converse` and `converse_stream` record them. I have a fix + regression test (asserting the streamed cache tokens reach the call summary) ready and will open a PR.

## System info

- weave `master` (`weave/integrations/bedrock/bedrock_sdk.py`)
- botocore `1.41.3`; field names verified against the `bedrock-runtime` service model + AWS docs.

ro31337:
## Description

- Fixes [WB-36951](https://coreweave.atlassian.net/browse/WB-36951)
- Fixes #7326

Bedrock reports cache token counts as `cacheReadInputTokens` / `cacheWriteInputTokens`. Our integration read `cacheReadInputTokenCount` / `cacheWriteInputTokenCount` — fields that don't exist. So cache tokens were never recorded, on both the `converse` and `converse_stream` paths.

The tests didn't catch it because the mocks used the same wrong names. That's how it shipped twice (#6516, then #7525). @kimnamu diagnosed this and opened #7328 with the right names — thank you! His PR now conflicts with master, so this one re-lands the rename (first commit co-authored) and fixes the mocks.

One more thing had to change: `prompt_tokens`. Bedrock's `inputTokens` does not include cache tokens ([AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html)). Our cost math subtracts cache counts from `prompt_tokens`, so it expects the total. A plain rename would make it subtract tokens that were never there. Fix: `prompt_tokens = inputTokens + cache read + cache write`. litellm maps Bedrock usage the same way. Calls without caching are not affected.

## Testing

All 15 bedrock tests pass, on both the in-memory and ClickHouse backends. I also checked the tests can actually catch the bug now: revert the source fix, keep the tests — the three cache tests fail; restore the fix — they pass. There is also a new test that checks mock usage keys against the botocore service model, so a made-up field name can't go green again.

[WB-36951]: https://coreweave.atlassian.net/browse/WB-36951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integrations/bedrock/bedrock_test.py
+++ b/tests/integrations/bedrock/bedrock_test.py
@@ -720,7 +720,9 @@ def test_bedrock_converse_cache_tokens(
     calls = client.get_calls()
     assert len(calls) == 1
     model_usage = calls[0].summary["usage"][model_id]
-    assert model_usage["prompt_tokens"] == 100
+    # prompt_tokens is gross: inputTokens (100) excludes cached tokens, so the
+    # cache read (80) and cache write (15) counts are added in.
+    assert model_usage["prompt_tokens"] == 195
     assert model_usage["completion_tokens"] == 20
     assert model_usage["total_tokens"] == 120
     assert model_usage["cache_read_input_tokens"] == 80
@@ -751,7 +753,8 @@ def test_bedrock_converse_stream_cache_tokens(
     calls = client.get_calls()
     assert len(calls) == 1
     model_usage = calls[0].summary["usage"][model_id]
-    assert model_usage["prompt_tokens"] == 100
+    # Same gross prompt_tokens math as the non-streaming cache test above.
+    assert model_usage["prompt_tokens"] == 195
     assert model_usage["completion_tokens"] == 20
     assert model_usage["total_tokens"] == 120
     assert model_usage["cache_read_input_tokens"] == 80
@@ -783,6 +786,8 @@ def test_bedrock_converse_stream_cache_tokens_zero(
     model_usage = calls[0].summary["usage"][model_id]
     assert model_usage["cache_read_input_tokens"] == 0
     assert model_usage["cache_creation_input_tokens"] == 0
+    # Zero cache counts add nothing: prompt_tokens stays at inputTokens.
+    assert model_usage["prompt_tokens"] == 100
 
 
 @mock_aws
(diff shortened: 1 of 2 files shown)


rgao-coreweave (approved):
thanks for fixing!

PR #7541, opened 2026-07-08
chore(weave_ts): migrate node SDK to OpenTelemetry JS 2.x

ash0ts:
### Why

The published `weave` npm package depends on the OTel **1.x** trace stack
(`@opentelemetry/sdk-trace-base@^1.26.0`, `@opentelemetry/resources@^1.26.0`,
`@opentelemetry/exporter-trace-otlp-proto@^0.53.0`). Host applications that
force-resolve `@opentelemetry/core@2.x` via npm `overrides` crash the SDK at
load time:

```
TypeError: Cannot read properties of undefined (reading 'AlwaysOn')
```

(`TracesSamplerValues` was removed from `@opentelemetry/core` in 2.x, but
sdk-trace-base 1.x still imports it.)

This is not hypothetical: OpenClaw's plugin installer writes managed
dependency overrides — including `@opentelemetry/core: 2.8.0` — into every
installed plugin's package.json, which makes `weave-openclaw` (and any other
OpenClaw plugin that depends on `weave`) crash on gateway startup with the
error above. Reproduced on a fresh OpenClaw install; fixed by this migration.

### What

- `@opentelemetry/sdk-trace-base` / `@opentelemetry/resources`: `^1.26.0` -> `^2.8.0`
- `@opentelemetry/exporter-trace-otlp-proto`: `^0.53.0` -> `^0.219.0`
- `new Resource({...})` -> `resourceFromAttributes({...})` (2.x API)
- Tests migrated for 2.x renames/privatization:
  - `ReadableSpan.parentSpanId` -> `parentSpanContext?.spanId`
  - `BasicTracerProvider.resource` is private now; the resource-attribute
    assertion reaches `_resource`
  - the exporter-internals probe in provider.test.ts follows the new
    otlp-exporter-base 0.2xx chain (`_delegate`, async `headers()`)
- CHANGELOG entry under Unreleased.

No public API changes. `@opentelemetry/api` stays `^1.9.1` (unchanged, 2.x
compatible).

### Validation

- `pnpm typecheck:cjs`, `pnpm typecheck:esm`, `pnpm lint`: clean
- `jest --selectProjects default modern legacy`: 53 suites / 421 tests pass
- End-to-end: packed this branch (`pnpm pack`), overrode `weave` in an
  OpenClaw `weave-openclaw` plugin install (which forces core 2.8.0), ran an
  agent turn through the gateway — plugin loads cleanly and spans land in the
  Weave Agents store (`agents/spans/query`) with a consistent OTel 2.x tree
  (core 2.8.0, sdk-trace-base 2.9.0, exporter 0.219.0).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/sdks/node/CHANGELOG.md
+++ b/sdks/node/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to the Weave TypeScript SDK will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Changed
+
+- Upgrade OpenTelemetry JS SDK dependencies to the 2.x line (`@opentelemetry/sdk-trace-base` / `@opentelemetry/resources` `^2.8.0`, `@opentelemetry/exporter-trace-otlp-proto` `^0.219.0`). Fixes a load-time crash (`Cannot read properties of undefined (reading 'AlwaysOn')`) in host applications that force-resolve `@opentelemetry/core@2.x` via npm overrides (e.g. OpenClaw's managed dependency pins), where the 1.x trace SDK is incompatible with the 2.x core.
+
 ## [0.16.2] - 2026-07-06
 
 ### Added
--- a/sdks/node/package.json
+++ b/sdks/node/package.json
@@ -165,9 +165,9 @@
   },
   "dependencies": {
     "@opentelemetry/api": "^1.9.1",
-    "@opentelemetry/exporter-trace-otlp-proto": "^0.53.0",
-    "@opentelemetry/resources": "^1.26.0",
-    "@opentelemetry/sdk-trace-base": "^1.26.0",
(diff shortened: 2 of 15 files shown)


ro31337 (changes requested):
The description says **"No public API changes"**, but bumping to OpenTelemetry JS 2.x also raises the minimum supported Node.js to `^18.19.0 || >=20.6.0` (per the OTel JS SDK 2.0 release), and `sdks/node/package.json` still doesn't declare an `engines.node` range — so anyone consuming the published `weave` package on Node 16 or an early Node 18 can silently update into an unsupported runtime with no signal from us. Is that raised Node floor intended for consumers of the package? If so, could we add `engines.node: "^18.19.0 || >=20.6.0"` to `sdks/node/package.json` plus a short CHANGELOG line calling out the new minimum, so the "No public API changes" note stays accurate for package consumers?

PR #7542, opened 2026-07-09
fix(weave): drop stalling calls_complete reclaim DELETE

gtarpenning:
## Summary
- The post-soft-delete reclaim `DELETE FROM calls_complete` runs as a lightweight delete on the same rows the soft-delete UPDATE just patched; on CH Cloud it stalls past the 60s `max_execution_time` (code 394) and, being un-guarded, 502s `/calls/delete` even though the soft-delete already hid the rows.
- Drop the inline DELETE; the soft-delete UPDATE now also sets `expire_at = now64(3)` so the table's native `TTL expire_at DELETE` reclaims. Deletes reduce to one non-blocking statement.
- Reclamation is now eventual (rides on the lightweight-update patch materializing into the base part, then TTL); rows are hidden immediately by the `deleted_at` filter. Not a regression: prod currently reclaims nothing since the DELETE always 502s.
- Removes dead `build_calls_complete_delete_query` + `CLICKHOUSE_ASYNC_DELETE_SETTINGS`.

## Testing
functional test (real CH): after delete the call is hidden, `deleted_at` + `expire_at` are both set to the delete instant, sibling keeps its sentinels.

QA'd on zoo-qa (real CH Cloud) via `/qa-this`. Server-side A/B of `/calls/delete`, master vs this build: master's `_delete_calls_complete` fires 2 ClickHouse commands (UPDATE + `DELETE FROM calls_complete`), this build fires 1 (UPDATE only). Both 200, zero code-394 in the window. UI delete of 10 calls succeeds and persists across reload.

[datadog before/after](https://github.com/user-attachments/assets/1ed76557-36cc-4864-840e-6a8a48762436)

[delete via weave UI](https://github.com/user-attachments/assets/140d242e-d24b-495a-adb0-2f04f0233ada)

### DIFF
--- a/tests/trace_server/query_builder/test_calls_query_builder.py
+++ b/tests/trace_server/query_builder/test_calls_query_builder.py
@@ -15,7 +15,6 @@ from weave.trace_server.calls_query_builder.calls_query_builder import (
     _invalid_field_message,
     _is_minimal_filter,
     _maybe_convert_datetime_operands,
-    build_calls_complete_delete_query,
     build_calls_complete_soft_delete_query,
     build_calls_complete_update_end_query,
     build_calls_complete_update_query,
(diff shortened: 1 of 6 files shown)


ro31337 (approved):
Looks good, I like it

PR #7549, opened 2026-07-09
fix(weave): price W&B Inference models under the id usage records

gtarpenning:
## Summary
Every W&B Serverless Inference call (including the LLM-as-a-judge scorers behind agent Signals) shows **no cost**. This fixes the seed so those calls price correctly.

### Why it's broken, step by step
1. **Cost is computed at read time by an exact-match join.** For each call, usage rows are joined to `llm_token_prices` on `usage.llm_id == llm_token_prices.llm_id` (see `token_costs.py::get_ranked_prices`). No matching `llm_id` row -> no price -> the UI renders `-`.
2. **What a call actually records.** The inference service is OpenAI-compatible and echoes the model back as its bare playground id. A real judge call in prod records its usage under:
   ```
   "openai/gpt-oss-20b"
   ```
   Both the scorer path and the playground path strip any `coreweave/` prefix before the request (`clickhouse_trace_server_batched.py`, `chat/completions.py`), so the bare id is always what lands in usage.
3. **What the seed contains.** `fetch_models_begin_costs()` reads the inference catalog (`modelsBegin.json`) and keys every price under a `coreweave/` prefix:
   ```
   modelsBegin idPlayground:  "openai/gpt-oss-20b"
   seeded price key:          "coreweave/openai/gpt-oss-20b"   # $0.05 / $0.20 per 1M
   ```
4. **The mismatch.** The join looks up `openai/gpt-oss-20b` but the only priced row is `coreweave/openai/gpt-oss-20b`. Exact match fails, so cost is null. This hits **all 37** inference models, not just gpt-oss.

### The fix
Emit **both** keys from the same `modelsBegin` entry:
```
"openai/gpt-oss-20b"            -> $0.05 / $0.20 per 1M   # NEW: matches recorded usage
"coreweave/openai/gpt-oss-20b"  -> $0.05 / $0.20 per 1M   # kept, unchanged
```
- The bare id makes recorded usage price correctly.
- The `coreweave/` alias is left in place, so anything ever logged under the prefixed id keeps pricing (no existing match dropped).
- Both come from one catalog entry, so the numbers are identical by construction (no drift). Adds the 37 bare-id rows to `cost_checkpoint.json`.

## Testing
new unit test asserts both keys emit with identical prices + the cents-per-billion conversion; existing `update_costs` suite green.


jwlee64 (approved):
This makes sense to me
I thought we used to do cw-prefix everything, but this is a good fix, this might cause some overlap if some other providers have the oss models under the same slug, but that might just be a gap in our cost system, maybe @jamie-rasmussen for another look

PR #7555, opened 2026-07-10
feat(weave): denormalize scorer_trace_id onto feedback

gtarpenning:
## Summary
Adds one column, `feedback.scorer_trace_id` (migration 039), so agent signals can price a scorer invocation off the judge's own agent span instead of the calls model. Consumer (worker populate + FE) lands in wandb/core#47751.

## Migration 039: why one new column
`scorer_trace_id String DEFAULT ''` on `feedback`.

- **What:** the trace id of the scorer's judge LLM call (the invocation that wrote this feedback row).
- **Why:** signals wants to show what each scorer cost. That cost lives on the judge's agent span, keyed by `trace_id`. A feedback row today points at the *scored turn* (`weave_ref`) and the scorer *wrapper call* (`call_ref`, a calls URI), but never the judge span. This is the missing pointer.
- **Why a typed column, not payload:** it is a join key, not scorer output. Mirrors the `span_*` denormalization (033, 038); `spans` stays the source of truth.
- **Why no cost stored:** cost is priced at read time (`span_costs.py` joins the span's model against `llm_token_prices`). We persist a key, never a dollar figure, so prices can't go stale.
- **Blast radius:** additive, populated going forward (no backfill), no index (resolved against `spans.trace_id`, already bloom-indexed by 030). Down-migration drops the column.

## Testing
feedback round-trip asserts `scorer_trace_id` persists independently of `span_trace_id`.

### DIFF
--- a/tests/trace/test_feedback.py
+++ b/tests/trace/test_feedback.py
@@ -215,6 +215,7 @@ def test_annotation_feedback(client: WeaveClient) -> None:
         "span_status_code": "UNSET",
         "span_conversation_id": "",
         "span_trace_id": "",
+        "scorer_trace_id": "",
     }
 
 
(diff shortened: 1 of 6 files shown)


jtschoonhoven (approved):
LGTM

neutralino1 (approved):

PR #7561, opened 2026-07-11
fix(weave): include Anthropic prompt cache tokens in summary input_tokens

ro31337:
## Description

- Fixes [WB-37043](https://coreweave.atlassian.net/browse/WB-37043)

Anthropic reports prompt tokens in three separate counters: `input_tokens` covers only the uncached part, while cache hits and cache writes arrive as `cache_read_input_tokens` / `cache_creation_input_tokens` ([docs](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#tracking-cache-performance)). Our cost math subtracts the cache counts from the summary's input count, so it expects the gross total there. The Anthropic integration stores the uncached count, and the subtraction removes tokens that were never in it. Cached calls get undercounted, and once a cache read is bigger than the uncached part the input cost goes negative: `input_tokens=50` with a 100k cache read prices as (50 − 100000) × the regular rate.

This is the same producer-side mismatch we fixed for Bedrock in #7540, and the fix follows that convention: an `on_finish` handler rewrites the call summary's `input_tokens` to the gross sum (uncached + cache read + cache creation) at the integration boundary. The raw response payload stays provider-native — `call.output.usage.input_tokens` keeps the uncached count. No trace server changes. Calls without cache activity see no difference, and a call whose summary was rolled up from child calls is left alone.

## Testing

New tests cover sync/async `create`, `create(stream=True)`, the `messages.stream()` context manager, and the beta API. Their cassettes are recorded from the live API: each test makes two calls sharing a cacheable prefix, so the first recorded response carries a real `cache_creation_input_tokens` and the second a real `cache_read_input_tokens`. The assertions then check that the summary carries the gross sum while the raw output keeps the provider's net count. The new recordings use a current model, since `claude-3-haiku-20240307` no longer serves requests; the older tests keep replaying their existing cassettes. The prefix is ~6k tokens because caching needs at least 4k — a shorter one records real zeros.

Also covered: a below-minimum cacheable prompt (real zero cache counts), a stream abandoned before its final message, a traced call nested through an httpx response hook (the child-rollup summary must not be rewritten), and unexpected output shapes.

The full anthropic shard passes on both the in-memory and ClickHouse backends (28 tests), and lint (ruff, mypy, pyright, ty) is clean.

[WB-37043]: https://coreweave.atlassian.net/browse/WB-37043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integrations/anthropic/anthropic_test.py
+++ b/tests/integrations/anthropic/anthropic_test.py
@@ -1,13 +1,19 @@
 import os
 from collections.abc import Generator
+from unittest.mock import Mock
 
 import pytest
 from anthropic import Anthropic, AsyncAnthropic
+from anthropic.types import Message, TextBlock, Usage
 from pydantic import BaseModel
 
 import weave
-from weave.integrations.anthropic.anthropic_sdk import get_anthropic_patcher
+from weave.integrations.anthropic.anthropic_sdk import (
+    anthropic_on_finish,
+    get_anthropic_patcher,
+)
 from weave.integrations.integration_utilities import op_name_from_call
+from weave.trace.call import Call
 
 model = "claude-3-haiku-20240307"
 parse_model = "claude-sonnet-4-5"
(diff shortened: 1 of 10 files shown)


rgao-coreweave (approved):
> The fix normalizes the call summary at the integration boundary: its input_tokens becomes the gross sum.

Good to establish that we fix this in the client instead of doing the math in the backend.

LGTM!

PR #7564, opened 2026-07-13
perf(weave): batch content-object writes on the v1 call/upsert_batch path

gtarpenning:
## Summary
- Extends the calls_complete content-object batching to the v1 `/call/upsert_batch` path. `call_start_batch` now threads one `PendingContentObjs` through the batch: content objects are deduped and written in a single `obj_create_batch` (in parallel with file chunks, before the calls that reference them) instead of one inline synchronous `obj_create` per base64 blob.
- Exposes `call_start_batch` on `ExternalTraceServer` so the HTTP handler (wandb/core#48000) can reach this batched path in one call; without it the handler looped `call_start` per item and never batched.
- Root cause of a 55s prod `upsert_batch` trace: 16 inline blobs = 16 serial `obj_create` round-trips (~1.2s each). This collapses them to one batched insert.
- Converges the two content walks into one `process_call_content`; standalone (non-batch) `call_start`/`call_end` keep inline writes.

## Testing
adds v1 batching/failure-injection/collision tests to test_calls_complete.py (real CH); the adapter path is covered by the retargeted upsert_batch test.

---

### QA A/B (zoo-qa, /qa-this)

Deployed `ee192c21` (weave a950fa7) to zoo-qa, A/B vs master `43f264d` on server-side `@duration` of `POST /call/upsert_batch` (`service:weave-trace env:qa`), identical input (8-call batch, 120 content blobs each).

| build | P50 | MAX | count | errors |
| master `43f264d` | 60.8s | 63.9s | 12 | none |
| PR `ee192c21` | 62.0s | 73.7s | 11 | none |

Parity + zero new error classes = no regression from the concurrent file‖content-obj flush. All requests 200 OK; trace structure unchanged. This workload uses large (>8KB) distinct blobs so it is byte-upload-bound (the intended per-blob `obj_create` round-trip win needs many small blobs to surface), so the A/B validates no-regression rather than the speedup.

| master | PR |
| --- | --- |
| master | pr |

### DIFF
--- a/tests/trace_server/conftest_lib/trace_server_external_adapter.py
+++ b/tests/trace_server/conftest_lib/trace_server_external_adapter.py
@@ -111,6 +111,12 @@ class UserInjectingExternalTraceServer(
         req.start.wb_user_id = self._user_id
         return super().call_start(req)
 
+    def call_start_batch(self, req: tsi.CallCreateBatchReq) -> tsi.CallCreateBatchRes:
+        for item in req.batch:
+            if isinstance(item, tsi.CallBatchStartMode):
+                item.req.start.wb_user_id = self._user_id
+        return super().call_start_batch(req)
+
     def calls_delete(self, req: tsi.CallsDeleteReq) -> tsi.CallsDeleteRes:
         req.wb_user_id = self._user_id
         return super().calls_delete(req)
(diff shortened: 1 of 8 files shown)


jtschoonhoven (approved):
Sic dope

amwarrier (approved):

PR #7570, opened 2026-07-14
fix(weave): trace direct openai calls under otel_v2

gtarpenning:
## Summary
Under the default `use_otel_v2=True`, `weave.init()` did not patch `openai`, so **direct** (non-Agents-SDK) openai calls were silently untraced: no `openai.chat.completions.create` child span, no token usage, `$0` cost. This patches `openai` by default again and suppresses it **only** while the OpenAI Agents OTel processor is active (that processor already emits those calls via `ResponseSpanData`/`GenerationSpanData`), so in-agent calls are not double-logged.

Mirrors the existing `patch_google_adk` -> google-genai suppression: drop the blanket `should_use_otel_v2()` gate in `_dispatch_openai`, have `patch_openai_agents_otel` add `"openai"` to `_SUPPRESSED_INTEGRATIONS`, and order `agents` ahead of `openai` in the implicit-patch mapping so suppression lands first.

Jira: [WB-37240](https://coreweave.atlassian.net/browse/WB-37240)

## Behavior (default `use_otel_v2=True`)

| Scenario | Before | After |
| --- | --- | --- |
| Direct `openai` call inside a `@weave.op` | no child span, no usage/cost | child span + usage + cost |
| Direct `openai` + explicit `patch_openai()` | works (was required) | works (no longer required) |
| LLM call inside the OpenAI Agents SDK (OTel) | single span | single span (still no double-log) |
| `use_otel_v2=False` (legacy path) | works | works (unchanged) |

### What did not work before, now does
```python
import weave
from openai import OpenAI

weave.init("entity/project")   # use_otel_v2=True (the default)

@weave.op
def summarize(text: str) -> str:
    resp = OpenAI().chat.completions.create(
        model="gpt-4o-mini",
        messages=[{"role": "user", "content": text}],
    )
    return resp.choices[0].message.content

summarize("hello")
```
Before: `summarize` is logged, but there is no `openai.chat.completions.create` child span and no token usage or cost. The only way to trace it was to add `weave.integrations.patch_openai()` after `init`. After: the child span is logged with usage and cost rolls up, with no extra call.

### What must keep working, and does
```python
import weave
from agents import Agent, Runner   # OpenAI Agents SDK

weave.init("entity/project")
# run an agent whose model/tools call openai internally
```
The Agents OTel processor already logs each in-agent LLM call. Before and after this PR that stays a single span per call: once `patch_openai_agents_otel` runs it adds `"openai"` to `_SUPPRESSED_INTEGRATIONS`, so direct-openai patching is skipped while the agents processor is live.

## Testing
unit tests pin all three paths: direct openai patched by default without agents, suppressed once the OTel agents processor patches (implicit, explicit, and import-hook), and the legacy calls-based `patch_openai_agents` leaving openai enabled.

<details>
<summary>Background: how the gap arose</summary>

- The `should_use_otel_v2()` gate in `_dispatch_openai` was added in #6917 (OpenAI Agents OTel integration). At that point `use_otel_v2` defaulted to `False`, so the suppression only affected opt-in users. Rationale (from the docstring): the Agents OTel processor already exposes in-agent LLM calls via `ResponseSpanData`/`GenerationSpanData`, so patching `openai` too would double-log them.
- #7048 flipped `use_otel_v2` to `True` by default, which generalized the suppression to all direct-openai users. The workaround (`weave.integrations.patch_openai()`) was documented in the settings docstring, but direct calls were untraced by default from then on.
- This PR keeps the double-log protection but narrows it to exactly the agents-active case, restoring default tracing for direct openai use.

Note: the Claude Agent SDK OTel path (`patch_claude_agent_sdk_otel`) has a similar shape but does not add `"anthropic"` to `_SUPPRESSED_INTEGRATIONS`, and `anthropic` was never gated by `use_otel_v2`. If that processor emits equivalent LLM spans, direct-anthropic calls may double-log when both are active. Out of scope here, flagged for that integration's owner.
</details>

[WB-37240]: https://coreweave.atlassian.net/browse/WB-37240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integrations/test_autopatch.py
+++ b/tests/integrations/test_autopatch.py
@@ -12,6 +12,7 @@ import pytest
 import weave.integrations.patch as patch_module
 from weave.integrations.patch import (
     _PATCHED_INTEGRATIONS,
+    _dispatch_openai,
     _patch_if_needed,
     _patch_integration,
     implicit_patch,
@@ -407,7 +408,7 @@ def test_direct_openai_patched_by_default_without_agents(setup_env, monkeypatch)
     Pins the WB-37240 regression where ``_dispatch_openai`` skipped openai
     entirely in OTel V2 mode, silently dropping spans/usage for direct calls.
     """
-    assert patch_module.INTEGRATION_MODULE_MAPPING["openai"] is patch_openai
+    assert patch_module.INTEGRATION_MODULE_MAPPING["openai"] is _dispatch_openai
     _reset_import(monkeypatch, "agents")
     _inject_fake_module(monkeypatch, "openai")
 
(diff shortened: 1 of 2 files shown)


andrewtruong (commented):
Can you add a test case where the `agents` import is deferred?

```py
import weave

weave.init(...)

from agents import ...  # inside, this imports openai
```

andrewtruong (approved):

PR #7572, opened 2026-07-14
chore(weave): file storage presign_read

gtarpenning:
## Summary
- Jira: [WB-37450](https://coreweave.atlassian.net/browse/WB-37450)
- Stacked on `griffin/mbe-1-contract`. Adds `presign_read(uri, ttl)` to `FileStorageClient` + S3/GCS/Azure impls: a self-signed, read-only, single-object download URL minted with the module's first-party bucket creds (no gorilla hop).
- S3 `generate_presigned_url`, GCS v4 signed URL, Azure read-only blob SAS.

## Testing
moto round-trip: presigned GET returns the object bytes; distinct keys -> distinct urls; ttl honored; out-of-prefix raises. IAM write-denial + out-of-prefix-read denial proven against MinIO in the local de-risk harness.

[WB-37450]: https://coreweave.atlassian.net/browse/WB-37450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/file_storage.py
+++ b/weave/trace_server/file_storage.py
@@ -150,6 +150,16 @@ class FileStorageClient:
         assert isinstance(base_uri, FileStorageURI)
         self.base_uri = base_uri
 
+    def _validate_presign_read_uri(self, uri: FileStorageURI) -> None:
+        """Ensure a presigned URL is only minted for an object below base_uri."""
+        if type(uri) is not type(self.base_uri):
+            raise ValueError(
+                f"Cannot presign {uri.to_uri_str()} with {self.base_uri.to_uri_str()}"
+            )
+        base_prefix = self.base_uri.to_uri_str().rstrip("/") + "/"
+        if not uri.to_uri_str().startswith(base_prefix):
+            raise ValueError(f"Cannot presign {uri.to_uri_str()} outside {base_prefix}")
+
     @abstractmethod
     def store(self, uri: FileStorageURI, data: bytes) -> None:
         """Store data at the specified URI location in cloud storage."""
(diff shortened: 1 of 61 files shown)


weave/trace_server/file_storage.py
```diff
@@ -328,6 +338,15 @@ def read(self, uri: FileStorageURI) -> bytes:
         response = self.client.get_object(Bucket=uri.bucket, Key=uri.path)
         return response["Body"].read()
 
+    def presign_read(self, uri: FileStorageURI, ttl: int) -> str:
+        assert isinstance(uri, S3FileStorageURI)
+        assert uri.to_uri_str().startswith(self.base_uri.to_uri_str())
+        return self.client.generate_presigned_url(
```
amwarrier, line 354: The `startswith` prefix check has a boundary bug: if `base_uri = s3://bucket/exports`, then `s3://bucket/exports-secret/foo` also matches. Not currently exploitable (no bucket has sibling prefixes named this way), but latent — the fix is `startswith(self.base_uri.to_uri_str().rstrip("/") + "/")` or a proper path-segment check. Same pattern repeats in the GCS and Azure implementations below, so worth fixing all three together.

Separately (feel free to defer): using `assert` for a security-relevant check means the guard disappears under `python -O`. Prod pods don't run with `-O` today, but converting these three asserts to `if not ...: raise ValueError(...)` costs nothing and removes the interpreter-flag footgun.

amwarrier (approved):
LGTM overall — the round-trip test proves the S3 path end-to-end and the backend parity is clean. One inline nit worth fixing (latent path-boundary bug in the prefix check) before this compounds up the stack; not a blocker.

amwarrier (approved):

PR #7573, opened 2026-07-14
chore(weave): export target queries

gtarpenning:
## Summary
- Jira: [WB-37450](https://coreweave.atlassian.net/browse/WB-37450)
- Replaces the static export-target registry with named query builders plus minimal target dispatch.
- The `calls` target follows the canonical project table routing: raw `calls_complete` rows for current projects and raw `calls_merged` physical rows for legacy projects. The legacy query only finalizes `display_name`, whose AggregateFunction state cannot be encoded as Parquet; it does not GROUP BY, FINAL, merge, deduplicate, or hide tombstones.
- Calls export every physical column. Objects retain newest-visible-version handling; feedback remains a project-scoped raw select.

## Testing
- Real ClickHouse coverage for complete and legacy projects, cross-project isolation, raw deletion metadata, object visibility, and direct legacy Parquet serialization.
- Exact query-builder assertions, including unknown-target rejection.

[WB-37450]: https://coreweave.atlassian.net/browse/WB-37450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/export_targets.py
+++ b/weave/trace_server/export_targets.py
@@ -1,0 +1,56 @@
+"""Canonical queries for managed-bucket bulk export targets.
+
+Every query is project-scoped with the bound ClickHouse parameter
+``{project_id:String}``. Calls follow the project's canonical storage routing,
+while objects and feedback have one physical source each.
+"""
+
+from weave.trace_server.project_version.types import ReadTable
+
+EXPORT_TARGET_NAMES = frozenset({"calls", "objects", "feedback"})
+
+
+def build_export_query(target_name: str, calls_read_table: ReadTable) -> str:
+    """Build one target query after the project's calls table is resolved."""
+    if target_name == "calls":
+        return build_calls_export_query(calls_read_table)
+    if target_name == "objects":
+        return build_objects_export_query()
+    if target_name == "feedback":
+        return build_feedback_export_query()
+    raise ValueError(f"unknown export target {target_name!r}")
+
+
+def build_calls_export_query(read_table: ReadTable) -> str:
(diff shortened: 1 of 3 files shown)


amwarrier (approved):

weave/trace_server/export_targets.py
```diff
@@ -0,0 +1,47 @@
...
+    partition_by: str | None = None
+
+
+REGISTRY: dict[str, ExportTarget] = {
+    "calls": ExportTarget(
+        name="calls",
+        # Epoch-zero deleted_at is the calls_complete "not deleted" sentinel;
+        # rows mutate via lightweight UPDATE, so no dedup/FINAL is needed.
+        source_sql=(
+            "SELECT * FROM calls_complete "
```
amwarrier: Are all columns ok to be exported? should we have a column allow list?

amwarrier: don't we need calls_merged? or this is a calls_complete only feature?

amwarrier (approved):

weave/trace_server/export_targets.py
gtarpenning: Yes — this is intentionally a project-scoped raw-table export, so every calls_complete column is included. I dropped the allowlist: SELECT * means newly added call data is not silently omitted. A stable normalized/customer-facing schema would be a separate export contract; this one also retains operational/deletion metadata.

PR #7574, opened 2026-07-14
chore(weave): clickhouse export write path

gtarpenning:
## Summary
- `export_start` persists `manifest.json` through the configured file-storage client, then starts one daemon worker for the job. The worker executes target `INSERT INTO FUNCTION s3()` commands serially, so a single export does not fan out across targets.
- Export data goes directly from ClickHouse to Parquet through the server-configured `weave_exports` named collection. The application does not create or insert into any ClickHouse job table.
- The manifest is the durable job record; it contains the selected targets and deterministic object names without exposing the internal project ID.

## Testing
- Real ClickHouse query-log coverage, real S3 manifest persistence, and explicit proof that `export_jobs` is absent.
- Serial-worker ordering, target-failure continuation, invalid/empty/duplicate target rejection, and exact native export SQL.

### DIFF
--- a/weave/trace_server/export.py
+++ b/weave/trace_server/export.py
@@ -1,0 +1,185 @@
+"""Detached ClickHouse -> S3 export write path.
+
+`start_export` writes the job's manifest object before starting one background
+worker. The worker runs each `INSERT INTO FUNCTION s3()` serially, while the
+manifest and `system.query_log` are the durable read path.
+"""
+
+import datetime
+import json
+import logging
+import re
+import threading
+import uuid
+from collections.abc import Callable
+from dataclasses import dataclass
+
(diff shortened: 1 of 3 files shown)


weave/trace_server/export.py
```diff
@@ -0,0 +1,124 @@
...
+                job_id,
+                project_id,
+                target_names,
+                datetime.datetime.now(datetime.timezone.utc),
+            ]
+        ],
+        column_names=["job_id", "project_id", "targets", "created_at"],
+    )
+    for target in targets:
+        threading.Thread(
```
amwarrier: Pretty sure this is going to be improved soon.

amwarrier (approved):

PR #7575, opened 2026-07-14
chore(weave): export status via query_log

gtarpenning:
## Summary
- `export_status` reads the authorized project-scoped `manifest.json` object to discover selected targets, then derives each target state from `system.query_log`.
- No process-local job map and no ClickHouse metadata table are used. A fresh server can serve status from object storage plus query log.
- The project prefix and UUID validation keep unknown and cross-project jobs as 404; completed targets receive presigned read-only object URLs.

## Testing
- Real ClickHouse done/error/running query-log cases.
- Real S3 manifest recovery from a fresh server, cross-project/unknown/traversal guards, corrupt-manifest rejection, and a presigned download round trip.

### DIFF
--- a/weave/trace_server/export.py
+++ b/weave/trace_server/export.py
@@ -2,7 +2,7 @@
 
 `start_export` writes the job's manifest object before starting one background
 worker. The worker runs each `INSERT INTO FUNCTION s3()` serially, while the
-manifest and `system.query_log` are the durable read path.
+manifest and replica-wide `system.query_log` are the durable read path.
 """
 
 import datetime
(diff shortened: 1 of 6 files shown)


weave/trace_server/export.py
```diff
@@ -122,3 +225,5 @@ def _resolve_target(name: str) -> ExportTarget:
 # Bounds CH cost per detached insert via max_execution_time.
 EXPORT_MAX_EXECUTION_SECONDS = 300
 JOB_ID_RE = re.compile(r"^[0-9a-f-]{36}$")
+# Download-link lifetime for presigned GETs in the status manifest.
+PRESIGN_TTL_SECONDS = 3600
```
amwarrier, line 341: Do we plan to have a GC mechanism after the TTL? we should add a bucket lifecycle policy to GC old exported stuff.

amwarrier (approved):

PR #7577, opened 2026-07-14
fix(weave_ts): propagate conversation attributes across runIsolated frames

rgao-coreweave:
Design doc: https://docs.google.com/document/d/1HzfNC0UV59xUfeMWws1x5ve4WAq4iYvrjsjSXhk0Bwk/edit?tab=t.0

## What

`weave.startConversation({ attributes })` is documented to propagate custom attributes to every span in the conversation. In practice they were only read from the ambient `GenAIState` at span creation, so they reached only spans created in the conversation's own `runIsolated` frame. Callback-driven hosts that open each span in its own frame (correlating turns by held handles and a shared conversation id) lost the attributes on nearly every span.

The conversation id did not have this problem, because it is already passed down the handle chain. This PR passes the attributes down the same way.

## Why

Callback-driven integrations (for example the Claude Agent SDK hooks behind the weave-openclaw and weave-claude-code plugins) receive decoupled callbacks with no shared async chain and trace many sessions at once, so each callback runs in its own `runIsolated` frame. The conversation lives in the first frame and every later span is created in another, so integration name, tenant, and agent version silently dropped from most spans and could not be filtered on in the Agents tab.

## How

- `common.ts`: `ChildSpanContext` now carries `attributes`, alongside `parentContext` and `conversationId`.
- `conversation.ts`: `Conversation.startTurn` seeds the turn from the conversation handle (a per-turn value wins on a key collision).
- `turn.ts` / `llm.ts` / `subagent.ts`: each emitter stores its inherited attribute set and forwards it to every child it creates.
- Every emitter reads attributes from the `opts` passed down the handle chain, and the per-emitter ambient read of `state.conversation` is removed, so nothing depends on the frame.
- New `TurnInit.attributes` lets a rootless `startTurn({ attributes })` seed its children too.

Backward compatible: single-frame code is unchanged, since a turn started through `startConversation` is seeded with the same attributes the ambient read used to supply. The value now also reaches spans created in other frames.

## Testing

Full genai suite green (`jest genai`), build (typecheck) and eslint clean. New regression tests in `api.test.ts`, all confirmed to fail against the previous ambient-read implementation:

- cross-frame propagation to a tool under a turn;
- cross-frame propagation to a tool nested under an LLM;
- cross-frame propagation to a subagent and the tools nested under it;
- per-turn attributes override conversation attributes on key collision;
- rootless `startTurn({ attributes })` propagates to children.

The existing single-frame test passes both before and after the change.


sdks/node/src/genai/turn.ts
```diff
@@ -128,7 +136,10 @@ export class Turn extends SpanBase {
       );
     }
     const tracer = getWeaveTracer(WEAVE_GENAI_TRACER_NAME);
-    const attributes: Attributes = {...(state.conversation?.attributes ?? {})};
```
drtangible, line 131:
```ts
    const state = getGenaiState();
    if (state.turn !== null) {
      throw new Error(
        'A Turn is already active in this async chain. End it before starting a new one.'
      );
    }
```

This block above makes it seem like it's safe to rely on `state` here, but the bug described in this PR shows otherwise. Wonder if we can move that check into where we're actually using on that state -- presumably in the top-level API fns?

That might let us encapsulate all of the "gen ai" state dependencies in a one place, which might allow us to more clearly instruct customers when to use `runIsolated`.

sdks/node/src/genai/llm.ts
```diff
@@ -129,7 +130,9 @@ export class LLM extends SpanBase {
       );
     }
     const tracer = getWeaveTracer(WEAVE_GENAI_TRACER_NAME);
-    const attributes: Attributes = {...(state.conversation?.attributes ?? {})};
+    // Attributes arrive from the parent handle, not ambient state, so they
+    // survive across runIsolated frames.
+    const attributes: Attributes = {...(opts.attributes ?? {})};
```
drtangible: Same thought as in `turn.ts` about whether the error above should actually live here.

sdks/node/src/genai/common.ts
```diff
@@ -1,11 +1,14 @@
...
 
 /**
- * Internal options threaded from a parent emitter into a child class's
- * `create()` factory. Carries the OTel parent Context (which already has
- * the parent span attached) and the propagating conversation id.
+ * Internal options passed from a parent emitter into a child class's
+ * `create()` factory. Carries the OTel parent Context (which already has the
+ * parent span attached), the propagating conversation id, and the propagating
+ * custom attributes. All three travel with the handle rather than being read
+ * from ambient state, so they survive across `runIsolated` frames.
```
drtangible: Nit: I think talking about "ambient state" and "runIsolated" here spreads understanding of it in too many places. Since it's confusing, would be awesome if we could document that stuff all in one place, and let most of the unrelated code not have to reference it.

sdks/node/src/genai/subagent.ts
```diff
@@ -106,12 +107,14 @@ export class SubAgent extends SpanBase {
...
     this._agentVersion = opts.agentVersion;
+    this._attributes = opts.attributes;
   }
 
   static create(opts: SubAgentInit & ChildSpanContext): SubAgent {
-    const state = getGenaiState();
     const tracer = getWeaveTracer(WEAVE_GENAI_TRACER_NAME);
-    const attributes: Attributes = {...(state.conversation?.attributes ?? {})};
+    // Attributes arrive from the parent handle, not ambient state, so they
+    // survive across runIsolated frames.
```
drtangible: Same note, would be nice if none of these classes needed to have any understanding or documenation about these "ambient state" or "runIsolated"

drtangible (approved):
nice! 💯 💯 💯

Just added some non-blocking comments / followup ideas that might be able to further clarify some of the confusing global state / `runIsolated` stuff.

andrewtruong (approved):

sdks/node/src/genai/llm.ts
rgao-coreweave: done, thank you!

sdks/node/src/genai/subagent.ts
rgao-coreweave: Good point, moved these docs into a central place

sdks/node/src/genai/turn.ts
rgao-coreweave: good question, will follow up on this.

sdks/node/src/genai/common.ts
rgao-coreweave: done, thank you!

PR #7578, opened 2026-07-15
fix(weave): guard ref converter recursion against deeply-nested payloads

gtarpenning:
## Summary
- `calls/complete` (and any endpoint through the ref converter) 500s with an unhandled `RecursionError` on a deeply-nested payload: the copy-on-write walk in `trace_server_converter.py` had no depth guard. WB-37414.
- Add a total structure-depth cap (`_MAX_STRUCTURE_DEPTH=200`) threaded through `_walk` and the JSON-in-JSON re-walk, raising the existing typed `RequestTooLarge` (which the error registry already maps to 413) instead of blowing the stack.
- Owner is the converter itself: nesting depth is independent of body size (a ~50-byte payload reproduces it), so a request-size limit would not fix this.

## Testing
red/green: test commit fails with RecursionError on master, passes after the guard. covers dict/list/model + a within-cap positive control.

### DIFF
--- a/weave/trace_server/trace_server_converter.py
+++ b/weave/trace_server/trace_server_converter.py
@@ -30,7 +30,7 @@ from typing import Any, NamedTuple, TypeVar, cast
 from pydantic import BaseModel
 
 from weave.shared import refs_internal as ri
-from weave.trace_server.errors import InvalidExternalRef
+from weave.trace_server.errors import InvalidExternalRef, RequestTooLarge
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


weave/trace_server/trace_server_converter.py
```diff
@@ -47,12 +47,23 @@
 # ``chat_view._MAX_REF_SEARCH_DEPTH``.
 _MAX_REF_SEARCH_DEPTH = 8
 
+# Caps the copy-on-write walk over the payload's own container nesting
+# (dict / list / tuple / set / pydantic model). Without it a deeply-nested
+# body exhausts the Python stack and surfaces as an unhandled RecursionError
+# (HTTP 500); the guard converts that into a typed RequestTooLarge (HTTP 413).
+# The measured natural failure depth is ~330 for models on a shallow stack, so
+# 200 fires first with headroom for the request's own call stack, while staying
+# far above any realistic payload (real nesting is a handful of levels).
```
jtschoonhoven, line 56: Nit: no one has time to read this 🙃

jtschoonhoven (approved):
Clean

PR #7579, opened 2026-07-15
fix(weave): normalize Claude Code chat projection

bcsherma:
## Summary

- coalesce exact assistant text mirrored by a model `chat` span and its direct `assistant_text` transcript child
- use the transcript child's timestamp for response ordering while inheriting model, reasoning, token, cost, duration, and status metadata from the parent
- render Claude Code async `<task-notification>` inputs and async subagent-launch envelopes as tool activity instead of human/assistant chat bubbles
- add regression coverage for mirroring, distinct replies, transcript ordering, task notifications, and launch envelopes

## Root cause

Claude Code represents assistant output twice in the agent span tree: a richly attributed model `chat` span and a direct `assistant_text` child used for precise transcript ordering. The projection only suppressed mirrored output on `invoke_agent` wrappers, so both ordinary content spans became assistant bubbles. Some sibling model spans also share a coarse start time even though their transcript children record the true response order.

Claude Code additionally feeds async task completion notifications back to the model with provider role `user`, and records async subagent launch results as assistant JSON. Those execution events therefore appeared as human prompts or assistant chat bubbles.

The fix requires exact mirrored text plus direct parent/child linkage, and recognizes only the complete Claude task-notification wrapper and async-launch envelope. Distinct replies and ordinary invoke-agent output remain unchanged.

## Validation

- `nox -e lint` — passed all pre-push hooks, including ty, import-linter, pyright, mypy, Ruff, and format checks
- `python -m pytest tests/trace_server/test_genai_chat_view.py -q` — 41 passed, 1 expected xfail
- replayed all six turns of production Claude Code conversation `82736943-ae35-414e-859c-bd51de3055f2` through the patched projector:
  - all 13 exact `chat` → `assistant_text` mirrors emitted once at the child position, with parent model metadata inherited
  - both task notifications rendered as tools and none rendered as users
  - zero duplicate assistant texts; coarse sibling ties followed transcript-child timestamps
- passed the patched turn-1 projection through the actual frontend grouping functions: 17 visible assistant messages became 17 response segments, no assistant text remained inside a work fold, and both async launches remained in tool activity
- `bun test frontends/weave/src/components/PagePanelComponents/Home/Browse3/pages/AgentsPage/chat/sections.test.ts` — 12 passed

### DIFF
--- a/tests/trace_server/test_genai_chat_view.py
+++ b/tests/trace_server/test_genai_chat_view.py
@@ -678,6 +678,184 @@ def test_invoke_agent_mirrors_child_llm_output_messages() -> None:
     assert _assistant_payload(agent_messages[0]).text == "It's 72°F."
 
 
+def test_chat_span_mirrors_assistant_text_child() -> None:
+    """Claude Code emits assistant text on both the model chat span and a
+    direct assistant_text transcript child. The child keeps its timeline
+    position and inherits the parent model's richer metadata.
+    """
+
+    def at(milliseconds: int) -> datetime.datetime:
+        return datetime.datetime(
+            2026,
+            1,
+            1,
+            0,
+            0,
+            0,
+            milliseconds * 1000,
+            tzinfo=datetime.timezone.utc,
+        )
+
+    final_text = [
+        {
+            "role": "assistant",
+            "content": _parts(_text_part("I'll inspect the trace.")),
+        }
+    ]
+    spans = [
+        _span(
+            span_id="chat",
+            operation_name="chat",
+            agent_name="claude-code",
+            request_model="claude-opus-4-8",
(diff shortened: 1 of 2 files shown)


bcsherma:
* **#7579** <a href="https://app.graphite.com/github/pr/wandb/weave/7579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7579?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

rgao-coreweave (approved):
Super interesting.

Thanks for fixing.

PR #7581, opened 2026-07-15
fix(weave): allow span_name in agent-spans group_by/filter

gtarpenning:
## Summary
- `span_name` was missing from `SPAN_FILTERABLE_COLS`, which `SPAN_GROUP_BY_COLS` unions from, so the "Span name" alert filter dropdown was always empty.
- Added `span_name` as the first entry in the allowlist.

## Testing
unit test


bcsherma (approved):
hehe

PR #7584, opened 2026-07-16
fix(weave): stop bool/int param collision that 400s calls queries

ro31337:
## JIRA Issue(s)

[WB-37505](https://coreweave.atlassian.net/browse/WB-37505)

## Description

Some `/calls/stream_query` requests return a 400. ClickHouse says `Value True cannot be parsed as UInt64`. The whole calls list fails, not just one row.

The ticket read this as a missing cast in the filter: a boolean compared to a numeric column. The suggested fix was to coerce `bool -> 0/1` in the query builder.

But the filter side is already fine. A boolean literal is bound as a `Bool` param, and a test covers it. So the fix has to be somewhere else.

The real cause is in `ParamBuilder`. It reuses a parameter when the same value shows up twice. In Python `True == 1`, and they hash the same. One `ParamBuilder` is shared by the filter and the cost query. So when a request has `True` in the `query` and also asks for costs, the boolean literal and the cost query's `rank = 1` become the same parameter. That parameter is a `Bool` in one place and a `UInt64` in another, and it holds `True`. ClickHouse can't read `True` as `UInt64`, so it drops the request.

The fix is small: key the dedup cache on `(type, value)` instead of just `value`. Now `True` and `1` are separate parameters (this also splits `1` and `1.0`). Nothing else changes — dedup is only an optimization, so query results stay the same.

## Testing

Two tests, both fail on `master` and pass with this change:

- `ParamBuilder` unit test: `add_param(True)` and `add_param(1)` return different params, and each keeps its own type.
- Query test: a calls query with a boolean filter plus `include_costs=True`, checking that `rank` is bound to the integer `1`, separate from the boolean.

All 576 query-builder tests still pass, so no SQL changed. Locally clean: `ruff`, `ruff format`, `mypy`.

[WB-37505]: https://coreweave.atlassian.net/browse/WB-37505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## UPD: addressing "other paths"

Griffin asked whether this only affects the cost query. It doesn't.

The boolean is bound in the shared param builder, which the filter, the `query` literals, and the cost query all use. The cost query is just the most common trigger — it always binds `rank = 1` as `UInt64`. The same 400 also happens with no costs, e.g. `enabled == true AND count == 1` in one query, because the boolean and the integer collapse into one param.

Other builders (feedback, usage) go through the same `add_param`, so keying the dedup on `(type, value)` fixes the whole class in one place, not just the cost query. Added a no-costs functional test to cover it.


tests/trace_server/query_builder/test_costs_query.py
```diff
@@ -1243,3 +1244,116 @@ def test_query_calls_complete_with_costs_and_trace_name_order() -> None:
...
+    cq.add_field("id")
+    cq.add_field("started_at")
+    cq.add_condition(
+        tsi_query.EqOperation.model_validate(
+            {"$eq": [{"$getField": "inputs.enabled"}, {"$literal": True}]}
+        )
+    )
+    assert_sql(
+        cq,
+        """WITH filtered_calls AS
```
ro31337: Below is literal output of the cost-query builder `CallsQuery(project_id=.., include_costs=True)` with run through `.as_sql()`... -  every cost test in this file is the same giant blob checked. So I assume it's OK?

gtarpenning (commented):
can we add a functional test from the client or server that exercises this path? ideally we would generate the error as a caller first, (super ideally in CI so we can reference that commit) and then show that we are fixing. that way we have a clean red/green repro/fix. 

also, any other places that this might bite us? other paths that the query builder takes? seems odd that this would only affect us in the cost query.

gtarpenning (approved):

PR #7585, opened 2026-07-16
feat(weave): limit active exports per user

gtarpenning:
## Summary
- Redis is used only as a per-user cross-worker admission lease; it is not used as export/job metadata.
- The lease is acquired before preflight, compare-and-deleted on any preflight failure, and released in the detached worker `finally` after serial execution. Its 20-minute crash TTL covers three five-minute target limits.
- Redis outage fails closed with 503 and a second active export for the same authenticated user gets 409.

## Testing
- Same-user/different-user behavior, unavailable Redis, token-safe cleanup, preflight cleanup, and worker cleanup after both target failure and client-start failure.

### DIFF
--- a/weave/trace_server/export_rate_limit.py
+++ b/weave/trace_server/export_rate_limit.py
@@ -1,0 +1,81 @@
+"""Redis-backed cross-worker concurrency guard for managed-bucket exports."""
+
+import hashlib
+import logging
+import uuid
+from dataclasses import dataclass
+
+from weave.trace_server.redis_client import get_redis_client
+
+logger = logging.getLogger(__name__)
+
+# A single job may serialize all three targets, each limited to five minutes.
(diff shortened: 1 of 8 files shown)


weave/trace_server/export_rate_limit.py
```diff
@@ -0,0 +1,79 @@
...
+import hashlib
+import logging
+import uuid
+from dataclasses import dataclass
+
+from weave.trace_server.redis_client import get_redis_client
+
+logger = logging.getLogger(__name__)
+
+EXPORT_LOCK_TTL_SECONDS = 600
```
amwarrier: EXPORT_MAX_EXECUTION_SECONDS was 300. Deliberate mismatch?

amwarrier (approved):

PR #7588, opened 2026-07-16
chore(inference): use uv and add sync_model_catalog target

jamie-rasmussen:
## Description

Jira: https://coreweave.atlassian.net/browse/INF-807

Adds a `sync_model_catalog` Makefile target that is the same as `make update_model_catalog` except it passes an additional new `--skip-hugging-face` flag. Running with this flag enabled is much faster, doesn't need a network connection, and reduces Meticulous AI failures and merge conflicts caused by Hugging Face popularity metric updates that are unrelated to the main point of a PR.

Both targets updated to invoke the program through `uv`.

### DIFF
--- a/weave/trace_server/model_providers/catalog_augment_hf.py
+++ b/weave/trace_server/model_providers/catalog_augment_hf.py
@@ -122,19 +122,6 @@ def _fetch_hf_model_json(model_name: str) -> dict[str, Any] | None:
         return None
 
 
-def _hf_likes_downloads_from_response(d: dict[str, Any]) -> dict[str, Any]:
-    """Extract likes and downloads from a Hub API model payload."""
-    return pick_keys(d, HF_KEYS_TO_KEEP)
-
-
-def _hf_license_from_response(d: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


weave/trace_server/model_providers/catalog_augment_hf.py
```diff
@@ -126,27 +144,27 @@ def _hf_fields_from_response(d: dict[str, Any]) -> dict[str, Any]:
     Returns:
         Dict with ``likesHuggingFace``, ``downloadsHuggingFace``, and optionally ``license``.
     """
-    out = _hf_likes_downloads_from_response(d)
+    out: dict[str, Any] = _hf_likes_downloads_from_response(d)
```
gtarpenning: we can't type `_hf_likes_downloads_from_response` response? seems like we might know roughly the shape?

gtarpenning (approved):

weave/trace_server/model_providers/catalog_augment_hf.py
jamie-rasmussen: I'm actually going to inline _hf_likes_downloads_from_response - code is trivial

PR #7589, opened 2026-07-16
fix(weave): support $size in ORM

jtschoonhoven:
## Summary

Adds a Mongo-style `$size` operand to the trace-server Query DSL. This lets clients express non-empty structured fields with normal Query composition. 

This is required by https://github.com/wandb/core/pull/48318 which needs a way to query for empty maps.

### DIFF
--- a/tests/trace/test_feedback.py
+++ b/tests/trace/test_feedback.py
@@ -2330,7 +2330,7 @@ def test_feedback_query_returns_tz_aware_created_at(client: WeaveClient) -> None
     )
 
 
-def test_feedback_query_scored_only(client: WeaveClient) -> None:
+def test_feedback_query_by_collection_size(client: WeaveClient) -> None:
     project_id = client.project_id
 
     def create_feedback(
@@ -2365,7 +2365,26 @@ def test_feedback_query_scored_only(client: WeaveClient) -> None:
         tsi.FeedbackQueryReq(
             project_id=project_id,
             fields=["id"],
-            scored_only=True,
+            query=tsi.Query.model_validate(
+                {
+                    "$expr": {
+                        "$or": [
+                            {
+                                "$gt": [
+                                    {"$size": {"$getField": "scorer_tags"}},
+                                    {"$literal": 0},
+                                ]
+                            },
+                            {
+                                "$gt": [
+                                    {"$size": {"$getField": "scorer_ratings"}},
+                                    {"$literal": 0},
+                                ]
+                            },
+                        ]
+                    }
+                }
+            ),
         )
     )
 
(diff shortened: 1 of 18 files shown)


weave/trace_server/orm.py
```diff
@@ -373,7 +373,6 @@ def prepare(
                 datetime_columns=self.datetime_columns,
             )
             conditions.extend(query_conds)
-
```
gtarpenning, line 376: 🔪

weave/trace_server/interface/query.py
```diff
@@ -348,11 +363,15 @@ class ContainsSpec(BaseModel):
...
 )
-Operand = LiteralOperation | GetFieldOperator | ConvertOperation | Operation
+Operand = (
+    LiteralOperation | GetFieldOperator | ConvertOperation | SizeOperation | Operation
+)
 
 # Update the models to include the recursive types
 LiteralOperation.model_rebuild()
 GetFieldOperator.model_rebuild()
+ConvertOperation.model_rebuild()
```
gtarpenning: interesting that we need this?

tests/trace_server/query_builder/test_agent_query_compiler.py
```diff
@@ -167,6 +167,18 @@ def test_custom_attr_rejected_field_vs_field(self) -> None:
 
 
 class TestOperatorShapes:
+    def test_size_composes_with_comparison(self) -> None:
```
gtarpenning, line 170: rather presumptuous to put your test first ;)

gtarpenning (approved):
lgtm

tests/trace/test_feedback.py
```diff
@@ -2328,3 +2328,71 @@ def test_feedback_query_returns_tz_aware_created_at(client: WeaveClient) -> None
...
+    rated_id = create_feedback("rated", ratings={"_rating_": 0.8})
+
+    all_feedback = client.server.feedback_query(
+        tsi.FeedbackQueryReq(project_id=project_id, fields=["id"])
+    )
+    scored_feedback = client.server.feedback_query(
+        tsi.FeedbackQueryReq(
+            project_id=project_id,
+            fields=["id"],
+            query=tsi.Query.model_validate(
```
gtarpenning, line 2368: maybe one more test that uses a different literal value? (negative, float, non-sensical value, etc...)

weave/trace_server/interface/query.py
jtschoonhoven: We don't. Removed.

tests/trace_server/query_builder/test_agent_query_compiler.py
jtschoonhoven: I've been to nice to my agent lately and it's getting entitled. I'll deal with it.

PR #7591, opened 2026-07-17
feat(weave): link stamped agent spans to eval results

bcsherma:
## Summary

- Hydrates evaluation result trials from spans stamped with the evaluation run and predict-and-score call IDs.
- Links one root stamped span per trace while preserving explicit `genai_span_ref` values.
- Keeps hydration best-effort and batches/paginates agent span queries using existing trace-server limits.
- Covers merge behavior and the ClickHouse query path.

Companion UI PR: wandb/core#48399.

## Validation

- `uv run --group test python -m pytest tests/trace_server/test_eval_results_genai_span_ref.py` (7 passed)
- `uvx ruff check weave/trace_server/eval_results_helpers.py tests/trace_server/test_eval_results_genai_span_ref.py`
- Seeded local eval run `019f7155-0e31-785c-9b4c-270060511eab`; `eval_results/query` returned 3 linked root span refs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -87,6 +87,12 @@ When trace-server request/response models or route schemas change, refresh the A
 2. Copy that schema into the tracked Node SDK schema: `cp ../../weave-trace/openapi.json sdks/node/weave.openapi.json`.
 3. Regenerate the TypeScript client from `sdks/node`: `pnpm run generate-api`.
 
+Evaluation result rows merge agent span links from two sources: legacy
+`weave.genai_span_ref` call attributes and OTel spans whose promoted
+`eval_run_id` plus `eval_predict_and_score_call_id` columns identify the
+trial. Keep the promoted-column hydration best-effort so eval results remain
+available during rolling deploys.
+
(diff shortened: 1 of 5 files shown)


nikumar1206:
[P2] Batch eval IDs too. [_query_eval_agent_spans](https://github.com/wandb/weave/blob/b36480b4390681ed0b6244ae90242b65b6731eb5/weave/trace_server/eval_results_helpers.py#L481) batches predict_and_score_call_ids but puts every eval_call_id into each query. Eval-result requests do not bound that list, bypassing the repository’s 1,000-item filter convention. Batch correlated (eval_id, predict_and-score_id) pairs (or both dimensions) to keep query size bounded and avoid cross-product overfetch.

[P2] Don’t make the in-memory backend scan every agent span. [InMemoryTraceServer.eval_results_query](https://github.com/wandb/weave/blob/b36480b4390681ed0b6244ae90242b65b6731eb5/weave/trace_server/in_memory_trace_server.py#L6842) now calls hydration through agent_spans_query(query=...), but that in-memory implementation ignores req.query. Every eval-results request therefore pages through all agent spans in the project before post-filtering. Either apply this fixed eval filter in-memory or avoid hydration there until generic agent-query filtering exists.

nikumar1206 (approved):

PR #7593, opened 2026-07-17
fix(weave): resolve Bedrock inference profile region from ARN for cost tracking

ro31337:
## Description

Bedrock calls made through an inference profile show tokens but no cost: https://coreweave.atlassian.net/browse/WB-29454 . It was reported as an async race — the slower of two concurrent calls loses its cost — but the handler has no shared state. It's a model-name problem.

Weave computes cost at read time by matching the stored model name against the price table. For an inference-profile modelId (an ARN), the integration first resolves the ARN to a real model name via `GetInferenceProfile`, which must run in the profile's region. That region came from `AWS_REGION_NAME`, which boto3 never reads — and it doesn't read `AWS_REGION` either, only `AWS_DEFAULT_REGION`. So with `AWS_REGION` or nothing set, the resolve failed, the stored name matched no price row, and the call kept its tokens but lost its cost.

The region is already in the ARN (`arn:aws:bedrock:<region>:...`), so we read it from there. This also handles the case where the caller's default region differs from the profile's.

## Testing

Verified the boto3 behavior directly (1.41.3): `AWS_REGION` alone raises `NoRegionError`, `AWS_DEFAULT_REGION` alone resolves, and botocore maps the region to `AWS_DEFAULT_REGION` only. Added a unit test asserting the control-plane client is built in the ARN's region and the model name resolves. Full Bedrock suite passes (16), plus ruff, mypy, and pyright.

### DIFF
--- a/tests/integrations/bedrock/bedrock_test.py
+++ b/tests/integrations/bedrock/bedrock_test.py
@@ -10,6 +10,9 @@ from moto import mock_aws
 
 import weave
 from weave.integrations.bedrock import patch_client
+from weave.integrations.bedrock.bedrock_sdk import (
+    extract_model_name_from_inference_profile_arn,
+)
 
 model_id = "anthropic.claude-3-5-sonnet-20240620-v1:0"
 inference_profile_id = "arn:aws:bedrock:us-east-1::foundation-model/anthropic.claude-3-5-sonnet-20240620-v1:0"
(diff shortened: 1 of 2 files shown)


ro31337:
Closing for now — I'm not confident the bug still triggers, and I couldn't verify the fix end-to-end. It looks harmless, but I'd rather not ship something I can't confirm. Can reopen if a real repro comes up.

PR #7594, opened 2026-07-17
fix(weave): preserve user prompt in conversation previews

mattliu-mygit:
## Description

- Fixes [WB-37420](https://coreweave.atlassian.net/browse/WB-37420), which is caused by WB-agent prepending logged content with JSON that includes info such as page context. After we truncate, we may parse this JSON incompletely, resulting in preview text being ","
- Core creates [one page-context part](https://github.com/wandb/core/blob/a2832038c8bd66b7b38a7bb90d517863317a047d/frontends/app/src/agentChat/pageContext/buildPromptContext.ts#L18-L33) and [prepends it to the customer's prompt parts](https://github.com/wandb/core/blob/a2832038c8bd66b7b38a7bb90d517863317a047d/frontends/app/src/agentChat/promptAssembly.ts#L44-L53). Conversation previews truncated that combined value before the frontend separated structured data from prose; at certain JSON boundaries, cleanup removed objects but left a separator comma as the title.
- Remove complete leading JSON objects or arrays before applying the preview limit. Complete JSON later in the prose is represented as `[JSON]`, preserving the fact that the customer referenced structured data without putting the full blob in the title.
- Plain text, malformed leading JSON, and comma-separated top-level JSON remain unchanged. JSON-only input produces no user preview, allowing the existing assistant, agent-name, or conversation-ID fallback to apply.

For example:

```text
Input:
{"page": "..."}
Compare {"run": "a"} with {"run": "b"}

Preview: Compare [JSON] with [JSON]
```

## Testing

- Unit coverage for plain text, one and multiple leading JSON values, JSON in the middle of prose, multiple embedded values, arrays, JSON-only input, malformed JSON, and comma-separated JSON.
- ClickHouse-backed regression reproducing the production preview cutoff and asserting the stored question becomes the conversation title.
- `106 passed, 1 xfailed`; Ruff check and format clean.

Local reproduction using the same conversation before and after the change:

| before (current behavior) | after (this PR) |
|---|---|
| Conversation title shown as a comma before the fix | Conversation title derived from the user question after the fix |

[WB-37420]: https://coreweave.atlassian.net/browse/WB-37420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/weave/trace_server/agents/chat_view.py
+++ b/weave/trace_server/agents/chat_view.py
@@ -788,14 +788,60 @@ def _extract_user_text(
 
 
 def first_user_preview_text(messages: list[NormalizedMessage]) -> str:
-    """Public: opening user-prompt text from a span's `input_messages`.
+    """Public: readable opening user-prompt text from `input_messages`.
 
     The opening user prompt is the first user entry of the earliest
     message-bearing span; that span's `input_messages` begins at the
     conversation opening.
     """
     texts = _user_prompt_texts(messages)
-    return texts[0] if texts else ""
+    return _json_aware_preview_text(texts[0]) if texts else ""
+
+
+def _json_aware_preview_text(text: str) -> str:
+    """Remove leading JSON and replace later values before preview truncation.
+
+    Conversation titles use prose, so JSON-only input produces no user preview.
+    Malformed leading JSON is left unchanged rather than partially stripped.
+    """
+    stripped = text.lstrip()
+    cursor = 0
+    parsed_json = False
+    decoder = json.JSONDecoder()
+    while cursor < len(stripped) and stripped[cursor] in "[{":
+        try:
+            _, cursor = decoder.raw_decode(stripped, cursor)
+        except json.JSONDecodeError:
+            return text
+        parsed_json = True
+        while cursor < len(stripped) and stripped[cursor].isspace():
+            cursor += 1
+        if cursor < len(stripped) and stripped[cursor] == ",":
+            return text
+
+    trailing = stripped[cursor:].strip() if parsed_json else text
+    return _replace_json_values(trailing)
+
+
+def _replace_json_values(text: str) -> str:
+    """Replace complete JSON objects/arrays in preview prose with `[JSON]`."""
+    parts: list[str] = []
+    cursor = 0
+    decoder = json.JSONDecoder()
+    while cursor < len(text):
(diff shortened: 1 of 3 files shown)


mattliu-mygit:

weave/trace_server/agents/chat_view.py
```diff
@@ -788,14 +788,33 @@ def _extract_user_text(
...
+    This runs before preview truncation so JSON punctuation cannot become the
+    conversation title.
+    """
+    stripped = text.lstrip()
+    try:
+        value, end = json.JSONDecoder().raw_decode(stripped)
+    except json.JSONDecodeError:
+        return text
+
+    if not isinstance(value, dict) or value.get("type") != "wandb_web_page_state":
```
bcsherma: Why are we hard coding `wandb_web_page_state`? This is a data key specific to the the wb agent implementation that ends up in the user message

mattliu-mygit: There is a cleaner upstream solution. For structured WB Agent prompts, we currently create the opening trace span with an [empty user message](https://github.com/wandb/core/blob/a2832038c8bd66b7b38a7bb90d517863317a047d7/services/wb_agent/src/weave_tracing/logger.py#L533-L541). We could change tracing to extract the actual question from the prompt parts and record that instead. However, this would have a much larger blast radius on the rest of our flow. Not sure if we want to limit blast radius with a change like this or go for this change instead.

We need to check specifically for [wandb_web_page_state](https://github.com/wandb/core/blob/a2832038c8bd66b7b38a7bb90d517863317a047d7/services/wb_agent/src/agent_interface_contract/prompt_parts.py#L246-L251) because Weave receives the page state as plain JSON text, not as structured data. There is no shared Core schema or registry available to Weave that we can use to identify it automatically. Not sure how else to identify this part of the input without hardcoding

mattliu-mygit: Actually, we could just remove the first JSON in logging, assuming that Weave will always prepend some sort of context JSON to its list input. Adding that as a followup commit. This would decouple our preview processing code from hardcodes like wandb_web_page_state.

bcsherma (approved):
🔥

PR #7600, opened 2026-07-21
feat(weave): emit agent events for insights

jtschoonhoven:
## Summary

- Adds a new kafka event, `weave.embed_agent_span` that is consumed by the new insights worker.
- Decouples `ENABLE_ONLINE_EVAL` from `ENABLE_AGENT_SCORING` so that agent scoring can be enabled independently.
- Adds `ENABLE_AGENT_INSIGHTS` which, if enabled, will emit `weave.embed_agent_span`.

### DIFF
--- a/tests/trace_server/test_clickhouse_trace_server_batched.py
+++ b/tests/trace_server/test_clickhouse_trace_server_batched.py
@@ -2056,9 +2056,7 @@ def _turn_ended_span_row() -> AgentSpanCHInsertable:
         "all-consumers-off",
     ],
 )
-def test_genai_otel_export_emit_gate(
-    monkeypatch, online_eval, scoring, insights
-):
+def test_genai_otel_export_emit_gate(monkeypatch, online_eval, scoring, insights):
     """OTel ingest emits for scoring or Insights, independent of online eval."""
     mock_producer = MagicMock()
     monkeypatch.setattr(
--- a/tests/trace_server/test_clickhouse_trace_server_batched.py
+++ b/tests/trace_server/test_clickhouse_trace_server_batched.py
@@ -1999,14 +1999,18 @@ def _turn_ended_span_row() -> AgentSpanCHInsertable:
     [
         (True, True, False, True),
         (True, False, True, True),
-        (False, False, True, True),
+        (True, True, True, True),
+        (False, False, True, False),
+        (False, True, True, False),
         (False, True, False, False),
         (True, False, False, False),
     ],
     ids=[
         "online-eval-and-scoring",
         "online-eval-and-insights",
-        "insights-only",
+        "online-eval-with-both-consumers",
+        "insights-without-online-eval",
+        "both-consumers-without-online-eval",
         "scoring-without-online-eval",
         "all-consumers-off",
     ],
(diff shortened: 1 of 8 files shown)


tests/trace_server/test_clickhouse_trace_server_batched.py
```diff
@@ -1995,18 +1995,26 @@ def _turn_ended_span_row() -> AgentSpanCHInsertable:
...
 
 @pytest.mark.parametrize(
-    ("online_eval", "scoring", "should_emit"),
+    ("online_eval", "scoring", "insights", "should_emit"),
     [
-        (True, True, True),
-        (True, False, False),  # scoring off -> skip
-        (False, True, False),  # online eval off -> skip
+        (True, True, False, True),
+        (True, False, True, True),
```
ro31337: Sooo....

```text
online_eval = true
scoring = false
insights = true
```

and `should_emit = true or (true and false)` - true, right?

But scoring is off. And scorer will receive the event (and won't see the flag `scoring = false` and will score), because the topic is the same. But the scoring is set to "False".

So the scoring kill switch is bypassed?

jtschoonhoven: I just updated the PR so the event gets emitted if `ENABLE_ONLINE_EVAL` _and_ (`ENABLE_AGENT_INSIGHTS` or `ENABLE_AGENT_SCORING`)

ro31337: For the same flag combination:

```text
online_eval = true
scoring = false
insights = true
```

 the condition evaluates to:

 ```text
should_emit = online_eval and (scoring or insights)
                      = true and (false or true)
                      = true
 ```

Therefore, when online eval and agent insihgts are enabled, the event is published to the shared topic (`weave.score_agent_spans` -- even though ENABLE_AGENT_SCORING is disabled).

Botth the insights worker and the agent scoring worker consume this topic using different consumer groups, for example:

1) https://github.com/wandb/core/blob/master/services/weave-trace/src/workers/agent_scoring_worker/__main__.py#L59-L61

2) https://github.com/wandb/core/blob/master/services/weave-trace/src/workers/insights_worker/__main__.py#L52-L54

Because the consumer groups are independent, both groups receive the event. And the scoring worker will still perform scoring...

Before this PR, scoring = false prevented these events from being published, so the scoring worker did not start any scoring jobs.

In this PR, when we enable insights it will cause the events to be published anyway, so scoring jobs can run even though scoring is disabled.

Solution is... Actually I haven't thought about this. We either need separate topics (ideally, IMO), or a consumer-side scoring gate, changing only the producer condition does not preserve the scoring kill switch.

Who is the best Kafka expert on the team, @gtarpenning ? We need an extra pair of eyes here - I'm curious if we can just change the Kafka topic name and it will be created automagically? In this case something tells me we should somehow migrate previous events? Hm...

jtschoonhoven: I don't think these should be separate events.

In practice, the agent scoring worker won't be installed into a deployment at all unless `ENABLE_AGENT_SCORING=true`, and same with `ENABLE_AGENT_INSIGHTS`. But I can also add a consumer-side gate just to be sure.

jtschoonhoven: Nevermind you're right, thanks for blocking on this. Making changes now.

ro31337 (approved):

PR #7603, opened 2026-07-21
feat(weave_ts): add agent conversation URL helper

gtarpenning:
## Summary
- export `agentConversationPath(projectId, conversationId)` from the Node SDK
- use the initialized W&B domain, with `wandb.ai` as the pre-initialization default
- URL-encode app-provided conversation IDs

This adds the TypeScript equivalent of Python's `agent_conversation_path()` used by wandb/docs#2913.

## Testing
- `pnpm test -- urls.test.ts --runInBand`
- `pnpm lint`
- `pnpm prettier-check`
- `pnpm typecheck:cjs`
- `pnpm typecheck:esm`
- `pnpm build`
- runtime export checks against both `dist/index.js` and `dist/index.mjs`

## Breaking changes
None.


drtangible (commented):
Not opposed to adding, but one question: before we reference the Python counterpart to this in [documentation](https://github.com/wandb/docs/pull/2913/changes#diff-9bd1b0c4cb3c3643b8f1324b557bc497c3b6af52c7ab9ae9b13b9ba9ffee0b35R426-R435), do we want to add it to `__init__.py`, to make it part of the supported public API for the Python SDK?

drtangible:
> Not opposed to adding, but one question: before we reference the Python counterpart to this in [documentation](https://github.com/wandb/docs/pull/2913/changes#diff-9bd1b0c4cb3c3643b8f1324b557bc497c3b6af52c7ab9ae9b13b9ba9ffee0b35R426-R435), do we want to add it to `__init__.py`, to make it part of the supported public API for the Python SDK?

Followed up on this offline w/ @andrewtruong -- for this change, could we expose this url helper namespaced in some capacity, instead of directly on the top-level? Maybe just `urls.agentConversationPath(...)`?

PR #7608, opened 2026-07-22
feat(weave): allow custom runtimes without API keys

mattliu-mygit:
## Summary

- allow custom runtime endpoints to omit an API-key secret
- preserve configured custom headers, including alternate authentication headers
- keep existing secret lookup and missing-secret behavior unchanged when a secret is configured

Paired with wandb/core#48619.

## Validation

- `.venv/bin/python -m pytest -p no:rerunfailures tests/trace_server/test_llm_completion.py::TestGetCustomProviderInfo tests/trace_server/test_custom_provider.py::test_get_custom_provider_info -q` (7 passed)
- `.venv/bin/python -m compileall -q weave/trace_server/llm_completion.py tests/trace_server/test_llm_completion.py`
- `git diff --check`

### DIFF
--- a/tests/trace_server/test_llm_completion.py
+++ b/tests/trace_server/test_llm_completion.py
@@ -169,6 +169,47 @@ class TestGetCustomProviderInfo(unittest.TestCase):
         finally:
             _secret_fetcher_context.reset(token)
 
+    def test_provider_without_api_key_skips_secret_lookup(self):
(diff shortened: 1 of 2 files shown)


mattliu-mygit:
Consolidated into #7614 as the first of two commits. The combined PR also adds coverage for the configured-but-missing-secret path while preserving existing configured-secret behavior.

PR #7612, opened 2026-07-22
feat(weave): add totals to feedback queries

jtschoonhoven:
## Summary

To unlock server-side pagination for the Signals tab we first need to return the page size for every FeedbackQueryRes. Blocks https://github.com/wandb/core/pull/48768.

- return `total_count` from every existing `/feedback/query` response
- compute the filtered count before `limit` and `offset`
- match the behavior in ClickHouse and the in-memory trace server

## Testing

- [x] Tested locally

### DIFF
--- a/tests/trace/test_feedback.py
+++ b/tests/trace/test_feedback.py
@@ -1162,6 +1162,48 @@ def test_agent_monitor_feedback_sort_by_map_column(client: WeaveClient) -> None:
     ]
 
 
+def test_feedback_query_total_count_precedes_pagination(client: WeaveClient) -> None:
+    project_id = client.project_id
+    for index, feedback_type in enumerate(["keep", "skip", "keep"]):
+        client.server.feedback_create(
+            tsi.FeedbackCreateReq(
+                project_id=project_id,
+                weave_ref=f"weave:///{project_id}/call/{index}",
+                feedback_type=feedback_type,
+                payload={},
+            )
+        )
+
+    without_total = client.server.feedback_query(
+        tsi.FeedbackQueryReq(project_id=project_id, fields=["id"], limit=1)
+    )
+    assert without_total.total_count is None
+
+    result = client.server.feedback_query(
+        tsi.FeedbackQueryReq(
+            project_id=project_id,
+            fields=["weave_ref"],
+            query=Query(
+                **{
+                    "$expr": {
+                        "$eq": [
+                            {"$getField": "feedback_type"},
+                            {"$literal": "keep"},
+                        ]
+                    }
+                }
+            ),
+            sort_by=[SortBy(field="weave_ref", direction="asc")],
+            limit=1,
+            offset=1,
+            include_total=True,
+        )
+    )
+
+    assert result.total_count == 2
+    assert result.result == [{"weave_ref": f"weave:///{project_id}/call/2"}]
+
+
 async def populate_feedback(client: WeaveClient) -> None:
     @weave.op
     def my_scorer(x: int, output: int) -> int:
--- a/tests/trace/test_feedback.py
+++ b/tests/trace/test_feedback.py
@@ -1174,11 +1174,6 @@ def test_feedback_query_total_count_precedes_pagination(client: WeaveClient) -> 
             )
         )
 
-    without_total = client.server.feedback_query(
-        tsi.FeedbackQueryReq(project_id=project_id, fields=["id"], limit=1)
-    )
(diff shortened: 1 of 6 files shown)


gtarpenning (approved):

gtarpenning:
## 🔴 Red-team report

> [!NOTE]
> **Verdict: no red test warranted.** 8 vectors exercised against the in-memory backend (`--trace-server=fake`) plus static analysis of the ClickHouse path. The count is computed before `limit`/`offset` and is self-consistent with the paged result set on every vector tried. Existing PR test + 3 edge-case probes green.

### Attack matrix

| # | Attack vector | Hypothesis it would break | Result | Evidence |
|---|---|---|:--:|---|
| 1 | In-memory count on a **zero-match** filter | `count_result[0][...]` IndexError (empty aggregate → 500) | ✅ holds | `_orm_select` count branch returns `[{"count(*)": 0}]` for empty match; probe → `total_count=0`, `result=[]` |
| 2 | Missed `FeedbackQueryRes(...)` constructor (field now required, no default → 500) | some path builds the res without `total_count` | ✅ holds | only 2 non-test constructors exist, both updated by the PR |
| 3 | `limit=0` "count-only" page (natural Signals pattern) | falsy-`0` treated as "no limit" → returns all rows | ✅ holds | both ORMs gate on `is not None`; probe → `total_count=3`, `result=[]` |
| 4 | `offset` past end of filtered set | crash / wrong total | ✅ holds | probe → `total_count=2`, `result=[]` |
| 5 | `count(*)` over `ReplacingMergeTree` without `FINAL` | count over-counts vs deduped paged list | ✅ holds | see diagram — neither path uses `FINAL`; `id` is unique per create; `feedback_replace` purges then re-creates |
| 6 | count vs list filter divergence (JSON/dynamic field) | count filter ≠ list filter | ✅ holds | both call the same `.where(req.query)` / `_orm_eval_query`; identical predicate by construction |
| 7 | Backend divergence CH vs in-memory count | two backends disagree on total | ✅ holds | count applies only `project_id` + shared `where`; list adds sort/limit/offset which don't affect count |
| 8 | Concurrent insert between count and list round-trips | transient count≠result | ⚠️ n/a | contrived race, needs artificial interleaving — suppressed by severity filter |

### Why #5 is safe

```mermaid
flowchart LR
    A["feedback_query"] --> C["count(*) query<br/>project_id + where, NO FINAL"]
    A --> L["list query<br/>+ sort/limit/offset, NO FINAL"]
    C --> R{"same raw<br/>(unmerged) rows?"}
    L --> R
    R -->|"yes: neither dedups"| OK["total_count == rows the<br/>user pages through ✅"]
    K["feedback_create → fresh UUID id<br/>feedback_replace → purge then re-create"] --> U["ORDER BY key incl. id<br/>never collapses distinct feedback"]
```

The overcount worry only bites if the list path deduped while `count(*)` did not. Neither uses `FINAL`, so both see identical raw rows and `total_count` equals what pagination actually traverses.

<details>
<summary>Repro - edge cases probed (in-memory backend)</summary>

```python
# zero-match filter
r = server.feedback_query(FeedbackQueryReq(project_id=pid, fields=["weave_ref"],
        query=eq("feedback_type", "nonexistent")))
assert r.total_count == 0 and r.result == []          # no IndexError

# limit=0 count-only page
r = server.feedback_query(FeedbackQueryReq(project_id=pid, fields=["weave_ref"], limit=0))
assert r.total_count == 3 and r.result == []          # 0 not swallowed as "no limit"

# offset past end
r = server.feedback_query(FeedbackQueryReq(project_id=pid, fields=["weave_ref"],
        query=eq("feedback_type", "keep"), limit=10, offset=99))
assert r.total_count == 2 and r.result == []
```
All three pass, alongside the PR's own `test_feedback_query_total_count_precedes_pagination`.
</details>

> [!WARNING]
> **Residual (low, no test):** making `total_count` a required field with no default means a **newer SDK parsing an older server's** feedback response (which omits the field) hits a `model_validate` ValidationError in `RemoteHTTPTraceServer`. Reachable only in a self-hosted SDK-ahead-of-server skew window, and it matches this repo's established convention (all 56 `*Res` fields, incl. sibling `CallsQueryStatsRes.count` / `EvalResultsQueryRes.total_rows`, are required-no-default). Flagged for awareness, not action.

_ClickHouse path validated by static analysis only (no Docker in this run); its `count(*)` always returns exactly one row, so the empty-aggregate crash class (#1) cannot occur there either._

PR #7614, opened 2026-07-22
feat(weave): add custom runtime registration API

mattliu-mygit:
## Summary

- Add one desired-state Custom Runtime operation over the existing `Provider` and `ProviderModel` objects.
- Preserve existing object IDs, Provider digest links, and `custom::<provider>::<id>` Playground selectors.
- Allow team-secret, header-only, and unauthenticated OpenAI-compatible endpoints.
- Reconcile the complete runtime-ID list and remove omitted IDs only after desired writes succeed.

## API

The companion Core change exposes:

```http
PUT /v2/{entity}/{project}/runtimes/{runtime_name}
```

Repeating the same request is logically idempotent. If `max_tokens` is omitted, it defaults to `4096`. Updated endpoint configuration may remain cached by an inference replica for up to 60 seconds.

## Compatibility

Existing custom providers retain the same representation and behavior:

- `Provider` and `ProviderModel` remain authoritative.
- Base URLs, team-secret references, headers, endpoint IDs, and token limits retain their existing fields.
- Existing UI-created configurations can be updated through this operation.
- Configured team secrets retain their existing lookup and missing-secret behavior.

## Related changes

- Core route: [wandb/core#49011](https://github.com/wandb/core/pull/49011)
- Python SDK wrapper: #7617
- Frontend migration: [wandb/core#49164](https://github.com/wandb/core/pull/49164)

## Verification

- `18 passed` across focused Custom Runtime and API-routing tests.
- `51 passed` across focused completion tests.
- Local E2E covered public SDK registration, persistence, Playground selection, and completion with header-only and named-secret authentication.

The E2E harness supplied test user/project authentication and secret lookup and allowed localhost; production authentication and public-URL validation were unchanged.


weave/trace_server/constants.py
```diff
@@ -3,6 +3,7 @@
 COMPLETIONS_CREATE_OP_NAME = "weave.completions_create"
 IMAGE_GENERATION_CREATE_OP_NAME = "weave.image_generation_create"
 
+DEFAULT_CUSTOM_RUNTIME_MAX_TOKENS = 4096
```
mattliu-mygit, line 6: This is to be in parity with the UI, which also populates 4096 tokens on default (ref https://github.com/wandb/core/blob/agent/agent-runtime-cosmetic-mvp/frontends/app/src/weave/components/PagePanelComponents/Home/Browse3/pages/PlaygroundPage/llmMaxTokens.ts#L1427-L1456).

weave/trace_server/llm_completion.py
```diff
@@ -608,17 +608,15 @@ def get_custom_provider_info(
 
     secret_name = provider_obj.api_key_name
 
-    # Get the API key
```
mattliu-mygit, line 611: Making API key optional to be in parity with UI changes for custom runtime: https://github.com/wandb/core/pull/48619

weave/trace_server/llm_completion.py
```diff
@@ -608,17 +608,15 @@ def get_custom_provider_info(
...
-        raise InvalidRequest(f"No secret name found for provider {provider_name}")
-
-    api_key = secret_fetcher.fetch(secret_name).get("secrets", {}).get(secret_name)
-
-    if not api_key:
-        raise MissingLLMApiKeyError(
-            f"No API key {secret_name} found for provider {provider_name}",
-            api_key_name=secret_name,
-        )
+    api_key = ""
```
ro31337: I understand the intent, but what happens when a custom runtime has no API-key secret and this empty string is passed into LiteLLM? 

Could its OpenAI-compatible client construct an empty Authorization header and fail before the request reaches the endpoint? Just asking...

mattliu-mygit: This is a good callout. I just tested locally and it seems that LiteLLM calls do fail with no api key since it treats the lack of api key as a missing credential. We were mocking before so this was not tested for.

The OpenAI client actually allows for empty api keys and for users to configure auth themselves via headers. I'll assess that as an option for hitting custom customer runtimes.

mattliu-mygit: We can send custom-runtime requests directly through AsyncOpenAI (rather than LiteLLM). Custom runtimes already implement the OpenAI protocol, so LiteLLM’s provider translation is unnecessary; the SDK still handles streaming, tools, and response parsing.
```
client = AsyncOpenAI(
    base_url=runtime.base_url,
    api_key=resolved_api_key or "",
    default_headers=resolved_headers,
)
response = await client.chat.completions.create(**inputs)
```
Braintrust uses the same OpenAI-compatible custom-endpoint model with custom authentication headers: [Braintrust custom providers](https://www.braintrust.dev/docs/integrations/ai-providers/custom). Vercel provides the clearest code precedent for optional bearer auth plus header-only requests: [implementation](https://github.com/vercel/ai/blob/main/packages/openai-compatible/src/openai-compatible-provider.ts#L146-L149)

Evaluating this option through a Codex implementation locally to definitely see if it works.

mattliu-mygit: This worked; tested both API-key and non-API key routes for custom runtimes. Non-custom runtimes will not be affected by this.

weave/trace_server/llm_completion.py
```diff
@@ -288,14 +289,25 @@ def lite_llm_completion(
     extra_headers: dict[str, str] | None = None,
     vertex_credentials: str | None = None,
 ) -> tsi.CompletionsCreateRes:
+    if _use_direct_openai_client(provider, api_key, inputs.model):
```
mattliu-mygit, line 292: Using OpenAI Client in cases of custom + not Ollama (since it requires API key) + API key to allow all usecases except custom runtimes users configure without API key to work since LiteLLM throws validation error on empty API key field.

weave/trace_server/custom_runtime.py
```diff
@@ -0,0 +1,209 @@
...
+from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.errors import InvalidRequest
+from weave.trace_server.interface.builtin_object_classes.provider import (
+    Provider,
+    ProviderModel,
+    sanitize_name_for_object_id,
+)
+
+
+def apply_custom_runtime(
```
mattliu-mygit, line 12: Similar to FE's handle save logic for custom models/runtimes:
https://github.com/wandb/core/blob/e6342a5919d86ee5f61a0d5bc974732849d68e13/frontends/app/src/weave/components/PagePanelComponents/Home/Browse3/pages/OverviewPage/AddProviderDrawer.tsx#L121-L220

Planning to migrate FE to use this in the future once this is merged.

mattliu-mygit: This entire transaction is non-atomic, but worst case is user sees partially created or updated resources with an error that states what happened. This is also consistent with past UI behavior. Should be ok for MVP since non-atomic runs for custom runtime creation is not catastrophic and not too bad of a customer experience.

weave/trace_server/custom_runtime.py
```diff
@@ -0,0 +1,209 @@
...
+                    builtin_object_class="ProviderModel",
+                    wb_user_id=req.wb_user_id,
+                )
+            )
+        )
+        result_ids.append(
+            tsi.CustomRuntimeIDRes(
+                id=runtime_id.id,
+                max_tokens=runtime_id.max_tokens,
+                playground_id=f"custom::{req.runtime_name}::{runtime_id.id}",
```
ro31337, line 193: nite: what happens if `runtime_name` is "custom"? We have code that would do `replace("custom::", "")`, and in this case it would become something like `gpt-4`.

there is a code that says `name_part = model_name.replace("custom::", "")` - would be nice if you can change it to something like `removeprefix("custom::")`.

A tiny follow-up fix would work better, IMO since this PR is already big, approving!

ro31337 (approved):

weave/trace_server/custom_runtime.py
mattliu-mygit: Sounds good, thanks for the callout! Will raise a followup!

mattliu-mygit: Ref: https://github.com/wandb/weave/pull/7641

PR #7617, opened 2026-07-23
feat(weave): add custom runtime SDK wrapper

mattliu-mygit:
## Summary

- Add `WeaveClient.register_custom_runtime(...)` for registering or replacing a Custom Runtime configuration.
- Accept string runtime IDs with the default token limit or `weave.CustomRuntimeID` values with explicit limits.
- Implement the manual HTTP transport against `PUT /v2/{entity}/{project}/runtimes/{runtime_name}`.
- Configure the generated Node client to URL-encode Custom Runtime names.

## Usage

```python
import weave

client.register_custom_runtime(
    name="support-agent",
    base_url="https://agent.example/v1",
    api_key_secret="AGENT_API_KEY",
    headers={"X-Tenant-ID": "customer-1"},
    runtime_ids=[
        "support-v12",
        weave.CustomRuntimeID(id="support-canary", max_tokens=8192),
    ],
)
```

`runtime_ids` is the complete desired list. IDs omitted from a later successful call are removed.

## Related changes

- Runtime implementation: #7614
- Shared request validation: #7649
- Core HTTP route: [wandb/core#49011](https://github.com/wandb/core/pull/49011)

## Verification

- Python SDK and HTTP transport: `24 passed`.
- Generated Node client URL regression: `1 passed`.
- Node CJS/ESM typechecks, ESLint, Prettier, Ruff, and Python formatting passed.

### DIFF
--- a/weave/trace/weave_client.py
+++ b/weave/trace/weave_client.py
@@ -7,7 +7,7 @@ import logging
 import os
 import re
 import time
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Iterable, Mapping, Sequence
 from concurrent.futures import Future
 from functools import cached_property
 from typing import TYPE_CHECKING, Any, TypedDict, TypeVar, cast
@@ -146,6 +146,9 @@ from weave.trace_server.trace_server_interface import (
     CostPurgeReq,
     CostQueryOutput,
     CostQueryReq,
+    CustomRuntimeApplyReq,
+    CustomRuntimeApplyRes,
+    CustomRuntimeID,
     EndedCallSchemaForInsertWithStartedAt,
     FeedbackCreateReq,
     FileCreateReq,
@@ -538,6 +541,40 @@ class WeaveClient:
             pass
 
     ################ High Level Convenience Methods ################
+
+    @trace_sentry.global_trace_sentry.watch()
+    def register_custom_runtime(
+        self,
+        *,
+        name: str,
+        base_url: str,
+        runtime_ids: Iterable[str | CustomRuntimeID],
+        api_key_secret: str | None = None,
+        headers: Mapping[str, str] | None = None,
+    ) -> CustomRuntimeApplyRes:
+        """Register a custom runtime, replacing its configuration if it exists.
+
+        Runtime IDs omitted from this call are removed. String IDs use the
+        default maximum token count; pass ``CustomRuntimeID`` to set it explicitly.
+        Inference may use a prior configuration for up to 60 seconds after an
+        update.
+        """
+        normalized_runtime_ids = [
+            CustomRuntimeID(id=runtime_id)
+            if isinstance(runtime_id, str)
+            else runtime_id
+            for runtime_id in runtime_ids
+        ]
+        return self.server.custom_runtime_apply(
+            CustomRuntimeApplyReq(
+                project_id=self.project_id,
+                runtime_name=name,
+                base_url=base_url,
+                api_key_secret=api_key_secret,
+                headers=headers or {},
+                runtime_ids=normalized_runtime_ids,
+            )
+        )
 
     @trace_sentry.global_trace_sentry.watch()
     def save(self, val: Any, name: str, branch: str = "latest") -> Any:
--- a/weave/trace_server_bindings/remote_http_trace_server.py
+++ b/weave/trace_server_bindings/remote_http_trace_server.py
@@ -3,6 +3,7 @@ import io
 import logging
 from collections.abc import Iterator
 from typing import Any, cast
+from urllib.parse import quote
 from zoneinfo import ZoneInfo
 
 import httpx
@@ -1471,6 +1472,24 @@ class RemoteHTTPTraceServer(TraceServerClientInterface):
             tsi.DatasetDeleteRes,
             method="DELETE",
             params=params,
+        )
+
+    @validate_call
+    def custom_runtime_apply(
+        self, req: tsi.CustomRuntimeApplyReq
+    ) -> tsi.CustomRuntimeApplyRes:
+        entity, project = from_project_id(req.project_id)
+        runtime_name = quote(req.runtime_name, safe="")
+        url = f"/v2/{entity}/{project}/runtimes/{runtime_name}"
+        body = tsi.CustomRuntimeApplyBody.model_validate(
+            req.model_dump(exclude={"project_id", "runtime_name", "wb_user_id"})
+        )
+        return self._generic_request(
+            url,
+            body,
+            tsi.CustomRuntimeApplyBody,
+            tsi.CustomRuntimeApplyRes,
+            method="PUT",
         )
 
     @validate_call
--- a/sdks/node/package.json
+++ b/sdks/node/package.json
@@ -28,7 +28,7 @@
     "format": "prettier --write \"src/**/*.ts\" \"examples/**/*.ts\"",
     "run": "tsx",
     "prettier-check": "prettier --check \"src/**/*.ts\" \"examples/**/*.ts\"",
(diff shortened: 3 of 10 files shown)


weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -1473,6 +1474,24 @@ def dataset_delete(self, req: tsi.DatasetDeleteReq) -> tsi.DatasetDeleteRes:
...
+    @validate_call
+    def custom_runtime_apply(
+        self, req: tsi.CustomRuntimeApplyReq
+    ) -> tsi.CustomRuntimeApplyRes:
+        entity, project = from_project_id(req.project_id)
+        runtime_name = quote(req.runtime_name, safe="")
+        url = f"/v2/{entity}/{project}/custom-runtimes/{runtime_name}"
+        body = tsi.CustomRuntimeApplyBody.model_validate(
+            req.model_dump(exclude={"project_id", "runtime_name", "wb_user_id"})
+        )
```
tonyyli-wandb, line 1486: **suggestion (if-minor):** unless there's a lot of custom `WrapValidator` logic (or other bespoke customizations), model_validate_json/model_dump_json are typically faster if the roundtripping is necessasry
```suggestion
        body = tsi.CustomRuntimeApplyBody.model_validate_json(
            req.model_dump_json(exclude={"project_id", "runtime_name", "wb_user_id"})
        )
```

weave/trace/weave_client.py
```diff
@@ -539,6 +542,40 @@ def _on_wal_send(self, record_type: str, record: dict) -> None:
...
 
+    @trace_sentry.global_trace_sentry.watch()
+    def apply_custom_runtime(
+        self,
+        *,
+        name: str,
+        base_url: str,
+        runtime_ids: Sequence[str | CustomRuntimeID],
+        api_key_secret: str | None = None,
+        headers: dict[str, str] | None = None,
```
tonyyli-wandb: **nitpick:** really minor, but the argument type hints can be loosened here (to reduce noisy/spurious future typechecking errors)

```suggestion
        runtime_ids: Iterable[str | CustomRuntimeID],
        api_key_secret: str | None = None,
        headers: Mapping[str, str] | None = None,
```

weave/trace_server_bindings/remote_http_trace_server.py
mattliu-mygit: I think this should be fine. We currently don't roundtrip since we already have a Python object (the request object we constructed) and this should be a bit faster. It also matches the patterns we have in other routes in this codebase.

tonyyli-wandb: tangential: not critical or blocking, but def curious what ends up making the python<->python round trip faster here

for (naive) context: [pydantic usually roundtrips python<->JSON faster](https://pydantic.dev/docs/validation/latest/concepts/performance/#in-general-use-model_validate_json-not-model_validatejsonloads), since most of that logic goes through Rust.  this is less true for pydantic classes that heavily customize validation/serialization more though, and I'm def still getting familiar with the existing code

mattliu-mygit: Hmmm yea I did some light research with ChatGPT. I think the definitive optimal solution is to not have this time transitioning in the first place. 

See https://chatgpt.com/share/e/6a680b6d-102c-8012-a611-e8f9ffbd681b

However, I believe that it makes sense for a separate request and response model here since they are slightly different. In that the body is richer.

weave/trace/weave_client.py
```diff
@@ -539,6 +542,40 @@ def _on_wal_send(self, record_type: str, record: dict) -> None:
...
+        default maximum token count; pass ``CustomRuntimeID`` to set it explicitly.
+        Inference may use a prior configuration for up to 60 seconds after an
+        update.
+        """
+        normalized_runtime_ids = [
+            CustomRuntimeID(id=runtime_id)
+            if isinstance(runtime_id, str)
+            else runtime_id
+            for runtime_id in runtime_ids
+        ]
```
tonyyli-wandb, line 567: **suggestion (non-blocking):** optional, but I'd consider absorbing this into the validator logic for `CustomRuntimeID` if there's no ambiguity.  E.g.

```python3
class CustomRuntimeID(...):

    ...

    @model_validator(mode="before")
    def _wrap_str_id(cls, v: Any) -> Any:
        return cls(id=v) if isinstance(v, str) else v

    ...
```

This way you can pass `runtime_ids` as-is and delegate to the model validator here -- and anywhere else -- to coerce the argument appropriately.

Reduces boilerplate here and anywhere else CustomRuntimeID is used, tradeoff is that the conversion is less explicit/transparent at the call sites themselves -- hence totally optional

tonyyli-wandb (approved):
extra optional suggestion, but LGTM

weave/trace/weave_client.py
mattliu-mygit: Hmmm makes sense; thanks for calling it out! But here CustomRuntimeID is also part of the shared REST schema. Moving coercion there would silently allow string IDs at every validation boundary while OpenAPI still documents an object. I think making it clear for users to call this API with just the CustomRuntimeID is clearest here.

I also asked Codex to test the suggestion: returning cls(id=v) from the before validator fails nested Pydantic validation. Returning {"id": v} works, but creates the undocumented API behavior above.

weave/trace/weave_client.py
```diff
@@ -539,6 +542,40 @@ def _on_wal_send(self, record_type: str, record: dict) -> None:
 
     ################ High Level Convenience Methods ################
 
+    @trace_sentry.global_trace_sentry.watch()
+    def apply_custom_runtime(
```
neutralino1: Why "apply"? Should it be "create"?

weave/trace/weave_client.py
tonyyli-wandb: > Moving coercion there would silently allow string IDs at every validation boundary while OpenAPI still documents an object.

yeah I figured, that's a hell of a blast radius so all the more reason to keep this optional.

EDIT: To be clear, I'm totally OK with ignoring the orig suggestion

> I also asked Codex to test the suggestion: returning cls(id=v) from the before validator fails nested Pydantic validation. Returning {"id": v} works

This is actually a little surprising.  Gut guess is that we don't [configure models](https://pydantic.dev/docs/validation/latest/api/pydantic/config/#pydantic.config.ConfigDict.from_attributes) with `from_attributes=True`?  Need to verify, and not critical at any rate :)

tonyyli-wandb: **note(non-blocking):** re the "undocumented" OpenAPI concern -- for future reference, this is almost always solvable (edit: when justified ofc) :) 

there are weird complicated corner cases, but usually the existing API lets you tweak the generated JSON schemas for fields, models, etc if it's necessary for accuracy: https://pydantic.dev/docs/validation/dev/concepts/json_schema/#customizing-json-schema

weave/trace/weave_client.py
mattliu-mygit: Following the similar past behavior in the UI for create/update, I'm keeping this runtime creation flow simple with a single PUT endpoint. Since we're using this for both create and update, I thought "apply" is a bit more apt for applying creation/update requests.

weave/trace/weave_client.py
mattliu-mygit: Makes sense!

weave/trace/weave_client.py
neutralino1: As a user, I don't know what "apply" means. Apply to what?
`register_custom_runtime` is what I would expect. That still covers both create and update.

mattliu-mygit: Makes sense. Yea ``egister_custom_runtime`` is much better naming. Applied!

neutralino1 (approved):
Thank you!

PR #7623, opened 2026-07-23
fix(weave): avoid duplicate OpenAI Agents telemetry

ro31337:
## JIRA Issue(s)

[WB-37905](https://coreweave.atlassian.net/browse/WB-37905)

## Description

PR [#7570](https://github.com/wandb/weave/pull/7570) restored automatic tracing for direct OpenAI calls. When Agents was already loaded, it suppressed OpenAI patching to avoid duplicate records.

I checked whether Agents would always load first. It does not. This valid order installs the OpenAI wrapper before Agents:

```python
import openai
import weave

weave.init("entity/project")

import agents
```

I reproduced this order in a fresh process with the same patch and import-hook steps as `weave.init()`. I then ran `Runner.run()` once against a mocked OpenAI response.

That one model request produced two records. One was an `openai.responses.create` call in Weave Calls. The other was an OpenTelemetry (OTel) `chat` span in the Agents view. Both had the same OpenAI response ID, which confirmed that they described the same request.

The cause was import timing. The late Agents import installed its OTel processor and suppressed future OpenAI patch attempts. It could not remove the OpenAI wrapper that was already installed, so both paths recorded the request.

The extra record inflates call counts. It can also overstate token and cost totals when an aggregate includes both records. Disabling OpenAI tracing for the whole process would hide direct OpenAI calls made outside Agents, so that is not a safe fix.

## Before

The screenshots below come from one agent run in the import order above, against the live API. The agent calls a tool, so it makes two model requests, and a separate direct OpenAI request runs after it. Traces holds three calls: the boxed pair are the duplicates of the two agent model requests, and `direct_openai_call` is the direct request that belongs there.

Traces before the fix: direct_openai_call plus two duplicate openai.responses.create calls

The Agents view already records those same two requests as `chat` spans, which is where the duplication comes from.

Agents spans before the fix: seven spans, two of them chat

## After

Same script, same import order. The two duplicate calls are gone and the direct OpenAI request is still traced.

Traces after the fix: only direct_openai_call

The Agents view is unchanged, so the agent run loses nothing.

Agents spans after the fix: the same seven spans

## Why this approach

Both integrations stay installed. During a supported Agents model request, the processor stores the current model span in OTel context. Only `AsyncResponses.create` and `AsyncCompletions.create` check this signal.

If the span is live, recording, and sampled, those methods call OpenAI directly. The Agents `chat` span records the request once. Otherwise, the existing Weave wrapper runs as before.

The signal belongs to the OTel context, not the whole process. Independent tasks stay traced. Child tasks created inside the model span inherit the signal.

This covers the built-in Responses and Chat Completions models, which make one OpenAI request per model span. Custom multi-request models are out of scope. Sync methods, other OpenAI endpoints, and `use_otel_v2=False` do not use this bypass.

## Testing

- Manual clean-process reproduction: 1 passed
- Live agent run against the API in both directions, screenshots above
- Local tests: 1,491 passed, 9 skipped
- Local lint, type, and schema checks: 7 passed
- GitHub check runs: 106 passed, 1 skipped

[WB-37905]: https://coreweave.atlassian.net/browse/WB-37905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rgao-coreweave (approved):
Could you please add screenshots of duplicated spans before and after the fix?

otherwise LGTM.

ro31337:
Added `## Before` / `## After` screenshots to the description, from a live agent run in that import order.

PR #7633, opened 2026-07-24
fix(weave): honor explicit wandb.is_turn over OTel conversation id

ro31337:
## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-37944

## Description

Weave was splitting one conversation into several separate turns. In the Threads view, one agent run showed up as duplicated, broken-up turns instead of one clean thread. That is the bug in #7066.

The worse part - some senders already say "this span is not a turn" by setting `wandb.is_turn=false`. We ignored it and made a turn anyway.

I thought that flag should just win. If a sender takes the trouble to say `false`, we should listen, not override it with a guess.

The fix is small. After the normal parsing, we read `wandb.is_turn` directly. If it is a real true/false, we use that.

So why was it broken? I dug in, and it was subtle. To decide if a span is a turn, Weave checks a few fields in order, and `gen_ai.conversation.id` is first. Just having a conversation id was enough to mark the span as a turn.

But that id is really just a session/thread id for [grouping messages](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-conversation-id). It does not mean the span is a turn. So using it that way is only a guess.

The explicit `false` was checked later. But a bare `false` looks "empty" to the shared parser, so it got dropped. The guess always won.

The clean fix was not to reorder or rewrite that parser (lots of other fields use it). I just read the one flag at the end and let it win.

Bottom line: before, `wandb.is_turn=false` was ignored and threads fell apart. Now it is respected and the turns stay whole. Spans that do not set the flag work exactly like before. Only a real true/false counts, not the text "false" and not 0/1. Nothing else changes - no producer, storage, schema, or UI.

## Testing

Extended three tests in `tests/trace_server/test_opentelemetry.py` for the explicit true/false cases, plus a full span that keeps its thread but gets no turn. All pass (71 passed, 1 pre-existing skip), and lint and type checks are clean.

### DIFF
--- a/weave/trace_server/opentelemetry/attributes.py
+++ b/weave/trace_server/opentelemetry/attributes.py
@@ -121,13 +121,18 @@ def get_weave_outputs(_: list[SpanEvent], attributes: dict[str, Any]) -> dict[st
     return parse_weave_values(attributes, OUTPUT_KEYS)
 
 
+# Read wandb.is_turn directly: parse_weave_values checks the conversation/session-id aliases first and drops a literal False
+def _explicit_is_turn(attributes: dict[str, Any]) -> bool | None:
+    value = get_attribute(attributes, "wandb.is_turn")
+    return value if isinstance(value, bool) else None
+
+
 # Custom attributes for weave to enable setting fields like wb_user_id otherwise unavailable in OTEL Traces
 def get_wandb_attributes(attributes: dict[str, Any]) -> dict[str, Any]:
     result = parse_weave_values(attributes, WB_KEYS)
-    # An explicit Boolean wandb.is_turn overrides the conversation/session-id turn
-    # fallback; read it directly since parse_weave_values drops a literal False.
-    explicit_is_turn = get_attribute(attributes, "wandb.is_turn")
-    if isinstance(explicit_is_turn, bool):
+    # An explicit Boolean wandb.is_turn overrides the conversation/session-id turn fallback
+    explicit_is_turn = _explicit_is_turn(attributes)
+    if explicit_is_turn is not None:
         result["is_turn"] = explicit_is_turn
     return result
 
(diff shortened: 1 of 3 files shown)


weave/trace_server/opentelemetry/attributes.py
```diff
@@ -123,7 +123,13 @@ def get_weave_outputs(_: list[SpanEvent], attributes: dict[str, Any]) -> dict[st
 
 # Custom attributes for weave to enable setting fields like wb_user_id otherwise unavailable in OTEL Traces
 def get_wandb_attributes(attributes: dict[str, Any]) -> dict[str, Any]:
-    return parse_weave_values(attributes, WB_KEYS)
+    result = parse_weave_values(attributes, WB_KEYS)
+    # An explicit Boolean wandb.is_turn overrides the conversation/session-id turn
+    # fallback; read it directly since parse_weave_values drops a literal False.
+    explicit_is_turn = get_attribute(attributes, "wandb.is_turn")
+    if isinstance(explicit_is_turn, bool):
```
gtarpenning: do we want to handle `"false"` too?

gtarpenning: super nit: could also make this like a little helper `_explicit_is_turn()` or something, would keep the `get_wandb_attributes` a bit cleaner

gtarpenning (approved):

weave/trace_server/opentelemetry/attributes.py
ro31337: my thinking is: no producer sends a string today. If they send a string, should we also accept `"False"`, `"0"`, `"no"`, etc? So I think we don't want to handle it, strong contract is actually the better preference, IMO

PR #7640, opened 2026-07-27
feat(weave): support Azure workload identity

zacharyblasczyk:
Weave trace file storage now falls back to `DefaultAzureCredential`, which enables AKS workload identity. A configured connection string still wins, followed by an account key. Workload identity is used only when neither is present.

Token credentials produce read-only user-delegation SAS URLs; connection strings and account keys stay on the existing account-key SAS path.

The storage URI remains `az://<account>/<container>`. This PR does not enable object storage in Helm; the chart switch is separate and defaults to off.

Deployments without `WF_FILE_STORAGE_URI` continue to store trace files in ClickHouse.

## Checks

- 6 focused tests passed.
- Ruff passed for the changed implementation and tests.
- The dependency lock was generated with Python 3.12.

Jira: WB-38011


weave/trace_server/file_storage_credentials.py
```diff
@@ -90,25 +98,25 @@ def get_gcp_credentials() -> GCPCredentials | None:
...
     if connection_string is not None:
         return AzureConnectionCredentials(connection_string=connection_string)
     access_key = environment.wf_storage_bucket_azure_access_key()
-    if access_key is None:
-        raise ValueError("Azure access key not set")
-    account_url = environment.wf_storage_bucket_azure_account_url()
-    return AzureAccountCredentials(access_key=access_key, account_url=account_url)
+    if access_key is not None:
+        account_url = environment.wf_storage_bucket_azure_account_url()
+        return AzureAccountCredentials(access_key=access_key, account_url=account_url)
```
amwarrier, line 124: nit: add a logger.info("using DefaultAzureCredential chain (no WF_FILE_STORAGE_AZURE_CONNECTION_STRING or WF_FILE_STORAGE_AZURE_ACCESS_KEY set)") in the default-credential branch of get_azure_credentials().

will help catching problems when its misconfigured/typo etc.

amwarrier (approved):

PR #7644, opened 2026-07-28
feat(weave): promote source/integration attribution to columns

gtarpenning:
## Summary

- Adds `source_name`, `source_version`, and `ingest_source` to `spans`, `call_parts` -> `calls_merged`, and `calls_complete` (migration 040), so source attribution is a column query instead of a JSON scan over `attributes_dump` / `custom_attrs_string`.
- Uses one resolution ladder in `trace_server/source_attribution.py`: explicit `weave.source.*` / integration attributes, then the OTel instrumentation scope, then `''`. Resource `service.*` and `telemetry.sdk.*` identity is intentionally excluded because it describes the producing service rather than its instrumentation.
- Records the authoritative server ingest path as `ingest_source`: direct call APIs write `weave`, while OTel exports write `otlp`. Client-supplied attributes and `otel_dump` cannot override it.
- Threads OTel scope and full wire attributes through `Span.as_dict()` so the spans and calls paths resolve the same producer, including Node's `weave.integration.*` spelling.
- Regenerates the tracked Node OpenAPI schema and TypeScript trace-server client from the current trace-service API.

## Testing

- Real ClickHouse source-attribution coverage across `spans`, `call_parts` / `calls_merged`, and `calls_complete`, including both OTel routing modes, resource-only spans, and a direct-call spoof attempt.
- In-memory and resolver coverage for precedence, empty sentinels, and authoritative ingest-source selection.
- Query-builder suites: 220 passed.
- Migration parser coverage for all shipped migrations.
- Node generated-client test, CommonJS typecheck, ESM typecheck, and Prettier.
- Full staged pre-push gate: ty, import-linter, Pyright, mypy, JSON validation, Fixit, Ruff lint, and Ruff format.

## Design notes

- Existing rows read `''` / NULL; there is no read-time fallback or backfill.
- `source_version` remains a plain `String`; `source_name` and `ingest_source` use `LowCardinality`.
- No skip index is added because a project typically emits only a few interleaved source names, so granules would rarely be skipped.
- `User-Agent` attribution remains out of scope; supporting it would require threading the header from the FastAPI layer.


ro31337:

PR #7645, opened 2026-07-28
fix(weave): align integration otel keys and retain instrumentation scope

gtarpenning:
## Summary
- Python `as_otel_attributes()` emitted `integration.*` while the node SDK emits `weave.integration.*`, so a consumer keyed on either spelling missed the other SDK's spans. Python now emits the `weave.integration.*` form `AGENTS.md:322` documents as canonical.
- Instrumentation scope was parsed and then discarded for every ingested span. `ScopeSpans.from_proto` was the only site threading it, and it is reachable only from `TracesData.from_proto`, which has no non-test callers; all three real ingest loops dropped it. `iter_proto_spans` now pairs each span with its enclosing scope in one place, so the three loops cannot disagree.
- Scope absence is a `Scope | None` mirroring `resource` rather than `""`, so "sender omitted the scope" and "sender sent an empty scope" stay distinguishable in the persisted `otel_dump`. This adds `attributes.otel_span.scope` as a permanent user-visible key.
- `gen_ai_aliases` -> `wire_aliases`: it already held non-`gen_ai` keys (`ERROR_TYPE` takes `error.type`), so the name carries the contract instead of a docstring carve-out.

No server-side ladder accepting both spellings: no weave server code consumes either key, so converging the producers now means a consumer needs one key rather than two forever. Note the old spelling is already persisted as custom attrs in real projects, so those will list both until the old spans age out, and a saved filter on `integration.name` stops matching new spans. #7644 should rebase and drop its dual-key rung.

## Testing
The scope assertion runs through the real `otel_export` path, not `TracesData.from_proto`; reverting the threading in `clickhouse_trace_server_batched.py` fails it with `assert None == {...}`. A parity test reads the node literal out of `sdks/node/src/genai/semconv.ts` so the SDKs cannot drift again; the import-linter forbids sharing the constant, so the literals are restated per the `google_adk/_semconv.py` precedent.

### DIFF
--- a/tests/integrations/claude_agent_sdk/claude_agent_sdk_otel_test.py
+++ b/tests/integrations/claude_agent_sdk/claude_agent_sdk_otel_test.py
@@ -63,15 +63,15 @@ def get_attrs(span: Any) -> dict[str, Any]:
 
 
 def check_integration_and_strip(attrs: dict[str, Any]) -> dict[str, Any]:
-    """Assert + remove the flattened integration.* provenance keys.
+    """Assert + remove the flattened weave.integration.* provenance keys.
 
     The agent OTel processor stamps integration provenance on every span; pop it
     here so the exact-shape assertions below stay focused on the GenAI semconv keys.
     """
-    assert attrs["integration.name"] == "claude_agent_sdk"
-    assert attrs["integration.version"]  # weave SDK version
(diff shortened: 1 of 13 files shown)


jtschoonhoven (approved):
You might want a stamp from the sdk team on this, but LGTM

PR #7646, opened 2026-07-28
chore(weave): ignore (and log) extra fields from `BaseModelStrict`

jtschoonhoven:
## Summary

Instead of raising an exception, log at WARNING level if an extra field is passed to BaseModelStrict.

Context: https://coreweave.slack.com/archives/C0880TX4U5V/p1785259342111189

### DIFF
--- a/weave/trace_server/common_interface.py
+++ b/weave/trace_server/common_interface.py
@@ -4,16 +4,44 @@ This module contains base classes and common types used by both
 trace_server_interface.py and http_service_interface.py to avoid circular dependencies.
 """
 
-from typing import Literal
+import logging
+from typing import Any, Literal
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 WB_USER_ID_DESCRIPTION = (
     "Do not set directly. Server will automatically populate this field."
 )
 
+logger = logging.getLogger(__name__)
 
-class SortBy(BaseModel):
+
+class BaseModelStrict(BaseModel):
+    """API model that tolerates and reports unknown additive fields."""
+
+    model_config = ConfigDict(extra="ignore")
+
+    @model_validator(mode="before")
+    @classmethod
+    def warn_on_extra_fields(cls, values: Any) -> Any:
+        if not isinstance(values, dict):
+            return values
+
+        known_fields = set(cls.model_fields)
+        known_fields.update(
+            field.alias for field in cls.model_fields.values() if field.alias is not None
+        )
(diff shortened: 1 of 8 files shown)


weave/trace_server/common_interface.py
```diff
@@ -6,20 +6,14 @@
...
 
 WB_USER_ID_DESCRIPTION = (
     "Do not set directly. Server will automatically populate this field."
 )
 
 
-class BaseModelStrict(BaseModel):
-    """Base model with strict validation that forbids extra fields."""
-
-    model_config = ConfigDict(extra="forbid")
```
gtarpenning, line 19: Can we make this `warn` instead of `forbid`

gtarpenning (changes requested):
i think we should keep the class and just warn

weave/trace_server/common_interface.py
jtschoonhoven: `warn` isn't a thing

gtarpenning: okay maybe `ignore` + a post validator that logs a warning

weave/trace_server/trace_server_interface.py
```diff
@@ -831,7 +830,7 @@ class ObjQueryReq(BaseModelStrict):
     )
 
 
-class ObjDeleteReq(BaseModelStrict):
+class ObjDeleteReq(BaseModel):
```
ro31337: what happens if a caller misspells `digests` on this delete request? (for example, send `digest`, without `s`). Could the ignored field leave `digests=None` and delete every version?

If a clients sends:

 ```json
   {
     "object_id": "my-object",
     "digest": ["version-2"]
   }
 ```

Before the PR - send an error back, after this PR - swallow and delete all.

Sounds like nit, but it's about mass-delete, so I wanna leave a comment here in case you want to improve that part.

weave/trace_server/common_interface.py
jtschoonhoven: Updated

weave/trace_server/trace_server_interface.py
jtschoonhoven: Yes, all digests would be deleted

jtschoonhoven: I think that's acceptable. If the client is making a DELETE request then there are lots of ways that a typo could be destructive. Clients should be careful with DELETE requests 🤷

gtarpenning (approved):

ro31337 (approved):

PR #7647, opened 2026-07-28
chore(weave): enforce top-level Python imports

gtarpenning:
## Summary

Enable Ruff's `PLC0415` (`import-outside-top-level`) check without changing intentional lazy, optional, circular, or test-isolation import behavior.

Before this change, the rule was selected through the `PLC` family but globally ignored, so the AGENTS.md import rule relied entirely on self-enforcement.

## Audit

A fresh `master` scan found 556 violations across 141 files:

- 370 in tests, including 235 in optional-integration tests
- 185 in production code, dominated by lazy optional dependencies and circular-import boundaries
- 1 in a benchmark script

This PR:

- removes 31 redundant nested imports whose identical binding already exists at module scope
- adds line-level `# noqa: PLC0415` suppressions to the remaining 525 baseline occurrences
- removes `PLC0415` from the global Ruff ignore list

Line-level suppressions are deliberate: unlike generated per-file ignores, they keep Ruff enforcement active everywhere in each affected file, so any new function-level import fails lint immediately.

## Behavior

An AST comparison against `master` found semantic changes in only six test files. Those changes are exactly the 31 redundant import removals; all production Python ASTs are unchanged.

Four apparent duplicate imports remain intentionally local because they are part of traced function bodies, import-patching coverage, or isolated subprocess execution.

## Validation

- `uvx ruff@0.15.5 check .`
- `uvx ruff@0.15.5 format --check .`
- `uvx ruff@0.15.5 check --select PLC0415 .`
- pre-push suite: ty, import-linter, pyright, mypy, Fixit, Ruff check/format
- targeted tests: 70 passed, 3 skipped

One additional OpenAI realtime assertion failed locally and was reproduced unchanged on the exact base commit; it is unrelated to this diff.


jtschoonhoven (approved):
😻

PR #7649, opened 2026-07-28
fix(weave): use standard custom runtime validation

mattliu-mygit:
## Summary

- Move runtime-name and runtime-ID checks to reusable Pydantic field validation.
- Trim leading/trailing whitespace, then reject empty values or identifiers containing internal whitespace.
- Keep duplicate-ID validation after normalization.
- Keep the combined storage-name limit in the Custom Runtime service as an `InvalidRequest` raised before writes.

## Why

Core route validation currently has to catch and reformat an internal `ValidationError` after combining its path and body. This keeps structural validation at the request boundary and the cross-field storage constraint at the domain boundary.

## Related change

- Core route cleanup: [wandb/core#49011](https://github.com/wandb/core/pull/49011)

## Verification

- `20 passed` in `tests/trace_server/test_custom_runtime_api.py`
- Weave pre-push type, import-boundary, lint, and format checks passed
- Regression coverage verifies normalization occurs before length and duplicate checks

### DIFF
--- a/weave/trace_server/trace_server_interface.py
+++ b/weave/trace_server/trace_server_interface.py
@@ -7,11 +7,13 @@ from pydantic import (
     BaseModel,
     ConfigDict,
     Field,
+    StringConstraints,
     field_serializer,
+    field_validator,
     model_validator,
     with_config,
 )
-from typing_extensions import TypedDict
+from typing_extensions import Self, TypedDict
 
 if TYPE_CHECKING:
     from opentelemetry.proto.trace.v1.trace_pb2 import ResourceSpans
@@ -33,6 +35,7 @@ from weave.trace_server.constants import (
     DEFAULT_CUSTOM_RUNTIME_MAX_TOKENS,
     MAX_OBJECT_NAME_LENGTH,
 )
+from weave.trace_server.errors import InvalidRequest
 from weave.trace_server.interface.query import Query
 
 # Re-exported from service_interface for backwards compatibility.
@@ -2703,10 +2706,28 @@ class DatasetDeleteRes(BaseModel):
     num_deleted: int = Field(..., description="Number of dataset versions deleted")
 
 
+CustomRuntimeName: TypeAlias = Annotated[
+    str,
+    StringConstraints(
+        strip_whitespace=True,
+        min_length=1,
+        max_length=MAX_OBJECT_NAME_LENGTH,
+        # Forbid whitespace and "::".
+        pattern=r"^[^\s:]*(?::[^\s:]+)*:?$",
+    ),
+]
+
+
 class CustomRuntimeID(BaseModelStrict):
-    id: str = Field(
-        description="Value sent in the OpenAI-compatible request model field"
-    )
+    id: Annotated[
+        str,
+        StringConstraints(
+            strip_whitespace=True,
+            min_length=1,
+            # Forbid whitespace.
+            pattern=r"^\S+$",
+        ),
+    ] = Field(description="Value sent in the OpenAI-compatible request model field")
     max_tokens: int = Field(
         default=DEFAULT_CUSTOM_RUNTIME_MAX_TOKENS,
         gt=0,
@@ -2728,33 +2749,31 @@ class CustomRuntimeApplyBody(BaseModelStrict):
         description="Complete desired list of IDs exposed by the endpoint"
     )
 
-
-class CustomRuntimeApplyReq(CustomRuntimeApplyBody):
-    project_id: str
-    runtime_name: str
-    wb_user_id: str | None = Field(None, description=WB_USER_ID_DESCRIPTION)
-
-    @model_validator(mode="after")
-    def validate_runtime_identity(self) -> "CustomRuntimeApplyReq":
-        if not self.runtime_name.strip():
-            raise ValueError("runtime_name must not be empty")
-        if len(self.runtime_name) > MAX_OBJECT_NAME_LENGTH:
-            raise ValueError(
-                f"runtime_name cannot exceed {MAX_OBJECT_NAME_LENGTH} characters"
-            )
-        if "::" in self.runtime_name:
-            raise ValueError("runtime_name cannot contain '::'")
-
+    @field_validator("runtime_ids")
+    @classmethod
+    def validate_unique_runtime_ids(
+        cls, runtime_ids: list[CustomRuntimeID]
+    ) -> list[CustomRuntimeID]:
         seen_ids: set[str] = set()
-        for runtime_id in self.runtime_ids:
-            if not runtime_id.id.strip():
-                raise ValueError("runtime ID must not be empty")
+        for runtime_id in runtime_ids:
             if runtime_id.id in seen_ids:
                 raise ValueError(f"duplicate runtime ID: {runtime_id.id}")
             seen_ids.add(runtime_id.id)
+        return runtime_ids
+
+
+class CustomRuntimeApplyReq(CustomRuntimeApplyBody):
+    project_id: str
+    runtime_name: CustomRuntimeName
+    wb_user_id: str | None = Field(None, description=WB_USER_ID_DESCRIPTION)
+
+    @model_validator(mode="after")
+    def validate_storage_name_length(self) -> Self:
+        for runtime_id in self.runtime_ids:
             if len(f"{self.runtime_name}/{runtime_id.id}") > MAX_OBJECT_NAME_LENGTH:
-                raise ValueError(
-                    f"runtime name and ID cannot exceed {MAX_OBJECT_NAME_LENGTH} characters"
+                raise InvalidRequest(
+                    f"Runtime name and ID together cannot exceed "
+                    f"{MAX_OBJECT_NAME_LENGTH} characters"
                 )
         return self
 
--- a/tests/trace_server/test_custom_runtime_api.py
+++ b/tests/trace_server/test_custom_runtime_api.py
@@ -290,16 +290,14 @@ def test_custom_runtime_apply_objects_work_with_existing_completion_lookup(
     "request_data",
     [
         {"runtime_name": "", "runtime_ids": [{"id": "runtime"}]},
+        {"runtime_name": "runtime name", "runtime_ids": [{"id": "runtime"}]},
         {"runtime_name": "name::part", "runtime_ids": [{"id": "runtime"}]},
         {"runtime_name": "a" * 129, "runtime_ids": [{"id": "runtime"}]},
         {"runtime_name": "runtime", "runtime_ids": [{"id": ""}]},
+        {"runtime_name": "runtime", "runtime_ids": [{"id": "runtime\tid"}]},
         {
             "runtime_name": "runtime",
             "runtime_ids": [{"id": "duplicate"}, {"id": "duplicate"}],
-        },
-        {
-            "runtime_name": "r" * 64,
-            "runtime_ids": [{"id": "i" * 64}],
         },
     ],
 )
@@ -311,6 +309,59 @@ def test_custom_runtime_apply_validates_names_before_writes(
             project_id=PROJECT_ID,
             base_url="https://agent.example.com/v1",
             **request_data,
+        )
+
+
+def test_custom_runtime_apply_trims_identifiers_before_validation() -> None:
+    request = tsi.CustomRuntimeApplyReq(
+        project_id=PROJECT_ID,
+        runtime_name=f" {'r' * 64} ",
+        base_url="https://agent.example.com/v1",
+        runtime_ids=[
+            {"id": " runtime "},
+        ],
+    )
+
+    assert request.runtime_name == "r" * 64
+    assert request.runtime_ids[0].id == "runtime"
+
+    with pytest.raises(ValidationError, match="duplicate runtime ID"):
+        tsi.CustomRuntimeApplyReq(
+            project_id=PROJECT_ID,
+            runtime_name="runtime",
+            base_url="https://agent.example.com/v1",
+            runtime_ids=[
+                {"id": "duplicate"},
+                {"id": " duplicate "},
+            ],
+        )
+
+
+def test_custom_runtime_id_constraints_are_reflected_in_schema() -> None:
+    id_schema = tsi.CustomRuntimeID.model_json_schema()["properties"]["id"]
+
+    assert id_schema["minLength"] == 1
+    assert id_schema["pattern"] == r"^\S+$"
+
+
+def test_custom_runtime_name_constraints_are_reflected_in_schema() -> None:
+    request_schema = tsi.CustomRuntimeApplyReq.model_json_schema()
+    name_schema = request_schema["properties"]["runtime_name"]
+
+    assert name_schema["minLength"] == 1
+    assert name_schema["pattern"] == r"^[^\s:]*(?::[^\s:]+)*:?$"
+
+
+def test_custom_runtime_request_validates_storage_name_length() -> None:
+    with pytest.raises(
+        InvalidRequest,
+        match="Runtime name and ID together cannot exceed 128 characters",
+    ):
+        tsi.CustomRuntimeApplyReq(
+            project_id=PROJECT_ID,
+            runtime_name="r" * 64,
+            base_url="https://agent.example.com/v1",
+            runtime_ids=[{"id": "i" * 64}],
         )
 
 


weave/trace_server/trace_server_interface.py
```diff
@@ -2728,35 +2751,23 @@ class CustomRuntimeApplyBody(BaseModelStrict):
...
-        if not self.runtime_name.strip():
-            raise ValueError("runtime_name must not be empty")
-        if len(self.runtime_name) > MAX_OBJECT_NAME_LENGTH:
-            raise ValueError(
-                f"runtime_name cannot exceed {MAX_OBJECT_NAME_LENGTH} characters"
-            )
-        if "::" in self.runtime_name:
-            raise ValueError("runtime_name cannot contain '::'")
-
+    @field_validator("runtime_ids")
```
mattliu-mygit, line 2752: Moved validation to request body model so to move validation from Core to FastAPI

weave/trace_server/trace_server_interface.py
```diff
@@ -2703,8 +2705,29 @@ class DatasetDeleteRes(BaseModel):
...
+    if "::" in runtime_name:
+        raise ValueError("runtime_name cannot contain '::'")
+    return runtime_name
+
+
+CustomRuntimeName: TypeAlias = Annotated[
+    str,
+    Field(max_length=MAX_OBJECT_NAME_LENGTH),
+    AfterValidator(_validate_custom_runtime_name),
+]
```
tonyyli-wandb, line 2718: **question:** Is the intent to keep leading/trailing whitespace if it exists, or do we want to also drop it in the validated value?

Asking since this informs the suggestion I was going to make -- it looks different depending on the answer. (Will follow up)

(There's also a weird pydantic v2 gotcha around string constraints that may or may not have been fixed, checking in the meantime)

weave/trace_server/trace_server_interface.py
```diff
@@ -2703,8 +2705,29 @@ class DatasetDeleteRes(BaseModel):
...
+    if not runtime_id.strip():
+        raise ValueError("runtime ID must not be empty")
+    return runtime_id
+
+
 class CustomRuntimeID(BaseModelStrict):
-    id: str = Field(
+    id: Annotated[str, AfterValidator(_validate_custom_runtime_id)] = Field(
         description="Value sent in the OpenAI-compatible request model field"
     )
```
tonyyli-wandb: **question:** same question here -- do we intend to keep vs trim leading/trailing whitespace

weave/trace_server/trace_server_interface.py
```diff
@@ -2728,35 +2751,23 @@ class CustomRuntimeApplyBody(BaseModelStrict):
...
+        cls, runtime_ids: list[CustomRuntimeID]
+    ) -> list[CustomRuntimeID]:
         seen_ids: set[str] = set()
-        for runtime_id in self.runtime_ids:
-            if not runtime_id.id.strip():
-                raise ValueError("runtime ID must not be empty")
+        for runtime_id in runtime_ids:
             if runtime_id.id in seen_ids:
                 raise ValueError(f"duplicate runtime ID: {runtime_id.id}")
             seen_ids.add(runtime_id.id)
```
tonyyli-wandb, line 2761: I realize this is existing code, but should duplicate IDs be fatal, or is it fine if we just automatically dedup IDs?

If it's find to just auto-dedup IDs, we can probably simplify this (and speed it up) with a set comprehension

weave/trace_server/custom_runtime.py
```diff
@@ -70,6 +72,14 @@ def apply_custom_runtime(
     )
 
 
+def _validate_storage_name_lengths(req: tsi.CustomRuntimeApplyReq) -> None:
+    for runtime_id in req.runtime_ids:
+        if len(f"{req.runtime_name}/{runtime_id.id}") > MAX_OBJECT_NAME_LENGTH:
+            raise InvalidRequest(
+                f"Runtime name and ID cannot exceed {MAX_OBJECT_NAME_LENGTH} characters"
+            )
```
tonyyli-wandb: Can this just be a `@model_validator(mode="after")` on the `CustomRuntimeApplyReq` class? https://pydantic.dev/docs/validation/latest/concepts/validators/#model-after-validator

tonyyli-wandb (commented):
Some initial comments + will follow up

weave/trace_server/trace_server_interface.py
mattliu-mygit: We want to trim out whitespaces. Thanks for the catch! Will update the code to make sure that strings are stripped of trailing/leading whitespaces.

weave/trace_server/trace_server_interface.py
mattliu-mygit: no intention to keep;  same as above!

weave/trace_server/trace_server_interface.py
mattliu-mygit: We should keep them fatal. The existing Custom Provider UI treats duplicate model/runtime IDs as validation failures and disables Save. I think we should preserve that behavior since users might want to keep their ids named in a specific way, which also follows patterns for other many other UXs where users input strings that should be globally unique (ie. many username creation UXs). Maybe they even added a duplicate id or forgot to modify one from copy/paste. I think instead of failing at this level, there should be some UI-level check for duplication before form submission, but that can be done as a followup.

weave/trace_server/custom_runtime.py
mattliu-mygit: A model validator would run when Core constructs CustomRuntimeApplyReq inside the route, after FastAPI has separately validated the URL path and body. This would move the validation back outside FastAPI, breaking from the pattern other routes follow and forcing us to handle the error in the Core code again.

weave/trace_server/trace_server_interface.py
mattliu-mygit: Actually, will reject! UI already does trimming for us. If they try to create a custom runtime with whitespaces via SDK, we should fail and call it out rather than mutate their input.

tonyyli-wandb: I mean -- if the UI already trims for us, that seems to signal the intent is to normalize by trimming.  Does it hurt for us to align w/the same intended behavior here?

I get there's an argument to avoid duplication and fragmentation of validation logic, but -- in this case --  if the intent is to trim anyway (backend or frontend), gut feeling is it doesn't hurt to follow the same logic here.

Also, this acts as a safeguard on existing behavior in case e.g. some freak bug removes the trimming logic from the UI

mattliu-mygit: Makes sense; I'll update the code to trim.

weave/trace_server/custom_runtime.py
tonyyli-wandb: > This would move the validation back outside FastAPI

not totally sure I understand this yet, might be worth an offline chat/huddle to clear up, will follow up

weave/trace_server/trace_server_interface.py
```diff
@@ -2703,10 +2706,37 @@ class DatasetDeleteRes(BaseModel):
...
+    if "::" in runtime_name:
+        raise ValueError("runtime_name cannot contain '::'")
+    return runtime_name
+
+
+CustomRuntimeName: TypeAlias = Annotated[
+    str,
+    StringConstraints(strip_whitespace=True, max_length=MAX_OBJECT_NAME_LENGTH),
+    AfterValidator(_validate_custom_runtime_name),
+]
```
tonyyli-wandb, line 2718: suggestion: good call using `StringConstraints`!  We probably don't need `_validate_custom_runtime_name` considering it's only being used to exclude (a) whitespace and (b) the `::` pattern: 

```suggestion
CustomRuntimeName: TypeAlias = Annotated[
    str,
    StringConstraints(
        strip_whitespace=True, 
        min_length=1, 
        max_length=MAX_OBJECT_NAME_LENGTH,
        pattern=r"^(?!.*(\s|::)).*$",
    ),
]
```

weave/trace_server/trace_server_interface.py
```diff
@@ -2703,10 +2706,37 @@ class DatasetDeleteRes(BaseModel):
...
+
 class CustomRuntimeID(BaseModelStrict):
-    id: str = Field(
-        description="Value sent in the OpenAI-compatible request model field"
-    )
+    id: Annotated[
+        str,
+        StringConstraints(strip_whitespace=True),
+        AfterValidator(_validate_custom_runtime_id),
+    ] = Field(description="Value sent in the OpenAI-compatible request model field")
```
tonyyli-wandb, line 2730: **suggestion:** similarly here, we probably don't need the `_validate_custom_runtime_id` helper



```suggestion
    id: Annotated[
        str,
        StringConstraints(
            strip_whitespace=True,
            min_length=1,
            pattern=r"^\S+$",  # inner whitespace not allowed
        ),
    ] = Field(description="Value sent in the OpenAI-compatible request model field")
```

weave/trace_server/trace_server_interface.py
tonyyli-wandb: > move validation from Core to FastAPI

I think I'm missing something about the architecture here -- not totally sure I understand this either but curious what this means

Separately, makes sense to me to collocate the field validator with the field/model 👍

weave/trace_server/trace_server_interface.py
mattliu-mygit: Makes sense!

weave/trace_server/trace_server_interface.py
mattliu-mygit: Makes sense!

weave/trace_server/custom_runtime.py
tonyyli-wandb: from offline discussion -- we think it makes sense to absorb this into `CustomRuntimeApplyReq` via `@model_validator(mode="after")`, but will leave this as an optional/non-blocking suggestion in case there are complications

mattliu-mygit: Done!

weave/trace_server/trace_server_interface.py
```diff
@@ -2703,10 +2706,26 @@ class DatasetDeleteRes(BaseModel):
     num_deleted: int = Field(..., description="Number of dataset versions deleted")
 
 
+CustomRuntimeName: TypeAlias = Annotated[
+    str,
+    StringConstraints(
+        strip_whitespace=True,
+        min_length=1,
+        max_length=MAX_OBJECT_NAME_LENGTH,
+        pattern=r"^[^\s:]*(?::[^\s:]+)*:?$",
```
tonyyli-wandb, line 2716: **suggestion:** Probably worth double-checking this pattern still does what you're expecting with some extra test cases (eg leading/trailing "::", " ::", ":: ", etc.

No need to go overboard, but the pattern itself is kind of opaque. I think my original suggestion had a bug in it, so this one is probably more correct, but it's harder to tell from a quick inspection.

Regardless, I'd definitely add a quick comment to clarify intent, in case of future bugs


```suggestion
        # Forbid whitespace, "::"
        pattern=r"^[^\s:]*(?::[^\s:]+)*:?$",
```

weave/trace_server/trace_server_interface.py
```diff
@@ -2703,10 +2706,26 @@ class DatasetDeleteRes(BaseModel):
...
 class CustomRuntimeID(BaseModelStrict):
-    id: str = Field(
-        description="Value sent in the OpenAI-compatible request model field"
-    )
+    id: Annotated[
+        str,
+        StringConstraints(
+            strip_whitespace=True,
+            min_length=1,
+            pattern=r"^\S+$",
```
tonyyli-wandb, line 2728: **suggestion:** this one's easier to read, but a quick comment is still a good idea:

```suggestion
            # Forbid whitespace
            pattern=r"^\S+$",
```

weave/trace_server/trace_server_interface.py
```diff
@@ -2728,33 +2747,31 @@ class CustomRuntimeApplyBody(BaseModelStrict):
         description="Complete desired list of IDs exposed by the endpoint"
     )
 
+    @field_validator("runtime_ids")
```
tonyyli-wandb, line 2752: **suggestion:** I think "after" is the default mode anyway for field validators, but for clarity

```suggestion
    @field_validator("runtime_ids", mode="after")
```

weave/trace_server/trace_server_interface.py
```diff
@@ -2728,33 +2747,31 @@ class CustomRuntimeApplyBody(BaseModelStrict):
...
+    @classmethod
+    def validate_unique_runtime_ids(
+        cls, runtime_ids: list[CustomRuntimeID]
+    ) -> list[CustomRuntimeID]:
+        seen_ids: set[str] = set()
+        for runtime_id in runtime_ids:
+            if runtime_id.id in seen_ids:
+                raise ValueError(f"duplicate runtime ID: {runtime_id.id}")
+            seen_ids.add(runtime_id.id)
+        return runtime_ids
```
tonyyli-wandb, line 2762: **suggestion:** the happy path can be checked much faster.  

If it fails, then we can worry about the (more expensive) step of identifying the duplicate ids.

```suggestion
    @classmethod
    def validate_unique_runtime_ids(
        cls, ids: list[CustomRuntimeID]
    ) -> list[CustomRuntimeID]:
        if len(set(obj.id for obj in ids)) != len(ids):
            counts = Counter(obj.id for obj in ids)
            dup_ids = [v for v, n in counts.items() if n > 1]
            raise ValueError(f"duplicate runtime IDs: {dup_ids!r}")
        return ids
```

weave/trace_server/trace_server_interface.py
```diff
@@ -2728,33 +2747,31 @@ class CustomRuntimeApplyBody(BaseModelStrict):
...
-            raise ValueError("runtime_name must not be empty")
-        if len(self.runtime_name) > MAX_OBJECT_NAME_LENGTH:
-            raise ValueError(
-                f"runtime_name cannot exceed {MAX_OBJECT_NAME_LENGTH} characters"
-            )
-        if "::" in self.runtime_name:
-            raise ValueError("runtime_name cannot contain '::'")
-
-        seen_ids: set[str] = set()
+    def validate_storage_name_length(self) -> "CustomRuntimeApplyReq":
```
tonyyli-wandb: **suggestion:** return `Self` since this method can be inherited by subclasses.  Also, it's just shorter 🙂 

Note: if we're supporting Python < 3.11 here, import `Self` from `typing_extensions` rather than `typing` 

```suggestion
    def validate_storage_name_length(self) -> Self:
```

weave/trace_server/trace_server_interface.py
```diff
@@ -2728,33 +2747,31 @@ class CustomRuntimeApplyBody(BaseModelStrict):
...
-                raise ValueError("runtime ID must not be empty")
-            if runtime_id.id in seen_ids:
-                raise ValueError(f"duplicate runtime ID: {runtime_id.id}")
-            seen_ids.add(runtime_id.id)
             if len(f"{self.runtime_name}/{runtime_id.id}") > MAX_OBJECT_NAME_LENGTH:
-                raise ValueError(
-                    f"runtime name and ID cannot exceed {MAX_OBJECT_NAME_LENGTH} characters"
+                raise InvalidRequest(
+                    f"Runtime name and ID together cannot exceed "
+                    f"{MAX_OBJECT_NAME_LENGTH} characters"
```
tonyyli-wandb, line 2776: **suggestion:** I can't make an inline suggestion on the full block, but this is more performant and avoids the nested if-statement with something like

```python3
if any(
    len(f"{self.runtime_name}/{r.id}") > MAX_OBJECT_NAME_LENGTH
    for r in self.runtime_ids
):
    raise InvalidRequest(...)
```

tonyyli-wandb (approved):
looks much better, thanks!  Left some follow up suggestions, but preapproving to unblock

weave/trace_server/trace_server_interface.py
mattliu-mygit: Makes sense!

weave/trace_server/trace_server_interface.py
mattliu-mygit: Sounds good

weave/trace_server/trace_server_interface.py
mattliu-mygit: I think “After” is already the default and nearby Weave validators conventionally omit it so I think it should be ok

weave/trace_server/trace_server_interface.py
mattliu-mygit: Sounds good!

weave/trace_server/trace_server_interface.py
mattliu-mygit: I think this should be ok since we shouldn't be dealing with large lists of runtime ids (though I agree that it's more performant for large lists). I personally feel like it's a bit overkill for small lists of ids that we're expecting for custom runtimes.

weave/trace_server/trace_server_interface.py
mattliu-mygit: I agree that it's more performant, but in our case, we're not expecting many ids. I'm biasing a bit more towards the current code since it's a bit more readable.

PR #7652, opened 2026-07-28
fix(weave_ts): WB-38082

ro31337:
## JIRA Issue(s)

- Fixes [WB-38082](https://coreweave.atlassian.net/browse/WB-38082)

[WB-38082]: https://coreweave.atlassian.net/browse/WB-38082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdks/node/src/weaveClient.ts
+++ b/sdks/node/src/weaveClient.ts
@@ -445,6 +445,27 @@ type CompletedCallParams = StartedCallSchemaForInsert & {
 // We count characters item by item, and try to limit batches to about this size.
 const MAX_BATCH_SIZE_CHARS = 10 * 1024 * 1024;
 
+// A plain object's prototype is `Object.prototype`, whose own prototype is
+// null. Anything with a deeper chain than that is an instance of some class.
+// Checked this way rather than comparing against `Object.prototype` directly so
+// objects built in another realm (a `vm` context, a worker) still count as
+// plain.
+function isPlainObject(val: object): boolean {
+  const proto = Object.getPrototypeOf(val);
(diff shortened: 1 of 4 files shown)


sdks/node/src/weaveClient.ts
```diff
@@ -1826,7 +1874,11 @@ export class WeaveClient {
       inputs['self'] = thisArg;
     }
     if (parameterNames === 'useParam0Object') {
-      inputs = {...inputs, ...params[0]};
+      // Spreading an instance would lift its fields into the inputs before the
+      // serializer sees it, so record it under a name instead.
+      inputs = isUnownedInstance(params[0])
+        ? {...inputs, arg0: params[0]}
```
gtarpenning, line 1880: will this `arg0` appear in user-space? they will see this in their traces in the UI right?

ro31337: I checked all 9 places in the SDK that use useParam0Object: evaluation.ts, evaluationLogger.ts, googleGenAI.ts, and openai.ts - alll of them pass a plain `{ ... }` object as the first argument (request with fields like `model`, `contents`, or something like `{self: scorer, output: ...}`).

So none at themoment passes a class instance, so no integration hits this branch.

SO here it only fires when a user declares `useParam0Object` on their own op and calls it with an object made with `new` (class instance pretty much). Even then `arg0` is not something new, an op with no declared parameter names already shows `arg0` in traces today.

rgao-coreweave (approved):
👍👍👍 💯💯💯

PR #7642, opened 2026-07-27
refactor(weave): migrate Claude Agent SDK tracing to GenAI SDK

rgao-coreweave:
## Summary

- Migrate the Python Claude Agent SDK OTel integration to Weave GenAI `Conversation`, `Turn`, `LLM`, and `Tool` handles.
- Preserve conversation IDs, model and tool spans, usage, errors, integration metadata, and custom agent names.
- Keep `gen_ai.provider.name` on child `chat` spans only. The root `invoke_agent` Turn is an internal span and has no single provider because one turn can use multiple model providers.
- Leave the legacy calls-based integration unchanged.

## Standards alignment

The root Turn follows the current OpenTelemetry internal-agent convention, which omits `gen_ai.provider.name`: https://github.com/open-telemetry/semantic-conventions-genai/pull/289

## Stack

- Base of the Python stack.
- Followed by #7665, #7666, #7667, and #7660.

## Testing

- `uv run --no-sync --group test python -m pytest tests/integrations/claude_agent_sdk/claude_agent_sdk_otel_test.py -q` (15 passed)
- `nox --no-install -e lint` (ty, import-linter, pyright, mypy, Ruff, and formatting passed before restack)

## Breaking changes

None.

PR #7654, opened 2026-07-28
feat(weave): add provider name to turns

rgao-coreweave:
## Summary

- Add optional `provider_name` to the Python `Turn` model.
- Thread it through `Conversation.start_turn()`, module-level `start_turn()`, `Turn.record()`, and `log_turn()`.
- Emit it as `gen_ai.provider.name` on root `invoke_agent` spans in both streaming and batch flows.
- Add focused coverage for each public entry point.

## Testing

- `uv run --no-sync --group test python -m pytest tests/conversation/test_conversation_otel.py -q` (179 passed)
- `uvx ruff check weave/conversation/conversation.py tests/conversation/test_conversation_otel.py`
- `uvx ruff format --check weave/conversation/conversation.py tests/conversation/test_conversation_otel.py`

## Breaking changes

None. The new parameter and field are optional and default to an empty string.

## Related

Prerequisite for #7642, which uses the typed parameter in the Claude Agent SDK integration.

### DIFF
--- a/weave/conversation/conversation.py
+++ b/weave/conversation/conversation.py
@@ -961,6 +961,7 @@ class Turn(_SpanBase):
     """
 
     agent_name: str = ""
+    provider_name: str = ""
     model: str = ""
     agent_id: str = ""
     agent_description: str = ""
@@ -1078,6 +1079,7 @@ class Turn(_SpanBase):
         output_messages: list[Message] | None = None,
         system_instructions: list[str] | None = None,
         agent_name: str | None = None,
+        provider_name: str | None = None,
         model: str | None = None,
         agent_id: str | None = None,
         agent_description: str | None = None,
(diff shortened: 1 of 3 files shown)


weave/conversation/conversation.py
```diff
@@ -1148,6 +1152,7 @@ def _build_attrs(
             agent_name=self.agent_name,
             conversation_id=conversation_id,
             conversation_name=conversation_name,
+            provider_name=self.provider_name,
```
ro31337, line 1155: Seems the current standard has removed this label? https://github.com/open-telemetry/semantic-conventions-genai/pull/289

Curious about what should `gen_ai.provider.name` represent on this internal `invoke_agent` span when one turn can use multiple model providers?

could setting it here, for example, conflict with the current OTel convention, which removed this attribute from internal agent spans?

rgao-coreweave: oh true, good findings!

From PR's design, the turn's provider name should propagate down to the children spans, and the children spans can override the provider name. 

From current state, it wouldn't make sense to show this info at the turn level since one turn can clearly use multiple model providers.

I am okay dropping this PR as well!

PR #7656, opened 2026-07-29
security(weave): validate ORM field qualifiers

gtarpenning:
## Summary

- require table-qualified ORM fields to use simple SQL identifiers
- require the qualified field portion to exactly match an allowed column
- preserve valid qualified fields while rejecting unsafe prefixes and suffixes

## Testing

- `uv run --group test python -m pytest tests/trace_server/test_orm.py -q` (31 passed)
- `uvx ruff check weave/trace_server/orm.py tests/trace_server/test_orm.py`
- `uvx ruff format --check weave/trace_server/orm.py tests/trace_server/test_orm.py`
- `uv run ty check weave/trace_server/orm.py`
- `git diff --check`

## Breaking changes

None for valid field names. Previously accepted malformed qualified fields now fail validation.

### DIFF
--- a/weave/trace_server/orm.py
+++ b/weave/trace_server/orm.py
@@ -15,6 +15,7 @@ from weave.trace_server.errors import InvalidFieldError
 from weave.trace_server.interface import query as tsi_query
 
 param_builder_count = 0
+_SQL_IDENTIFIER_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 
 
 class ParamBuilder:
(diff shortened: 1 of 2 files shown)


jtriggscw (approved):

weave/trace_server/orm.py
```diff
@@ -761,18 +762,11 @@ def _transform_external_field_to_internal_field(
...
-        field not in all_columns
-        and field != "*"
-        and field.lower() != "count(*)"
-        and not any(
-            # Check if field starts with a valid column name as prefix
-            unprefixed_field.lower().startswith(col_name.lower() + ".")
-            for col_name in all_columns
-        )
-    ):
+    if field not in all_columns and field != "*" and field.lower() != "count(*)":
```
tonyyli-wandb, line 769: **suggestion (non-blocking):** 
```suggestion
    if field not in {*all_columns, "*"} and field.lower() != "count(*)":
```

tonyyli-wandb (approved):

weave/trace_server/orm.py
tonyyli-wandb: whoops -- optional suggestion anyway

gtarpenning: will get to this in a bit

PR #7664, opened 2026-07-29
fix(weave): WB-38110

ro31337:
## JIRA Issue(s)

- Fixes [WB-38110](https://coreweave.atlassian.net/browse/WB-38110)

[WB-38110]: https://coreweave.atlassian.net/browse/WB-38110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rgao-coreweave (approved):
A credential inside a JSON string is not redacted, and parses are only the *top level*, so JSON string nested one level deeper is also not redacted. For example `tool-call arguments: tool_calls[].function.arguments`.

 LGTM for this current scope, but might need to cover more edge cases.

👍👍👍

weave/trace_server/opentelemetry/genai_extraction.py
```diff
@@ -465,6 +466,34 @@ def _flatten_attrs(attrs: dict[str, Any], prefix: str = "") -> list[tuple[str, A
...
+    every one of them in ``raw_span_dump``, and the attribute-derived columns read
+    the same values. That includes the output ones, so a credential-shaped name
+    inside a structured output is replaced too.
+
+    Runs before ``strip_inline_blobs_from_span``, not after: that step replaces a
+    long base64-shaped value with a ref, so this pass has to see the value while
+    it is still inline. Runs outside the caller's file-storage guard as well --
+    the blob strip needs a trace server, this does not.
+    """
+    span.attributes = redact_sensitive_keys(span.attributes)
```
gtarpenning, line 488: i think it's still possible for key:value pairs to slip through if they aren't exactly in the shape that we expect. is it worthwhile to add a test for:
```
gen_ai.input.messages = "[{\"openai_api_key\":\"sk-test-marker\"}]"
```

ro31337: I added a test, tests are OK, will push soon after local verification

PR #7678, opened 2026-08-03
feat(weave): instrument invoke_model_with_response_stream for token tracking

### LINKED ISSUE #5767: [Feature]: Patch bedrock client method `invoke_model_with_streaming_response`
### Description

_I did not find any existing issues when searching "invoke_model_with_streaming_response"._

# Feature request

The current Bedrock patch client message [only patches invoke_model](https://github.com/wandb/weave/blob/627819e63709bf8718ecaaf0dd395f501a986616/weave/integrations/bedrock/bedrock_sdk.py#L264-L268). It would be great if token usage and costs could also be automatically tracked for [invoke-model-with-response-stream ](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model_with_response_stream.html#invoke-model-with-response-stream) as well

### Suggested Solution

I think a possible first place to look would be patching the generator returned by `invoke_model_with_response_stream` t check each chunk into and when the final chunk is returned grab / finish collecting the token usage stats. 

### Weave Project Link

https://wandb.ai/feedback-aide/dev/weave

### Screenshots or Mockups

The following two screenshots are from the same overall trace. No automatic spans are created for calls to `model_invoke_with_response_stream`.

**Two calls to model_invoke**
Image

**One call to model_invoke and one call to model_invoke_with_response_stream**
Image

Aftabbs:
Fixes #5767

## Summary

`invoke_model_with_response_stream` was the only bedrock-runtime method
with no weave tracing. This PR closes that gap with the same pattern
already used for `invoke_model` and `invoke_agent`.

### What changed

**`weave/integrations/bedrock/bedrock_sdk.py`**

- `postprocess_output_invoke_stream` — consumes the `body` EventStream
  before the call returns, buffers all events, puts a generator of the
  original events back on `outputs["body"]` (so callers can still
  iterate it), and returns a logged copy with the concatenated content
  and event count. `ResponseMetadata` is preserved on the original
  `outputs` dict for the finish handler.
- `bedrock_on_finish_invoke_stream` — reads token counts from the
  `x-amzn-bedrock-input-token-count` / `x-amzn-bedrock-output-token-count`
  HTTP response headers, matching the pattern in `bedrock_on_finish_invoke`.
- `_patch_invoke_model_stream` — wires the op with `postprocess_inputs_invoke`
  (reused from `invoke_model`), the new `postprocess_output` and finish handler.
- `patch_client` — calls `_patch_invoke_model_stream` when the method
  is present on the client (guarded with `hasattr` so it degrades
  gracefully on older boto3 versions).

**`tests/integrations/bedrock/bedrock_test.py`**

- `MOCK_INVOKE_STREAM_EVENTS` — two chunk events with JSON text bytes.
- `mock_invoke_stream_make_api_call` — returns a dict with
  `ResponseMetadata` (including the token-count headers) and `body`
  as an iterator over the mock events.
- `test_bedrock_invoke_model_stream` — verifies that after calling
  `invoke_model_with_response_stream` the caller can still consume the
  stream, the trace is captured, and token counts are populated in the
  call summary.

## Test plan

```
pytest tests/integrations/bedrock/bedrock_test.py --trace-server=fake -v
```

All 13 existing tests still pass; the 3 ARN-parametrized errors are a
pre-existing Windows path-length limitation unrelated to this change.

### DIFF
--- a/tests/integrations/bedrock/bedrock_test.py
+++ b/tests/integrations/bedrock/bedrock_test.py
@@ -386,6 +386,19 @@ MOCK_INVOKE_AGENT_RESPONSE = {
     "sessionId": "test-session",
 }
 
+MOCK_INVOKE_STREAM_EVENTS = [
+    {
+        "chunk": {
+            "bytes": b'{"type":"content_block_delta","delta":{"type":"text_delta","text":"Hello"}}'
+        }
+    },
+    {
+        "chunk": {
+            "bytes": b'{"type":"content_block_delta","delta":{"type":"text_delta","text":" World"}}'
+        }
+    },
+]
+
 # Original botocore _make_api_call function
 orig = botocore.client.BaseClient._make_api_call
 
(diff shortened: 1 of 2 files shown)


Aftabbs:

PR #7686, opened 2026-08-04
fix(weave_ts): label a fetched image with its real format

ro31337:
## JIRA Issue(s)

[WB-38335](https://coreweave.atlassian.net/browse/WB-38335)

## Description

Right now, if you publish an image from Python and then read it back with the TypeScript client, the format is wrong. A JPEG comes back as PNG, and so does a WEBP. The bytes are right, only the label is wrong. That is a problem for any code that trusts the label. You end up naming the file `photo.png` when a JPEG is inside it, or serving the bytes as `image/png` when they are not a PNG.

This sits on top of #7685, which is what makes `get()` return a real image at all. That PR always labels the image `png`.

The obvious fix is to look at the bytes, because a JPEG and a PNG start with different magic numbers. We do not need that. Python saves the image under its own name: `image.png`, `image.jpg` or `image.webp`. That name comes down together with the file, we were just throwing it away.

So now we read the name and believe it. `ImageType` accepts `png`, `jpeg` and `webp`. The file on disk says `jpg` while the label says `jpeg`, because we build the content type out of the label (`image/jpeg`). An extension we do not know keeps `png`, same as today, so nothing new can break.

Publishing did not change. It still stores the file as `image.png`, so an image published from TypeScript reads back exactly as before.

Before this, the TypeScript client called every image a PNG. Now it reports the format the image was actually saved in.

## Testing

Tests cover `image.jpg`, `image.webp` and an extension we do not know. Two of them fail on the parent branch. I also published a JPEG, a WEBP and a PNG from Python to a live server and read them back here: right labels, same bytes.

[WB-38335]: https://coreweave.atlassian.net/browse/WB-38335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdks/node/src/__tests__/weaveClient.get.test.ts
+++ b/sdks/node/src/__tests__/weaveClient.get.test.ts
@@ -1,4 +1,5 @@
 import {Api as TraceServerApi} from '../generated/traceServerApi';
+import {type ImageType} from '../media';
 import {WeaveClient} from '../weaveClient';
 import {ObjectRef} from '../weaveObject';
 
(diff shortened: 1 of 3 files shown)


rgao-coreweave (approved):
LGTM  👍 👍 👍

PR #7689, opened 2026-08-04
fix(weave_ts): return the real audio from client.get()

ro31337:
## JIRA Issue(s)

- Fixes [WB-38338](https://coreweave.atlassian.net/browse/WB-38338)

## Description

This is the third part of a stack.

First part, #7685: `client.get()` returns real images.

Second part, #7686: with the right format label.

This part: audio still comes back as the string `Coming soon!`. This fixes that. The audio branch sits
right below the image one in the same function.

#7685 already added the file download and `media.ts` has a `weaveAudio()` wrapper, so this downloads
the stored file, wraps the bytes and returns them.

Python writes a `_metadata.json` next to the audio, so I expected to pick the audio file by name. Turns
out Python stopped using this type name in May 2025, so there's only ever one file here. Details in
`AGENTS.md`.

Before: `get()` on audio gave you a string. Now it gives you your bytes.

Audio from today's Python still comes back unread. That's a separate PR — it needs its own branch and
an mp3 type.

## Testing

Two unit tests next to the image ones check that `get()` hands back the audio instead of the
`Coming soon!` string, plus a round trip against a live server that came back byte-identical.

[WB-38338]: https://coreweave.atlassian.net/browse/WB-38338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -339,6 +339,19 @@ pnpm exec tsx examples/claudeAgents.ts
   flattened `weave.integration.meta.*` provenance. OTel scalar metadata stays
   typed; non-scalar values are stringified.
 
+### TypeScript media round trip
+
+- `client.get()` rebuilds a `WeaveImage` from a `PIL.Image.Image` payload and a
+  `WeaveAudio` from a `wave.Wave_read` one, downloading the stored file from
+  `ref.projectId`, not the client's project.
+- `wave.Wave_read` is what this SDK writes, and what Python wrote until May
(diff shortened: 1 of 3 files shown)


rgao-coreweave (approved):
Very nice, LGTM 👍👍👍

PR #7690, opened 2026-08-05
fix(weave): keep chat completion streams open for iteration

mattliu-mygit:
## Summary

- keep the HTTP response and client open until the returned chat completion stream is consumed
- close both resources after successful iteration, parsing failures, setup failures, or explicit context-manager cleanup
- add focused regression coverage for both Playground and Inference streaming

## Root cause

Simplified, the old implementation did this:

```python
with client.stream(...) as response:
    return ChatCompletionChunkStream(response)
```

A Python `with` block cleans up its resource when the block ends. Returning from inside the block still runs that cleanup before the caller receives the returned value:

1. Open the streaming response.
2. Construct `ChatCompletionChunkStream(response)`.
3. Exit the `with` block.
4. Close the response.
5. Give `ChatCompletionChunkStream` to the caller.

The caller therefore received an object that looked like a stream, but its underlying response had already been closed:

```python
stream = client.chat.completions.create(..., stream=True)

stream.response.is_closed
# True

list(stream)
# httpx.StreamClosed
```

The network connection was disconnected before the caller could consume the first chunk. Pre-buffered response bodies can hide this because their bytes are already in memory, but a genuine network-backed stream needs the response to remain open. The old path also left its separately-created `httpx.Client` open.

## How the fix works

The streaming path now opens the HTTP client and response through one `ExitStack`. If setup or status validation fails, normal context-manager unwinding closes everything that was opened. After successful setup, `pop_all()` transfers those cleanup callbacks to `ChatCompletionChunkStream`, so returning the stream no longer closes its underlying response.

The returned stream then owns both resources. Iterating it uses the stream's context-manager path, while explicit `close()` and `with stream:` usage share the same cleanup implementation. In every case, the response is closed before its client, including when response cleanup raises.

## Reproduction

This reproduces against the pre-fix `origin/master` implementation with a network-like HTTPX byte stream:

```python
stream = client.chat.completions.create(
    endpoint="inference",
    model="coreweave/runtime-model",
    messages=[{"role": "user", "content": "Hello"}],
    stream=True,
    track_llm_call=False,
)

print(stream.response.is_closed)
list(stream)
```

Before this PR:

```text
True
httpx.StreamClosed: Attempted to read or stream content, but the stream has been closed.
```

After this PR:

```text
response closed before iteration: False
received content: Hello
response closed after iteration: True
```

The regression test fails consistently on `origin/master` for both `endpoint="inference"` and `endpoint="playground"`, where the returned response is already closed.

## Impact

Callers can now iterate streamed chat completions normally. The response and client remain open only while needed and are closed after iteration or on an error. This PR contains no Custom Runtime conversation-context changes.

## Validation

- Reproduced the closed-before-iteration failure on untouched `origin/master`
- `python -m pytest tests/trace/test_chat_completions.py tests/trace/test_op_generators.py tests/trace/test_op_decorator_behaviour.py tests/trace/test_llm_structured_completion_model.py -q` — 66 passed
- `pyright weave/chat/completions.py weave/chat/stream.py` — passed
- `ruff check weave/chat/completions.py weave/chat/stream.py tests/trace/test_chat_completions.py` — passed

### DIFF
--- a/tests/trace/test_chat_completions.py
+++ b/tests/trace/test_chat_completions.py
@@ -1,0 +1,210 @@
+"""Regression coverage for the public chat completions client."""
+
+from __future__ import annotations
+
+import json
+from unittest.mock import MagicMock
+
+import httpx
+import pytest
+from pydantic import ValidationError
+
+from weave.chat import completions as completions_module
+from weave.chat.completions import Completions
+from weave.trace.weave_client import WeaveClient
+
+
+def _client() -> WeaveClient:
+    client = MagicMock(spec=WeaveClient)
+    client.entity = "entity"
+    client.project = "project"
+    return client
+
+
+def _chunk_payload(content: str = "Hello") -> dict:
+    return {
+        "id": "completion-1",
+        "choices": [
+            {
+                "delta": {"role": "assistant", "content": content},
+                "finish_reason": None,
+                "index": 0,
+            }
+        ],
+        "created": 1,
+        "model": "runtime-model",
+        "object": "chat.completion.chunk",
+    }
+
+
+class _TrackingByteStream(httpx.SyncByteStream):
+    def __init__(self, content: bytes, close_error: Exception | None = None) -> None:
+        self.content = content
+        self.close_error = close_error
+        self.closed = False
+
+    def __iter__(self):
+        yield self.content
+
+    def close(self) -> None:
+        self.closed = True
+        if self.close_error is not None:
+            raise self.close_error
+
+
+def _install_mock_transport(monkeypatch, handler) -> list[httpx.Client]:
+    real_client = httpx.Client
+    transport = httpx.MockTransport(handler)
+
+    clients = []
+
+    def client_factory(**kwargs):
+        client = real_client(transport=transport, **kwargs)
+        clients.append(client)
+        return client
+
+    monkeypatch.setattr(
+        completions_module.httpx,
+        "Client",
+        client_factory,
+    )
+    monkeypatch.setattr(completions_module, "get_wandb_api_context", lambda: "key")
+    return clients
+
+
+def _create_stream(
+    completions: Completions,
+    endpoint: completions_module.Endpoint = "inference",
+    model: str = "coreweave/runtime-model",
+):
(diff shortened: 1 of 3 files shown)


tests/trace/test_chat_completions.py
```diff
@@ -0,0 +1,206 @@
...
+        if endpoint == "inference"
+        else chunk + b"\n"
+    )
+    body = _TrackingByteStream(content)
+
+    def handler(_request: httpx.Request) -> httpx.Response:
+        return httpx.Response(200, stream=body)
+
+    clients = _install_mock_transport(monkeypatch, handler)
+    completions = Completions(_client())  # type: ignore[arg-type]
```
gtarpenning: even in tests this is a smell'

tests/trace/test_chat_completions.py
```diff
@@ -0,0 +1,206 @@
...
+
+    assert not stream.response.is_closed
+    assert not clients[0].is_closed
+    assert [chunk.choices[0].delta.content for chunk in stream] == ["Hello"]
+    assert stream.response.is_closed
+    assert body.closed
+    assert clients[0].is_closed
+
+
+def test_stream_context_manager_closes_unconsumed_resources(monkeypatch) -> None:
```
gtarpenning, line 121: i think a few of these tests can either be paramterized or collapsed into fewer denser tests, which is my personal preference

gtarpenning (approved):
lgtm

tests/trace/test_chat_completions.py
mattliu-mygit: Replaced the SimpleNamespace and type: ignore suppressions with a MagicMock(spec=WeaveClient), so the test double now follows the declared client contract!

tests/trace/test_chat_completions.py
mattliu-mygit: The identical inference/playground behavior is already parameterized. I'm erring on keeping the remaining cases separate because they exercise distinct lifecycle and failure points; combining them would require conditional setup and assertions, making failures less specific.

PR #7691, opened 2026-08-05
chore(weave): add signature cluster tables

gtarpenning:
## Summary
- Add project-sharded run, cluster, and assignment tables keyed to existing signature-row UUIDs.
- Store extraction and clustering provenance as config digests; clustering settings live in typed checked-in `clustering.yaml`.
- Keep scope to base storage: no daily rollup, projection, visualization fields, or reader queries.

## Testing
- ClickHouse lifecycle plus replicated/distributed round-trip and cloud/replicated/distributed migration idempotency.


weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,52 @@
...
+SETTINGS min_bytes_for_wide_part = 0;
+
+-- One assignment per signature-table row in a run.
+CREATE TABLE IF NOT EXISTS signature_cluster_assignments
+(
+    project_id String,
+    cluster_run_id UUID,
+    signature_record_id UUID,
+    cluster_id Int32 DEFAULT -1,
+    cluster_confidence Float32 DEFAULT 0,
```
jtschoonhoven: "distance"

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,52 @@
...
+    inserted_at DateTime64(6, 'UTC') DEFAULT now(),
+    expire_at DateTime DEFAULT '2100-01-01 00:00:00'
+)
```
jtschoonhoven: Let's include umap coords so we can rebuild the umap

gtarpenning: done

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
+-- A cluster run groups one signature type and exact extraction and clustering configs.
+CREATE TABLE IF NOT EXISTS signature_cluster_runs
+(
+    project_id String,
+    id UUID DEFAULT generateUUIDv7(),
+    signature_type LowCardinality(String),
```
jtschoonhoven: This is `intent` or `failure` right?

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    signature_config_sha LowCardinality(String),
+    cluster_config_sha LowCardinality(String),
+    window_start DateTime64(6, 'UTC'),
+    window_end DateTime64(6, 'UTC'),
+    status LowCardinality(String) DEFAULT 'running',
```
jtschoonhoven: I like giving ourselves flexibility, but can we make this an ENUM?

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    started_at DateTime64(6, 'UTC'),
+    completed_at Nullable(DateTime64(6, 'UTC')) DEFAULT NULL,
+    inserted_at DateTime64(6, 'UTC') DEFAULT now(),
```
jtschoonhoven: Do we really need both `completed_at` and `inserted_at`? Won't these be milliseconds apart at most??

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    expire_at DateTime DEFAULT '2100-01-01 00:00:00'
+)
+ENGINE = ReplacingMergeTree(inserted_at)
+ORDER BY (project_id, id)
```
jtschoonhoven: I know this table is small and that `id` is sort of a timestamp already, but shouldn't we still sort on the datetime field we query on?
```suggestion
ORDER BY (project_id, window_end, id)
```

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+TTL expire_at DELETE
+SETTINGS min_bytes_for_wide_part = 0;
+
+-- One generated label and occurrence count per cluster in a run.
+CREATE TABLE IF NOT EXISTS signature_clusters
+(
+    project_id String,
+    cluster_run_id UUID,
+    cluster_id Int32,
+    label String DEFAULT '',
```
jtschoonhoven: This probably shouldn't default to an empty string (unless that has some special meaning that I'm not aware of)

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+ORDER BY (project_id, id)
+TTL expire_at DELETE
+SETTINGS min_bytes_for_wide_part = 0;
+
+-- One generated label and occurrence count per cluster in a run.
+CREATE TABLE IF NOT EXISTS signature_clusters
+(
+    project_id String,
+    cluster_run_id UUID,
+    cluster_id Int32,
```
jtschoonhoven: Why is this not a UUID7 like the others? I'm guessing it's something to do with stable cluster IDs, but it would be good to see that documented here.

Also, how do we make sure these are unique? I don't think clickhouse is into UNIQUE constraints.

Also nit, for consistency this should probably just be named `id`.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    inserted_at DateTime64(6, 'UTC') DEFAULT now(),
+    expire_at DateTime DEFAULT '2100-01-01 00:00:00'
+)
+ENGINE = ReplacingMergeTree(inserted_at)
+ORDER BY (project_id, id)
+TTL expire_at DELETE
+SETTINGS min_bytes_for_wide_part = 0;
+
+-- One generated label and occurrence count per cluster in a run.
+CREATE TABLE IF NOT EXISTS signature_clusters
```
jtschoonhoven, line 33: Maybe `signature_cluster_labels`?

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+ORDER BY (project_id, cluster_run_id, cluster_id)
+TTL expire_at DELETE
+SETTINGS min_bytes_for_wide_part = 0;
+
+-- One assignment per signature-table row in a run.
+CREATE TABLE IF NOT EXISTS signature_cluster_assignments
+(
+    project_id String,
+    cluster_run_id UUID,
+    signature_record_id UUID,
```
jtschoonhoven, line 74: Ok so `spans` joins to `intent_signatures` joins to `signature_cluster_assignments`. How unique are signatures and is it really worth deduplicating them? Could we save ourselves a join and just store `trace_id` here?

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+CREATE TABLE IF NOT EXISTS signature_clusters
+(
+    project_id String,
+    cluster_run_id UUID,
+    cluster_id Int32,
+    label String DEFAULT '',
+    occurrence_count UInt64 DEFAULT 0,
+    inserted_at DateTime64(6, 'UTC') DEFAULT now(),
+    expire_at DateTime DEFAULT '2100-01-01 00:00:00'
+)
```
jtschoonhoven, line 60: I kinda think we should denormalize cost and duration and token count here. Or if that's expensive, could we put it on `signature_cluster_assignments instead`?

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    label String DEFAULT '',
+    occurrence_count UInt64 DEFAULT 0,
+    inserted_at DateTime64(6, 'UTC') DEFAULT now(),
+    expire_at DateTime DEFAULT '2100-01-01 00:00:00'
+)
+ENGINE = ReplacingMergeTree(inserted_at)
+ORDER BY (project_id, cluster_run_id, cluster_id)
```
jtschoonhoven: Should we also sort on `inserted_at`?

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    -- Distance from the signature vector to its cluster centroid, 0 for noise rows.
+    cluster_distance Float32 DEFAULT 0,
+    -- UMAP 2-D projection of the signature vector, reprojected by every run.
+    umap_x Float32 DEFAULT 0,
+    umap_y Float32 DEFAULT 0,
+    inserted_at DateTime64(6, 'UTC') DEFAULT now(),
+    expire_at DateTime DEFAULT '2100-01-01 00:00:00'
+)
+ENGINE = ReplacingMergeTree(inserted_at)
+ORDER BY (project_id, cluster_run_id, signature_record_id)
```
jtschoonhoven: Should we also sort on inserted_at?

jtschoonhoven (commented):
A few questions before merging. Looks pretty good.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: yeah lets store the trace_id

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: yep, now on `signature_cluster_assignments`

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: potentially...

jtschoonhoven: I don't feel strongly about this, up to you.

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,94 @@
...
+
+    signature_type Enum8('intent' = 1, 'failure' = 2),
+    signature_config_sha LowCardinality(String),
+    cluster_config_sha LowCardinality(String),
+    naming_config_sha LowCardinality(String),
+
+    window_start DateTime64(6, 'UTC'),
+    window_end DateTime64(6, 'UTC'),
+
+    status Enum8('unset' = 1, 'running' = 2, 'completed' = 3, 'failed' = 4) DEFAULT 'unset',
```
jtschoonhoven: I don't see any reason to allow "unset", how about replacing "unset" with "pending"? I'd also add "canceled" since we might want that later. Optional, but I prefer "succeeded" vs "completed".

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,94 @@
...
+(
+    project_id String,
+    cluster_run_id UUID,
+    id UUID,
+    -- Copied from the run. Fixed at run creation, so it is partition-safe.
+    run_window_end DateTime64(6, 'UTC'),
+
+    -- Stable across runs. The writer matches each cluster to the previous run's
+    -- clusters by centroid similarity. Nil means no match, so this topic is new.
+    topic_id UUID,
```
jtschoonhoven: > Nil means no match, so this topic is new

So this should be `Nullable(UUID)` then right?

Does this then join on `signature_clusters.id`? That should probably be made more clear. Maybe this column should be named `stable_cluster_id` or `original_cluster_id`. Or maybe, since we're not solving stable clusters yet, we should just drop this column for now.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,94 @@
...
+    -- Distance from the signature vector to its cluster centroid.
+    cluster_distance Float32 DEFAULT 0,
+    -- UMAP 2-D projection of the signature vector.
+    umap_x Float32 DEFAULT 0,
+    umap_y Float32 DEFAULT 0,
+
+    -- `failure_signatures.current_trace_id` or `intent_signatures.trace_id`
+    trace_id String,
+    turn_duration_ms UInt32 DEFAULT 0,
+    turn_cost_usd Float64 DEFAULT 0,
```
jtschoonhoven, line 103: Token count?

jtschoonhoven: Agent name? Model version?

jtschoonhoven (approved):
See comments, but LGTM

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
gtarpenning: in clickhouse we really want to avoid nullable if at all possible, i think the 0 uuid should be fine to represent that. 
also yes this is speculative but I think we are going to at least want this for testing, like thats probably one of the next things i'm going to want to think about. could be wrong

weave/trace_server/migrations/042_add_signature_cluster_tables.up.sql
gtarpenning: that all makes sense

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: we didn't add tokens to the underlying signature tables, should we do that?

gtarpenning: A turn can call several models right?

jtschoonhoven: > We didn't add tokens to the underlying signature tables, should we do that?

Probably yes, but low-priority.

> A turn can call several models right?

Yes

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,126 @@
+-- A cluster run groups one signature type and exact extraction and clustering configs.
+CREATE TABLE IF NOT EXISTS signature_cluster_runs
```
benjreinhart, line 2: Once the MySQL DB is in, thoughts on using it for a table like this? These wouldn't be problematic for mysql any time soon. IMO it's a better fit for this type of data (small records, `status` updates).

Also just curious, do we plan to use this table anywhere today (UI, API), or is this internal bookkeeping that nothing is leveraging now?

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,126 @@
...
+(
+    project_id String,
+    -- Intentionally omit UUID default, tracking in-progress runs requires re-inserting 
+    -- with the same PK, consumers should generate uuidv7 before insert
+    id UUID, -- UUIDv7
+
+    signature_type Enum8('intent' = 1, 'failure' = 2),
+    signature_config_sha LowCardinality(String),
```
benjreinhart, line 10: How do you plan to generate these SHAs? Just the raw file contents, or will do you plan to sort the config keys/values to keep it a little more robust to irrelevant updates?

Also, is this better than just requiring these files to have an explicit version? I don't feel strongly either way, but wondering your thoughts.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,126 @@
...
+    run_window_end DateTime64(6, 'UTC'),
+
+    -- Not an FK: a topic identity carried forward by centroid match across runs, so it
+    -- outlives the per-run `id`. Nil until the writer links a run to its predecessor.
+    topic_id UUID,
+    -- The signature category this cluster was fit within, empty when the run fit the
+    -- whole space at once. Set by the clustering config's `scope`.
+    category LowCardinality(String) DEFAULT '',
+    -- Mean signature vector. Enables assigning a new signature to an existing cluster.
+    centroid Array(Float32),
```
benjreinhart, line 51: How do we handle drift and recentering a cluster as it evolves? Are we planning to update these records with a new centroid?

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,126 @@
...
+
+    label String,
+    description String DEFAULT '',
+    occurrence_count UInt64 DEFAULT 0,
```
benjreinhart, line 55: I assume this is the number of signatures in the cluster?

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,126 @@
...
+    signature_type Enum8('intent' = 1, 'failure' = 2),
+
+    -- Distance from the signature vector to its cluster centroid.
+    cluster_distance Float32 DEFAULT 0,
+    -- UMAP 2-D projection of the signature vector.
+    umap_x Float32 DEFAULT 0,
+    umap_y Float32 DEFAULT 0,
+
+    -- `failure_signatures.current_trace_id` or `intent_signatures.trace_id`
+    trace_id String,
```
benjreinhart, line 90: The UI and API endpoints work primarily off of conversation_id. For example, the UI routes are /conversations/:id. Without that present here, I assume that will require extra lookups to link to conversations.

We're also probably going to want metrics like:

* Distinct conversations in cluster X
* Distinct users in cluster X

Adding conversation_id and user_id here I believe enables that, or were were you thinking we'd first load up other records / have different query patterns?

Lastly. I think we might eventually want that over a time period as well. "In the last day, N conversations were bucketed into X". I'm guessing we don't want to use this record's inserted at for that since it may be computed later. Should we add a timestamp from the trace here, perhaps `trace_ended_at`?

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,126 @@
...
+)
+ENGINE = ReplacingMergeTree(inserted_at)
+-- Every key column is fixed at run creation, so the terminal rewrite of `status` collapses
+-- onto the running row instead of landing beside it.
+ORDER BY (project_id, signature_type, window_end, id)
+TTL expire_at DELETE
+SETTINGS min_bytes_for_wide_part = 0;
+
+-- One generated label and occurrence count per cluster in a run.
+CREATE TABLE IF NOT EXISTS signature_clusters
```
benjreinhart, line 33: Does this need signature_type? Or do we not need to know that on this table?

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
```diff
@@ -0,0 +1,126 @@
...
+    signature_record_id UUID,
+    -- FK to `signature_cluster_runs.window_end`.
+    run_window_end DateTime64(6, 'UTC'),
+    -- FK to `signature_clusters.id`. The nil uuid is noise, which has no cluster row,
+    -- so a run's noise rows are one contiguous range of the sorting key.
+    cluster_id UUID,
+    signature_type Enum8('intent' = 1, 'failure' = 2),
+
+    -- Distance from the signature vector to its cluster centroid.
+    cluster_distance Float32 DEFAULT 0,
```
benjreinhart, line 82: IIRC, hdbscan has a probability associated with assignment. If so, I assume at some point we may want to know that?

benjreinhart (commented):
I'm going to ignore anything surrounding cluster stability as we aren't solving that right now, but I am also assuming we may need some updates for that.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: I agree it could/should be in mysql. Its not critical, but definitely useful, like "get me the latest clustering run for x project" obviously you can reconstruct that from the table below. Especially useful if we do multiple versions of the clustering run, or if the user triggers one, etc. bookkeeping and auditability.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: sha generation should be yaml hash, which is deterministic I think? we can play with that. 

I don't really think it's better than the files having an explicit version. This way you don't have to remember to bump the version, any change will get captured. but also it generates a hash which is a less visually understandable than version. We could always have a version in the yaml too actually.

benjreinhart (commented):
Perhaps hard to do at this time, but if at all possible, it helps if we know how we plan to query this data in the UI (as well as perhaps the primary API exposure). That can help inform how these tables should be designed wrt denormalized columns and ordering keys.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: this is mostly for visual analysis, so that you can plot the centroid. Also used as mentioned in the comment for new records to find a home. 

if we need to update the centroid, we can always insert a new record that replaces/updates with this one, thats the nice part about replacing merge tree. we would just insert a new record matching `(project_id, cluster_run_id, id)` and then it would be updated in the background.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: all good thoughts, will include those denormed fields, and confirm the ability to grab performantly by conversation id.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: good point, optional but nice to have. will add it.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: i think thats actually the same thing as cluster distance, but I will confirm that.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
benjreinhart: What I'm imagining is probably not something to worry much about, but I was more thinking if we ever update the config file with something that doesn't change the outcome of the pipeline. If someone adds a comment in the file, or moves something around in the file, or maybe even adds something that isn't yet used. Would we want those cases to cause a new version? The explicit version solves for that.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
benjreinhart: Yeah agree about bookkeeping / auditing. I was more curious if we had an immediate plan for using it, or were just covering all the bases

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
benjreinhart: 👍🏻 

Generally, with all this "agent span data," I think we're usually going to have the following hierarchy:

```
conversation_id
└── trace_id (a "turn," optional)
    └── span_id (super optional)
```

Leaving as a general point for any data modeling related to referencing the agent span data.

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: hmm, i was wrong, added that

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: as written, that wouldn't create any issues, it would merely tag the row with the updated hash. I don't think that updating the config would ever like deprecate previous rows, its more of a way to concretely answer "hey for this cluster how did we generate it?"

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
benjreinhart: I think this is more nuanced than it seems, but i also think that's something that matters only when we get to cluster stability. So for now, I'm not sure it matters, but it may eventually.

benjreinhart (approved):
This may take more iterations to get right but I think this is as good of a pass as we're going to get now.

:shipit:

weave/trace_server/migrations/043_add_signature_cluster_tables.up.sql
gtarpenning: agree, i think this is not a one-way door though.

neutralino1 (approved):

PR #7695, opened 2026-08-06
fix(weave_ts): return a real Date from client.get()

ro31337:
## JIRA Issue(s)

- Fixes [WB-38379](https://coreweave.atlassian.net/browse/WB-38379)

## Description

This tiny PR is another step toward TypeScript/Python parity in the SDK — the fourth in a row in `client.get()`, taking the most visible gaps first.

`client.get()` in the TypeScript SDK returns a `datetime` published from Python as a raw internal record instead of a `Date`:

```ts
// today
{_type: 'CustomWeaveType', weave_type: {type: 'datetime.datetime'},
 val: '2025-01-01T00:00:00+00:00', load_op: ...}

// with this change
2025-01-01T00:00:00.000Z
```

The call succeeds, so the failure lands later — the first time someone compares or formats the value.

Weave stores rich types in a record tagged with a type name, and `client.get()` reads that tag to rebuild the object. Images and audio had branches, both downloading a stored file; a date fell through to the default, which returns the record as is. A date is the cheap case — Python writes the ISO string straight into the record's `val` — so the new branch is one line: read `val`, return a `Date`.

A `Date` holds milliseconds and an instant, so `.999999` becomes `.999` and `10:15:30+05:30` becomes `04:45:30Z`. Tests pin both; `AGENTS.md` has the rest.

Separate work, still open: publishing a `Date` from TypeScript, and datetimes inside table rows.

## Testing

Three new tests — a plain UTC date, a `+05:30` offset, and one with microseconds — fail on `master` and pass here. I also ran the round trip against a live server. The helper split lets a datetime record omit the `files` key the way the server does; image and audio tests still get the same bytes.

[WB-38379]: https://coreweave.atlassian.net/browse/WB-38379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -339,11 +339,18 @@ pnpm exec tsx examples/claudeAgents.ts
   flattened `weave.integration.meta.*` provenance. OTel scalar metadata stays
   typed; non-scalar values are stringified.
 
-### TypeScript media round trip
+### TypeScript custom type round trip
 
 - `client.get()` rebuilds a `WeaveImage` from a `PIL.Image.Image` payload and a
   `WeaveAudio` from a `wave.Wave_read` one, downloading the stored file from
   `ref.projectId`, not the client's project.
+- It also rebuilds a `Date` from a `datetime.datetime` payload, which stores no
+  file: Python writes the ISO string inline into the record's `val`. A `Date`
+  is a UTC instant with millisecond resolution, so the round trip keeps the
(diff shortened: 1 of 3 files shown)


rgao-coreweave (approved):
👍 👍 👍  LGTM

PR #7665, opened 2026-07-29
fix(weave): trace Claude Agent SDK async-iterable queries per turn

rgao-coreweave:
## Problem

The Claude Agent SDK accepts either a string or an `AsyncIterable` as the prompt to `query()`. An async-iterable prompt can submit multiple inputs and produce a separate `ResultMessage` for each one.

The #7666 base traces the entire output iterator as one turn. A replay containing two submitted prompts and two results therefore produces only one `invoke_agent` span, collapsing both interactions together.

## Stack

```text
master
└── #7666 subagent tracing
    └── #7665 async-iterable multi-turn tracing (this PR)
```

This PR contains one commit and is intentionally based on the #7666 branch. Its diff is only the async-iterable turn-boundary change.

## Approach

Keep the ordinary one-result APIs simple and isolate multi-result coordination to the only path that needs it:

| SDK API | Tracing path | Turn boundary |
| --- | --- | --- |
| `query(prompt=str)` | Linear single-turn tracer | The first `ResultMessage` |
| `query(prompt=AsyncIterable)` | Async-iterable FIFO tracer | Each `ResultMessage` |
| `ClaudeSDKClient.receive_response()` | Linear single-turn tracer | The method stops after its first `ResultMessage` |

Both paths share message processing and turn finalization. The async-iterable path additionally:

- Captures user inputs as the SDK consumes them without pre-consuming or cloning the iterable.
- Buffers messages marked `shouldQuery: false` into the next query-triggering input.
- Pairs submitted inputs with results in FIFO order.
- Reuses the SDK session ID as the conversation ID while keeping each turn in its own trace.
- Records a pending input as an errored turn when transport or parsing fails before its result.
- Leaves a completed turn successful if teardown fails after its terminal result.

String prompts and the public `ClaudeSDKClient.receive_response()` contract remain on the straightforward single-turn implementation.

## Before and after

The same replay submits `Hello`, then `What is the capital of France?`, and receives one `ResultMessage` for each input.

| Run | Spans | Conversation turns | Trace IDs | Result |
| --- | ---: | ---: | ---: | --- |
| [Base #7666](https://wandb.ai/wandb/materials-weave-demo/weave/agents/conversations/pr7665-before-multiturn-1785786801227003000?convTurn=1&chatTrace=48647f4f4e77bee55b23a9924eb557a2) | 3 | 1 | 1 | Both responses are collapsed into one turn. |
[...]

PR #7697, opened 2026-08-06
fix(weave): prevent Claude async turn misattribution

rgao-coreweave:
## Summary

- Prevent promptless turns when `system/init` precedes a delayed async prompt.
- Defer only bootstrap init messages, keeping late background-task output out of the next turn.
- Add regressions for delayed init, transport failure, and stale task output.

Follow-up to #7665 and its [review thread](https://github.com/wandb/weave/pull/7665#discussion_r3716355213).

## Validation

- Claude Agent SDK shard: 44 passed
- Ruff check and format: passed

| Scenario | `master` | This PR |
| --- | --- | --- |
| Delayed prompt | [Mispaired turns](https://wandb.ai/wandb/materials-weave-demo/weave/agents/conversations/pr7697-delayed-master-1786045230124629001?convTurn=1&chatTrace=ee23b0f2c4df932a58cede8056eca256) | [Correctly paired turns](https://wandb.ai/wandb/materials-weave-demo/weave/agents/conversations/pr7697-delayed-dev-1786045230124629001?convTurn=1&chatTrace=dbf0c0d0350d15108322420f113a309f) |
| Late task output | [Attributed to next turn](https://wandb.ai/wandb/materials-weave-demo/weave/agents/conversations/pr7697-stale-task-master-1786045230124629001?convTurn=2&chatTrace=42dd68d44240f621a8bc740400c5fb18) | [Excluded from next turn](https://wandb.ai/wandb/materials-weave-demo/weave/agents/conversations/pr7697-stale-task-dev-1786045230124629001?convTurn=2&chatTrace=4e4838296869e5306c7d7d127418bbf1) |

## Breaking changes

None.

### DIFF
--- a/weave/integrations/claude_agent_sdk/otel_integration.py
+++ b/weave/integrations/claude_agent_sdk/otel_integration.py
@@ -737,6 +737,7 @@ async def _trace_async_iterable_query(
         attributes=_INTEGRATION_OTEL_ATTRS,
     ) as conversation:
         message_iterator = aiter(messages)
+        deferred_messages: list[Any] = []
         while True:
             try:
                 next_message = await anext(message_iterator)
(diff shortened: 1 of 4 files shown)


weave/integrations/claude_agent_sdk/otel_integration.py
```diff
@@ -747,12 +748,28 @@ async def _trace_async_iterable_query(
...
+                        for msg in deferred_messages:
+                            _process_message_for_turn(msg, state)
+                        raise
+                    finally:
+                        _finalize_turn(state)
+
+            # Input tracking happens before the SDK writes each prompt, but
+            # bootstrap output can arrive before the first prompt is consumed.
+            if not inputs.has_pending_turn():
+                deferred_messages.append(next_message)
```
ro31337: Just a question. Not saying I'm 100% confident, but look at what I found https://github.com/anthropics/claude-agent-sdk-python/issues/788

What if `TaskNotificationMessage` from a completed background task arrives _after_ the previous `ResultMessage`, but _before_ the next async prompt?

Could it leak into the next turn (and make its trace include work from the previous turn)?

Up to you to decide if it's the legit case we should address.

ro31337 (approved):
Approving with a caveat to think about the corner case here, maybe improve here, or follow-up, or ignore if it's not relevant. Other changes look good

weave/integrations/claude_agent_sdk/otel_integration.py
rgao-coreweave: Good catch! this is a real edge case. Output from a completed background task could be attributed to the next async prompt. I’ve updated the code to defer only bootstrap system/init messages and added a regression for late task notifications and stale output. 

Proper cross-turn task ownership can be handled separately. I will file a ticket to follow up.

rgao-coreweave: ty!

ro31337 (approved):

PR #7699, opened 2026-08-06
fix(weave): preserve typed message fields during PII redaction

ash0ts:
## Why

I hit this while testing a private HiveMind chat review mirror with `redact_pii=True`.

The Conversation SDK currently serializes each typed `Message`, recursively sends every string through Presidio, and then validates the result back into the typed model. That means a false positive on a protocol value can corrupt the payload before any span is emitted. In the concrete failure, Presidio classified the literal `"uri"` discriminator from `UriPart.type` as a person name, changed it to `"<PERSON>"`, and Pydantic rejected the message union.

This makes URI-backed conversation content impossible to log with PII redaction enabled, even when the URI itself contains no sensitive text.

## What changed

- Walk typed message parts directly instead of redacting a dumped protocol object.
- Preserve roles, union discriminators, IDs, MIME/modality metadata, blobs, file references, and URIs byte-for-byte.
- Continue using the same configured Presidio engines for user-authored text:
  - flat message content;
  - text and reasoning parts;
  - tool-call arguments; and
  - tool-call responses.
- Keep configured field-name redaction behavior for those textual payload fields.
- Add a regression where the fake analyzer deliberately marks the exact string `"uri"` as PII while also redacting an email in adjacent text. The resulting span must retain a valid `UriPart` and contain the redacted email.

## Validation

- `23 passed` in `tests/conversation/test_conversation_settings.py`
- Ruff lint passed
- Ruff formatting check passed
- `git diff --check` passed

This is a small companion fix for the review-mirror work in [wandb/hivemind#8](https://github.com/wandb/hivemind/pull/8). It is independent of the proposed atomic historical-turn API and does not change OTLP delivery or idempotency semantics.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -490,6 +490,10 @@ pnpm exec tsx examples/claudeAgents.ts
 - SDK output can fail before the first message or between completed turns. Keep
   the submitted/pending input observable by ending its `Turn` through the
   exception path so the span records the error before the exception propagates.
+- Conversation message PII redaction walks typed parts directly. Preserve
+  roles, discriminators, identifiers, MIME/modality metadata, and media/file
+  references; only flat text, text/reasoning parts, and tool arguments/results
+  are Presidio-scanned.
 
 ### Claude Agent SDK token accounting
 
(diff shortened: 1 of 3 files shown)


ash0ts:
## Post-rebase validation

I cleaned this draft onto current `master`; it now contains exactly one commit and changes three files. The current branch result is 27 passing conversation-settings tests, plus clean Ruff lint, Ruff format, and `git diff --check`.

The HiveMind review mirror consumes the same fix on its combined pinned companion commit because it also needs the separate no-create initialization capability.

PR #7700, opened 2026-08-06
feat(weave): allow score tracing to be disabled in imperative evals

domphan-wandb:
## JIRA Issue(s)

- https://coreweave.atlassian.net/browse/WB-37813

## Description

This PR adds a `trace_scores` argument to imperative evaluations so callers can record score values without emitting a separate scorer call for every score.

Previously, `ScoreLogger` always traced each score unless tracing was disabled globally. For eval tables, each score is already a value on the prediction, so the extra call adds write overhead without adding scoring-code provenance. Benchmarked a 1500 row table with 30 score cols taking 63 seconds vs 6 seconds with scoring disabled.

Passing `trace_scores=False` now keeps scores on `Evaluation.predict_and_score` and includes them in automatic summaries while skipping their individual calls. The default remains `True` to preserve existing behavior for other imperative evaluation callers.

- Adds `trace_scores` to `EvaluationLogger.__init__` and `EvaluationLogger._create_with_meta`.
- Propagates the setting to `ScoreLogger` for direct, asynchronous, and deferred score logging.
- Preserves score values and automatic summaries when score calls are disabled.
- Keeps nested traced operations under `Evaluation.predict_and_score` when a deferred score uses a placeholder call.

## Testing

- [x] _Added unit tests_
  - [x] Direct untraced scores retain values and summaries.
  - [x] `log_example` retains untraced scores.
  - [x] Deferred untraced scores preserve nested operation parentage.
  - [x] Asynchronous untraced scores retain values.
  - [x] Score tracing remains enabled by default.
  - [x] All 5 PR-specific tests passed with the in-memory trace server.


weave/evaluation/eval_imperative.py
```diff
@@ -392,6 +401,25 @@ def _apply_eval_meta(self, eval_meta: dict[str, Any]) -> None:
         """Internal-only: Merge potential caller-provided fields into the eval meta."""
         self._score_meta = {**eval_meta, **IMPERATIVE_SCORE_META_MARKER}
 
+    def _scores_untraced(self) -> bool:
```
kelu-wandb: I would lean toward calling this _scores_traced and using a `not` at the caller. Positive names are less confusing than negative ones.

kelu-wandb (approved):
approving with suggestion

weave/evaluation/eval_imperative.py
```diff
@@ -688,6 +729,14 @@ class EvaluationLogger:
     # Finish the evaluation
     ev.log_summary({"avg_score": 0.9})
     ```
+
+    Args:
+        trace_scores: Whether each logged score also emits its own scorer call.
```
andrewtruong, line 736: Can you specifically call out this is for the EvalTable case?

weave/evaluation/eval_imperative.py
```diff
@@ -370,13 +376,15 @@ def __init__(
         predict_call: Call,
         predefined_scorers: list[str] | None = None,
         eval_span_context: EvalSpanContext | None = None,
+        trace_scores: bool = True,
```
andrewtruong, line 381: This should be kw only, put `*,` before it.  We should also call out this is for the EvalTable case.

andrewtruong (approved):

weave/evaluation/eval_imperative.py
domphan-wandb: updated, thanks.

weave/evaluation/eval_imperative.py
domphan-wandb: updated

PR #7702, opened 2026-08-06
feat(weave_ts): accept JSON tool values

rgao-coreweave:
## Summary

- add exported `JsonValue` and `JsonObject` types for JSON-compatible tool data
- accept a JSON object or pre-serialized string in `ToolInit.args`
- accept a JSON value in `Tool.end({result})` and serialize structured values inside the SDK
- record `result`, `error`, and `errorType` together when a tool ends
- preserve string arguments/results and deprecated mutable `Tool.result` compatibility

## Testing

- `pnpm exec jest --runInBand src/__tests__/genai` (9 suites, 101 tests, 13 snapshots)
- `pnpm exec jest --runInBand src/__tests__/genai/tool.test.ts` (8 tests)
- `pnpm run typecheck:cjs`
- `pnpm run typecheck:esm`
- `pnpm run lint`
- `pnpm run prettier-check`

## Breaking changes

None. Existing string arguments and mutable `Tool.result` assignments continue to work.

## Related

- wandb/weave-claude-code#147

### DIFF
--- a/sdks/node/src/genai/tool.ts
+++ b/sdks/node/src/genai/tool.ts
@@ -13,22 +13,23 @@ import {
   ATTR_GEN_AI_TOOL_NAME,
   WEAVE_GENAI_TRACER_NAME,
 } from './semconv';
+import type {JsonObject, JsonValue} from './types';
 
 export interface ToolInit extends SpanInitBase {
   name: string;
-  /** String values are recorded as-is; other values are JSON-serialized. */
-  args?: unknown;
+  /** A JSON object, or a pre-serialized string recorded as-is. */
+  args?: JsonObject | string;
   toolCallId?: string;
 }
 
 export interface ToolEndOptions extends SpanEndOptions {
-  /** String values are recorded as-is; other values are JSON-serialized. */
-  result?: unknown;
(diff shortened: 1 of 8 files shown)


sdks/node/src/genai/tool.ts
```diff
@@ -12,36 +13,60 @@ import {
...
+}
+
+function serializeToolValue(value: JsonValue | undefined): string | undefined {
+  if (value === undefined) {
+    return undefined;
+  }
+  if (typeof value === 'string') {
+    return value;
+  }
+  try {
```
ro31337, line 39: I see the PR is draft, so apologies if it is too early to ask and if this is something you had in mind already.

Is there anything that prevents a tool argument or result containing a secret or customer data from being serialized into these span attributes in prod?

Should we make content capture opt-in or redact it before recording?

rgao-coreweave: Thanks, marking the PR ready for review.

Good call out that we should add pii redaction to the tool args. However I don't think we have this capability to do so in typescript yet.

ro31337 (approved):

PR #7704, opened 2026-08-07
feat(weave): link a @weave.op call to the agent spans it produced

ro31337:
> **Superseded by [#7730](https://github.com/wandb/weave/pull/7730).** This PR shows as merged because an auto-merge queued on it fired into its new base branch when the migration was split out into [#7729](https://github.com/wandb/weave/pull/7729). Nothing reached `master`. #7730 is the same branch and the same commits; the review below still applies to it.

Four PRs, in review order:

1. TypeScript evals: [#7708](https://github.com/wandb/weave/pull/7708) (merged), a prerequisite, separate from this work: it fixes existing eval linking
2. Migration: [#7729](https://github.com/wandb/weave/pull/7729), split out of this PR so the columns land before the code that writes them
3. Python + server: **this PR**
4. TypeScript ops: [#7711](https://github.com/wandb/weave/pull/7711)

## JIRA Issue(s)

## Description

Today if you wrap an agent run in a `@weave.op`, the call lands in one table and the spans the agent produced land in another. Nothing links them. So you cannot ask "what did this call actually do" — you get two tabs and no way across.

This adds that link. One new column says which call produced the span, and finding them is a normal filter on the agent-spans query we already have. No new endpoint, no UI change.

I wanted the smallest thing that works: write the link down when the span is created, and stop there. The read side — an API for it and the UI affordance — is WB-38427 and comes next.

The main idea behind this work is a span processor, the same trick we already use for evals. A span starts, the processor asks "is an op running right now", and if yes it writes that call's id onto the span. The server copies the ids into the two `spans` columns #7729 adds. The processor is exported as `weave.OpLinkSpanProcessor`, so someone who configured OTel before `weave.init()` — where weave installs nothing of its own — can register it themselves. That is the whole feature.

## Why this approach

There is no per-integration code here, and that is the point of doing it this way. The processor sits on the tracer provider that every integration writes through, so registering it once covers all of them.

First I tried to be clever about storage: stamp only the top span of each subtree, not every span. I wrote it and it does not work. The rule only compares against the direct parent, so it stamps every second level and one call ends up with several "top" spans. Doing it properly needs bookkeeping I did not want on this path, so I stamp everything, same as evals.

Then I found something I did not expect. When a span has too many attributes, OTel throws away the oldest ones. So the order you write them in decides what you lose. I write the trace id first and the call id second, because the call id is the one you filter on and I would rather lose the other. Same reason this processor is registered before the eval one: if something has to go, it should be the new thing, not eval data that has been shipping since May.

One limit worth knowing. This only works when the span starts in the same context as the op. If a library runs the agent on its own thread, there is nothing to read and the column stays empty. Google ADK's `Runner.run()` does that, so wrap `run_async` instead. The realtime integration is the one source that never links: it emits from a timer thread, and reaching it means carrying the call through its own queue, which is not this PR. Empty is a normal state here, not a lost link.

Before this, a call and its spans had no field in common. Now they share two, and getting from one to the other is one filter.

## Backward compatibility

Additive. The two columns come from #7729 and default to empty; this PR is what starts filling them, and nothing reads them yet. Existing rows read `''`; there is no read-time fallback or backfill.

## Testing

New test file for the processor: inside an op, outside one, nested ops, Conversation SDK, `log_turn`, threads, and a crowded span. The CH query test asserts the rows that come back, not just the count. Also ran the `openai_agents` and `google_adk` shards.

### DIFF
--- a/weave/trace/otel_op_linker.py
+++ b/weave/trace/otel_op_linker.py
@@ -11,8 +11,15 @@ Register this processor alongside the ``BatchSpanProcessor`` during
 
 The link needs the weave call stack to be visible where the span starts. That
 stack is a ``ContextVar``, so a span emitted from a bare thread carries none —
-and neither does one from a provider weave did not install. An empty column is
-a normal state, and so is a link to a call that has already finished.
+and neither does one from a provider weave did not install. Google ADK's
+synchronous ``Runner.run()`` and the realtime integration both emit off the
+caller's thread and so carry no link at all. An empty column is a normal
+state, and so is a link to a call that has already finished, and so is one
+evicted from a span that was already at its attribute limit.
+
+A user who configured their own ``TracerProvider`` before ``weave.init()``
+gets no processor, and can register this one themselves: it is exported as
+``weave.OpLinkSpanProcessor``.
 """
 
 from __future__ import annotations
@@ -36,8 +43,9 @@ class OpLinkSpanProcessor(SpanProcessor):
 
     def on_start(self, span: Span, parent_context: Context | None = None) -> None:
         call = call_context.get_current_call()
-        # A call that has not been created yet has no ids to stamp.
-        if call is None or call.id is None or call.trace_id is None:
+        # A placeholder call carries an empty trace id and no id at all, and
+        # half a link is worse than none: the empty half reads back as unset.
+        if call is None or not call.id or not call.trace_id:
             return
 
         # A crowded span evicts oldest first, so spend the trace id before the id.
--- a/weave/trace/otel_op_linker.py
+++ b/weave/trace/otel_op_linker.py
@@ -36,7 +36,7 @@ class OpLinkSpanProcessor(SpanProcessor):
 
     def on_start(self, span: Span, parent_context: Context | None = None) -> None:
         call = call_context.get_current_call()
-        # set_attribute(key, None) is dropped silently, reading back as "no op".
+        # A call that has not been created yet has no id to stamp.
         if call is None or call.id is None:
             return
 
--- a/weave/trace/otel_op_linker.py
+++ b/weave/trace/otel_op_linker.py
@@ -13,11 +13,6 @@ The link needs the weave call stack to be visible where the span starts. That
 stack is a ``ContextVar``, so a span emitted from a bare thread carries none —
 and neither does one from a provider weave did not install. An empty column is
 a normal state, and so is a link to a call that has already finished.
-
-Register this processor before the eval linker, and write the trace id before
-the call id: a span that overflows OTel's attribute limit evicts oldest first,
-so this order spends the least useful attribute first and leaves the eval
-linkage untouched.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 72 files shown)


weave/trace/otel_op_linker.py
```diff
@@ -0,0 +1,45 @@
...
+    from opentelemetry.context import Context
+
+
+class OpLinkSpanProcessor(SpanProcessor):
+    """OTel SpanProcessor that stamps the enclosing weave call onto spans."""
+
+    def on_start(self, span: Span, parent_context: Context | None = None) -> None:
+        call = call_context.get_current_call()
+        # A call that has not been created yet has no id to stamp.
+        if call is None or call.id is None:
```
rgao-coreweave: A ContextVar read can silently fail.

weave/trace/otel_op_linker.py
```diff
@@ -0,0 +1,45 @@
...
+
+if TYPE_CHECKING:
+    from opentelemetry.context import Context
+
+
+class OpLinkSpanProcessor(SpanProcessor):
+    """OTel SpanProcessor that stamps the enclosing weave call onto spans."""
+
+    def on_start(self, span: Span, parent_context: Context | None = None) -> None:
+        call = call_context.get_current_call()
```
rgao-coreweave, line 45: This gets from ambient state, so this is fragile when starting new thread with a fresh, empty context.

Use case like threading.Thread, ThreadPoolExecutor won't have propagated context vars.

Would explicit parents be better?

weave/trace/otel_op_linker.py
ro31337: `_call_stack` is defined with `default=[]`, so reading ContextVar can't really fail? Or did you mean "silently won't link"? This case is explained in a docstring, covered by a test, and works the same way as it is in eval-processor (since May).

ro31337: I'll add a guard just in case

weave/trace/otel_op_linker.py
ro31337: It's true, and this fact is explained in a docstring as well, ~20-25 lines above. I like your "would explicit parents be better?" though. Let me get back to you.

ro31337: So, to make explicit parents work we need a line of code to pass the parent to, and these spans are not produced by our code.

Take Google ADK. Between the user's `Runner.run()` and the internal `tracer.start_as_current_span` there is not a single line that belongs to us. ADK creates the spans inside 11 of its own modules (counted them), and the `tracer` object is bound at import time, so there is no single seam to swap either. Our integration only adds attributes to spans that already exist. There is nowhere to pass a parent to.

The ambient processor does reach those spans, because it sits on the provider they all go through. So ambient is not the poor man's choice here, it is the only thing that works for third-party code. A hybrid is possible: in `openai-agents` the `start_span` line is ours. But the parent would still come out of the same ContextVar, on the same thread, at the same moment. Identical result, more code.

`openai_realtime` is the one case that genuinely needs the explicit model, since its spans start on a timer thread. That is a different kind of work and belongs in its own PR, so I just opened https://coreweave.atlassian.net/browse/WB-38559 for it. There is a prior question there though: we cannot currently tell which integration produced a given span, so that ticket starts by establishing whether the integration is used at all, and only does the work if it is.

rgao-coreweave (approved):
LGTM



thanks for addressing the comments!

PR #7706, opened 2026-08-07
fix(weave): insert pending model costs when the schema is already current

gtarpenning:
## What breaks today

Model cost changes shipped in `cost_checkpoint.json` (new model ids, updated price history) with no accompanying schema migration were never applied once a database's schema was already at the latest migration version. New model ids render NULL costs and updated prices stay stale.

## Mechanism

`BaseClickHouseTraceServerMigrator.apply_migrations` has a lock-free pre-check that returns before acquiring the migration lock when there are no pending schema migrations:

```python
if not self._has_migrations_to_apply(target_db, target_version):
    logger.info("No migrations to apply to `%s`; skipping lock", target_db)
    return
```

That `return` happens before the post-migration hook (which is what inserts checkpoint costs) ever runs. The hook only fires from inside `_apply_migrations_locked`, which is unreachable once the pre-check short-circuits. So on any database that is already caught up on schema, a checkpoint-only cost change is silently never applied.

## What applies on merge

Three coordinated changes:

1. **`weave/trace_server/costs/insert_costs.py`**: extracted a `pending_costs()` helper (load, then diff against what's already in the table) shared by `insert_costs()` and a new read-only `has_pending_costs()` check. `has_pending_costs` saves/restores the caller's `client.database` (it runs on the shared migrator client outside the normal flow) and never raises, since a broken check must not break server startup. The old `should_insert_costs()` version gate is replaced by `costs_schema_is_ready()`, a `system.columns` probe: the gate's only real question is whether this database has the columns the cost code reads and writes, so it now asks the schema instead of comparing against a hardcoded migration number. `COST_COLUMNS` is the single source of truth, used both as the probe's column set and as the insert's `column_names`.

2. **`weave/trace_server/clickhouse_trace_server_migrator.py`**: added an optional `post_migration_work_check` threaded alongside the existing `post_migration_hook` through all migrator constructors and the factory function. The lock-free pre-check now also asks `_has_post_migration_work`, so pending cost work takes the lock even when there's no schema migration to apply. The check is short-circuited to run only when no migration is pending (the target db may not have the relevant tables otherwise), and is a no-op unless both a hook and a work check are configured, so callers that already pass `post_migration_hook=None` to opt out of the costs hook are unaffected by this change.

3. **`tests/trace_server_migrator/test_migrator_functional.py`**: two real-ClickHouse regression tests. There was previously no test covering the post-migration hook at all, and the existing migrator unit tests use mocked clients, which is why this shipped unnoticed. `tests/trace_server/test_clickhouse_trace_server_migrator.py` also gains a case asserting the work check is never invoked when no hook is configured.

The actual insert still runs inside the migration lock, unchanged from before. Running `insert_costs` unlocked on N rolling-deploy replicas would make the read-then-write cost diff a race, inserting N duplicate rows per model on every deploy.

## Why the version gate went away

`should_insert_costs` required migration 5, where `llm_token_prices` is created. But `get_current_costs` selects `cache_read_input_token_cost` and `cache_creation_input_token_cost`, which migration 27 adds. The gate therefore approved versions 5 through 26, where the read fails and `filter_out_current_costs` swallows the error. That is how the stale 5 survived migration 27: the failure mode is a caught exception nobody reads.

Bumping 5 to 27 would fix the instance and leave the mechanism, so the version number is gone instead. A new parameterized functional test pins the coupling at the three interesting points, and the middle row is exactly the case the old gate got wrong:

| Schema | `llm_token_prices` exists | Gate ready | Cost read works |
| --- | --- | --- | --- |
| 4 | no | no | raises |
| 26 | yes | no | raises |
| latest | yes | yes | yes |

Two related notes:

- `has_pending_costs` is new steady-state work. The checkpoint diff used to run only when a migration was actually applied; it now runs once per replica boot, alongside the small `_read_migration_status` read the pre-check needs. `llm_token_prices` holds a few thousand `system` rows, so this is a cheap read on a path that already talks to ClickHouse, but it is new per-boot work.
- `PostMigrationHookContext` keeps its `current_version` / `target_version` fields even though the default costs hook no longer reads them. They describe the migration that just ran and remain useful to any other hook, so they stay part of the hook contract rather than being removed as dead code.

## Testing

- `uv run --group test python -m pytest tests/trace_server/costs/test_insert_costs.py tests/trace_server/test_clickhouse_trace_server_migrator.py -q`: 118 passed
- `uv run --group test python -m pytest tests/trace_server_migrator/test_migrator_functional.py -q` (real ClickHouse): 25 passed, 1 pre-existing unrelated failure deselected (see below). Includes the backfill regression test, which migrates a scratch database to latest, deletes a model's seeded `created_by='system'` rows to simulate a checkpoint change made after the schema was already current, and asserts the row is restored by a second `apply_migrations` call with no version bump. Verified this test fails (0 rows restored) against the pre-fix code and passes against the fix.
- The schema-gate test was strengthened after that full-file run (adding the table-existence column of the table above) and re-run with the backfill test: 4 passed. The edit touched only that test function and an import, not the source or any shared fixture.
- `uvx ruff check` / `uvx ruff format --check` on touched files: clean
- `nox --no-install -e lint` (ty, import-linter, pyright, mypy, ruff): passed, with the changes staged so `prek` did not stash them out from under the type checkers
- Deselected `test_distributed_legacy_replicated_management_db`: pre-existing failure in this environment unrelated to this change (an `ORDER BY (db_name)` vs `ORDER BY db_name` engine-string formatting difference, reproduces identically on an unmodified checkout with this branch's changes absent, so it is a local ClickHouse version difference in how the engine string is rendered).

### DIFF
--- a/tests/trace_server/costs/test_insert_costs.py
+++ b/tests/trace_server/costs/test_insert_costs.py
@@ -177,7 +177,6 @@ class TestInsertCosts(unittest.TestCase):
         mock_insert_db.assert_called_once_with(
             self.client, mock_filter_costs.return_value
         )
-        mock_logger.info.assert_any_call("Loaded %d costs from json", 2)
         mock_logger.info.assert_any_call(
             "There are %d costs to insert, after filtering out existing costs", 1
         )
@@ -258,7 +257,6 @@ class TestInsertCosts(unittest.TestCase):
         mock_load_json.assert_called_once()
         mock_filter_costs.assert_called_once_with(self.client, self.sample_costs_json)
         mock_insert_db.assert_not_called()
-        mock_logger.info.assert_any_call("Loaded %d costs from json", 2)
         mock_logger.info.assert_any_call(
             "There are %d costs to insert, after filtering out existing costs", 0
         )
--- a/tests/trace_server/test_clickhouse_trace_server_migrator.py
+++ b/tests/trace_server/test_clickhouse_trace_server_migrator.py
@@ -466,6 +466,32 @@ def test_apply_migrations_costs_disabled_does_not_call_costs(mock_migration_lock
     mock_insert_costs.assert_not_called()
 
 
+def test_apply_migrations_work_check_inert_without_a_hook():
+    """A work check is meaningless without a hook to act on it.
+
+    A caller that opts out of the hook (post_migration_hook=None) but leaves a
+    work check configured must not have that check invoked by the lock-free
+    pre-check, even at a schema version where it would otherwise fire.
+    """
+    ch_client = _make_ch_client()
+    work_check = Mock(return_value=True)
+    migrator = trace_server_migrator.get_clickhouse_trace_server_migrator(
+        ch_client, post_migration_hook=None, post_migration_work_check=work_check
+    )
+    migrator._has_migrations_to_apply = Mock(return_value=False)
+    migrator._read_migration_status = Mock(
+        return_value={"curr_version": 10, "partially_applied_version": None}
+    )
+
+    with patch(
+        "weave.trace_server.clickhouse_trace_server_migrator.migration_lock"
+    ) as mock_lock:
+        migrator.apply_migrations("test_db")
+
+    work_check.assert_not_called()
+    mock_lock.assert_not_called()
+
+
 def test_apply_migrations_skips_lock_when_nothing_pending():
     """The lock-free pre-check short-circuits when no schema migration is due,
     so rolling replicas don't serialize through the lock for a no-op.
@@ -475,6 +501,7 @@ def test_apply_migrations_skips_lock_when_nothing_pending():
         ch_client, post_migration_hook=None
     )
     migrator._has_migrations_to_apply = Mock(return_value=False)
+    migrator._has_post_migration_work = Mock(return_value=False)
     migrator._apply_migrations_locked = Mock()
 
     with patch(
--- a/tests/trace_server/test_clickhouse_trace_server_migrator.py
+++ b/tests/trace_server/test_clickhouse_trace_server_migrator.py
@@ -113,7 +113,7 @@ def mock_migration_lock():
 def mock_costs():
     with (
         patch(
-            "weave.trace_server.clickhouse_trace_server_migrator.should_insert_costs",
+            "weave.trace_server.clickhouse_trace_server_migrator.costs_schema_is_ready",
             return_value=False,
         ),
         patch("weave.trace_server.clickhouse_trace_server_migrator.insert_costs"),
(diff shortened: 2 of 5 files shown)


nikumar1206 (commented):
<html><head></head><body><p><code dir="ltr">llm_token_prices</code> is created in migration 5, but the cost check reads two extra columns:</p><pre dir="ltr"><code>cache_read_input_token_cost
cache_creation_input_token_cost</code></pre><p>Those columns are not added until migration 27.</p><p>So if someone runs:</p><pre dir="ltr"><code>apply_migrations(db, target_version=10)</code></pre><p>the database has the costs table, but not those two columns. The new <code dir="ltr">should_insert_costs</code> returns <code dir="ltr">True</code> because <code dir="ltr">10 &gt;= 5</code>, then the cost check queries columns that do not exist. It catches the database error and returns no pending costs.</p><p>Separately, a rollback such as version 40 down to 30 now also returns <code dir="ltr">True</code>. That makes the migration hook insert the new checkpoint’s cost rows while rolling the schema backwards. The old condition avoided this because it only allowed the hook when moving forward.</p><p>The desired behavior is:</p><div><div><div>
Migration action | Run cost hook?
-- | --
26 → 27 | Yes
30 → 30, checkpoint-only backfill | Yes
40 → 30 | No
0 → 10 | No, required columns do not exist

</div></div></div><p>That is why I suggested both:</p><ul><li>require schema version 27, not merely 5</li><li>require <code dir="ltr">target_version &gt;= current_version</code> when an explicit target is supplied</li></ul></body></html>

nikumar1206 (approved):

gtarpenning:
@nikumar1206 both claims check out against the code, different conclusions on each.

**5 vs 27:** right that `get_current_costs` reads the two cache columns and only `027_add_cache_token_costs` adds them. But the old gate returned True there too: `should_insert_costs(0, 10)` is `10 >= 5 and 0 < 10`, so `target_version=10` never inserted costs before this PR either, the SELECT failed and `filter_out_current_costs` swallowed it. The only newly reachable case is `5 <= target <= 26 and curr == target`, where the work check throws, `has_pending_costs` catches it, and no lock is taken. Spurious traceback, not a data change.

Fixing it anyway, for the rot rather than the bug: 27 added columns to the cost query and nobody bumped the 5, and nothing caught it because the failure is a caught exception. Bumping 5 to 27 fixes this instance and leaves the mechanism in place. The gate only ever asks "does this db have the columns I'm about to read and write", so I'll ask `system.columns` that directly and delete the hardcoded version. One cheap query, covers table-missing and columns-missing, no version threaded into the hook context, no except path.

**40 to 30:** down migrations are disabled. `_determine_migrations_to_apply` logs and returns `[]` when `target < current` (the loop is commented out), so that call is a schema no-op: the work check runs at 40, the hook inserts at 40, same as a normal boot. Nothing rolls backwards. And if downs came back, `target >= current` is the wrong invariant. The hook already evaluates `applied_target_version`, so a capability gate handles a real 40 to 30 on its own: the columns exist at 30, the insert is legal, and `llm_token_prices` is append-only and effective_date keyed. A schema rollback is not a request for last quarter's prices. Skipping that guard.

Separately, something neither of us flagged: `has_pending_costs` is new per-boot work, since the diff query used to run only when a migration was actually applied. Tiny table, but it is a new read on every replica boot and belongs in the PR body.

gtarpenning:
@nikumar1206 pushed `eb61b1fa`.

The gate is now a `system.columns` probe over `COST_COLUMNS`, the same tuple the insert passes as `column_names`, so there is no migration number left to keep in sync and the table-missing case falls out of the same check. `should_insert_costs` is gone.

New parameterized functional test pins v4 / v26 / latest. The v26 row asserts table exists, gate refuses, cost read raises, which is precisely the range the hardcoded 5 approved.

Left the `target >= current` guard out, for the reasons above. Body now also flags the two things worth knowing: the checkpoint diff is new per-boot work, and `PostMigrationHookContext` keeps its version fields even though the costs hook stopped reading them.

PR #7708, opened 2026-08-09
fix(weave_ts): link an evaluation to the agent spans it produced

ro31337:
Four PRs, in review order:

1. TypeScript evals: **this PR**. A prerequisite, separate from this work: it fixes existing eval linking.

## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-38444

## Description

In TypeScript an evaluation never links to the agent spans its predictions produce. No "View spans" button, and the eval columns on `spans` stay empty. Python has had this link since May.

The code for it was here all along, just never connected. The processor registered on the OTel global provider, but weave emits through its own provider, which is kept out of that registry. The call never even ran: with no user-installed provider, `trace.getTracerProvider()` has no `addSpanProcessor`, so the helper returned `false` and nobody checked. Nothing regressed; the wire was never there.

The fix registers the processor where the provider is built, because that happens lazily on the first span and again when `init()` switches projects.

One attribute was missing too. The server counts a span as linked only with both `weave.eval.run_id` and `weave.eval.predict_and_score_call_id`, and TypeScript wrote only the second. It now writes both, run id first, since a full span drops whatever arrives next.

That is enough for the "View spans" drill-in. Full parity is a separate ticket, and so is `EvaluationLogger`, which cannot be linked this way at all.

## Why this approach

Turning the processor on also turns on its `onEnd` hook, which appended a span ref to the prediction call's `childSummary`. That hook has never run, and once it does the result is wrong: `childSummary` propagates to the parent, and merging two rows' arrays gives an object with numeric keys, so every multi-row eval would ship a malformed `weave.genai_span_ref` on its root call. So it is gone. An existing test already says an SDK must not write that field; the promoted columns carry the link.

The linker goes after the exporting processor, the same order Python uses. That is neutral for the default pipeline. A custom `settings.genai.spanProcessor` now runs before it and will not see `weave.eval.*`, but no processor has ever seen those attributes, so nothing is lost today.

## Testing

The new tests go through the public path: run an `Evaluation`, emit a real GenAI span, read the exported span. That is the only kind of test that catches this, since the old ones drove the processor directly and passed for months against a processor nobody had attached; the full `sdks/node` suite (503 tests), eslint, prettier and both typechecks are green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -381,10 +381,10 @@ deterministic.
   builds, never the OTel global registry. A processor that must see them belongs
   in that provider's `spanProcessors`, which is built lazily on the first span
   and rebuilt on a project switch — registering once from `init()` misses both.
-- OTel JS drops the *incoming* attribute once a span is at its attribute limit,
-  the opposite of Python's evict-the-oldest `BoundedAttributes`. Write order is
-  therefore load-bearing and inverted between the two SDKs: whatever a
-  processor sets first is what survives.
+- At a span's attribute limit OTel JS drops the *incoming* attribute, where
+  Python's `BoundedAttributes` evicts the oldest. A processor writing several
+  attributes therefore keeps its most important one first in JS and last in
+  Python.
 - For callback or generator integrations, create `Conversation`, `Turn`, and
   `LLM` in short `runIsolated()` scopes, then retain and pass explicit handles.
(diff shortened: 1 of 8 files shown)


rgao-coreweave (approved):
Amazing work, learned a lot!

LGTM

PR #7711, opened 2026-08-10
feat(weave_ts): link a weave op call to the agent spans it produced

ro31337:
Four PRs, in review order:

4. TypeScript ops: **this PR**.

## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-38448

## Description

A weave op call and the agent spans it produces share no field, so you cannot ask which spans a call produced. #7730 adds that link in Python. This does the same in Node: a span processor writes the enclosing call's id and trace id onto each span as it starts. The attribute names are Python's, so the server needs no change.

## Why this approach

The processor goes into the `spanProcessors` array `genai/provider.ts` builds. That array is the only path weave's spans take, and it is rebuilt on a project switch, so registering from `init()` would miss it. That same indirection is what hid the eval bug #7708 fixes.

Two orderings are the reverse of Python's, because OTel JS drops the *incoming* attribute on a full span where Python drops the oldest: call id before trace id, and this processor after the eval one. Both have a test.

## Coverage

This covers the spans weave emits through its own provider, not every span in the process. Node keeps its own so that a user's provider cannot take weave's over. The call stack is an `AsyncLocalStorage`: a worker thread gets no link, and a span emitted while a streaming op's result is consumed links to the consuming op, an existing `op.ts` gap. Existing spans stay empty.

## Testing

The tests run an op, emit a real span and read the exported span, so a processor that is never reached cannot pass them. They cover nested ops, `runIsolated`, a re-init and both orderings, and the full `sdks/node` suite (510 tests), eslint, prettier and both typechecks are green.

## Merge order

Stacked on #7730, so the columns exist before this starts writing to them. Nothing here depends on their code; the order is only so the attributes land in the promoted columns rather than the overflow map.


rgao-coreweave (approved):
LGTM

Good to see typescript parity

PR #7712, opened 2026-08-10
fix(weave): map OpenAI cache write tokens

rgao-coreweave:
## What

Map OpenAI prompt-cache usage to Weave's cache usage fields:

| API | OpenAI usage fields |
| --- | --- |
| Responses | `input_tokens_details.cached_tokens`, `input_tokens_details.cache_write_tokens` |
| Chat Completions | `prompt_tokens_details.cached_tokens`, `prompt_tokens_details.cache_write_tokens` |

Weave maps these to `cache_read_input_tokens` and `cache_creation_input_tokens`. Responses OTel spans also include cache-write usage. Calls without `cache_write_tokens` are unchanged, and pricing remains in Weave's standard cost registry.

## Why

[OpenAI documents](https://developers.openai.com/api/docs/guides/prompt-caching#requirements) that GPT-5.6 and later report cache writes in `cache_write_tokens` and charge 1.25× the uncached input token rate. Without this mapping, Weave leaves `cache_write_tokens` in `input_tokens` or `prompt_tokens` and applies the Input rate.

## How

Map `cache_write_tokens` when present instead of parsing model names:

| Model | Behavior |
| --- | --- |
| Before GPT-5.6 | Field absent; unchanged |
| GPT-5.6 | Map the field and use the model's `cache_creation_input_token_cost` |
| Later model family | Same mapping; use that model's registry cost |

The cost example aggregates ten calls, each with 100K input tokens and 10K output tokens. The same amounts appear as `input_tokens` / `output_tokens` in Responses and `prompt_tokens` / `completion_tokens` in Chat Completions.

| Value | Calculation |
| --- | --- |
| Total input | `10 × 100,000 = 1,000,000` |
| `cached_tokens` | `200,000`, included in total input |
| `cache_write_tokens` | `300,000`, included in total input |
| Input-rate amount before this fix | `1,000,000 - 200,000 = 800,000` |
| Input-rate amount after this fix | `1,000,000 - 200,000 - 300,000 = 500,000` |
| Total output | `10 × 10,000 = 100,000` |
| `total_tokens` | `1,000,000 + 100,000 = 1,100,000` |

At [GPT-5.6 Sol rates](https://developers.openai.com/api/docs/models/gpt-5.6-sol):

- Before: `800K × $5/M + 200K × $0.50/M + 100K × $30/M = $7.10`
- After: `500K × $5/M + 200K × $0.50/M + 300K × $6.25/M + 100K × $30/M = $7.475`

## Testing

- 18 OpenAI integration tests passed with the in-memory backend.
- 5 Responses OTel tests passed.
- The Chat Completions pre-5.6, GPT-5.6, and later-model cases passed with ClickHouse.
- Ruff and formatting checks passed.

No API, schema, dependency, configuration, or breaking changes.


weave/integrations/openai/openai_sdk.py
```diff
@@ -454,6 +455,14 @@ def _normalize_openai_cache_tokens(usage: dict[str, Any]) -> None:
         usage.setdefault(
             "cache_read_input_tokens", input_tokens_details["cached_tokens"]
         )
+    if (
+        isinstance(input_tokens_details, dict)
+        and input_tokens_details.get("cache_write_tokens") is not None
```
ro31337: OpenAI docs  https://developers.openai.com/api/docs/guides/prompt-caching#requirements  say that Chat Completions returns `cache_write_tokens` under `prompt_tokens_details`.

However, it seems we only map it here from `input_tokens_details` ?

So, GPT-5.6 Chat Completions cache writes would still be charged at the regular input rate instead of `cache_creation_input_token_cost`, right?

The exact quote from the docs:

> The following Chat Completions usage example shows both fields.

I think we could normalize `prompt_tokens_details.cache_write_tokens` as well and add a Chat Completions case. What do you think?

rgao-coreweave: Good catch. Added `prompt_tokens_details.cache_write_tokens` normalization in 5d730e3526 and applied the pre-5.6, GPT-5.6, and later-model cost cases to both Chat Completions and Responses. The Chat Completions cases pass against ClickHouse.

rgao-coreweave: ty!

weave/integrations/openai/openai_sdk.py
```diff
@@ -432,28 +432,23 @@ def openai_on_input_handler(
 
 
 def _normalize_openai_cache_tokens(usage: dict[str, Any]) -> None:
-    """Flatten OpenAI's nested cache token fields to canonical names.
+    """Copy OpenAI's nested cache token fields to Weave usage fields.
```
ro31337, line 436: Are we sure we're OK merging this without a corresponding pricing entry? There is no gpt-5.6 / gpt-5.6-sol entry in manual_costs.json with a real cache_creation_input price, and LiteLLM doesn't currently provide one either.

I see tests set it to `6.25e-6` that's why they pass

ro31337 (approved):

weave/integrations/openai/openai_sdk.py
rgao-coreweave: Yep the cost is generated in `cost_checkpoint.json` file. 

https://github.com/wandb/weave/blob/dbeed9af4392f38b4f791cb6b0c1148dbda72373/AGENTS.md?plain=1#L95

rgao-coreweave: The 6.25e-6 value in this test is only a project-scoped fixture for verifying.

PR #7714, opened 2026-08-10
refactor(weave): give the intent_records taxonomy one source of truth

gtarpenning:
## Summary
- Stacked on #7598. The category taxonomy lived in the DDL's Enum and constraint and (soon) in the judge prompt, with nothing tying them together, so a label added to one would drift from the other silently.
- Adds `intent_records_taxonomy.py` as the writer's copy and derives the test's expectations from it. The column-type assertion renders the Enum from the module instead of repeating it, and the round-trip is exhaustive: every declared `(lens, category)` pair must insert, and every lens-exclusive label must be rejected under the opposite lens.
- Promotes `'other'` to a real member of both taxonomies instead of a `category = 'other'` disjunct sitting outside them. Same accepted values, but it now reads as what it is: the label a writer falls back to when it does not recognize a judge's output.

## Testing
Both drift directions were confirmed to fail before being reverted: a label added to the module alone fails the Enum assertion, and a label moved to the wrong lens fails the constraint round-trip.


ro31337:

PR #7716, opened 2026-08-10
fix(weave): warn once per ignored-field set instead of per request

ro31337:
## JIRA Issue(s)

[WB-38488](https://coreweave.atlassian.net/browse/WB-38488)

## Description

Our trace server writes a warning every time a request carries a field it does not know. Turns out we do this to ourselves. The batch ingest endpoint takes a list of items, each one either a call start or a call end, and that union has no discriminator — so pydantic tries every item against both shapes, and the shape that loses logs a warning about the other one's field. That is one warning for every call we ingest. Over the last seven days it came to 13.2M lines, out of the 13.6M warnings the whole service wrote — 97% of them, about 2M a day — see [datadog](https://us5.datadoghq.com/logs?query=service%3Aweave-trace%20env%3Aprod%20%22Ignoring%20unexpected%20fields%20while%20validating%22&live=true). Everything else at warn level put together is 397k. The warning channel is dead as a channel.

The easy fix is to delete the warning. I did not want to. In those same seven days it also caught nine lines — nine, out of 13.2M — where a caller misspelled a query parameter: `fields` instead of `columns`, `digest` instead of `digests`. The server drops those and still returns success, so they are real bugs. Nobody is ever going to find nine lines in thirteen million. WB-38398 covers fixing that part.

So I kept the warning and killed the repeating. The server now remembers what it already said, and says each thing once.

Giving that union a discriminator would remove these warnings at the source, and it is the same `Field(discriminator=...)` we already use elsewhere in this repo. I left it out: it turns `mode` into a `Literal`, which narrows the accepted API and means regenerating the schema and the Node SDK. That belongs in its own PR.

The obvious move was to reuse `log_once`, which already exists in `weave/trace/util.py`. Turned out I could not: the trace server is not allowed to import the client SDK, and there is an import-linter rule that fails the build if you try. It also appends a suffix to the message, which would change the text. So I wrote a small memo in the same file — a set of what has already been reported, behind a lock, the same shape as the other caches in that package.

One thing needed care. The field names come from the request body, so a caller decides what goes into that set. It is capped at 1024 entries and it never evicts. Evicting is the normal thing to do, but here it would undo the fix: drop an entry and that caller's next request warns again. When the cap fills, it says so once and then stays quiet.

Bottom line: same warning, same wording, but you see it once per process instead of two million times a day. One thing to know — a repeat now shows up once, so if you are tailing logs it will look like the problem stopped.

## Testing

`uv run --group test python -m pytest tests/trace_server/ tests/shared/` passes. Seven new tests cover it: the same field set warns once, a different set or a different model still warns, a full memo says so and then goes quiet, and nothing is remembered that never got logged. They are plain unit tests and need no ClickHouse.

[WB-38488]: https://coreweave.atlassian.net/browse/WB-38488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace_server/test_trace_server_interface_util.py
+++ b/tests/trace_server/test_trace_server_interface_util.py
@@ -8,6 +8,7 @@ import pytest
 from pydantic import Field, ValidationError
 
 from weave.shared.trace_server_interface_util import extract_refs_from_values
+from weave.trace_server import common_interface
 from weave.trace_server import trace_server_interface as tsi
 from weave.trace_server.common_interface import BaseModelStrict
 from weave.trace_server.errors import NotFoundError
@@ -23,6 +24,21 @@ REF_B = "weave-trace-internal:///test_project/object/obj_b:def456"
 class _BaseModelStrictTestModel(BaseModelStrict):
     required_field: str
     aliased_field: str = Field(alias="aliasedField")
+
+
+class _SecondBaseModelStrictTestModel(BaseModelStrict):
+    required_field: str
+
+
+def _validate_with_extra_field(extra_field: str) -> None:
+    """Validate `_BaseModelStrictTestModel` with one unknown field attached."""
+    _BaseModelStrictTestModel.model_validate(
+        {
+            "required_field": "required",
+            "aliasedField": "aliased",
+            extra_field: "value",
+        }
+    )
 
 
 def test_base_model_strict_ignores_and_warns_on_unknown_fields(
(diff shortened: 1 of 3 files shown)


weave/trace_server/common_interface.py
```diff
@@ -15,6 +17,49 @@
 
 logger = logging.getLogger(__name__)
 
+# One warning per distinct model-and-fields message for the life of the process instead
+# of one per request. The names are caller-controlled, so the memo is capped — far above
+# the handful of distinct messages a fleet actually produces — and never evicted, since
+# forgetting one would let it warn all over again.
```
mscavezze-cw, line 23: Consider allowing re-warns on some cadence. We don't need spam every second, but we probably also don't want to silence the warning forever. Maybe once a day or once a week would be reasonable to remind us of the warning without obscuring the more meaningful ones.

If some additional complexity was acceptable, you could detect how many warnings we suppressed when re-warning, and then maybe alert in cases when that number is high (spam needs fixing) or low (there may be a newly introduced problem).

weave/trace_server/common_interface.py
```diff
@@ -15,6 +17,49 @@
...
+_warned_field_sets: set[tuple[str, str]] = set()
+_warned_field_sets_lock = threading.Lock()
+
+
+def _warn_ignored_fields_once(model_name: str, extra_fields: set[str]) -> None:
+    """Warn about a model's ignored fields the first time this warning would be sent."""
+    # Built before anything else, so a payload the warning cannot render fails the same
+    # way it did before this memo existed, whatever the log level. Sorted because equal
+    # sets can still iterate in different orders, which would key the same fields twice.
+    message = ", ".join(sorted(extra_fields))
```
mscavezze-cw, line 34: nit: this could probably happen after the early-out. Moving it below that would place it closer to where it is actually used without any apparent change in functionality.

tests/trace_server/test_trace_server_interface_util.py
```diff
@@ -51,6 +67,127 @@ def test_base_model_strict_preserves_required_field_validation() -> None:
...
+    """
+    with caplog.at_level("WARNING", logger="weave.trace_server.common_interface"):
+        _validate_with_extra_field("\ud800")
+
+    assert caplog.messages == [
+        "Ignoring unexpected fields while validating _BaseModelStrictTestModel: \ud800"
+    ]
+
+
+def test_base_model_strict_keeps_raising_on_unformattable_field_names(
```
mscavezze-cw, line 129: Should we raise on unformattable field names, or would it be better to use some placeholder text and output the warning?

mscavezze-cw (approved):
Looks good. I left a couple of comments about potential enhancements.

Do we also have another task to cleanup the spammy warnings? It seems like that could be a legitimate issue worth addressing.

weave/trace_server/common_interface.py
ro31337: Thanks! I think this already happens for us, just implicitly. The memo lives in the process, and we deploy often. Prod seems to be running 32 single-process pods and every rollout replaces all of them. So each distinct warning comes back a few times a day on its own, without any timer. I don't like my own answer to your comment, because I would prefer this to be defined, not assumed.

However, whether we want something explicit on top of that, I need to think about. Honestly, I do not have a design in mind I like yet, so I filed https://coreweave.atlassian.net/browse/WB-38553 to let it sit if that's OK

weave/trace_server/common_interface.py
ro31337: so, the join is the only step here that can fail: a non-string field name throws a TypeError. Before the early out, that happens at any log level, same as master. 

If you move it after, and with warnings disabled the same request goes through instead. The test `test_base_model_strict_keeps_raising_on_unformattable_field_names` fails with DID NOT RAISE if the line moves...

tests/trace_server/test_trace_server_interface_util.py
ro31337: I do not think so, but I do not have a strong opinion. It would change the behavior compared to master, which raised here too. And a loud error feels a bit better: a non-string field name cannot come from JSON, so it only happens when our own code passes one. And I would rather see that than a warning with a placeholder in it.

I'd be happy to address this part if you think we should fix it this way though.

PR #7718, opened 2026-08-11
feat(weave): add tables for tags monitoring

benjreinhart:
We're considering a design for a new way to specify tags that would be applied either by humans in the UI or LLM judges. 

This is a simpler data model for labeling traces (turns) of conversations in the spans table. It does away with the ref system and object versions. Importantly, it improves the UI's ability to query these quickly at scale across both conversations a user may be viewing, as well as filtering down all conversations by specific tags.


benjreinhart:
* **#7718** <a href="https://app.graphite.com/github/pr/wandb/weave/7718?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wandb/weave/7718?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

weave/trace_server/migrations/040_add_trace_tags.up.sql
```diff
@@ -0,0 +1,90 @@
...
+
+-- Current tag-centric state, ordered efficiently by the backing table's key.
+CREATE VIEW IF NOT EXISTS trace_tags_by_tag_current AS
+SELECT
+    project_id,
+    tag,
+    trace_ts,
+    conversation_id,
+    trace_id,
+    argMax(source, trace_tags_by_tag.inserted_at) AS source,
```
ro31337: nit: what happens if two audit rows for the same project, trace, and tag share the same inserted_at value? I don't know if it is something even possible, maybe not.

I'm wondering if argMax could choose a different row after merges and leave the current tag view showing the wrong add/remove state?

PS it's not a review, just a question, feel free to ignore

ro31337 (approved):
not requesting any changes, I have one question. The change itself looks good, so approving.

weave/trace_server/migrations/040_add_trace_tags.up.sql
benjreinhart: The intention behind inserted_at is to always be increasing without duplicates. Technically, nothing prevents that, but I don't see a reason to purposely write duplicate values and I don't expect this to happen otherwise.

weave/trace_server/migrations/040_add_trace_tags.up.sql
```diff
@@ -0,0 +1,90 @@
...
+CREATE TABLE IF NOT EXISTS trace_tags
+(
+    project_id      String,
+    conversation_id String,
+    trace_id        String,
+    tag             LowCardinality(String),
+
+    source          Enum8('judge' = 1, 'human' = 2),
+    source_version  LowCardinality(String),
+    wb_user_id      String,
```
gtarpenning, line 26: consider skip index on wb_user_id, probably set

weave/trace_server/migrations/040_add_trace_tags.up.sql
```diff
@@ -0,0 +1,90 @@
...
+    rationale       String DEFAULT '' CODEC(ZSTD(3)),
+    inserted_at     DateTime64(6),
+    is_removed      UInt8 DEFAULT 0
+)
+ENGINE = MergeTree
+PARTITION BY toYYYYMM(trace_ts)
+ORDER BY (project_id, conversation_id, trace_id, tag);
+
+-- Tag-centric copy for querying recent traces with a given tag.
+CREATE TABLE IF NOT EXISTS trace_tags_by_tag
```
gtarpenning: why not use a `PROJECTION`?

weave/trace_server/migrations/040_add_trace_tags.up.sql
```diff
@@ -0,0 +1,90 @@
...
+    tag             LowCardinality(String),
+
+    source          Enum8('judge' = 1, 'human' = 2),
+    source_version  LowCardinality(String),
+    wb_user_id      String,
+
+    -- Removal and re-add rows for a (trace_id, tag) must preserve trace_ts.
+    trace_ts        DateTime64(3),
+    rationale       String DEFAULT '' CODEC(ZSTD(3)),
+    inserted_at     DateTime64(6),
```
gtarpenning: maybe calculate these in the db? with materialized `now`?

weave/trace_server/migrations/040_add_trace_tags.up.sql
```diff
@@ -0,0 +1,90 @@
...
+    trace_id,
+    tag,
+    argMax(source, trace_tags.inserted_at) AS source,
+    argMax(trace_ts, trace_tags.inserted_at) AS trace_ts,
+    argMax(source_version, trace_tags.inserted_at) AS source_version,
+    argMax(wb_user_id, trace_tags.inserted_at) AS wb_user_id,
+    argMax(rationale, trace_tags.inserted_at) AS rationale,
+    max(trace_tags.inserted_at) AS inserted_at
+FROM trace_tags
+GROUP BY project_id, conversation_id, trace_id, tag
```
gtarpenning: i think the group by in this view makes it basically non-functional. as tags grow this group by does more and more work. i measured locally at even 1 million tags this view query takes 240MiB and 100+ms (scaling linearly with ingest). 
limits always apply after the groupby, so even limiting 50 runs an entire aggregation. this is to say, i don't think this is necessary at all. in fact I suspect this will actually overall harm perf. in general we will be querying by conversation right? or at least filtering down by project_id, which lead to performant queries already. id say skip this.

gtarpenning: fwiw if you did really want this you could use a RMT pushing the collapse into background merges, which my claude benchmarked at: 
```
┌────────────┬────────────────┬─────────────┬─────────┐
│ Tag volume │ RMT: rows read │ RMT: memory │ RMT: ms │
├────────────┼────────────────┼─────────────┼─────────┤
│ 50,000     │ 53,546         │ 193 KiB     │ 5       │
├────────────┼────────────────┼─────────────┼─────────┤
│ 200,000    │ 211,746        │ 221 KiB     │ 10      │
├────────────┼────────────────┼─────────────┼─────────┤
│ 800,000    │ 809,284        │ 334 KiB     │ 15      │
└────────────┴────────────────┴─────────────┴─────────┘
```

comparison of twelve concurrent top-50 queries:
```
┌─────────┬──────────────┬────────────────┬────────┬────────┐
│         │ Total memory │ Peak per query │ Avg ms │ Max ms │
├─────────┼──────────────┼────────────────┼────────┼────────┤
│ PR view │ 3.01 GiB     │ 269 MiB        │ 738    │ 985    │
├─────────┼──────────────┼────────────────┼────────┼────────┤
│ RMT     │ 14.7 MiB     │ 4.44 MiB       │ 50     │ 84     │
└─────────┴──────────────┴────────────────┴────────┴────────┘
```

weave/trace_server/migrations/040_add_trace_tags.up.sql
benjreinhart: Is this something we can add later if needed? I don't currently have a need to query for users, but it's of course plausible we will want to expose a "tags you've added" style view or filter in the UI at some point. For now, the most important queries (and the ones we've designed for) are:

1. Selecting all tags that have been applied to conversation `<conversation_id>`
2. Filtering down all conversations by `{tag, trace_ts}` (find recent traces tagged with `<tag>`)

weave/trace_server/migrations/040_add_trace_tags.up.sql
benjreinhart: Agree we should probably rely on Clickhouse to set this value. However, what are your thoughts on using `DEFAULT` for that instead? MATERIALIZED is problematic I think if backfilling or moving between tables should we ever need to do that? Maybe overthinking it

weave/trace_server/migrations/040_add_trace_tags.up.sql
benjreinhart: Mostly because that's how I've done this type of thing before. On a technical level, I like an explicit table because then there's no relying on the query planner to do the right thing. That might not really be an issue in practice but feels nice to me. Happy to move to one if that's how we do things here or you have a strong preference.

I just ran this through claude and additionally it stated (among of few other less interesting reasons):

> Your headline query is "recent traces with tag X" — ORDER BY trace_ts DESC LIMIT 50. Reading from an explicit table whose sorting key matches gets the streaming optimize_read_in_order top-N optimization reliably. Projection support for read-in-order has improved but is version-dependent; worth testing on your version with EXPLAIN PIPELINE before committing.

weave/trace_server/migrations/040_add_trace_tags.up.sql
benjreinhart: Good catch, thanks, will reassess this.

weave/trace_server/migrations/040_add_trace_tags.up.sql
gtarpenning: indices are easier to add now than later, but just a thought

weave/trace_server/migrations/040_add_trace_tags.up.sql
gtarpenning: yeah default seems fine

weave/trace_server/migrations/040_add_trace_tags.up.sql
benjreinhart: Spent a bit more time thinking about this. Thoughts:

* We need to collapse values in this table to support the addition/removal of tags by humans or judges. Logic that deduplicates and grabs the most recent value must live somewhere. It would be nice if the application layer did not have to understand any nuance here. A view is only worth it if there is at least one very obvious query that will be reused again and again. Here we have two: grab all tags for a conversation and grab all conversations given a tag.
* If used with the correct filters (e.g. project_id, conversation_id for `trace_tags_current`), it seems clickhouse might correctly push the predicate down into the view ([enable_optimize_predicate_expression](https://clickhouse.com/docs/reference/settings/session-settings/enable-optimize-predicate-expression#enable_optimize_predicate_expression), default on). If that's true (I haven't tested), it would perform as well as if we wrote out the whole query by hand, including the predicates. This assumes clients correctly include the proper predicates, but that's also true if we do not provide a view in which case there is even more logic for the client to include.
* One solution that might be nice is to include the required predicates in the view as query parameters. E.g.:
  ```
  CREATE VIEW trace_tags_current AS SELECT ... WHERE project_id = {project_id:String} GROUP BY ...
  ```
  but I haven't really used these before. Thoughts?

All that said, I don't feel super strongly. This is a preference not a hill I'll die on. If we think this is a bad pattern or we avoid these then I'm happy to follow suit.

p.s., one interesting note from claude:

> For the tag-centric listing, exploit the sort order. Note that trace_tags_by_tag_current's GROUP BY keys are exactly the backing table's sorting key, and your schema comment about removals preserving trace_ts exists precisely so all audit rows for a (trace_id, tag) land adjacent in that sort order. That means optimize_aggregation_in_order = 1 can stream the aggregation with bounded memory instead of building a hash table. Whether it also short-circuits with ORDER BY trace_ts DESC LIMIT 50 is worth testing.

ro31337:

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
+-- Source of truth and full audit history for conversation tags. Rows are append-only.
+CREATE TABLE IF NOT EXISTS conversation_tags
+(
+    project_id      String,
+    conversation_id String,
+    trace_id        String,
+    tag             LowCardinality(String),
```
jtschoonhoven: These are user-defined and can be generated programmatically or randomly from an LLM, so I expect we shouldn't assume LowCardinality
```suggestion
    tag             String,
```

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
...
+
+    source          Enum8('judge' = 1, 'human' = 2),
+    source_version  LowCardinality(String),
```
jtschoonhoven: What's this?

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
+-- Source of truth and full audit history for conversation tags. Rows are append-only.
+CREATE TABLE IF NOT EXISTS conversation_tags
+(
+    project_id      String,
+    conversation_id String,
+    trace_id        String,
+    tag             LowCardinality(String),
+
+    source          Enum8('judge' = 1, 'human' = 2),
```
jtschoonhoven, line 20: Nit: could be more specific here
```suggestion
    judge_type          Enum8('llm' = 1, 'human' = 2),
```

jtschoonhoven, line 20: We probably also want something like `judge_host` or `source_type` so we can distinguish between direct API calls, self-hosted code scorers, and our own `agent_scoring_worker` etc.

weave/trace_server/migrations/040_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    conversation_id String,
+    trace_id        String,
+    tag             LowCardinality(String),
+
+    source          Enum8('judge' = 1, 'human' = 2),
+    source_version  LowCardinality(String),
+    wb_user_id      String,
+
+    -- Removal and re-add rows for a (trace_id, tag) must preserve trace_ts.
+    trace_ts        DateTime64(3),
```
jtschoonhoven: Traces have several timestamps (started_at, created_at, ended_at) so it would be good to be more specific. Probably this should be denormalized from `spans.ended_at`.

```suggestion
    trace_ended_at        DateTime64(6),
```

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    source_version  LowCardinality(String),
+    wb_user_id      String,
+
+    -- Removal and re-add rows for a (trace_id, tag) must preserve trace_ts.
+    trace_ts        DateTime64(3),
+    rationale       String DEFAULT '' CODEC(ZSTD(3)),
+    inserted_at     DateTime64(6) DEFAULT now64(6),
+    is_removed      UInt8 DEFAULT 0
```
jtschoonhoven: Nit: I prefer `deleted_at` for this

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
+-- Source of truth and full audit history for conversation tags. Rows are append-only.
+CREATE TABLE IF NOT EXISTS conversation_tags
```
jtschoonhoven, line 2: This is called `conversation_tags` but it appears to only work for turns. Probably it should be able to tag both conversations and turns, unless you've decided that tags are always turn-only.

In either case we should pick a different name.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    trace_id        String,
+    tag             LowCardinality(String),
+
+    source          Enum8('judge' = 1, 'human' = 2),
+    source_version  LowCardinality(String),
+    wb_user_id      String,
+
+    -- Removal and re-add rows for a (trace_id, tag) must preserve trace_ts.
+    trace_ts        DateTime64(3),
+    rationale       String DEFAULT '' CODEC(ZSTD(3)),
```
jtschoonhoven, line 29: I think we both agree that `confidence` isn't reliable, but I _do_ think it's useful for auditing and alignment. E.g. I might want to make a dataset of low-confidence tags, manually relabel them, and then tune the prompt based on that. So I'd add a column for confidence here (it can default to `1.0`)

weave/trace_server/migrations/040_add_conversation_tags.up.sql
jtschoonhoven: From Slack it sounds like you do want this table to be able to handle conversation-level tags. What timestamp would you associate with the conversation then?

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    inserted_at     DateTime64(6) DEFAULT now64(6),
+    is_removed      UInt8 DEFAULT 0
+)
+ENGINE = MergeTree
+PARTITION BY toYYYYMM(trace_ts)
+ORDER BY (project_id, tag, trace_ts, conversation_id, trace_id);
+
+CREATE MATERIALIZED VIEW IF NOT EXISTS conversation_tags_by_tag_mv
+TO conversation_tags_by_tag AS
```
jtschoonhoven, line 73: Is this whole MV just to support the use case "Show me all the tags for this project"? Is that not overkill?

Two other approaches could be:
1. In the UI we show the user only tags that have been applied in the last 30 days
2. We keep a registry of tags in MySQL. 

I recommend option 2. I think customers will want a source of truth for approved tags. E.g. this MV only knows about tags that have already been used at least once. And it might still show you historic tags that aren't relevant anymore to the project.

jtschoonhoven (commented):
If we're planning to use this table to tag conversations as well as turns, then it would be helpful to see more comments explaining how that will work. 

Also, I've been assuming that you don't ever plan to support tagging at the span level (like a single tool call), but it might be a good idea to be explicit about that.

weave/trace_server/migrations/040_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    project_id      String,
+    conversation_id String,
+    trace_id        String,
+    tag             LowCardinality(String),
+
+    source          Enum8('judge' = 1, 'human' = 2),
+    source_version  LowCardinality(String),
+    wb_user_id      String,
+
+    -- Removal and re-add rows for a (trace_id, tag) must preserve trace_ts.
```
jtschoonhoven: Griffin has been adding an `expire_at` column that defaults to `2100-01-01` which is a pattern to consider here. Someday we will want customers to be able to configure their retention policy.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: From slack

>The 'conversation' naming is intentional. I think of what we're doing in this new product as being largely based around the concept of conversations. These tags are scoped first to conversation (see ordering key) and secondarily by trace_id.
>
>This allows us to query by conversation id without a trace id. In the future, we could even decide to leave the trace_id key an empty string and then it would be considered a 'conversation level tag'.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: No this if for tag filtering: "Show me all conversations/turns for tag \<tag\>"

weave/trace_server/migrations/042_add_conversation_tags.up.sql
jtschoonhoven, line 2: Can you add some extra context on why we can't proceed using the `feedback` table?

Obviously your approach is cleaner and more suited to the task, but AFAICT it's a marginal improvement and it doesn't unlock anything new. _Not_ switching would allow us to continue sharing tags between the prototype and _Weave Classic_ ™️

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: Given there's currently no intention to use this value, I'd rather leave it out. If I'm not mistaken, this is an easy migration to run later if/when we want to make use of it.

If we must add it now, then I would add it to the table but not populate the value until we need to. Populating it requires the judge to emit it, which is something I'd not have the judge do unless necessary for a product feature we're building.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    conversation_id String,
+    trace_id        String,
+    source          Enum8('judge' = 1, 'human' = 2),
+    source_version  LowCardinality(String),
+    wb_user_id      String,
+    rationale       String DEFAULT '' CODEC(ZSTD(3)),
+    inserted_at     DateTime64(6) DEFAULT now64(6),
+    is_removed      UInt8 DEFAULT 0
+)
+ENGINE = MergeTree
```
jtschoonhoven, line 67: Why not `ReplacingMergeTree`?

weave/trace_server/migrations/040_add_conversation_tags.up.sql
benjreinhart: Good question. I would use the conversation's created at time, which I assume is the first trace started_at. In either case, I'm sure there is some value we'll be able to use here if/when we add that functionality.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
jtschoonhoven: Similar things to consider (we don't need all of these)
- Judge ID
- Judge Display Name
- Judge SHA1
- Judge Version

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: I normally do too as the timestamp adds an extra bit of info. I suppose there's no longer a good reason i can't use it here, but previously I was trying to use a replacing merge tree and some clickhouse functionality, which requires he 'deleted' indicator to be uint8. 

That said, what should we default it too? Clickhouse discourages nulls. We could still use it, or use a fixed timestamp and compare against that in queries?

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: "Judge version" is source_verrsion

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: I started with that and then removed it due to complexity.

We have two tables, two ordering keys, two types (judge vs human), and the ability to mark deleted. When reasoning over the different replacement rules and how that affected the ordering key it ended up feeling not worth it. For example, at one point claude was encouraging me to add the `source` to the ordering key to ensure human records have precedence over the judges, but that threw off my queries.

I want a record of what happened. There's the possibility a judge or human applies a tag and is later shadowed by the other. RMT removes the record of that happening eventually. This can get especially confusing if you leverage the deleted behavior of RMT.

Lastly, RMT is dedup eventually. You may still need to add logic to your queries to ensure you always get the right results. If I need to add this logic to my queries, then the RMT and its complexity bought me essentially nothing, as we don't expect tons of replacements.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: It is a feature not a bug that they are different.

1. The queries enabled by this model are much improved. Specifically:
    1. Query efficiently by conversation_id. Useful when loading all tags for N conversations when browsing conversations. this is a query we need to make today for the UI we designed. Also useful when pulling up individual conversations.
    2. Filter all conversations by a given tag (a tag-centric view).
3. This data model nicely fits the feature we want to build without additional complexity or legacy cruft (like the ref system) which i'm trying to avoid carrying forward.

As for sharing tags:

1. I don't know when or how the existing people will move over. When they do, there are issues either way. Signals will not be carried over because that's a different feature that gave users additional flexibility. So they wouldn't show up in the new UI. We can most likely do a one-time manual backfill for the handful of customers that may need this if/when they're ready.
2. We have an opportunity to redesign some features from first principles. I want to take that opportunity and knowledge of the existing systems and design something that is better suited to the task moving forward. It's not perfect, but I think it's an improvement that we'll leverage.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: Maybe we should chat offline about this, but i would prefer to capture that in something like a "conversation_tag_runs" table which captures metadata surrounding the worker that runs judges (or code scoring in the future). That can also capture things like latency, cost, errors/failures, sampling, etc. that we can surface to users if needed for reasons of transparency (e.g., you're getting charged for runs... here's all the runs and the cost of each).

I briefly mentioned this table to @gtarpenning but decided against it for now as we don't have plans to use those features. I can be convinced otherwise.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: As of now they're not random, the user must define a known set in the UI and the LLM must choose from that set.

That said, you're not wrong. I had switched this at one point and somehow claude convinced me that in the case where there are a lot it falls back to behaving like a `String`. But I think that sounds sus.

jtschoonhoven: It's hard to evaluate because it's hard to anticipate how this will work in the future. Obviously a project with 10k different tags is totally pathological and ideally we won't even allow that.

Maybe we keep this `LowCardinality` and we try and put some guard rails in place that limit the number of unique tags projects can create.

I'll defer to you. `String` is more future-proof but `LowCardinality(String)` will have better performance for well-behaved projects.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
jtschoonhoven: I like the idea of `conversation_tag_runs` but I don't see any downside to capturing some of that data here too. Clickhouse tends to encourage denormalizing anyway.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
jtschoonhoven: Oh, right, the Nullable thing is annoying. Maybe `deleted_at_ms UInt8 DEFAULT 0`?

But I'm ok with the `is_removed` pattern too.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
jtschoonhoven: I like that the new product is organized around `conversation_id` but I still think it's confusing and unnecessary to name this table `conversation_tags`. Why not something more generic like `agent_tags` or something?

weave/trace_server/migrations/042_add_conversation_tags.up.sql
jtschoonhoven: I'm ok with this, feel free to resolve

weave/trace_server/migrations/040_add_conversation_tags.up.sql
jtschoonhoven: Conversations might also get scored multiple times as new turns arrive, so alternatively this could be the timestamp of the most recent turn at the time the conversation was tagged. Just an idea, I haven't fully thought this through.

benjreinhart: > timestamp of the most recent turn at the time the conversation was tagged.

👍🏻 

@gtarpenning and I discussed this. I think we can use the trace_ended_at here too and keep it the last turn ended at at the time of labeling.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: Switched to string

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: renamed judge_version

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: Since these are all the same now, this will be easy to add once needed. Today we only have one judge (ours).

weave/trace_server/migrations/040_add_conversation_tags.up.sql
benjreinhart: Good call, thanks

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: I was about to switch to deleted_at but then realized that we already have inserted at, which would be the same value. So in this case we do have the timestamp. This is easier to user in queries with a more sane default and also allows us to switch the MV table to a ReplacingMergeTree using the uint8 deleted at should we want to do that.

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: IMO, if "conversation_tags" is confusing, then "agent_tags" is no less confusing. I think conversation_tags makes sense given how the data is organized.

weave/trace_server/migrations/040_add_conversation_tags.up.sql
benjreinhart: Added, thanks

weave/trace_server/migrations/042_add_conversation_tags.up.sql
```diff
@@ -0,0 +1,88 @@
...
+
+    -- When source=human, this is the id of the user.
+    wb_user_id      String,
+
+    -- The reasoning behind the tag being applied.
+    rationale       String DEFAULT '' CODEC(ZSTD(3)),
+
+    -- 1 = this row records the tag being removed
+    --
+    -- deleted_at in this case will be the same as inserted_at,
```
gtarpenning, line 33: but we dont have a `deleted_at`?

gtarpenning (approved):

weave/trace_server/migrations/042_add_conversation_tags.up.sql
benjreinhart: Yeah sorry the comment is maybe not clear. I was trying to communicate why `is_removed` over `deleted_at`. So this comment was trying to say "not using deleted_at because..."

PR #7720, opened 2026-08-11
feat(weave): add insights signature read and write APIs

gtarpenning:
## Summary
- Adds the trace-server interface, Python HTTP client, external project-id adapter, and ClickHouse handlers for signature writes and reads; wandb/core#50334 contains the authenticated route layer for these methods.
- Each write names one signature type and one config digest. Row `id` is `uuid5` over `(project_id, config_sha, conversation_id, turn, signature)`, so a retried batch reuses its ids and the merge collapses them; a read landing between the retry and the merge still sees both. Malformed candidates are dropped and unknown taxonomy labels are repaired, each counted per gate without failing the batch.
- The read surface covers turn and conversation drilldown, grouped facets and ranking, and keyset-paged vector export. The day window is required, `limit` is capped and capped lower when `include_vector` is set, and a field the chosen `mode` does not read is rejected rather than ignored.
- Migration 042 defaults `intent_signatures.sentiment` to `''`, the sentinel for a label the judge did not emit.

## Performance
On ClickHouse 26.6, the tuple cursor kept 49/49 granules and read 24,576 rows at page 1, 122,880 at page 1,000, and 408,192 at page 3,900. The explicit `OR` predicate pruned to 13/49 granules and held pages 1,000 and 3,900 at 16,384 rows.

## Testing
Every SQL builder is pinned to its full query and parameter dictionary. Real-ClickHouse endpoint tests cover both signature types, project isolation, every writer gate, an idempotent replay through a forced merge, turn-scoped measures on both signature types, and a 25-row cursor walk across a day boundary.


weave/trace_server/insights/clickhouse.py
```diff
@@ -0,0 +1,126 @@
...
+            "one write carries one signature type: intents and failures name "
+            "different configs"
+        )
+    if req.failures:
+        writer.validate_config_sha("failure", req.config_sha)
+        rows, gate_counts = writer.prepare_failures(
+            req.project_id, req.config_sha, req.failures
+        )
+        written = InsightWriteCounts(failures=_insert(server, FAILURE_TABLE, rows))
+    else:
```
jtschoonhoven, line 56:
```suggestion
    elif req.intents:
```

jtschoonhoven: ...then maybe also raise `InvalidRequest` if req.intents and req.failures are both empty

weave/trace_server/insights/writer.py
```diff
@@ -0,0 +1,224 @@
...
+        raise RuntimeError(f"unsupported normalization_version {normalization_version}")
+    collapsed = " ".join(unicodedata.normalize("NFKC", text).split())
+    return collapsed.rstrip(" .").casefold()
+
+
+def prepare_intents(
+    project_id: str,
+    config_sha: str,
+    candidates: list[IntentSignatureCandidate],
+) -> tuple[list[Row], GateCounts]:
```
jtschoonhoven, line 64: Coming in cold, I have no idea what GateCounts is. Probably worth adding a docstring.

weave/trace_server/insights/writer.py
```diff
@@ -0,0 +1,224 @@
...
+        candidate.signature,
+        normalization_version=cfg.extraction.normalization_version,
+    )
+    if not signature:
+        gate_counts["empty_signature"] += 1
+        return None
+    if len(candidate.vector) != cfg.embedding.dimensions:
+        gate_counts["vector_dimensions"] += 1
+        return None
+    return signature
```
jtschoonhoven, line 205: If the point of this function is to generate a canonical signature, then why is it called `_gate`? Seems needlessly opaque.

weave/trace_server/insights/writer.py
```diff
@@ -0,0 +1,224 @@
...
+    rows: dict[uuid.UUID, Row] = {}
+
+    for candidate in candidates:
+        signature = _gate(cfg, candidate, gate_counts)
+        if signature is None:
+            continue
+        affected = sorted(set(candidate.affected_trace_ids))
+        if candidate.current_trace_id not in affected:
+            gate_counts["ungrounded_attribution"] += 1
+            continue
```
jtschoonhoven, line 117: It seems that `gate_counts` is actually tracking validation errors, and a "gate" is a label for a specific validation failure. Can we call these `validation_errors` instead of `gate_counts`? I think that would be much more clear.

weave/trace_server/insights/write_types.py
```diff
@@ -0,0 +1,184 @@
...
+        return self
+
+
+class InsightSignaturesWriteRes(BaseModelStrict):
+    written: int = Field(
+        description="Rows that reached the table, after gating and in-batch dedupe. "
+        "One request writes one signature type, so one count.",
+        examples=[42],
+    )
+    dropped: dict[InsightWriteGate, int] = Field(
```
jtschoonhoven: Tiny nit: I think `success_count` and `failure_counts` would be a bit more descriptive.

weave/trace_server/insights/writer.py
```diff
@@ -0,0 +1,224 @@
...
+from weave.trace_server.insights import config
+from weave.trace_server.insights.enums import InsightSignatureType, InsightWriteGate
+from weave.trace_server.insights.write_types import (
+    FailureSignatureCandidate,
+    InsightSignatureCandidateBase,
+    IntentSignatureCandidate,
+)
+
+FALLBACK_LABEL = "other"
+UNKNOWN_LANGUAGE = "und"
```
jtschoonhoven, line 24: This is the default from the migration, right? Do we need to store this?

weave/trace_server/insights/writer.py
```diff
@@ -0,0 +1,224 @@
...
+
+# Namespace for the derived row id. Changing it re-mints every id and so orphans
+# every stored row from its replays; it is permanent.
+ROW_ID_NAMESPACE = uuid.UUID("320faf23-fcac-48f0-9161-54f2ab433304")
```
jtschoonhoven, line 28: I haven't read ahead yet, but on first read this is confusing

weave/trace_server/insights/writer.py
```diff
@@ -0,0 +1,224 @@
...
+            f"{signature_type} config_sha {claimed!r} does not match the deployed "
+            f"config {actual!r}"
+        )
+
+
+def canonicalize_signature(text: str, *, normalization_version: int) -> str:
+    """The stored form, so `GROUP BY signature` groups identity.
+
+    Versioned by the config's `normalization_version`, which lives inside
+    `config_sha`, so changing it is a visible digest move.
```
jtschoonhoven, line 52: Does anything enforce this? Or would I need to read this docstring and remember to update `normalization_version` manually?

weave/trace_server/insights/writer.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+    if normalization_version != 1:
+        raise RuntimeError(f"unsupported normalization_version {normalization_version}")
```
jtschoonhoven, line 55: Is this check worthwhile? e.g. I can imagine us incrementing the version and forgetting to update this check and having it blow up in prod

weave/trace_server/insights/write_types.py
```diff
@@ -0,0 +1,184 @@
...
+        default_factory=dict,
+        description="Gate name to count, for the gates that fired. "
+        "`empty_signature`, `vector_dimensions`, `ungrounded_attribution`, and "
+        "`duplicate_in_batch` discard the candidate; `unknown_category`, "
+        "`unknown_sentiment`, and `unknown_severity` repair the label and keep it. "
+        "Neither outcome fails the batch, so callers watch this to catch a drifting "
+        "judge.",
+        examples=[{"unknown_category": 3, "empty_signature": 1}],
```
jtschoonhoven, line 183: How about `success_count`, `warnings_count` ("unknown_category", "unknown_sentiment", and "unknown_severity") and `failure_count` ("empty_signature", "vector_dimensions", "ungrounded_attribution", "duplicate_in_batch")?

E.g. it's confusing that "unknown_severity" is included in "dropped" if nothing is actually dropped.

weave/trace_server/insights/enums.py
jtschoonhoven, line 1: "enums.py" is probably not the best name for a module that contains Literal types and constants, but no Enums. Probably should be renamed "types.py" and optionally the constants could live in "constants.py"

weave/trace_server/insights/read_types.py
```diff
@@ -0,0 +1,257 @@
...
+        examples=["2026-08-01"],
+    )
+    day_end: datetime.date = Field(
+        description="Window end, inclusive.", examples=["2026-08-14"]
+    )
+    mode: InsightQueryMode = Field(
+        default="rows",
+        description="Result shape over the same window and filters. `rows` returns one "
+        "record per occurrence into `rows`, for drilldown and vector export, and is the "
+        "only mode that pages. `groups` returns one record per distinct `group_by` "
```
jtschoonhoven, line 162: If "row mode" and "group mode" have different shapes and different behavior (paging) then that strongly suggests they should be two different endpoints.

Our current APIs are all hard to use (even for agents) and I think we should try really hard to keep things simple. I don't see any benefit to this kind of polymorphism.

weave/trace_server/insights/read_types.py
```diff
@@ -0,0 +1,257 @@
...
+        "`order='key'`.",
+    )
+
+    group_by: list[InsightGroupField] = Field(
+        default_factory=lambda: ["signature"],
+        description="`groups` only: the facets to aggregate on. `day` buckets by "
+        "`toDate(trace_started_at)`. `sentiment` is intent-only and `severity` "
+        "failure-only; the wrong pairing is an error, not an empty column.",
+        examples=[["signature"], ["category", "day"]],
+    )
```
jtschoonhoven, line 213: Does the client really need this level of control over the GROUP BY? Before working at W&B I pretty much never saw an HTTP API that accepted custom "group by" but we do it all the time.

If there are 2 or 3 ways we expect clients to want to group this data then I'd recommend we break this up into 2 or 3 different endpoints. This will also help us keep the query builder logic nice and simple and less prone to injection.

weave/trace_server/insights/read_types.py
```diff
@@ -0,0 +1,257 @@
...
+        description="Restrict to these canonicalized signatures. The way to fetch the "
+        "occurrences behind an offline cluster.",
+        examples=[["cancel my subscription and get a refund"]],
+    )
+
+    include_vector: bool = Field(
+        default=False,
+        description="`rows` only. Off by default because the embedding dwarfs the rest "
+        f"of the row, and caps `limit` at {MAX_VECTOR_PAGE_LIMIT}.",
+    )
```
jtschoonhoven, line 193: This is less severe than "mode" but I still don't love it when an endpoint's response changes shapes depending on the payload it receives.

What's the use case where the client _sometimes_ wants vectors and sometimes doesn't? Why do we think this optimization is necessary?

weave/trace_server/insights/read_types.py
```diff
@@ -0,0 +1,257 @@
...
+        if named:
+            raise ValueError(f"mode={self.mode!r} does not read {', '.join(named)}")
+        if self.cursor is not None and self.order != "key":
+            raise ValueError(f"cursor requires order='key', got {self.order!r}")
+        if self.include_vector and self.limit > MAX_VECTOR_PAGE_LIMIT:
+            raise ValueError(
+                f"include_vector caps limit at {MAX_VECTOR_PAGE_LIMIT}, got {self.limit}"
+            )
+        if not self.group_by:
+            raise ValueError("group_by needs at least one field")
```
jtschoonhoven, line 240: See earlier comments about how polymorphism makes APIs harder to use 🙃

jtschoonhoven (commented):
`InsightSignaturesQueryReq` is trying to do too much and it should be factored out into more than one endpoint.

PR #7721, opened 2026-08-11
fix(weave_ts): skip non-succeeded Anthropic batch results

ro31337:
## Description

- Fixes [WB-38528](https://coreweave.atlassian.net/browse/WB-38528)

This is a prerequisite for the next change in this file, which fixes token and cost counts for cached Anthropic calls. That change adds to the same summarizer, so this path needs to stop throwing first.

Anthropic has a batch API: you send a pile of requests, then read the results back one by one. Some can fail, get canceled, or expire — normal, the API has a counter for each.

Only a successful result carries a `message`. Our code read `message` off every result, so for the failed ones it got `undefined`. Then it crashed.

The crash lands badly. It comes out of the user's own `for await` loop, so their process dies unless they wrapped it. And the trace vanishes — not a broken one, nothing at all, because a call's start and end are sent together by default. One failed request out of a hundred is enough.

I found it reading this file for that token change, not from a report — there are none, so I do not know how often it hits. One hit is bad enough.

The obvious patch is to skip the empty ones in the summarizer, where the crash happens. That stops the crash, but the bad value is already in the recorded state by then. So the fix goes earlier, where results are collected: the type now spells out the four kinds the API really returns, and the code checks which one it got before reading it.

There is a test for batches and it passes. It used `result.type: 'message'`, which the API never sends — a batch that cannot exist, so the failed case was never covered.

Bottom line: one failed request used to kill the caller and lose the trace. Now the failed ones are skipped and the rest is recorded. The trace still does not say how many were skipped; that needs a new field.

## Testing

Three new tests cover a batch with all four outcomes, a batch where everything failed, and a `break` halfway through; all three crash on master and pass here. Locally that file is 11/11 and the full default Jest project is 490 passed, with 4 pre-existing failures in the OpenAI instrumentation suites that also fail on master, and tsc, eslint and prettier are clean.

[WB-38528]: https://coreweave.atlassian.net/browse/WB-38528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -344,6 +344,20 @@ deterministic.
   flattened `weave.integration.meta.*` provenance. OTel scalar metadata stays
   typed; non-scalar values are stringified.
 
+### TypeScript Anthropic message batches
+
+- A batch result is a discriminated union and only its `succeeded` variant
+  carries a `message`; a batch ends when every request has succeeded, errored,
+  been canceled or expired. Read `message` without checking the discriminator
+  and the reducer stores `undefined`, which the usage summarizer then throws on,
+  out of the stream iterator's `finally`.
+- `patchAnthropicMessagesCreate` wraps `Messages.prototype.create` in a `Proxy`,
+  so property reads still reach the mock behind it and a test can reconfigure it
+  through the patched prototype. `patchBatchApi` assigns `op()`'s plain wrapper
+  to `Batches` `create`, `retrieve` and `results` instead, so for those a test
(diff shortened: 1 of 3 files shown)


rgao-coreweave (approved):
LGTM

thanks for fixing!

PR #7722, opened 2026-08-11
chore(weave_ts): hide deprecated Session const from TypeDoc output

dbrian57:
## Description
Adds the `@hidden` tag to `export type Session = Conversation;` to keep TypeDoc from duplicating it i[n the docs](https://wb-21fd5541-typedoc-upgrade.mintlify.site/weave/reference/typescript-sdk/type-aliases/session-1).


drtangible (approved):
👍 👍 👍

PR #7725, opened 2026-08-11
fix(weave_ts): report inclusive Anthropic prompt tokens

ro31337:
## Description

- Fixes [WB-38555](https://coreweave.atlassian.net/browse/WB-38555)

If you turn on prompt caching with Anthropic in the TypeScript SDK, the traces lie about tokens. Say a call has 21,500 tokens of cached prompt and only 4 fresh ones. The trace shows 4. The cost is wrong too: near zero on streaming calls, negative on plain ones.

That is because Anthropic does not send one prompt number. From their [prompt caching docs](https://platform.claude.com/docs/en/build-with-claude/prompt-caching):

> The `input_tokens` field represents only the tokens that come **after the last cache breakpoint** in your request - not all the input tokens you sent.

> `total_input_tokens = cache_read_input_tokens + cache_creation_input_tokens + input_tokens`

So the real prompt size is those three added up. Weave wants that single total and subtracts the two cache numbers itself, because they are billed at their own rates. We were passing Anthropic's numbers through untouched, so Weave subtracted them from a total that never had them in it.

Python already fixed this in [#7561](https://github.com/wandb/weave/pull/7561) ([WB-37043](https://coreweave.atlassian.net/browse/WB-37043)), so I figured TypeScript should just do the same. It adds the three counters up now, and keeps the two cache counts next to the total so the server can still subtract them. That covers plain calls, streaming and batches.

Streaming still did not work after that, so I went digging. During a stream Anthropic keeps resending the usage as a running total, and puts `null` in the fields it is not restating. Our code copied each update over the stored usage, nulls and all, which wiped the cache counts that came in at the start. It now copies only the fields that have a value, which is what Anthropic's own client does.

I also found the Claude Agent SDK tracer already had this same sum written out by hand. I did not want to write it twice, so it moved into a small shared helper and both use it.

Bottom line: a cached call used to record 4 prompt tokens and a nonsense cost, now it records the real 21,504. The response your own code gets back still has Anthropic's numbers in it, untouched. The only other change is that streaming and batch summaries now always carry the two cache counts, zero when nothing was cached, which is why four old tests move.

## Testing

Three new tests, one for each way you can call Anthropic — plain, streaming and batch — all with caching on, and all three fail on master. Two more cover the helper on its own; locally this file is 16/16, the full Node suite picks up no new failures, and tsc, eslint and prettier are clean.

[WB-38555]: https://coreweave.atlassian.net/browse/WB-38555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-37043]: https://coreweave.atlassian.net/browse/WB-37043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -358,6 +358,22 @@ deterministic.
   reconfigures the mock it captured before patching, or reaches it through
   `__wrappedFunction`.
 
+### TypeScript Anthropic token accounting
+
+- Anthropic reports `input_tokens` as fresh, uncached prompt only and bills the
+  rest through `cache_read_input_tokens` and `cache_creation_input_tokens`.
+  Weave's cost math subtracts those two from the prompt total, so the usage
+  summary has to carry an inclusive `input_tokens`. `totalInputTokens()` in
+  `integrations/anthropicUsage.ts` is where that sum lives, for this integration
+  and for the Claude Agent SDK OTel tracer; it mirrors the Python
+  `total_input_tokens()`.
+- Only the summary is normalized. The provider's own `usage` object stays
+  untouched, so the response the caller receives keeps Anthropic's own numbers.
+- Streaming `message_delta` usage is cumulative, and every field except
+  `output_tokens` is nullable. Skip the null ones, as the vendor's own client
+  does, and overwrite rather than add — a delta carries the running total, not
+  an increment.
+
 ### TypeScript custom type round trip
 
 - `client.get()` rebuilds a `WeaveImage` from a `PIL.Image.Image` payload and a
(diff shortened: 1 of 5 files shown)


rgao-coreweave (approved):
LGTM

PR #7726, opened 2026-08-11
feat(weave): federated identity support for weave

RyanBuccellato:
## Description
https://coreweave.atlassian.net/browse/WB-33746

Adds support for[ federated identity](https://docs.wandb.ai/platform/hosting/iam/identity_federation) to weave. We currently support this in the standard wandb SDK, sort of reimplementing it for the weave SDK since they dont really share code today.

The way federated identity works is the user provides a JWT, we send that to gorilla to exchange for a Bearer token. This token is saved in a CREDENTIALS_FILE like the standard wandb SDK. This token will then get fed into the weave backend, which then forwards it to gorilla (which already knows how this credential works). The token will eventually expire and then the SDK will need to exchange it out for a new one.

## Testing

Have a local script that went thru all the weave endpoints to test that they authed properly with federated identity. The one exception is inference (since local testing for that is a bit more involved) but inference requires a separate backend change to support anyways.


RyanBuccellato:

weave/wandb_interface/context.py
```diff
@@ -1,8 +1,21 @@
...
+    WandbCredentials,
+    get_wandb_credentials,
+)
+
+
+def get_wandb_auth_context() -> WandbCredentials | None:
+    return get_wandb_credentials()
 
 
 def get_wandb_api_context() -> str | None:
```
RyanBuccellato, line 10: not really used anymore but kept for compatibility (auth_context returns a typed value so we know what type of cred it is). Could also just remove/consolidate not sure how big of a concern that actually is. The current name is sort of hard to understand though, `api context` but returns credentials. I would personally just remove in favor auth_context but didn't want to add potentially breaking changes

weave/wandb_interface/auth.py
```diff
@@ -0,0 +1,266 @@
...
+    def wal_seed(self) -> str:
+        return f"{self.base_url}\0{self.token_file.expanduser().resolve()}"
+
+    def access_token(self) -> str:
+        with self._lock:
+            cached = self._read_cached_token()
+            if cached is not None:
+                return cached
+            now = time.monotonic()
+            if self._exchange_failure is not None:
```
RyanBuccellato, line 97: added some backoff since I don't want our token endpoint getting spammed when auth fails for w/e reason. It auto-refreshes when the credential is rotated. Our models SDK has no auto-refresh mechanic but wanted to add it for weave since I'd imagine weave is used more like a middleware/server situation (where it needs to be able to recover, can't just "offline sync" later).

weave/wandb_interface/auth.py
```diff
@@ -0,0 +1,266 @@
...
+        self.credentials_file.parent.mkdir(parents=True, exist_ok=True)
+        temporary_file = self.credentials_file.with_name(
+            f".{self.credentials_file.name}.{os.getpid()}.{threading.get_ident()}.tmp"
+        )
+        try:
+            temporary_file.write_text(
+                json.dumps(data, indent=2) + "\n",
+                encoding="utf-8",
+            )
+            os.chmod(temporary_file, 0o600)
```
amwarrier, line 232: nit: on Windows, `os.chmod(..., 0o600)` is effectively a no-op. but I guess the gap is there in wandb sdk as well. so I think its ok

weave/wandb_interface/context.py
```diff
@@ -1,8 +1,22 @@
...
-    """Return the current W&B API key, reading from env var or netrc."""
-    return weave_wandb_api_key()
+    """Return the current W&B credential, reading from env var or local files"""
+    credentials = get_wandb_auth_context()
+    if isinstance(credentials, ApiKeyCredentials):
```
amwarrier: nit: both subclasses return the right thing from `bearer_token()` already, so this can collapse to 
```python
return credentials.bearer_token() if credentials else None. 
```

Avoids adding a third isinstance next time we grow a credential type.

weave/trace/weave_init.py
```diff
@@ -228,14 +242,21 @@ def _setup_conversation_tracing(entity: str, project: str, api_key: str | None)
     _conversation_span_exporter = exporter
 
 
-def _conversation_headers(project_id: str, api_key: str | None) -> dict[str, str]:
-    """Export headers routing agent spans: project via project_id, creds via Basic auth."""
-    # Auth mirrors the rest of weave (BasicAuth base64("api:<key>")); one source
-    # of truth so setup and reroute never diverge into stale creds.
+def _coerce_credentials(
+    auth: str | WandbCredentials | None,
```
amwarrier: nit: `_coerce_credentials` exists to let these functions accept either `str` or `WandbCredentials`, but every internal caller in  this PR already passes a WandbCredentials. Is the str branch covering an external caller I'm missing? If not, would prefer to drop the union from the signatures and require a WandbCredentials  — otherwise the new abstraction leaks back into "sometimes a string, sometimes an object" everywhere it's used, which is what the  abstraction was meant to eliminate.

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -105,15 +110,26 @@ def ensure_project_exists(
...
+    def set_auth(self, auth: tuple[str, str] | WandbCredentials) -> None:
         """Set authentication credentials.
 
         Args:
             auth: Tuple of (username, password) for authentication.
         """
-        self._username, self._password = auth
+        self._credentials = None
+        if isinstance(auth, ApiKeyCredentials):
+            self._username = "api"
```
amwarrier:
```python
WANDB_API_BASIC_USERNAME = "api"
```

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -105,15 +110,26 @@ def ensure_project_exists(
...
+            self._password = auth.api_key
+        elif isinstance(auth, WandbCredentials):
+            self._username = ""
```
amwarrier: good to add some comments here on whats the magic here.

weave/trace_server_bindings/stainless_remote_http_trace_server.py
```diff
@@ -105,15 +110,26 @@ def ensure_project_exists(
     def from_env(cls, should_batch: bool = False) -> Self:
         return cls(weave_trace_server_url(), should_batch)
 
-    def set_auth(self, auth: tuple[str, str]) -> None:
+    def set_auth(self, auth: tuple[str, str] | WandbCredentials) -> None:
```
amwarrier, line 100:
```python

def set_auth(self, auth: tuple[str, str] | WandbCredentials) -> None:
    if isinstance(auth, tuple):
        auth = ApiKeyCredentials(auth[1])   # normalize at the door
    self._credentials = auth
    self._rebuild_client()

def _rebuild_client(self) -> None:
    # Stainless auth is disabled; we own the Authorization header.
    self._stainless_client = StainlessClient(
        base_url=self.trace_server_url,
        username="",
        password="",
        default_headers=self._compose_headers(),
        batch_requests=False,
    )

def _compose_headers(self) -> dict[str, str]:
    headers = self._extra_headers.copy()
    headers["Authorization"] = self._credentials.authorization_header()
    if retry_id := get_current_retry_id():
        headers["X-Weave-Retry-Id"] = retry_id
    return headers

```

weave/trace/weave_init.py
RyanBuccellato: makes sense yeah I can update that Was trying to keep support for the old way because didn't want to break any existing users who were for w/e reason importing internal fns but it's probably fine to assume ppl aren't doing that or if they are they will just need to update. And should make it less confusing to just have 1 type

weave/trace_server_bindings/remote_http_trace_server.py
```diff
@@ -66,7 +67,7 @@ def __init__(
         should_batch: bool = False,
         *,
         remote_request_bytes_limit: int = REMOTE_REQUEST_BYTES_LIMIT,
-        auth: tuple[str, str] | None = None,
+        auth: tuple[str, str] | httpx.Auth | None = None,
```
amwarrier: Nit: RemoteHTTPTraceServer isn't in `weave/__init__.py` and its package's `__init__.py` is empty — it's internal transport. Both auth= callers are in-repo (`weave_init.py`, `wal_sender.py`) and there are none in `wandb/core`, so tightening both `__init__` and `set_auth` to `WandbCredentials | None`, migrating `wal_sender.py:454` to `ApiKeyCredentials(args.api_key)`, and dropping the dead `httpx.Auth` arm is safe — matches the cleanup already done on the Stainless side.

weave/wandb_interface/auth.py
```diff
@@ -0,0 +1,266 @@
...
+    def requests_auth(self) -> _WandbRequestsAuth:
+        return _WandbRequestsAuth(self)
+
+
+class ApiKeyCredentials(WandbCredentials):
+    def __init__(self, api_key: str) -> None:
+        self.api_key = api_key
+
+    def authorization_header(self) -> str:
+        encoded = base64.b64encode(f"api:{self.api_key}".encode()).decode()
```
amwarrier: Nit: the magic string "api" (the W&B Basic-auth username sentinel) also appears in weave/durability/wal_sender.py:454 as auth=("api", args.api_key). Would be worth extracting to a module-level constant here — something like:
```python
WANDB_API_BASIC_USERNAME = "api"
```

so there's one source of truth. If you also apply the RemoteHTTPTraceServer tightening (other comment), wal_sender.py:454 becomes ApiKeyCredentials(args.api_key) and this constant becomes the sole reference.

amwarrier (approved):

weave/wandb_interface/auth.py
```diff
@@ -0,0 +1,266 @@
...
+                EXPIRES_AT_FORMAT
+            ),
+        }
+
+        self.credentials_file.parent.mkdir(parents=True, exist_ok=True)
+        temporary_file = self.credentials_file.with_name(
+            f".{self.credentials_file.name}.{os.getpid()}.{threading.get_ident()}.tmp"
+        )
+        try:
+            temporary_file.write_text(
```
amwarrier, line 228: a read-only rootfs container without a writable volume mount at ~/.config/wandb/ every request becomes an uncaught OSError. Consider catching once and falling back to in-process caching with a warn-once log — the read path already treats OSError as a cache miss.

ro31337 (approved):

PR #7728, opened 2026-08-12
fix(weave): include OTEL spans in root usage stats

gtarpenning:
## Summary
- Root-only call stats keep native SDK root summaries and include OTEL span usage, where usage remains on descendants.
- Call metrics remain root-only; OTEL detection uses `otel_dump` in both ClickHouse and in-memory backends.

## Testing
ClickHouse and fake-backend call-stats tests; complete SQL assertions for `calls_merged` and `calls_complete`.

### DIFF
--- a/weave/trace_server/calls_query_builder/usage_query_builder.py
+++ b/weave/trace_server/calls_query_builder/usage_query_builder.py
@@ -106,14 +106,14 @@ def build_usage_query(
     tz_param = pb.add_param(req.timezone or "UTC")
     bucket_interval_param = pb.add_param(granularity_seconds)
 
-    # Root-only usage keeps native root summaries plus individual OTEL span usage.
-    calls_filter = req.filter
-    include_otel_span_usage = bool(calls_filter and calls_filter.trace_roots_only)
-    if calls_filter is not None and calls_filter.trace_roots_only:
-        calls_filter = calls_filter.model_copy(update={"trace_roots_only": None})
+    # Root-only usage keeps native root summaries plus individual OTEL span usage,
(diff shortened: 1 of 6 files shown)


weave/trace_server/calls_query_builder/usage_query_builder.py
```diff
@@ -105,8 +106,14 @@ def build_usage_query(
...
 
+    # Root-only usage keeps native root summaries plus individual OTEL span usage.
+    calls_filter = req.filter
+    include_otel_span_usage = bool(calls_filter and calls_filter.trace_roots_only)
+    if calls_filter is not None and calls_filter.trace_roots_only:
+        calls_filter = calls_filter.model_copy(update={"trace_roots_only": None})
+
     # Build filter clauses - applied directly in WHERE
-    where_filter_sql = build_calls_filter_sql(req.filter, pb, read_table)
+    where_filter_sql = build_calls_filter_sql(calls_filter, pb, read_table)
```
jtschoonhoven: This seems kinda janky. Cloning `req.filter` so you can set `trace_roots_only = None` _just_ for `build_calls_filter_sql` seems surprising. I'd rather this did something more explicit, like passing a `force_all_spans=True` kwarg to `build_calls_filter_sql`.

jtschoonhoven (approved):

weave/trace_server/calls_query_builder/usage_query_builder.py
gtarpenning: Agreed, done in fd870652b9.

PR #7729, opened 2026-08-12
feat(weave): add parent-call columns to spans

ro31337:
Four PRs, in review order:

2. Migration: **this PR**.

## JIRA Issue(s)

## Description

Migration `040`: two String columns on `spans`, `parent_call_id` and `parent_call_trace_id`, with a skip index on each. They hold the id of the `@weave.op` call that was running when a span started, which is what turns "show me the spans this call produced" into a normal filter.

This started inside #7704 and is split out on review feedback: if a migration fails we do not want merged code that references it. So the columns land first and the code that writes them follows in #7730. Nothing references them yet.

Several open PRs also claim `040`, so whoever merges second renames. Worth re-checking right before this one goes in.

## Backward compatibility

Additive and empty. Both columns default to `''`, existing rows need no backfill, and no code path reads or writes them until #7730 merges. The down migration drops both indexes and both columns.

## Testing

CI applies this against a real ClickHouse in three topologies (1s1r, 1s3r, 2s2r) and re-runs the ups to diff the schema either side; all three are green. Locally the migrator unit suite is 110 passed.

### DIFF
--- a/weave/trace_server/migrations/040_add_span_parent_call_columns.up.sql
+++ b/weave/trace_server/migrations/040_add_span_parent_call_columns.up.sql
@@ -1,18 +1,3 @@
--- Link a @weave.op call to the agent spans it produced. The SDK stamps
--- weave.parent_call.id / weave.parent_call.trace_id on spans that start while
--- a call is running, and these columns are the promoted form. Populated going
--- forward only, while `spans` remains the source of truth.
---
--- bloom_filter(0.001) rather than the 0.01 of idx_eval_run_id in 037: both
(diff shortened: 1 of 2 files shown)


weave/trace_server/migrations/040_add_span_parent_call_columns.up.sql
```diff
@@ -0,0 +1,24 @@
+-- Link a @weave.op call to the agent spans it produced. The SDK stamps
```
gtarpenning: i think we probably don't need this preemable, its all in git pr/notes if we need it

gtarpenning (approved):

PR #7730, opened 2026-08-12
feat(weave): link a @weave.op call to the agent spans it produced

ro31337:
Four PRs, in review order:

3. Python + server: **this PR**.

This PR replaces [#7704](https://github.com/wandb/weave/pull/7704) — same branch, same commits, and the earlier review is still there.

[#7729](https://github.com/wandb/weave/pull/7729) merges first. Once it is on `master`, merge `master` into this branch before merging this one.

## JIRA Issue(s)

## Description

Today if you wrap an agent run in a `@weave.op`, the call lands in one table and the spans the agent produced land in another. Nothing links them. So you cannot ask "what did this call actually do" — you get two tabs and no way across.

This adds that link. Two new columns say which call produced the span, and finding them is a normal filter on the agent-spans query we already have. No new endpoint, no UI change.

I wanted the smallest thing that works: write the link down when the span is created, and stop there. The read side — an API for it and the UI affordance — is WB-38427 and comes next.

The main idea behind this work is a span processor, the same trick we already use for evals. A span starts, the processor asks "is an op running right now", and if yes it writes that call's id onto the span. The server copies the ids into the two `spans` columns #7729 adds. The processor is exported as `weave.OpLinkSpanProcessor`, so someone who configured OTel before `weave.init()` — where weave installs nothing of its own — can register it themselves. That is the whole feature.

## Why this approach

There is no per-integration code here, and that is the point of doing it this way. The processor sits on the tracer provider that every integration writes through, so registering it once covers all of them.

First I tried to be clever about storage: stamp only the top span of each subtree, not every span. I wrote it and it does not work. The rule only compares against the direct parent, so it stamps every second level and one call ends up with several "top" spans. Doing it properly needs bookkeeping I did not want on this path, so I stamp everything, same as evals.

Then I found something I did not expect. When a span has too many attributes, OTel throws away the oldest ones. So the order you write them in decides what you lose. I write the trace id first and the call id second, because the call id is the one you filter on and I would rather lose the other. Same reason this processor is registered before the eval one: if something has to go, it should be the new thing, not eval data that has been shipping since May.

One limit worth knowing. This only works when the span starts in the same context as the op. If a library runs the agent on its own thread, there is nothing to read and the column stays empty. Google ADK's `Runner.run()` does that, so wrap `run_async` instead. The realtime integration is the one source that never links: it emits from a timer thread, and reaching it means carrying the call through its own queue, which is not this PR. Empty is a normal state here, not a lost link.

Before this, a call and its spans had no field in common. Now they share two, and getting from one to the other is one filter.

## Backward compatibility

Additive. The two columns come from [#7729](https://github.com/wandb/weave/pull/7729) and default to empty; this PR fills them and puts them in the spans list projection, so it needs #7729 on `master` first. Existing rows read `''`; there is no read-time fallback or backfill.

## Testing

New test file for the processor: inside an op, outside one, nested ops, Conversation SDK, `log_turn`, threads, and a crowded span. The CH query test asserts the rows that come back, not just the count. Also ran the `openai_agents` and `google_adk` shards.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,13 +29,13 @@ jobs:
         run: |
           uv build
       - name: upload test distribution
-        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # release/v1
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # release/v1
         if: ${{ inputs.is_test }}
         with:
           repository-url: https://test.pypi.org/legacy/
           verbose: true
       - name: upload distribution
-        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # release/v1
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # release/v1
         if: ${{ !inputs.is_test }}
 
   # this job exists so that there's some traceability between weave SDk releases
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -344,6 +344,20 @@ deterministic.
   flattened `weave.integration.meta.*` provenance. OTel scalar metadata stays
   typed; non-scalar values are stringified.
 
+### TypeScript Anthropic message batches
+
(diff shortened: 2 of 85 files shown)


rgao-coreweave (approved):
Approved!

LGTM from https://github.com/wandb/weave/pull/7704

PR #7732, opened 2026-08-12
fix(weave): guard None extra_headers in openai should_use_accumulator

### LINKED ISSUE #7731: OpenAI streaming silently not traced when extra_headers=None is passed explicitly
### Description

`weave.integrations.openai.openai_sdk.should_use_accumulator` crashes (and the crash is swallowed by the op wrapper) when a streaming call is made with an explicit `extra_headers=None`. `extra_headers` is typed `Headers | None`, so `None` is a valid value. When it's passed, the accumulator check raises `AttributeError`, the exception is caught silently, and the streaming response is **not** accumulated — the trace stores the raw un-consumed iterator instead of the assembled `content`/`usage`. Silent observability loss with no error surfaced to the user.

weave version: 0.53.5.dev0 (master @ 1213bfa) · python 3.13 · openai 2.30.0

### Reproducer

```python
from weave.integrations.openai.openai_sdk import should_use_accumulator
# extra_headers: Headers | None  ->  None is valid and reaches here
should_use_accumulator({"stream": True, "extra_headers": None})
# AttributeError: 'NoneType' object has no attribute 'get'
```

Equivalent end-user trigger:

```python
client.chat.completions.create(model="gpt-4o", messages=[...], stream=True, extra_headers=None)
```

### Root cause

`weave/integrations/openai/openai_sdk.py:374`

```python
and not inputs.get("extra_headers", {}).get("X-Stainless-Raw-Response") == "true"
```

The `{}` default only applies when the key is absent; an explicit `None` value passes through, so `None.get(...)` raises.

### Fix

`(inputs.get("extra_headers") or {}).get(...)`. Happy to open a PR (branch ready, test included).

Shailendra005:
### What

`should_use_accumulator` did `inputs.get("extra_headers", {}).get("X-Stainless-Raw-Response")`. When a streaming call passes `extra_headers=None` explicitly (valid — SDK type is `Headers | None`), the key is present with value `None`, the `{}` default doesn't apply, and `None.get(...)` raises `AttributeError`.

### Impact

The exception is swallowed by the op wrapper, so streaming is silently **not** accumulated: the trace records the raw un-consumed iterator instead of assembled content/usage. No error is surfaced — pure observability loss.

### Fix

One-line guard: `(inputs.get("extra_headers") or {}).get(...)`. This is the only occurrence of the pattern in the integration; sync + async both route through this function; the Responses API path (`should_use_responses_accumulator`) is unaffected.

### Tests

Added `test_should_use_accumulator_with_none_extra_headers` (regression: fails on master with `AttributeError`, passes with the fix) plus a small variants test covering absent / empty-dict / `stream=False` / raw-response-header cases. Verified locally with `pytest tests/integrations/openai/test_openai_sdk.py -k should_use_accumulator` (2 passed; the regression test fails without the source change) and `ruff check` clean.

Fixes #7731

### DIFF
--- a/tests/integrations/openai/test_openai_sdk.py
+++ b/tests/integrations/openai/test_openai_sdk.py
@@ -10,6 +10,7 @@ from weave.integrations.openai.openai_sdk import (
     create_wrapper_sync,
     openai_on_input_handler,
     serverless_inference_call_display_name,
+    should_use_accumulator,
 )
 from weave.trace.autopatch import OpSettings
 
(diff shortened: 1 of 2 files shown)


Shailendra005:

ro31337 (approved):

PR #7738, opened 2026-08-13
feat(weave_ts): add GenAI span error helpers

rgao-coreweave:
## Summary

- `recordError(error)` records a failure without ending the span.
- `end({error})` records the failure and closes the span.
- `error.type` is derived from `error.name`.
- The Claude Agent SDK integration uses the helper for deferred subagent closure.

Supports wandb/weave-claude-code#167.

## Testing

- Focused GenAI tests: 46 passed
- Full checks run in CI


sdks/node/src/genai/spanBase.ts
```diff
@@ -94,15 +105,23 @@ export abstract class SpanBase {
...
-        message: opts.error.message,
-      });
+      this._recordError(opts.error);
     }
     this.span.end(opts?.endTime);
   }
 
+  private _recordError(error: Error): void {
+    this.span.setAttribute(
+      ATTR_ERROR_TYPE,
```
ro31337, line 115: nit: mentioned in the sister PR as well. This change seems to overwrite an existing `error.type`. For example, wandb/weave-claude-code#167 sets `error.type` to `rate_limit` and then calls `end({error: new Error(details)})`. This method changes the final value to `Error`.

Its CI does not detect the problem because the lockfile uses `weave@0.16.3`. A fresh install can select 0.16.6 because the dependency is `^0.16.3`.

I would recommend to preserve an existing `error.type`, or support an explicit error type.

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
```diff
@@ -745,12 +737,10 @@ export class ClaudeAgentOtelTracer {
       return;
     }
 
-    openSubagent.outcome = {
-      status: 'error',
-      errorType: msg.status === 'stopped' ? 'aborted' : 'subagent_error',
-      error: new Error(msg.summary || `Background subagent ${msg.status}`),
-      endedAt,
-    };
+    openSubagent.span.recordError(
```
ro31337, line 740: nit: am I right that this change removes the task-status discriminator? For example, a failed task and a stopped task now both have `error.type = Error`, and no other attribute stores `msg.status`. The Python implementation in #7744 stores this value in `claude_agent_sdk.task.status`.

I would recommend to record `claude_agent_sdk.task.status` for `completed`, `failed`, and `stopped` tasks. Up to you tho.

ro31337 (commented):
comment

sdks/node/src/genai/spanBase.ts
rgao-coreweave: Simplified in f824d16406: reverted both explicit-override experiments. `recordError()` and `end({error})` derive standard error metadata from the `Error`; the `StopFailure` path in wandb/weave-claude-code#167 no longer sets or preserves a failure-type classification. There is no warning or attribute sanitizer.

ro31337 (approved):

sdks/node/src/integrations/claude-agent-sdk/otelTracer.ts
rgao-coreweave: Good call, thank you!

rgao-coreweave: added `Claude task status`

PR #7739, opened 2026-08-13
feat(weave): record the agent span that invoked a call

ro31337:
Jira: [WB-38616](https://coreweave.atlassian.net/browse/WB-38616)

## Description

We have two models. `calls` holds every `@weave.op` call. `spans` holds the OTel spans an agent run makes. An agent that calls a tool writes to both, because the tool itself is a `@weave.op`. Nothing used to join the two. The call row has no weave parent, so it opens a trace of its own, and the UI shows it as an orphan next to the agent trace.

#7730 did the first half. A span now carries the id of the call it ran under, so from a call you can find its spans. This PR does the second half. We want to start at an agent span and find the call it invoked, and for that the id of the span has to sit on the call.

So that is what this PR writes. When a call starts, `create_call` reads the OTel span that is open at that moment and saves its id on the call:

```json
{"weave": {"invoking_span": {"trace_id": "41512b73...", "span_id": "31b3c094..."}}}
```

The ids are lowercase hex, the same form `spans.trace_id` and `spans.span_id` already use, so a reader compares strings and nothing more. To find the calls a span made, filter `/calls/stream_query` on the pair. No new endpoint, and no work on the read side at all.

## Why this approach

I first looked for the place to add code per integration. There is none to add. Every framework opens a span for the tool and makes it current before it runs the tool body. So `create_call` only has to read the span that is already there. `openai_agents`, `google_adk` and the Conversation SDK's `weave.Tool` all work that way, and all three link for free. `log_turn` and `openai_realtime` build their spans after the fact, so they link nothing, which is what they did before.

Then I tried to save writes. The plan was to store the id on the outermost call only and skip the calls under it. It does not work. The check reads what the parent recorded, and the parent may have skipped as well, so you mark every second level. So every call started inside the span gets the id. It costs about 90 bytes per call, and a test fails if someone brings the old plan back.

## Why it is safe

The one case that had to be cut out is our own server. `weave.trace_server` opens OTel spans, and with an in-process server one of them is open while user code runs. Without the fix, every call of every hosted evaluation would name Weave's own telemetry as its caller. The server now marks its span on the OTel context, and `create_call` skips that one span. It compares the span itself, not a "we are inside the server" flag, because an agent span opened inside a server frame is a real target. It reads the context, not the span object: under the OTel to Datadog bridge the current span is a `ddtrace` span, so anything read off the object would pass straight through.

Nothing changes shape. Call attributes are free-form JSON all the way into `attributes_dump`, so there is no schema change, no migration and no new bindings. A caller cannot supply the key, because it is dropped before it is written. An ended span, an unsampled span and an all-zero id all record nothing. The check costs about 0.5 µs when a span is open. The OTel import sits under `try/except`, since `import weave` works today with `opentelemetry` absent. One thing a reader has to accept: the pair can name a span we never stored, when someone runs their own OTel and exports it elsewhere. A missing key and a dead pair are both normal.

Before this, a call and an agent span shared no field. Now the call carries the span it came from, and one filter takes you across. The read API, the UI and TypeScript parity are separate work.

## Testing

New file `tests/trace/test_op_span_linking.py`: the id is there inside a live span, and absent with no span, an unsampled span, an all-zero id, an ended span, or a Weave server span. The `openai_agents` and `google_adk` tests round-trip through ClickHouse and check the pair names the `execute_tool` span, not the agent span above it.

[WB-38616]: https://coreweave.atlassian.net/browse/WB-38616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -124,6 +124,18 @@ Evaluation result rows merge agent span links from two sources: legacy
 trial. Keep the promoted-column hydration best-effort so eval results remain
 available during rolling deploys.
 
+`weave.invoking_span` (`{trace_id, span_id}`, hex) points the other way, from a
+call to the OTel span that was current when it started — an agent's
+`execute_tool` span in the case it is built for, but any ambient instrumentation
+qualifies. `create_call` writes it on *every* call started inside that span, not
+just the outermost one: skipping when the parent call already carries it marks
+every other level, because the check reads the parent's recorded state and the
+parent may itself have skipped. A reader must treat a missing key and a pair
+that matches no span as normal — the write does not check that the span reaches
+our backend. Spans opened by `trace_server.tracing` are the one exclusion, and
+they are recognised through `WEAVE_SERVER_SPAN_KEY` on the context rather than
+off the span object, which under the OTel→DD bridge is not an SDK span.
+
 If `sdks/node/node_modules` is missing, run `pnpm install --frozen-lockfile` in `sdks/node` first. Do not use `npm install`; this SDK is pinned to pnpm.
 
 ## Python Testing Guidelines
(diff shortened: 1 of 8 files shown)


rgao-coreweave (approved):
Good work!

PR #7740, opened 2026-08-13
security(weave): redact PII in Call and Agent writes

rgao-coreweave:
## What

Apply the parent stack's `pii-v1` policy to supported Weave Call models and parsed Agent OTel spans.

The Call helpers become active when the Core integration PR pins this commit. Agent OTel ingest consumes the internal request policy in this PR. An omitted policy remains `off` through the parent request contract.

## Why

Supported PII must be replaced before covered trace content reaches persistence. The backend policy covers direct Call writes and Agent OTel writes without trusting client-side redaction.

## How

- Add field-selective adapters for Call start, end, complete, and display-name update models, typed with constrained TypeVars so v1 and v2 requests keep their exact type. Structural IDs and project names remain unchanged. The batch endpoint keeps its Core-side adapter over Core's local batch types.
- Redact Agent span and event names, span, event, and link attributes, and the status message per span, and each shared resource once per resource-spans group. A PII-bearing name changes `span_name`, `operation_name`, and `agent_name` grouping; keeping PII out of storage wins that trade-off.
- Parse and redact every span in a request before ingest sampling, credential replacement, blob stripping, derived-column extraction, or insertion. The sampler sees redacted values even for spans it later drops, and a failing span cannot leave Content files behind for its siblings.
- Reject a value nested too deeply to scan as `RequestTooLarge` (HTTP 413). Any other detector failure also fails the complete request before any write.
- Leave standard OTel, completions, evaluation and scoring workers, objects, files, existing stored data, and operational telemetry unchanged.

## Testing

- 45 focused tests passed across the sensitive-data adapters, Agent policy contract, ingest sampler, and ClickHouse-backed Agent ingest path, including regression coverage for redaction before sampling at a zero rate, request rejection with no stored rows or extraction side effects, and name redaction end to end.
- The full repo lint session passes (`nox -s lint`: ruff, pyright, mypy, ty, import-linter).

## Anything Else

- Parents: https://github.com/wandb/weave/pull/7788 and https://github.com/wandb/weave/pull/7789 (merged). Core activation: [wandb/core#50579](https://github.com/wandb/core/pull/50579).
- Request logging and Datadog, OpenTelemetry, or Sentry operational telemetry are not scrubbed by this PR.
- No runtime dependency, migration, permission change, or public schema change.

### DIFF
--- a/tests/trace_server/test_sensitive_data_adapters.py
+++ b/tests/trace_server/test_sensitive_data_adapters.py
@@ -2,7 +2,10 @@ from __future__ import annotations
 
 import datetime
 
+import pytest
+
 from weave.trace_server import trace_server_interface as tsi
+from weave.trace_server.errors import RequestTooLarge
 from weave.trace_server.opentelemetry.python_spans import (
     Event,
     Link,
@@ -14,14 +17,16 @@ from weave.trace_server.opentelemetry.python_spans import (
     Resource as SpanResource,
 )
 from weave.trace_server.sensitive_data.call_redaction import (
-    redact_call_batch,
     redact_call_end,
     redact_call_start,
     redact_call_update,
     redact_calls_complete,
 )
 from weave.trace_server.sensitive_data.policy import SensitiveDataPolicy
-from weave.trace_server.sensitive_data.span_redaction import redact_pii_from_span
+from weave.trace_server.sensitive_data.span_redaction import (
+    redact_pii_from_resource,
+    redact_pii_from_span,
+)
 
 NOW = datetime.datetime(2026, 8, 12, tzinfo=datetime.timezone.utc)
 
@@ -50,7 +55,7 @@ def test_call_start_preserves_ref_op_names() -> None:
     assert redacted.start.op_name == "weave:///entity/project/op/my-op:v1"
 
 
-def test_call_end_update_complete_and_batch_redact_content() -> None:
+def test_call_end_update_and_complete_redact_content() -> None:
     end_req = _call_end_req()
     update_req = tsi.CallUpdateReq(
         project_id="ada@example.com/project",
@@ -58,17 +63,10 @@ def test_call_end_update_complete_and_batch_redact_content() -> None:
         display_name="Email ada@example.com",
     )
     complete_req = tsi.CallsUpsertCompleteReq(batch=[_completed_call()])
-    batch_req = tsi.CallCreateBatchReq(
-        batch=[
-            tsi.CallBatchStartMode(req=_call_start_req()),
-            tsi.CallBatchEndMode(req=end_req),
-        ]
-    )
 
     redacted_end = redact_call_end(end_req, SensitiveDataPolicy.PII_V1)
     redacted_update = redact_call_update(update_req, SensitiveDataPolicy.PII_V1)
     redacted_complete = redact_calls_complete(complete_req, SensitiveDataPolicy.PII_V1)
-    redacted_batch = redact_call_batch(batch_req, SensitiveDataPolicy.PII_V1)
 
     assert redacted_end.end.output == {"card": "<CREDIT_CARD>"}
     assert redacted_end.end.summary == {"contact": "<EMAIL_ADDRESS>"}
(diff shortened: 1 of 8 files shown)


weave/trace_server/agents/clickhouse.py
```diff
@@ -861,6 +862,10 @@ def parse_span(protobuf_span: Any, resource: Resource) -> Span | None:
                 rejected += 1
                 errors.append(str(e))
                 return None
+            # Outside the parse-error handler: a detector failure rejects the
+            # whole request before sampling or any storage preparation.
+            redact_pii_from_span(span, req.sensitive_data_policy)
```
ro31337, line 878: Probably worth redacting all spans before extracting any of them, for example if the second span fails, the first span may have already written files. Doing all redaction first would help.

Imagine span 1 has inline blob. Span 2 has JSON string with a depth of 2000. Result is gonna be recursion error. But before it happens we'll have file(s) and object(s) created, and zero CH inserts.

weave/trace_server/sensitive_data/span_redaction.py
```diff
@@ -0,0 +1,35 @@
...
+    ``strip_inline_blobs_from_span`` so values are scanned while still inline,
+    and before extraction so derived columns read the redacted span. Span and
+    event names, IDs, trace state, and timestamps are structural and stay
+    unchanged.
+    """
+    if not pii_enabled(policy):
+        return
+    span.attributes = redact_pii_value(span.attributes)
+    if span.resource is not None:
+        span.resource.attributes = redact_pii_value(span.resource.attributes)
```
ro31337: IMO, worth redacting the resource once before the span loop. Say we have 5000 spans. What happens? Currently we scan the same resource 5000 times. Moving this outside the loop would help, you know.

In other words, `resource.attributes` seems to be the same object and redacting it outside the loop would save us a lot of CPU cycles.

weave/trace_server/sensitive_data/span_redaction.py
```diff
@@ -0,0 +1,35 @@
...
+def redact_pii_from_span(span: Span, policy: SensitiveDataPolicy) -> None:
+    """Redact supported PII in a span's customer-authored string values.
+
+    Mirrors ``redact_credentials_from_span``: covers all four attribute
+    containers a span carries (its own, its resource's, its events' and its
+    links') plus the status message, because ``raw_span_dump`` and every
+    attribute-derived column read from them. Must run before
+    ``strip_inline_blobs_from_span`` so values are scanned while still inline,
+    and before extraction so derived columns read the redacted span. Span and
+    event names, IDs, trace state, and timestamps are structural and stay
```
ro31337: nit: think about redacting span and event names too, for example an email in a span name is stored as-is. What should we do in this case? It's more of a product question tho

ro31337 (approved):
Approving, with a caveat that we need a couple of fixes, IMO

tests/trace_server/test_sensitive_data_adapters.py
```diff
@@ -0,0 +1,192 @@
...
+        "gen_ai.prompt": "Email <EMAIL_ADDRESS>",
+        "payload": {"image": "data:image/png;base64,QUJD"},
+    }
+    assert span.resource is not None
+    assert span.resource.attributes == {"service.owner": "<EMAIL_ADDRESS>"}
+    assert span.events[0].attributes == {"note": "call <PHONE_NUMBER>"}
+    assert span.links[0].attributes == {"context": "ssn <US_SSN>"}
+    assert span.status.message == "card <CREDIT_CARD>"
+    assert span.name == "agents_pii_surface"
+    assert span.span_id == "b2" * 8
```
andrewtruong: Why not just assert the actual span shape?

rgao-coreweave: it was constructed as `span_id="b2" * 8`, switched to whole-span and changed everything to be string literals.

weave/trace_server/agents/clickhouse.py
rgao-coreweave: GOOD CALL! Made this change so extraction never starts, and nothing is stored if there is error.

weave/trace_server/sensitive_data/span_redaction.py
rgao-coreweave: redact_pii_from_resource now runs once per ResourceSpans before the span loop.

Super happy that you caught this optimization! thank you!

weave/trace_server/sensitive_data/span_redaction.py
rgao-coreweave: Good call, I think redacting span and event names makes sense.

PR #7751, opened 2026-08-14
chore(weave): remove signature normalization claim

gtarpenning:
## Summary
- Remove migration comments claiming insight signatures are rewritten before insertion.
- Core PR [wandb/core#50387](https://github.com/wandb/core/pull/50387) stores the judge signature unchanged and uses the same text for embedding and ClickHouse.

## Testing
non-functional SQL comment change; executable migration text is unchanged after comment stripping.


gtarpenning:
whatever

PR #7752, opened 2026-08-14
chore(weave): drop clickhouse-connect 0.x workarounds, pin driver to 1.6

nikumar1206:
## Summary

Bump clickhouse-connect pin 0.11.0 → 1.6.0 and deletes the workarounds the 1.x CH driver makes dead code:

- `should_retry_empty_query` / `insert_with_empty_query_retry` + `INSERT_MAX_RETRIES` — the 1.0 driver rebuilds the insert body on connection retry, fixing the "Empty query" replay bug this loop existed for. Callers go back to plain `ch_client.insert`.
- rely on the now structured int `exc.code` instead of the message-regex we used to rely on.

## Testing

- Unit tests
- QA testing


pyproject.toml
```diff
@@ -462,7 +462,7 @@ test = [
   "pytest-rerunfailures>=14.0",
   "pytest-timeout>=2.3.1",
   "pytest-shard==0.1.2",
-  "clickhouse_connect==0.11.0",
+  "clickhouse_connect==1.6.0",
```
gtarpenning: maybe `>=1.6.0,<1.7`?

also do we need to pin this anywhere else? like `[dependency-groups] trace_server` (above)

gtarpenning: and ` [project.optional-dependencies] trace_server`?

gtarpenning (approved):

pyproject.toml
nikumar1206: yep 100% i didnt even see we had optional dep above. 
but moving fastapi, ch-connect, sql-parse up there breaks some of the integration tests.

so im electing to fix in a separate PR. 
added a todo comment for now.

nikumar1206: also these dependency requirements for the [trace-server] group are not 1-1 with core/services/weave-trace/pyproject.toml

which means CI can (probably is) is using different versions of some dependencies compared to what runs deployed.

PR #7755, opened 2026-08-15
fix(weave): stop duplicating custom summaries in log_summary

### LINKED ISSUE #7741: [Bug]: `EvaluationLogger.log_summary(summary=..., auto_summarize=False)` duplicates the custom summary in the logged output
### Describe the bug

**weave version:** 0.52.25

### Description

When calling `EvaluationLogger.log_summary(summary, auto_summarize=False)`, the custom
`summary` dict ends up duplicated inside the value that gets logged to Weave: once
flattened at the top level, and once again nested under an `"output"` key. This
duplicated dict is then used as the output for *two* separate things in the UI — the
parent evaluation's summary, and the child `Evaluation.summarize` call — so the same
custom keys visibly appear twice in each of those views.

### Root cause

In `weave/evaluation/eval_imperative.py`, `log_summary` does:

```python
if auto_summarize:
    data_to_summarize = [pred._captured_scores for pred in self._accumulated_predictions]
    summary_data = auto_summarize_fn(data_to_summarize)
else:
    summary_data = summary          # <-- summary_data is now literally `summary`

final_summary = {}
if summary_data:
    final_summary = summary_data    # <-- final_summary == summary (same dict)
if summary is not None:
    final_summary = {**final_summary, "output": summary}   # <-- summary merged in AGAIN, nested this time
```

When `auto_summarize=True`, `summary_data` comes from the scorer-derived auto-summary,
which is a distinct object from the user-supplied `summary`, so nesting `summary` under
`"output"` alongside it is fine — no overlap.

When `auto_summarize=False`, `summary_data` is set to `summary` itself, so
`final_summary` starts out equal to the user's dict, and the last `if` re-adds that same
dict again under `"output"`. The result: `final_summary` contains every custom key twice
— once at the top level, once nested at `final_summary["output"]`.

That single duplicated `final_summary` dict is then used twice more downstream:
- `self._pseudo_evaluation.summarize()` is a weave op whose body returns
  `current_summary.get()` (i.e. `final_summary`) — this call shows up as a child
  `Evaluation.summarize` call in the trace tree with `final_summary` as its output.
- `self._finalize_evaluation(output=final_summary)` sets the *parent* evaluation call's
  output/summary to the same `final_summary`.

So in the UI, the evaluation's summary panel and the nested `Evaluation.summarize` call
both render `final_summary`, and each of those already has the custom summary appearing
twice within itself.

### Repro

```python
import weave

weave.init("repro-log-summary-dup")

ev = weave.EvaluationLogger(name="dup-repro")
pred = ev.log_prediction(inputs={"q": "2+2"}, output="4")
pred.log_score("correctness", 1.0)
pred.finish()

ev.log_summary(
    summary={"correctness": {"mean": 1.0}},
    auto_summarize=False,
)
```

Inspect what gets logged, e.g. by printing right before `wc.finish_call` runs, or just
open the run in the Weave UI: the evaluation summary (and the nested
`Evaluation.summarize` call) show:

```python
{
    "correctness": {"mean": 1.0},       # flattened copy
    "output": {
        "correctness": {"mean": 1.0},   # same dict, nested again
    },
}
```

### Expected

With `auto_summarize=False`, the logged summary should just be the caller-supplied
`summary` dict, without being duplicated under an extra `"output"` key.

### Weave Project Link

_No response_

### Screenshots

_No response_

feiiiiii5:
## Problem

With `auto_summarize=False`, `EvaluationLogger.log_summary(summary=...)` set `summary_data` to the caller's dict and then nested that same dict again under an `"output"` key. Every custom key appeared twice in the logged summary and in the child `Evaluation.summarize` call output.

Fixes #7741.

## Fix

The caller-supplied summary is only nested under `"output"` when an auto-computed summary was merged alongside it (`auto_summarize=True`), which is the distinct-object case the nesting exists for. With `auto_summarize=False` the logged summary is exactly the caller's dict.

## Test

Updated `test_evaluation_no_auto_summarize` (expects `{}` instead of `{"output": {}}`) and `test_evaluation_no_auto_summarize_with_custom_dict` (expects the custom dict without the duplicated nested copy) in `tests/flow/test_evaluation_imperative.py`.

### DIFF
--- a/tests/flow/test_evaluation_imperative.py
+++ b/tests/flow/test_evaluation_imperative.py
@@ -537,7 +537,7 @@ def test_evaluation_no_auto_summarize(client):
     calls = client.get_calls()
     # assert len(calls) == 1
     summarize_call = calls[4]
-    assert summarize_call.output == {"output": {}}
+    assert summarize_call.output == {}
 
 
 def test_evaluation_fail_with_exception(client):
@@ -566,11 +566,7 @@ def test_evaluation_no_auto_summarize_with_custom_dict(client):
     calls = client.get_calls()
     # assert len(calls) == 1
     summarize_call = calls[4]
-    assert summarize_call.output == {
-        "something": 1,
-        "else": 2,
-        "output": {"something": 1, "else": 2},
-    }
+    assert summarize_call.output == {"something": 1, "else": 2}
 
 
 def test_evaluation_logger_model_inference_method_handling(weave_active):
--- a/weave/evaluation/eval_imperative.py
+++ b/weave/evaluation/eval_imperative.py
@@ -1083,7 +1083,7 @@ class EvaluationLogger:
         final_summary = {}
         if summary_data:
             final_summary = summary_data
-        if summary is not None:
+        if summary is not None and auto_summarize:
             final_summary = {**final_summary, "output": summary}
 
         # Call the summarize op


feiiiiii5:

feiiiiii5:
Closing as part of a contribution-portfolio consolidation: my effort is now focused on a single project (failroute, a static analyzer for silent-failure anti-patterns) and its upstream follow-through. Happy to revisit if there is interest.

PR #7757, opened 2026-08-17
feat(weave): log a query_id on every ClickHouse query, including failures

nikumar1206:
## What

Stacked on #7752. Logs a `query_id` for every ClickHouse query, command, and insert — including the ones that fail.

We already log the id on success, because it comes back in the driver's result summary. A failed query has no result, so those log lines had nothing to correlate with. This mints the id before firing and passes it through settings, so both the success and the error line carry it. A caller-supplied `query_id` still wins — `export.py` sets its own.

## Why

When a query fails or times out, the id is what lets you find it in `system.query_log` and see what ClickHouse thought was happening.

## Testing

191 tests passing. New cases cover the id appearing in both the success and failure log lines, matching what was actually sent to ClickHouse, and a caller's own id not being overwritten.


weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -7758,6 +7767,8 @@ def _insert(
         sanitized_invalid_utf8 = False
         # At most two attempts: the original, plus one retry after sanitizing
         # invalid client UTF-8.
+        settings = dict(settings or {})
+        query_id = settings.setdefault("query_id", generate_id())
```
gtarpenning, line 7776: we intentionally re-use the same id?

gtarpenning: and do we want to add this to the DD span?

gtarpenning: (instead of just in the log)

gtarpenning (approved):

weave/trace_server/clickhouse_trace_server_batched.py
nikumar1206: only retry is a client-side encode failure, so ClickHouse never saw the first one technically.
should be safe to re-use the query_id.

PR #7763, opened 2026-08-18
chore(weave): refresh generated model data

mattliu-mygit:
## Description

Refresh the generated model-provider routing and token-cost checkpoints from the current upstream sources. This is the repository's periodic generated-data sync and does not add manual routing or pricing logic.

The refresh adds 75 routing keys and 34 cost keys, including Gemini 3.7 Flash routing for Gemini and Vertex AI and its introductory $0.75/M input and $3.75/M output pricing. All 34 added prices match the current LiteLLM source data exactly.

The generated routing changes are backend lookup data, not an advertised model catalog. This PR does not change the Core Playground registry, the W&B-hosted model catalog, or SDK model discovery, so newly recognized Inkling, DeepSeek, GPT, audio, or other upstream IDs do not automatically appear for customers. An API/SDK caller that already submits an exact added ID may now reach its provider route when the required provider secret is configured.

The existing Gemini 3.6 Flash checkpoint remains at the price currently published by LiteLLM. Google's temporary promotional repricing for 3.6 is not overridden here because LiteLLM is the authoritative source in the existing generator. This can affect Weave's displayed trace-cost estimate for Gemini 3.6, but not the external provider's billing.

## Testing

- Re-ran `make update_model_providers` and `make update_costs`; output hashes were unchanged and the cost generator reported zero additional changes.
- 26 focused cost update/insertion tests passed.
- Parsed both generated JSON files and asserted the exact Gemini 3.5 Flash-Lite and 3.7 Flash routing and prices.
- Compared the full generated outputs with the current upstream registry: no current recognized provider route or priced model is missing, and no current price differs.

## Rollout

Related Core Playground PR: https://github.com/wandb/core/pull/51097.

The PRs are not stacked and can be reviewed independently. This generated routing data must be deployed before the Core UI refresh so newly selectable Gemini IDs resolve to the intended provider and secret.

### DIFF
--- a/weave/trace_server/costs/cost_checkpoint.json
+++ b/weave/trace_server/costs/cost_checkpoint.json
@@ -29472,5 +29472,345 @@
       "cache_creation_input": 0.0,
       "created_at": "2026-08-12 22:36:42"
     }
+  ],
+  "azure_ai/FW-DeepSeek-V3.2": [
+    {
+      "provider": "azure_ai",
+      "input": 6.2e-07,
+      "output": 1.85e-06,
+      "cache_read_input": 3.1e-07,
+      "cache_creation_input": 0.0,
+      "created_at": "2026-08-18 17:54:20"
+    }
+  ],
+  "azure_ai/FW-DeepSeek-V4-Pro": [
+    {
+      "provider": "azure_ai",
+      "input": 1.925e-06,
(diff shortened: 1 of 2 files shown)


jwlee64 (approved):
lgtm looks like we hadnt refreshed model_providers in a bit

PR #7764, opened 2026-08-18
fix(weave): keep an evaluation's derived name off its dataset's object_id

ro31337:
## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-38822

## Description

If you don't name an `Evaluation`, we name it: the dataset's name plus `-evaluation`. Names are cut at 128 characters, so if the dataset's name already fills that, the suffix goes with it and both objects share one `object_id`. The server rejects the second one (WB-30574, #6804) with an id you never typed. Through `EvaluationLogger` it doesn't even raise, it just never saves.

207 of these in the last seven days ([datadog](https://us5.datadoghq.com/apm/traces?query=env%3Aprod%20service%3Aweave-trace%20resource_name%3A%22POST%20%2Fobj%2Fcreate%22%20%40http.status_code%3A400)), up from one or two a day before 7 August. It is not spread out, though: of 200 I sampled, 179 are this bug and every one of them is a single internal project retrying. No customer hit it in that window. Any dataset whose name reaches the limit will do the same, which is why it is worth fixing now rather than when it lands on someone else.

Leaving room for the suffix doesn't work: a dataset named `<117 chars>-evaluation` collides anyway. So compare the two ids, rewrite the name only if they match, and cut to 127 characters with a hash in the middle. A colliding dataset's id is always exactly 128, so a shorter one can't be it.

Long dataset names stop breaking unnamed evaluations, and only already-broken ids move. The general two-name collision stays, but now it is visible: the client warns on a cut name, and the server's error names the object.

## Testing

Ran the trace, eval and `trace_server` suites locally, plus ruff, mypy, pyright, ty and import-linter. Green, apart from two flakes that pass when run on their own and six `tests/flow/` failures that are the same on master.

### DIFF
--- a/tests/trace/conftest.py
+++ b/tests/trace/conftest.py
@@ -13,6 +13,7 @@ from google.api_core import exceptions as gcp_exceptions
 from google.auth.credentials import AnonymousCredentials
 
 from tests.trace.test_utils import FailingSaveType, failing_load, failing_save
+from weave.trace import util as trace_util
 from weave.trace.serialization import serializer
 from weave.trace_server.file_storage import reset_stored_key_cache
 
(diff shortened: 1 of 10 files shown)


gtarpenning (approved):

rgao-coreweave (approved):
Looks good, thanks for fixing.

PR #7770, opened 2026-08-19
chore(weave): add a skill for converting classic calls to agent spans

gtarpenning:
## Summary
- A project that only logged `@weave.op` calls is empty in the Agents tab. `skills/calls-to-agent-spans` converts a time window of those calls into agent spans over `/calls/stream_query` and `/agents/otel/v1/traces`.
- The skill entrypoint is the dry-run procedure. `references/conventions.md` is the full Agents-tab span contract (every `weave.*` key, including what must stay empty) and links [`semconv.py`](https://github.com/wandb/weave/blob/master/weave/trace_server/agents/semconv.py) as the registry.
- The converter refuses the source project and a target that already has agent spans, reports per-op coverage on dry-run, reads the last `role=user` message, and takes assistant text from the last model child when the root is a wrapper.
- `skills/weave-instrument` teaches the same contracts for new instrumentation: conversation id on the turn and model calls, user text on `start_turn` / `startTurn`, usage on the LLM span. `references/span-shape.md` is the short form; Session SDK and raw OTEL examples follow it.

## Testing
`uv run --group test python -m pytest skills/calls-to-agent-spans/tests/test_convert.py -v` covers wrapper-plus-child, model-only roots, history lists, missing session keys, and same-project writes. weave-instrument evals 1, 3, and 6 expect user text on the turn and usage on the model call. `evals/evals.json` is not wired into CI.


skills/calls-to-agent-spans/scripts/calls_client.py
```diff
@@ -0,0 +1,115 @@
...
+    after: str,
+    before: str,
+    columns: list[str] | None,
+    roots_only: bool = False,
+    op_names: list[str] | None = None,
+    max_rows: int | None = None,
+) -> list[Row]:
+    """Every call in the window, roots and children alike, so the trace tree survives."""
+    terms: list[dict[str, object]] = [
+        {"$gt": [{"$getField": "started_at"}, {"$literal": after}]}
```
ro31337, line 25: is it possible the selected window starts after the root call but still contains its child calls? curious if we silentry drop them here in this case

ro31337 (approved):
Approving, but see the comment - other stuff LGTM

skills/calls-to-agent-spans/scripts/calls_client.py
gtarpenning: mmm interesting. we can probably construct a failure case there.

gtarpenning: but i suspect very very rare

PR #7185, opened 2026-06-11
chore(weave): remove experimental StainlessRemoteHTTPTraceServer

andrewtruong:
## Description

Removes the experimental `StainlessRemoteHTTPTraceServer` and all of its opt-in plumbing. The binding was already dormant on master: the `weave_server_sdk` dependency group it needs is commented out (it pinned a dead git generation of the SDK), and its CI matrix entries are disabled — a clean install could never actually use it.

The next PR in this stack (#7177) rebuilds `RemoteHTTPTraceServer` itself on the published weave-server-sdk, so the parallel class goes first:

- `weave/trace_server_bindings/stainless_remote_http_trace_server.py`, its test file, the dormant `stainless` nox shard, and the commented-out CI matrix entries
- the `use_stainless_server` setting, its `weave.init` docstring entry, and the stainless branch in `init_weave_get_server`
- the `--remote-http-trace-server` pytest flag, the `server_class` fixture, the collection filter implementing the `_remote`/`_stainless` test-file split, and the now-meaningless `remote_only`/`stainless_only` markers
- `test_http_behavior_remote.py` → `test_http_behavior.py` (pure rename; the suffix only existed to pair with `_stainless`)
- the commented-out stainless dependency group in `pyproject.toml` and the AGENTS.md section documenting the flag
- a stale docstring mention of the class in `weave_client.py`

No dependency or lockfile changes: the removed group was commented out, so resolution is untouched.

## Testing

- `tests/trace_server_bindings` — 55 passed (against the surviving `RemoteHTTPTraceServer`, no SDK installed)
- `import weave` smoke test with `weave-server-sdk` absent from the env (proves no hidden dependency)
- Full pre-push hook suite (ty, pyright, mypy, ruff check/format, fixit, typos) — all pass
- Grep audit: zero remaining references to stainless / `--remote-http-trace-server` / `use_stainless_server` / the dead markers (excluding unrelated OpenAI-SDK "Stainless" header mentions)

PR #7177, opened 2026-06-11
feat(weave): rebuild RemoteHTTPTraceServer on weave-server-sdk

andrewtruong:
## Description

Swaps `RemoteHTTPTraceServer`'s hand-rolled HTTP mechanics for the generated weave-server-sdk client (0.0.1, temporarily resolved from test PyPI via a uv index pin), keeping the class and its full tsi-typed method surface intact — a drop-in replacement. (The dormant experimental `StainlessRemoteHTTPTraceServer` was removed in the parent PR, #7185.)

Transport design:

- One injected `httpx.Client` (env-proxy default transport, `ssl_verify`/`http_timeout` honored) shared by SDK-routed and raw paths; a response event hook routes errors through `handle_response_error` so callers keep seeing `httpx.HTTPStatusError` / `CallsCompleteModeRequired` (retry predicates, 413 batch splitting, and the calls_complete auto-upgrade key off these); a request hook injects `X-Weave-Retry-Id` per attempt.
- Typed SDK resource methods wherever the generated client is correct; raw requests, each with a documented reason, for endpoints excluded from the OpenAPI spec (calls_complete v2, eager v2 call start/end, completions, project stats, TTL, feedback aggregate) and for weave-server-sdk 0.0.1 codegen bugs (duplicate generated method names where the last definition wins, lost multipart/otel bodies). Streaming endpoints keep line-by-line streaming (the published SDK buffers jsonl).
- `files_stats` now hits `/files/query_stats` (the real route); the old implementation targeted `/files/stats`, which does not exist.
- The parallel-table-upload endpoint probe uses a new explicit `unretried_table_create_from_digests` method instead of poking `_post_request_executor.__wrapped__` internals.
- The `network_proxy_client` fixture routes the binding's transport into the FastAPI TestClient via the binding's transport test seam instead of monkeypatching `weave.utils.http_requests.post`.

Also starts the incremental client/server import lint (ruff TID251 banned-api): `weave.trace_server.http_service_interface` is now unused by client code and banned; the list grows as the client moves to weave_server_sdk models in the next PRs.

## Testing

- `tests/trace_server_bindings` — 58 passed, including transport-boundary tests for retry behavior, 413 splitting, calls_complete/eager batching, and the auto-upgrade path
- Full pre-push hook suite (ty, pyright, mypy, ruff check/format) — all pass standalone on this PR (an earlier revision had a latent standalone mypy failure in `weave_init.py` that was masked at the stack tip; fixed here)
[...]

PR #7771, opened 2026-08-19
chore(weave): vendor the generated trace server SDK

ro31337:
## JIRA Issue(s)

[WB-38837](https://coreweave.atlassian.net/browse/WB-38837)

## Description

This is the next step towards moving Weave onto the Stainless-generated trace server SDK. To get there, we need that SDK in the repo. This PR puts it there, which unblocks the Python side.

Here is the problem. `StainlessRemoteHTTPTraceServer` has never worked. It has been in the repo since November. `weave/trace_server_bindings/stainless_remote_http_trace_server.py:12` imports `weave_server_sdk`, and that package is not installed anywhere.

Why it is not installed: the SDK is generated in core and it is not on PyPI. It used to be pinned from git. PyPI publishing does not allow that, so #5798 pulled the pin and turned the `stainless` CI shard off with it. Nothing has moved since, because you cannot test code that will not import.

Publishing the SDK for real would fix this properly. Someone tried — #7177, on top of #7185 — but that needs a release agreed first, and the stack went stale in draft. So this follows the same idea minus the publishing — and minus that stack's other move, deleting `StainlessRemoteHTTPTraceServer`, which this PR leaves alone.

So we vendor instead. A copy inside `weave/` is not a dependency, so the PyPI rule stops applying. It goes in `weave/vendor/weave_server_sdk/`, because only `weave/` ends up in the wheel. `scripts/vendor_weave_server_sdk.py` copies it out of a codegen build in core.

The copy needs one thing changed. The package refers to itself by name, and that name moved. I thought it was one line: where it imports its own `resources` module. It is two. `__init__.py` also rewrites `__module__` on everything it exports, and some of those are re-exported httpx classes. So leaving it alone broke pickling of `httpx.Timeout` for the whole process, not just for us. The script does both now, and shouts if either stops matching.

Two new dependencies, `distro` and `sniffio`. Both come from `stlc`, not us. The rest already arrive with `pydantic` and `gql[httpx]`.

One more thing. The `typos` hook runs with `--write-changes`. On its way past it renamed `plain_serializer_function_ser_schema` to `..._set_schema` inside the generated code. So it is excluded now, next to the type checkers and coverage. That trade is real — from here on nobody type-checks or covers that tree.

Bottom line. The package is in the repo and imports. Nothing is turned on: `use_stainless_server` is still `False`, the shard is still off, the old client is still the only one in use. Moving the binding onto the copy is the next PR, and it is real work — 36 of its 78 client calls do not resolve against what the generator produces today.

## File map

This PR changes 221 files. 211 of them are generated — the copy under `weave/vendor/weave_server_sdk/`. The other 10 are ours.

```
weave/vendor/
├── weave_server_sdk/                 [GENERATED] 211 files, verbatim from stlc
│   ├── __init__.py                   [GENERATED] 1 line rewritten
│   ├── _utils/_resources_proxy.py    [GENERATED] 1 line rewritten
│   ├── resources/                    [GENERATED] 23 files
│   ├── types/                        [GENERATED] 157 files
│   └── _base_client.py, _client.py   [GENERATED] and the rest
├── weave_server_sdk.origin.json      written by the script
├── README.md                         do not hand-edit the copy
└── __init__.py                       empty

scripts/vendor_weave_server_sdk.py    the copy and the 2 rewrites
tests/trace_server_bindings/test_vendored_weave_server_sdk.py    2 tests

touched:
pyproject.toml           2 deps, 6 tool exclusions
uv.lock                  re-locked
.pre-commit-config.yaml  mypy exclusion, kept in sync
.gitattributes           collapses the generated diff
AGENTS.md                do-not-hand-edit note
```

## Testing

Two new tests in `tests/trace_server_bindings/test_vendored_weave_server_sdk.py`, one per rewrite. They run in the `trace_server_bindings` shard, which is already on:

- the lazy proxy imports the vendored `resources` module
- a rewritten `__module__` still names something importable, checked on `httpx.Timeout`

Checked by hand:

- bindings shard passes, 77 tests
- `prek run --hook-stage=pre-push --all-files` is clean and leaves the generated code alone
- built a wheel, installed it into an empty venv, imported the client — works
- removed `distro` from that venv, then `sniffio` — each one breaks the import, so both are needed
- wheel gains 214 files against master and loses none, +0.21 MB

The dependency part had to be by hand. Every test env installs `weave[trace_server]`, that brings in LiteLLM and OpenAI, and those supply `distro` and `sniffio` regardless — so CI cannot tell you whether the two lines are needed.

[WB-38837]: https://coreweave.atlassian.net/browse/WB-38837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rgao-coreweave (approved):
👍👍👍

PR #7773, opened 2026-08-19
fix(weave): unbox traced values before the google.genai call

ro31337:
## JIRA Issue(s)

[WB-33439](https://coreweave.atlassian.net/browse/WB-33439)

## Description

If your eval model calls Gemini, the prompt never arrives. Google returns `400 Request has empty input` on every row. The eval does not crash. It catches the error, writes `None`, and finishes, so the run looks fine and every score is empty.

The ticket has been open since April, with a repro and the captured body in it: `{"contents": [{}]}`. I reran it, and it still breaks. You do not need an API key to see why. Build `_GenerateContentParameters` by hand, pass a `str` subclass as `contents`, dump it, and the contents come back `{}`. The prompt is gone before anything leaves the process.

The value is a string, just not a plain one. Weave wraps values read out of a saved object to attach a ref, so a string comes back as `BoxedStr`. `google.genai`'s `Content` model sets `from_attributes=True` with both fields optional, so pydantic takes any object, finds nothing on it, and returns an empty `Content`. A plain `str` takes its own branch. A subclass does not.

## Why this approach

The ticket suggests unboxing at the eval boundary, in `weave/flow/model.py`. That fixes every integration at once, but the recorded inputs lose their refs to the dataset rows. Refs are why boxing exists, so that call is the ticket owner's.

So it sits in the two Gemini wrappers, right before the SDK call, the same shape as `_add_stream_options` in the OpenAI integration. It runs inside the op, after the inputs are recorded, so traces keep their refs. It recurses: on the version we pin, a list of boxed strings breaks the same way. I tried `vals.unwrap` on top and dropped it: `WeaveList` already subclasses `list`, and on a `WeaveTable` it would fetch the whole table. `AGENTS.md` gets the same note.

## Why it is safe

One line is outside Gemini, in `weave/trace/box.py`. `unbox()` had no callers in the repo, and its datetime branch dropped the tzinfo that `box()` sets. This PR is its first caller, so that bug becomes ours: a boxed datetime in a free-form payload goes out shifted by the local offset.

The generator branch matters too. `weave.op` dispatches on `inspect.isgeneratorfunction`, which `functools.wraps` does not carry over, so a plain wrapper would take `generate_content_stream` off the streaming path. One case stays broken: a `GenerateContentConfig` built by hand is already empty before we see it. A dict config is fine.

## Testing

Nine new tests. Reverting the fix fails seven of eight, and I read the outgoing body off the SDK transport for all eight patched symbols.

Bottom line. Gemini used to get `{"contents": [{}]}`. Now it gets the prompt.

[WB-33439]: https://coreweave.atlassian.net/browse/WB-33439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AGENTS.md
```diff
@@ -602,6 +602,27 @@ deterministic.
   ```
 - Some integrations (like instructor) may need to patch multiple libraries
 
+### Wrapping `fn` to change what an SDK receives
```
ro31337, line 625: I have mixed feelings about this part

weave/integrations/google_genai/gemini_utils.py
```diff
@@ -206,6 +208,52 @@ def google_genai_gemini_accumulator(
...
+        return {k: _unbox_traced_values(v) for k, v in obj.items()}
+    elif isinstance(obj, list):
+        return [_unbox_traced_values(v) for v in obj]
+    elif isinstance(obj, tuple):
+        return tuple(_unbox_traced_values(v) for v in obj)
+
+    return unbox(obj)
+
+
+def _unbox_inputs(fn: Callable) -> Callable:
```
gtarpenning, line 229: surprised we dont have a helper/utils for this somewhere

weave/trace/box.py
```diff
@@ -107,7 +107,7 @@ def unbox(
     elif type(obj) == BoxedStr:
         return str(obj)
     elif type(obj) == BoxedDatetime:
-        return datetime.datetime.fromtimestamp(obj.timestamp())
+        return datetime.datetime.fromtimestamp(obj.timestamp(), tz=obj.tzinfo)
```
gtarpenning, line 110: what else does this affect?

tests/integrations/google_genai/test_google_genai.py
andrewtruong, line 1: these tests seem oddly coupled.  I think they make sense as regression tests for this specific case, but don't you really want to have tests to ensure unboxed values are passed in all other funcs too?

andrewtruong (approved):

PR #7777, opened 2026-08-20
chore(weave): bring the Stainless binding up to the vendored SDK

ro31337:
## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-25925

## Description

`StainlessRemoteHTTPTraceServer` did not import. It asked for a top-level `weave_server_sdk`, but the copy #7771 vendored is `weave.vendor.weave_server_sdk`. So nothing in that module had ever run, and no test had ever reached one of its methods.

I thought the import path was the whole job. It was the smallest part.

Once it ran, the constructor raised `TypeError` on an argument the generated client does not take, 34 call sites used a nested resource style the config does not declare, and the tag and alias methods passed a `wb_user_id` that those routes read from the auth header instead.

Then the real bugs showed up. The five object deletes never forwarded `req.digests`, and those routes delete every version of an object when that filter is missing, so asking to delete one version asked to delete all of them. Three creates silently dropped fields, and `call_start_batch` read its response from the wrong key.

Nothing changes for users: `use_stainless_server` still defaults to `False`, so the hand-written client is the default. Eleven methods raise `NotImplementedError`, down from thirteen.

Still broken: filters crash inside the generated SDK, `file_create` cannot upload, list query params comma-join instead of repeating, and the stream routes cannot read their response. wandb/core#51436 fixes the first three. The last one needs a re-vendor.

## Testing

The `stainless` shard is switched back on and passes on 3.10, 3.11 and 3.13. The new `tests/trace_server_bindings/test_stainless_routes.py` checks the request each fixed method actually sends, over an `httpx.MockTransport`.


tests/trace_server_bindings/test_stainless_routes.py
```diff
@@ -0,0 +1,719 @@
...
+    StainlessRemoteHTTPTraceServer,
+)
+from weave.vendor.weave_server_sdk import Client as StainlessClient
+
+BASE_URL = "http://example.com"
+PROJECT = "entity/project"
+V2 = "/v2/entity/project"
+
+# One body serves the whole route table: it carries the union of the v2 response
+# models' required fields, and those models ignore the fields they do not declare.
```
andrewtruong: Let's try to avoid generic claude comments if we can :)

tests/trace_server_bindings/test_stainless_routes.py
```diff
@@ -0,0 +1,719 @@
...
+    "success": True,
+    "trials": 1,
+    "value": None,
+    "version_index": 0,
+}
+
+
+def _mock_server(
+    response: httpx.Response,
+) -> tuple[StainlessRemoteHTTPTraceServer, list[httpx.Request]]:
```
andrewtruong: prefer returning dataclass instead of tuple for better readability/types

andrewtruong (approved):

PR #7778, opened 2026-08-20
chore(weave): return the turn root's facts and the trace's token totals from the chat view

gtarpenning:
## Summary
- A consumer that needs one turn's identity, clock, or usage now gets it from the chat view it already calls, instead of a second query over the same spans. The insights worker in wandb/core is the caller driving this.
- `wb_user_id` is the only field that widens the read: `started_at` and `ended_at` were already selected and `agent_name` was already returned. All three root fields come from `_select_root_span`, so the single-trace and conversation paths get them from the one `build_trace_chat` call.
- Token totals are a plain per-span sum, the same definition `total_input_tokens` carries on a grouped spans query, so one turn means one number wherever it is read. Not the subtree walk `_sum_descendant_tokens` does: that attributes a subtree's usage to the one message standing in for it, and summing its results would double count an enclosed llm span.

## Testing
unit test. The token assertion pins the root span's own usage as counted, which is what separates this sum from the per-message attribution walk, and the facts assertion uses a child with a later start, earlier end, and different identity so root-sourcing is what passes it.

### DIFF
--- a/tests/trace_server/test_agent_scorer_events.py
+++ b/tests/trace_server/test_agent_scorer_events.py
@@ -27,8 +27,6 @@ def test_from_row() -> None:
         ended_at=_ENDED_AT,
         conversation_id="c",
         operation_name="invoke_agent",
-        wb_user_id="u",
-        agent_name="coding-agent",
     )
     event = ScoreAgentSpansEvent.from_row(root_row, "acme")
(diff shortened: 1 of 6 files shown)


bcsherma (approved):
I am down

PR #7783, opened 2026-08-23
chore(weave): refresh the vendored SDK and finish the stream binding

ro31337:
## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-25925

## Description

The Stainless binding talks to a generated client. That client is a copy in the repo, because it is not on PyPI. #7777 made the binding import and call the copy. The copy was old.

Here is what was still broken. Six of the nine list methods could not read their stream at all. File upload could not send bytes. `scorer_list` and `evaluation_list` could not pass `limit` or `offset`. wandb/core#51436 fixes those in the generator. This PR brings the new copy in and points the binding at it.

I thought a re-vendor would be the whole job. It is not. After the new copy, each stream line is a generated model, not a dict. `model_validate` wants a dict, so the nine loops need `item.model_dump()` first. You cannot put that dump on the old copy. Three methods there still yield dicts, and a dict has no `model_dump`. The new tree and the nine lines have to land together.

The same refresh types the file upload as a real file, so `file_create` can send `(name, bytes)`. It also puts `limit` and `offset` on the two list methods that were missing them. Two helpers in the binding had no callers. I deleted those.

This sits on #7777. Merge it after wandb/core#51436. That is the core commit this copy was built from.

Bottom line. The copy can stream, upload, and paginate. The binding uses those three. Nothing is turned on: `use_stainless_server` is still `False`, the other ten stubs still raise, and `calls_complete` is still missing.

## File map

This PR changes 64 files. 62 of them are the vendored copy. The other 2 are ours.

```
weave/vendor/
├── weave_server_sdk/                 [GENERATED] 244 files, verbatim from stlc
│   ├── __init__.py                   [GENERATED] 1 line rewritten
│   ├── _utils/_resources_proxy.py    [GENERATED] 1 line rewritten
│   ├── resources/                    [GENERATED] 30 files
│   ├── types/                        [GENERATED] 183 files
│   └── _base_client.py, _client.py   [GENERATED] and the rest
├── weave_server_sdk.origin.json      written by the script
├── README.md                         do not hand-edit the copy
└── __init__.py                       empty

weave/trace_server_bindings/stainless_remote_http_trace_server.py    the nine dumps, file_create, pagination, two dead helpers
tests/trace_server_bindings/test_stainless_routes.py                 stream table, file_create, limit+offset
```

## Testing

The stainless shard is `tests/trace_server_bindings/` over a mock HTTP client, and it passed on 3.10, 3.11 and 3.13. The stream table now covers all nine methods, `file_create` checks the filename on the multipart part, and the two list methods assert both `limit` and `offset`.


andrewtruong (commented):
How will we keep this up to date with core?

andrewtruong (approved):

PR #7785, opened 2026-08-24
feat(weave_ts): vendor the Stainless client and switch Node onto it

ro31337:
## JIRA Issue(s)

https://coreweave.atlassian.net/browse/WB-30089

## Description

The Node SDK calls the Weave trace server through a swagger client we generate and check in. That client is a 9k-line file from `swagger-typescript-api`. Python already left that stack. This PR copies the Stainless TypeScript client into the Node SDK and points every live call at it. There is no flag: the npm package is what people install, and two clients would mean keeping the swagger file.

Public method signatures stay the same. Stainless returns the payload and throws `APIError`; the old client returned `{data, error}` and threw that object. Nine public methods still do that, via `withResponse()`, and the six exported types stay in swagger form. Uploads pass a named Blob and `expected_digest` rather than `toFile`, because the package supports Node 18. `username` is always `'api'`. Stainless retries are off because we already retry in `fetch`, and the timeout is five minutes instead of 60 seconds. `v2Calls.complete` takes the project first, then `{entity, batch}`.

### Second iteration

Master moved after this PR opened. We merged it. That brought in the Python Stainless finish (#7777, #7783) and the rest of the week's weave master. The Node SDK files in this PR did not change.

The vendored TypeScript tree was not regenerated. OpenAPI and Stainless config hashes on core master still match the stamp from after wandb/core#51436. Re-ran typecheck (cjs and esm), prettier, lint, and jest locally: 540 passed, including hostApps.

## Testing

In `sdks/node`: typecheck for cjs and esm, prettier, lint, and jest, 540 passed, including hostApps against the Python mock server.


rgao-coreweave (approved):
Looks good 👍👍👍

PR #7786, opened 2026-08-24
fix(weave): return stored external refs on read instead of 500

gtarpenning:
## Summary
- `POST /calls/stats` 500s with `InvalidInternalRef: Encountered unexpected ref format.` whenever a project's `summary.usage` holds a key that looks like a `weave:///` ref: `model` comes from `JSONExtractKeysAndValuesRaw(summary.usage)`, so a user op returning `{"model": "<a weave ref>", "usage": {...}}` puts one in the response and takes the whole stats panel down. 11 requests over the last 14 days, one project.
- `universal_int_to_ext_ref_converter` now logs and returns a ref already stored in external form on every read, deduped to one warning per distinct ref so a value repeated across buckets can't flood the logs. The stored string is what the caller wants back, and only the write path can stop it from being persisted.
- `tolerate_external_refs` and the `embedded` flag threaded through `_make_ref_string_mapper` are gone with it, along with the nine agent-endpoint opt-ins in `ExternalTraceServer`. A malformed *internal* ref still raises; nothing can externalize it.

## Testing
unit test plus the existing ClickHouse read-back guard, extended to cover a bare external ref alongside one nested in a JSON leaf. The `/calls/stats` shape itself is not exercised end to end.

### DIFF
--- a/tests/trace_server/test_trace_server_converter.py
+++ b/tests/trace_server/test_trace_server_converter.py
@@ -235,11 +235,11 @@ def test_replace_external_weave_ref_rejects_malformed_tail():
         replace_external_weave_ref("weave:///just-entity", lambda p: p)
 
 
-def test_universal_int_to_ext_ref_converter_tolerate_external_refs(caplog):
-    """Egress: internal refs externalize and unresolvable projects fall back to
-    a private ref. A stored external ref raises by default (surfacing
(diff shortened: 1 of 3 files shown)


gtarpenning:
we should do this but whatever

PR #7787, opened 2026-08-24
fix(weave): let ClickHouse mint query_id

nikumar1206:
## Description

the bump from ch 0.x to 1.6 fixed a very longstanding [bug](https://github.com/ClickHouse/clickhouse-connect/pull/732) which now enables the ch-connect driver to auto-retry failed connections to clickhouse.

this surfaced this new [issue](https://us5.datadoghq.com/apm/trace/9c388f8c37dabfd5b6052762315092e1?spanID=6092360037035044935&timeHint=1787467474661).

this is due to the fact that ch-connect now auto-retries requests that failed, keeping the same query_id when calling CH. was not an issue before because the retry mechanism was just broken.

so we need to stop client-side generation of query_ids. this is in 2 steps:

1. [i introduced setting query_id explicitly](https://github.com/wandb/weave/pull/7757). this was so even on exceptions, we could log/trace the query_id (oops)
2. the ch-connect client also autogenerates the query_id by default. u need to set `autogenerate_query_id=False` to disable this.

however, if we do these changes we lose the ability to see the query_id on exceptions. the way to fix that is by passing a separate id (this pr calls it the `correlation_id`) that is safe even if it collides and pass it into `log_comment`.

so this pr does:
1. undo passing query_id explicitly in settings.
2. set `autogenerate_query_id=False`. similar to [griffin's pr here](https://github.com/wandb/weave/pull/6655).
3. use `correlation_id` instead of `query_id` when dealing with exceptions, but log/trace both ids.

so we can now do the following if we see a query blew up, and we dont have the query_id:
```sql
SELECT
    type,
    query_id,
    log_comment                       AS correlation_id,
    query_duration_ms,
    read_rows,
    formatReadableSize(read_bytes)    AS read,
    formatReadableSize(memory_usage)  AS memory,
    exception_code,
    exception
FROM system.query_log
WHERE log_comment = '01a034f8-2dfd-7910-a1eb-704986ae6350'
  AND event_date >= today() - 1
ORDER BY event_time_microseconds
```

```
Output:

type        | query_id                             | query_duration_ms | read_rows | read      | memory
QueryStart  | 47833a2a-0c71-4607-8715-0ddf75a56691 |                 0 |         0 | 0.00 B    | 0.00 B
QueryFinish | 47833a2a-0c71-4607-8715-0ddf75a56691 |                 9 |   5000000 | 38.15 MiB | 4.18 MiB
```

## Testing

- [x] Manual testing
- [x] Unit tests
- [x] QA

## QA Validation

Deployed to multi-qa as core `d7abe31676c0` (version `suddenly-champion-gannet-d7abe31`, versionId `d71dc570-cd7c-4616-8a3c-03d52a36b9d7`), pinning this branch at `8516dff2f897`. Rollout verified homogeneous — 25/25 `/version` probes on the new sha before, during and after measurement, with the Minecart lock held across the window.

Both ids present on QA spans, and the version nibbles are the point:

| build | `clickhouse.correlation_id` | `clickhouse.query_id` |
| --- | --- | --- |
| before (`daba808bf5`) | — | `01a0352f-20fd-7dbe-…` (uuid**7**, client-minted) |
| after (`d7abe31676`) | `01a0353e-7fb2-74d7-…` (uuid**7**, ours) | `14293d32-7fe5-4220-…` (uuid**4**, server-minted) |

### DIFF
--- a/weave/trace_server/clickhouse/utilities.py
+++ b/weave/trace_server/clickhouse/utilities.py
@@ -308,10 +308,12 @@ def set_correlation_id(settings: dict[str, Any]) -> str:
 def record_query_id(result: QueryResult | QuerySummary) -> str | None:
     """Tag the span with the server-minted query_id off a driver result.
 
-    Both types read it off the `X-ClickHouse-Query-Id` response header, which is
-    absent on a summary the driver built without a response.
+    Read through `summary`, the accessor both types share and where the driver
+    stores the `X-ClickHouse-Query-Id` response header: `query_id` itself is a
+    property on QueryResult but a plain method on QuerySummary. It is absent on
+    a summary the driver built without a response.
     """
-    query_id = result.query_id or None
+    query_id = result.summary.get("query_id") or None
     if query_id:
         set_current_span_dd_tags({"clickhouse.query_id": query_id})
     return query_id
--- a/weave/trace_server/clickhouse/utilities.py
+++ b/weave/trace_server/clickhouse/utilities.py
@@ -294,13 +294,18 @@ def convert_to_insert_too_large(e: Exception) -> Exception:
 def set_correlation_id(settings: dict[str, Any]) -> str:
     """Stamp our own id on a ClickHouse call via `log_comment` and return it.
 
+    The trace server owns `log_comment`: adopting an inherited value would make
+    the correlation id something that can repeat across queries, which breaks
+    the system.query_log pivot without any sign in the logs.
+
     We never send a `query_id`: the driver reuses the URL across its internal
     retries, and a client-supplied id that the server has already registered
     fails with QUERY_WITH_SAME_ID_IS_ALREADY_RUNNING. ClickHouse mints the
     query_id per attempt; `log_comment` carries ours and is queryable in
     system.query_log without a uniqueness constraint.
     """
-    correlation_id = str(settings.setdefault("log_comment", generate_id()))
+    correlation_id = generate_id()
+    settings["log_comment"] = correlation_id
     set_current_span_dd_tags({"clickhouse.correlation_id": correlation_id})
     return correlation_id
 
--- a/weave/trace_server/clickhouse/utilities.py
+++ b/weave/trace_server/clickhouse/utilities.py
@@ -323,7 +323,7 @@ def log_and_raise_insert_error(
     e: Exception,
     table: str,
     data: Sequence[Sequence[Any]],
-    correlation_id: str | None = None,
+    correlation_id: str,
 ) -> None:
     """Log insert error with data size info and re-raise."""
     data_bytes = sum(num_bytes(row) for row in data)
--- a/weave/trace_server/clickhouse/utilities.py
+++ b/weave/trace_server/clickhouse/utilities.py
@@ -13,6 +13,8 @@ from collections.abc import Sequence
 from typing import Any, TypeVar, cast
 
 import sqlparse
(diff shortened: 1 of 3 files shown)


weave/trace_server/clickhouse/utilities.py
```diff
@@ -288,11 +291,39 @@ def convert_to_insert_too_large(e: Exception) -> Exception:
...
+def set_correlation_id(settings: dict[str, Any]) -> str:
+    """Stamp our own id on a ClickHouse call via `log_comment` and return it.
+
+    We never send a `query_id`: the driver reuses the URL across its internal
+    retries, and a client-supplied id that the server has already registered
+    fails with QUERY_WITH_SAME_ID_IS_ALREADY_RUNNING. ClickHouse mints the
+    query_id per attempt; `log_comment` carries ours and is queryable in
+    system.query_log without a uniqueness constraint.
+    """
+    correlation_id = str(settings.setdefault("log_comment", generate_id()))
```
amwarrier: A few things worth thinking about on `settings.setdefault("log_comment", generate_id())`:

1. `setdefault` means if a caller has already put something in `log_comment`, we'd end up using their string as our correlation id and returning it. Probably want to just overwrite here since we own the key.

2. Retry-wise — ch-connect 1.6 reuses the `settings` dict across attempts. Since we're mutating in place, attempt N+1 would inherit attempt N's stamp. Worth double-checking whether `merge_default_query_settings` returns a fresh dict.

3. `log_comment` is a shared free-form String (1 MB cap), and right now we're claiming all of it for the correlation id. As soon as we want to stamp anything else — tenant_id, retry_counter, deploy_sha — we'd be overloading a single string.

What do you think about a typed envelope? Something like:

```python
# weave/trace_server/clickhouse/log_comment.py
class ClickHouseLogComment(BaseModel):
    model_config = ConfigDict(frozen=True, extra="forbid")
    weave_schema_version: int = 1
    weave_correlation_id: str
    # future: tenant_id, request_id, retry_counter, ...

    def merge(self, **updates) -> "ClickHouseLogComment": ...
```

Plus a `stamp_log_comment(settings, *, weave_correlation_id, **extra)` helper that parses-or-fresh, merges, writes back. Frozen + `extra="forbid"` on read gives us mypy on both sides, and it's easy to add fields later without breaking old envelopes.

Still queryable — `JSONExtractString(log_comment, 'weave_correlation_id')` in `system.query_log` works — and it'd let us do retry-detection queries (`GROUP BY corr HAVING count() > 1`) that we can't write today.

Happy to sketch the full class if useful.

nikumar1206: i think points 1 and 2 are good.
fixed point 1. we now own the log_comment key.

2. is already addressed. merge_default_query_settings always returns a fresh dict.

3. im gonna push back on this. we are simply just using the log_comment as a backup in case query_id cannot be used. i think creating a pydantic base model, serializing into log_comment is overkill. even for future-casing. 
also makes querying harder. the current sql query is just log_comment = `id-here` vs a jsonextractstring.

we can always change this later if we need.

nikumar1206: happy to add now too

amwarrier: sure sounds good

amwarrier (approved):

amwarrier (approved):
All good — thanks for the fast turnaround on 1 and 2. Fair call on 3, we can revisit if/when we actually want to stamp another field.

PR #7788, opened 2026-08-24
security(weave): add PII detection and redaction helpers

rgao-coreweave:
## What

Add the versioned `off` and `pii-v1` policy enum, bounded PII detectors, and a copy-on-write traversal for trace values.

This PR adds dormant library code only. No write path calls these helpers when this PR merges.

## Why

Calls and Agent OTel writes need one shared implementation for recognizing and replacing supported PII before storage. Keeping the primitives separate lets reviewers evaluate the detector and traversal contract without route or storage integration.

## How

- Replace supported email addresses, phone numbers, US Social Security numbers, and Luhn-valid payment card numbers with typed markers.
- Use a cheap character check to run only applicable recognizers for each string.
- Combine credential-key replacement and PII replacement in one copy-on-write traversal. Clean subtrees retain their identity, and dictionary keys are not scanned.
- Preserve complete Weave refs, valid base64 data URLs, and valid standalone base64 payloads. Malformed lookalikes remain eligible for scanning.
- Reject unknown policy values instead of choosing a fallback.

## Testing

58 focused tests passed for detector boundaries, false-positive exclusions, structured and escaped JSON values, copy-on-write behavior, credential-key handling, and preserved encoded content.

## Anything Else

- No runtime dependency, migration, permission change, public schema change, or active redaction behavior.
- Child: https://github.com/wandb/weave/pull/7789.

### DIFF
--- a/weave/trace_server/sensitive_data/detectors.py
+++ b/weave/trace_server/sensitive_data/detectors.py
@@ -1,0 +1,278 @@
+"""Bounded recognizers for the versioned ``pii-v1`` entity set.
+
+The first version covers plausible ASCII email addresses, formatted North
+American phone numbers, formatted international phone numbers, compact
+``+``-prefixed numbers containing 10 to 15 digits, dashed US SSNs, and
+13-to-19-digit card candidates that pass Luhn validation.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Literal, NamedTuple
+
+PIIEntity = Literal["EMAIL_ADDRESS", "PHONE_NUMBER", "US_SSN", "CREDIT_CARD"]
+
+REPLACEMENT_MARKERS: dict[PIIEntity, str] = {
+    "EMAIL_ADDRESS": "<EMAIL_ADDRESS>",
+    "PHONE_NUMBER": "<PHONE_NUMBER>",
+    "US_SSN": "<US_SSN>",
+    "CREDIT_CARD": "<CREDIT_CARD>",
+}
+
+_DETECTOR_PRIORITY: dict[PIIEntity, int] = {
+    "EMAIL_ADDRESS": 0,
+    "US_SSN": 1,
+    "CREDIT_CARD": 2,
+    "PHONE_NUMBER": 3,
+}
+
+_ASCII_DIGIT_RE = re.compile(r"[0-9]")
+_MIN_NUMERIC_CANDIDATE_DIGITS = 9
+# One C-level search for "at least 9 digits in the window".
+_MIN_NUMERIC_DIGITS_RE = re.compile(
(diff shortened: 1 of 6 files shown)


weave/trace_server/sensitive_data/walker.py
```diff
@@ -0,0 +1,208 @@
...
+        ("[", "]"),
+        ('"', '"'),
+    }:
+        return False
+    return "@" in stripped or any("0" <= character <= "9" for character in stripped)
+
+
+def _redact_json_string(value: str) -> str:
+    try:
+        parsed = json.loads(value)
```
ro31337: probably worth handling duplicate JSON keys here, for example `{"x":"ada@example.com","x":"safe"}` can pass through unchanged, so a duplicate key check would help

ro31337 (approved):

weave/trace_server/sensitive_data/detectors.py
```diff
@@ -0,0 +1,278 @@
...
+
+class Detection(NamedTuple):
+    start: int
+    end: int
+    entity: PIIEntity
+
+
+def detect_pii(text: str) -> list[Detection]:
+    """Return non-overlapping PII spans without retaining matched text."""
+    detections: list[Detection] = []
```
ro31337, line 79: a few hundred of MB of repeated emails can allocate hundreds of MB. A match limit would help, do you think it's worth bounding the number of detections here?

Maybe it's something for the follow-up. I tested with 3.5 MB input, and peak allocation was ~125 MB.

For route limits with tens of MB we might get around hundreds of gigs of decompressed OTEL and  it's around 1-3 GB memory footprint for a query.

Authentication kinda helps, so it won't be anonymous DoS. But it's still something, isn't it? Or am I overthinking the problem?

Streaming redaction would be a solution, not sure if it's implementable. Or maybe scan/match budget with an error like "RequestTooLarge". We don't have a real budget now, except the limiting HTTP body.

ro31337 (approved):

weave/trace_server/sensitive_data/walker.py
rgao-coreweave: Good catch! 

`json.loads` keeps only the last duplicate, so the shadowed value could leak through unchanged. 

Fixed that duplicate keys now force re-serialization, which drops shadowed values. Added `test_walker_drops_shadowed_duplicate_json_keys`.

weave/trace_server/sensitive_data/detectors.py
rgao-coreweave: Decompressed OTel is already capped at 100 MB. not a hard blocker for now. we can optimize in the future. thanks for the feedback!

PR #7799, opened 2026-08-26
fix(weave_ts): honor Retry-After to distinguish 429 WAIT vs STOP

mayuriphad:
## Summary
`createFetchWithRetry` (`sdks/node/src/utils/retry.ts`) and its caller in `clientApi.ts` treated every 429 the same way: always retry with exponential backoff. As the issue notes, different 429s call for different handling:

- Transient pressure → wait and retry
- Quota / billing exhaustion → do not retry

Collapsing these means a caller that hits a hard quota limit just keeps retrying into it instead of surfacing the failure, amplifying pressure rather than absorbing it.

## Fix
On a 429 response, parse the `Retry-After` header (seconds or an HTTP date):

- If it exceeds `maxDelay`, treat it as quota/billing exhaustion and return the response immediately instead of retrying (STOP).
- Otherwise, use it as the actual wait instead of the exponential-backoff guess, since the server knows its own rate-limit window (WAIT).
- No `Retry-After` header, or a non-429 status: unchanged behavior (exponential backoff / existing `retryOnStatus`).

This only touches the shared `retry.ts` primitive, so both call sites named in the issue (`clientApi.ts:108` and `retry.ts:15`) get the fix from one place.

Fixes #6543

## Test plan
- [x] Added two tests: a 429 with a long `Retry-After` is not retried (STOP), a 429 with a short `Retry-After` is retried using that delay (WAIT)
- [x] `npx jest retry.test.ts` — 6/6 passed (4 existing + 2 new)
- [x] `npx eslint` / `npx prettier --check` clean on the changed files
- [ ] Full CI (could not run the whole monorepo suite locally)

### DIFF
--- a/sdks/node/src/__tests__/util/retry.test.ts
+++ b/sdks/node/src/__tests__/util/retry.test.ts
@@ -62,4 +62,45 @@ describe('retry', () => {
     expect(mockFetch).toHaveBeenCalledTimes(2);
     expect(response).toEqual(mockSuccess);
   });
+
+  test('429 with a long Retry-After is treated as quota exhaustion and not retried', async () => {
+    const mockRateLimited = {
+      ok: false,
+      status: 429,
+      headers: {get: (name: string) => (name === 'retry-after' ? '600' : null)},
+    } as unknown as Response;
+    const mockFetch = jest.fn(() => Promise.resolve(mockRateLimited));
+    global.fetch = mockFetch;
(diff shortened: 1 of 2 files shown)


mayuriphad:

mayuriphad:
I have recreated this PR here: https://github.com/wandb/weave/pull/7903

PR #7801, opened 2026-08-26
feat(weave): add a native async ClickHouse transport, one client per event loop

nikumar1206:
## Description

`clickhouse-connect` 1.6 ships a native `aiohttp` client. This PR adds `AsyncClickHouseTransport` beside the existing blocking driver and gives `AsyncClickHouseTraceServer` native `_query_async`, `_command_async` and `_insert_async` primitives over it. Nothing in weave calls them yet; the first consumers are the workers in wandb/core.

**Structure.** `clickhouse/transport.py` holds the connection config, the two transports, and the call logic they share:

```
ClickHouseTraceServer      -> SyncClickHouseTransport  -> clickhouse-connect Client
AsyncClickHouseTraceServer -> AsyncClickHouseTransport -> clickhouse-connect AsyncClient
```

Settings, correlation id, logging, error mapping and the insert sanitize-and-retry live in `prepare_*` / `record_*` / `raise_*` functions that both servers call. The sync server's `_query`, `_command`, `_insert` and `_query_stream` bodies now go through the same helpers, so the two primitives differ only in the transport call.

**Design points that are not obvious from the diff:**

- `AsyncClickHouseTransport` keeps **one client per event loop**, keyed weakly by loop. An aiohttp session belongs to the loop that opened it, and this server is reached from more than one loop in a process: FastAPI's, each worker's, and every pytest-asyncio test loop. `close()` drains the calling loop's client only; each loop owner closes before its loop ends.
- `query_stream` uses **`query_row_block_stream`, not `query_rows_stream`**. The driver's async stream context hops to the executor once per yielded item. Per row that measured about 50x slower than the blocking driver at 1 CPU (5000 rows: 232ms vs 5ms). Per block it is at parity or better (50000 rows: 12.5ms vs 19.6ms sync). The transport flattens blocks into rows so callers see the same shape as the sync stream.
- `ensure_database` is a module-level provisioning step, not a transport method, per review.
- The thread-local `ch_client` stays. 32 call sites hand a raw client to helpers; that goes away as those methods move to async in the follow-up stack rather than here.

This is the second prerequisite for moving the trace server's ClickHouse I/O onto one async body per method; #7861 (context-local batch state) is the first. It supersedes the `a*` twin approach in #7803 and #7804, which will be closed.

## Performance

One CPU, matching the `cpu: 1` a trace-server or worker pod requests, against local ClickHouse. Same SQL, only the transport differs. The thread-pool arm is the two shapes production runs, 40 threads (Starlette default in the HTTP service) and 32 (the workers' pool), so this varies load, not pool shape.

Inserts, 50 rows per batch:

| in flight | threadpool40 | native async | change | p95 |
| 16 | 70.0 b/s | 73.1 b/s | +4% | 256 → 234 ms |
| 64 | 152.9 b/s | 263.7 b/s | **+72%** | 421 → 250 ms |
| 256 | 175.3 b/s | 377.4 b/s | **+115%** | 1446 → 670 ms |
| 1024 | 189.1 b/s | 420.3 b/s | **+122%** | 5187 → 2220 ms |

Reads, 40 narrow rows per query:

| in flight | threadpool40 | native async | change | p95 |
| 16 | 1363 q/s | 1736 q/s | +27% | 11.9 → 8.8 ms |
| 64 | 1309 q/s | 2697 q/s | **+106%** | 49.5 → 19.0 ms |
| 256 | 1181 q/s | 2223 q/s | **+88%** | 186 → 101 ms |
| 1024 | 1711 q/s | 2154 q/s | +26% | 565 → 383 ms |

At one request in flight native is equal or a few percent slower: the driver parses responses on the default executor, which is one thread hop the sync client does not pay. Streaming, 50000 rows, one in flight: sync 19.6 ms, native by row 2286 ms, native by block 12.5 ms.

## Testing

`tests/trace_server/test_clickhouse_transport.py` checks parity between the two primitives through recording transports (same prepared insert, same single sanitize-retry, same raised types), then drives the real `AsyncClickHouseTransport` against ClickHouse for `start`, `command`, `insert`, `query`, `query_stream` and `close`, and checks that two loops get two clients while one loop reuses its own and `close` is per loop. `trace_server` and `trace` shards pass locally.


weave/trace_server/clickhouse/transport.py
```diff
@@ -0,0 +1,461 @@
...
+            pool_mgr=self._pool_mgr,
+            autogenerate_session_id=False,
+            autogenerate_query_id=False,
+            **optional_kwargs,
+        )
+        self._ensure_database(client)
+        client.database = self._config.database
+        return client
+
+    def _ensure_database(self, client: CHClient) -> None:
```
amwarrier: I don't think _ensure_database belongs in the transport. A transport's job is moving bytes to and from ClickHouse; whether the database exists is a startup/provisioning concern, same as running migrations. `AsyncClickHouseTransport` has no such concept at all

Proposed shape (either now or as a follow-up):
```python
# transport.py — module-level
def ensure_database(config: ClickHouseConfig, pool_mgr) -> None: ...
async def aensure_database(config: ClickHouseConfig) -> None: ...
```

weave/trace_server/async_clickhouse_trace_server.py
```diff
@@ -113,6 +131,76 @@ async def _acompletion_call(
...
         return _CompletionCall(prep, res, start_time, end_time)
 
+    # -- Native async transport ------------------------------------------------
+
+    async def aclose(self) -> None:
+        """Drain the aiohttp session. A worker calls this on shutdown."""
+        await self._atransport.close()
+
+    @traced(name="async_clickhouse_trace_server._aquery")
+    async def _aquery(
```
gtarpenning, line 141: nit: `aquery` is a bit less readable than `query_async`

tests/trace_server/test_clickhouse_transport.py
```diff
@@ -0,0 +1,163 @@
...
+
+TABLE = "intent_signatures"
+COLUMNS = ["project_id"]
+DATA = [["project-1"]]
+SUMMARY = QuerySummary({"query_id": "qid"})
+# Minted per call, so parity compares everything else.
+PER_CALL_SETTINGS = ("insert_deduplication_token", "log_comment")
+
+
+class RecordingTransport:
```
gtarpenning, line 32: this fixture means `AsyncClickHouseTransport.start / _connect / query / command / insert / close` is never run against a real `AsyncClient` right? 

the test: `test_deferred_span_flow_lands_queryable_spans` builds an actual one and really tests it, maybe consider doing the same. does that make sense? like ideally we can test ALL the commands, even stuff like `CREATE DATABASE` etc (from `aclose`) using the real new methods. 

depending on ch version that might uncover quirks

tests/trace_server/test_clickhouse_transport.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+@pytest.mark.disable_logging_error_check
+@pytest.mark.parametrize("arm", ["sync", "async"])
+def test_both_paths_convert_an_oversized_insert(arm):
+    with pytest.raises(InsertTooLarge):
+        drive_one(arm, fail_times=1, error=ValueError("negative shift count"))
+
+
+def test_async_transport_close_clears_the_session():
```
gtarpenning: 🔪

tests/trace_server/test_clickhouse_transport.py
```diff
@@ -0,0 +1,163 @@
...
+def test_both_paths_propagate_a_non_retryable_error(arm):
+    with pytest.raises(RuntimeError, match="driver down"):
+        drive_one(arm, fail_times=1, error=RuntimeError("driver down"))
+
+
+@pytest.mark.disable_logging_error_check
+@pytest.mark.parametrize("arm", ["sync", "async"])
+def test_both_paths_convert_an_oversized_insert(arm):
+    with pytest.raises(InsertTooLarge):
+        drive_one(arm, fail_times=1, error=ValueError("negative shift count"))
```
gtarpenning: can we combine these

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -7496,53 +7494,12 @@ def export_status(self, req: tsi.ExportStatusReq) -> tsi.ExportStatusRes:
...
-            return
-        with self._init_lock:
-            if self._database_ensured:
-                return
-            client.command(f"CREATE DATABASE IF NOT EXISTS {self._database}")
-            self._database_ensured = True
+        """This thread's blocking ClickHouse client."""
+        return self._transport.client
 
     def _mint_client(self, send_receive_timeout: int | None = None) -> CHClient:
```
gtarpenning: i think we might be able to completely delete this. just use `self._transport.mint` everywhere? how many callsites is this?

weave/trace_server/clickhouse_trace_server_batched.py
```diff
@@ -7651,20 +7608,22 @@ def _query_stream(
...
+        correlation_id = prepared.correlation_id
+        parameters = prepared.parameters
 
         summary = None
         query_id = None
-        parameters = process_parameters(parameters)
-        start = time.monotonic()
+        start = prepared.started
         try:
             with self.ch_client.query_rows_stream(
```
gtarpenning: do we need to deal with this ch_client caller?

weave/trace_server/clickhouse_trace_server_batched.py
gtarpenning: totally not in scope but we could clean this up with helpers/logging module to mkae this look a lot nicer

gtarpenning: er, maybe it is in scope, you did that cleanup below

weave/trace_server/clickhouse/transport.py
```diff
@@ -0,0 +1,461 @@
...
+`ClickHouseConfig` is the connection. `SyncClickHouseTransport` and
+`AsyncClickHouseTransport` are the two ways to run a call against it: the
+blocking driver on a thread-local client, and clickhouse-connect's aiohttp
+`AsyncClient` bound to one event loop.
+
+Everything a call does apart from the driver invocation — settings merge,
+correlation id, logging, error mapping, the insert retry decision — lives in the
+`prepare_*` / `record_*` / `raise_*` functions below, which both transports'
+callers share.
+"""
```
gtarpenning, line 8: llmsloppppp

weave/trace_server/clickhouse/transport.py
```diff
@@ -0,0 +1,461 @@
...
+logger = logging.getLogger(__name__)
+
+CLICKHOUSE_DEFAULT_PORT = 8123
+CLICKHOUSE_SECURE_PORT = 8443
+
+# aiohttp connector caps for the native async client. `connector_limit_per_host`
+# is the real concurrency ceiling: weave-trace talks to one host, so this is the
+# async analogue of the thread pool's width.
+ACH_CONNECTOR_LIMIT = 200
+ACH_CONNECTOR_LIMIT_PER_HOST = 100
```
gtarpenning: id prefer `ASYNC_CH...`

weave/trace_server/clickhouse/transport.py
```diff
@@ -0,0 +1,461 @@
...
+    log_and_raise_insert_error(
+        error, prepared.table, prepared.data, prepared.correlation_id
+    )
+    raise error
+
+
+# --- Transports -------------------------------------------------------------
+
+
+class SyncClickHouseTransport:
```
gtarpenning: this wants to be at the top of the file ish (above the helpers at least)

weave/trace_server/clickhouse/transport.py
```diff
@@ -0,0 +1,461 @@
...
+    """The blocking driver, one client per thread over a shared pool manager.
+
+    Per-thread because the driver refuses concurrent queries on one client.
+    """
+
+    def __init__(self, config: ClickHouseConfig, pool_mgr: Any) -> None:
```
gtarpenning: `Any`? maybe `urllib3.PoolManager`?

weave/trace_server/clickhouse/transport.py
```diff
@@ -0,0 +1,461 @@
...
+    def mint(self, send_receive_timeout: int | None = None) -> CHClient:
+        """Create a new client on the shared pool manager.
+
+        autogenerate_session_id=False: weave-trace uses no session features,
+        and the default collides on overlapping queries with SESSION_IS_LOCKED
+        (code 373). See PR #6655.
+        autogenerate_query_id=False: the default collides on a resent request
+        with QUERY_WITH_SAME_ID_IS_ALREADY_RUNNING (code 216). See PR #7787.
+        `send_receive_timeout` overrides the HTTP read timeout (migration clients
+        need to outlast replicated-DDL propagation); None keeps the library default.
```
gtarpenning: this could be 1 line to a jira or something

gtarpenning (commented):
some thoughts, looks good overall, ping me again when you are ready

nikumar1206:
Reworked in place after the design discussion that superseded #7803/#7804. Squashed the earlier 11 commits into one and rebased onto master, then added a second commit: one aiohttp client per event loop (weakly keyed), `query_stream` over `query_row_block_stream` (the per-row async stream measured ~50x slower at 1 CPU), and per-loop `close()`. Body updated with the 1-CPU numbers at 16 through 1024 in flight. The thread-local `ch_client` question from review: it stays for now, its 32 call sites go away as those methods move to async in the follow-up stack rather than here.

PR #7803, opened 2026-08-26
feat(weave): add native async agent span, search and feedback reads [2/4]

nikumar1206:
## Description

Adds async versions of the agent span, search and feedback reads, so the agent scoring worker can await a socket instead of a thread.

- `aagent_spans_query`, `aagent_search` and `afeedback_create` sit next to their sync twins. Same SQL builders, same row transforms — only the round trip is different.
- `aagent_spans_query` only handles ungrouped queries, which is all a worker asks for. Grouped queries need extra reads to fill in distributions and conversation previews, so they stay on `agent_spans_query`. Passing `group_by` raises instead of quietly giving you half an answer.
- `afeedback_create` still forces a sync insert. Someone is waiting to see that write land.
- Pulled `ungrouped_spans_res`, `search_res_from_rows` and `_prepare_feedback_create` out so both paths use them instead of growing their own copies. Sync behaviour is unchanged.

Stacked on wandb/weave#7801, which adds the transport. Nothing calls these yet — the consumer is wandb/core#52039.

## Testing

`tests/trace_server/test_async_agent_queries.py` checks the async path runs the same two statements as `_run_paginated` with one shared `ParamBuilder`, rejects `group_by`, and keeps `do_sync_insert=True` on feedback.

402 existing agent/feedback/trace-server tests pass. 7 `test_feedback_stats` failures also fail on master and aren't related to this.

- [x] Manual testing
- [x] Unit tests

### DIFF
--- a/tests/trace_server/test_async_agent_queries.py
+++ b/tests/trace_server/test_async_agent_queries.py
@@ -7,17 +7,17 @@ or what comes back -- not that the transport is faster.
 """
 
 from types import SimpleNamespace
-from unittest.mock import AsyncMock, MagicMock
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
 from weave.trace_server.agents.clickhouse import (
     AgentQueryHandler,
     search_res_from_rows,
-    ungrouped_spans_res,
 )
 from weave.trace_server.agents.types import (
     AgentGroupByRef,
+    AgentSearchReq,
     AgentSpansQueryReq,
 )
 from weave.trace_server.async_clickhouse_trace_server import AsyncClickHouseTraceServer
@@ -34,13 +34,13 @@ def result(rows, columns):
 
 
 @pytest.mark.asyncio
-async def test_arun_paginated_issues_the_same_two_statements_as_the_sync_pair():
-    """The count and list SQL must match what `_run_paginated` would run."""
+async def test_arun_paginated_issues_the_sync_pair_with_one_param_builder():
+    """Count then page, built by the same builders and bound to one ParamBuilder."""
     req = AgentSpansQueryReq(project_id=PROJECT, limit=5)
-    seen: list[str] = []
+    seen: list[tuple[str, dict]] = []
 
     async def aquery(sql, params):
-        seen.append(sql)
+        seen.append((sql, params))
         return result([[0]], ["c"]) if len(seen) == 1 else result([], [])
 
     handler = AgentQueryHandler(MagicMock(), MagicMock())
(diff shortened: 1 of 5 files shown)


amwarrier (approved):

amwarrier:
Nit on the PR body: says "the consumer is wandb/core#51996" but the actual consumer PR is wandb/core#52039. Small thing, but a wrong number makes the stack harder to trace for future readers.

amwarrier:
The runtime raise here makes grouped-vs-ungrouped a runtime distinction where it should be a type-level one — two requests of the same type, one silently works, one blows up at call time. Two ways I'd rather see it:

1. **Best:** split the request type. `AgentSpansUngroupedReq` (no `group_by` field at all) and `AgentSpansGroupedReq` (requires it). `aagent_spans_query` takes the ungrouped type; callers get compile-time safety.
2. **Cheaper for this PR:** keep the current shape, but use a named exception. `ValueError` is the "I didn't know what to raise" default and gives callers nothing useful to catch on.

Not blocking on option 1 — that's a follow-up. But option 2 is a small change and worth doing before this lands so we don't ossify `ValueError` into the async contract. Concrete proposal:

**`weave/trace_server/errors.py`** — add:

```python
class GroupedQueryNotSupportedAsync(InvalidRequest):
    """Raised when the async agent-spans endpoint receives a grouped request.

    Grouped queries require per-group hydration reads (distribution buckets,
    conversation previews) that only the sync `agent_spans_query` path knows
    how to do. The async twin refuses grouped requests rather than return a
    partial result. Callers with `group_by` should use `agent_spans_query`.
    """

    def __init__(self) -> None:
        super().__init__(
            "aagent_spans_query does not support group_by; "
            "use agent_spans_query for grouped requests."
        )
```

**`weave/trace_server/async_clickhouse_trace_server.py`** — swap the raise inside `aagent_spans_query`:

```python
if req.group_by:
    raise GroupedQueryNotSupportedAsync()
```

**Existing test** — swap `pytest.raises(ValueError, match=...)` for `pytest.raises(GroupedQueryNotSupportedAsync)`. Stronger check, and the type is the invariant we care about.

Bonus: because it extends `InvalidRequest`, any higher-layer error handler that already catches `InvalidRequest` (FastAPI, etc.) handles this for free.

nikumar1206:
Closing in favour of the one-body-per-method approach: the transport lands via #7801 (reworked in place) and #7861 moves batch state to a contextvar. The `a*` twins and the "not supported async" guards here are superseded; `agent_spans_query` and `calls_query` themselves will become async, grouped and hydrated paths included.

PR #7804, opened 2026-08-26
feat(weave): add native async calls and calls-stats reads [3/4]

nikumar1206:
## Description

Adds async versions of the calls and calls-stats reads, so the scoring worker can await a socket instead of a thread.

- `acalls_query` and `acalls_query_stats` sit next to their sync twins and run the same SQL.
- `acalls_query` returns the unhydrated shape a worker asks for. `calls_query` collects its whole stream anyway, so reading it in one round trip loses nothing. Ref expansion and feedback hydration need extra reads per batch, so those stay on `calls_query_stream` — ask for them here and it raises rather than handing back rows that are missing them.
- `resolve_read_table` still hops to a thread. It's cached and only hits ClickHouse on a miss.
- Pulled `_build_calls_query` and `calls_stats_res` out so both paths build the same query and share the same transform. Sync behaviour is unchanged.

Stacked on wandb/weave#7803. The consumer is wandb/core#52045.

## Testing

565 calls/agent/trace-server tests pass, including the 190 covering `calls_query_stream`, so the extraction didn't change it. `tests/trace_server/test_async_calls_queries.py` covers the two things that would break quietly: cost queries add sort-only columns to the SELECT, and `ch_client` has to be resolved on the pool thread rather than the event loop.

7 `test_feedback_stats` failures also fail on master and aren't related to this.

- [x] Manual testing
- [x] Unit tests

### DIFF
--- a/weave/trace_server/async_clickhouse_trace_server.py
+++ b/weave/trace_server/async_clickhouse_trace_server.py
@@ -57,6 +57,7 @@ from weave.trace_server.query_builder.agent_query_builder import (
     make_spans_count_query,
     make_spans_list_query,
 )
+from weave.trace_server.token_costs import get_cost_result_columns
 from weave.trace_server.tracing import traced
 
 _T = TypeVar("_T")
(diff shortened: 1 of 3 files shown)


weave/trace_server/async_clickhouse_trace_server.py
```diff
@@ -260,6 +269,58 @@ async def afeedback_create(
...
+
+    @traced(name="async_clickhouse_trace_server.acalls_query")
+    async def acalls_query(self, req: tsi.CallsQueryReq) -> tsi.CallsQueryRes:
+        """Native-async twin of `calls_query`, for the unhydrated shape.
+
+        Ref expansion and feedback hydration issue further reads per batch and
+        stay on `calls_query_stream`.
+        """
+        if req.expand_columns or req.include_feedback:
+            raise ValueError(
```
amwarrier: lets use the custom exception mentioned in #7803 here.

amwarrier (approved):

nikumar1206:
Closing in favour of the one-body-per-method approach: the transport lands via #7801 (reworked in place) and #7861 moves batch state to a contextvar. The `a*` twins and the "not supported async" guards here are superseded; `agent_spans_query` and `calls_query` themselves will become async, grouped and hydrated paths included.

PR #7805, opened 2026-08-27
fix(weave): mark agent response defaults required in the serialization schema

ro31337:
## JIRA Issue(s)

[WB-39210](https://coreweave.atlassian.net/browse/WB-39210)

Related: [WB-39109](https://coreweave.atlassian.net/browse/WB-39109).

## Description

https://github.com/wandb/core/pull/51853 (draft) fills in missing fields on the client after each agent call. That is the coalescing we said we should not ship.

The generated client treats those fields as optional because OpenAPI does. The server already sends them. This PR marks the defaults on agent responses as required. Requests and constructors do not change.

Once this merges, the usual spec and client sync will pick it up. Then the draft can return the SDK response as-is, without coalescing.

## Testing

`pytest tests/trace_server/test_agent_response_serialization_required.py tests/trace_server/test_genai_schema.py -q --trace-server=fake` — 7 passed. The new tests check that agent responses require every serialized field, and that request schemas did not get stricter.

[WB-39210]: https://coreweave.atlassian.net/browse/WB-39210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-39109]: https://coreweave.atlassian.net/browse/WB-39109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/trace_server/test_agent_response_serialization_required.py
+++ b/tests/trace_server/test_agent_response_serialization_required.py
@@ -1,0 +1,74 @@
+"""Serialization JSON Schema for agent *Res models marks defaults required."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from weave.trace_server.agents.types import (
+    AgentConversationChatRes,
+    AgentConversationSpansRes,
+    AgentCustomAttrsSchemaRes,
+    AgentSearchRes,
+    AgentSpanSchema,
+    AgentSpansQueryReq,
+    AgentSpansQueryRes,
+    AgentSpanStatsRes,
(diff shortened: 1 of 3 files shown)


rgao-coreweave (approved):
Great PR, thanks for fixing!

